PR #109, opened 2021-03-07
Updated Relationship Manager to include a check for Morph Many and stop the attach/detach panels showing if so.

jamessessford:
Hey folks,

The attach/detach methods aren't available on a MorphMany relationship so I created a similar check for this relationship and stopped the buttons showing if the relationship was present.

There's a good chance I've misunderstood what you were going for in the other check so throw this in the sea if it's way off.

Love what you've all done with this - had a great weekend playing with it!

### DIFF
--- a/resources/views/relation-manager.blade.php
+++ b/resources/views/relation-manager.blade.php
@@ -9,7 +9,7 @@
                 {{ __(static::$createButtonLabel) }}
             </x-filament::button>
 
-            @unless ($this->isHasMany())
+            @unless ($this->isHasMany() || $this->isMorphMany())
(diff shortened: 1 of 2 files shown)


danharrin:
Hey James! Thanks for the contribution.

Maybe it's actually best to refactor `isHasMany()` to `canAttach()` and `canDetach()`, then call both type checks within those methods? When we soon implement granular authorisation checks, that will provide a better foundation.

jamessessford:
No problem Dan,

Thanks to all of you for making something fantastic to work with! I've made the requested change.

Cheers

danharrin:
Thank you James! We're glad you love it.

PR #113, opened 2021-03-07
Adding the ability to overrride the default layout

ngtranthanhtoan:
Hi team, 

Sometimes I don't want to use the header or the footer, So I think it is a good idea to add the ability to override the default layout.


danharrin:
Thanks! The static variable is now called `$layout` btw.

PR #116, opened 2021-03-07
feature: add new Icon and Boolean column

ryangjchandler:
This pull request adds support for a new `Icon` and `Boolean` column type.

The `Icon` column is simply a wrapper around `x-dynamic-component`.

You can use it like this:

```php
Icon::make('editable')
    ->options([
        'heroicon-o-check-circle' => fn ($value) => $value === 1,
    ])
```

The `options` array shoud follow the `icon => callback` pattern, such that the callback returns a `bool`. The return value of the callback will determine whether or not the icon should be used.

You could also just specify the name of a custom component as the key, it doesn't _have_ to be an icon component.

The `Boolean` column is a wrapper around `Icon` with some sensible defaults.

By default it will use the `heroicon-s-check-circle` for the `true` icon and `heroicon-o-x-circle` icon for `false`. These icons can be customised using the `trueIcon` and `falseIcon` methods on the column.

You can also hook up an action to both columns using `->action()`. This is handy if you want to perhaps toggle the value or update another piece of state when something is changed.


ryanscherler:
Love this! I JUST tried an icon - using this:

```
Columns\Text::make('slug')
                    ->label('Visit')
                    ->formatUsing(fn ($slug) => svg('heroicon-o-globe', 'w-4 h-4'))
```

But an icon method is even easier!

ryangjchandler:
> Love this! I JUST tried an icon - using this:
> 
> ```
> Columns\Text::make('slug')
>                     ->label('Visit')
>                     ->formatUsing(fn ($slug) => svg('heroicon-o-globe', 'w-4 h-4'))
> ```
> 
> But an icon method is even easier!

Yeah, the only thing that it doesn't support at the moment is custom classes on the icon. It defaults to `w-5 h-5`...

danharrin:
Custom classes could be dangerous as they won't be included in the CSS build...

roni-estein:
looks great! I would definitely remove h-6 w-6 in the component. If you feel you need something there add a default to the classes variable or use something much smaller like h-3 w-3 or even 2 but again last resort this stems from the structure and order of tailwind if you have h-3 and h-6 on an element it will defer to h-6 based on the position in the tailwindcss compiled or downloaded file. So you can always for sizes higher but not constrain them lower.

ryangjchandler:
@roni-estein I think the default classes are actually a good size, but I will likely added a classes() method too.

danharrin:
Again, custom classes won't be in the CSS build unfortunately. We should create a `$size` abstraction and give options like `small`, `base` and `large`.

Ref: https://github.com/laravel-filament/filament/blob/main/resources/views/components/button.blade.php#L19-L22

ryangjchandler:
> Again, custom classes won't be in the CSS build unfortunately. We should create a `$size` abstraction and give options like `small`, `base` and `large`.
> 
> Ref: https://github.com/laravel-filament/filament/blob/main/resources/views/components/button.blade.php#L19-L22

I will message you RE classes. I've got a few thoughts.

roni-estein:
> Again, custom classes won't be in the CSS build unfortunately. We should create a `$size` abstraction and give options like `small`, `base` and `large`.
> 
> Ref: https://github.com/laravel-filament/filament/blob/main/resources/views/components/button.blade.php#L19-L22

I love the ugly components, they are starting to look surprisingly familiar ;)

PR #118, opened 2021-03-07
feature: View column

ryangjchandler:
This pull request introduces a new `View` column that can be used to render a custom view inside of a column cell.

It can be used like this:

```php
View::make('Custom Column')
    ->view('path.to.custom.view')
```

By default, the view you specify with have access to the current record in the table (`$record`) as well as the column class itself (`$column`).

You can also pass through additional data using the `data()` method, or pass an array of data as a second argument to `view()`:

```php
View::make('Custom Column')
    ->view('path.to.custom.view')
    ->data([
        'user' => Filament::auth()->user(),
    ])
```

### DIFF
--- a/packages/tables/src/Columns/Column.php
+++ b/packages/tables/src/Columns/Column.php
@@ -30,6 +30,8 @@ class Column
 
     public $view;
 
+    public $viewData = [];
+
     protected $pendingExcludedContextModifications = [];
 
     protected $pendingIncludedContextModifications = [];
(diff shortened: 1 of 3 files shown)


ryangjchandler:
@danharrin Instead of having the separate `view` and `data` methods, could we also accept an array of data in `view()` - what do you think?

danharrin:
> @danharrin Instead of having the separate `view` and `data` methods, could we also accept an array of data in `view()` - what do you think?

Second view param could call data internally?

ryangjchandler:
> > @danharrin Instead of having the separate `view` and `data` methods, could we also accept an array of data in `view()` - what do you think?
> 
> Second view param could call data internally?

Yeah, that's all it would do just a nice shortcut.

danharrin:
Cool, that keeps it consistent with Laravel then.

ryanscherler:
I absolutely love this idea! Nice @ryangjchandler !

PR #119, opened 2021-03-07
Feature: Add external url capability for a text column URL

ryanscherler:
This PR allows for adding external URL's to the ->url() method of a text column.

Example for viewing an external `Page` resource from the table view:

```php
                Columns\Text::make('slug')
                    ->label('Visit')
                    ->url(function ($page) {
                        $path = $page->slug === 'home' ? '/' : $page->slug;
                        return url($path);
                    }, true),
                    
// or

                Columns\Text::make('slug')
                    ->label('Visit')
                    ->externalUrl(function ($page) {
                        $path = $page->slug === 'home' ? '/' : $page->slug;
                        return url($path);
                    }),
```

### DIFF
--- a/packages/tables/src/Columns/Text.php
+++ b/packages/tables/src/Columns/Text.php
@@ -14,6 +14,8 @@ class Text extends Column
 
     public $url;
 
+    public $externalUrl;
+
     public function action($action)
     {
         $this->action = $action;
@@ -105,10 +107,17 @@ class Text extends Column
         return $this;
     }
 
-    public function url($url)
+    public function url($url, $external = false)
     {
         $this->url = $url;
+        $this->externalUrl = $external;
 
+        return $this;
+    }
+
+    public function externalUrl($url)
+    {
+        $this->url($url, true);
         return $this;
     }
 }
--- a/packages/tables/src/Columns/Text.php
+++ b/packages/tables/src/Columns/Text.php
@@ -12,9 +12,9 @@ class Text extends Column
 
     public $formatUsing;
 
-    public $url;
+    public $shouldOpenUrlInNewTab = false;
 
-    public $openInNewTab = false;
+    public $url;
 
     public function action($action)
     {
@@ -94,6 +94,13 @@ class Text extends Column
         return $value;
     }
 
+    public function openUrlInNewTab()
+    {
+        $this->shouldOpenUrlInNewTab = true;
+
+        return $this;
+    }
+
     public function options($options)
     {
         $this->formatUsing = function ($value) use ($options) {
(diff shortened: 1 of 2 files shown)


ryangjchandler:
I wonder if there's something smarter we could do here that would avoid the need for an additional method. Perhaps something like:

```php
$this->isExternalUrl = ! Str::startsWith($url, url('/')
```

ryanscherler:
> I wonder if there's something smarter we could do here that would avoid the need for an additional method. Perhaps something like:
> 
> ```
> $this->isExternalUrl = ! Str::startsWith($url, url('/'));
> ```

I thought about that, but sometimes we may want to link to something completely different than the app? e.g. a headless CMS / netlify URL etc.?

danharrin:
I think that the terminology "external URL" could be problematic too - you may want to open admin records in a new tab but that's not an external URL. Maybe `$shouldOpenInNewTab` is more explicit.

ryanscherler:
> I think that the terminology "external URL" could be problematic too - you may want to open admin records in a new tab but that's not an external URL. Maybe `$shouldOpenInNewTab` is more explicit.

Oh that makes sense. Ok - maybe a separate method then? 

```php

->url()
-> shouldOpenInNewTab();
``` 
?

danharrin:
Separate method, and second argument shortcut?

ryanscherler:
> Separate method, and second argument shortcut?

Second argument shortcut yeah, but also a separate method seems more explicit as to what you're doing. So maybe not?

packages/tables/src/Columns/Text.php
```diff
@@ -111,4 +113,11 @@ public function url($url)
 
         return $this;
     }
+    
+    public function shouldOpenInNewTab()
```
ryangjchandler, line 117: I think the method name should be `openInNewTab` - `shouldOpenInNewTab` sounds like it returns a boolean, instead of instructing the field to open in a new tab.

ryangjchandler (changes requested):

ryangjchandler:
@ryanscherler I think a second argument to `url()` is a good idea.

packages/tables/src/Columns/Text.php
ryanscherler: I agree. And second param in URL is nice too - making the changes now.

ryanscherler:
Ok - changes are in:

```php
->url('', true) // now accepts second argument as short hand to open in new tab
->openInNewTab() // new fluent method for opening in new tab
```

danharrin:
Sorry to be a pain, I think that the method should be the only thing called `openInNewTab()` - the internal variables should all be `shouldOpenInNewTab` as they are a boolean.

ryanscherler:
Makes sense to use `shouldOpenInNewTab` - thanks for updating and merging in @danharrin ! 🍻

PR #148, opened 2021-03-09
feature: basic theming support

ryangjchandler:
This pull request adds support for very basic theming. With the changes here, you will be able to change the three main colours used by Filament.

### Customising colours

The main colours used by Filament can be customised using CSS variables.

The colours available are:
* `primary`
* `success`
* `danger`
* `white`
* `blue`
* `gray`

All of the colours, apart from `white`, can be adjusted on a scale of 100-900 (just like Tailwind), e.g.

```css
:root {
    --f-primary-100: #450fff;
}
```

To customise the colour `white`, use `--f-white: #ffffff` as the CSS variable.

### Quickly generating a theme

If you want to quickly generate a theme with all of the variable names stubbed out, you can use the `make:filament-theme` command:

```bash
php artisan make:filament-theme custom-theme
```

This will generate a new CSS file `resources/css/filament/custom-theme.css` with all of the CSS variables commented out. Uncomment the variables you wish to modify and insert your own colours.

You can then use the instructions below to include / enqueue the CSS file on the Filament pages.

### Enqueuing custom scripts and styles

This pull request also adds support for including your own styles and scripts inside of Filament, this can be done using the 2 methods:

* `Filament::registerScript($name, $script)`
* `Filament::registerStyle($name, $style)`

These scripts will be included after Filament's own scripts and Livewire's scripts but before any stacks. @danharrin Do we want to change this so that they come _after_ stacks instead?

### Passing data to the front-end

If your custom scripts rely on some data from the server, you can pass this along using `Filament::provideToScript($variables)`:

```php
Filament::provideToScript([
    'model' => Model::first()
])
```

### Calling these new methods

You should call these new methods inside of a service provider's `boot` method. This will ensure Filament's core classes as well as Laravel's are registered with the container.

### `Filament::serving`

There is also a new `Filament::serving($callback)` method that can be used to only run your Filament related logic when Filament is actually being served to the user.

You can use it like this:

```php
    Filament::registerScript('script-name, asset('js/script.js'));
});
```

Behind the scenes, this will just register an event listener for the `Filament\Events\ServingFilament` event. This means you could also pass through a Laravel-style callable array too.

```php
Filament::serving([ListenerClass::class, 'method']);
```

The callback function will receive the `ServingFilament` event too, which has a `public $request` variable so you can run some logic based on the current request context.

> You could also register an event using `Event::listen(ServingFilament::class)` directly, but the `Filament::serving` method is recommended since any under-the-hood changes won't result in any code breaks.


resources/views/components/layouts/base.blade.php
```diff
@@ -14,6 +14,11 @@
         'id' => Filament\get_asset_id('/css/filament.css'),
         'path' => 'css/filament.css',
     ]) }}" />
+
+    @foreach (\Filament\Filament::getStyles() as $style)
+        <link rel="stylesheet" href="{{ $style }}">
```
danharrin, line 19:
```suggestion
        <link rel="stylesheet" href="{{ $path }}" />
```

resources/views/components/layouts/base.blade.php
```diff
@@ -14,6 +14,11 @@
         'id' => Filament\get_asset_id('/css/filament.css'),
         'path' => 'css/filament.css',
     ]) }}" />
+
+    @foreach (\Filament\Filament::getStyles() as $style)
```
danharrin, line 18:
```suggestion
    @foreach (\Filament\Filament::getStyles() as $path)
```

resources/views/components/layouts/base.blade.php
```diff
@@ -23,10 +28,19 @@
...
+    <script>
+        window.filamentConfig = @json(\Filament\Filament::getScriptData());
+    </script>
+
     <script src="{{ route('filament.asset', [
         'id' => Filament\get_asset_id('/js/filament.js'),
         'path' => 'js/filament.js',
     ]) }}"></script>
+
+    @foreach (\Filament\Filament::getScripts() as $script)
```
danharrin, line 40:
```suggestion
    @foreach (\Filament\Filament::getScripts() as $path)
```

resources/views/components/layouts/base.blade.php
```diff
@@ -23,10 +28,19 @@
...
+        <script src="{{ $script }}"></script>
```
danharrin, line 41:
```suggestion
        <script src="{{ $path }}"></script>
```

src/FilamentManager.php
```diff
@@ -55,13 +89,37 @@ public function registerRole($role)
         $this->roles = array_merge($this->roles, [$role]);
     }
 
+    public function registerScript($name, $script)
+    {
+        $this->scripts[$name] = $script;
```
danharrin, line 94:
```suggestion
        $this->scripts[$name] = array_merge($this->scripts, [$path]);
```

src/FilamentManager.php
```diff
@@ -55,13 +89,37 @@ public function registerRole($role)
         $this->roles = array_merge($this->roles, [$role]);
     }
 
+    public function registerScript($name, $script)
```
danharrin, line 92:
```suggestion
    public function registerScript($name, $path)
```

src/FilamentManager.php
```diff
@@ -55,13 +89,37 @@ public function registerRole($role)
...
+    {
+        $this->scripts[$name] = $script;
+    }
+
+    public function registerStyle($name, $style)
```
danharrin, line 97:
```suggestion
    public function registerStyle($name, $path)
```

src/FilamentManager.php
```diff
@@ -55,13 +89,37 @@ public function registerRole($role)
...
+    {
+        $this->styles[$name] = $style;
```
danharrin, line 99:
```suggestion
        $this->styles[$name] = array_merge($this->styles, [$path]);
```

src/FilamentManager.php
```diff
@@ -16,6 +19,14 @@ class FilamentManager
 
     public $roles = [];
 
+    public $styles = [];
+
+    public $scripts = [];
+
+    public $scriptData = [];
+
+    public $servingCallbacks = [];
```
danharrin, line 28: Is this used?

src/FilamentManager.php
ryangjchandler: @danharrin This isn't right. The point of the `$name` variable is so that each script / style registered has a unique name and the same `$name` can't be present more than once (allows you to override vendor scripts in an extension package for example).

PR #152, opened 2021-03-10
Added trim functionality to Text Column

jamessessford:
Morning folks,

TLDR: Added trim functionality to Text Column

I found that I was using the getValueUsing callback a lot of the time just to trim the length of the content but if I did this I lost the ability to search or sort on that field.

I've added a trim helper to the Text column that takes a trim length and then trims the value after it's been taken from the record.

Example usage:

```php
Columns\Text::make('reference')->primary()->searchable()->sortable()->trim(6),
```

This allows me to only show the first six characters of the reference in the table but retain search and sort functionality.

Cheers!

### DIFF
--- a/packages/tables/src/Columns/Text.php
+++ b/packages/tables/src/Columns/Text.php
@@ -90,9 +90,7 @@ class Text extends Column
     public function limit($length = -1)
     {
         $this->formatUsing = function ($value) use ($length) {
-            $value = Str::of($value)->limit($length);
-
-            return $value;
+            return Str::limit($value, $length);
         };
 
         return $this;
--- a/packages/tables/src/Columns/Text.php
+++ b/packages/tables/src/Columns/Text.php
@@ -14,10 +14,6 @@ class Text extends Column
 
     public $formatUsing;
 
-    public $trimLength = -1;
-
-    public $trimmable = false;
-
     public function currency($symbol = '$', $decimalSeparator = '.', $thousandsSeparator = ',')
     {
         $this->formatUsing = function ($value) use ($decimalSeparator, $symbol, $thousandsSeparator) {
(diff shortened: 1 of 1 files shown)


danharrin:
Hey James! This sounds like a job for `formatUsing` rather than `getValueUsing`. We should use the `trim()` method to set that callback, similar to `date()`. Let me know how that goes.

ryangjchandler:
I think this method should probably be called `truncate` or `limit`, since PHP's `trim` function is for removing characters at each end, not actually limiting the length of a string.

jamessessford:
Totally agreed @ryangjchandler - Updating now

jamessessford:
I've changed the function to limit rather than trim (that was the str helper it's piggybacking) and updated to use formatUsing rather than getValueUsing.

Cheers!

packages/tables/src/Columns/Text.php
```diff
@@ -85,4 +86,15 @@ public function options($options)
 
         return $this;
     }
+
+    public function limit($length = -1)
+    {
+        $this->formatUsing = function ($value) use ($length) {
+            $value = Str::of($value)->limit($length);
```
ryangjchandler, line 93: No need for the `Str::of()` here - you could just use `Str::limit` directly, right?

ryangjchandler (changes requested):

packages/tables/src/Columns/Text.php
jamessessford: You can indeed! I didn't realise you could do that, buzzing I've learned something new this morning :)

jamessessford: I've updated this to:

```php
return Str::limit($value, $length);
```

ryangjchandler: Nice one @jamessessford!

ryangjchandler (approved):

PR #168, opened 2021-03-10
Fix PostgreSQL search issues

### LINKED ISSUE #75: Postgres search issues
**Describe the bug**
I have a BelongsToSelect form relation in my post model. `Components\BelongsToSelect::make('user_id')->relationship('user', 'username')->required(),` 

I can search any user who have the username all lowercase. But for those Uppercase one this never match, even if I search with the Uppercase letter.

**To reproduce**
Having uppercase username in a table with a BelongsToSelect relation.

**Expected behavior**
I expected to see any user matching the search term.

**Screenshots**
Screen Shot 2021-03-04 at 9 22 41 AM
Screen Shot 2021-03-04 at 9 22 50 AM

**Context**
- Browser:  Chrome
- Server OS: Localhost on MacOS Catalina 10.15.7
- PHP version: 8.0

**Additional details**

As you can see, when I type `e` I get EQuimper there. But the reason is because of the `er` at the end. This never match the big `E`

danharrin:
Fixes #75.


ryangjchandler (approved):

ryangjchandler:
Have tested the changes here with a local PGSQL database and can confirm it is working. Following scenarios tested:

* Searching by full word.
* Searching by partial word.
* Searching using integers/numeric values.

bram-pkg:
It seems this feature never made it into Filament 3 :sweat_smile: 
Any plans to port it to the new version?

danharrin:
@bram-pkg whats the problem?

bram-pkg:
It seems that in Filament 2 the "ilike" operator was used when using PostgreSQL, this does not seem to be the case in Filament 3. Am I missing something?

Is the expected usage now to add the method to each table column to force case insensitivity?

Using the "ilike" when using PostgreSQL, like in Filament 2, was quite nice behaviour to be honest.

danharrin:
The problem is that when using it, indexes are not used. Which kills performance for everyone. Maybe you could use a case-insensitive collation instead. But this original behaviour is still possible using the method to force case insensitivity.

PR #172, opened 2021-03-11
Allow custom middleware to be registered

Larsklopstra:
This allows users to register their own middleware.


ryangjchandler:
I'm wondering whether we should actually put the default middleware stack inside of the config array, instead of doing the merge / spread. Will wait for @danharrin to review too, but in it's current state I can't see any issues.

danharrin:
> I'm wondering whether we should actually put the default middleware stack inside of the config array, instead of doing the merge / spread. Will wait for @danharrin to review too, but in it's current state I can't see any issues.

Yeah let's do that please @Larsklopstra 👍

ryangjchandler:
Closing in favour of #180.

PR #175, opened 2021-03-11
Allow additional recordActions

iantasker:
Currently, if you set any `recordActions` on a table they are overwritten in ListRecords as `recordActions` just replaces the array instead of merging the arrays.


ryangjchandler (approved):

danharrin:
Thanks!

PR #180, opened 2021-03-11
Allow custom middleware to be registered

ryangjchandler:
Copy of #172 with some extra changes. Adds support for customising middleware.


danharrin:
Thanks @Larsklopstra!

PR #185, opened 2021-03-12
feature: KeyValue field type

ryangjchandler:
> Opening this PR now so people can jump in a give suggestions too, as well as design feedback.

This pull request introduces a new field type, `KeyValue`. This is essentially a glorified single-level JSON editor that is useful for simple meta columns that are cast using Laravel's `array` or `json` casts.

To use this field type, you would do the following:

```php
KeyValue::make('meta')
```

This would add the following field to your form:

image

You can add a key in the left hand input and the corresponding value on the right.

If deleting rows is enabled, you'll also see a little "bin" icon on the right hand side. This removes the row.

image

The ability to add and delete rows can be disabled using the `disableAddingRows()` and `disableDeletingRows()` methods. These will completely hide the buttons and prevent people from manually calling the appropriate JavaScript methods from the front-end.

You can also disable editing keys using the `disableEditingKeys()` method.  This will disable any key inputs and only allow the user to update the values. Perfect when you need to ensure certain key names can't be changed.

All of the labels, placeholders and column labels are customisable using their respective methods too. By default, they use translation strings.

Here's what needs to be done still:
1. Get some design feedback from @ryanscherler, especially responsive design and UX.
2. Get any extra ideas or improvements from people who might need this sort of field in their app.

### To-do

* [x] Implement drag & drop / sorting.
* [x] Allow sorting to be enabled / disabled on field.
* [x] Fix issue where data is being duplicated on sort & save.
* [x] Fix issue where input focus is lost after saving.


ryanscherler:
This field is going to be awesome! I think per our feedback on Discord its ready to roll design-wise!

ryangjchandler:
@ryanscherler This is ready for full review now. I've implemented sorting / dragging for table rows, this is disabled by default but can easily be enabled using the `sortable` method on the field.

Here's how it looks:

https://user-images.githubusercontent.com/41837763/110935322-2e60ef80-8327-11eb-9d8f-ad29ccb339ad.mov

ryanscherler:
@ryangjchandler this is soooo awesome! Great work! Only visual feedback would be to remove the vertical borders between columns - keep it more clean / minimal like our other existing table layout.

ryangjchandler:
@ryanscherler - Yeah, happy to do that. It does make the focus styles on the forms look a bit off though.. I'll remove them and see what you think.

yob-yob:
Hi..

I'm getting `Array to string conversion` error when using this component...

I already have setup a `$casts` on my model

```php
protected $casts = [
    'schema' => 'array'
  ];
```

but still I get an error.. so I made a `mutator`...

```php
public function setSchemaAttribute($value)
  {
    $this->attributes['schema'] = json_encode($value);
  }
```

it goes well.. it creates the model... problem is that when editing, the current values for `KeyValue` are not shown...

this is my column for the table
```php
$table->json('schema');
```

Php Version: 8.0.9
Laravel Version: 8.56.0
Filament Version: 1.10.12

yob-yob:
Found the culprit...

you can't use `schema` column name... you need to rename your column...  it's the one that fixed for me... I don't really know why this happens thou... but it fixed both issue..


`Array to string conversion` and Empty Fields while editing...

PR #198, opened 2021-03-12
Support Icons on RecordActions

iantasker:
This PR allows Icons to be added to RecordActions. 

There is a change in the logic of how RecordActions are displayed, previously any RecordActions added to a table would be displayed in addition to the default edit action. 

To allow the default edit action to be replaced it will now only be shown if no RecordActions are supplied.


danharrin:
Hey @iantasker! I've rewritten this in #203 and merged into the develop branch. If you could test it (as well as the rest of the large refactor), that would be amazing. Thank you!

PR #209, opened 2021-03-13
Pluck id from record before passing to destroy

### LINKED ISSUE #186: Resource listing page deletion issue
**Describe the bug**
When selecting one or more items from a resource list page and trying to delete them an exception happens.

**To reproduce**
Steps to reproduce the behaviour:
1. Install Filament version >= 1.50
2. Go to a resource listing page
3. Select 1 or more items to delete
4. Click on the delete button
5. Confirm the deletion from the modal
6. The exception happens

**Expected behaviour**
The selected items in the resource list should be deleted from the database and the change reflected in the page

**Screenshots**

Deleting a single item
Single item delete result

Deleting multiple items - long numbers in query are fake id and sku values from a product record
Multi item select

Multiple item delete result

**Context**
- Package version: [>= 1.5.0]
- Livewire version: [2.3.0]
- Laravel version: [>= 8.12.0]
- Browser: [Chrome latest]
- Server OS: [OSX 10.11.6]
- PHP version: [7.4.14]
- Database: [MySQL 8.0.23]

**Additional details**
Issue seems to be different for single and multiple selections - with multiple selections, the ids passed into the IN clause in the query do not reflect the record ids that should be selected. Instead they look like they contain the other field values from one of the selected records and not a sequence of integers as expected.  

database entry

Delete buttons in the item edit pages still work as expected.

I rolled back my repo until I found the version were this began - version 1.4.8 worked as expected and no other packages etc have been updated, only filament version was increased. Happens in all versions from 1.5.0 upwards.

Also still happens if I then update all other packages to the current latest versions.

I have also cleared and rebuilt the database, rerun all migrations, reinstalled all vendor packages and cleared all caches etc

jamessessford:
Fixes #186 

Full record was being passed to destroy function, not the ID of the record

### DIFF
--- a/src/Resources/Pages/ListRecords.php
+++ b/src/Resources/Pages/ListRecords.php
@@ -68,6 +68,7 @@ class ListRecords extends Page
                 ->filter(function ($record) {
                     return Filament::can('delete', $record);
                 })
+                ->pluck('id')
                 ->toArray(),
         );
 
--- a/src/Resources/Pages/ListRecords.php
+++ b/src/Resources/Pages/ListRecords.php
@@ -68,7 +68,7 @@ class ListRecords extends Page
                 ->filter(function ($record) {
                     return Filament::can('delete', $record);
                 })
-                ->pluck('id')
+                ->map(fn ($record) => $record->getKey())
                 ->toArray(),
         );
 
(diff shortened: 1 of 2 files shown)


danharrin:
Thanks James! I've modified this to use `getKey()` instead of hard-coding `id` to account for other primary key columns.

This has been merged into develop. I encourage you to test it out, we are pending a large refactor and want to encounter as many issues as possible before we release it.

PR #210, opened 2021-03-13
Allow Title Property on RecordAction Links and Icons

iantasker:
If you only display icons, a title property in needed to populate the browser tooltip. 

The title property should also be settable on links for consistency.


danharrin:
Thanks!

PR #212, opened 2021-03-13
Cannot read property 'split' of null issue fixed

shuvroroy:


ryangjchandler:
Using the index is generally a bad idea since it won't be unique to each record. You'd be better off using the record ID / primary key.

danharrin:
Thanks!

PR #214, opened 2021-03-13
chore: php-cs-fixer workflow

### LINKED ISSUE #207: Setup PHP-CS-Fixer GitHub Action
It would be great to get some code formatting setup for the PHP files. There's 2 approaches we could take here:

1. Automatically run the formatting action when somebody pushes to a PR (annoying because you would have to pull down between each push if changes have been committed back).
2. Automatically run the formatting action when a PR is merged into `develop` (can cause merge conflicts if develop's formatting is newer than the branches).
3. Manually trigger the action using a comment on the PR, such as `/format-php` (good because we can still run on an individual branch just before merging in).

Would be interested in your thoughts @danharrin and @ryanscherler.

ryangjchandler:
Closes #207. Went with the "after merge into develop" approach, so this workflow should run when any pull request is closed **and** merged into develop.

The rules are very basic and can be expanded upon in the future, full credit to Spatie for providing a great default in their package skeleton.

### DIFF
--- a/.github/workflows/php-cs-fixer.yml
+++ b/.github/workflows/php-cs-fixer.yml
@@ -7,7 +7,7 @@ on:
     types: [closed]
 
 jobs:
-  phpcs-fixer:
+  php-cs-fixer:
     runs-on: ubuntu-latest
     if: github.event.pull_request.merged == true
     steps:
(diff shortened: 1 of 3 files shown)


danharrin:
Is it worth also scanning, `/config`, `/packages` and `/routes`?

PR #215, opened 2021-03-14
New section form component features

ryanscherler:
This PR introduces some subtle styling, a11y markup changes and expand / collapse functionality for the `Components\Section` component. Some additional work is required to ensure that the section is expanded if a validation error occurs.


ryanscherler:
One additional item is to have a section be open / closed by default. This is likely done via a param and method on the section:

```php
// via additional param
Components\Section::make($heading, $subheading, $schema = [], $open = true);
// or via method
Components\Section::make()->collapsed() 
// this could even have a callback to pass in true / false depending on a condition?
Components\Section::make()->collapsed(true|false) 
```

ryangjchandler:
@ryanscherler I think the method with a callback / boolean argument is best.

packages/forms/resources/views/components/section.blade.php
```diff
@@ -16,20 +16,31 @@
...
+            </p>
+        @endif
         </div>
-    @endif
-
-    <x-forms::layout :schema="$formComponent->getSchema()" :columns="$formComponent->getColumns()" />
-</div>
+        
+        <div class="flex">
+            <button aria-controls="section-content-{{ $formComponent->getId() }}" @click.prevent="open = !open" class="flex p-2 -m-2 text-gray-400 transition-colors duration-200 hover:text-gray-700">
```
ryangjchandler, line 36: Think this should be wrapped in an "if collapsible" check.

ryangjchandler, line 36: There should probably be some extra JS validation for the collapsible logic too, just to be sure nobody could inadvertently collapse a section when they shouldn't be able to.

ryangjchandler (changes requested):

danharrin:
@ me when it's ready for the expand-on-validation to be implemented and I'll take a look.

packages/forms/resources/views/components/section.blade.php
ryanscherler: I think I have the logic in there now to ensure the state sync's up (its in the PHP class on the `collapsed()` method.

ryanscherler: Done! Just pushed an update to allow the following:

```php
// collapsible false by default when creating a section to maintain current functionality and not hide things inadvertently
Components\Section::make($heading, $subheading, $schema = [], $collapsible = false)

Components\Section::make()->collapsible() // fluent method for enabling collapsible functionality (I need some help withe how to create a callback on this method to allow for logic to set true|false

Components\Section::make()
    ->collapsed() // sets the initial state of the section to collapsed, and enables 'collapsible' to ensure it can be toggled back open again
```

ryanscherler:
Example implemented in the demo:

```php
Components\Section::make(
    'Information',
    'Additional order information',
    [
        Components\FileUpload::make('invoice'),
        Components\RichEditor::make('notes')
             ->placeholder('Notes'),
     ]
)->collapsed(),
```

ryanscherler:
@danharrin i think this is ready to go design wise from my end - have at it!

packages/forms/src/Components/Section.php
```diff
@@ -48,12 +62,13 @@ public function heading($heading)
         return $this;
     }
 
-    public static function make($heading, $subheading = null, $schema = [])
+    public static function make($heading, $subheading = null, $schema = [], $collapsible = false)
```
ryangjchandler, line 65: I don't know how I feel about having 4 arguments to a method... generally speaking, if I've got a method that has more than 3 arguments, there always a better way to do it. Everything becomes a bit too dense, especially when the previous argument is an array that's likely going to split the method call across multiple lines.

In this case, I think we can get rid of the `$collapsible` argument and just rely on the `collapsible` method instead. What do you think?

ryangjchandler (changes requested):
Small thought @ryanscherler.

packages/forms/src/Components/Section.php
danharrin: This sounds good to me 👍

ryanscherler: Me too - happy with the collapsible() method

ryangjchandler (approved):

PR #219, opened 2021-03-15
fix: dont merge middleware config

ryangjchandler:
This fixes an issue where the `middleware` config options were being deeply merged, so any custom middleware would still be applied alongside the original middleware, i.e. `Authenticate` from Filament would also be applied alongside any custom middleware in `middleware.auth`.


iantasker:
@ryangjchandler I have tested this and it resolves the issue I had.

PR #221, opened 2021-03-15
feature: generate resourceUrls in columns

ryangjchandler:
This pull request introduces a new convenience method for resource-based table columns called `resourceUrl`.

It can be used like so:

```php
Columns\Text::make('name')
    ->resourceUrl(resource: OrderResource::class, route: 'edit', shouldOpenInNewTab: false)
```

/cc @edgrosvenor

### DIFF
--- a/src/Resources/Tables/Columns/Concerns/InteractsWithResource.php
+++ b/src/Resources/Tables/Columns/Concerns/InteractsWithResource.php
@@ -1,0 +1,15 @@
+<?php
+
+namespace Filament\Resources\Tables\Columns\Concerns;
+
+trait InteractsWithResource
+{
+    public function resourceUrl($resource, $route = 'edit', $shouldOpenInNewTab = false)
+    {
+        $this->url(function ($record) use ($resource, $route) {
+            return $resource::generateUrl($route, ['record' => $record]);
+        }, $shouldOpenInNewTab);
+
+        return $this;
+    }
+}
--- a/src/Resources/Tables/Columns/Boolean.php
+++ b/src/Resources/Tables/Columns/Boolean.php
@@ -4,5 +4,5 @@ namespace Filament\Resources\Tables\Columns;
 
 class Boolean extends \Filament\Tables\Columns\Boolean
 {
-    //
+    use Concerns\InteractsWithResource;
 }
--- a/src/Resources/Tables/Columns/Column.php
+++ b/src/Resources/Tables/Columns/Column.php
@@ -4,5 +4,5 @@ namespace Filament\Resources\Tables\Columns;
 
 class Column extends \Filament\Tables\Columns\Column
 {
-    //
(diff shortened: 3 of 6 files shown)


danharrin:
Sorry I'm hesitant on merging this - something about this feature just seems really specific, but I agree its a common use case.

The same effect can be achieved using...
```
url(fn ($order) => OrderResource::generateUrl('edit', $order->id))
```

ryangjchandler:
> Sorry I'm hesitant on merging this - something about this feature just seems really specific, but I agree its a common use case.
> 
> The same effect can be achieved using...
> 
> ```
> url(fn ($order) => OrderResource::generateUrl('edit', ['record' => $order->id]))
> ```
> 
> Is the saving that large?
> 
> We could instead introduce a feature that removes the need for the parameter array:
> 
> ```
> url(fn ($order) => OrderResource::generateUrl('edit', $order))
> ```

I agree, it does seem specific but it's only going to be available on `Resources\Tables\Columns` and the maintenance burden is basically zero. Removing a few words or maybe lines of code (if you don't like using arrow functions), I'd say it's worth it.

src/Resources/Tables/Columns/Concerns/InteractsWithResource.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Filament\Resources\Tables\Columns\Concerns;
+
+trait InteractsWithResource
+{
+    public function resourceUrl($resource, $route = 'edit', $shouldOpenInNewTab = false)
+    {
+        $this->url(function ($record) use ($resource, $route) {
```
danharrin, line 9: This will need to be wrapped in a `$this->configure()` to ensure that state modifications are made in an orderly way.

danharrin (changes requested):

src/Resources/Tables/Columns/Concerns/InteractsWithResource.php
ryangjchandler: I don't think it does since `$this->url()` is already calling `$this->configure()`, no? It's purely an alias for another method.

danharrin: Technically you're right, but do you think its worth using `$this->configure()` everywhere to ensure that we're following a pattern? There are some places where we can't avoid using `$this->configure()` unnecessarily due as some parts of the method require it and some already call it in another method.

I'm not sure how big of a consistency issue it is.

ryangjchandler: I don't think it's a huge problem in this instance, because `resourceUrl` is just a proxy alias for the `url` method anyway. It would be different if `resourceUrl` was doing something different and not using `url`, because it would _need_ the configure.

PR #222, opened 2021-03-15
feature: add new Hidden field

ryangjchandler:
This pull request adds a new dedicated `Hidden` field type that can be used for static data. It can be used like this:

```php
Hidden::make('name')
    ->value('static data')
```

`Hidden::value()` is just an alias for `Hidden::default()`, so they are interchangeable but the `value` method feels more natural since it's not technically a default value.

### DIFF
--- a/packages/forms/src/Components/Hidden.php
+++ b/packages/forms/src/Components/Hidden.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Filament\Forms\Components;
+
+class Hidden extends Field
+{
+    public function value($value)
+    {
+        $this->default($value);
+
+        return $this;
+    }
+}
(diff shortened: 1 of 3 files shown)


packages/forms/src/Components/Hidden.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Filament\Forms\Components;
+
+class Hidden extends Field
+{
+    public function value($value)
+    {
+        $this->default($value);
```
danharrin, line 9: This will need to be wrapped in `$this->configure()`.

ryangjchandler: I don't think it does, since `$this->default()` already has the `configure()` logic, right?

PR #231, opened 2021-03-16
add default table sorting

eminos:


danharrin:
Thanks for this! I've refactored quite a lot, so these changes have now moved to #281.

PR #234, opened 2021-03-17
Add parameter overrides to column searchable and sortable

Blackpig:
In reference to this discussion https://github.com/laravel-filament/filament/discussions/51

This PR allows the `->searchable()` and `->sortable()` methods of the Column to accept a parameter to override a column's default search or sort behaviour.

Main example use case is where a resource column's  value is accessed via a model's `getXXXXXAttribute` where the underlying field does not exist in the table definition.  Refer to the above linked discussion for an example use case.

### DIFF
--- a/packages/tables/src/Columns/Column.php
+++ b/packages/tables/src/Columns/Column.php
@@ -20,7 +20,11 @@ class Column
 
     protected $isSearchable = false;
 
+    protected $searchUsing = false;
+
     protected $isSortable = false;
+
+    protected $sortUsing = false;
 
     protected $label;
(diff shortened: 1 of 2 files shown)


danharrin:
Hey! Thanks for this. I've just rewritten this as part of #281, with some adjustments...

Originally, sorting on relationship columns was broken. I have fixed this, and you can now able to use multiple sort and search columns for them:

```php
Column::make('customer.name')
    ->searchable(['first_name', 'last_name'])
    ->sortable(['first_name', 'last_name']);
```

PR #238, opened 2021-03-17
Add decimal quantity to currency in Text column

sebapastore:
Some currencies doesn't use two decimal after decimal separator, so it will be useful if it's customizable.


ryangjchandler (approved):

danharrin:
Thanks!

PR #243, opened 2021-03-18
allow absolute URL as avatar

### LINKED ISSUE #242: Avatar component does not accept a full image URL
**Describe the bug**
If `\App\Models\User::getFilamentAvatar()` return s a full URL - for example a Gravatar, Unavatar or similar service URL - the avatar blade component prepends the already absolute URL with the app URL.

**Expected behavior**
It should render the `` tag with the given full URL.

**Screenshots**
Bildschirmfoto 2021-03-18 um 17 50 33

**Fix**

Add another condition to the `\Filament\View\Components\Avatar::src()` method.

```php
if (Str::startsWith($avatar, ['http://', 'https://'])) {
    return $avatar;
}
```

Gummibeer:
fixes #242

### DIFF
--- a/src/View/Components/Avatar.php
+++ b/src/View/Components/Avatar.php
@@ -3,7 +3,6 @@
 namespace Filament\View\Components;
 
 use Filament\Filament;
-use Illuminate\Support\Str;
 use function Filament\get_image_url;
 use Illuminate\View\Component;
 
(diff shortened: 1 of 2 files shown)


danharrin:
Hey, should we try `filter_var($avatar, FILTER_VALIDATE_URL)` here instead?

danharrin:
Also, I feel like this fix should be dealt with in `get_image_url()`?

Gummibeer:
It has to be a combination as `ftp://ftp.example.com/image.jpg` is a valid URL.

danharrin:
Thank you!

PR #245, opened 2021-03-18
use resolveRouteBinding method to resolve model

### LINKED ISSUE #244: Model should be resolved by resolveRouteBinding method
**Describe the bug**
Right now the model in `\Filament\Resources\Pages\EditRecord::mount()` is resolved with a custom query that doesn't behave the same like in application route binding.

```php
$routeKeyName = (new $model())->getRouteKeyName();
$this->record = $model::where($routeKeyName, $record)->firstOrFail();
```

This current implementation will work in most cases but not if the user customizes `resolveRouteBinding()` method which is common to do if you want to use something like http://hashids.org 

The better way would be to use the model `resolveRouteBinding()` method instead of a custom query.

```php
$this->record = (new $model())->resolveRouteBinding($record) 
    ?? throw (new ModelNotFoundException())->setModel($model, [$record]);
```

Gummibeer:
fixes #244

### DIFF
--- a/src/Resources/Pages/EditRecord.php
+++ b/src/Resources/Pages/EditRecord.php
@@ -78,9 +78,8 @@ class EditRecord extends Page
 
         $model = static::getModel();
 
-        $routeKeyName = (new $model())->getRouteKeyName();
-
-        $this->record = $model::where($routeKeyName, $record)->firstOrFail();
+        $this->record = (new $model())->resolveRouteBinding($record) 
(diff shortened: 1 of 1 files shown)


danharrin:
Hey, using throw as an expression like this is only supported in PHP 8. We're supporting 7.4 at the moment as well, is it possible to make the required adjustments?

Gummibeer:
Sorry, sure. Was just copy-paste from my local environment to GitHub webUI.^^

danharrin:
Thanks!

PR #246, opened 2021-03-18
value should be casted to string

Gummibeer:
Right now the exact model value is used - with a package like https://github.com/spatie/laravel-enum this leads to problems.
This casts the real model value to string - this still works with numeric array keys as PHP casts the string back to integer if needed.

### DIFF
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -26,7 +26,7 @@ class Select extends Field
         $this->placeholder('forms::fields.select.placeholder');
 
         $this->getDisplayValueUsing(function ($value) {
-            return $this->getOptions()[strval($value)] ?? null;
+            return $this->getOptions()[(string) $value] ?? null;
         });
 
         $this->getOptionSearchResultsUsing(function ($search) {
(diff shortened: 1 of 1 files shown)


danharrin:
What setup is causing you problems?

Gummibeer:
The problem occurs on edit view if I have a model with enum attribute cast.
So `$post->status` is an instance of `StatusEnum`. The corresponding select has all enum values as options. But as objects can't be array keys it uses the string value.

Now if I open the edit view of that model I get an exception.

I can provide real code if wanted. But am on phone right now, so first tried to describe.

danharrin:
Do you think it'd be better to cast it with `(string)` instead of using `strval()`?

Gummibeer:
I think that the result is the same. So I don't care. Could also be casted with `string` argument typehint (I think).
I don't have performance benchmarks but this function isn't called a million times so it's unimportant.

Whatever feels better for you!? 😉

danharrin:
Yeah, if you could switch to using `(string) $value` that would be awesome. Thanks!

ryangjchandler:
@Gummibeer You okay making the change @danharrin asked for? If not, I can do it for you so we can get this merged in.

danharrin:
All done, thanks :)

Gummibeer:
Sorry, have planned to do it this evening.^^

PR #250, opened 2021-03-19
Fix dependent field for relation manager in record creation form

### LINKED ISSUE #240: Dependent Field not works for relation manager on the record creation form
Conditions are not triggered on the record creation form.

hadolen:
Fix #240


danharrin:
Thanks!

PR #252, opened 2021-03-20
Adds drag and drop reordering to tables

jellycow:
Uses https://github.com/livewire/sortable to manage table reordering. 

Adds new method to Table : reorderable($action)

Passing a valid livewire action into the method activates it and includes the https://github.com/livewire/sortable library to handle all the drag and drop functionality.

Demo
https://user-images.githubusercontent.com/209072/111874531-b67d6f80-898d-11eb-9947-4a70999b1237.mov

### DIFF
--- a/packages/tables/resources/views/components/table.blade.php
+++ b/packages/tables/resources/views/components/table.blade.php
@@ -50,6 +50,10 @@
                 @endforeach
 
                 <th scope="col"></th>
+
+                @if($table->isReorderable())
+                    <th scope="col"></th>
+                @endif
             </tr>
         </thead>
 
@@ -60,7 +64,7 @@
                     wire:loading.class="opacity-50"
                     class="{{ $loop->index % 2 ? 'bg-gray-50' : 'bg-white' }}"
                     @if($table->isReorderable())
-                      wire:sortable.item="{{ $record->getKey() }}"
+                        wire:sortable.item="{{ $record->getKey() }}"
                     @endif
                 >
                     <td class="p-4 whitespace-nowrap">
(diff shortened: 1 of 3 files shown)


ryangjchandler:
If we're going to do this, I think there should be a dedicated "drag" trigger, like a small handle or something. Making the entire row draggable is risky business.

jellycow:
Yup, sounds fair :-) The library actually has support for a drag handle element, so it would just be a case of deciding what the best option for specifying that was. Could add something like the dots-vertical hero-icon to the start or end of each row?

ryangjchandler:
> Yup, sounds fair :-) The library actually has support for a drag handle element, so it would just be a case of deciding what the best option for specifying that was. Could add something like the dots-vertical hero-icon to the start or end of each row?

Yeah, I've used the `sortable` library before for tables. I'll wait to see what @danharrin thinks too, because we've got the checkboxes on the left as well, we don't it to be too crowded.

jellycow:
Something like this

2021-03-20 07 30 35 pm

ryangjchandler:
@ryanscherler is the man to ask for design help - I personally think the 3 dots look like a menu icon, similar to what Google et al use.

danharrin:
I think we use a double hamburger for sorting elsewhere:

Screenshot 2021-03-20 at 20 14 48

jellycow:
That make sense, image was for placement reference really, double bar would be better I think but I'm not a designer by any stretch of the imagination so any input from the design inclined would probably be best :-)

ryangjchandler:
> I think we use a double hamburger for sorting elsewhere:
> 
> Screenshot 2021-03-20 at 20 14 48

Yeah, we use this on the `KeyValue` field for sorting.

jellycow:
Looks better with the suggested icon

2021-03-20 08 25 49 pm

danharrin:
Out of interest, how well does this work with relation managers? And if a pivot table is used, are we able to sort the records through that?

jellycow:
Don't know about pivot tables but it did work with a hasMany relationship

danharrin:
Hey! Thanks for this. I've rewritten it in #281 as part of a larger refactor, and made it easier to use:

- Call `reorderOn($column)` on a table: reordering will be enabled, a default sort (#231) will be added for that `$column`, and Livewire will automatically write to that column when the sort order is changed.
- Call `reorderUsing($callback)` on a table: reordering will be enabled, and you may specify a `$callback` that is used to save the sort order using the `$order` and `$livewire` callback parameters.
- Overwrite the `reorder()` method on the Livewire component: specify completely custom reordering logic.

jellycow:
Very nice work - thanks for including this.

cardentey:
hi, is this implemented in the latest version?
I can't find anything in the documentation.

Thanks for this job.

cardentey:
> Very nice work - thanks for including this.

hi, is this implemented in the latest version?
I can't find anything in the documentation.

Thanks for this job.

diazsasak:
Hi all,

how to implement reorder on the filament page/resource? i can't find it in the docs. thank you

danharrin:
Added to v2 in #3211

ryanscherler:
Awesome!

PR #256, opened 2021-03-21
feature: Markdown field auto bullet

ryangjchandler:
This pull request adds support for auto-bullet and auto-numbering lists in the Markdown field. That means if you've started an unordered list (for example) and hit enter, the field will automatically add in a new list item for you.

If you don't want that list item, you can just hit Enter again and it will be removed. This will also work for the `- ` syntax for unordered list.

Ordered lists will behave in the same way, the only difference being the numbers will increment on each new line item.


ryangjchandler:
I'm aware of how these nested if/elses look, but there's not really any other idiomatic way of doing it (or I'm being a fool). :(

PR #266, opened 2021-03-21
Changes to markdown editor

snellingio:
Several changes.

1. Normalize fonts. `font-mono leading-normal`. Unfortunately, if you have non-monospaced fonts, bolding and other operations will not line up properly in the editor / preview.
2. Hide scrollbars on textarea with `overflow-y-hidden`. 
3. Add a 1px offset to the overlay with `mt-px`. Because the input has a border, we're ever so slightly off. This fixes it. We could also add a transparent border or something.
4. Stop resizing the entire element, and resize the texture and overlay directly. This has the advantage of not messing with the flow of the document, and growing directly. 
5. Use the resize function in the init to make sure we have the correct height when loading in content.

You may wonder why we're resetting the height. It's a problem with the browser scrollHeight calculation that has been around for a long time (see: https://stackoverflow.com/questions/57953324/wrong-scrollheight-on-auto-grow-textarea-after-input-deletion) ... This little bug is infuriating, and really only affects the input when deleting content after resizing it larger.

### DIFF
--- a/packages/forms/resources/views/components/markdown-editor.blade.php
+++ b/packages/forms/resources/views/components/markdown-editor.blade.php
@@ -78,7 +78,7 @@
 
                     this.$refs.overlay.style.padding = window.getComputedStyle(this.$refs.textarea).padding
 
-                    this.overlay = mdhl.highlight(this.$refs.textarea.value)
+                    this.resize()
 
                     this.$watch('tab', () => {
                         if (this.tab !== 'preview') return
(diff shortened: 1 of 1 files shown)


snellingio:
A clearer illustration of the cursor not lining up, as well as incorrect scroll / resizing behavior

CleanShot 2021-03-21 at 15 49 27

ryangjchandler:
@snellingio We've already made a lot of these fixes on `develop` - would you mind checking out that branch and checking?

snellingio:
Okay, so on develop, it's close, but still not quite right.

without overlay including `border border-transparent` or `mt-px`:
CleanShot 2021-03-21 at 16 01 44@2x

with overlay including `border border-transparent` or `mt-px`:
CleanShot 2021-03-21 at 16 01 59@2x

It's close, but notice where the spellcheck lines up on the j's. 

The jelly scroll is still there without `overflow-y-hidden` on the textarea
CleanShot 2021-03-21 at 16 02 26

You got the rest!

snellingio:
I'm going to close this PR, and either let you take care of those two minor styles, or I can make a new PR

ryangjchandler:
> I'm going to close this PR, and either let you take care of those two minor styles, or I can make a new PR

Feel free to make a new PR from `develop`, looking to get this spot on!

PR #274, opened 2021-03-23
feature: create command to upgrate filament to the latest version

icarojobs:
#### Command to upgrade Filament to the latest version
I thought about adding this contribution because instead of typing 5 commands to perform this task, we just typed 1.


danharrin:
I like this! We won't need the alias though, the main command should be `filament:upgrade`.

src/Commands/MakeUpgradeCommand.php
```diff
@@ -0,0 +1,30 @@
...
+
+    protected $description = 'Upgrade Filament to the latest version.';
+
+    protected $signature = 'make:filament-upgrade';
+
+    public function handle()
+    {
+        system("composer update");
+
+        system("php artisan migrate");
```
ryangjchandler, line 20: Can we replace these calls with `$this->call` instead - it's definitely more reliable?

ryangjchandler (changes requested):

src/Commands/MakeUpgradeCommand.php
danharrin: Whoops only just seen this and I merged locally - will update.

PR #286, opened 2021-03-24
Add hooks to relations

sebapastore:
This PR tries to provide hooks to create, edit and attach related records.

To implement a hook, you have to add it in the parent relation manager, i.e:

```
class OrdersRelationManager extends RelationManager
{

    ...

    public static function afterCreate($record)
    {
        // do some stuff and access the record data if you need 
    }
}
```

 I will be happy to receive any comments on this in order to improve it.

### DIFF
--- a/resources/views/relation-manager.blade.php
+++ b/resources/views/relation-manager.blade.php
@@ -52,7 +52,7 @@
                 <x-filament::card class="space-y-5">
                     <x-filament::card-header :title="static::$createModalHeading" />
 
-                    @livewire(\Livewire\Livewire::getAlias(static::getCreateRecordClass()), [
+                    @livewire(\Livewire\Livewire::getAlias(Filament\Resources\RelationManager\CreateRecord::class), [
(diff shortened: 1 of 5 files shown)


ryangjchandler:
I think there might be a cleaner way of achieving this, without having to override each "action". Each of the modal handlers, i.e. `CreateRecord` should have access to the parent relation manager, so we could probably let users define hooks on that single class, reducing the amount of bloat.

sebapastore:
Thanks for the insights! I will work on that.

sebapastore:
There are cases in which is necessary to access record data in a hook. The relation manager doesn't have access to the record data. Should we pass the record in each hook?

sebapastore:
Now each of the modal handler has access to the parent relation manager. I have updated the use case in my first comment. Let me know if there are other improvements to make.

danharrin:
Hey! I've had to change the way relation managers work, so I've moved these changes to #281.

Thanks for your work!

sebapastore:
Cool, thanks for considering this!

PR #297, opened 2021-03-26
feature: bulk table actions

ryangjchandler:
This pull request provides a very basic implementation of bulk table actions. Currently, actions are only run on selected records but in the future, we might want to generalise it so that you can almost have "run whenever" actions too.

Actions can be defined inline, similar to a `Filter`, or they can be reusable classes.

To register an action, you can use the `$table->actions()` method in your `Resource` class.

```php
use Filament\Tables\BulkAction;

$table->columns([
    // ...
])->bulkRecordActions([
    BulkAction::make('Mark as Delivered', fn ($records) => $records->each->markAsDelivered()),
])
```

In the case of a `Resource` / `ListRecords` page, `$records` will be an instance of `Illuminate\Database\Eloquent\Collection` that you loop over, map over. Do what you want!

If you want to create a reusable `BulkAction` class, simply extend the `Filament\Tables\BulkAction` class and define a `run` method.

```php
use Filament\Tables\BulkAction;

class MarkAsDeliveredAction extends BulkAction
{
    public function run($records) {

    }
}
```

If you want to require a confirmation before running the action, you can chain on the `->requiresConfirmation()` method or define a `protected $requiresConfirmation = true` property.


ryangjchandler:
@danharrin The only things outstanding are translation strings, as well as the question above about destructive actions?

danharrin:
This looks great. One thing to start with - how do you feel about calling this feature `bulkRecordActions` instead? We already have `recordActions` so `actions` on its own could be confusing.

ryangjchandler:
@danharrin I've renamed everything to bulk record actions (or a variant of based on context). I think we can leave out the separation between destructive and non-destructive actions for now, probably best to get some users to test this once released and see if there's a need for that separation.

danharrin:
Closing until v2.

PR #300, opened 2021-03-28
Feature: Add the ability to define a post logout route

RVxLab:
This PR adds the ability to define a route after which users will be redirected to after logging out.

This can be changed in `filament.php` and defaults to the route that was defined in the `Logout` Livewire component as to maintain backwards compatibility.

I wasn't sure what exact style you maintain due to lack of a contributions guide, so I attempted to stick as close to what I saw as possible.


danharrin:
Thank you very much!

PR #301, opened 2021-03-29
Added Directory option to Image column in Tables

ClausMunch:
Directory attribute was missing.

### DIFF
--- a/packages/tables/src/Columns/Image.php
+++ b/packages/tables/src/Columns/Image.php
@@ -12,6 +12,8 @@ class Image extends Column
 
     protected $disk;
 
+    protected $directory;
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks for this! Where is the `$directory` actually being used here though? Isn't the directory saved by the `FileUpload` component in the database?

ClausMunch:
No, that is for some reason omitted. Might be the right place to fix it?

danharrin:
Yeah this is not an issue, since if your file upload field has a directory set, it's saved to the database. The directory is not actively consumed by the image column so it does not need to be set. Thanks!

PR #305, opened 2021-03-30
Fix issue in BelongsToSelect when initialDipslayValue contains a single quote

s-sadiq:
**Describe the bug**
Using BelongsToSelect field and the initialDisplayValue contains a single quote in the string for example "A'ali" or "
O'Reilly" will cause this alpine error.

```
Alpine Error: "SyntaxError: missing } after property list"

Expression: "select({
            autofocus: false,
            emptyOptionsMessage: 'Start typing to search...',
            initialDisplayValue: 'A'ali',
            initialOptions: [],
            name: 'record.area_id',
            noSearchResultsMessage: 'No options match your search.',
            required: true,
                            value:  window.Livewire.find('olD0buI7uf4vyKJTx7fF').entangle('record.area_id') .defer,
                    })"
```

**Proposed fix**

use `addslashes()` to handle this in case the initialDisplayValue string contains a single quote or double quote.

### DIFF
--- a/packages/forms/resources/views/components/select.blade.php
+++ b/packages/forms/resources/views/components/select.blade.php
@@ -192,7 +192,7 @@
         x-data="select({
             autofocus: {{ $formComponent->isAutofocused() ? 'true' : 'false' }},
             emptyOptionsMessage: '{{ __($formComponent->getEmptyOptionsMessage()) }}',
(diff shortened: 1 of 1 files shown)


ryangjchandler:
Wouldn't a simpler fix be wrapping the output in `json_encode()`? This should handle the escaped slashes.

s-sadiq:
Yes, I think `json_encode()` is a better option to solve this issue and works better with other value types.

danharrin:
Thanks!

PR #318, opened 2021-04-04
Add ability to disable RelationManager list edit action

mallardduck:
This just adds a very simple method to complement the `canCreate`, `canDelete`, and `canDetach` methods. The benefit of this is a very easy way to make the relation manager only allow attach and detach actions.

### DIFF
--- a/resources/views/relation-manager.blade.php
+++ b/resources/views/relation-manager.blade.php
@@ -33,20 +33,22 @@
             </x-filament::modal>
         @endif
 
-        <x-filament::modal
-            class="w-full max-w-4xl"
-            :name="static::class . 'RelationManagerEditModal'"
-        >
-            <x-filament::card class="space-y-5">
-                <x-filament::card-header :title="static::$editModalHeading" />
+        @if ($this->canEdit())
+            <x-filament::modal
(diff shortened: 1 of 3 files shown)


monoman81:
I made the same question earlier today on the filament discord server. What i have to do is this:

```
use Filament\Resources\RelationManager\ListRecords;
class ListMediaRecords extends ListRecords
{
    public function getRecordActions()
    {
        return [
        ];
    }
}
```
And in the RelationManager class override the $components static variable to tell filament that use my custom ListMediaRecords:

```
protected static $components = [
        'attach' => RelationManager\AttachRecord::class,
        'create' => CreateMediaRecord::class,
        'edit' => RelationManager\EditRecord::class,
        'list' => ListMediaRecords::class,
];
```
It works but seems to me a little bit of overkill if only need to remove the Edit action.

mallardduck:
@monoman81 - Thanks for providing that input, I think this helps to show how this PR could be a simpler alternative to that method. This method will also make the control of edit abilities match the others as mentioned in my OP.

monoman81:
I agree with you. Regards.

danharrin:
Thank you!

PR #343, opened 2021-04-12
Swedish translation

tanthammar:
As a thank you, for letting me "borrow" your Alpine date picker. (tnx @danharrin)


danharrin:
Thanks so much for these!

*(enjoy the date picker 😉)*

PR #344, opened 2021-04-12
Fix hour, minute, seconds in date-time-picker

tanthammar:
If you increment or decrement hour, minute, seconds in the date-time-picker and exceed the allowed range, the value falls back to current dayjs().

Example if the hour is 14, incrementing to 24 resets the value to 14 instead of 0.
The same goes for minutes, seconds, if current is, let's say 45 and you increment to 60 it resets to 45 instead of 0.

This PR fixes that.

### DIFF
--- a/packages/forms/resources/views/components/date-time-picker.blade.php
+++ b/packages/forms/resources/views/components/date-time-picker.blade.php
@@ -187,6 +187,8 @@
                     })
 
                     this.$watch('hour', () => {
+                        if (this.hour > 23) this.hour = 0
+                        if (this.hour < 0) this.hour = 23
                         this.hour = Number.isInteger(+this.hour) && this.hour >= 0 && this.hour < 24 ? +this.hour : dayjs().hour()
 
                         let date = this.getSelectedDate()
@@ -197,6 +199,8 @@
                     })
 
                     this.$watch('minute', () => {
+                        if (this.minute > 59) this.minute = 0
+                        if (this.minute < 0) this.minute = 59
                         this.minute = Number.isInteger(+this.minute) && this.minute >= 0 && this.minute < 60 ? +this.minute : dayjs().minute()
 
                         let date = this.getSelectedDate()
@@ -207,6 +211,8 @@
                     })
 
                     this.$watch('second', () => {
+                        if (this.second > 59) this.second = 0
+                        if (this.second < 0) this.second = 59
                         this.second = Number.isInteger(+this.second) && this.second >= 0 && this.second < 60 ? +this.second : dayjs().second()
 
                         let date = this.getSelectedDate()
--- a/packages/forms/resources/views/components/date-time-picker.blade.php
+++ b/packages/forms/resources/views/components/date-time-picker.blade.php
@@ -187,9 +187,17 @@
                     })
 
                     this.$watch('hour', () => {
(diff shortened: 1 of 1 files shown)


danharrin:
Hey! Thanks for these fixes. I've made a couple of adjustments - namely adding `min` and `max` attributes to the inputs, so that it's impossible to go to `-1` or `61`, for example. When a value out of range is typed manually, it is now corrected as in your original PR. Suitable adjustments are also made to the preceding unit also - exceeding seconds in a minute will increment the minute as well as setting the seconds to 0.

tanthammar:
Hi @danharrin 
I didn't add the min/max, thought you deliberately skipped it to avoid stopping at 0 on arrow down key :)

tanthammar:
Is it up for discussion, to auto switch day when changing the time?
It's a cool feature, but when thinking about how I would use the picker; I would first click on a day, then use the arrow keys to adjust time. 
Probably hitting the key to fast, so I'd first fast forward, then having to rewind, without noticing that the day I picked has changed. 

Maybe it's just me...

Thoughts?

tanthammar:
@danharrin 

Your changes returned the bug I intended to fix.
The problem is that the code resets the field to current time.
The value must be altered before this line
Example hour.
```
this.hour = Number.isInteger(+this.hour) && this.hour >= 0 && this.hour < 24 ? +this.hour : dayjs().hour()
```

tanthammar:
If you want the feature to go to next/previous day and fix the bug,
Here is a suggestion:

For the hour code:
```js
                this.$watch('hour', () => {
                    if (Number.isInteger(+this.hour)) {
                        if (this.hour > 23) this.hour = 0 & this.focusNextDay()
                        if (this.hour < 0) this.hour = 23 & this.focusPreviousDay()
                        if (this.hour >= 0 && this.hour < 24) this.hour = +this.hour
                    } else {
                        this.hour = dayjs().hour()
                    }

                    let date = this.getSelectedDate()

                    if (date === null) return

                    this.setValue(date.set('hour', this.hour))

                })
```

tanthammar:
Here is my full suggestion to avoid repetition:

Add this method
```js
incrementTime: function (field, switchValue, max) {
                if (Number.isInteger(+field)) {
                    if (field > switchValue) field = 0 & this.focusNextDay()
                    if (field < 0) field = switchValue & this.focusPreviousDay()
                    if (field >= 0 && field < max) field = +field
                } else {
                    field = dayjs().hour()
                }
                return field
            },
```

Then change the watchers
```js
this.$watch('hour', () => {
                    this.hour = this.incrementTime(this.hour, 23, 24)

                    let date = this.getSelectedDate()

                    if (date === null) return

                    this.setValue(date.set('hour', this.hour))

                })

                this.$watch('minute', () => {
                    this.minute = this.incrementTime(this.minute, 59, 60)

                    let date = this.getSelectedDate()

                    if (date === null) return

                    this.setValue(date.set('minute', this.minute))
                })

                this.$watch('second', () => {
                    this.second = this.incrementTime(this.second, 59, 60)

                    let date = this.getSelectedDate()

                    if (date === null) return

                    this.setValue(date.set('second', this.second))
                })
```

danharrin:
Hi @tanthammar, could you please confirm that #347 also fixes the issue? Many thanks.

PR #347, opened 2021-04-13
fix: Exceeding range with time picker

danharrin:
See #344.


tanthammar:
Yes this would do it :)

PR #349, opened 2021-04-14
Refactor CanCallHooks for greater reuse

mallardduck:
The same method in the trait is used in the same way on `Resources/Pages/Page.php`, so this just refactors to keep things uniform.


danharrin:
Thank you, this is great.

PR #354, opened 2021-04-16
Added two missing german translations

dominicbachmann:
I added two german transaltions which were missing. I checked that the translations styles fit the other translations that already exist.


danharrin:
Thanks!

PR #383, opened 2021-04-20
[Feature] Allow the checkbox and toggle to render as column fields

mallardduck:
Most fields by default render as a column, in other words the form label and input are in consecutive rows.
The exception to this seems to be checkbox and toggle which render inline with the field's label.

This PR adds a simple trait to allow those fields to render as a column. See examples here:

### Checkbox
```php
                Components\Checkbox::make('is_featured')
                    ->label('Featured Post')
                    ->stacked(),
                Components\Checkbox::make('is_featured')
                    ->label('Featured Post'),
```

Screen Shot 2021-04-20 at 10 16 45 AM

### Toggles
```php
                Components\Toggle::make('is_featured')
                    ->label('Featured Post')
                    ->stacked(),
                Components\Toggle::make('is_featured')
                    ->label('Featured Post'),
```
Screen Shot 2021-04-20 at 10 16 30 AM

---

The biggest benefit comes when used in a grid on a form:
Screen Shot 2021-04-20 at 10 20 51 AM


danharrin:
Hey, how about calling this `stacked()` or `vertical()`? I guess you could call the existing layout `inline()`.

mallardduck:
I like the sound of either of those. I'll make some changes now.

mallardduck:
@danharrin - decided to go with `stacked` and `inline`. Added methods for both so that it can be done either way.
This also seems like a good move because the "default" is for fields to be stacked. This way the trait conforms to the 90% of other fields.

Then for these fields that are "inline by default" we just call `inline()` in the `mount()` methods.

danharrin:
Thanks!

PR #388, opened 2021-04-22
Fix translation string from create-record resource page

dllobell:
This PR just fixes a small typo on the `CreateRecord` resource page.


danharrin:
Many thanks David

PR #397, opened 2021-04-26
Add missing return $this on name() and added type declarations on mul…

bernhardh:
…tiple methods.

I have only done it for methods, which aren't used in child or parent classes at the moment and only for the Field class. If you accept PR I would do it for other classes as well.

### DIFF
--- a/packages/forms/src/Components/Field.php
+++ b/packages/forms/src/Components/Field.php
@@ -6,26 +6,61 @@ use Illuminate\Support\Str;
 
 class Field extends Component
 {
-    protected $bindingAttribute = 'wire:model.defer';
+    /**
+     * @var string
+     */
+    protected string $bindingAttribute = 'wire:model.defer';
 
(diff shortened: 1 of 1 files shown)


danharrin:
Hey, can you remove the docblocks and types you added please. We're looking to fully type the codebase when we release a PHP 8+ version, but at the moment that is not a requirement.

bernhardh:
Removing the docblocks and types wouldn't be a problem. 

But found out, that the `name` method shouldn't be called (at least in the current form) from outside the constructor method at all, because you would need to add "record." as a prefix as in `ManipulatesResourceRecord` trait. So I don't know whats your prefered way to handle this. Either make this logic directly in the `name` method and remove the trait or make the `name` method protected or rethink the whole logic behind this?

bernhardh:
I will close this pull request and create an issue ticket for it. Thanks.

PR #409, opened 2021-05-06
DatePicker set first day of week and remove manual translation files

tanthammar:
**This PR suggests the following changes to the `DatePicker` field:**

* Remove the need for manual translation files
* Set any weekday as first day of week
* Datepicker language defaults to `app()->getLocale()`
* config option: `first-day-of-week`

Added methods:
```php
/**
     * Default = app()->getLocale()
     * DayJS i18 via browser: https://day.js.org/docs/en/i18n/loading-into-browser
     * Available methods: https://day.js.org/docs/en/plugin/locale-data
     * Supported locales: https://github.com/iamkun/dayjs/tree/dev/src/locale
     * @param string $dayJSlocale
     * @return $this
     */
    public function locale(string $dayJSlocale): self
    {
        $this->locale = $dayJSlocale;
        return $this;
    }

    /**
     * 0 to 7 are allowed. Monday = 1, Sunday = 0 or 7
     * @param int $day
     * @return $this
     */
    public function firstDayOfWeek(int $day = 1): self
    {
        if($day < 0 || $day > 7) $day = 1;
        $this->firstDayOfWeek = $day;
        return $this;
    }

    /**
     * Set Sunday as first day of week
     * @return $this
     */
    public function weekStartSunday(): self
    {
        $this->firstDayOfWeek = 7;
        return $this;
    }

    /**
     * This method is otiose because Monday is default
     * <br>But sometimes you want it for code readability
     * @return $this
     */
    public function weekStartMonday(): self
    {
        $this->firstDayOfWeek = 1;
        return $this;
    }


// + equivalent getters...
```


packages/forms/resources/views/components/date-time-picker.blade.php
```diff
@@ -425,20 +435,7 @@ class="absolute z-10 w-64 p-4 my-1 bg-white border border-gray-300 rounded shado
                             x-model="focusedMonth"
                             class="flex-grow p-0 text-lg font-medium text-gray-800 border-0 cursor-pointer focus:ring-0 focus:outline-none"
```
tanthammar, line 436: Ouch, forgot to add `class` "capitalize"

danharrin:
Thanks Tina! I've removed the unnecessary translation strings as well.

PR #410, opened 2021-05-07
remove Sushi as it's not used by filament

mallardduck:
This dependency is not actually being used by the project at all currently. However it's in the requirements so it downloads the package when users don't need it - and it also prevents users that do need it from tagging custom versions.


danharrin:
Thanks!

PR #411, opened 2021-05-07
Add placeholder component

mallardduck:
This PR allows for using a placeholder element in a form. The intention of this is to provide a UX similar to what LinkedIn does here:

So with this PR, if this UX were needed you could do:

```php
  Components\Grid::make([
      Components\Toggle::make('is_active')
          ->label('Currently working this role')
          ->default(true)
          ->dependable()
          ->stacked(),
      Components\DatePicker::make('started_at'),
// Our conditional field and equivalent placeholder
      Components\DatePicker::make('departed_at')
          ->when(fn ($record) => $record->is_active !== true),
      Components\Placeholder::make('departed_at', 'Present')
          ->when(fn ($record) => $record->is_active === true), // Note the callback logic here is inverted from above.
  ])->columns(3),
```

The end results would be:
Screen Shot 2021-05-07 at 5 21 01 PM
Screen Shot 2021-05-07 at 5 20 56 PM

### DIFF
--- a/packages/forms/resources/views/components/placeholder.blade.php
+++ b/packages/forms/resources/views/components/placeholder.blade.php
@@ -1,0 +1,14 @@
+<x-forms::field-group
+    :column-span="$formComponent->getColumnSpan()"
+    :error-key="$formComponent->getName()"
+    :for="$formComponent->getId()"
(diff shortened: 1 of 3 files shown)


danharrin:
Thanks!

PR #415, opened 2021-05-12
Allow to use html in prefixes plus add postfix method

bernhardh:
This change allow to use html in field prefixes. This is usefull, if you wanna add an icon as a prefix or if you have two fields with different length prefix (for example short currencies like `€` vs `CAD`) and want to make both of them same width with css classes.

# Icon Example:

```php
Components\TextInput::make("price")
   ->numeric()
   ->prefix('<svg xmlns="http://www.w3.org/2000/svg" class="h-6 w-6" fill="none" viewBox="0 0 24 24" stroke="currentColor">
  <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M14.121 15.536c-1.171 1.952-3.07 1.952-4.242 0-1.172-1.953-1.172-5.119 0-7.072 1.171-1.952 3.07-1.952 4.242 0M8 10.5h4m-4 3h4m9-1.5a9 9 0 11-18 0 9 9 0 0118 0z" />
</svg>')
```

# Result

image


bernhardh:
And second commit adds a `postfix` method to input fields, with the same logic as the `prefix` just on the other side:

image

danharrin:
Nice, thanks!

PR #418, opened 2021-05-12
Allow to change classes of components (example)

bernhardh:
This is a draft. I will finalize it, if you will accept the PR later. Atm I have implemented it just for `Components\Section`.

What does it do? It gives the user the possibility to change the css classes on components. For example, you can now change the background color of a section. This is how a regular section looks like (if outside a card):

image

Now you can change the bg-color:

```php
Components\Section::make(
    'Heading',
    'Subheading',
    [],
)
    ->changeClass('bg-blue-200', 'bg-gray-50')
```

image

Or you can even remove all stylings:

```
Components\Section::make(
    'Heading',
    'Subheading',
    [],
)
    ->changeClass('', true)
```

image

I can add this functionality to all components if you like. What do you think?


danharrin:
Hey, I'm going to leave this, since we've decided on a different system for v2. Many thanks for your work!

PR #425, opened 2021-05-23
Allow overwrite phpunit.xml only locally

Muhammad-Sarfaraz:
As the phpunit looks for phpunit.xml then fallbacks to phpunit.xml.dist. We can prevent the local phpunit.xml to overwrite the master phpunit configuration. It will allows us to have some variants in local development.


danharrin:
Thanks!

Muhammad-Sarfaraz:
Thanks to you too for this nice package.

PR #429, opened 2021-05-26
fix file upload isn't cleared after create

dark-wind:
file upload box isn't cleared after created ， [look at last comment](https://github.com/laravel-filament/filament/issues/417)


danharrin:
Thank you! This PR also covers the same fix for the select component.

PR #488, opened 2021-07-19
Add "expanded" attribute to card component

danilopolani:
This PR introduces a `expanded` attribute that can be passed to cards to set them full-width. It's particularly useful when building a widget that needs to display large data, such as charts etc.

An example of the default widget with the attribute:

```blade
<x-filament::card class="flex" expanded>
    ...
</x-filament::card>
```

image

**Please note** that a new assets build is required to include the `col-span-full` class into the minified CSS 🌈

### DIFF
--- a/resources/views/components/card.blade.php
+++ b/resources/views/components/card.blade.php
@@ -1,3 +1,5 @@
-<div {{ $attributes->merge(['class' => 'bg-white shadow-xl rounded p-4 md:p-6']) }}>
+@props(['expanded' => false])
+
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks!

danilopolani:
> Thanks!

Hey Dan, I guess we forgot to rebuild the assets 😞

danharrin:
> > Thanks!
> 
> Hey Dan, I guess we forgot to rebuild the assets 😞 Right now the class `col-span-full` is not available in the purged current CSS of Filament

Good spot! Fixed it.

PR #520, opened 2021-08-06
add code to check if a page should be shown in the navigation 

rabol:
Add a check to see if the page class have a method called showInNav, if it has this method, call it so see if the pages should be registered.
It is not a breaking change.


ryangjchandler:
@rabol No code changes needed to achieve this. You can simply overwrite the `navigationItems` method on the class and return an empty array, `[]`.

```php
public static function navigationItems()
{
    return [];
}
```

PR #526, opened 2021-08-16
Fix date column always shown in english

Vextil:
Changes the use of Carbon's ```format``` for ```translatedFormat``` so ```Text``` columns containing dates are shown in the correct language when changing locales.


danharrin:
Thanks!

PR #536, opened 2021-08-21
[MOD]: modify card && table by remove shadow-xl class

moh-abdullah-it:
remove big shadow from card && table


danharrin:
Hi, whats the reason behind this change?

moh-abdullah-it:
For a better user experience and user interface, I think the shadows suggest layers.
And I don't prefer having a shadow in large elements.
And the absence of shadows or a little percentage works to relax the eye and keep distractions away.
I see that the design has become cleaner after removing the big shadow

danharrin:
Thanks!

PR #544, opened 2021-08-24
Allow specifying file name for uploaded files

chiiya:
Similar to Nova, I want to be able to specify a custom filename for uploaded files. Use cases:
- Store the file with the original filename
- Use a field on the model (e.g. slug or id) to set the filename

This is how it works for Nova:

```php
File::make('icon')
    ->storeAs(function (Request $request) {
        $slug = Str::slug($request->input('slug'));
        $file = $request->file('icon');

        return $slug.'.'.$file->getClientOriginalExtension();
    }),
``` 

Instead of working with the request object, I believe we can only use the Livewire `$record`? Which when when creating a record is an array but when updating is a model instance. Please let me know if there is a better way to handle this. I also wasn't sure where to place the logic (form builder or admin panel), since a resource is always involved.

This is how it would look like with my changes:

```php
Components\FileUpload::make('icon')
    ->storeAs(function (UploadedFile $file, ParameterBag $input) {
        return $input->get('slug').'.'.$file->getClientOriginalExtension();
    }),
```

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,1 +1,1 @@
-github: [danharrin, ryangjchandler]
+github: [danharrin]
--- a/.github/workflows/monorepo-split.yml
+++ b/.github/workflows/monorepo-split.yml
@@ -18,7 +18,6 @@ jobs:
               run: echo "::set-output name=matrix::$(vendor/bin/monorepo-builder packages-json)"
         outputs:
             matrix: ${{ steps.packages-list.outputs.matrix }}
-
     split-monorepo:
         needs: provide-packages-list
         runs-on: ubuntu-latest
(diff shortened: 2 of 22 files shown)


danharrin:
Thanks for your work on this 🙏

I think it's a bit tricky to implement this feature neatly in v1, so I will take your work here and manipulate it into a feature for v2, where it will be much easier. Of course you could always create a custom field and use this in your own app.

Hope that's ok!

chiiya:
Updated the PR

danharrin:
Thanks!

danharrin:
This feature has been implemented already in upcoming v2 btw - you can use `saveUploadedFileUsing()` to customise save behaviour, including setting a custom filename.

https://github.com/laravel-filament/filament/blob/2.x/packages/forms/src/Components/FileUpload.php#L297-L325

PR #564, opened 2021-09-07
Remove transition effect for sections

martinmildner:
In my opinion, the backend looks more professional and cleaner if the transition effect is not used when form sections are opened.

### DIFF
--- a/packages/forms/resources/views/components/section.blade.php
+++ b/packages/forms/resources/views/components/section.blade.php
@@ -54,7 +54,7 @@
         </div>
     </div>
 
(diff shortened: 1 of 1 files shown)


ryangjchandler:
@martinmildner This is technically a breaking change as we're removing functionality / abilities. Can we make this optional instead, with a method on the field type.

danharrin:
I'd rather not have a an additional methods for this. I will remove the animation in v2 if this reply gets more 👍 than 👎 .

danharrin:
This will be in v2.

PR #566, opened 2021-09-08
feature: add meta to fields

ryangjchandler:
This pull request allows `Component` instances to have additional meta stored against them.

A good example would be a `macro` on the `Component` class. You might want to place a marker on the `Component` to reference at a later date.

This can be done at the moment with dynamic class properties, but that'll likely be deprecated in a future version (https://wiki.php.net/rfc/deprecate_dynamic_properties).

### DIFF
--- a/packages/forms/src/Components/Component.php
+++ b/packages/forms/src/Components/Component.php
@@ -23,6 +23,7 @@ class Component extends ViewComponent implements Htmlable
     use Concerns\HasExtraAttributes;
     use Concerns\HasId;
     use Concerns\HasLabel;
+    use Concerns\HasMeta;
     use Concerns\HasState;
     use Concerns\HasView;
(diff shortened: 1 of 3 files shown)


ryangjchandler:
@danharrin The only thing I'm not sure about right now is the behaviour when you pass an array of keys. It's using `Arr::only` so will return a keyed array.

```php
$field->getMeta(['foo', 'bar']);

// Would return this.
['foo' => 'bob', 'bar' => 'baz'];
```

Not sure if it should return that, or an array without the keys. If you did it without the keys, you'd be able to list the values out:

```php
[$foo, $bar] = $field->getMeta(['foo', 'bar']);
```

danharrin:
It makes sense the way it was written. It's better to return more information than less. You can just run `array_values()` and destructure that instead.

PR #595, opened 2021-09-17
feature: support reactive/lazy modifiers on container components

ryangjchandler:
This PR will allow container components, e.g. `Grid`, `Section`, etc to accept `reactive()`, `lazy()` and `stateBindingModifiers()` calls and then proxy those calls down to child components.

This removes multiple calls to those methods on the child components.


danharrin:
I've just refactored this so that the component inherits its state from its container, and the container inherits its state from its parent component, instead of the state being passed down the component tree.

What do you think @ryangjchandler?

ryangjchandler:
LGTM!

PR #606, opened 2021-09-24
Fix password reset with laravel users

underdpt:
This PR fixes the password reset procedure when using Laravel users table instead of Filament's one.

It also fixes the wrong lang file being used for reset password errors (this got merged into this PR by mistake)

### DIFF
--- a/docs/01-getting-started.md
+++ b/docs/01-getting-started.md
@@ -96,6 +96,8 @@ public function isFilamentAdmin()
 }
 ```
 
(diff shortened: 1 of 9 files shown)


danharrin:
Thanks for this. I think I realise why you suggested we should separate the traits btw, could you do that for me please? Otherwise this would be a breaking change I think. Unless, we can detect in `sendPasswordResetNotification()` if the request is coming from Filament or not.

danharrin:
Thanks!

underdpt:
Thank you!

PR #618, opened 2021-09-28
New TimePicker component

titantwentyone:
New TimePicker component. Usage:

```
return $form
    ->schema([
        Components\TimePicker::make('start_time')
                ->withoutSeconds(), //optional
    ]);
```


danharrin:
In #654, I've refactored the entire implementation so that all components extend `DateTimePicker` instead of `DatePicker`.

Thanks for your work here, it formed the base of that PR!

PR #629, opened 2021-10-04
document resource page code snippet updated

zobay:
## Description

- In resource page of docs, There is a mention of `OrdersRelationManager.php` and so on after the code snippet. 
- But, In code snippet, this is `PostsRelationManager`. 
- I think, If the class name is `OrdersRelationManager`, it will be more realistic with docs.


danharrin:
Thanks!

PR #634, opened 2021-10-04
Fix for labels not working for relationship fields (select and multiselect)

shamil8124:
- If a label is present, it should be used, instead of using the relationship name.


danharrin:
Thank you!

PR #635, opened 2021-10-04
fixing collapsed states

shamil8124:
This PR should fix referenced in this discussion item here: https://github.com/laravel-filament/filament/discussions/632

Basically, if we pass a condition or callable to the `collapsed()` method, it should only affect the collapse state, and should not convert the section back to a non-collapsible one.


danharrin:
Thanks!

PR #641, opened 2021-10-07
Fix small bug in MultiSelect template

pmartelletti:
When an item was selected, it was not being displayed properly because the x-text was using an unexistent array.

### DIFF
--- a/packages/forms/resources/views/components/multi-select.blade.php
+++ b/packages/forms/resources/views/components/multi-select.blade.php
@@ -154,7 +154,7 @@
                             'cursor-default' => $isDisabled(),
                         ])
                     >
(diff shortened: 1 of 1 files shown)


pmartelletti:
After a brief discussion with @danharrin on discord, I think the issue is valid but the solution is not the appropriate one (won't work for every component that uses multi-select).

I think the issue is around how the "labels" property is being filled up on the multi-select.js file.

danharrin:
Hey, you need to specify a `getOptionLabelsUsing()` method I think.

```
$select->getOptionLabelsUsing(function (array $values) {
    $labels  = []; // Do something with `$values` and transform it into a key => value array of values and labels.

    return $labels;
})
```

Let me know if this fixes your issues!

PR #642, opened 2021-10-07
Update Docs

yob-yob:
Inform users that when using `getValueUsing` they could not use `sortable` and `searchable`


danharrin:
Thanks!

PR #663, opened 2021-10-20
Fix example field names for consistancy

roni-estein:
Super odd, i can't tell whats happening with the repository.

I made some changes to field name examples for it to read more consistently and make sense in only one file and I've destroyed and re-forked my own filament repo twice trying to figure this out. However only 1 file has changed.

packages/forms/docs/03-fields.md

Please check only this file. Not sure where all the other changes are coming from


danharrin:
Ah, I think you forked the `main` branch instead of the `2.x` branch.

roni-estein:
Yeah I figured it out in bed... it's like the gremlins. Never PR past midnight :) it's an 80's film reference in case you are scratching your head 😁

PR #675, opened 2021-11-02
Update variable names to work with intended example

roni-estein:
Changers to some example field names to clarify intentions


danharrin:
Awesome, thanks!

PR #679, opened 2021-11-03
Allows pattern definition to work

roni-estein:
Without this flag, json_encode will escape the slashes that the configuration. Consider the example from the iMask docs

```
var zipMask = IMask(element, {
  mask: '#00000',
  definitions: {
    // <any single char>: <same type as mask (RegExp, Function, etc.)>
    // defaults are '0', 'a', '*'
    '#': /[1-6]/
  }
});
```


danharrin:
Thanks!

PR #684, opened 2021-11-03
Add autocapitalize attribute to  maintained concerns

roni-estein:
It creates a more direct way to overwrite autocapitalize and deals with at the concerns level rather than in extraAttributes for greater clarity. The side effect of this should be on mobile devices that the keyboard is always in caps mode.


danharrin:
Thanks!

PR #688, opened 2021-11-05
Added an object override for direct access to imask configuration

roni-estein:
A good example of  this would be when you want to make something a little more complicated from imask, here is an example from the modified from the docs for a [pattern](https://imask.js.org/guide.html#masked-pattern)

The example
```js

var zipMask = IMask(element, {
  mask: '#00000',
  definitions: {
    // <any single char>: <same type as mask (RegExp, Function, etc.)>
    // defaults are '0', 'a', '*'
    '#': /[1-6]/
  }
});

```

We are going to extract the second argument javascript object and drop it in as our override string using [nowdoc syntax](https://wiki.php.net/rfc/flexible_heredoc_nowdoc_syntaxes)

Usage:

```php

TextInput::make('numbers')
    ->mask(fn (TextInput\Mask $mask) => $mask->overrideWithObject(<<<field
    {
      mask: '#00000',
      definitions: {
        // <any single char>: <same type as mask (RegExp, Function, etc.)>
        // defaults are '0', 'a', '*'
        '#': /[1-6]/
      }
    }
    field)
)

```

There is at least one area where this will fail. Using the Date configuration will work, if you object only consists of `{ mask: Date }`which uses a default pattern of DD.MM.YYYY actually denoted as the pattern 'd{.}m{.}Y' however once you try change the pattern to 'm{/}d{/}Y' or anything else in the Date configuration, odd things start happening. There is also a warning that the state is being modified outside of imask.

'Element value was changed outside of mask. Syncronize mask using `mask.updateValue()` to work properly'

Looking further into `getMaskOptionsUsing` may resolve this. If someone stronger in js could take a peek :) that would be awesome, if not I will poke around next weekend for a bit.

PR'd with ❤️


danharrin:
Thanks!

PR #697, opened 2021-11-11
(fix) Fix type hinting for sortable()

dsturm:
As stated in the [docs](https://filamentadmin.com/docs/2.x/tables/columns#sorting), argument $condition could be `bool` or `array`. Because only `bool` is currently allowed, we cannot use an accessor column.


danharrin:
Thanks!

PR #702, opened 2021-11-12
Change method cache from boot to booted Trait 

wit3:
Move all table entities from

- bootInteractsWithTable 
to 
- bootedInteractsWithTable

Now the public properties setted in mount method is accessible.

#701 Fixed


wit3 (commented):
Sorry for leave this empty method,

Reading the documentation on Livewire site, it's better to use `mountInteractsWithTable` method?

danharrin:
Oh, `mountInteractsWithTable()` works? I didn't know that existed.

danharrin:
Thanks!

PR #707, opened 2021-11-16
(fix) Use extraAttributes in Card component

dsturm:
As stated in the [docs](https://filamentadmin.com/docs/2.x/forms/layout#custom-attributes) you should be able to set custom attributes (i.e.) to the Card component. This fix ensures these are used in the view file.


danharrin:
Thanks!

PR #712, opened 2021-11-17
Update 03-fields.md

jeffgreco13:
Update docs to include passing boolean value to disabled method on TextInput


danharrin:
Thanks!

PR #719, opened 2021-11-22
Allow unescaping labels

ralphjsmit:
This PR adds functionality for labels to not be escaped when they are rendered. This is useful for things like emitting events (see the example in the docs) or adding links to other pages.

### DIFF
--- a/packages/forms/docs/03-fields.md
+++ b/packages/forms/docs/03-fields.md
@@ -35,6 +35,17 @@ use Filament\Forms\Components\TextInput;
 TextInput::make('name')->label(__('fields.name'))
 ```
 
+You may additionally pass a boolean to the `label()` method to determine whether the label should be escaped:
+
+```php
+use App\Models\Post;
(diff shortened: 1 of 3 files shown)


danharrin:
Thanks for the PR. IMO, we should instead pass the label through `Str::markdown()` (same as with hints) and not provide this as a configurable option. Labels are developer-defined and not user-defined so it's not very unsafe.

ralphjsmit:
Okay, so you're saying to replace the `{!! $label !!}}` with `{{ Str::markdown($label) }}`?

Would that work with `wire:click=""` attributes as well?

danharrin:
I think it should be `{!! \Illuminate\Support\Str::markdown($label) !!}` right? Isn't that what we already have for helper text?

danharrin:
Yes please, move to `Str::markdown()` for consistency.

danharrin:
Thanks!

PR #722, opened 2021-11-24
Make Markdown editor controls wrap on smaller devices

ralphjsmit:
I noticed that on smaller devices the controls for the markdown editor overflow the container and cause horizontal scroll. This PR fixes that by adding a `flex-wrap` utility for the smaller devices. 

**Before:**
Schermafbeelding 2021-11-24 om 08 27 57

**After:**
Schermafbeelding 2021-11-24 om 08 27 18


danharrin:
I think that these buttons should instead scroll horizontally without affecting the form width - this keeps it in line with Trix.

danharrin:
Thanks!

Issue #726, opened 2021-11-25
SpatieTagsInput error when  Registering a model class


budiardianata:
### Package

filament/spatie-tags-plugin

### Package Version

v 2.43.7

### Laravel Version

v 8.73

### Livewire Version

v 2.8

### PHP Version

v  8.0

### Bug description

SpatieTagsInput error bind model when i [Registering a model class](https://filamentadmin.com/docs/2.x/forms/building-forms#registering-a-model-class)
![Untitled](https://user-images.githubusercontent.com/17608807/143375168-9c28f551-5556-481f-9130-d2378a237bc7.png)

### Steps to reproduce

change  `$model`  parameter type to `Model | string | callable | null` on the `afterStateHydrated` callback function.

### Relevant log output

```shell
Filament\Forms\Components\SpatieTagsInput::Filament\Forms\Components\{closure}(): Argument #2 ($model) must be of type ?Illuminate\Database\Eloquent\Model, string given, called in D:\CODE\LARAVEL\boilerplate\vendor\laravel\framework\src\Illuminate\Container\BoundMethod.php on line 36
```

PR #727, opened 2021-11-25
fix SpatieTagsInput error

budiardianata:
This PR  try to fix SpatieTagsInput error when Registering a model class #726.


danharrin:
Thanks!

PR #728, opened 2021-11-25
Update forms documentation with additional search options

nkeena:
This PR documents the `getSearchResultsUsing()` and `getOptionLabelUsing()` methods to further customize the `searchable()` method on a select field.

These methods are very helpful when needing to search a very large result set either directly from the database or an external data source such as an API.


danharrin:
Thanks!

PR #729, opened 2021-11-25
Radio field

nkeena:
This PR adds a new field called `Radio`.  This allows you to define a radio button group using the following API:

```php
use Filament\Forms\Components\Radio;

Radio::make('status')
    ->options([
        'draft' => 'Draft',
        'scheduled' => 'Scheduled',
        'published' => 'Published'
    ])
```

The HTML output would look something like this:

Screen Shot 2021-11-24 at 10 40 46 PM

You can optionally add descriptions to the options:

```php
use Filament\Forms\Components\Radio;

Radio::make('status')
    ->options([
        'draft' => 'Draft',
        'scheduled' => 'Scheduled',
        'published' => 'Published'
    ])
    ->descriptions([
        'draft' => 'Is not visible.',
        'scheduled' => 'Will be visible.',
        'published' => 'Is visible.'
    ]);
```

The HTML output would look something like this:

Screen Shot 2021-11-24 at 10 20 34 PM

If you only want a simple `Yes` or `No` radio button group that returns a `boolean` value:

```php
use Filament\Forms\Components\Radio;

Radio::make('is_public')->boolean()
```

HTML Output:

Screen Shot 2021-11-24 at 10 20 08 PM

I've also updated the documentation to include a new field called `Radio`.

### DIFF
--- a/packages/forms/docs/03-fields.md
+++ b/packages/forms/docs/03-fields.md
@@ -408,7 +408,7 @@ class App extends Model
     protected $casts = [
         'technologies' => 'array',
     ];
-    
+
     // ...
 }
 ```
@@ -475,7 +475,7 @@ class User extends Model
     protected $casts = [
         'is_admin' => 'boolean',
     ];
-    
+
     // ...
 }
 ```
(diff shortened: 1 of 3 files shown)


danharrin:
This is great. Sorry for the delay merging, been busy with v2 admin work :)

PR #736, opened 2021-11-26
Add default filter

underdpt:
Hello,

This PR adds the ability to set a default filter on tables. That filter is applied by default if no other filter is selected by the user, and the filter name is shown on the filter dropdown.

From https://github.com/laravel-filament/filament/discussions/551#discussioncomment-1688544

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -398,34 +398,30 @@
         },
         {
             "name": "doctrine/inflector",
-            "version": "2.0.3",
+            "version": "2.0.4",
             "source": {
                 "type": "git",
                 "url": "https://github.com/doctrine/inflector.git",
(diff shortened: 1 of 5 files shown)


danharrin:
Hey! Similar to how we have a `defaultSort()` method on the table object, maybe we should have a `defaultFilter()`? Then, we can set the filter property in LW straight from that value and don't need to loop through all filters to find the default one. Thoughts?

underdpt:
Yeah, this way feels cleaner.

danharrin:
Thanks!

underdpt:
My pleasure!

Thanks for your edits, I've learnt something new!

danharrin:
Sorry for the delay merging - been buried in work for v2!

PR #738, opened 2021-11-30
SelectFilter - Fix bug with falsy values

robertorinaldi-dev:
As agreed on discord, I made a PR to solve the issue with falsy values ignored by SelectFilter.

### DIFF
--- a/packages/tables/src/Filters/SelectFilter.php
+++ b/packages/tables/src/Filters/SelectFilter.php
@@ -18,7 +18,7 @@ class SelectFilter extends Filter
             return parent::apply($query, $data);
         }
 
-        if ($data['value'] === null || $data['value'] === '') {
+        if (blank($data['value'])) {
             return $query;
         }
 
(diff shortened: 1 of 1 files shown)


packages/tables/src/Filters/SelectFilter.php
```diff
@@ -18,7 +18,7 @@ public function apply(Builder $query, array $data = []): Builder
             return parent::apply($query, $data);
         }
 
-        if (! $data['value']) {
+        if ($data['value'] === null || $data['value'] === '') {
```
ryangjchandler, line 21: Could we not use the `blank()` helper function here?

ryangjchandler (changes requested):

packages/tables/src/Filters/SelectFilter.php
roberto-mgquadro: Yup!

danharrin:
Thanks!

PR #745, opened 2021-12-01
Add ability to sort widgets

underdpt:
Hello,

This PR adds the ability to sort widgets, similar how you can sort the right menu options.

It's backwards compatible, so it won't break or change ordering of existing widgets if no order is set.

### DIFF
--- a/src/FilamentManager.php
+++ b/src/FilamentManager.php
@@ -147,7 +147,9 @@ class FilamentManager
 
     public function getWidgets()
     {
-        return $this->widgets;
+        return collect($this->widgets)
+            ->sortBy(fn ($widget) => $widget::$sort ?? 0)
+            ->toArray();
     }
 
     public function ignoreMigrations()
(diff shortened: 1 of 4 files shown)


danharrin:
Hey, please move the sorting logic to `FilamentManager` so that plugin-registered widgets are also sorted.

Also, add the `$sort` attribute to the base `Widget` class, not the stub.

underdpt:
Hello,

I've just taken a look into this, and I'm not sure which way is preferable, on `FilamentManager`, having in mind that

- `$widgets` is a public property, so it should be ordered by default in case someone is using it directly
- Being `$widgets` a public property, it can be modified by the developer, so the order would be lost

So the two options are:

- Order `$widgets` when calling `registerWidget()`. This would be less performant, and would ignore those two points before
- Order the widgets when calling `getWidgets()`. This is only called one time, so it should be more perfomant. The widgets would be ordered also in case the developer tinkers with `$widgets` property, but not if he uses that property directly.

I think the second one is better, what do you think?

danharrin:
Hi,

Yeah, exactly. I was hinting at using `getWidgets()` to process sorting.

underdpt:
Done.

I'm using a intermediate collection because it will allow for cleaner code when adding #746 to the same method.

danharrin:
Thanks!

PR #746, opened 2021-12-01
Add ability to hide widgets

underdpt:
Hello,

This PR adds the ability to hide widgets with a static property. Use case is when you have developed some widgets that are not required anymore, but you don't want to remove them as they may be needed on the near future.

I think this can be made more useful by adding a static function `hide()` which defaults to `false`, so you can show or hide the widget at runtime (for example, show a widget based on user roles, or on some metrics). I would add it (backwards compatible) if that adds more value to this PR.

### DIFF
--- a/src/FilamentServiceProvider.php
+++ b/src/FilamentServiceProvider.php
@@ -255,7 +255,8 @@ class FilamentServiceProvider extends ServiceProvider
             })
             ->filter(function ($class) {
                 return is_subclass_of($class, Widget::class) &&
-                    ! (new ReflectionClass($class))->isAbstract();
+                    ! (new ReflectionClass($class))->isAbstract() &&
+                    ($class::$hidden ?? false) === false;
             })
             ->each(fn ($widget) => Filament::registerWidget($widget));
     }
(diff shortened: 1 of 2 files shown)


danharrin:
Hey, similar to #745, please move the hidden logic to `FilamentManager`, and the property onto the base class.

Also, please rename the property to `$isHidden` for consistency. Thanks!

underdpt:
Hello,

Similar to https://github.com/laravel-filament/filament/pull/745#issuecomment-984443090, there's two ways to accomplish this. In this case I clearly prefer to add the logic to `getWidgets()` function, so the `$widgets` property has all the widgets (no BC) and maybe add a `$onlyVisible` optional parameter to `getWidgets()` and modify `resources\views\dashboard.blade.php` to use that parameter, so we make sure nothing changes if someone is calling `getWidgets()` on his own implementation.

danharrin:
Yup! Please go ahead with the `getWidgets()` method.

underdpt:
I'll wait until #745 is merged and start from there because both PR targets the same method and the merge might be a mess.

danharrin:
All merged :)

PR #749, opened 2021-12-02
Add ability to hide widgets

underdpt:
This is a refactored version of #745 with the changes suggested by @danharrin and a `isVisible()` static method on base `Widget` class to allow for better readability and further customization by the developer.


danharrin:
Since widgets are visible by default, we can safely remove the extra `$onlyVisible` parameter as it's not a BC.

underdpt:
Updated!

danharrin:
Thank you

PR #762, opened 2021-12-07
SelectFilter using whereRelation in related columns

dmandrade:
Filter data by related column using whereRelation.

this will allow simpler filters without  query(), example:

`SelectFilter::make('client.name')->query(fn ($query, $data) => $query->whereRelation('client', 'name', $data['value']))`

only need to write:

`SelectFilter::make('client.name')`


danharrin:
Hey! I've just made some changes to your PR that make it even easier to create relationship selects. However, it does break the original behaviour in this PR.

Select filters are now able to automatically populate themselves based on a `BelongsTo` relationship. For example, if your table has a `client` relationship with a `name` column, you may use `relationship()` to filter the records belonging to an client:

```php
use Filament\Tables\Filters\SelectFilter;

SelectFilter::make('client')->relationship('client', 'name')
```

Dot-syntax is not supported in the `make()` method, as it causes issues with scoping the filter form's state.

kurucu:
Unless I've done something wrong, I can't see how to filter those clients (using the above example) e.g. to only show those for the current user. Have I missed something?

danharrin:
@kurucu You can still specify custom filter `->options()` with your own list if you wish?

kurucu:
If you mean like this:

```
SelectFilter::make('project')
    ->options(auth()->user()->projects()->pluck('name', 'id'))
    ->column('project_id'),
```

Then yep - that's what I've gone for! Seems to work a charm; just wanted to make sure I wasn't missing anything.

danharrin:
Yeah, the `relationship()` method is really just a shortcut :)

PR #763, opened 2021-12-07
Refactor applySearchConstraint and applySort methods

dmandrade:
Use whereRelation in search clause and update sort to use `getRelationExistenceQuery()`, a builtin method from Eloquent's relation class.

Also, replace IF ELSE by $query->when()


danharrin:
Thanks!

PR #768, opened 2021-12-08
+setHidden properties to empty array on fillForm record

daverydan:
Please update the fillForm() in Filament\Resources\Pages\EditRecord + Filament\Resources\Pages\ViewRecord as the password field wasn't available as a value passed on the form submit due to it being in the $hidden attribute array in the User model.


danharrin:
Hey, can you see if my new changes instead still fix the issue you were facing?

danharrin:
I assume it did :)

PR #777, opened 2021-12-08
Add comparison rules for dates and times

ralphjsmit:
This pull request adds comparison rules for dates and times, similar to the `gte()`, `lte()`, etc methods.


danharrin:
Can you add these to do docs please? :)

ralphjsmit:
Yeah of course, forgot about that. Added them!

danharrin:
These methods don't actually work, right? Because they're not field comparison rules, they're normal rules in their own right and don't depend on other fields. Maybe create a new `dateComparisonRule()`:

```php
    protected function dateComparisonRule(string $rule, string $date): static
    {
        $this->rule(
            fn () => "{$rule}:{$date}",
            fn (): bool => (bool) $this->evaluate($date),
        );

        return $this;
    }
```

ralphjsmit:
I'll test it tomorrow again and let you know. Initially I created the helpers in the vendor directory of a project I was working on, before copying them over to the repository itself. I thought that they worked back then.

I only tested it on an array though, could that have an effect? Like:
```php
$array = [
[
   'start_time' => '10:00',
   'end_time' => '12:00',
],
[
   'start_time' => '14:00',
   'end_time' => '15:00',
]
];

**Comment by danharrin** [comment]
Not sure what you mean there. Judging by your docs, it cant be used with other fields, just set.

**Comment by ralphjsmit** [comment]
Hey Dan, I tried the new functions out and I noticed that:

1. They work correctly with top-level fields.
2. They don't work inside repeaters.

I tried adding your proposed function and changing the implementation, but that didn't have any effect on both conclusions.

Any suggestion how we should make this work inside repeaters?

```php
<?php

namespace App\Http\Livewire;

use Filament\Forms\Components\DatePicker;
use Filament\Forms\Components\Repeater;
use Livewire\Component;

class FormDateComparisonsTest extends Component implements HasForms
{
    use InteractsWithForms;

    public function render()
    {
        return view('livewire.form-date-comparisons-test');
    }

    public $start_date;
    public $end_date;
    public $dates;

    public function getFormSchema(): array
    {
        return [
            // Works nice
            DatePicker::make('start_date')->required(),
            DatePicker::make('end_date')->afterOrEqual('start_date')->required(),

            // Doesn't work
            Repeater::make('dates')->schema([
                DatePicker::make('repeater_start_date')->required(),
                DatePicker::make('repeater_end_date')->afterOrEqual('start_date')->required(),
            ]),
        ];
    }

    public function save()
    {
        $this->form->getState();
    }

}
```
```blade
<div>
    {{ $this->form }}
    <button wire:click="save">Save</button>
</div>
```

danharrin:
Hi, just gonna ignore that repeater issue for now, just something else to discuss:

I think there's a bit of confusion, the original Laravel validation methods (`after`, `before` etc) do not compare against other fields, just set dates like "tomorrow". Is this PR different? Check https://laravel.com/docs/8.x/validation#rule-after

ralphjsmit:
I see in the docs that they can be used for both, and so does this PR.

The documentation says:

> The field under validation must be a value after a given date. The dates will be passed into the strtotime PHP function in order to be converted to a valid DateTime instance:

PR implementation:
```php
// Works
DatePicker::make('end_date')->afterOrEqual('tomorrow')->required(),
```

The Laravel docs also say:

> Instead of passing a date string to be evaluated by strtotime, you may specify another field to compare against the date:

```php
// Works 
DatePicker::make('start_date')->required(),
DatePicker::make('end_date')->afterOrEqual('start_date')->required(),
```

What I understand from the docs – and see in practice – is that this allows both comparing to dates like 'tomorrow', 'now' etc, and álso comparing to other fields (just like things as `required_if:field`).

danharrin:
Great, thanks

PR #785, opened 2021-12-09
[2.x] Add MultiSelect table filter

dllobell:
This pull request adds a MultiSelect based table filter. It simply renders the MultiSelect form component and it uses the `whereIn` method from the `Builder` to filter the results.

I had to make two little modifications to the MultiSelect form component:

1. I added the flex-wrap class to the list of select options since it breaks when too many are selected
image
image

2. I added the stop modifier to the button click event when deselecting an option, without it for some reason it hides the filter dialog on click.

If this pull request is accepted I would happily add the corresponding docs. 👍


dllobell:
Sure!, I just added relationship support, I believe there is no `whereInRelation` method or something like that in the `Builder` class so I used the `whereHas` method directly. Let me know if something is missing.

danharrin:
This is great, thank you!

PR #790, opened 2021-12-09
Use ISO 15897 standard for languages that differ by territory

dmandrade:
Laravel use ISO 15897 standard for languages that differ by territory. For example, "en_GB" should be used for British English rather than "en-gb".


danharrin:
Thanks :)

PR #793, opened 2021-12-09
Add maxFiles and minFiles rules to MultipleFileUpload

dmandrade:
This PR adds two rules in MultipleFileUpload to limit the maximum and minimum number of files (including the spatie plugin).

I'm currently using this implementation in a project and I believe it will be useful for everyone.


iantasker:
@dmandrade I have rolled support for your PR into #806 so there is a consistent approach. 

Your PR highlighted a couple of areas I had previous missed in my approach, great work!

dmandrade:
@danharrin I will close this PR in favor of #806

PR #794, opened 2021-12-10
Wrap tags

marcreichel:
If there are many tags, so they do not fit into one row, the tags are cut off currently. By using `.flex-wrap` the tags wrap into the next row.


danharrin:
Thanks!

PR #796, opened 2021-12-10
White background

martin-ro:
Add `bg-white` to align with other inputs white background. Not sure if that was left out intentionally for some reason?


danharrin:
Thanks. Not intentional, just hadn't tested the input on a non-white background :)

PR #799, opened 2021-12-10
Make classes configurable

bernhardh:
As discussed yesterday on discord, I have made the classes of the button row on forms and the sidebar configurable. I have wrapped it into `styles` array, so someone could add more class-configs in the future.

With this pull request, a developer can change for example the placement  of the buttons to center:

```
'form_action_button_row' => 'flex flex-wrap items-center gap-4 justify-center'
```

image

and he can for example remove the (in my opinion) weird rounded corners in the sidebar:

```
'sidebar_container' => 'fixed inset-y-0 left-0 z-20 flex flex-col h-screen overflow-hidden shadow-2xl transition duration-300 bg-white lg:border-r w-80 lg:z-0 lg:translate-x-0',
```

image

As default value, I have used your existing classes (so the buttons are left for now and there are still these rounded corners on the sidebar).

What do you think?

### DIFF
--- a/packages/admin/config/filament.php
+++ b/packages/admin/config/filament.php
@@ -177,4 +177,17 @@ return [
         ],
     ],
 
+
+    /*
+    |--------------------------------------------------------------------------
+    | Default classes
+    |--------------------------------------------------------------------------
+    |
+    | You may customise some parts of filaments default classes
+    |
+    */
+
+    'styles' => [
+        'form_action_buttons' => 'flex flex-wrap items-center gap-4'
+    ]
 ];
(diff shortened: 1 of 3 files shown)


danharrin:
Hey, I've gone ahead and removed the sidebar rounding in the core.

I'm not sure this is the best solution as then classes get separated into config files from their views, and when the config file is published we don't have control over those classes anymore.

If you want to add a more specific config option for form button alignment, go ahead. But we should keep the options to `left` (default) or `right`. Maybe `filament.layout.forms.actions.alignment`? Make sure to add a new `align` attribute to the `x-filament::actions` component so it doesn't get applied everywhere else, only the forms.

bernhardh:
Thank you!

Obiously there should be a centered option as well. I will create a pull request with the options `left` (default), `center` and `right`.

PR #804, opened 2021-12-10
Disable registering certain resources to navigation

sawirricardo:
Currently, the way to disable resources is to add this
```
public static  function registerNavigationItems(): void
    {
        return;
    }
```
ref: https://discord.com/channels/883083792112300104/883085267383226478/918873716782403615

However, as discussed in Discord, ideally we can use static property directly.

### DIFF
--- a/packages/admin/docs/02-resources.md
+++ b/packages/admin/docs/02-resources.md
@@ -132,9 +132,9 @@ php artisan make:filament-has-many CategoryResource posts title
 php artisan make:filament-morph-many CategoryResource posts title
 ```
 
--   `CategoryResource` is the name of the resource class for the parent model.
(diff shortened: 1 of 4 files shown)


danharrin:
Hey, this looks good. Maybe add this to the Resources docs though, as the Navigation docs should be solely about the actual Navigation API IMO. Thoughts?

sawirricardo:
> Hey, this looks good. Maybe add this to the Resources docs though, as the Navigation docs should be solely about the actual Navigation API IMO. Thoughts?

Agree @danharrin

danharrin:
Thanks! I've just added the same API to custom pages, so moved the docs back to the Navigation section :)

sawirricardo:
🙏🏻

PR #807, opened 2021-12-10
Fix typo for Tailwind CSS init command

ericp-mrel:
The command should use `npx` instead of `npm` as documented in the official Tailwind CSS docs: https://tailwindcss.com/docs/installation


danharrin:
Thanks!

PR #810, opened 2021-12-10
Use a temporary url for private uploads

martin-ro:
If we upload a private file to S3 we should use a signed temporary URL for the preview to work.

See: https://github.com/laravel-filament/filament/blob/fca0ea26c1166ce405c5e8eda12dea85a3900282/packages/forms/src/Components/FileUpload.php#L471


danharrin:
Thanks!

PR #811, opened 2021-12-10
 Allow to display a conversion

martin-ro:
Pass a conversion name along (e.g. 'thumb') to display a converted image rather then original.

- if no conversion provided: return empty string (will display the original image)
- if conversion doesn't exist on collection: return empty string (will display the original image)
- if conversion exist: profit


danharrin:
Great, thanks :)

PR #812, opened 2021-12-10
Add widget support for pages.

dmandrade:
Add support to display Widgets in any filament page below the header and above footer.

Also update MakeWidgetCommand to allow create  a widget specifically for a Resources (like page command)

`artisan make:filament-widget CustomerOverview --resource=CustomerResource`

This is a solution to create widgets only for a resource page and dont display it on dashboard.

Maybe we need an option in the widget to define which ones will be automatically loaded in dashboard.


danharrin:
Thanks!

PR #817, opened 2021-12-11
Use slug syntax for dayjs locale files

dmandrade:
There was a small bug causing wrong locale url for DayJS.

Wrong url:
https://unpkg.com/dayjs@1.10.4/locale/pt_br.js

the correct is:
https://unpkg.com/dayjs@1.10.4/locale/pt-br.js

### DIFF
--- a/packages/forms/resources/views/components/date-time-picker.blade.php
+++ b/packages/forms/resources/views/components/date-time-picker.blade.php
@@ -7,7 +7,7 @@
 
             window.dayjs_locale = dayjs.locale()
         </script>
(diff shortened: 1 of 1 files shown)


lolen:
One more line should be added:
`dayjs.locale ('$lang')`
But that doesn't solve the next problem. Month names are retrieved with `dayjs.months()`, but the days of the week `getDayLabels()`, which are unaffected by the set location.

danharrin:
As per #818, should we not use that approach here too?

PR #818, opened 2021-12-11
Keep a consistent approach using locale in lang tag

dmandrade:
Update docs and scaffolding  to use same approach for setting locale in lang tag.


danharrin:
Thank you :)

PR #823, opened 2021-12-12
ADD: Make form button alignment configurable

bernhardh:
Hello, 

as discussed in #799, here the new version of it. Now only `left`, `center` and `right` is possible and I used the (long?) config key `filament.layout.forms.actions.alignment` as suggested.

Since `justify-end` wasn't in the css yet, I had to recompile filament. Not sure if this was the right way to do it.


danharrin:
Thank you!

PR #824, opened 2021-12-12
Polish translation

lolen:
This is a simplified version of the Polish translation because this language contains many pluralisms, for example
| :count  | Select all :count records |
| ------------- | ------------- |
| 2-4  | Zaznacz wszystkie :count **rekordy**  |
| 5+  | Zaznacz wszystkie :count **rekordów**  |


danharrin:
Thanks!

PR #825, opened 2021-12-12
Makes ChartWidget type Dynamic

bezhanSalleh:
Fixes the chart widget type which is hard coded as `line`.


danharrin:
Thank you so much!

PR #838, opened 2021-12-13
Simplified Chinese Translation

mugennsou:
Hey, filament team, Thanks for your work.

This PR is Simplified Chinese translation for admin panel package & table package.


danharrin:
Thank you!

PR #842, opened 2021-12-13
Dashboard label translation and icon customization

Zagna:
Currently title is translated.
Should the label also use filament::pages/dashboard.title?
Or separate for the option of them being different?

### DIFF
--- a/packages/admin/config/filament.php
+++ b/packages/admin/config/filament.php
@@ -73,6 +73,9 @@ return [
         'register' => [
             Pages\Dashboard::class,
         ],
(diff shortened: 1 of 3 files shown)


danharrin:
Hi, thanks for this.

I'm afraid that I'm gonna have to decline this one. You can do this by creating your own `Dashboard` page that extends the original `Dashboard` class, and changing these navigation options. I just don't believe that always adding config options, which can get out of date, is our best bet.

Zagna:
That indeed worked.

PR #846, opened 2021-12-14
fix: Badge columns unexpectedly ignoring falsy values

capibawa:
Badge columns will currently ignore any falsy values such as 0 or '0'.

My proposed solution is to explicitly check that the formatted state is null.


danharrin:
Thanks! Switched to `blank()` which also fails empty strings as well as `null` :)

PR #865, opened 2021-12-14
Update CanRequireConfirmation.php

pokmot:
This was forced to be true rather than evaluate the condition sent through


danharrin:
Good spot, thanks

Issue #877, opened 2021-12-15
Table sorting icons


telkins:
### Package

filament/filament

### Package Version

v2.5.26

### Laravel Version

v8.76.1

### Livewire Version

v2.8.2

### PHP Version

PHP 8.1.0

### Bug description

Not 100% sure if this is a bug, but the icons that are show when sorting on a column seem to be backward.

When sorting small-to-big (ascending), we get the following:
![image](https://user-images.githubusercontent.com/53731/146188789-a7d8f289-9470-4ef9-b440-f92bc0afc5cc.png)
and the query that's executed orders `asc`.

When sorting big-to-small (descending), we get the following:
![image](https://user-images.githubusercontent.com/53731/146189084-f0c987a0-4dcb-4106-8859-35b894ed0795.png)
and the query that's executed orders `desc`.

This seems backward to me.  I can't find anything in the docs, issues, PRs, or general search that helps.  It's possible that I've missed something or I'm simply wrong in my expectations.

Finally, I use Laravel DebugBar to view the query that's run, fwiw.

### Steps to reproduce

Create a resource and make a column sortable.  Compare expectations to "indicator".

### Relevant log output

```shell
n/a
```

danharrin:
Ha, you're right, good spot. Mind fixing in a PR?

shuvroroy:
@telkins and @danharrin I just create a [PR](https://github.com/laravel-filament/filament/pull/882) regarding this

telkins:
> @telkins and @danharrin I just create a [PR](https://github.com/laravel-filament/filament/pull/882) regarding this

LOL!  Looks like you beat me to it by about 8 minutes.  😅 

In any case, it's fixed. 🤓

PR #883, opened 2021-12-15
Fix: Simple switch of sort icons to correctly reflect sort direction

telkins:


telkins:
Related to #877.

danharrin:
Thanks, @shuvroroy beat you ;)

PR #884, opened 2021-12-15
Add selected records indicator

Majkie:
Add indicator with selected records count and quick actions to select or deselect all records.

Screenshot 2021-12-15 at 15 19 31

Also added deselect all action to bulk actions where currently is only a select all action.

Both indicator/bulk select actions can be optionally disabled.


danharrin:
Thanks for this! Instead of specifying an arbitrary colspan="10", we should work out how to pass the actual number of columns to this component.

Majkie:
> Thanks for this! Instead of specifying an arbitrary `colspan="10"`, we should work out how to pass the actual number of columns to this component.

Yea, I kind of forgot on this placeholder. Now it should be pretty much okay.

danharrin:
Thanks. I've updated this to be default behaviour and removed the bulk action. The maintenance burden is high for two features that do the same thing.

Screenshot 2021-12-17 at 20 03 46

PR #885, opened 2021-12-15
Add fully support for rtl languages

maherelgamil:
- fix issues with rtl languages by using tailwindcss rtl modifiers.

Screen Shot 2021-12-15 at 5 56 27 PM


danharrin:
Thank you so much!

PR #891, opened 2021-12-15
fully localized resources phrases

maherelgamil:
Developers are now able to add direct translations for resources

for example :

if your resource names `UserResource`

just add your translations to your lang file ex. `ar.json`

```json
{
    "Users": "المستخدمين",
    "User": "مستخدم" 
}
```

So, the result reflects in all admin panel

image

image

image

image


danharrin:
Hey, not gonna merge this in case it results in people having strings automatically translated that they don't want to be translated (this was feedback from v1). You can use translation strings in resources by overriding resource methods instead of properties, for example `getPluralLabel()` or `getNavigationLabel()`.

maherelgamil:
@danharrin okay, no problem.

PR #902, opened 2021-12-16
fix: use exists property

fabpl:
Using ->exists() on eloquent model execute query on each call.
Replace by property to use cached value.


danharrin:
Thanks!

PR #904, opened 2021-12-16
Fix some RTL issues

maherelgamil:


danharrin:
Hi, thanks for these improvements. We noticed that they actually break the standalone forms and tables packages, as they now require a `dir="ltr"` on the `<html>` tag. I think we need to remove the `ltr:` variants and just keep the `rtl:` as overrides. I'll make these changes. Thanks!

PR #907, opened 2021-12-16
feature: support hiding filament logo

ryangjchandler:
Supports hiding the Filament logo in the footer via a config variable / environment variable.


wychoong:
i think footer with filament logo should be a separate view or the checking of the config should be inside of the view. by doing the ifelse at app.blade.php break the possibility of publishing footer blade and customize over there

danharrin:
Agreed @wychoong, I made this change.

PR #908, opened 2021-12-17
Feature: support remembering per page

mansoorkhan96:
Supports remembering the default per page option on the table via a config variable.

Issue:
If a user selects the per page option on a table and reloads the page, it would not remember the previously selected option. Hence the user is forced to select the per page option on each page load for a table.

Solution:
This pull request solves the above issue by optionally storing the per page option in the cache.


wychoong:
i might be wrong but the caching of rememberpage seems applying for every user, as its using global cache key

mansoorkhan96:
@wychoong Good catch. I should make the key specific to current authenticated user, right?
Let me push another commit.
Thanks for reviewing.

wychoong:
> @wychoong Good catch. I should make the key specific to the currently authenticated user, right?
> Let me push another commit.
> Thanks for reviewing.

Not sure would this be a good addition as it breaks the possibility of using table package without auth. Maybe Auth::check() before prepend id would be better

danharrin:
Hey, check out [sessions](https://laravel.com/docs/8.x/session). We don't need to use caching and user account keys here.

mansoorkhan96:
> Hey, check out [sessions](https://laravel.com/docs/8.x/session). We don't need to use caching and user account keys here.

ok cool. I will remove Auth specific things and replace cache with sessions.
Thanks.

mansoorkhan96:
> Hey, check out [sessions](https://laravel.com/docs/8.x/session). We don't need to use caching and user account keys here.

Hi, I have updated the pull request with new commits

danharrin:
Thanks! I've made this default behaviour.

mansoorkhan96:
> Thanks! I've made this default behaviour.

You are welcome :)
Thanks for merging the pr. I have learned a few nice things from your refactor.

PR #922, opened 2021-12-18
ADD: Allow extra attributes to table columns

bernhardh:
This pull request is to allow extra attributes - especially classes - to table columns. 

Usecase: If you link a column with `url` method right now, nothing shows you, that the cell is linked (except the cursor). With this commit you can style it like a link:

```php
Tables\Columns\TextColumn::make("owner.name")
    ->withAttributes(["class" => "text-green-500"])
    ->url(fn ($record): string => route("...")),
```


danharrin:
Thanks! I had to add the `extraAttributes()` method to columns.

PR #927, opened 2021-12-19
Not working when used other directory structure

aymenel:
If i use other directory structure livewire cant find components.


danharrin:
Great, thanks!

PR #875, opened 2021-12-15
Merge extraAttributes with DatePicker button 

daggemarren:
Merge extraAttributes with DatePicker button  so you can modify roundness and paddings etc.

PR #935, opened 2021-12-19
Enhanced extra attributes

danharrin:
All `extraAttributes()` are now always merged onto the outer HTML element of the component. This was previously VERY inconsistent between fields, causing unexpected merging behaviour.

I've also introduced a new `extraAlpineAttributes()` method - which always merges attributes onto the root Alpine element (the one containing `x-data`).

Additionally, I've added the first "dedicated" attribute bag - for merging onto the Date Picker's trigger button. You may use `extraTriggerAttributes()` method for this. See #875.


ekandreas:
Since this update there is no longer possible for us to change the styling of inputs in a form. We will not use rounded attributes and need to remove them from the forms via extraAttributes. Any help would be nice, thanks!
Eg, 
```
TextInput::make('sort_order')
            ->label(__('groups.sort_order'))
            ->required()
            ->numeric()
            ->extraAttributes([
                'class' => 'border rounded-none pl-2',
            ]);
```

danharrin:
Hey, we should add `extraInputAttributes()` to the text input component :) Do you feel comfortable PRing (check out `extraAlpineAttributes()` for inspiration), or should I?

ekandreas:
It would be fun for me to contribute but perhaps it will take a couple of days for me. I will try.
And once again, thanks for magic packages, so much fun.
Any discord channel connected to this lib?

danharrin:
Yup! Join us on https://filamentadmin.com/discord.

danharrin:
> It would be fun for me to contribute but perhaps it will take a couple of days for me. I will try.
> And once again, thanks for magic packages, so much fun.
> Any discord channel connected to this lib?

I've implemented `extraInputAttributes()` in #1039 :)

PR #949, opened 2021-12-20
fix select input loading spinner display

achetibi:
The spinner is not showing correctly in the select input when searching the data.


danharrin:
Thanks!

PR #953, opened 2021-12-21
Unique Chart JS Variable

MazzMazz:
This MR Provides a possible solution for chart.js rerender bug.

When having multiple charts within one page and triggering a rerendering with 'updateChartData' - only the last Chart is update. It seems like the chart js variable gets overwritten if multiple charts are on one page and only one chart gets updated (instead of every chart)

```
<select class="float-right" wire:model="selectedChartOption" wire:change="updateChartData">
```

The images explain the problem. While the left chart should have been updated, the right chart gets rerendered with the Data of the left chart.

image
image


ryangjchandler:
Hm, I think there's likely a better solution for this. 

I'll have to take a look later on tonight / tomorrow.

danharrin:
Thanks @ryangjchandler, just mark this as approved when you're finished.

danharrin:
Uh, I did not mean to merge that. I will revert it now.

PR #960, opened 2021-12-22
RTL Enhancements

mohamedsabil83:
In general, added missing `ltr:` to prevent those classes related to LTR direction from applying to the RTL one


danharrin:
Hey, we can't use `ltr` variants as those rely on users of the forms & tables packages having a `dir="ltr"` on their `<html>`. That's a breaking change. What issues are you facing? The `rtl:` version should always override the normal LTR classes.

mohamedsabil83:
Ok, instead of using **ltr:**, I'll remove its effects by using **rtl:**.
I'll push a new PR.

PR #979, opened 2021-12-23
changed the color of placeholder to gray-500

felagund18:
Placeholder color is changed to `text-gray-500` and added `select-none`.

Old:
Screen Shot 2021-12-23 at 14 09 56

New:
Screen Shot 2021-12-23 at 14 08 16


danharrin:
Hey, we deliberately didn't have this placeholder color set as it was inconsistent with non-searchable select inputs. Since we use an empty `<option>` for the placeholder, it's difficult to consistently style the placeholder to be the correct color without affecting the color of the other options. We decided that it was better to have a consistent placeholder color across both select inputs.

PR #983, opened 2021-12-23
Fix syntax error in toggle field example

iDiegoNL:
When copy-pasting the `Toggle` on/off icon snippet, I noticed that there's a comma in a place it doesn't belong, causing a syntax error.

This PR fixes that :)


danharrin:
Thanks!

PR #994, opened 2021-12-24
Add MonthPicker based on date-Time Picker

aliloubm:
Is not clean but it works
Capture Maxthon20211224162905
forms/src/Components/MonthPicker.php is not optimized.

we need to move @stack('scripts') to <head> in base.blade  without that month() work only in English.


danharrin:
Hi, thanks for the contribution! I'm not gonna accept this new component into the core of Filament, as it will add a maintenance burden for us when this is likely not a feature that alot of people will benefit from. Please release it as a third party plugin and we will advertise it on our upcoming marketplace :)

PR #995, opened 2021-12-24
feature: Integrated multiple file uploads

danharrin:
`FileUpload::make('files')->multiple()`

This also allows you to drag / select multiple files to upload at once, unlike `MultipleFileUpload` which forces you to do this one at a time.

- [x] Write docs.
- [x] Remove `MultipleFileUpload` from docs.
- [x] Generate UUID in JS instead of random `fXXXX` number.
- [x] Integrate with Spatie media library.
- [x] Ensure no breaking changes to `MultipleFileUpload`.
- [x] Validation, including `minFiles()` and `maxFiles()`.


robertorinaldi-dev:
I saw it quickly, but how did you fix the revert problem?

danharrin:
> I saw it quickly, but how did you fix the revert problem?

Not sure exactly what you were doing wrong :)

robertorinaldi-dev:
Are you sure that if you add a file and then delete it (without saving the form) the temporary file is removed from livewire-tmp? This thing was not working for me. I also noticed strange behaviors saving the form after several revert actions.
I tried to call $wire.remove but i found that bug (?) i sent you on discord, i.e livewire which checks only the first element of the array while the newly added files are at the end.
However i haven't tried with these changes yet so i will let you know

danharrin:
> Are you sure that if you add a file and then delete it (without saving the form) the temporary file is removed from livewire-tmp? This thing was not working for me. I also noticed strange behaviors saving the form after several revert actions.
> I tried to call $wire.remove but i found that bug (?) i sent you on discord, i.e livewire which checks only the first element of the array while the newly added files are at the end.
> However i haven't tried with these changes yet so i will let you know

Yup, works fine :) The thing about the first element might be because you were using numeric keys instead of string keys? So Livewire might have been dehydrating the files as a multi-upload instead of single uploads.

robertorinaldi-dev:
Since you go very fast, I'll add my wish list 😁
- [ ] sortable files (at least for spatie https://spatie.be/docs/laravel-medialibrary/v9/advanced-usage/ordering-media)
- [ ] image editing with some free library (https://pqina.nl/filepond/docs/api/plugins/image-edit/)
- [ ] settable file name, alt and/or title

Thanks!

shuvroroy:
@danharrin @robertorinaldi-dev and @ryangjchandler  In my opinion `BaseFileUpload` class only have properties and method that requires so that others can extends this class and use it. We can make it much more cleaner by moving some of the logic from `BaseFileUpload` to `FileUpload` class.

robertorinaldi-dev:
I agree with the first part but what logic are you referring to? Everything in BaseFileUpload seems to me related to a normal upload. Which method or property would you move to FileUpload?

danharrin:
> Since you go very fast, I'll add my wish list 😁
> - [ ] sortable files (at least for spatie https://spatie.be/docs/laravel-medialibrary/v9/advanced-usage/ordering-media)
> - [ ] image editing with some free library (https://pqina.nl/filepond/docs/api/plugins/image-edit/)
> - [ ] settable file name, alt and/or title
> 
> Thanks!

Once this is merged, feel free to create PRs for these. Not my priority right now though :)

shuvroroy:
> I agree with the first part but what logic are you referring to? Everything in BaseFileUpload seems to me related to a normal upload. Which method or property would you move to FileUpload?

Properties: 
`isMultiple`
Methods:
`multiple`

danharrin:
> > I agree with the first part but what logic are you referring to? Everything in BaseFileUpload seems to me related to a normal upload. Which method or property would you move to FileUpload?
> 
> Properties: `isMultiple` Methods: `multiple`

IMO, those are core features of the file upload now :)

robertorinaldi-dev:
Hi @danharrin , I'm trying now, but unfortunately revert is not working for me.
It calls `removeUploadedFile` that removes the file from the state, but I don't see any deletion from livewire-tmp.
This is not a major problem, but in the long run the folder will fill up with images.
Actually the temporary file is not deleted even in case of saving, so I don't know if this is a desired behavior; previously you said that the temporary file is removed from livewire-tmp for you, so I guess not.

robertorinaldi-dev:
As for the revert, with the new configuration it shouldn't be complicated; I think adding `$file->delete();` to `removeUploadedFile` should be enough

danharrin:
Thanks for the PR. For files that aren't reverted, Livewire should clean up temporary uploads after 24 hours. Check out https://laravel-livewire.com/docs/2.x/file-uploads#auto-cleanup

robertorinaldi-dev:
Ah, good to know! :)

robertorinaldi-dev:
I keep writing in here, being on topic ... may I ask you why you used removeUploadedFileUsing in revert, instead of changing the state directly? (I'm working on sorting)

Thanks

danharrin:
We use that method as then the PHP behaviour can be customised for reverting files (`removeUploadedFileUsing(closure)`)

PR #1001, opened 2021-12-25
Allow overwrite forms for views

lolen:
I wanted to be able to overwrite the form for each view (create, edit, view). The documentation does not contain information about this. I found the simplest solution that may not be correct. And I decided to share it. Now I can implement a method in each view that will overwrite the default form. For example:
```
class ViewCustomer extends ViewRecord
{
    protected static string $resource = CustomerResource::class;

    public static function customForm(Form $form): Form
    {
        return $form
            ->schema([]);
    }
}
```


danharrin:
Hey, I wasn't able to edit this PR for some reason so I opened a new one - #1002.

I've made some changes - I've cleaned up the code by using this method to fetch the resource table by default. It's been renamed to `form()`, and made not-static so that you can customize the form easier based on the current Livewire component / record.

PR #1004, opened 2021-12-26
Allow mutate record data before fill form

lolen:
By overloading the `mutateRecordDataBeforeFill` method, we can format the data with which the form will be filled.


danharrin:
Thanks!

PR #1034, opened 2021-12-28
SpatieMediaLibraryFileUpload - Reorder files

robertorinaldi-dev:
I made a draft PR because I have doubts about some things and I am not sure you are ok with how I implemented it.

With `allowReorder()` it activate the corresponding property on FilePond.
On every reorder action it calls `reorderFiles()` which updates the state.
On save it calls `reorderFilesUsing()` which is preconfigured for spatie but can be customized.

One problem I encountered is that the files are currently appended to the last position in the state, but they appeare as first because this is the automatic FilePond configuration; the consequence of this is that after saving and refreshing the page, the order of the files changes from how it had been left.
For this reason I have added `appendFiles()` which sets whether files should be inserted 'after' and reversed the files in file-upload.js by default.

I'm not very sure about the choice to put all these functions and properties in BaseFileUpload, except for `appendFiles()` which seemed to me a more specific FilePond related thing. I'm also not very convinced by the choice of names.

Let me know what you think (when you'll have time, I know this is not a priority for you).

Thanks


danharrin:
Thanks for your work on this. I've definitely not forgotten about this, but not a priority for the project right now. Thank you for your patience.

robertorinaldi-dev:
No problem, I'm already using it anyway 😆

packages/forms/src/Components/FileUpload.php
```diff
@@ -34,6 +36,13 @@ class FileUpload extends BaseFileUpload
 
     protected string | Closure $uploadProgressIndicatorPosition = 'right';
 
+    public function appendFiles(bool | Closure $appendFiles = true): static
```
danharrin, line 39: Please rename `$appendFiles` to `$condition`.

packages/forms/src/Components/FileUpload.php
```diff
@@ -137,6 +146,11 @@ public function uploadProgressIndicatorPosition(string | Closure | null $positio
         return $this;
     }
 
+    public function getAppendFiles(): bool
```
danharrin, line 149: This should be `shouldAppendFiles()`, and `$this->appendFiles` should be `$this->shouldAppendFiles`

packages/forms/src/Components/BaseFileUpload.php
```diff
@@ -234,6 +252,11 @@ public function getAcceptedFileTypes(): ?array
         return $types;
     }
 
+    public function getAllowReorder(): bool
```
danharrin, line 255: We can rename this to `canReorder()`, and the property can be renamed from `$this->allowReorder` to `$this->canReorder`

packages/forms/src/Components/BaseFileUpload.php
```diff
@@ -127,6 +131,13 @@ public function acceptedFileTypes(array | Arrayable | Closure $types): static
         return $this;
     }
 
+    public function allowReorder(bool | Closure $allowReorder = true): static
```
danharrin, line 134: Please rename this to `enableReordering()`, and the argument should be `$condition`.

danharrin:
Does this work for a mixture of already-uploaded and newly-uploaded files?

robertorinaldi-dev:
Yes, in this part
```.​map​(​file​ ​=>​ ​file​.​source​ ​instanceof​ ​File​ ? ​file​.​serverId​ : ​this​.​uploadedFileUrlIndex​[​file​.​source​]​ ​??​ ​null​)​ 
```
The only problem I have encountered is if I reorder a file whose upload hasn't even started yet, because serverId is still null; in this case the order of the file is simply ignored.
I couldn't find a way to handle this, maybe the only solution would be to temporarily block the reorder, but I have doubts that it could be a viable solution

saade:
👀 needing this for my next project, to be able to reorder carousel images

danharrin:
Thank you so much! This is great. Also, I fixed an n+1 problem and added docs.

PR #1039, opened 2021-12-28
feature: Merge extra input attributes

danharrin:


mokhosh:
`RichEditor` doesn't have this?
Is there any way currently I can tell `trix` to use `ring-primary-400` instead of `ring-primary-500`?

danharrin:
Its being introduced gradually as people need it, feel free to PR it in for the rich editor.

PR #1041, opened 2021-12-28
Set default locale when all translatable fields are blank

quintenbuis:
When all fields in the `$translatable` property on a model are empty it throws an `undefined offset 0` error. 

When this happens it will now use the default locale so the form still loads and can be filled.


danharrin:
Thanks!

PR #1059, opened 2021-12-29
asHtml method added to placeholders

ekandreas:
To gain the possible render as Html to content in a placeholder the method asHtml is added and the default behaviour (non html) is preserved.


danharrin:
Hey, you can do this already by passing an `HtmlString` into the `content()`:

```php
use Illuminate\Support\HtmlString;

Placeholder::make('Label')
    ->content(new HtmlString('<h2>Heading<h2>This is rendered as <span class="font-bold">HTML</span'))
```

I will add this to the docs now :)

PR #1060, opened 2021-12-29
Possible hide of the select placeholder option

ekandreas:
I would like to hide the option serving as a placeholder for the Select component via a new method noPlaceholder added.


danharrin:
Hi, thanks for this. For some reason, I don't have permission to make tweaks to this PR - so I've pushed it to a new PR here - #1065.

I've renamed the method to `disablePlaceholderSelection()` - since `noPlaceholder()` implies that the placeholder is removed also for searchable selects, which is not true.

ekandreas:
Great!
Thanks!

PR #1071, opened 2021-12-30
Add support for enum casts in PHP 8.1

ralphjsmit:
This PR adds support for enums/casts in PHP 8.1. 

I encountered this issue when working with an Eloquent model that has an enum cast. In the below example, the `$record->status_type` is cast as an enum, leading to errors about an illegal offset type. (It was trying to use an enum as an array key.)
```php
BadgeColumn::make('status_type')->label('Status')->enum((array) StatusType::labels()),
```

This PR adds a simple check for the cases where we have an enum that's backed by a string. In that situation it will use the backed value.

```php
if ($this->formatStateUsing instanceof Closure) {
    $state = app()->call($this->formatStateUsing, [
        'livewire' => $this->getLivewire(),
        'record' => $this->getRecord(),
        'state' => (interface_exists(\BackedEnum::class) && $state instanceof \BackedEnum::class) ? $state->value : $state,
    ]);
}
```

I tested it and this implementation also works on PHP 8.0.

### DIFF
--- a/packages/tables/src/Columns/Concerns/CanFormatState.php
+++ b/packages/tables/src/Columns/Concerns/CanFormatState.php
@@ -74,7 +74,7 @@ trait CanFormatState
             $state = app()->call($this->formatStateUsing, [
                 'livewire' => $this->getLivewire(),
                 'record' => $this->getRecord(),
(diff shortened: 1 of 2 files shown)


danharrin:
Thanks :)

PR #1075, opened 2021-12-30
fixed some translations in Kurdish locale

Xoshbin:
fixed some translations in Kurdish locale, Because of the plural issue we can't use the label variable in some places. it makes  wrong translations


danharrin:
Thanks

PR #1121, opened 2022-01-04
fix typo

maherelgamil:


ryangjchandler:
@maherelgamil You don't actually need the quotes around the version constraint.

maherelgamil:
@ryangjchandler without qouts, we have issue when copy this line directly to terminal

danharrin:
I can't replicate this, and all guides online use the example without quotes. I'm going to close this - let me know if there is another reason why we should consider merging this.

ryangjchandler:
Yeah, I use zsh and can't replicate either.

maherelgamil:
@danharrin @ryangjchandler okay, no problem

PR #1124, opened 2022-01-04
Fix input flickering problem

ralphjsmit:
This PR fixes a flickering problem I noticed on Safari, which also seems to appear on Chrome. (Other browsers not tested.)

It basically comes down to the fact that an `<input>` or `<select>` element flickers when it only has a `transition-duration` specified and a user selects it. If you add the Tailwind `transition` property, everything works fine. See the below short screencasts.

**[Before (it flickers on select)](https://stream.new/v/7HufSaAwVOdzV1LIkqpDW2VIiT7d4vWco1zM6VCnbZU)**
**[After (no flickering, it goes smooth)](https://stream.new/v/ZUCmfg3r2icVWAn6bY9QrXuhvKQX8fN4cEfEoSgV02Vo)**

I didn't test this yet on the admin panel, but perhaps we should add the `transition` property there as well.

### DIFF
--- a/packages/admin/resources/views/components/actions/select-action.blade.php
+++ b/packages/admin/resources/views/components/actions/select-action.blade.php
@@ -7,7 +7,7 @@
         id="{{ $getId() }}"
         wire:model="{{ $getName() }}"
         {{ $attributes->class([
-            'text-gray-900 border-gray-300 invalid:text-gray-400 block w-full h-9 py-1 duration-75 rounded-lg shadow-sm focus:border-primary-600 focus:ring-1 focus:ring-inset focus:ring-primary-600',
+            'text-gray-900 border-gray-300 invalid:text-gray-400 block w-full h-9 py-1 transition duration-75 rounded-lg shadow-sm focus:border-primary-600 focus:ring-1 focus:ring-inset focus:ring-primary-600',
         ]) }}
     >
         @if (($placeholder = $getPlaceholder()) !== null)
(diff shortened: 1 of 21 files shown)


danharrin:
Thanks, if you could add these classes everywhere else too that would be great.

ryangjchandler:
@danharrin Can't see this being a problem in admin, I'm not able to reproduce it at least.

ralphjsmit:
@ryangjchandler I'll try to double-check it in an admin panel this afternoon, I've been mostly using the forms and tables package. 

Think that it doesn't harm to have those classes in the admin panel anyway...?

ryangjchandler:
@ralphjsmit The admin package uses the tables & forms package under the hood so it might be fine with these changes.

ralphjsmit:
@ryangjchandler That's true, but when you do a search on the whole project for "duration-", then you see that there are also a few occurrences in the admin panel files that have a duration without a `transition` class.

ralphjsmit:
I added the transition class on all the other places I could find in the project.

ryangjchandler:
@ralphjsmit Awesome, thanks!

ryangjchandler (approved):

danharrin:
While I can't replicate these issues, the classes don't hurt. Thanks!

ralphjsmit:
Thank you both as well! 🙌

ralphjsmit:
@wychoong Looks this is indeed the reverse. Personally I don't see the point of removing those classes..? ([See](https://stackoverflow.com/questions/36962193/whats-causing-my-select-box-to-flicker-when-selected))

ryangjchandler:
@ralphjsmit Tailwind 3.x no longer requires those classes to make transitions work but it's clear that something else was missing, perhaps a vendor prefix or similar.

PR #1125, opened 2022-01-04
feature: configuring and configured hooks for PluginServiceProvider

ryangjchandler:
Adds 2 new hooks for the `PluginServiceProvider`:

* `configuringPackage`
* `configuredPackage`

These can be used by plugin developers to easily register migrations, etc with the `PackageServiceProvider` that we extend. At the minute, the only way to do this is overriding the `configurePackage` method and calling `parent::configurePackage` which could cause problems if people forget to call it, just like the old `parent::boot()` problems from `Model` files :)

### DIFF
--- a/packages/admin/src/PluginServiceProvider.php
+++ b/packages/admin/src/PluginServiceProvider.php
@@ -22,19 +22,9 @@ abstract class PluginServiceProvider extends PackageServiceProvider
 
     protected array $widgets = [];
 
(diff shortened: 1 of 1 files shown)


danharrin:
I've renamed these methods to match the names of the others - `packageConfigured()` instead of `configuredPackage()` - like `packageRegistered()`.

PR #1135, opened 2022-01-04
adding new line after form/table components

maherelgamil:
adding new line after form/table components to make it more readable when use generator


danharrin:
Thanks for this. At the moment, all of our docs examples & our demo don't format fields this way, it is a bit of a personal preference. If I see the community widely shift towards putting new lines in the schema, I'll merge this. But at the moment - it would just mean that people's code is inconsistent.

PR #1136, opened 2022-01-04
fix: Clearing Filepond's state

### LINKED ISSUE #1133: FileUpload field not clearing state
### Package

filament/forms

### Package Version

2.8

### Laravel Version

8.78.0

### Livewire Version

2.8.2

### PHP Version

8.1.0

### Bug description

The FileUpload field is not clearing Filepond's state after saving. 

### Steps to reproduce

- Register a simple FileUpload field in Filament admin: `Forms\Components\FileUpload::make('test')`
- Click "Create & add another"
- Filepond state is not cleared and still showing the selected files

This check is causing the issue:
https://github.com/laravel-filament/filament/blob/2.x/packages/forms/resources/js/components/file-upload.js#L129-L131

### Relevant log output

_No response_

pxlrbt:
Closes #1133

### DIFF
--- a/packages/forms/resources/js/components/file-upload.js
+++ b/packages/forms/resources/js/components/file-upload.js
@@ -48,6 +48,8 @@ export default (Alpine) => {
 
             pond: null,
 
+            shouldUpdateState: true,
+
             state,
 
             init: async function () {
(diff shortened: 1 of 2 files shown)


danharrin:
Thank you so much for this!

packages/forms/resources/views/components/file-upload.blade.php
```diff
@@ -35,7 +35,7 @@
             uploadProgressIndicatorPosition: '{{ $getUploadProgressIndicatorPosition() }}',
             uploadUsing: async (fileKey, file, success, error, progress) => {
                 $wire.upload(`{{ $getStatePath() }}.${fileKey}`, file, () => {
-                    success()
+                    success(fileKey)
```
pxlrbt, line 38: Is it really necessary to pass `fileKey` from the component when we don't modify it?

packages/forms/resources/js/components/file-upload.js
```diff
@@ -105,8 +105,9 @@ export default (Alpine) => {
                                 (c ^ crypto.getRandomValues(new Uint8Array(1))[0] & 15 >> c / 4).toString(16)
                             )
 
-                            uploadUsing(fileKey, file, () => {
+                            uploadUsing(fileKey, file, (fileKey) => {
```
pxlrbt, line 108: We already have `fileKey` from outside the function, right?

pxlrbt (commented):
@danharrin Not sure whether these 2 changes were necessary.

packages/forms/resources/views/components/file-upload.blade.php
danharrin: It's not necessary in our implementation, but if someone has a custom view or JS then this could be useful to them. It doesn't hurt to pass the key anyway :)

PR #1148, opened 2022-01-05
fix: HasAction: Pass all data to actions created from string

pxlrbt:
As mentioned in Discord: Actions created from a string received different data than callables.
https://discord.com/channels/883083792112300104/883085267383226478/928258984081784892

### DIFF
--- a/packages/tables/src/Actions/BulkAction.php
+++ b/packages/tables/src/Actions/BulkAction.php
@@ -74,9 +74,7 @@ class BulkAction
         $action = $this->action;
 
         if (is_string($action)) {
-            $action = function (HasTable $livewire, Collection $records) use ($action) {
-                return $livewire->{$action}($records);
-            };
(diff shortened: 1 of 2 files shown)


danharrin:
We should probably do this for bulk actions as well

pxlrbt:
> I might be wrong here, but aren't you using first class callables? Does this work in PHP 8.0?

It's only arrow functions (PHP 7.4) and variable parameter/array unpacking (PHP 5.6).

danharrin:
Ah right fair enough, but does dependency injection still detect those properties correctly on the original closure? You've tested this, right?

pxlrbt:
> Ah right fair enough, but does dependency injection still detect those properties correctly on the original closure? You've tested this, right?

No it required the arguments in the right direction. Since you introduced the `$this->evaluate()` here, we can apparently create a Closure on the fly since it's evaluated at a later point 🤔 Before I got Livewire errors.

danharrin:
Thanks :)

PR #1149, opened 2022-01-05
Notifications outside of page context

pxlrbt:
I was trying to wrap my head around how to get notifications to work with ButtonActions. Feel free to hit me up on Discord if you want to discuss this or let me know what you think.

### DIFF
--- a/packages/admin/src/Pages/Page.php
+++ b/packages/admin/src/Pages/Page.php
@@ -3,6 +3,7 @@
 namespace Filament\Pages;
 
 use Closure;
+use Filament\Concerns\CanNotify;
 use Filament\Facades\Filament;
 use Filament\Navigation\NavigationItem;
 use Illuminate\Contracts\View\View;
@@ -12,6 +13,8 @@ use Livewire\Component;
 
 class Page extends Component
 {
+    use CanNotify;
+
     protected static string $layout = 'filament::components.layouts.app';
 
     protected static ?string $navigationGroup = null;
@@ -29,6 +32,8 @@ class Page extends Component
     protected static ?string $title = null;
 
     protected static string $view;
+
+    protected ?array $notification = null;
 
     public static function registerNavigationItems(): void
     {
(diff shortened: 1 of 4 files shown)


packages/admin/src/Pages/Page.php
```diff
@@ -73,14 +78,6 @@ public static function getUrl(array $parameters = [], bool $absolute = true): st
         return route(static::getRouteName(), $parameters, $absolute);
     }
 
-    protected function notify(string $status, string $message): void
```
ryangjchandler, line 76: Removing this method is a breaking change.

ryangjchandler (changes requested):

ryangjchandler:
I think there might be a better way to do this...

@danharrin Could we not refresh the `Page` component with an event listener instead, i.e. emit an event when an action is finished and then:

```php
protected $listeners = [
    'someActionCompleteEvent' => '$refresh',
];
```

This would then trigger a re-render and the notification will appear.

packages/admin/src/Pages/Page.php
pxlrbt: It extracted into a Trait.

ryangjchandler: Ah, yeah. Had that files collapsed for some reason.

pxlrbt:
> My reasoning behind this is that the Page component shouldn't really hold any state regarding notifications. Having the method to notify() is fine, but actually holding those notifications on the component feels a bit iffy.

Yeah, I felt the same. Ideally we would have a NotificationManager. I'm still quite new to Livewire so I didn't know how to approach this otherwise.

danharrin:
Yeah, I don't think this is the right solution. I think we should have a `NotificationManager` component which we then emit events to to add notifications. This could also support multiple stacked notifications at a time maybe? I'll add this to my low priority todos, but if you have chance I'd appreciate a PR (if you want to attempt this).

PR #1150, opened 2022-01-05
Update Mask.php

MACscr:
->padFractionalZeros() is needed to stop ensure ->decimalPlaces($decimalPlaces) is respected and always there. Found that when editing an existing number, it would remove decimals while typing, etc. This solves the issue.

### DIFF
--- a/packages/forms/src/Components/TextInput/Mask.php
+++ b/packages/forms/src/Components/TextInput/Mask.php
@@ -149,6 +149,7 @@ class Mask implements Jsonable
                     ->numeric()
                     ->thousandsSeparator($thousandsSeparator)
                     ->decimalPlaces($decimalPlaces)
+                    ->padFractionalZeros()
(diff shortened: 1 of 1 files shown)


MACscr:
You might be able to ignore this update as the money mask is still completely broken for me. Seems fine on display, but try to edit or write out a full amount like `500.50` and you pretty much have to trick the input into taking the data by click after the decimals. Im fully giving up on the imaskjs library at this point. Wasted too many hours on it.

danharrin:
Gonna merge this anyway, can you remember the name of the number masking library that I found to integrate with a new `MoneyInput`? I can't find it through Discord search.

MACscr:
> Gonna merge this anyway, can you remember the name of the number masking library that I found to integrate with a new `MoneyInput`? I can't find it through Discord search.

https://github.com/RobinHerbots/Inputmask

PR #1156, opened 2022-01-07
Update grid layout docs

jeffgreco13:
Explain the ability to use a string classname in columnSpan to remove a column in a grid.


danharrin:
Thanks!

PR #1166, opened 2022-01-07
File Upload - Keep original filename

robertorinaldi-dev:
Gives the user a way to keep the original filename of files on upload.
The default behavior remains the current one to avoid breaking changes, but it can be changed with `->keepFilename()`.

Perhaps users should be warned of the risk of collisions in the event of duplicate filenames or we should implement a way to avoid them.

With images it still doesn't work as `getClientOriginalName()` always returns 'blob', I don't know if you want to wait for this PR ( https://github.com/livewire/livewire/pull/4486), hoping it will be accepted soon.

### DIFF
--- a/packages/forms/src/Components/BaseFileUpload.php
+++ b/packages/forms/src/Components/BaseFileUpload.php
@@ -23,8 +23,6 @@ class BaseFileUpload extends Field
 
     protected bool | Closure $isMultiple = false;
 
-    protected bool | Closure $preserveFilename = false;
-
     protected int | Closure | null $maxSize = null;
 
     protected int | Closure | null $minSize = null;
@@ -32,6 +30,8 @@ class BaseFileUpload extends Field
     protected int | Closure | null $maxFiles = null;
 
     protected int | Closure | null $minFiles = null;
+
+    protected bool | Closure $shouldPreserveFilenames = false;
 
     protected string | Closure $visibility = 'public';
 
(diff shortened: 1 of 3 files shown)


packages/forms/src/Components/BaseFileUpload.php
```diff
@@ -104,6 +106,12 @@ protected function setUp(): void
         });
 
         $this->saveUploadedFileUsing(function (BaseFileUpload $component, TemporaryUploadedFile $file): string {
+            if ($this->getKeepFilename()) {
```
danharrin, line 109: Instead of duplicating the code here, should we use `$component->getKeepFilename() ? $file->getClientOriginalName() : $file->getFilename()`?

Also, all instances of `$this` inside closures need to be replaced with `$component` to ensure that repeaters aren't broken with this new functionality.

packages/forms/src/Components/BaseFileUpload.php
```diff
@@ -137,6 +145,13 @@ public function disk($name): static
         return $this;
     }
 
+    public function keepFilename(bool | Closure $keepFilename = true): static
```
danharrin, line 148: Instead of "keep filename", maybe "preserve filename" more eloquently describes this feature?

danharrin (changes requested):

packages/forms/src/Components/BaseFileUpload.php
robertorinaldi-dev: Fixed, if I understand correctly

packages/forms/src/Components/BaseFileUpload.php
robertorinaldi-dev: I was also undecided with keepOriginalFilename.
Confirmed preserveFilename?

danharrin: Yeah - `preserve` is synonymous to `keepOriginal`.

danharrin:
Great, thanks!

PR #1181, opened 2022-01-09
Adds placeholder function to SelectFilter and MultiSelectFilter

DistantOrbit:
Added HasPlaceholder trait
Added setUp() function (to assign default placeholder string)
Added default placeholder strings to lang files, with translations for EN and SV


danharrin:
Thanks!

PR #1185, opened 2022-01-10
added code by @DanHarrin to fix the problem with limit() on TextColumns

Cigoler:
Previously if the table had any entries with a NULL value it would error out, Dan provided the following fix on Discord.


danharrin:
Thank you :)

PR #1188, opened 2022-01-10
Remove shadow from sidebar

jeroenvanrensen:
In line with an earlier PR, https://github.com/laravel-filament/filament/pull/536, I think it would also make the design better if the shadow on the left sidebar would be removed.


ryangjchandler:
These sort of changes are subjective. You could probably publish the views and make this change yourself.

danharrin:
Yeah, this is subjective. However, if the community reacts with enough 👍 to this PR, then I will consider it again.

PR #1195, opened 2022-01-11
Add option to define the font Filament uses separately

ralphjsmit:
This PR adds the option to set a separate font for the Filament Admin panel.

**Context**
I'm trying the v2 admin panel out (mainly used the separate packages until now) and I wanted to change the colors of the admin panel.

I registered a new theme via the ServiceProvider, based on my project's `tailwind.config.js` and I changed the colors. This worked fine.

For this project, I also needed to add a few custom fonts, so in the `tailwind.config.js` file I updated my font configurations, among them the `font-sans`. Unfortunately this also changed the font in the admin panel.

The sans font I used didn't look that pretty, so that's why I went looking for ways to customize this further.

I noticed that the DM Sans font was automatically added to the `html` element by Tailwind. So I came up with the idea of adding a `font-filament` class to the `html` element, so that the font-family can be easily overridden only for the admin-panel.

In the Discord you said that you were looking for ways to let users customize the theme, without having to publish the theme, so this would be a good addition IMO.

Thanks!


danharrin:
Hi, this is definitely a valid concern, but I don't think that an extra class is the right solution here. Instead, we should encourage people to compile a separate Admin-only theme stylesheet using Laravel Mix, instead of sharing it with their main app. Does that make sense? In which case, you have a custom font just for your Admin panel.

PR #1200, opened 2022-01-11
Send flash notifications from Widgets

jeffgreco13:


danharrin:
Hi, thanks for this. As discussed in #1149, we're actually going to refactor the notification system. In which case, that would make this method a breaking change. While this works for now, it probably won't in the future, so I'm not going to add it to the docs yet. Thanks though

PR #1205, opened 2022-01-12
[Admin] Extract update and create method

nuradiyana:
As in the discussion, I extracted create and update to dedicated method. I don't know if there is a need for additional testing, if needed I will add it.


danharrin:
Thanks

PR #1221, opened 2022-01-12
Fixes RTL forever🔥

babakfp:
Hi🖐️

Using the `rtl` modifier isn't a standard way of doing RTL, and it's a pain for the developers.

In this PR, I removed all `rtl` modifier classes and I added the [RTLCSS](https://rtlcss.com) package so the RTL CSS code automatically get created.

RTLCSS will do everything you need. So, you don't need to worry about RTL anymore.
- You are not going to get any RTL related issues
- No extra work like `rtl:pr-4`
- The source code going to be much cleaner
- And much more...

I set it up so RTLCSS will create a file named `app-rtl.css` in `dist` folder. See `package.json`'s `build` script.

This PR isn't complete because I'm not a Laravel developer and I didn't know how to handle the rest. I only prepared and added some necessary changes. What else needs to be done:
1. Add a config option (for RTL) to users be able to switch between `app.css` and `app-rtl.css` files.
2. Do `npm run build`
3. Please read [RTLCSS docs](https://rtlcss.com/learn) for the config options and more...

---

This isn't 100% necessary but it's good to have. Some icons also need to support RTL. For example, if you are using an icon like this "👉", it needs to look like this "👈" in RTL.

## How to accomplish this?

1. Add the SVG shape name as a class to the `<svg>` tag. For example: `<svg class="svg-dashboard" />`.
2. Add this config to `tailwind.config.js`:
```
theme: {
	extend: {
		scale: {
			'-px': '-1',
		},
	},
},
```
3. add this CSS:
```css
.svg-dashboard {
  @apply -scale-x-px;
}
```
This will do the trick.
You can also add this class `-scale-x-px` to make this "👇" look like this "👆".

---
Thanks for reading🥲 Mention me if you need any help.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "filament",
     "scripts": {
-        "build": "node bin/build && mix --production",
+        "build": "rtlcss packages/admin/dist/app.css packages/admin/dist/app-rtl.css && node bin/build && mix --production",
         "watch": "node bin/build && mix watch"
     },
     "devDependencies": {
@@ -29,6 +29,7 @@
         "marked": "^3.0.0",
         "mdhl": "^0.0.6",
         "postcss": "^8.4.4",
+        "rtlcss": "^3.5.0",
         "tailwindcss": "^3.0.0",
         "trix": "^1.3.1"
     }
--- a/packages/admin/resources/views/components/button.blade.php
+++ b/packages/admin/resources/views/components/button.blade.php
@@ -25,12 +25,12 @@
         'w-6 h-6' => $size === 'md',
         'w-7 h-7' => $size === 'lg',
         'w-5 h-5' => $size === 'sm',
-        'mr-1 -ml-2 rtl:ml-1 rtl:-mr-2' => ($iconPosition === 'before') && ($size === 'md'),
-        'mr-2 -ml-3 rtl:ml-2 rtl:-mr-3' => ($iconPosition === 'before') && ($size === 'lg'),
-        'mr-1 -ml-1.5 rtl:ml-1 rtl:-mr-1.5' => ($iconPosition === 'before') && ($size === 'sm'),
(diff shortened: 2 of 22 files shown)


robertorinaldi-dev:
There is also this option https://github.com/20lives/tailwindcss-rtl

robertorinaldi-dev:
Nvm https://github.com/20lives/tailwindcss-rtl/issues/42#issuecomment-1004777163 😂

danharrin:
Hi, I do appreciate the effort here. Thank you.

However - I believe this is a breaking change. Not everyone is using the Filament assets as-is - some are compiling their own themes (which then in turn also need this plugin to support RTL now)
- some people are using the forms and tables packages outside the admin panel and compiling their own Tailwind classes

Am I correct in saying that all people now compiling this CSS need the new package? In which case, we'll need to hold off on this change.

babakfp:
> There is also this option https://github.com/20lives/tailwindcss-rtl

It needs extra work and a learning curve. Also, you need to replace some of the currently used classes. It has other problems like, it's not documented officially in tailwind docs and new people can get confused and etc.

babakfp:
I think I need more info on how developers work with this package. I'll keep you informed...

danharrin:
@mohamedsabil83 and @maherelgamil - you've both been maintaining our RTL. This is a breaking change, right? And using more `space-x-***` would help reduce RTL-specific classes too, correct?

mohamedsabil83:
> @mohamedsabil83 and @maherelgamil - you've both been maintaining our RTL. This is a breaking change, right? And using more `space-x-***` would help reduce RTL-specific classes too, correct?

yes, it is a breaking change. The **RTL** modifier made for situations as we @maherelgamil and I did. using rtlcss it's good but not for this. cause you have to do a lot of change for any additional package the dev may use. Also, if anyone uses a library that controls **direction** in js the result will become wrong with CSS.

IMHO, the best is to keep things as we did.

danharrin:
Thank you very much @mohamedsabil83, and thank you @babakfp.

PR #1228, opened 2022-01-13
Docs: Explain how to conditionally hide pages

DistantOrbit:
Added a paragraph about how to use shouldRegisterNavigation() in page classes to docs.


danharrin:
Thanks :)

robertorinaldi-dev:
@danharrin what are the cons in directly making a method that accepts a closure and takes care of the rest automatically? Maybe canView() like for widgets?

danharrin:
The cons are that every page has different `mount()` parameters, so if we add `mount()` to the base Page class we effectively stop people from using mount parameters.

PR #1230, opened 2022-01-13
Include hidden component containers when hydrating a form

travisobregon:
When calling `$this->form->fill()` to set initial default values or to clear a form after submission, hidden component containers are ignored. 

This PR ensure all fields are hydrated.

### DIFF
--- a/packages/forms/src/Components/Concerns/HasChildComponents.php
+++ b/packages/forms/src/Components/Concerns/HasChildComponents.php
@@ -46,7 +46,7 @@ trait HasChildComponents
 
     public function hasChildComponentContainer(bool $withHidden = false): bool
     {
-        if ($this->isHidden() && ! $withHidden) {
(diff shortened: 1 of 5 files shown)


packages/forms/src/Components/Concerns/HasChildComponents.php
```diff
@@ -35,17 +35,17 @@ public function getChildComponentContainer(): ComponentContainer
...
         }
 
         return [$this->getChildComponentContainer()];
     }
 
-    public function hasChildComponentContainer(): bool
+    public function hasChildComponentContainer(bool $withHidden = false): bool
     {
-        return ! $this->isHidden() && $this->getChildComponents() !== [];
+        return $withHidden ?: ! $this->isHidden() && $this->getChildComponents() !== [];
```
danharrin, line 49: Can we refactor this line to use early returns please?

PR #1233, opened 2022-01-14
Field - Icon

robertorinaldi-dev:
Picking up an old question from discord, I made a draft PR of how an icon could be added to fields.

I left it generic for any icon, maybe then I can add a `translatable()` method that directly sets `icon('heroicon-o-translate')`.

field-icon

For now I have configured it only on the TextFields, if it's ok for you I can proceed with the other fields.

Let me know what you think.

Thanks


danharrin:
Hmm, I'm tempted to say that this should be slightly smaller? Also, maybe call it `hintIcon()`.

robertorinaldi-dev:
Yeah, I had simply copied the size from the menu icon. I try to reduce it.

Regarding the function name, I set icon() to use the same trait used elsewhere. At this point maybe better to add it to HasHint? We could also use hint() itself adding a second parameter.

danharrin:
I think we should use `HasHint` yes, but it should be its own method in case people don't want text but want an icon.

robertorinaldi-dev:
Better?
field-icon

danharrin:
Thank you very much

PR #1235, opened 2022-01-14
Add hook to register plugin macros

tanthammar:
Instead of overriding the `packageBooted()` method when extending the `Filament/PluginServiceProvider`, this PR adds a method that also makes it possible for Filament to control _when_ `macros` should be registered.

Also suggest to add the method to the Plugin development documentation.


danharrin:
Thanks :)

PR #1238, opened 2022-01-14
fix: CanViewRecord hiding all relations in edit page

drhoussem:
This fixes hidden relationship table when using CanViewRecord.


danharrin:
Hey, thank you but this isn't the correct solution. You're just switching the array key to one that works in your project.

I've fixed the issue in #1250.

PR #1246, opened 2022-01-16
Adds rows / keys / values editing to KeyValue form component

saade:
When we disable key editing using `->disableEditingKeys()`, values cannot be edited too.

This PR introduces two new methods to handle editing:

`->disableEditingValues()` - Disables only values editing.
`->disableEditingRows()` - Disables both key and value editing.

### DIFF
--- a/packages/forms/resources/views/components/key-value.blade.php
+++ b/packages/forms/resources/views/components/key-value.blade.php
@@ -48,7 +48,7 @@
                                     x-model="row.key"
                                     x-on:input="updateState"
                                     {!! ($placeholder = $getKeyPlaceholder()) ? "placeholder=\"{$placeholder}\"" : '' !!}
-                                    @if ((! $canEditRows()) || (! $canEditKeys()) || $isDisabled())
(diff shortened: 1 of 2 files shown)


packages/forms/resources/views/components/key-value.blade.php
```diff
@@ -86,7 +86,7 @@ class="text-danger-600 hover:text-danger-700"
                 </tbody>
             </table>
 
-            @if ($canAddRows() && $canEditKeys() && (! $isDisabled()))
+            @if ($canAddRows() && $canEditRows() && (! $isDisabled()))
```
saade, line 89: I've changed this, since the user can disables only keys or values editing.

If the user disables row editing, then it cannot add new rows, since this rows will cannot be edited.
If the user disables only key or value editing, then they can add new rows, because they can edit them separately

danharrin:
Hey, surely `disableEditingRows()` is just `disabled()`?

saade:
> Hey, surely `disableEditingRows()` is just `disabled()`?

Done

danharrin:
Thanks!

PR #1288, opened 2022-01-18
Fix slug translatable

mohamedsabil83:
with `->setLocale()` method, the slug field - [Spatie Sluggable](https://github.com/spatie/laravel-sluggable) - is translated using the latest locale selected and that's for all translations. By using `->setTranslation()` method instead, the slug translation works as expected.


danharrin:
As mentioned in DM, I'm pretty sure that this is a bug with the Spatie package unfortunately. I'm hesitant to merge this change as it would decrease performance on larger forms (having to loop through every field and set the translation manually). And for the small amount of people who are likely using Sluggable & Translatable, I don't think it's worth it.

If anyone else comes across this PR, please check the code and override the `handleRecordUpdate()` method on your page in the same way if you need a fix.

PR #1291, opened 2022-01-19
Add CSS classes to Filament so that people can more easily tweak little design choices

ralphjsmit:
As discussed in the Discord earlier this week, this PR adds CSS-classes to all the Blade components of the three packages.

This allows people to make small tweaks to Filament, without having to maintain the burden of publishing views.

For example, in a current project, we wanted to tighten the columns in a table somewhat more. We're now able to add the following to our CSS and it's done:

```css
.filament-tables-columns-text-column {
    @apply !px-2;
}
```

In all cases I followed the namespace. If a view didn't have a PHP-counterpart, I followed the folder structure. Items like `modal/index.blade.php` are named as `-modal`. Whenever I had to add classes via `$attributes->class(...)`, I added the class as a separate item in the array, so that it stands apart from the default classes.

_PS.: I also added a commit that added the `vendor.nosync` and `node_modules.nosync` folders to `.gitignore` as well. I use those folder names on my Mac because otherwise iCloud Drive has a hard time. I'd appreciate it if you could accept this commit, otherwise my `git status` always floods with 'added' files whenever I make a change :). But if you don't want to add it, you can revert the first commit._


danharrin:
Hey, thanks for this. Please bear with me, this is low priority and will take a while to review. I'm not ignoring it :)

ralphjsmit:
Ok, I understand! Would be nice if you could find a moment for it, because we'd prefer to use them in some projects as well 👍

danharrin:
Excellent. I've made a few tweaks so that the class names use more natural English instead of using namespaces like I suggested originally. Thank you so much for your hard work!

danharrin:
Now time to see how many merge issues this creates in #1363 😭

danharrin:
Update: uh oh

Screenshot 2022-01-28 at 20 41 04

ralphjsmit:
Thank you also for your hard work! The naming system has indeed improved with your changes, good idea. 👍

Hope it didn't took you too long, but thanks for merging and have a good weekend!😄

PR #1313, opened 2022-01-23
chore: make `Page::notify()` public

ryangjchandler:
Makes the `Page::notify()` method public. Will allow people to request the `Component $livewire` in a callback and call `$livewire->notify()` without having to guarantee execution context of the callback.


danharrin:
Thank you

PR #1316, opened 2022-01-23
Add register beforeScripts

jeffgreco13:
The ability to register scripts before the core js file; specifically for Alpine plugins. + docs.

### DIFF
--- a/packages/admin/docs/07-appearance.md
+++ b/packages/admin/docs/07-appearance.md
@@ -36,8 +36,7 @@ module.exports = {
     ],
     theme: {
         extend: {
-            colors: {
-                // [tl! focus:start]
+            colors: { // [tl! focus:start]
                 danger: colors.rose,
                 primary: colors.blue,
                 success: colors.green,
--- a/packages/admin/docs/07-appearance.md
+++ b/packages/admin/docs/07-appearance.md
@@ -101,23 +101,25 @@ The default is `7xl`.
 
 ## Including frontend assets
 
-You may register your own scripts and styles by including the following in the `AppServiceProvider`'s `boot()` method:
+You may register your own scripts and styles using the `registerScripts()` and `registerStyles()` methods in a service provider's `boot()` method:
 
 ```php
 use Filament\Facades\Filament;
+
+Filament::registerScripts([
+    asset('js/my-script.js'),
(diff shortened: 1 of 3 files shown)


danharrin:
I'm gonna need to think about this one. I don't think that `beforeScripts()` is very clear, and I think this should probably be in the Plugins docs and not Appearance.

Maybe rename it to `alpinePlugins()`? But I don't even know if thats clear for all Alpine plugins. @ryangjchandler?

ryangjchandler:
@danharrin I actually recommended the `$before` approach as an alternative to having a separate method.

Keeps it slightly more ergonomic imo. The `::beforeScripts` method is just an internal detail, so doesn't really matter what it's called. It might not be restricted to Alpine plugins though, so maybe a more generalised name like it has right now makes more sense.

ryangjchandler:
@jeffgreco13 Can you check why PHPStan is failing too please :)

danharrin:
@ryangjchandler PHPStan has been failing since our L9 upgrade, I think there's an issue with Testbench. Nothing to do with this PR

ryangjchandler:
Ah, in which case don't worry!

jeffgreco13:
Thanks @danharrin happy to adjust if something else comes to mind. My thought process:

- Similar to what @ryangjchandler said, I did have `alpinePlugins` before but changed to `beforeScripts` since that property and method call are abstract anyway and could have other uses. The naming might be a bit unclear but since it behaves identical to the existing `registerScripts` flow I figure it should be close.

- In the docs, thoughts on a **Customisation** heading which would encapsulate appearance, branding, theming? I know I've personally found many ways to do slight customisations - like this tooltip where I didn't want to have to setup a whole package just to register the style and script.

danharrin:
Thank you very much, I've renamed it slightly to `beforeCoreScripts()` but not changed the functionality or anything.

pathros:
That's what you needed in your Kernel part 1 video, right? When you were trying to add your $clipboard package before the alpineJS ...

jeffgreco13:
> Thank you very much, I've renamed it slightly to `beforeCoreScripts()` but not changed the functionality or anything.

Sounds good @danharrin happy to help!

PR #1323, opened 2022-01-23
Overwrite create button label

lolen:
Possibility to overwrite the label of the create button in list page.


ManojKiranAppathurai:
can't you override the entire button or maybe you can override in translation.

danharrin:
Thank you very much, but for simplicity: I'd like to make it a convention that you override action methods to customise them. For example:

```php
protected function getCreateButtonAction(): ButtonAction
{
    return parent::getCreateButtonAction()->label('Label here');
}
```

PR #1332, opened 2022-01-24
Allow string rules()

tanthammar:
This PR adds support for validation `rules()` as `string`. 
Less typing :)

Example:
```php
TextInput::make('name')->rules('required|string|min:2|max:125'),
```

### DIFF
--- a/packages/forms/src/Components/Concerns/CanBeValidated.php
+++ b/packages/forms/src/Components/Concerns/CanBeValidated.php
@@ -62,8 +62,12 @@ trait CanBeValidated
         return $this;
     }
 
-    public function rules(array $rules, bool | Closure $condition = true): static
+    public function rules(string | array $rules, bool | Closure $condition = true): static
     {
+        if(is_string($rules)) {
+            $rules = explode('|', $rules);
+        }
+        
         $this->rules = array_merge(
             $this->rules,
             array_map(fn (string | object $rule) => [$rule, $condition], $rules),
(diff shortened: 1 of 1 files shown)


tanthammar:
@danharrin ;) I love it because I don't need to write [],''''''''' 
Thank you for permitting it.

danharrin:
By the way - I'd advise using the dedicated methods (like `->required()`) etc whereever possible - lots of them provide frontend validation.

pathros:
> By the way - I'd advise using the dedicated methods (like `->required()`) etc whereever possible - lots of them provide frontend validation.

Understood. I'll use the dedicated methods first.

farhad321:
how to pass variable in Rule class
same below
image
method rules accept only array, unlike other methods that accept closure

pathros:
@farhad321 
From the [docs](https://filamentphp.com/docs/2.x/forms/validation#custom-rules)

I can show you an example I did:

```php
                TextInput::make('custom_email')
                    ->unique(table: Participants::class, column: 'email_custom' ,callback: function (Unique $rule)use($meeting){
                        return $rule->where('meeting_id',$meeting->id);//make it unique only per meeting, not the whole table
                    })
```
It's more confortable to ask right in the discord channel. Go to https://filamentphp.com/ and click on the Discord channel.

danharrin:
> how to pass variable in Rule class
> same below
> image
> method rules accept only array, unlike other methods that accept closure

It accepts an array of closures.

PR #1351, opened 2022-01-26
feature: advanced (isolated) navigation API

ryangjchandler:
This pull request lays the foundations for a more advanced Navigation API that allows complete control over the navigation sidebar.

It introduces a new `Filament::navigation()` method which can be called from the `boot` method of a `ServiceProvider`.

```php
use Filament\Navigation\NavigationBuilder;

// The `$builder` is required since the callback is invoked through the container.
Filament::navigation(function (NavigationBuilder $builder) {

});
```

Once you add this callback function, Filament's default automatic navigation will be disabled and your sidebar will be empty. This is done on purpose, since this new API is designed to give you complete control over the navigation.

If you want to register a new group, you can call the `NavigationBuilder::group` method.

```php
Filament::navigation(function (NavigationBuilder $builder) {
    $builder->group('Settings', [
        // An array of `NavigationItem` objects.
    ]);
});
```

You provide the name of the group and an array of `NavigationItem` objects to be rendered. If you've got a `Resource` or `Page` you'd like to register in this group, you can use the following syntax:

```php
Filament::navigation(function (NavigationBuilder $builder) {
    $builder->group('Content', [
        ...PageResource::getNavigationItems(),
        ...CategoryResource::getNavigationItems(),
    ]);
});
```

The `Resource::getNavigationItems()` and `Page::getNavigationItems()` methods were introduced in this PR, so anybody who has overwritten the original `registerNavigationItems()` method will need to do some extra work to get this syntax working.

You can also register ungrouped items using the `NavigationBuilder::item()` and `NavigationBuilder::items()` methods.

### DIFF
--- a/packages/admin/src/FilamentManager.php
+++ b/packages/admin/src/FilamentManager.php
@@ -34,7 +34,7 @@ class FilamentManager
 
     protected array $widgets = [];
 
-    protected ?Closure $navigationBuilder;
+    protected ?Closure $navigationBuilder = null;
 
     public function auth(): Guard
     {
--- a/packages/admin/src/FilamentManager.php
+++ b/packages/admin/src/FilamentManager.php
@@ -6,7 +6,6 @@ use Closure;
 use Filament\Events\ServingFilament;
 use Filament\Models\Contracts\HasAvatar;
 use Filament\Models\Contracts\HasName;
-use Filament\Navigation\NavigationBuilder;
 use Illuminate\Contracts\Auth\Guard;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Arr;
--- a/packages/admin/src/FilamentManager.php
+++ b/packages/admin/src/FilamentManager.php
@@ -6,6 +6,7 @@ use Closure;
 use Filament\Events\ServingFilament;
 use Filament\Models\Contracts\HasAvatar;
 use Filament\Models\Contracts\HasName;
+use Filament\Navigation\NavigationBuilder;
 use Illuminate\Contracts\Auth\Guard;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Arr;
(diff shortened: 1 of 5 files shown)


packages/admin/src/FilamentManager.php
```diff
@@ -33,18 +34,37 @@ class FilamentManager
...
 
+    public function navigation(Closure $builder): void
+    {
+        $this->navigationBuilder = $builder;
+    }
+
+    public function buildNavigation(): array
+    {
+        /** @var \Filament\Navigation\NavigationBuilder $builder */
+        $builder = app()->call($this->navigationBuilder, [
```
danharrin, line 52: We could remove the `app()->call()` params and allow the container to instantiate a `NavigationBuilder`, right?

packages/admin/src/Pages/Page.php
```diff
@@ -40,15 +40,20 @@ public static function registerNavigationItems(): void
             return;
         }
 
-        Filament::registerNavigationItems([
+        Filament::registerNavigationItems(static::getNavigationItems());
+    }
+
+    public static function getNavigationItems(): array
```
danharrin, line 46: Pretty sure they aren't - but these method name changes to resources and pages arent breaking are they? If people have overridden the `registerNavigationItems()` methods.

ryangjchandler: I wouldn't have said so. If you overwrite the method completely, then you won't be using the underlying method. If you are using it, then the behaviour is still the same just extracted.

packages/admin/src/FilamentManager.php
ryangjchandler: Yeah we could do actually. There's no dependencies so it'll just make a new instance anyway.

packages/admin/src/Pages/Page.php
danharrin: Awesome

packages/admin/src/FilamentManager.php
danharrin: Great

PR #1352, opened 2022-01-26
feature: collapsible navigation groups

ryangjchandler:
Adds support for collapsible navigation groups. 

The collapsed status is persisted between pages loads with `Alpine.$persist`.

https://cln.sh/Zg08Xx


SebastiaanKloos:
You are amazing🎉

PR #1363, opened 2022-01-28
feature: darkmode

Majkie:
So, I had enough time to make dark mode, hopefully you will appreciate almost 2 days spent just on this. I believe and know there is still enough space to improve, but I don't have that much of a free time to fix all known and unknown issues. That's why I added possibility to turn off darkmode for now. I believe this awesome community will send PR as soon as they find something light in their project, this is just supposed to be some kind of kick off to ease our eyes.

## Known issues

**Chart widgets** - Best to see in Filament demo, if you switch theme it won't update colors of the graph, specifically line, either it stays light or dark based on what was the initial theme. Also as you can see in the screenshots below I'm not exactly sure how to set color of line charts, it might be sufficient, but I believe it would be easier to read if labels were brighter.

**Trix / RichEditor** - Toolbar buttons use image as a icon, so the color of icons may not be changed that easily. Also I couldn't figure out why there is white line in between them.

**Input autocomplete** - Whenever input is filled with suggestion from browser the input is white, I tried to use `autofill:bg-dark-700` presented by Tailwind `Style an input when it has been autofilled by the browser using the autofill modifier:`, but without success (Chrome).

I believe there is more of them, but to do everything at once is not that easy, especially if you have to go through every single component. Thanks y'all for future PR's resolving all of these issues.

## Screenshots

Screenshot 2022-01-27 at 23 56 43

Screenshot 2022-01-27 at 23 56 57

Screenshot 2022-01-27 at 23 57 11


saade:
This looks beautiful! Nice work

danharrin:
Love this. Thank you so much. I will review as soon as I can.

ralphjsmit:
Great job!❤️

pxlrbt:
This is looking neat. I'd only change the text color for items with bg-primary

Majkie:
> This is looking neat. I'd only change the text color for items with bg-primary

I tried to go for some other text color for bg-primary but nothing seemed to fit as good as white, afterall even Github uses white text color on buttons and badges in the darkmode, so I ended up leaving it as it is.

danharrin:
Thanks so much! I'm going to merge this in and fix any issues I see going forward before the v2.10.0 release.

danharrin:
After testing this, I've realised that it's actually a breaking change for non-admin users.

Since Tailwind's dark mode default is to rely on the system preference (🙄), this causes fields to turn dark even in light-mode apps if `class` mode is not enabled in the Tailwind build.

I won't be reverting this as it's a killer feature, but I will have to make some changes:

1) It will not be default. You may enable dark mode via a config option, which adds the classes to your HTML and allows your users to switch between dark and light mode.

2) Because of this, all `dark:` classes need to be merged onto the elements conditionally (usually with the `@class` directive). Which is alot of work but again, worth it for such a feature.

Thank you very much, again!

PR #1375, opened 2022-01-29
FileUpload: Add button to download file

saade:
Using `nielsboogaard/filepond-plugin-get-file` to add a download button to filepond component


ryangjchandler:
@saade Can you provide a screenshot of what it looks like please? :)

saade:
I'm not with my main pc this weekend, but it looks like this:


https://nielsboogaard.github.io/filepond-plugin-get-file/


With/without the image preview, the download button stays at left of the image name

saade:
The only thing is that we cannot disable the download button, it will be always present.

If thats an issue, i can fork the plugin, add a condition to disable it and use it instead of the original one

danharrin:
Looks like this plugin has not had a release since 2020, so I am hesitant to merge it in considering that NPM packages often require frequent security patches to fix new vulnerabilities. Wdyt @ryangjchandler?

saade:
@danharrin no worries, i can fork and maintain it, perhaps add a automatic dependabot to it, since i'm definately will use filament in every single laravel project i make in the future :) 

i rlly want to be part of filament community forever <3 haha

saade:
i'm interested on these too, from the same author

https://github.com/nielsboogaard/filepond-plugin-media-preview
and
https://github.com/nielsboogaard/filepond-plugin-image-overlay

ralphjsmit:
@saade Would it be possible to combine the functionality of those three plugins into one plugin? If so, we could tailor that plugin to integrate nicely with Filament and maintain it ourselves.

saade:
> @saade Would it be possible to combine the functionality of those three plugins into one plugin? If so, we could tailor that plugin to integrate nicely with Filament and maintain it ourselves. That would be less 'messy' and easier IMO than to maintain three forks of someone else's project.

yeah, thats what i tought, i just dont get it how to "Plug'n Play" this plugin into the Filepond component without touching app.js and the blade file, since enabling this to everyone won't be nice

ralphjsmit:
Filament has a nice way for plugins to share data with the frontend, perhaps you could use that?

saade:
> Filament has a nice way for plugins to share data with the frontend, perhaps you could use that?

Gonna try it! Thanks

ryangjchandler:
Could we not just add a download link to the input ourselves if a file is present?

saade:
Gonna close this, and reopen using another aproach, thanks!

PR #1383, opened 2022-01-29
Update CanBeHidden traits for a more consistent implementation everywhere

ralphjsmit:
I just noticed that the `CanBeHidden` traits do not share the same implementation everywhere. For example, I wanted to dynamically determine whether to display `ButtonAction` A or B.

To fix that, I copied over the IMO best implementation of `CanBeHidden`, which is the implementation that offers both the `hidden()` and `visible()` methods.

For the page actions there wasn't a way to evaluate closures yet, so I added an `EvaluatesClosures` trait as well for the page actions. I saw that a page action also had a `BelongsToLivewire` trait that exposes a `getLivewire()` method. I used that method to add a `$livewire` parameter to the `EvaluatesClosures` trait. I couldn't think of a use case for that, but I though that it couldn't do any harm either.


danharrin:
Hey, the visible() additions are fine here, but theres no need to accept a closure on pages - you can just use `$this` to access the LW component.

ralphjsmit:
Hey, thanks for the feedback! I updated the PR and removed the Closure support from the page actions component.

danharrin:
Thanks!

PR #1400, opened 2022-01-31
[2.x] Add enum support for Select and Select Filters

ManojKiranA:
Lets say i have a Enum class like this.

```
<?php

enum UserStatus
{
    case UnVerified;
    case Verified;
    case InActive;
    case Disabled;
}
```

So right now i have to do something like this

```
Select::make('user_status')
                        ->label('Status')
                        ->options(
                            collect(UserStatus::cases())
                                ->map(function (UserStatus $userStatus) {
                                    return [
                                        'name' => $userStatus->name,
                                        'value' => $userStatus->value,
                                    ];
                                })
                                ->pluck('name', 'value')
                        ),
```

So this PR add support for passing enum key values into SelectFilter.

```
SelectFilter::make('Status')
            ->column('user_status')
            ->enum(UserStatus::class),
```

### DIFF
--- a/packages/forms/src/Components/Concerns/HasOptions.php
+++ b/packages/forms/src/Components/Concerns/HasOptions.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace Filament\Forms\Components\Concerns;
+
+use Closure;
+use Illuminate\Contracts\Support\Arrayable;
+
+trait HasOptions
+{
+    protected string |array | Arrayable | Closure $options = [];
+
+    public function options(string |array | Arrayable | Closure $options): static
+    {
+        $this->options = $options;
+
+        return $this;
+    }
+
+    public function getOptions(): array
+    {
+        $options = $this->evaluate($this->options);
+
+        if (is_string($options)) {
+            $options = collect($options::cases())
+                ->mapWithKeys(function ($eachEnumInstance) {
+                    return [$eachEnumInstance?->value ?? $eachEnumInstance->name => $eachEnumInstance->name];
+                })
+                ->toArray();
+        }
+
+        if ($options instanceof Arrayable) {
+            $options = $options->toArray();
+        }
+
+        return $options;
+    }
+}
--- a/packages/forms/src/Components/Concerns/HasOptions.php
+++ b/packages/forms/src/Components/Concerns/HasOptions.php
@@ -7,9 +7,9 @@ use Illuminate\Contracts\Support\Arrayable;
 
 trait HasOptions
 {
-    protected string |array | Arrayable | Closure $options = [];
+    protected array | Arrayable | string | Closure $options = [];
 
-    public function options(string |array | Arrayable | Closure $options): static
+    public function options(array | Arrayable | string | Closure $options): static
     {
         $this->options = $options;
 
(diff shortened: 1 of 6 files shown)


ManojKiranAppathurai:
I am not sure whre to add test cases.

danharrin:
Instead of introducing a new method, we should allow people to pass an enum directly into `options()`. Also, on the MultiSelectFilter, Select form component and MultiSelect form component as well.

ManojKiranAppathurai:
So I need to accept the string in options method. And if the option is string I need to consider this as enum class right 🤔

danharrin:
Exactly

ManojKiranAppathurai:
As per request https://github.com/laravel-filament/filament/pull/1400#issuecomment-1025617651

ManojKiranAppathurai:
Shall we move those methods to separate traits 🤔

danharrin:
> Shall we move those methods to separate traits 🤔

For the filters - yes. Extract to a `Concerns\HasOptions` maybe.

Please add enum support to Select and Multiselect form components too :)

ManojKiranAppathurai:
> Please add enum support to Select and Multiselect form components too :)

In forms ?

danharrin:
> > Please add enum support to Select and Multiselect form components too :)
> 
> In forms ?

Yes please mate

danharrin:
In fact, maybe we could completely move the logic to the form components only - pass the enum class from the filter to the `options()` method on the form component.

ManojKiranAppathurai:
> In fact, maybe we could completely move the logic to the form components only

Sounds logical. I will do it.

ManojKiranAppathurai:
@danharrin any update on this

danharrin:
Thanks :)

danharrin:
I didn't remove the string type, I just moved it later on in the definition. Is that what you mean?

Have you tested this on the `v2.10.0-beta1` tag?

ManojKiranAppathurai:
I haven't tested it yet

ManojKiranAppathurai:
everyting works fine. Thanks a ton

aeq-dev:
> Lets say i have a Enum class like this.
> 
> ```
> <?php
> 
> enum UserStatus
> {
>     case UnVerified;
>     case Verified;
>     case InActive;
>     case Disabled;
> }
> ```
> 
> So right now i have to do something like this
> 
> ```
> Select::make('user_status')
>                         ->label('Status')
>                         ->options(
>                             collect(UserStatus::cases())
>                                 ->map(function (UserStatus $userStatus) {
>                                     return [
>                                         'name' => $userStatus->name,
>                                         'value' => $userStatus->value,
>                                     ];
>                                 })
>                                 ->pluck('name', 'value')
>                         ),
> ```
> 
> So this PR add support for passing enum key values into SelectFilter.
> 
> ```
> SelectFilter::make('Status')
>             ->column('user_status')
>             ->enum(UserStatus::class),
> ```

Hi,
Why did you set 'value' as key instead of 'name' set pluck('name','value') ? shouldn't be pluck->('value', 'name') ?

danharrin:
No, its correct

aeq-dev:
@danharrin thanks for quick reply,
As example, I have this : 
```
enum Os: string
{
    case UBUNTU22 = 'Ubuntu 22.04';
    case UBUNTU20 = 'Ubuntu 20.04';
    case UBUNTU18 = 'Ubuntu 18.04';
}

```
So using` ->options(Os::class)`,  and set casts in the model, it saves the 'value' not the the 'name', also it displays 'name' in the form.

danharrin:
The value should be saved into the DB, thats how enums work. The "backed" value (string) is used to serialise. Also check Laravel's enum casts.

aeq-dev:
Oh thanks ! sorry then :))))

PR #1405, opened 2022-01-31
Add extra classes to the main layout and sidebar components

jacksleight:
This PR adds some additional classes to the main layout and sidebar components where there were a few gaps, I think due to the elements not being their own components so not receiving classes in the previous PR?

For example there's a `filament-sidebar-footer` class, but no `filament-sidebar-header`, and none of the main elements within `filament-app-layout` have a class, which leaves them quite tricky to target still.

The additional classes (in bold) are:

* filament-app-layout
    * **filament-sidebar-close**
    * **filament-main**
        * **filament-main-header**
            * **filament-sidebar-open**
            * **filament-breadcrumbs**
        * **filament-main-content**
        * **filament-main-footer**
* filament-sidebar
    * **filament-sidebar-header**
    * **filament-sidebar-nav**
    * filament-sidebar-footer

I'm sure you'll want to tweak some of the naming.

### DIFF
--- a/packages/admin/resources/views/components/layouts/app.blade.php
+++ b/packages/admin/resources/views/components/layouts/app.blade.php
@@ -6,7 +6,7 @@
             x-show="$store.sidebar.isOpen"
             x-transition.opacity.500ms
             x-on:click="$store.sidebar.close()"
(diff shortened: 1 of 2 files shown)


danharrin:
Cool, only a couple of tweaks. Thanks!

PR #1406, opened 2022-01-31
Locale DatePicker and DateTimePicker

### LINKED ISSUE #1335: Locale datepicker
### Package

filament/forms

### Package Version

v2.9.9

### Laravel Version

v8.1

### Livewire Version

_No response_

### PHP Version

PHP 8.1

### Bug description

The locale configuration is not taken in account for the datepicker that is in english only.
I have try many locale: fr, it, es, id, ...

### Steps to reproduce

_No response_

### Relevant log output

_No response_

dmandrade:
With this change I'm passing the translation object directly, for some reason using locale name dont work.

This fix #1335 

I also removed the localeData script call that was being made in the blade file because it is already imported in js.

image
image

### DIFF
--- a/packages/forms/resources/views/components/date-time-picker.blade.php
+++ b/packages/forms/resources/views/components/date-time-picker.blade.php
@@ -1,13 +1,10 @@
 @once
     @push('scripts')
         <script src="//unpkg.com/dayjs@1.10.4/dayjs.min.js"></script>
-        <script src="//unpkg.com/dayjs@1.10.4/plugin/localeData.js"></script>
-        <script>
-            dayjs.extend(window.dayjs_plugin_localeData)
-
-            window.dayjs_locale = dayjs.locale()
-        </script>
         <script src="//unpkg.com/dayjs@1.10.4/locale/{{ strtolower(str_replace('_', '-', app()->getLocale())) }}.js"></script>
+        <script>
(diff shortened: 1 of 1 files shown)


packages/forms/resources/views/components/date-time-picker.blade.php
```diff
@@ -1,13 +1,10 @@
 @once
     @push('scripts')
         <script src="//unpkg.com/dayjs@1.10.4/dayjs.min.js"></script>
```
danharrin, line 3: Do we still need the CDN version too?

dmandrade: I did a test removing the cdn call from dayjs.min.js and the DatePicker still works but the translations don't load.

danharrin: And does this work outside the admin panel also?

dmandrade: I tested with admin and without it
Removing the localeData cdn didn't cause any problem but for some reason the translations stop working when  remove the dayjs cdn .

I think this would need further investigation as I think it's redundant to keep this cdn when it's included in the bundle.

danharrin:
Thanks!

PR #1409, opened 2022-02-01
fixed docs for admin installation remove version to fix no matches found issue on composer require.

sayasuhendra:
fixed docs for admin installation remove version to fix no matches found issue on composer require.

image

### DIFF
--- a/packages/admin/docs/01-installation.md
+++ b/packages/admin/docs/01-installation.md
@@ -6,9 +6,9 @@ title: Installation
 
 Filament has a few requirements to run:
 
-- PHP 8.0+
-- Laravel v8.0+
-- Livewire v2.0+
+-   PHP 8.0+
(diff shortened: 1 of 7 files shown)


danharrin:
As far as I know, this is actually a Composer bug on Windows. The reason why we specify the version is that often people will have an unsupported PHP version and then complain that Filament v1 installs.

danharrin:
Does it work if you try `composer require filament/filament:"^2.0"`?

sayasuhendra:
Sorry for the late response.

> As far as I know, this is actually a Composer bug on Windows. The reason why we specify the version is that often people will have an unsupported PHP version and then complain that Filament v1 installs.

I use MacOS Catalina.

> Does it work if you try `composer require filament/filament:"^2.0"`?

Yes It work.

PR #1422, opened 2022-02-02
Add time() method to CanFormatState

ralphjsmit:
This PR adds a `time()` method to the `CanFormatState` trait, in addition to the already existing `date()` and `dateTime()` formatters.

### DIFF
--- a/packages/tables/config/tables.php
+++ b/packages/tables/config/tables.php
@@ -4,6 +4,7 @@ return [
 
     'date_format' => 'M j, Y',
     'date_time_format' => 'M j, Y H:i:s',
+    'time_format' => 'H:i',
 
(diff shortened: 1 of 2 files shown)


danharrin:
The default in config should be `H:i:s`, right?

ralphjsmit:
If you ask me, `H:i` is somewhat more logical, because displaying seconds in an app almost never makes sense. That's why I choose this format, but feel free to pick the other one.

ryangjchandler:
@ralphjsmit I'd say go with `H:i:s`.

ralphjsmit:
No problem, updated it to `H:i:s`👍

danharrin:
Awesome, thanks

PR #1428, opened 2022-02-02
Fix inline radio item alignment

jacksleight:
Inline radio buttons aren't aligned correctly:

Screenshot 2022-02-02 at 15 00 23

This fixes that:

Screenshot 2022-02-02 at 15 00 30

Changes are:

* Only set the `space-y-2` class if the radio is not inline


danharrin:
Cheers

PR #1431, opened 2022-02-02
Navbar slots

tanthammar:
This PR adds slots in the navbar, to the right of the search field.

Typical use case,
- Button icons
- Notifications dropdown
- Language switcher
- User settings menu
- Team switcher

Usage
Filament config file gets two new items, where one can add paths to Livewire components or a blade views
```php
    'layout' => [
        //...
        'navbar' => [
            'livewire-components' => [],
            'included-views' => [],
        ],
    ],
```

Example
```php
    'layout' => [
       //...
        'navbar' => [
            'livewire-components' => ['navbar.switch-locale', 'navbar.switch-team'],
            'included-views' => ['filament.navbar.quick-links'],
        ],
    ],
```

The specified paths are iterated in app.blade.php
```
@foreach($lwComponents as $lwComponent)
    @livewire($lwComponent)
@endforeach
@foreach($includedViews as $includedView)
    @include($includedView)
@endforeach
```

Non-breaking change because the `config(..., [])` in `app.blade.php` returns an empty array as default value.


ryangjchandler:
I don't think this is the best way to do this really. I'd honestly just recommend publishing the Filament views and adding customisations in there instead.

Using the config file means you can't pass data into the Livewire components or partials, so you instead have to do everything in the mount method. It also disables the global search which is a bit too opinionated I think.

Not sure what you think @danharrin?

tanthammar:
Why does it disable the global search?

ryangjchandler:
My bad, didn't see the global search component in the if statement.

tanthammar:
If you do not want a config setting, can we add an empty view that is always included?

I really want to avoid publishing the views for this.
Think it is a very natural position - UX - for things like language switchers, notifications and more.

If there is an empty view, that is meant to be overridden we could add a publish command that only publishes this "navbar-top-right" view or something.

I would prefer a config setting...

pxlrbt:
> If you do not want a config setting, can we add an empty view that is always included?

I'd prefer this way. Extract as much stuff as possible in smaller views, so it's easier to overwrite.

ryangjchandler:
Yeah, I think the empty view that can be published is probably a little bit better. You could just publish the vendor views as normal and delete the ones you don't need / keep the empty one for the navbar slot.

danharrin:
Yeah, I'm happy with the empty views as well. We've done the same on the sidebar I think (check `start.blade.php` and `end.blade.php`);

tanthammar:
Great. I'll get back w an updated design.

danharrin:
Spoke about this on DC, gonna close.

PR #1435, opened 2022-02-02
Allow dynamically changing the settings property.

mxm1070:
This pull request adds the ability to dynamically set the settings property of a settings page:

```
public static function getSettings(): string
{
    return config('multitenancy.tenant_model')::checkCurrent()
        ? \App\Settings\Tenant\StoreSettings::class
        : \App\Settings\Landlord\StoreSettings::class;
}
```

in your class the same way you can use the getModel function in the Resource class. This is part of my discussion on multitenancy (#1407).


danharrin:
Perfect, thanks!

PR #1437, opened 2022-02-03
Fix sidebar flash

jacksleight:
This is a bit of a random one, and might only bug me so no worries if you ignore it 😄 .

When you load a page on desktop there’s sometimes a very brief flash when the sidebar isn’t visible:

https://user-images.githubusercontent.com/126740/152324821-333161f5-9915-44aa-aed7-549e6fd95719.mov

This only happens sometimes, and just depends on how quickly the browser evaluates the JS.

It's due to Alpine’s `x-cloak` attribute, which isn’t actually necessary on desktop (as the sidebar is always visible), but is necessary on mobile to avoid the opposite problem.

This PR adds two `x-cloak` variants that only apply below or above the `lg` breakpoint:

* `x-cloak=“-lg”`
* `x-cloak=“lg”`

The `x-cloak` on the sidebar has been changed to the `-lg` variant, preventing it from applying on desktop and fixing the flash.

## Known Issue

It would be really nice to use the actual `lg` breakpoint size into the media query, in case it's being overidden by the user and not the default `1024px`. Unfortunately I don't think this is possible while the CSS is in the HTML, it would need to be moved to a CSS file where you could use `theme('screens.lg')`.

### DIFF
--- a/packages/admin/resources/views/components/layouts/app/sidebar/index.blade.php
+++ b/packages/admin/resources/views/components/layouts/app/sidebar/index.blade.php
@@ -1,6 +1,6 @@
 <aside
     x-data="{}"
-    x-cloak
+    x-cloak="-lg"
     x-bind:class="$store.sidebar.isOpen ? 'translate-x-0' : '-translate-x-full rtl:lg:-translate-x-0 rtl:translate-x-full'"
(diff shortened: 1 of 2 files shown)


danharrin:
This is a great solution, I've never thought about including the breakpoint inside `x-cloak` like that. Thanks ❤️

jacksleight:
Sorry just spotted one tweak! The `max-width` on `-lg` should be `1023px`, otherwise at exactly `1024px` both rules would apply.

danharrin:
No worries, I will fix this. Thanks for the heads up :)

PR #1441, opened 2022-02-03
Fix contribution guide

pxlrbt:
Composer requires branches to be prefixed with `dev-`

> you should prefix your custom branch name with "dev-" (without making it part of the actual branch name)

https://getcomposer.org/doc/05-repositories.md#loading-a-package-from-a-vcs-repository


danharrin:
Thanks!

PR #1443, opened 2022-02-03
Allow file extensions restriction on FilePond

pxlrbt:
The default FilePond validation plugin only checks against MIME types. Sometimes you want to limit on extensions as well. While the `<input type="file">` field supports file extensions via `accept` and works out of the box, FilePond tries to compare the files MIME type against the file extension which always fails. 

This PR adds a custom validator which extracts the allowed extensions from the `allowedFileTypes` and returns a valid MIME type when the files extension is in the list of allowed extensions.


danharrin:
Thanks!

kinosang:
why this commit reverted?

PR #1446, opened 2022-02-03
Dont output actions column if there aren't any actions

jacksleight:
Dont output the table actions column if there aren't any actions defined. Cleans up some empty space at the end of the rows.


danharrin:
On line 11, you can see we add 1 to the columns count. We need to make this conditional, only if actions are defined.

danharrin:
Made the change myself, thanks!

jacksleight:
Ah thanks 👍

PR #1449, opened 2022-02-03
fix dayjs localization

saade:
dayjs localization was broken since the last update (#1406 )

### This pr fixes:
- Dayjs localization
> Adding `async defer` to the script we can download the locale script in parallel and load it immediately when it finishes downloading

- Create a custom local dayjs plugin to update and fire events when the locale changes
> Then we can use it to update whatever we want

- Remove redundant scripts from the page, since we've already included them in `date-time-picker.js`

- Add spacing to the month select to give it a left padding


saade:
just a dumb question, do i need to include the build files when creating a PR? or you build yourself before releasing a new version?

PR #1452, opened 2022-02-04
[2.x] Add some disabled styles

ryangjchandler:
The text input and select fields didn't have any disabled styles.


danharrin:
Hmm, lets do `75` instead.

danharrin:
Fair enough, I just like to stick to more conventional numbers. 70 is good if Chrome uses it haha.

PR #1454, opened 2022-02-04
generate slug with resource name not model name

maherelgamil:
If you need to generate two resources with the same model, so the result is duplication in the slug for each resource

I know you can override the slug of resources, but I think, it's more efficient to generate correctly at the main function from the first time and leave override to customization only.

so, I created this pull request to generate the default slug with resource name not with model name
so no more needed code for editing the default slug every time.

I think we need to change the `getLabel` function for the same reason if you accept this pull request.

thanks

### DIFF
--- a/packages/admin/src/Resources/Resource.php
+++ b/packages/admin/src/Resources/Resource.php
@@ -267,9 +267,9 @@ class Resource
 
     public static function getSlug(): string
     {
-        return static::$slug ?? (string) Str::of(class_basename(static::getModel()))
-            ->plural()
-            ->kebab();
+        return static::$slug ?? (string)Str::of( str_replace('Rescource', '', class_basename(new static)))
+                ->plural()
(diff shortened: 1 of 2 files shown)


danharrin:
Hi, this would be a breaking change, I think?

maherelgamil:
@danharrin, I don't think that, 
as an example, if you have `PostResource` the old slug was `posts` and now it also `posts`
And also this commit passed all tests. so if you have any old links with `posts` as hardcoded as an example you don't need to change anything.

The only change is the way of generating

danharrin:
Yeah but if someone was using a resource with a different name to the model, their default slug will now change?

danharrin:
Yeah, so for that reason I think this is too risky. But I agree that this does make more sense 👍

maherelgamil:
@danharrin so it will be in a major version update next time ??!

PR #1463, opened 2022-02-05
Fix dropdown wrapping issue on larger labels

ralphjsmit:
Current:

Schermafbeelding 2022-02-05 om 15 33 17

With this PR:

Schermafbeelding 2022-02-05 om 15 33 09

Without `truncate` on the `<span>`:

Schermafbeelding 2022-02-05 om 15 31 12

### DIFF
--- a/packages/tables/resources/views/components/dropdown/item.blade.php
+++ b/packages/tables/resources/views/components/dropdown/item.blade.php
@@ -8,7 +8,7 @@
 
 @php
     $buttonClasses = \Illuminate\Support\Arr::toCssClasses([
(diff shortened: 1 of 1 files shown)


danharrin:
Swapped `min-w-6` for `flex-shrink-0` which should do the same job without being width-dependant. Thanks!

ralphjsmit:
Thanks! Looks like you forgot to commit that flex-shrink-0 @danharrin ..?

danharrin:
Ah weird, I did do it. Just doesn't show up on this PR :)

PR #1465, opened 2022-02-05
Add scripts before core

titantwentyone:
Had a use case where I wanted to hook into Alpine with `document.addEventListener('alpine:init', () => {...});`. This change enables scripts to placed before the Filament JS to enable hooking (and potentially other things).


danharrin:
Thanks :)

PR #1466, opened 2022-02-06
:sparkles: Get a resource by his model FQCN

Ashk2a:
New method `getResource(string $model): ?Resource` on `FilamentManager` to get a resource by his associated model FQCN. 

In my case this method is useful when I want make a link to a resource page on morph relation. I can get directly the resource associated to the `model_type` in morph relation.

Example of my usage : 

```php
Tables\Columns\TextColumn::make('relation')
    ->label(__('relation'))
    ->getStateUsing(fn(Media $record) => $record->model_type . ': ' . $record->model_id)
    ->url(fn(Media $record) => Filament::getResource($record->model)::getUrl('edit', ['record' => $record->model]), true)
```

Render of this example : 

image

### DIFF
--- a/packages/admin/src/FilamentManager.php
+++ b/packages/admin/src/FilamentManager.php
@@ -182,7 +182,7 @@ class FilamentManager
         return array_unique($this->resources);
     }
 
-    public function getResource(string | Model $model): ?string
+    public function getModelResource(string | Model $model): ?string
     {
         if ($model instanceof Model) {
             $model = $model::class;
@@ -192,7 +192,7 @@ class FilamentManager
             if ($model !== $resource::getModel()) {
                 continue;
             }
-            
+
             return $resource;
         }
 
--- a/packages/admin/src/FilamentManager.php
+++ b/packages/admin/src/FilamentManager.php
@@ -6,6 +6,7 @@ use Closure;
 use Filament\Events\ServingFilament;
 use Filament\Models\Contracts\HasAvatar;
 use Filament\Models\Contracts\HasName;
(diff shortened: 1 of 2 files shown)


packages/admin/src/FilamentManager.php
```diff
@@ -181,6 +182,26 @@ public function getResources(): array
...
+            ? $model::class
+            : $model;
+
+        foreach ($this->getResources() as $resourceClass) {
+            /**
+             * @var Resource $instance
+             */
+            $instance = new $resourceClass();
+
+            if ($modelFqcn === $instance::getModel()) {
```
ryangjchandler, line 197: The `new $resourceClass` is pointless because you're just calling a static method. You can do that on the `$resourceClass` already.

packages/admin/src/FilamentManager.php
```diff
@@ -181,6 +182,26 @@ public function getResources(): array
...
+                return new $instance;
```
ryangjchandler, line 198: `$instance` is already an instance of the resource class. Just return the class string directly, the user can instantiate the object if they want to.

ryangjchandler (changes requested):

packages/admin/src/FilamentManager.php
Ashk2a: Oh god, little mistake... I fixed it. 

I want return `$instance` and not `$resourceClass` because I want keep completion and be sure to manipulate a `Resource` and not a string (fqcn)

danharrin:
Thanks, I've renamed it to `getModelResource()`.

PR #1467, opened 2022-02-06
fix(date-time-picker): missing html closing caret

saade:
Hi Dan, one of the latest merge commit that introduced dark mode (d490a7e42b0d5f4774bc2a16b62704c8c00abf84), broke the datetime picker. It removed the closing caret `>` for the select element, resulting in alpine rendering issues

Here's the fix :)


danharrin:
Thanks :)

PR #1468, opened 2022-02-06
fix: resolve original file type even if extension does not match

saade:
Recently this commit (fba57700e358571230d2e4974a07db18c0dd5bb1) broke the file validation, since it will not return the default `type` that filepond need to do validation.

Currently, given the example:
```php
->acceptedFileTypes(['image/png', '.jpg'])
```
if uploading a `.png` file, the validation will fail, because it will return `''` [(See)](https://github.com/laravel-filament/filament/commit/fba57700e358571230d2e4974a07db18c0dd5bb1#diff-9ca487f871dd58846f89f01cb20f5e468ca577dfe6446695a573d6b7704174e1R140), since '.png' is not in the accepted array (even though image/png is)

another example:
```php
->acceptedFileTypes(['image/png'])
```
this will fail too, since when checking the extension (`.png`) it will not be present on the array (event though `image/png` is)

This PR fixes the behavior, returning the original file `type` (that should always be returned) according to [Filepond Docs](https://pqina.nl/filepond/docs/getting-started/examples/validate-file-type/)

### DIFF
--- a/packages/forms/resources/js/components/file-upload.js
+++ b/packages/forms/resources/js/components/file-upload.js
@@ -131,13 +131,13 @@ export default (Alpine) => {
                     },
                     fileValidateTypeDetectType: (source, type) => {
                         return new Promise((resolve, reject) => {
-                            // Detect valid file extensions and return an accepted type if matching
-                            let fileName = source.name
-                            let acceptedExtensions = acceptedFileTypes.filter((type) => type.startsWith('.'))
-                            let extension = fileName.substr(fileName.lastIndexOf('.'))
+                            // Detect valid file extensions and return any accepted type if matching
+                            let filename = source.name
+                            let extension = filename.substr(filename.lastIndexOf('.'))
 
-                            let isValid = acceptedExtensions.indexOf(extension) >= 0
-                            resolve(isValid ? acceptedFileTypes[0] : '')
+                            if( acceptedFileTypes.includes(extension) ) resolve(acceptedFileTypes[0])
+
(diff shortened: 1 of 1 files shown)


saade:
I've just fixed the filepond validation, Laravel will fail the validation, since `.jpg` will not work with the laravel `mimetypes:` validation rule, it should use `mimes:` but without the `.` in the extension...

For passing mimes:
`mimetypes:image/png`

For passing extensions (but still will check the mime):
`mimes:png`

or for only extension checking:

`extensions:png`

as now, this is invalid:
`mimetypes:image/png,.jpeg`

I cannot make it work since i cannot do "or" validations (for checking `mimetypes` OR `mimes`)


maybe someone can come up with a solution to this :) But the frontend needs this patch since its only checking file extensions and not the mime anymore..

danharrin:
I'm tempted to revert the file extension validation... the validation rule thing is a big oversight. @pxlrbt?

saade:
Yeah, i've tried to solve by using two validation rules, but afaik its not possible to make "or" validation, then would require a bit of work refactoring some code...

Idk if supporting file extensions are a a big thing, mimetypes are good enough.

Or maybe we can make another method, `->acceptedFileExtensions()`, but as i said... it wold require more work

pxlrbt:
@danharrin Please revert this and I'll see whether I can come up with something better.

danharrin:
All done, it's been reverted.

kinosang:
this is why #1443 reverted

PR #1478, opened 2022-02-07
feature: Simple (modal) resources

danharrin:


packages/admin/src/Resources/Pages/ListRecords/Concerns/CanCreateRecords.php
```diff
@@ -0,0 +1,116 @@
...
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Str;
+
+trait CanCreateRecords
+{
+    use UsesResourceForm;
+
+    protected function hasCreateAction(): bool
+    {
+        if (! static::getResource()::canCreate()) {
```
ryangjchandler, line 17: This can be a one liner - just return the condition without the `!`

packages/admin/src/Resources/Pages/ListRecords/Concerns/CanCreateRecords.php
```diff
@@ -0,0 +1,116 @@
...
+        return parent::getCreateAction()
+            ->url(null)
+            ->form($this->getCreateFormSchema())
+            ->mountUsing(fn () => $this->fillCreateForm())
+            ->modalActions([
+                Modal\Actions\ButtonAction::make('submit')
+                    ->label(__('filament::resources/pages/list-records.actions.create.modal.actions.create.label'))
+                    ->submit('callMountedAction')
+                    ->color('primary'),
+                Modal\Actions\ButtonAction::make('submit')
```
ryangjchandler, line 35: Does it matter that these buttons have the same ID/name?

ryangjchandler (changes requested):

packages/admin/src/Resources/Pages/ListRecords/Concerns/CanCreateRecords.php
danharrin: Hah - must have been a bad refactor. Thanks.

packages/admin/src/Resources/Pages/ListRecords/Concerns/CanCreateRecords.php
danharrin: eh, probably not in this case. but it's better practice to have them unique all the time. I will do that now

PR #1479, opened 2022-02-07
fix dark mode for minutes and seconds

saade:
Fix dark mode on minutes and seconds input, it was being dark even with light theme, also the seconds colon was not respecting the config

ps: on line [73](https://github.com/laravel-filament/filament/pull/1479/files#diff-ea8114ac781958e1aea949dc578736fdc36f72f813c7f7395a5fff5a69015f36R73) Headwind just reordered the classes :)


danharrin:
Thanks :)

PR #1482, opened 2022-02-07
fix(hydrate-issue): missing child component check

FurkanGM:
When I use fill function for the form in custom page, I got the this error when I use layout fields like Section, Tabs. 

I'm solved problem with this code lines.

Error Message: Cannot assign null to property \ClassName::$data of type array


danharrin:
Thanks :)

PR #1486, opened 2022-02-07
Reduce uninitialised UI flashes and reflows

jacksleight:
Related to my previous sidebar PR, this reduces uninitialised UI flashes and reflows on page load in a few other spots, to make things feel a little snappier and smoother. The goal is to make the initial render before JS initialisation as close as possible to how it looks after initialisation, within reason. The video below demonstrates what I'm referring to (after the changes is around the half-way point):

https://user-images.githubusercontent.com/126740/152842352-81ba4159-8f32-4a5e-8310-62d81824cfb1.mov

These pre-initialisation UI changes have been implemented:

- [x] Table headers are visible (when applicable)
- [x] Table filter button is visible (when applicable)
- [x] Navigation close group button is hidden
- [x] These fields have tweaked UI:
    - [x] Textarea height is set correctly (when blank)
        - [x] There's also a small bug fix to the `scrollHeight` calculation, which wasn't taking the border into account
    - [x] File Upload height is set correctly (when blank and no custom ratio is set)
    - [x] Multi Select field dropdown lists and loading icons are hidden 
    - [x] Tag dropdown lists and loading icons are hidden 
    - [x] Select loading icons are hidden
    - [x] Rich Editor UI is visible
    - [x] Markdown Editor UI is visible
    - [x] Toggle UI is visible
- [x] These fields have initial values/state set:
    - [x] Checkbox
    - [x] Checkbox List
    - [x] Datetime
    - [x] Markdown
    - [x] Markdown Editor
    - [x] Radio
    - [x] Rich Editor
    - [x] Select
    - [x] Text
    - [x] Textarea
    - [x] Toggle

This won't be able to cover absolutely everything. Some things probably aren't worth the hassle of pre-rendering server side, like the initial value/state of a multi-select. And some things simply aren't known server side, like the scroll height of a textarea. But I think even just these changes make things feel much smoother.

### DIFF
--- a/packages/admin/resources/views/components/layouts/app/sidebar/group.blade.php
+++ b/packages/admin/resources/views/components/layouts/app/sidebar/group.blade.php
@@ -18,7 +18,7 @@
             <x-heroicon-o-chevron-down :class="\Illuminate\Support\Arr::toCssClasses([
                 'w-3 h-3 text-gray-600',
                 'dark:text-gray-300' => config('filament.dark_mode'),
-            ])" x-show="$store.sidebar.groupIsCollapsed(label)" />
+            ])" x-show="$store.sidebar.groupIsCollapsed(label)" x-cloak="" />
 
             <x-heroicon-o-chevron-up :class="\Illuminate\Support\Arr::toCssClasses([
                 'w-3 h-3 text-gray-600',
--- a/packages/admin/resources/views/components/layouts/app/sidebar/group.blade.php
+++ b/packages/admin/resources/views/components/layouts/app/sidebar/group.blade.php
@@ -18,7 +18,7 @@
             <x-heroicon-o-chevron-down :class="\Illuminate\Support\Arr::toCssClasses([
                 'w-3 h-3 text-gray-600',
                 'dark:text-gray-300' => config('filament.dark_mode'),
-            ])" x-show="$store.sidebar.groupIsCollapsed(label)" x-cloak="" />
+            ])" x-show="$store.sidebar.groupIsCollapsed(label)" x-cloak />
 
             <x-heroicon-o-chevron-up :class="\Illuminate\Support\Arr::toCssClasses([
                 'w-3 h-3 text-gray-600',
--- a/packages/forms/resources/js/components/textarea.js
+++ b/packages/forms/resources/js/components/textarea.js
@@ -7,7 +7,7 @@ export default (Alpine) => {
         render: function () {
             if (this.$el.scrollHeight > 0) {
                 this.$el.style.height = '150px'
-                this.$el.style.height = this.$el.scrollHeight + 'px'
+                this.$el.style.height = this.$el.scrollHeight + 2 + 'px'
             }
         },
     }))
--- a/packages/forms/resources/js/components/textarea.js
+++ b/packages/forms/resources/js/components/textarea.js
@@ -7,7 +7,7 @@ export default (Alpine) => {
         render: function () {
             if (this.$el.scrollHeight > 0) {
                 this.$el.style.height = '150px'
-                this.$el.style.height = this.$el.scrollHeight + 2 + 'px'
+                this.$el.style.height = (this.$el.scrollHeight + 2) + 'px'
             }
         },
     }))
--- a/packages/forms/resources/views/components/toggle.blade.php
+++ b/packages/forms/resources/views/components/toggle.blade.php
@@ -21,22 +21,25 @@
                     'bg-primary-600': state,
                     'bg-gray-200 @if (config('forms.dark_mode')) dark:bg-gray-700 @endif': ! state,
                 }"
-                x-cloak
                 {!! $isAutofocused() ? 'autofocus' : null !!}
                 {!! $isDisabled() ? 'disabled' : null !!}
                 id="{{ $getId() }}"
                 type="button"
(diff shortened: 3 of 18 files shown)


packages/admin/resources/views/components/layouts/app/sidebar/group.blade.php
```diff
@@ -18,7 +18,7 @@ class="flex items-center justify-between w-full"
             <x-heroicon-o-chevron-down :class="\Illuminate\Support\Arr::toCssClasses([
                 'w-3 h-3 text-gray-600',
                 'dark:text-gray-300' => config('filament.dark_mode'),
-            ])" x-show="$store.sidebar.groupIsCollapsed(label)" />
+            ])" x-show="$store.sidebar.groupIsCollapsed(label)" x-cloak="" />
```
ryangjchandler, line 21: `x-cloak` shouldn't have `""` in this case.
```suggestion
            ])" x-show="$store.sidebar.groupIsCollapsed(label)" x-cloak />
```

packages/forms/resources/js/components/textarea.js
```diff
@@ -7,7 +7,7 @@ export default (Alpine) => {
         render: function () {
             if (this.$el.scrollHeight > 0) {
                 this.$el.style.height = '150px'
-                this.$el.style.height = this.$el.scrollHeight + 'px'
+                this.$el.style.height = this.$el.scrollHeight + 2 + 'px'
```
ryangjchandler, line 10:
```suggestion
                this.$el.style.height = (this.$el.scrollHeight + 2) + 'px'
```

Would like to avoid all weird JS `number + string` bugs if possible.

packages/forms/resources/views/components/toggle.blade.php
```diff
@@ -21,30 +21,37 @@
...
                 {!! $isDisabled() ? 'disabled' : null !!}
                 id="{{ $getId() }}"
                 type="button"
                 {{ $attributes->merge($getExtraAttributes())->class([
-                    'relative inline-flex shrink-0 h-6 w-11 border-2 border-transparent rounded-full cursor-pointer transition-colors ease-in-out duration-200 focus:outline-none focus:border-primary-500 focus:ring-2 focus:ring-primary-500 bg-gray-200 filament-forms-toggle-component',
+                    'relative inline-flex shrink-0 h-6 w-11 border-2 border-transparent rounded-full cursor-pointer transition-colors ease-in-out duration-200 focus:outline-none focus:border-primary-500 focus:ring-2 focus:ring-primary-500 filament-forms-toggle-component',
                     'border-gray-300' => ! $errors->has($getStatePath()),
                     'border-danger-300 ring-danger-500' => $errors->has($getStatePath()),
+                    'bg-primary-600' => $getState(),
+                    'bg-gray-200' . (config('forms.dark_mode') ? 'dark:bg-gray-700' : null) => ! $getState(),
```
ryangjchandler, line 33:
```suggestion
                    'bg-gray-200' . (config('forms.dark_mode') ? ' dark:bg-gray-700' : null) => ! $getState(),
```
Needs a space.

ryangjchandler (commented):
Some minor things right now - general idea is good though. Let's see how far you can get with it, thanks @jacksleight!

danharrin:
Also - are there unintended side effects of setting the initial `value=""` when using `wire:model` as well? For example, when dynamically creating instantiating half way through the LW lifecycle?

jacksleight:
Great, I'll make those tweaks and continue with the remaining bits.

I'm not too sure about the empty `value` attributes, I've not (yet) spotted any issues... But I'll change it to only output them if there's a value, just in case.

packages/admin/resources/views/components/layouts/app/sidebar/group.blade.php
jacksleight: This was due to it being on a blade component. If it's left as `x-cloak` Blade renders it as `x-cloak="1"` which didn’t match the CSS selector. That’s actually a bug I accidentally introduced in the sidebar PR. 😕

I’ve implemented a [better solution](https://github.com/laravel-filament/filament/pull/1486/commits/c52fc9a70c8a602e6a563de93b5d5499eaaeff79) which fixes that bug.

jacksleight:
I think I’ve got this about as far as I can, shout if you spot anything I’ve missed.

These items are intentionally excluded, as implementing them reliably is probably more work than it’s worth, or just not possible:

* Textarea field height when a long value is set
* File Upload field UI and height when a value or custom ratio is set
* Multi Select field when a value is set
* Tags field when a value is set 
* Key Value field when a value is set

There are a couple of things I’m a little unsure of that you may want to review:

* The value of a Date Time field does not always seem to match the format in `$getFormat()`, causing errors if you try to use that with `createFromFormat`, so I’m currently just relying on `DateTime` to figure out the format.
* I’ve set the initial `min-height` of the File Upload container to `76px`, as this matches the initial render height of FilePond, preventing a height change and reflow when the field is blank. However I’m not exactly sure where this `76px` comes from, it feels a bit arbitrary.
* Searchable select fields with a value set briefly flash the “Select an option” text during initialisation, I’m not quite sure on the best way to fix that.

danharrin (approved):

danharrin:
Thank you so much! These are great.

PR #1497, opened 2022-02-08
Add resource name flag

wychoong:
add {--N|name=} option to specify resource name 

`make:filament-resource {model?} {--N|name=}`

current behaviour: Model name is the Resource name 

with -N option: filenames, slug and labels will use the resource name passed in


danharrin:
Personally, I actually think this is a layer of complexity that we don't need at the moment. I can see why someone would want it, in case they have multiple resources of the same model, but I think it will make things too confusing for new users.

PR #1500, opened 2022-02-08
add darkMode facade

wychoong:


wychoong:
@danharrin personally i think feature flags toggled by methods is easier to configure as no config files require to be published. 

regarding your concerns mentioned on discord:
- other features are not available in methods.
i think more can be added if there are requests

- not something usually controlled programatically 
yes and no. taking dark mode for example, can store in DB to persist the preference across devices.
but still its just more of convenience 

anyway feel free to close it if you don't see a need

danharrin:
Storing this setting in the DB will not persist it. Enabling dark mode like this just lets the user decide between light & dark :)

As with other static config options, lets keep this in the config file.

PR #1508, opened 2022-02-08
Add method to easily override initial table records per page

jacksleight:
Allows you to override the default records per page count (10) without having to override the entire `getDefaultTableRecordsPerPageSelectOption` method and all its other logic.

There may be a better name than "initial".


danharrin:
Maybe we should just move it to a property?

jacksleight:
That would certainly be simpler, I only did it that way to mirror `getTableRecordsPerPageSelectOptions()`.

danharrin:
Thanks :)

PR #1511, opened 2022-02-09
fix: Remove initial field values

danharrin:
Removes the initial state set from #1486, as it is causing Livewire DOM-diffing issues in reactive components.


jacksleight:
Ah I didn't realise that would cause an issue.

FYI there's also `{{ $getOptionLabel() }}` on line 87 of `select.blade.php`, and a bunch of `$getState()` calls were added to apply the on/off classes in `toggle.blade.php`.

jacksleight:
Ah didn't realise that would cause an issue.

FYI there's also a `{{ $getOptionLabel() }}` call in select, and bunch of `$getState()` calls in toggle to set the on/off classes.

danharrin:
Thanks!

PR #1516, opened 2022-02-09
extract create modal actions

wychoong:


packages/admin/src/Resources/Pages/ListRecords/Concerns/CanCreateRecords.php
```diff
@@ -3,7 +3,7 @@
 namespace Filament\Resources\Pages\ListRecords\Concerns;
 
 use Filament\Pages\Actions\ButtonAction;
-use Filament\Pages\Actions\Modal;
+use Filament\Pages\Actions\Modal\Actions\ButtonAction as ModalButtonActions;
```
ryangjchandler, line 6: Can we use a relative namespace instead of an aliased import please.

ryangjchandler (changes requested):

danharrin:
Method names like `getCancelModalAction()` aren't appropriate for traits like this, as there are many cancel actions on the class. Prefix each method name with `getCreateAction`, so `getCreateActionCancelModalAction()` or similar. Also, do you need to do the same for `CanAttachRecords`?

wychoong:
> Method names like `getCancelModalAction()` aren't appropriate for traits like this, as there are many cancel actions on the class. Prefix each method name with `getCreateAction`, so `getCreateActionCancelModalAction()` or similar. Also, do you need to do the same for `CanAttachRecords`?

Will look into both.

Actually the purpose of extracting is I want to remove createAndCreateAnother. Do you think it would be better to:
- a config to disable createAndCreateAnother globally 
- Add prepend/append form actions method (if not available yet)

wychoong:
> Prefix each method name with `getCreateAction`,

this will result to 
```php
$this->getCreateActionCreateModalAction(),
$this->getCreateActionCreateAndCreateAnotherModalAction(),
$this->getCreateActionCancelModalAction(),
```
is that ok?

danharrin:
> > Prefix each method name with `getCreateAction`,
> 
> this will result to
> 
> ```
> $this->getCreateActionCreateModalAction(),
> $this->getCreateActionCreateAndCreateAnotherModalAction(),
> $this->getCreateActionCancelModalAction(),
> ```
> 
> is that ok?

Yup that's fine

wychoong:
> > > Prefix each method name with `getCreateAction`,
> > 
> > 
> > this will result to
> > ```
> > $this->getCreateActionCreateModalAction(),
> > $this->getCreateActionCreateAndCreateAnotherModalAction(),
> > $this->getCreateActionCancelModalAction(),
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is that ok?
> 
> Yup that's fine

the PR is ready, it's now is more than just extracting modal actions 😅  but it's to make resource actions consistent with RM.

one part I'm not sure is it the best to set the static property globally with FilamentManager and RelationManager

PR #1517, opened 2022-02-09
Support custom admin panel paths with other parameters

ralphjsmit:
I wanted to use Filament admin with a URL like: `example.com/organizations/7165e010-0f3d-4f2f-86ae-253b6bb3778b`.

In the config file we can set the url:
```php
'path' => 'organizations/{organization}',
```

I have a middleware applied on every single route via the `$middleware` in the kernel:
```php
// ...

URL::defaults([
    'organization' => $organizationUuid,
]);
```

I confirmed with a `dd()` that this code is being run, so that should be good. Having this works for 80%, the only problem was that the JavaScript was not being loaded:

```
(404) https://example.test/organizations/7165e010-0f3d-4f2f-86ae-253b6bb3778b/core/assets/app.js?id=937c2814f9d988613ece
```

I took a look at the AssetController and I found out that the way Filament uses route model binding is the error here:
```php
    public function __invoke(string $file, string $b)
    {
        ray($file); // This is the organization Eloquent model cast to json
        ray($b); // This turns out to be the correct filename.
         
        // This is correct
        $file = request()->file;
```

To fix that, I updated the AssetController like:

```php
    public function __invoke(string $file)
    {
        // The first argument to __invoke will be the first parameter in the complete route,
        // and not the parameter named 'file'. This means that the original $file parameter will
        // not be the correct parameter in cases where people use the admin panel with a custom
        // path like "/admin/{userUUID}/".
        $file = request()->file ?? $file;
```

I tried creating a test, but I'm not aware of a way to flush the registered routes and run the registration process again:

```php
it('can support admin panel paths with other parameters', function () {
    config()->set('filament.path', 'admin');

    $this->get(route('filament.asset', [
        'id' => Filament\get_asset_id('app.js'),
        'file' => 'app.js',
    ]))->assertOk();

    config()->set('filament.path', 'admin/{param}');

    // Flush registered routes and register them again

    $this->get(route('filament.asset', [
        'id' => Filament\get_asset_id('app.js'),
        'file' => 'app.js',
    ]))->assertOk();
});
```

### DIFF
--- a/packages/admin/config/filament.php
+++ b/packages/admin/config/filament.php
@@ -4,7 +4,6 @@ use Filament\Http\Middleware\Authenticate;
 use Filament\Http\Middleware\DispatchServingFilamentEvent;
 use Filament\Http\Middleware\MirrorConfigToSubpackages;
 use Filament\Pages;
-use Filament\Resources;
 use Filament\Widgets;
 use Illuminate\Cookie\Middleware\AddQueuedCookiesToResponse;
 use Illuminate\Cookie\Middleware\EncryptCookies;
--- a/packages/admin/config/filament.php
+++ b/packages/admin/config/filament.php
@@ -28,6 +28,20 @@ return [
 
     'path' => env('FILAMENT_PATH', 'admin'),
 
+
+    /*
+    |--------------------------------------------------------------------------
+    | Filament Asset Path
+    |--------------------------------------------------------------------------
+    |
+    | This is the path which Filament will use to load it's assets. Default is
+    | `_filament`, but you may change it if it conflicts with your other routes.
+    |
+    */
+
+    'asset_path' => env('FILAMENT_PATH', '_filament'),
+
+
     /*
     |--------------------------------------------------------------------------
     | Filament Domain
--- a/packages/admin/config/filament.php
+++ b/packages/admin/config/filament.php
@@ -35,11 +35,11 @@ return [
     |--------------------------------------------------------------------------
     |
     | This is the path which Filament will use to load it's assets. Default is
-    | `_filament`, but you may change it if it conflicts with your other routes.
+    | `filament`, but you may change it if it conflicts with your other routes.
     |
     */
 
-    'asset_path' => env('FILAMENT_PATH', '_filament'),
+    'asset_path' => env('FILAMENT_ASSET_PATH', 'filament'),
 
 
     /*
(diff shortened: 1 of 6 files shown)


ryangjchandler:
@ralphjsmit Thanks for this. @danharrin What do you think about registering the assets route on a pseudo-private URL like `/_filament` instead?

danharrin:
Yeah I agree with Ryan here - we should just move the assets away from the main admin path.

danharrin:
Even just `/filament` will work - without the underscore.

ralphjsmit:
Good idea! Can I help with this or should this be just as simple as updating the routes file?

danharrin:
You can update the routes file if you want

ralphjsmit:
Done! Tested it and works smooth.

packages/admin/src/Http/Controllers/AssetController.php
```diff
@@ -9,6 +9,12 @@ class AssetController
 {
     public function __invoke(string $file)
     {
+        // The first argument to __invoke will be the first parameter in the complete route,
+        // and not the parameter named 'file'. This means that the original $file parameter will
+        // not be the correct parameter in cases where people use the admin panel with a custom
+        // path like "/admin/{userUUID}/".
+        $file = request()->file ?? $file;
```
ryangjchandler, line 16: Do we still need this?

ryangjchandler (changes requested):

packages/admin/src/Http/Controllers/AssetController.php
ralphjsmit: Not necessarily, I removed it.

ryangjchandler (changes requested):

packages/admin/config/filament.php
```diff
@@ -28,6 +28,20 @@
...
+    |--------------------------------------------------------------------------
+    | Filament Asset Path
+    |--------------------------------------------------------------------------
+    |
+    | This is the path which Filament will use to load it's assets. Default is
+    | `_filament`, but you may change it if it conflicts with your other routes.
+    |
+    */
+
+    'asset_path' => env('FILAMENT_PATH', '_filament'),
```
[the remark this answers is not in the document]
ralphjsmit: Done! Thanks👍

ryangjchandler:
@ralphjsmit I think we can leave the docs for now. Not sure if we've got any tests that make sure the asset paths return the correct response, etc. If not, would you be able to add some?

ralphjsmit:
@ryangjchandler No worries😊 I added a test for loading the asset files. Indeed better to have tests!

Btw, I saw that Filament currently does `uses(TestCase::class)` in every test file. Is there a particular reason for that or could we just switch to `uses(TestCase::class)->in(__DIR__)` in Pest? If needed, we can override the testcase in a specific file.

danharrin:
Thank you!

PR #1518, opened 2022-02-09
Fix french translations

marcbelletre:
This fixes a bug introduced by #1509.


danharrin:
Thanks, I wonder why CI didn't pick this up. Maybe because these translation files aren't autoloaded?

PR #1519, opened 2022-02-09
added method to use customized route key name in record route binding.

FurkanGM:
to easily change routeKeyName in resources.


danharrin:
Thanks!

PR #1520, opened 2022-02-09
Additional outline, text and gray button variants

jacksleight:
This PR adds outline and text variants for all button colours, plus a `gray` option:

Screenshot 2022-02-09 at 11 40 37
Screenshot 2022-02-09 at 11 40 44

Rather than introduce a new property on the button component the variants are implemented with a suffix on the `color` property. For example `warning-outline` and `primary-text`. Can change that if a separate property would be better.


danharrin:
Love it! I think we should have it as a separate `outline` property though, and maybe drop the "Text" variants since we already have Link?

jacksleight:
Sure. I was on the fence about the Text option anyway, but it was the easiest to add 😄  .

I *think* I've added the new property and interfaces everywhere they need to be.

packages/admin/src/Pages/Actions/Concerns/CanBeOutline.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Filament\Pages\Actions\Concerns;
+
+trait CanBeOutline
```
ryangjchandler, line 5: `CanBeOutline` sounds weird - `CanHaveOutline` or `CanBeOutlined` perhaps?

packages/admin/src/Pages/Actions/Modal/Actions/Concerns/CanBeOutline.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Filament\Pages\Actions\Modal\Actions\Concerns;
+
+trait CanBeOutline
```
ryangjchandler, line 5: Same as comment above.

ryangjchandler (changes requested):

packages/admin/src/Pages/Actions/Concerns/CanBeOutline.php
jacksleight: I feel like `CanHaveOutline` sounds like the outline is in addition to the existing style, whereas it's really instead of. Shall I go with `CanBeOutlined`?

danharrin:
Thanks :)

PR #1523, opened 2022-02-09
controlling spacing

swarakaka:
You may want buttons only to have icons and not write anything in labels, for example in tables icons will take less space and mean more.

Before:
Bildschirmfoto 2022-02-09 um 14 03 34

After:
Bildschirmfoto 2022-02-09 um 14 02 31


swara-mohammed:
Because IconButtonAction exists

PR #1524, opened 2022-02-09
controlling spacing classes from button component

swarakaka:


swara-mohammed:
Because IconButtonAction exists

PR #1526, opened 2022-02-09
add sort indicator for all sortable columns

maherelgamil:
how can users know the sortable columns? 
So I added a sort indicator to all sortable columns and when the user uses these, it will be shown with different text colors depending on the sorting direction.

image

### DIFF
--- a/packages/tables/resources/views/components/header-cell.blade.php
+++ b/packages/tables/resources/views/components/header-cell.blade.php
@@ -25,17 +25,20 @@
             {{ $slot }}
         </span>
 
-        @if ($isSortColumn)
-            <span @class([
-                'relative flex items-center',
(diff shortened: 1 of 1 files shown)


maherelgamil:
@ryangjchandler So, how can users know if this column is sortable or not?

maherelgamil:
we can increase the size little

danielbidala:
I agree with @ryangjchandler I would suggest something like this

column-sorting

danharrin:
I agree. Add an `opacity-50` down arrow to sortable columns.

maherelgamil:
Thanks all for reviewing, 

I added `opacity-50` for the down arrow to sortable columns

image

danharrin:
Thanks :)

PR #1528, opened 2022-02-09
Allow to trigger delete events on child relations

### LINKED ISSUE #1527: HasManyRepeater field doesn't trigger deleting, deleted oberver events
### Package

filament/forms

### Package Version

v2.10.5

### Laravel Version

v8.83.0

### Livewire Version

v2.10.1

### PHP Version

PHP 8.0.9

### Bug description

When using the **HasManyRepeater** form component, during the delete event of the records, the framework doesn't trigger the deleting, deleted events of the model observer attached.

This is because in the file `Filament\Forms\Components\RelationshipRepeater.php` in the `setUp()` method at line `49` we are running a delete query:

```
$relationship->whereIn($localKeyName, $recordsToDelete)->delete();
```

And in order to trigger the eloquent events, we must retrieve and then execute the delete method instead:

```
$relationship->whereIn($localKeyName, $recordsToDelete)->get()->each(function ($record) {
    $record->delete();
});
```

>Is very important to trigger the propper laravel events because when leading with file uploads inside repeater or items, in order to delete the file from the storage we must have the propper event as specified in the documentation.

### Steps to reproduce

1. Create a model with child relations.
2. Create an observer to the child relation and look for the delete, deleting methods (call the `dd($model)` method).
3. Register the observer in the `EventServiceProvider `boot method.
4. Create a resource and implement the `Filament\Forms\Components\HasManyRepeater` component.
5. When using the application, try to create a resource with child relations (items) and then delete the item.
6. The delete, deleting event of the observer won't execute.

### Relevant log output

_No response_

romaldyminaya:
This fixes the issue #1527 that specifies that when deleting child relations, deleting and delete events from model observer are not triggered.

### DIFF
--- a/packages/forms/src/Components/RelationshipRepeater.php
+++ b/packages/forms/src/Components/RelationshipRepeater.php
@@ -46,7 +46,9 @@ class RelationshipRepeater extends Repeater
                 $recordsToDelete[] = $keyToCheckForDeletion;
             }
 
-            $relationship->whereIn($localKeyName, $recordsToDelete)->delete();
+            $relationship->whereIn($localKeyName, $recordsToDelete)->get()->each(function ($record) {
+                $record->delete();
+            });
 
             $childComponentContainers = $component->getChildComponentContainers();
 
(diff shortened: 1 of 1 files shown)


danharrin:
Thank you :)

PR #1533, opened 2022-02-09
fix: Reset masked input

### LINKED ISSUE #1289: Masked TextInput does not update displayed value in relation edit modal
### Package

filament/filament

### Package Version

v2.9.7

### Laravel Version

v8.80.0

### Livewire Version

_No response_

### PHP Version

PHP 8.1

### Bug description

I have the following form on a relation manager:

```php
{
    return $form
        ->schema([
            TextInput::make('name'),
            Textarea::make('description'),
            TextInput::make('price')
                ->postfix('kr.')
                ->mask(fn (TextInput\Mask $mask) => $mask
                ->patternBlocks([
                    'money' => fn (Mask $mask) => $mask
                        ->numeric()
                        ->integer()
                        ->thousandsSeparator('.')
                ])
                ->pattern('money')),
        ]);
}
```

When I click edit on the first item, I get this modal:

billede

Whether I update the price field or not, that value it has when the modal is closed, is "transferred"/"persisted" to the next edit modal.

billede

The actual price should be 3000 (which is the value stored in the database), and not 1234, as the other package was updated to. However, if I click save, the price is not updated to the price that is displayed. So the problem seems to be front end related.

### Steps to reproduce

Step 1 - Create a relation manager for a resource

Step 2 - Create a masked TextInput field for the related resource
```php
TextInput::make('price')
    ->mask(fn (TextInput\Mask $mask) => $mask
    ->patternBlocks([
        'money' => fn (Mask $mask) => $mask
            ->numeric()
            ->integer()
            ->thousandsSeparator('.')
    ])
    ->pattern('money')),
```

Step 3 - Create two or more related resource, so a list of 2+ resources is displayed

Step 4 - Click edit on the first resource, then close modal and click edit on the second

The masked value from the first modal is "persisted" between any of the other edit modals until full page refresh.

### Relevant log output

_No response_

### LINKED ISSUE #1525: Masked TextInput does not reset after clicking "Create & create another"
### Package

filament/filament

### Package Version

v2.10.3

### Laravel Version

v8.83.0

### Livewire Version

v2.10.1

### PHP Version

PHP 8.0.13

### Bug description

While this issue is similar to #1289, it's slightly different and I wanted to make sure that I reported it.

I have the following form on a relation manager:

```
{
    return $form
        ->schema([
            Forms\Components\TextInput::make('amount')
                ->mask(fn (Forms\Components\TextInput\Mask $mask) => $mask->money('$', ',', 2)),
        ]);
}
```

The first time that I click the "Create" button, both the `title` and `amount` fields are empty, as they should be. However, if I put values into both fields and click "Create & create another", only the `title` field resets to become empty again. It seems that the mask is preventing the `amount` field from resetting to be empty. One important thing to note is that, in this scenario, while the `amount` field appears to have a value (at least according to what is visible in the UI) when I submit the form the field is actually null.

I'm out of my league here, but I wonder if the solution might be as easy as updating the imask options via their `updateOptions()` method and then doing a field destroy via `mask.destroy()`? I'm stabbing in the dark here, but trying to be helpful.

### Steps to reproduce

1. Create a relation manager for a resource
2. Create a masked TextInput field for the related resource, like this:

```
{
    return $form
        ->schema([
            Forms\Components\TextInput::make('amount')
                ->mask(fn (Forms\Components\TextInput\Mask $mask) => $mask->money('$', ',', 2)),
        ]);
}
```

3. Click "attach" button, fill out fields and then click "Create & create another"
4. The `amount` field will retain the prior value while the `title` field will properly reset

### Relevant log output

_No response_

danharrin:
Fixes #1525.


danharrin:
Also fixes #1289.

PR #1536, opened 2022-02-09
feature: global notification api

ryangjchandler:
This plugin introduces a new global Notification API for Filament.

From any Livewire component rendered by Filament (any component on any `Page`), you can call the `Filament::notify()` method.

Here's an example from a RelationManager:

```php
public function afterSave()
{
    Filament::notify('success', 'Saved!');
}
```

This will tell the `NotificationManager` component to render this notification.

If you are redirecting inside of an action on a table or page, you probably want the notification to be displayed on the next page. The `Filament::notify()` method has a third parameter, `afterRedirect` that is `false` by default.

```php
ButtonAction::make(...)
    ->action(function () {
        Filament::notify('success', 'All done!', afterRedirect: true);

        return redirect()->to('/');
    })
```

> The old `Page::notify()` method that could be called from any `Page` in the admin has been updated to use this method too. There's no need to update your `$this->notify()` calls unless you've overwritten the method.


danharrin:
Yes! And relation managers & other modals now have notifications.

PR #1541, opened 2022-02-10
Added defer keyword in dayjs locale

wit3:
Added defer in tag script


danharrin:
What advantage does this give?

ryangjchandler:
@danharrin It'll fix a problem where the onload is being called at the wrong time I believe. I've seen this myself.

danharrin:
Thanks

PR #1547, opened 2022-02-10
Add Turkish (tr) translation files for admin, forms, tables and translatable packages

trk:
Add Turkish (tr) translation files for admin, forms, tables and translatable packages


danharrin:
Thank you :)

PR #1554, opened 2022-02-10
Enable modal on submit buttonAction

jvkassi:
Here is a demo about what this PR enables

```
class EditDevice extends EditRecord
{
    protected static string $resource = DeviceResource::class;

    protected function getSaveFormAction(): ButtonAction
    {
        return ButtonAction::make('save')
            ->requiresConfirmation()
            ->action('save'); // We must use action instead of submit for the modal to work
            // ->submit('save');
    }
    
}
```


danharrin:
I'm confused about how these caching changes allow you to submit the action differently?

Also, please remove the whitespace formatting changes from this PR.

jvkassi:
This code wasn't showing the confirmation modal.

```
class EditDevice extends EditRecord
{
    protected static string $resource = DeviceResource::class;

    protected function getSaveFormAction(): ButtonAction
    {
        return ButtonAction::make('save')
            ->requiresConfirmation()
            ->submit('save');
    }
    
}
```

I search in the code to find why this wasn't working. And i found those caching mechanism.
The HasAction train has this function

```
    protected function cacheActions(): void
    {
        $this->cachedActions = collect($this->getActions())
            ->mapWithKeys(function (Action $action): array {
                $action->livewire($this);

                return [$action->getName() => $action];
            })
            ->toArray();
    }
```

It enable to cache the action but only for the header  button actions in the forms on the bottom action buttons.
It's called here in the file views/components/page.blade.php

```

            <x-filament::header :actions="$this->getCachedActions()">
```

So i figured i should create a cache function for the bottom button actions.
like this

```
        <x-filament::form.actions :actions="$this->getCachedFormActions()" />
```

danharrin:
Ah ok. Not all pages have form actions, so maybe we should extract this into a `HasFormActions` trait instead?

jvkassi:
I was thinking about making it a default.
So i have added a default function in HasAction trait 'packages/admin/src/Pages/Concerns/HasActions.php' like this

<pre>
    protected function getFormActions(): array
    {
        return [];
    }
</pre>

I think it's okay like that.

danharrin:
Yeah, not every page has a form so this is not default behaviour. Please add it to a separate trait.

jvkassi:
Okay, now it should be okay.
I've added the trait by default in the UsesResourceForm.php

bezhanSalleh:
should be removed from `UsesResourceFrom.php`

danharrin:
Thanks :)

PR #1557, opened 2022-02-11
fix issue in repeater icons with rtl

maherelgamil:
Before
image

Now 
image

### DIFF
--- a/packages/forms/resources/views/components/repeater.blade.php
+++ b/packages/forms/resources/views/components/repeater.blade.php
@@ -23,7 +23,7 @@
 
                         @unless ($isItemDeletionDisabled() && ($isItemMovementDisabled() && ($loop->count <= 1)))
                             <div @class([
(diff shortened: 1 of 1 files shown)


danharrin:
How about the builder? Also, please ensure that this does not change the LTR styling, it looks like you're removing a border class? Also, keep all RTL classes together at the end of the class list.

maherelgamil:
thanks for this review @danharrin 

- keep all RTL classes together at the end of the class list. [done]

And, borders not removed I fixed it in the last commit.

image

danharrin:
And the builder? :) Have you checked that the border is still there in LTR? You've removed the `border-l` class and added an extra space.

maherelgamil:
@danharrin OMG, i am so sorry for that mistake, i fixed it 

Thanks 🙏

danharrin:
Thanks :)

PR #1561, opened 2022-02-11
Fix toolbar's height wrongly when scrollbar appears

minhkhoablieu:
Before
image
After
image


ryangjchandler (changes requested):

minhkhoablieu:
@danharrin 
Markdown editor isn’t affected by this bug.

danharrin:
Thanks

PR #1564, opened 2022-02-11
Allow custom classes in icon column

wit3:


danharrin:
Hm not sure about this. Would need to do the same everywhere else (for consistency) and I think we probably have cases where it would be a breaking change..

danharrin:
Decided this will be fine for the icon column. Thank you :)

Wit3:
Thanks

Wit3:
If i create a PR for BadgeColumn, same implementation, what do you think about?

danharrin:
Hmm, sure 👍

PR #1566, opened 2022-02-11
Update french translations

marcbelletre:
This fixes an HTML special char issue with the "no results" message in french. I also changed "mode nuit" to "mode sombre" which is more adequate and added a missing accent.


danharrin:
Thanks :)

PR #1568, opened 2022-02-11
Flip logout icon

mohamedsabil83:


danharrin:
I think we should avoid using a separate Tailwind customisation in this case and just use a standard class, as it's a bit of a breaking change for those using custom themes.

mohamedsabil83:
Applied the flip effect directly in the blade file and revert tailwind configuration

PR #1570, opened 2022-02-11
Reduce space between action's cell

swarakaka:
Before:
Bildschirmfoto 2022-02-11 um 16 48 37

After:
Bildschirmfoto 2022-02-11 um 16 49 48


danharrin:
How does this look for link actions? Button actions? That was the original reason it was such a large gap.

swara-mohammed:
The space between them is also very low, and they may not look very nice, but it is better than all that space. They may be separated by checking the action type condition.

PR #1581, opened 2022-02-12
Change to generate paragaphs when hitting return in RichEditor

titantwentyone:
Currently Trix generates divs when return is entered. I think the expectation is that paragraphs should be generated. This commit generates `p` tags instead. Code pulled from an example of a custom Trix config [here](https://github.com/lnu-norge/lokaler.lnu.no/pull/26/commits/fd56cf37a76f05e2888a2042358b66e414813c33).


danharrin:
Haha, this is a much more sensible default. Thank you!

zumbidoweb:
Shift + Enter should make a <br/>
But also make a <p>
I thinks is a big issue.

PR #1582, opened 2022-02-12
Added ability to specify name when uploading using SpatieMediaLibraryFileUpload

titantwentyone:
Provides ability to specify the name of a file to be stored in the `media` table. Usage:

```
SpatieMediaLibraryFileUpload::make('myfiles')
    ->collection('myfiles')
    ->usingName('the_name_of_my_file')

//we can now use:
$post->getFirstMediaUrl('myfiles', 'the_name_of_my_file')

```


danharrin:
Thanks, I've renamed this to `mediaName` since that's what Spatie uses internally.

titantwentyone:
Ah! Just seen that you saw the case of specifying null as default! Sorry I missed that! :)

DanjBethel:
@danharrin What if I wanted to set the name, using another form input field? How would something like that be possible using the "customProperties" handler?

danharrin:
Probably passing a closure and using `$get` like normal

PR #1584, opened 2022-02-12
feat(tables): added alignment support to table cell.

FurkanGM:


packages/tables/resources/views/components/cell.blade.php
```diff
@@ -1,12 +1,13 @@
...
+'action' => null,
+'name',
+'record',
+'shouldOpenUrlInNewTab' => false,
+'url' => null,
+'alignment' => null,
 ])
 
-<td {{ $attributes->class(['filament-tables-cell', 'dark:text-white' => config('tables.dark_mode')]) }}>
+<td {{ $attributes->class(['filament-tables-cell', 'dark:text-white' => config('tables.dark_mode'), "text-$alignment" => $alignment]) }}>
```
ryangjchandler, line 10: This won't work with Tailwind since it can't accurately determine what classes should be kept/removed in the purge. You'll have to use a match statement or `Arr::toCssClasses` instead. There's plenty of examples in the code base.

packages/tables/src/Columns/Concerns/HasAlignment.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Filament\Tables\Columns\Concerns;
+
+trait HasAlignment
+{
+    protected ?string $alignment = null;
+
+    public function alignment(?string $alignment): static
```
ryangjchandler, line 9: Can we also add some convenience methods like `alignLeft`, etc?

ryangjchandler (changes requested):

packages/tables/src/Columns/Concerns/HasAlignment.php
FurkanGM: done

packages/tables/resources/views/components/cell.blade.php
FurkanGM: it's now correctly?

danharrin:
Thanks :)

PR #1585, opened 2022-02-12
Dispatch an event when toggling dark mode

mohamedsabil83:


packages/admin/resources/views/components/layouts/app/topbar/user-menu.blade.php
```diff
@@ -35,8 +35,10 @@
 
                 if (this.theme === 'dark' && (! document.documentElement.classList.contains('dark'))) {
                     document.documentElement.classList.add('dark')
+                    $dispatch('dark-mode-changed', { mode: 'dark' })
```
ryangjchandler, line 38: I prefer `dark-mode-toggled` personally.

ryangjchandler (changes requested):

packages/admin/resources/views/components/layouts/app/topbar/user-menu.blade.php
mohamedsabil83: Good advice 👌🏽

ryangjchandler (changes requested):
I think we just need to fix some grammar in the docs and it'll be good to merge. Can imagine this will be handy for people who are building custom components etc. 

Will take a look at the docs tomorrow/Monday.

danharrin:
Thanks! I've removed the event payload object and just replaced it with a `dark` or `light` string.

mohamedsabil83:
You're welcome. Yes, that's better 👍🏽

PR #1587, opened 2022-02-13
Add malay translation (ms)

sumardi:
Added malay translation (ms) for admin, forms, tables and translatable packages.


danharrin:
Thanks!

PR #1596, opened 2022-02-13
remove default image caption in rich editor

mokhosh:


danharrin:
I think we need a method to enable this, since it breaks non-image uploads (which Trix doesn't preview).

mokhosh:
I agree, but I don't know where or how to add that :D

danharrin:
I've added the PHP method to enable this, but since the editor is loaded before the config can take effect, there is no longer any effect on the caption. Trix doesnt allow you to dynamically set the config AFAIK :(

So, I don't think this is possible. If you wish, play around and try and get this working again with the current PHP implementation. But I don't think it is possible. Let me know if you get somewhere and I can reopen the PR.

PR #1605, opened 2022-02-13
feature: support custom global search providers

ryangjchandler:
This pull request adds support for custom global search providers. This API will allow developers to change how the global search works and how results are generated.

In theory, somebody could implement a Scout provider for Filament and share it as a package or perhaps a completely custom search provider that uses an alternative service.

Here's how the API works:

```php
public function boot()
{
	Filament::globalSearchProvider(MyCustomSearchProvider::class);
}
```

`MyCustomSearchProvider` is a `class` that implements the `Filament\GlobalSearch\Contracts\GlobalSearchProvider` interface. This interface enforces the class to have a `getResults(string $query): ?array` method.

The array that gets returned should have the following shape:

```php
[
	{category} => [
		[
			'details' => [
				{label} => {value}...
			],
			'title' => {string},
			'url' => {string},
		]...
	]...
]
```

The `category` should be a string that is used to group the results. `details` is an array of `key => value` pairs that will display underneath the result's `title` in the dropdown. The `url` is where the user goes when they click the result.

### DIFF
--- a/packages/admin/src/GlobalSearch/DefaultGlobalSearchProvider.php
+++ b/packages/admin/src/GlobalSearch/DefaultGlobalSearchProvider.php
@@ -6,9 +6,9 @@ use Filament\Facades\Filament;
 
 class DefaultGlobalSearchProvider implements Contracts\GlobalSearchProvider
 {
-    public function getResults(string $query): ?array
+    public function getResults(string $query): ?GlobalSearchResults
     {
-        $results = [];
+        $builder = GlobalSearchResults::make();
 
         foreach (Filament::getResources() as $resource) {
             if (! $resource::canGloballySearch()) {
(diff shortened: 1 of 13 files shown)


danharrin:
Could we maybe get some tests for the global search as part of this? Since the API we're using is solid now.

packages/admin/src/GlobalSearch/DefaultGlobalSearchProvider.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Filament\GlobalSearch;
+
+use Filament\Facades\Filament;
+
+class DefaultGlobalSearchProvider implements Contracts\GlobalSearchProvider
+{
+    public function getResults(string $query): ?array
```
SirMishaa, line 9: Small question, why the return type is nullable? It shouldn't be always an array?

ryangjchandler: The original logic could potentially return null.

ryangjchandler:
The API has now been updated. We've introduced a new `GlobalSearchResults` builder object that can be used instead of the flimsy array syntax.

```php
GlobalSearchResults::make()
	->category('Posts', [
		new GlobalSearchResult(title: 'foo', url: '#', details: []),
	])
```

PR #1609, opened 2022-02-14
fix: badge/tags column alignment, header cell alignment

jacksleight:
This PR fixes these issues:

* Alignment in badge and tags columns (these  columns use flexbox so `text-align` doesn't work as intended)
* Alignment in table header cells
* The header cells have also been tweaked so the button fills the cell, so the whole cell is clickable when sorting

Before:

Screenshot 2022-02-14 at 15 03 32

After:

Screenshot 2022-02-14 at 15 03 42


danharrin:
Thanks!

PR #1611, opened 2022-02-14
feature: add prefix/postfix support and url link feature to select fields

jacksleight:
This PR adds prefix/postfix support to select fields.

My use case for this is adding links to related records when using a BelongsToSelect, eg:

Screenshot 2022-02-14 at 15 42 02

Notes:

* prefix/postfix properties and methods have been moved to a new `HasAffixes` trait
* `getPrefixLabel` and `getPostfixLabel` can now return more than just a string, so that an `HtmlString` object may be used
* As the `<select>` is no longer the root element the `filament-forms-select-component` class has been moved up to the new wrapper `<div>`. This is potentially a breaking change, but it does keep it consistent with other fields such as text-input.
    * Also the searchable version didn't include `filament-forms-select-component` at all, which this fixes
* I've added `whitespace-nowrap` to the searchable select label, which helps with longer values and is consistent with how a regular select works
* I have added a `suffix` alias for `postfix`, as I think that's a bit more intuitive name, but no worries if you want that removed

### DIFF
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -30,6 +30,8 @@ class Select extends Field
 
     protected string | Closure | null $searchPrompt = null;
 
+    protected string | Closure | null $url = null;
+
     protected function setUp(): void
     {
         parent::setUp();
@@ -114,6 +116,13 @@ class Select extends Field
         return $this;
     }
 
+    public function url(string | Closure | null $url): static
+    {
+        $this->url = $url;
+
+        return $this;
+    }
+
     public function getOptionLabel(): ?string
     {
         return $this->evaluate($this->getOptionLabelUsing, [
@@ -151,6 +160,11 @@ class Select extends Field
         }
 
         return $results;
+    }
+
+    public function getUrl(): ?string
+    {
+        return $this->evaluate($this->url);
     }
 
     public function isOptionDisabled($value, string $label): bool
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -11,6 +11,7 @@ class Select extends Field
     use Concerns\HasOptions;
     use Concerns\HasPlaceholder;
     use Concerns\HasAffixes;
+    use Concerns\CanOpenUrl;
 
     protected string $view = 'forms::components.select';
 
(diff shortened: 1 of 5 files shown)


jacksleight:
> @danharrin I wonder if linking from a `BelongsToSelect` is something we should support anyway?

It's quite useful, essentially the reverse of a Relation Manager (when set up to link directly to record pages), lets you jump back and fourth very easily.

ryangjchandler:
@jacksleight Yeah, I could do with this myself on an app tbh. The best way I think would be a `url()` method on the `BelongsToSelect` field that shows an icon or button.

jacksleight:
I could try adding that to this PR if you like?

ryangjchandler:
@jacksleight Feel free to give it a go - I'd definitely see it as a worthy addition.

jacksleight:
Example of how it all looks:

Screenshot 2022-02-14 at 17 02 38

packages/forms/src/Components/Select.php
```diff
@@ -30,6 +30,8 @@ class Select extends Field
 
     protected string | Closure | null $searchPrompt = null;
 
+    protected string | Closure | null $url = null;
```
ryangjchandler, line 33: The Tables package has a `CanOpenUrl` trait, might be worth making a similar one for the forms package.

ryangjchandler (changes requested):

jacksleight:
Another feature that would be really handy (but probably way beyond the scope of this PR) would be the ability to create a new related record via a `BelongsToSelect`, just like the relation manager.

saade:
> Another feature that would be really handy (but probably way beyond the scope of this PR) would be the ability to create a new related record via a `BelongsToSelect`, just like the relation manager.

Yeah! that would be awesome, i asked about this a time ago [Here](https://discord.com/channels/883083792112300104/883085267383226478/936432730026172497) and Dan replied [Here](https://discord.com/channels/883083792112300104/883085267383226478/936524755048214588) i think that could be easy (maybe) with the new resource modals feature

danharrin:
Thanks for your work on this, but I'm going to remove the `url()` feature as I want to approach this differently in the future, so that we can support modal create / edit forms for select fields 👍

PR #1617, opened 2022-02-14
Fix belongstoselect styles

felipe-balloni:
lines 78, 86: to match background, text color and selectable with disabled styles;

line 85: if remove absolute the caret position goes to center when in edit form;

line 87: to correct ring focus at right that cutting a little;
focus-cutting

lines 109 to match position and size with default select;


danharrin:
Thanks!

PR #1625, opened 2022-02-15
Update of Polish translations

lolen:
Adding missing translation strings.

ps. Maybe you should use some external translation management system?


danharrin:
Thanks. I want to build a CLI for translation management. Haven't found any other good systems yet.

Issue #1410, opened 2022-02-01
Page modal action: Undefined array key "mountedActionForm"


tanthammar:
### Package

filament/filament

### Package Version

v2.9.16

### Laravel Version

v8.81.0

### Livewire Version

v2.10.1

### PHP Version

v8.1.2

### Bug description

Trying to add a modal action button with confirmation on a custom admin Page.
No problem rendering the page, it is when I click the generated button I get an error:
```shell 
Undefined array key "mountedActionForm"
```

![image](https://user-images.githubusercontent.com/21239634/151954773-3c253e3f-93e4-4342-8bf3-59cd3064cea0.png)

### Steps to reproduce

Page:
```php
<?php

namespace App\Filament\Pages;

use Filament\Forms\ComponentContainer;
use Filament\Pages\Actions\ButtonAction;
use Filament\Pages\Page;
use Illuminate\Support\Facades\Auth;
use Filament\Forms\Components\TextInput;

/**
 * @property ComponentContainer|mixed|null $passwordForm
 * @property ComponentContainer|mixed|null $userNameEmailForm
 */
class Profile extends Page
{
    protected static ?string $navigationIcon = 'heroicon-o-user';
    protected static ?string $navigationGroup = 'Account';
    protected static string $view = 'user.profile';
    public string $password = '';
    public string $deleteMsg = '';

    protected function getForms(): array
    {
        return [
            'userNameEmailForm' => $this->makeForm()
                ->schema([/* name email fields */])
                ->model(Auth::user()),
            'passwordForm' => $this->makeForm()
                ->schema([/* psw fields */])
                ->model(Auth::user()),
        ];
    }

    protected function getActions(): array
    {
        return [
            ButtonAction::make(__('Delete account'))
                ->action('deleteUser')
                ->form([
                        TextInput::make($password)
                            ->label('Password')
                            ->password()
                            ->required()
                            ->currentPassword()
                            ->autocomplete('off')
                ])
                ->color('danger')
                ->icon('heroicon-o-key')
                ->requiresConfirmation(),
        ];
    }

    private function deleteUser()
    {
        //return route()
    }
}
```

view
```blade
<x-filament::page>
    @if($deleteMsg)
        <div class="rounded bg-danger text-white p-6">
            <p>{{ $deleteMsg }}</p>
        </div>
    @endif
    <form>
        {{ $this->userNameEmailForm }}
        <!-- submit btn -->
    </form>
    <form>
        {{ $this->passwordForm }}
[...]

PR #1635, opened 2022-02-16
Fixes issue #1410 update

romaldyminaya:
I've realized that I did try to edit the concern behavior instead of the main page component.


danharrin:
Thanks, but I'm going to solve this a different way.

PR #1642, opened 2022-02-16
fix: year input allowing more or less than 4 chars

### LINKED ISSUE #1507: When selecting year first in datepicker, it breaks
### Package

filament/forms

### Package Version

v2.10.3

### Laravel Version

v8.82.0

### Livewire Version

v2.10.1

### PHP Version

8.0

### Bug description

When you have a datepicker and you click the year first on mobile. it breaks

### Steps to reproduce

_No response_

### Relevant log output

_No response_

saade:
this fixes #1507

- The year input was allowing to input more or less than 4 numbers, which strangely causes errors only on mobile.
- Use `inputmode="numeric"` to open correct keyboard

### DIFF
--- a/packages/forms/resources/js/components/date-time-picker.js
+++ b/packages/forms/resources/js/components/date-time-picker.js
@@ -116,10 +116,11 @@ export default (Alpine) => {
                 })
 
                 this.$watch('focusedYear', () => {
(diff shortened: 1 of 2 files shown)


danharrin:
Thanks!

PR #1655, opened 2022-02-18
Adds `Filtering` ability to stats overview widget

bezhanSalleh:
This PR implements the Filtering ability for stats overview widgets.
Now the value, label, description, icon, color and chart can react to the filter based on the devs preference.

https://user-images.githubusercontent.com/10007504/154627530-f25bf7be-9b03-4ede-8dbc-2eddcef12b39.mov


bezhanSalleh:
@danharrin the select design might not be good and includes this bit `style="padding-block: 4px"` so if you have any suggestions do let me know.

danharrin:
Hey! Thanks for your work on this.

Firstly, please remove the "action" parts of this. We should instead store the filter values on the Livewire component, in an `$filters` array property. The keys of the array should be the keys from the `getCards()` array, and the value should be the filter value for that card.

Also, please replace inline styles with Tailwind JIT classes if you need custom values 👍

bezhanSalleh:
> Hey! Thanks for your work on this.
> 
> Firstly, please remove the "action" parts of this. We should instead store the filter values on the Livewire component, in an `$filters` array property. The keys of the array should be the keys from the `getCards()` array, and the value should be the filter value for that card.
> 
> Also, please replace inline styles with Tailwind JIT classes if you need custom values 👍

I have removed the `action` part. didn't like tbh.
The filter values are stored in a $filters array on the component.
Since we already had `id` for each card I just wired it to the `$filters` property `wire:model="filters.{{$id}}"`.
Just be declaring the `$filters` property everything works. 
Now if the user wants to set a default for a card's `->filters()` the can do so by adding the card id and default filters value in the `$filters` property like so.

```php
    public ?array $filters = ['days' => 'y'];

    protected function getCards(): array
    {
        return [
            Card::make('Unique views', '192.1k')
                ->id('days')
                ->filters([
                    't' => 'Today',
                    'y' => 'Yesterday',
                    'lw' => 'Last Week'
                ])
                ->chart([rand(1,10),rand(100,11),rand(10,1),rand(11,100)])
                ->chartColor('primary'),
            Card::make('Bounce rate', '21%')
                ->id('months')
                ->filters(['jan','feb','march'])
                ->chart([rand(1,100),rand(1000,101),rand(100,1),rand(101,1000)])
                ->chartColor('success'),
            Card::make('Average time on page', '3:12')
                ->filters(['blue','green','black']),
        ];
  }
  ```
  I have also added the ability to change the chart data by dispatching a browser event with the card id, if the users choose so.

bezhanSalleh:
Will make the last part like what we have for `ChartWidgets`.
Should've done it already but been jumping between alpine and livewire codebases alot lately, got 'em moxed up😇

danharrin:
As discussed on Discord, this probably isn't the best solution and we should reserve this functionality for custom widgets only, for now.

PR #1658, opened 2022-02-18
feat: Allow customization for BelongsToSelect label

pxlrbt:
As discussed on Discord this allows to easily set a label for the BelongsToSelect without overwriting the whole `getSearchResultsUsing()` and `getOptionLabelUsing()` callbacks.

**Outlook**
This could also applied to `BelongsToManyMultiSelect` and `BelongsToManyCheckboxList `. Both are missing the `searchable()` method or `$searchColumns` property this feature relies on. While it makes sense to add this to `BelongsToManyMultiSelect` there is no search on `BelongsToManyCheckboxList` and we need to define the fields for the SELECT statement differently.


pxlrbt:
Thanks for your additions and merging!

PR #1659, opened 2022-02-18
Added alignment config to table row actions.

ap1969:
- Added left/right/center and default options to align table row actions.
- Config options added to filament config file.
- Table docs updated.


danharrin:
The config for this actually needs moving to the tables package config file, as the admin panel is not always installed when you use tables in your apps.

dmitrijsmihailovs:
Looks like there is a typo in docs. In config it is called "action_alignment" but in docs just "alignment"

ap1969:
> @ap1969 I've made some changes for you so we can get this one released. @danharrin Any problems?

Thanks guys. I've been heads-down in the day job so didn't get a chance to follow-up on this.

PR #1672, opened 2022-02-21
Fix select listboxes z-index

marcbelletre:
This fixes a small bug causing the Select and Multiselect listboxes to appear behind the "add between" button when used within a Builder field.

Capture d’écran 2022-02-21 à 11 57 38


danharrin:
Thank you!

PR #1676, opened 2022-02-21
[1.X] fix: added record prefix to same validation

chrisAlkemy:
The "same" validation has an error because the object "record" doesn't exist


danharrin:
Thanks!

PR #1684, opened 2022-02-22
Cache chart data and stats cards to prevent multiple executions

dmandrade:
The getData() method on ChartWidget was being called twice without cache, the same occurs in the getCards() in StatsOverviewWidget which is called in 4 different places.

With this PR I called the respective methods in component mount() to cache data and reuse the result where needed using a different method: getCachedCards in StatsOverview and getCachedData in Chart.


danharrin:
Thank you

PR #1689, opened 2022-02-22
Locale filepond

dmandrade:
I load the filepond locale as a JavaScript module using <script type="module"> (https://caniuse.com/es6-module) and dispatch "filepond-locale-updated" event with the json data.

The field component listens for the event and calls pond.setOptions() to update locale data.

pt_BR
image

zh_CN
image

fr
image

Tested using the filament-demo


danharrin:
Thanks!

PR #1693, opened 2022-02-23
feature: selectable table columns

jszobody:
This is a first rough stab at adding column selection to tables. This is what it looks like:

CleanShot_2022-02-22_at_18 36 20

To make a column selectable, add `->selectable()` when defining a column. If any columns in a table are selectable, the action icon appears on the table.

I very closely followed the way filters are implemented, copied and pasted quite a bit of that code. 

I'm looking for input on a number of items:

1. Naming things is hard. I have terms like `ColumnSelection` and `CanBeSelectable` and `isColumnSelectionVisible` and I'm really not sure these are well named. Is there a better single-word term for this "column selectability" behavior anyone can think of?
2. I decided not to use any of the existing visible/hidden functionality of columns. It seems we should leave that for situations where a developer wants full control over whether a column appears at all or not. This selectable stuff is completely separate.
3. Currently when a column is `->selectable()` it is hidden by default until you select it. Need a way to also indicate that it should default to visible, thoughts on a good method name for that? 
4. Does it make sense to add this behavior on a column-by-column basis? I initially thought so, but now I'm wondering if this is overkill. Would it make sense to just have a table-level flag that simply makes ALL columns selectable?
5. Changing column selections triggers a livewire request. We COULD make this entirely client-side instead, by rendering all table columns and controlling the visibility with alpinejs. This would be faster. It would mean though that we can't save selections server-side, say, in session. Thoughts?
6. Speaking of saving column selections: _should_ this be handled in session server-side? Should we persist to localstorage client-side? (Related: are there any plans for a database table where filament can store user preferences/settings/etc?)
7. I'll need help with the language string, right now it's still using `tables::table.buttons.filter.label` which obviously will need to change.

Feedback welcome. =)


packages/tables/src/Columns/Concerns/CanBeSelectable.php
```diff
@@ -0,0 +1,26 @@
...
+
+use Illuminate\Support\Str;
+
+trait CanBeSelectable
+{
+    protected bool $isSelectable = false;
+
+    public function selectable(bool | array $condition = true): static
+    {
+        if (is_array($condition)) {
```
ryangjchandler, line 13: This can be a ternary instead of if/else blocks

ryangjchandler: Can we also accept a `Closure`?

packages/tables/src/Columns/Concerns/CanBeSelectable.php
```diff
@@ -0,0 +1,26 @@
...
+        } else {
+            $this->isSelectable = $condition;
+        }
+
+        return $this;
+    }
+
+    public function isSelectable(): bool
+    {
+        return $this->isSelectable;
```
ryangjchandler, line 24: If we add support for closures ^ then we need to send this through `$this->evaluate()`

packages/tables/src/Concerns/CanSelectColumns.php
```diff
@@ -0,0 +1,55 @@
...
+trait CanSelectColumns
+{
+    public $tableColumnSelections = [];
+
+    public function isTableColumnSelectable(): bool
+    {
+        return count($this->getCachedSelectableColumns());
+    }
+
+    public function getTableColumnSelectionForm(): Forms\ComponentContainer
```
ryangjchandler, line 23: `ComponentContainer` is already in the namespace so can use that instead of relative

ryangjchandler (changes requested):

ryangjchandler:
Regarding storage of selected/visible columns, I'd opt for `$persist` with Alpine personally so that it exceeds the lifetime of the user's session. 

I'd prefer to see a table-level flag that makes columns selectable, as well as the method on the column that can be used to disable the functionality for that particular column. The inverse of what you have now basically.

dmandrade:
I think the behavior is like display toggle, for me "selectability" refers to another type of action other than showing/hiding something.

I like the option of having a table-level method to enable the function globally and two column-level methods to change the default display and disable the feature.

ryangjchandler:
Toggleable doesn't roll of the tongue...

jszobody:
Ok so let's think through the DX, if we flip this to a table-level flag. 

```php
$table
    ->columns(...)
    ->filters(...)
    ->selectable() // ??
```

My thinking is that by making the table `selectable` it will, by default, _hide_ columns. I'm guessing that if a developer wants this feature, it's because they have a ton of columns. Let's treat 'not-selected' as the default for any column that isn't otherwise specified. 

Perhaps the `selectable` method would take an integer of how many columns to display for starters:

```php
public function selectable(int $initiallySelected = 1)
```

As for the individual column option, perhaps...

```php
Tables\Columns\TextColumn::make('name')->alwaysSelected(),
Tables\Columns\TextColumn::make('email')->defaultSelected(),
```

Adding `->defaultSelected()` to a column would then make it initially selected by default, and adding `->alwaysSelected()` would lock the column in a selected state, cannot be unselected by the user.

Thoughts?

dmandrade:
> Toggleable doesn't roll of the tongue

Agree. I just use word "toggle" for enable/disable something

Example
`
$table->enableColumnToggle(bool | Closure)
$column->canToggle(bool | Closure)
$column->defaultToggleState(bool | Closure)
`

jszobody:
Ok this has been significantly reworked, based on my [comment yesterday](https://github.com/laravel-filament/filament/pull/1693#issuecomment-1048775175). 

The simplest way to make use of the new column selection is to add the `selectabe()` method when defining a table:

```php
$table
    ->columns(...)
    ->filters(...)
    ->selectable();
```

By default, this will auto-select the first column in the table, and hide the rest. You can pass in an integer to this method to change the default number of columns selected. `->selectable(3)` will show the first three columns, for example.

You can also control selection on an individual column if you want. Two methods are available:

```php
Tables\Columns\TextColumn::make('name')->alwaysSelected(),
Tables\Columns\TextColumn::make('email')->selected(),
Tables\Columns\TextColumn::make('phone')
```

The `selected()` method makes the column selected by default. The `alwaysSelected()` method also locks this, such that the user cannot un-select. This is what the above example looks like:

CleanShot 2022-02-24 at 11 25 33@2x

Note: if you explicitly set selection on any individual columns, the table-level `selectable()` count is ignored, column-level selection rules take precedence.

@dmandrade While I still have mixed feelings about the 'selectable' verbiage, I want to try and keep to a terse term if possible, rather than multi-word methods like `enableColumnToggle`. Just a preference I suppose. 

@ryangjchandler Most of the code you commented on has now changed, feel free to re-visit. However, before getting too detailed with the code specifics, I'd like to hear your thoughts on this updated API and whether you like the high-level approach. I still haven't focused much on closure arguments and such, wanting to get the big picture in place first.

packages/tables/src/Columns/Concerns/CanBeSelectable.php
jszobody: This method doesn't exist on the column anymore

danharrin:
Hey, I've just had a chance to review this and quickly fix some formatting things. Overall, its great, you're definitely on the right track.

However. I dont think the term "select" or "selectable" is suitable here. Since that could easily get confused with selecting rows that we used for bulk actions. I'm not sure of an alternative though... tbh I think I prefer "toggleable" but I'm open to other suggestions. Maybe `toggleColumns()` `alwaysToggled()` etc.

Also, persistence of toggled columns between page reloads didn't seem to work for me, not sure if its a bug or you haven't thought about it yet.

joshbenham:
> Hey, I've just had a chance to review this and quickly fix some formatting things. Overall, its great, you're definitely on the right track.
> 
> However. I dont think the term "select" or "selectable" is suitable here. Since that could easily get confused with selecting rows that we used for bulk actions. I'm not sure of an alternative though... tbh I think I prefer "toggleable" but I'm open to other suggestions. Maybe `toggleColumns()` `alwaysToggled()` etc.
> 
> Also, persistence of toggled columns between page reloads didn't seem to work for me, not sure if its a bug or you haven't thought about it yet.

Could you use Visible? Then you could have toggleVisibility() and alwaysVisible().

danharrin:
I think `toggleVisibility()` is clear, yeah. But we need to make sure we don't get mixed up with existing `visible*` column methods that deal with responsive columns.

nkeena:
Anything I can do to help get this PR merged? I would love to use this feature in a current project

danharrin:
> Anything I can do to help get this PR merged? I would love to use this feature in a current project

Not sure if @jszobody wants us to take over or not, he's probably very busy. If you're interested though, you could maybe clone his branch and rename any terminology from "select" to "toggleVisibility". Might require quite a few changes but all of the real work to get the feature working has been done.

jszobody:
Sorry folks, yes I've been out of town lately and too busy to revisit this. I won't be able to put much effort in until after next week.

There are a few open questions in my mind that perhaps @danharrin can help clarify for me.

1) Since there are already `visible()` methods on columns, it seems super confusing to use `visibility` for this behavior. Don't you think future devs are going to get those mixed up?

2) If you're sure you like visibility, I'll go with it. But I'm unclear on the desired API. Would `toggleVisibility()` be at the table or column level? `$table->columns(...)->toggleVisibility()` feels awkward to me. I have a table-level switch to enable this behavior in general, and then a column-level switch to control whether it is toggled/selected by default or not. Interested in nailing down the precise API you think would be developer-friendly, before refactoring the code again.

3) I have not implemented any persistence yet. Can you provide me some thoughts on how you think that should be handled:

    a) Server-side in session. Easiest to implement, obviously only preserves column selections within a specific session.

    b) Client-side in local storage. If we do this though, it might mean revisiting how this entire feature is currently designed. Right now it's the _server_ where default toggling is specified. That means we'd initially render the table with certain columns toggled, _then_ check local storage for stored column preferences, then trigger an immediate livewire request to modify the column toggles. That immediate second trip to the server seems undesirable, no? If we want to store the column selections in local storage, I think that might suggest handling the entire interaction of toggling/untoggling columns client-side. The server would render ALL columns, and in Alpine we'd hide any columns we don't want to see, and toggle those entirely client-side with no more LW trips to the server at all, and updating local storage on all toggle changes. 

    c) Some kind of user-specific preference storage on the server. This is frankly my favorite, but I don't think Filament has anything to handle user preferences currently, and would need to introduce that first. If you thought that storing user preferences is something Filament needs to have in the near future, might be worth building that first (DB table migration, presumably) so then the table column selections can be stored there.

ryangjchandler:
Okay, this is now ready for testing and review.

Picking up where @jszobody left off, I've done the following:
1. Renamed all references of "selectable" / "selected" to "toggleable" / "toggled".
2. Removed all changes related to the admin package, all changes are scoped to the tables package.
3. Added persistence of column toggle states to the session, using a similar system to the pagination logic.
4. Moved the logic of toggling columns from the table itself to the column, no longer need to enable at the table level.

Will let @danharrin tear this code apart, but functionality wise it's all up to scratch and close to what has been discussed above.

packages/tables/resources/views/index.blade.php
```diff
@@ -184,6 +185,13 @@ class="mr-2"
                                 class="shrink-0"
                             />
                         @endif
+
+                        @if ($isColumnToggleFormVisible)
+                            <x-tables::toggleable
+                                :form="$getColumnToggleForm()"
```
ryangjchandler, line 191: Think we need to pass in a width here based on the number of columns the form has...

ryangjchandler (changes requested):

packages/tables/resources/views/index.blade.php
danharrin: Maybe it will be ok for just 1 column in the columns list for now, until it we get other requests. People try to minimise the number of columns in tables anyway to avoid scroll.

packages/tables/src/Columns/Concerns/CanBeToggled.php
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Filament\Tables\Columns\Concerns;
+
+use Closure;
+
+trait CanBeToggled
+{
+    protected bool | Closure $isToggleable = false;
+
+    protected bool $isToggledByDefault = false;
```
danharrin, line 11:
```suggestion
    protected bool | Closure $isToggledByDefault = false;
```

packages/tables/src/Columns/Concerns/CanBeToggled.php
```diff
@@ -0,0 +1,42 @@
...
+
+    public function toggleable(bool | Closure $condition = true, bool $isToggledByDefault = false): static
```
danharrin, line 13:
```suggestion
    public function toggleable(bool | Closure $condition = true, bool | Closure $isToggledByDefault = false): static
```

packages/tables/src/Columns/Concerns/CanBeToggled.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        $this->isToggleable = $condition;
+        $this->isToggledByDefault = $isToggledByDefault;
+
+        return $this;
+    }
+
+    public function toggledByDefault(bool $condition = true): static
```
danharrin, line 21:
```suggestion
    public function toggledByDefault(bool | Closure $condition = true): static
```

packages/tables/src/Concerns/CanToggleColumns.php
```diff
@@ -0,0 +1,95 @@
...
+
+use Filament\Forms\ComponentContainer;
+use Filament\Forms\Components\Checkbox;
+
+/**
+ * @property ComponentContainer $toggleColumnForm
+ */
+trait CanToggleColumns
+{
+    public array $toggledColumns = [];
```
danharrin, line 13:
```suggestion
    public array $toggledTableColumns = [];
```

packages/tables/src/Columns/Concerns/CanBeToggled.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public function isToggleable(): bool
+    {
+        return (bool) $this->evaluate($this->isToggleable);
+    }
+
+    public function isToggled(): bool
+    {
+        return $this->getTable()->getLivewire()->isToggled($this->getName());
```
danharrin, line 40:
```suggestion
        return $this->getTable()->getLivewire()->isTableColumnToggled($this->getName());
```

packages/tables/src/Concerns/CanToggleColumns.php
```diff
@@ -0,0 +1,95 @@
...
+
+    public function prepareToggledColumns(): void
```
danharrin, line 15:
```suggestion
    public function prepareToggledTableColumns(): void
```

packages/tables/src/Concerns/CanToggleColumns.php
```diff
@@ -0,0 +1,95 @@
...
+    {
+        $this->toggledColumns = session()->get(
```
danharrin, line 17:
```suggestion
        $this->toggledTableColumns = session()->get(
```

packages/tables/src/Concerns/CanToggleColumns.php
```diff
@@ -0,0 +1,95 @@
...
+            $this->getColumnToggleFormStateSessionKey(),
```
danharrin, line 18:
```suggestion
            $this->getTableColumnToggleFormStateSessionKey(),
```

packages/tables/src/Concerns/CanToggleColumns.php
```diff
@@ -0,0 +1,95 @@
...
+            $this->getDefaultColumnToggleState()
```
danharrin, line 19:
```suggestion
            $this->getDefaultTableColumnToggleState()
```

packages/tables/src/Concerns/CanToggleColumns.php
```diff
@@ -0,0 +1,95 @@
...
+        );
+    }
+
+    protected function getDefaultColumnToggleState(): array
```
danharrin, line 23:
```suggestion
    protected function getDefaultTableColumnToggleState(): array
```

packages/tables/src/Concerns/CanToggleColumns.php
```diff
@@ -0,0 +1,95 @@
...
+                continue;
+            }
+
+            $state[$column->getName()] = true;
+        }
+
+        return $state;
+    }
+
+    public function updatedToggledColumns(): void
```
danharrin, line 38:
```suggestion
    public function updatedToggledTableColumns(): void
```

packages/tables/src/Concerns/CanToggleColumns.php
```diff
@@ -0,0 +1,95 @@
...
+    {
+        session()->put([
+            $this->getColumnToggleFormStateSessionKey() => $this->toggledColumns,
```
danharrin, line 41:
```suggestion
            $this->getTableColumnToggleFormStateSessionKey() => $this->toggledTableColumns,
```

packages/tables/src/Concerns/CanToggleColumns.php
```diff
@@ -0,0 +1,95 @@
...
+        ]);
+    }
+
+    public function hasToggleableColumns(): bool
```
danharrin, line 45:
```suggestion
    public function hasToggleableTableColumns(): bool
```

packages/tables/src/Concerns/CanToggleColumns.php
```diff
@@ -0,0 +1,95 @@
...
+                continue;
+            }
+
+            return true;
+        }
+
+        return false;
+    }
+
+    public function getColumnToggleForm(): ComponentContainer
```
danharrin, line 58:
```suggestion
    public function getTableColumnToggleForm(): ComponentContainer
```

packages/tables/src/Concerns/CanToggleColumns.php
```diff
@@ -0,0 +1,95 @@
+<?php
+
+namespace Filament\Tables\Concerns;
+
+use Filament\Forms\ComponentContainer;
+use Filament\Forms\Components\Checkbox;
+
+/**
+ * @property ComponentContainer $toggleColumnForm
```
danharrin, line 9:
```suggestion
 * @property ComponentContainer $toggleTableColumnForm
```

packages/tables/src/Concerns/CanToggleColumns.php
```diff
@@ -0,0 +1,95 @@
...
+
+            return true;
+        }
+
+        return false;
+    }
+
+    public function getColumnToggleForm(): ComponentContainer
+    {
+        return $this->toggleColumnForm;
```
danharrin, line 60:
```suggestion
        return $this->toggleTableColumnForm;
```

packages/tables/src/Concerns/CanToggleColumns.php
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    protected function getColumnToggleFormSchema(): array
```
danharrin, line 63:
```suggestion
    protected function getTableColumnToggleFormSchema(): array
```

packages/tables/src/Concerns/CanToggleColumns.php
```diff
@@ -0,0 +1,95 @@
...
+            }
+
+            $schema[] = Checkbox::make($column->getName())
+                ->label($column->getLabel());
+        }
+
+        return $schema;
+    }
+
+    protected function getColumnToggleFormColumns(): int | array
```
danharrin, line 79:
```suggestion
    protected function getTableColumnToggleFormColumns(): int | array
```

danharrin, line 79:
```suggestion
    protected function getTableColumnToggleFormColumns(): int | array
```

packages/tables/resources/views/index.blade.php
```diff
@@ -7,6 +7,7 @@
     $heading = $getHeading();
     $isSearchVisible = $isSearchable();
     $isFiltersDropdownVisible = $isFilterable();
+    $isColumnToggleFormVisible = $hasToggleableColumns();
```
danharrin, line 10:
```suggestion
    $isColumnToggleFormVisible = $hasToggleableTableColumns();
```

packages/tables/resources/views/index.blade.php
danharrin, line 191:
```suggestion
                                :form="$getTableColumnToggleForm()"
```

packages/tables/src/Concerns/CanToggleColumns.php
```diff
@@ -0,0 +1,95 @@
...
+    {
+        return 1;
+    }
+
+    public function isToggled(string $name): bool
```
danharrin, line 84:
```suggestion
    public function isTableColumnToggled(string $name): bool
```

packages/tables/src/Concerns/CanToggleColumns.php
```diff
@@ -0,0 +1,95 @@
...
+    {
+        return isset($this->toggledColumns[$name]) && (bool) $this->toggledColumns[$name];
```
danharrin, line 86:
```suggestion
        return isset($this->toggledTableColumns[$name]) && (bool) $this->toggledTableColumns[$name];
```

packages/tables/src/Concerns/CanToggleColumns.php
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    public function getColumnToggleFormStateSessionKey(): string
```
danharrin, line 89:
```suggestion
    public function getTableColumnToggleFormStateSessionKey(): string
```

packages/tables/src/Concerns/InteractsWithTable.php
```diff
@@ -42,6 +43,8 @@ public function bootedInteractsWithTable(): void
 
         $this->cacheTableFilters();
         $this->getTableFiltersForm()->fill($this->tableFilters);
+
+        $this->prepareToggledColumns();
```
danharrin, line 47:
```suggestion
        $this->prepareToggledTableColumns();
```

packages/tables/src/Concerns/InteractsWithTable.php
```diff
@@ -113,6 +116,11 @@ protected function getTableForms(): array
                 ->columns($this->getTableFiltersFormColumns())
                 ->statePath('tableFilters')
                 ->reactive(),
+            'toggleColumnForm' => $this->makeForm()
```
danharrin, line 119:
```suggestion
            'toggleTableColumnForm' => $this->makeForm()
```

packages/tables/src/Concerns/InteractsWithTable.php
```diff
@@ -113,6 +116,11 @@ protected function getTableForms(): array
...
+                ->schema($this->getColumnToggleFormSchema())
```
danharrin, line 120:
```suggestion
                ->schema($this->getTableColumnToggleFormSchema())
```

packages/tables/src/Concerns/InteractsWithTable.php
```diff
@@ -113,6 +116,11 @@ protected function getTableForms(): array
...
+                ->columns($this->getColumnToggleFormColumns())
```
danharrin, line 121:
```suggestion
                ->columns($this->getTableColumnToggleFormColumns())
```

packages/tables/src/Concerns/InteractsWithTable.php
```diff
@@ -113,6 +116,11 @@ protected function getTableForms(): array
...
+                ->statePath('toggledColumns')
```
danharrin, line 122:
```suggestion
                ->statePath('toggledTableColumns')
```

packages/tables/src/Contracts/HasTable.php
```diff
@@ -52,4 +52,10 @@ public function isTableFilterable(): bool;
     public function isTableSearchable(): bool;
 
     public function isTableSelectionEnabled(): bool;
+
+    public function hasToggleableColumns(): bool;
```
danharrin, line 56:
```suggestion
    public function hasToggleableTableColumns(): bool;
```

packages/tables/src/Contracts/HasTable.php
```diff
@@ -52,4 +52,10 @@ public function isTableFilterable(): bool;
...
+
+    public function isToggled(string $name): bool;
```
danharrin, line 58:
```suggestion
    public function isTableColumnToggled(string $name): bool;
```

packages/tables/src/Contracts/HasTable.php
```diff
@@ -52,4 +52,10 @@ public function isTableFilterable(): bool;
...
+
+    public function getColumnToggleForm(): ComponentContainer;
```
danharrin, line 60:
```suggestion
    public function getTableColumnToggleForm(): ComponentContainer;
```

packages/tables/src/Table.php
```diff
@@ -326,6 +341,11 @@ public function isSearchable(): bool
         return $this->getLivewire()->isTableSearchable();
     }
 
+    public function hasToggleableColumns(): bool
+    {
+        return $this->getLivewire()->hasToggleableColumns();
```
danharrin, line 346:
```suggestion
        return $this->getLivewire()->hasToggleableTableColumns();
```

packages/tables/resources/views/index.blade.php
danharrin:
```suggestion
                                :form="$getColumnToggleForm()"
                                :width="$getColumnToggleFormWidth()"
```

danharrin:
> Will let @danharrin tear this code apart, but functionality wise it's all up to scratch and close to what has been discussed above.

It's actually perfect, it looks like I just made tons of comments but I just added `Table` to the name of some Livewire methods so they're consistent with others. Also I passed in the width to the popover.

ryangjchandler:
@danharrin Sweet, will sort these tomorrow. 🎉

ryangjchandler (approved):
This all looks good to me now.

danharrin:
And it all works with the new changes?

ryangjchandler:
@danharrin Yeah, all good still. Just tested :)

danharrin:
Just pulled this down for testing, looks like the columns are visible oppositely to what their checkbox says?

When the table first load, the columns should be hidden / visible based on `$isToggledByDefault`, and if a checkbox is checked the column should be visible. Right?

https://user-images.githubusercontent.com/41773797/164007817-05b6a41a-ba3c-406d-8344-e22e7baf2b51.mov

ryangjchandler:
Oh yeah, I've been thinking of this the wrong way. Just need to invert the `isToggled` logic.

ryangjchandler:
Actually, I'm not sure now... to me you're toggling the column so the checkbox being checked is almost indicating that you've toggled it.

Happy either way though!

danharrin:
"Toggling" to me sounds like the process of showing/hiding the column, not an indication of whether the column is visible or not.

To me, I think it makes more sense if checked = visible, as on the original video https://github.com/laravel-filament/filament/pull/1693#issue-1147538318

saade:
Don't you guys think that the `view-boards` icon looks better representing what the button does?

image

danharrin:
I do @saade

PR #1696, opened 2022-02-23
Fixed z-index of block picker when using child Builders

titantwentyone:
A convoluted example but given:

```
return $form
    ->schema([
        TextInput::make('title'),
        Builder::make('content')
            ->blocks([
                Block::make('builder')
                    ->schema(fn() => [
                        Builder::make('anotherbuilder')
                            ->schema([
                                Block::make('sometextfield_block')
                                    ->schema([
                                        TextInput::make('sometextfield')
                                    ]),
                                Block::make('sometextfield_block2')
                                        ->schema([
                                            TextInput::make('sometextfield2')
                                        ])
                            ])
                    ])
            ])
    ]);
```
When creating two child blocks, the block picker on the first item is obfsucated with the "add block" icon of the parent block making it unselectable.

image

This fix just changes the z-index of the block picker to z-20 ensuring it displays above the "add block" icon:

image

### DIFF
--- a/packages/forms/resources/views/components/builder/block-picker.blade.php
+++ b/packages/forms/resources/views/components/builder/block-picker.blade.php
@@ -9,7 +9,7 @@
     x-on:click.away="isCreateButtonDropdownOpen = false"
     x-transition
     {{ $attributes->class([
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks

PR #1706, opened 2022-02-24
Add close button to filter popover

ralphjsmit:
This PR adds a Close button to the filters popover, because some people do not know how to close the filter popover.

Example: 

Schermafbeelding 2022-02-24 om 17 16 21

PS: I figured out that I could add the new translation key as well to the Dutch translation file. So I compared the Dutch file to the English one and updated all missing keys.


danharrin:
I think this should be an X button in the top right corner of the modal instead of a link button?

drhoussem:
@danharrin I personally think it's better to put it next to reset filter from a UX perspective

PR #1667, opened 2022-02-21
[2.x] Extract notification message

ManojKiranA:
After the discussion in discord [here](https://discord.com/channels/883083792112300104/883083792653381695/942718176242851881)

PR #1711, opened 2022-02-25
[2.x] Show notification if only available

ManojKiranA:
followup #1667

### DIFF
--- a/packages/admin/src/Resources/Pages/CreateRecord.php
+++ b/packages/admin/src/Resources/Pages/CreateRecord.php
@@ -62,7 +62,7 @@ class CreateRecord extends Page implements HasFormActions
 
         $this->callHook('afterCreate');
 
-        if($this->getCreatedNotificationMessage()){
+        if( $this->getCreatedNotificationMessage()) {
             $this->notify(
                 'success',
                 $this->getCreatedNotificationMessage(),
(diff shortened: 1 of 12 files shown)


danharrin:
Can you search through all other instances of `$this->notify()` and do the same please?

Also, please format if statements with proper spacing, and use the `filled()` helper - `if (filled($this->getCreatedNotificationMessage())) {`

ManojKiranAppathurai:
I have updated it

packages/admin/src/Resources/Pages/CreateRecord.php
```diff
@@ -81,7 +83,7 @@ public function create(bool $another = false): void
         $this->redirect($this->getRedirectUrl());
     }
 
-    protected function getCreatedNotificationMessage(): string
+    protected function getCreatedNotificationMessage(): ?string
```
ryangjchandler, line 86: This is a breaking change I believe.

ryangjchandler (changes requested):

packages/admin/src/Resources/Pages/CreateRecord.php
danharrin: This isn't a breaking change - existing overrides to this method would have more narrow (but compatible typehints) - `?string -> string`

ryangjchandler: Yeah, got it the wrong way round. If the base class had a `-> string` and the child class implemented the `-> ?string` it would fail.

Ignore me!

ManojKiranAppathurai:
Do I need to change something ?

danharrin:
Thank you

PR #1721, opened 2022-02-27
docs: return navigation builder

ryangjchandler:
The `NavigationBuilder` instance has been to be returned from the callback function since we're no longer sending our own object through.


danharrin:
Cheers

PR #1722, opened 2022-02-27
feat: Add ability to add custom tags to the head

jeffgreco13:


ryangjchandler:
@jeffgreco13 Can you undo all of the formatting changes in the docs please.

packages/admin/resources/views/components/layouts/base.blade.php
```diff
@@ -13,6 +13,10 @@ class="filament antialiased bg-gray-100 js-focus-visible"
         <meta name="viewport" content="width=device-width, initial-scale=1">
         <meta name="csrf-token" content="{{ csrf_token() }}">
 
+        @foreach (\Filament\Facades\Filament::getMeta() as $tag)
+        {!! $tag !!}
```
ryangjchandler, line 17: This doesn't feel right... we should probably use a normal `{{ }}` here instead and in the documentation use `new HtmlString` for the tags instead. Reduces the likelihood of injections etc.

ryangjchandler: This should also be indented.

packages/admin/src/FilamentManager.php
```diff
@@ -82,6 +84,11 @@ public function registerNavigationItems(array $items): void
         $this->navigationItems = array_merge($this->navigationItems, $items);
     }
 
+    public function registerMeta(array $meta): void
```
ryangjchandler, line 87: `registerMeta` sounds weird too - perhaps we'd be better off with a method name like `pushMeta` or `addMeta`, thoughts @danharrin?

ryangjchandler (changes requested):

packages/admin/src/FilamentManager.php
jeffgreco13: I like `pushMeta` too. If that's the chosen option I can update and get it pushed.

packages/admin/resources/views/components/layouts/base.blade.php
```diff
@@ -13,6 +13,10 @@ class="filament antialiased bg-gray-100 js-focus-visible"
         <meta name="viewport" content="width=device-width, initial-scale=1">
         <meta name="csrf-token" content="{{ csrf_token() }}">
 
+        @foreach (\Filament\Facades\Filament::getMeta() as $tag)
+            {{ new \Illuminate\Support\HtmlString($tag) }}
```
ryangjchandler, line 17: We shouldn't wrap the `$tag` in a `HtmlString` - we should just output `$tag` normally and let the user decide if they want to pass an `HtmlString` or not.

ryangjchandler (changes requested):

packages/admin/resources/views/components/layouts/base.blade.php
jeffgreco13: I can agree with this. My thought was that there shouldn't ever be a scenario where it wouldn't be an HTML tag since it's in the head, but perhaps this just needs to be documented.

I'll update it and we can go from there!

danharrin:
Thank you :)

PR #1725, opened 2022-02-27
[2.x] Fix simple resources not respecting policy methods

### LINKED ISSUE #1720: Simple (model) resources doesn't seems to take the delete policy into consideration
### Package

filament/filament

### Package Version

v2.10.14

### Laravel Version

v9.2.0

### Livewire Version

v2.10.4

### PHP Version

PHP 8.1

### Bug description

While I'm returning `false` from the `delete()` method in a policy class, when I use a simple resource, I can still see and use the delete button. If I switch to a regular resource, everything seems to be working as expected.

### Steps to reproduce

_No response_

### Relevant log output

_No response_

ryangjchandler:
Closes #1720.


danharrin:
Thanks, I think we need to double check the relation manager traits to make sure we haven't made the same mistakes there too.

ryangjchandler:
@danharrin Yeah, they all looked good to me. That's what I referenced to double check how we were doing it elsewhere.

Issue #1740, opened 2022-03-01
Version v2.10.15 missing semicolon in CanCreateRecords


mallardduck:
### Package

filament/filament

### Package Version

v2.10.15

### Laravel Version

v9

### PHP Version

PHP 8.1.3

### Bug description

When setting up Filament I found that the current latest release is flawed and has a broken line.

### Steps to reproduce

Install the affected version and load a page that uses this trait.

### Relevant log output

```shell
> During class fetch: Uncaught ParseError: syntax error, unexpected token "}", expecting ";" in vendor/filament/filament/src/Resources/Pages/ListRecords/Concerns/CanCreateRecords.php:100
```

https://github.com/laravel-filament/filament/blob/819d69e0371c8b886c710601673572059753bad0/packages/admin/src/Resources/Pages/ListRecords/Concerns/CanCreateRecords.php#L99

ManojKiranAppathurai:
@danharrin I think I have missed the semicolon. And CI failed to catch the bug. We need to start adding more tests

mallardduck:
Could be good to re-enable the phpstan workflows.
Seems they are manually disabled currently.

danharrin:
They've been disabled since the L9 upgrade, because I haven't had time to fix them yet (even though the code works fine). Otherwise yeah, PHPstan will have caught this.

I'll start manually running PHPstan before every release, until we have the workflow working again.

ryangjchandler:
@danharrin Since I fixed up the PHPStan stuff, the workflow should mostly run fine I think.

mallardduck:
I have a PR to fix the handful of issues PHPStan is alerting on locally. Want me to send that up?

PR #1741, opened 2022-03-01
Fix bug #1740

mallardduck:
https://github.com/laravel-filament/filament/issues/1740


danharrin:
Whoops, looks like this was introduced in #1711. Thank you very much.

danharrin:
I will create a release right now.

Issue #1730, opened 2022-02-28
Table filter dropdown not properly visible on mobile devices


rajatpatelz:
### Package

filament/tables

### Package Version

v2.10

### Laravel Version

v9.1.0

### PHP Version

PHP 8.0.2

### Bug description

Table filter dropdown is not visible properly on mobile device (small and extra small screens) when table search field is not present.

danharrin:
Can I have screenshot/s of the issue?

rajatpatelz:
![image](https://user-images.githubusercontent.com/11793408/155976942-895ef7fd-2562-4a25-a7c4-f63ead5d24c3.png)

PR #1743, opened 2022-03-01
fix: responsive compatibility problem on table filters

FurkanGM:
for #1730


danharrin:
This isn't an issue with the positioning of the popover, we actually just need to make sure that the filter button is always on the right hand side of the table so that there is enough space to see the popover.

danharrin:
Thank you

PR #1744, opened 2022-03-01
Add Dutch translations for settings and translatable plugins

RVxLab:
This PR adds Dutch translations for the Settings and Translatable plugins.

While Dutch has multiple locale strings available (such as nl_NL for the Netherlands and nl_BE and Flemish Belgium), the ones here should be applicable to all of them.


danharrin:
Thanks!

PR #1757, opened 2022-03-02
Allow for sections to be hidden

philipmclifton:
Currently the section component uses the hideable trait but the template does not.


danharrin:
Hmm, sections can be hidden for me just fine. There's no need to hide the Blade for each component, that is done by the component container view.

Have you tried this basic example?

```php
Forms\Components\Section::make('Heading')->hidden()
```

PR #1759, opened 2022-03-02
fix: wrong slug generation on pages

FurkanGM:
This pr fixes wrong slug generation problem for pages.

Case: in currently code lines, not converts ü-ğ-ö etc. characters and addes this like this characters to directly url.

Solution: I'm converted first to kebab case because if page dont have custom title will get from class name. Later, i used Str::slug() helper function for fix ü-ğ-ö etc. characters.


danharrin:
I think we just need `Str::slug()`? as that uses kebab case anyway. Also, we should do the same for resources.

danharrin:
Ignore me lol, after testing you were right.

PR #1762, opened 2022-03-03
Add enabling dark mode through environment variable

marcoboers:
This PR adds the ability to set the dark mode setting for both admin, forms and tables through a single environment variable.

```dotenv
FILAMENT_DARK_MODE=true
```

It does however also allow overriding by `FORMS_DARK_MODE` for the form builder and `TABLES_DARK_MODE` for the table builder or for uniformity when not used in combination with filament.


danharrin:
Hi, thanks for your contribution, I do appreciate it. Personally though, I don't think that this is "environment" data - especially if you're using the form and table builders on their own. The custom parts of your app will either support dark mode styling or not, that's not something that could change between app environments.

PR #1769, opened 2022-03-03
Fix searchable select width

### LINKED ISSUE #1734: Select field size doesn't adjust to the specified in the colspan
### Package

filament/filament

### Package Version

v2.10.14

### Laravel Version

^8.75

### Livewire Version

_No response_

### PHP Version

PHP 8.1.2

### Bug description

When using the select field with various columsn ,when the content of the select field surprasses the colSpan size, it doesn't truncate as it did before. See image below:

image

### Steps to reproduce

Example code:

                HasManyRepeater::make('items')
                    ->label('')
                    ->relationship('items')
                    ->schema([
                        Select::make('vehicle_type')
                            ->label(__('labels.vehicle_type'))
                            ->searchable()
                            ->options(VehicleType::toArray())->required(),
                        TextInput::make('quantity')
                            ->label(__('labels.quantity'))
                            ->numeric()
                            ->default(1)
                            ->required(),
                        Toggle::make('is_helpersrequired')
                            ->label(__('labels.is_helpersrequired'))
                            ->inline(false),
                        TextInput::make('pallets')
                            ->label(__('labels.pallets'))
                            ->numeric()
                            ->default(0),
                    ])->columns(4),

> As you can see, it has 4 colums and it should fix perfectly.

### Relevant log output

_No response_

jacksleight:
Fixes https://github.com/laravel-filament/filament/issues/1734

The fix here was to add `min-w-0` to the select's wrapper div. Flex items have `min-width: auto` by default, which is based on the content width and doesn't play nice with long values that should be truncated.

This PR also addresses a separate issue with the internal search label/input. Both the label and input have `w-full` so are each given 50% of the width, which looks odd when you have long values:

Screenshot 2022-03-03 at 15 02 53

Simplest fix I could think of was to allow the label to take up all of the space by collapsing the input with `w-0`, but allow that to expand when the search is open with `grow`. Then adding `cursor-text` to the label.

Screenshot 2022-03-03 at 15 03 26
Screenshot 2022-03-03 at 15 04 16

I also removed the redundant `whitespace-nowrap`.

### DIFF
--- a/packages/forms/resources/views/components/select.blade.php
+++ b/packages/forms/resources/views/components/select.blade.php
@@ -23,7 +23,7 @@
             </span>
         @endif
 
-        <div class="flex-1">
+        <div class="flex-1 min-w-0">
             @unless ($isSearchable())
                 <select
                     {!! $isAutofocused() ? 'autofocus' : null !!}
@@ -97,7 +97,7 @@
                             x-show="! isOpen"
                             x-text="label ?? '{{ addslashes($getPlaceholder()) }}'"
                             @class([
-                                'w-full bg-white whitespace-nowrap truncate',
+                                'w-full bg-white truncate cursor-text',
                                 'dark:bg-gray-700' => config('forms.dark_mode'),
                             ])
                         ></span>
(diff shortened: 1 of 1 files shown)


romaldyminaya:
Thanks

PR #1771, opened 2022-03-03
Allow stat cards to be linked, allow `HtmlString` objects in stat card labels and descriptions

jacksleight:
First go at https://github.com/laravel-filament/filament/discussions/1768

Not sure about the way the optional link has been implemented in `card.blade.php`... Duplicating the entire contents just to switch the wrapping element seems a bit over the top, but also doing it this way is not pretty. Let me know the preferred approach.

### DIFF
--- a/packages/admin/resources/views/components/stats/card.blade.php
+++ b/packages/admin/resources/views/components/stats/card.blade.php
@@ -1,22 +1,28 @@
 @props([
+    'tag' => 'div',
     'chart' => null,
     'chartColor' => null,
     'color' => null,
     'description' => null,
     'descriptionColor' => null,
     'descriptionIcon' => null,
+    'href' => null,
+    'target' => null,
     'flat' => false,
     'label' => null,
     'value' => null,
 ])
 
-<div {{ $attributes->class([
-    'relative p-6 rounded-2xl filament-stats-card',
-    'bg-white shadow' => ! $flat,
-    'dark:bg-gray-800' => (! $flat) && config('filament.dark_mode'),
-    'border' => $flat,
-    'dark:border-gray-700' => $flat && config('filament.dark_mode'),
-]) }}>
+<{{ $tag }} 
+    {!! ($tag === 'a') ? "href=\"{$href}\" target=\"{$target}\"" : null !!}
+    {{ $attributes->class([
+        'relative p-6 rounded-2xl filament-stats-card',
+        'bg-white shadow' => ! $flat,
+        'dark:bg-gray-800' => (! $flat) && config('filament.dark_mode'),
+        'border' => $flat,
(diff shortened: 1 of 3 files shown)


packages/admin/resources/views/components/stats/card.blade.php
```diff
@@ -1,22 +1,28 @@
...
 
-<div {{ $attributes->class([
-    'relative p-6 rounded-2xl filament-stats-card',
-    'bg-white shadow' => ! $flat,
-    'dark:bg-gray-800' => (! $flat) && config('filament.dark_mode'),
-    'border' => $flat,
-    'dark:border-gray-700' => $flat && config('filament.dark_mode'),
-]) }}>
+<{{ $tag }} 
+    {!! ($tag === 'a') ? "href=\"{$href}\" target=\"{$target}\"" : null !!}
```
ryangjchandler, line 17: Should we be using raw echos here?

ryangjchandler (changes requested):

packages/admin/resources/views/components/stats/card.blade.php
danharrin: Yes, Blade throws a fit if you attempt to use quotes and stuff if not

jacksleight: Good point, fixed.

packages/admin/resources/views/components/stats/card.blade.php
```diff
@@ -1,22 +1,28 @@
...
 ])
 
-<div {{ $attributes->class([
-    'relative p-6 rounded-2xl filament-stats-card',
-    'bg-white shadow' => ! $flat,
-    'dark:bg-gray-800' => (! $flat) && config('filament.dark_mode'),
-    'border' => $flat,
-    'dark:border-gray-700' => $flat && config('filament.dark_mode'),
-]) }}>
+<{{ $tag }}
```
danharrin: Instead of adding `{{ $tag }}` here - I usually prefer to write the tags out in conditions fully. Check out `button.blade.php`. You can extract the array of classes to a variable elsewhere.

danharrin: Ignore me. It would be excessive here.

danharrin:
Thank you!

jacksleight:
Just realised this is completely unnecessary:

```php
    'href' => null,
    'target' => null,
```
```php
{!! ($tag === 'a') ? "href=\"{$href}\" target=\"{$target}\"" : null !!}
```

They'd be passed through as attributes anyway and wouldn't affect the div if no url is set.

danharrin:
I'll fix it. Thank you mate.

PR #1779, opened 2022-03-04
Change type of min/max value from int to float

wit3:
This can manage an input mask with min/max value a decimal number


danharrin:
Thanks

PR #1784, opened 2022-03-04
Update card borders

ralphjsmit:
I noticed that you updated the `Card` component to have borders. On the admin panel, this currently looks like this:

Schermafbeelding 2022-03-04 om 12 52 06

 
To be honest, I think this is quite noisy/incoherent. In the Discord-channel, Dan said that Filament was always internally using the `<x-filament::card>` component, also for the stats overview widget. I could not find that relation, because for the stats overview, you have a separate Blade component in `packages/admin/resources/views/components/stats/card.blade.php`. 

### Restoring admin panel design

Anyways, I updated the `packages/admin/resources/views/components/card/index.blade.php` component, so that it doesn't have a border anymore. **This change did not impact the `Card` component from the Forms package**, so it basically restores the dashboard design to its 'old' look.

### `Card` component

Second, I took a look at the `Card` component. This is an example from the admin panel with a dummy card and section about how it currently looks:

Schermafbeelding 2022-03-04 om 12 59 22

Personally I think that the shadows on the grey background are bit too heavy when you compare it with the relationmanager below it. It is not really the same style. 

The fix for this is very simple: remove the `shadow` from both the Card and the Section component. This makes the design of both components more consistent with the design/border of the relationmanagers: 

Schermafbeelding 2022-03-04 om 13 05 25

If I'm really honest, I still prefer the old design with only the `shadow` and no borders, but otherwise I don't have a good idea on how to fix the problem that the `Card` isn't visible on white backgrounds.

### DIFF
--- a/packages/admin/resources/views/components/card/index.blade.php
+++ b/packages/admin/resources/views/components/card/index.blade.php
@@ -6,7 +6,7 @@
 ])
 
 <div {{ $attributes->class([
-    'p-2 space-y-2 bg-white rounded-xl shadow border border-gray-300',
+    'p-2 space-y-2 bg-white rounded-xl shadow',
     'dark:border-gray-600 dark:bg-gray-800' => config('filament.dark_mode'),
 ]) }}>
     @if ($actions || $header || $heading)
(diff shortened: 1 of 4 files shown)


ralphjsmit:
PS: can you explain what the variable `$flat` exactly does in the `Card` component?

ralphjsmit:
Nice, I removed the `$flat` variable.

danharrin:
Thank you

PR #1796, opened 2022-03-05
Specify pluralization intervals

alperenersoy:
Languages that don't exist in Symfony's [getPluralizationRule function](https://github.com/symfony/translation-contracts/blob/main/TranslatorTrait.php#L139) always return the singular translation for trans_choice() function. So it's needed to specify intervals.


danharrin:
Thank you

PR #1797, opened 2022-03-06
fix selected items count in Arabic translation

maherelgamil:
After reviewing this #1796 Pull Request, I think it's solved also in Arabic translation.


danharrin:
The change you made doesn't look like it's in the correct format? But I assume it's different in your language and that you've tested it 👍

maherelgamil:
@danharrin tested don't worry and solve our problem in with translation when selected  multi rows 

thanks for merged .

PR #1800, opened 2022-03-06
remove default value from `->columns()` method

maherelgamil:


danharrin:
This is a breaking change, and we use 2 columns deliberately in these places. If you want to customise the columns in your project, use `->columns(1)` on those components yourself.

maherelgamil:
@danharrin the value added by default 2 in the attribute of the method , so no need to re assign value again

And if any one need to customize the value , he will change it to the new value

I think no breaking here

But as you need 👌

danharrin:
Ah, I understand. The reason we added the parameters was to be explicit that we're using 2 columns without having to check the default.

maherelgamil:
@danharrin , so I think you should accept this request 😂😁

maherelgamil:
@danharrin 
https://github.com/laravel-filament/filament/blob/2.x/packages/forms/src/Concerns/HasColumns.php#L31

here is what I mean, this code in filament core package, not changed from the developers' side in any way

PR #1813, opened 2022-03-07
Add a new table widget type

jacksleight:
Adds a new type of widget called `TableWidget`. Simply implements `InteractsWithTable` and outputs the result.

Includes a new `--T|table` option to the `make:filament-widget` command.

Example (with pagination disabled):

Screenshot 2022-03-07 at 10 44 39


packages/admin/stubs/TableWidget.stub
```diff
@@ -0,0 +1,28 @@
...
+class {{ class }} extends BaseWidget
+{
+    protected function getTableHeading(): string | Closure | null
+    {
+        return 'Table';
+    }
+
+    protected function getTableQuery(): Builder
+    {
+        return null;
```
ryangjchandler, line 19: Perhaps the stub should be empty instead of having an invalid `return` type?

packages/admin/stubs/TableWidget.stub
```diff
@@ -0,0 +1,28 @@
...
+use Closure;
+use Filament\Tables;
+use Filament\Widgets\TableWidget as BaseWidget;
+use Illuminate\Database\Eloquent\Builder;
+
+class {{ class }} extends BaseWidget
+{
+    protected function getTableHeading(): string | Closure | null
+    {
+        return 'Table';
```
ryangjchandler, line 14: Could we replace `Table` with the class name perhaps?

ryangjchandler (changes requested):

ryangjchandler:
@jacksleight This is a great idea I think, assuming @danharrin agrees, would you be able to add this to the docs too?

packages/admin/stubs/TableWidget.stub
```diff
@@ -0,0 +1,28 @@
...
+use Closure;
+use Filament\Tables;
+use Filament\Widgets\TableWidget as BaseWidget;
+use Illuminate\Database\Eloquent\Builder;
+
+class {{ class }} extends BaseWidget
+{
+    protected function getTableHeading(): string | Closure | null
+    {
+        return '{{ class }}';
```
danharrin, line 14: I think that Ryan meant that the default heading should be generated by `Str::title()` instead of being copied from the class name - like we do on Page and Resource classes.

danharrin:
Those command flags are relatively new so haven't been documented yet.

packages/admin/stubs/TableWidget.stub
ryangjchandler: I would leave it as an empty `// ...` line or something.

packages/admin/stubs/TableWidget.stub
jacksleight: Gotcha, that's been added.

packages/admin/stubs/TableWidget.stub
jacksleight: Cool, done.

ryangjchandler (changes requested):

ryangjchandler:
Ignore that last comment - GH showed me an old commit from the notifications, lol.

danharrin:
Maybe we should also use `->beforeLast('Widget')` on the heading? Don't we do the same for chart widgets?

packages/admin/docs/04-dashboard.md
```diff
@@ -147,6 +147,51 @@ Alternatively, you may disable polling altogether:
...
+use Closure;
+use Filament\Tables;
+use Filament\Widgets\TableWidget as BaseWidget;
+use Illuminate\Database\Eloquent\Builder;
+
+class OrdersOverview extends BaseWidget
+{
+    protected function getTableQuery(): Builder
+    {
+        return Order::query()->orderBy('created_at', 'desc')->limit(10);
```
ryangjchandler, line 179: We could use `->latest()` here to it's more idiomatic.

ryangjchandler (changes requested):

packages/admin/docs/04-dashboard.md
```diff
@@ -147,6 +147,51 @@ Alternatively, you may disable polling altogether:
...
 ```
 
+## Table widgets
+
+Filament comes with a "table" widget template, which you can use to display a table of data without needing to write a custom view.
+
+Start by creating a widget with the command:
+
+```bash
+php artisan make:filament-widget OrdersOverview
```
pxlrbt, line 157: Shouldn't this use the new `--table`flag instead of explaining how to replace the BaseWidget?

PR #1817, opened 2022-03-07
Fix custom order in HasManyRepeater

jacksleight:
Possible fix for https://github.com/laravel-filament/filament/issues/1816

Livewire re-orders numerically indexed array keys, breaking the order of repeater items that aren't ordered by ascending ID. This changes the array to an associative array, preventing that. See https://github.com/laravel-filament/filament/issues/1816#issuecomment-1060925406

From initial testing this appears to be all that's necessary to fix this with no side effects that I can see. Creating, updating and deleting items all work as expected. But I may well have missed something!

### DIFF
--- a/packages/forms/src/Components/RelationshipRepeater.php
+++ b/packages/forms/src/Components/RelationshipRepeater.php
@@ -40,7 +40,7 @@ class RelationshipRepeater extends Repeater
             $recordsToDelete = [];
 
             foreach ($existingRecords->pluck($localKeyName) as $keyToCheckForDeletion) {
-                if (array_key_exists('id-'.$keyToCheckForDeletion, $state)) {
+                if (array_key_exists("record-{$keyToCheckForDeletion}", $state)) {
                     continue;
                 }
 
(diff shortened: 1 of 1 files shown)


ryangjchandler:
Hm, @danharrin - can we not use a UUID here like we do for the normal block editor and repeater?

jacksleight:
I assumed it had to use the model IDs so it can match changes with the existing DB records? But I'm not super familiar with how it works so maybe that isn't the case.

danharrin:
We use UUIDs for new items & real numeric IDs for existing ones, so we are able to keep track of records that gave been updated and not just created.

ryangjchandler:
Hm, this should be fine then.

jacksleight:
Yeah I know it’s not the most elegant solution.

But, it sounds like the only fix is going to involve avoiding unsorted numeric keys somehow, see https://github.com/livewire/livewire/issues/900#issuecomment-652080300 and https://github.com/livewire/livewire/issues/900#issuecomment-652080626.

I tried to see if there was an alternative approach by using UUIDs as keys but not actually using them at all when checking for updates/deletions. Instead checking the actual attribute values inside the item data. Thought that should work in theory, but it doesn't as the keys changing on each request seems to mess with Livewire and you end up with loads of duplicate records. There could be a way to make that work but I couldn't figure it out.

At the end of the day, is having the array keyed like `id-123`, `id-456` much different from the way standard repeaters are keyed by UUID.

danharrin:
Alright, I've renamed it to `record-` instead of `id-`. This should be fine. Thank you very much.

PR #1820, opened 2022-03-07
docs: Update StatsOverviewWidget command

pxlrbt:
`php artisan make:filament-widget StatsOverview` will create `StatsOverview.php` but the example class is named `StatsOverviewWidget` so it causes an issue with autoloading.

### DIFF
--- a/packages/admin/docs/04-dashboard.md
+++ b/packages/admin/docs/04-dashboard.md
@@ -44,7 +44,7 @@ Filament comes with a "stats overview" widget template, which you can use to dis
 Start by creating a widget with the command:
 
 ```bash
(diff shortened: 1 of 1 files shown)


danharrin:
Not sure what you mean... the name of the class doesn't need to include `Widget`, and it should ways match the filename... I think this is something wrong with the command that generates the file & not the example?

I actually think that widget class names without `Widget` at the end look better, that's why the examples are like that.

pxlrbt:
Someone on Discord had an autoloader issue. He said he created a StatsOverview widget via command and then copied the example code and the class name didn't match the file name. I thought the command appends "Widget" to the name. Just tried it myself and it doesn't. You can ignore this.

PR #1828, opened 2022-03-08
File naming increment

titantwentyone:
As discussed, this PR enables the automatic incrementing of filenames when files are uploaded with `FileUpload`. You can disable the functionality by passing `false` to `getFileNameForStorageUsing` or override with a closure.


danharrin:
Hey, thanks for this. I've reviewed it, and although this works well, we should go for a different solution.

The actual `getFileNameForStorageUsing()` is fine, it should probably be called `getUploadedFileNameForStorageUsing()` though.

By default in `setUp()`, `getUploadedFileNameForStorageUsing()` should just return `$file->getFilename()`. I like the incrementing check, but it will never be used for `getFilename()` as those are random already, so I think it should be up to the user to define that in their own code.

Instead of checking `$component->shouldPreserveFilenames` in `getUploadedFileNameForStorageUsing()`, the `shouldPreservesFilename()` method should set the `getUploadedFileNameForStorageUsing()` method callback internally to just use `getClientOriginalName()` instead of `getFileName()` - nothing else.

Thanks!

titantwentyone:
Thanks Dan. Are you saying that we set this as sensible default if `shouldPreseveFilenames` evaluates to `true`? I think (if I have understood you correctly), there may be an issue depending on the order in which these are applied:

```
FileUpload::make('whatever')
   ->getUploadedFileNameForStorageUsing(function() { //custom code })
   ->shouldPreseveFilenames() // will overwrite custom code above
```
Perhaps in `setUp` we just have a guard:

```
$this->getFileNameForStorageUsing(function(BaseFileUpload $component, TemporaryUploadedFile $file) {

    if(!$component->shouldPreserveFilenames()) {
        return $file->getFilename();
    }

    ....
}
```

danharrin:
Yeah, order of methods would matter. But even in your proposed solution, the custom code would still never be ran. Let's keep it simple :)

titantwentyone:
The proposed solution should be fine. The config of the form just sets the value of the save method and whether we should preserve filenames - `getUploadedFileNameForStorageUsing` is not executed until the control is in use at which time `shouldPreserveFilename` is checked (I'm assuming `getUploadedFileNameForStorageUsing` is not set in `shouldPreserveFilename` - if it was as suggested, then, yes, it wouldn't execute).

I think really I'm not massively sold on setting `getUploadedFileNameForStorageUsing` from `shouldPreserveFilenames` because:

1. It feels anomolous to have these methods be dependent on order. AFAIK, no other form method in any other control behaves this way. It also breaks the "fluency" of the API due to this order constraint.
2. It feels like it possibly breaks the "single responsibility principle" by allowing one method to set the value of another.
3. I think long term this may cause headaches. If this was commonplace, as the API provides more functionality, we may be looking at problems unravelling what method set what and why and a list of rules pertaining to what order methods should be applied. Imagine if someone suggested new functionality `doWhatever()` which must preserve filenames in addition to this PR - we'd have a kind of "cascade method dependency hell" going on - i.e. methods stepping out of their area of responsibility. Imagine further if some other functionality depended on that - not totally outside the realms of possibility!

Guards will remove all these issues. Let methods decide for themselves what logic to employ by inspecting others. Genuinely believe this is the simpler and more logical solution - and opens up opportunities for complex (but sane!) behaviour.

Respectfully my 2 cents (plus a bit more!) :)

danharrin:
I don't really see the issue of setting `getUploadedFileNameForStorageUsing()` inside `shouldPreserveFilenames()` as you'd never use both methods together anyway so the order doesn't matter. The `getUploadedFileNameForStorageUsing()` callback accepts a `$file` and not a `$fileNameAlreadyRetrievedFromFile`. But fair enough on your point about being able to use `shouldPreserveFilenames()` elsewhere, since this could be useful. Let's keep the simple conditional check.

However, we still need to remove the logic that generates a new filename for existing files. In your app, you could macro the field to provide this functionality, I don't think it should be in core though.

danharrin:
Thanks. You can now use `getUploadedFileNameForStorageUsing()` to implement your custom file name functionality via a macro 👍

titantwentyone:
Thanks Dan. Bit of a full week and weekend so couldn't look sooner.

PR #1832, opened 2022-03-08
Fix issue with from() method in Input Mask

intrepidws:
Fixing a typo that was preventing the `from()` method from working properly in an Input Mask range.


danharrin:
Thanks

PR #1837, opened 2022-03-09
Added Spanish missing translations

pathros:
I've just found some other files with missing Spanish translations.


danharrin:
Thanks

PR #1839, opened 2022-03-09
label for translation dropdown

panakour:
This will let developer to define label in translations dropdown list to be more friendly for the users. 
```
    public static function getTranslatableLocales(): array
    {
        return ['en' => "English", 'el' => "Greek"];
    }
```
image

Also I've add a [condition](https://github.com/laravel-filament/filament/pull/1839/files#diff-615e561435fc8c8276bc2c640a37d6fed8cb2d1d1bf4f4f03e339e9771050d32R18) to prevent introducing breaking changes for anyone that was using the translatable locales only with keys.


danharrin:
Instead of changing the format of `getTranslatableLocales()`, maybe we should just add a in a translatable lang file that maps language keys to labels, starting with English-only? Then we can use that lang file to translate the keys without any extra effort from the developer.

panakour:
@danharrin you mean a translatable.php file under [config directory](https://github.com/laravel-filament/filament/tree/2.x/packages/spatie-laravel-translatable-plugin/config )  like: 
```
return [
    'locales' => [
        'en' => 'English', 
        'el' => 'Greek',
    ],
]
```
and register this in service provider?

danharrin:
In a translation file like `locales.php`:

```php
return [
    'en' => 'English', 
    'el' => 'Greek',
];
```

panakour:
shouldn't be better to put this in the existing config https://github.com/laravel-filament/filament/blob/2.x/packages/spatie-laravel-translatable-plugin/config/filament-spatie-laravel-translatable-plugin.php

danharrin:
IMO I don't think this is config related, languages don't change names but they do have different names in different languages.

panakour:
Can you please explain me a bit more how you want this to be implemented without a config file

danharrin:
Turns out that `locale_get_display_name()` exists in PHP, we can get the name of the language very easily!

Issue #764, opened 2021-12-07
Policies ignored in bulk table actions


dive-michiel:
### Package

filament/tables

### Package Version

v2.5.4

### Laravel Version

v8.74

### Livewire Version

v2.8.1

### PHP Version

PHP 8.0.0

### Bug description

The policy is ignored for bulk actions, and I'm able to see the delete action and delete all rows (even-though it is denied by the policy)

Example policy:

```php
<?php

namespace App\Policies;

use Illuminate\Auth\Access\HandlesAuthorization;

class ProjectPolicy
{
    use HandlesAuthorization;

    public function viewAny()
    {
        return true;
    }

    public function view()
    {
        return true;
    }

    public function create()
    {
        return false;
    }

    public function update()
    {
        return false;
    }

    public function delete()
    {
        return false;
    }

    public function restore()
    {
        return false;
    }

    public function forceDelete()
    {
        return false;
    }
}

```

Result:

![bulk-delete](https://user-images.githubusercontent.com/2650975/145092231-fc7e7770-4ee7-4f8f-97b5-e077cec300cd.png)

### Steps to reproduce

• Create a policy for a resource to test with that denies everything except viewing

danharrin:
That bulk action relies on the deleteAny method of the resource, as the delete method usually checks a specific record.

danharrin:
I've added more clarification [here](https://filamentadmin.com/docs/2.x/admin/resources#authorization).

dive-michiel:
Ok, thanks for the clarification! 

In my opinion: I do find it a bit strange to have a `deleteAny` method. Deleting is a destructive action and if you don't have permission to delete a single record, why would you have permission to delete `any`/all record(s)? 

P.S. congrats on the v2 release, looking very promising.

danharrin:
If we used the `delete` method, we'd have to loop through every selected record and check if you have permission to delete them. Since this is a "bulk action", this probably isn't very performant. So, a `deleteAny()` method that doesn't accept a `Model` instance parameter is better IMO. Does that make more sense? Or am I missing something? ;)

dive-michiel:
That's true, you could show the action as long as at least one record has the delete permission, if no records have the permission this will have to loop all - like you said. But if the page sizes aren't huge, this should have only minimal performance implications? I fully understand your concerns and adding the policy list to the docs is good enough for me :)

danharrin:
[...]

Issue #1843, opened 2022-03-09
Policies ignored in bulk table actions


romaldyminaya:
### Package

filament/filament

### Package Version

v2.10.24

### Laravel Version

V9.1.1

### PHP Version

PHP 8.1.2

### Bug description

The policy is ignored for bulk actions, and I'm able to see the delete action and delete all rows (even-though it is denied by the policy).

The thing is that even if the user has permissions to deleteAny, the system needs to verify each record individually to confirm if he can achieve the desired task.

> I copied the information from the clossed issue #

### Steps to reproduce

Example Policy

```
<?php

namespace App\Policies;

use Illuminate\Auth\Access\HandlesAuthorization;

class ProjectPolicy
{
    use HandlesAuthorization;

    public function viewAny()
    {
        return true;
    }

    public function view()
    {
        return true;
    }

    public function create()
    {
        return false;
    }

    public function update()
    {
        return false;
    }

    public function delete()
    {
        return false;
    }

    public function restore()
    {
        return false;
    }

    public function forceDelete()
    {
        return false;
    }
}
```

![image](https://user-images.githubusercontent.com/2809147/157455124-a6513126-ed82-40e0-b28d-11802d4a66f2.png)

danharrin:
https://github.com/laravel-filament/filament/pull/1844#issuecomment-1064962052

PR #1844, opened 2022-03-09
Fixes issue #764 and #1843

romaldyminaya:
Allow the system to verify the individual delete method's policy even if the user has permissions to deleteAny.

### DIFF
--- a/packages/admin/src/Resources/Pages/ListRecords.php
+++ b/packages/admin/src/Resources/Pages/ListRecords.php
@@ -136,7 +136,7 @@ class ListRecords extends Page implements Tables\Contracts\HasTable
 
     protected function handleRecordBulkDeletion(Collection $records): void
     {
-        $records->each(fn (Model $record) => $record->delete());
+        $records->each(fn (Model $record) => static::getResource()::canDelete($record) ? $record->delete() : null);
     }
 
     protected function getTitle(): string


danharrin:
As mentioned in #764, we deliberately do not check the `delete()` policy for bulk deletes, as it is not performant to loop through every record & potentially check the database for permissions or whatever your policy does. We use the `deleteAny()` method for checking bulk access permissions, and if you wish to check each record policy before you delete yourself, you can override the method on your List page:

```php
protected function handleRecordBulkDeletion(Collection $records): void
{
     $records->each(fn (Model $record) => static::getResource()::canDelete($record) ? $record->delete() : null);
}
```

romaldyminaya:
Understood

adrianlzx1996:
> As mentioned in #764, we deliberately do not check the `delete()` policy for bulk deletes, as it is not performant to loop through every record & potentially check the database for permissions or whatever your policy does. We use the `deleteAny()` method for checking bulk access permissions, and if you wish to check each record policy before you delete yourself, you can override the method on your List page:
> 
> ```
> protected function handleRecordBulkDeletion(Collection $records): void
> {
>      $records->each(fn (Model $record) => static::getResource()::canDelete($record) ? $record->delete() : null);
> }
> ```

How can i do this if i'm just using filament/table ONLY, instead of filament/admin?

PR #1849, opened 2022-03-09
add ItemBetweenCreationDisabled

Z3d0X:
https://github.com/laravel-filament/filament/discussions/1840


danharrin:
Hmm, instead of adding a new method, why don't we limit between creation based on `disableItemMovement()`? Because otherwise you could just create a new item at the end and move it up anyway. Both settings seem linked.

danharrin:
Thanks mate

PR #1850, opened 2022-03-09
Adding Id to notification

brunolipe-a:
With this change, we can dispatch many notifications.


ryangjchandler:
@brunolipe-a What problem does this fix?

brunolipe-a:
> 

It is not possible to send call multiple `$this->notify()` and show in the page, only one.

danharrin:
Thank you

PR #1854, opened 2022-03-10
[Feature] Ability to display icon on widget

mohamedsabil83:
Screenshot from 2022-03-10 14-09-42

### DIFF
--- a/packages/admin/resources/views/components/stats/card.blade.php
+++ b/packages/admin/resources/views/components/stats/card.blade.php
@@ -22,10 +22,13 @@
         'space-y-2',
     ])>
         <div @class([
-            'text-sm font-medium text-gray-500',
+            'flex items-center space-x-2 rtl:space-x-reverse text-sm font-medium text-gray-500',
             'dark:text-gray-200' => config('filament.dark_mode'),
         ])>
-            {{ $label }}
+            @if ($icon)
+                <x-dynamic-component :component="$icon" class="w-4 h-4" />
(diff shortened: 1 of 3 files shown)


ryangjchandler:
I don't think this visually looks great. I get the reasoning behind it but maybe we're better off with a different design.

mohamedsabil83:
Yep, it's a simple use until we confirm which styles we can add to core of Filament

danharrin:
I think that maybe a small icon on the left of the label might be better.

mohamedsabil83:
ok, I'll change it

mohamedsabil83:
What about this look? if it's ok, I'll push it

Screenshot from 2022-03-10 16-35-15

danharrin:
I think that the icon needs to be a little smaller & there needs to be a little more space between it and the label.

danharrin:
And I don't think we should have `bigIcon()` - you can always create your own Card components if you wish with a custom view.

mohamedsabil83:
Ok, just the small one. I have a plan to multi-style widget but don't know if accept to the core or a separate package but now I'm sure it'll be a separate package 😄 

For the current one, I think it's good now? (w-4 h-4) and (space-x-2) the previous was (w-6 h-6) and (space-x-1)
Screenshot from 2022-03-11 13-37-57

danharrin:
LGTM, thanks!

mohamedsabil83:
You're welcome

PR #1857, opened 2022-03-10
FilePond get all the urls at once

robertorinaldi-dev:
FileUpload becomes very slow if there are many files.

This is due to the fact that the getUploadedFileUrl function is called for each of them.

Replacing it with a single getUploadedFilesUrl function that is called only once solves the problem.

I also reorganized the part of the code that creates the array of files to be passed to the component in order to avoid repetitions.


robertorinaldi-dev:
Yes, in BaseFileUpload getUploadedFileUrls calls the function getUploadedFileUrlUsing for each file as it was before.
It is only from Alpine that instead of calling it once for each file it is called once for all.

PS: I'm using Spatie Media Library plugin

danharrin:
Thank you :)

PR #1861, opened 2022-03-10
Fix file upload spacing

jacksleight:
Fixes https://github.com/laravel-filament/filament/pull/1626#issuecomment-1064079610

Screenshot 2022-03-10 at 14 51 20


danharrin:
Thank you

SebastianUhlig:
@jacksleight @danharrin could it be that this change no longer exists in v3? For me, they are still listed one below the other without spacing

danharrin:
In theory it is possible that its broken in v3, but it probably wasn't deliberate

PR #1867, opened 2022-03-10
Fix  overwrite on tables w/ multiple filters using  method

stiknoltz:
Hey @danharrin I'm really enjoying Filament on a current project and thus far have only run into this one issue for which I _hope_ I've provided the proper solution.

The steps to reproduce this are: 
* Add a multiple filters to a table that each use the `->query()` method. 
* In each query, apply a join.
* With Xdebug on:
    * lines 18 and 20 of HasRecords.php 
    * line 65 of HasFilters.php
    * lines 26 and 29 of InteractsWithTableQuery.php 
    you can see that the joins are dropped once the loop is exited. 

I believe this is because the $query is re-instantiated entering the `->where()` method. I'm not 100%, but when I experimented by passing the $query object through in the `use` clause, it solved the issue. 

Hope this helps.

### DIFF
--- a/packages/tables/src/Concerns/HasFilters.php
+++ b/packages/tables/src/Concerns/HasFilters.php
@@ -60,7 +60,7 @@ trait HasFilters
     {
         $data = $this->getTableFiltersForm()->getState();
 
-        return $query->where(function (Builder $query) use ($data) {
+        return $query->where(function () use ($query, $data) {
             foreach ($this->getCachedTableFilters() as $filter) {
                 $filter->apply(
                     $query,


ryangjchandler:
Hm, the original code should work since Laravel passes the `$query` through to allow grouped where clauses. What errors were you running into?

stiknoltz:
That's honestly what I thought too. The issue I was encountering was in a table where the filters required joins when executing the `->query()` method. 

Reviewing it now, I can see that perhaps there is a step in the grouped `where` clause that removes those joins. 

The error I encounter is an SQL error that relates to a column I'm using a conditional not existing because the join is stripped out. 

The following is the code of the filter that has the join removed. I'll dig in a little deeper myself into what happens during that `where()` clause using a closure. That's probably the answer. 

```
                Tables\Filters\SelectFilter::make('vendor')
                    ->options(function () {
                        $options = FacilityProduct::query()->whereFacilityId(current_session_facility())
                            ->get()
                            ->pluck('product')->flatten()
                            ->pluck('vendor')->flatten()
                            ->unique()
                            ->sortBy('name')
                            ->pluck('name', 'id');
                        return $options;
                    })
                    ->query(function(Builder $query, $data) {
                        if (blank($data['value'] ?? null)) {
                            return $query;
                        }

                        $query->join('products as products_vendor_filter', 'facility_product.product_id', '=', 'products_vendor_filter.id');
                        $query->where('products_vendor_filter.vendor_id', '=', $data['value']);

                        return $query;
                    }),
```

stiknoltz:
Now that I think through the ramifications of having that query enveloped in part of a `where` clause, I'm almost certain that's is the root of the issue, however, I'm not uncertain as to how to proceed. I'll take a look at other options for handling the joins.

stiknoltz:
Ok, just to boil this down a bit more, the error I was receiving had to do w/ the aggregation query for pagination and not the main query. Although, I do think that the `where()` filter aggregation likely washes out any joins declared therein. 

In either event, I will try to further analyze if I'm unable to workaround in the future. In this case, the workaround was to:
* Utilize `getTableQuery()` in the `ListRecords` class of the resource
* Add the joins there
* Utilize the joined tables in the filter `query()` calls. 

This could be an optimization issue in other scenarios where the joins are too taxing to add when the filter is not in place, but in this case, it's fine. 

In either event, the after further evaluation, the proposed fix is inappropriate and detrimental.

danharrin:
Yeah, so we use grouped `where()` clauses so that different filters' queries dont get mixed up with each other. For example, a mixed `orWhere()` which causes an unexpected order of clauses.

I'd add joins in `getTableQuery()` as you mentioned. We could also add an `applyJoinsUsing()` method to filters, which accepts a callback to apply joins for that column. We then loop through each column and apply that callback to the query. Then you don't need to override anything?

stiknoltz:
I like the `applyJoinsUsing()` concept and I'd be happy to work on a PR for that unless you think it'd be trivial to knock in an upcoming release. For now, the use of `getTableQuery()` is working fine. 

For the table I was working on, the filters in question are linchpins of functionality so they will used frequently and the joins are simple and indexed, so it's not a major issue.

PR #1868, opened 2022-03-10
Auto detect redirect from Livewire response

pxlrbt:
The solution I came up with for auto-detecting redirects for notification. It checks the Livewire response for a redirect. 
I tested this with the demo app with saving, saving with redirect and saving from a modal.

Disclaimer: I have no idea why this doesn't flash messages sent via Browserevent two times, but it somehow works 😅

What do you think? And would you want to deprecate that `isAfterRedirect` param? I thought about `#[Deprecated]` but that is PhpStorm specific. Alternatively we could trigger a deprecation error.

### DIFF
--- a/packages/admin/resources/views/components/notification-manager.blade.php
+++ b/packages/admin/resources/views/components/notification-manager.blade.php
@@ -1,6 +1,6 @@
 <div
     x-data="{
-        notifications: {{ Js::from(session()->pull('notifications', [])) }},
+        notifications: @js(session()->pull('notifications', [])),
         add (event) {
             this.notifications = this.notifications.concat(event.detail)
         },
--- a/packages/admin/resources/views/components/notification-manager.blade.php
+++ b/packages/admin/resources/views/components/notification-manager.blade.php
@@ -2,13 +2,7 @@
     x-data="{
         notifications: {{ Js::from(session()->pull('notifications', [])) }},
         add (event) {
-            for (let notification of event.detail) {
-                this.notifications.push({
-                    id: notification.id,
-                    status: notification.status,
-                    message: notification.message,
-                })
-            }
+            this.notifications = this.notifications.concat(event.detail)
         },
         remove: function (notification) {
             this.notifications = this.notifications.filter(i => i.id !== notification.id)
--- a/packages/admin/resources/views/components/notification-manager.blade.php
+++ b/packages/admin/resources/views/components/notification-manager.blade.php
@@ -1,8 +1,14 @@
 <div
     x-data="{
-        notifications: {{ json_encode(session()->pull('notifications', [])) }},
-        add: function (event) {
-            this.notifications.push(event.detail)
+        notifications: {{ Js::from(session()->pull('notifications', [])) }},
+        add (event) {
+            for (let notification of event.detail) {
+                this.notifications.push({
+                    id: notification.id,
+                    status: notification.status,
+                    message: notification.message,
+                })
+            }
         },
         remove: function (notification) {
             this.notifications = this.notifications.filter(i => i.id !== notification.id)
--- a/packages/admin/src/FilamentServiceProvider.php
+++ b/packages/admin/src/FilamentServiceProvider.php
@@ -104,6 +104,8 @@ class FilamentServiceProvider extends PackageServiceProvider
             $this->app->singleton(Component::class, fn () => $component);
         });
 
+        Livewire::listen('component.dehydrate', [NotificationManager::class, 'handleLivewireResponses']);
+
         Livewire::component('filament.core.auth.login', Login::class);
         Livewire::component('filament.core.global-search', GlobalSearch::class);
         Livewire::component('filament.core.pages.dashboard', Dashboard::class);
--- a/packages/admin/src/FilamentServiceProvider.php
+++ b/packages/admin/src/FilamentServiceProvider.php
@@ -100,10 +100,6 @@ class FilamentServiceProvider extends PackageServiceProvider
             MirrorConfigToSubpackages::class,
         ]);
 
-        Livewire::listen('component.hydrate', function ($component) {
-            $this->app->singleton(Component::class, fn () => $component);
-        });
-
         Livewire::listen('component.dehydrate', [NotificationManager::class, 'handleLivewireResponses']);
 
         Livewire::component('filament.core.auth.login', Login::class);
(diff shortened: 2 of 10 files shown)


pxlrbt:
@ryangjchandler or @danharrin could anyone of you have a look at this?

danharrin:
Yeah, didn't think this was a priority so I requested a review from Ryan as he masterminded the notification system. Now it's not marked as a draft, I'll review it soon. Only just caught up on all the issues and PRs from the last few days 👍

pxlrbt:
Definitly not a priority, so take your time. I tagged this as draft to collect some feedback and it's not merged right away. I tested it with the the Filament Demo but I'm not sure whether I overlooked something.

packages/admin/resources/views/components/notification-manager.blade.php
```diff
@@ -1,12 +1,14 @@
...
-        notifications: {{ json_encode(session()->pull('notifications', [])) }},
-        add: function (event) {
-            this.notifications.push({
-                id: event.timeStamp,
-                status: event.detail.status,
-                message: event.detail.message,
-            })
+        notifications: {{ Js::from(session()->pull('notifications', [])) }},
+        add (event) {
+            for (let notification of event.detail) {
```
ryangjchandler, line 5: This for loop can just be an `Array.concat` no?

packages/admin/src/NotificationManager.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace Filament;
+
+use Livewire\Response;
+
+class NotificationManager
+{
+    protected static $notifications = [];
```
ryangjchandler, line 9:
```suggestion
    protected static array $notifications = [];
```

packages/admin/src/NotificationManager.php
```diff
@@ -0,0 +1,36 @@
...
+        self::$notifications = session()->get('notifications');
+    }
+
+    public static function handleLivewireResponses($component, Response $response): Response
+    {
+        if ($component->redirectTo !== null) {
+            return $response;
+        }
+
+        $notifications = self::$notifications;
```
ryangjchandler, line 28:
```suggestion
        $notifications = static::$notifications;
```

packages/admin/src/Http/Livewire/Concerns/CanNotify.php
```diff
@@ -8,6 +8,6 @@ trait CanNotify
 {
     public function notify(string $status, string $message, bool $isAfterRedirect = false): void
```
ryangjchandler, line 9: Well done for keeping the parameter here, removing them would be a breaking change 👍

packages/admin/src/NotificationManager.php
```diff
@@ -0,0 +1,36 @@
...
+        session()->push('notifications', [
+            'id' => uniqid(),
+            'status' => $status,
+            'message' => $message,
+        ]);
+
+        self::$notifications = session()->get('notifications');
+    }
+
+    public static function handleLivewireResponses($component, Response $response): Response
```
ryangjchandler, line 22: Can we typehint the `$component` prop to be a `Livewire\Component` object?

ryangjchandler (changes requested):

packages/admin/src/NotificationManager.php
pxlrbt: I think so. Not sure why I haven't done it.

packages/admin/src/Http/Livewire/Concerns/CanNotify.php
pxlrbt: Any feedback for this?

> And would you want to deprecate that isAfterRedirect param? I thought about #[Deprecated] but that is PhpStorm specific. Alternatively we could trigger a deprecation error.

ryangjchandler: Just use a deprecated DocBlock instead.

pxlrbt:
Thanks @ryangjchandler. I added the requested changes.

danharrin:
Could you rebase when you have a minute please mate

packages/admin/src/Http/Livewire/Concerns/CanNotify.php
pxlrbt: `@deprecated` only works for full methods not for params, I think?

pxlrbt:
Messed up the rebase and started fresh from 2.x with a cherry pick 😅

packages/admin/resources/views/components/notification-manager.blade.php
```diff
@@ -1,8 +1,8 @@
 <div
     x-data="{
-        notifications: {{ json_encode(session()->pull('notifications', [])) }},
-        add: function (event) {
-            this.notifications.push(event.detail)
+        notifications: {{ Js::from(session()->pull('notifications', [])) }},
```
danharrin, line 3: Can we swap this for `@js`? Or, use the fully qualified class name, I don't like aliases

packages/admin/src/FilamentServiceProvider.php
```diff
@@ -98,6 +98,8 @@ protected function bootLivewireComponents(): void
             $this->app->singleton(Component::class, fn () => $component);
```
danharrin, line 98: We can get rid of this binding now AFAIK

packages/admin/resources/views/components/notification-manager.blade.php
pxlrbt: Yeah sure.

packages/admin/src/FilamentServiceProvider.php
pxlrbt: ✅

danharrin:
Nice, ready for review now.

danharrin:
These changes look fine to me. Maybe we should add tests for the notification system generally / the NavigationManager component?

pxlrbt:
I am all in for more tests 😅

danharrin:
This feels like a good opportunity, considering this PR doesn't fix bugs so theres not much time pressure.

ryangjchandler:
@pxlrbt Did you find some time to write tests for this? Don't worry if not, I can help out too :)

pxlrbt:
@ryangjchandler Not yet, as I am currently on vacation until 25th of April. If you want help out, go for it. Otherwise I might look into it in May.

ryangjchandler:
@pxlrbt Not a problem, enjoy your vacation! I'll pick this up later this week.

ryangjchandler:
@danharrin Have written some basic tests for this now:

* Test that notifications get pushed to the session correctly.
* Test that a `notify` event gets dispatched to the browser when no redirect present.
* Test that the `notify` event **doesn't** get dispatched if a redirect is detected.

danharrin:
Thanks to both of you!

PR #1869, opened 2022-03-10
Badges for navigation items

pxlrbt:
Allow users to set a badge on the navigation items. This could be helpful for showing resources that require an action like a reminder resource.

Usage on resources could look like this:
```php
    public static function getNavigationItems(): array
    {
        [$item] = parent::getNavigationItems();
        $item->badge(Reminder::where('is_open')->count());
        return [$item];
    }
```

This is how it looks. I used the styles from the badge column.

Bildschirmfoto 2022-03-10 um 23 13 58

Bildschirmfoto 2022-03-10 um 23 22 12

### DIFF
--- a/packages/admin/docs/05-navigation.md
+++ b/packages/admin/docs/05-navigation.md
@@ -4,7 +4,7 @@ title: Navigation
 
 ## Getting started
 
-By default, Filament will register navigation items for each of your [resources](resources) and [custom pages](pages). These classes contain static properties that you can override, to configure that navigation item and its order:
+By default, Filament will register navigation items for each of your [resources](resources) and [custom pages](pages). These classes contain static properties and methods that you can override, to configure that navigation item and its order:
 
 ```php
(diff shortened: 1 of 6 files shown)


danharrin:
We should define a `getNavigationBadge()` method for resources and pages that makes setting `badge()` easier - like we do for labels, icons and groups.

pxlrbt:
I'm on it.

pxlrbt:
@danharrin Done.

danharrin:
Thank you!

asefsoft:
Thanks but what about logout link on this widget view?
[resources](https://github.com/laravel-filament/filament/tree/2.x/packages/admin/resources)/[views](https://github.com/laravel-filament/filament/tree/2.x/packages/admin/resources/views)/[widgets](https://github.com/laravel-filament/filament/tree/2.x/packages/admin/resources/views/widgets)/account-widget.blade.php

GET route is removed so this logout link throws exception:
The GET method is not supported for this route. Supported methods: POST.

PR #1870, opened 2022-03-11
Improve the logout flow

RVxLab:
This PR makes a couple of improvements to the logout flow, as mentioned in #1821.

Previously logging out happened by means of a GET request which in would technically count as a CSRF vulnerability. This PR changes the HTTP verb to POST to prevent fix this.

Because of this, I expanded the `<x:filament::dropdown.item>` component to allow form tags, which will render an inline form with a single submit button.

Additionally I added a LogoutResponse class with corresponding interface to allow users to modify the logout flow slightly. This still requires a redirect and by default redirects you to the login page.

This is useful for when the application has a custom authentication flow that doesn't rely on Filament's, for example when OAuth2 is being used.


danharrin:
Thanks!

asefsoft:
Thanks but what about logout link on this widget view?
[resources](https://github.com/laravel-filament/filament/tree/2.x/packages/admin/resources)/[views](https://github.com/laravel-filament/filament/tree/2.x/packages/admin/resources/views)/[widgets](https://github.com/laravel-filament/filament/tree/2.x/packages/admin/resources/views/widgets)/account-widget.blade.php

GET route is removed so this logout link throws exception:
The GET method is not supported for this route. Supported methods: POST.

danharrin:
Good point.

alexjustesen:
Opened a PR to fixed `account-widget.blade.php`.

PR #1874, opened 2022-03-11
[BUG] focusedYear null value

wit3:
I don't know how to replicate bug but if you try to use datetime in form, when save ( or reload state ) 
the input trigger error in javascript like this:

```
Uncaught TypeError: Cannot read properties of null (reading 'length')
    at eval (module.esm.js?cbbd:6149:32)
    at eval (module.esm.js?027e:2724:1)
```

i've added a simple optional chaining to property.

[Option chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)


danharrin:
Thank you

PR #1878, opened 2022-03-11
Added custom validation rule for minLength() and maxLength()

alperenersoy:
As you specified on the [documentation](https://filamentphp.com/docs/2.x/forms/fields#text-input) "min" and "max" rules behave differently for numeric inputs. So in order to fix that we can use the "digits_between" rule but it needs minimum and maximum values together. To solve this problem, we can extend the validator with a custom rule that parses value to a string and checks the length of it which is what actually laravel does in [validateDigitsBetween function.](https://github.com/laravel/framework/blob/c0f3ee8c8779d5cfb6bb3c37c93b84672e67f236/src/Illuminate/Validation/Concerns/ValidatesAttributes.php#L578) And for the message we can use the validation messages for "min" and "mix" rules not to add any message for all the languages. You may think it's laravel's behavior but we also have minValue() and maxValue() for inputs so it makes it confusing.

### DIFF
--- a/packages/forms/src/Components/Concerns/CanBeLengthConstrained.php
+++ b/packages/forms/src/Components/Concerns/CanBeLengthConstrained.php
@@ -3,6 +3,7 @@
 namespace Filament\Forms\Components\Concerns;
 
 use Closure;
+use Illuminate\Support\Facades\Validator;
 
 trait CanBeLengthConstrained
 {
@@ -12,12 +13,15 @@ trait CanBeLengthConstrained
 
     public function maxLength(int | Closure $length): static
     {
+        Validator::extend('max_length', fn ($attribute, $value, $parameters) =>  strlen((string) $value) <= $parameters[0], __('validation.max.string'));
+        Validator::replacer('max_length', fn ($message, $attribute, $rule, $parameters) =>  str_replace(':max', $parameters[0], $message));
+
         $this->maxLength = $length;
 
         $this->rule(function (): string {
             $length = $this->getMaxLength();
 
-            return "max:{$length}";
+            return "max_length:{$length}";
         });
 
         return $this;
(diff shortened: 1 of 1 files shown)


danharrin:
Hey, thanks for this. In #1879, we're gonna introduce an interface for possibly-numeric inputs. Instead of adding macros to the validator, we should check for this interface (similar to that PR, in `rule()`), and then use the [`digits_between`](https://laravel.com/docs/9.x/validation#rule-digits-between) rule for numeric inputs.

danharrin:
Hey, thanks for this. In #1879, we're gonna introduce an interface for possibly-numeric inputs. Instead of adding macros to the validator, we should check for this interface (similar to that PR, in `rule()`), and then use the [`digits_between`](https://laravel.com/docs/9.x/validation#rule-digits-between) rule for numeric inputs.

alperenersoy:
> Hey, thanks for this. In #1879, we're gonna introduce an interface for possibly-numeric inputs. Instead of adding macros to the validator, we should check for this interface (similar to that PR, in `rule()`), and then use the [`digits_between`](https://laravel.com/docs/9.x/validation#rule-digits-between) rule for numeric inputs.

hey, adding interface makes sense but I think `digits_between` rule is a little bit problematic for this case because for `maxLength()` we can use "digits_between: 0, max" but for `minLength()` how can we determine the maximum?

danharrin:
Have you tested it with not specifying a max value for min length?

alperenersoy:
Yes I have, it needs both to work properly.

danharrin:
Sounds like we need to PR a better solution to Laravel then

alperenersoy:
okay thanks for your time

danharrin:
laravel/framework#43797

PR #1879, opened 2022-03-11
Added length() function for inputs

alperenersoy:
Sometimes we need exact length for inputs. So instead of using minLength() and maxLength() together we can call length(). To validate, it uses "digits" for numeric inputs and "size" for strings. It just sets minLength and maxLength to the given value since HTML doesn't have an attribute for exact input length. But for numeric inputs, it looks like minlength and maxlength attributes don't work. It may also be fixed with some alpine logic.

### DIFF
--- a/packages/forms/src/Components/Concerns/CanBeLengthConstrained.php
+++ b/packages/forms/src/Components/Concerns/CanBeLengthConstrained.php
@@ -4,11 +4,32 @@ namespace Filament\Forms\Components\Concerns;
 
 use Closure;
 
+
 trait CanBeLengthConstrained
 {
+    protected int | Closure | null $length = null;
+
     protected int | Closure | null $maxLength = null;
 
     protected int | Closure | null $minLength = null;
+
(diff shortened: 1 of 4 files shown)


packages/forms/src/Components/Concerns/CanBeLengthConstrained.php
```diff
@@ -4,12 +4,33 @@
...
+    public function length(int | Closure $length): static
+    {
+        $this->length = $length;
+        $this->maxLength = $length;
+        $this->minLength = $length;
+
+        $this->rule(function (): string {
+            $length = $this->getLength();
+
+            if (method_exists($this, 'isNumeric') && $this->isNumeric())
```
danharrin: Maybe we can refactor this to a `Components\Contracts\CanHaveNumericState` interface for the text input? I like to avoid `method_exists()` checks where possible.

danharrin:
Thank you :) I've written docs

danharrin:
Thank you :) I've written docs

PR #1880, opened 2022-03-11
Fix BelongsToSelect key with custom OptionLabel

MissaelAnda:
When using a custom OptionLabelFromRecord it used the collection id instead of model's id, fixed with mapWithKeys.

### DIFF
--- a/packages/forms/src/Components/BelongsToManyMultiSelect.php
+++ b/packages/forms/src/Components/BelongsToManyMultiSelect.php
@@ -71,7 +71,9 @@ class BelongsToManyMultiSelect extends MultiSelect
             if ($component->hasOptionLabelFromRecordUsingCallback()) {
                 return $relationshipQuery
                     ->get()
-                    ->map(fn (Model $record) => $component->getOptionLabelFromRecord($record))
+                    ->mapWithKeys(fn (Model $record) => [
+                        $record->{$relatedKeyName} => $component->getOptionLabelFromRecord($record)
+                    ])
                     ->toArray();
             }
 
@@ -108,7 +110,9 @@ class BelongsToManyMultiSelect extends MultiSelect
             if ($component->hasOptionLabelFromRecordUsingCallback()) {
                 return $relationshipQuery
                     ->get()
-                    ->map(fn (Model $record) => $component->getOptionLabelFromRecord($record))
+                    ->map(fn (Model $record) => [
+                        $record->getKey() => $component->getOptionLabelFromRecord($record)
+                    ])
                     ->toArray();
             }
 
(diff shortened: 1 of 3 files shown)


pxlrbt:
This needs to be fixed in `BelongsToManySelect` and `BelongsToManyCheckboxes` too.
See this commit:
https://github.com/laravel-filament/filament/pull/1658/commits/138793fe20e94e1d78bd9908b83d8e01fcdf6917

packages/forms/src/Components/BelongsToSelect.php
```diff
@@ -126,7 +128,9 @@ public function relationship(string | Closure $relationshipName, string | Closur
             if ($component->hasOptionLabelFromRecordUsingCallback()) {
                 return $relationshipQuery
                     ->get()
-                    ->map(fn (Model $record) => $component->getOptionLabelFromRecord($record))
+                    ->mapWithKeys(fn (Model $record) => [
+                        $record->{$relationship->getOwnerKeyName()} => $component->getOptionLabelFromRecord($record)
```
pxlrbt, line 132: Do we need `$relationship->getOwnerKeyName()` here? Isn't that the same as `$record->getKey()`?
Before we had:

```suggestion
                        $record->getKey() => $component->getOptionLabelFromRecord($record)
```

packages/forms/src/Components/BelongsToSelect.php
```diff
@@ -99,7 +99,9 @@ public function relationship(string | Closure $relationshipName, string | Closur
...
```
pxlrbt, line 103:
```suggestion
                        $record->getKey() => $component->getOptionLabelFromRecord($record)
```

MissaelAnda:
Added the solution to all BelongsTo coponents and replaced $relationship->getOwnerKeyName() with getKey()

packages/forms/src/Components/BelongsToManyMultiSelect.php
```diff
@@ -71,7 +71,9 @@ public function relationship(string | Closure $relationshipName, string | Closur
             if ($component->hasOptionLabelFromRecordUsingCallback()) {
                 return $relationshipQuery
                     ->get()
-                    ->map(fn (Model $record) => $component->getOptionLabelFromRecord($record))
+                    ->mapWithKeys(fn (Model $record) => [
+                        $record->{$relatedKeyName} => $component->getOptionLabelFromRecord($record)
```
pxlrbt, line 75: Do we need `$relatedKeyName` here or can we use `->getKey()` like on the other components?

```suggestion
                        $record->getKey() => $component->getOptionLabelFromRecord($record)
```

danharrin:
I actually think `getOwnerKeyName()` is a more correct approach than `getKey()`, as you may be using a non-primary column as the owner key in the relationship.

pxlrbt:
> I actually think `getOwnerKeyName()` is a more correct approach than `getKey()`, as you may be using a non-primary column as the owner key in the relationship.

Will you? Isn't that just the key of the related `$record`

MissaelAnda:
getOwnerKeyName() will return the primary key name, whereas getKey() will return the actual value of that column, I think it's the same result but with one less step (except if the model has a mutator on the primary key)

pxlrbt:
I guess `->getKey()` won't work with relations that don't use the primary key of the relation, right? Is this even a thing? 😅 Maybe your first version was better.

MissaelAnda:
Ohhh you're right, reverting now

danharrin:
Thanks gents

PR #1883, opened 2022-03-11
feature: Inline field labels / custom field wrappers

danharrin:
Calling `->usesInlineWrapper()` on any form component will make the all child components have an inline field wrapper:

```php
Forms\Components\Group::make([
    TextInput::make('test'),
    TextInput::make('test2')
        ->hint('Hint')
        ->helperText('Helper text')
        ->prefix('https://')
        ->suffix('.com')
        ->required(),
    Select::make('test3'),
])->usesInlineWrapper(),
```

You can also pass a custom field wrapper view to `->fieldWrapperView()`, which has the same behaviour.


packages/forms/src/Components/Concerns/HasFieldWrapper.php
```diff
@@ -0,0 +1,32 @@
...
+    protected string | Closure | null $fieldWrapperView = null;
+
+    public function fieldWrapperView(string | Closure | null $view): static
+    {
+        $this->fieldWrapperView = $view;
+
+        return $this;
+    }
+
+    public function usesInlineWrapper(bool $condition = true): static
```
ryangjchandler, line 18: Should this accept a `Closure` too, not sure why you'd want it to but it would be consistent then?

ryangjchandler (changes requested):

PR #1885, opened 2022-03-11
Collapsible navigation sidebar

pxlrbt:
Adds an option for a collapsible sidebar. State is persisted across page loads.

https://user-images.githubusercontent.com/22632550/157986876-a4387d98-89a8-4f04-8a74-216da34c7caf.mov


mustafa-online:
This looks great 👌

ralphjsmit:
Very nice Dennis!

danharrin:
Yeah, this is awesome. Thanks!

intrepidws:
Amazing - thanks for doing this. This is exactly what I needed.

intrepidws:
Just noticing that this doesn't work when your `max_content_width` is set to `full`. It seems like an offset is probably needed for the main content area, equal to (or slightly large than) the sidebar width.

pxlrbt:
@intrepidws I will look into this.

asefsoft:
I dont know why it does not show icons correctly on collapse:

image

danharrin:
is it the Windows scrollbar? maybe they need a `shrink-0`

asefsoft:
> is it the Windows scrollbar? maybe they need a `shrink-0`

no it is scrollbar of sidebar

danharrin:
yeah, I was just pointing out that on Windows the scrollbar takes up space that it does not on MacOS. could be causing issues?

asefsoft:
yes scrollbar causing this issue. when there is no scrollbar there is no such issue
I also using Windows

pxlrbt:
@danharrin Not sure what's the best way to solve this. `shrink-0` would solve the issue for icons, but the padding on the right side would be missing and therefore the hover/active state would look weird.

I tried to simulate this with `padding-right: 20px` on the sidebar.

Bildschirmfoto 2022-03-21 um 14 10 31

intrepidws:
Thanks for the quick fix @pxlrbt. I can't stress how helpful this is - I really love it.

A couple more ideas that I think would improve this even more:

- Ability to specify a square logo for the collapsed sidebar
- Tooltips when you mouse over the icons

Feel free to implement either/both if you'd like, otherwise I might try to PR them next week when I have some more time.

asefsoft:
sure tooltips are very necessary for collapse state

PR #1896, opened 2022-03-13
fix localization of datetime pickers

saade:
This fixes the bug where only the last DateTimePicker was being translated to the chosen locale

ps: sorry for the delay on this, been really busy =/


danharrin:
Thank you!

PR #1898, opened 2022-03-14
Fix TimePicker darkmode

Z3d0X:
Darkmode is applied to Timepicker only when used along with DatePicker.
This PR fixes that


danharrin:
Thanks

PR #1901, opened 2022-03-14
Add @ryangjchandler/alpine-tooltip to table's actions and cells, and page's actions

invaders-xx:
add tooltip('Test') functionnality to Page's and Table's actions and also to table's cells.


packages/admin/src/Pages/Actions/Concerns/HasTooltip.php
```diff
@@ -0,0 +1,23 @@
...
+    public function tooltip(string | Closure | null $tooltip): static
+    {
+        $this->tooltip = $tooltip;
+
+        return $this;
+    }
+
+    public function getTooltip(): ?string
+    {
+        return $this->tooltip;
```
ryangjchandler, line 21: This should return `$this->evaluate($this->tooltip)` so that `Closure` arguments get resolved correctly.

packages/tables/resources/views/components/button.blade.php
```diff
@@ -66,6 +67,10 @@
 @if ($tag === 'button')
     <button
         type="{{ $type }}"
+        @if($tooltip)
+        x-data="{ tooltip: '{{ $tooltip }}' }"
```
ryangjchandler, line 71: If the `$tooltip` has any unescaped characters, it will break the rendering in the browser. We should either use `json_encode` or `@js` to properly escape the characters.

This applies to all usages of `{{ $tooltip }}` and similar.

ryangjchandler (changes requested):

packages/admin/src/Pages/Actions/Concerns/HasTooltip.php
invaders-xx: $this->evaluate is not available here

ryangjchandler: Ah, in which case it should be `return value($this->tooltip)`.

packages/tables/resources/views/components/button.blade.php
invaders-xx: my apologies : I missed the last "}"

packages/tables/resources/views/components/button.blade.php
```diff
@@ -66,6 +67,10 @@
 @if ($tag === 'button')
     <button
         type="{{ $type }}"
+        @if($tooltip)
+        x-data="{ tooltip: {{ json_encode($tooltip) }} }"
```
ryangjchandler, line 71: @danharrin What are your thoughts on using the `@js()` directive instead of `json_encode`? It makes debugging things slightly more difficult as Laravel converts more complex structures into base64 encoded strings that get sent through `atob`, but the Blade syntax is a little nicer to look I think...

invaders-xx: Well, the rendering with @js is better than json_code :
x-data="{ tooltip:'Test'}" vs x-data="{ tooltip:"Test"}"

invaders-xx: @danharrin Any thoughts ?

invaders-xx: I finally changed json_encode to @js

invaders-xx:
Can you please tell me what are those conflicts. It might comes from the fact that I added import Tooltip from "@ryangjchandler/alpine-tooltip" to app.js et ran npm. Did I make something wrong ?

ryangjchandler:
@invaders-xx You need to rebase on `2.x` and recompile the assets, that's all.

invaders-xx:
When you say rebase, you mean merge ? Sorry but I am not familiar with all of that

ryangjchandler:
Merging master in is fine, rebases are generally cleaner.

invaders-xx:
I am completely lost on how to rebase on and from. Any help ?

packages/admin/resources/css/app.css
```diff
@@ -1,5 +1,5 @@
 @import '../../../forms/dist/module.esm.css';
-
+@import 'https://unpkg.com/tippy.js@6/dist/tippy.css';
```
ryangjchandler, line 2: Should this not be imported from the `node_modules` directory?

ryangjchandler (changes requested):

ryangjchandler:
@invaders-xx I've rebased for you. You'll need to fetch and pull down your branch again before making further changes.

packages/admin/resources/css/app.css
invaders-xx: sure, my apologies

invaders-xx:
@ryangjchandler , Oups I didn't see your message, as I was making the change for tippy.css. I am sorry, I am so a newbie with such git functionalities and github

invaders-xx:
@ryangjchandler , is it ok for you now ?

invaders-xx:
Do I have to do something to get it approved ?

danharrin:
Thanks!

asefsoft:
Thanks 
there is a problem when we are using custom theme, tippy styles are not available and tooltips are not display correctly.
I think  you should add to [theme documentations](https://filamentphp.com/docs/2.x/admin/appearance#building-themes) that we need to import tippy in our app.css and also needs to install tippy.
`@import '~tippy.js/dist/tippy.css';`

danharrin:
You're right 👍

asefsoft:
It also would be great if we can set HTLM as tooltip
can u enable that?

PR #1906, opened 2022-03-14
Added missing mapWithKeys with relation keys

MissaelAnda:
Accidently left some missing mapWithKeys fields, also changed them to use the relation key instead model's pkey.


danharrin:
Thank you

PR #1907, opened 2022-03-14
Changes logout on the account-widget to POST

alexjustesen:
#1870 changed the logout route to no longer support a `GET` request. This updates `account_widget.blade.php` to send a `POST` request to the logout route.

### DIFF
--- a/packages/admin/resources/views/widgets/account-widget.blade.php
+++ b/packages/admin/resources/views/widgets/account-widget.blade.php
@@ -15,14 +15,16 @@
                     {{ __('filament::widgets/account-widget.welcome', ['user' => \Filament\Facades\Filament::getUserName($user)]) }}
                 </h2>
 
-                <p class="text-sm">
+                <form method="POST" action="{{ route('filament.auth.logout') }}" class="text-sm">
+                    @csrf
+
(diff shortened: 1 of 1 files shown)


packages/admin/resources/views/widgets/account-widget.blade.php
```diff
@@ -15,14 +15,16 @@ class="w-10 h-10 rounded-full bg-gray-200 bg-cover bg-center"
                     {{ __('filament::widgets/account-widget.welcome', ['user' => \Filament\Facades\Filament::getUserName($user)]) }}
                 </h2>
 
-                <p class="text-sm">
+                <form method="POST" action="{{ route('filament.auth.logout') }}" class="text-sm">
+                    @csrf
+
                     <a href="{{ route('filament.auth.logout') }}" @class([
```
danharrin, line 21: Instead of using JavaScript to submit the form, surely we can just make this tag a `<button type="submit">`?

alexjustesen: I ended up going the same route as the breeze methods, but if you prefer a button I can swap it over.

danharrin:
Thank you

PR #1927, opened 2022-03-16
fix: table search problem when use translatable package

FurkanGM:
When use spatie/translatable package, table search not working because LIKE operator is not correctly working when column type is json.

Query;

before:
select count(*) as aggregate from `collections` where (`name` like '%more%')

after:
select count(*) as aggregate from `collections` where (json_unquote(json_extract(`name`, '$."tr"')) like '%more%')

### DIFF
--- a/packages/tables/src/Columns/Concerns/InteractsWithTableQuery.php
+++ b/packages/tables/src/Columns/Concerns/InteractsWithTableQuery.php
@@ -52,8 +52,15 @@ trait InteractsWithTableQuery
             default => 'like',
         };
 
+        $model = $query->getModel();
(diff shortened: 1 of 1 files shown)


danharrin:
I don't think this is the right solution, it'd have issues if the `app.locale` is not a locale in the database, right? We need to read from the locales defined on the resource probably

FurkanGM:
The reason for using app->getLocale() is that records are already listed in the list page according by current language. So, that records already if not has default language translation show another language translation.

danharrin:
Can you fix PHPStan's failing test please?

FurkanGM:
> Can you fix PHPStan's failing test please?

👍 fixed phpstan errors

danharrin:
Thanks

PR #1928, opened 2022-03-16
Added $record parameter to colors callback

tmverkuijlen:
Small change in the callback to activate a color, so you are able to set a color based on a value of an other column, something like:

```
BadgeColumn::make('sum')
    ->colors([
        'success' => fn ($state, $record) => $state < $record['budget'],
        'danger' => fn ($state, $record) => $state >= $record['budget']
     ])
```

### DIFF
--- a/packages/tables/src/Columns/Concerns/HasColors.php
+++ b/packages/tables/src/Columns/Concerns/HasColors.php
@@ -23,7 +23,7 @@ trait HasColors
         foreach ($this->getColors() as $color => $condition) {
             if (is_numeric($color)) {
                 $stateColor = $condition;
-            } elseif ($condition instanceof Closure && $condition($state)) {
+            } elseif ($condition instanceof Closure && $condition($state, $this->getRecord())) {
                 $stateColor = $color;
             } elseif ($condition === $state) {
                 $stateColor = $color;


ryangjchandler (approved):

packages/tables/src/Columns/Concerns/HasColors.php
```diff
@@ -23,7 +23,7 @@ public function getStateColor(): ?string
         foreach ($this->getColors() as $color => $condition) {
             if (is_numeric($color)) {
                 $stateColor = $condition;
-            } elseif ($condition instanceof Closure && $condition($state)) {
+            } elseif ($condition instanceof Closure && $condition($state, $this->getRecord())) {
```
pxlrbt, line 26: Shouldn't we use `$this->evaluate()` here instead?

```suggestion
            } elseif ($condition instanceof Closure && $this->evaluate($condition, ['state' => $state])) {
```

This could break stuff for people relying on argument position instead of name though.

ryangjchandler: Yeah, this would be a breaking change unfortunately.

danharrin: Really? If people are using `$state` as documented, it shouldn't be

pxlrbt: Probably depends on what you see as breaking change. When people adhere strictly to the docs, this shouldn't be an issue. I'd prefer to break it now instead of waiting longer.

ryangjchandler: I'd still avoid it, I know I have at least 1 app that doesn't call it $state.

danharrin: Shame. I'll add it to a breaking changes list for v3.

danharrin:
Thank you

PR #1929, opened 2022-03-16
Add disabled action option

a21ns1g4ts:
Now we can disable action :)
disable

### DIFF
--- a/packages/admin/src/Pages/Actions/Concerns/CanBeDisable.php
+++ b/packages/admin/src/Pages/Actions/Concerns/CanBeDisable.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Filament\Pages\Actions\Concerns;
+
+use Closure;
+
+trait CanBeDisable
+{
+    protected bool | Closure $isDisabled = false;
+
+    public function disable(bool | Closure $condition = true): static
+    {
+        $this->isDisabled = $condition;
+
+        return $this;
+    }
+
+    public function isDisabled(): bool
+    {
+        return $this->isDisabled;
+    }
+}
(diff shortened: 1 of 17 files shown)


packages/admin/src/Pages/Actions/Concerns/CanBeDisable.php
```diff
@@ -0,0 +1,22 @@
...
+    public function disable(bool | Closure $condition = true): static
+    {
+        $this->isDisabled = $condition;
+
+        return $this;
+    }
+
+    public function isDisabled(): bool
+    {
+        return $this->isDisabled;
```
ryangjchandler, line 20:
```suggestion
        return value($this->isDisabled);
```

`isDisabled` can be a `Closure` so we need to evaluate it.

ryangjchandler (changes requested):

packages/admin/src/Pages/Actions/Concerns/CanBeDisable.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Filament\Pages\Actions\Concerns;
+
+use Closure;
+
+trait CanBeDisable
+{
+    protected bool | Closure $isDisabled = false;
+
+    public function disable(bool | Closure $condition = true): static
```
ryangjchandler, line 11:
```suggestion
    public function disabled(bool | Closure $condition = true): static
```

`disabled` here instead. This is the same name we use elsewhere in Filament.

ryangjchandler (changes requested):

ryangjchandler (approved):

danharrin:
Thank you!

PR #1932, opened 2022-03-17
Allow opening navigation item URLs in new tabs

ralphjsmit:
Just like the title.


packages/admin/src/Navigation/NavigationItem.php
```diff
@@ -3,9 +3,12 @@
 namespace Filament\Navigation;
 
 use Closure;
+use Filament\Pages\Actions\Concerns\CanOpenUrl;
```
ryangjchandler, line 6: We shouldn't really leak Page/Action traits into the Navigation space. I think it's fine to duplicate those methods onto this class.

ryangjchandler (changes requested):

packages/admin/src/Navigation/NavigationItem.php
danharrin: Yes. It's completely fine to duplicate code if the abstraction isn't completely perfect 👍

ralphjsmit: Okay, nice, will do!

Personally I hate duplicating code, LOL. Did you also consider creating a sort of general Support namespace or sth. It could be used by all three packages and would be the perfect place for traits like this IMO.

ralphjsmit: PR is updated.

danharrin:
Thanks!

PR #1935, opened 2022-03-18
Fix simple pagination arrows in RTL

asefsoft:
Now arrows are display correctly on both LTR & RTL

was discussed here: https://github.com/laravel-filament/filament/issues/1899

### DIFF
--- a/packages/tables/resources/views/components/pagination/index.blade.php
+++ b/packages/tables/resources/views/components/pagination/index.blade.php
@@ -5,6 +5,11 @@
 
 @php
     $isSimple = ! $paginator instanceof \Illuminate\Pagination\LengthAwarePaginator;
+
+    // set next/prev icon base on language direction
+    $direction = __('filament::layout.direction') ?? 'ltr';
+    $nextIcon = $direction == 'rtl' ? 'left' : 'right';
+    $prevIcon = $direction == 'rtl' ? 'right' : 'left';
 @endphp
 
 <nav
@@ -18,7 +23,7 @@
                 <x-tables::icon-button
                     :wire:click="'previousPage(\'' . $paginator->getPageName() . '\')'"
                     rel="prev"
-                    icon="heroicon-o-chevron-left"
+                    icon="heroicon-o-chevron-{{$prevIcon}}"
                     :label="__('tables::table.pagination.buttons.previous.label')"
                 />
             @endif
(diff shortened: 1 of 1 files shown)


danharrin:
I'm interested, is there a way to do this in CSS instead of PHP? `rtl:rotate-180`

mohamedsabil83:
> I'm interested, is there a way to do this in CSS instead of PHP? `rtl:rotate-180`

Also, could use `rtl:scale-x-[-1]`

asefsoft:
> I'm interested, is there a way to do this in CSS instead of PHP? `rtl:rotate-180`

Yes that would be better if possible
but I am confusing about this file, why used 3 different components to show pagination icons? => x-tables::icon-button, x-tables::button, x-tables::pagination.item

Interesting thing is that last one has `rtl:rotate-180` on its classes but two others dont have that class.
x-tables::icon-button and x-tables::button are generic components and I dont know how can I add this functionality from outside to them.

p.s: as I noted, x-tables::pagination.item already has `rtl:rotate-180` class on it so current PR commit should be changed and should not push.

danharrin:
AFAIK:

`x-tables::icon-button` - mobile pagination buttons
`x-tables::button` - desktop simple pagination buttons
`x-tables::pagination.item` - desktop standard pagination button

`rtl:rotate-180` should do the trick for all of these

danharrin:
Well, turns out that CSS isn't the easiest option here, as the icon abstractions are locked inside of those Blade components. Thank you @asefsoft

asefsoft:
> Well, turns out that CSS isn't the easiest option here, as the icon abstractions are locked inside of those Blade components. Thank you @asefsoft

As I told you [here](https://github.com/laravel-filament/filament/pull/1935#issuecomment-1073223178) x-tables::pagination.item already has `rtl:rotate:180` so then you **should not** add this changes for that component.
did u consider that?
if did not please dont release otherwise there will be issue on that component for arrows.

danharrin:
I'll fix that in a bit

danharrin:
see, this is why i dont do RTL lol

asefsoft:
so far u did great job on rtl 👍👍

mohamedsabil83:
as I mentioned before, for icons, just `rtl:scale-x-[-1]` did the trick. no need for any rotations.

@danharrin you can revert it and I'll review and fix it.

danharrin:
Sure, I've reverted it now @mohamedsabil83

PR #1936, opened 2022-03-18
Allow mutation of settings data

ajnsn:
Add `mutateFormDataBeforeFill` and `mutateFormDataBeforeSave` option as we have with EditRecord.


danharrin:
Perfect, thank you

PR #1938, opened 2022-03-18
fix issue of dark style in time box

maherelgamil:
before:

image

after:

image

And also, test the DateTime box, and everything works fine 

image


danharrin:
Fixed by #1898

PR #1964, opened 2022-03-22
Feature: ActiveRelationManager keys

Z3d0X:
image

Usage is similar to the syntax i suggested in https://github.com/laravel-filament/filament/discussions/1724#discussioncomment-2349148
> ```
> public static function getRelations(): array
> {
>     return [
>         'permissions' => RelationManagers\PermissionsRelationManager::class,
>         'users' => RelationManagers\UsersRelationManager::class,
>     ];
> }
> ```

If no keys are provided it will retain the old way, with the namespace
`activeRelationManager=App%5CFilament%5CResources%5CUserResource%5CRelationManagers%5CPostsRelationManager`


packages/admin/src/Resources/Pages/Concerns/HasRelationManagers.php
```diff
@@ -2,28 +2,33 @@
...
     public $activeRelationManager = null;
 
     protected function getRelationManagers(): array
     {
         $managers = $this->getResource()::getRelations();
-
-        return array_values(array_filter(
+        if(!Arr::isAssoc($managers)) {
+            $managers = array_combine($managers, $managers);
+        };
```
ryangjchandler, line 16: semi-colon?
```suggestion
        }
```

ryangjchandler (changes requested):

danharrin:
I feel like we can simplify this and always use the keys of the array instead of checking `isAssoc`. Would result in relation managers without a key to have `activeRelationManager=1` in the query string, but this is arguably cleaner (but less clear) than using FQNs

ZedoXmv:
I tried without checking  for `isAssoc` at first, but couldn't get it working properly.
checkout to https://github.com/laravel-filament/filament/pull/1964/commits/f4d91af327a15971940797fed76c60e794bdf362
image


not sure if im doing something wrong 🤷. any suggestions?

ryangjchandler:
`===` would probably be checking an `int === string`?

ZedoXmv:
I think the issue is due to non-unique id `id="{{ $activeManager }}" ` . 

https://github.com/laravel-filament/filament/blob/c9cd361474e3477ca8f20cabb2e89153e3005215/packages/admin/resources/views/components/resources/relation-managers/index.blade.php#L37-L46

danharrin:
Thanks

PR #1974, opened 2022-03-22
chore: build assets

ryangjchandler:
Fixes the new tooltip changes.


ryangjchandler:
Dan is faster than me. :(

PR #1977, opened 2022-03-23
fix/discuss: FilePond doesn't initialize correctly when not visible

jacksleight:
This is a potential fix for https://github.com/laravel-filament/filament/issues/1967.

This issue occurs when FilePond is in a secondary tab. In my testing I've found it happens when FilePond initializes inside any element set to `display: none`. I feel like this may actually be a FilePond bug rather than something that should be fixed in Filament, but I came up with this potential solution anyway.

This change uses Alpine's `x-intersect` directive to defer the FilePond initialisation until the element is actually visible. it works, but the delayed initialisation does mean that FilePond doesn't kick in until you switch to the tab, which doesn't look very slick. It also unnecessarily defers fields further down the page until they're scrolled into view.

Not ideal, but something for discussion.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,7 @@
             "devDependencies": {
                 "@alpinejs/collapse": "^3.8.1",
                 "@alpinejs/focus": "^3.8.1",
+                "@alpinejs/intersect": "^3.9.2",
                 "@alpinejs/persist": "^3.8.1",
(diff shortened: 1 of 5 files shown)


jacksleight:
Closing this as it's a FilePond bug: https://github.com/pqina/filepond/issues/819

PR #1978, opened 2022-03-23
Allow ordering of RelationshipRepeater

bernhardh:
Usage:

```php
Forms\Components\HasManyRepeater::make('relationName')
    ->enableOrdering('custom_order_column_name')
```


danharrin:
Thank you very much. The method is now called `orderable()`, and the default column name is `order`.

bernhardh:
Thank you! Just one question: Is it a good idea to use (my)SQL [reserved words](https://dev.mysql.com/doc/refman/8.0/en/keywords.html) like `order` as table column name?

bernhardh:
Sounds good. :)

danharrin:
I'll change the default now

PR #1979, opened 2022-03-23
Extract theme logic to global store and apply theme to button tooltips

pxlrbt:
- Extract theme logic to a global store. 
- Change button/cell tooltip theme based on global theme

### DIFF
--- a/packages/tables/resources/views/components/button.blade.php
+++ b/packages/tables/resources/views/components/button.blade.php
@@ -67,8 +67,8 @@
 @if ($tag === 'button')
     <button
         @if ($tooltip)
-            x-data="{}"
-            x-tooltip.raw="{{ $tooltip }}"
+            x-data="tooltip(@js($tooltip))"
+            x-tooltip="tooltip"
         @endif
         type="{{ $type }}"
         wire:loading.attr="disabled"
@@ -101,8 +101,8 @@
 @elseif ($tag === 'a')
     <a
         @if ($tooltip)
-            x-data="{}"
-            x-tooltip.raw="{{ $tooltip }}"
+            x-data="tooltip(@js($tooltip))"
+            x-tooltip="tooltip"
         @endif
         wire:loading.attr="disabled"
         {!! $hasLoadingIndicator ? 'wire:loading.class="opacity-70 cursor-wait"' : '' !!}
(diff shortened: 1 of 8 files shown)


pxlrbt:
Done.

danharrin:
Do these tooltips work outside of the admin panel now?

pxlrbt:
I don't think so. I only added the theme logic to it. If they didn't work before they probably won't work now.

danharrin:
They did work before, but I wonder how the theming logic that isnt available outside the admin affects that

pxlrbt:
Ah ... damn didn't think about that. I am only using the admin panel.

danharrin:
> Ah ... damn didn't think about that. I am only using the admin panel.

Not expecting you to work further on fixing this if you're not actively benefiting from your work, but I need to mark this as Draft so that I remember to do some work on it myself if you don't want to.

pxlrbt:
@danharrin As I have no experience with the Form and Table Builder as separate packages I'll leave this up to you.

packages/tables/resources/views/components/button.blade.php
```diff
@@ -101,8 +101,8 @@
 @elseif ($tag === 'a')
     <a
         @if ($tooltip)
-            x-data="{}"
-            x-tooltip.raw="{{ $tooltip }}"
+            x-data="tooltip(@js($tooltip))"
+            x-tooltip="tooltip"
```
wychoong, line 105: can we just typeof tooltip !== 'undefined'  check
as the concern is on the new `Alpine.data('tooltip')`
and keep `x-tooltip.raw` as fallback for those not using admin panel

`
x-data="typeof tooltip !== 'undefined' ? tooltip(@js($tooltip)) : {}"
x-tooltip.raw="{{ $tooltip }}"
x-tooltip="typeof tooltip !== 'undefined' ? tooltip : null"
`

however i also don't have standalone table project, but i simulated it seems fine

danharrin: So multiple `x-tooltip` attributes on the same element works?

zepfietje:
What's required for this PR to move forward?

pxlrbt:
Currently, this solution only works for the admin panel, but not for the forms package. I don't have any experience with the standalone forms package, therefore I didn't continue work on this. 
I guess the main issue was detecting dark/light mode on the forms package.

zepfietje:
Hmm yeah, so we basically need the theme logic in the packages as well? 🤔

Did you already think of a solution, @danharrin?

danharrin:
Gonna close this, since it's quite far behind `main` and needs to be reworked taking into account recent changes. We need to find a solution for the subpackages as well.

PR #1987, opened 2022-03-23
Add brand-icon for collapsed navigation

pxlrbt:
Allows a brand icon for the collapsed sidebar. Default is 2 initials of the brand name;

https://user-images.githubusercontent.com/22632550/159794380-73f37e43-f28b-4548-b215-db5778468448.mov

By overwriting the `brand-icon.blade.php` you could add nice SVG icons:

https://user-images.githubusercontent.com/22632550/159794392-e6176421-ad79-4dec-9b09-9b954996e1ab.mov

### DIFF
--- a/packages/admin/resources/views/components/layouts/app/sidebar/index.blade.php
+++ b/packages/admin/resources/views/components/layouts/app/sidebar/index.blade.php
@@ -34,12 +34,10 @@
             <a
                 class="block w-full text-center"
                 href="{{ config('filament.home_url') }}"
-                @if (config('filament.layout.sidebar.is_collapsible_on_desktop'))
-                    x-show="! $store.sidebar.isOpen"
-                    x-transition:enter="lg:transition delay-100"
-                    x-transition:enter-start="opacity-0"
-                    x-transition:enter-end="opacity-100"
-                @endif
+                x-show="! $store.sidebar.isOpen"
+                x-transition:enter="lg:transition delay-100"
+                x-transition:enter-start="opacity-0"
+                x-transition:enter-end="opacity-100"
             >
                 <x-filament::brand-icon />
             </a>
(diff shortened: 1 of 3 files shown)


danharrin:
We should get some simple docs in [this section](https://filamentphp.com/docs/2.x/admin/appearance#changing-the-brand-logo) about how to switch between versions.

pxlrbt:
Not sure what you mean by "how to switch between versions."
But I provided some info on how to set the brand icon.

Not sure whether it should be `components/brand-icon.blade.php` or `components/layouts/app/sidebar/brand-icon.blade.php` as it's only used in the sidebar header so far. What do you think?

packages/admin/resources/views/components/layouts/app/sidebar/index.blade.php
```diff
@@ -30,6 +29,21 @@
         >
             <x-filament::brand />
         </a>
+
+        @if (config('filament.layout.sidebar.is_collapsible_on_desktop'))
+            <a
+                class="block w-full text-center"
+                href="{{ config('filament.home_url') }}"
+                @if (config('filament.layout.sidebar.is_collapsible_on_desktop'))
```
danharrin, line 37: This is always true here

danharrin:
> Not sure whether it should be `components/brand-icon.blade.php` or `components/layouts/app/sidebar/brand-icon.blade.php` as it's only used in the sidebar header so far. What do you think?

`components/brand-icon.blade.php` is good, I think. Then we can use it elsewhere.

packages/admin/resources/views/components/layouts/app/sidebar/index.blade.php
pxlrbt: Oops. You're right. Time for some sleep

danharrin:
Stellar as always. Thank you.

asefsoft:
Hi
`str` helper function just added to Laravel 9 and it is not available on older versions of Laravel. so using it here will break compatibility of filamnet with Laravel 8 and ...

Please use another methods or just use Str facade.

I just updated filament and faced with this exception on all pages:
`Call to undefined function str() (View: ..\vendor\filament\filament\resources\views\components\brand-icon.blade.php`

pxlrbt:
@asefsoft Thanks. Already got a report on Discord and opened a PR with a fix: https://github.com/laravel-filament/filament/pull/1990

PR #1990, opened 2022-03-24
Fix: str() helper not compatible with L8

pxlrbt:
`str()` was introduced with L9 and causes issues with L8.

### DIFF
--- a/packages/admin/resources/views/components/brand-icon.blade.php
+++ b/packages/admin/resources/views/components/brand-icon.blade.php
@@ -8,7 +8,7 @@
                 ->snake()
                 ->upper()
                 ->explode('_')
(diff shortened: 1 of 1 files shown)


asefsoft:
Thanks,
I suggest not using `substr` which is not multi-byte safe and will cause issue on Unicode strings. you can use mb_substr or just use Laravel Str facade here too.

pxlrbt:
Thanks. Adjusted.

asefsoft:
Thanks. 
do you think is it not better that to have a tooltip on logo too when it is collapsed or not?

pxlrbt:
I don't see any use for a tooltip here. It's just aesthetical.

danharrin:
Thanks

PR #1991, opened 2022-03-24
Slight hover effect for table rows

pxlrbt:
As table rows are clickable, I think they would benefit from a slight hover effect:

https://user-images.githubusercontent.com/22632550/159869426-990cb598-3724-4e56-8282-7374cf506208.mov

### DIFF
--- a/packages/tables/resources/views/components/row.blade.php
+++ b/packages/tables/resources/views/components/row.blade.php
@@ -1,3 +1,13 @@
-<tr {{ $attributes->class(['hover:bg-primary-500/5 dark:hover:bg-primary-500/5 filament-tables-row']) }}>
+@props([
+    'recordUrl' => null,
+])
+<tr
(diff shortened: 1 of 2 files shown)


pxlrbt:
Yep, you're right. Fixed.

packages/tables/resources/views/components/row.blade.php
```diff
@@ -1,3 +1,13 @@
-<tr {{ $attributes->class(['filament-tables-row']) }}>
+@props([
+    'recordUrl' => null,
+])
```
ryangjchandler, line 3:
```suggestion
])

```

ryangjchandler (changes requested):

danharrin:
Do we need to run the closure or just check it exists?

danharrin:
I think you're right actually, ignore me

PR #1993, opened 2022-03-24
feature: show icons in link actions

ryangjchandler:
Button actions have icons and the `LinkAction` class has an `icon()` method too, but the icons don't actually show up. This sorts that out :)

CleanShot 2022-03-24 at 09 23 23@2x


pxlrbt:
Nice one.

PR #1994, opened 2022-03-24
fix select action of pages with dark mode

maherelgamil:
Before:

image

After:

image

### DIFF
--- a/packages/admin/resources/views/pages/actions/select-action.blade.php
+++ b/packages/admin/resources/views/pages/actions/select-action.blade.php
@@ -6,10 +6,10 @@
     <select
         id="{{ $getId() }}"
         wire:model="{{ $getName() }}"
-        @class([
-            'text-gray-900 border-gray-300 block h-10 transition duration-75 rounded-lg shadow-sm focus:border-primary-600 focus:ring-1 focus:ring-inset focus:ring-primary-600',
+        {{ $attributes->class([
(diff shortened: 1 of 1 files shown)


packages/admin/resources/views/pages/actions/select-action.blade.php
```diff
@@ -6,9 +6,10 @@
     <select
         id="{{ $getId() }}"
         wire:model="{{ $getName() }}"
-        {{ $attributes->class([
-            'text-gray-900 border-gray-300 invalid:text-gray-400 block w-full h-9 py-1 transition duration-75 rounded-lg shadow-sm focus:border-primary-600 focus:ring-1 focus:ring-inset focus:ring-primary-600',
-        ]) }}
+        @class([
```
danharrin, line 9: This can remain as `{{ $attributes->class([]) }}`, which supports the same syntax as `@class`.

packages/admin/resources/views/pages/actions/select-action.blade.php
```diff
@@ -6,9 +6,10 @@
...
+            'text-gray-900 border-gray-300 block h-10 transition duration-75 rounded-lg shadow-sm focus:border-primary-600 focus:ring-1 focus:ring-inset focus:ring-primary-600',
```
danharrin, line 10:
```suggestion
            'text-gray-900 border-gray-300 block h-9 py-1 transition duration-75 rounded-lg shadow-sm focus:border-primary-600 focus:ring-1 focus:ring-inset focus:ring-primary-600',
```

danharrin: we did this so that the select is the same height as buttons. please check

danharrin:
Thanks

Issue #1899, opened 2022-03-14
Simple patination arrows in RTL should be swapped


asefsoft:
### Package

filament/tables

### Package Version

2.10.27

### Laravel Version

8

### PHP Version

8

### Bug description

Hello
In RTL environment this arrows of simple pagination must be swapped with each other.

current status in RTL:
![image](https://user-images.githubusercontent.com/46367425/158148850-b39cc619-c32e-4dad-879a-a12e07d5b7dd.png)

it shoud be like this:
![image](https://user-images.githubusercontent.com/46367425/158149928-0fa205ec-f7fc-4e93-9bf1-0c078030baf8.png)

Thanks

### Steps to reproduce

Set locale of your laravel app to a RTL language in config.php file
`    'locale' => 'fa',`

then create a filement table with simple pationation to see problem

danharrin:
Hey, while this project has an RTL mode, this completely depends on community support. None of the maintainers have experience designing RTL interfaces. However, we'll never reject PRs from anyone looking to improve the RTL experience themselves. If you'd like to see this fixed, please submit a PR.

PR #1998, opened 2022-03-24
Fix simple pagination arrows in RTL

mohamedsabil83:
This PR replacing #1935 for fixing issue #1899

### DIFF
--- a/packages/tables/resources/views/components/button.blade.php
+++ b/packages/tables/resources/views/components/button.blade.php
@@ -45,7 +45,7 @@
     ]);
 
     $iconClasses = \Illuminate\Support\Arr::toCssClasses([
-        'filament-button-icon',
+        'filament-button-icon rtl:scale-x-[-1]',
         'w-6 h-6' => $size === 'md',
         'w-7 h-7' => $size === 'lg',
         'w-5 h-5' => $size === 'sm',
--- a/packages/tables/resources/views/components/icon-button.blade.php
+++ b/packages/tables/resources/views/components/icon-button.blade.php
@@ -20,7 +20,7 @@
         'opacity-70 cursor-not-allowed' => $disabled,
     ];
 
-    $iconClasses = 'w-5 h-5 filament-tables-icon-button-icon';
+    $iconClasses = 'w-5 h-5 rtl:scale-x-[-1] filament-tables-icon-button-icon';
 @endphp
 
 @if ($tag === 'button')
--- a/packages/tables/resources/views/components/pagination/item.blade.php
+++ b/packages/tables/resources/views/components/pagination/item.blade.php
@@ -23,7 +23,7 @@
         ]) }}
     >
         @if ($icon)
-            <x-dynamic-component :component="$icon" class="w-5 h-5 rtl:rotate-180" />
+            <x-dynamic-component :component="$icon" class="w-5 h-5 rtl:scale-x-[-1]" />
         @endif
 
         <span>{{ $label ?? ($separator ? '...' : '') }}</span>


danharrin:
Should all buttons have their icons flipped in RTL like that? 🤔

asefsoft:
> Should all buttons have their icons flipped in RTL like that? 🤔

I really don't think so and it is very risky!

danharrin:
Which is why I didn't do `rotate-180` :D

mohamedsabil83:
> Should all buttons have their icons flipped in RTL like that? thinking

With this, it's like a mirror. until now, it useful with all icons I have used

danharrin:
Okay 🤷‍♂️ you're the RTL user so I trust you :)

asefsoft:
image

Which one do u think is ok?
first one has `rtl:scale-x-[-1]`
as i said it is risky and not recommended

mohamedsabil83:
the three components mentioned are for arrows in simple paginations, right?

danharrin:
😬 

you two let me know your final decision, if you need to change anything, submit a new PR

mohamedsabil83:
@danharrin which component renders the trademark shown above?

pxlrbt:
You mean the checkmark? It's the `BooleanColumn`

mohamedsabil83:
Is it use one of these to render icon?
- packages/tables/resources/views/components/button.blade.php
- packages/tables/resources/views/components/icon-button.blade.php
- packages/tables/resources/views/components/pagination/item.blade.php

asefsoft:
it is obvious that we should not mirror all icons just bcoz they are in rtl mode!
two of those components are generic components

mohamedsabil83:
They are mentioned as pagination buttons for that I update them all.
Can you confirm which one is related to the simple pagination specifically?

mohamedsabil83:
> image
> 
> Which one do u think is ok? first one has `rtl:scale-x-[-1]` as i said it is risky and not recommended

@asefsoft can you share the schema to try re-introduced that issue?

mohamedsabil83:
@danharrin can you check if it works as expected? it's ok with me 🤷🏽‍♂️ 
Screenshot from 2022-03-24 15-55-59

asefsoft:
I did it manually to show why we should not mirror every icons. 

if u search filament code, u will see that x-tables::button and x-tables::icon-button are generic and used in many places like actions and filters... and if we flip those icons, then whenever an icon is used in those places it will be flipped and potentially makes situations like that image above

asefsoft:
if u need real example:
image
just add this icon to an action button and see what's happen

```
->prependActions([
            
            Tables\Actions\ButtonAction::make('add_process')->label("ثبت بررسی")
                ->icon('heroicon-o-check')
                ->action('addProcess')
```

mohamedsabil83:
@danharrin I think we can do one of those to fix this conflict:
- new component for pagination button only. or,
- new tag option named e.g. **pagination** or **arrow** similar to **button** tag just apply the class `scale-x-[-1]` which used only for pagination or places that render arrows only

asefsoft:
> we just need to pass a different icon in for RTL, like the original PR. Except we shouldn't do it for the pagination-specific buttons as they aready use CSS.

yes, please 🤷‍♂️🤦‍♂️👍

mohamedsabil83:
Back to the future 😆 
It's up to you @asefsoft, check if only one you changed before or if there is another and send a PR.

asefsoft:
> Back to the future 😆 It's up to you @asefsoft, check if only one you changed before or if there is another and send a PR.

I just want kill myself 😂😂

asefsoft:
> If neither of you want to fix it, I can

you already fixed it, please apply it again. 
thanks

mohamedsabil83:
I'll check and fix it

PR #1999, opened 2022-03-24
show sidebar group collapse icon when sidebar is collapsed

maherelgamil:
this is the final result

image

### DIFF
--- a/packages/admin/resources/views/components/layouts/app/sidebar/group.blade.php
+++ b/packages/admin/resources/views/components/layouts/app/sidebar/group.blade.php
@@ -6,9 +6,9 @@
     @if ($label)
         <button
             x-on:click.prevent="$store.sidebar.toggleCollapsedGroup(label)"
-            class="flex items-center justify-between w-full"
+            class="flex items-center justify-between w-full rtl:space-x-1"
         >
-            <p 
+            <p
                 @if (config('filament.layout.sidebar.is_collapsible_on_desktop'))
                     x-show="$store.sidebar.isOpen"
                 @endif
--- a/packages/admin/resources/views/components/layouts/app/sidebar/group.blade.php
+++ b/packages/admin/resources/views/components/layouts/app/sidebar/group.blade.php
@@ -20,8 +20,7 @@
                 {{ $label }}
             </p>
 
-            <div class="ml-3">
-
+            <div class="ml-3 rtl:ml-auto rtl:mr-3">
                 <x-heroicon-o-chevron-down :class="\Illuminate\Support\Arr::toCssClasses([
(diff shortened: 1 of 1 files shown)


danharrin:
Thoughts @pxlrbt?

maherelgamil:
> Thoughts @pxlrbt?

Just we help his thoughts 😄

pxlrbt:
I can see that the collapsed sidebar is not that good when you have a lot of collapsed groups. But I feel like this clutters the collapsed sidebar. You still have no idea which caret belongs to which group. This would at least require a tooltip for the carets.

In my case I'd prefer to toggle the sidebar open the group I need and collapse it again. But I also don't have projects with that many resources yet.

maherelgamil:
@pxlrbt @danharrin 

I think 
> toggle the sidebar open the group I need and collapse it again

is a good solution for this case, I thought about this solution also and tried to solve the current state.

so If there is consensus on this, I will start to change it

ryangjchandler:
I wonder if maybe we uncollapse groups when the sidebar is closed?

pxlrbt:
> I thought about this solution also and tried to solve the current state.

What do you mean? That's the current state. Manually open the sidebar, adjust your collapsed groups and collapse the sidebar again. But it's more manual work.

pxlrbt:
> I wonder if maybe we uncollapse groups when the sidebar is closed?

Thought about this too when I first implemented this. But I thought that groups are collapsed for a reason and didn't want to interfere with the users decision. At least we should restore them when opening the sidebar again.

ryangjchandler:
Yeah, definitely restore them. If we use `$store.sidebar.isOpen` inside of the collapse logic for the groups, it'll just work anyway.

maherelgamil:
@pxlrbt 

I mean, trying to fix the current shape, like this 

image

maherelgamil:
> I wonder if maybe we uncollapse groups when the sidebar is closed?

I think also, it's a good solution in this case

ryangjchandler:
How about this: 
CleanShot 2022-03-24 at 11 15 57

danharrin:
I back Ryan's suggestion

pxlrbt:
Yeah, uncollapsing them is probably the best solution.

pxlrbt:
I thought it would shift the items too much but the current solutions looks nasty with many collapsed groups 😅

pxlrbt:
@ryangjchandler As you already have it up and running do you wanna PR that? 😅

ryangjchandler:
Yeah, might as well open the pr haha

ryangjchandler:
Closing in favour of https://github.com/laravel-filament/filament/pull/2003.

PR #2005, opened 2022-03-24
feat: added translation select action to list page

FurkanGM:
This pr, adds translation filter feature to listing page

### DIFF
--- a/packages/spatie-laravel-translatable-plugin/resources/lang/ar/actions.php
+++ b/packages/spatie-laravel-translatable-plugin/resources/lang/ar/actions.php
@@ -2,7 +2,7 @@
 
 return [
 
-    'active_form_locale' => [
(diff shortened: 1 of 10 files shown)


danharrin:
Hmm, the issue with this solution is that it leaks translatable code into the tables package... 🤔 You'll need to find a way to keep it all within the translatable package, maybe you need to introduce a new API to work around that limitation

FurkanGM:
is there any problem if we use the macro ``applySearchConstraint`` method for the Column class in SpatieLaravelTranslatablePluginServiceProvider? Using this, we are not directly leaks with the tables package.

example
```php
\Filament\Tables\Columns\Column::macro('applySearchConstraint', function(){});
```

danharrin:
Fixed it up, thanks!

PR #2009, opened 2022-03-24
📚 Docs: Replace simple `configurePackage()` examples with `static $name`

pxlrbt:
`PluginServiceProvider::configurePackage()` does a lot of bootstrapping but in the examples it's always overwritten.
Suggest using the static properties for automatic registration.

### DIFF
--- a/packages/admin/docs/08-plugins.md
+++ b/packages/admin/docs/08-plugins.md
@@ -59,31 +59,6 @@ Much like a normal Laravel package, you should add your service provider's fully
 
 This will ensure your service provider is automatically loaded by Laravel when the package is installed.
 
-## FilamentServing Event
-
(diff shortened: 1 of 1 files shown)


ryangjchandler:
@pxlrbt Think you made these changes from your other branch instead of 2.x btw.

pxlrbt:
> @pxlrbt Think you made these changes from your other branch instead of 2.x btw. You can probably cherry pick the commit into a new branch.

Ah damn. Made these changes via GitHub directly. Not sure how it ended up in here as well. It's just 1 commit 😅

Will have a look at it later

pxlrbt:
Fixed.

danharrin:
Thanks, but I'm not gonna merge this. Our docs were like this before, but people were then getting confused about how to interact with the package object to register other assets. I haven't had any complaints about the docs being this way yet.

PR #2011, opened 2022-03-24
Extend GlobalSearchResult.php to make it more useful for plugins

pxlrbt:


ryangjchandler:
@pxlrbt What makes this more useful for plugins?

pxlrbt:
E.g. for Filament Spotlight I can use `$resource::getGlobalSearchResults($query)` and create my own url based on the result set. Currently I get the fixed view or edit url. 

Would be a BC as we changed the order of params, but Dan expected nobody to use it 😅

PR #2014, opened 2022-03-25
Improve fa translation

themahdavi:
this PR includes improvements to **'fa' translation** and **add associate and dissociate** .


danharrin:
Thanks

PR #2017, opened 2022-03-25
feature: repeater and builder item drag and drop ordering

jacksleight:
Drag and drop ordering for repeater items:

https://user-images.githubusercontent.com/126740/160114202-a1a2a39b-44e6-438b-a466-0bf419be9c4c.mov

### DIFF
--- a/packages/forms/src/Components/Repeater.php
+++ b/packages/forms/src/Components/Repeater.php
@@ -115,6 +115,26 @@ class Repeater extends Field
                     data_set($livewire, $statePath, $items);
                 },
             ],
+            'repeater::moveItems' => [
+                function (Repeater $component, string $statePath, array $uuids): void {
+                    if ($component->isDisabled()) {
+                        return;
+                    }
+
+                    if ($component->isItemMovementDisabled()) {
+                        return;
+                    }
+
+                    if ($statePath !== $component->getStatePath()) {
+                        return;
+                    }
+
+                    $items = array_merge(array_flip($uuids), $component->getState());
+
+                    $livewire = $component->getLivewire();
+                    data_set($livewire, $statePath, $items);
+                },
+            ],
         ]);
 
         $this->createItemButtonLabel(function (Repeater $component) {
--- a/packages/forms/src/helpers.php
+++ b/packages/forms/src/helpers.php
@@ -9,6 +9,7 @@ if (! function_exists('Filament\Forms\array_move_after')) {
 
         $indexToMoveAfter = array_search($keyToMoveAfter, $keys);
         $keyToMoveBefore = $keys[$indexToMoveAfter + 1] ?? null;
+        
         if (filled($keyToMoveBefore)) {
             return $array;
         }
--- a/packages/forms/src/helpers.php
+++ b/packages/forms/src/helpers.php
@@ -9,7 +9,7 @@ if (! function_exists('Filament\Forms\array_move_after')) {
 
         $indexToMoveAfter = array_search($keyToMoveAfter, $keys);
         $keyToMoveBefore = $keys[$indexToMoveAfter + 1] ?? null;
-        if (!isset($keyToMoveBefore)) {
+        if (filled($keyToMoveBefore)) {
             return $array;
         }
 
--- a/packages/forms/src/helpers.php
+++ b/packages/forms/src/helpers.php
@@ -34,7 +34,7 @@ if (! function_exists('Filament\Forms\array_move_before')) {
 
         $indexToMoveBefore = array_search($keyToMoveBefore, $keys);
         $keyToMoveAfter = $keys[$indexToMoveBefore - 1] ?? null;
-        if (!isset($keyToMoveAfter)) {
+        if (filled($keyToMoveAfter)) {
             return $array;
         }
 
--- a/packages/forms/src/helpers.php
+++ b/packages/forms/src/helpers.php
@@ -8,7 +8,10 @@ if (! function_exists('Filament\Forms\array_move_after')) {
         $keys = array_keys($array);
 
         $indexToMoveAfter = array_search($keyToMoveAfter, $keys);
-        $keyToMoveBefore = $keys[$indexToMoveAfter + 1];
+        $keyToMoveBefore = $keys[$indexToMoveAfter + 1] ?? null;
+        if (!isset($keyToMoveBefore)) {
+            return $array;
+        }
 
         $keys[$indexToMoveAfter + 1] = $keyToMoveAfter;
         $keys[$indexToMoveAfter] = $keyToMoveBefore;
@@ -30,7 +33,10 @@ if (! function_exists('Filament\Forms\array_move_before')) {
         $keys = array_keys($array);
 
         $indexToMoveBefore = array_search($keyToMoveBefore, $keys);
-        $keyToMoveAfter = $keys[$indexToMoveBefore - 1];
+        $keyToMoveAfter = $keys[$indexToMoveBefore - 1] ?? null;
+        if (!isset($keyToMoveAfter)) {
+            return $array;
+        }
 
         $keys[$indexToMoveBefore - 1] = $keyToMoveBefore;
         $keys[$indexToMoveBefore] = $keyToMoveAfter;
--- a/packages/forms/src/helpers.php
+++ b/packages/forms/src/helpers.php
@@ -34,6 +34,7 @@ if (! function_exists('Filament\Forms\array_move_before')) {
 
         $indexToMoveBefore = array_search($keyToMoveBefore, $keys);
         $keyToMoveAfter = $keys[$indexToMoveBefore - 1] ?? null;
+        
         if (filled($keyToMoveAfter)) {
             return $array;
         }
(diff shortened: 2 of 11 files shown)


danharrin:
Thanks! Few points...
- Test failing
- I don't think we need extra button?
- Does this work with nested repeaters?
- How about the builder?

pxlrbt:
Maybe we should integrate our own Livewire integration of SortableJS. It's less than 100 LOC in this case and we don't need to rely on 3rd party packages. There is not much activity on that repo currently.

pxlrbt:
Btw tests are failing because you bumped min PHP version on 2.x 😅

jacksleight:
Yeah test error seems to be because of the composer requirements?

Works fine with nested repeaters.

Will change the drag handle as @pxlrbt suggested, and look at builder as well.

jacksleight:
OK, made the suggested changes. I'm a bit unsure on how/where is best to import the custom Livewire directive, shout if it's wrong.

New drag handle style:

Screenshot 2022-03-25 at 15 17 12

pxlrbt:
Not sure how to improve this, but the drag handle feels a bit off 🙈

jacksleight:
Mmmm yeah, I do feel it's a bit close to the field label.

pxlrbt:
> Edit: Also the Livewire directive is basically a copy paste from the other package, but could be customised now it's in Filament.

Wasn't the other implementation about 75 lines instead of 30? 🤔

jacksleight:
Oh, no they're self-contained at the moment, you can't move something from one nested to repeater to a parent/child repeater at the moment... Do the regular up/down buttons support that?

pxlrbt:
Sure, you're right. Maybe that's out of scope for now.

jacksleight:
Yeah, certainly be nice to have, but maybe one for another day 🙂 .

jacksleight:
Now we're using a custom Livewire directive we shouldn't need the extra dispatch event methods anymore. I'll refactor that.

wychoong:
maybe have the drag handle at the top center, heroicon-s-dots-horizontal, like ipad multitask dots?

panakour:
or maybe with a settings tooltip which let have more actions in the future, something like octobercms repeater
Screenshot from 2022-03-26 13-14-56

jacksleight:
I've knocked up a few drag handle variations, 1 is the current version.

I'd personally lean towards 4, maybe 3, but I'm happy to implement whichever people prefer, or something else entirely. Let me know your thoughts.

Screenshot 2022-03-26 at 16 04 22
Screenshot 2022-03-26 at 16 03 45
Screenshot 2022-03-26 at 16 06 47
Screenshot 2022-03-26 at 16 07 41

panakour:
I also prefer the 4

martinmildner:
What about an up and down arrow?
Screenshot-2022-03-27 at 14 25 53@2x

pxlrbt:
> What about an up and down arrow? Screenshot-2022-03-27 at 14 25 53@2x

Check out the video in the initial PR comment ;)

danharrin:
I think 3 is my favourite, but we can get rid the old up / down buttons maybe? What do you think?

wychoong:
I feel 4 is crowded but of course I’m biased towards 3 😆 
If it’s not too hard to maintain, maybe ->showDragHandle(left/right/center) and ->hideSortButtons()

pxlrbt:
I guess with Drag & Drop we could remove the Up/Down arrow and get some space.

What about something like this? Inspired by other block editors and the October CMS solution. It takes up some more space but also gives a bigger dragging handle (whole grey background area)

1) Left
Bildschirmfoto 2022-03-27 um 17 11 32

2) Right
Bildschirmfoto 2022-03-27 um 17 19 42

3) Top
Bildschirmfoto 2022-03-27 um 17 30 20

martinmildner:
+1 for 5 (aligned top left)

jacksleight:
I really like @panakour and @pxlrbt’s ideas of having a larger handle that could include the other actions as well. And possibly making this configurable as @wychoong suggested. 

But that probably needs more discussion, and I think it’d be good to get the core drag and drop functionality in place, so I’m gonna focus on that for this PR.

> I think 3 and 4 are my favourites, but we can get rid the old up / down buttons maybe? What do you think?

I’ve implemented 4 for now. That’s the simplest solution with the smallest UI impact.

I’ve removed the up/down buttons but added arrow up/down key listeners to the drag handle, so it’s still keyboard accessible. Livewire seems to get really glitchy if you add/remove these listeners on changes, so instead of toggling them in the view like the old buttons were I've added a check to the `array_move_after` and `array_move_before` functions.

packages/forms/src/helpers.php
```diff
@@ -8,7 +8,10 @@ function array_move_after(array $array, $keyToMoveAfter): array
         $keys = array_keys($array);
 
         $indexToMoveAfter = array_search($keyToMoveAfter, $keys);
-        $keyToMoveBefore = $keys[$indexToMoveAfter + 1];
+        $keyToMoveBefore = $keys[$indexToMoveAfter + 1] ?? null;
+        if (!isset($keyToMoveBefore)) {
```
danharrin, line 12:
```suggestion
        if (filled($keyToMoveBefore)) {
```

packages/forms/src/helpers.php
```diff
@@ -30,7 +33,10 @@ function array_move_before(array $array, $keyToMoveBefore): array
         $keys = array_keys($array);
 
         $indexToMoveBefore = array_search($keyToMoveBefore, $keys);
-        $keyToMoveAfter = $keys[$indexToMoveBefore - 1];
+        $keyToMoveAfter = $keys[$indexToMoveBefore - 1] ?? null;
+        if (!isset($keyToMoveAfter)) {
```
danharrin, line 37:
```suggestion
        if (filled($keyToMoveAfter)) {
```

packages/forms/src/helpers.php
```diff
@@ -30,7 +33,10 @@ function array_move_before(array $array, $keyToMoveBefore): array
         $keys = array_keys($array);
 
         $indexToMoveBefore = array_search($keyToMoveBefore, $keys);
-        $keyToMoveAfter = $keys[$indexToMoveBefore - 1];
+        $keyToMoveAfter = $keys[$indexToMoveBefore - 1] ?? null;
```
danharrin, line 36:
```suggestion
        $keyToMoveAfter = $keys[$indexToMoveBefore - 1] ?? null;
        
```

packages/forms/src/helpers.php
```diff
@@ -8,7 +8,10 @@ function array_move_after(array $array, $keyToMoveAfter): array
         $keys = array_keys($array);
 
         $indexToMoveAfter = array_search($keyToMoveAfter, $keys);
-        $keyToMoveBefore = $keys[$indexToMoveAfter + 1];
+        $keyToMoveBefore = $keys[$indexToMoveAfter + 1] ?? null;
```
danharrin, line 11:
```suggestion
        $keyToMoveBefore = $keys[$indexToMoveAfter + 1] ?? null;
        
```

packages/forms/src/Components/Repeater.php
```diff
@@ -111,6 +111,26 @@ function (Repeater $component, string $statePath, string $uuidToMoveUp): void {
 
                     $items = array_move_before($component->getState(), $uuidToMoveUp);
 
+                    $livewire = $component->getLivewire();
+                    data_set($livewire, $statePath, $items);
+                },
+            ],
+            'repeater::moveItems' => [
```
danharrin, line 118: Since we add this new event listener for moving items, can we remove the old up/down button listeners for Repeater & Builder?

invaders-xx: Maybe I am late in the conversation, but would it be possible to have the Drap&Drop placement driven by config. I don't know the complexity in implementing that

jacksleight: The up/down listeners are still used by the arrow-up/arrow-down key events on the drag handle. Without those you wouldn't be able to move items via the keyboard.

a21ns1g4ts:
> I think 3 and 4 are my favourites, but we can get rid the old up / down buttons maybe? What do you think?

Yep. we can keep all clean

danharrin:
Thank you very much! 🍻

danharrin:
Hey @jacksleight, after multiple bug reports it seems that this breaks sorting for nested repeaters. I've opened a PR which removes this feature here - #2156.

I really don't want to merge it, but I've spent a while trying to find a solution and I cannot. It feels like Livewire is conflicting with the sortable package, which causes the moved items to appear above the parent repeater (DOM-diffing issue?). @ryangjchandler or @jacksleight if you're able to look into this today or tomorrow, that would be great. But if you don't have any time then that's absolutely fine, but I will need to merge #2156, probably tomorrow afternoon.

danharrin:
To replicate the issue:

```php
Repeater::make('options')
    ->schema([
        Repeater::make('values')
            ->schema([
                TextInput::make('value'),
            ])
    ]),
```

Add two items to the `values` repeater and then swap their positions.

danharrin:
Thanks @ryangjchandler, fixed by #2164.

jacksleight:
Hey, sorry I've been away so didn't catch this, wasn't ignoring it 😄 . Glad you found a fix!

jacksleight:
@danharrin Hey, sorry I've been away so didn't catch this, wasn't ignoring it 😄 . Glad you found a fix!

danharrin:
No worries mate, all sorted! Hope you had a good break

Issue #1985, opened 2022-03-23
FileUpload Locale it does not work correctly


themahdavi:
### Package

filament/forms

### Package Version

2.10

### Laravel Version

9.2

### Livewire Version

2.6

### PHP Version

8.0.3

### Bug description

When change Laravel locale the Filament all packages set to new locale, design and more thing change dynamically.
but in FileUpload form component when change the locale of Laravel project this component try to set correct locale of Filepond package and search **[https://cdn.skypack.dev/filepond/locale/](https://cdn.skypack.dev/filepond/locale/)** CDN to get locale translation and not find some languages because locale filename of all locale not same as (en-en) 

from this page [https://github.com/pqina/filepond/tree/master/locale](https://github.com/pqina/filepond/tree/master/locale)
for example : da-dk.js , fa_ir.js, no-nb.js

and shown error in console get request for locale js file not respond!

for solution : define option to FileUpload component can handle filename of locale of Filepond package 🤙 
anyone have best solution ?

themahdavi:
in FileUpload form component Link for handle locale : 
[https://github.com/laravel-filament/filament/blob/7865a0f2d97c645e99497de94fc110ee83735d36/packages/forms/resources/views/components/file-upload.blade.php#L14](https://github.com/laravel-filament/filament/blob/7865a0f2d97c645e99497de94fc110ee83735d36/packages/forms/resources/views/components/file-upload.blade.php#L14)

Zagna:
Skypack seems to only have Filepond, no locales. Switch to https://cdn.jsdelivr.net/npm/filepond@4/locale/?

themahdavi:
CDN link not problem

filename cause problem

[in this ](https://github.com/laravel-filament/filament/blob/7865a0f2d97c645e99497de94fc110ee83735d36/packages/forms/resources/views/components/file-upload.blade.php#L14) link logic of send filename to CDN not completely load every links


because for example code generate fa-fa.js but in filepond fa_ir.js is correct

Zagna:
Simple switch which checks for both 'fa' and 'fa-ir' and sets $locale to 'fa_ir'?
Same for norwegian and swedish.

themahdavi:
yes switch need fix for support all locale

PR #2027, opened 2022-03-26
fix issue #1985 - FileUpload locale not work correctly

themahdavi:
this change for fix issue #1985 .

my solution:
1- define array of available locale filename from filepond
2- search array for find site locale
3- set the correct name to url

### DIFF
--- a/packages/forms/resources/views/components/file-upload.blade.php
+++ b/packages/forms/resources/views/components/file-upload.blade.php
@@ -6,19 +6,22 @@
             $defaultLocaleData = ($placeholder = $getPlaceholder()) ? "{ labelIdle: '{$placeholder}' }" : '{}';
 
             $allAvailableLocale = ['ar-ar', 'cs-cz', 'da-dk', 'de-de', 'en-en', 'es-es', 'fa_ir', 'fr-fr', 'he-he', 'hu-hu', 'id-id', 'it-it', 'ja-ja', 'lt-lt', 'nl-nl', 'no-nb', 'pl-pl', 'pt-br', 'ru-ru', 'sk-sk', 'sv_se', 'tr-tr', 'uk-ua', 'zh-cn', 'zh-tw'];
-
+            
+            $isContainLocale = false;
             foreach ($allAvailableLocale as $availableLocale) {
                 if (str_contains($availableLocale, $locale)) {
                     $locale = $availableLocale;
+                    $isContainLocale = true;
                 }
             }
         @endphp
+        @if($isContainLocale)
+            <script type="module">
+                import localeData from 'https://cdn.skypack.dev/filepond/locale/{{$locale}}.js'
 
-        <script type="module">
-            import localeData from 'https://cdn.skypack.dev/filepond/locale/{{$locale}}.js'
-
-            window.FilePond && window.FilePond.setOptions({ ...localeData, ...{!! $defaultLocaleData !!} })
-        </script>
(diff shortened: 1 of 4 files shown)


pxlrbt (changes requested):
Check the styling

themahdavi:
fix formatting on second commit (wip)

packages/forms/resources/views/components/file-upload.blade.php
```diff
@@ -5,8 +5,12 @@
 
             $defaultLocaleData = ($placeholder = $getPlaceholder()) ? "{ labelIdle: '{$placeholder}' }" : '{}';
 
-            if (! str_contains($locale, '-')) {
-                $locale .= '-' . $locale;
+            $allAvailableLocale = ['ar-ar', 'cs-cz', 'da-dk', 'de-de', 'en-en', 'es-es', 'fa_ir', 'fr-fr', 'he-he', 'hu-hu', 'id-id', 'it-it', 'ja-ja', 'lt-lt', 'nl-nl', 'no-nb', 'pl-pl', 'pt-br', 'ru-ru', 'sk-sk', 'sv_se', 'tr-tr', 'uk-ua', 'zh-cn', 'zh-tw'];
+
+            foreach ($allAvailableLocale as $availableLocale){
+                if(str_contains($availableLocale, $locale)){
```
pxlrbt, line 11:
```suggestion
            foreach ($allAvailableLocale as $availableLocale) {
                if (str_contains($availableLocale, $locale)) {
```

pxlrbt (changes requested):

pxlrbt:
> fix formatting on second commit (wip)
> 
> meaning about styling ?

Sorry, somehow it didn't show the requested changes 🤷🏼‍♂️ Tried it again.

Also I think we don't need to try loading the locale, if it didn't match (Lines 18–20)
```html
<script type="module">
    import localeData from 'https://cdn.skypack.dev/filepond/locale/{{$locale}}.js'
    window.FilePond && window.FilePond.setOptions({ ...localeData, ...{!! $defaultLocaleData !!} })
</script>
        ```

themahdavi:
this pr just for load all available locale from filepond ... 
if not load and get error from get request haven't any idea
can use fallback_locale from laravel app.php config

pxlrbt:
> this pr just for load all available locale from filepond ...
> if not load and get error from get request haven't any idea
> can use fallback_locale from laravel app.php config

I don't think we need to check the request. Just whether the locale is in `allAvailableLocales`. If it's not there we don't need to import it.

themahdavi:
yep the default locale is EN and if the locale not contain in allAvailableLocale can ignore from send request.

commit now

danharrin:
Tbh, this is more of a Filepond issue than Filament. Only 3 Filepond locales use `_` instead of `-`. Surely, we should encourage that project to fix the inconsistency, instead of introducing this extra code to Filament?

danharrin:
Instead, maybe we should add a new translation key that allows you to define a Filepond locale string for that locale. Then we dont need that custom code, we can just read the locale from `__('forms::components.file_upload.filepond_locale')`

pxlrbt:
> Instead, maybe we should add a new translation key that allows you to define a Filepond locale string for that locale. Then we dont need that custom code, we can just read the locale from `__('forms::components.file_upload.filepond_locale')`

I guess that would be the best solution 😅

PR #2032, opened 2022-03-27
Fix error in global search when user can not view or edit record.

### LINKED ISSUE #2031: getGlobalSearchResultUrl() returns null, but GlobalSearchResult does not accept anything but a string
### Package

filament/filament

### Package Version

2.10.39

### Laravel Version

9.1.2

### Livewire Version

_No response_

### PHP Version

8.1.3

### Bug description

when using the global search function and typing in a search which results in a match I get:

`Filament\GlobalSearch\GlobalSearchResult::__construct(): Argument #2 ($url) must be of type string, null given, called in /Users/heno/Documents/filamentpackagetest/vendor/filament/filament/src/Resources/Resource.php on line 204`

### Steps to reproduce

This does only happen if I use the global search with a user having restricted policy access to resources. 

### Relevant log output

_No response_

pxlrbt:
@ryangjchandler Can you give feedback on this one?

Ideally data is filtered by `getEloquentQuery()` before, but you cannot rely on that entirely I think.

Fixes https://github.com/laravel-filament/filament/issues/2031

### DIFF
--- a/packages/admin/src/Resources/Resource.php
+++ b/packages/admin/src/Resources/Resource.php
@@ -198,11 +198,20 @@ class Resource
         return $query
             ->limit(50)
             ->get()
-            ->map(fn (Model $record): GlobalSearchResult => new GlobalSearchResult(
-                title: static::getGlobalSearchResultTitle($record),
-                url: static::getGlobalSearchResultUrl($record),
-                details: static::getGlobalSearchResultDetails($record),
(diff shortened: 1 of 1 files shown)


pxlrbt:
@ryangjchandler Can you give feedback on this one?

danharrin:
Thank you

PR #2033, opened 2022-03-28
Tweak forms validation attribute generation

austincarpenter:
This commit tweaks the behaviour of the `getValidationAttribute` method for fields, to use the entire lowercased version of the field label (and therefore name, if not explicitly set) as a fallback, rather than simply lowercasing the first character.

For a label consisting of a single word, this wouldn't normally be a problem, however for labels with multiple words, aesthetically it looks better to have all words lowercased when used for error messages.

Before:
Screen Shot 2022-03-28 at 11 38 01 am

After:
Screen Shot 2022-03-28 at 11 38 13 am

### DIFF
--- a/packages/forms/src/Components/Concerns/CanBeValidated.php
+++ b/packages/forms/src/Components/Concerns/CanBeValidated.php
@@ -170,7 +170,7 @@ trait CanBeValidated
 
     public function getValidationAttribute(): string
     {
-        return $this->evaluate($this->validationAttribute) ?? lcfirst($this->getLabel());
+        return $this->evaluate($this->validationAttribute) ?? strtolower($this->getLabel());
     }
 
     public function getValidationRules(): array


danharrin:
Hey, thank you very much for the contribution.

I agree  - if you're capitalising the first letter of each word in the label, then your method works perfectly - "contract name" looks far better than "contract Name".

However, Filament's default label generator does not capitalise first letters of each word. There are cases where you will pass a custom label with words that deliberately need capitalisation for the validation attribute AND label - for example with abbreviations:

```php
TextInput::make('contract_id')
    ->label('Contract ID') // validation attribute is now `contract ID` and not `contract id`
```

So now, hopefully you can see why this would be a breaking change for some people.

I would recommend not capitalising the each word in the label, as this gets rid of the issue for you and (IMO) also looks a bit neater.

PR #2034, opened 2022-03-28
Add columnSpan for table filters

ttungbmt:
XYZ
ABC


danharrin:
I'm implementing this in a new PR. You've added a ton of other changes here that are not related to the original PR.

PR #2042, opened 2022-03-29
Docs: Fix widget polling examples

intrepidws:


danharrin:
Thanks, the old way did work though, right? Because `string` is a subset of `?string`

intrepidws:
Not for me, no. I got the following error on PHP 8.1 and Filament v2.10.39:

>Type of App\Filament\Widgets\StatsOverview::$pollingInterval must be ?string (as in class Filament\Widgets\StatsOverviewWidget)

PR #2043, opened 2022-03-29
Fix FileUpload field overlaps buttons

nikita2608:
If the file upload field is set to multiple, buttons overlaps and becomes unclickable. 
The problem is solved by adding "overflow: hidden;" for ".filepond--root" class


danharrin:
Thanks

PR #2044, opened 2022-03-29
Update TextColumn.php

pschilly:
Addition of ->html() modifier. Purpose of this is to display the returned text from the database with HTML intact (rich text) rather than outputting raw data.

Sample usage in a Resource:

```
    {
        return $table
            ->columns([
                Tables\Columns\TextColumn::make('name'),
                Tables\Columns\TextColumn::make('description')->html()
            ])
            ->filters([
                //
            ]);
    }
```


danharrin:
Thank you! And well done on the first PR ;)

PR #2054, opened 2022-03-30
Fixes overflow issue when block picker is at bottom of screen.

awcodes:
Adds support to Builder Block Picker to determine if it should drop down or up. This prevents overflowing at bottom of screen when open.

Before:
filament-builder-block-overflow

After:
block-picker-drop-up


danharrin:
I actually want to move to Popper.js for all popups in the entire form package, inc for our Select, MultiSelect and DatePicker components. How would you feel about integrating that instead?

We currently use Alpine to calculate available window space, but that's not a reliable solution and can be replaced.

danharrin:
I understand that this issue is only affecting you for this component, so if you want just integrate Floating UI for the Builder, and then I can handle the rest if you don't feel up to it?

awcodes:
Would you want to implement floating ui as an alpine plugin?

danharrin:
If you can find a nice abstraction, yes!

awcodes:
I’ll see what I can do. Next couple of days are pretty busy at work. Should have some time this weekend. 

Thanks for your feedback.

danharrin:
Thank you!

PR #2055, opened 2022-03-30
Laravel Octane Support: Scoped service binding

coleshirley:
See https://github.com/laravel-filament/filament/issues/2010

@danharrin As I was working on this I found there is an issue with octane itself where scoped services aren't being flushed correctly because of a discrepancy between octane and laravel itself. https://github.com/laravel/octane/issues/500

There is a workaround where after this pull request is submitted you can add the filament service to octane's flush services in the config like so:

config/octane.php:
```php
<?php

...

return [

   ...

    /*
    |--------------------------------------------------------------------------
    | Warm / Flush Bindings
    |--------------------------------------------------------------------------
    |
    | The bindings listed below will either be pre-warmed when a worker boots
    | or they will be flushed before every new request. Flushing a binding
    | will force the container to resolve that binding again when asked.
    |
    */

    'warm' => [
        ...Octane::defaultServicesToWarm(),
    ],

    'flush' => [
        'filament'
    ],

  ...

];

```

Is that something that should be added to the documentation or do you want to just refer people here? Once the issue in the octane package is resolved we won't need to add that item to the flush in the config

### DIFF
--- a/packages/admin/src/FilamentServiceProvider.php
+++ b/packages/admin/src/FilamentServiceProvider.php
@@ -72,7 +72,13 @@ class FilamentServiceProvider extends PackageServiceProvider
 
     public function packageRegistered(): void
     {
-        $this->app->singleton('filament', function (): FilamentManager {
+        $this->app->resolving('filament', function () {
+            $this->discoverPages();
+            $this->discoverResources();
+            $this->discoverWidgets();
+        });
+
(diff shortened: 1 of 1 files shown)


danharrin:
Hey, let's refer people here. I don't want to add documentation that we then have to remove later when Laravel fixes the issue.

Thanks for the PR!

PR #2062, opened 2022-03-31
TextInput field tel() method to have basic phone number validation

wychoong:
as suggested on dicord https://discord.com/channels/883083792112300104/883085267383226478/956839096075051008

currently `tel()` does no validation
now it will be using a basic check `/^[+]*[(]{0,1}[0-9]{1,4}[)]{0,1}[-\s\.\/0-9]*$/`

if a more specific regex for a country, can override with `regex()`
```php
Forms\Components\TextInput::make('mobile_phone')
             ->tel()
            ->regex('/\(?\d{3}\)?-? *\d{3}-? *-?\d{4}/')
```

### DIFF
--- a/packages/forms/src/Components/Concerns/CanBeValidated.php
+++ b/packages/forms/src/Components/Concerns/CanBeValidated.php
@@ -12,6 +12,8 @@ trait CanBeValidated
 {
     protected bool | Closure $isRequired = false;
 
+    protected string | Closure | null $regexPattern = null;
+
     protected array $rules = [];
 
     protected string | Closure | null $validationAttribute = null;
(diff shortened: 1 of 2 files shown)


ryangjchandler:
The `regex()` method makes sense, but adding a pattern for the `tel()` method by default is too opinionated in my opinion. Happy to accept this with just the `regex()` method.

danharrin:
I'd say it was too opinionated too, but the default regex is very loose and just rejects alphabetical and special characters really. So I think it's fine?

ryangjchandler:
Do `type="tel"` inputs not allow `+` characters for country code?

danharrin:
Yeah, which is part of the regex pattern. Along with ()

ryangjchandler:
Yeah, let's go for it then. I've just tested on a few phone numbers and it seems to work quite well.

PR #2064, opened 2022-03-31
increase shadow on select, rotate icon while open

blite:
Before
image

After:

image

Not sure about how the app.css and app.js are generated as I used  ``` npm run watch``` while doing these changes.

The icon change puts it in line with the following libraries:
https://vue-select.org/
https://select2.org/dropdown
and many other libraries.

We could consider animating it, but I didn't know if other parts of the UI have much animation currently.


danharrin:
You might wanna make the same change/s to the multiselect & date picker :)

blite:
@danharrin currently looking at how to get tag input same icon, but I'll look at those too if time permits

blite:
I added this same change to multi select.

on datepicker

before:
image

after:

image

I checked out 
https://flatpickr.js.org/examples/
they don't use an icon to retract but they do have increased shadow so I added that just like other elements.

PR #2069, opened 2022-03-31
set TagsInput webkit-calendar-picker icon to 0 opacity

blite:
removes webkit-calendar-picker icon on tagsinput

before:
161101068-d8f95234-aea9-47d0-9523-3fc5471a6d53

after:
image

I did not like that I had to add another css file for this.  Also don't love all the auto generated files in this commit.


blite:
This is per discussion at https://github.com/laravel-filament/filament/discussions/2067

Maybe we should add back in a SVG icon, but for now just having that weird hard to debug icon out is IMO a good thing.  

I sort of agree on the datalist thing which is why I created an issue before doing that.  My only hesistation on the datalist is other unknown bugs, but we could get that with either solution.

packages/forms/resources/views/components/tags-input.blade.php
```diff
@@ -42,7 +42,7 @@
                         x-on:blur="createTag()"
                         x-model="newTag"
                         {{ $getExtraInputAttributeBag()->class([
-                            'block w-full border-0',
+                            'tags-input block w-full border-0',
```
danharrin, line 45: Lets stick with utility classes and rename this `webkit-calendar-picker-indicator-opacity-0` maybe?

danharrin, line 45: Lets stick with utility classes and rename this `webkit-calendar-picker-indicator-opacity-0` maybe?

PR #2084, opened 2022-04-04
View Page Title Improvement

intrepidws:
This PR makes the formatting of the title for the view page consistent with the formatting of the title for the create and edit pages. For the examples below, let's assume we have a Registrant Record with a title of Craig Anderson. Currently, the page titles for these three pages would be as follows:

- _Create Page:_ Create Registrant
- _Edit Page:_ Edit Craig Anderson
- _View Page:_ Craig Anderson

I found this to be confusing for the user experience, especially in situations when some resources have a View page only, some have an Edit page only, and some have both.

This PR changes the View page to be consistent with the formatting of the other two, as follows:

- _Create Page:_ Create Registrant
- _Edit Page:_ Edit Craig Anderson
- _View Page:_ View Craig Anderson

### DIFF
--- a/packages/admin/resources/lang/ar/resources/pages/view-record.php
+++ b/packages/admin/resources/lang/ar/resources/pages/view-record.php
@@ -2,6 +2,8 @@
 
 return [
 
+    'title' => 'عرض :label',
+
     'breadcrumb' => 'عرض',
 
     'actions' => [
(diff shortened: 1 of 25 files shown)


danharrin:
Wow... you went into the effort to do this for every language? I have to applaud you. Thank you so much.

PR #2088, opened 2022-04-05
Improved search variable name for selects

intrepidws:
Per [this discussion](https://github.com/laravel-filament/filament/discussions/2007), this PR improves the default search string variable for the Form Select Components. Previously, the variable had to be `$query`, which could sometimes be confusing when the Laravel Query builder was also used. This PR allows either `$search` or `$query` to be used, and updates the docs to default to `$search`.

I also changed the variable name in a few other related methods for the sake of consistency.

### DIFF
--- a/packages/admin/src/Resources/RelationManagers/Concerns/CanAssociateRecords.php
+++ b/packages/admin/src/Resources/RelationManagers/Concerns/CanAssociateRecords.php
@@ -61,7 +61,7 @@ trait CanAssociateRecords
         return Select::make('recordId')
             ->label(__('filament::resources/relation-managers/associate.action.modal.fields.record_id.label'))
             ->searchable()
-            ->getSearchResultsUsing(function (Select $component, RelationManager $livewire, string $query): array {
+            ->getSearchResultsUsing(function (Select $component, RelationManager $livewire, string $search): array {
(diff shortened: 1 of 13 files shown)


danharrin:
I know I said that I like `$search` the other day, I just checked in the codebase for other instances to make sure we don't miss any, and I saw that we use `$searchQuery` in quite a few places.

Problem is, if we changed `$searchQuery` to `$search` in those places, some of those changes would break existing apps (as methods could be overridden, like `getGlobalSearchResults(string $searchQuery)` on the resource to implement Laravel Scout).

So we either make it `$search` in some places and `$searchQuery` in others, or we use `$searchQuery` everywhere. Thoughts? I'm not expecting you to make this change, you've done loads of work already and I told you specifically to use `$search` the other day.

ryangjchandler:
@danharrin I say we use `$searchQuery` and put it on our list of breaking changes to make in the future.

intrepidws:
That certainly makes sense. I have updated the PR to reflect `$searchQuery` instead. I do think it makes sense as a future breaking change though, since the variable name implies it's the actual query of the search, rather than the text being searched for.

danharrin:
Thanks :)

PR #2089, opened 2022-04-05
Remove Unused Method Arguments

intrepidws:
This PR removes a couple unused method arguments that I noticed while writing a different PR.


danharrin:
Cheers

PR #2090, opened 2022-04-05
feature: allow disabling collapsible nav groups

ryangjchandler:
`->group()` now accepts a `$collapsible` argument that defaults to `true`. Default needs to be `true` to avoid a breaking UI/UX change.

If you set it to `false`, then the group won't be collapsible.

### DIFF
--- a/packages/admin/resources/views/components/layouts/app/sidebar/group.blade.php
+++ b/packages/admin/resources/views/components/layouts/app/sidebar/group.blade.php
@@ -1,11 +1,14 @@
 @props([
     'label' => null,
+    'collapsible' => true,
 ])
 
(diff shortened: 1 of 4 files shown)


danharrin:
i dont like it

intrepidws:
It would be great if there was a way to do this for people that aren't registering their own custom navigation items. Maybe something similar to how `Filament::registerNavigationGroups()` lets you enforce a  [custom navigation group order](https://filamentphp.com/docs/2.x/admin/navigation#ordering-navigation-groups).

ryangjchandler:
Yeah, although I'd generally encourage people to use the navigation builder. There's a lot more freedom with it.

PR #2096, opened 2022-04-06
fix: use `resolveRouteBinding` to resolve model

innocenzi:
This PR reintroduces the fix from https://github.com/laravel-filament/filament/pull/245, that was apparently lost. My use case is the same, my model uses hashid by extending `resolveRouteBinding`.

### DIFF
--- a/packages/admin/src/Resources/Resource.php
+++ b/packages/admin/src/Resources/Resource.php
@@ -86,9 +86,8 @@ class Resource
 
     public static function resolveRecordRouteBinding($key): ?Model
     {
-        return static::getEloquentQuery()
-            ->where(static::getRecordRouteKeyName(), $key)
-            ->first();
+        return (new (static::getModel()))
+            ->resolveRouteBinding($key);
     }
 
     public static function can(string $action, ?Model $record = null): bool
--- a/packages/admin/src/Resources/Resource.php
+++ b/packages/admin/src/Resources/Resource.php
@@ -86,7 +86,13 @@ class Resource
 
     public static function resolveRecordRouteBinding($key): ?Model
     {
-        return (new (static::getModel()))
+        if (static::shouldUseRecordRouteKeyName()) {
+            return static::getEloquentQuery()
+                ->where(static::getRecordRouteKeyName(), $key)
+                ->first();
+        }
+
+        return app(static::getModel())
             ->resolveRouteBinding($key);
     }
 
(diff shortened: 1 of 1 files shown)


ryangjchandler:
Isn't this a breaking change for people who have already defined `$recordRouteKeyName` on their resource?

innocenzi:
@ryangjchandler Good catch, thanks! I changed the PR so it keeps using `$recordRouteKeyName` just like before if it was defined. Does that look right?

danharrin:
Thank you very much

danharrin:
Hey, I've just had to revert this change, apologies. Issue #2177.

I now remember why we changed this in v2... it's because we now use the `getEloquentQuery()` method on the resource as the base query for the route binding. This allows people to apply a custom scope (https://filamentphp.com/docs/2.x/admin/resources#disabling-global-scopes) just for the admin panel, without affecting their main app. For example, removing a tenancy scope so that admins can access all records.

If you'd like to still use `resolveRouteBinding()`, you can override the method on the resource yourself, and even extract it to a trait to use across multiple resources:

```php
public static function resolveRecordRouteBinding($key): ?Model
{
    return app(static::getModel())->resolveRouteBinding($key);
}
```

innocenzi:
I didn't look into it yet, but is there not any way to support both? It'd not be ideal to add a trait to all pages using hashids, but maybe a property in the resource would do?

Or maybe we can detected if the `getEloquentQuery` method has been overwritten and put that in the `shouldUseRecordRouteKeyName` heuristic?

danharrin:
You wouldn't need to add a trait to all pages, just resources (or create a base resource class like HashidRoutedResource)

```php
public static function resolveRecordRouteBinding($key): ?Model
{
    return parent::resolveRecordRouteBinding(Hashids::decode($key)[0] ?? $key);
}
```

innocenzi:
I meant resources yes, and I'll do that for now, it's not that big of a deal. Do you think you'd accept a PR for a `$useRecordRouteBinding` property on the resource to force using it though? I'd understand if not, but I think it'd be nice to have.

danharrin:
The alternative is overriding `resolveRouteBindingQuery()` instead of `resolveRouteBinding()`, which accepts an existing query instance. Then we can use `resolveRouteBindingQuery()` inside the resource's `resolveRecordRouteBinding()` method?

danharrin:
I've implemented this in #2223, you can now override `resolveRouteBindingQuery()` instead of `resolveRouteBinding()`, both do the same thing!

innocenzi:
Thanks a lot Dan!

PR #2116, opened 2022-04-09
🎨 Color picker

saade:
This PR adds a new ColorPicker input using [web-padawan/vanilla-colorful](https://github.com/web-padawan/vanilla-colorful).

Chose this because i've used before and:
- 🗜 Small: Just 2,7 KB (minified and gzipped).
- 🚀 Fast: Built with standards based Custom Elements.
- 💬 Accessible: Follows the [WAI-ARIA](https://www.w3.org/WAI/standards-guidelines/aria/) guidelines to support users of assistive technologies.
- 📲 Mobile-friendly: Works well on mobile devices and touch screens.
- 💨 No dependencies

It's a vanilla web component (no html-in-js bullshit), so all its functionality can be managed via Alpine.js (show, hide, state, style ...).

Originally i've made this as a external plugin (saade/filament-colorful), but its so lightweight that i've included it in the core since filament doesn't have one. (I still can release as a plugin if you don't want it to ship with filament)

#### Popup with preview
```php
Forms\Components\ColorPicker::make('color')
```
image

#### Popup without preview
```php
Forms\Components\ColorPicker::make('color')
    ->preview(false)
```
image

#### Inline
```php
Forms\Components\ColorPicker::make('color')
    ->inline(true)
```
image

#### Color modes
```php
Forms\Components\ColorPicker::make('color')
    ->format('rgba') // hex, rgb, rgba, hsl
```
or using method shortcuts
```php
Forms\Components\ColorPicker::make('color')
    ->rgba() // ->hex(), ->hsl(), ->rgb(), ->rgba()
```
image


packages/forms/src/Components/ColorPicker.php
```diff
@@ -0,0 +1,51 @@
...
+    protected bool | Closure $preview = true;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->inline(false);
+    }
+
+    public function format(string | Closure $format): static
```
ryangjchandler, line 28: Could we get some extra shortcut methods for doing `->format('hex')`, etc?

ryangjchandler (changes requested):

ryangjchandler:
Looking really good @saade, I think this is a worthwhile addition to Filament.

packages/forms/src/Components/ColorPicker.php
saade: Sure!

saade:
> Looking really good @saade, I think this is a worthwhile addition to Filament.
> 
> 
> 
> Does this work correctly with default field values?

Yes, setting a default value, both input and picker get the correct default value

ryangjchandler:
Awesome, nice work!

packages/forms/src/Components/ColorPicker.php
saade: Done, added 
`->hex()`
`->hsl()`
`->rgb()`
`->rgba()`

ryangjchandler (approved):

invaders-xx:
Nice work! Is this plugin also provides a button to clear the input?

saade:
> @saade Nice work! Does this plugin also provide a button to clear the input?

It does not, the component itself doesn't have a input either, only the picker.

I have a wip here locally where i've implemented this clearable behavior, gonna PR it later tonight

danharrin:
Thanks!

I added docs, and removed the preview & inline options. I didn't feel like the naming of `inline()` was consistent with other inline fields.

PR #2128, opened 2022-04-11
feature: introduce new `ReplicatorAction`

ryangjchandler:
This pull request introduces a new first-party `ReplicatorAction`. I originally built this as a package for a project but figured it might be worth adding to core.

Here's the API:

```php
ReplicatorAction::make('replicate')
	// The `exclude()` method is used to tell the action which columns to be exclude when replicating.
	->exclude(['slug'])
	// `beforeSaving` can be used to provide a callback that will be invoked before saving the replica.
	->beforeSaving(function (Model $replica) {
		
	})
	// `afterSaving` can be used to provide a callback that will be invoked after saving the replica.
	->afterSaving(function (Model $replica) {

	});
```

By default, the action will appear as like a `LinkAction` but calling the `->button()` method will make it look like a button instead.

```php
ReplicatorAction::make('replicate')
	->button()
```

### Getting user input

Just like normal actions, you can provide a form schema that can be used to modify the replica before saving.

```php
ReplicatorAction::make('replicate')
	->exclude(['title'])
	->form([
		TextInput::make('title')
			->required(),
	])
	->beforeSaving(function (Model $replica, array $data) {
		$replica->fill($data);
	})
```

### Future scope

In the future, this could be expanded to support replicating relationships as well as the model.


pxlrbt:
Really useful. Thanks.

packages/tables/src/Actions/ReplicatorAction.php
```diff
@@ -0,0 +1,102 @@
...
+use Filament\Facades\Filament;
+use Illuminate\Database\Eloquent\Model;
+
+class ReplicatorAction extends ButtonAction
+{
+    protected bool $isButton = false;
+
+    protected array | Closure | null $exclude = null;
+
+    protected ?Closure $beforeSavingCallback = null;
```
danharrin, line 15:
```suggestion
    protected ?Closure $beforeReplicaSaved = null;
```

packages/tables/src/Actions/ReplicatorAction.php
```diff
@@ -0,0 +1,102 @@
...
+
+    protected ?Closure $afterSavingCallback = null;
```
danharrin, line 17:
```suggestion
    protected ?Closure $afterReplicaSaved = null;
```

packages/tables/src/Actions/ReplicatorAction.php
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public function beforeSaving(Closure $callback): static
+    {
+        $this->beforeSavingCallback = $callback;
+
+        return $this;
+    }
+
+    public function callBeforeSaving(Model $replica, array $data = []): void
```
danharrin, line 59:
```suggestion
    public function callBeforeReplicaSaved(Model $replica, array $data = []): void
```

packages/tables/src/Actions/ReplicatorAction.php
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public function afterSaving(Closure $callback): static
+    {
+        $this->afterSavingCallback = $callback;
+
+        return $this;
+    }
+
+    public function callAfterSaving(Model $replica, array $data = []): mixed
```
danharrin, line 74:
```suggestion
    public function callAfterReplicaSaved(Model $replica, array $data = []): mixed
```

packages/tables/src/Actions/ReplicatorAction.php
```diff
@@ -0,0 +1,102 @@
...
+
+        return $this;
+    }
+
+    public function getExcludedAttributes(): ?array
+    {
+        return $this->evaluate($this->exclude);
+    }
+
+    public function getView(): string
```
danharrin, line 94: See first comment

danharrin (changes requested):

packages/tables/src/Actions/ReplicatorAction.php
```diff
@@ -0,0 +1,102 @@
...
+
+namespace Filament\Tables\Actions;
+
+use Closure;
+use Filament\Facades\Filament;
+use Illuminate\Database\Eloquent\Model;
+
+class ReplicatorAction extends ButtonAction
+{
+    protected bool $isButton = false;
```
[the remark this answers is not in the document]
ryangjchandler: Yeah, we spoke about this. Will add.

danharrin:
Also, we'll need a short paragraph about this in the Actions docs I think

danharrin:
Any reason why the `type()` methods / property can't be on the base Action class (`Concerns\HasView`) instead?

ryangjchandler:
@danharrin That type method is gone now anyway. It wasn't needed since the methods are instead changing the `$view` directly.

danharrin:
Alright, just docs & translation strings to go then :) Lets set the default `label()` in `setUp()`.

PR #2130, opened 2022-04-12
Merge form rules with Component rules

tanthammar:
Sometimes you want to add custom rules to your Livewire component, then you want them merged with the Filament form rules.

This PR follows the Livewire native code style to get the rules from the component and then merges them with the Filament rules.

### Example:

```php
protected function rules(): array
{
    return [
        'terms' => 'accepted'
    ];
}

//or

protected array $rules = [ 'terms' => 'accepted' ];


{
    return [
        //Filament form fields with rules
    ];
}
```


ryangjchandler:
@tanthammar Can this not be done by merging the parent method with your own rules already?

tanthammar:
@ryangjchandler Maybe it is just me, but I think that the form trait should behave like native Livewire without having to add a parent::override method

ryangjchandler:
This PR doesn't do that though, Livewire uses `getRules`, not `rules()`.

tanthammar:
Livewire uses getRules() and the Filament Form trait overrides that method.
The two rows I added to Filament form trait getRules() method are the rows missing from the method it overrides.

tanthammar:
ValidatesInput

```
    protected function getRules()
    {
        if (method_exists($this, 'rules')) return $this->rules();
        if (property_exists($this, 'rules')) return $this->rules;

        return [];
    }
```

Or do I misunderstand you?

ryangjchandler:
Sorry, I misunderstood you.

tanthammar:
My point is that I don't want the Filament form trait to remove the default behaviour of the 
Livewire `rules() `/`$rules `. 
My bad, not to give a better example or explanation.

PR #2137, opened 2022-04-12
Feature: Add visible method to Table Filters

Z3d0X:
This adds the missing `visible()` method to` Filament\Tables\Filters\Concerns\CanBeHidden`

I have implemented it same as [Filament\Tables\Actions\Concerns\CanBeHidden](https://github.com/laravel-filament/filament/blob/4d6826132158a59ed880819c291494edee7de024/packages/tables/src/Actions/Concerns/CanBeHidden.php)


danharrin:
Thanks

PR #2155, opened 2022-04-15
Fix: Add missing disabled classes to toggle.blade.php

robertdrakedennis:
Currently when a toggle is disabled, it still shows the `cursor-pointer` class. I've added a fix for that specifically and also included a lower opacity to match the `TextInput` disabled classes.


danharrin:
Thanks

PR #2167, opened 2022-04-17
Add Spanish translation for associate and dissociate

pathros:
I updated the new translation strings for Spanish lang


danharrin:
Thanks

PR #2168, opened 2022-04-17
Return null when image not found on FileUpload component

nurmuhammet-ali:
When FileUpload component does not find the file, it throws an exception. Sometimes files can be deleted for some reasons like updating the server, etc. Very short screen sharing video expressing what I mean: [youtube link](https://www.youtube.com/watch?v=BdwH_XhQjmw)

Open to any ideas.


packages/forms/src/Components/BaseFileUpload.php
```diff
@@ -99,6 +101,13 @@ protected function setUp(): void
             // An ugly mess as we need to support both Flysystem v1 and v3.
             $storageAdapter = method_exists($storage, 'getAdapter') ? $storage->getAdapter() : (method_exists($storageDriver = $storage->getDriver(), 'getAdapter') ? $storageDriver->getAdapter() : null);
             $supportsTemporaryUrls = method_exists($storageAdapter, 'temporaryUrl') || method_exists($storageAdapter, 'getTemporaryUrl');
+            
+            try {
```
danharrin, line 105: Instead of catching the exception & logging, let's just use an `exists()` check for the file to make sure it exits before we make the `getVisibility()` check.

packages/forms/src/Components/BaseFileUpload.php
```diff
@@ -99,6 +101,13 @@ protected function setUp(): void
...
+                $storage->getVisibility($file);
+            } catch (UnableToRetrieveMetadata $e) {
+                Log::error($e);
+                return '';
+            }
 
             if ($storage->getVisibility($file) === 'private' && $supportsTemporaryUrls) {
```
danharrin, line 112: See above comment. Does this work instead? I haven't tested it.

```suggestion
            if ($storage->exists($file) && $storage->getVisibility($file) === 'private' && $supportsTemporaryUrls) {
```

nurmuhammet-ali:
@danharrin, of course you're right, it's better to check for the existence of a file than to catch an exception. Made a commit. Please check it.

danharrin:
Thanks!

PR #2169, opened 2022-04-18
Extract user command methods

dododedodonl:
Extracts logic to methods for easier overriding of the defaults.


danharrin:
Instead of passing in `$userModel` to these methods, lets define `protected static function getUserModel()`, `getUserProvider()`, `getAuthGuard()` methods instead :)

dododedodonl:
Like this?

danharrin:
Thanks!

PR #2170, opened 2022-04-18
Feature: Add view option to MakeResourceCommand

Z3d0X:
This PR adds a `--view` option to `MakeResourceCommand` to make resource with ViewPage
```bash
php artisan make:filament-resource User --view
```

PS: Since pages are conditionally added `SimpleResource` stub is no longer needed.


danharrin:
Simple resources also shouldn't have a relations method

Z3d0X:
Oh sorry I've missed that. How would you like to see it fixed? 
Bring back `SimpleResource` or conditionally include relations

danharrin:
I'm thinking about moving our stubs system to use `Blade::render()` instead, then we can use Blade to construct templates instead of messy PHP strings?

https://laravel.com/docs/9.x/blade#rendering-inline-blade-templates

Z3d0X:
@danharrin isn't inline blade templates (`Blade::render`) a laravel 9 feature?

danharrin:
Ah thats ok mate, annoying though. Thanks for your work!

wychoong:
it was introduced in Laravel v8.80 but  facade Blade::render is in 8.81, probably not worth to version bump

PR #2174, opened 2022-04-19
feature: support statically setting nav options

ryangjchandler:
This will allow developers to call `setNavigationX` methods on `Resource` classes without needing to extend third-party resources.

I'm doing this in a plugin at the moment but figured it would make sense to add to core as an extra point of customisation / extension for third-party plugins.

### DIFF
--- a/packages/admin/src/Resources/Resource.php
+++ b/packages/admin/src/Resources/Resource.php
@@ -369,9 +369,19 @@ class Resource
         return static::$navigationGroup;
     }
 
+    public static function setNavigationGroup(string $group): void
+    {
+        static::$navigationGroup = $group;
+    }
+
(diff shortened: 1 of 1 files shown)


packages/admin/src/Resources/Resource.php
```diff
@@ -369,11 +369,21 @@ protected static function getNavigationGroup(): ?string
         return static::$navigationGroup;
     }
 
+    public static function setNavigationGroup(string $group): void
```
danharrin, line 372: Maybe just `navigationGroup()` instead? Like other setter methods?

ryangjchandler: Yeah, good shout.

PR #2057, opened 2022-03-30
feature: Refactor select inputs to use Choices.js, add action support to form components, including `BelongsToSelect` create action / modal

ryangjchandler:

PR #2178, opened 2022-04-19
Add mousetrap library and save and create shortcuts

buzkall:
Attempt to bring back the v1 shortcuts.
Can't use the same click action than in [PR 264](https://github.com/laravel-filament/filament/pull/264/) because the logout submit is triggered.

I've also added a shortcut for the create action. Can't bind Cmd+N, so I suggest Cmd+O


danharrin:
Let's pass an `id="form"` to the `<form>` and using `.submit()` instead of targetting the button?

danharrin:
Let's pass an `id="form"` to the `<form>` and using `.submit()` instead of targetting the button?

danharrin:
Also we should install the NPM package and add it to our JS bundle instead of using the CDN

ryangjchandler:
Hey @buzkall, have you committed all changes? I can't see any changes made to the `app.js` source file, only the file in `dist`.

ryangjchandler:
Another thought, should we disable these by default and add an option to the config for the package to enable them and perhaps customise the shortcuts?

buzkall:
Hi @ryangjchandler, I've commited my changes now following your suggestion of adding it to the filament config file

danharrin:
Just had an even better idea (and sorry for wasting your time with previous suggestions), we could add this to the admin panel Action object.

`Action::make('save')->keyBindings(['ctrl+s', 'command+s'])`

saade:
> Just had an even better idea (and sorry for wasting your time with previous suggestions), we could add this to the admin panel Action object.
> 
> `Action::make('save')->keyBindings(['ctrl+s', 'command+s'])`

That's nice, that way we can open modals using keybinding 😍

buzkall:
Hi @danharrin, I think that's a great idea. I've created a HasKeyBindings trait (not sure about the name) and moved the Alpine code to the button.blade.php file.
I'm not sure about my solution when the button is a link. What do you think?

packages/admin/config/filament.php
```diff
@@ -211,6 +211,25 @@
...
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | Shortcuts
+    |--------------------------------------------------------------------------
+    |
+    | Using Mousetrap, you can bind keyboard shortcuts to your admin panel.
+    | Ctrl|Cmd + S will save the create or edit form
+    | Ctrl|Cmd + 0 will open the create form from the list page
```
zepfietje, line 221: @buzkall shouldn't this `0` be `o`?
```suggestion
    | Ctrl|Cmd + O will open the create form from the list page
```

buzkall: My bad. Fixed now

danharrin:
I've had to completely refactor the actions implementation in #2057, so I'm gonna add this as a todo there. Thanks for your work!

PR #2181, opened 2022-04-20
feature: render hooks

ryangjchandler:
At the moment, extending the admin panel from a third-party package is only really possible with fields/resources/pages. If somebody wanted to add a bit of text or view somewhere, the developer would need to publish a vendor Blade view and most likely modify it.

An example of where we support this already is with the global search `end` and `start` views. 

This PR introduces a new "render hook" concept. These are small hooks that we can add into our own Blade views and allow developers to register their own callbacks from packages, etc.

An example of this might be a context switcher rendered in the header of the admin panel. Registering a render hook will allow developers to render their own Blade views here without worrying about overwriting Blade views from the package.

```php
public function boot()
{
	Filament::registerRenderHook('header::start', function () {
		return view('my-package::header-start');
	});
}
```

**Inside of a Filament view**

```blade
<header>
	{{-- ... --}}

	{{ \Filament\Facades\Filament::renderHook('header::start') }}
</header>
```

Mostly looking for feedback on this, as I have a few potential plugins that would benefit from this.

### DIFF
--- a/packages/admin/src/FilamentManager.php
+++ b/packages/admin/src/FilamentManager.php
@@ -47,6 +47,8 @@ class FilamentManager
 
     protected ?Closure $navigationBuilder = null;
 
+    protected array $renderHooks = [];
+
     public function auth(): Guard
     {
         return auth()->guard(config('filament.auth.guard'));
@@ -109,6 +111,11 @@ class FilamentManager
         $this->pages = array_merge($this->pages, $pages);
     }
 
+    public function registerRenderHook(string $name, Closure $callback): void
+    {
+        $this->renderHooks[$name][] = $callback;
+    }
+
     public function registerResources(array $resources): void
     {
         $this->resources = array_merge($this->resources, $resources);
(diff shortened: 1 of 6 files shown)


SebastiaanKloos:
This is a very welcome feature!

ryangjchandler:
A sensible point of extension might also be sending data through to the render hooks...

packages/admin/src/FilamentManager.php
```diff
@@ -168,6 +175,21 @@ public function getGlobalSearchProvider(): GlobalSearchProvider
...
+            return null;
+        }
+
+        $output = '';
+
+        foreach ($this->renderHooks[$name] as $renderHook) {
+            $output .= (string) app()->call($renderHook);
+        }
+
+        return $output;
```
danharrin, line 190: We should return an `HtmlString()` object instead so you don't need to escape the output in Blade

packages/admin/src/FilamentManager.php
```diff
@@ -168,6 +175,21 @@ public function getGlobalSearchProvider(): GlobalSearchProvider
...
+    public function renderHook(string $name): ?string
+    {
+        if (! array_key_exists($name, $this->renderHooks)) {
+            return null;
+        }
+
+        $output = '';
+
+        foreach ($this->renderHooks[$name] as $renderHook) {
+            $output .= (string) app()->call($renderHook);
```
danharrin, line 187: no improvement here, just TIL that view objects can be cast to strings to be rendered? 🤯

ryangjchandler: They sure can (`__toString()`).

packages/admin/src/FilamentManager.php
ryangjchandler: Yeah, good shout.

danharrin:
Refactored the render function a bit

PR #2218, opened 2022-04-25
Feature/make columns globally configurable

bernhardh:
As mentioned in discord, there is currently no way to configure table columns globally. With this pull request, you can make for example all table columns `sortable` and/or `toggleable` without copy+pasting it on all field definitions. 

If this pull request is ok for you, I could do the same for form fields as well. Usecases: 
- All fields should be required by default. 
- All `DatePicker` fields should have a the central european format `format('d.m.Y')`
- All fields should have a specific custom attribute
- All `Toggle` fields should be inline.


danharrin:
Thanks, I love this. I've extracted it into a new "support" package which we share between packages. Looking forward to adding it to other objects, like form components!

danharrin:
#2229

bernhardh:
Thank you for enhancing/merging and your great work.

PR #2221, opened 2022-04-25
fix: unable to close modal when click on the empty side

wychoong:


danharrin:
We might also need to make the same changes to the tables package?

wychoong:
applied the same for tables modal, but not tested

danharrin:
Thanks!

PR #2225, opened 2022-04-26
FIX: SpatieMediaLibraryFileUpload should use existing `getUploadedFil…

bernhardh:
…eNameForStorageUsing` and not copy+paste code ;)

Otherwise `getUploadedFileNameForStorageUsing` will not work.


danharrin:
Thanks

PR #2239, opened 2022-04-27
feat: improved Builder design

Larsklopstra:
This is the improved design for the builder featuring:

- Visually better grouping
- More breathing space
- Show "Add block" button on mouseover
- Improved dropdown contrast
- Show iteration count in header (useful for scanning blocks)

image


packages/forms/resources/views/components/builder.blade.php
```diff
@@ -126,18 +131,12 @@ class="flex items-center justify-center h-8 w-8 rounded-full border border-gray-
 
         @if ((! $isItemCreationDisabled()) && (blank($getMaxItems()) || ($getMaxItems() > $getItemsCount())))
             <div x-data="{ isCreateButtonDropdownOpen: false }" class="relative flex justify-center">
-                <button
+                <x-filament::button
```
danharrin, line 134: We cannot use `filament::button` within the forms package, however a `forms::button` button component is being introduced in #2057

Larsklopstra:
> Can you post a dark mode screenshot too please? :)

The well uses a darker gray so the cards and inputs remain the same as other components

image

packages/forms/resources/views/components/builder.blade.php
Larsklopstra: Will this be merged soon? Then I'll refactor it already

Larsklopstra:
> Can we make the same changes to the repeater?

Yes!

ryangjchandler:
I wonder if we should replicate the border around the block actions on the right hand side too (delete button), perhaps a `border-l`?

Larsklopstra:
> I wonder if we should replicate the border around the block actions on the right hand side too (delete button), perhaps a `border-l`?

Would look too "blocky", there's no need for a divider on that side because the whitespace already makes for a great divider :)

packages/forms/resources/views/components/builder.blade.php
danharrin: This will be merged first

Larsklopstra:
@danharrin added the opt-in containment that will add the well, by default it has none

johncarter-:
Any chance of a collapse/expand on section headers, with an option to toggle expand collapse on all at the top?

ryangjchandler:
> Any chance of a collapse/expand on section headers, with an option to toggle expand collapse on all at the top?

Yeah, I like this idea. Being able to collapse a block, etc.

Larsklopstra:
> > Any chance of a collapse/expand on section headers, with an option to toggle expand collapse on all at the top?
> 
> Yeah, I like this idea. Being able to collapse a block, etc.

Sounds good, maybe for a later PR? Let's get this merged first 👀

danharrin:
We already have the x-collapse plugin installed, so it shouldn't be too time consuming to do it now @Larsklopstra?

danharrin:
Todo - merge in #2261

Larsklopstra:
Added expand (all) & collapse (all)

image

packages/forms/src/Components/Builder.php
```diff
@@ -225,6 +227,13 @@ public function disableItemDeletion(bool | Closure $condition = true): static
         return $this;
     }
 
+    public function contained(bool | Closure $condition = true): static
```
danharrin, line 230: Maybe we should call this feature "inset"? "Contained" sounds not clear enough

Larsklopstra: Sounds good

danharrin:
Todo: merge in #2057 and utilize `<x-forms::button>`

danharrin:
Todo: reflex changes on repeater

danharrin:
Todo: utilitize the dropdown.item blade component

danharrin:
Merged #2057 into this, when that is merged this will be too.

PR #2241, opened 2022-04-27
fix: card consistency

Larsklopstra:
**Before:**

Before

**After:**

After

**Other card elements:**

Table

Card
^ _This example is still missing a `shadow-sm`_

### DIFF
--- a/packages/admin/resources/views/components/card/index.blade.php
+++ b/packages/admin/resources/views/components/card/index.blade.php
@@ -6,7 +6,7 @@
 ])
 
 <div {{ $attributes->class([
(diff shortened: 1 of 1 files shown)


danharrin:
Ref: #1784

danharrin:
Is there a way we can keep the shadow as it looks 100x better on gray backgrounds, but also make it look ok on white backgrounds too?

danharrin:
Then we can change all the other cards to use the new styling that works on gray and white backgrounds

danharrin:
Gonna close this, I don't wanna lose the widget shadows just yet 👍

PR #2242, opened 2022-04-27
fix: button text legibility

Larsklopstra:
- Remove tracking
- Decrease text size to fit in better with UI

**Before:**

Before

**After:**

After


danharrin:
Have you checked that the icons are still a good size?

danharrin:
Replaced by #2057

PR #2251, opened 2022-04-28
fix: tab.blade.php padding

mattkibbler:
Fixes issue where "p-6" class persists and overrides "p-0" class when tabs are hidden, resulting in the padding of hidden tabs adding extra white space.

E.g..

Tab 1, being the first tab, appears at the top and shows the padding from each of the other tabs below.

DTV_Dashboard

Tab 2, being the second tab, has the padding from tab 1 above it and the padding of the rest of the tabs below.

DTV_Dashboard

etc...


danharrin:
Thanks!

PR #2253, opened 2022-04-28
refactor: move `HasExtraXAttributes` concerns into support package

ryangjchandler:


danharrin:
Mark the old traits as deprecated, and import the support traits `as BaseTrait`

PR #2257, opened 2022-04-29
Feature: Show filters on top of the table

dhruva81:
Greetings,

While building one application, I have the requirement to display filters on top of the table.

image

By adding following line to table builder component, the filters will be displayed on the top of table.

```
public bool $showFiltersOnTop = true;
```

By default, the filters will show as usual. 

Thank you.


lfleigl:
Works perfect for me. Thanks a lot.

ralphjsmit:
Great idea!

dhruva81:
> Does this also work in the admin panel for resources?

Yes! this also works with admin panel for resources.

Add `public bool $showFiltersOnTop = true; `  to ListRecords page.

ralphjsmit:
> > Does this also work in the admin panel for resources?
> 
> Yes! this also works with admin panel for resources.
> 
> Add `public bool $showFiltersOnTop = true; ` to ListRecords page.

Ah, yes of course. Nice!

dhruva81:
Hi @danharrin 

According to this PR, one can use property on livewire component to show filters above table.

```
public bool $showFiltersOnTop = true;
```

But, according to documentation, filament's standard way of customization is using protected method on component.
For example, (https://filamentphp.com/docs/2.x/tables/filters#customizing-popover-columns)

```
protected function getTableFiltersFormColumns(): int
{
    return 3;
}
```

So kindly let me know, should I update this PR from  using property `public bool $showFiltersOnTop = true; `  to protected method like below.

```
protected function getTableFiltersOnTop(): bool 
{
    return true;
}
```

Please guide.

danharrin:
Yes please, because public properties are editable via Livewire. We should use a protected method 👍

dhruva81:
Hi @danharrin  

As suggested by you, I have changed the public property `$showFiltersOnTop` to protected method `showFiltersAboveTable()`.

By default, filters will be shown at there default position. To show filters above table,  override the showFiltersAboveTable() method on table livewire component:


```
protected function showFiltersAboveTable(): bool 
{
    return true;
}
```

Thank you.

danharrin:
Thanks! I've slightly changed the syntax for more flexibility in the future, and added docs.

PR #2259, opened 2022-04-29
feature: Ternary filter

dododedodonl:
I often find myself writing a filter that does this. I think it is a useful addition to core.

```php
<?php
use Filament\Tables;

// Before
Tables\Filters\SelectFilter::make('is_activated')
    ->options([
        'yes' => 'Yes',
        'no' => 'No',
    ])
    ->placeholder('-')
        if (blank($data['value'])) {
            return $query;
        }

        return ($data['value'] == 'yes')
            ? $query->whereNotNull('activated_at')
            : $query->whereNull();
    }),
    
// After
Tables\Filters\TernaryFilter::make('is_activated')
    ->nullableColumn('activated_at'),

```

I used this as a base for a `TrashedFilter` which allows to easily show soft deleted records.

Edit: renamed the filter, made use of the default column name that has been set and added documentation


dododedodonl:
The normal `Filter` has only two states, applied or not applied, but a third state is needed: not-applied, "true" applied, "false" applied. For a `TrashedFilter` that would be `withoutTrashed()`, `withTrashed()` and `onlyTrashed()` respectively.

ryangjchandler:
@dododedodonl Ah, yeah sorry. I'm following now.

zepfietje:
I ran into this multiple times as well. In fact, I think I'd use the type of filter added in this PR instead of the default checkbox filter in almost all cases.

A few suggestions:

Wouldn't it make sense to add a default placeholder like `-`?

Shouldn't `nullableAttribute()` and `booleanAttribute()` be renamed to `column()` to be consistent with the existing `SelectFilter` (which isn't required as it should default to the filter name)? Then something like `nullable()` could optionally be chained to change the conditions.

Also, I'm not sure if `BooleanFilter` is the right name for this type of select filter.

dododedodonl:
I added a placeholder in the language file, so overwriting is easier. Thanks for the suggestion!

I'm not sure about naming either, this was the most obvious to me.

dododedodonl:
Made some changes to better match how other filters work in filament. Added some documentation with examples as well.

danharrin:
Thanks for this.

How about instead of introducing a new class, we do something like:

```php
Tables\Filters\SelectFilter::make('is_activated')
    ->ternary([optionally specify custom option labels]),
```

danharrin:
`ternary()` can internally set up `options()`, `query()` etc for you

danharrin:
Also, select form components include a `boolean()` method that does something similar setting the options automatically, so I'm leaning towards suggesting that we use `boolean()` instead of `ternary()` for consistent naming.

danharrin:
I've refactored this a bit, added support for relationships, and updated the docs. Thanks!

PR #2261, opened 2022-04-29
fix: disabled repeater still showing 'Add', 'Delete', 'Drag' button

wychoong:


danharrin:
Thanks, but lets merge these changes into #2239

PR #2269, opened 2022-04-30
add missing datatype declaration

maherelgamil:


ryangjchandler:
Livewire properties are better left untyped unless they're an object type / class type. Thanks though!

PR #2272, opened 2022-04-30
feat: adds multi action button to List, Edit and Create Resource pages

awcodes:
This adds the ability to have a multi action drop down on page resources for collapsing actions to save space.

Implemented with a public property on each resource:
```
public bool $hasMultiActionButton = true;
```
Screen Shot 2022-04-30 at 4 25 27 PM
Screen Shot 2022-04-30 at 4 25 33 PM
Screen Shot 2022-04-30 at 4 25 41 PM
Screen Shot 2022-04-30 at 4 25 48 PM
Screen Shot 2022-04-30 at 4 26 10 PM
Screen Shot 2022-04-30 at 4 26 18 PM


awcodes:
Sorry if a lot of the diffs look worse than they are. That’s just my ide formatting. I assume your cs fixer actions will handle that. Please let me know if and what I can do it the future to make these better for you.

danharrin:
I like this, but we do have plans for a `Group` for action which renders the actions inside a dropdown, and I feel like supporting both options could result in maintenance that we don't need. Wdyt @ryangjchandler ?

ryangjchandler:
Yeah, I generally prefer the idea of having an `ActionGroup` or something that is a single button that wraps multiple actions. Very similar concept, but instead of being an add-on to an existing action, it's completely separate.

awcodes:
No worries. Thank you for taking a look at it. Can't wait to see your implementation. I'll close the PR.

ralphjsmit:
Really love the idea of this and how it looks. ❤️

I'm just putting this as a reference here. Would it be possible to support forms as well? @danharrin @ryangjchandler 

I'm not sure about the intended code architecture, but it would definitely be very powerful. It could be very similar to how we can now define a `->form()` on a filter. 

Consider this example:

Schermafbeelding 2022-05-04 om 18 36 34


An action at the top of the page would be perfect for this sort of functionality.

danharrin:
That example is identical to using an action modal atm, but with a popover instead of a modal?

ralphjsmit:
> That example is identical to using an action modal atm, but with a popover instead of a modal?

Hmm, yeah, that's another way of looking at it. Didn't realize that yet. On the other hand, there is certainly a UX-difference between  a modal and a popover.

PR #2285, opened 2022-05-04
fix: Use the "resolving" service event in the PluginServiceProvider

coleshirley:
Currently the plugin service provider force-resolves the filament service whenever the first filament plugin's ServiceProvider is registered. This interrupts the normal service lifecycle which allows for calls to `$app->extend()` (such as in my multi-context package).

This is also related to the changes made in https://github.com/laravel-filament/filament/pull/2055

### DIFF
--- a/packages/admin/src/PluginServiceProvider.php
+++ b/packages/admin/src/PluginServiceProvider.php
@@ -59,21 +59,18 @@ abstract class PluginServiceProvider extends PackageServiceProvider
 
     public function packageRegistered(): void
     {
-        $this->app->singletonIf(
-            'filament',
-            fn (): FilamentManager => app(FilamentManager::class),
-        );
+        $this->app->resolving('filament', function () {
+            Facades\Filament::registerPages($this->getPages());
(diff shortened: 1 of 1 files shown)


ryangjchandler:
Can we be certain that this won't introduce any breaking changes for other plugin developers? I don't think it will, but want to be certain.

coleshirley:
@ryangjchandler I don't believe it will either. This brings also more conformity with what the `Register` and `Boot` methods of a service provider are supposed to be used for.

https://laravel.com/docs/providers#the-register-method
> "As mentioned previously, within the register method, you should only bind things into the [service container](https://laravel.com/docs/container). You should never attempt to register any event listeners, routes, or any other piece of functionality within the register method."

If this does break something because a plugin is calling the Filament service in their provider's register method then we can point them to this documentation saying that they should do so in the `Boot` method instead

ryangjchandler:
Hm, yeah.. the reason we're forced into using the register method is because of the order of execution of route loading in packages.

danharrin:
Thanks!

PR #2290, opened 2022-05-04
Updates Akaunting dependency.

aidan-casey:
This PR simply updates the Akaunting dependency version so the latest version may be used.


danharrin:
Thanks!

PR #2302, opened 2022-05-06
Fix issue with id not being prefixed

gavinhewitt:
The issue is caused when you use a MultiSelectFilter: `Column 'id' in where clause is ambiguous`. This fix prefixes the id column correctly.


danharrin:
Thanks!

PR #2307, opened 2022-05-07
fix: missing polish translation

kamilst96:
A few missing words here and there, and associate/dissociate were missing completely.
I am not confident in associate/dissociate, because I don't have any actual use case for it.

@lolen  was the last one to work on translations, so if he can opinion it, then great.
If not, it's better than having sudden English on page.

### DIFF
--- a/packages/admin/resources/lang/pl/login.php
+++ b/packages/admin/resources/lang/pl/login.php
@@ -4,6 +4,8 @@ return [
 
     'title' => 'Logowanie',
 
+    'heading' => 'Zaloguj się',
+
     'buttons' => [
 
         'submit' => [
@@ -29,10 +31,8 @@ return [
     ],
 
     'messages' => [
-        'failed' => 'Błędny login lub hasło.',
-        'throttled' => 'Za dużo nieudanych prób logowania. Proszę spróbować za :seconds sekund.',
+        'failed' => 'Nieprawidłowe poświadczenia.',
+        'throttled' => 'Zbyt wiele nieudanych prób logowania. Spróbuj ponownie za :seconds sekund.',
     ],
-
-    'heading' => 'Zaloguj się do swojego konta',
 
 ];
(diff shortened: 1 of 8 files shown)


danharrin:
Thanks!

packages/admin/resources/lang/pl/login.php
```diff
@@ -29,10 +31,8 @@
     ],
 
     'messages' => [
-        'failed' => 'Błędny login lub hasło.',
```
lolen, line 32: The previous values were identical to one of the most popular translation packages for Laravel [https://github.com/Laravel-Lang/lang/blob/main/locales/pl/auth.php#L15](auth.php#L15)

packages/admin/resources/lang/pl/pages/dashboard.php
```diff
@@ -2,6 +2,6 @@
 
 return [
 
-    'title' => 'Dashboard',
```
lolen, line 5: I didn't translate this phrase because it is known in Polish. And the translation "Pulpit" can be misleading, because "pulpit" = "desktop", and the "dashboard" = "deska rozdzielcza" or "tablica rozdzielcza". Is not used on the Internet, it is more common in the automotive industry

packages/admin/resources/lang/pl/resources/pages/create-record.php
```diff
@@ -2,9 +2,9 @@
 
 return [
 
-    'title' => 'Utwórz :label',
```
lolen, line 5: I used "Utwórz" because it is in many popular applications, e.g. in Wordpress it has "Dodaj wpis" and not "Dodawanie Wpisów".

lolen:
@kamilst96 I added a comment with my opinion and my way of thinking. The other translations are great :+1:

danharrin:
If you can both agree on these changes, would you mind sending another PR to fix?

kamilst96:
Yeah, sorry about that. I'll make a new PR. We need to address 2.12.0 changes in **forms** anyway.

PR #2318, opened 2022-05-08
Forms\Components\Hidden: CanBeAutocompleted

DSpeichert:
Useful for hidden field as `username` field on password change forms (https://goo.gl/9p2vKq)

### DIFF
--- a/packages/forms/resources/views/components/hidden.blade.php
+++ b/packages/forms/resources/views/components/hidden.blade.php
@@ -2,6 +2,7 @@
     id="{{ $getId() }}"
     {!! $isRequired() ? 'required' : null !!}
     type="hidden"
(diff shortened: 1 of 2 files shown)


DSpeichert:
This isn't the right implementation, it should be `type="text"` with `hidden` attribute instead.
See: https://stackoverflow.com/a/65672949

ryangjchandler:
I'd recommend trying something like below instead:

```php
TextInput::make('username')
    ->withExtraAttributes(['hidden'])
    ->autocomplete('username')
```

danharrin:
Yeah, if this isn't the right implementation then use Ryan's solution instead.

DSpeichert:
That looks like exactly what's needed, but it does not work:
> Method Filament\Forms\Components\TextInput::withExtraAttributes does not exist.

Or am I using the wrong Input component class?

danharrin:
I think the method is `extraInputAttributes()`

PR #2336, opened 2022-05-10
fix: update pl translation for v2.12

kamilst96:
- Translation strings synced to match v2.12
- All actions are worded in imperative case: _Edycja_ vs _Edytuj_
- Restored login messages from #2307
- Restored "Dashboard" string to #2307
- The `selected_count` in tables package didn't actually use `[5,*]` case, so I improvised an alternative.

@lolen what do you think? Let's not merge this until consensus is reached.


danharrin:
Mark this as "ready for review" when you want it merged

packages/tables/resources/lang/pl/table.php
```diff
@@ -107,7 +125,7 @@
 
     'selection_indicator' => [
 
-        'selected_count' => '{1} 1 rekord zaznaczony.|[2,4]:count rekordy zaznaczone.|[5,*]:count rekordów zaznaczonych.',
```
lolen, line 110: I would keep the previous version. It is grammatically correct, e.g. for 3 pieces. And the user has value in one place which is more intuitive, and there's no need to look for values. 👀 
X rekord zaznaczony.
Zaznaczono rekordów: X

lolen (changes requested):
I would only change where there is a comment and that's all correct. Overall great job! 👌

packages/tables/resources/lang/pl/table.php
kamilst96: Yes, it is correct and I'd rather keep it, but it doesn't appear to be working.
- 1 rekord zaznaczony.
- 2 rekordy zaznaczone.
- 4 rekordy zaznaczone.
- 5 rekordy zaznaczone (!)

lolen: The problem is not in the translation, but in another place. You can run `tinker`, and paste:
`Lang::choice('{1} 1 rekord zaznaczony.|[2,4]:count rekordy zaznaczone.|[5,*]:count rekordów zaznaczonych.', 5)`
And it works well.

kamilst96: I see, so something in tables? Picking your version, then.

lolen (approved):

lolen (approved):

lolen (approved):

lolen (approved):

lolen (approved):

lolen (approved):

lolen (approved):

lolen (approved):

PR #2349, opened 2022-05-10
Raise z-index for select dropdown

Cannonb4ll:
This is because sometimes 3rd party plugins will work with `z-index` and have this as a higher value than `1` which choices does now. You will also always want to have the choices dropdown at first on the layers (always on top)

**Broken:**
CleanShot 2022-05-10 at 15 50 22

**Fixed:**
CleanShot 2022-05-10 at 15 50 36

I'm also open for a z-index 10 so you're even more sure the choices list will overlap other forms/tools.

### DIFF
--- a/packages/forms/resources/css/components/select.css
+++ b/packages/forms/resources/css/components/select.css
@@ -111,7 +111,7 @@
 }
 
 .is-active.choices__list--dropdown, .is-active.choices__list[aria-expanded] {
-    @apply visible;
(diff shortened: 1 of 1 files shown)


danharrin:
`z-10` is consistent with other fields like the Date Picker, so let's use that. I just suggested `z-1` because that was the default for Choices.js

danharrin:
Thanks!

PR #2368, opened 2022-05-11
Add view record to Simple (modal ) Resources

invaders-xx:
Just add functionality to view records for the simple (modal) resources


packages/admin/src/Resources/Pages/ManageRecords.php
```diff
@@ -7,4 +7,5 @@ class ManageRecords extends ListRecords
     use ListRecords\Concerns\CanCreateRecords;
     use ListRecords\Concerns\CanDeleteRecords;
     use ListRecords\Concerns\CanEditRecords;
+    use ListRecords\Concerns\CanViewRecords;
```
danharrin, line 10: Don't think this should be a default trait IMO

invaders-xx: IMO, from a CRUD perspective you should be able to Create/Read/Update/Delete if you have the right to do so. So for me, this should be a default trait.

PR #2374, opened 2022-05-11
Evaluate action icon

titantwentyone:
Just a small amend to provide the ability to dynamically set the icon of an Action.


titantwentyone:
Okay. So PHPStan failed here (I should have run prior to raising a PR - apologies!). Seems the location of the trait `EvaluatesClosures` is not accessible in the abstract class `Filament\Support\Actions\BaseAction`. Would an idea be to defer the inclusion of `Filament\Support\Actions\HasIcon` to those child classes that implement `EvaluatesClosures`?

danharrin:
fixed it!

titantwentyone:
Sweet. This opens up a few cother options as well! Thank you.

PR #2375, opened 2022-05-11
fix modal component taking empty space

wychoong:
i'm using the modal in custom blade file 
```php
<x-filament::modal id="page-action" display-classes="block" width="2xl">
```
recent extraction to Support package causes missing attributes to pass into components

**Before**
image

**After**
image

p/s: only notice `displayClasses` is missing for now

### DIFF
--- a/packages/admin/resources/views/components/dropdown/item.blade.php
+++ b/packages/admin/resources/views/components/dropdown/item.blade.php
@@ -3,7 +3,7 @@
 ])
 
 <x-filament-support::dropdown.item
-    :attributes="$attributes->merge($slots)"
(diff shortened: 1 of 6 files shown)


danharrin:
I don't think that was the correct solution, I've pushed the one I think is correct, can you test my changes?

wychoong:
yeah, your changes is working.

is @captureSlots for controlling what slots can be passed in? not sure why my change previously is working LOL

danharrin:
Yeah, since slots do not appear in the `$attributes` bag, I made a custom directive that captures them into an array that we can merge into the bag. Not sure why your changes worked either lol ahahah

PR #2377, opened 2022-05-11
fix: Duplicate BelongsToMany pivot records

### LINKED ISSUE #1804: Belongs to Many: Prevent Attaching already Attached Model
### Discussed in https://github.com/laravel-filament/filament/discussions/1510

Due to laravel/framework#4962, belongs to many relation managers have unexpected behaviour when dealing with multiple relationships between the same 2 Eloquent records.
- Detaching one instance detaches all.
- Editing pivot data does not work except for the first instance.

<sup>Originally posted by **intrepidws** February  8, 2022</sup>
I have a Belongs to Many relationship between Registrants and Codes. I have a `CodesRelationManager` class (which extends `BelongsToManyRelationManager`) and am using that to allow codes to be attached to registrants. Currently, there is nothing to prevent someone from attaching a code to a registrant if that code is already attached to the registrant. Is there a way to prevent this?

I have the following code for my `attachForm` method:

```
public static function attachForm(Form $form): Form
{
    return $form
        ->schema([
            static::getAttachFormRecordSelect()
                ->searchable(['code', 'type', 'title']),
            Forms\Components\Select::make('quantity')
                ->required()
                ->options(array_combine(range(1, 100), range(1, 100)))
                ->default(1),
        ]);
}
```</div>

### LINKED ISSUE #2356: RelationManager: use pivot's primaryKey if available
### Package

filament/filament

### Package Version

v2.12.1

### Laravel Version

v9.11.0

### Livewire Version

v2.3

### PHP Version

PHP 8.1.5

### Bug description

If I create LocationUser with two same location_id and user_id:
| pivot_id | location_id | user_id |
|---|---|---|
| 11 | 2 | 7 |
| 12 | 2 | 7 |

Filament will fail to distinguish between them, assign them user_id for x-model, so they both will be selected for bulk operations. Would it be possible to configure the column used by tables to distinguish records?

My best guess would be to check whether getKey() on pivot returns null, which would be usual for pivot table, and then apply the current logic. If it returns something, use it as key.

Table config as below. All tables have models.

```
User:
- id

Location:
- id
- date_from
- date_to
- is_active

LocationUser:
- pivot_id, PK
- location_id
- user_id
- pivot_date_from
- pivot_date_to
- pivot_is_active
```

### Steps to reproduce

- Create tables as in description
- Treat pivot_id as PK for LocationUser
- Avoid inbuilt `attach() / detach()` methods
- Create two records, bridging the same records

### Relevant log output

_No response_

### PR DESCRIPTION
**Pull request by danharrin**
Fixes #1510.
Fixes #1804.
Fixes #2356.

### CODE REVIEW

**Comment by kamilst96** [comment]
It appears to be working, though I failed to use my custom PK, so I removed prefixes from intermediate table. They won't be needed directly anyway if this pulls off, but I'll keep trying.

**Working:**
- Attaching / detaching / editing records.
- Table selector and actions correctly use PK
- Saving pivot data through $record->pivot directly in custom actions

**Issues:**
- For all of this to work, I have to use `Model`, not `Pivot`. Maybe it's just my messed config.
- Pivot's own `id` key masks record's `id`, so creating url links for them is impossible. `Collection::pluck()` is a little confused because of masked `id` allowing duplicates
- There is a pivot available on `$record` now, but I can't access it in columns and inputs, using dot notation,
though it works in custom actions, when injecting the collection.
- You are also using `pivot_` prefixed properties, and removing them before passing record to user. My `pivot_is_active` became `pivot_pivot_is_active`, and still qualified for purging from record, so I was a little confused when I didn't see my keys.

Would it be possible to ditch merging resource with relation and make pivot available with dot notation?
Alternatively, we could have `pivot_` prefixed attributes, and smash then into their relation when saving.

**Comment by danharrin** [comment]
> I failed to use my custom PK

Did you set up a custom Pivot model, and set the `$primaryKey` to be your custom key?

> I have to use Model, not Pivot

Not sure what you mean by this?

> Pivot's own id key masks record's id, so creating url links for them is impossible.

Yeah, this is solved with a custom PK on the pivot table though. So that Filament is able to understand that two duplicate records are different, we need to overwrite the original ID with the pivot ID.

> There is a pivot available on $record now, but I can't access it in columns and inputs, using dot notation

It's always been the case that you can simply alter pivot data without the `pivot.` prefix, so binding to pivot columns and inputs is as simple as `TextInput::make('pivot_field_name')`. This works, right?

> You are also using pivot_ prefixed properties, and removing them before passing record to user.

I don't remember doing that at all, where are these `pivot_` properties coming from in your example? It must be an internal Eloquent thing, I just searched the codebase for `pivot_` with no results.

> Would it be possible to ditch merging resource with relation and make pivot available with dot notation?

This would be a breaking change, and we only need to make sure the primary key is customisable for this to not be an issue.

**Comment by kamilst96** [comment]
Okay, I messed up by extending `Model` instead of `Pivot`.
- Custom PK works.
- Edit and Detach actions get null instead of Model. Custom Action gets Model properly. They work when I disable `$allowsDuplicates`.
- Selecting still uses related model's id, instead of pivot's, so problem persists. Previously, pivot's `id` masked record's `id`, so it was placebo.

I'm pasting my models, just in case.

```php
class Person extends Model
{
    use HasFactory;

    protected $table = 'people';

    protected $guarded = [
        'id',
        'created_at',
        'updated_at',
    ];

    public function shelters(): BelongsToMany
    {
        return $this->belongsToMany(Shelter::class, ShelterHasPerson::class)
            ->withPivot([
                'role_id',
                'pivot_id',
                'pivot_date_from',
                'pivot_date_to',
                'pivot_is_active',
            ]);
    }
}
```

```php
class Shelter extends Model
{
    use HasFactory;

    protected $table = 'shelters';

    protected $guarded = [
        'id',
        'created_at',
        'updated_at',
    ];

    public function people(): BelongsToMany
    {
        return $this->belongsToMany(Person::class, ShelterHasPerson::class)
            ->withPivot([
                'role_id',
                'pivot_id',
                'pivot_date_from',
                'pivot_date_to',
                'pivot_is_active',
            ]);
    }
}
```

```php
class ShelterHasPerson extends Pivot
{
    use HasFactory,;

    protected $table = 'shelter_has_people';

    protected $primaryKey = 'pivot_id';
    public $incrementing = true;

    protected $guarded = [
        'pivot_id',
        'created_at',
        'updated_at',
    ];
}
```

danharrin:
> Edit and Detach actions get null instead of Model. Custom Action gets Model properly. They work when I disable $allowsDuplicates.

Where are you testing this? The edit form gets filled with data, so why would the model be null?

> Selecting still uses related model's id, instead of pivot's, so problem persists

Which problem are you referring to, and how can I replicate it?

kamilst96:
Here's a test repo
https://github.com/kamilst96/filament.test

Seeded user:
U: test@example.com
P: test

There's only one resource, one relation, and three models, including pivot. Minimal inputs to keep classes simple.

Person belongsTo Shelter
Shelter belongsTo Person
ShelterHasPerson as pivot

After confirming they work as expected, I cloned PR and switched `$allowsDuplicates` to true.

Selected rows still behave wrong:

Zrzut ekranu 2022-05-12 o 16 07 29

But now, the Edit/Detach/Delete actions on table rows result in this error:

```
Filament\Resources\RelationManagers\BelongsToManyRelationManager::Filament\Resources\RelationManagers\Concerns\{closure}():
Argument #1 ($record) must be of type Illuminate\Database\Eloquent\Model, null given,
called in /var/www/test/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php on line 36
```

danharrin:
Alright, fixed that! Also make sure to use `->using(PivotClassName::class)` on the relationship method.

kamilst96:
It works perfect - I didn't even need to alter my repos.

As for `using()`, I define relationship by specifying pivot class instead of pivot table in second argument, so it picks the class automatically.

```php
return $this->belongsToMany(Shelter::class, ShelterHasPerson::class)
    ->withPivot([
        'pivot_id',
        'pivot_is_active',
    ]);
```

Thanks for your patience, by the way - I wasn't the best at giving you any pointers.

danharrin:
Great!

GarethSomers:
Hi @danharrin this doesn't work for me... might be my fault, but I can't see why?

_Just to confirm I've downloaded v2.12.10_

I have a Collection which can belong to a Collection defined by a parents and children methods:

```
class Collection extends Model
{
    // ...

    public function parents (): \Illuminate\Database\Eloquent\Relations\BelongsToMany
    {
        return $this->belongsToMany(Collection::class, 'collection_parents', 'collection_id', 'parent_id')
            ->withPivot(["id", "is_primary"])
            ->using(CollectionParent::class);
    }

    public function children (): \Illuminate\Database\Eloquent\Relations\BelongsToMany
    {
        return $this->belongsToMany(Collection::class, 'collection_parents', 'parent_id', 'collection_id')
            ->withPivot(["id", "is_primary"])
            ->using(CollectionParent::class);
    }
}
```

The intermediate table

```
class CollectionParent extends Pivot
{
    // Assumes it's `collection_parent`?
    public $table = 'collection_parents';
}
```

My relationship manager:

```
class ParentsRelationManager extends BelongsToManyRelationManager
{
    protected static string $relationship = 'parents';
    protected static ?string $inverseRelationship = 'children';
}
```

And when attaching it still offers me the same options:
Screenshot 2022-05-16 at 17 12 35

danharrin:
Hmm potentially this is an issue with relating the table to itself? If you can provide a demo repo with a seeder then I can test it out and work on a fix

GarethSomers:
Hey @danharrin here you go

https://github.com/GarethSomers/laravel-filament-belongs-to-many-bug

1. Naturally you'll need to migrate, seed and create a user.
2. Navigate to `/admin/collections/3/edit`
3. Under "Parents" click "Attach"
4. Search for "Foo Recipes" (which is already attached)
5. Click "Attach"

Expected result: It shouldn't appear in search by the sounds of it?
Actual result: Let's you attach.

---

Were you testing with the same model @danharrin or was it 2 different models?

Additionally I found that you can associate the Collection model with itself, not a big deal but undesirable. Though you can only associate the Collection with itself once!? So something is working there...

kamilst96:
@GarethSomers `ParentsRelationManager` in demo repo doesn't have `protected bool $allowsDuplicates = true;` set.

kamilst96:
Ah, you mean attach search results don't ignore already attached pivots.

danharrin:
I've replicated this issue outside of Filament. I think it could be an Eloquent bug...

image

GarethSomers:
You might be right @danharrin as I have encountered many issues with this type of relationship before.

What I've done to overcome this is simply make 2 relationships on each model (see below). I use `children` on the frontend, and the `pivotChildren` with a `HasManyRelationManager` on the backend. I also use custom rules to prevent parents having children which are parents and vice-versa.

e.g.

```
    public function children (): BelongsToMany
    {
        return $this->belongsToMany(Example::class, 'example_children', 'parent_id', 'child_id')
            ->using(ExampleChild::class);
    }

    public function pivotChildren(): HasMany
    {
        return $this->hasMany(CollectionChild::class, 'parent_id', 'id');
    }
```

PR #2382, opened 2022-05-11
Feature: Updates Section component styling

awcodes:
Updates Section component styling per discussion #2305.

I also did not remove the Card component as I think that could be a breaking change to existing installs, so I feel it best to let you handle that. :)

Cheers!

Screen Shot 2022-05-11 at 3 51 19 PM
Screen Shot 2022-05-11 at 3 51 36 PM


kamilst96:
When the header is clearly distinguishable from content, the height difference between all variants becomes apparent - they have different height when default, with collapse toggle, and with description. Maybe it's just me noticing things though.

awcodes:
Thanks for the feedback @kamilst96. I’m not sure what you are getting at. Can you explain further?

SebastiaanKloos:
This is a good improvement!

danharrin:
Thanks!

awcodes:
Cheers @danharrin 

Thank you!

PR #2391, opened 2022-05-11
fix: missed labels from PR #2385

kamilst96:
These labels referenced strings from **tables** that were removed in #2385


danharrin:
Cheers

PR #2392, opened 2022-05-12
fix: [forms] block builder when block non-existant

phpsa:
Fixes an issue - though rare as it might be -- could be only during development of a resource Block:

How to replicate:
- Using Block builder - create an record
- Change the Block Builder and try and edit the record - Throws exception

To Fix:
Updated the GetChildComponentContainers method to filter our null blocks


danharrin:
Thanks!

PR #2401, opened 2022-05-12
Disable instant filepond upload

wychoong:
not sure how to proceed as not familiar with the livewire upload and state, i'm stuck with on submit it saw field is required for obvious reason that the field is not updated with file


danharrin:
As discussed on Discord, closing this

PR #2402, opened 2022-05-12
[ar] Translate move label

mohamedsabil83:


mohamedsabil83:
Better to make a new PR for all new translations according to latest update

PR #2403, opened 2022-05-12
[ar] Update translations - Translate new phrases - Revert missing phrase

mohamedsabil83:


mohamedsabil83:
Where is the new place of the key?

PR #2409, opened 2022-05-12
Use `getRecordTitle()` for relation manager

pxlrbt:
@danharrin We already use `getRecordTitle()` for  the option labels, but not for the search results. I'd like to introduce the functionality to customize the label based on the record as in `Select` or `BelongsToSelect`.

As `getRecordTitle()` is already used and has a fallback there is no way to check whether it was defined, so we need to apply it to every query instead of the bit more performant `->pluck()`.

What do you think?

### DIFF
--- a/packages/admin/src/Resources/RelationManagers/Concerns/CanAttachRecords.php
+++ b/packages/admin/src/Resources/RelationManagers/Concerns/CanAttachRecords.php
@@ -108,6 +108,7 @@ trait CanAttachRecords
                     return [];
                 }
 
+                /** @var BelongsToMany $relationship */
                 $relationship = $livewire->getRelationship();
 
                 $displayColumnName = static::getRecordTitleAttribute();
--- a/packages/admin/src/Resources/RelationManagers/Concerns/CanAttachRecords.php
+++ b/packages/admin/src/Resources/RelationManagers/Concerns/CanAttachRecords.php
@@ -90,16 +90,14 @@ trait CanAttachRecords
                     $isFirst = false;
                 }
 
-                $relatedKey = $relationship->getRelatedKeyName();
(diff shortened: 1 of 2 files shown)


packages/admin/src/Resources/RelationManagers/Concerns/CanAttachRecords.php
```diff
@@ -92,7 +93,10 @@ protected static function getAttachFormRecordSelect(): Select
                     ->whereDoesntHave($livewire->getInverseRelationshipName(), function (Builder $query) use ($livewire): void {
                         $query->where($livewire->ownerRecord->getQualifiedKeyName(), $livewire->ownerRecord->getKey());
                     })
-                    ->pluck($displayColumnName, $relationship->getRelated()->getKeyName())
+                    ->get()
+                    ->mapWithKeys(static fn (Model $record) => [
+                        $record->{$relationship->getRelated()->getKeyName()} => static::getRecordTitle($record),
```
danharrin, line 98: Here and on line 121, we can store the result of `$relationship->getRelated()->getKeyName()` in a variable so that we don't call it more than once.

pxlrbt: That's 2 different methods. We would need to extract `$relationship` and the new variable outside the `Select` component scope then.

danharrin: I meant two variables, one in each closure.

pxlrbt: Ah, now I get it. We have a loop with `mapWithKeys()` 😅

PR #2411, opened 2022-05-12
Do not process schema when hidden

titantwentyone:
Small change to submit an empty array as the schema for anything that has `childComponents` when it is `hidden`. I noticed that if the `schema` was generated dynamically with a callback, this would run even if the component was hidden. For my use case, I wanted to display a `Section` with dynamically created components when editing but not when creating as the generation was dependent on a property that could not be calculated at the time of creation.

### DIFF
--- a/packages/forms/src/Components/Concerns/HasChildComponents.php
+++ b/packages/forms/src/Components/Concerns/HasChildComponents.php
@@ -25,7 +25,7 @@ trait HasChildComponents
 
     public function getChildComponents(): array
     {
-        return $this->evaluate($this->childComponents);
+        return $this->isHidden() ? [] : $this->evaluate($this->childComponents);
     }
 
     public function getChildComponentContainer(): ComponentContainer


danharrin:
Hmm I think this is a breaking change and usually expected behaviour to return hidden values even if they're not being used. You can prevent this behaviour using `->dehydrated(fn (Component $component): bool => $component->isVisible())`

titantwentyone:
I tried your suggestion. My code is (I may have misunderstood!):

```
Section::make('Attendance Option Info')
    ->hiddenOn(Pages\CreateCampEvent::class)
    ->dehydrated(fn (Component $component): bool => $component->isVisible())
    ->schema(function($record) {
        $components = [];
        foreach($record->getAttendanceOptionsBreakdown() as $option => $count) {
            $components[] = Placeholder::make($option)
                ->content($count);
        }
        return $components;
    })
```
This still seems to execute the callback in schema however when accessing the Create page.

Can you elaborate on why this may be a breaking change? If the return value of `schema` is essentially discarded if hidden, I'm not sure where/why it may be problematic? Just curious to see if there is another way and understand more of Filament internals! :)

danharrin:
> This still seems to execute the callback in schema however when accessing the Create page.

Sorry, I did't understand what you meant in the original post. I thought that you didn't want the hidden fields saved.

> Can you elaborate on why this may be a breaking change?

`getChildComponents()` is not just used to render the form, it's also used to fill & retrieve data from it. By skipping hidden fields, users will not be able to save data from hidden fields, or load it in when the form is initialised.

Unfortunately I don't believe that there is a way for you to only skip hidden fields when the form is rendered, as the closure will still be executed before that during the filling process.

I suggest using something like this in the Closure instead:

```php
if (! $record) {
    return [];
}
```

PR #2404, opened 2022-05-12
fix: Markdown editor diff issues

danharrin:
Fixes #2398.

PR #2412, opened 2022-05-12
fix: MarkdownEditor preview

themahdavi:
Improve markdown editor preview ui
just remove `border border-gray-300  dark:border-gray-600 shadow-sm`


danharrin:
Already implemented in #2404

PR #2441, opened 2022-05-16
doc: correct hidden() to visible() method

Michael-HEIW:
- the `hidden()` method was incorrect:

```php
// was:
->hidden(fn (Post $record): bool => auth()->user()->can('update', $record))
// should be:
->hidden(fn (Post $record): bool => ! auth()->user()->can('update', $record))
```

- I think the `visible()` method is easier to read, in this example.

```php
// now:
->visible(fn (Post $record): bool => auth()->user()->can('update', $record))
```

### DIFF
--- a/packages/tables/docs/05-actions.md
+++ b/packages/tables/docs/05-actions.md
@@ -224,7 +224,7 @@ BulkAction::make('delete')
 
 ## Authorization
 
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks!

PR #2453, opened 2022-05-18
fix: disable filepond transform

wychoong:
In the current implementation filepond is transforming the image before posting to livewire, this is causing issue for my use case where i rely on the image's `md5(binary)` as filename for uniqueness.

for normal use cases uploading the transformed image shouldn't cause any issue but my application allow user to upload images individually or using ZIP. This is where the transformed image binary will cause uniqueness issue when the same image is uploaded individually as well as in ZIP file.

TL;DR
- filepond transform image by default
- image is transform even no Crop or Resize properties are set

Fix:
- introduce shouldTransformImage variable to check for Crop and Resize related properties

https://pqina.nl/filepond/docs/api/plugins/image-transform/


danharrin:
Thanks

PR #2455, opened 2022-05-18
Feature: add date time to config for DateTime and DateTimePicker

Michael-HEIW:
Allows the config for DateTime to be set in config\forms.php, or defaults to the same current standard

I hope this is acceptable.

### DIFF
--- a/packages/forms/config/forms.php
+++ b/packages/forms/config/forms.php
@@ -1,6 +1,19 @@
 <?php
 
 return [
+    /*
+    |--------------------------------------------------------------------------
+    | Date / Time Formatting
+    |--------------------------------------------------------------------------
+    |
+    | These are the formats that Filament will use to display dates and times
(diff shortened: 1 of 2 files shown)


danharrin:
I believe this breaks the TimePicker, we also need formats for `time` and `time_with_seconds`

packages/forms/config/forms.php
```diff
@@ -1,6 +1,19 @@
...
+    |--------------------------------------------------------------------------
+    |
+    | These are the formats that Filament will use to display dates and times
+    | by default.
+    |
+    */
+
+    'date_format' => 'M j, Y',
+    'date_time_format' => 'M j, Y H:i',
+    'date_time_second_format' => 'M j, Y H:i:s',
```
danharrin, line 16: Maybe rename to `date_time_with_seconds_format`

PR #2456, opened 2022-05-18
New render hook before and after main content

mohamedsabil83:
Sometimes we need to put an alert for the user and one a good place to do this is inside the main content above the widgets


danharrin:
Thanks

mohamedsabil83:
You're Welcome

PR #2469, opened 2022-05-19
Adds support for `extraImgAttributes()` on ImageColumn

johncarter-:
Allows us to add attributes to the `img` tag on an `ImageColumn`, so we can do stuff like:
```php
        return $table
            ->columns([
                ImageColumn::make('logo')
                    ->extraAttributes(['class' => 'flex items-center h-16 w-32'])
                    ->extraImgAttributes(['class' => 'object-contain w-full h-full']),
            ]);
```

Getting logos to contain within the bounds of predefined area:

image

### DIFF
--- a/packages/tables/resources/views/columns/image-column.blade.php
+++ b/packages/tables/resources/views/columns/image-column.blade.php
@@ -11,12 +11,12 @@
             ></div>
         @else
             <img
-                {!! $attributes->merge($getExtraImgAttributes()) !!}
                 src="{{ $path }}"
                 style="
                     {!! ($height = $getHeight()) !== null ? "height: {$height};" : null !!}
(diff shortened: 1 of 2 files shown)


packages/tables/resources/views/columns/image-column.blade.php
```diff
@@ -11,6 +11,7 @@ class="rounded-full bg-center bg-cover"
             ></div>
         @else
             <img
+                {!! $attributes->merge($getExtraImgAttributes()) !!}
```
danharrin, line 14: This will merge the main attribute bag onto the extra attributes

Check out the HasExtraAlpineAttributes trait - there's a method to get the `AttributeBag` which we can use here instead

PR #2481, opened 2022-05-19
Update dutch translations for form components

sten:
Dutch translations for the wizard fields and some other dutch translation improvements.


danharrin:
Thanks

PR #2482, opened 2022-05-20
Add view action to relation's manager

invaders-xx:
Add a view action to RelationManager. I know that you will not accept to have this action by default, but I didn't know how to implement that as an option with the relation manager. 
Please share your thoughts and I will adapt


danharrin:
Check how we did the optional "associate" action for RMs, we can do the same thing.

invaders-xx:
I will check and adapt. Thanks for your feedback

invaders-xx:
@danharrin , are you ok ?

danharrin:
Thanks :)

PR #2483, opened 2022-05-20
[2.X] Add Limit to property

ManojKiranA:
I have been using `limit` in my `TextColumn` but I can't reuse it on my other methods such as `tooltip`

I have been trying to show the tooltip only if the limit exceeds

Before

```
TextColumn::make('instance_name')
->limit(20)
->tooltip(fn ($record): string => strlen($record->instance_name) >=20 ? $record->instance_name : "")
->searchable()
->sortable()
->default('N/A'),
```

After

```
TextColumn::make('instance_name')
->limit(20)
->tooltip(fn ($column,$record): string => strlen($record->instance_name) >= $column->limit ? $record->instance_name : "")
->searchable()
->sortable()
->default('N/A'),
```

### DIFF
--- a/packages/tables/src/Columns/Concerns/CanFormatState.php
+++ b/packages/tables/src/Columns/Concerns/CanFormatState.php
@@ -14,7 +14,7 @@ trait CanFormatState
 {
     protected ?Closure $formatStateUsing = null;
 
-    private ?int $limit = null;
+    protected ?int $limit = null;
 
     protected string | Closure | null $prefix = null;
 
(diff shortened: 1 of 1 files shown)


packages/tables/src/Columns/Concerns/CanFormatState.php
```diff
@@ -14,6 +14,8 @@ trait CanFormatState
 {
     protected ?Closure $formatStateUsing = null;
 
+    private ?int $limit = null;
```
danharrin, line 17: should be protected, and there should be a getLimit() public method

danharrin: done

danharrin:
Thanks

PR #2491, opened 2022-05-20
Add loading state and styles to table and global search 

brunolipe-a:
Now the tables has loading state with skeleton: 
ezgif-1-9620ef9124
ezgif-1-ef829a068f

And global search has a loading icon:
ezgif-1-d01120fa0b


brunolipe-a:
I forgot that... think now is right. I added to config all the targets to don't repeat in the code, but don't know if is the best solution.

danharrin:
Updated the table loading design so that the content shift for individual columns isn't presen
Screenshot 2022-05-22 at 12 53 48
t

danharrin:
I've also added a delay to every loading indicator to prevent <200ms flashes

danharrin:
Thanks!

PR #2498, opened 2022-05-22
Feature: Allow user timezone to be different to System time for display

phpsa:
Adds functionality to allow all Date pickers and table columns to use the users timezone for date,

This way Database can be kept in UTC and each user can view the time in their timezone,

Changes to note:
1) New Middleware to set the users timezone
2) new config variable to default a timezone for users

Documentation updates:
User model can now have a new attribute called timezone-- this can be either a column or use the laravel get attribute method: `function getTimezoneAttribute()` or `function timezone(): Attribute`


danharrin:
I like this, but instead of the middleware, we could use an approach like we do for avatars - `HasTimezone` interface for the User model, and then a `Filament::getUserTimezone()` method to fetch the timezone from the user. Then, Filament users can customise the timezone attribute if they wish :)

danharrin:
Apologies, I have refactored this PR and removed quite alot of it. While the idea is great, setting a user timezone in the admin panel and then accessing it from the forms and tables packages will not work, since not everyone has the admin panel installed. I would suggest that you pass in the user timezone to the new DateTimePicker `timezone()` method, and column's existing `dateTime()` method.

You can even do this automatically for all DateTimePickers in a service provider:

```php
DateTimePicker::configureUsing(fn (DateTimePicker $component) => $component->timezone(auth()->user()->timezone))
```

PR #2501, opened 2022-05-22
fix: allow collapse by default on repeater

phpsa:
This PR allows you to set `->collapse()` on a repeater field to automatically collapse all repeaters on page load.


danharrin:
Thanks!

PR #2518, opened 2022-05-23
fix: adding $set target to loading in table

brunolipe-a:
When we have a form or a select with `->searcheable()` in filters it don't start the loading state on the table.

### DIFF
--- a/packages/tables/src/Table.php
+++ b/packages/tables/src/Table.php
@@ -59,7 +59,7 @@ class Table extends ViewComponent
 
     protected string $viewIdentifier = 'table';
 
-    public const LOADING_TARGETS = ['gotoPage', 'tableFilters', 'resetTableFiltersForm', 'tableSearchQuery', 'tableRecordsPerPage'];
+    public const LOADING_TARGETS = ['gotoPage', 'tableFilters', 'resetTableFiltersForm', 'tableSearchQuery', 'tableRecordsPerPage', '$set'];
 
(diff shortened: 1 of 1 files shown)


danharrin:
Could you explain this a little more? What is being `$set`? If it's a specific property, we can use that as the target instead. Otherwise, all `$set` calls will trigger a loading state.

danharrin:
Could you explain this a little more? What is being `$set`? If it's a specific property, we can use that as the target instead. Otherwise, all `$set` calls will trigger a loading state.

brunolipe-a:
In filters, when we use a `Filter::make()->form()` it use the `$set` method to change the values instead of `wire:model`. So, to get this event and trigger the loading state, we have to add the `$set` to the LOADING_TARGETS. I tested with a form builder outside the table that calls the `$set` method, and it do not trigger the loading inside table. I think that the Table Component kind of create a context and only observes the targets inside of it self.

I don't know how to target the specific property of the `$set` method when it has different parameters... 

An example: 


https://user-images.githubusercontent.com/49805116/170047234-ccc7f6b9-27eb-4be6-b2c9-707e4b4b5180.mp4

pxlrbt:
Unfortunately this triggers loading indicator when closing a modal that doesn't update anything.
Actions already have a loading indicator on their own.

https://user-images.githubusercontent.com/22632550/203955575-6b866969-a0de-40ef-b330-79addf535218.mov

PR #2521, opened 2022-05-24
Make sidebar group collapsing configurable

johncarter-:
Switch off sidebar group collapse globally with config option.


danharrin:
Thanks John :)

PR #2524, opened 2022-05-24
fix: syntax error on table search query

FurkanGM:
Mariadb doesn't support ``->`` operator on queries. I'm changed ``->`` operator to ``json_extract`` function.

I'm tested Mariadb 10.x and MySQL 5.x versions and working correctly


danharrin:
Thanks

PR #2534, opened 2022-05-24
Use brand config for base layout title

ajnsn:
Unifies the appearance with `brand.blade.php` and `brand-icon.blade.php`


danharrin:
Great, thanks

PR #2535, opened 2022-05-24
Fix to match Spatie Media Library default behaviour when media name is not specified.

### LINKED ISSUE #2523: If you do not call ->mediaName(...) when specifying spatie media library field, the name column in media table will always be EMPTY
### Package

filament/spatie-laravel-media-library-plugin

### Package Version

2.11

### Laravel Version

9.13

### Livewire Version

2.10.5

### PHP Version

8.1.6

### Bug description

If you do not call ->mediaName(...) when specifying spatie media library field, the name column in media table will always be EMPTY

This shouldn't be the case because when I just use spatie media library normally, if I didn't set the name, the name will automatically be the file name (with .extension removed)

Another side question i like to ask is. if I want to modify the media name, what parameter does the closure take?

In the docs I see no mention of mediaName(). I only knew about the existence of it in the source code. But it is not obvious what arguments are passed into the closure if I want to have more control over the naming. I tried TemporaryFileUpload but it did not work.

```php
->mediaName(function (TemporaryUploadedFile $file): string {
// Exception
                            return pathinfo($file->getClientOriginalName(), PATHINFO_FILENAME);
                        })
```

https://filamentphp.com/docs/2.x/spatie-laravel-media-library-plugin/form-components

### Steps to reproduce

Just don't call ->mediaName(...) anywhere

```php
SpatieMediaLibraryFileUpload::make('photo')
                        ->required()
                        ->collection('profile-photos'),
```

### Relevant log output

_No response_

ziming:
In Spatie Media Library

if you never call ->usingName() while adding to the collection

the name column will by default be the file name without the extension. That is 'abc.pdf' will be 'abc' in the name column of media table. See Line 103 of this file

https://github.com/spatie/laravel-medialibrary/blob/f98d24ac31fcd63e04e9f44389695ba4fa6eaa87/src/MediaCollections/FileAdder.php#L103

However in filament SpatieMediaLibraryFileUpload field, if ->mediaName('value') is not called, it pass in '' to ->usingName() which is almost certainly not wanted.

This PR preserves the original name spatie media library will use.

### DIFF
--- a/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
+++ b/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
@@ -94,7 +94,7 @@ class SpatieMediaLibraryFileUpload extends FileUpload
 
             $media = $mediaAdder
                 ->usingFileName($filename)
-                ->usingName($component->getMediaName() ?? '')
+                ->usingName($component->getMediaName() ?? pathinfo($file->getClientOriginalName(), PATHINFO_FILENAME))
                 ->withCustomProperties($component->getCustomProperties())
                 ->toMediaCollection($component->getCollection(), $component->getDiskName());
 
--- a/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
+++ b/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
@@ -92,15 +92,15 @@ class SpatieMediaLibraryFileUpload extends FileUpload
 
             $filename = $component->getUploadedFileNameForStorage($file);
             
-            $media = $mediaAdder
+            $mediaAdder
                 ->usingFileName($filename)
                 ->withCustomProperties($component->getCustomProperties());
 
             if (filled($mediaName = $component->getMediaName())) {
-                $media->usingName($mediaName);
+                $mediaAdder->usingName($mediaName);
             }
             
-            $media->toMediaCollection($component->getCollection(), $component->getDiskName());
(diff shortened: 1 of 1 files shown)


danharrin:
Instead, why dont we just conditionally call `usingName()` ?

ziming:
Optionally calling it look good to me too

danharrin:
Okay, lets do that instead

ziming:
Do u edit it or do I edit it?

ziming:
Okay, I changed it @danharrin

ziming:
Hey @danharrin 

It turns out livewire (or is it filament?) actually edits the name value before handling it off to spatie media library. This result in the name column end up as having a value like "media-library41zWjf" 

So while this is still better than the previous state of passing in empty string.

my original solution is actually the more correct one. pathinfo($file->getClientOriginalName(), PATHINFO_FILENAME);

So now I can make a new PR with my original solution. That is pathinfo($file->getClientOriginalName(), PATHINFO_FILENAME)

or we add a new method call preserveOriginalMediaName() as the matching counterpart to preserveFilenames()

Which do you think is better?

Apologies

danharrin:
Hm alright, PR your original solution, this can be customised using `mediaName()` anyway.

ziming:
Ok will PR my original solution.

Yes it can be customized with mediaName($state)

It's a bit weird to me however that the $state variable is an array instead of TemporaryFileUpload instance

It look something like this

["d29f66ef-a1c4-4139-a6ba-5ab8c48bda18" => Livewire\TemporaryUploadedFile]

but changing it might break backward compatibility.

danharrin:
`$state` is an array because we also support multi-uploads. We can pass in `$file` to closures maybe, which is the current file object?

danharrin:
See #2577, you can now use `$file` instead of `$state`

PR #2536, opened 2022-05-24
Added an alpine magic function $pluralize(string, number, replace) for pluralization support.

lukas-frey:
I replicated Laravels `trans_choice()` function in an alpine magic function to be able to pluralize strings from JavaScript.

Everything that trans_choice supports is implemented, including:

- Singular form
- Plural form
- Plural conditions
      - {x} for one specific number, such as {0}
      - [x,y] for a range, such as [2,4]
      - [x,\*] and [\*,y] for a range from x to "unlimited" or from "unlimited" to y.
- Replaces variables just like laravel, ie:
      - :var - replaces all occurences of :var 
      - :VAR - replaces all occurrences and transform to upper case.
      - :Var - replaces all occurences and changes starting character to uppercase.

There is only one difference between the Laravel implementation and this one, which is in [MessageSelector.php#getPluralIndex()](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Translation/MessageSelector.php#L108) in Laravels source. I didn't know whether it is a good idea to reimplement this whole function and add 1000 lines of code just to replicate it.

This function enables shorter syntax, so for the czech language for example, instead of writing
```{1} 1 pivo|[2,4] :count piva|{5,*} :count piv.```
you could simply write
```1 pivo| :count piva| :count piv.```

because the function specifies the syntax for every language already.

Some languages have an unreal complex syntax though, so there it might make sense to implement this function also, for example [here](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Translation/MessageSelector.php#L352).

If you think it's a good idea to implement this class too, to fully support everything trans_choice() does, I will implement it. :)

### CODE REVIEW

**Comment by LukasFreyCZ** [comment]
https://user-images.githubusercontent.com/10926334/170105848-dd47de38-1d6f-4993-ba94-581fccc99a1c.mov

Screen recording of what it does. :)

**Comment by ryangjchandler** [comment]
I don't think `$pluralize` is the best name for this helper, given that it's a translation of Laravel's own `trans_choice()` helper. Perhaps we should just call it `$transChoice` instead?

File: packages/tables/resources/views/components/selection-indicator.blade.php [hunk 1]
```diff
@@ -12,14 +12,7 @@
...
 
-            <span @class(['dark:text-white' => config('tables.dark_mode')]) x-text="
-                singularText = '{{ trans_choice('tables::table.selection_indicator.selected_count', 1, ['count' => 1]) }}'
-                pluralText = '{{ trans_choice('tables::table.selection_indicator.selected_count', 2, ['count' => 2]) }}'
-
-                return (selectedRecords.length === 1) ?
-                    singularText.replace('1', selectedRecords.length) :
-                    pluralText.replace('2', selectedRecords.length)
-            "></span>
+            <span @class(['dark:text-white' => config('tables.dark_mode')]) x-text="$pluralize('{{__('tables::table.selection_indicator.selected_count')}}', selectedRecords.length, {'count': selectedRecords.length})"></span>
```
ryangjchandler, line 15:
```suggestion
            <span @class(['dark:text-white' => config('tables.dark_mode')]) x-text="$pluralize(@js(__('tables::table.selection_indicator.selected_count')), selectedRecords.length, {'count': selectedRecords.length})"></span>
```

ryangjchandler, line 15:
```suggestion
            <span @class(['dark:text-white' => config('tables.dark_mode')]) x-text="$pluralize('{{__('tables::table.selection_indicator.selected_count')}}', selectedRecords.length, { count: selectedRecords.length })"></span>
```

packages/tables/resources/views/index.blade.php
```diff
@@ -464,4 +464,96 @@ class="hidden"
...
+
+                    if (value !== null && value !== undefined) {
+                        return replace(value.trim(), variables);
+                    }
+
+                    segments = stripConditions(segments);
+
+                    if (segments.length > 1) {
+                        return replace(number > 1 ? segments[1] : segments[0], variables);
+                    } else {
```
ryangjchandler, line 552: `else` is redundant given that both blocks return.

ryangjchandler (changes requested):

packages/tables/resources/views/index.blade.php
LukasFreyCZ: Removed else block in latest commit.

packages/tables/resources/views/components/selection-indicator.blade.php
LukasFreyCZ: Couldn't merge anymore since I merged the previous suggestion, but it is added in the latest commit now.

packages/tables/resources/views/index.blade.php
```diff
@@ -466,4 +466,92 @@ class="hidden"
...
+                    function replace(line, replace) {
+
+                        if (replace.length === 0) {
+                            return line;
+                        }
+
+                        const shouldReplace = {};
+
+                        for (let [key, value] of Object.entries(replace)) {
+                            shouldReplace[":" + ucfirst(key ?? "")] = ucfirst(value ?? "");
```
danharrin, line 526: Please replace all `"` with `'`

danharrin (changes requested):

packages/tables/resources/views/index.blade.php
```diff
@@ -466,4 +466,92 @@ class="hidden"
...
+         * https://github.com/laravel/framework/blob/5299c22321c0f1ea8ff770b84a6c6469c4d6edec/src/Illuminate/Translation/MessageSelector.php#L15
+         */
+        document.addEventListener('alpine:init', () => {
+            window.Alpine.magic('transChoice', () => {
+                return (text, number, variables) => {
+                    function extract(segments, number) {
+                        for (const part of segments) {
+                            const line = extractFromString(part, number);
+                            if (line !== null) {
+                                return line;
```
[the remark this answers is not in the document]
LukasFreyCZ: Done

packages/tables/resources/views/index.blade.php
```diff
@@ -466,4 +466,92 @@ class="hidden"
...
+                    }
+
+                    function extractFromString(part, number) {
+                        const matches = part.match(/^[\{\[]([^\[\]\{\}]*)[\}\]](.*)/s);
+
+                        if (matches === null || matches.length !== 3) {
+                            return null;
+                        }
+
+                        var condition = matches[1];
```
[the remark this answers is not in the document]
LukasFreyCZ: Done

packages/tables/resources/views/index.blade.php
```diff
@@ -466,4 +466,92 @@ class="hidden"
     </form>
 
     {{ $this->modal }}
+
+    <script>
+        /**
```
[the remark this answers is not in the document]
LukasFreyCZ: Changed it. :)

packages/tables/resources/views/index.blade.php
LukasFreyCZ: Done

danharrin:
Thanks!

PR #2537, opened 2022-05-24
fix: avoiding repeated queries on table

brunolipe-a:
When the table is loaded it query the count of items on table 3 times the same query:

Captura de tela de 2022-05-24 16-11-03

I made the CanSelectRecords verify if the pagination is enable to get the total count of items from the LengthAwarePaginator. With change:

Captura de tela de 2022-05-24 16-13-52


danharrin:
Thanks, but PhpStan is failing

danharrin:
Fixed, thanks

PR #2538, opened 2022-05-24
fix: removing repeated queries on edit action

brunolipe-a:
When finding the model from `$record` on edit action (ManageRecords) o table it was calling the same query many times.

Before:
Captura de tela de 2022-05-24 18-08-49

After:
image


danharrin:
I think this will break BelongsToManyRelationManagers, since these depend on certain model properties being hydrated based on the pivot record selected

Instead of using `$mountedTableActionRecord` to store the model instance, we need to create a `$cachedMountedTableActionRecord` **protected** property, and check if this is present or not before executing the query.

Because this property is protected, it will not be rehydrated using Livewire, so it will not cause issues with the BelongsToManyRM.

danharrin:
I fixed it, thanks

PR #2545, opened 2022-05-25
close modal in response to dynamic event fired when modal backdrop clicked or escape key is pressed

jojostx:
Currently trying to create a plugin for filament that makes use of modals;

I specifically needed to hook into the **`close-modal`** event to run some code when the modal is closed, this work fine when the modal is closed by clicking the button that dispatches the event responsible for closing the modal, but fails when the modal is closed by clicking the backdrop or pressing the escape key. 

I know this is a very niche use case but I would love to use the modal provided by Filament.

I've implemented a working fix for this and I would be glad if it is considered or improved.

Thank you and thanks for the awesome package.


danharrin:
Great, thanks!

PR #2555, opened 2022-05-25
feat: Option to disable outputting the default DM Sans font

realslimsutton:
Having the ability to optionally disable outputting the default DM Sans as the front-end assets, helps developers to optimise their panels when they're loading their own fonts. Otherwise, they'll be forced to load multiple fonts, even if they're using a custom theme (thus custom styles, and fonts).

This is especially important for dashboards that are publicly accessible for all users, rather than just administrators.

Notes:
- Config option is set to true by default to not change the current behaviour
- When setting to the config option to false, it will also remove the preconnect links for Google Fonts, but these can be added back in by the developer (if they wish to still use Google Fonts).

### DIFF
--- a/packages/admin/config/filament.php
+++ b/packages/admin/config/filament.php
@@ -259,7 +259,7 @@ return [
     |
     */
 
-    'output_default_font' => env('FILAMENT_DEFAULT_FONT', false),
+    'output_default_font' => env('FILAMENT_DEFAULT_FONT', true),
 
     /*
     |--------------------------------------------------------------------------
--- a/packages/admin/config/filament.php
+++ b/packages/admin/config/filament.php
@@ -251,6 +251,18 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Output Default Font
+    |--------------------------------------------------------------------------
+    |
+    | Output the default `DM Sans` font to the front-end assets. You may disable
+    | this and load your own fonts.
+    |
+    */
+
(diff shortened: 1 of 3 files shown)


danharrin:
How about, instead we allow customisation of the Google fonts URL in config? Then you can load a different font

realslimsutton:
> How about, instead we allow customisation of the Google fonts URL in config? Then you can load a different font

I did actually consider that, but decided this route offered more flexibility by not locking people in on using Google Fonts.

I think both options are good and have their benefits/drawbacks, so I'm not particularly sure which would be overall best-fit here.

danharrin:
How about:

```php
'google_fonts' => '<url here>',
```

If you set it to `null`, Google fonts aren't loaded at all. But by default, this is set to the Google fonts URL that contains DM Sans.

realslimsutton:
> How about:
> 
> ```
> 'google_fonts' => '<url here>',
> ```
> 
> If you set it to `null`, Google fonts aren't loaded at all. But by default, this is set to the Google fonts URL that contains DM Sans.

Ooh I like that approach. I'll change it to that shortly.

Thanks!

realslimsutton:
> How about:
> 
> ```
> 'google_fonts' => '<url here>',
> ```
> 
> If you set it to `null`, Google fonts aren't loaded at all. But by default, this is set to the Google fonts URL that contains DM Sans.

That's been pushed. I'm definitely happy with your idea, I think that works nicely.

danharrin:
Made the changes and added docs. Thanks!

PR #2559, opened 2022-05-26
Fix Card $color style

wxfjamdc:


danharrin:
It's actually part of the design that the `$color` is not used for the main heading on each card. The color is used for the description, the icon and the chart in the background :)

Thanks for the PR though

PR #2564, opened 2022-05-26
feature: replace custom Markdown editor with EasyMDE

ryangjchandler:
This pull request aims to replace our custom Markdown editor with something a little more battle tested, EasyMDE.

The process here is quite simple, EasyMDE is more flexible than I had originally thought. At the time of writing this, the editor will do almost exactly the same thing as before (text formatting, insert image and link wrappers, preview markdown). Below is a list of things that still need to be done.

**TODO:**
* [x] Get EasyMDE setup
* [x] Make toolbar buttons do the same thing 
* [x] Get drag & drop image uploads working (take code from old component)
* [x] Get file picker image uploads working (use same code as above)
* [x] Fix preview pane not going full content height when toggled
* [x] Add heading button?
* [x] Dark mode
* [x] RTL

Quite a bit to do still, but it's functional right now.


packages/forms/resources/css/components/markdown-editor.css
```diff
@@ -0,0 +1,31 @@
...
+.EasyMDEContainer .editor-toolbar {
+    @apply hidden;
+}
+
+.EasyMDEContainer .CodeMirror {
+    @apply w-full h-full min-h-full transition duration-75 rounded-lg shadow-sm border border-gray-300 focus:border-primary-500 focus:ring-1 focus:ring-inset focus:ring-primary-600 caret-black !important;
+}
+
+.EasyMDEContainer .CodeMirror .CodeMirror-scroll {
+    @apply tracking-normal whitespace-pre-wrap font-mono text-sm !important;
```
danharrin, line 12: !important does not work with @apply in Sass, so these need to be normal CSS declarations :(

ralphjsmit:
Will this keep the same appearance as the custom Markdown editor?

ryangjchandler:
@ralphjsmit Yeah, all looks the same! There's a new header button in the toolbar, but otherwise it's the same 😊

packages/forms/resources/css/components/markdown-editor.css
ryangjchandler: Ah, that sucks. Important probably isn't even needed because of specificity.

themahdavi:
Nice 👌
I tested and two bug :
1- dark mode not work for textarea
2- for RTL need direction property on `new EasyMDE()`

ryangjchandler:
Thanks for testing! I'll sort those out today :)

ryangjchandler:
@themahdavi Are you able to test again - I've verified dark mode fixes but would like to check RTL support.

ziming:
The current markdown field look kind of ugly when ->disabled() is called on it. I hope this new one look nicer!

ziming:
I see. Can the Preview mode not have that weird circle? @ryangjchandler 

https://share.getcloudapp.com/jkuX1Rnk

Thank you!

ryangjchandler:
@ziming Yeah, I've fixed that in this pull request. I think that was a conflict between some padding classes.

themahdavi:
@ryangjchandler Dark mode and RTL work correctly but two thing need fix:
1- on darkmode preview show preview box color smaller 
image

2- on RTL preview all html items go left and text on right
image

themahdavi:
And it's Nice if showing (uploading image) text when image uploading progress...

ryangjchandler:
Hm, the RTL problem on preview is down to EasyMDE..

ryangjchandler:
@danharrin I think this is functionally ready. I might have to open an issue on the EasyMDE repo regarding the RTL problem with lists though.

themahdavi:
@ryangjchandler how for implement uploading state like github when uploading in progress?
is it possible with easymde ?

ryangjchandler:
Sorry missed your comment about that. Good idea. We can definitely do something similar to the file upload fields where we set the buttons to disabled with "file is uploading" text. We can also add a small status bar to the Markdown editor perhaps?

themahdavi:
@ryangjchandler yes status bar is good but i think have state on button or text like github is better

themahdavi:
And i think increase dependencies of project is maybe reduce control ...
smallest MarkdownEditor is perfect with simple features
and have seprate plugin is better
Of course this is my opinion

ryangjchandler:
@themahdavi Using an existing solution takes some of the maintenance pressure off of us, especially when it does things that our existing editor couldn't do / didn't do.

jeffgreco13:
@ryangjchandler how's this coming along? Just had a need to include headings. No pressure, but if this is imminent then I'll wait instead of using RichText :)

danharrin:
Sorry @jeffgreco13, theres no block AFAIK, I am just slow at reviewing

jeffgreco13:
@danharrin Hey no worries at all! Lots on your plate.

ryangjchandler:
Headings will work if you use the regular MD syntax for them!

ryangjchandler:
Needed a rebase - ready for review whenever :)

danharrin:
Just reviewed this. It's perfect, except when you change the state from the server.

For example, when using a simple resource / relation manager, and switching between records.

To make the editor contents update, you have to click on it, scroll, and then it glitches in.

PR #2566, opened 2022-05-26
Add support for overriding the maximum content width on a page

ralphjsmit:
This PR adds support for overriding the maximum content width on an individual page:

```php
protected static ?string $maxContentWidth = 'full';
```

Useful for example to limit the width of certain small "create"-pages, whilst keeping index pages full width.

### DIFF
--- a/packages/admin/docs/03-pages/01-getting-started.md
+++ b/packages/admin/docs/03-pages/01-getting-started.md
@@ -48,12 +48,6 @@ protected static ?string $navigationLabel = 'Custom Navigation Label';
 protected static ?string $slug = 'custom-url-slug';
 ```
 
-You may override the [maximum content width](https://filamentphp.com/docs/2.x/admin/appearance#changing-the-maximum-content-width) of a page using the `$maxContentWidth` property:
-
-```php
-protected static ?string $maxContentWidth = 'full';
-```
-
 You may also specify a custom header and footer view for any page. You may return them from the `getHeader()` and `getFooter()` methods:
 
 ```php
--- a/packages/admin/docs/03-pages/01-getting-started.md
+++ b/packages/admin/docs/03-pages/01-getting-started.md
@@ -48,6 +48,12 @@ protected static ?string $navigationLabel = 'Custom Navigation Label';
 protected static ?string $slug = 'custom-url-slug';
 ```
 
+You may override the [maximum content width](https://filamentphp.com/docs/2.x/admin/appearance#changing-the-maximum-content-width) of a page using the `$maxContentWidth` property:
+
+```php
+protected static ?string $maxContentWidth = 'full';
+```
+
 You may also specify a custom header and footer view for any page. You may return them from the `getHeader()` and `getFooter()` methods:
 
 ```php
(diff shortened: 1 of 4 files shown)


packages/admin/resources/views/components/layouts/app.blade.php
```diff
@@ -29,7 +33,7 @@ class="hidden transition-all w-screen space-y-6 flex-1 flex flex-col filament-ma
             ])>
                 <div @class([
                     'flex items-center w-full px-2 mx-auto sm:px-4 md:px-6 lg:px-8',
-                    match (config('filament.layout.max_content_width')) {
+                    match ($maxContentWidth ?: config('filament.layout.max_content_width')) {
```
danharrin, line 36:
```suggestion
                    match ($maxContentWidth ?? config('filament.layout.max_content_width')) {
```

packages/admin/resources/views/components/layouts/app.blade.php
```diff
@@ -64,7 +68,7 @@ class="hidden transition-all w-screen space-y-6 flex-1 flex flex-col filament-ma
 
             <div @class([
                 'flex-1 w-full px-4 mx-auto md:px-6 lg:px-8 filament-main-content',
-                match (config('filament.layout.max_content_width')) {
+                match ($maxContentWidth ?: config('filament.layout.max_content_width')) {
```
danharrin, line 71:
```suggestion
                match ($maxContentWidth ?? config('filament.layout.max_content_width')) {
```

packages/admin/src/Pages/Page.php
```diff
@@ -37,6 +37,8 @@ class Page extends Component implements Forms\Contracts\HasForms
 
     protected static string | array $middlewares = [];
 
+    protected static ?string $maxContentWidth = null;
```
danharrin, line 40:
```suggestion
    protected ?string $maxContentWidth = null;
```

packages/admin/src/Pages/Page.php
```diff
@@ -177,11 +179,17 @@ protected function getTitle(): string
             ->title();
     }
 
+    protected function getMaxContentWidth(): ?string
+    {
+        return static::$maxContentWidth;
```
danharrin, line 184:
```suggestion
        return $this->$maxContentWidth;
```

packages/admin/docs/03-pages/01-getting-started.md
```diff
@@ -48,6 +48,12 @@ protected static ?string $navigationLabel = 'Custom Navigation Label';
 protected static ?string $slug = 'custom-url-slug';
 ```
 
+You may override the [maximum content width](https://filamentphp.com/docs/2.x/admin/appearance#changing-the-maximum-content-width) of a page using the `$maxContentWidth` property:
```
danharrin, line 51: These docs probably belong with the other max width docs instead of here

ralphjsmit: What you think is better. My reasoning was that this is part of "customizing a page", so that's why I put it here.

danharrin:
Also, failing tests?

ralphjsmit:
> Also, failing tests?

Failing tests because of one of your suggested changes 😂 Is fixed👍

ralphjsmit:
For me it's ready.

PR #2567, opened 2022-05-26
Add support for setting the 'defaultTableRecordsPerPageSelectOption' via a config file

ralphjsmit:
Currently, the `$defaultTableRecordsPerPageSelectOption` is set to 10 records per page. In almost all circumstances I'd prefer to show more records per page, e.g. like 50 or so.

In Discord I proposed some time ago to allow setting this via a config-file. There's not a perfect solution without breaking changes, but there is a pretty good solution.

Personally I'd prefer to make the `$defaultTableRecordsPerPageSelectOption` `null` by default. However, that would change the typehint `int` to `?int` and thus result in a breaking change for all the people who manually set a `protected int $defaultTableRecordsPerPageSelectOption = 999`.

The best solution I could come up with was setting the default value to `0`, which is still an integer. Zero records per page would never make sense, so that's a pretty good way to distinguish between values set by the user and the default value.


ralphjsmit:
PS: is it an idea to also move the following values to a configuration option?

```php
protected function getTableRecordsPerPageSelectOptions(): array
{
    return [5, 10, 25, 50];
}
```

packages/admin/config/filament.php
```diff
@@ -211,6 +211,7 @@
             'actions' => [
                 'type' => Action::class,
             ],
+            'default_table_records_per_page_select_option' => 10,
```
danharrin, line 214: This belongs in the tables config, `tables.pagination.default_records_per_page`

packages/tables/src/Concerns/CanPaginateRecords.php
```diff
@@ -15,7 +15,7 @@ trait CanPaginateRecords
 
     public $tableRecordsPerPage;
 
-    protected int $defaultTableRecordsPerPageSelectOption = 10;
+    protected int $defaultTableRecordsPerPageSelectOption = 0;
```
danharrin, line 18:
```suggestion
    protected ?int $defaultTableRecordsPerPageSelectOption = null;
```

packages/tables/src/Concerns/CanPaginateRecords.php
```diff
@@ -50,7 +50,10 @@ protected function getTableRecordsPerPageSelectOptions(): array
 
     protected function getDefaultTableRecordsPerPageSelectOption(): int
     {
-        $perPage = session()->get($this->getTablePerPageSessionKey(), $this->defaultTableRecordsPerPageSelectOption);
+        $perPage = session()->get(
+            $this->getTablePerPageSessionKey(),
+            $this->defaultTableRecordsPerPageSelectOption ?: config('filament.layout.tables.default_table_records_per_page_select_option')
```
danharrin, line 55:
```suggestion
            $this->defaultTableRecordsPerPageSelectOption ?? config('filament.layout.tables.default_table_records_per_page_select_option')
```

packages/admin/config/filament.php
ralphjsmit: Done!

packages/tables/src/Concerns/CanPaginateRecords.php
ralphjsmit: Are you sure? This is a breaking change for people who did:

```php
protected int $defaultTableRecordsPerPageSelectOption = 50;
```

danharrin: Doesn't PHP allow this, since `int` is a subset of `?int` ? testing this would be great

ralphjsmit: Schermafbeelding 2022-05-26 om 22 12 31

ralphjsmit: Good idea!

PR #2572, opened 2022-05-27
fix: tables also use a user timezone setting same as forms

phpsa:
@danharrin -- Thanks for the form - your solution was more elegant :-)
I feel we should be able to do the same with table date columns so replicated it to the hasFormatters trait so that can be set in a service provider aswell.


danharrin:
Thanks :)

PR #2578, opened 2022-05-27
Fix redundant focus event on modals

shalior:
I noticed when I try to close a modal by clicking outside it triggers a focus event on modal elements which isn't desired IMO. I looked into it and figured this happens because of the alpine x-trap plugin and the presence of a button tag to close the overlay.

this PR fixes it, also I've added [innert](https://alpinejs.dev/plugins/focus#inert) and [noscroll](https://alpinejs.dev/plugins/focus#noscroll) modifiers to x-trap 

I tested the issue by registering event listeners in the console and clicking outside the modal window:
`document.getElementById('mountedTableActionData.THE_NAME').addEventListener('focus', ()=> console.log('focused'));
document.getElementById('mountedTableActionData.THE_NAME').addEventListener('click', ()=> console.log('clicked'));
`

### DIFF
--- a/packages/support/resources/views/components/modal/index.blade.php
+++ b/packages/support/resources/views/components/modal/index.blade.php
@@ -46,16 +46,15 @@
         x-cloak
         class="fixed inset-0 z-40 flex items-center min-h-screen p-4 overflow-y-auto transition"
     >
-        <button
+        <div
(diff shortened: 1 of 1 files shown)


danharrin:
Great, thanks

PR #2592, opened 2022-05-28
shortcut to set table filter layout in Resource class

wychoong:


danharrin (changes requested):

danharrin:
Thanks! Also added a `layout:` shortcut param for the `filters()` method

PR #2598, opened 2022-05-29
Added 500ms debounce to key value field inputs

alperenersoy:
@danharrin As we discussed in the discord channel key value field is now pretty hard to type when used with ->reactive(). Until we can debounce $entangle statement this should do the trick.


danharrin:
Thanks

PR #2603, opened 2022-05-29
Added ->download() option for file uploads.

alperenersoy:
It provides download functionality for preloaded files with FilePond's onactivatefile event. We mentioned previewing files in a new tab in the Discord channel, but downloading would be more generic for different file types and devices. I wouldn't say I liked downloading files by creating anchor tag but I couldn't find a better way. I am open to suggestions about that.

Here is how it works:

https://user-images.githubusercontent.com/83382417/170879439-f9edc0b8-1a83-4a7e-8252-28ad3015b6b4.mp4


ryangjchandler:
I believe there's a FilePond plugin that can do this as part of the FilePond UI.. think it's called "get file" or something.

alperenersoy:
> I believe there's a FilePond plugin that can do this as part of the FilePond UI.. think it's called "get file" or something.

As you discussed in #1375 that package is not well-maintained. Maybe we can use the same approach and prepend a download button before file name. This way works but does not seem the best way. Wdyt @danharrin?

alperenersoy:
For downloading process yes but it adds a download icon before the file name instead onactivatefile event.

ryangjchandler:
Why don't we roll our own plugin...

alperenersoy:
I updated it with the download button approach like the mentioned plugin and copied styling from it. But since avatars don't have names on them, it doesn't prepend if the field is an avatar. Couldn't test with s3 disk but [we give ](https://github.com/laravel-filament/filament/blob/c2251d3bb7d1ecc895125c086d36091ebf32e7b3/packages/forms/resources/js/components/file-upload.js#L211) uploadedFileUrl to the source property of files so it should work. I somehow didn't like this PR. If you feel the same way we can close it otherwise you can merge it. Thanks.

Here is how it looks now:
image

danharrin:
Great, renamed the method to `enableDownload()`, thanks :)

ziming:
Currently the download button is dangerously close to the delete button.

Wonder if there is a way to put 1 of them on the opposite end.

alperenersoy:
I think that's ok but wdyt @danharrin?

alperenersoy:
> Currently the download button is dangerously close to the delete button.
> 
> Wonder if there is a way to put 1 of them on the opposite end.

I think that's ok but wdyt @danharrin?

danharrin:
Personally I think it's fine, but maybe we could position it similar to the delete button, on the other side? If anyone wants to PR that I'd merge it, but I'm not working on it myself as I don't think it's very important.

ziming:
I just tried to delete a file some time ago, and the moment I click on the x button, it instantly delete the file without any warning confirmation prompt. Given how download and delete button are so close together, if we are not moving the delete button to the other end, could we at least add a confirmation dialog when the X button is pressed?

danharrin:
I don't think Filepond allows a delete confirmation, but feel free to send a PR that moves the download button 👍

PR #2607, opened 2022-05-29
fix: pass an array to the Spatie TagInput state like taginput

ReeceM:
Hello Filament team.

This fix makes the Spatie Tag plugin render like the normal tag input by passing an array like the normal `TagInput` does.

It seems that when passing a collection it isn't converting to an array. This was confirmed by going to the blade component and rendering the states length `<code x-html="state.length"></code>` this had 0 as the value when there were tags and with the old code. But adding the toArray to force that being passed down it actually gave a length and showed tags.

So this basically adds the one method call. I haven't noticed any adverse affects from doing this.

Reece


danharrin:
Thank you :)

PR #2629, opened 2022-06-01
Feature/make repeater and builder collapsible

bernhardh:
As discussed in discord...

I moved the `isCollapsible` logic from `Section` into the `CanBeCollapsed` trait and added the necessary `if`'s to the views of `Repeater` and `Builder`. 

For the docs, I also added (copy+paste from section) the documentation for `Repeater`. Not sure, if I should also add it to the 'Builder` block as well.


danharrin:
Thanks!

PR #2630, opened 2022-06-01
Only dispatch notifications via LW on Livewire requests

pxlrbt:
The issue with the notifications was, that every component would call the `dehydrate` hook on render and therefore trigger `NotificationManager::handleLivewireResponses()`. But we only want this to happen on LiveWire requests, as otherwise a rendering component will clear the notifications while it cannot dispatch browser events.

### DIFF
--- a/packages/admin/src/NotificationManager.php
+++ b/packages/admin/src/NotificationManager.php
@@ -2,8 +2,8 @@
 
 namespace Filament;
 
-use Filament\Pages\Page;
 use Livewire\Component;
+use Livewire\Livewire;
 use Livewire\Response;
 
 class NotificationManager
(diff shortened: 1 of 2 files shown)


packages/admin/src/NotificationManager.php
```diff
@@ -35,6 +35,9 @@ public function handleLivewireResponses(Component $component, Response $response
         session()->forget('notifications');
 
         if (count($notifications) > 0) {
+            if (! Livewire::isLivewireRequest()) {
```
danharrin, line 38: Why is the check twice? The method wont reach this part if the first check fails

pxlrbt: Yeah I was asking the same 😅 Probably Cmd + V at the wrong time and didn't notice.

danharrin:
Great, thank you

PR #2637, opened 2022-06-02
Prefix session keys

pxlrbt:
Session data may collide with other data otherwise.


danharrin:
Should we do the same for other session properties within Filament?

pxlrbt:
Yeah I think that would be good. Didn't know any others though 😅

danharrin:
Just search `session` in the codebase ig

pxlrbt:
Found the `per_page` and `toggled_columns` keys. I guess that's it.
They are more specific than `notifications`, but I guess it's good to prefix them all just in case.

Just prefixing will reset the current settings of users. Do you think that's okay? Or should we "migrate" the old to the new value?

PR #2644, opened 2022-06-02
Prebuilt CRUD table actions & soft delete filter

pxlrbt:
Introduce actions and filters for SoftDeletion functionality:

`Pages\Actions\ForceDeleteAction`
`Pages\Actions\RestoreAction`

`Tables\Actions\ForceDeleteAction`
`Tables\Actions\RestoreAction`

`Tables\Filters\SoftDeletesFilter`

## Todo

- [x] ForceDeleteBulkAction
- [x] RestoreBulkAction
- [x] Fix PhpStan
- [x] Add `@deprecated` notices everywhere
- [ ] Write docs
- [ ] Implement SoftDelete in Filament Demo


ryangjchandler:
I'll wait for you to sort the PHPStan errors out and then give this a review, the diff is a bit hard to look at with all the comments/notices.

pxlrbt:
@ryangjchandler PHPStan issues should be resolved now.

packages/admin/src/Resources/Pages/EditRecord.php
```diff
@@ -165,7 +165,7 @@ protected function getViewAction(): Action
 
     protected function getDeleteAction(): Action
     {
-        return DeleteAction::make('delete')
+        return DeleteAction::make()
             ->action(fn () => $this->delete());
```
pxlrbt, line 169: @danharrin I think it would be good if the action is also contained inside the action. That would clean up the pages even more, and make it easy to overwrite the action globally.

packages/admin/src/Resources/Pages/EditRecord.php
```diff
@@ -161,15 +165,8 @@ protected function getViewAction(): Action
 
     protected function getDeleteAction(): Action
```
pxlrbt, line 166: Since these are all single line functions then, we could deprecate them now and remove them with V3.

packages/support/src/Actions/Concerns/HasLifecycleHooks.php
```diff
@@ -0,0 +1,92 @@
...
+
+    protected ?Closure $after = null;
+
+    protected ?Closure $beforeFormFilled = null;
+
+    protected ?Closure $afterFormFilled = null;
+
+    protected ?Closure $beforeFormValidated = null;
+
+    protected ?Closure $afterFormValidated = null;
```
pxlrbt, line 19: I know these are "LifecycleHooks", but shouldn't these be move to `HasFormSchema` as there is no use for them without a form?

packages/admin/src/Resources/Pages/EditRecord.php
danharrin: Do you mean the `->action()` call? its unclear in GitHub which line you're referring to

packages/support/src/Actions/Concerns/HasLifecycleHooks.php
danharrin: not sure, I think I want to keep all hooks in the same place tbh

packages/admin/src/Resources/Pages/EditRecord.php
danharrin: yup 👍

packages/admin/src/Resources/Pages/EditRecord.php
pxlrbt: Yes the `->action()`. I guess this it because of compatibility with the tables package as we discussed on Discord?

danharrin: It's because people might already be overriding the `delete()` method or using hooks so we need to preserve that behaviour.

pxlrbt: I guess it would be better to use the `::configureUsing()` method and checking whether the class was extended with a `deleted()` method. See my original PR that was merged into this one.

pxlrbt: Shouldn't that be covered the way I did it? It looks a bit hackier, but would support old syntax only if overwritten, and use new default if not overwritten:

https://github.com/pxlrbt/filament/blob/f99f59bde307dbbb6670e3c5fc800a126a3815a7/packages/admin/src/Pages/Actions/DeleteAction.php#L65-L76

danharrin: The getDeleteAction() method needs to stay too as well, so I think this is cleaner.

pxlrbt: Alright.

ryangjchandler:
Replaced by #2706.

PR #2646, opened 2022-06-03
Refactor PageActions

pxlrbt:
As discussed with Dan on Discord this is a first POC how we could refactor the page actions into separate classes while keeping them compatible with V2. 

This could clean up Page classes a lot and also make Actions easier to modify globally by using Laravel's container, e.g.:

```php
$this->app->bind(
    \Filament\Pages\Actions\DeleteAction::class,
    \App\Filament\Actions\DeleteAction::class,
);
```

Page actions could then be defined and modified like this:

```php

protected function getActions(): array
{
    return [
        DeleteAction::make('delete')->color('primary'),
        EditAction::make('edit')->color('danger'),
    ];
}
```

### DIFF
--- a/packages/admin/src/Pages/Actions/DeleteAction.php
+++ b/packages/admin/src/Pages/Actions/DeleteAction.php
@@ -1,0 +1,76 @@
+<?php
+
+namespace Filament\Pages\Actions;
+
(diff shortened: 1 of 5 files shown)


danharrin:
Gonna merge this into #2644

PR #2652, opened 2022-06-03
Add GridRepeater component

flxsource:
With a `Repeater` component, the `columns` config is passed to child components. The `GridRepeater` intercepts the config and applies it to the layout of the `Repeater` child components.

Currently works by overriding the `getColumns` and `getColumnsConfig` methods from `HasColumns` to both return the column config despite not being an instance of `ComponentContainer`, and the clearing the `$columns` property once the config has been fetched (to prevent it being fetched by the children). The "feels" like a hack though, so would welcome a better method.

Further improvements could involve a `childColumns` method or similar, for column configs you _do_ want to pass onto children.

image
image


ryangjchandler:
@flxsource Would you be able to provide some screenshots in the PR description please?

danharrin:
As mentioned on Discord, it would be great to refactor this to `Repeater->grid(columns)`.

We can extract the grid part of `component-container.blade.php` into a Blade component in our /support package (x-filament::grid), then we can use those classes whereever we want by just passing in the columns config array to the component.

danharrin:
Refactored it. Thanks!

PR #2656, opened 2022-06-03
change calendar width for persian languade

ariaieboy:
in persian language (maybe it occur in other language's too based on their locale and the size of the days label) we have a overlap on label's of the days and make it unreadable :
before

for fixing this i made a simple change to the width of the calendar it will set the min of the width to 64 and will fit the width according to the content this is the result after this change for persian language:
after
it's may be a little bit ugly but atleast it's readabble

### DIFF
--- a/packages/forms/resources/views/components/date-time-picker.blade.php
+++ b/packages/forms/resources/views/components/date-time-picker.blade.php
@@ -93,7 +93,7 @@
                 @class([
                     'absolute z-10 my-1 bg-white border border-gray-300 rounded-lg shadow-md',
                     'dark:bg-gray-700 dark:border-gray-600' => config('forms.dark_mode'),
-                    'p-4 w-64' => $hasDate(),
+                    'p-4 min-w-64 w-fit' => $hasDate(),
                 ])
             >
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks

themahdavi:
If we can add jalali persian is very nice
Like this package: https://github.com/zoomit-org/Dayjs-Jalali-Plugin/

ariaieboy:
> 

already done it https://github.com/ariaieboy/filament-jalali-datetimepicker

asefsoft:
> in persian language (maybe it occur in other language's too based on their locale and the size of the days label) we have a overlap on label's of the days and make it unreadable : before
> 
> for fixing this i made a simple change to the width of the calendar it will set the min of the width to 64 and will fit the width according to the content this is the result after this change for persian language: after it's may be a little bit ugly but atleast it's readabble

Hi
Thanks but it's not fixed for me
and I should note that this class: `min-w-64` is not included in https://github.com/laravel-filament/filament/blob/2.x/packages/admin/dist/app.css
I also compiled assets myself but there is no such class

PR #2671, opened 2022-06-05
TableBuilder : styling, rtl and more

themahdavi:
some improve for TableBuilder :

- [x] Styling improve to same as KeyValue Field
- [x] Change delete button from X to trash same as KeyValue Field
- [x] Change row `overflow-hidden` to `overflow-x-auto` to best view if have more columns
- [x] RTL improvements
- [x] Add Row and Add Column button now have Dynamic Label
- [x] Add Row and Add Column button now translatable
- [ ]  Delete Row  buttons Header need have sr-only
- [ ] Move inline Alpine Logic to seprate JS file

### DIFF
--- a/packages/forms/resources/lang/en/components.php
+++ b/packages/forms/resources/lang/en/components.php
@@ -230,4 +230,22 @@ return [
 
     ],
 
+    'table_builder' => [
+        'buttons' => [
+
+            'row' => [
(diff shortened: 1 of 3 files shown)


ryangjchandler:
Thanks @themahdavi!

PR #2705, opened 2022-06-08
add pointer-events-none class to disabled button

wychoong:
this only add pointer-events-none to elements that already have cursor-not-allowed


flolanger:
@wychoong can you explain the reasoning behind this change? It does make some trouble with title tags. See this discussion https://github.com/filamentphp/filament/discussions/11636

PR #2713, opened 2022-06-09
Determine plural label based on locale

NathanaelGT:
Some languages do not have a "plural rule",
this PR improves how developers write code

before this PR:
```php
class ProductResource extends Resource
{
    protected static ?string $label = 'Produk';
    protected static ?string $pluralLabel = 'Produk';
    ...
}
```

after this PR:
```php
class ProductResource extends Resource
{
    protected static ?string $label = 'Produk';
    ...
}
```

those codes will produce the same output in a language that does not have a "plural rule",
in a language with "plural rule", everything will work as usually


pathros:
How do you determine whether a locale has plural or not? 
Several languages do have plural forms: English, Skandinavian, Slavian, Latin ...

NathanaelGT:
Laravel has a built-in [method](https://github.com/laravel/framework/blob/1c0b3d2814b4ea2d49414ed3070c150f5e46f8eb/src/Illuminate/Translation/MessageSelector.php#L108) to check the pluralization index.

This PR makes use of that method to determine whether a locale has a plural or not

danharrin:
Just got around to reviewing this. Thanks!

PR #2717, opened 2022-06-09
fix: datetime picker afterStateHydrated blank state

### LINKED ISSUE #2681: Hidden closure on grid makes values null even if hidden is false
### Package

filament/forms

### Package Version

v2.12.27

### Laravel Version

v9.16.0

### Livewire Version

v2.10.5

### PHP Version

8.1.2

### Bug description

I created a selectbox with two options. I have a grid that I hide when one of the options (multiple in this case) is not selected. When multiple is selected and I submit the form. The values of the fields are null. When I remove the hidden method it works fine.

### Steps to reproduce

           Select::make('type')
                    ->options([
                        'single' => __('labels.single_day'),
                        'multiple' => __('labels.multiple_days'),
                    ])
                    ->reactive(),
                Grid::make()
                    ->schema([
                        DatePicker::make('start_date')
                            ->label(__('labels.start_date'))
                            ->required(),

                        DatePicker::make('end_date')
                            ->label(__('labels.end_date'))
                            ->required(),
                    ])->hidden(function (\Closure $get) {
                                return $get('type') !== 'multiple';
                            })

### Relevant log output

_No response_

wychoong:
closes #2681


wychoong:
@danharrin  not sure which is better, the following line in `setUp` works too

```php
   $this->default(static fn (DateTimePicker $component) =>  null);
```
* similar to Select component

danharrin:
Neither fix is correct, but I pushed the right one :)

wychoong:
ouch 😅 was reluctant to pass `withHidden: true` as I'm not sure will it cause any unintended behaviour

PR #2720, opened 2022-06-10
Fix section component flicker

austincarpenter:
Before Alpine.js loads, a programmatically collapsed section component briefly shows its contents. I've added an `x-cloak` attribute via Blade to prevent this from happening.

Before:
section-flicker

Additionally, for the same reason, the disclosure indicator for a collapsed section is briefly rotated 180º. I've included the correct class via Blade, which on subsequent changes is handled by Alpine.

Before:
disclosure-indicator-flicker

(Apologies the GIFs are so short – each loop is a page refresh)


danharrin:
Perfect, thank you!

austincarpenter:
No worries!

PR #2722, opened 2022-06-10
Make excluded field list more specific

joseph-d:
Make the list of fields excluded from table column generation more specific so that custom field names such as "starts_at" or "ends_at" are generated if they exist in the schema but Laravel system fields still remain hidden.


danharrin:
Thanks

PR #2740, opened 2022-06-12
button component isDisabled getter

wychoong:
sorry that previous fix is incomplete, missed out during test


danharrin:
What is the issue?

wychoong:
in my blade file, i'm using this component as below
```php
@php
            $isDisabled = $isDisabled();
        @endphp
        <x-forms::button
            type="button"
            x-on:click="{{ 'state += 1' }}"
            disabled="{{ $isDisabled }}"
            color="{{ $getColor() }}"
            :attributes="$getExtraAttributeBag()"
        >
            {{ $getLabel() }}
        </x-forms::button>
```

so the disabled property can be changed based on condition, but the opacity is not removed even with getter

i tried to fix with
```js
x-bind:class="{ 'opacity-70': isUploadingFile || isDisabled(), ... }"
```
this works but there are a bunch of console error

danharrin:
Can you test my changes?

wychoong:
with my latest changes then it works well with the disabled state being toggled multiple times

i also tested with FileUpload field. There is a opacity problem with the following steps:
1) show button on visible but disabled
2) change button to enabled
*up to now things works fine
3) upload file
*here the button doesn't change opacity but is not clickable

not sure how to proceed with a fix

danharrin:
I don't understand why the disabled state needs to be in Alpine, it's purely affected by Blade and having it stored in two places is obviously causing issues, thats why I removed it

wychoong:
I'm so sorry Dan, while testing on first time something is not right, now when i test again it works perfectly fine. 🤦🏻‍♂️

danharrin:
Its ok! Glad it works :)

PR #2741, opened 2022-06-12
feat:mutate relationship data before create or save

wychoong:
this PR would allow modifying data before create or save
```php
Forms\Components\HasManyRepeater::make('repeater')
        ->mutateRelationshipDataBeforeCreate(function ($data): array {
            $data['user_id'] = auth()->id();
            return $data;
        })
```


danharrin:
Apologies for the delay on this, in #2788 I am deprecating HasManyRepeater and allowing users to just use `Repeater::make()->relationship()`. So once that is merged I will look at these changes

mohamedsabil83:
A simple reminder

MACscr:
Can we get this for Select and MultiSelect relationships too? Seems no way to mutate/dehydrate ->relationships() in those field types before save.

wychoong:
> Can we get this for Select and MultiSelect relationships too? Seems no way to mutate/dehydrate ->relationships() in those field types before save.

What is your use case? With `createOptionForm`?

PR #2742, opened 2022-06-12
Feature: Disable URL on table column

studiowizjo:
Sometimes it could be usefull if column doesn`t link to anything, eg: if you want to copy it or attach own action:

```php
TextColumn::make('is_paid')
    ->disableUrl()
    ->extraAttributes(function (Invoice $record) {
        return [
            'wire:click' => 'mountTableAction("settle", ' . $record->getKey() . ')',
            'class' => 'cursor-pointer',
        ];
    })
```

This change allows to do it 2 ways by: `url(false)` or `disableUrl()`. However only the second method is documented as I think it is more transparent.


ryangjchandler:
@studiowizjo Pretty sure calling `->action()` or `url()` on a table will already do this, no?

ryangjchandler:
Also, you've deleted a Blade file?

studiowizjo:
@[ryangjchandler](https://github.com/ryangjchandler) I think using `action()` it is not possible to call one of defined table actions. At least I couldn`t figure out how to do this.

The second use case is, as I mentioned, to allow easy selecting and copying data from table without accidentaly clicking link.

> Also, you've deleted a Blade file?

It shouldnt be part of commit, it is my oversight. I want to withdraw it, but this is probably not how it should be done.

PR #2744, opened 2022-06-13
Fix extra spacing above wizard

austincarpenter:
A tiny PR to remove the initial whitespace in the wizard component. The hidden input field is causing the parent's `space-y-*` utility to apply a top margin to the first visible element which is unneeded. Using `grid` and `gap-y-*` is preferable here because it ignores hidden first child elements.

Edit: After looking at Tailwinds's `space-*` utilities again, it correctly handles elements with an attribute of `hidden`, so alternatively that can be added to the `input` component, but I haven't seen that used in Filament (or anywhere really) so the first method may be the way to go.


themahdavi:
This change best for RTL languages too.

danharrin:
Thanks

PR #2749, opened 2022-06-13
fix(admin): remove duplicate display properties and use flex

MaximeWillinger:
This PR fixes an issue on the `app.blade.php` layout.
Two different CSS display properties were applied with `hidden` and `flex`. This was ambiguous.

Also, the custom `x-cloak` behaviour was implemented using `!block` instead of `!flex`.
Using `!flex` here allows to make custom pages's root element full height using `h-full`.

### DIFF
--- a/packages/admin/resources/views/components/layouts/app.blade.php
+++ b/packages/admin/resources/views/components/layouts/app.blade.php
@@ -18,7 +18,7 @@
         @if (config('filament.layout.sidebar.is_collapsible_on_desktop'))
             <div
                 x-data="{}"
-                class="transition-all w-screen space-y-6 flex-1 flex-col filament-main lg:pl-80 rtl:lg:pl-0 rtl:lg:pr-80"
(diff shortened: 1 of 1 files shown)


packages/admin/resources/views/components/layouts/app.blade.php
```diff
@@ -18,9 +18,9 @@ class="fixed inset-0 z-20 w-full h-full bg-gray-900/50 lg:hidden filament-sideba
         @if (config('filament.layout.sidebar.is_collapsible_on_desktop'))
             <div
                 x-data="{}"
-                class="hidden transition-all w-screen space-y-6 flex-1 flex flex-col filament-main lg:pl-80 rtl:lg:pl-0 rtl:lg:pr-80"
```
danharrin, line 21: The `hidden` class needs to remain on this line, it's not ambiguous as it's then being overridden by `!flex` to act like `x-cloak`

danharrin (changes requested):

packages/admin/resources/views/components/layouts/app.blade.php
MaximeWillinger: Done

PR #2761, opened 2022-06-14
Update Swedish translations

adevade:
All Swedish translations updated from the current English files.

- Added newer, missing translations.
- Tweaked existing translations to follow the same style throughout.
- Changed `Skapa ny => Skapa`. Equivalent would be `Create new => Create`.
- Changed `Ändra => Redigera`. Equivalent would be `Change => Edit`.
- Changed `Ta bort => Radera`. Equivalent would be `Remove => Delete`.

## Feedback wanted

I changed the use of `Poster => Rader`, which would be more like `Posts => Rows`.
I can't come up with a better translation for `Records`. Maybe `Poster` is better after all?


adevade:
I don't understand why the conflicts... 😕

danharrin:
Thank you!

PR #2765, opened 2022-06-14
Prepare Polish language for v2.13

kamilst96:
**Changes:**
- SoftDeletes & Wizard support.
- Changed "Ładowanie..." to "Wczytywanie...". Seems more appropriate.
- [?] Move from `action & action another` to `action & continue` for consecutive actions.
- [?] Made a few changes to action result messages.

@lolen will you have a moment to review? Let's keep it as draft until review or until v2.13.

### DIFF
--- a/packages/admin/docs/02-resources/06-relation-managers.md
+++ b/packages/admin/docs/02-resources/06-relation-managers.md
@@ -380,6 +380,7 @@ AttachAction::make()->preloadRecordSelect()
 When you attach record with the `Attach` button, you may wish to define a custom form to add pivot attributes to the relationship:
 
 ```php
+use Filament\Forms;
 use Filament\Tables\Actions\AttachAction;
 
 AttachAction::make()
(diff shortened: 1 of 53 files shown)


lolen:
I'm worried about is returns: 
- Rekord utworzony
- Rekord usunięty
- Rekord przywrócony

They only concern a single record. I don't know if multiple records can be selected. And this is universal:
- Utworzono
- Usunięto
- Przywrócono

> [?] Move from action & action another to action & continue for consecutive actions.
Another concern I have is that "Continue" may be difficult for the user to read. And in the original it is "Create & create another" and the user knows what the next action is. 

The rest without complaint. Great job.

kamilst96:
@lolen I was expecting push back on these, hence the question marks. Reverted both.

**Rekord utworzony** - I am not aware of being able to create more than one object at once. **Rekord usunięty** and **Rekord przywrócony** also have plural forms **Rekordy usunięte** and **Rekordy przywrócone**. It depends on whether **single** or **multiple** index is used. At least that's what I'm thinking.

kamilst96:
@danharrin I marked it by accident, but I think it's alright. @lolen ?

lolen (approved):
Good job 👍

PR #2769, opened 2022-06-14
Feature: Disable URL on table column

studiowizjo:
This is reopen of this: https://github.com/laravel-filament/filament/pull/2742 as I messed up with commits and couldn`t put it in order. Now things are cleaned up.

Sometimes it could be usefull if column doesn`t link to anything, eg: if you want to copy it or attach own action:

```php
TextColumn::make('is_paid')
    ->disableUrl()
    ->extraAttributes(function (Invoice $record) {
        return [
            'wire:click' => 'mountTableAction("settle", ' . $record->getKey() . ')',
            'class' => 'cursor-pointer',
        ];
    })
```

This change allows to do it 2 ways by: `url(false)` or `disableUrl()`. However only the second method is documented as I think it is more transparent.

### DIFF
--- a/packages/tables/docs/03-columns.md
+++ b/packages/tables/docs/03-columns.md
@@ -153,15 +153,6 @@ TextColumn::make('title')
     ->openUrlInNewTab()
 ```
 
-Alternatively, you can also disable URL entirely:
-
-```php
-use Filament\Tables\Columns\TextColumn;
-
-TextColumn::make('title')
-    ->disableUrl()
-```
-
 ### Setting a default value
 
 To set a default value for fields with a `null` state, you may use the `default()` method:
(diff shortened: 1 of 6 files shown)


danharrin:
I want to propose a slightly different syntax for this, `url(false)`. We need to allow the `$url` property to store `bool` items too.

studiowizjo:
It already works that way. `$url` can receive now bool values. I mentioned about that at the begining:

> This change allows to do it 2 ways by: url(false) or disableUrl(). 

`disableUrl()` is only syntax sugar:

```php
public function disableUrl(): static
{
    $this->url = false;

    return $this;
}
```

danharrin:
Silly me lol, sorry

studiowizjo:
Maybe if your concern is that using `disableUrl()` doesn`t encompasses also disabling action, it should be thrown exception when trying to use it both?

I thought about the name a lot and apart from what was mentioned, I have only one more reasonable proposal, although I am not sure if it is better than the previous one: `notClickable()` . As always, naming is the hardest thing...

danharrin:
I think `disableClick()` is our best bet

And its a dedicated feature, not relying on `url(false)`

We need a `isClickDisabled` property and method for it

PR #2774, opened 2022-06-15
Fix Spatie translatable plugin getRecordTitle() error

tanthammar:
Declaration of Filament\Resources\Pages\EditRecord\Concerns\Translatable::getRecordTitle(): **?string** must be compatible with Filament\Resources\Pages\EditRecord::getRecordTitle(): **string**

  Access level to Filament\Resources\Pages\EditRecord\Concerns\Translatable::getRecordTitle() **must be public** (as in class Filament\Resources\Pages\EditRecord)


danharrin:
Thanks, unsure why PHPStan missed this

PR #2776, opened 2022-06-15
Allow HTML for notification messages

pxlrbt:
This would allow us to put links or a bit of formatting inside the message. For example, a link to a downloaded file.

### DIFF
--- a/packages/admin/resources/views/components/notification.blade.php
+++ b/packages/admin/resources/views/components/notification.blade.php
@@ -44,7 +44,7 @@
 
         <div class="flex-1 space-y-1">
             <div class="flex items-center justify-between font-medium">
-                <p
+                <div
                     class="text-sm leading-6"
                     x-bind:class="{
                         'text-danger-900 @if (config('filament.dark_mode')) dark:text-danger-200 @endif': notification.status === 'danger',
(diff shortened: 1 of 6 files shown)


ryangjchandler:
If we do this, we should probably change the surrounding `<p>` tag to a `<div>`, so it's a little more semantic. Thoughts?

pxlrbt:
You are right. Totally missed that.

danharrin:
Is there any chance of an XSS from allowing HTML like this?

pxlrbt:
> Is there any chance of an XSS from allowing HTML like this?

Since the HTML comes from Filament or plugins I don't see much risk here. Only if stuff gets injectected into the session data, but I guess you have other issues then 😅

danharrin:
I'm thinking, if people put the name of the record in the notification from user input, then it could be vulnerable.

If we allow markdown instead with `Str::markdown()`, does that strip dangerous HTML for us?

ralphjsmit:
> If we allow markdown instead with `Str::markdown()`, does that strip dangerous HTML for us?

Yes, I just tested that earlier this week, whether it's save to `{!! Str::markdown($userInput) !!}`. The result was that things like `<iframes>` were stripped, but normal Markdown was rendered correctly.

danharrin:
Great, lets do that @pxlrbt @ryangjchandler ?

pxlrbt:
I wrapped the notification messages inside `Str::markdown()`. As a side effect the message is wrapped in <p>-Tags but I don't think that's an issue as we don't have any general styling.

PR #2784, opened 2022-06-15
fix: replace hardcoded page namespace with config value

### LINKED ISSUE #2494: Hard coded namespace on make commands.
### Package

filament/filament

### Package Version

v2.0

### Laravel Version

v9.0

### Livewire Version

v2.0

### PHP Version

PHP 8.0

### Bug description

namespace is hard coded on the commands, so if i change it in the config file will remain the same.

### Steps to reproduce

1. create a fresh laravel + filament project
2. publish filament config file.
3. change pages namespace
4. php artisan make:filament-page Example

the generated page will have the default namespace, App\\Filament\\Pages

### Relevant log output

_No response_

nerdroid23:
This PR intends to fix #2494. There are some additional things I'd like to discuss, when the user changes the default namespace for the pages in the config: 
1. should the view path also be updated to reflect the new namespace?
2. should the path for the page class also be updated to reflect the namespace?

or will those just be left for the user to decide?


danharrin:
Yes and yes, we should use `pages.path` as the path of the page class, and generate a view path from that as well?

Also, we probably need this fix for resources and widgets too, both also have config.

nerdroid23:
Should those changes be made in another PR or just add them to this one?

danharrin:
This one is fine :)

nerdroid23:
Hey @danharrin! I have been taking a stab at this and hit a wall. We said that when `pages.namespace` changes, the path for both the _page class_ and _page view_ should reflect the new `pages.namespace`.

However, you mentioned that:
> we should use `pages.path` as the path of the page class, and generate a view path from that as well?

Hence the wall I hit, should the `pages.namespace` or `pages.path` be prioritized when the stubs are being copied? If I am understanding correctly from what you've mentioned, it should be `pages.path`, _which I already have working_. But then again, this PR was meant to fix the `pages.namespace` being hardcoded, which would mean that it should be used to generate the paths.

Let me know what you think, please!

danharrin:
I think we should use `pages.path` to copy the page class into the Pages directory, but `pages.namespace` to generate the view path from. Because the `pages.path` includes the directory path to the Laravel project, so it's more difficult to extract.

Same can be done for resources and widgets.

danharrin:
Gonna close this for now, I think there's alot more complexity that needs to be considered. Thanks for the PR :)

PR #2793, opened 2022-06-16
feat: add the possibility to configure namespace for make page command

### LINKED ISSUE #2494: Hard coded namespace on make commands.

OfenHuryes:
BREAKING CHANGE: Change configuration file :
  pages->path becomes string instead app_path()
  resources->path becomes string instead app_path()
  possibility to use a ```$JOKER$``` for resources->path and namespace
  e.g :
```php
  resources' => [
          'namespace' => 'App\\Filament\\$RESOURCE$\\Resources',
          'path' => 'Filament/Resources/$RESOURCE$/Pages',
          'register' => [],
      ],
```
resolve #2494

### DIFF
--- a/packages/admin/src/Commands/Concerns/CanTransposeStringPath.php
+++ b/packages/admin/src/Commands/Concerns/CanTransposeStringPath.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Filament\Commands\Concerns;
+
+use Illuminate\Support\Str;
+
+trait CanTransposeStringPath
+{
+    protected function changeForwardSlashToBackSlashes(string $value): string
+    {
+        return (string)Str::of($value)
+            ->trim('/')
+            ->trim('\\')
+            ->trim(' ')
+            ->replace('/', '\\')
+            ->append('\\');
+    }
+}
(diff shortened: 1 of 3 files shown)


danharrin:
Thanks, but duplicate of #2784, which does not have breaking changes.

nicobernardet:
Alright but #2784 resolve only one problem. 
In future the same problem will be always present for the path, namespace,... 

 Thinking for the next major version

danharrin:
Honestly I am very confused about what this PR introduces and why we need a wildcard

nicobernardet:
It introduces a customization of the namespace and path.
Currently, the path is not customizable because it is hardcoded. This allows to give back this possibility via the config file. In particular for :

 ``filament\resources
\\{{$resource}}\\pages```

danharrin:
I'm not really interested in customising the path to files inside the resource, the original issue was filed because the path to the resources parent directory was not being observed

PR #2794, opened 2022-06-16
Allow mutate view record data before fill form

itsfaqih:
Apparently, the method to [customize the data before filling the form in view record](https://filamentphp.com/docs/2.x/admin/resources/viewing-records#customizing-data-before-filling-the-form) is not implemented yet. 

This pull request will fix that.


danharrin:
Thanks

PR #2804, opened 2022-06-17
fix: wrong key used in failure notification

jonquihote:
The css class used to display failure notification is `danger` instead of `failure`.

This PR fixes that wrong class/key use.


danharrin:
Thanks

Issue #2791, opened 2022-06-16
BulkAction Authorization return error type


massimomarazzi:
### Package

filament/tables

### Package Version

v2.13.14

### Laravel Version

8.83.16

### Livewire Version

v2.10.5

### PHP Version

php 8.1

### Bug description

Visible method called by BulkAction get error.
Visible methot called by Action work. 

### Steps to reproduce

BulkAction::make('delete')
                ->action(fn(Collection $records) => $records->each->delete())
                ->deselectRecordsAfterCompletion()
                ->color('danger')
                ->icon('heroicon-o-trash')
                ->visible(fn($records) => $this->allowDelete())

### Relevant log output

```shell
Filament\Tables\Actions\BulkAction::Filament\Tables\Actions\{closure}(): Return value must be of type Illuminate\Database\Eloquent\Collection, null returned (View: /shared/httpd/sitemanagerpro/vendor/filament/tables/resources/views/index.blade.php)
```

danharrin:
Please share the Flare exception URL

massimomarazzi:
Filament\Tables\Actions\BulkAction::Filament\Tables\Actions\{closure}
vendor/filament/tables/src/Actions/BulkAction.php:63

danharrin:
Click the "share" button on the Laravel exception page and send the link here please

massimomarazzi:
https://flareapp.io/share/oPRM19Qm

G3z:
this is what I found: when adding `->visible` the method `getRecords()` inside the `BulkAction` class returns `null` while return type of the closure inside `getDefaultEvaluationParameters()` is expected to be a `Collection`

danharrin:
Fixed by #2806

PR #2806, opened 2022-06-17
fix error with BulkActions

G3z:
fix for error #2791
```
 Filament\Tables\Actions\BulkAction::Filament\Tables\Actions\{closure}(): Return value must be of type Illuminate\Database\Eloquent\Collection, null returned
```

returning value from https://github.com/laravel-filament/filament/blob/a79660175a204d1188ccca59446f5fc8ce21a11a/packages/tables/src/Actions/Concerns/InteractsWithRecords.php#L64 can be null


danharrin:
thanks!

PR #2808, opened 2022-06-17
Feature: support icon-only buttons

ralphjsmit:
This PR allows to support buttons that only have a visible icon. This would make it easy to add an unobtrusive "back" button or something to the icons:

Schermafbeelding 2022-06-17 om 14 33 29

In principle, this should work on every action in the interface, because I added it to the `Filament\Support\Actions\Concerns\HasLabel`-trait.

Thanks!


ralphjsmit:
> I appreciate the PR, but why not ->iconButton()? I dont think I want two different icon button styles in Filament

Didn't know about `->iconButton()`, thanks for letting me know!

The difference is that the output looks like this, but for me at least that's fine: 

Schermafbeelding 2022-06-17 om 17 06 14

bernhardh:
@ralphjsmit 

I discussed this issue (the lack of icon only buttons, not links) with @danharrin and @awcodes in discord last week, since I need/want this feature. There is a clear difference in UI between having an link with an icon (using `iconButton()`) and a button with an icon.

 It would be great, if you can reopen your PR and just change the `disableLabel` to `disabledLabel` since there is already the same logic for form fields with the name `disabledLabel`.

If you don't have time, I will "steal" your PR and make it myself of course. But you already did 99% the hard work ;)

ralphjsmit:
@bernhardh Reopened👍

Are you sure that `disabledLabel()` is the correct name, since [the other file](https://github.com/filamentphp/filament/blob/2.x/packages/forms/src/Components/Concerns/HasLabel.php) also uses the original name in this pr, `disableLabel()`?

bernhardh:
Upps. Sorry. Was just coping it from discord. 

image

Haven't checked it in source :). So you are right, it is `disableLabel`. 

But maybe it would be a better idea, to move the existing `HasLabel` trait from forms to support package and use it in both packages. Right?

ralphjsmit:
Np, let's keep the name how it is. 

@danharrin Where do you prefer the trait?

ralphjsmit:
Allright, in that case it's ready.

danharrin:
Please remove the docs, I don't want people getting confused and not using `iconButton()`, since icon-only buttons like this are against our design system 😂

ralphjsmit:
Okay, removed it. I love rigid design systems 💫😝

bernhardh:
Imho, I think it looks better, then just simple icons in the following scenario, where just links look a bit "lost". Compare yourself

image

image

ralphjsmit:
I think that this looks cleaner as well. Imo the `->iconButton()` is perfect inside table rows and the `->disableLabel()` is perfect for real buttons.

danharrin:
> Btw (came up randomly), just curious, do you really think that this looks bad (as the reason for your comment)?

Yes, I think it looks quite bad, icon buttons arent really meant to be put in the header, just the table rows. For the header I would always choose a normal button, or an `ActionGroup` if there are too many

PR #2811, opened 2022-06-17
Added filepond  imageResizeMode properties

qalainau:
Added filepond  imageResizeMode properties.

https://pqina.nl/filepond/docs/api/plugins/image-resize/


danharrin:
Thanks!

PR #2813, opened 2022-06-17
Add page default actions

ralphjsmit:
This PR allows to add a default set of page actions to every page. I used it to add a back-button to every page in the application (some clients...😆).

```php
Page::defaultActions([
    Action::make('back')
        ->url(fn (): string => Url::previous())
        ->hidden(fn(Component $livewire) => $livewire instanceof ListRecords || Url::previous() === Url::current())
]);
```

PS: the sentence about the Livewire URLs package is totally optional, but I at least wanted to mention it in the PR for people who try to use the normal URL facade. Feel free to remove.

### DIFF
--- a/packages/admin/docs/03-pages/02-actions.md
+++ b/packages/admin/docs/03-pages/02-actions.md
@@ -195,3 +195,32 @@ Action::make('save')
     ->action(fn () => $this->save())
     ->keyBindings(['command+s', 'ctrl+s'])
 ```
+
+## Default actions
+
+You may add default actions to every page by calling the `Page::defaultActions()` method inside the `FilamentServing` event. 
+
+```php
+use Filament\Pages\Page;
+
+Page::defaultActions([
+    Action::make('back')
+        ->url(fn (): string => Url::previous())
+]);
+```
+
+Paired with the `visible()` and `hidden()` methods, this makes for a very powerful tool:
+
+```php
+use Livewire\Component;
+use Filament\Pages\Page;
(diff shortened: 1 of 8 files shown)


zepfietje:
What's the advantage of having a `Page::defaultActions()` method over creating your own `Page` class with customizations which you extend?

Depending on your use case, there's probably more you'd want to customize for your default page. This potentially leads to a lot of `defaultXxx()` methods, which to me is really just an anti-pattern. In my opinion, inheritance is the way to go. Look at the recent `Action` classes refactor for a similar case.

ralphjsmit:
> What's the advantage of having a `Page::defaultActions()` method over creating your own `Page` class with customizations which you extend?
> 
> Depending on your use case, there's probably more you'd want to customize for your default page. This potentially leads to a lot of `defaultXxx()` methods, which to me is really just an anti-pattern. In my opinion, inheritance is the way to go. Look at the recent `Action` classes refactor for a similar case.

There's no particular reason, except for that fact that I didn't think of that route yet. Certainly makes sense, so thanks for sharing!

What do you think @danharrin?

danharrin:
Yeah sorry, I think this is too niche to be a feature. You can make a custom action class for this, then just use it on every page. `getActions()` is exposed everywhere now, and you can even add your custom action to the stubs.

ralphjsmit:
No problem, next time will immediately start using custom base classes for such things, just like people usually do with a base Eloquent Model and a base Livewire component. That's an excellent route as well.

PR #2825, opened 2022-06-18
UI: Add borders to table popovers; Invert tabs dark mode; Fix DateTimePicker disabled state

### LINKED ISSUE #2818: Visibility: Add borders for table popovers; Buttons in dark mode blend with bg
### Package

filament/filament

### Package Version

v2.13.5

### Laravel Version

v9.17

### Livewire Version

v2.10.5

### PHP Version

PHP 8.1.7

### Bug description

First, the filter popover blends into table header. Adding `border border-gray-300 dark:border-gray-700` and changing `mt-2` to `mt-4` in `filters/popover.blade.php` made it much more distinguishable.

I also changed dark background to not blend with buttons `dark:bg-gray-800`.

A second issue, with identical nature exists in tabs. I failed to hack it, because I didn't dig enough to find it yet, but the point is eliminating the buttons blending into background. It looks good on light, but not on dark. In this case, background could be darker, and inactive tabs brighter. The current feels like it's disabled.

Disabled DateTimePicker has focus outline.

Apologies for compound issue, but I wanted to avoid opening a few separate ones. I am not confident with theming, so I avoided opening PR. Can you intervene in PRs, when someone lacks skills to finish the job?

### Steps to reproduce

- Open table filter
- See DateTimePicker in View mode
- View tabs in dark mode

### Relevant log output

_No response_

kamilst96:
Closes #2818

- [x] **Filters:** add popover border; fix dark mode
- [x] **Toggleable:** add popover border; fix dark mode
- [x] **Tabs:** fix dark mode
- [x] **DateTimePicker:** fix disabled state
- [x] **RelationManager tabs:** add active tab dark mode variant
- [x] **Select:** Set `opacity-70` in disabled state

Currently, filters and toggleable popovers blend into table header. Adding borders around them and bigger margin-top looks better.

filters_light

toggleable_light

Dark mode in popovers and tabs is inconsistent with the rest of UI elements. Fixes as below.

filters_dark

toggleable_dark

tabs_dark

I failed to disable the outline in DateTimePicker view mode, as `disabled:` directives didn't apply. There's some logic here out of my bounds of understanding, sadly.

timedatepicker_outline

I also found this beauty, but I'm not sure what the expected design would be. Could you take a look?

relationmanager_dark


kamilst96:
I managed to fix the outline in **DateTimePicker**. At the same time, I stole `opacity-70` trick from **TextInput**, so it properly darkens when in disabled state. I noticed the same issue occurs in **Select** inputs, but I'm not sure how it relates to #2788.

Enabled:


Disabled:

kamilst96:
Relation manager tabs could be something like this:

kamilst96:
The `Select` somehow doesn't cooperate, completely ignoring me, even after removing all dark variants from **select.blade.php**. Could you take a look at this one?

danharrin:
I think we should use the RM tab style for form tabs as well, what do you think?

kamilst96:
Do you mean something like this? Seems good to me. Less to maintain if tabs are standardized. Not sure if left or centered. Either way, clean implementation is out of my league, sadly.



Will it include tab header? If so, I wonder if it would be possible to squeeze tab-specific actions into it, the same way it's done in RM. `Tab::schema()->headerActions()`.

PS: RM uses Livewire to switch tabs, and Tabs uses Alpine. Livewire remembers with GET, but Alpine doesn't remember which tab was open when moving from View to Edit, since it's out of its scope. Should I push an issue of this, or is it expected behavior?

danharrin:
I've fixed the opacity for disabled selects, let me know when you finish the tabs.

danharrin:
Thanks, merged the current set of changes. If you want to proceed with the tabs redesign, please open another PR.

kamilst96:
Thanks. Sorry about that. I was on sick leave for two weeks.

danharrin:
All good, hope you're feeling better 🙂

PR #2828, opened 2022-06-18
Fix: cannot edit pivot data in belongs to many relationships unless they allow duplicates

coleshirley:
see https://github.com/laravel-filament/filament/issues/2827


danharrin:
Thanks! Refactored it a little bit

coleshirley:
> Thanks! Refactored it a little bit

Nice! much cleaner than what I had

PR #2832, opened 2022-06-18
Add id column

zepfietje:
Often I need a column for the id of a model. Essentially it's just a text column, but with a few defaults:

- `id` as name
- `ID` as label
- Searchable; not too sure about this one, but personally I always want to be able to find models by their id
- Sortable
- Link to the resource view or edit page when available

I've talked to @dododedodonl, another Filament user, about this field and he was about to PR something similar. We've both been using a field like this for some time and would love to see it in the Filament core. So let me know what you think!

### DIFF
--- a/packages/tables/src/Columns/IdColumn.php
+++ b/packages/tables/src/Columns/IdColumn.php
@@ -20,8 +20,6 @@ class IdColumn extends TextColumn
             $this->label('ID');
         }
 
-        $this->searchable();
-
         $this->sortable();
 
         $this->url(function (Model $record): ?string {
--- a/packages/tables/src/Columns/IdColumn.php
+++ b/packages/tables/src/Columns/IdColumn.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace Filament\Tables\Columns;
+
+use Filament\Facades\Filament;
+use Illuminate\Database\Eloquent\Model;
+
+class IdColumn extends TextColumn
+{
+    public static function make(string $name = 'id'): static
+    {
+        return parent::make($name);
+    }
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        if ($this->getName() === 'id') {
+            $this->label('ID');
+        }
+
+        $this->searchable();
+
+        $this->sortable();
+
+        $this->url(function (Model $record): ?string {
+            $resource = Filament::getModelResource($record);
+
+            if ($resource === null) {
+                return null;
+            }
+
+            if ($resource::hasPage('view') && $resource::canView($record)) {
+                return $resource::getUrl('view', ['record' => $record]);
+            }
+
+            if ($resource::hasPage('edit') && $resource::canEdit($record)) {
+                return $resource::getUrl('edit', ['record' => $record]);
+            }
+
+            return null;
+        });
+    }
+}


ryangjchandler:
All columns will link to the resources edit page by default, fyi.

zepfietje:
> All columns will link to the resources edit page by default, fyi.

I know, but does it also link to the resource view page if one exists?

kamilst96:
@zepfietje yes, basically it's pseudocode of `hasViewPage ? viewUrl : editUrl`.

dododedodonl:
First thing I do when using Filament, is disable the row click, because this prevents copying text from a table. I use an IdColumn to replace this link (that defaults to the view page if it exists). It could also be a modifier on `TextColumn`, but I chose an IdColumn to reduce code as much as possible on a resource.

zepfietje:
> When searching within any numeric set of data (phone numbers, passports, etc), searching by ID is rendered useless, isn't it? And if you already know ID, you are already capable of constructing proper URL to the resource.

Then it still isn't useless, and you can't expect end users of your application to manually construct a URL. That said, I do understand this might differ heavily per project.

> Btw, they seem to be focusing on reducing classes and adding modifiers (see RelationManager, Repeater, Select), and you want a class just to use less modifiers, so this might get closed regardless.

I'm very well aware. However, this doesn't mean it's a bad idea to move regularly used code into a component.

If this PR ends up being closed regarded as too opinionated, I'll just package it up. But let's first discuss if there might be value in having this in core in some sort or form.

kamilst96:
So I am a weird one, then. I even force enable URL in RelationManager tables.

@zepfietje for search, I wonder if a filter "Show exact **match/matches** in **row**" would be safer then.

zepfietje:
> @zepfietje for search, I wonder if a filter "Show exact **match/matches** in **row**" would be safer then. Search might be too loose for such precise requirement.

Not sure if that would warrant a filter; it works fine in my case as is. But I'd already removed the default `searchable()` for this PR: [aae57e7](https://github.com/laravel-filament/filament/pull/2832/commits/aae57e73ee3cd4f5b171a3fdd3a5664e1d3bd18c).

ryangjchandler:
I don't have a problem with the overall idea of a dedicated ID column. I would definitely remove searchable and sortable by default though and make them opt-in/user controlled.

ryangjchandler:
I'd maybe also rename it to `ID`.. not sure though cause the other fields have PascalCase.

zepfietje:
> I'd maybe also rename it to `ID`.. not sure though cause the other fields have PascalCase.

Yeah that's actually what I had previously. PascalCase doesn't seem to have strict rules on uppercasing abbreviations though. I took a look at the Laravel codebase and there it's also not too consistent...

kamilst96:
I know you said you didn't want to make a filter, but I got curious anyway. You can type multiple IDs into the list, and it will spit all matches.

```php
Filter::make('matching_ids')
    return $query->when(
        $data['ids'],
        function (Builder $query, array $values) {
            $query->whereIn('id', $values);
        }
    );
})
->form([
    TagsInput::make('ids')
        ->label('IDs')
        ->placeholder('Type IDs...'),
]),
```

To be honest, I missed the part where I can make only one column clickable, so I'll try it out, thanks.

zepfietje:
> I don't have a problem with the overall idea of a dedicated ID column. I would definitely remove searchable and sortable by default though and make them opt-in/user controlled.

Hmm, yes it's hard to see when it gets too opinionated. I already removed searchable by default, and agree, maybe sortable should be left out as well.


But then there's not so much left of the field... Especially if the row link issue could be solved in a neat way.

kamilst96:
@zepfietje and automatic label wouldn't be language-universal. In my country its "Lp.".

zepfietje:
> I know you said you didn't want to make a filter...

Let's please stay on topic of this PR and not bloat the thread, @kamilst96.

zepfietje:
> @zepfietje and automatic label wouldn't be language-universal. In my country its "Lp.".

If that's the case, that means it's perfect to have a separate ID column, right? Then we could just add localization in the Filament language files.

packages/tables/src/Columns/IdColumn.php
```diff
@@ -0,0 +1,43 @@
...
+        parent::setUp();
+
+        if ($this->getName() === 'id') {
+            $this->label('ID');
+        }
+
+        $this->sortable();
+
+        $this->url(function (Model $record): ?string {
+            $resource = Filament::getModelResource($record);
```
danharrin, line 26: Not all table builder installs have the admin panel installed, so we can't do this check unfortunately

danharrin (changes requested):

packages/tables/src/Columns/IdColumn.php
zepfietje: Hmm fair, what would you suggest instead @danharrin?

ryangjchandler: We can wrap the entire url() call in a class_exists check if we wanted to.

danharrin: Hmm I don't like that, feels hacky

danharrin:
Thank you for the contribution, but I'm going to give this one a pass. The implementation feels very subjective, and it's not entirely clear how this would play with the tables package outside of the admin panel.

Since you are willing to distribute it as a plugin, that would be the best option IMO. For such a simple feature, a plugin would give you the flexibility to add whatever functionality you wished without having to have it approved and compromising with the Filament core team.

packages/tables/src/Columns/IdColumn.php
zepfietje: Maybe hacky in a way, but probably still the cleanest solution.

zepfietje:
@danharrin, I understand your reasoning. No problem for me to distribute it as a plugin. 👍🏻

However, maybe we could spend a minute looking at the bigger picture? As @dododedodonl explained above, the need for this PR comes from an annoyance with table rows being links by default:

> First thing I do when using Filament, is disable the row click, because this prevents copying text from a table. I use an IdColumn to replace this link (that defaults to the view page if it exists).

Aside from causing problems when copying text, the current linked-by-default behavior also makes tooltips useless on mobile. When tapping (hovering) the column, it will show the tooltip only very briefly before navigating to the resource view/edit page. I know it's probably hard to change the default at this point, but I think it's important for you to be aware of this problem and its usability impact in Filament. Let me know your thoughts!

ryangjchandler:
Can you not remove all URLs from columns by default using `Column::configureUsing()`?

```php
Column::configureUsing(function (Column $column) {
    $column->url(null);
});
```

zepfietje:
@ryangjchandler, nope, you can't. That does work for manually set URLs, but not for the `$recordUrl` set on the cell by Filament.

ryangjchandler:
We should probably support calling `->url()` with `false` then, like we do for other objects that can have URLs.

zepfietje:
Hmm, not sure if I like that solution. Especially since the column and record URL are two different concepts (the record URL is set on the cell just like the column URL, but that's just an implementation detail). Also it would be confusing what happens when passing `null` vs. `false` to `->url()`.

Currently, I just create my own `ListRecords` class with:
```php
protected function getTableRecordUrlUsing(): ?Closure
{
    return null;
}
```
Downside is you need to extend that class for all resources.

Not sure which classes support `configureUsing()`, but might that be useful here as well?

ryangjchandler:
Anything that extends ViewComponent currently supports configuration callbacks.

zepfietje:
> Anything that extends ViewComponent currently supports configuration callbacks.

Yeah so that doesn't work for the `ListRecords` class then unfortunately.

danharrin:
I do understand the frustration with the table record URL, so I think this should be customizable in the config, so you don't have to override every list page class. Also #2769 introduces disabling the URL for individual columns.

zepfietje:
> I do understand the frustration with the table record URL, so I think this should be a setting to disable in the config, so you don't have to override every list page class. Also #2769 introduces disabling the URL for individual columns.

Calling `disableUrl()` in `configureUsing()` would already work then. 👍🏻

I'm unsure when to create a config item for something... it could be nice for disabling the URL indeed, but once I want to customize the default pagination records per page setting, I'll end up with a custom `ListRecords` class anyway. We might want to think this through globally.

PR #2842, opened 2022-06-20
Use qualified key name for BelongsToMany relationship

pxlrbt:
I have an "ambiguous id" error (https://flareapp.io/share/DPygnXg5#F70) with BelongsToMany relationship that I cannot replicate inside the demo. But using the fully qualified key name instead of the key name fixes the issue. I only changed this on line but I guess we could use `getQualifiedKeyName()` instead of `getKeyName()` everywhere?

### DIFF
--- a/packages/tables/src/Concerns/HasRecords.php
+++ b/packages/tables/src/Concerns/HasRecords.php
@@ -65,7 +65,7 @@ trait HasRecords
         $relationship = $this->getRelationship();
 
         $pivotClass = $relationship->getPivotClass();
-        $pivotKeyName = app($pivotClass)->getKeyName();
+        $pivotKeyName = app($pivotClass)->getQualifiedKeyName();
 
         $query = $this->allowsDuplicates() ?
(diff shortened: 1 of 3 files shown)


danharrin:
Yea, changing it everywhere sounds right to me

packages/tables/src/Concerns/HasRecords.php
```diff
@@ -91,7 +91,7 @@ public function getTableRecordKey(Model $record): string
         $relationship = $this->getRelationship();
 
         $pivotClass = $relationship->getPivotClass();
-        $pivotKeyName = app($pivotClass)->getKeyName();
```
kamilst96, line 94: This kills my (BelongsToMany)RelationManager when `$allowsDuplicates = true;`.

pxlrbt: Thanks for the heads up. I'll create a PR to remove the breaking changes.

PR #2849, opened 2022-06-20
Feature: makes sidebar width configurable

awcodes:
This feature adds a config key to Filament Config file to support custom sidebar sizes set by the user.

Option can be set in any CSS unit.

```php
'sidebar' => [
    ...
    'width' => '320px'
],
```


packages/admin/config/filament.php
```diff
@@ -205,6 +205,7 @@
             'groups' => [
                 'are_collapsible' => true,
             ],
+            'width' => '320px'
```
danharrin, line 208: IMO the default should be null and set in the view instead, so we have the flexibility to customise it ourselves internally if we wish

danharrin (changes requested):

packages/admin/config/filament.php
awcodes: updated with latest commit

danharrin:
Needs to be recompiled, then we're good to go

awcodes:
Should be ok now.

danharrin:
Thanks!

PR #2863, opened 2022-06-21
Store filters in query string by default

zepfietje:
Currently, `ListRecords` table filters are not being stored in the query string by default, while sort and search are. As a user I'd expect all or none of them to be restored on a page reload, but not just some of them.

This has been discussed on Discord and @danharrin stated its a design choice as he's not 100% sure that all filters can be serialized to the query string yet.

We agreed it's probably better UX to store the filters in the query string by default. So I checked the filters for serialization issues and couldn't find any.

### DIFF
--- a/packages/admin/src/Resources/Pages/ListRecords.php
+++ b/packages/admin/src/Resources/Pages/ListRecords.php
@@ -22,6 +22,7 @@ class ListRecords extends Page implements Tables\Contracts\HasTable
     protected static string $view = 'filament::resources.pages.list-records';
 
     protected $queryString = [
+        'tableFilters',
         'tableSortColumn',
         'tableSortDirection',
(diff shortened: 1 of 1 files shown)


Cannonb4ll:
I was actually running into the same thing this morning, when going to another page, and going back, you'd lose the filters. It would be nice to have this inside the query so it's possible even to share links with others and filters enabled.

Cannonb4ll:
> I also want an option to store filters in the session as well, but thats for another PR

The downside of that though is that you can't share them with others, which I do regularly (or my customers do). Not sure if you were thinking for a complete replacement on that?

danharrin:
Nah, both. They would be stored in the query string by default, then you can enable session storage using `ListRecords::persistTableFilters()` or by overriding `areTableFiltersPersisted()` on an individual page

PR #2868, opened 2022-06-21
Add CSS Class to Filters

mustafa-online:
I wanna make some changes on the filters, 
But there is no unique class to target it. so I added this one.

### DIFF
--- a/packages/tables/resources/views/components/filters/index.blade.php
+++ b/packages/tables/resources/views/components/filters/index.blade.php
@@ -2,7 +2,7 @@
     'form'
 ])
 
-<div class="space-y-6 filament-tables-filters">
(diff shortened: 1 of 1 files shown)


packages/tables/resources/views/components/filters/index.blade.php
```diff
@@ -2,7 +2,7 @@
     'form'
 ])
 
-<div class="space-y-6">
+<div class="space-y-6 filament-tables-filters">
```
danharrin, line 5:
```suggestion
<div class="space-y-6 filament-tables-filters-form">
```

`filament-tables-filters` is already used for the main popover

danharrin:
Renamed to `filament-tables-filters-form`

PR #2871, opened 2022-06-21
refactor: replace `Str::title` with `Str::headline`

innocenzi:
This PR replaces occurrences of `Str::title` with `Str::headline` so `snake_cased` names are properly converted to `Title Case`.

image


ryangjchandler (approved):

danharrin:
Thanks, is that really the only difference between those methods lol

PR #2875, opened 2022-06-21
Draft: Allow multiple table sorts

bumbummen99:
This PR is intended to add the ability to use multiple defaultSort columns on Tables.

- [x] Basic change to data structure
- [ ] Actually test it


danharrin:
Unfortunately, there are many breaking changes in this PR

bumbummen99:
> Unfortunately, there are many breaking changes in this PR

Yeah i figured half way trought that this might not be the way to go with what i originally intended

PR #2879, opened 2022-06-22
feat: floating action groups

### LINKED ISSUE #2783: ActionGroup creates scrollable overflow and hides below table
### Package

filament/filament

### Package Version

v2.13.2

### Laravel Version

v9.17

### Livewire Version

v2.10.5

### PHP Version

PHP 8.1.7

### Bug description

Expanded ActionGroup is restricted to table container's bounds, so any potential overflow creates scrollbar and actions are pushed under it. Picture below:

1:
Zrzut ekranu z 2022-06-15 22-38-33

2:
Zrzut ekranu z 2022-06-15 22-30-52

### Steps to reproduce

- Create action group and add actions to it
- Propagate enough data for table to render
- Open ActionGroup in table's last row/record

### Relevant log output

_No response_

saade:
This PR fixes the action group behavior inside the table. As now, the table has a overflow-x to allow horizontal scroll, but then the action group modal "clips" inside de table container.

### Before this PR:
image

### After this PR:
image

As discussed on [discord](https://discord.com/channels/883083792112300104/883085291722788964/988356164985159710) Dan suggested Floating UI as an alternative to the Popper.js lib, so this PR uses Floating UI to position the dropdown near its trigger button, and Alpine's [x-teleport](https://alpinejs.dev/directives/teleport) to teleport the dropdown outside the table to prevent the overflow clipping.

@danharrin please see my comments down below.

Alpine.js docs states that it wont work with Livewire (because of the teleport to outside the Livewire component). But i've found a way around by making the Alpine.js do the call to Livewire using $wire.

### DIFF
--- a/bin/build.js
+++ b/bin/build.js
@@ -12,3 +12,15 @@ esbuild.build({
     mainFields: ['module', 'main'],
     watch: shouldWatch,
 }).catch(() => process.exit(1))
+
+esbuild.build({
+    define: {
+        'process.env.NODE_ENV': shouldWatch ? `'production'` : `'development'`,
+    },
+    entryPoints: ['packages/support/resources/js/index.js'],
+    outfile: 'packages/tables/dist/module.esm.js',
+    bundle: true,
+    platform: 'neutral',
+    mainFields: ['module', 'main'],
+    watch: shouldWatch,
+}).catch(() => process.exit(1))
--- a/packages/support/resources/js/components/action-group.js
+++ b/packages/support/resources/js/components/action-group.js
@@ -1,0 +1,55 @@
+import {computePosition, flip, shift, autoUpdate} from '@floating-ui/dom'
+
+export default (Alpine) => {
+    Alpine.data('actionGroupTableComponent', ({
+        mountTableActionUsing
+    }) => {
+        return {
+            isOpen: false,
+
+            mountTableActionUsing,
+
+            cancelAutoUpdate: function () {},
+
+            init: function () {
+                //
+            },
+
+            update: function () {
+                const trigger = this.$refs.trigger
+                const dropdown = this.$refs.dropdown
+
+                this.cancelAutoUpdate = autoUpdate(trigger, dropdown, () => {
+                    computePosition( trigger, dropdown, {
+                        placement: 'bottom-end',
+                        middleware: [
+                            flip(),
+                            shift({ padding: 5 })
+                        ],
+                    }).then(({x, y}) => {
+                        Object.assign(dropdown.style, {
+                            left: `${x}px`,
+                            top: `${y}px`,
+                        })
+                    })
+                })
+            },
+
+            toggleDropdown: function () {
+                this.isOpen = ! this.isOpen
+
+                if( this.isOpen ){
+                    this.update()
+                } else {
+                    this.cancelAutoUpdate()
+                }
+            },
+
+            wireClickAction: function (actionName, tableRecordKey) {
+                if(! actionName) return
+
+                this.mountTableActionUsing(actionName, tableRecordKey)
+            }
+        }
+    })
+}
--- a/packages/support/resources/views/components/actions/group.blade.php
+++ b/packages/support/resources/views/components/actions/group.blade.php
@@ -7,9 +7,17 @@
     'tooltip' => null,
 ])
 
-<div x-data="{ isOpen: false }" {{ $attributes->class(['relative']) }}>
+<div
+    x-data="actionGroupTableComponent({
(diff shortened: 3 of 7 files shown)


packages/support/resources/js/components/action-group.js
```diff
@@ -0,0 +1,55 @@
...
+                this.isOpen = ! this.isOpen
+
+                if( this.isOpen ){
+                    this.update()
+                } else {
+                    this.cancelAutoUpdate()
+                }
+            },
+
+            wireClickAction: function (actionName, tableRecordKey) {
```
saade, line 48: This is the function that replaced `wire:click="mountTableAction"`.

We need the alpine component to call $wire, since when the dropdown is teleported outside de Livewire component, it loses the Livewire reference to make the call using wire:click

packages/support/resources/views/components/actions/group.blade.php
```diff
@@ -7,9 +7,17 @@
     'tooltip' => null,
 ])
 
-<div x-data="{ isOpen: false }" {{ $attributes->class(['relative']) }}>
+<div
+    x-data="actionGroupTableComponent({
+        mountTableActionUsing: async (...args) => {
```
saade, line 12: Using the same approach as the other components. We need this here because of `$wire`

packages/tables/resources/views/components/actions/action.blade.php
```diff
@@ -19,14 +19,14 @@
...
     :tag="$action->getUrl() ? 'a' : 'button'"
-    :wire:click="$action->isEnabled() ? $wireClickAction : null"
     :href="$action->isEnabled() ? $action->getUrl() : null"
     :target="$action->shouldOpenUrlInNewTab() ? '_blank' : null"
     :disabled="$action->isDisabled()"
     :color="$action->getColor()"
     :tooltip="$action->getTooltip()"
     :icon="$icon ?? $action->getIcon()"
     :size="$action->getSize() ?? 'sm'"
+    x-on:click="{!! $action->isEnabled() ? $wireClickAction : null !!}"
```
saade, line 29: replaced `wire:click` with `x-on:click` because these actions will be teleported outside the Livewire component using Alpine `x-teleport`. So wire:click will not be available anymore.

Now the alpine component calls the `mountTableAction` function.

saade (commented):
Some explanations.

bin/build.js
```diff
@@ -12,3 +12,15 @@ esbuild.build({
     mainFields: ['module', 'main'],
     watch: shouldWatch,
 }).catch(() => process.exit(1))
+
+esbuild.build({
```
saade, line 16: Not sure about this tho. It works for me, but idk if thats the way to go.

I'll leave the build process up to you 😛

ryangjchandler:
Does `x-teleport` work with Livewire now? I know there were issues at first.

saade:
> Does `x-teleport` work with Livewire now? I know there were issues at first.

Let try to explain why it does not work, but why its working here.

Lets say i have this Livewire component:
```html
<div wire:id="abc" x-data="{}">
   <button>Open dropdown</button>
   
   <ul class="dropdown">
           <li wire:click="action(1)">Item 1</li>
           <li wire:click="action(2)">Item 2</li>
           <li wire:click="action(3)">Item 3</li>
       </ul>
</div>
```

If i teleport the `ul` outside de Livewire component, it will lose the reference to the component (`wire:id=abc`), so any of the `li` wire:clicks, will not work.

But, note that the Livewire component, is a Alpine component too, so even if the `ul` get teleported outside of it, the "parent" it will continue to work because it is still inside the Livewire component, an thus the teleported component can communicate with its parent using the x-teleport proxy.

So its just a matter of making the `ul` call its parent (that continues inside the Livewire) (using x-on:click instead of wire:click), and them the parent makes the call to Livewire.  Thats why its working.

1 - User clicks on a action
2 - The teleported action will make the call to its parent (via the x-teleport proxy)
3 - The parent makes the call to the Livewire component

ryangjchandler:
Even if it _does_ work here, I'd be hesitant to use it until Caleb says that it work with Livewire. Thoughts @danharrin?

saade:
> Even if it _does_ work here, I'd be hesitant to use it until Caleb says that it work with Livewire. Thoughts @danharrin?

i completly understand your point! Up to you guys.

But then, action groups seems a useless adition, since we cant really use them.

ryangjchandler:
They're usable, we'll just have to find an alternative solution, that's all.

saade:
Up to you. thanks anyway

danharrin:
Replaced by #2994. Thanks!

PR #2880, opened 2022-06-22
Added missing translations to Turkish language

suleymanozev:


packages/tables/resources/lang/tr/table.php
```diff
@@ -59,6 +59,31 @@
 
     ],
 
+
+    'bulk_actions' => [
```
danharrin, line 63: These translations are outdated, they have moved to support/actions

danharrin (changes requested):

packages/tables/resources/lang/tr/table.php
zepfietje: They're also still in the `en` translations file. Should they be removed there as well, @danharrin?

danharrin: oh, really? that must be a merge problem then, yeah they need removing everywhere

zepfietje: https://github.com/filamentphp/filament/pull/2886 @danharrin

PR #2885, opened 2022-06-22
Normalize English actions translations

zepfietje:


zepfietje:
Waiting for https://github.com/filamentphp/filament/pull/2886 to be merged.

danharrin:
Hmm, not sure if I agree with some of these? What was the thought process behind this?

zepfietje:
I'm also open to discussion for sure. When I was updating the Dutch translations, I copied over the English ones. Then I noticed that some inconsistencies in the English translations.

Firstly, some actions' messages had "Record" prepended and some didn't. In particular, the most recently added ones like replicate and soft deletes actions were inconsistent with already existent messages. Having "Record deleted" or ":Label deleted" instead of "Deleted" is fine by me as well, but I think the point is just to be consistent across all actions.

Similar thing with some 'multiple' actions, where some had "selected" appended while others used just the verb as label.

Lastly, I found the create action label to be inconsistent with the others where it was just set to the action verb.

Let me know what you think, @danharrin.

packages/support/resources/lang/en/actions/create.php
```diff
@@ -4,7 +4,7 @@
 
     'single' => [
 
-        'label' => 'New :label',
```
danharrin, line 7: I prefer the old one for this action, since its usually on its own and not associated with a particular record like "Edit" is, so its clear what you are creating

Also `New` instead of `Create` since `Create` is the action of saving

packages/support/resources/lang/en/actions/detach.php
```diff
@@ -37,7 +37,7 @@
             'actions' => [
 
                 'detach' => [
-                    'label' => 'Detach selected',
```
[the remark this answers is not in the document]
zepfietje: So now you actually did merge it without `selected`?

packages/support/resources/lang/en/actions/create.php
zepfietje: I understand your thinking. However, this is a little harder in many languages (like Dutch) where the word `New` should be inflected based on the word passed to `:label`.

So for example:
- Nieuwe reactie
- Nieuw bericht

And as I already saw this key being translated as just `Create`, I thought it would be good to make this consistent across locales.

zepfietje:
Sorry for not getting back to this PR, @danharrin!

danharrin:
Its alright, all fixed 👍

zepfietje:
Did you see the comments I just added to your review comments though? 😄

packages/support/resources/lang/en/actions/detach.php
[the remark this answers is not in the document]
danharrin: Yeah, I thought you had removed it from the bulk action name, not the modal button label, apologies

packages/support/resources/lang/en/actions/create.php
danharrin: I'm fine with it being different across locales if that works best, but in English it's clearer to use `New :label` IMO

packages/support/resources/lang/en/actions/detach.php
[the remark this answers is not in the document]
zepfietje: Ah no, just the button label. 👍

packages/support/resources/lang/en/actions/create.php
zepfietje: Makes sense to use the best possible option for every language!

PR #2889, opened 2022-06-22
chore: deprecated `MultiSelect`

ryangjchandler:
`MultiSelect` can now be replaced with `Select::multiple()` so we should mark as deprecated before v3.x.

### DIFF
--- a/packages/forms/src/Components/MultiSelect.php
+++ b/packages/forms/src/Components/MultiSelect.php
@@ -2,6 +2,10 @@
 
 namespace Filament\Forms\Components;
 
+/**
+ * @deprecated Use `\Filament\Forms\Components\Select` and `->multiple()` instead.
+ * @see Select
+ */
 class MultiSelect extends Select
 {
     public function isMultiple(): bool


danharrin:
Personally, I think I prefer `MultiSelect` though 🤔 as the only exception to the modifier rule. Looks clearer to me that it outputs an array of values and not just a single one, do you not agree?

But tbh, we do have `FileUpload::multiple()`, so maybe I am wrong.

ryangjchandler:
For consistency's sake, I'd say deprecate `MultiSelect` too...

danharrin:
Need to also swap out every mention of MultiSelect in the codebase & documentation, and potentially deprecate MultiSelectFilter as well.

iddir94:
I prefer keeping Select and MultiSelect separated

iddir94:
I prefer keeping Select and MultiSelect separated

iddir94:
I prefer keeping Select and MultiSelect separated

danharrin:
Yeah I feel like this is personal preference, `Select::make()->multiple()` is already possible for those who wanna use it

PR #2892, opened 2022-06-22
Update id translation

NathanaelGT:


packages/support/resources/lang/id/actions/force-delete.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+return [
+
+    'single' => [
+
+        'label' => 'Hapus paksa',
```
akunbeben, line 7: I think it should be either 'Hapus permanen' or 'Hapus selamanya'.

PR #2903, opened 2022-06-23
Modularize Composer scripts

szepeviktor:
Just that :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -55,7 +55,12 @@
         }
     },
     "scripts": {
-        "test": "./vendor/bin/pest --parallel && ./vendor/bin/phpstan analyse"
+        "test:pest": "pest --parallel",
+        "test:phpstan": "phpstan analyse",
+        "test": [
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -55,7 +55,12 @@
         }
     },
     "scripts": {
-        "test": "./vendor/bin/pest --parallel && ./vendor/bin/phpstan analyse"
+        "test:pest": "pest --parallel",
```
danharrin, line 58: Don't we still need the `./vendor/bin` prefix though?

szepeviktor: Composer is witty: puts `./vendor/bin` in the beginning of `$PATH`
https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands (purple callout)
Even on Windows!

danharrin: Awesome, TIL!

szepeviktor:
Thank you.

PR #2906, opened 2022-06-23
feature: Validation notifications

danharrin:
Unknown-2


ryangjchandler:
Can we make this an opt-in?

danharrin:
Yeah that is the plan, why I haven't merged it yet

intrepidws:
This is great! My only thought is whether it might make more sense to have a more general/generic error message, simply alerting the user that errors exist. Especially in cases where multiple fields have validation errors, this might make more sense?

intrepidws:
Would this message occur on form submit? Or automatically as fields change?

danharrin:
On submit, when the fields are validated

danharrin:
```php
use Filament\Pages\Page;

Page::$reportValidationErrorUsing = function (ValidationException $exception) {
    Filament::notify('danger', $exception->getMessage());
};
```

tanthammar:
If I understand this correctly, I would have to add something to every Resource, Create/Update Page if I want to show the errors as notifications?
Or could we combine it with a new config, like "show-form-errors-as-notifications"?

intrepidws:
@tanthammar I was able to add the code that @danharrin provided above to a service provider, instead of adding it to every resource.

That said, I do think this would make sense as a config option. To me, this is expected functionality just like the success notification is.

danharrin:
Trying to move away from more config options as code-driven config is much more flexible IMO, and we just end up polluting the config with so many options

tanthammar:
> @tanthammar I was able to add the code that @danharrin provided above to a service provider, instead of adding it to every resource.
> 
> That said, I do think this would make sense as a config option. To me, this is expected functionality just like the success notification is.

@intrepidws Can you please tell me how you added it to the service provider?

intrepidws:
@tanthammar No problem, the code that I'm using is below but it came more or less directly from what Dan provided here: https://github.com/filamentphp/filament/pull/2906#issuecomment-1181960399:

```php
\Filament\Pages\Page::$reportValidationErrorUsing = function (ValidationException $exception) {
    Filament::notify('danger', $exception->getMessage());
};
```

tanthammar:
@intrepidws just like that, not wrapped in anything? In `register()` or `boot()`? Which service provider?

intrepidws:
@tanthammar In the `boot()` method of the AppServiceProvider would work fine.

tanthammar:
@intrepidws ok tnx. Nothing happens for me. No messages in Admin panel when validating a resource form.

PR #2911, opened 2022-06-24
Simplify getCommands

szepeviktor:
By one simple `array_reduce()`.

### DIFF
--- a/packages/forms/src/FormsServiceProvider.php
+++ b/packages/forms/src/FormsServiceProvider.php
@@ -25,18 +25,18 @@ class FormsServiceProvider extends PackageServiceProvider
             Commands\MakeLayoutComponentCommand::class,
         ];
 
-        $aliases = [];
+        return array_reduce(
+            $commands,
+            static function ($commands, $command) {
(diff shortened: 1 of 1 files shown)


danilopolani:
Don't mean to be rude, but IMO the before is better: "simplicity" to me does not always mean to write less code/use less functions, but also to write a code as readable as possible.

Also please note that the code is present also in [FilamentServiceProvider](https://github.com/filamentphp/filament/blob/2.x/packages/admin/src/FilamentServiceProvider.php#L64-L76), [TableServiceProvider](https://github.com/filamentphp/filament/blob/2.x/packages/tables/src/TablesServiceProvider.php#L27-L39) and also in [another Filament repository](https://github.com/filamentphp/spatie-laravel-settings-plugin/blob/2.x/src/SpatieLaravelSettingsPluginServiceProvider.php#L29-L41) so a change should affect all of those occurrences to keep the consistency high

szepeviktor:
The ultimate solution is to use an array handling package: `Arr::make($data)->doThis()->doThat()`

PR #2918, opened 2022-06-24
Fixes typo in FA translations.

fsamapoor:
It's a common typo in FA. Although it might be hard to see the difference, you can check its correctness with [google translate](https://translate.google.com/?sl=en&tl=fa&text=loading&op=translate).

### DIFF
--- a/packages/forms/resources/lang/fa/components.php
+++ b/packages/forms/resources/lang/fa/components.php
@@ -202,7 +202,7 @@ return [
             'false' => 'خیر',
         ],
 
-        'loading_message' => 'درحال بارگزاری...',
+        'loading_message' => 'درحال بارگذاری...',
 
         'no_search_results_message' => 'هیچ گزینه ای با جستجوی شما مطابقت ندارد',
 


danharrin:
Thanks!

themahdavi:
https://lamtakam.com/qanda/1087/%DA%A9%D9%84%D9%85%D9%87-%D8%A8%D8%A7%D8%B1%DA%AF%D8%B2%D8%A7%D8%B1%DB%8C-%D8%B5%D8%AD%DB%8C%D8%AD-%D8%A7%D8%B3%D8%AA-%DB%8C%D8%A7-%D8%A8%D8%A7%D8%B1%DA%AF%D8%B0%D8%A7%D8%B1%DB%8C-%D8%9F

@fsamapoor in this article equivalent of UPLOAD is بارگذاری 

@danharrin need to be revert pr older version is correct

fsamapoor:
Hi Ali.
I'm sure it's the correct one for both cases. They're so commonly mistaken for each other that you might get confused about which one to choose. When you're actually "doing" something like uploading the file or something is being "done" like being loaded, it's always the right choice.

themahdavi:
I think if uploading or loading are doing thing older version is relevant from newer translate

google say that older version is best enough ;)

PR #2927, opened 2022-06-26
create slug based on path and model class to prevent duplicate slugs

mvenghaus:
if you have the same model name in different folders a duplicate slug is created. You can handle it manually in the resource .. but I think this approach should work and is backward compatible


danharrin:
The tests are failing because this is a breaking change. You need to check if the path starts with `App\Models` first, before using this new code. Otherwise, use `class_basename()` instead.

Also, nested resource slugs should be in the format of `shop/products` rather than `shop-products`, for `Models\Shop\Product`

mvenghaus:
more complex than i thought .. should i close the pull request? .. because with this information there are several ways to solve this .. in my case i will override the slug in the resource

PR #2945, opened 2022-06-27
faking livewire entangle.lazy

tanthammar:
This is a base for faking `$wire.entangle(...).lazy `.
A 2nd argument `$except`, has been added to `applyStateBindingModifiers()`. (Suggested by Dan)

The `TextInput` has been updated to support `lazy()` on fields using `TextInput\Mask`.

It will not be as easy as adding an `x-on:blur` to all other entangled alpine fields, but this is a start.


packages/forms/resources/views/components/text-input.blade.php
```diff
@@ -42,10 +42,11 @@
                 @else
                     x-data="textInputFormComponent({
                         {{ $hasMask() ? "getMaskOptionsUsing: (IMask) => ({$getJsonMaskConfiguration()})," : null }}
-                        state: $wire.{{ $applyStateBindingModifiers('entangle(\'' . $getStatePath() . '\')') }},
+                        state: $wire.{{ $applyStateBindingModifiers('entangle(\'' . $getStatePath() . '\')', except: 'lazy') }},
                     })"
                     type="text"
                     wire:ignore
+                    @if (in_array('lazy', $getStateBindingModifiers())) x-on:blur="$wire.$refresh" @endif
```
danharrin, line 49: We could maybe extract this into a `hasStateBindingModifier('lazy')` method?

packages/forms/src/Concerns/HasStateBindingModifiers.php
```diff
@@ -29,11 +29,17 @@ public function stateBindingModifiers(array $modifiers): static
         return $this;
     }
 
-    public function applyStateBindingModifiers($expression): string
+    public function applyStateBindingModifiers($expression, ?string $except = null): string
     {
         $modifiers = $this->getStateBindingModifiers();
 
-        return implode('.', array_merge([$expression], $modifiers));
+        $merged = array_merge([$expression], $modifiers);
```
danharrin, line 36: Logically, it makes more sense to filter the array to remove the modifiers, then merge the expression

We can probably do it with a collection, chaining `->implode()` at the end

danharrin (changes requested):

packages/forms/resources/views/components/text-input.blade.php
tanthammar: I it is going to be a separate method, I like short names ;) `@if(isLazy()) ... @endif`

tanthammar:
@danharrin 
I filtered the modifiers.

packages/forms/src/Concerns/HasStateBindingModifiers.php
tanthammar: I changed it, see comment below.

danharrin:
Thanks!

PR #2952, opened 2022-06-28
FIX: Respect action align setting

bernhardh:
Also respect the config on pages


danharrin:
What issue is this fixing? It works fine for me already

Screenshot 2022-07-08 at 21 36 47

bernhardh:
I am sorry. I have no idea, where the error occured and can't reproduce it. So I will close the pull request. Thanks!

bernhardh:
@danharrin I now found again, why I created this pull request, but I am not sure, if this is the best solution.

Currently, when using your `settings` plugin (https://filamentphp.com/docs/2.x/spatie-laravel-settings-plugin/installation) this plugin doesn't respect the align from the settings, while all other pages does:

image

So we could either add the config to the `settings-page.blade.php`:

```
<x-filament::pages.actions :actions="$this->getFormActions()" :align="config('filament.layout.forms.actions.alignment')" />
```

or just make it default in the actions-component itself, instead of left, like I have done it in my Pull request.

bernhardh:
Ok. I think I have removed it from the two other code locations

packages/admin/resources/views/components/modal/actions.blade.php
```diff
@@ -1,6 +1,5 @@
 <x-filament-support::modal.actions
     :attributes="\Filament\Support\prepare_inherited_attributes($attributes)"
-    :align="config('filament.layout.forms.actions.alignment')"
```
danharrin, line 3: Hasn't this broken modal actions now? This line probably needs to stay

bernhardh:
Sorry, I have now set the config in the corresponding view as well, so that its the same logic then in the other one.

danharrin:
We can't use the admin config in the support package.

Worked out what was wrong, we were just missing an `actions.modal.actions.alignment` option for the admin panel, like we do for forms and tables. I've added that.

To change the alignment of all actions, you can now use `forms.actions.alignment` and `actions.modal.actions.alignment` together.

danharrin:
@bernhardh #3179

PR #2956, opened 2022-06-28
Make Card Width configurable

ralphjsmit:
This PR adds a simple prop to the `components/layouts/card.blade.php` template to specify the width of the card (the template used for the login card).

This is useful for plugin authors who want to display a wider form or wizard in those components. For example, in the onboarding plugin I'm working, I allow people to add wizards or forms into that card. Making the width dynamic would be very handy. 

Thanks!

### DIFF
--- a/packages/admin/resources/views/components/layouts/card.blade.php
+++ b/packages/admin/resources/views/components/layouts/card.blade.php
@@ -1,6 +1,6 @@
 @props([
     'title' => null,
-    'width' => 'md'
+    'width' => 'md',
 ])
 
 <x-filament::layouts.base :title="$title">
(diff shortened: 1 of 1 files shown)


packages/admin/resources/views/components/layouts/card.blade.php
```diff
@@ -1,13 +1,29 @@
...
+                'sm' => 'max-w-sm',
+                'md' => 'max-w-md',
+                'lg' => 'max-w-lg',
+                'xl' => 'max-w-xl',
+                '2xl' => 'max-w-2xl',
+                '3xl' => 'max-w-3xl',
+                '4xl' => 'max-w-4xl',
+                '5xl' => 'max-w-5xl',
+                '6xl' => 'max-w-6xl',
+                '7xl' => 'max-w-7xl',
```
ryangjchandler, line 24: We should add a `default => $width` here too since somebody could use a custom class and generate their own theme too.

ryangjchandler (changes requested):

packages/admin/resources/views/components/layouts/card.blade.php
ralphjsmit: Done, thanks👍

PR #2965, opened 2022-06-29
Feature: allow storing filters in session

ralphjsmit:
This PR allows people to store the values of the filters on their resource in the session and retrieve the query with it. That means that e.g. on the edit page, people can replicate the exact query that was being used on the `index` page. 

This is very useful for stuff like 'previous' and 'next' buttons, but there might also be other handy use cases that I didn't think of yet. 

```php
Action::make('previous')
    ->url(function() {
        $previousRecord = YourResource::getLastFilteredEloquentQuery()
            ->where('id', '<', $this->record->id)
            ->take(1)
            ->first();
            
        return YourResource::getUrl('edit', ['record' => $previousRecord]);
    })
    ->icon('heroicon-o-arrow-left')
    ->color('secondary')
    ->label('Previous')
```

Thanks!


danharrin:
Thanks, managed to simplify the implementation quite a bit though, just to 1 method. Don't want to add anything we don't especially need right now

PR #2976, opened 2022-06-29
feature: Dynamic builder block labels

danharrin:


invaders-xx:
@danharrin, How do we activate this functionality ? Thanks for your feedback

danharrin:
```php
Block::make('heading')
    ->label(function ($state) {
        return $state['label'] ?? 'Heading';
    })
    ->schema([
        TextInput::make('label')->lazy()
    ])
```

webbipage:
> ```
> Block::make('heading')
>     ->label(function ($state) {
>         return $state['label'] ?? 'Heading';
>     })
>     ->schema([
>         TextInput::make('label')->lazy()
>     ])
> ```

Hello @danharrin, how can we avoid the ordering number of the blocks appear after the label? For example, I have only one block with the lazy label called "Topic A". The result of the label will be "Topic A 1" now. Thank you!

danharrin:
`Builder::make('content')->withBlockNumbers(false)`

PR #2980, opened 2022-06-30
Change UiAvatarProvider logic

themahdavi:
For fix problem on UTF8 locales with `urlencode()` method on UiAvatarProvider

use **+** instead of **space**


wychoong:
before:
image

after:
image

Issue #2834, opened 2022-06-19
Error RestoreAction on EditPage


themahdavi:
### Package

filament/filament

### Package Version

2.13.6

### Laravel Version

9.17

### Livewire Version

2.10

### PHP Version

8.1.5

### Bug description

RestoreAction On Edit Page functionality work and item restored but after click restore modal not close and have this console error
`Uncaught (in promise) DOMException: String contains an invalid character`

themahdavi:
how to reproduce this error ?

on table actions and bulk actions work nice but on page action give console error

danharrin:
It works fine for me in Chrome and Safari :(

themahdavi:
test on chrome : 
![image](https://user-images.githubusercontent.com/31320104/174882819-0acf3ff0-9e8f-4ddb-ad6a-07a04fea94b7.png)

themahdavi:
I find problem 
https://github.com/laravel-filament/filament/blob/3b7b71cf95e06ae9c057dc0998ecbe508319f90c/packages/admin/src/Pages/Actions/DeleteAction.php#L35

`$this->keyBindings(['mod+d']);` cause error
when comment this line work fine 

what means mod+d ?
and why on RestoreAction affect from DeleteAction ?

themahdavi:
hey @danharrin on new version i am losing RestoreAction on my 2 projects and downgrade version
can you review ` $this->keyBindings(['mod+d']); ` to find solution ?

pxlrbt:
Having the same issue: MacOS, Chrome/Safari, latest Filament

> Unhandled Promise Rejection: InvalidCharacterError: The string contains invalid characters.

https://user-images.githubusercontent.com/22632550/176637429-ae329f40-ff00-44fd-85e6-7f800e69bc83.mov

pxlrbt:
The issue is the `+` symbol which apparently causes issues on DOM diffing.

danharrin:
@pxlrbt is there a list of safe attribute characters anywhere?

pxlrbt:
> @pxlrbt is there a list of safe attribute characters anywhere?

@danharrin I was searching for "valid attribute names" but didn't find any. What I tried: `-`, `_`, `.` are good. So maybe underscore or a dash`

danharrin:
Fixed by #3039.

PR #2982, opened 2022-06-30
Remove keybindings from DeleteAction

pxlrbt:
The keybinding on DeleteAction (which is the only PageAction with a keybinding) is causing issues as `+` is not a valid char for attribute names (https://github.com/filamentphp/filament/issues/2834).

Removing this until there is a solution for `x-mousetrap`


themahdavi:
Hey @danharrin on my all projects RestoreAction not working and all customers say website crash and when need restore i am changing manually from database .
Please deside for `+` or remove until mousetrap update 
Thanks

danharrin:
Replaced by #3039.

PR #2986, opened 2022-06-30
Add `enableOpen` to FileUpload field

pxlrbt:
Add a new feature `enableOpen()`, which adds an `external-link` icon to the files and opens them in a new tab (or same tab if you wish)

Bildschirmfoto 2022-06-30 um 12 50 59


danharrin:
Removed the new tab option, want to protect anyone from making that poor UX choice 😂

PR #2999, opened 2022-07-01
fix: Stale Spatie tags

### LINKED ISSUE #2797: stale UI values after update
### Package

filament/spatie-tags-plugin

### Package Version

^2.13.4

### Laravel Version

v9.17.0

### Livewire Version

v2.10.5

### PHP Version

PHP 8.1.4

### Bug description

When updating a form and remaining on the page, the tags input field will either retain removed values or will remove newly added values from the UI itself. The database receives the correct values and refreshes to the page will show the correct values. Currently using `spatie/laravel-tags` 4.3.2

I have tried with saving the relationship manually, and that causes the db to lose the correct values after `update()` was called. Also tried a combination of including `tags` in my form fill or leaving it empty/null

Attached are two screen caps, one for removing a tag and one for adding tags

Removing tag with manual refresh after

tags_issue

Adding tag with manual refresh after

tags_add_issue

### Steps to reproduce

```php
use Filament\Forms\Components\SpatieTagsInput;

class EditDemoComponent extends Component implements Forms\Contracts\HasForms {

use Forms\Concerns\InteractsWithForms;

    public $tags;
    public ModelWithHasTags $modelWithHasTags;

    {
        return [
            SpatieTagsInput::make('tags')
                ->type('support'),
        ];
    }
    
    protected function getFormModel(): ModelWithHasTags
    {
        return $this->modelWithHasTags;
    }    

    public function submit()
    {
        $this->modelWithHasTags->update($this->form->getState());
    }
}
```

### Relevant log output

_No response_

danharrin:
Fixes #2797.

### DIFF
--- a/packages/spatie-laravel-tags-plugin/src/Forms/Components/SpatieTagsInput.php
+++ b/packages/spatie-laravel-tags-plugin/src/Forms/Components/SpatieTagsInput.php
@@ -21,7 +21,7 @@ class SpatieTagsInput extends TagsInput
             }
 
             $type = $component->getType();
-            $tags = $record->tagsWithType($type);
+            $tags = $record->load('tags')->tagsWithType($type);
 
             $component->state($tags->pluck('name')->toArray());
         });


danharrin:
Yeah I can confirm this. Please open a new issue.

danharrin:
@ainesophaur I have tracked the issue down, and it's not on Filament's side.

In the view, replace this...

```blade
@if(!empty($result))
    <p>Result: {{$result}}</p>
@endif
```

with this

```blade
<div>
    @if(!empty($result))
        <p>Result: {{$result}}</p>
    @endif
</div>
```

it's an issue with Livewire's DOM diffing.

More info - https://laravel-livewire.com/docs/2.x/troubleshooting#dom-diffing-issues

Check the part that says "Wrap Blade conditionals (@if, @error, @auth) in an element"

ainesophaur:
Confirmed @danharrin. Thank you for your diligence in my reported issues and for the wonderful packages.

danharrin:
You're welcome! ❤️

PR #3017, opened 2022-07-03
[Enhance] move Khmer language code to standard ISO-639

reksmeysrey:
[List of ISO 639 codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)


danharrin:
Thanks

PR #3021, opened 2022-07-03
Add Dusk selectors to actions and notification

tanthammar:
In a multilingual app it is easier to write language neutral tests.

With dusk selectors added to actions and notifications you can do something like this:
```php
    public function testOrganizerUpdate()
    {
        $this->browse(function (Browser $browser) {
            $browser->visit('/admin/organizers')
                ->pause(500)
                ->click('@filament.tables.action.edit')
                ->pause(500)
                ->clear('@filament.forms.data.name')
                ->type('@filament.forms.data.name', fake()->company())
                ->click('@filament.admin.action.save')
                ->pause(1000)
                ->assertPresent('@filament.admin.notification.success');
        });
    }
```


danharrin:
Slightly tweaked the notification key, since we will soon be integrating filament/notifications and asserting a particular status might not be appropriate any more, so I don't want to commit to keeping that in a future version

The rest are good though, thanks

PR #3022, opened 2022-07-03
Support Debounce

tanthammar:
The goal with this PR is to support Livewire model debounce modifier and to take control over the debounce value in general.

When ->reactive() is to fast, and .lazy is to slow... you want to reach for debounce.

Non breakable change.

New method examples
```php
SomeField::make()->debounce(delay: '700ms')

x-on:input.debounce.{{ $getDebounce() }}="updateState"  // for blade files
```

Fields that are `entangled` doesn't support `.lazy` or `.debounce` therefore the `applyStateBindingModifiers()` has been updated to return different values based on if the field is entangled or not. (The `$except` parameter from "fake entangle" has been removed, it becomes obsolete)

Debounced alpine fields have been updated to use a dynamic value instead of fixed "500ms", to let the developer decide the delay.  (Only applied to `KeyVal` field because it is not applicable in the `DatePicker`)

The new `$getDebounce()` method can be used to pass the debounce value to alpine components.

Any field that has a `wire.model="foo"` attribute, can now use the new debounce method.

The default value is set to `'500ms'`

If one tries to use the `.debounce()` method on an entangled field the modifier is ignored.

(Replaces #3002 because I merged two PRs by mistake)

### DIFF
--- a/packages/forms/resources/views/components/key-value.blade.php
+++ b/packages/forms/resources/views/components/key-value.blade.php
@@ -25,32 +25,32 @@
                 'dark:divide-gray-700' => config('forms.dark_mode'),
             ])>
                 <thead>
-                <tr @class([
-                        'bg-gray-50',
-                        'dark:bg-gray-800/60' => config('forms.dark_mode'),
-                    ])>
-                    <th @class([
-                            'px-4 py-2 whitespace-nowrap font-medium text-sm text-gray-600',
-                            'dark:text-gray-300' => config('forms.dark_mode'),
-                        ]) scope="col">
-                        {{ $getKeyLabel() }}
-                    </th>
-
-                    <th @class([
-                            'px-4 py-2 whitespace-nowrap font-medium text-sm text-gray-600',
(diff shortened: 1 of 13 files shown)


packages/forms/resources/views/components/key-value.blade.php
```diff
@@ -25,32 +25,32 @@
                 'dark:divide-gray-700' => config('forms.dark_mode'),
             ])>
                 <thead>
-                    <tr @class([
```
zepfietje, line 28: Could you fix these indentation changes?

zepfietje (changes requested):

packages/forms/resources/views/components/key-value.blade.php
tanthammar: Yes, sure :)

danharrin:
Please mark this as "ready for review" when the formatting changes have been removed :)

zepfietje:
I've fixed the formatting changes, so this PR is now ready for review.

danharrin:
Thanks

PR #3027, opened 2022-07-04
feature: Testing helpers

danharrin:


packages/tables/src/Concerns/HasFilters.php
```diff
@@ -67,7 +73,7 @@ public function resetTableFiltersForm(): void
 
     protected function applyFiltersToTableQuery(Builder $query): Builder
     {
-        $data = $this->getTableFiltersForm()->getState();
+        $data = $this->getTableFiltersForm()->getRawState();
```
juliomotol, line 76: @danharrin Sorry to dig up an old PR, but why the change to `getRawState()`?

We have a `Filter` with a custom form and we'd want to use `dehydrateStateUsing()` on a component in that form but its doesn't pass through it.

danharrin: Because otherwise validation happens. Instead of `dehydrateStateUsing()`, you can just mutate the data inside `query()`

PR #3030, opened 2022-07-04
Remove `final` from the constructor

### LINKED ISSUE #3029: Filament\Tables\Columns\Column::__construct is final
### Package

filament/forms
filament/tables

### Package Version

v2.13.21

### Laravel Version

v9.2

### PHP Version

8.1.7 (cli)

### Bug description

The class `Filament\Tables\Columns\Column` has its constructor marked as final. This is a huge problem if you want to overwrite some behavior in a sub class. For example, i have a lot of `TextColumn` elements in my forms and all of them need certain attributes set. To keep code clean i don't want to copy and paste the setters all the time. Instead i created my own class extending `TextColumn` and wanted to set them in one place.

# Example:
I have a lot of those make calls in my code:
```php
TextColumn::make('email')
  ->searchable()
  ->sortable(),
```

To avoid the duplication i have created `LeoTextColumnEmail` with the following code:
```php

class LeoTextColumnEmail extends TextColumn
{
  public function __construct(string $name)
  {
    parent::__construct($name);
    
    $this->searchable();
    $this->sortable();
  }
}
```
But sadly i cannot do that, because your `__construct` is set to final. 

Could you please be so kind an change that in the next release? It would not harm your code and it would not break compatibility in any way. 

Thanks and Greetings
Leo

PS: I you would not mind, the same principle would apply to `Filament\Forms\Components\Field` as well. It would allow the creation of a lot of custom fields with easy defaults.

leo-unglaub:
This allows for more fine controll when sub-classing an existing Column or Field. 

This implements the request from #3029

### DIFF
--- a/packages/forms/src/Components/Field.php
+++ b/packages/forms/src/Components/Field.php
@@ -15,7 +15,7 @@ class Field extends Component implements Contracts\HasValidationRules
 
     protected string $viewIdentifier = 'field';
 
-    final public function __construct(string $name)
+    public function __construct(string $name)
     {
         $this->name($name);
         $this->statePath($name);
--- a/packages/tables/src/Columns/Column.php
+++ b/packages/tables/src/Columns/Column.php
@@ -30,7 +30,7 @@ class Column extends ViewComponent
     protected string $evaluationIdentifier = 'column';
     protected string $viewIdentifier = 'column';
 
-    final public function __construct(string $name)
+    public function __construct(string $name)
     {
         $this->name($name);
     }


ryangjchandler:
The right way to have this customisation is by overwriting the `setUp()` method in your base class.

leo-unglaub:
I see. I did not find that in the docs. Sorry for the noise. :)

So i asume i should use `setUp` because the signature of `__construct()` can change and would create more breakage? If that's the case, i withdraw my PR. 

Thanks!

danharrin:
Yeah, avoid overriding the constructor. We use final to prevent the signature from changing, as we use `new static()` to instantiate. PHPStan doesn't like it not being final.

As Ryan said, use `setUp()`, which is designed for this:

```php
{
    parent::setUp();

    $this->searchable();
    $this->sortable();
}
```

danharrin:
Sorry, we both commented the same thing at the same time

leo-unglaub:
Makes sense, thanks to both of you!

PR #3043, opened 2022-07-06
Striped tables

G3z:
Add optional striped tables


danharrin:
Yeah, I have moved this to a method on the table trait

PR #3050, opened 2022-07-07
Pass old state to form field callback

NathanaelGT:
This PR provide a way to get the old state from field callback
usage:
```php
->afterStateUpdated(function ($old) {
    // do something that require old state
})
```

### DIFF
--- a/packages/forms/src/Components/Component.php
+++ b/packages/forms/src/Components/Component.php
@@ -38,6 +38,7 @@ class Component extends ViewComponent
             'get' => $this->getGetCallback(),
             'livewire' => $this->getLivewire(),
             'model' => $this->getModel(),
+            'old' => $this->getOld(),
             'record' => $this->getRecord(),
             'set' => $this->getSetCallback(),
             'state' => $this->shouldEvaluateWithState() ? $this->getState() : null,
--- a/packages/forms/src/Components/Component.php
+++ b/packages/forms/src/Components/Component.php
@@ -38,7 +38,6 @@ class Component extends ViewComponent
             'get' => $this->getGetCallback(),
             'livewire' => $this->getLivewire(),
             'model' => $this->getModel(),
-            'old' => $this->getOldState(),
             'record' => $this->getRecord(),
             'set' => $this->getSetCallback(),
             'state' => $this->shouldEvaluateWithState() ? $this->getState() : null,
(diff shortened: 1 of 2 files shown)


packages/forms/src/Components/Concerns/HasState.php
```diff
@@ -186,6 +187,17 @@ public function getState()
         return $state;
     }
 
+    public function getOld()
+    {
+        $state = request('serverMemo.data.' . $this->getStatePath());
+
+        if (blank($state) || ! Livewire::isLivewireRequest()) {
```
danharrin, line 194: `Livewire::isLivewireRequest()` should probably be the first check before `$state = `

packages/forms/src/Components/Concerns/HasState.php
```diff
@@ -186,6 +187,17 @@ public function getState()
         return $state;
     }
 
+    public function getOld()
```
danharrin, line 190:
```suggestion
    public function getOldState()
```

packages/forms/src/Components/Component.php
```diff
@@ -38,6 +38,7 @@ protected function getDefaultEvaluationParameters(): array
             'get' => $this->getGetCallback(),
             'livewire' => $this->getLivewire(),
             'model' => $this->getModel(),
+            'old' => $this->getOld(),
```
danharrin, line 41: Instead of passing this in all the time, we can just pass it into the `$this->evaluate()` found in the `callAfterStateUpdated()` method

danharrin (changes requested):

packages/forms/src/Components/Component.php
NathanaelGT: Sorry, I'm not really sure about this one.
So instead of passing the old state, do we pass a closure that returns the old state?

packages/forms/src/Components/Concerns/HasState.php
```diff
@@ -59,7 +60,9 @@ public function callAfterStateHydrated(): static
     public function callAfterStateUpdated(): static
     {
         if ($callback = $this->afterStateUpdated) {
-            $this->evaluate($callback);
+            $this->evaluate($callback, [
+                'old' => $this->getOldState(),
```
danharrin, line 64: Like this

PR #3054, opened 2022-07-07
Update getRules() to stay compatible with future v. of Livewire

tanthammar:
There's been som chatter about changing `Livewire\Component::getRules()` to support this'n'that...

Thought best to change the current `InteractsWithForms::getRules()` to pull in the original Livewire method instead of manually replicating it. To stay compatible with however it changes.

I chose not to use `parent::getRules()` as there may be multiple parents, thus causing potential errors.

### DIFF
--- a/packages/forms/src/Concerns/InteractsWithForms.php
+++ b/packages/forms/src/Concerns/InteractsWithForms.php
@@ -300,15 +300,7 @@ trait InteractsWithForms
 
     protected function getRules(): array
     {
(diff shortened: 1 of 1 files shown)


danharrin:
Sorry, could you elaborate more on why we're not using `getRules()`?

tanthammar:
I got warnings in php storm that there there are multiple parents. So I addressed Livewire directly to avoid any potential inheritance problems.

PR #3057, opened 2022-07-07
Docs: Fix renderHook names

pxlrbt:


pxlrbt:
It's already implemented, so it's too late for that. And I think both have their room. Because `body.start` != `body.before`. Maybe

Z3d0X:
You're right!

In the context of `body`(or any other element) it could be meant:
```html
//body.before
<body>
    //body.start
</body>
```

However in the context of `scripts` or `styles`, `before` & `start` means the the same. I guess.

danharrin:
Tbh I feel like the render hooks are incorrectly named, they should be `start` and `end`. But we have to keep BC, so lets add the correct hook above the incorrect one, and not change the docs.

danharrin:
Only added these hooks yesterday (?), and since the docs are our source of truth, I'm just gonna rename them instead of duplicating.

PR #3058, opened 2022-07-08
Fixing idea #3053

dejury:
Fixing idea #3053 

Adds the possibility to create a tab with an icon.

Usage:

```php
Tab::make('Devices')
    ->icon('heroicon-o-desktop-computer')
    ->schema([]);
```

### DIFF
--- a/packages/forms/docs/04-layout.md
+++ b/packages/forms/docs/04-layout.md
@@ -273,6 +273,7 @@ Tabs::make('Heading')
                 // ...
             ]),
         Tabs\Tab::make('Label 3')
+            ->icon('heroicon-o-eye') // Optional icon
             ->schema([
                 // ...
             ]),
--- a/packages/forms/resources/views/components/tabs.blade.php
+++ b/packages/forms/resources/views/components/tabs.blade.php
@@ -57,6 +57,10 @@
                 class="shrink-0 p-3 text-sm font-medium"
                 x-bind:class="{ 'bg-white @if (config('forms.dark_mode')) dark:bg-gray-800 @endif': tab === '{{ $tab->getId() }}' }"
             >
+                @if ($tab->hasIcon())
+                    <x-dynamic-component :component="$tab->getIcon()" class="h-5 w-5 mr-1 text-gray-500 @if (config('forms.dark_mode')) dark:text-gray-400 @endif inline"/>
+                @endif
+
                 {{ $tab->getLabel() }}
             </button>
         @endforeach
--- a/packages/forms/resources/views/components/tabs.blade.php
+++ b/packages/forms/resources/views/components/tabs.blade.php
@@ -54,11 +54,17 @@
                 x-on:click="tab = '{{ $tab->getId() }}'"
                 role="tab"
                 x-bind:tabindex="tab === '{{ $tab->getId() }}' ? 0 : -1"
-                class="shrink-0 p-3 text-sm font-medium"
+                class="flex items-center gap-2 shrink-0 p-3 text-sm font-medium"
                 x-bind:class="{ 'bg-white @if (config('forms.dark_mode')) dark:bg-gray-800 @endif': tab === '{{ $tab->getId() }}' }"
             >
-                @if ($tab->hasIcon())
-                    <x-dynamic-component :component="$tab->getIcon()" class="h-5 w-5 mr-1 text-gray-500 @if (config('forms.dark_mode')) dark:text-gray-400 @endif inline"/>
+                @if ($icon = $tab->getIcon())
+                    <x-dynamic-component
+                        :component="$icon"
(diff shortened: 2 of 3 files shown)


dejury:
Schermafbeelding 2022-07-09 om 22 34 01

packages/forms/src/Components/Tabs/Tab.php
```diff
@@ -34,6 +44,14 @@ public function getColumnsConfig(): array
         return $this->columns ?? $this->getContainer()->getColumnsConfig();
     }
 
+    public function getIcon(): ?string {
+        return $this->evaluate($this->icon);
+    }
+
+    public function hasIcon(): bool {
+        return (bool) $this->getIcon();
```
ryangjchandler, line 52:
```suggestion
        return $this->icon !== null;
```

packages/forms/resources/views/components/tabs.blade.php
```diff
@@ -57,6 +57,10 @@
                 class="shrink-0 p-3 text-sm font-medium"
                 x-bind:class="{ 'bg-white @if (config('forms.dark_mode')) dark:bg-gray-800 @endif': tab === '{{ $tab->getId() }}' }"
             >
+                @if ($tab->hasIcon())
+                    <x-dynamic-component :component="$tab->getIcon()" class="h-5 w-5 mr-1 text-gray-500 @if (config('forms.dark_mode')) dark:text-gray-400 @endif inline"/>
```
ryangjchandler, line 61: I feel like the icon should be a bit smaller, looks a bit too big next to the text.

ryangjchandler (changes requested):

packages/forms/resources/views/components/tabs.blade.php
danharrin: I agree, also the solid Heroicons should be used instead of outline in this context IMO, which could also affect sizing

zepfietje: Exactly what @danharrin says. The sizing classes are in fact correct, but the solid icons should be used.

packages/forms/docs/04-layout.md
```diff
@@ -273,6 +273,7 @@ Tabs::make('Heading')
                 // ...
             ]),
         Tabs\Tab::make('Label 3')
+            ->icon('heroicon-o-eye') // Optional icon
```
zepfietje, line 276:
```suggestion
            ->icon('heroicon-s-eye') // Optional icon
```

packages/forms/resources/views/components/tabs.blade.php
zepfietje, line 61: Probably cleaner to use `:class="\Illuminate\Support\Arr::toCssClasses([])"`

packages/forms/src/Components/Tabs/Tab.php
zepfietje: Do we actually really need this method? `@if ($tab->getIcon())` should suffice in the Blade view.

zepfietje (changes requested):

packages/forms/resources/views/components/tabs.blade.php
```diff
@@ -57,6 +57,10 @@
                 class="shrink-0 p-3 text-sm font-medium"
```
zepfietje, line 57: We should probably make sure to vertically center the icon and label.

zepfietje (changes requested):

packages/forms/src/Components/Tabs/Tab.php
ryangjchandler: We use `hasX` methods elsewhere to prevent unnecessarily evaluating closures. Best to keep the pattern imo.

packages/forms/resources/views/components/tabs.blade.php
ryangjchandler: Yeah, I don't think directives work inside of component attributes anyway?

packages/forms/src/Components/Tabs/Tab.php
danharrin: We actually still need to evaluate the closure in this case, as it could return null. So `getIcon()` can be used for the check and set to `$icon`

zepfietje: Just wanted to comment what @danharrin said. Also, in other places it's used like I suggested above.

packages/forms/resources/views/components/tabs.blade.php
danharrin: I prefer `Arr::toCssClasses()` in this context I think, directives look messy

danharrin:
Fixed

Screenshot 2022-07-17 at 10 32 05

PR #3062, opened 2022-07-08
Added repeater item label option

SebastiaanKloos:
This PR adds the ability to add labels to repeater items. It's very helpful when collapsing items.

### DIFF
--- a/packages/forms/src/Components/Repeater.php
+++ b/packages/forms/src/Components/Repeater.php
@@ -37,6 +37,8 @@ class Repeater extends Field
 
     protected string | Closure | null $relationship = null;
 
+    protected string | Closure | null $itemLabel = null;
+
     protected ?Closure $modifyRelationshipQueryUsing = null;
 
     protected ?Closure $mutateRelationshipDataBeforeCreateUsing = null;
@@ -352,6 +354,13 @@ class Repeater extends Field
         return $this;
     }
 
+    public function itemLabel(string | Closure | null $label): static
+    {
+        $this->itemLabel = $label;
+
+        return $this;
+    }
+
     public function fillFromRelationship(): void
     {
         $this->state(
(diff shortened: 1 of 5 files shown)


packages/forms/src/Components/Repeater.php
```diff
@@ -440,6 +449,15 @@ public function getCachedExistingRecords(): Collection
         );
     }
 
+    public function getItemLabel(string $uuid): ?string
+    {
+        $itemData = data_get($this->getLivewire(), "{$this->getStatePath()}.{$uuid}");
```
pxlrbt, line 454: I think I'd prefer just `$item`, but let's see what the others think.

danharrin: Yup

packages/forms/src/Components/Repeater.php
```diff
@@ -440,6 +449,15 @@ public function getCachedExistingRecords(): Collection
...
+
+        return $this->evaluate($this->itemLabel, [
+            'itemData' => $itemData,
```
danharrin, line 457: I think we should potentially just override `state` instead of using an additional `itemData` variable

SebastiaanKloos: @danharrin Ahh yeah much better! So just change the `itemData` to `state` would do the trick?

pxlrbt: Might there be any edge case where people would want to have the whole repeaters state? 😅

danharrin: I think that is probably unexpected behaviour though to return the entire repeater, state is most useful as the current item here

If they really need all items they can just use `$component->getState()`

pxlrbt: Alright. `$state` is good ;)

Issue #2398, opened 2022-05-12
Markdown editor doesn't get a red border upon validation errors


Cannonb4ll:
### Package

filament/forms

### Package Version

v2.0.0

### Laravel Version

v9.0.0

### PHP Version

PHP 8.1

### Bug description

The markdown editor does not have a red border when a validation error occurs like inputs or selects.

Upon investigation with @danharrin we noticed `wire:ignore` is the cause of this:

https://github.com/laravel-filament/filament/blob/417e359c7023ed2b9cba319eb3529b16a456b47c/packages/forms/resources/views/components/markdown-editor.blade.php#L17

Removing that line fixes the $error state, but introduces a error within action modals.

danharrin:
Can you test out #2398? There are a few other issues but most have been fixed I think

Cannonb4ll:
As far as I can see and test seems all good!

tanthammar:
#3649

PR #3065, opened 2022-07-08
Make collapsible Section header clickable + scrollIntoView onClick

tanthammar:
It is an expected UX to be able to click on the Section header to expand/collapse the section.
This PR also fixes that the section scrolls into focus when it is expanded. It didn't before.

PR #3069, opened 2022-07-08
Make Alpine aware of validation errors and auto-expand Sections

tanthammar:
The goal of this PR is to be able to access validation errors from Alpine.

```js
//in any alpine script
$store.errors.any //boolean 
$store.errors.all //Array
```

I chose the same naming as Blade 
```blade
$errors->any()
$errors->all()
```

There are multiple use cases, but my initial thoughts are
1. animate alpine components, based on errors or not
2. be able to notify, all or selected, validation errors (as popups)
3. plugins can access validation errors or validation status.

My first implementation, also in this PR, is to **make the Section form component expand if there are any errors in the form**.  Better UX if all Sections are collapsed and there is no popup or anything telling you that there are errors in the form.

### Help please:
I am trying to find where where the `create` and `save` form buttons are?
Need to add  `x-on:click="$store.errors.any = false"`, so the value is reset before form submit. 
Otherwise a `$watch` would only work on the first form submission (if there are still errors in the form)


ryangjchandler (commented):
Couple of thoughts from my phone:
1. This only affects the admin panel by the looks of it. Really, it should work for any form component. 
2. If there are multiple forms on the same page (multiple LW components), errors from one form will overwrite the errors in the other form. This would be problematic. 
3. Would you be using this for custom form components? If so, you'd be better off getting the errors for your component in a custom view. If you're not using it for custom components, how exactly would another developer use this?

danharrin:
The multiple Livewire component limitation also hit me immediately

tanthammar:
Agree on all points.

_"Would you be using this for custom form components? If so, you'd be better off getting the errors for your component in a custom view. If you're not using it for custom components, how exactly would another developer use this?"_
- My first goal was to be able to open all form Sections that contains errors, not just the first found.
- Then I'm also interested in displaying notifications for all form error messages. If that should be added to the main notification feature or as a plugin, let's see when this topic is closed :)
- The feature to let Alpine be aware of $errors->any() can be of interest to any field plugin.

_"If there are multiple forms on the same page (multiple LW components), errors from one form will overwrite the errors in the other form. This would be problematic."_
- I updated so each form only replaces its own errors.

"This only affects the admin panel by the looks of it. Really, it should work for any form component."
- Why I placed it in app Is because I couldn't find a suitable blade file to place it in... thought best to start somewhere to show my thoughts. 
- Maybe it should go into form.index.js?

Not sure this is the way to go about it but I'm interested in getting the discussion going and to find a solution.

tanthammar:
Added an example use, to show all form errors as notifications.
This could be in combo w a config, "show-all-errors-as-notifications"
Doing so made me revert back from grouping error notifications with form->id.
Makes sense to only show the latest validation errors.

tanthammar:
Ok, piling up comments here...
Got another idea, updated Section so that all sections that has field validation errors opens.
Then we don't have to emit from backend, because it is based on watching the $store.

tanthammar:
Sunday...
I went through all fields and there are some with `wire:ignore` that also needs to get errors from Alpine, 
because `$errors->has($getStatePath())` won't update, meaning that the field never gets the error border.

Still don't know where to put this, in admin panel `app.blade` looks like the best option, but if you use only the form package, I guess it has to go into the installer and the docs on how to setup the package manually?

```html
x-init="
            Alpine.store('errors', {
                any: @js($errors->any()),
                all: @js($errors->all()),
                entries: {},
            })
            Livewire.hook('message.processed', (message, component) => {
                $store.errors.any = Object.keys(message.response.serverMemo.errors).length > 0
                $store.errors.all = Array.from(Object.values(message.response.serverMemo.errors).flat())
                $store.errors.entries = message.response.serverMemo.errors
            })
        "
```

Other thoughts
- I tried it with multiple forms on the page. It only keeps the latest errors, but in terms of UX that doesn't feel awkward because the users focus is on the field one is currently filling out. Notifications also feels best to show only from the latest form submit.
- would be nice with a $hasErrors(key) Alpine directive, did one, but again, where to put it...

danharrin:
I think we can create a new `errors.js` file in the forms plugin and import that in `index.js` as an Alpine plugin. Then we will have access to `Alpine` to register the store. Then we don't need the user to add anything else to their view to set this feature up.

About showing form errors as notifications, this is already possible without using JavaScript - #2906

About the wire:ignore for components like the markdown editor, this fixes #2398

danharrin:
Fixes #2398

danharrin:
> I am trying to find where where the create and save form buttons are?
> Need to add x-on:click="$store.errors.any = false", so the value is reset before form submit.
> Otherwise a $watch would only work on the first form submission (if there are still errors in the form)

Are these still questions?

packages/admin/resources/views/components/notification-manager.blade.php
```diff
@@ -4,10 +4,21 @@
         add: function (event) {
             this.notifications = this.notifications.concat(event.detail)
         },
+        addError: function (notificationObject) {
```
danharrin, line 7: All this code can be removed from the notification manager

packages/forms/resources/views/components/section.blade.php
```diff
@@ -1,13 +1,14 @@
 <div
     @if ($isCollapsible())
         x-data="{ isCollapsed: {{ $isCollapsed() ? 'true' : 'false' }} }"
+        x-init="$watch('$store.errors?.all', () => isCollapsed = $el.getElementsByClassName('filament-forms-field-wrapper-error-message')[0] ? false : isCollapsed )"
```
danharrin, line 4: Does this code need changing since #3065?

danharrin (changes requested):

tanthammar:
> > I am trying to find where where the create and save form buttons are?
> > Need to add x-on:click="$store.errors.any = false", so the value is reset before form submit.
> > Otherwise a $watch would only work on the first form submission (if there are still errors in the form)
> 
> Are these still questions?

No

packages/forms/resources/views/components/section.blade.php
tanthammar: Yes, this is the magic :), it opens all sections that contains field errors, not just the first one.
- `x-on:expand-concealing-component.window` only opens the first found section with field errors.
- `$watch('$store.errors?.all'` opens all other sections that contains errors
- then the page scrolls to the **first** section containing field errors.
- sections without errors, remains closed.

packages/admin/resources/views/components/notification-manager.blade.php
tanthammar: > About showing form errors as notifications, this is already possible without using JavaScript - #2906

Using the javascript version in notification.manager.blade.php would mean less overhead in backend.
The part with dispatching a browser event  from `InteractsWithForms`, could actually be removed entirely... the Tab field could easily be modified to look for errors in js instead.

packages/forms/resources/views/components/section.blade.php
danharrin: So could you fix the merge conflict for these changes please?

packages/admin/resources/views/components/notification-manager.blade.php
danharrin: It's a breaking change if we now don't allow sending notifications for validation errors in the backend, and its more configurable this way. Also the notification manager is no longer used any more because we have a new notifications package, so please remove all of the notifications logic from this PR.

tanthammar:
@danharrin 
Been on vacation ... 
- updated notification-manager to match current version.
- updated classes in markdown-editor to match current version.

Please let me know of other required changes.

tanthammar:
Closing this as we might resolve it with Livewire v3. For reference, this is the latest version. 
It has a conflict with the notification plugin.
https://github.com/tanthammar/filament/tree/notification-store-session-conflict

PR #3082, opened 2022-07-10
Support BelongsToMany in HasRelationship

drifteaur:
Support BelongsToMany relationships
(I had an error in a MultiSelectFilter, adding support for BelongsToMany fixed it)


danharrin:
Thanks

PR #3084, opened 2022-07-10
feat: Select component can allow HTML in the choices.js config

consignr:
This PR adds a method to the select component to enable the allowHtml option in the choices.js config. The developer can then render html for the select options which is useful for users with avatars, countries with flags...etc

image

``` 
            Select::make('user_id')
                    ->label('User')
                    ->allowHtml()
                    ->searchable() 
                    ->getSearchResultsUsing(function (string $search) {
                        $users = User::where('name', 'like', "%{$search}%")->limit(50)->get();
                        return $users->mapWithKeys(function ($user) {
                            return [
                                $user->getKey() => 
                                    view('filament.components.select-user-result')
                                        ->with('name', $user->name)
                                        ->with('email', $user->email)
                                        ->with('image', $user->image)
                                        ->render()
                                    ];
                        })->toArray();
                    }) 
                    ->getOptionLabelUsing(function ($value): string {
                        $user = User::find($value);
                        return view('filament.components.select-user-result')
                            ->with('name', $user->name)
                            ->with('email', $user->email)
                            ->with('image', $user->image)
                            ->render();
                    }) ```


danharrin:
Thanks, but...

Be safe! Sanitise your options HTML! This will introduce XSS attacks :)

danharrin:
Please submit a code example (for example the user avatar & email) to `https://filamentphp.com/tricks`, many people will find it useful :)

consignr:
Thank for merging! I've submitted a trick (with sanitised HTML). I couldn't upload my screenshots though.

danharrin:
Did you try dragging and dropping your image onto the textarea?

consignr:
Sorted. Thanks! 👍

blite:
great work!, example works for me 

I wonder if example should use profile_photo_url from jetpack, but I don't know how many people continue using that.  

I am digging in now for my actual use case.

PR #3087, opened 2022-07-11
Ability to set step start

Cannonb4ll:
This allows you to set the step start on a wizard, for example if you have a wizard that saves in between, you might want to start the wizard in a later step automatically and not at the start (going through the whole wizard again).

### DIFF
--- a/packages/forms/resources/views/components/wizard.blade.php
+++ b/packages/forms/resources/views/components/wizard.blade.php
@@ -10,7 +10,7 @@
         step: null,
 
         init: function () {
-            this.step = this.getSteps()[{{ $getStartOnStep() }}]
+            this.step = this.getSteps()[{{ ($getStartOnStep() - 1) }}]
         },
 
         nextStep: function () {
--- a/packages/forms/resources/views/components/wizard.blade.php
+++ b/packages/forms/resources/views/components/wizard.blade.php
@@ -10,7 +10,7 @@
         step: null,
 
         init: function () {
-            this.step = this.getSteps()[{{ ($getStartOnStep() - 1) }}]
+            this.step = this.getSteps()[{{ $getStartOnStep() }} - 1]
         },
 
         nextStep: function () {
--- a/packages/forms/resources/views/components/wizard.blade.php
+++ b/packages/forms/resources/views/components/wizard.blade.php
@@ -10,7 +10,7 @@
         step: null,
 
         init: function () {
-            this.step = this.getSteps()[0]
+            this.step = this.getSteps()[{{ $getStartOnStep() }}]
         },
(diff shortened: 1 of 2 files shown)


SebastiaanKloos:
I need this so much! Please merge this :) 🎉

packages/forms/src/Components/Wizard.php
```diff
@@ -65,6 +67,18 @@ public function steps(array | Closure $steps): static
...
     }
 
+    public function startOnStep(int $startOnStep): static
+    {
+        $this->startOnStep = $startOnStep - 1;
+
+        return $this;
+    }
+
+    public function getStartOnStep()
```
danharrin, line 77:
```suggestion
    public function getStartStep()
```

packages/forms/src/Components/Wizard.php
```diff
@@ -16,6 +16,8 @@ class Wizard extends Component
 
     protected string | Htmlable | null $submitAction = null;
 
+    public int | null $startOnStep = 0;
```
danharrin, line 19:
```suggestion
    public int | Closure | null $startStep = 1;
```

packages/forms/src/Components/Wizard.php
```diff
@@ -65,6 +67,18 @@ public function steps(array | Closure $steps): static
         return $this;
     }
 
+    public function startOnStep(int $startOnStep): static
```
danharrin, line 70:
```suggestion
    public function startOnStep(int | Closure $startOnStep): static
```

packages/forms/src/Components/Wizard.php
```diff
@@ -65,6 +67,18 @@ public function steps(array | Closure $steps): static
...
+    {
+        $this->startOnStep = $startOnStep - 1;
```
danharrin, line 72: This logic belongs in the view

```suggestion
        $this->startStep = $startStep;
```

packages/forms/resources/views/components/wizard.blade.php
```diff
@@ -10,7 +10,7 @@
         step: null,
 
         init: function () {
-            this.step = this.getSteps()[0]
+            this.step = this.getSteps()[{{ ($getStartOnStep() - 1) }}]
```
danharrin, line 13: I know this is nitpicky, but it's Alpine's responsibility to assign the step number correctly, not PHP

```suggestion
            this.step = this.getSteps()[{{ $getStartOnStep() }} - 1]
```

danharrin (changes requested):

SebastiaanKloos:
Thanks @danharrin that's really nice!

PR #3091, opened 2022-07-12
✨ `filament/notifications` package

zepfietje:
- [x] Finalize dark mode configuration
- [x] Accept `Closure` in addition to `bool`
- [x] Support action icon position
- [x] Support notification duration
- [x] Support Markdown in notification content
- [x] Update config comment
- [x] [Dusk attribute](https://github.com/filamentphp/filament/pull/3091#issuecomment-1186297096)
- [x] Check accessibility (old notification had some attributes)
- [x] Copy over notifications config from admin
- [x] Consider accepting multiple event arguments instead of array
- [x] Make notifications work with redirects
- [x] Fix link action icon overflowing
- [x] Send notifications in admin panel using new `Notification` object
- [x] Notification transition direction based on alignment
- [x] Finalize notification animations
- [x] Add tests and tests for existing admin panel API
- [x] Update Tailwind Forms plugin
- [x] Check notifications responsiveness
- [x] Write documentation


packages/notifications/src/Http/Livewire/Concerns/CanNotify.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Filament\Notifications\Http\Livewire\Concerns;
+
+use Filament\Notifications\Notification;
+
+trait CanNotify
+{
+    protected function notify(Notification $notification): void
```
danharrin, line 9: This needs to have the same signature as in the existing `CanNotify` trait, but you can add the `Notification | string $status` union type to the first parameter

packages/notifications/src/Http/Livewire/Notifications.php
```diff
@@ -0,0 +1,44 @@
...
+    public function add(array $notification): void
+    {
+        $this->notifications->push(Notification::fromLivewire($notification));
+    }
+
+    public function close(array $notification): void
+    {
+        $notification = Notification::fromLivewire($notification);
+
+        $this->notifications->map(
```
danharrin, line 33: I don't think a map is appropriate here, we can probably just use a `foreach`

packages/notifications/src/Http/Livewire/Notifications.php
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public function add(array $notification): void
+    {
+        $this->notifications->push(Notification::fromLivewire($notification));
+    }
+
+    public function close(array $notification): void
+    {
+        $notification = Notification::fromLivewire($notification);
```
danharrin, line 31: How about `$notificationToClose` so we can avoid the ambiguous `$i` variable later, which can be `$notification`?

packages/notifications/resources/views/components/notification.blade.php
```diff
@@ -0,0 +1,27 @@
...
+            :component="$getIcon()"
+            :class="Arr::toCssClasses([
+                'h-6 w-6',
+                'text-success-400' => $getStatus() === 'success',
+                'text-warning-400' => $getStatus() === 'warning',
+                'text-danger-400' => $getStatus() === 'danger',
+            ])"
+        />
+    @endif
+    <div class="grid flex-1">
```
danharrin, line 22:
```suggestion
    
    <div class="grid flex-1">
```

packages/notifications/resources/views/components/notification.blade.php
```diff
@@ -0,0 +1,27 @@
...
+        <span class="font-medium">{{ $getTitle() }}</span>
+        <p class="text-neutral-500 dark:text-neutral-400">{{ $getDescription() }}</p>
```
danharrin, line 24:
```suggestion

        <p class="text-neutral-500 dark:text-neutral-400">{{ $getDescription() }}</p>
```

packages/notifications/resources/views/components/notification.blade.php
```diff
@@ -0,0 +1,27 @@
...
+    </div>
+    <x-heroicon-s-x class="h-6 w-5 text-neutral-400 dark:text-neutral-500" x-on:click="close" />
```
danharrin, line 26:
```suggestion
    
    <x-heroicon-s-x class="h-6 w-5 text-neutral-400 dark:text-neutral-500" x-on:click="close" />
```

packages/notifications/src/Notification.php
```diff
@@ -0,0 +1,170 @@
...
+
+    public ?string $description = null;
+
+    public ?string $icon = null;
+
+    public ?string $status = null;
+
+    public function __construct()
+    {
+        $this->id(uniqid());
```
danharrin, line 26: How about `Str::uuid()`?

packages/notifications/src/Notification.php
```diff
@@ -0,0 +1,170 @@
...
+use Filament\Support\Components\ViewComponent;
+use Livewire\Wireable;
+
+class Notification extends ViewComponent implements Wireable
+{
+    protected string $view = 'notifications::components.notification';
+
+    public string $id;
+
+    public bool $closed = false;
```
danharrin, line 14:
```suggestion
    public bool $isClosed = false;
```

packages/notifications/src/Notification.php
```diff
@@ -0,0 +1,170 @@
...
+
+        return $this;
+    }
+
+    public function getId(): string
+    {
+        return $this->id;
+    }
+
+    public function closed(bool $closed): self
```
danharrin, line 79:
```suggestion
    public function closed(bool $condition): self
```

packages/notifications/src/Notification.php
```diff
@@ -0,0 +1,170 @@
...
+    {
+        $this->closed = $closed;
+
+        return $this;
+    }
+
+    public function getClosed(): bool
```
danharrin, line 86:
```suggestion
    public function isClosed(): bool
```

packages/notifications/src/Notification.php
```diff
@@ -0,0 +1,170 @@
...
+    public ?string $icon = null;
+
+    public ?string $status = null;
+
+    public function __construct()
+    {
+        $this->id(uniqid());
+    }
+
+    public static function make(): self
```
danharrin, line 29: All `self` types in these files can be switched to `static`

danharrin (changes requested):

packages/notifications/src/Http/Livewire/Concerns/CanNotify.php
zepfietje: Fixed in 96380d19325096b3625822e7bda43d5114c86ba1.

packages/notifications/src/Http/Livewire/Notifications.php
zepfietje: Hmm yeah, I realize this is kind of leftover dead code. We never retrieve the notification closed state, since it's set in JS based on the Livewire event. We probably do need this however, since I still have to implement notifications after redirects.

Will update for now, but great chance it'll change anyway.

zepfietje: 31e7fd8f5ce05daf52e94a3e59c6107f8216e08c

packages/notifications/src/Http/Livewire/Notifications.php
zepfietje: 31e7fd8f5ce05daf52e94a3e59c6107f8216e08c

packages/notifications/resources/views/components/notification.blade.php
zepfietje: Huh, that looks weird.

packages/notifications/resources/views/components/notification.blade.php
zepfietje: 033d13276c798d4fc78b3be68a189376726fca65

packages/notifications/resources/views/components/notification.blade.php
zepfietje: 033d13276c798d4fc78b3be68a189376726fca65

packages/notifications/src/Notification.php
zepfietje: No longer relevant: 31e7fd8f5ce05daf52e94a3e59c6107f8216e08c.

packages/notifications/src/Notification.php
zepfietje: Good one! 9a8c19fe4f3dc59125b935898b7ad6ef279f4a21

packages/notifications/src/Notification.php
zepfietje: No longer relevant: 31e7fd8f5ce05daf52e94a3e59c6107f8216e08c.

packages/notifications/src/Notification.php
zepfietje: That's what I opted for myself, until I found the current implementation used `uniqid()`. 😂

3dcedd0cf061748f32e6cb6be6c0fec812e9d8fe

packages/notifications/resources/views/components/notification.blade.php
danharrin: Perfect, blank lines between siblings

packages/notifications/src/Notification.php
```diff
@@ -0,0 +1,151 @@
...
+    public function __construct()
+    {
+        $this->id(Str::uuid());
+    }
+
+    public static function make(): static
+    {
+        $static = app(static::class);
+        $static->setUp();
+        $static->success();
```
danharrin, line 32: This configuration should be in the `setUp()` method

zepfietje: Whoops, f01b3138fcbc94dc3a12d95727ef5dd955305dcb.

danharrin:
Todo: dusk attribute https://github.com/filamentphp/filament/pull/3021/files#diff-b47e4fa0fad96d5958d55be8aa9831d27852358af348d025e33f1f99edaaecb1R56

zepfietje:
> Todo: dusk attribute https://github.com/filamentphp/filament/pull/3021/files#diff-b47e4fa0fad96d5958d55be8aa9831d27852358af348d025e33f1f99edaaecb1R56

Only for the title? The old notifications just had a message, whereas the new can have both a title and description.

ryangjchandler:
@zepfietje We can fix the PHPStan errors about `view-string` etc by adding the new service provider to the `bootstrap/app.php` file.

zepfietje:
> @zepfietje We can fix the PHPStan errors about `view-string` etc by adding the new service provider to the `bootstrap/app.php` file.

Thanks, @ryangjchandler! Fixed it. ✅

packages/admin/src/Pages/Actions/Action.php
```diff
@@ -78,9 +77,9 @@ public static function makeModalAction(string $name): ModalAction
         return $action;
     }
 
-    public function notify(string | Closure | null $status, string | Closure | null $message): void
+    public function notify(Notification $notification): void
```
zepfietje, line 80: @danharrin, this would be a breaking change, right?

zepfietje: It's also strange that it accepted closures, since `Filament::notify()` did and still does not support that.

danharrin:
I think the websockets / polling solution would come when we have database notification support, right? cause all other notifications are stored in the session

pxlrbt:
Oh, I didn't see DB support was removed from the scope.

zepfietje:
Database support has been removed from this PR indeed. Will be added in a later version and then something like polling will be introduced yeah.

NathanaelGT:
Instead of using `->send()` to "send" the notification, what if it automagically "send" using [__destruct](https://www.php.net/manual/en/language.oop5.decon.php#object.destruct)?

zepfietje:
> Instead of using `->send()` to "send" the notification, what if it automagically "send" using [__destruct](https://www.php.net/manual/en/language.oop5.decon.php#object.destruct)?

This wouldn't work.

ryangjchandler:
Why wouldn't `__destruct()` work?

zepfietje:
> Why wouldn't `__destruct()` work?

It would send all notifications you construct, even if you just assign them to a variable which is used in a conditional later to determine which notification to send. Additionally, this probably wouldn't work well with database notifications in the future.

I love a clean API, but explicitly sending might be a nice thing to have in fact.

themahdavi:
Hey @zepfietje 
This package is awesome 
NICE WORK ❤

We can use this package standalone with livewire project ?

danharrin:
We can use this package standalone with livewire project ?

yes

composer.json
```diff
@@ -68,5 +69,11 @@
         ]
     },
     "minimum-stability": "dev",
-    "prefer-stable": true
+    "prefer-stable": true,
+    "repositories": [
```
zepfietje, line 73: @danharrin does this actually need to be in here?

themahdavi:
Hello 
on question : this package break `$this->notify()` create by user on custom actions and pages ?

packages/notifications/composer.json
```diff
@@ -12,6 +12,7 @@
         "blade-ui-kit/blade-heroicons": "^1.2",
         "filament/support": "self.version",
         "illuminate/contracts": "^8.6|^9.0",
+        "illuminate/filesystem": "^8.6|^9.0",
```
zepfietje, line 15: What's this required for?

packages/notifications/resources/views/notifications.blade.php
```diff
@@ -1,6 +1,6 @@
 <div
     @class([
-        'filament-notifications pointer-events-none fixed inset-4 z-50 mx-auto flex justify-end gap-3',
+        'pointer-events-none fixed inset-4 z-50 mx-auto flex justify-end gap-3 filament-notifications',
```
zepfietje, line 3: Custom classes should be at the start according to Tailwind's ordering convention.

packages/notifications/composer.json
danharrin: install command

danharrin:
> What if people are using this currently, @danharrin?

its not documented, we only used it internally

packages/notifications/resources/views/notifications.blade.php
danharrin: Ah.

danharrin (approved):

danharrin:
🎉

alperenersoy:
Congrats @zepfietje @danharrin!

saade:
woot woot!

romaldyminaya:
Yesssssss, thank you  @zepfietje and @danharrin !!!!!!

PR #3097, opened 2022-07-12
feature: allow configuring "close on backdrop click" for modals

hxnk:
As discussed in #3089 

This stops modals from closing when clicking the backdrop.

I wasn't sure to remove the `filament-modal-close-overlay` class, so I left it in.

### DIFF
--- a/packages/support/config/filament-support.php
+++ b/packages/support/config/filament-support.php
@@ -1,0 +1,9 @@
+<?php
+
+return [
+
+    'modal' => [
+        'is_closed_by_clicking_away' => true,
+    ],
+
+];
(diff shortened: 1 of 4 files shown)


ryangjchandler:
I'm gonna say we should make this configurable... I personally find clicking outside the modal useful and this is somewhat of a functional breaking change. 

Could we maybe make it globally configurable - thoughts @danharrin?

ryangjchandler:
@danharrin Have made this configurable through config.

danharrin:
@ryangjchandler had to fix the config option, the file was the incorrect name and it wasn't registered in the service provider?

ryangjchandler:
Maybe I missed it from my commit? 😂

PR #3100, opened 2022-07-13
Vite support

Z3d0X:
Adds a 2nd argument to `registerTheme`

```diff
- public function registerTheme(string $url): void
+ public function registerTheme(string $url, bool $vite = false): void
```

### Example
```php
    Filament::registerTheme(
        'resources/css/filament.css',
        vite: true,
    );
});
```

### TODO
- [ ] update docs


ArtMin96:
maybe
```php
if ($vite) {
    $this->themeUsesVite = true;
}
```

to

```php
 $this->themeUsesVite = $vite;
```

danharrin:
Removed the extra param, now you can just do `Filament::registerTheme(app(Vite::class)('resources/css/filament.css')(`

awcodes:
This look good to me. Works as expected from a Filament standpoint.

One thing I would add is to put autoprefixer in the npm install command and the postcss.config.js file in the docs, since it is still recommended by tailwind in their postcss docs.

zepfietje (changes requested):
Haven't tested locally, but code looks good to me.

As @awcodes mentioned, autoprefixer should be added to the postcss config in the docs ([this is also how Laravel Jetstream sets this up](https://github.com/laravel/jetstream/blob/2.x/stubs/livewire/postcss.config.js)).

pxlrbt:
Tested this with the demo and works.

pxlrbt (approved):

danharrin:
Added autoprefixer, thanks

packages/admin/docs/07-appearance.md
```diff
@@ -107,7 +107,37 @@ module.exports = {
...
+                // ...
+                'resources/css/filament.css',
+            ],
+            // ...
+        }),
+    ],
+})
+```
+
+And add Tailwind to the `postcss.config.js` file:
```
zepfietje, line 129: Should we add autoprefixer to this sentence as well, @danharrin?

pxlrbt:
@danharrin Seems like this causes issues with Laravel Mix. I didn't have that issue when testing with the default stylesheet, but when registering a theme `Filament::registerTheme(mix('css/filament.css'))`. 

Apparently `mix()` returns a HtmlString,

danharrin:
Oh... right. We can fix that.

danharrin:
hmmm

danharrin:
how do we differentiate between mix and vite now...

danharrin:
fixed

danharrin:
I hate it but its fine

Screenshot 2022-07-17 at 22 42 08

pxlrbt:
Looks nasty, but I guess it works. Had the same idea about checking for `<script`

PR #3102, opened 2022-07-13
CHG: Make filament app.js defer

bernhardh:
As discussed in discord, I just added `defer` to the script-tag. The reason is, because Alpine.js recomends adding `defer` to its script include and I have some third party JS Scripts, which will not work, if Alpine isn't loaded `defer`. 

See https://github.com/wireui/wireui/issues/326

### DIFF
--- a/packages/admin/resources/views/components/layouts/base.blade.php
+++ b/packages/admin/resources/views/components/layouts/base.blade.php
@@ -98,7 +98,7 @@
 
         {{ \Filament\Facades\Filament::renderHook('scripts.start') }}
 
-        <script src="{{ route('filament.asset', [
+        <script defer src="{{ route('filament.asset', [
             'id' => Filament\get_asset_id('app.js'),
             'file' => 'app.js',
         ]) }}"></script>
--- a/packages/admin/resources/views/components/layouts/base.blade.php
+++ b/packages/admin/resources/views/components/layouts/base.blade.php
@@ -78,6 +78,8 @@
 
         {{ $slot }}
 
+        {{ \Filament\Facades\Filament::renderHook('scripts.start') }}
+
         @livewireScripts
 
         <script>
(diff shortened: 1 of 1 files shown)


danharrin:
Yeah, but I am concerned that this could break plugins as well in some way

ryangjchandler:
Deferring a script that is basically already at the end of the body is somewhat pointless, since it will only be executed milliseconds later than it is currently.

bernhardh:
@ryangjchandler I am not a js expert. I just tried it out with the code and the mentioned third party script doesn't work without defer and throws a couple of errors. Just adding it to the script tag, the code works and no error is thrown. So there has to be a difference, otherwise it wouldn't change the result. 

And when I have a look at the alpinejs installation instruction, it clearly states that you have to add defer:

> Don't forget the "defer" attribute in the <script> tag.

See https://alpinejs.dev/essentials/installation

ryangjchandler:
Yeah, defer on the CDN. We're bundling Alpine. 

What custom JS are you trying to use?

bernhardh:
I am trying to use wireui notifications (https://livewire-wireui.com/docs/notifications). Here is the corresponding issue from me: https://github.com/wireui/wireui/issues/326

And this is the situation with and without defer:

https://user-images.githubusercontent.com/642292/178781156-7136994b-2697-4806-adf1-ae8ba210675a.mp4

ryangjchandler:
Can you also provide the code where you're registering the script tag from the package?

bernhardh:
I installed wireui including the modifications on tailwind, added the scripts in the AppServiceProvider.php

```php
        Filament::registerRenderHook(
            'head.start',
            fn (): string => Blade::render("@wireUiScripts(['nonce' => 'csp-token']) "),
        );
        Filament::registerRenderHook(
            'body.start',
            fn (): string => Blade::render("<x-wireui-notifications />"),
        );
```
and created a custom page for testing:

```html
<x-filament::page>
    <div>
        <h1>Just some buttons</h1>
        <x-wireui-button primary label="Primary"></x-wireui-button>
        <x-wireui-button secondary label="Secondary"></x-wireui-button>
    </div>
    <div x-data="{ msg: 'Hallo, dies ist eine Notification. Wie get es dir?' }">
        <h1>Notifications</h1>
        <x-wireui-input class="mt-3" label="Notification-Message" x-model:value="msg" />
        <x-wireui-button secondary class="mt-3" type="button" label="Open Alert" x-on:click="window.$wireui.notify({
            title: 'Profile saved!',
            description: msg,
            icon: 'success'
        })"></x-wireui-button>
    </div>
</x-filament::page>
```

Attached you will find the resulting code (without the defer).

[Wireui - Filament Demo.txt](https://github.com/filamentphp/filament/files/9104504/Wireui.-.Filament.Demo.txt)

bernhardh:
A published the demo (based on regular filament demo) for easier bugfixing:

https://github.com/bernhardh/filamentdemo

danharrin:
I've deferred the other scripts as well

PR #3120, opened 2022-07-14
Add default value fallback for $get()

NathanaelGT:
And maybe move the $default parameter to the 2nd parameter so it's similar to `data_get()` for v3?


ryangjchandler:
Moving the `$default` parameter would be a breaking change so we can't do that. I'm sure adding an optional 3rd one is fine though.

danharrin:
Thanks, but I'm going to say no. The `default` parameters in Laravel were added before `??` was really a thing in PHP. Now, it's really easy to specify a default like `$get('field') ?? 'default'`, which is clearer to people who are familiar with PHP.

PR #3121, opened 2022-07-14
Add `laravel/pint` for code style fixer

datlechin:
Add `laravel/pint` package to support code style


danharrin:
We already have PHP CS Fixer set up, with a GitHub action. Also, we have a couple of custom styling rules. You'll need to migrate that setup over too.

zepfietje:
Wouldn't it be better to take the Laravel preset as the default and add our custom rules on top of that? That would allow us to stay up-to-date with the Laravel style and just have our own opinionated style on top.

Think that would be worth it if we're migrating over to Pint anyway.

Also, you'd probably like to remove the PHP CS Fixer setup after migrating everything to Pint.

danharrin:
Closing in favour of #3126

PR #3129, opened 2022-07-15
Add not exists testing helpers for action, bulk action, and columns

p-vinayak:


danharrin:
So these check if the actions / columns exist by checking if they're cached, I remember on Discord you wanted to find our if they were visible to the user, surely we need to call `isHidden()` instead as our assertion?

p-vinayak:
gotcha, that makes sense and i can do that.

perhaps i can add methods such as `assertColumnIsHidden()` along with `assertColumnNotExists()`?

in the "assertColumnIsHidden" method i can check for `column->isHidden()` and leave `assertColumnNotExists()` to check in cache as it is right now.

also, when i do create a method like `assertColumnIsHidden()`, i would still fetch the column using `$livewire->getCachedTableColumn($name)` correct? or is there a different way to fetch the column that exists/should be created?

danharrin:
Yeah, you can use `getCachedTableColumn()->isHidden()` fine I think. However, I don't see the reason you'd ever use `assertColumnNotExists()`? So maybe we should just remove it

p-vinayak:
So, from what it seems like `getCachedTableBulkActions()`, `getCachedTableColumn()` filters out any hidden actions/columns, so calling `isHidden()` on either of those is redundant since they always return null on hidden actions/columns. Oddly enough, though, `getCachedTableActions()` does not filter for hidden actions.

I can either add an optional parameter to those methods or I can create new methods on HasColumns/HasBulkActions that fetch hidden columns/actions. What do you think is the better approach?

Here are the files I am referring to: 
- [getCachedTableColumns()](https://github.com/filamentphp/filament/blob/2.x/packages/tables/src/Concerns/HasColumns.php#L43-L48)
- [getCachedTableActions() (no filter)](https://github.com/filamentphp/filament/blob/2.x/packages/tables/src/Concerns/HasActions.php#L157-L160)
- [getCachedTableBulkActions()](https://github.com/filamentphp/filament/blob/2.x/packages/tables/src/Concerns/HasBulkActions.php#L136-L141)

danharrin:
I see, interesting. I think we should stop filtering inside the cache method and filter inside the view instead, to only show actions that are not hidden. What do you think?

p-vinayak:
From some very quick testing, I can say for certain that removing filtering from the cache method will introduce breaking changes to filament. 

The cache method, from what I can tell, is used to determine whether a column should be rendered in the frontend. I tried removing the filter method from `getCachedTableColumn()` and the `visible()` method stopped working altogether. Meaning that columns that were hidden before, using the visible method, started showing up again.

danharrin:
If we remove the cache filtering, we can just filter in the view instead, which seems more appropriate

danharrin:
Made these changes myself, thanks.

p-vinayak:
Sorry, got busy with work and hadn't gotten time to look at this D:
Thanks for taking it up, I hope to contribute in other places around filament

PR #3131, opened 2022-07-15
Update turkish translations

deligoez:
This PR adds the missing Turkish translations and fixes problematic ones.

### DIFF
--- a/packages/admin/resources/lang/tr/global-search.php
+++ b/packages/admin/resources/lang/tr/global-search.php
@@ -7,6 +7,6 @@ return [
         'placeholder' => 'Ara',
     ],
 
-    'no_results_message' => 'Arama sonucu bulunamadı.',
+    'no_results_message' => 'Sonuç bulunamadı.',
 
 ];
--- a/packages/admin/resources/lang/tr/layout.php
+++ b/packages/admin/resources/lang/tr/layout.php
@@ -2,7 +2,7 @@
 
 return [
 
-    'direction' => 'ltr',
+    'direction' => 'soldan sağa',
 
     'buttons' => [
 
--- a/packages/admin/resources/lang/tr/login.php
+++ b/packages/admin/resources/lang/tr/login.php
@@ -4,7 +4,7 @@ return [
 
     'title' => 'Giriş yap',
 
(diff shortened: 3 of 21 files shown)


danharrin:
Thanks

zepfietje:
@deligoez, just something I noticed: shouldn't the `:label` be `:Label` when used at the start of a translation value to ensure correct capitalization?

danharrin:
oh really, I didn't know that was a feature haha

deligoez:
@zepfietje that would be quite logical, but the problem is auto-capitalization of some Turkish letters is problematic.

In Turkish;
- lowercase of the letter "I" is "ı"
- lowercase of the letter "İ" is "i"

and they are different letters.

Nonetheless I double checked and tried to use `:Label` and see the problem really exists.

So I suggest that we should leave it as it is 😁

danharrin:
Maybe there is a way to rephrase the strings to ensure correct capitalisation?

deligoez:
As far as I know, such a thing is not possible yet If you don't want to dive into Laravel's localization templating 😅

These problems can be deeper than Laravel's localization templating because even `mb_strtolower` has these problems.

https://www.php.net/manual/tr/function.mb-strtolower.php#108346

See the timestamp 😂

danharrin:
I meant that you could rephrase your Turkish maybe to ensure that the `:label` is always in a lowercase position in the string? Sorry I wasn't clear

deligoez:
Oh yes, previous translator/s had tried that in some places. Not a bad idea actually.

In English, right order of the words in a sentence: `Create :label` 
In Turkish, right order of the words in a sentence: `:label Oluştur` 

But; It is possible to write `Oluştur: :label`. That's correct and definitely a solution. But In my opinion, it seems to disrupt the reading flow a bit.

If we come to a conclusion that it is more appropriate to use them this way, I can gladly convert the translations.

zepfietje:
Hmm, I see what you mean. I think it's similar in Dutch, where generally you'd prefer to have the `:Label` first and then the verb.

It's hard for me to judge what's best in Turkish, so I trust you can make a well-founded decision!

PR #3139, opened 2022-07-17
Fix broken filterTable testing helper

austincarpenter:
Currently it's not possible to use the `filterTable()` testing helper in the way described in the docs for anything other than a basic `Filter` filter) 

[https://github.com/filamentphp/filament/blob/c104493ffabed3f10c00bd3bd7d03dc435c82aab/packages/tables/docs/06-testing.md?plain=1#L114-L130](https://github.com/filamentphp/filament/blob/c104493ffabed3f10c00bd3bd7d03dc435c82aab/packages/tables/docs/06-testing.md?plain=1#L114-L130)

This is due to a few incorrect parameter/return types referencing `Filter` instead of `BaseFilter`. This PR fixes those.

May also be worth it to document the usage of testing a `MultiSelectFilter` with an example by passing in an array, although I'll leave this one for the reviewer to decide and implement how they wish.

### DIFF
--- a/packages/tables/.stubs.php
+++ b/packages/tables/.stubs.php
@@ -33,7 +33,7 @@ namespace Livewire\Testing {
 
         public function searchTable(?string $search = null): static {}
 
-        public function filterTable(string $name, ?array $data = null): static {}
+        public function filterTable(string $name, mixed $data = null): static {}
 
         public function resetTableFilters(): static {}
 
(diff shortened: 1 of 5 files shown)


danharrin:
Thanks, I've added another test for this too

austincarpenter:
Awesome!

PR #3143, opened 2022-07-17
Add CSS classes to resource pages

pxlrbt:
Add CSS classes to target specific resouces/records. For example, if avatart for users needs a different rounding then avatar for other resources.

Should we extract that into a `getCssClasses()` on the `Page`?


danharrin:
I added the toCssClasses in a hotfix

PR #3150, opened 2022-07-18
Adds closeOnDateSelection() to date time picker

johncarter-:
👆


packages/forms/src/Components/DateTimePicker.php
```diff
@@ -30,6 +30,8 @@ class DateTimePicker extends Field
 
     protected bool | Closure $isWithoutTime = false;
 
+    protected bool | Closure $closeOnDateSelection = false;
```
zepfietje, line 33:
```suggestion
    protected bool | Closure $shouldCloseOnDateSelection = false;
```

packages/forms/src/Components/DateTimePicker.php
```diff
@@ -182,6 +184,13 @@ public function withoutTime(bool | Closure $condition = true): static
         return $this;
     }
 
+    public function closeOnDateSelection(bool | Closure $condition = true): static
+    {
+        $this->closeOnDateSelection = $condition;
```
zepfietje, line 189:
```suggestion
        $this->shouldCloseOnDateSelection = $condition;
```

packages/forms/src/Components/DateTimePicker.php
```diff
@@ -273,6 +282,11 @@ public function hasTime(): bool
         return ! $this->isWithoutTime;
     }
 
+    public function shouldCloseOnDateSelection(): bool
+    {
+        return $this->closeOnDateSelection;
```
zepfietje, line 287:
```suggestion
        return $this->evaluate($this->shouldCloseOnDateSelection);
```

zepfietje (changes requested):

zepfietje:
I've added some comments to update naming to make it similar to existing code.

`closeOnDateSelection` still feels a bit long to me though. It is clear, but would something like `closeOnSelect` also suffice? @danharrin

johncarter-:
I'll get those changes made asap. Just for reference: https://github.com/filamentphp/filament/discussions/3060#discussioncomment-3108418

zepfietje:
> I'll get those changes made asap. Just for reference: [#3060 (comment)](https://github.com/filamentphp/filament/discussions/3060#discussioncomment-3108418)

Makes sense!

danharrin:
`closeOnDateSelection()` is appropriate here because it should not close when we select a time, in which case `closeOnSelection()` is inappropriate

zepfietje:
@johncarter-, I've updated the implementation to make sure the panel also closes when selecting dates using the keyboard. This is ready to be merged if @danharrin approves.

johncarter-:
Thank you, mate.

PR #3151, opened 2022-07-18
feat: add configurable filament locale

FurkanGM:
This pr for use specific locale on filament dashboard.


packages/admin/src/FilamentManager.php
```diff
@@ -144,7 +144,7 @@ public function registerStyles(array $styles): void
         $this->styles = array_merge($this->styles, $styles);
     }
 
-    public function registerTheme(string | Htmlable | null $theme): void
```
danharrin, line 147: Please remove all formatting changes from this PR

danharrin (changes requested):

packages/admin/src/FilamentManager.php
furkangzk: done?

FurkanGM: done?

FurkanGM:
can you add config description?

packages/admin/config/filament.php
```diff
@@ -27,6 +27,16 @@
...
 
+    /*
+    |--------------------------------------------------------------------------
+    | Filament Locale
+    |--------------------------------------------------------------------------
+    |
+    | ...
+    |
+    */
+    'locale' => null,
```
zepfietje, line 38:
```suggestion

    'locale' => null,
```

zepfietje (changes requested):

packages/admin/config/filament.php
FurkanGM: done

FurkanGM:
> This PR doesn't have any effect on the actual used locale, apart from the `locale_has_pluralization` function?

Yes. But we can set app locale by filament::getLocale in provider

pxlrbt:
That's what I would expect if I set the locale. What's the use of this PR otherwise? 🤔

zepfietje:
@pxlrbt, this is exactly what I questioned as well. I've discussed it with @danharrin and we've decided to fix the use case of this PR in a different way. Dan has just opened a new pull request with that implementation #3155.

PR #3159, opened 2022-07-18
Fix overflow hidden issues with floating ui integration

awcodes:
Adds teleport option to alpine-floating-ui to fix overflow issues on tables

Usage:
x-float.teleport


danharrin:
This is such a nice solution, thank you!

PR #3166, opened 2022-07-19
fix: PageClass property nullable by default

Quadrubo:
Makes the `$pageClass` property `null` by default, to allow for the `RelationManager` getting used without being mounted on any page. Previously the `RelationManager` would break if it's being used independently from a page, by example when testing.


danharrin:
Thanks!

PR #3175, opened 2022-07-19
chore: console ui improvement added

shuvroroy:


Saifallak:
i guess this needs L9 to run,.
but filament still supporting L8

danharrin:
Yeah, unfortunately that is the case :(

PR #2869, opened 2022-06-21
feat:filament doctor

wychoong:
basic troubleshoot and fix for common inquiry

image

PR #3176, opened 2022-07-19
feature: add useful info to new `about` command

ryangjchandler:
Should be backwards compatible with the `class_exists` check.

Replaces #2869.

### DIFF
--- a/packages/support/src/SupportServiceProvider.php
+++ b/packages/support/src/SupportServiceProvider.php
@@ -2,6 +2,8 @@
 
 namespace Filament\Support;
 
+use Composer\InstalledVersions;
+use Filament\Facades\Filament;
 use Filament\Support\Testing\TestsActions;
 use HtmlSanitizer\Sanitizer;
 use HtmlSanitizer\SanitizerInterface;
@@ -11,6 +13,7 @@ use Illuminate\Support\Stringable;
 use Livewire\Testing\TestableLivewire;
 use Spatie\LaravelPackageTools\Package;
 use Spatie\LaravelPackageTools\PackageServiceProvider;
+use Illuminate\Foundation\Console\AboutCommand;
 
 class SupportServiceProvider extends PackageServiceProvider
 {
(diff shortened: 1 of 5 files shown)


packages/support/src/SupportServiceProvider.php
```diff
@@ -60,5 +62,17 @@ public function packageBooted()
...
+
+        if (class_exists(AboutCommand::class)) {
+            AboutCommand::add('Filament', [
+                'Version' => InstalledVersions::getPrettyVersion('filament/support'),
+                'Packages' => collect([
+                    'admin' => InstalledVersions::isInstalled('filament/filament'),
+                    'forms' => InstalledVersions::isInstalled('filament/forms'),
+                    'tables' => InstalledVersions::isInstalled('filament/tables'),
+                ])->filter()->keys()->join(', '),
+                'Views' => is_dir(resource_path('views/vendor/filament')) ? '<fg=red;options=bold>PUBLISHED</>' : '<fg=green;options=bold>NOT PUBLISHED</>',
```
danharrin, line 74: I think we also need to check for `forms` and `tables` and `support`

danharrin (changes requested):

danharrin:
Also you should bump the min Laravel version

pxlrbt:
> Also you should bump the min Laravel version
Don't think this is necessary:

> Should be backwards compatible with the class_exists check.

danharrin:
yeah but maybe we should still do that so more people have access to it

pxlrbt:
As we currently also support L8, I think we should keep it that way. If we were on L9 already, I guess it would be okay.

danharrin:
Thats not what I mean, we can bump the Laravel 9 *only* version so L8 users can still use it

danharrin:
It just forces current L9 users to upgrade Laravel at the same time as Filament

ryangjchandler:
I'd say leave as is.

pxlrbt:
I guess that would be fine.

zepfietje:
It would be possible that a user has different versions installed for the forms/tables packages if installed separately from `filament/filament`, right?

Not sure how often that happens, but it might be useful to show the versions per package. What do you think?

danharrin:
No, we use self.version so all packages have the same version

zepfietje:
> No, we use self.version so all packages have the same version

Even if you manually install just the forms and tables package? What about when pinning those to a specific version?

danharrin:
As far as I know there isnt a way around it

packages/support/src/SupportServiceProvider.php
danharrin: @ryangjchandler could you take a look at this when you have time please?

danharrin (changes requested):

danharrin:
Fixed it all, thanks @ryangjchandler

PR #3190, opened 2022-07-20
doc fix: update example money TextColumn

blite:
This makes the example on text column compatible with the ones on https://filamentphp.com/docs/2.x/forms/fields#text-input

Also makes it compatible with the ones on the demo app.  Longer term all these examples should expect integer money values, but that isn't really possible with one simple commit

### DIFF
--- a/packages/tables/docs/03-columns.md
+++ b/packages/tables/docs/03-columns.md
@@ -341,7 +341,7 @@ The `money()` method allows you to easily format monetary values, in any currenc
 ```php
 use Filament\Tables\Columns\TextColumn;
 
-TextColumn::make('price')->money('eur')
+TextColumn::make('price')->money('eur', true)
 ```
 
 You may `limit()` the length of the cell's value:


zepfietje (dismissed):
I think this is a good solution for now.

danharrin:
So I don't actually think there's anything wrong with the text column here, there is something wrong with the mask. Integer storage is clearly superior to the decimal format, but the masking library we use is bad, and doesn't support that.

There are a couple of solutions:
- Using the mutate functions on form pages to transform the decimal value into an integer (https://filamentphp.com/docs/2.x/admin/resources/creating-records#customizing-data-before-saving, https://filamentphp.com/docs/2.x/admin/resources/editing-records#customizing-data-before-filling-the-form, https://filamentphp.com/docs/2.x/admin/resources/editing-records#customizing-data-before-saving)
- Add a MoneyCast to all money columns, which transforms the value to and from integer and decimal in your database, and then pass `true` as the second parameter of the `TextColumn::money()` method:

```php
<?php

namespace App\Casts;

use Illuminate\Contracts\Database\Eloquent\CastsAttributes;

class MoneyCast implements CastsAttributes
{
    public function get($model, string $key, $value, array $attributes)
    {
        return intval($value) / 100;
    }

    public function set($model, string $key, $value, array $attributes)
    {
        if (blank($value)) {
            return null;
        }

        return round(floatval($value) * 100);
    }
}
```

PR #3204, opened 2022-07-22
Fix: Inability to mutate form data when using Translatable package

ciamarro:
Adds a call to mutateFormDataBeforeFill()


danharrin:
Does this need adding to `ViewRecord` too?

ciamarro:
Didn't encounter it but let me check

ciamarro:
It was indeed missing

danharrin:
Thanks

PR #3206, opened 2022-07-22
fix: relation manager table description colour

### LINKED ISSUE #3205: Relation manager table description text colour
### Package

filament/tables

### Package Version

2.13.24

### Laravel Version

9.20.0

### Livewire Version

2.10.6

### PHP Version

8.1.8

### Problem description

When adding a table description for a relations manager, the description text is hardcoded to `gray-900` in `description.blade.php`. This makes it very difficult to read when running Filament in dark mode.

I can see other parts of  Filament address this with with calls to config for `filament.dark_mode`, happy to provide you with a PR with a more appropriate colour :)

### Expected behavior

Would expect Filament to adjust the colours

### Steps to reproduce

Implementing `getTableDescription` on a relations manager.

### Reproduction repository

_No response_

### Relevant log output

_No response_

AndyTWF:
fix #3205

Adds an alternative text colour for when Filament is running in dark mode.

### DIFF
--- a/packages/tables/resources/views/components/header/description.blade.php
+++ b/packages/tables/resources/views/components/header/description.blade.php
@@ -1,4 +1,4 @@
-<p {{ $attributes->class([
+<p {{ $attributes->class(['filament-tables-header-description']) }}>
     'text-gray-900 filament-tables-header-description',
     'dark:text-gray-100' => config('filament.dark_mode')
 ]) }}>
--- a/packages/tables/resources/views/components/header/description.blade.php
+++ b/packages/tables/resources/views/components/header/description.blade.php
@@ -1,3 +1,6 @@
-<p {{ $attributes->class(['text-gray-900 filament-tables-header-description']) }}>
+<p {{ $attributes->class([
+    'text-gray-900 filament-tables-header-description',
+    'dark:text-gray-100' => config('filament.dark_mode')
+]) }}>
     {{ $slot }}
 </p>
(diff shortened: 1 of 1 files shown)


packages/tables/resources/views/components/header/description.blade.php
```diff
@@ -1,3 +1,6 @@
-<p {{ $attributes->class(['text-gray-900 filament-tables-header-description']) }}>
+<p {{ $attributes->class([
```
zepfietje, line 1: Actually we could just fall back to the default text color like the table heading does.

```suggestion
<p {{ $attributes->class(['filament-tables-header-description']) }}>
```

zepfietje (changes requested):

packages/tables/resources/views/components/header/description.blade.php
```diff
@@ -1,3 +1,6 @@
-<p {{ $attributes->class(['text-gray-900 filament-tables-header-description']) }}>
+<p {{ $attributes->class(['filament-tables-header-description']) }}>
+    'text-gray-900 filament-tables-header-description',
+    'dark:text-gray-100' => config('filament.dark_mode')
+]) }}>
```
zepfietje, line 4: Whoops, I just found out how to suggest changes for multiple lines. These lines need removal of course.

```suggestion
```

zepfietje (changes requested):

packages/tables/resources/views/components/header/description.blade.php
AndyTWF: Done! :)

zepfietje (approved):

PR #3210, opened 2022-07-23
Fix: Replicate action on EditRecord not working

blite:
Adding a replicate button on an Edit page was not working properly.  It appears the reason is that 'record' was not being passed in.

Example of how I tested this in demo app:

https://github.com/filamentphp/demo/compare/main...blite:filament-demo:main

I also caught a bug in excludeAttributes documentation.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -9,16 +9,16 @@
     "packages-dev": [
         {
             "name": "akaunting/laravel-money",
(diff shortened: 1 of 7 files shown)


danharrin:
Thanks! Ended up implementing a different fix for the missing `$record`, although yours did work

PR #2907, opened 2022-06-24
Feature: Navigation group objects

oddvalue:
This PR adds a NavigationGroup class to allow navigation groups to be collapsed by default and to add icons to them.

The previous functionality of just specifying the group name is maintained.

## Usage

Registering groups:

```php
use Filament\Facades\Filament;
use Illuminate\Support\ServiceProvider;
use Filament\Navigation\NavigationGroup;

class AppServiceProvider extends ServiceProvider
{
    public function boot(): void
    {
        Filament::serving(function () {
            Filament::registerNavigationGroups([
                NavigationGroup::make('Shop')
                    ->icon('heroicon-o-shopping-cart')
                    ->collapsed(),
                NavigationGroup::make('Blog')
                    ->icon('heroicon-o-pencil')
                    ->collapsed(),
                NavigationGroup::make('Settings')
                    ->icon('heroicon-o-cog')
                    ->collapsed(),
            ]);
        });
    }
}
```

Using groups in manual nav:

```php
use Filament\Facades\Filament;
use Filament\Navigation\NavigationBuilder;

Filament::navigation(function (NavigationBuilder $builder): NavigationBuilder {
    return $builder->group(NavigationGroup::make()
        ->label('Content')
        ->items([
            ...PageResource::getNavigationItems(),
            ...CategoryResource::getNavigationItems(),
        ]);
});
```

PR #3216, opened 2022-07-24
feature: Navigation group objects

danharrin:
See #2907 for usage


packages/admin/resources/views/components/layouts/app/sidebar/item.blade.php
```diff
@@ -48,16 +49,28 @@
...
+                        ],
+                        'warning' => [
+                            'text-warning-700 bg-warning-500/10' => ! $active,
+                            'dark:text-warning-500' => (! $active) && config('filament.dark_mode'),
+                        ],
+                        default => [
+                            'text-primary-700 bg-primary-500/20' => ! $active,
+                            'dark:text-primary-500' => (! $active) && config('filament.dark_mode'),
+                        ],
+                    },
```
pxlrbt, line 72: Could this be cleaned up by extracting this into a method? Would also allow people to add custom colors.

danharrin: This was inherited from #3218

danharrin: We could use the `default` arm to allow custom colors

danharrin: Done

packages/admin/resources/views/components/layouts/app/sidebar/index.blade.php
```diff
@@ -49,15 +49,31 @@ class="block w-full text-center"
...
+
+            $collapsedNavigationGroupLabels = collect($navigation)
+                ->filter(fn (\Filament\Navigation\NavigationGroup $group): bool => $group->isCollapsed())
+                ->map(fn (\Filament\Navigation\NavigationGroup $group): string => $group->getLabel())
+                ->values();
+        @endphp
+
+        <script>
+            if (localStorage.getItem('collapsedGroups') === null) {
+                localStorage.setItem('collapsedGroups', JSON.stringify(@js($collapsedNavigationGroupLabels)))
```
zepfietje, line 63: What's the `JSON.stringify()` for in addition to `@js()`?

packages/admin/resources/views/components/layouts/app/sidebar/item.blade.php
```diff
@@ -48,16 +49,35 @@
...
-                        'text-white bg-white/20' => $active,
-                        'dark:text-primary-500' => (! $active) && config('filament.dark_mode'),
-                    ])
-                >
-                    {{ $badge }}
-                </span>
+                <span @class([
+                    'inline-flex items-center justify-center ml-auto rtl:ml-0 rtl:mr-auto min-h-4 px-2 py-0.5 text-xs font-medium tracking-tight rounded-xl whitespace-normal',
+                    'text-white bg-white/20' => $active,
+                    ...match ($badgeColor ?? null) {
```
zepfietje, line 55: `?? null` is redundant, right?

packages/admin/resources/views/components/layouts/app/sidebar/item.blade.php
```diff
@@ -48,16 +49,35 @@
...
+                        'warning' => [
+                            'text-warning-700 bg-warning-500/10' => ! $active,
+                            'dark:text-warning-500' => (! $active) && config('filament.dark_mode'),
+                        ],
+                        'primary', null => [
+                            'text-primary-700 bg-primary-500/10' => ! $active,
+                            'dark:text-primary-500' => (! $active) && config('tables.dark_mode'),
+                        ],
+                        default => [
+                            $badgeColor => ! $active,
```
zepfietje, line 77: What does this do?

packages/admin/resources/views/components/layouts/app/sidebar/item.blade.php
zepfietje, line 55: Array unpacking with string keys is [only supported from PHP 8.1](https://wiki.php.net/rfc/array_unpacking_string_keys).

packages/admin/resources/views/components/layouts/app/sidebar/item.blade.php
```diff
@@ -48,16 +49,35 @@
...
-                    ])
-                >
-                    {{ $badge }}
-                </span>
+                <span @class([
+                    'inline-flex items-center justify-center ml-auto rtl:ml-0 rtl:mr-auto min-h-4 px-2 py-0.5 text-xs font-medium tracking-tight rounded-xl whitespace-normal',
+                    'text-white bg-white/20' => $active,
+                    ...match ($badgeColor ?? null) {
+                        'danger' => [
+                            'text-danger-700 bg-danger-500/10' => ! $active,
```
zepfietje, line 57: I feel like we should refactor this and move `! $active` outside the match expression.

packages/admin/src/Navigation/NavigationItem.php
```diff
@@ -16,19 +16,30 @@ class NavigationItem
...
     protected bool $shouldOpenUrlInNewTab = false;
 
     protected ?int $sort = null;
 
     protected string | Closure | null $url = null;
 
-    final public function __construct()
+    final public function __construct(?string $label = null)
+    {
+        if (filled($label)) {
```
zepfietje, line 29: Why the `filled` check here and not for groups?

zepfietje (changes requested):

packages/admin/resources/views/components/layouts/app/sidebar/item.blade.php
zepfietje: Hmm, custom colors would need to be hard-coded for Tailwind CSS to compile them.

zepfietje:
I haven't checked out this PR locally yet. Just reviewed the code.

packages/admin/resources/views/components/layouts/app/sidebar/item.blade.php
danharrin: Yes, the user passes in custom color classes and adds them to the tailwind config

packages/admin/resources/views/components/layouts/app/sidebar/item.blade.php
danharrin: No, because each `! $active` is color-specific

packages/admin/resources/views/components/layouts/app/sidebar/item.blade.php
danharrin: Supports custom color classes

packages/admin/resources/views/components/layouts/app/sidebar/index.blade.php
danharrin: Because `@js` passes it to JS as an object, it needs to be stringified as well to be stored in the session

packages/admin/src/Navigation/NavigationItem.php
danharrin: NavigationItem labels are required, NavigationGroup labels are nullable. It's a BC if we force users to pass a label in the NavigationItem constructor so it needs to be nullable

packages/admin/resources/views/components/layouts/app/sidebar/item.blade.php
zepfietje: Sorry if I wasn't clear, but Tailwind needs the color classes to be in the Blade view. Otherwise they aren't compiled into the production CSS.

packages/admin/resources/views/components/layouts/app/sidebar/item.blade.php
zepfietje: https://github.com/filamentphp/filament/pull/3216#discussion_r928812885

packages/admin/resources/views/components/layouts/app/sidebar/item.blade.php
zepfietje: How do you mean? They all have `! $active`, so it could be extracted, right?

packages/admin/resources/views/components/layouts/app/sidebar/item.blade.php
danharrin: I know, they need to add the resource to the tailwind.config content, so the classes from the file are added to the CSS.

zepfietje: Okay, if that's the intended route, then it's fine.

PR #3228, opened 2022-07-25
Fixed translated direction problem

alperenersoy:
The layout direction should be ltr or rtl. It is translated in Turkish translation files.


zepfietje (approved):
Good catch, @alperenersoy!

alperenersoy:
Thanks!

PR #3242, opened 2022-07-26
Added translations for ordening

sweebee:


packages/tables/resources/lang/nl/table.php
```diff
@@ -45,6 +45,14 @@
 
     'buttons' => [
 
+        'disable_reordering' => [
+            'label' => 'Ordenen afronden',
+        ],
+
+        'enable_reordering' => [
+            'label' => 'Opnieuw ordenen',
```
zepfietje, line 53: I think this could just be:
```suggestion
            'label' => 'Records ordenen',
```

zepfietje (changes requested):

packages/tables/resources/lang/nl/table.php
```diff
@@ -45,6 +45,14 @@
 
     'buttons' => [
 
+        'disable_reordering' => [
+            'label' => 'Ordenen afronden',
```
zepfietje, line 49: Not sure about `afronden` vs. `voltooien`, but I guess both are fine.

zepfietje: Maybe I do like`voltooien` better. What do you think, @sweebee?

zepfietje (approved):

PR #3254, opened 2022-07-27
allow size of the icon-button to be changed

saade:
currently icon buttons have a fixed size of `w-5 h-5` (md), this PR makes use of the `size` property that is already being passed to the component, but not used


danharrin:
Should the size of the button on line 16 also be changed?

saade:
yeah, good catch

danharrin:
Thanks!

PR #3269, opened 2022-07-27
Move upgrade command to support package

zepfietje:
Not sure if it was intended to only use the `php artisan filament:upgrade` command for the admin panel. I think it should be used for all packages.

### DIFF
--- a/packages/forms/docs/01-installation.md
+++ b/packages/forms/docs/01-installation.md
@@ -240,7 +240,6 @@ To do this automatically, we recommend adding these commands to your `composer.j
 ```json
 "post-update-cmd": [
     // ...
-    "@php artisan config:clear",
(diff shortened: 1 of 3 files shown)


pxlrbt:
The upgrade command is part of the admin package, so it can't be used if people installed only single packages.

zepfietje:
Oh sorry, didn't think about that. The command could be moved to the support package, though. But seems like it's in the admin package on purpose. 👍

danharrin:
We should move it to the support package I think

zepfietje:
> We should move it to the support package I think

Agree. Will reopen as draft and move it to support package.

zepfietje:
Closing in favor of https://github.com/filamentphp/filament/pull/3336.

PR #3272, opened 2022-07-28
add label locale

ariaieboy:
```PHP
    {
        return $form
            ->schema([
                Forms\Components\TextInput::make('name')->label('نام')
                    ->required()
                    ->maxLength(255),
                Forms\Components\TextInput::make('email')->label('ایمیل')
                    ->email()
                    ->required()
                    ->maxLength(255),
                Forms\Components\TextInput::make('password')->label('پسورد')
                    ->password()
                    ->required()
                    ->maxLength(255)
            ]);
    }

    {
        return $table
            ->columns([
                Tables\Columns\TextColumn::make('name')->label('نام'),
                Tables\Columns\TextColumn::make('email')->label('ایمیل'),
                Tables\Columns\TextColumn::make('email_verified_at')
                    ->dateTime(),
            ]);
    }
```

In projects that use another locale than EN for each Form field and Table Column, we must set the label value.
 in the example above we can see that `label('نام')` repeated twice for the column name in the table schema and field name in form schema.

imagine you have 10 resources that all of them have a field and column called `name` you should repeat `label('نام')` on each of these columns or fields. 

now imagine you decide to change the name translation from `نام` to something else. you need to go and find any label method call and change the translation. 

with this PR any Label that returns from the filament will go throw laravel localization using the `__()` function.

so now I just need to add the `name` to my translation strings and filament will return the translation for me.

since I am not that familiar with the filament core it's might be some missing label return statements but it's a good approach to add localization to the labels. and this PR can be a start point

### DIFF
--- a/packages/admin/src/Navigation/NavigationGroup.php
+++ b/packages/admin/src/Navigation/NavigationGroup.php
@@ -75,7 +75,7 @@ class NavigationGroup
 
     public function getLabel(): ?string
     {
-        return $this->label;
+        return __($this->label);
     }
 
     public function isCollapsed(): bool
--- a/packages/admin/src/Navigation/NavigationItem.php
+++ b/packages/admin/src/Navigation/NavigationItem.php
@@ -116,7 +116,7 @@ class NavigationItem
 
     public function getLabel(): string
     {
-        return $this->label;
+        return __($this->label);
     }
 
     public function getSort(): int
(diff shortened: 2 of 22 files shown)


pxlrbt:
I think we had this in V1, and it was causing issues.

ariaieboy:
> I think we had this in V1, and it was causing issues.

we can make it optional by introducing an `isLabelLocalized` flag

ariaieboy:
in https://github.com/ariaieboy/filament/tree/patch I've added a `localizeLabel()` to Table and Form components so users can control the label localization. the default behavior is not changed and only if users add this function to their field or column definition is it's gonna use a localized string instead of the default label.

PR #3273, opened 2022-07-28
add localizeLabel to Table and Form components

ariaieboy:
instead of [#3272](https://github.com/filamentphp/filament/pull/3272), I've created this PR that's introduce a new `localizeLabel()` customization to Form and Table components and users can manage the localization of each field on the form or table schema.

```php
                    ->localizeLabel()
                    ->required()
                    ->maxLength(255)
```

code above will now return the label using `__()`.

### DIFF
--- a/packages/forms/src/Components/Builder/Block.php
+++ b/packages/forms/src/Components/Builder/Block.php
@@ -50,11 +50,11 @@ class Block extends Component
     {
         $label = $this->evaluate($this->label, array_merge(
             $this->labelState ? ['state' => $this->labelState] : [],
-        ));
-
-        return $label ?? (string) Str::of($this->getName())
+        ))??(string) Str::of($this->getName())
             ->kebab()
             ->replace(['-', '_'], ' ')
             ->ucfirst();
+
+        return ($this->isLabelLocalized())?__($label):$label;
     }
 }
--- a/packages/forms/src/Components/Builder/Block.php
+++ b/packages/forms/src/Components/Builder/Block.php
@@ -50,11 +50,11 @@ class Block extends Component
     {
         $label = $this->evaluate($this->label, array_merge(
             $this->labelState ? ['state' => $this->labelState] : [],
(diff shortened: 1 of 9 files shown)


ariaieboy:
we can introduce a config value to set default localizeLabel to `true` instead of false so we don't need to add `localizeLabel()` on all components

pxlrbt:
The default should definitely be `false`. Otherwise, this would be the same result as just including it: People are coming with support requests because of it. 

This could be enabled globally with `TextColumn::configureUsing()` and similar.

zepfietje:
What's the problem with explicitly passing in a label through the translation function `__()`?

pxlrbt:
> What's the problem with explicitly passing in a label through the translation function `__()`?

From the original PR:

> imagine you have 10 resources that all of them have a field and column called name you should repeat label('نام') on each of these columns or fields.
> now imagine you decide to change the name translation from نام to something else. you need to go and find any label method call and change the translation.

ariaieboy:
> > What's the problem with explicitly passing in a label through the translation function `__()`?
> 
> From the original PR:
> 
> > imagine you have 10 resources that of them have a field and column called name you should repeat the label('نام') on each of these columns or fields.
> > now imagine you decide to change the name translation from نام to something else. you need to go and find any label method call and change the translation.

> What's the problem with explicitly passing in a label through the translation function `__()`?

🤔

I didn't want to use duplicate code all around my project but I managed to fix the problem with this piece of code:

```PHP
        Component::configureUsing(function (Component $component) {
            $component->label(fn() => __($component->getName()));
        });
        Column::configureUsing(function (Column $column) {
            $column->label(fn()=>__($column->getName()));
        });
```

but if I want to use a custom label I need to pass it like this:

```PHP
                    ->label( __('fullname'));
```

and again it's kinda duplicate code. but it's not a big problem for me anymore

pxlrbt:
Do you want to add this to the Tricks section on the Filament website? I guess that would be better.

PR #3276, opened 2022-07-28
Auto register prefix/suffix actions

DougSisk:
As mentioned in #2641, prefix/suffix actions are very clunky and requires registering the action before referencing it. This will automatically register the action if it's not already.


danharrin:
Thanks! Also simplified the manual registration syntax.

PR #3284, opened 2022-07-29
Give default items an uuid

Edsardio:
In the current version when re-ordering a repeater field it will throw the following error:

`Filament\Forms\ComponentContainer::getRawState(): Return value must be of type array, int returned`

Giving the default item an uuid fixes this.

### DIFF
--- a/packages/forms/src/Components/Repeater.php
+++ b/packages/forms/src/Components/Repeater.php
@@ -172,7 +172,7 @@ class Repeater extends Field
             }
 
             foreach (range(1, $count) as $index) {
-                $items[] = [];
+                $items[(string) Str::uuid()] = [];
             }
 
             return $items;


danharrin:
Thanks!

ariaieboy:
is there any way to get this Id for testing purposes?

after this PR some of my tests for Repeaters failed cause we use UUID instead of incremental integer

danharrin:
Sorry, may I know more context about what the test does?

ariaieboy:
> Sorry, may I know more context about what the test does?

this is my form schema :

```PHP
Forms\Components\TextInput::make('title')
                    ->required()
                    ->maxLength(125),
                Forms\Components\Select::make('type')
                    ->options(FeatureTypes::toNameArray())
                    ->required(),
                Forms\Components\Repeater::make('values')->relationship('values')->minItems(1)->schema([
                    Forms\Components\TextInput::make('value')->required()
                    ]
                )
```

before this Pr I could test the create form using this test case :
```PHP
\Pest\Livewire\livewire(CreateFeature::class)
        ->set('data.title', $newData->title)
        ->set('data.type', $newData->type->value)
        ->set('data.values.0.value', 'red')
        ->call('create')
        ->assertHasNoErrors();
```
right now i cant set the repeater value cause its use random UUID

danharrin:
I see, yeah

Maybe you could do `set('data.values', [Str::uuid() => ['value' => 'red']])`

Which will then replace the array keys of the default items, which should not be a problem

Also you could replace it all with `fillForm()` which would encourage you to replace the array as well:

https://filamentphp.com/docs/2.x/admin/testing#creating

PR #3296, opened 2022-07-29
Feature/duplicate repeater field

Edsardio:
Added an option to duplicate a record in the repeater field.
To disable the feature:
`
Repeater::make('posts')->disableItemDuplication()
`


Edsardio:
As requested in the discord changed it to cloneable, it's now an opt-in instead of an opt-out

`Repeater::make('posts')->isCloneable()`

danharrin:
Thanks

PR #3308, opened 2022-07-30
Update CreateAction.php - Reset Record After Creation

### LINKED ISSUE #3070: RelationManager Select field modifying previous record on create
### Package

filament/filament

### Package Version

v2.13.24

### Laravel Version

9.19.0

### Livewire Version

v2.10.6

### PHP Version

8.1.7

### Bug description

I have a RelationManager with a Select->relationship() field. When creating multiple records in a row, the value of the previously created record gets changed to the value that's selected for the new record. This change gets persisted to the database. If I refresh the page between creating records or use the "Create & Create Another" button, the issue doesn't happen.

After chatting with Dan in discord, he suggested adding `->saveRelationshipsUsing(null)` to the Select field which resolved the issue.

https://user-images.githubusercontent.com/46914709/178060380-21d7aac4-3920-489f-b5fb-518b2001d37f.mov

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #3307: Relation Manager: unique() callback not working unless the whole page is Refreshed after the first item
### Package

filament/forms

### Package Version

v2.15.4

### Laravel Version

v9.22.1

### Livewire Version

_No response_

### PHP Version

PHP 8.1.3

### Problem description

I was trying to add a callback to the unique() validator method in a 1-n relation manager so in order to implement the logic of enforcing uniqueness only within the current owner record. here is My code:
```php
Forms\Components\Select::make('material_id')->required()
                    ->label('Material')
                    ->searchable()
                    ->preload()
                    ->relationship('material','name')
                    ->reactive()
                    ->unique(callback: function(Component $livewire, Unique $rule) {
                        return $rule->where('recipe_id', $livewire->ownerRecord?->id);
                    }, ignoreRecord: true)
```
This Validator works fine, but still allows duplication if I create the first item then the second item without refreshing the page.
Here is a gif of the the problematic behavior:
**Screen 1: First Two items allow duplication when added back to back without refreshing the page first**
first-two-duplicate
**Screen 2: The unique constraint works fine if the page is reloaded after the first item is added**
works-after-refresh

I am inclined to believe that the root cause of this issue is somehow the same as the root cause for issue #3070.

### Expected behavior

The code should disallow duplication of the same item for the same record, even without having to refresh the page first.

### Steps to reproduce

1. Add a filament resource
2. Add items for it with a belongsTo Relationship
3. Implement the code snippet above in the belongsTo field inside the relation manager's form schema

### Reproduction repository

N/A

### Relevant log output

_No response_

coolsam726:
Ensure the relation manger record is reset to null after creation to avoid changing the previous record when creating a new one.
Fixes https://github.com/filamentphp/filament/issues/3070
Fixes https://github.com/filamentphp/filament/issues/3307

### DIFF
--- a/packages/admin/src/Pages/Concerns/HasActions.php
+++ b/packages/admin/src/Pages/Concerns/HasActions.php
@@ -62,6 +62,8 @@ trait HasActions
         } finally {
             $this->mountedAction = null;
 
+            $action->record(null);
+
             $action->resetArguments();
             $action->resetFormData();
 
--- a/packages/tables/src/Actions/CreateAction.php
+++ b/packages/tables/src/Actions/CreateAction.php
@@ -86,7 +86,7 @@ class CreateAction extends Action
 
                 return;
             }
-            $this->record(null); // Ensure the record is reset to solve issue #3070
+
             $this->success();
         });
     }
--- a/packages/tables/src/Actions/CreateAction.php
+++ b/packages/tables/src/Actions/CreateAction.php
@@ -86,7 +86,7 @@ class CreateAction extends Action
 
                 return;
             }
-
+            $this->record(null); // Ensure the record is reset to solve issue #3070
             $this->success();
         });
     }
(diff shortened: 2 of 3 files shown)


danharrin:
Thanks!

PR #3312, opened 2022-07-30
fix duplicated index issue

egyjs:
# issue :
using relation with multiple **same** related model (in morph relations  ) will cause an issue with "record-id" 
# 1
image
# 2 :
image

# fix 
using random ID:

image

### DIFF
--- a/packages/forms/src/Components/Repeater.php
+++ b/packages/forms/src/Components/Repeater.php
@@ -438,7 +438,7 @@ class Repeater extends Field
         $relatedKeyName = $relationship->getRelated()->getKeyName();
 
         return $this->cachedExistingRecords = $relationshipQuery->get()->mapWithKeys(
-            fn (Model $item): array => ["record-{$item[$relatedKeyName]}" => $item],
+            fn (Model $item): array => ["record-".Str::uuid()."-{$item[$localKeyName]}" => $item],
         );
     }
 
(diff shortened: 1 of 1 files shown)


danharrin:
I think this causes the records to be deleted and recreated every time the repeater is saved because Filament is not able to recognise the random ID as the ID of an existing record in the database? Am I correct?

If so, I think this is not the solution, and we will need to close the PR unfortunately

egyjs:
> I think this causes the records to be deleted and recreated every time the repeater is saved because Filament is not able to recognize the random ID as the ID of an existing record in the database? Am I correct?
> 
> If so, I think this is not the solution, and we will need to close the PR unfortunately

no, this not the issue

I have a relation table, which is morph table, (many to many)
first screen
Here I have (one to many) relations, each item can be related more than once
The items were retrieved in the old case as follows: 
`record-$id`,there are multiple record with the same $id , so the indexes cancel each other except last one, which is logical.

so the fix was to add a random unique string to the indexes

danharrin:
I understand the problem, I was explaining a new problem that this PR has introduced

egyjs:
> I understand the problem, I was explaining a new problem that this PR has introduced

Hmmmm Oh, I understand now.. i will try to fix this

danharrin:
Hi, I'm going to close this since the fix doesn't work. Feel free to open a new PR with a working fix.

PR #3314, opened 2022-07-30
Add support for descriptions to Text Columns

ralphjsmit:
This PR allows you to add a description above or below the label in a regular text column. I've wanted this on several occasions already and I think it should really be a core feature. It's both very useful and beautiful at the same time. 

```php
TextColumn::make('title')
    ->description(fn(Model $record) => $record->description)
```

By default, the description is shown below the actual label. However, you can choose to display the description above the label using the `->showDescriptionOnTop()` method:

```php
TextColumn::make('title')
    ->description(fn(Model $record) => $record->description)
    ->showDescriptionOnTop()
```

Example below: 

Schermafbeelding 2022-07-30 om 16 08 07

Example on top:

Schermafbeelding 2022-07-30 om 16 06 21

Looks also fine in dark mode:

Schermafbeelding 2022-07-30 om 16 11 37

Thanks!

### DIFF
--- a/packages/tables/docs/03-columns.md
+++ b/packages/tables/docs/03-columns.md
@@ -51,18 +51,13 @@ TextColumn::make('title')->label('Post title')
 
 ### Setting a description
 
-You can display a description of the column below or above the actual column text using the `description()` method. By default, the description is displayed below the column text, but you can change the position by using the `showDescriptionOnTop()` method.
+You can display a description of the column below or above the actual column text using the `description()` method. By default, the description is displayed below the column text, but you can change the position by using the `descriptionPosition()` method and passing `above` as it's parameter. Accepted values are `above` and `below`. You may also update the position by passing it as the second argument to the `description()` method. 
 
 ```php
 use Filament\Tables\Columns\TextColumn;
 
 TextColumn::make('title')
-    ->description(fn(Post $record) => $record->description);
-
-// Show the description above the actual column content:
-TextColumn::make('title')
-    ->description(fn(Post $record) => $record->description)
-    ->showDescriptionOnTop();
+    ->description(fn(Post $record) => $record->description, descriptionPosition: 'below');
 ```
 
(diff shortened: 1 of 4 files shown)


ryangjchandler (changes requested):
We should probably allow setting the position from the `description` method too, as a second argument with a default value.

ralphjsmit:
> We should probably allow setting the position from the `description` method too, as a second argument with a default value.

@ryangjchandler Added, thanks!

zepfietje:
What about renaming `description` to `sublabel`?

danharrin:
Not sure `->label()` in Ralph's original example is correct, that is changing the label of the entire column... the main cell content is fetched from the record automatically based on the column name.

zepfietje:
> Not sure `->label()` in Ralph's original example is correct, that is changing the label of the entire column... the main cell content is fetched from the record automatically based on the column name.

Yeah, I think that led me to make the bad `sublabel` suggestion. Only see what's wrong in the example now. 😅

sohaibafifi:
is it relevant to add the HtmlString type too?

danharrin:
Yes, and we should be parsing markdown & sanitising by default as well, probably

ralphjsmit:
> Yes, and we should be parsing markdown & sanitising by default as well, probably

– Added support for `HtmlString`
– Added support for parsing markdown.
– Added HTML sanitization.

ralphjsmit:
Added docs as well.

danharrin:
I think we should switch out `showDescriptionOnTop()` with `descriptionPosition('above' / 'below')`.

`description('Description', string | Closure | null position: 'above')`

`getDescriptionPosition(): ?string`

In the view: `$getDescriptionPosition() ?? 'below'`

ralphjsmit:
Done!

danharrin:
Thanks

PR #3323, opened 2022-07-31
fix: Results not found select

### LINKED ISSUE #2931: Searchable Select No Longer Showing a No Results Found Message
### Package

filament/forms

### Package Version

v2.13.10

### Laravel Version

v9.18

### Livewire Version

v2.10

### PHP Version

PHP 8.1

### Bug description

A select field using the `searchable` method is not returning a no results message when no options are found. Instead it only says `keep typing to search...`. I'm not sure when this behavior started but seems to have been introduced a few versions back.

This bug only seems to occur when the select field is using the `getSearchResultsUsing` method to query the database to find selection options. If the options are static, the no results message is displayed as expected.

### Steps to reproduce

1. Add a select field to your project with the `searchable` and `getSearchResultsUsing` methods. 
2. Try searching for an option that doesn't exists. 
3. Observe there is not a "no results message" displayed.

### Relevant log output

_No response_

danharrin:
Fixes #2931


danharrin:
This could have other potential side effects, so we need to keep an eye on new incoming issues with select searching, as this is probably the cause

PR #3324, opened 2022-07-31
fix: Admin panel registration Livewire boot

### LINKED ISSUE #3283: Livewire::component does not get called on registered resources
### Package

filament/filament

### Package Version

v2.15.3

### Laravel Version

v9.22.1

### Livewire Version

v2.10.6

### PHP Version

PHP 8.1.8

### Problem description

I am currently developing an internal plugin for Filament that has resources. In the beginning I registered the resources through the `PluginServiceProvider`, like recommended in the documentation.

Consumers of the plugin will need to override the resources to implement their own logic though, e.g. by extending the plugin resource and putting it in their app's resources folder. 

`FilamentManager` calls `array_unique` within `getResources()` but as the plugin uses it's own namespace, resources will be registered twice as the whole namespace will appear within the `$this->resources` array.

I decided to remove the registration within the `PluginServiceProvider` to let users register the resources on demand within `filament.resources.register`.

This works as expected except now the Livewire component seems to be missing as every page returns a `ComponentNotFoundException`. 

This might be because the `PluginServiceProvider` registers each of the components [by looping through the resources array.](https://github.com/filamentphp/filament/blob/2.x/packages/admin/src/PluginServiceProvider.php#L96-L108) whereas the `FilamentServiceProvider` [only loops](https://github.com/filamentphp/filament/blob/2.x/packages/admin/src/FilamentServiceProvider.php#L119) through the directory defined within the `'filament.livewire.path`.

The registering of the resources is called [somewhere else](https://github.com/filamentphp/filament/blob/2.x/packages/admin/src/FilamentServiceProvider.php#L161).

### Expected behavior

I expected that registering resources with different namespaces than `App\\Filament` (or as defined within `filament.livewire` will work.

### Steps to reproduce

Register a resource within a different namespace via `filament.resources.register` and visit the resource list records page.

### Reproduction repository

-

### Relevant log output

_No response_

danharrin:
Fixes #3283.

The aim of this PR is improve how Filament components (resources / pages / widgets) are registered.

Instead of looping through the registration directories several times, we only loop through them once if that is possible, and do all the component checks within that loop.

### DIFF
--- a/packages/admin/src/FilamentServiceProvider.php
+++ b/packages/admin/src/FilamentServiceProvider.php
@@ -11,7 +11,6 @@ use Filament\Http\Responses\Auth\Contracts\LoginResponse as LoginResponseContrac
 use Filament\Http\Responses\Auth\Contracts\LogoutResponse as LogoutResponseContract;
 use Filament\Http\Responses\Auth\LoginResponse;
 use Filament\Http\Responses\Auth\LogoutResponse;
-use Filament\Pages\Dashboard;
 use Filament\Pages\Page;
 use Filament\Resources\RelationManagers\RelationManager;
 use Filament\Resources\Resource;
@@ -19,8 +18,6 @@ use Filament\Tables\Actions\Action as TableAction;
 use Filament\Tables\Actions\ButtonAction;
 use Filament\Tables\Actions\IconButtonAction;
 use Filament\Testing\TestsPages;
-use Filament\Widgets\AccountWidget;
-use Filament\Widgets\FilamentInfoWidget;
 use Filament\Widgets\Widget;
 use Illuminate\Filesystem\Filesystem;
 use Illuminate\Support\Arr;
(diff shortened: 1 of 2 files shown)


coleshirley:
@danharrin I'm having an issue from this PR in my Multi-Context package tests. In my test suite I have modified the filament Pages, Resources, and Widget path configuration but this PR makes it so those overrides stop functioning.

I believe the cause of this is the new `registerComponent` function since it utilizes the `config` helper inside of the `packageRegistered` function. This means that components are loaded from the filesystem before my changes to the path config takes place. I believe this function should be moved into the `packageBooted` function instead so that any changes to the app config have taken place first. 

```php
public function packageBooted(): void
{
    // move the call here so that it runs as part of the boot process 
    // but before the pluginServiceProvider packageBooted function is called
    $this->registerComponents(); 

    parent::packageBooted();

    $this->bootLivewireComponents();

    $this->bootTableActionConfiguration();

    TestableLivewire::mixin(new TestsPages());
}
```

danharrin:
Yeah I've just reviewed this and I think it would also work. Do you want to PR it?

PR #3326, opened 2022-07-31
fix: Dropdown item icon size

zepfietje:


danharrin:
Screenshot?

zepfietje:
The solid icons (which should be used in this case, so that's correct) are designed to be rendered at 20px size.

## Before

Screen Shot 2022-08-01 at 10 35 28

## After

Screen Shot 2022-08-01 at 10 38 12

PR #3331, opened 2022-08-01
chore: Optimize array handling (and remove simple collections)

danharrin:


pxlrbt (approved):

packages/support/src/SupportServiceProvider.php
```diff
@@ -67,7 +67,8 @@ public function packageBooted()
                     ->filter(fn (string $package): bool => InstalledVersions::isInstalled("filament/{$package}"))
                     ->join(', '),
                 'Views' => function () use ($packages): string {
-                    $publishedViewPaths = collect($packages)->filter(fn (string $package): bool => is_dir(resource_path("views/vendor/{$package}")));
+                    $publishedViewPaths = collect($packages)
```
zepfietje, line 70: @danharrin you could use `array_filter` here.

zepfietje (changes requested):

packages/support/src/SupportServiceProvider.php
danharrin: Nah, we're using `count()` and `join()` after

zepfietje: Should be able to implode.

PR #3335, opened 2022-08-01
add configuration for default select options

dododedodonl:
Adds the possibility to globally define different options for the records per page select when using pagination.


danharrin:
I think we can stick with config here, but rename to `records_per_page_select_options`. We can't use `configureUsing()` as this is a Livewire component, not Filament.

`default_` is only used in the first config option because its changeable by the user

danharrin:
Thanks

PR #3336, opened 2022-08-01
chore: Upgrade command in support

zepfietje:


packages/forms/docs/01-installation.md
```diff
@@ -232,15 +232,23 @@ To upgrade the package to the latest version, you must run:
 ```bash
 composer update
 php artisan config:clear
+php artisan livewire:discover
+php artisan route:clear
 php artisan view:clear
 ```
 
-To do this automatically, we recommend adding these commands to your `composer.json`'s `post-update-cmd`:
+Alternatively, you may use the `filament:upgrade` command to do this all at once:
```
pxlrbt, line 240: I would only provide them `php artisan filament:upgrade`. Or do you want to keep this as information on what it does? Then I would suggest to list `php artisan filament:upgrade` first and the rest afterwards.

zepfietje: Thought about this as well. Think it's a good idea to list what it does. Also, this is probably still the best order, don't you think?

pxlrbt: My thought behind changing the order was this:

- If you just want to upgrade, do this ...
- If you care about what's happening, here is your explanation 

People are lazy 😅

zepfietje: Agree. People can always check manually what it does and running the command also gives you output. Will remove the redundant docs.

pxlrbt (approved):

danharrin:
This has broken all `make:` commands. #3343

PR #3338, opened 2022-08-01
fix production error on environments with Composer version 1

bfiessinger:
The `Composer\InstalledVersions` Class was first introduced in the composer-runtime-api 2.0.
Environments running on Composer 1 would therefore throw a Fatal Error because the class was not found.


pxlrbt (approved):
I wouldn't say that we support Composer v1, since it is EOL for almost 2 years now. As the fix is really minimal, I guess we can merge this.

PR #3345, opened 2022-08-01
hiddenOn() / visibleOn() for actions

bomshteyn:
As highlighted in discussion #3171 there is an issue where the `mountedTableAction` is set by the edit modal, and then when opening the create modal the `mountedTableAction` is still set to `edit`. This will fix this issue.

I understand that this might not be the prettiest way of accomplishing this, but we either have to merge the `mountedAction` and `mountedTableAction` or reference one in the others code...

Another alternative solution would be to under it when closing the modal, but when i tried hooking into the `close-modal` event nothing got triggered.

### DIFF
--- a/packages/admin/docs/02-resources/01-getting-started.md
+++ b/packages/admin/docs/02-resources/01-getting-started.md
@@ -139,11 +139,11 @@ To view a full list of available [layout components](../../forms/layout), see th
 
 You may also build your own completely [custom layout components](../../forms/layout#building-custom-layout-components).
 
-### Hiding components based on the page
+### Hiding components contextually
 
-The `hiddenOn()` method of form components allows you to dynamically hide fields based on the current page.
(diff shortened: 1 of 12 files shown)


danharrin:
Okay so I've changed this so that the existing `hiddenOn()` and `visibleOn()` methods will now work with actions!

Now you can use `hiddenOn('edit')`, the same as a resource page

khanakia:
@danharrin  `->hiddenOn('create')` when i use create it does not hides the fields on create page but when i use `->hiddenOn('edit')` it works perfectly on edit page.

I copied the code from this example

https://filamentphp.com/docs/2.x/admin/resources/getting-started#hiding-components-contextually

danharrin:
I've just checked, it works fine for me

khanakia:
Yes it works now after i upgraded from `v2.15.10` to `v2.15.13`

PR #3350, opened 2022-08-02
fix: Refactor select component to use array of objects for options

### LINKED ISSUE #3268: Select relationship orderBy is without function
### Package

filament/forms

### Package Version

v2.0.0

### Laravel Version

all

### Livewire Version

all

### PHP Version

all

### Problem description

Select field orders items by `$displayTitleColumn`. But as the primary key is usually an integer, the order gets lost when returning the JSON.

For example:

```php
Route::get('/test', fn () => [
    2 => 'N2',
    1 => 'N1',
]);
```

will result in:

```json
{
  "1": "N1",
  "2": "N2"
}
```

We would need to wrap every result in an array to prevent this.

### Expected behavior

Correctly ordered results.

### Steps to reproduce

`Select::make()->relationship()`

### Reproduction repository

_No response_

### Relevant log output

_No response_

pxlrbt:
Addressing the issue of non-sortable select options.

This works in cases where the options are dynamic:
- initital order of `->options()->searchable()`
- `->options()->getSearchResultsUsing()`
- `->relationship()->searchable()`.

Limitations are results that are sorted client side, for example `->options()->searchable()` **when searching**, as these results are ordered via FuseJS.

**Works for initial order**

```php
Forms\Components\Select::make('test')
    ->searchable()
    ->options([
        3 => 'First entry',
        1 => 'Second entry',
        2 => 'Third entry',
    ]),
```

**Works for initial order and search**

```php
Forms\Components\Select::make('blog_author_id')
    ->relationship('author', 'name', fn ($query) => $query->orderByDesc('id'))
    ->getOptionLabelFromRecordUsing(fn ($record) => $record->name . ' (' . $record->id . ')')
    ->searchable()
```

Fixes #3268


packages/forms/src/Components/Select.php
```diff
@@ -395,7 +395,10 @@ public function getSearchResults(string $search): array
             $results = $results->toArray();
         }
 
-        return $results;
+        return collect($results)
+            ->map(fn ($option, $key) => [$key => $option])
+            ->values()
+            ->all();
```
pxlrbt, line 401: This should be probably moved to a higher level as people might overwrite `getSearchResults()`?

MACscr:
I am assuming this would be needed for MultiSelect as well?

danharrin:
> I am assuming this would be needed for MultiSelect as well?

`MultiSelect instanceof Select === true`

packages/forms/resources/js/components/select.js
```diff
@@ -167,11 +167,15 @@ export default (Alpine) => {
...
 
             transformOptionsIntoChoices: function (options) {
-                return Object.entries(options)
-                    .map(([value, label]) => ({
-                        label,
-                        value,
-                    }))
+                return Object.values(options)
+                    .map((option) => {
+                        let [value, label] = Object.entries(option)[0]
```
saade, line 172:
```suggestion
                        const [value, label] = Object.entries(option)[0]
```

zepfietje:
@pxlrbt, what's required for this PR to move forward? 😊

pxlrbt:
@zepfietje This is only waiting on a thorough review and some feedback from @danharrin. I only set it on "draft" so it's not merged "too quickly" 😅

zepfietje:
@pxlrbt could you resolve merge conflicts?

danharrin:
I've just fixed the merge conflicts @pxlrbt, could you double check it all still works as intended before I give it a review?

packages/forms/src/Components/Select.php
danharrin: Done

pxlrbt:
Sorry, didn't have the time this week and will be on vacation the next two weeks.

PR #3355, opened 2022-08-02
Add disabled:opacity-70 to checkbox and radio

ajnsn:
Almost every form component has these disabled-modifier except a few, including radio and checkbox. 

checkbox

radio


danharrin:
Have you checked `CheckboxList`?

danharrin:
Thank you

PR #3365, opened 2022-08-03
Added content hooks to footer.blade.php

ap1969:
Added footer.start and footer.end content hooks to footer.blade.php

### DIFF
--- a/packages/admin/docs/07-appearance.md
+++ b/packages/admin/docs/07-appearance.md
@@ -295,6 +295,8 @@ The available hooks are as follows:
 - `head.end` - before `</head>`
 - `content.start` - before page content
 - `content.end` - after page content
+- `footer.start` - before footer content
+- `footer.end` - after page content
 - `sidebar.start` - before [sidebar](navigation) content
 - `sidebar.end` - after [sidebar](navigation) content
 - `scripts.start` - before scripts are defined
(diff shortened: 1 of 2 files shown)


danharrin:
Can you add the hooks to the docs too please?

ap1969:
Of course. Done!

danharrin:
Cheers

pxlrbt:
Not sure how we want to continue with the naming. Technically this is before and after the footer

danharrin:
Whats the benefit over start and end?

ryangjchandler:
Start and end would be inside of the footer in my mind - before and after would be around the actual elements / rendering.

pxlrbt:
Yep. We already had this discussion, when you wanted to rename some hooks, and make them all `.start`/`.end`

pxlrbt:
Do we want to change this before releasing?

danharrin:
> Yep. We already had this discussion, when you wanted to rename some hooks, and make them all `.start`/`.end`

This is start and end already, so what is there to change?

pxlrbt:
> This is start and end already, so what is there to change?

Yes. We had the discussion to change them all to `.start`/`.end`. But I don't think it makes sense, as there is a difference between `start` and `before`. And in this case `before` would be better suited.

danharrin:
I'll shuffle it around

PR #3368, opened 2022-08-03
Use Filament auth middleware for LW requests.

pxlrbt:
When using a custom auth guard, users are not authenticated during Livewire requests.

I expected issues with the `Login` component, but it's working fine for me. Any other cases, where this global approach could lead to issues?

Reproduction Repo: https://github.com/utsavsomaiya/demo
Create request is not authenticated.

### DIFF
--- a/packages/admin/src/FilamentServiceProvider.php
+++ b/packages/admin/src/FilamentServiceProvider.php
@@ -218,6 +218,7 @@ class FilamentServiceProvider extends PluginServiceProvider
     protected function bootLivewireComponents(): void
     {
         Livewire::addPersistentMiddleware([
+            ...config('filament.middleware.auth'),
             DispatchServingFilamentEvent::class,
             MirrorConfigToSubpackages::class,
         ]);
--- a/packages/admin/src/FilamentServiceProvider.php
+++ b/packages/admin/src/FilamentServiceProvider.php
@@ -5,6 +5,7 @@ namespace Filament;
 use Filament\Facades\Filament;
 use Filament\Http\Livewire\Auth\Login;
 use Filament\Http\Livewire\GlobalSearch;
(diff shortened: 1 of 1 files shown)


ryangjchandler:
Would we want to push the entire `auth` stack or just the `Authenticate` middleware?

pxlrbt:
I had this first, but shouldn't it be the same auth middleware for Filament and Livewire requests?

wychoong:
Or should they just update their livewire config?

https://laravel-livewire.com/docs/2.x/authorization
`If you use a different guard to authenticate your users then also add an entry to middleware_group in the livewire config file:`

danharrin:
I also suggested the same thing as Ryan on Discord. @wychoong we cant make them update the LW config, as their frontend LW site might need a different guard to Filament.

ajnsn:
> Just trying to think of unintended side-effects if somebody were to insert their own auth middleware into the config.

fyi: We are using Filament as the "frontend" on the root of the site and have set our own `Authenticate` middleware within Filament's config file. 

Basically because we are using own login redirect route because of a SingleSignOn/Socialite implementation and moreover we are not implementing the `FilamentUser`. 

We are not using a custom guard though.

Without having tested this PR personally, do you guys think this static addition of your `Authenticate` middleware to Livewire will affect us somehow? Thank you!

danharrin:
Nope, this will not affect you. Livewire persistent middleware is only applied to requests IF the middleware was also applied to the original request. So in your case, because Filament's built-in Authenticate was not applied to your request, this won't affect you.

You're probably fine, but if you are struggling to retrieve the `auth()->user()` in Livewire requests, you should also add your Authentication middleware to persistent middleware in a service provider. This has not been changed by this PR at all though.

PR #3379, opened 2022-08-04
Add title and alt attributes for iconbutton

G3z:
this adds the little native popover with the label for iconbuttons
Schermata 2022-08-04 alle 18 27 05

### DIFF
--- a/packages/support/resources/views/components/icon-button.blade.php
+++ b/packages/support/resources/views/components/icon-button.blade.php
@@ -48,6 +48,9 @@
             x-data="{}"
         @endif
         {{ $attributes->class($buttonClasses) }}
+        @if ($label)
+            title="{{ $label }}"
+        @endif
     >
         @if ($label)
             <span class="sr-only">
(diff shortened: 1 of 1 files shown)


pxlrbt:
We already have `->tooltip()`. Would this work as well? Or do you need the native version.
At least, this should only be applied if `$tooltip` is null. Otherwise they might collid.

G3z:
sorry, I didn't know about `tooltip()`.
My guess is that `title` is more accessible but I'm no expert here

danharrin:
Thanks

andrewdwallo:
Why was @pxlrbt comment not addressed here for this? I'm surprised I just noticed this as well.. When using `tooltip()` on an Action, for example, both the title and the tooltip show, which doesn't look good. 

Screenshot 2024-03-14 at 6 47 23 PM

Now anytime I use a tooltip on an icon button action, I have to do this:
```php
Tables\Actions\Action::make('markAsReviewed')
        ->label('Mark as Reviewed')
        ->iconButton()
        ->icon('heroicon-o-check-circle')
        ->tooltip('Mark as Reviewed')
        ->extraAttributes([
                'title' => '',
        ])
```

danharrin:
I think the suggestion should have been applied

andrewdwallo:
@danharrin Can I make a PR to have the title not applied when a user is using a tooltip?

danharrin:
For sure

andrewdwallo:
Will do soon. Thanks.

PR #3385, opened 2022-08-04
fixes date time picker disappearing when set to reactive

### LINKED ISSUE #3381: Reactive/lazy not properly working on DateTimePicker
### Package

filament/forms

### Package Version

v2.15.12

### Laravel Version

v9.23.0

### Livewire Version

v2.10.6

### PHP Version

8.0.20

### Problem description

When you make a DateTimePicker reactive it keeps closing on every action you make in the picker. When you select a date it closes, when you change the hour/minute it closes.

https://user-images.githubusercontent.com/10496485/182912068-39997d89-60fe-42d5-a74f-e8a5035abb95.mov

### Expected behavior

Picker shouldn't be closed when changing the date/time. It's working fine on v2.13.27 but since v2.13.28 its broken.

### Steps to reproduce

See attached repository.

```php
    {
        return [
            Forms\Components\DateTimePicker::make('date')->reactive(),
            Forms\Components\TextInput::make('title')->visible(fn(Closure $get) => $get('date')),
        ];
    }
```

### Reproduction repository

https://github.com/sweebee/filament-datetimepicker-bug

### Relevant log output

_No response_

awcodes:
Fixes #3381.

### DIFF
--- a/packages/forms/resources/views/components/date-time-picker.blade.php
+++ b/packages/forms/resources/views/components/date-time-picker.blade.php
@@ -88,9 +88,9 @@
                 x-ref="panel"
                 x-cloak
                 x-float.placement.bottom-start.offset.flip.shift="{ offset: 8 }"
-                @if (!$getStateBindingModifiers())
-                wire:ignore.self
-                wire:key="reactiveDateTimeInput-{{ $this->id }}"
+                @if (! $getStateBindingModifiers())
+                    wire:ignore.self
+                    wire:key="reactiveDateTimeInput-{{ $this->id }}"
                 @endif
                 @class([
                     'absolute hidden z-10 my-1 bg-white border border-gray-300 rounded-lg shadow-md',
--- a/packages/forms/resources/views/components/date-time-picker.blade.php
+++ b/packages/forms/resources/views/components/date-time-picker.blade.php
@@ -89,7 +89,7 @@
                 x-cloak
                 x-float.placement.bottom-start.offset.flip.shift="{ offset: 8 }"
                 wire:ignore.self
-                wire:key="dateTimePickerPanel-{{ $getStatePath() }}"
+                wire:key="{{ $this->id }}.{{ $getStatePath() }}.panel"
                 @class([
                     'absolute hidden z-10 my-1 bg-white border border-gray-300 rounded-lg shadow-md',
                     'dark:bg-gray-700 dark:border-gray-600' => config('forms.dark_mode'),
--- a/packages/forms/resources/views/components/date-time-picker.blade.php
+++ b/packages/forms/resources/views/components/date-time-picker.blade.php
@@ -90,7 +90,7 @@
                 x-float.placement.bottom-start.offset.flip.shift="{ offset: 8 }"
                 @if (! $getStateBindingModifiers())
                     wire:ignore.self
-                    wire:key="reactiveDateTimeInput-{{ $this->id }}"
+                    wire:key="dateTimePickerPanel-{{ $this->id }}"
                 @endif
                 @class([
                     'absolute hidden z-10 my-1 bg-white border border-gray-300 rounded-lg shadow-md',
--- a/packages/forms/resources/views/components/date-time-picker.blade.php
+++ b/packages/forms/resources/views/components/date-time-picker.blade.php
@@ -88,10 +88,8 @@
                 x-ref="panel"
                 x-cloak
                 x-float.placement.bottom-start.offset.flip.shift="{ offset: 8 }"
-                @unless ($isDeferred())
-                    wire:ignore.self
-                    wire:key="dateTimePickerPanel-{{ $getStatePath() }}"
-                @endif
+                wire:ignore.self
+                wire:key="dateTimePickerPanel-{{ $getStatePath() }}"
                 @class([
                     'absolute hidden z-10 my-1 bg-white border border-gray-300 rounded-lg shadow-md',
                     'dark:bg-gray-700 dark:border-gray-600' => config('forms.dark_mode'),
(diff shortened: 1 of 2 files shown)


packages/forms/resources/views/components/date-time-picker.blade.php
```diff
@@ -88,6 +88,10 @@
                 x-ref="panel"
                 x-cloak
                 x-float.placement.bottom-start.offset.flip.shift="{ offset: 8 }"
+                @if (!$getStateBindingModifiers())
+                wire:ignore.self
+                wire:key="reactiveDateTimeInput-{{ $this->id }}"
+                @endif
```
zepfietje, line 94:
```suggestion
                @if (! $getStateBindingModifiers())
                    wire:ignore.self
                    wire:key="reactiveDateTimeInput-{{ $this->id }}"
                @endif
```

zepfietje (changes requested):

packages/forms/resources/views/components/date-time-picker.blade.php
```diff
@@ -88,6 +88,10 @@
                 x-ref="panel"
                 x-cloak
                 x-float.placement.bottom-start.offset.flip.shift="{ offset: 8 }"
+                @if (! $getStateBindingModifiers())
+                    wire:ignore.self
+                    wire:key="reactiveDateTimeInput-{{ $this->id }}"
```
danharrin, line 93: What happens if there are multiple date time inputs on the same component? Because then the wire:key will be the same. Maybe use `$getStatePath()`?

danharrin (changes requested):

packages/forms/resources/views/components/date-time-picker.blade.php
awcodes: Is $this->id not the id of the component?

danharrin: It's the ID of the Livewire component, no?

danharrin: `$this` in all views is always the Livewire component

zepfietje: Hmm, but `$this->id` is used as `wire:key` in more places.

danharrin: Yup, but in both cases there's only one chance of it being included on the Livewire component at a time, you can only have 1 table at a time. So it's fine in those places

zepfietje:
Shouldn't we use `getStatePath` in other places as well?

awcodes:
I would think so. The only other place I know we had this issue was the Filters on the table. that is using $this->id

awcodes:
Togglable in Tables is using $this->id too.

Do you want me to update those two as part of this commit?

awcodes:
Never mind. Those won't have a statePath.

zepfietje:
> Never mind. Those won't have a statePath.

Yeah just wanted to comment this. But do they work fine with `$this->id`?

awcodes:
They work ok as far as I know. My concern with it initially was if there were multiple tables on the page i.e. Multiple Relation Managers, etc.

I'm not sure what else could be used there though to make it unique.

danharrin:
I'm interested, is `wire:key` scoped to the current Livewire component, or the entire page?

danharrin:
I always assumed it was just the Livewire component (much easier to handle than the page) but maybe I'm wrong

awcodes:
According to the docs "the value to wire:key must be unique across the page".

https://laravel-livewire.com/docs/2.x/troubleshooting#dom-diffing-cures

awcodes:
The value you pass to wire:key must be entirely unique to that page. Meaning that you should prefix it, like wire:key="item-{{ $item->id }}", and avoid using $loop->index to track the individual elements where you can.

danharrin:
Thanks - I will now be prefixing every single wire:key with a Livewire component ID in the future 😂

zepfietje (approved):

zepfietje:
So this PR is now ready to be merged, @danharrin?

awcodes:
Do the binding attributes allow for denounce? Is so we might need modify the conditional since it’s only checking for an empty array. Sorry just though about this.

zepfietje:
> Do the binding attributes allow for debounce? Is so we might need modify the conditional since it’s only checking for an empty array. Sorry just though about this.

Yeah, check `HasStateBindingModifiers.php`. In what cases do we not want to add the `wire:key` by the way?

awcodes:
If it’s empty of has ‘debounce’ then it would need to be ignored. Other cases are ok since they don’t send a livewire request.

awcodes:
I can revisit it in the morning unless you feel like taking care of it.

zepfietje:
> If it’s empty of has ‘debounce’ then it would need to be ignored. Other cases are ok since they don’t send a livewire request.

So only don't add it if it's `['defer']`.

zepfietje:
> I can revisit it in the morning unless you feel like taking care of it.

I've updated the implementation already.

However, I found another issue: when using `debounce()` with the `DatePicker`, the input shows `undefined` as value.

awcodes:
honestly we may be able to just use ignore and key without the check all together. 

The undefined value sounds like an unrelated issue, though.

danharrin:
DatePicker doesnt support debounce, as `$wire.entangle()` doesn't support debounce. So we can get rid of that handling.

zepfietje:
> honestly we may be able to just use ignore and key without the check all together.

Yeah I think that should actually be fine.

packages/forms/src/Concerns/HasStateBindingModifiers.php
```diff
@@ -68,7 +68,7 @@ public function getStateBindingModifiers(): array
 
     public function isLazy(): bool
     {
-        return in_array('lazy', $this->getStateBindingModifiers(), false);
+        return in_array('lazy', $this->getStateBindingModifiers());
```
zepfietje, line 71: I've kept this change, as `false` is the default for the `$strict` parameter.

PR #3394, opened 2022-08-05
Separate user-avatar

ajnsn:
Uses the same component in `user-menu` and `account-widget`

If you do not have images or do not want to use a 3rd party service (Greetings from the GDPR), this simplifies overriding the avatar when you do not want the background-image-`getUserAvatarUrl()` thing.


danharrin:
On the GDPR side of things, I think our current avatar generator is fine, as we only send the initials of the user :)

ajnsn:
You are always sending the IP address without prior consent. Not a lawyer but we're are careful here.

But my particular use case is that I currently need to override the whole user-menu (with all the Alpine stuff) just to use a string-based approach :)

PR #3402, opened 2022-08-06
ColorPicker fixes

sweebee:
**Problem**
When calling ```$this->form->getState();``` the ColorPicker value becomes empty. Also when reactive, the panel gets hidden.

https://user-images.githubusercontent.com/10496485/183253012-04f87993-82f1-4c4c-a105-1b1f1031fe9b.mov

**Fix**

Added the wire:model attribute. and added an ignore to the colorpalette panel.

### DIFF
--- a/packages/forms/resources/views/components/color-picker.blade.php
+++ b/packages/forms/resources/views/components/color-picker.blade.php
@@ -45,6 +45,7 @@
                 type="text"
                 dusk="filament.forms.{{ $getStatePath() }}"
                 id="{{ $getId() }}"
+                {{ $applyStateBindingModifiers('wire:model') }}="{{ $getStatePath() }}"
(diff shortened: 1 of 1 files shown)


packages/forms/resources/views/components/color-picker.blade.php
```diff
@@ -45,6 +45,7 @@
                 type="text"
                 dusk="filament.forms.{{ $getStatePath() }}"
                 id="{{ $getId() }}"
+                {{ $applyStateBindingModifiers('wire:model') }}="{{ $getStatePath() }}"
```
danharrin, line 48: How about swapping this line for `x-model="state"`?

danharrin (changes requested):

PR #3405, opened 2022-08-06
Create CreatedDateRangeFilter.php

ziming:
Feel free to make it into a more generic DateRangeFilter that is not just for created_at

### DIFF
--- a/packages/tables/src/Filters/CreatedDateRangeFilter.php
+++ b/packages/tables/src/Filters/CreatedDateRangeFilter.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace Filament\Tables\Filters;
+
+use Filament\Forms\Components\DatePicker;
+use Filament\Tables\Filters\Filter;
+use Illuminate\Database\Eloquent\Builder;
+
+class CreatedDateRangeFilter extends Filter
+{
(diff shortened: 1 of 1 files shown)


danharrin:
https://github.com/filamentphp/filament/pull/3406#discussion_r939550277

danharrin:
https://github.com/filamentphp/filament/pull/3406#discussion_r939626881

ziming:
Hi @danharrin 

i have made the DateRangeFilter class more generic now. Would this still be better outside core?

```php
class DateRangeFilter extends Filter
{
    public static function make(?string $name = 'created_at'): static
    {
        return parent::make($name)
            ->form([
                DatePicker::make($name . '_from'),
                DatePicker::make($name . '_until'),
            ])
            ->query(function (Builder $query, array $data) use ($name): Builder {
                return $query
                    ->when(
                        $data[$name . '_from'],
                        fn (Builder $query, $date): Builder => $query->whereDate($name, '>=', $date),
                    )
                    ->when(
                        $data[$name . '_until'],
                        fn (Builder $query, $date): Builder => $query->whereDate($name, '<=', $date),
                    );
            });
    }
}
```

danharrin:
It would still need to handle relationships etc, same as the SelectFilter

PR #3406, opened 2022-08-06
Create TextFilter.php

ziming:
A generic text filter

### DIFF
--- a/packages/tables/src/Filters/TextFilter.php
+++ b/packages/tables/src/Filters/TextFilter.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Filament\Tables\Filters;
+
+use Filament\Forms\Components\TextInput;
+use Filament\Tables\Filters\Filter;
+use Illuminate\Database\Eloquent\Builder;
+
+class TextFilter extends Filter
+{
+    public static function make(?string $name = null): static
+    {
+        return parent::make($name)
+            ->form([TextInput::make($name)])
+            ->query(
+                fn (Builder $query, array $data): Builder => $query->when(
+                    $data[$name],
+                    fn (Builder $query, $date): Builder => $query->where($name, $data[$name]),
+                )
+            );
+    }
+}


packages/tables/src/Filters/TextFilter.php
```diff
@@ -0,0 +1,22 @@
...
+class TextFilter extends Filter
+{
+    public static function make(?string $name = null): static
+    {
+        return parent::make($name)
+            ->form([TextInput::make($name)])
+            ->query(
+                fn (Builder $query, array $data): Builder => $query->when(
+                    $data[$name],
+                    fn (Builder $query, $date): Builder => $query->where($name, $data[$name]),
```
danharrin, line 18: Isn't this more useful as a `like` query? Like a per-column searchable filter

Also, it should work with relationship `dot.syntax` like our SelectFilter does

danharrin (changes requested):

packages/tables/src/Filters/TextFilter.php
ziming: Hmm how about renaming this to ExactTextFilter instead?

For me if i want LIKE match i would use the search bar instead of filter

danharrin: I feel like the implementation is too subjective, same with #3405. What one person's idea of a text filter is might differ from someone else, and it's hard to strike a balance. I think it's probably much easier if you put this in a package or something, or just within your own code and reuse it. Then people can choose whether they want to use it or not, or built their own. Sound good?

ziming: hmm so do I close these 2 PRs or wait for somebody else to suggest an approach to these 2 filters that is more suitable to be merged into core?

danharrin: So I think that because the implementations are subjective, it's better for the community that they remain as plugins, and then if your version doesnt suit some people they can release a different plugin with different behaviour. If it turns out that many people use these plugins when they are released, I'll consider merging the most popular version into the core, if that is going to benefit alot of people. Thank you for the PRs.

PR #3410, opened 2022-08-06
fix: wire:key consistency

### LINKED ISSUE #3000: The "category" field breaks after changing the payment method
When I change the plugin type from **Not paid** to **Paid** and return to **Not paid** the `category` field breaks, take a look to the demo video below.

https://user-images.githubusercontent.com/6197875/176866253-c1246f49-61bf-40bd-8d9f-8f7b0882d8a2.mp4

danharrin:
Fixes #3000.

### DIFF
--- a/packages/tables/src/Testing/TestsRecords.php
+++ b/packages/tables/src/Testing/TestsRecords.php
@@ -27,7 +27,7 @@ class TestsRecords
                         $record = $livewire->getTableRecordKey($record);
                     }
 
-                    return "wire:key=\"table.records.{$record}\"";
+                    return "{$livewire->id}.table.records.{$record}";
                 },
                 $records instanceof Collection ? $records->all() : $records,
             );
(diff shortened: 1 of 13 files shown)


packages/tables/src/Testing/TestsRecords.php
```diff
@@ -27,7 +27,7 @@ function ($record) use ($livewire) {
                         $record = $livewire->getTableRecordKey($record);
                     }
 
-                    return "wire:key=\"table.records.{$record}\"";
+                    return "{$livewire->id}.table.records.{$record}";
```
Saifallak, line 30: @danharrin isn't this wrong?, where is wire:key=xxxx

danharrin: No this isn't an issue, I want to decouple the tests from this specific attribute in the view. It still works fine for you, right?

Saifallak: Didn't have a chance to test,..
Just thought it was a mistypo to remove wirekey.

Issue #2969, opened 2022-06-29
DatePicker timezone issue


invaders-xx:
### Package

filament/filament

### Package Version

v2.13.12

### Laravel Version

v9.19.0

### Livewire Version

v2.10.5

### PHP Version

PHP 8

### Bug description

my app uses UTC as timezone

When I create a DatePicker, the date value returned is the day before the date selected when the field's attribute timezone is set to user's timezone ('Europe/Zurich') in my case or not set.
It works when this attribute is set to 'UTC'.
Below the form:
<img width="1283" alt="image" src="https://user-images.githubusercontent.com/604907/176412381-dd3d5145-6bf7-468a-bcc2-4ffe9be1d7ba.png">
Below the data : 
<img width="198" alt="image" src="https://user-images.githubusercontent.com/604907/176412589-d755967d-1bb5-4749-85b0-13fb0b426a23.png">

Saifallak:
maybe this helps ? https://filamentphp.com/tricks/multiple-user-timezones

zepfietje:
Can you provide us a reproduction repository URL and then we'll look at your issue?

That would allow us to download it and review your bug much easier, so it can be fixed quicker. Please make sure to give us some instructions on where to find the bug, and include a database seeder with everything we need to set it up quickly.

sweebee:
The timezone must be ignored by the picker if there is no time to select. Or just don't add the timezone attribute when it's a date only field.

danharrin:
@sweebee is right, the `timezone()` method is not appropriate for use on a date-only input - only time.

What you experienced is the desired behaviour of the date picker in this case - converting midnight Zurich time to UTC, which is a day before.

PR #3414, opened 2022-08-07
#2969: Ignore the timezone on a date only input.

sweebee:
The DatePicker just creates a datetime with ```2022-06-04 00:00:00``` (Europe/Amsterdam) and converts it to UTC which is ```2022-06-03 22:00:00,``` and then converts it to a date only ```2022-06-03```. This doesnt make any sense when you want a date only so the timezone must be ignored.

#2969

### DIFF
--- a/packages/forms/src/Components/DatePicker.php
+++ b/packages/forms/src/Components/DatePicker.php
@@ -2,6 +2,8 @@
 
 namespace Filament\Forms\Components;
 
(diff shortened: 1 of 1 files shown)


danharrin:
Hey, I don't think this is the correct solution - people should just not be using the timezone() method with the Date Picker. I don't think it fixes #2969.

PR #3424, opened 2022-08-08
Add notify event listener to `CanNotify` trait

bfiessinger:
This PR adds the ability to create Filament Notifications directly in blade files using Livewire Events.

This helps if someone wants to notify from child components (or even simple blade components).

Example:
```html
<!-- resources/views/code-highlighter.blade.php -->
<div
    x-data="{
        copy_data: $el.querySelector('code').innerText,
    }"
>
    <code>...</code>
    <button
            @click="$clipboard(copy_data)"
            wire:click="$emitUp('notify', 'success', 'Code copied to clipboard')"
    >
            <x-dynamic-component component="heroicon-o-duplicate" />
    </button>
</div>
```

### DIFF
--- a/packages/admin/src/Http/Livewire/Concerns/CanNotify.php
+++ b/packages/admin/src/Http/Livewire/Concerns/CanNotify.php
@@ -9,6 +9,13 @@ use Filament\Facades\Filament;
  */
 trait CanNotify
 {
+    protected function getListeners()
+    {
+        return array_merge($this->listeners, [
+            'notify' => 'notify',
+        ]);
+    }
+    
     public function notify(string $status, string $message, bool $isAfterRedirect = false): void
     {
         Filament::notify($status, $message);


zepfietje:
Thanks for opening a PR, @bfiessinger!

Note, the class you've added the listener to is deprecated.

Also, I'm not sure if we should introduce this functionality.

> This helps if someone wants to notify from child components

Currently, it's already possible to send notifications from anywhere in your application (not just Livewire components). So calling `Notification::make()->send()` from child components should also work just fine.

Additionally, I do see the use case of sending notifications from Blade. However, you could just call a Livewire method instead of emitting an event. Then you can send the notification from the method inside your Livewire component. Honestly, I think this is the cleaner way, as supporting the entire notifications API in a `notify` event from Blade will be rather messy. I'm using the method approach myself and don't feel the need for another approach.

What do you think about calling a method instead of sending an event, @bfiessinger?

Also, @danharrin and @pxlrbt, I'm curious about your opinions.

ryangjchandler:
I'd initially say that while there are probably a few use cases for this, the better way is to have a dedicated method and like @zepfietje said, send the Notification from there. With the event system, it'd be almost impossible to support the entire notification API so it kind of seems pointless if a few extra lines of code gives you something more powerful.

bfiessinger:
@ryangjchandler @danharrin Thanks for checking my PR. I do agree a dedicated method is for sure the better approach to achieve sending notifications and of course I was aware of the traits deprecation.

> Additionally, I do see the use case of sending notifications from Blade. However, you could just call a Livewire method instead of emitting an event. Then you can send the notification from the method inside your Livewire component. Honestly, I think this is the cleaner way, as supporting the entire notifications API in a notify event from Blade will be rather messy. I'm using the method approach myself and don't feel the need for another approach.

Yes, sending notifications from Blade was why I thought this could be interesting. And in my case the component that had need to send notifications was no Livewire component. But turning such things into Livewire Components is for sure no witchcraft. That being said I agree with not merging it.

Anyways thanks for checking.

danharrin:
@bfiessinger we've decided to implement this, so I opened a PR - #4239

danharrin:
This has been implemented in #4444

PR #3435, opened 2022-08-08
Move filesystem component registration to boot process in app lifecycle

coleshirley:
See the comment here: https://github.com/filamentphp/filament/pull/3324#issuecomment-1205603433

This PR moves the registration of components in the FilamentServiceProvider into the app Booting event. This allows all changes to the app configuration to take place inside of a test suite before Filament registers components from the filesystem.

For an example of how this works see the test suite here: https://github.com/coleshirley/filament-package-tests-lifecycle-example

```php
// Tests/TestCase.php

public function getEnvironmentSetUp($app)
    {
        config()->set('database.default', 'testing');

        config()->set('filament.resources', [
            'namespace' => 'Coleshirley\FilamentPackageTestsLifecycleExample\Tests\\Filament\\Resources',
            'path' => __DIR__ . '/Filament/Resources',
            'register' => [],
        ]);
    }
```

```php
it('can register components for tests', function () {
    expect(Filament::getResources())->toContain(UserResource::class);
});
```

The above test fails because the `registerComponents` function runs before the `TestCase` alters the filament configuration.

My proposed changes move the `registerComponents` function to the very beginning of the boot process of the application. This ensures that the components are registered before the filament service will be resolved but after any changes to the configuration are made.

I initially considered adding this to the `resolving` callback of the filament service but that would slow down apps with Octane enabled significantly since each request would cause the filesystem to be searched again.

I avoided moving this to the `bootingPackage` function in the service provider because the situation could arise where another `ServiceProvider` class could boot prior to filament and resolve the filament service prior to any filesystem components being registered.

So effectively this puts the `registerComponents` function call after the registration of all service providers but before the boot of all service providers in the app


danharrin:
Thanks

PR #3439, opened 2022-08-09
fix: Swap toArray() with attributesToArray()

danharrin:


theVannu:
Hello @danharrin, we have some problem with this fix. I try to explain.
Consider Resource file (for example UserResource) and suppose to implement getEloquentQuery() using eager loading

`User::with('post')`

Before the fix, when overriding method mutateFormDataBeforeFill(array $data), $data variable contains also eager loaded data (post relationship).

After the fix my variable $data doesn't have eager loaded data anymore, only user datas are present.

Is this the correct behaviour?
Thank you

danharrin:
Yes, that is the correct behaviour. Relationship data has always been accessible from `$this->record`, and I have always recommended that approach to others before this change was made. I wasnt even aware that the relationships were loaded before this was fixed 😂

theVannu:
Ok great, thank you for this clarification.

PR #3441, opened 2022-08-09
fix: Mask watcher

### LINKED ISSUE #3049: Numeric masks not working as intended
### Package

filament/forms

### Package Version

v2.13.9

### Laravel Version

v9.17.0

### Livewire Version

v2.10.5

### PHP Version

PHP 8.1.6

### Bug description

Masked numeric fields seem not to behave as intended via https://imask.js.org/guide.html#masked-number

Testing in a custom file as per the documentation, all works fine. Upon entering a number and then tabbing out, the value is being formatted correctly.

```
IMask(document.getElementById('test'), {
    mask: Number,
    padFractionalZeros: true,
});
```

working

When trying to replicate this behavior in a filament `TextInput` the field breaks. Immediately after typing a single digit, the field is being formatted. It is then not possible to move into the decimals by pressing comma/dot/tab but by either using the right arrow key or the mouse. Even if I do enter the decimals, it is not possible to override any of them. 

```
Components\TextInput::make('amount')
    ->mask(
        fn(Mask $mask) => $mask
            ->numeric()
            ->padFractionalZeros()
    ),
```
not_working

### Steps to reproduce

_No response_

### Relevant log output

_No response_

danharrin:
Fixes #3049.


fayne1337:
Thanks @danharrin for your work on this. I've just upgraded my project to the latest version (v2.15.16), however I still experience the same issue as per #3049. Is there any specific methods I need to add to my `TextInput` mask?

danharrin:
Have you recompiled your assets & cleared your view cache? I assume you're using the form builder and not the admin panel

fayne1337:
Those are the steps I've done:
`composer upgrade`
`php artisan filament:upgrade`
`php artisan cache:clear`

I don't use any custom JS but recompiled my assets just in case.
I am using the admin panel, the TextInput component is loaded normally via form schema.

danharrin:
I see, I wonder how I got this working ok yesterday...

danharrin:
Just released a hotfix for this, can you test out v2.15.17?

fayne1337:
It's working now, many thanks!

PR #3452, opened 2022-08-10
fix: Use Livewire ID in action modal ID

danharrin:


invaders-xx:
@danharrin many thanks this PR solved the issue with multiple tables in a Pages. I have the issue now on this page where I have a page's action using a modal and I have this behavior : 

image

The whole background is black and when I click on Cancel, I have this: 
image

And I need to refresh the page.

danharrin:
Please open another issue for that with a reproduction repo

PR #3453, opened 2022-08-10
fix: Hydrate state after loading from relationship

### LINKED ISSUE #3451: FileUpload does not work unless `->multiple()`
### Package

filament/filament

### Package Version

v2.15.16

### Laravel Version

v9.24.0

### Livewire Version

v2.10.7

### PHP Version

v8.1.6

### Problem description

`FileUpload` does not work unless being `->multiple()`.

Without being `->multiple()`, the file url gets saved as string on the database, and the component fails to parse the state because it expects a array but a string is passed.

### Expected behavior

The file should be saved as an array, as per Dan's reply [here](https://discord.com/channels/883083792112300104/883085278993068082/1006802122517516328)

### Steps to reproduce

1. Create a `FileUpload` without it being `->multiple()`
2. Create the resource (submit the form).
3. Try to edit the form by changing the image or just by clicking 'Save'. It will fail.

### Reproduction repository

https://github.com/saade/filament-issue-reproduction-repo/tree/issue/3451

### Relevant log output

```shell
https://flareapp.io/share/NPLKGw3m#F61
```

danharrin:
Fixes #3451.

### DIFF
--- a/packages/forms/src/Components/Concerns/BelongsToModel.php
+++ b/packages/forms/src/Components/Concerns/BelongsToModel.php
@@ -44,7 +44,15 @@ trait BelongsToModel
         if ($andHydrate) {
             $this->callAfterStateHydrated();
 
+            foreach ($this->getChildComponentContainers() as $container) {
+                $container->callAfterStateHydrated();
+            }
+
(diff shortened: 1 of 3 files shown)


saade:
thank you!

PR #3461, opened 2022-08-10
Allow javascript callback for chart options

margarizaldi:
The idea is by wrapping javascript callback inside `!!` then we remove the rendered `"!!` and `!!"` in the frontend.

For example:

```
public static ?array $options = [
    'scales' => [
        'x' => [
            'ticks' => [
                'callback' => "!!
                    function(value, index, ticks) {
                        return '$' + value;
                    }
               !!"
            ],
        ],
    ],
];
```

I know the code is ugly but if someone can make better PR please do it. Chart.js brings a lot of functionalities through callbacks.

Anyone can improve if `!!` or create better PR if this is not acceptable.

### DIFF
--- a/packages/admin/resources/views/widgets/chart-widget.blade.php
+++ b/packages/admin/resources/views/widgets/chart-widget.blade.php
@@ -49,7 +49,7 @@
                         return this.chart = new Chart($el, {
                             type: '{{ $this->getType() }}',
                             data: this.applyColorToData(data),
-                            options: {{ json_encode($this->getOptions()) }} ?? {},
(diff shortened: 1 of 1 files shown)


margarizaldi:
Well I don't expect too much that this PR will be accepted. At least the idea must go on.

danharrin:
Yeah I'm not going to merge this solution, but I can suggest another that you can implement.

Allow the `getOptions()` method to return a string instead, and then check if it's a string or an array before passing it to `json_encode()`. That way the user can provide their own pre-encoded JSON with callbacks and everything, that gets passed directly to the view.

PR #3472, opened 2022-08-11
Add loading indicator trait to text input component

Random72IsTaken:
Greetings,

Just couldn't resist customizing my text-input component with a Livewire loading indicator. And I thought it might be worth sharing and made it optional.

Here is how it looks:

https://user-images.githubusercontent.com/81492351/184070631-c41ccc58-29f5-40db-9abf-662ffa24e7a4.mp4

### DIFF
--- a/packages/forms/config/forms.php
+++ b/packages/forms/config/forms.php
@@ -14,14 +14,6 @@ return [
 
     'components' => [
 
-        'text_input' => [
-
-            'lazy' => [
-                'loading_indicator' => false,
-            ],
-
-        ],
-
         'actions' => [
 
             'modal' => [
--- a/packages/forms/config/forms.php
+++ b/packages/forms/config/forms.php
@@ -14,6 +14,14 @@ return [
 
     'components' => [
 
+        'text_input' => [
+
+            'lazy' => [
+                'loading_indicator' => false,
+            ],
+
+        ],
+
         'actions' => [
 
             'modal' => [
--- a/packages/forms/resources/views/components/text-input.blade.php
+++ b/packages/forms/resources/views/components/text-input.blade.php
@@ -6,6 +6,8 @@
         'text-gray-400' => ! $errors->has($getStatePath()),
         'text-danger-400' => $errors->has($getStatePath()),
     ];
+
+    $hasLazyLoadingIndicator = $isLazy() && config('forms.components.text_input.lazy.loading_indicator');
 @endphp
 
 <x-dynamic-component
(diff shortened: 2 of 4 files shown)


GoodM4ven:
@danharrin Cannot figure out how to make it `blur` compatible or something... So it only works now upon `change` (`lazy()`). 🥲

The `showValidity` state is important to bypass the first without errors, I thought! Yet, again, I have no clue how everything works in here; can you please teach me? 😅

danharrin:
Ah okay, I kinda misunderstood then. I didn't think it was just for displaying errors, I'm sorry. Since we don't really support realtime validation, I'm not sure if this feature fits in with the rest of our Filament features, until we properly implement realtime validation (which I don't know if we will yet). Please make it a plugin though, as it would be quite easy (just a custom $view for the existing TextInput), and I'm sure people will find it useful.

GoodM4ven:
Hey, @danharrin . Just a quick inquiry here...

I have noticed that the state for showing the visual error stuff is accounted for somewhere (the `danger` border). And unlike what I understood from the code, it doesn't work on `blur` somehow, which exactly what my `showValidity` state was for!

So all that is left is just pointing to where all of that takes place, and that's it, don't you think?! **Because that danger-border works well with `reactive`, `debounce`, and lazy altogether.** So I thought all I need was to understand how it works and where that styling is, and so I can add this logic of mine right there and we're done! :)

danharrin:
So we actually havent implemented anything special for checking validation errors, we just use Livewires `$errors` bag to check for an error on that field! Not sure why you wouldnt be able to hook into that, unless we use a wire:ignore somewhere, which could prevent anything inside that element from receiving validation errors?

GoodM4ven:
I thought about it. The current situation is styling when `$errors->has()` or not. But if we want to style when things fail first, we need to keep errors history...

Laravel doesn't care about modifying its errors [`MessageBag`](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Support/MessageBag.php) for something real-time like this, AND Livewire uses exactly that without extending it. So I thought that the only place to add such logic is in Livewire's [`ValidatesInput`](https://github.com/livewire/livewire/blob/master/src/ComponentConcerns/ValidatesInput.php) component concern, maybe. But even there, we're dealing with the entire component still, not the input field!

So I figured we could use another `$errorHistoryBag` next to the original `$errorBag`, which synchronizes with it for **additions only** and therefore keeping track of whatever that *had failed once*, I guess.

Finally, and hopefully, we can use something like:
- `$errorsHistory->had($getStatePath()) && $errors->has($getStatePath())`
- `$errorsHistory->had($getStatePath()) && ! $errors->has($getStatePath())`

What do you think, @danharrin ? 🙂

-But, before you ask, am I a chad enough to do the PR? No! I literally started learning OOP & Laravel 6 months ago. Yet I hope I could help more in the near future. 🥲

danharrin:
Does #3069 help you a bit maybe?

GoodM4ven:
> Does #3069 help you a bit maybe?

No, @danharrin . Listen, take a look at all of the following styling or at the default styling in text-input field, for instance.

image

You can clearly see that we have **only 2 styles, not 3** like it should; in my opinion!
- The default style **AND** when there's no errors.
- The `danger` style for errors.
- **The `success` style is missing!** Because we're not able to tell that there "was" an error, and now it was corrected, so display the successful (3rd) style!

I think I'm a bit more determined to make the PR to Livewire today.

PR #3476, opened 2022-08-11
add note for Enum column in resource --generate

wychoong:


ryangjchandler:
Just needs some grammar changes here, but should definitely be in the docs.

PR #3485, opened 2022-08-12
Add support for custom table row classes based on the record

marcoboers:
This PR adds the ability to dynamically add classes to a table row based on the record.  
It therefore provides a single method to alter for example text-color, background and borders on a table row.

It supports a flexible and even conditional return type:

```php
protected function getTableRecordClassUsing(): ?Closure
    {
        return function (Post $record) {
            return match ($record->status) {
                'draft' => 'opacity-30',
                'reviewing' => [
                    'border-l-solid',
                    'border-l-2',
                    'border-l-orange-600',
                    'dark:border-l-orange-300' => config('filament.dark_mode'),
                    'opacity-30',
                ],
                'published' => 'border-0 border-l-solid border-l-2 border-l-orange-400',
                default => null,
            } 
            return null;
        };
    }
```

Preview:  
image

### DIFF
--- a/packages/tables/docs/02-getting-started.md
+++ b/packages/tables/docs/02-getting-started.md
@@ -497,3 +497,74 @@ class ListPosts extends Component implements Tables\Contracts\HasTable
     }
 }
 ```
+
+## Row appearance
+
+To change the appearance of rows conditionally on the record data, you may add the `getTableRecordClassUsing()` method.
+
+The `getTableRecordClassUsing()` method may return a string or an array of classes that need to be applied to the row.
+
+```php
+<?php
+
+namespace App\Http\Livewire;
+
+use App\Models\Post;
+use Filament\Tables;
+use Illuminate\Contracts\View\View;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Collection;
+use Livewire\Component;
+
+class ListPosts extends Component implements Tables\Contracts\HasTable
+{
+    protected function getTableRecordClassUsing(): ?Closure
+    {
+        return function (Post $record) {
+            return match ($record->status) {
+                'draft' => 'opacity-30',
+                'reviewing' => [
+                    'border-l-solid',
+                    'border-l-2',
+                    'border-l-orange-600',
+                    'dark:border-l-orange-300' => config('filament.dark_mode'),
+                    'opacity-30',
+                ],
+                'published' => 'border-0 border-l-solid border-l-2 border-l-orange-400',
+                default => null,
+            } 
+            return null;
+        };
+    }
+}
+```
+
+These classes are not automatically discovered by tailwind.
+In order to use tailwind classes that are not already used in blade files, you may want to update your `tailwind.config.js` to include your php files:
+
+
+```js
+const colors = require('tailwindcss/colors')
+
+module.exports = {
+    content: [
+    './app/Filament/Resources/**/*.php',
+    ],
+}
+```
+
+Or add the custom classes to the safelist:
+
+```js
+const colors = require('tailwindcss/colors')
+
+module.exports = {
+    safelist: [
+        'border-l-solid',
+        'border-l-2',
+        'border-l-orange-600',
+        'dark:border-l-orange-300',
+        'opacity-30',
+    ],
+}
+```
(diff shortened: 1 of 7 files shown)


zepfietje:
I think we prefer `getRecordClasses` over `getRecordClass`. So could you change `Class` to plural in all places you added it?

packages/tables/docs/02-getting-started.md
```diff
@@ -500,9 +500,9 @@ class ListPosts extends Component implements Tables\Contracts\HasTable
 
 ## Row appearance
```
zepfietje, line 501: This should be moved just below the `Record URLs (clickable rows)` section.

zepfietje (changes requested):

packages/tables/resources/views/components/row.blade.php
```diff
@@ -10,7 +11,7 @@
         'dark:hover:bg-gray-500/10' => $recordUrl && config('tables.dark_mode'),
         'even:bg-gray-100' => $striped,
         'dark:even:bg-gray-900' => $striped && config('tables.dark_mode'),
-    ]) }}
+    ])->class($recordClasses) }}
```
zepfietje, line 14: Let's merge this in with the above using `array_merge` (need to support PHP 8.0, else could have used the spread operator)?

packages/tables/src/Table.php
```diff
@@ -373,6 +382,17 @@ public function getRecordAction(): ?string
...
 
+    public function getRecordClasses(Model $record): null|string|array
+    {
+        $callback = $this->getRecordClassesUsing;
+
+        if (! $callback) {
+            return null;
+        }
+
+        return $callback($record);
```
zepfietje, line 393: Let's wrap the callback return value and always return `?array` (`Arr::wrap()`).

zepfietje (changes requested):

packages/tables/docs/02-getting-started.md
```diff
@@ -308,6 +308,77 @@ In this example, clicking on each post will take you to the `posts.edit` route.
 
 If you'd like to [override the URL](columns#opening-urls) for a specific column, or instead [run a Livewire action](columns#running-actions) when a column is clicked, see the [columns documentation](columns#opening-urls).
 
+## Row appearance
```
zepfietje, line 311:
```suggestion
## Record classes
```

packages/tables/docs/02-getting-started.md
```diff
@@ -308,6 +308,77 @@ In this example, clicking on each post will take you to the `posts.edit` route.
...
+
+To change the appearance of rows conditionally on the record data, you may add the `getTableRecordClassesUsing()` method.
+
+The `getTableRecordClassesUsing()` method may return a string or an array of classes that need to be applied to the row.
+
+```php
+<?php
```
zepfietje, line 318: Let's keep this example simple and have it contain just the method (like the record URL example above).

packages/tables/docs/02-getting-started.md
```diff
@@ -308,6 +308,77 @@ In this example, clicking on each post will take you to the `posts.edit` route.
 
 If you'd like to [override the URL](columns#opening-urls) for a specific column, or instead [run a Livewire action](columns#running-actions) when a column is clicked, see the [columns documentation](columns#opening-urls).
 
+## Row appearance
+
+To change the appearance of rows conditionally on the record data, you may add the `getTableRecordClassesUsing()` method.
```
zepfietje, line 313:
```suggestion
You may want to conditionally style rows based on the record data. This can be achieved by specifying CSS classes to be applied to the row using the `getTableRecordClassesUsing()` method:
```

zepfietje (changes requested):

danharrin:
Related: #2746, #3465

packages/tables/src/Table.php
marcoboers: `array_merge` doesn't accept `null`, so I made the return type `array`

zepfietje: Even better!

zepfietje (approved):

zepfietje:
@danharrin, this should be ready to be merged. Tested it locally and cleaned up the new documentation.

danharrin:
Thanks @marcoboers

carestad:
Was this removed again?

zepfietje:
> Was this removed again?

Nope, and we wouldn't as that would be a breaking change.

yojosee:
Is it not more available? I can't find it, or I don't know how to apply it =(

danharrin:
Please just search the docs https://filamentphp.com/docs/3.x/tables/advanced#custom-row-classes

yojosee:
> Please just search the docs https://filamentphp.com/docs/3.x/tables/advanced#custom-row-classes

oh I see sorry, thank you!

0xyzone:
can we do that with grid layout as well?

PR #3496, opened 2022-08-13
Chore: deprecate `MultiSelect`

Z3d0X:
Replaces https://github.com/filamentphp/filament/pull/2889


danharrin:
As I mentioned in the original PR, I think this is too much of a personal preferences so I'd like to keep MultiSelect - sorry.

PR #3512, opened 2022-08-15
Modularity by wildcard

mgralikowski:
Many Laravels developers implement modularity in their apps using packages like https://github.com/nWidart/laravel-modules or manually by special folders structure (like me).

Assuming structure like this:
`app/Modules/ModuleName/*` 
I would like to have a dedicated folder for Filament:
`app/Modules/ModuleName/Filament/*` 

It is possible to achieve this in Laravel Nova using a provider `Nova::resourcesIn(app_path('Modules/*/Nova'))`.

Filament allows only manual registering resource classes using the `register` config.

This PR provides auto wildcard registering:

```php
'resources' => [
    'namespace' => 'App\\Modules\\*\\Filament\\Resources',
    'path' => app_path('Modules/*/Filament/Resources'),
    'register' => [ ],
],
```

Implementing a wildcard is tricky only for namespace because `allFiles` method under a hood uses the native `glob` so no tweaking is needed.

### DIFF
--- a/packages/admin/src/FilamentServiceProvider.php
+++ b/packages/admin/src/FilamentServiceProvider.php
@@ -207,7 +207,7 @@ class FilamentServiceProvider extends PluginServiceProvider
 
         $filesystem = app(Filesystem::class);
 
-        if (! $filesystem->exists($directory)) {
+        if (! $filesystem->exists($directory) && str_contains($directory, '*') === false) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks, just refactored a bit, please double check it still works fine

danharrin:
@mgralikowski could you check this again for me please?

@ziming @torosegon care to share why you 👎 this solution? Is there something I missed?

mgralikowski:
@danharrin sorry for the late answer. I tested Your changes on my existing app and my module is still auto-discovered correctly in the admin panel.

danharrin:
No worries, thank you very much!

PR #3513, opened 2022-08-15
feature: Support actions in global search results

dmandrade:
Add support to include actions to global search result.
This actions are simple button or link which may open a URL or emit a Livewire event, like the ones on notifications.

image


packages/admin/src/GlobalSearch/Actions/Action.php
```diff
@@ -0,0 +1,89 @@
...
+{
+    use CanBeOutlined;
+    use CanEmitEvent;
+    use CanOpenUrl;
+
+    protected string $view = 'filament::global-search.actions.link-action';
+
+    protected string $viewIdentifier = 'action';
+
+    public function toArray(): array
```
zepfietje, line 22: I don't think the actions need to be fully serializable like in notifications, right? Could just be similar to actions in the tables and forms package.

zepfietje (changes requested):
Generally, I'm not too sure if we should want this functionality from a design/UI/UX perspective.

Anyway, a quick comment on the implementation.

packages/admin/resources/views/components/global-search/actions/action.blade.php
```diff
@@ -0,0 +1,33 @@
+@props([
+'action',
```
zepfietje, line 2: Indentation should be fixed.

packages/admin/src/GlobalSearch/Actions/Action.php
dmandrade: Really don't need it

packages/admin/resources/views/components/global-search/actions/index.blade.php
```diff
@@ -0,0 +1,9 @@
+@props([
+'actions',
```
zepfietje, line 2: Fix indentation.

zepfietje (changes requested):

dmandrade:
@zepfietje Perhaps the best would be to allow the implementation of a custom GlobalSearchProvider where we can define a new class/view of GlobalSearchResult.

That way you are free to customize how the results would be presented.

zepfietje:
What exactly would you like to customize?
If there's missing CSS classes to hook into, we could just add those.
I'm not sure if we really need the ability to customize the entire view.

dmandrade:
> What exactly would you like to customize?
nothing, just thinking about other possibilities.

zepfietje:
Ah alright, let's keep things as simple as possible.

Also, just curious about what you would use global search results actions for, @dmandrade? 😄

dmandrade:
to display shortcuts for some actions like exporting csv/pdf, copy to clipboard, and links to resource pages.

Ex:
- an action to copy the task id in a project management system
- action to preview blog posts in a cms
- action to see all tickets/or the latest one for a client in a helpdesk system

zepfietje:
@dmandrade is this PR ready for proper review? If so, you can mark it as "Ready for review" and we'll have a look!

danharrin:
Thanks

thyseus:
Thanks a lot for _that_ feature. In special thanks for giving us that _IDEA_ for that feature :).
 I am adopting this in one of our projects now which will save us a lot of time !!!

PR #3522, opened 2022-08-16
Fixed secondary button in dark mode

realslimsutton:
I had some issues when using dark mode with the Wizard component. The back button in dark mode uses the light mode hover background, see examples below (ignore the shade of gray in my custom theme):

Before:
image

After:
image


danharrin:
Thanks

PR #3525, opened 2022-08-16
fix resetting tabs and wizard when closing modal

hxnk:
This fixes the issue of wizards and tabs not returning to their original state after a modal closes.

Since resetting them while closing the modal would cause a "flicker" during the transitioning out, the init method on the x-data object is called when reopening the modal instead.

### DIFF
--- a/packages/admin/resources/views/components/page.blade.php
+++ b/packages/admin/resources/views/components/page.blade.php
@@ -38,6 +38,15 @@
         <x-filament::modal
             id="page-action"
             :wire:key="$action ? $this->id . '.actions.' . $action->getName() . '.modal' : null"
+            x-init="
+                $watch('isOpen', () => {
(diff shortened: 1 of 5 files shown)


danharrin:
What are the consequences of watching this property, if it doesnt exist, if you use the wizard or tabs outside of a modal?

hxnk:
Good point!

Added a check to make sure the property is defined.

danharrin:
I'm trying to imagine, maybe it's better to add a modal hook that allows us to reset `mountedTableAction` / `mountedAction` when it is closed. That would mean the modal content is removed from the page and the wizard is reinitialised when the modal is reopened

danharrin:
I've built a different solution, can you please check it still solves your problem @hxnk?

hxnk:
Nice! This solves it too @danharrin :)

PR #3535, opened 2022-08-17
fix: remove padding when sidebar is collapsed

innocenzi:
I'm not sure why, but when using a custom theme, the class order is different and the padding using `--sidebar-width` has priority over the other one, which causes the layout to not shift when we collapse the sidebar.

image

I figured that instead of debugging why would the class order change, I'd just fix the template. So this PR removes the padding that uses `--sidebar-width` when the sidebar is collapsed.


danharrin:
Thanks!

PR #3540, opened 2022-08-17
Allow Toggle Field to have different onColor & offColor

bomshteyn:
The `BooleanColumn` supports both icons and colors.

The toggle field only supports icons. This pull request adds he same functionality to the toggle that already exists on the boolean column

so now you would be able todo something like this:

```PHP
    Toggle::make('is_paid')
        ->onColor('success')
        ->offColor('danger')
        ->onIcon('heroicon-s-check')
        ->offIcon('heroicon-s-x'),
```

### DIFF
--- a/packages/forms/resources/views/components/toggle.blade.php
+++ b/packages/forms/resources/views/components/toggle.blade.php
@@ -1,3 +1,21 @@
+@php
+    $offColorClass = match ($getOffColor()) {
+        'danger' => 'bg-danger-500',
+        'primary' => 'bg-primary-500',
+        'success' => 'bg-success-500',
+        'warning' => 'bg-warning-500',
+        default => 'bg-gray-200',
+    };
+
+     $onColorClass = match ($getOnColor()) {
+        'danger' => 'bg-danger-500',
+        'primary' => 'bg-primary-500',
+        'success' => 'bg-success-500',
+        'warning' => 'bg-warning-500',
+        default => 'bg-primary-600',
+    };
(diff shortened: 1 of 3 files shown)


packages/forms/src/Components/Toggle.php
```diff
@@ -37,30 +41,65 @@ public function offIcon(string | Closure | null $icon): static
...
+    {
+        return $this->evaluate($this->onColor);
+    }
+
     public function hasOffIcon(): bool
     {
         return (bool) $this->getOffIcon();
     }
 
+    public function hasOffColor(): bool
```
danharrin, line 91: These `has` functions are not being used and can be removed

packages/forms/resources/views/components/toggle.blade.php
```diff
@@ -19,8 +37,8 @@
                 x-bind:aria-checked="state.toString()"
                 x-on:click="state = ! state"
                 x-bind:class="{
-                    'bg-primary-600': state,
-                    'bg-gray-200 @if (config('forms.dark_mode')) dark:bg-white/10 @endif': ! state,
+                    '{{$onColorClass}}': state,
```
danharrin, line 40: Please add the match statements inside these `{{ match(...) }}` instead of adding extra variables

danharrin (changes requested):

danharrin:
Thanks

bomshteyn:
Thank you! :-)

PR #3544, opened 2022-08-18
Feature: `TagsInput` Handle paste

Z3d0X:
Just splits the tags by `,` on Paste


Z3d0X:
Oh that is an issue I didn't consider. However, I think we can handle it by checking an _existence_ of a comma in the pasted text, right?

If that ^ doesn't resolve any UX difficulties that arise from this, we can go ahead and close this.

zepfietje:
Still you might want to paste a string containing commas and alter part of it before adding it as tags. I do certainly see the use of this PR, but I'm just not sure if it's worth the trade-off.

zepfietje:
Something like `->createOnPaste()` might be an option if we really want to add it.

danharrin:
I think that they should be created when you paste, if a comma is present within the string.

PR #3545, opened 2022-08-18
add relation support to spatie media library table column

ariaieboy:
right now `SpatieMediaLibraryImageColumn` does not support relationship.

with this PR users can pass the relationship name to the `SpatieMediaLibraryImageColumn` to load the image from the relationship instead of the current record


packages/spatie-laravel-media-library-plugin/src/Tables/Columns/SpatieMediaLibraryImageColumn.php
```diff
@@ -63,13 +66,4 @@ public function getImagePath(): ?string
...
         return $record->getFirstMediaUrl($this->getCollection(), $this->getConversion());
     }
-
-    public function applyEagerLoading(Builder $query): Builder
-    {
-        if ($this->isHidden()) {
-            return $query;
-        }
-
-        return $query->with(['media']);
```
danharrin, line 73: Hi, I think this needs fixing before we merge this. Media needs to be eager-loaded as before.

danharrin:
Potentially, the syntax should be `Column::make('nested.relationship.media')` instead of `Column::make('nested.relationship')`, as column names need to be unique, and this way, you may have multiple media columns for the same related model, each with different collections.

It's the same syntax as we have elsewhere for accessing relationships.

danharrin:
I've just implemented this, also for the Spatie Tags column.

PR #3547, opened 2022-08-18
Fix attach/associate action select translation

pxlrbt:
The attach/associate action select only selects a single value at a time and the translation key ends with `record_id`, but all translation files contain `record_ids`. (https://github.com/filamentphp/filament/blob/d4c5fb0b0fb15f1d625a452d5e6d7f7683ae0d11/packages/tables/src/Actions/AttachAction.php#L197)

I renamed and translated all languages that I could read and verify that the singular form resembles the plural form. For the others I just renamed the translation string as I think plural instead of singular is still better than a missing translation string.

### DIFF
--- a/packages/support/resources/lang/ar/actions/associate.php
+++ b/packages/support/resources/lang/ar/actions/associate.php
@@ -12,7 +12,7 @@ return [
 
             'fields' => [
 
-                'record_ids' => [
+                'record_id' => [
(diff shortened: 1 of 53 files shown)


danharrin:
Wow, so much effort here translating all the phrases. You really didn't need to do that, I would have left that job to the natives. But thank you so much.

PR #3549, opened 2022-08-18
Feature: Add filter count indicator on tables

bernhardh:
On tables right now, if you use one of the filters or if you got a deeplinked page with active filters, you don't see anywhere if a filter is active. Example: https://demo.filamentphp.com/shop/customers?tableFilters[trashed][value]=0

image

I think this isn't a good UX, because you can't expect that the user looks into the address bar (especially on mobiles)

This PR adds an indicator to the table filter icon. It justs adds the number of the currently active filters. I am not really good in design, so maybe there should be some design improvements?

https://user-images.githubusercontent.com/642292/185431352-29f282ca-6b02-4acc-a995-f5d1849adf7a.mp4

This indicator is configurable, see config file `config('tables.filters.indicator')`


zepfietje:
Do we really need the count in the badge or would a simple badge dot be sufficient?
If so, we could just keep the icon in place and have an absolutely positioned badge in the top right.

Z3d0X:
This doesn't handle complex filter forms..

I recall this was discussed in discord a few weeks back.

IIRC it was something along the lines of, its not easy to accurately determine when the filter with custom forms.

bernhardh:
@zepfietje 
It is not necessary to show the amount of used filters, just an indicator that filter(s) are active. So a simple dot or whatever would be enough.

@Z3d0X 
Oh. Thanks for testing this. In my simple example it worked. Hadn't a closer look at complex filters.

So I think I close this PR, since it doesn't work.

zepfietje:
@bernhardh if you like, you can give it another shot by implementing something like discussed in linked Discord thread.

cobraprojects:
I think it good to have a `reset filter icon` instead of count bage would be good

zepfietje:
@cobraprojects, the filters dropdown panel already has a reset button.

cobraprojects:
> @cobraprojects, the filters dropdown panel already has a reset button.

sure it does, But as visual indecator that the table is filterd we shoud add it to the table header or replace the filter icon with reset filter icon.

danharrin:
If we replaced the filter icon, you wouldnt be able to edit an active filter without removing it and recreating it?

danharrin:
Reopening this as we can use it for counting the number of active filter indicators

danharrin:
Replaced by #3743.

PR #3560, opened 2022-08-19
add label name function

rizkyanfasafm:
I think we can custom label name in attach action


danharrin:
Hey, you can use

```php
->recordSelect(function (Select $select) {
    return $select
        ->label('New label here');
})
```

rizkyanfasafm:
Ah,, i am sorry, thank you for your answer

PR #3564, opened 2022-08-19
update fa language

ariaieboy:
update fa translation according to this PR https://github.com/filamentphp/filament/pull/3547
in associate.php and attach.php

in the edit translation, the word `اصلاح` is an Arabic word and in Farsi we usually use `ویرایش` instead.


fsamapoor:
Thank you aria. I just edited the word in my own translation files and I was wondering if others would agree with me about changing it to the Farsi version.

themahdavi:
Good 👍
Need to be change all related `edit` word

PR #3599, opened 2022-08-21
feature: Filter indicators

danharrin:
Screenshot 2022-08-21 at 15 54 01


FaridAghili:
@danharrin

image

Just a little problem on RTL layout.
Close icon (X) goes over the text.

Sicklou:
It seems that the deletion doesn't work.

I think it's related to the fact that my filters have the same name property inside the Checkbox class, which the value is "isActive", even if I precise the name when configuring the filter :

image

When I dd the ```$this->getFlatComponents($withHidden)``` inside the ```getFlatFields``` function of the ```HasComponents``` : 

image

So when I click on the "x" to delete the filter, nothing happens.

danharrin:
The name shouldnt matter in this case. Please open an issue with a reproduction repo and I can look into it

intrepidws:
I think that this has to do with the `default()` method actually. Basically, if you specify the `default()` method for any filter, any don't change it from that default - deleting the filter doesn't do anything because it simply reverts back to that default. If you were to change the filter from the default selection to something else, then click the delete button - it doesn't remove the filter, but it does change it back to the default filter.

I never realized it before but the "Reset Filters" functionality, which was in place before these filter indicators, did the same thing.

While I can understand the rationale here, this behavior doesn't feel intuitive to me. It feels like clicking "Reset Filters", or removing any specific filter via the Filter Indicator, should remove the filter entirely - not revert to the default.

Sicklou:
@intrepidws you are right. I removed the ```default()``` modifier and the removal works. Thank you for finding it :)

I don't know if it is the expected behavior for @danharrin. But on checkbox filter, I was expecting to be "unchecked", even if the default status of the checkbox is to be "checked".

intrepidws:
@danharrin I guess I can understand why the "Reset Filters" link itself would reset the filters to their default state, but it feels like clicking the "X" close icon of a filter indicator should remove the filter entirely. Likewise with clicking the "X" close icon of the entire filter indicator row. Thoughts?

danharrin:
Hm right, I think I understand why this isn't ideal behaviour for resetting the indicator individually, yes. I can see that there would be edge cases where even setting the value to `null` would cause problems, but ignoring the `default()` is probably a better solution, yes. I will prepare a PR now.

danharrin:
#3669

intrepidws:
Stellar work as always Dan, thank you.

PR #3600, opened 2022-08-21
feature: Table content polling

danharrin:


themahdavi:
Thanks @danharrin 
for more control if we have reload button beside filters icon is good for when user needed to reset table content without full refresh
something like `->reloadable()` on table

danharrin:
I've got no plans for that yet, the table header bar is already crowded with how many buttons we have at the moment

themahdavi:
I don't understand why many buttons crowded tables ?
this features not always needed and some of them not activate together

danharrin:
You can already create a refresh button with `headerActions()`:

```php
Action::make('refresh')
    ->iconButton()
    ->icon('heroicon-o-refresh')
    ->action(function () {})
```

PR #3601, opened 2022-08-21
fix: Min/max length validation for numeric inputs

danharrin:
Dependant on laravel/framework#43797.
Replaces #1878.

Compatible with L9 only.


zepfietje:
Don't we have to wait for the new Laravel version to be tagged, @danharrin?

danharrin:
Noone is using this yet anyway so it doesn't matter

zepfietje:
Fair, only breaks if you would already expect that rule to work with numeric state.

danharrin:
Which it doesn't, people shouldnt be using `maxLength()` with `numeric()`

PR #3612, opened 2022-08-22
feature: Per-column search

danharrin:
API TBC.

image


studiowizjo:
What do you think of a bit different layout: split existing search field that the right side will be the select to choose scope? Access to it will be a bit longer, but I think design will stay more clean.

zepfietje:
I like the idea, @studiowizjo!
We would need an indicator to show if any searches are active then though, since we would support simultaneous multiple columns search.

studiowizjo:
Hmm, they could be comma separated in select (something like this: https://mdbootstrap.com/docs/standard/extended/multiselect/) or maybe use similar layout to [new filter indicators](https://github.com/filamentphp/filament/pull/3599)?

danharrin:
Hi, I would prefer not to have a select input involved here, you can use filters to implement a column select search field already. This PR is purely for people who want searchable at the top of their columns, which is not possible with filters atm.

markosole:
Hi @danharrin, appology for throwing question on post from 2022. I am implementing search as shown above and have not found any details for specifying "local searcsh" attributes similar to `globallySearchAttributesUsing` that is available on Global Search. 

I have BlogPosts resource and Blog Model which has  `GlobalScope` configured on it in order to limit/isolate records per user.  Resource Table View works fine and table View is showing records just fine. 

But when I try to use Search and query is executed, that's where it breaks apart. Below is just close similar representation what I have, use case is slightly different but you can get the point.

### I have Model Scope:

```php
protected static function booted()
{
    static::addGlobalScope('author', function ($query) {
        // Limit what can be seen by the user to only their own posts
        $query->leftJoin('promoted_posts', 'blog_posts.author_id', '=', 'promoted_posts.author_id')            
        ->select('blog_posts.*')
        ->where('blog_posts.author_id', auth()->user()->author_id)
        ->addSelect(DB::raw("CASE WHEN promoted_posts.author_id IS NOT NULL THEN 'Yes' ELSE NULL END AS promoted"));
    });
}

```
Query above works perfectly fine for listing records in Table View, as well filtering `?tableSortColumn=post_name&tableSortDirection=asc`  and other columns. But, .... when search kicks in and adds extra parameter to query, that's where it falls apart.


### Search generates query: 
```sql
SELECT
  count(*) AS aggregate
FROM
  `blog_posts`
  LEFT JOIN `promoted_posts` ON `blog_posts`.`author_id` = `promoted_posts`.`author_id`
WHERE
  (`post_name` LIKE % Rock band %)
  AND `blog_posts`.`author_id` = 23
  
  ```
  
Crucial part attribute/column is  `post_name` which isn't found and where expected is **`blog_posts.post_name`**. 
Now, how do I override  `addGlobalScope('author')` scope with maybe a custom `Builder` scope that can execute desired query?
Is that even supported? 

I am using Filament 3.1.x and haven't found anything but globalSearchResults in docs.

**Comment by markosole** [comment]
Update:

I've figure it out. research and research and persistance :). Override query search by setting `blog_posts.post_name` in query. This does "extend" existing Scoped model. 

```php
  
use Illuminate\Database\Eloquent\Builder;
 
TextColumn::make('post_name')
    ->searchable(query: function (Builder $query, string $search): Builder {
        return $query
            ->where('blog_posts.post_name', 'like', "%{$search}%");
    })
    
```
    
Thanks @danharrin for creating such a great tool!

PR #3614, opened 2022-08-22
fix sidebar toggle color in dark-mode

egyjs:


zepfietje:
What's the bug here @egyjs?
The menu icon is shown in the primary color just fine in dark mode.

zepfietje:
Yeah so that's caused by your primary color being too dark. If you're using a well-designed scale like provided by Tailwind CSS it looks fine.

If you really can't update your palette, you can also style the menu button by targeting the `filament-sidebar-open-button` CSS class.

PR #3617, opened 2022-08-22
Prevent lazy and debounce modifiers on entangled fields

tanthammar:
The current implementation allows the user to add `lazy` and `debounce` modifiers to `entangled` fields, thus breaking functionality.
( Did it myself, added lazy to a toggle and wondered why Alpine complained about an undefined value)
This PR fixes that.

With this we also can remove the isLazy() check, on text-input and forget about it on future fields.

### DIFF
--- a/packages/forms/src/Concerns/HasStateBindingModifiers.php
+++ b/packages/forms/src/Concerns/HasStateBindingModifiers.php
@@ -41,6 +41,10 @@ trait HasStateBindingModifiers
     public function applyStateBindingModifiers(string $expression): string
     {
         $modifiers = $this->getStateBindingModifiers();
+        
+        if(str_starts_with($expression, 'entangle') && (in_array('lazy', $modifiers) || in_array('debounce', $modifiers))) {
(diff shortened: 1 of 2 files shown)


packages/forms/src/Concerns/HasStateBindingModifiers.php
```diff
@@ -42,6 +42,10 @@ public function applyStateBindingModifiers(string $expression): string
     {
         $modifiers = $this->getStateBindingModifiers();
 
+        if (str_starts_with($expression, 'entangle') && (in_array('lazy', $modifiers) || in_array('debounce', $modifiers))) {
+            $modifiers = ['defer'];
```
danharrin, line 46: I think I disagree, entangling lazy and debounce fields should mean that no modifiers are applied, as people are still expecting those fields to refresh the Livewire component when they are changed

danharrin: I've fixed this

danharrin:
Thanks

PR #3618, opened 2022-08-22
CSS classes cleanup

zepfietje:


ryangjchandler:
Filament classes should be at the end imo. I want the _actual_ styling first so I have to do less horizontal scanning. In most cases the difference is minimal but at high velocity it makes a difference.

zepfietje:
Having the Filament classes first makes it easier to find the classes to hook into for your custom styling when viewing the Filament source in your application in the browser. So that's basically Filament dev vs Filament user convenience.

Also, the list of Tailwind CSS classes is often quite long anyway. This means it doesn't matter that much if the Filament classes get moved in front, whereas it matters a lot if the Filament classes are all the way at the end or the beginning of the line.

sohaibafifi:
I don’t think it’s only a matter of readability. Rules that appear later override earlier rules.

danharrin:
Ah right, so this may be a breaking change?

zepfietje:
> I don’t think it’s only a matter of readability. Rules that appear later override earlier rules.

That's not true. It's about the order in which the classes appear in the compiled CSS file, not the CSS attribute on the element.

> Ah right, so this may be a breaking change?

Nope, not at all.

danharrin:
Please fix the merge conflicts

zepfietje:
Done, @danharrin!

zepfietje:
@danharrin, if we could get this merged, I'll not be fixing merge commits for the next two weeks. 😂

PR #3631, opened 2022-08-23
Make collapsible repeater headers clickable

tanthammar:
Just like Sections, it is nicer UX if the Repeaters header is clickable to collapse/expand the content.


danharrin:
Thanks

PR #3633, opened 2022-08-23
Add table actions column label helper

emmadesilva:
## Abstract

As discussed in Discord, this PR adds a way to add labels to the action table columns header. This is my first Filament PR so I hope I did everything right here, if not, let me know!

### Backend
The feature is implemented through a protected method `getTableActionsColumnLabel()` in the `HasActions` trait. This is then accessed in the Table DTO, then constructed and rendered in the table Blade file.

### Frontend
In the view, an if-else condition is made to see if the label is set. If it is, it is rendered through the `header-cell` component, making it have the same styling as the other headers. If no label is set, it defaults to the current behaviour, making it fully backwards compatible.

## Usage

To add a action column header label, simply implement the following method stub in your table class.

```php
protected function getTableActionsColumnLabel(): ?string
{
    return 'My label';
}
```

It will then automatically be rendered in the view. All very Filamenty!

image

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -4,6 +4,7 @@
 
     $actions = $getActions();
     $actionsPosition = $getActionsPosition();
+    $actionsColumnLabel = $getActionsColumnLabel();
     $columns = $getColumns();
     $content = $getContent();
     $contentFooter = $getContentFooter();
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -406,7 +406,15 @@
                             @endforeach
 
                             @if (count($actions) && (! $isReordering) && $actionsPosition === Position::AfterCells)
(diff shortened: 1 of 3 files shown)


dododedodonl:
The position of the actions column can differ based on a setting. To ensure support for the label in the other positions, the label should also be added at lines https://github.com/filamentphp/filament/blob/a76c96400f9944a502a27c5c6d39ff90a31f121e/packages/tables/resources/views/index.blade.php#L366 and https://github.com/filamentphp/filament/blob/a76c96400f9944a502a27c5c6d39ff90a31f121e/packages/tables/resources/views/index.blade.php#L390

caendesilva:
> To ensure support for the label in the other positions,

Missed that. Fixed in https://github.com/filamentphp/filament/pull/3633/commits/0ddafbc04ab1bb9bf72d0abb9dca9bf045b32d7b. Thanks for letting me know about it!

danharrin:
Thanks

PR #3649, opened 2022-08-24
Fix wire:ignore fields: Change border colors on validation errors

### LINKED ISSUE #2398: Markdown editor doesn't get a red border upon validation errors

tanthammar:
- Border colors will react to validation errors, even if the field has the wire:ignore attribute applied.
- Sections that are collapsed will expand if they contain fields with validation errors.
--------

Compatible w LW3
`$wire.__instance.serverMemo.errors` is the same as `$wire.errors` in LW3 (maybe).
Src: https://github.com/livewire/livewire/pull/5054 Thank you @zepfietje 

If it will be implemented in LW3, we can do a simple search'n'replace.

(This PR does not have any conflict with the notification plugin. It replaces the previous PR #3069)

### DIFF
--- a/packages/admin/resources/views/components/form/index.blade.php
+++ b/packages/admin/resources/views/components/form/index.blade.php
@@ -1,10 +1,5 @@
 <form
-    x-data="{
-        isUploadingFile: false,
-        get validationErrors() {
-            return $wire.__instance.serverMemo.errors
-        }
-    }"
+    x-data="{ isUploadingFile: false }"
     x-on:submit="if (isUploadingFile) $event.preventDefault()"
     x-on:file-upload-started="isUploadingFile = true"
     x-on:file-upload-finished="isUploadingFile = false"
--- a/packages/admin/resources/views/components/form/index.blade.php
+++ b/packages/admin/resources/views/components/form/index.blade.php
@@ -1,5 +1,10 @@
 <form
-    x-data="{ isUploadingFile: false }"
+    x-data="{
+        isUploadingFile: false,
+        get validationErrors() {
+            return $wire.__instance.serverMemo.errors
+        }
+    }"
     x-on:submit="if (isUploadingFile) $event.preventDefault()"
     x-on:file-upload-started="isUploadingFile = true"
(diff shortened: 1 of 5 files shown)


danharrin:
As explained, there is no way to put this onto the `<form>` tag for users of the form builder outside the admin panel.

So, since `$wire.errors()` is on the roadmap, lets just use `$wire.__instance.serverMemo.errors.{{ $getStatePath() }}` raw to retrieve the error inside the markdown editor, as that's the only component that needs fixing, right?

Not sure I like the solution proposed for the collapsed Section to check for an error element, that feels like a bad practice 🤔 What is the issue with our current implementation for expanding the section?

tanthammar:
This PR triggered a Livewire/Alpine issue within a Repeater. 
When you click the Repeater button to add an item you get this error on console:
```
Alpine: Cannot reference "$wire" outside a Livewire component.
``` 
I have two text-inputs in the Repeater.
When I add an item this triggers the error.
```
x-bind:class="{
                    'border-gray-300': ! (@js($getStatePath()) in $wire.__instance.serverMemo.errors),
                    'dark:border-gray-600': ! (@js($getStatePath()) in $wire.__instance.serverMemo.errors) && @js(config('forms.dark_mode')),
                    'border-danger-600 ring-danger-600': (@js($getStatePath()) in $wire.__instance.serverMemo.errors),
                }"
```

tanthammar:
Fix #3657

PR #3652, opened 2022-08-24
UX enhancement for filter icon if table is filtered or not

cobraprojects:
change filter icon to be solid only if table is filtered


cobraprojects:
fix getFilters count

danharrin:
Honestly I've tried doing this in the past and I'm just not a huge fan of the way it looks. Sorry. I would accept a PR that adds a number count badge in the top right of the icon, which counts the number of filter indicators that are active. Like #3549

PR #3655, opened 2022-08-24
feature: Expand and scroll to validation errors

tanthammar:
https://user-images.githubusercontent.com/21239634/186468668-a1d5c244-a6c7-47cc-9cf9-5da8977e55cc.mov

### Trigger =` expand-concealing-component.window `

- It only scrolls to the element that matches the `$event.detail.id` sent from backend
- It opens the first Tab in each Tab-set, that contains errors
- It expands all Sections containing errors
- When a Section or Repeater contain errors it can't be collapsed
- It expands all Repeaters containing errors

In the demo movie you see a form containing two tabs with multiple sections and nested repeaters.
1. I click on the second tab before submitting the form.
2. The first tab (that contain errors) opens
3. All sections with errors expands
4. I scroll down and you can see that nested Repeaters are expanded.


danharrin:
I've managed to clean this up a bit:

- Each time the event is received on any of the concealing elements, check if the element contains a validation error
- If it does, then expand it
- If the validation error is the first error that is found on the entire page, then scroll to it. This prevents multiple elements from attempting to scroll at once

PR #3657, opened 2022-08-24
Fix bug caused by x-bound classes to wire.errors

tanthammar:
A bit weird solution but it was not enough to just add an empty x-data object


tanthammar:
Closing as agreed. Fixed.

PR #3664, opened 2022-08-25
chore(lang): update pl translations

kamilst96:
Small lang/pl updates.

@lolen for TagsColumn limit, would you suggest something other than `'more' => 'i :count więcej'`, or is it fine?


lolen (commented):
It seems okay 👍

PR #3669, opened 2022-08-25
fix: Removing filters with default values

danharrin:
https://github.com/filamentphp/filament/pull/3599#issuecomment-1225390438


Sicklou:
Just update on my project. It is really nice. Filters indicators was so needed ! Thanks a lot Dan.

PR #3678, opened 2022-08-25
Fix getFlatComponents doesn't return all hidden components

craigkuhns:
I'm working on adding some form testing helpers and ran into what looks like a bug to me when implementing an `assertFormFieldExists` helper. 

The helper gets the form from the Livewire instance. Then to check for the existence of the field it calls`$form->getFlatFields(true)` to get all of the fields on the form including hidden fields. However when testing a form with a container like a Section or Fieldset that is hidden, the child components of that container are not returned.

I tracked this down to the `getFlatComponents` method not including `$withHidden` when calling `$component->getChildContainers()`. `getChildContainers` calls `hasChildComponentContainer` and if `$withHidden` is not passed in it will return false. 

This PR fixes that. I don't think this should break anything. Everything on my project is still working. All of the tests are passing and I've added a couple of extra tests that verify the fix. But I'm still trying to get familiar with the internals of Filament so I'm not sure if there are any side effects I'm not aware of.

### DIFF
--- a/packages/forms/src/Concerns/HasComponents.php
+++ b/packages/forms/src/Concerns/HasComponents.php
@@ -43,7 +43,7 @@ trait HasComponents
                         $component,
                         array_map(
                             fn (ComponentContainer $container): array => $container->getFlatComponents($withHidden),
-                            $component->getChildComponentContainers(),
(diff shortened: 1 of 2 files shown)


danharrin:
Thanks!

PR #3684, opened 2022-08-26
Fix issue to remove table filters.

sumardi:
Fix issue to remove all filters with multiple `SelectFilter`.


danharrin:
Hi, this fix looks wrong, and all of the tests are failing. Whats the issue?

sumardi:
`Tables\Filters\SelectFilter` with relationship and `Tables\Filters\TrashedFilter` both have the same name of "value". Only 1 can be removed.

sumardi:
Should use `$field->getId()` maybe?

danharrin:
I've pushed the correct fix

sumardi:
> I've pushed the correct fix

Tested and it fixes the issue. Thanks.

PR #3698, opened 2022-08-26
feature: Add some form testing helpers

craigkuhns:
The new table and action testing helpers are great. The current project I'm working on has some interactive forms with disabled/enabled fields and hidden/visible fields depending on who is logged in or the state of other fields on the form. I wanted an easy way to test them so I took inspiration from the table testing helpers and I added the following helpers:

- assertFormExists: I haven't used this outside of use in the other testing helpers
- assertFormFieldExists: I haven't used this outside of use in the other testing helpers
- assertFormFieldIsDisabled
- assertFormFieldIsEnabled:
- assertFormFieldIsHidden
- assertFormFieldIsVisible

You can look at tests/src/Forms/FormsTest.php to see examples of how these helpers can be used. I also added some documentation showing simple use cases of how they work.

Regarding tests, I wasn't sure of the best place to add them so I put them all in FormsTest.php. Most of the helpers are actually about form fields, but they didn't seem to fit in the FieldTest.php file. I can move them around though if needed.

Let me know if there are any changes needed or if this would be better off as a plugin.

### DIFF
--- a/packages/forms/docs/07-testing.md
+++ b/packages/forms/docs/07-testing.md
@@ -4,11 +4,11 @@ title: Testing
 
 All examples in this guide will be written using [Pest](https://pestphp.com). However, you can easily adapt this to a PHPUnit.
 
-Since the form builder works on Livewire components, you can use the [Livewire testing helpers](https://laravel-livewire.com/docs/testing). However, we have custom testing helpers that you can use for forms:
+Since the form builder works on Livewire components, you can use the [Livewire testing helpers](https://laravel-livewire.com/docs/testing). However, we have custom testing helpers that you can use with forms:
 
-## Forms
+## Form existance
 
-To check that a Livewire component has a form, use `assertFormExists`:
+To check that a Livewire component has a form, use `assertFormExists()`:
 
 ```php
 use function Pest\Livewire\livewire;
@@ -19,11 +19,11 @@ it('has a form', function () {
 });
 ```
 
-> Note that if you have multiple forms on a Livewire component, you can pass the name of a specific form like `assertFormExists('fooForm')`.
+> Note that if you have multiple forms on a Livewire component, you can pass the name of a specific form like `assertFormExists('createPostForm')`.
 
 ## Fields
 
-To ensure that a form has a given field pass the field name to `assertFormFieldExists`:
+To ensure that a form has a given field pass the field name to `assertFormFieldExists()`:
 
 ```php
 use function Pest\Livewire\livewire;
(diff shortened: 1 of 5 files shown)


packages/forms/docs/07-testing.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+Or to ensure that a field is hidden you can pass the name to `assertFormFieldIsHidden`:
+
+```php
+use function Pest\Livewire\livewire;
+
+test('title is hidden', function () {
+    livewire(CreatePost::class)
+        ->assertFormFieldIsVisible('title');
```
rschoonheim, line 59: Is this correct? I assume you want to show the working of `assertFormFieldIsHidden('title');`

craigkuhns: Thanks, you're right!

packages/forms/src/Testing/TestsForms.php
```diff
@@ -0,0 +1,160 @@
...
+            $this->assertFormExists($formName);
+
+            $livewire = $this->instance();
+            $livewireClass = $livewire::class;
+
+            /** @var ComponentContainer $form */
+            $form = $livewire->$formName;
+
+            /** @var ?Field $field */
+            $field = data_get($form->getFlatFields(true), $fieldName, null);
```
danharrin, line 51: Please used a named parameter for `getFlatFields(true)`

danharrin: Also this is used in other places, update those also

packages/forms/src/Testing/TestsForms.php
```diff
@@ -0,0 +1,160 @@
...
+            $form = $livewire->$formName;
+
+            $fields = $form->getFlatFields(true);
+
+            $fields = $form->getFlatFields();
+
+            Assert::assertArrayHasKey(
+                $fieldName,
+                $fields,
+                "Failed asserting that a field with the name [{$fieldName}] is hidden on the form named [{$formName}] on the [{$livewireClass}] component."
```
danharrin, line 154:
```suggestion
                "Failed asserting that a field with the name [{$fieldName}] is visible on the form named [{$formName}] on the [{$livewireClass}] component."
```

packages/forms/src/Testing/TestsForms.php
```diff
@@ -0,0 +1,160 @@
...
+            /** @phpstan-ignore-next-line  */
+            $this->assertFormFieldExists($fieldName, $formName);
+
+            $livewire = $this->instance();
+            $livewireClass = $livewire::class;
+
+            /** @var ComponentContainer $form */
+            $form = $livewire->$formName;
+
+            $fields = $form->getFlatFields();
```
danharrin, line 123: Please explicitly pass `false` here instead of relying on the default, also using a named parameter

danharrin (changes requested):
Also, please add a `.stubs.php` file for the forms package, like we have for Admin and Tables. This provides IDE autocompletion for the TestableLivewire class.

Apart from that, very nice job! I like it.

packages/forms/src/Testing/TestsForms.php
craigkuhns: This is updated now

packages/forms/src/Testing/TestsForms.php
craigkuhns: All of these are updated now

craigkuhns:
Thanks @danharrin ! I've added the .stubs.php file and made the other suggested changes. Thanks for all of your work on Filament, it's the best admin panel I've used.

danharrin:
Thank you very much for the kind words and the contribution

PR #3701, opened 2022-08-26
Fixed empty check on filter indication

Justin991q:
Not really sure if it is a bug, but I noticed that if you have a select filter with:

```php
SelectFilter::make('active')
    ->options([
        '1' => 'Yes',
        '0' => 'No',
    ]),
```

And you select 'No' it does not show up in the filter indications because of the '0' value.

### DIFF
--- a/packages/tables/src/Filters/MultiSelectFilter.php
+++ b/packages/tables/src/Filters/MultiSelectFilter.php
@@ -24,7 +24,7 @@ class MultiSelectFilter extends BaseFilter
         $this->placeholder(__('tables::table.filters.multi_select.placeholder'));
 
         $this->indicateUsing(function (array $state): array {
-            if (! ($state['values'] ?? false)) {
+            if (blank($state['value'] ?? null)) {
                 return [];
             }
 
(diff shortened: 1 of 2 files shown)


danharrin:
Thanks!

mikrosmile:
After updating to latest ver with this PR, indicators are not showing for multiple selection ...

danharrin:
I see, there was a mistake

packages/tables/src/Filters/MultiSelectFilter.php
```diff
@@ -24,7 +24,7 @@ protected function setUp(): void
         $this->placeholder(__('tables::table.filters.multi_select.placeholder'));
 
         $this->indicateUsing(function (array $state): array {
-            if (! ($state['values'] ?? false)) {
+            if (blank($state['value'] ?? null)) {
```
danharrin, line 27: This should have been `$state['values']`, I will fix

danharrin: 99d725670d6bb3533c61194cbb4cbd4e68d7b70a

PR #3705, opened 2022-08-27
Centralized contracts in the admin package.

rschoonheim:
Hello Everyone, 

While exploring the filament admin package I encountered contracts at multiple locations making it harder to know what contracts are available. My understanding of a contract is that it promotes interchangeability by defining a specific service (interfacing). 

With the current implementation I have to open multiple directories to “discover” what contracts are available making it harder to get an overview. With this pull request I organized all (admin) contracts in the `packages/admin/src/Contracts` folder making it easier to get an overview.

Thoughts?

Resources
1. https://laravel.com/docs/9.x/contracts#introduction 

**Warning: this change is breaking when considered useful we should implement it in the next major version.**


zepfietje:
This change would make it harder to maintain and split parts of the Filament codebase.
There's probably other ways for you to quickly view all available contracts in your IDE.
All in all, I don't see how this change would be a net positive for maintaining Filament.

Also, it seems like you introduced additional changes in this PR aside from moving the contracts (you added `declare(strict_types=1);`) which should not be handled in this PR if intended at all.

Closing this PR, unless @danharrin thinks this change actually is a good idea.

danharrin:
Yes I agree with Zep here 👍 Our current structure also helps to contextualize contracts since we have packages that are often used individually, unlike Laravel.

PR #3706, opened 2022-08-27
Update German/DE translations

fabianpnke:
Hi all,

In this PR, I rebased all German/DE translations based on the English/EN ones. This should make a comparison to the English/EN ones simpler in the future. For example, it will be easier to find new missing translations without the need to check if the translations may be in a different order.

Additionally, I added all current missing translations to the German/DE files. So this locale should be complete at the moment.

It would be great if someone could do a quick crosscheck.

Have a nice day,
Fabian

### DIFF
--- a/packages/admin/resources/lang/de/layout.php
+++ b/packages/admin/resources/lang/de/layout.php
@@ -1,6 +1,9 @@
 <?php
 
 return [
+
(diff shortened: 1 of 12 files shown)


danharrin:
Great, thank you

PR #3708, opened 2022-08-27
feature: Add affixes to BooleanColumn and IconColumn

ralphjsmit:
This PR adds the ability to add suffixes and prefixes to boolean columns. This is handy in situations where you want to add some more context to a boolean, but where you don't want to create a new text column (e.g. because there's too much whitespace).

Schermafbeelding 2022-08-27 om 15 35 00

### DIFF
--- a/packages/tables/resources/views/columns/boolean-column.blade.php
+++ b/packages/tables/resources/views/columns/boolean-column.blade.php
@@ -11,10 +11,6 @@
         'warning' => 'text-warning-500',
         default => 'text-gray-700',
     };
-
-    $affixLabelClasses = [
-        'whitespace-nowrap group-focus-within:text-primary-500',
-    ];
 @endphp
 
 <div
--- a/packages/tables/resources/views/columns/boolean-column.blade.php
+++ b/packages/tables/resources/views/columns/boolean-column.blade.php
@@ -11,13 +11,34 @@
         'warning' => 'text-warning-500',
         default => 'text-gray-700',
     };
+
+    $affixLabelClasses = [
+        'whitespace-nowrap group-focus-within:text-primary-500',
+    ];
 @endphp
 
-<div {{ $attributes->merge($getExtraAttributes())->class(['px-4 py-3 filament-tables-boolean-column']) }}>
+<div
+    {{ $attributes->merge($getExtraAttributes())->class([
+        'px-4 py-3 filament-tables-boolean-column',
+        'flex flex-row gap-x-2'
+    ]) }}
+>
+    @if ($prefix = $getPrefix())
+        <span class="whitespace-nowrap group-focus-within:text-primary-500">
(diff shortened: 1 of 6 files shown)


zepfietje:
It feels like for your use case you could use a badge column instead of boolean?
I'm not sure if we should 'bloat' the boolean column with a prefix and suffix, feels wrong to me in a way.

danharrin:
I think it should be on both, also IconColumn? do we have that or am i making it up

ralphjsmit:
> It feels like for your use case you could use a badge column instead of boolean?
> I'm not sure if we should 'bloat' the boolean column with a prefix and suffix, feels wrong to me in a way.

No, because a badge column really displays a badge with a colored background, which looks much uglier.

ralphjsmit:
> I think it should be on both, also IconColumn? do we have that or am i making it up

I'm not sure either, but I'll check and update the PR!

ralphjsmit:
Added IconColumn, so ready for review. @danharrin

danharrin:
I'm interested now... can we somehow refactor `BooleanColumn` to extend `IconColumn`?

zepfietje:
> I'm interested now... can we somehow refactor `BooleanColumn` to extend `IconColumn`?

Exactly my thought!

ralphjsmit:
> I'm interested now... can we somehow refactor `BooleanColumn` to extend `IconColumn`?

I think that it should be possible. I'll take a look later this week.

ArtMin96:
Maybe create a new component and call it IndicatorColumn?

zepfietje:
Agree. Let's just make `BooleanColumn` extend `IconColumn` and see how that works out first.

ArtMin96:
> > Maybe create a new component and call it IndicatorColumn?
> 
> Personally I don't see the benefit of renaming IconColumn to IndicatorColumn. Or am I understanding you incorrectly?
> 
> Not sure what @danharrin thinks.

For each component to be responsible for the actions assigned to it, BooleanColumn should be boolean, IconColumn should be an icon, and IndicatorColumn should be the indicator column. Also, in the case of further changes, problems should not arise due to connecting the BooleanColumn to the IconColumn. This is just an opinion, you know Filament better than me

zepfietje:
What's the status of this PR, @ralphjsmit? 🙂

We've decided to deprecate `BooleanColumn` in favor of `IconColumn::make()->boolean()`: https://github.com/filamentphp/filament/issues/4273. So that will be handled separately from this PR.

zepfietje:
https://github.com/filamentphp/filament/pull/4431 has been merged and deprecates the `BooleanColumn`. Could you update this PR to add affix support to the `IconColumn`?

danharrin:
Gonna close this, many merge conflicts so it would be easier to do the work on a new branch anyway

PR #3715, opened 2022-08-28
chore(lang): update pl translations

kamilst96:
Missing strings from recent updates:
- Records per page: all
- Filters: delete (all) filters


zepfietje:
Thanks!

PR #3716, opened 2022-08-28
Added Japanese translation

askdkc:
found this resource file has not been translated in Japanese so I just fixed it.

### DIFF
--- a/packages/support/resources/lang/ja/actions/attach.php
+++ b/packages/support/resources/lang/ja/actions/attach.php
@@ -4,16 +4,16 @@ return [
 
     'single' => [
 
-        'label' => 'Attach',
+        'label' => '追加',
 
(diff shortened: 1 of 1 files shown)


packages/support/resources/lang/ja/actions/attach.php
```diff
@@ -4,36 +4,36 @@
 
     'single' => [
 
-        'label' => 'Attach',
+        'label' => '追加',
 
         'modal' => [
 
-            'heading' => 'Attach :label',
+            'heading' => ':labelの追加',
```
zepfietje, line 11: I don't know Japanese, but shouldn't there be a space between the label and the rest of the string?

zepfietje:
Thanks!

PR #3719, opened 2022-08-28
Support Turbo navigation

tanthammar:
Please consider adding support for us who use hotwire Turbo.
This change has absolutely no impact for non Turbo users, but it is very important if you do.
The data attribute forces a full page reload when clicking on the navigation brand links.
(Which in most cases leads to a part of the application outside of the Filament Admin panel.)

It would be great, not having to override this file for only these lines.


danharrin:
Sure, you make a reasonable case

tanthammar:
Thank you!

PR #3720, opened 2022-08-28
Show loading state when sorting table

marksalmon:
I'm new to filament, so I might be missing something, but is there any other way to get 'loading' status when sorting?


danharrin:
Thanks

PR #3721, opened 2022-08-28
Change sidebar active item background color

margarizaldi:
Change active item background color from `bg-primary-500` to `bg-primary-600` to enhance white text visibility especially for filament default theme, and bring color consistency with primary buttons. This applies for both light and dark mode.

Before:
image

After:
image

Current primary button:
image

**Note:** I know this change is subjective & debatable, so discussions or rejections are totally welcome ✌️

### DIFF
--- a/packages/admin/resources/views/components/layouts/app/sidebar/item.blade.php
+++ b/packages/admin/resources/views/components/layouts/app/sidebar/item.blade.php
@@ -33,7 +33,7 @@
             'flex items-center justify-center gap-3 px-3 py-2 rounded-lg font-medium transition',
             'hover:bg-gray-500/5 focus:bg-gray-500/5' => ! $active,
             'dark:text-gray-300 dark:hover:bg-gray-700' => (! $active) && config('filament.dark_mode'),
-            'bg-primary-500 text-white' => $active,
+            'bg-primary-600 text-white' => $active,
         ])
     >
         <x-dynamic-component :component="$icon" class="h-5 w-5 shrink-0" />


danharrin:
I think we should swap the default color from `yellow` to `amber` instead, across the entirety of Filament. I'm already using it on the new marketing site WIP.

danharrin:
And we should swap all `-600` uses to `-500`

zepfietje:
Are you ready for a PR, @danharrin?

danharrin:
Sure

danharrin:
Please manually check all -600 uses when you switch in case its a bad move in some places

zepfietje:
@margarizaldi as for the active sidebar item, what about switching to a less obtrusive color for the highlight? Something like @ryangjchandler's flat theme: https://github.com/ryangjchandler/filament-flat-theme.

image

danharrin:
Replaced by #3722

margarizaldi:
> @margarizaldi as for the active sidebar item, what about switching to a less obtrusive color for the highlight? Something like @ryangjchandler's flat theme: https://github.com/ryangjchandler/filament-flat-theme.
> 
> image
> 
> 

It looks good but less attractive I think 😁 honestly first time I chose filament because it is yellow. Just little adjustments to make it even perfect.

I will happily wait to see @danharrin 's PR released. As I agree that primary-600 might not a good practice for widely used admin panel package.

zepfietje:
I think the solid primary background color for the active sidebar item might be a little too striking for what it does. But we'll what design changes will make it when revamping the sidebar.

PR #3722, opened 2022-08-28
Use amber as primary color for Filament admin

zepfietje:
This PR updates the primary color of Filament from yellow to amber, since amber looks better especially in the darker shades.

Additionally, usage of some color shades was improved, mainly the use of 500 instead of 600 for borders (following Tailwind UI).


themahdavi:
Why just change primary color from yellow to amber?

danharrin:
Because it gives better readability in lots of cases when used as a background color, and also the darker colours are generally less murky

PR #3728, opened 2022-08-29
Allow filterTable testing helper to accept array of data

austincarpenter:
Currently when using the `filterTable()` testing helper, a series of conditional checks happen to ensure data is being passed to the various filter types correctly, transforming it as necessary. This PR allows data of type `array` to pass through the checks without being transformed into data for a simple boolean filter, meaning custom filter forms can be tested, such as:

```php
livewire(QuoteResource\Pages\ListQuotes::class)
    ->filterTable('date_range', [
        'start_date' => today()->subDay(),
        'end_date' => today()->addDay(),
    ])
    // ...
```


danharrin:
Thanks!

PR #3729, opened 2022-08-29
Compact section style

G3z:
this PR provides an optional smaller style for sections


danharrin:
Instead of adding an entirely different view, please add a `compact()` method and `isCompact()` getter to our existing Section and just merge classes conditionally.

packages/forms/src/Components/Section.php
```diff
@@ -10,11 +10,16 @@
...
     use Concerns\CanBeCollapsed;
+    use Concerns\CanBeCompacted;
     use Concerns\EntanglesStateWithSingularRelationship;
     use HasExtraAlpineAttributes;
 
     protected string $view = 'forms::components.section';
 
+    public const REGULAR = 'forms::components.section';
+
+    public const SMALL = 'forms::components.section-small';
```
zepfietje, line 21: I think we'd rather have abstracted methods like `compact()` instead of having to pass the const to the `view()` method. Take a look at actions for example, with their `button()` and `link()` methods.

zepfietje (changes requested):

packages/forms/src/Components/Section.php
danharrin: Yes exactly

G3z: Sorry: I forgot to clean up those contants.
abstract methods are implemented in the `CanBeCompacted` trait

PR #3733, opened 2022-08-29
Add authorization on actions of create and edit

luilliarcec:
Add authorization check before each action on create and edit pages.

### DIFF
--- a/packages/admin/src/Resources/Pages/CreateRecord.php
+++ b/packages/admin/src/Resources/Pages/CreateRecord.php
@@ -30,11 +30,16 @@ class CreateRecord extends Page implements HasFormActions
 
     public function mount(): void
     {
+        $this->authorizeAccess();
+
+        $this->fillForm();
+    }
+
+    protected function authorizeAccess()
+    {
         static::authorizeResourceAccess();
 
         abort_unless(static::getResource()::canCreate(), 403);
-
-        $this->fillForm();
     }
 
     protected function fillForm(): void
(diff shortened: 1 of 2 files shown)


danharrin:
Hey, how about View and List?

packages/admin/src/Resources/Pages/CreateRecord.php
```diff
@@ -29,12 +29,17 @@ public function getBreadcrumb(): string
     }
 
     public function mount(): void
+    {
+        $this->authorizeAccess();
+
+        $this->fillForm();
+    }
+
+    protected function authorizeAccess()
```
danharrin, line 38:
```suggestion
    protected function authorizeAccess(): void
```

packages/admin/src/Resources/Pages/EditRecord.php
```diff
@@ -37,14 +37,19 @@ public function getBreadcrumb(): string
     }
 
     public function mount($record): void
+    {
+        $this->authorizeAccess($record);
+
+        $this->fillForm();
+    }
+
+    protected function authorizeAccess($record)
```
danharrin, line 46:
```suggestion
    protected function authorizeAccess(Model $record): void
```

luilliarcec:
> Hey, how about View and List?

I don't think there is a need, in those pages the DB is not affected, I did this thinking about the validation that post routes have in a normal application.  But if you explain your point a little better, I can do it no problem ✌🏻

packages/admin/src/Resources/Pages/EditRecord.php
luilliarcec: That `$record` refers to the identifier of the model, therefore it would not receive `Model`, but rather int or, failing that, string.

danharrin:
Ah ignore me, I misunderstood the PR a bit. Thanks!

PR #3740, opened 2022-08-29
Make global search case insensitive for translatable json fields

zunderbolt:
Made global search case insensitive for translated fields saved in json. Applied the same approach as in `\Filament\Tables\Columns\Concerns\InteractsWithTableQuery`. Ideally, some code could be shared, but I don't know where to put it given the files that belong to different packages.


danharrin:
Thanks

danharrin:
Thanks

PR #3742, opened 2022-08-29
Fix/per page option issue

aymanalareqi:
This PR is fix for the issue when per page all option active is not been selected because it is null but the first option selected instead


danharrin:
Sorry, I don't understand the problem this solves?

aymanalareqi:
> Sorry, I don't understand the problem this solves?

You can check the problem by choosing the select all option of per page the refresh the page
You will see the selected option of per page is the first option instead of all option 


this PR solve this problem

zepfietje:
The issue is that the `wire:model` on the select doesn't mark the `All` option as selected, since it's value is `null` (which converts to an empty string).

aymanalareqi:
> ```
> <option value="{{ $option ?? 'null' }}">{{ $option ?? __('tables::table.pagination.fields.records_per_page.options.all') }}</option>
> ```

Ok thanks
but I think my solution do the same as you say but using the string "*" instead of "null"

zepfietje:
Using a string like `'all'` or an integer value `-1` would be an easy fix, though potentially kind of breaking now. I'd prefer if we could keep `null` for `All`. Any idea how we can get that working?

aymanalareqi:
> 



> Using a string like `'all'` or an integer value `-1` would be an easy fix, though potentially kind of breaking now. I'd prefer if we could keep `null` for `All`. Any idea how we can get that working?

No idea right now  😢

https://datatables.net/examples/advanced_init/length_menu.html
look at this it use -1 as value for all option and another library use 9999999 as all value

zepfietje:
We could use `-1`. What do you think, @danharrin?

danharrin:
Or 0

packages/tables/resources/views/components/pagination/records-per-page-selector.blade.php
```diff
@@ -2,13 +2,13 @@
     'options',
 ])
 
-<div class="filament-tables-pagination-records-per-page-selector flex items-center space-x-2 rtl:space-x-reverse ">
+<div class="flex items-center space-x-2 filament-tables-pagination-records-per-page-selector rtl:space-x-reverse ">
```
danharrin, line 5:
```suggestion
<div class="flex items-center space-x-2 filament-tables-pagination-records-per-page-selector rtl:space-x-reverse">
```

zepfietje: And make sure to move the `filament-` class to the start!

packages/tables/src/Concerns/CanPaginateRecords.php
```diff
@@ -30,25 +30,25 @@ protected function paginateTableQuery(Builder $query): Paginator
...
         );
 
         return $records->onEachSide(1);
     }
 
-    protected function getTableRecordsPerPage(): ?int
+    protected function getTableRecordsPerPage(): int
     {
-        return filled($this->tableRecordsPerPage) ? intval($this->tableRecordsPerPage) : null;
+        return intval($this->tableRecordsPerPage);
```
zepfietje, line 43: I believe casting `(int)` is more performant than using `intval`.

zepfietje (changes requested):

packages/tables/src/Concerns/CanPaginateRecords.php
danharrin, line 43:
```suggestion
        return (int) $this->tableRecordsPerPage;
```

PR #3743, opened 2022-08-29
Add filter count on top of filter icon

cobraprojects:
for https://github.com/filamentphp/filament/pull/3549


packages/support/resources/views/components/icon-button.blade.php
```diff
@@ -59,6 +60,9 @@
         @endif
 
         <x-dynamic-component :component="$icon" :class="$iconClasses" />
+        @if($indicator)
+            <span class="bg-{{ $color }}-500 rounded-full text-xs text-white inline-block w-4 h-4 absolute top-0 right-0">{{ $indicator }}</span>
```
Z3d0X, line 64: instead of `bg-{{ $color }}-500` complete tailwind class names should be used . See an example how dynamic classes are handled, for `$buttonClasses` & `$iconClasses`, at top of this file

danharrin:
Thanks!

PR #3763, opened 2022-08-30
fixes issue with casting to string for non-incrementing primary keys

nkeena:
This PR transforms the primary key being plucked from the table query and casts it to a string type. For some non-incrementing primary keys (i.e snowflake), selecting all records was exhibiting unexpected behavior.

This fixes the issue with all records not being selected when using a non-incrementing primary key.

### DIFF
--- a/packages/tables/src/Concerns/CanSelectRecords.php
+++ b/packages/tables/src/Concerns/CanSelectRecords.php
@@ -21,7 +21,10 @@ trait CanSelectRecords
     {
         $query = $this->getFilteredTableQuery();
 
-        return $query->pluck($query->getModel()->getQualifiedKeyName())->toArray();
+        return $query
+            ->pluck($query->getModel()->getQualifiedKeyName())
+            ->transform(fn ($key) => (string) $key)
+            ->toArray();
     }
 
     public function getAllTableRecordsCount(): int
--- a/packages/tables/src/Concerns/CanSelectRecords.php
+++ b/packages/tables/src/Concerns/CanSelectRecords.php
@@ -24,7 +24,7 @@ trait CanSelectRecords
         return $query
             ->pluck($query->getModel()->getQualifiedKeyName())
             ->map(fn ($key): string => (string) $key)
-            ->toArray();
+            ->all();
     }
 
     public function getAllTableRecordsCount(): int
--- a/packages/tables/src/Concerns/CanSelectRecords.php
+++ b/packages/tables/src/Concerns/CanSelectRecords.php
@@ -23,7 +23,7 @@ trait CanSelectRecords
 
         return $query
             ->pluck($query->getModel()->getQualifiedKeyName())
-            ->transform(fn ($key) => (string) $key)
+            ->map(fn ($key): string => (string) $key)
             ->toArray();
     }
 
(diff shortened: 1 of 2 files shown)


zepfietje:
Seems like the keys are already mapped to string in JS? https://github.com/filamentphp/filament/blob/52cba0d5b9c07396a1a8c702b3dbf92fe632acc4/packages/tables/resources/views/index.blade.php#L137

zepfietje:
@pxlrbt any idea?

pxlrbt:
Haven't worked with non-numeric keys inside Filament, yet. I think it's good if we serve this server side and remove the JS mapping.

@nkeena why is this solving the issue? Are string keys modified when using `->toArray()`?

zepfietje:
I don't think it's about string keys necessarily; snowflakes are numeric, but not incrementing.

pxlrbt:
Sorry, noticed that while reading the original comment again, but forgot that I mentioned "non-numeric" 😅

nkeena:
Snowflakes are stored at integer types and usually get cast as strings to the frontend. I have a custom cast that does cast the snowflake ID to an integer when setting it on the model and then casting it to a string when retrieving it. However, it does not seem to cast the snowflake correctly when using Eloquent collection methods like `pluck` and `toArray()`. This is why I added the transform method.

I figured the `toString` in JS would do the trick but it seems the snowflake IDs are already invalid by the time they are sent to the frontend.

nkeena:
I did figure out a workaround to automatically cast the primary key as a string even when using the query builder. I created a global scope:

```php
    protected static function booted()
    {
        static::addGlobalScope('snowflake', function (Builder $builder) {
            $builder->withCasts([
                $builder->getModel()->getQualifiedKeyName() => SnowflakeCast::class
            ]);
        });
```

I do think the simple change to filament would make things easier for everyone.

pxlrbt:
Thanks for explaining what Snowflakes are. But I still don't understand, why integers were causing issues with record selection and how casting to string fixes this..

zepfietje:
Might it be that snowflake IDs overflow the [MAX_SAFE_INTEGER size](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER)?

danharrin:
Thats a very good spot

In which case, the JS cast wont work

We can remove that JS cast, and then merge this

zepfietje:
> Thats a very good spot
> 
> In which case, the JS cast wont work
> 
> We can remove that JS cast, and then merge this

Cool, I'll update it.

packages/tables/src/Concerns/CanSelectRecords.php
```diff
@@ -21,7 +21,10 @@ public function getAllTableRecordKeys(): array
     {
         $query = $this->getFilteredTableQuery();
 
-        return $query->pluck($query->getModel()->getQualifiedKeyName())->toArray();
+        return $query
+            ->pluck($query->getModel()->getQualifiedKeyName())
+            ->map(fn ($key): string => (string) $key)
+            ->toArray();
```
danharrin, line 27:
```suggestion
            ->all();
```

zepfietje: Was thinking about that one as well. 👍

zepfietje:
Ready to be merged, @danharrin!

zepfietje:
Lol, you're too fast.

PR #3764, opened 2022-08-30
Style the scrollbar in the side menu

dmandrade:
A simple style to improve the look of the side menu with a scrollbar.

NOTE: the scrollbar appears only on the sidebar hover

image
image

### DIFF
--- a/packages/admin/resources/css/app.css
+++ b/packages/admin/resources/css/app.css
@@ -49,3 +49,23 @@
         top: -1px;
     }
 }
+
+@layer utilities {
+
+    /* Scroll thumb styles */
+    .filament-sidebar-nav::-webkit-scrollbar {
+        width: .2rem;
+    }
+    .filament-sidebar-nav::-webkit-scrollbar-thumb {
+        background-color: transparent;
+        border-radius: 1rem;
+    }
+    .filament-sidebar-nav:hover::-webkit-scrollbar-thumb {
+        background-color: theme('colors.gray.300');
+    }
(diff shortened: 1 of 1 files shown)


awcodes:
Code looks good to me. 

I think it could be a problem on computers that have scrollbars always set to show (a11y issue). They may not support the styling then the scrollbar will probably creep into the icons.

Also an 8px wide bar is going to be hard to grab to use without a scroll wheel. But it is about the size of the default scrollbar on Mac Chrome. So probably not a big deal.

zepfietje:
I think the dark mode color needs tweaking.

danharrin:
Make changes to the PR if you want

packages/admin/resources/css/app.css
```diff
@@ -49,3 +49,23 @@
         top: -1px;
     }
 }
+
+@layer utilities {
```
zepfietje, line 53: I think the `utilities` layer is wrong for this goal.

zepfietje (changes requested):

packages/admin/resources/css/app.css
awcodes: I agree. Should either be in base or components.

zepfietje:
I think we should hold off for now and not customize the scrollbar by default in Filament admin. Browser support hasn't reached the point where we can achieve consistent styling between browsers.

We might revisit this in the future, but I'd say no for now.
@awcodes, @pxlrbt, @danharrin agree?

danharrin:
Agreed

PR #3768, opened 2022-08-31
Support allow use 'above', 'below' description for TextColumn at the same time

AAbosham:
Some time we need to use description on above TextColumn and below at the same time,

so that I added :

```
 ->descriptionAbove('Above description')
 ->descriptionBelow('Below description'))
```

### DIFF
--- a/packages/tables/resources/views/columns/text-column.blade.php
+++ b/packages/tables/resources/views/columns/text-column.blade.php
@@ -1,5 +1,6 @@
 @php
-    $description = $getDescription();
+    $descriptionAbove = $getDescriptionAbove();
+    $descriptionBelow = $getDescriptionBelow();
     $descriptionPosition = $getDescriptionPosition();
 @endphp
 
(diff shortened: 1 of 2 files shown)


Z3d0X:
Do we need dedicated methods?
With this implementation, wouldn't this work as well?
```php
TextColumn::make('name')
    ->description('Above description', 'above')
    ->description('Below description', 'below')
```

AAbosham:
> Do we need dedicated methods?
> With this implementation, wouldn't this work as well?

No, Its also working well,

```
TextColumn::make('name')
    ->description('Above description', 'above')
    ->description('Below description', 'below')
```
 or
```
TextColumn::make('name')
    ->descriptionAbove('Above description')
    ->descriptionBelow('Below description')
```

danharrin:
Yeah I don't think we need the extra methods or extra documentation, please remove those

AAbosham:
> Yeah I don't think we need the extra methods or extra documentation, please remove those. I'd rather not deprecate methods.

I was remove new method and use only description with position , also i remove descriptionPosition()
 
```
TextColumn::make('name')
    ->description('Above description', 'above')
    ->description('Below description', 'below')
```

Z3d0X:
> also i remove descriptionPosition()
This method existed before this PR, removing it would be a BC?

danharrin:
The documentation uses the parameter only, and this is a very new feature, so I'm not too worried

ralphjsmit:
> also i remove descriptionPosition()

It was a breaking change for my Notifications package :|

danharrin:
How do you suggest we add it back without breaking the existing functionality?

AAbosham:
It is working before i removed ,
when use descriptionPosision will only show description at above or below

ralphjsmit:
> How do you suggest we add it back without breaking the existing functionality?

I'm not suggesting you add it back, since it's already been released and I released a fix. I just would have preferred not to have any breaking changes until 3.x. Even though it's not documented, it is autocompleted. Plus, 3.x is only a few months away, and it would have been a simple deprecation.

I just wanted to note that even though things aren't documented, you can't fully trust that no one uses it.

ralphjsmit:
> It is working before i removed ,
> when use descriptionPosision will only show description at above or below

I like the idea of allowing a description both above and below. It's just that you _removed_ the method `descriptionPosition()`. That means that all existing code that used `->descriptionPosition()` will throw a fatal error after this change.

danharrin:
Yeah, it should be deprecated, and aliased to moving the `descriptionBelow` to `descriptionAbove` IF `above` is passed as a param

danharrin:
5f7072041d0490749ec69292ce18335ce9df8a11

Z3d0X:
Should also set below description to null in that 'if' condition, to get the intended old behaviour I think.

Right now it seems if you use `descriptionPosition('above')` it seems whatever set as belowDescription is also set to aboveDescription, hence repeated in both places. Correct me if I'm wrong

danharrin:
Good one

PR #3454, opened 2022-08-10
Support action groups in notifications

zepfietje:

PR #3443, opened 2022-08-09
Extract dropdown

zepfietje:
- [x] Update docs about configuring scripts
- [x] Dropdown panel width
- [x] Dropdown border opacity

PR #3771, opened 2022-08-31
fix: reintroduce dropdown changes

danharrin:
This PR contains the code from both #3443 and #3454 that needed to be reverted because all dropdowns ended up broken.

@zepfietje could you please investigate what was wrong, and add the fix into this branch? When I pulled those PRs down yesterday they worked fine for me, but not any more.


danharrin:
Updating NPM dependencies fixed the dropdown visual issue for whatever reason, but apparently we have reports of 404s after saving a form from this change as well

PR #3778, opened 2022-08-31
add property type

Kamandlou:


zepfietje:
It's not allowed to have a different signature for these properties than defined in the Laravel `Command` class. As you can see, this doesn't work and all checks are failing.

Kamandlou:
Yes, I got it.
Thanks for the explanation

PR #3786, opened 2022-08-31
heroicons v2

tanthammar:
This is a breaking change for plugins that use icons from Heroicons v1
But when the day comes where Filament wants to require heroicons v2, this PR is a good start.


tanthammar:
That was my though, a reference to changed icon names, a base for search-n-replace...

tanthammar:
But If you are reviewing all the icons, I guess you would catch the name changes then...

zepfietje:
This PR is definitely a valuable resource for when we switch to Heroicons v2, so thank you very much for that!

But as we can't merge it into `2.x` anyway, I'll close this PR.

PR #3788, opened 2022-08-31
fix rounded borders for dropdown panel

saade:
`overflow-hidden` was removed causing the items background to overflow the panel

before:
image

after:
image

### DIFF
--- a/packages/support/resources/views/components/dropdown/index.blade.php
+++ b/packages/support/resources/views/components/dropdown/index.blade.php
@@ -39,7 +39,7 @@
             wire:key="{{ $attributes->get('wire:key') }}.panel"
         @endif
         @class([
-            'filament-dropdown-panel absolute z-10 w-full rounded-lg bg-white shadow-lg ring-1 ring-black/5 transition',
+            'filament-dropdown-panel absolute z-10 w-full rounded-lg bg-white shadow-lg ring-1 ring-black/5 transition overflow-hidden',
             'dark:bg-gray-800 dark:ring-white/10' => $darkMode,
             match ($width) {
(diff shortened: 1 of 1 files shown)


zepfietje:
This will break select fields inside filters, as explained in Discord.

zepfietje:
Dropdowns with items should get some extra padding at the top and bottom. I'm planning to do some style fixes in a separate PR.

If this is really urgent for someone, you might implement a dirty fix of applying the border radius to the first and last items.

saade:
didnt see your message in discord, sorry

danharrin:
I don't think the filter should even use the dropdown Blade component... we should have a "popover" view for that without the overflow class

Or add a param to the dropdown Blade component that removes the overflow-hidden for the filters dropdown only

zepfietje:
Filters are not the only place where overflow should be allowed.

Fixing the padding on the items should fix the issue as well, and is in fact how that's handled in more places (e.g. Tailwind UI, GitHub).

saade:
> Filters are not the only place where overflow should be allowed.
> 
> Fixing the padding on the items should fix the issue as well, and is in fact how that's handled in more places (e.g. Tailwind UI, GitHub).

image

Since the border-radius is `0.5rem` adding a `0.5rem` padding does solves the issue, and actually its how github and tailwind takes care of that indeed. But the radius is smaller and thus the padding, wdyt?

i can reopen this with the padding if u guys want

zepfietje:
Feel free to reopen and then we'll discuss before merging.

Make sure to keep the dropdown base component without padding and only add it when there's items.

PR #3789, opened 2022-08-31
add pivot property to a record when a relationship is present

dododedodonl:
At this moment, filament gets all pivot data and all model data. When keys overlap (eg. created_at on the pivot table and on the model table), data from the model takes precedence. In order to access this data more easily, I tried to recreate the `pivot` property as laravel also uses it when a `BelongsToMany` relation is set.

I am not sure if the implementation is complete. Also, a mention in the docs needs to be added.

```php
<?php

namespace App\Filament\Resources\PostResource\RelationManagers;

use App\Models\Post;
use Filament\Resources\RelationManagers\RelationManager;
use Filament\Resources\Table;

class PostUserRelationManager extends RelationManager
{
    protected static string $relationship = 'users';
    //...

    {
        return $table
            ->columns([
                    ->getStateUsing(fn (Post $record) => $record->pivot?->created_at) // before this fix, this returns null
                    ->dateTime('d-m-Y H:i'),
            ]);
    }
    //...
}
```

### DIFF
--- a/packages/tables/src/Concerns/CanSelectRecords.php
+++ b/packages/tables/src/Concerns/CanSelectRecords.php
@@ -51,9 +51,9 @@ trait CanSelectRecords
         $pivotClass = $relationship->getPivotClass();
         $pivotKeyName = app($pivotClass)->getKeyName();
 
-        return $this->selectPivotDataInQuery(
+        return $this->hydratePivotRelation($this->selectPivotDataInQuery(
             $relationship->wherePivotIn($pivotKeyName, $this->selectedTableRecords),
-        )->get();
+        )->get());
     }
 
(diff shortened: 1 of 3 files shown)


kamilst96:
With this merged, I can't save pivot data in RelationManager. Has this changed the way we use pivots, or should I put an issue?

dododedodonl:
Could you provide a sample implementation? I am unable to reproduce it locally.

dododedodonl:
Could you try this branch: https://github.com/dododedodonl/filament/tree/fix/explicitly-filter-allows-duplicates

It adds a check to not select the extra when allowsDuplicates, which was the intention to.

zepfietje:
Related issue: #3826.

PR #3792, opened 2022-08-31
Document using `wire:model` in view fields

Shelob9:
I wasn't sure how to use `wire:model` today in a view field and wasn't sure how that would work. I wanted to document how this works for others.

### DIFF
--- a/packages/forms/docs/03-fields.md
+++ b/packages/forms/docs/03-fields.md
@@ -1912,13 +1912,13 @@ Using [Livewire's entangle](https://laravel-livewire.com/docs/2.x/alpine-js#shar
     :required="$isRequired()"
     :state-path="$getStatePath()"
 >
-    <div x-data="{ state: $wire.entangle('{{ $getStatePath() }}') }">
+    <div x-data="{ state: $wire.entangle('{{ $getStatePath() }}').defer }">
         <!-- Interact with the `state` property in Alpine.js -->
     </div>
 </x-forms::field-wrapper>
 ```
 
-You may bind the value to a Livewire property using [`wire:model`](https://laravel-livewire.com/docs/properties#data-binding):
+Or, you may bind the value to a Livewire property using [`wire:model`](https://laravel-livewire.com/docs/properties#data-binding):
 
 ```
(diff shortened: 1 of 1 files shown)


packages/forms/docs/03-fields.md
```diff
@@ -1916,6 +1918,23 @@ The `$getStatePath()` closure may be used by the view to retrieve the Livewire p
 </x-forms::field-wrapper>
 ```
 
+You can also bind with [`wire:model`](https://laravel-livewire.com/docs/properties#data-binding) :
```
zepfietje, line 1921:
```suggestion
You may bind the value to a Livewire property using [`wire:model`](https://laravel-livewire.com/docs/properties#data-binding):
```

packages/forms/docs/03-fields.md
```diff
@@ -1899,6 +1899,8 @@ Inside your view, you may interact with the state of the form component using Li
 
 The `$getStatePath()` closure may be used by the view to retrieve the Livewire property path of the field. You could use this to [`wire:model`](https://laravel-livewire.com/docs/properties#data-binding) a value, or [`$wire.entangle`](https://laravel-livewire.com/docs/alpine-js) it with Alpine.js:
```
zepfietje, line 1900:
```suggestion
The `$getStatePath()` closure may be used by the view to retrieve the Livewire property path of the field. You could use this to [`wire:model`](https://laravel-livewire.com/docs/properties#data-binding) a value, or [`$wire.entangle`](https://laravel-livewire.com/docs/alpine-js) it with Alpine.js.
```

zepfietje (changes requested):
Could you also move the `wire:model` example above the `$wire.entangle` example?

Shelob9:
@zepfietje I accepted your suggestions.

danharrin:
The entangle example should be first, that's much more common.

zepfietje:
Alright, I thought it made sense to have them in the order they're mentioned in the sentence above.

danharrin:
Ah I didn't see that, the sentence needs changing then

PR #3802, opened 2022-09-01
Automatically detect button and link tag

zepfietje:
This is especially useful when using the Filament button and link outside of Filament admin in your own project.

You no longer need to pass the `tag` attribute as it will automatically detect if an `href` has been passed and uses a button or link HTML tag accordingly.


danharrin:
Sorry, but I'd prefer if this is explicit.

zepfietje:
Hmm, you can always still pass it explicitly.

That being said, I've now added the logic to my own app where I extend the Filament support components.

PR #3805, opened 2022-09-01
fix: using record key instead of routeKey

saade:
PR #3733 introduced authorization into record create and edit actions.

While create seems fine, it broke the edit action because its not using the `routeKeyName`, so that leaves to a 404 while resolving the record.

on mount, it authorizes the `$record` wich is the "id" (in some cases uuid) from the record in the route
```php
    public function mount($record): void
    {
        $this->authorizeAccess($record); // record is the routeKey (in some cases the uuid)

        $this->fillForm();
    }
```

that passes since it resolves the records using its uuid (in my case)

on save, it authorizes again, but using the record key (primaryKey), wich leads to a 404, since its using the model id instead of the `routeKey` (uuid)

```php
public function save(bool $shouldRedirect = true): void
    {
        $this->authorizeAccess($this->getRecord()->getKey());
        ...
```

this PR fixes that behavior using the `routeKey` wich in my case is the uuid, but in other cases will default to the model id.

### DIFF
--- a/packages/admin/src/Resources/Pages/EditRecord.php
+++ b/packages/admin/src/Resources/Pages/EditRecord.php
@@ -72,7 +72,7 @@ class EditRecord extends Page implements HasFormActions
 
     public function save(bool $shouldRedirect = true): void
     {
-        $this->authorizeAccess($this->getRecord()->getKey());
+        $this->authorizeAccess($this->getRecord()->{$this->getRecord()->getRouteKeyName()});
 
         $this->callHook('beforeValidate');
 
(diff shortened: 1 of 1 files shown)


danharrin:
We don't need to resolve the record twice, just once when the component is mounted.

Your solution is flawed because if the user uses a route-specific binding attribute, the model's `routeKey` is incorrect.

I've fixed the PR, it still works, right?

danharrin:
Thanks for the investigation btw, not trying to be harsh, just dont want other problems 😅

saade:
no worries, glad i can unlock the filament dependency and use all the new stuff. thanks!

PR #3808, opened 2022-09-01
fix: Remove filters via the indicator icons

### LINKED ISSUE #3794: Table not updated when filter is removed via the filter indicator
### Package

filament/filament

### Package Version

v2.15.38

### Laravel Version

v9.27.0

### Livewire Version

_No response_

### PHP Version

PHP 8.0.14

### Problem description

I have a users model with a boolean 'confirmed' field and I've created a filament resource for it. I added this filter to the table in the UsersResource

```PHP
->filters([
    Tables\Filters\Filter::make('confirmed')
        ->query(fn($query) => $query->where('confirmed', true)),
])
```

I enable the filter and the table updates as expected to only show confirmed users. However when I click the 'x' icon in the 'Confirmed' filter indicator, the filter is removed but the table results stay the same until the page is reloaded.

### Expected behavior

The table should refresh to show all users and not just confirmed users.

### Steps to reproduce

Detailed steps are in the reproduction repository readme

### Reproduction repository

https://github.com/craigkuhns/filament-remove-filter-from-indicator-broken

### Relevant log output

_No response_

craigkuhns:
This fixes #3794.

The details bug are in that issue but essentially if you have a `Filament\Tables\Filters\Filter` added to a form, apply the filter, and then remove it either by clicking the icon in the filter indicator or the 'x' icon in the filter indicator row, the table does not update until the page is refreshed.

This fixes the issue by first making sure that the Filter sets a default state of false when it's using a Checkbox or Toggle component. Then when removing the filter it sets the state of the filter to the default state of the field.

I've added tests showing this fixes the issue and all other tests are passing. However I'm not sure if this breaks other filter types or filters with custom forms. I don't think it should since it only sets the default state to false for Checkbox and Toggle components and it should be null for all other components.

### DIFF
--- a/packages/tables/.stubs.php
+++ b/packages/tables/.stubs.php
@@ -97,6 +97,10 @@ namespace Livewire\Testing {
 
         public function resetTableFilters(): static {}
 
+        public function removeTableFilter(string $filter, ?string $field = null): static {}
+
+        public function removeTableFilters(): static {}
+
         public function assertCanSeeTableRecords(array | Collection $records, bool $inOrder = false): static {}
 
         public function assertCanNotSeeTableRecords(array | Collection $records): static {}
--- a/packages/tables/src/Concerns/HasFilters.php
+++ b/packages/tables/src/Concerns/HasFilters.php
@@ -84,7 +84,7 @@ trait HasFilters
 
         foreach ($fields as $field) {
             $field->state(
-                is_array($field->getState()) ? [] : null,
+                is_array($field->getState()) ? [] : $field->getDefaultState(),
             );
         }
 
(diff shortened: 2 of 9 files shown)


craigkuhns:
A few other things. I updated the `checkbox()` and `toggle()` methods of the Filter class to set the default state too false. The `formComponent()` method is also updated to set the default state to `null` which is what it defaulted to before this fix. A potential problem is that if someone registers a Filter and calls `default()` and then `formComponent()` like so:
```php
Filter::make('confirmed')
    ->default('some default value')
    ->formComponent(CustomFormComponent::class)
```
it will override the default value the user explicitly set themselves. I'm not sure of the best way to get around that.

danharrin:
Originally, we reset to the default state, but then there was this issue - https://github.com/filamentphp/filament/pull/3599#issuecomment-1225000368

I think we just need an extra check - if the state is `true`, set it to `false`. Else, set it to `null`.

Also we need to make sure that `default()` is only called on the form component if there is a non-null default value defined on the filter.

danharrin:
Can you double check if my changes still fix all your issues?

craigkuhns:
@danharrin that does fix the issue.

PR #3819, opened 2022-09-02
Move text input affixes inside border

fira03079508:
This is my first Pull Request on any project and actually my first time using TailwindCSS.

What I have in mind is that prefixIcon and suffixIcon should ideally be inside the input and not outside.
I've changed some classes to allow that putting a pl-10 only if there's a prefixIcon (so you don't write inside the icon)

Maybe we should approach this as an option and maybe allow this for prefix / suffix and maybe even prefixAction / suffixAction

As it's my first time using TailwindCSS. I don't know the impact of changing the outer div to relative instead of the classes that were there

image

So here's my first attempt


packages/forms/resources/views/components/text-input.blade.php
```diff
@@ -71,6 +73,7 @@
                 {{ $getExtraInputAttributeBag()->class([
                     'block w-full transition duration-75 rounded-lg shadow-sm focus:border-primary-600 focus:ring-1 focus:ring-inset focus:ring-primary-600 disabled:opacity-70',
                     'dark:bg-gray-700 dark:text-white dark:focus:border-primary-600' => config('forms.dark_mode'),
+                    $getPrefixIcon ? 'pl-10' : '',
```
fira03079508, line 76: Forgot that it would also make sense to put pr-10 if a suffixIcon is present

zepfietje:
If we're going to move the prefix icon into the box, we should consider moving the prefix label as well. Have a look at how Tailwind UI handles this: https://tailwindui.com/components/application-ui/forms/form-layouts#component-db11f83176d113e39bf2559da9344b1c.

Personally I like the design of containing everything within the bordered box better than having the affixes outside.

danharrin:
Yes, I agree.

fira03079508:
Does this mean I should bring the changes to a new commit or these are just things that could be improved upon whoever does it? As I said above I'm really not experienced with this.

packages/forms/resources/views/components/text-input.blade.php
cobraprojects: you should make another class for `rtl` please change `pl-10` to `ltr:pl-10 rtl:pr-10`

zepfietje:
> Does this mean I should bring the changes to a new commit or these are just things that could be improved upon whoever does it? As I said above I'm really not experienced with this.

You can just work on this PR/branch to implement mentioned improvements.
If you need help, we can take over.

fira03079508:
I've tried adding this to labels also (not just Icons) and I had some problems with my Tailwind config. I wasn't able to check if everything was working correctly... In theory from what I know so far it should work but if someone could check if it works nicely visually please do.

packages/forms/resources/views/components/text-input.blade.php
```diff
@@ -31,7 +31,7 @@
         @endif
 
         @if ($label = $getPrefixLabel())
-            <span @class($affixLabelClasses)>
+            <span @class(array_merge($affixLabelClasses , ['inline-flex items-center rounded-l-lg border border-r-0 border-gray-300 bg-gray-50 px-3 text-gray-500']))>
```
zepfietje, line 34: Could you add the common affix label classes to `$affixLabelClasses`?

zepfietje (changes requested):

packages/forms/resources/views/components/text-input.blade.php
fira03079508: Oh yeah totally my bad will do this right now

zepfietje (changes requested):
Is it possible to move the `rounded` classes to the wrapping `div`? I think it should, but I'm not sure just judging from the code.

fira03079508:
> Is it possible to move the `rounded` classes to the wrapping `div`? I think it should, but I'm not sure just judging from the code.

Not sure, It's my first time using Tailwind. I've tried my best and wouldn't be offended at all if anything here is changed but I don't really know what I could do. Also i'm having issues testing my changes at the moment

zepfietje:
> > Is it possible to move the `rounded` classes to the wrapping `div`? I think it should, but I'm not sure just judging from the code.
> 
> Not sure, It's my first time using Tailwind. I've tried my best and wouldn't be offended at all if anything here is changed but I don't really know what I could do. Also i'm having issues testing my changes at the moment

Alright, we'll have a look later!

zepfietje:
@danharrin, do you think this can be added to v2?

This is quite a design change, so might hold off for v3?

danharrin:
Yeah, let's hold off on merging until the Design team has reviewed this as part of wider changes in v3

zepfietje:
Closing this PR as v3 will require it to target a different branch.

We're also tracking this on our roadmap: https://github.com/orgs/filamentphp/projects/2/views/6?query=is%3Aopen+sort%3Aupdated-desc.

PR #3827, opened 2022-09-02
fix: Check if allows duplicates before hydrating pivot relation

### LINKED ISSUE #3826: Edit pivot attribute bug in v2.15.40
### Package

filament/filament

### Package Version

v2.15.40

### Laravel Version

v9.27.0

### Livewire Version

v2.3

### PHP Version

PHP 8.1

### Problem description

After upgrading to v2.15.40 pivot attributes are not updating when editing a relation in RM
If downgrading to v2.15.39 there is no issue

### Expected behavior

Update notes pivot attribute

### Steps to reproduce

Edit a record in RM

### Reproduction repository

https://github.com/molcsab/pivotbug

### Relevant log output

```shell
No errors, but the query in debugbar looks like this:
update `role_user` set `notes` = 'test' where `user_id` is null and `role_id` is null
```

dododedodonl:
Possible fix for issue #3826

### DIFF
--- a/packages/tables/src/Concerns/HasRecords.php
+++ b/packages/tables/src/Concerns/HasRecords.php
@@ -91,10 +91,6 @@ trait HasRecords
 
         $record = $this->selectPivotDataInQuery($query)->first();
 
-        if(! $this->allowsDuplicates()) {
-            invade($relationship)->hydratePivotRelation([$record]);
-        }
-
         return $record?->setRawAttributes($record->getRawOriginal());
     }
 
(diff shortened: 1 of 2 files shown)


danharrin:
In the reproduction issue @molcsab submitted, the relation manager doesn't allow duplicates? So I guess this isnt the right fix

I am tempted to revert the original PR for now

dododedodonl:
Removing the `hydratePivotRelation()` call from the `resolveTableRecord()` method solves the update issue

sumardi:
The latest changes fix the issue.

PR #3832, opened 2022-09-02
fix: Filter dropdown width

zepfietje:


packages/tables/resources/views/components/toggleable/index.blade.php
```diff
@@ -1,12 +1,13 @@
 @props([
     'form',
-    'width' => 'sm',
+    'width' => null,
```
danharrin, line 3: Null width as a default?

zepfietje: Yeah, then it uses a sensible default equal to `w-56` (but as a max width since it was hard to implement width using Floating UI).

Couldn't achieve that using the current width options that depend on Tailwind's `max-w-` classes, as it would be breaking if I changed those.

zepfietje: But we could actually remove that `null` from there and depend on the default of `null` returned from `getColumnToggleFormWidth` via `getTableColumnToggleFormWidth`.

PR #3835, opened 2022-09-02
Fix: changed from array_merge to array_replace to avoid reindexing in…

MACscr:
When using the new `hasCombinedRelationManagerTabsWithForm()` method, RM keys would get reindex by the use of array_merge() and if an RM is hidden due to canViewForRecord(), the keys would be wrong and cause an error of `Undefined array key 3` (or whatever RM is hidden) when selected.

### DIFF
--- a/packages/admin/resources/views/components/resources/relation-managers/index.blade.php
+++ b/packages/admin/resources/views/components/resources/relation-managers/index.blade.php
@@ -15,7 +15,7 @@
                     $tabs = $managers;
 
                     if ($form) {
-                        $tabs = array_merge([null => null], $tabs);
+                        $tabs = array_replace[null => null], $tabs);
                     }
                 @endphp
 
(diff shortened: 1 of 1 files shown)


packages/admin/resources/views/components/resources/relation-managers/index.blade.php
```diff
@@ -15,7 +15,7 @@
                     $tabs = $managers;
 
                     if ($form) {
-                        $tabs = array_merge([null => null], $tabs);
+                        $tabs = array_replace[null => null], $tabs);
```
ryangjchandler, line 18: Missing an open / left paren here.

ryangjchandler (changes requested):

MACscr (commented):
My fault. The change was so simple that i simply edited the file in github as my original PR was to the wrong repo.

PR #3846, opened 2022-09-03
DateFilter

dmandrade:
A date filter using DatePicker component.

- can set maxDate, minDate and timezone
- can create a range filter with `range()` method


danharrin:
Hey, thank you for the contribution, but let's keep this in a plugin. I think that it might even be a plugin already. The implementation of such a filter is just too subjective for an official feature IMO

PR #3847, opened 2022-09-03
feature: Database and Broadcast notifications

danharrin:
- [x] Polling
- [x] Echo
- [x] Include button view inside the Livewire component so notification counter can be dynamically updated
- [x] Extract views
- [x] Extract translations
- [x] Docs


packages/notifications/resources/views/components/close-button.blade.php
```diff
@@ -1,4 +1,4 @@
 <x-heroicon-s-x
-    class="filament-notifications-close-button h-6 w-5 cursor-pointer text-gray-400"
+    class="filament-notifications-close-button h-4 w-4 cursor-pointer text-gray-400"
```
zepfietje, line 2: I believe 20px size matches Tailwind UI.

packages/notifications/src/Http/Livewire/Notifications.php
```diff
@@ -34,9 +34,31 @@ public function pullNotificationsFromSession(): void
...
     }
 
+    public function clear(): void
+    {
+        auth()->user()->notifications()->delete();
+    }
+
+    public function markAllAsRead(): void
+    {
+        auth()->user()->notifications()->update(['read_at' => now()]);
```
zepfietje, line 44:
```suggestion
        auth()->user()->unreadNotifications()->update(['read_at' => now()]);
```

packages/support/resources/views/components/modal/index.blade.php
```diff
@@ -96,57 +113,71 @@ class="fixed inset-0 z-40 flex items-center min-h-screen p-4 overflow-y-auto tra
                     'max-w-7xl' => $width === '7xl',
                 ])
             >
-                @if ($header)
-                    <div class="filament-modal-header px-4 py-2">
-                        {{ $header }}
+                @if ($closeButton)
+                    <div class="absolute top-2 right-2">
+                        <x-heroicon-s-x
+                            class="filament-notifications-close-button h-4 w-4 cursor-pointer text-gray-400"
```
zepfietje, line 119:
```suggestion
                            class="filament-modal-close-button h-4 w-4 cursor-pointer text-gray-400"
```

zepfietje (changes requested):

packages/notifications/resources/views/components/close-button.blade.php
danharrin: But it looks bad in this case 😬

zepfietje: Hmm, the same icon will also be added to modals in #3911. We should try to be consistent, and I 20px size looks totally fine actually.

themahdavi:
A small suggestion for improve performance:

When installing pusherjs and laravel-echo by default on core of filament but many users don't use that on projects totally just increase js file size and useless js code.

I think if we have option when installing filament admin panel and notification package standalone like `--pusher` or `--database` it's nice when user use these option install pusherjs or use notification with database

danharrin:
Those libraries are roughly 24kb, so I am not worried at all about performance. Adding them as optional dependencies would be a bit of a pain that I don't want to worry about right now.

danharrin:
@themahdavi I did it anyway, wasn't actually too much work

themahdavi:
Thanks 🙏

PR #3849, opened 2022-09-04
Fixed broken table header layout while reordering when table has header actions.

theriftlab:
When a table defines both `getTableHeaderActions()` and `getTableReorderColumn()`, clicking the reorder icon to trigger reordering breaks the layout when the header is hidden. When the search bar is visible (which I'm sure is true for most use cases) it drops down under the reorder icon and some padding is also lost. This fix simply hides the header div via Tailwind instead of removing it from the DOM, which resolves the issue.

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -239,8 +239,10 @@
         >
             @if ($header)
                 {{ $header }}
-            @elseif ($heading || ($headerActions && (! $isReordering)))
-                <div class="px-2 pt-2">
+            @elseif ($heading || $headerActions)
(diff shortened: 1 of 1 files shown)


packages/tables/resources/views/index.blade.php
```diff
@@ -239,8 +239,10 @@ class="filament-tables-component"
         >
             @if ($header)
                 {{ $header }}
-            @elseif ($heading || ($headerActions && (! $isReordering)))
-                <div class="px-2 pt-2">
+            @elseif ($heading || $headerActions)
+                <div class="px-2 pt-2" x-bind:class="{
```
danharrin, line 243: Why x-bind the class instead of using `@class` in Blade?

PR #3851, opened 2022-09-04
Fixed bulk-select/deselect not working after reorder toggled.

theriftlab:
Select / deselect all functionality (both the select-all checkbox and the select / deselect links) breaks after a table reorder has been toggled. This appears to be due to the conditional switching of the reorder cell & checkbox cell affecting the `x-model` binding on the checkboxes. This fix inserts all cells unconditionally and then conditionally hides them with Tailwind. Reorder functionality is unaffected and bulk select / deselect functionality is now retained after reorder.


danharrin:
Thanks

theriftlab:
@danharrin this seems to have partially undone the fix in #3849 - toggling reordering now only hides the header actions instead of hiding the whole header as before. Is this intentional? (Line 243 in index.blade.php).

danharrin:
Yeah I think that's fine, what do you think?

danharrin:
Does it actually cause issues if the heading and description are visible?

theriftlab:
No issues & I agree it's a good idea to _not_ hide the heading & description when reordering, my only (small) niggle is that when there is no header or description, when the header action button disappears there is still a grey border visible.

Screen Shot 2022-09-10 at 11 02 23 AM

theriftlab:
I can submit a PR for this too if necessary but it is a very minor thing. Overall I'm very happy with it.

danharrin:
Yes please, a PR would be great

PR #3853, opened 2022-09-04
fix: Dropdown styling

zepfietje:
This PR adds a dropdown group component which solves the overflowing items.

I've done some research and think for now it's okay to still use the dropdown component for the other popovers like filters and toggleable columns. The new dropdown group handles the padding around items, so using the dropdown without it will give you a panel without padding.

Before:
Screen Shot 2022-09-04 at 14 28 58

After:
Screen Shot 2022-09-04 at 14 29 15

I went for the padded design with rounded dropdown items, since it's hard to make it work without overflowing the panel due to the large border radius we're using. I've added some additional CSS hook classes as well for people to change the design to their liking.

I've got some ideas for additional improvements (mostly colors as the hover style is too obtrusive right now). But I will leave that for a separate PR.


dododedodonl:
Tested it on one of my filament admin setups, works as expected and solves the UI bug.

danharrin:
Unfortunately, moving the item Blade component is a breaking change that will need to wait until v3

zepfietje:
> Unfortunately, moving the item Blade component is a breaking change that will need to wait until v3

This should work:

- https://github.com/filamentphp/filament/pull/3853/commits/dbcc72fa48a5ce8bb3429e8f6bfcd5cebd44fa5f
- https://github.com/filamentphp/filament/pull/3853/commits/23588a3745ecbb96ea4a839c7395b006df09611a

PR #3856, opened 2022-09-04
suggestion to make the class more flexible

egyjs:
Greetings,

 I made this commit, and I believe it's a better way than adding the whole logic to one function,


zepfietje:
Thanks for contributing!

However, I'm not sure about some of the naming changes.
Also, it seems like the code was structured this way originally on purpose to have a better overview of the code it writes from the stubs.

Thanks again, but I think I'll have to close this PR.

PR #3862, opened 2022-09-05
fix: button overflow

4m1n0s:
Dropdown have a rounded borders, and buttons inside will overflow

### Before
Before

### After
After


zepfietje:
Thanks for contributing! However, this is will already be solved by https://github.com/filamentphp/filament/pull/3853.

PR #3872, opened 2022-09-05
Absolute option possible for Resource::getUrl()

dejury:
Adds the possibility to set the $absolute parameter to false when retrieving a resource url.

### DIFF
--- a/packages/admin/src/Resources/Resource.php
+++ b/packages/admin/src/Resources/Resource.php
@@ -368,11 +368,11 @@ class Resource
             ->implode('/');
     }
 
-    public static function getUrl($name = 'index', $params = [], $absolute = true): string
+    public static function getUrl($name = 'index', $params = [], $isAbsolute = true): string
     {
         $routeBaseName = static::getRouteBaseName();
 
-        return route("{$routeBaseName}.{$name}", $params, $absolute);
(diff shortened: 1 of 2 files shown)


ryangjchandler (approved):

packages/admin/src/Resources/Resource.php
```diff
@@ -368,11 +368,11 @@ public static function getSlug(): string
             ->implode('/');
     }
 
-    public static function getUrl($name = 'index', $params = []): string
+    public static function getUrl($name = 'index', $params = [], $absolute = true): string
```
danharrin, line 371:
```suggestion
    public static function getUrl($name = 'index', $params = [], $isAbsolute = true): string
```

danharrin (changes requested):
Also, please add this parameter to the Page class `getUrl()`

packages/admin/src/Resources/Resource.php
danharrin: Yeah but since this code is in the Filament domain, we should keep the parameter names consistent with the rest of our code.

dejury: Agree, changed it!

danharrin:
Thanks :)

PR #3880, opened 2022-09-05
Enhancement: Sticky header integration

awcodes:
Integrates sticky header functionality into core.


packages/admin/resources/js/app.js
```diff
@@ -70,6 +70,29 @@ window
 Chart.defaults.font.family = `'DM Sans', sans-serif`
 Chart.defaults.color = '#6b7280'
 
+const stickyTrigger = document.querySelector('.filament-sticky-trigger')
```
zepfietje, line 73: We don't need JS for this, right?

awcodes: Yes we do. The intersection observer needs an element to watch. It's the only way to add a class that can be used to style elements when they are stuck. There's currently no way to target a stuck element in CSS.

zepfietje: What exactly do we need that styling target for?

Sorry if I'm missing something, but the sticky header can be achieved with pure CSS, right? Or is there more to it?

packages/admin/resources/views/components/layouts/app.blade.php
```diff
@@ -19,7 +19,7 @@ class="filament-sidebar-close-overlay fixed inset-0 z-20 w-full h-full bg-gray-9
             @if (config('filament.layout.sidebar.is_collapsible_on_desktop'))
                 x-data="{}"
                 x-bind:class="{
-                    'lg:pl-[5.4rem] rtl:lg:pr-[5.4rem]': ! $store.sidebar.isOpen,
+                    'lg:pl-[5.4em] rtl:lg:pr-[5.4em]': ! $store.sidebar.isOpen,
```
zepfietje, line 22: This seems like an unrelated change?

packages/admin/resources/js/app.js
```diff
@@ -70,6 +70,29 @@ window
...
+            },
+            {
+                threshold: [0],
+            },
+        )
+
+        observer.observe(stickyTrigger)
+    })
+}
+
```
zepfietje, line 95: Could this be moved into the header Blade component, @awcodes?

packages/admin/resources/css/app.css
```diff
@@ -48,4 +48,35 @@
...
+        padding-inline: theme('spacing.4');
+        padding-block: theme('spacing.3');
+        z-index: 20;
+        min-height: 4rem;
+        align-items: center;
+        background-color: theme(colors.white / 75%);
+        border-color: theme('colors.gray.200');
+        border-bottom: solid 1px;
+        backdrop-filter: blur(12px);
+    }
```
zepfietje, line 81: I think it would be better if this was moved to the respective Blade components.

zepfietje (changes requested):

danharrin:
This seems to be glitchy in Safari?

https://user-images.githubusercontent.com/41773797/190621297-8a3a4fc6-79b9-4ec8-b03a-d8a26ea06633.mov

zepfietje:
@danharrin, note also that we still have to discuss if and how we would like this in core. @awcodes set up this PR after I told him we would be open for sticky headers in core, but I misunderstood sticky headers for the sticky topbar.

I think we could actually have this in core, but maybe optional? Also, we could have the actions row stick below the topbar so that's also still visible?

awcodes:
This does have issues with Safari < 16. 

Bigger problem is the stacking context with the notifications being inside the topbar. 

So, either way, probably shouldn't worry about this in core until that is resolved.

Might be better to just wait for v3 on this one.

zepfietje:
Why would this be breaking in a way that it needs to be held off until v3, @awcodes?

packages/admin/resources/views/components/layouts/app.blade.php
awcodes: This is an artifact. I'll fix it.

packages/admin/resources/js/app.js
awcodes: Could be moved to the blade component. My reasoning is that it'd be need on the majority of the pages so it'd be better to have it minified as part of the bundle.

packages/admin/resources/css/app.css
awcodes: Could be moved to the blade component. My reasoning is that it'd be need on the majority of the pages so it'd be better to have it minified as part of the bundle.

awcodes:
Doesn't have to wait, just know that since it has conflicts with notifications and z index that it'd be easier to do with teleport in livewire v3.

In order to do it now we'd have to move notifications out of the top bar and split them into two components with listens and events.

zepfietje:
What if the sticky header is sticky in addition to the sticky topbar (so not 'replacing' the topbar)? Would that solve the issue as well?

awcodes:
I'd have to runs some tests for this. It's not possible to stick to a stick in css so it might only work if the topbar has a permanent height or some things get moved around in the dom tree. I'm open to it though.

zepfietje:
Hmm, yeah maybe makes it more difficult only.

packages/admin/resources/js/app.js
zepfietje: Hmm I see. Personally, I also like having it in the context it's used instead of the general `app.js` file.

packages/admin/resources/views/components/layouts/app.blade.php
zepfietje: Already fixed by @danharrin.

packages/admin/resources/css/app.css
zepfietje: That defies the purpose of Tailwind CSS, doesn't it? 😉

awcodes:
@danharrin @zepfietje Do we even want this in core? I'll need to redo some things with this if we do.

danharrin:
I think probably not atm, but its up to you. Would it be an easy plugin?

zepfietje:
@danharrin, there's already a plugin for this by @awcodes.

Might be better to keep it as a plugin instead of in core.
Is there anything we could change in core to fix the stacking context issue?

awcodes:
It's already a plugin, but the notifications are still a problem the scoping index.

I'm ok to leave it as a plugin.

danharrin:
Sure, you can fix the stacking issue if you want

awcodes:
@dan, you're ok with taking the notification out of the topbar?

danharrin:
Well it still needs to update the counter when there are new database notifications, but yes

PR #3882, opened 2022-09-06
Enhancement: make filament-main-topbar sticky

awcodes:


mikrosmile:
I just yesterday night was thinking about adding It )) thx. I tried it and noticed that header is transparent and overlapping content, so I added bg-white

packages/admin/resources/views/components/layouts/app.blade.php
```diff
@@ -3,7 +3,7 @@
 ])
 
 <x-filament::layouts.base :title="$title">
-    <div class="filament-app-layout flex w-full min-h-screen overflow-x-hidden">
+    <div class="filament-app-layout flex w-full min-h-screen overflow-x-clip">
```
zepfietje, line 6: What does this fix? Curious. 😊

packages/admin/resources/views/components/layouts/app.blade.php
```diff
@@ -31,7 +31,7 @@ class="filament-sidebar-close-overlay fixed inset-0 z-20 w-full h-full bg-gray-9
             ])
         >
             <header @class([
-                'filament-main-topbar h-[4rem] shrink-0 w-full border-b flex items-center',
+                'filament-main-topbar h-[4rem] shrink-0 w-full border-b flex items-center sticky top-0 z-10',
```
zepfietje, line 34: We're missing a background color, it's transparent now.

zepfietje (changes requested):

zepfietje:
> I just yesterday night was thinking about adding It )) thx. I tried it and noticed that header is transparent and overlapping content, so I added bg-white.
> Also would be great to have an option in config for future consideration

I've fixed the background issue!

Maybe we could add a config key, but I'm not sure if we really should add config items for everything. 🤔

packages/admin/resources/views/components/layouts/app.blade.php
awcodes: It's a quirk with sticky. To be sticky the parent element has to be overflow hidden. Since the main content elements aren't directly inside the app layout it fixes that problem allowing anything inside the scrollable area to have the same perceived top to stick too. Functionality it's the same as overflow hidden.

danharrin:
I am interested - if the `sticky` class is removed, does the topbar revert back to the exact same behaviour as before without regressions?

In which case, people could just turn this on themselves in CSS:

```css
.filament-main-topbar {
    position: sticky;
}
```

and we can document this as the official solution?

CSS for configuring the appearance >>> Laravel config IMO

awcodes:
I still think it's better to have this with the new menu toggle button. Since it is part of the topbar it also scrolls out of view creating a weird UI experience.

zepfietje:
I agree the sticky topbar could be the default.

Reverting to the old behavior is also an easy change using CSS.

danharrin:
Alright, so use this as default & then add docs to revert:

```css
.filament-main-topbar {
    position: relative;
}
```

zepfietje:
I've added docs; this is now ready for final review.

PR #3823, opened 2022-09-02
New navigation menu button design

tanthammar:
# This PR does two things:

1. Moves the navbar button to a separate blade file.
    1. To accommodate the requests to make this a configurable option.
3. Changes the design of the navbars menu button on @media size 'lg'
    1. Mobile device design remains the same

### Why?
The first time I opened Filament Admin dashboard, I felt that the hamburger icon was misplaced in desktop view. 
It didn’t look related to the expanded sidebar. But of course when clicking it… 

Yes, all developers knows what the "hamburger" menu icon means. But not all users do. 
We need to make self explanatory design wherever possible.

The Mobile device design remains unchanged because the 'hamburger' icon makes perfect sense there.

One may argue that users can get confused by this design change.
In this case you can publish the nav-menu-button.blade.php and keep the "old" design.

### Preview

https://user-images.githubusercontent.com/21239634/188135915-3df72c2b-bb1b-4670-92b4-8c3335c1324b.mov

### Poll response from Discord:

image

# Question
Do I need to `npm run build:styles`?

PR #3885, opened 2022-09-06
Improve collapsible sidebar on desktop

tanthammar:
### Adds config for different sidebar designs
- optional fully collapsed sidebar
- config for free sized, sidebar width
- optional arrow or hamburger icon

#### Collapsed `5.4em` width + arrow icon (default)
```php
'sidebar' => [
            'is_collapsible_on_desktop' => true,
            'hamburger_icon_on_desktop' => false,
            'width' => '20rem',
            'collapsed_width' => '5.4em',
        ],
```
image

#### Fully collapsed with hamburger icon
```php
'sidebar' => [
            'is_collapsible_on_desktop' => true,
            'hamburger_icon_on_desktop' => true,
            'width' => '20rem',
            'collapsed_width' => '0px',
        ],
```
image

#### Collapsible, 5.4em with hamburger icon (default before #3823 )
```php
'sidebar' => [
            'is_collapsible_on_desktop' => true,
            'hamburger_icon_on_desktop' => true,
            'width' => '20rem',
            'collapsed_width' => '5.4em',
        ],
```
image

**TODO**, add examples to documentation


packages/admin/config/filament.php
```diff
@@ -207,10 +207,12 @@
         ],
         'sidebar' => [
             'is_collapsible_on_desktop' => false,
+            'hamburger_icon_on_desktop' => false,
             'groups' => [
                 'are_collapsible' => true,
             ],
-            'width' => null,
+            'width' => '20rem',
```
danharrin, line 214: Lets keep the `width` & `collapsed_width ` defaults in the views instead, because then we can change them later if we want without making people republish configs

packages/admin/config/filament.php
```diff
@@ -207,10 +207,12 @@
         ],
         'sidebar' => [
             'is_collapsible_on_desktop' => false,
+            'hamburger_icon_on_desktop' => false,
```
danharrin, line 210:
```suggestion
            'has_hamburger_icon_on_desktop' => false,
```

margarizaldi: What about `'use_hamburger_icon_on_desktop'` since it is a replacement instead of an addition to existing chevron.

zepfietje:
@danharrin, @awcodes, @pxlrbt discussed the sidebar icon situation again on Discord.
We agreed that the chevron should probably be part of the sidebar instead of floating outside.

Essentially, the fully collapsible and partially collapsible (resizable) sidebar are two distinct features. What I suggested (and @awcodes confirmed):

- Collapsible (fully): show hamburger menu icon in topbar
- Resizable (non-fully collapsible): show chevron icon in sidebar

Technically, this way both features can even coexist at the same time.

packages/admin/config/filament.php
tanthammar: I understand. 
My thought, when adding the default value to config, was that we publish the config when we want to change the styling to ... whatever, and when doing so, we don't have to dig into the code to see what the current default value is?
I don't mind keeping it as `null`, just commenting on why I made the change.

packages/admin/resources/views/components/layouts/app/topbar/collapse-sidebar-button.blade.php
```diff
@@ -1,24 +1,37 @@
...
-        'filament-sidebar-open-button shrink-0 flex items-center justify-center w-10 h-10 transition text-primary-500 rounded-full hover:bg-gray-500/5 focus:bg-primary-500/10 focus:outline-none lg:z-10 lg:hover:bg-white lg:focus:bg-white lg:hover:text-gray-600 lg:focus:text-gray-600 lg:-ml-[44px] lg:w-6 lg:h-6 lg:text-gray-500 lg:bg-white lg:shadow',
-        'lg:dark:bg-gray-900 lg:dark:hover:text-gray-400 lg:dark:focus:text-gray-400' => config('filament.dark_mode'),
-        'lg:hidden' => ! config('filament.layout.sidebar.is_collapsible_on_desktop'),
-        'lg:mr-4 rtl:lg:ml-4 rtl:lg:-mr-[44px]' => config('filament.layout.sidebar.is_collapsible_on_desktop'),
-    ])
->
-    <div class="relative lg:absolute visible lg:invisible">
-        <x-heroicon-o-menu class="w-6 h-6" />
-    </div>
+@if(config('filament.layout.sidebar.hamburger_icon_on_desktop', true))
```
tanthammar, line 1: Don't forget to change this if config is renamed to: `has_hamburger_icon_on_desktop`

tanthammar:
> @danharrin, @awcodes, @pxlrbt and I just discussed the sidebar icon situation on Discord. We agreed that the chevron should probably be part of the sidebar instead of floating outside.
> 
> Essentially, the fully collapsible and partially collapsible (resizable) sidebar are two distinct features. What I suggested (and @awcodes confirmed):
> 
> * Collapsible (fully): show hamburger menu icon in topbar
> * Resizable (non-fully collapsible): show chevron icon in sidebar
> 
> Technically, this way both features can even coexist at the same time.

I played around with moving the chevron icon to the sidebar instead of floating it when i did the other PR.
It was very tricky to keep it in the correct position in relation to the logo, but I am sure that someone more prominent in CSS than I don't have any problem fixing that.
I am up for any solution that keeps things open for individual styling, such as collapsed/expanded width, hamburger/chevron icon. 

Does this mean that this PR is obsolete, or is it of interest to merge it until the day you decide to do the changes you describe?

packages/admin/resources/views/components/layouts/app/topbar/collapse-sidebar-button.blade.php
tanthammar: done

zepfietje:
@tanthammar, I think we can keep this PR, but change the goal to what I described as discussed by the maintainers.

So, to be clear, there's no config for hamburger vs. chevron. These trigger icons are tied to their function, either collapsing or resizing (the current collapse behavior) the sidebar.

I don't know what exactly you ran into with styling, but you can probably swap the chevron and logo positions using [`flex-row-reverse`](https://tailwindcss.com/docs/flex-direction).

If you need any help, just ask! 🙂

packages/admin/config/filament.php
```diff
@@ -207,10 +207,12 @@
         ],
         'sidebar' => [
             'is_collapsible_on_desktop' => false,
+            'has_hamburger_icon_on_desktop' => false,
```
zepfietje, line 210: So this config key will probably become something like:

```suggestion
            'can_be_hidden_on_desktop' => false,
```

packages/admin/config/filament.php
```diff
@@ -207,10 +207,12 @@
...
             'groups' => [
                 'are_collapsible' => true,
             ],
             'width' => null,
+            'collapsed_width' => null,
```
zepfietje, line 215: Do we need a config key for this, or is it easy to style with CSS using the `filament-main` and `filament-sidebar` hooks?

Edsardio:
Saw this PR and not sure if it's out of scope or not. But since this is being worked on anyways I wanted to give it a shot.
Is it possible to add an option that determines if the sidebar should be open or closed on first entering a website? I believe it's currently hardwired to be false (meaning always closed on first load).

tanthammar:
@zepfietje, @danharrin 

Please confirm that my task is to rework the PR into:
- a config for the sidebars collapse behaviour – fully (collapsible), partially (resizable)?
- move the chevron icon away from a separate blade component into the sidebar `<header>`?
- keep the hamburger icon, in a separate blade component, or move it back to where it was?
- which icon is visible, is controlled by the sidebar collapse behaviour configuration?
    - collapsible = hamburger
    - resizable = chevron
- remove config for sidebar width?
- remove config for sidebar collapsed_width?
- let sidebar widths be altered via custom css, not `css root:  --var` + config?
- config for whether the initial collapse state should be open or closed (requested by @Edsardio )?

I am not sure I agree with your strategy, but it is your repo ...

danharrin:
## Width

Originally we used config to control this, so we need to keep doing that. The default values should exist within the view files instead of the config files, which makes it easier for us to change them in the future.

The collapsed sidebar width can be added to the config file in exactly the same way as the width was configured previously.

## "Collapsing"

I see that the sidebar should have 3 states, "open", "closed", and "collapsed".

### Default

The sidebar is open on desktop. There is no option to close or collapse it.

On mobile, it is closed by default, and can be opened by clicking a "hamburger" icon.

### `is_collapsible_on_desktop` config option

Show the "chevron" icon on desktop. Clicking this allows you to toggle between open and collapsed states.

A new `has_hamburger_icon_on_desktop` is available to swap the "chevron" icon for a "hamburger" on desktop.

On mobile, it is closed by default, can be opened by clicking a "hamburger" icon.

### `is_closable_on_desktop` config option

Show the "hamburger" icon on desktop. Clicking this allows you to toggle between open and closed states, which completely hides the sidebar.

A new `has_hamburger_icon_on_desktop` is available to swap the "chevron" icon for a "hamburger" on desktop.

On mobile, it is closed by default, can be opened by clicking a "hamburger" icon.

## The hamburger icon

I think that the hamburger icon needs to change to adapt to the sidebar layout.

On mobile, it should always be positioned in the topbar of the layout, as it is already.

On desktop, it should change depending on whether the sidebar is closable or collapsable.
- If the sidebar is closable, then the hamburger should be positioned in the topbar of the layout, as it is already.
- If the sidebar is collapsable, then the hamburger should be **inside the sidebar**, on the right of the logo. It should not be in the topbar. When the sidebar is collapsed, the sidebar logo should be removed completely and replaced with the hamburger. This reduces the space used on the topbar.

zepfietje:
> * a config for the sidebars collapse behaviour – fully (collapsible), partially (resizable)?

Not sure about the naming yet, could be `can_be_closed` and `is_collapsible` (or can_be_collapsed).

> * move the chevron icon away from a separate blade component into the sidebar `<header>`?

Yes, unless there's a need to have a separate Blade component for it?

> * keep the hamburger icon, in a separate blade component, or move it back to where it was?

Should be fine back where it was. Again, if you have a good reason to extract it into a component, that should be fine as well.

> * which icon is visible, is controlled by the sidebar collapse behaviour configuration?
>   * collapsible = hamburger
>   * resizable = chevron

Yes, so that's actually the main conclusion: closing (named collapsible in your example) and collapsing (named resizable in your example) are two distinct features which come with their own trigger icon. On mobile, the menu can already be closed, which is why there's a hamburger menu icon. On desktop it's collapsible already, which uses the chevron (just needs to be moved into the sidebar).

> * remove config for sidebar width?

Wouldn't that be a breaking change? Deprecating it by removing it from the default config and adding docs on customization via CSS should be good for now though.

> * remove config for sidebar collapsed_width?

Same for this, if it's easy with CSS, that's the preferred route over config keys.

> * let sidebar widths be altered via custom css, not `css root:  --var` + config?

What would be the pros and cons?

> * config for whether the initial collapse state should be open or closed (requested by @Edsardio )?

I'm not too sure about this yet, though I do see the use. Would be an easy separate PR though?

> I am not sure I agree with your strategy, but it is your repo ...

What in particular do you not agree with? We're always open to discussion to see what works best. 🙂

> Please help me understand why the current solution isn't enough, the outcome is the same and is easier personalized via config.

I think it's bad practice to tie the config to the specific icon being used. When reasoning about the features, you find that the chevron is used for collapsing and the hamburger menu icon for closing/opening the menu. So having just the config for enabling/disabling those two features sounds most obvious to me.

> If you want to change the collapse behaviour, you have to publish the config. When doing so, I think it is much easier to just change the widths, and icon in the same code position, then having to create custom styles in an external css file.

If custom styling using CSS would be too difficult, we could of course add a config key instead. However, generally we're moving away from config keys in favor of configuring custom styling using CSS. You could imagine how messy it would get if we'd have to add config keys for a lot of styling tweaks. Also, customizing with CSS gives you way more power and freedom to tweak the exact way you want.

> Also the combination if you want to keep the hamburger icon even though the sidebar is "resizable" is much easier to config with my solution than the one you suggest.

The hamburger menu icon shouldn't be used for the collapsing functionality. The menu icon is used for opening and closing the sidebar. The chevron icon is used for expanding and collapsing the sidebar, and is therefore shown inside the sidebar. If people really want to change the icons for any of those features, that will soon be possible: #3568.

zepfietje:
@danharrin, what if you'd configure the sidebar to be both collapsible and closable?
That's also why I outlined it as I did above.

tanthammar:
@zepfietje can you please see the discord group chat

tanthammar:
**Updating as discussed yesterday.**
- Sidebar is open as default on desktop.
- Hamburger icon replaces brand icon in collapsed sidebar.
- New hamburger/chevron icon before logo, in expanded sidebar.
    - Closes sidebar on all screen sizes (on desktop, if collapsible is enabled). 
    - Makes it easier to collapse the sidebar on xs screen.
- `collapsed_width` is added to config
- Chevron icon from previous PR is removed.
- Feature to fully collapse the sidebar on desktop is postponed to v3.
- Mobile screen remains unchanged, except that the new icon can close the sidebar, not just click outside.

tanthammar:
Don't understand the failing tests. Can't see that I changed anything that would cause it.

tanthammar:
Updated to support fully collapsed on desktop.

danharrin:
Collapsible sidebar now:

https://user-images.githubusercontent.com/41773797/190858369-20c028ed-1e28-463f-a0ba-93117398c89a.mov

With `collapsed_width` set to `0`:

https://user-images.githubusercontent.com/41773797/190858399-6b3d9819-8ca8-4ff1-bff5-80f7c68216e1.mov

packages/admin/resources/views/components/layouts/app/sidebar/index.blade.php
```diff
@@ -2,48 +2,74 @@
...
+                <button
+                    type="button"
+                    class="filament-sidebar-collapse-button shrink-0 hidden lg:flex items-center justify-center w-10 h-10 text-primary-500 rounded-full hover:bg-gray-500/5 focus:bg-primary-500/10 focus:outline-none"
+                    x-on:click.stop="$store.sidebar.isOpen ? $store.sidebar.close() : $store.sidebar.open()"
+                    x-transition:enter="lg:transition delay-100"
+                    x-transition:enter-start="opacity-0"
+                    x-transition:enter-end="opacity-100"
+                    data-turbo="false"
+                >
+                    {{-- https://github.com/codeat3/blade-google-material-design-icons/blob/main/resources/svg/menu-open-r.svg --}}
```
zepfietje, line 38: @danharrin we should really change this icon to the Heroicons chevron left before releasing this. Looks super inconsistent now.

danharrin: Okay will sort it

danharrin: Based on Heroicons v2

Screenshot 2022-09-17 at 19 25 29

Screenshot 2022-09-17 at 19 25 37

studiowizjo:
Is it still possible to use arrow to collapse sidebar?



I tried many combinations, eg this:

```php
'sidebar' => [
  'is_collapsible_on_desktop' => true,
  'hamburger_icon_on_desktop' => false,
  'width' => '20rem',
  'collapsed_width' => '5.4em',
],
```

but it is not working for me.

I also didnt found in package this config key "hamburger_icon_on_desktop".

danharrin:
No, the implementation for this feature is already complex, and we decided to aim for icon consistency across all sidebar configurations, rather than providing an option for the arrow icon.

The arrow icon was only introduced into Filament for a couple of weeks, and we had feedback that it was not as clear as the original hamburger icon for users. So we decided this was a better approach instead, and haven't had complaints yet.

studiowizjo:
=( I loved this subtle icon.

pxlrbt:
>  Would you consider PR for possibility of configuration?

@studiowizjo I think Dan made it clear, that we already have a lot of config options and it makes the config too complex.

> Or maybe any way to customize view for this?

Would there be an easy way with 2 view files or render hooks? Then this could maybe be an option.

studiowizjo:
I understand you don't want a view like that anymore, but I mean would you accept a PR that gives user the ability to define your own view (so not maintain two views by package, but on my own). There are now a lot of places where package has API for such things. I can't quickly say what the API could be. I would have to look at the matter, but would you consider merging method that accepts own view for this?

danharrin:
I would accept a PR that adds new views which you can more easily override, yes... atm you would need to override the entire `sidebar.index` which isn't very safe

incoming-th:
I was waiting for this icon as it looks easier for my end users, with the hamburgers they were waiting for a dropdown or modal to show not to expand the sidebar.

Anyway, tag me if this is back in the dev cycle one day.

jimiero:
@danharrin anything about this? I find it very interesting lot of people would love it I'm sure.

PR #3888, opened 2022-09-06
More generic translation for drag and drop repeater field & tooltips …

Edsardio:
More generic translation for drag and drop repeater field & tooltips for the icons in repeater field.


danharrin:
Thank you, but not sure if the tooltips might be a bit of an overkill for all repeater buttons?

Also, tooltips should have their own `tooltip` translation string alongside label.

Edsardio:
I do agree that for me as a developer the buttons are pretty self explanatory but when building a system for customers they might not always understand since they don't have to be tech-savy. What we could do is make them optional? Would be really annoying if you have to overwrite the template just for simple tooltips.

I updated the translations so they're their own thing.

zepfietje:
I think we should have as few tooltips as possible by default (it should all be accessible and clear of course). Wouldn't it be really annoying if a tooltip popped up immediately every time you hover one of the repeater actions?

danharrin:
Thats what I'm thinking. We could swap the tooltip for `title=""`

zepfietje:
> Thats what I'm thinking. We could swap the tooltip for `title=""`

Exactly what I was thinking. However, then there's no gain on touch devices, which makes me prefer not having it at all...

danharrin:
Theres no mobile gain for tooltips either... it doesnt hurt adding the title anyway

zepfietje:
Fair, but makes me wonder when we should add titles vs tooltips.

adrianlzx1996:
@danharrin there's a typo in this commit for builder.blade.php 

line 111 & line 168, missing a single quote " ' "

danharrin:
Fixed in the next release

PR #3889, opened 2022-09-06
Font size tweaks

zepfietje:
- [x] [Markdown editor](https://filamentphp.com/docs/2.x/forms/fields#markdown-editor)
- [x] [Tags input](https://filamentphp.com/docs/2.x/forms/fields#tags-input)
- [x] [Key-value](https://filamentphp.com/docs/2.x/forms/fields#key-value)
- [x] [Color picker](https://filamentphp.com/docs/2.x/forms/fields#color-picker)

---

https://discord.com/channels/883083792112300104/883083792653381695/1016710565286117527


zepfietje:
Will look into it! Just got this PR up as a start.

zepfietje:
@danharrin, I think this PR has some conflicts with #3974 and the new editable columns need font size fixing too?

danharrin:
I disagree, the editable column font size should match the normal cell font size

zepfietje:
And the normal cell font size should be `sm` as well, so then that's consistent.

danharrin:
I think we've tried that before and it looked bad

zepfietje:
`text-base` for the table cells actually looks worse, especially after reducing the input font size.

danharrin:
Oh, I didn't see you had already changed the cell font size

danharrin:
Markdown editor font size now looks bad

Screenshot 2022-09-12 at 14 47 02

Also the multiselect badges and placeholder

Screenshot 2022-09-12 at 14 47 31

zepfietje:
Yeah Markdown editor was tricky. What looks bad in your opinion?

zepfietje:
Hmm I thought I fixed the Choices.js placeholder, but that's different from multiselect?

danharrin:
> Yeah Markdown editor was tricky. What looks bad in your opinion?

The font just looks too big

zepfietje:
It's the same font size as the other inputs, so that's an issue with the used font if you think it looks bad.

danharrin:
What does xs look like?

zepfietje:
I think we shouldn't do that. The monospace font seems like the system font instead of one that pairs well with the DM family. We could use DM Mono instead for example, which has a similar x-height to DM Sans. Tweaking the font size will result in bad UI/UX for people who use a proper pair of fonts that match well.

All in all, I think this is a fonts issue rather than a problem with the font size of the Markdown editor.

danharrin:
I think a switch to DM Mono wouldn't be bad at all

realslimsutton:
Did this PR get reverted at some point (or am I just being blind and need to go to sleep already)? I can't seem to find these changes on the 2.x branch.

PR #3890, opened 2022-09-06
Feat: Validation Rule Methods

Z3d0X:
Adds validation rule methods.
Let me know which other rules that should be added.


zepfietje:
I'm not sure if we should have a method for every single available validation rule, since you can just pass it to the `rule()` or `rules()` method already.

danharrin:
I think we should cover every single Laravel rule 🤷‍♂️ It doesnt add overhead and makes the syntax consistent

zepfietje:
Okay, let's go all the way then.

Z3d0X:
What rules like `exclude` (we already have that functionality from `->dehydrate(false)`)?

Z3d0X:
Sorry I actually completely forgot about this, let me try to get this PR ready for review

danharrin:
A list of rules we dont need:

- Any rule variants that use "if" or "unless" - you can do this with closures
- Bail
- Distinct
- Exclude
- Prohibited
- Required *
- String

zepfietje:
Why don't we need the `string` rule, @danharrin? I wondered this recently as I usually add that rule to my validation but couldn't find any reference in Filament.

danharrin:
I don't see why anyone would really need it?

zepfietje:
Couldn't potentially an array be passed as the value for an input that expects a string? Maybe that's guarded by a type hint, but I'm just evaluating the extremes here.

danharrin:
Apart from those, all done?

danharrin:
Thanks so much for your work on this!

PR #3923, opened 2022-09-09
Active filters text dark mode color adjustment

belzaaron:
Ensures readability while on dark mode.

Before
before
After
after


danharrin:
Thanks

PR #3930, opened 2022-09-10
Adding menu item override code + sample config.

robertgdev:
This patch allows you to override/change menu item properties through the filament.php config file. 

The rationale for this change is that currently, it is difficult to change these properties because: 
1) Some plugins define these in a translation file
2) Some plugins simply hardcode these in code
Either way, you have to change the plugin's code or translation file to change these options. 

This way, you can override this behavior from the central filament config file without having to touch plugin code. For me, this is a useful addition, I hope you find it useful as well.

### DIFF
--- a/packages/admin/config/filament.php
+++ b/packages/admin/config/filament.php
@@ -201,6 +201,21 @@ return [
             'should_show_logo' => true,
         ],
         'max_content_width' => null,
+        /*
+         * Through these settings, you can override the following (empty string values have no effect)
+         * 1) The menu group the item is placed under
+         * 2) The icon used for the menu item
+         * 3) Hide the menu item by setting "display" to false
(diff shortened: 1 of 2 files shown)


danharrin:
Hey, thanks for the PR, but I'm not really a fan of using the config for this. We already have a similar API that you can use to override this sort of thing:

https://filamentphp.com/docs/2.x/admin/navigation#advanced-navigation-customization

rgasch:
I appreciate and respect your position but changing this through the API (which I did see before coding this), requires you to alter actual code ... for what it's worth, I prefer configurable code over requiring actual code changes to code that resides under /vendor.

danharrin:
You don't need to change any code within vendor, the Advanced API can be put in your AppServiceProvider, so there aren't risks involved there :)

rgasch:
OK, thank you, I will check/try this out.

PR #3931, opened 2022-09-10
fix: Ternary filter indicator

### LINKED ISSUE #3922: TernaryFilter indicators showing as "Yes" or "No" by default and no `->indicator()` changes the displayed text
### Package

filament/tables

### Package Version

v2.15.45

### Laravel Version

v9.28.0

### Livewire Version

v2.10.7

### PHP Version

v8.1.8

### Problem description

Creating a `TernaryFilter` for a table as such:
```php
TernaryFilter::make('email')->label('Has Email')->nullable()
```
or
```php
TernaryFilter::make('email')->label('Has Email')->indicator('Has Email')->nullable()
```
Will result in the "Active filters" indicator to be "Yes" or "No" respectively:
Example of Active filters text

As you could imagine some TernaryFilters can really be as simple as "Yes" or "No" for their values (which is default), and regarding their "Active filter" labels being the same this can be quite confusing:
confusing Active filter indicators

### Expected behavior

As per the documentation I would expect "Has Email: Yes" or "Has Email: No", considering the [documentation](https://filamentphp.com/docs/2.x/tables/filters#active-indicators) and the `->label()`/`->indicator()` I've set.

### Steps to reproduce

Create a simple TernaryFilter without using something to the effect of `->trueLabel('Has Email')->falseLabel('Has No Email')`.

danharrin:
Fixes #3922

### DIFF
--- a/packages/tables/src/Filters/TernaryFilter.php
+++ b/packages/tables/src/Filters/TernaryFilter.php
@@ -21,18 +21,6 @@ class TernaryFilter extends SelectFilter
         $this->placeholder('-');
 
         $this->boolean();
-
-        $this->indicateUsing(function (array $state): array {
-            if ($state['value'] ?? null) {
-                return [$this->getTrueLabel()];
-            }
-
-            if (blank($state['value'] ?? null)) {
-                return [];
-            }
-
-            return [$this->getFalseLabel()];
-        });
     }
 
     public function trueLabel(string | Closure | null $trueLabel): static
--- a/packages/tables/src/Filters/TrashedFilter.php
+++ b/packages/tables/src/Filters/TrashedFilter.php
@@ -29,6 +29,18 @@ class TrashedFilter extends TernaryFilter
             false: fn ($query) => $query->onlyTrashed(),
             blank: fn ($query) => $query->withoutTrashed(),
         );
+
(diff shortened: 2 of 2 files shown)


titantwentyone:
Hey @danharrin. Just commenting here as was using a TernaryFilter today and no indicators are being rendered at all. The filter itself works fine but no feedback displayed. Seems to be fine in 2.15.49 before this commit however?

danharrin:
Hey, it should now inherit the behaviour of the SelectFilter, so it should indicate when Yes or No is selected

titantwentyone:
Because of the state of the filter, which could very well be a boolean, it never reaches `return ["{$this->getIndicator()}: {$label}"];`. I think what happens is (thoughts in comments)...

```
        $this->indicateUsing(function (array $state): array {
            if (blank($state['value'] ?? null)) { // if false, will return empty array
                return [];
            }

            $label = $this->getOptions()[$state['value']] ?? null; //no options as its a TernaryFilter so no label

            if (blank($label)) { //...so returns an empty array
                return [];
            }

            return ["{$this->getIndicator()}: {$label}"]; //never reached
        });
```

danharrin:
@titantwentyone #4040 should fix this

PR #3943, opened 2022-09-10
fix: notification tooltip indicator color

shuvroroy:


danharrin:
This breaks the indicator for non-primary colors, I think it looks fine how it is?

PR #3944, opened 2022-09-10
Adding ability to exclude self from relationship

titantwentyone:
This is a tentative PR to provide the ability to exclude the current record for selection in a self-referential relationship. For example, if you have a `Category` model which has `parent` and `children` relationships which are mapped to other `Category` objects, you will most likely not want to to set the same record as a parent or a child. This PR includes `excludeSelf()` which removes the record from the options in a field where `relationship` is used.

There is probably more to do to this but wanted to just sound it out and see if it was worth continuing with.

Unsure where to put a test for this. If this looks okay, let me know where the best place to put that and I'll add.


danharrin:
Hi, gonna hold off on this one. I think that you can achieve this already using the `relationship(query: fn ($query, $record) => ...)` parameter?

PR #3952, opened 2022-09-10
fix: Database notifications docs

zepfietje:


packages/notifications/docs/03-database-notifications.md
```diff
@@ -142,7 +142,7 @@ Instead of rendering the trigger button as described above, you can always open
 
 ```blade
 <button
-    x-data
+    x-data="{}"
```
zepfietje, line 145: `x-data` should be fine actually for non-Blade components.

danharrin: I think it's good to be consistent everywhere without needing to explain where you can and cannot omit the value

PR #3974, opened 2022-09-12
feature: Editable table columns

danharrin:
This introduces 4 new table columns:
- SelectColumn
- TextInputColumn
- ToggleColumn
- CheckboxColumn

These column names correspond to the names of the form components they mimic.

Data can be written to the main Eloquent table, a relationship, or a pivot column.

These do not directly utilise the form builder, but do consume traits from that package for certain features

Validation can be added to the columns using `rules(['required', 'max:255'])`. Validation rules like `required()` and `maxLength()` that are used in the form builder are not necessary in this context, as there is no frontend validation.

Screenshot 2022-09-12 at 12 52 09

Screenshot 2022-09-12 at 12 51 33


zepfietje:
Wouldn't a checkbox be useful as well, in addition to the toggle?

PR #3982, opened 2022-09-12
feature: Allow wizard steps to be clickable

wdarins:


danharrin:
Thanks

ziming:
My skippable() is set to false but i can still click the wizard step headers to jump steps. Is there another method I should call to disable that?

```php
Wizard::make([...])
->skippable(false)
```

danharrin:
@ziming just fixed this in the latest version

PR #3993, opened 2022-09-13
Add documentation about echo to the notification docs.

bramr94:
While configuring the database notifications with Laravel Echo I couldn't find the part about uncommenting the code in the filament config.


zepfietje:
Notifications can be used outside the Filament admin panel as well. The documentation you're looking for is under the admin package: https://filamentphp.com/docs/2.x/admin/notifications#echo.

PR #3994, opened 2022-09-13
Consider relationship callback in select option label/s

ajnsn:


danharrin:
Thanks!

mikrosmile:
Hello. Just updated to this ver and select field with relationship is broken:

Before -> selected value displays correctly on view and edit pages just as should be 

After -> selected value is actually the first value from table or if (in my case I'm using sorting by name) it shows me the first option from table and not the one I chosen 

To confirm that, I reverted all changes by this PR in latest build

PR #4021, opened 2022-09-14
Adding alias for table filter method `column`

titantwentyone:
New method `field` added as alias to `column` in `SelectFilter` and `MultiSelectFilter` to better represent how the method can be used as per #3998. Docs also amended.

### DIFF
--- a/packages/tables/docs/04-filters.md
+++ b/packages/tables/docs/04-filters.md
@@ -97,7 +97,7 @@ SelectFilter::make('status')
     ])
 ```
 
-Select filters do not require a custom `query()` method. The column name used to scope the query is the name of the filter. To customize this, you may use the `field()` method:
+Select filters do not require a custom `query()` method. The column name used to scope the query is the name of the filter. To customize this, you may use the `attribute()` method:
 
 ```php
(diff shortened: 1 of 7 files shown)


zepfietje:
If we're going this direction, shouldn't `column()` call `field()` instead of the other way around? Also, we might want to mark `column()` as deprecated in that case.

Honestly, I'm not too sure about this change actually.

zepfietje:
Yeah I just think, if we're going for it, we should go all the way. As long as it's not a breaking change of course (that's why we could just mark the old method as deprecated).

danharrin:
Also we need a deprecated docblock with a replacement message on the old method

zepfietje:
> Also we need a deprecated docblock with a replacement message on the old method

Yep, that's what I meant.

titantwentyone:
Was also wondering if we should pull out `column()`, `getColumn()`, `attribute()`, `getAttribute()` into a trait?

danharrin:
> Was also wondering if we should pull out `column()`, `getColumn()`, `attribute()`, `getAttribute()` into a trait?

Probably overkill atm, but thanks

PR #4022, opened 2022-09-14
Adding assertions to test column data

titantwentyone:
```
assertTableColumnDataSet($record, $name, $value)
assertTableColumnDataNotSet($record, $name, $value)
```
Went for passing in the record rather than the key as it was simpler but can look at reworking if it is preferred. Am thinking the context of any test would have the object to hand anyway.

### DIFF
--- a/packages/tables/.stubs.php
+++ b/packages/tables/.stubs.php
@@ -95,6 +95,10 @@ namespace Livewire\Testing {
 
         public function assertTableColumnHidden(string $name): static {}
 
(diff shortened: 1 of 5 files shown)


titantwentyone:
If the `id` is preferential, we can add in something like...

```
return function ($record_id, $name, $value): static {

        ...
        $record = $column->getLivewire()->getTableModel()::find($record_id);
        Assert::assertNotEmpty($record);

        $column->record($record);
        ...
}
```

danharrin:
I handled both record objects and keys

PR #4036, opened 2022-09-15
Adding reset and deselect records on individual search

Sicklou:
Hi,

I propose this PR in order to add the same behavior that the global search when doing a search action.

I was on the sixth page of a table, then I made an indivudiual search and no records were appearing, because the new number of page was below one.


danharrin:
Thanks

PR #4054, opened 2022-09-16
🐛 Use appropriate can* method when check for resource pages

juliomotol:
This aims to use the appropriate `can*()` method when it falls back to checking the resources pages on `ListRecord::getTableRecordUrlUsing()` that was implemented on [this commit](https://github.com/filamentphp/filament/commit/d9cc6d23f9d7328e9c704f66a940478d1dd340d0)

Given the following table setup:
```php
{
    return $table
        ->columns([
            ...
        ])
        ->actions([
            Tables\Actions\EditAction::make()
                ->visible(fn (Admin $record): bool => Auth::user()->can('update', $record))
        ]);
}
```
Since I have a `visible()` in my `EditAction`, when the closure returns `false` it treats as if there were no actions for the table.

Now on the resource pages fallback, it checks for a `view` and `edit` policy. But `edit` is not a standard ability for a policy, but instead we have `update`. Thus it still allows the clicking of the row to open the edit page for the resource. And upon opening it just shows a 403 error.

To remedy this, we will be calling the appropriate `canEdit()` method on the `Resource`.

You may also want see my [other PR](https://github.com/filamentphp/filament/pull/4053) that solves the same issue

### DIFF
--- a/packages/admin/src/Resources/Pages/ListRecords.php
+++ b/packages/admin/src/Resources/Pages/ListRecords.php
@@ -421,7 +421,7 @@ class ListRecords extends Page implements Tables\Contracts\HasTable
                     continue;
                 }
 
-                if (! $resource::{'can'.ucfirst($action)}($record)) {
+                if (! $resource::{'can' . ucfirst($action)}($record)) {
                     continue;
                 }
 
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks

PR #4061, opened 2022-09-17
fix: Check translations command pluralization

zepfietje:


zepfietje:
Forgot about that string helper. Thanks for cleaning up, @danharrin.

PR #4086, opened 2022-09-18
feature: Table column layouts ("List style")

### LINKED ISSUE #4070: [2.x] List view for tables
https://twitter.com/danjharrin/status/1562839703422144513

danharrin:
Screenshot 2022-09-18 at 20 04 45

Screenshot 2022-09-18 at 20 05 07

Closes #4070.


studiowizjo:
Isn't it too similiar to the current layout? You can achieve layout close to this with `formatStateUsing`, but I suppose change will bring more clean api.

What do you think about layout more [similiar to cards](https://tailwindui.com/components/application-ui/lists/grid-lists)? Or maybe there is a plan for this, but it is different topic?

danharrin:
> Isn't it too similiar to the current layout? You can achieve layout close to this with `formatStateUsing`, but I suppose change will bring more clean api.
> 
> What do you think about layout more [similiar to cards](https://tailwindui.com/components/application-ui/lists/grid-lists)? Or maybe there is a plan for this, but it is different topic?

Nah, it can't be done using the existing columns. This is much more complex.

Columns can now be nested within layouts, which preserves sorting and searching. Using `formatStateUsing()` and custom columns doesnt allow that. Also allowing columns to stack on top of each other responsively is not easy with the existing builder.

I have no plans to make grid lists. At that point, it's not actually a table builder, as theres no table columns. You can achieve something similar with our existing [custom content feature](https://filamentphp.com/tricks/use-custom-view-for-table-content).

danharrin:
Well @studiowizjo... I take back what I said. I made grid cards 😂

Issue #3481, opened 2022-08-11
Markdown editor not show characters correctly on RTL


themahdavi:
### Package

filament/forms

### Package Version

2.15.18

### Laravel Version

9.24

### Livewire Version

2.10.7

### PHP Version

8.1.7

### Problem description

This problem accord when using markdown editor into Grid and RTL base locale
shown like this :
![image](https://user-images.githubusercontent.com/31320104/184193047-45b1d81b-dfc5-4186-9c2f-a0cf6a365996.png)

after testing some changes find problem but don't know how to solve that
 

### Expected behavior

If use MarkdownEditor in Grid with columns on specific screen break and change locale to RTL base locale like `fa` or `ar` on `config/app.php` issue shown

### Steps to reproduce

1- clone repository
2- under `database` folder have `database.sqlite` file just have users table with one record
3-change `.env.example` to `.env` or own .env file with sqlite database driver
4- nothing to do just run project and go to Edit page of user and see on bio field

### Reproduction repository

https://github.com/themahdavi/markdown-problem

danharrin:
Chrome-only issue.

danharrin:
Fixed by #3526. Let's see what other issues that change introduces, I can't find other effects of this change.

PR #4092, opened 2022-09-19
fixed markdown-editor list ident

mathijsm0:
Fixed an overlay issue with indents on ordered lists.

before: 
Schermafbeelding 2022-09-19 om 10 50 10

after:
Schermafbeelding 2022-09-19 om 10 49 16


danharrin:
Hi, I believe that this would reintroduce #3481 for RTL on Chrome?

falcomnl:
This would quite possibly fix #3481. I have no means to confirm this on RTL though.

danharrin:
No, the fix for #3481 was to remove that class - https://github.com/filamentphp/filament/pull/3526/files

You can test RTL by switching your locale to `ar` in `config/app.php`. The bug only exists in Chrome as far as I am aware.

falcomnl:
Ok, fixed it to keep support for RTL in #3481

danharrin:
Thanks :)

PR #4096, opened 2022-09-19
Add tables testing method assertCannotRenderTableColumn

miguel-is-coding:
- This feature will allow user to test if a table is not showing a determined column by default. 
- Added function at `TestColumns`, `.stubs` and documented at `06-testing.md`

This feature is useful in situations where you have a table with columns with 
```php
TextColumn::make('comments')
->toggleable(isToggledHiddenByDefault: true)
```
and you want to test that column is not showing by default in the table view.


danharrin:
Thanks!

PR #4105, opened 2022-09-19
Records selection on current page only

frankyso:
Bulk actions are very helpful, but the selection function in bulk actions takes all the records in the database, maybe with small records it is still fine, but it will reduce performance as more records in the table are increased.

image

see the `select  count(*) as aggregate from 'blog_posts'`

so I think it would be better to add options where we can only select the records that appear on the current page instead of selecting all existing records.

image
image
image

thank you


danharrin:
Thanks!

PR #4108, opened 2022-09-19
Adding getContext() shortcut for form builde

cheesegrits:
As discussed with @danharrin added getContext() method, and docs for $context in the closure customization section.  Lmk if you want the wording of the doc section modified.

### DIFF
--- a/packages/forms/src/Concerns/InteractsWithForms.php
+++ b/packages/forms/src/Concerns/InteractsWithForms.php
@@ -278,10 +278,10 @@ trait InteractsWithForms
     {
         return [
             'form' => $this->makeForm()
-                ->context($this->getFormContext())
                 ->schema($this->getFormSchema())
                 ->model($this->getFormModel())
-                ->statePath($this->getFormStatePath()),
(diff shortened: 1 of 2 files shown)


packages/forms/src/Concerns/InteractsWithForms.php
```diff
@@ -278,12 +278,18 @@ protected function getForms(): array
...
         return [
             'form' => $this->makeForm()
+                ->context($this->getContext())
                 ->schema($this->getFormSchema())
                 ->model($this->getFormModel())
                 ->statePath($this->getFormStatePath()),
         ];
     }
 
+    protected function getContext(): ?string
```
danharrin, line 288:
```suggestion
    protected function getFormContext(): ?string
```

danharrin (changes requested):

packages/forms/src/Concerns/InteractsWithForms.php
cheesegrits: Done

PR #4110, opened 2022-09-20
Support BelongsToThrough Relationship

wit3:
Support Table Filter relationship custom with BelongsToThrough relationship
[BelongsToThrough Package](https://github.com/staudenmeir/belongs-to-through)

### DIFF
--- a/packages/tables/src/Filters/Concerns/HasRelationship.php
+++ b/packages/tables/src/Filters/Concerns/HasRelationship.php
@@ -32,7 +32,7 @@ trait HasRelationship
         } elseif ($relationship instanceof HasOneThrough) {
             $keyColumn = $relationship->getQualifiedForeignKeyName();
         } elseif ($relationship instanceof \Znck\Eloquent\Relations\BelongsToThrough) {
-            $keyColumn = $relationship->getRelated()->getKeyName();
+            $keyColumn = $relationship->getRelated()->getQualifiedKeyName();
         } else {
             /** @var BelongsTo $relationship */
             $keyColumn = $relationship->getQualifiedOwnerKeyName();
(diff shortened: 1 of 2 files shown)


danharrin:
Please fix PHPStan by requiring the package in the base composer.json as a dev dependency

Wit3:
Thanks Dan, 've compiled the required package in dev dep.
All test are ok

packages/tables/src/Filters/Concerns/HasRelationship.php
```diff
@@ -31,6 +31,8 @@ public function getRelationshipKey(): string
             $keyColumn = $relationship->getQualifiedRelatedKeyName();
         } elseif ($relationship instanceof HasOneThrough) {
             $keyColumn = $relationship->getQualifiedForeignKeyName();
+        } elseif ($relationship instanceof \Znck\Eloquent\Relations\BelongsToThrough) {
+            $keyColumn = $relationship->getRelated()->getKeyName();
```
danharrin, line 35:
```suggestion
            $keyColumn = $relationship->getRelated()->getQualifiedKeyName();
```

Please check the integration still works after making this change

danharrin:
Thanks

mahfoudfx:
Thanks @Wit3, lot of people underestimate this relationship, which my opinion should be a part of Laravel Core.
Having this on Filament, it is just a great enhancement.

Wit3:
I've found a bug in

https://github.com/filamentphp/filament/blob/6ccf78a2f277b4f6a1dcdf8be12908b77cd4b35b/packages/forms/src/Components/Select.php#L477

It's necessary to check if relationship is an instance of `BelongsToThrough` abd get the correct `keyName`

Like this
```php
 if ($relationship instanceof \Znck\Eloquent\Relations\BelongsToThrough) {
      $keyName = $relationship->getRelated()->getQualifiedKeyName();
 } else {
      $keyName = $component->isMultiple() ? $relationship->getRelatedKeyName() : $relationship->getOwnerKeyName();
 }
```

Maybe it's necessary to revert this PR and find all key that need to change when relationship is loaded
From Form Docs
```md
Field relationships
Some fields, such as the Select, Repeater, SpatieMediaLibraryFileUpload, or SpatieTagsInput are able to interact with model relationships.
```
What do you think? @danharrin

danharrin:
Sorry, but I am not really interested in integration with this library, I don't use it. Any work will need to be done by yourself in a new PR, please.

PR #4115, opened 2022-09-20
Feature: Page SelectOption default selection

yob-yob:
This PR will allow setting the default selected option for the `SelectAction` under Page Actions.

I had a problem where when I refresh the page, the default option I wanted it to show is not selected, making it confusing to my users on what is the selected default option, because my table is showing a different data.

### DIFF
--- a/packages/admin/resources/views/pages/actions/select-action.blade.php
+++ b/packages/admin/resources/views/pages/actions/select-action.blade.php
@@ -16,9 +16,15 @@
         @endif
 
         @foreach ($getOptions() as $value => $label)
-            <option value="{{ $value }}">
-                {{ $label }}
-            </option>
+            @if(($selected = $getSelected()) !== null)
+              @if($selected == $value)
+                <option value="{{ $value }}" selected>{{ $label }}</option>
+              @else
+                <option value="{{ $value }}">{{ $label }}</option>
+              @endif
+            @else
+              <option value="{{ $value }}">{{ $label }}</option>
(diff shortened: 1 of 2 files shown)


packages/admin/resources/views/pages/actions/select-action.blade.php
```diff
@@ -16,9 +16,15 @@
         @endif
 
         @foreach ($getOptions() as $value => $label)
-            <option value="{{ $value }}">
-                {{ $label }}
-            </option>
+            @if(($selected = $getSelected()) !== null)
+              @if($selected == $value)
```
Saifallak, line 20: Maybe use @selected blade directive ?

danharrin: Is that supported in Laravel 8?

Saifallak: Ops, nope, it was added 22 feb 2022, so L9.
Maybe for filament v3 then

danharrin:
Hey, you should actually accomplish this by setting the Livewire public property that this action is bound to inside `mount()`. This action is actually for internal use, which is why it's not been documented, but that's how we use it 👍 Setting a `selected` attribute like this could conflict with the `wire:model`

yob-yob:
Oh My God... I'm honestly sorry, I did not noticed the `wire:model`... Big Mistake on part...

Sorry...

PR #4121, opened 2022-09-21
Feature: make brand text internationalized

aymanalareqi:
This PR is to make brand text internationalized


danharrin:
Hey, can't you just use `__()` in the config file instead?

aymanalareqi:
> Hey, can't you just use `__()` in the config file instead?

no, and it give me error
Target class [translator] does not exist.

aymanalareqi:
> > Hey, can't you just use `__()` in the config file instead?
> 
> no, and it give me error Target class [translator] does not exist.

you can check this 
https://stackoverflow.com/questions/35671717/laravel-string-localization-in-config-files/35672394#35672394

danharrin:
I know, but this still might be a breaking change for some people, I'm not sure we should force people to translate the brand name.

In a service provider `boot()`, try this:

```php
config()->set('filament.brand', __(config('filament.brand')));
```

danharrin:
But if someone has a translation string with their brand name key as an array, the entire admin panel will break. Does the service provider method not work?

aymanalareqi:
> But if someone has a translation string with their brand name key as an array, the entire admin panel will break. Does the service provider method not work?

oh, you are right
what do you mean _Does the service provider method not work?_

aymanalareqi:
> > But if someone has a translation string with their brand name key as an array, the entire admin panel will break. Does the service provider method not work?
> 
> oh, you are right what do you mean _Does the service provider method not work?_ do you mean to override filament.brand config from service provider?

it works with this
`config([
            "filament.brand"=>__("common.app_name"),
           ]);`
in service provider boot method

aymanalareqi:
> Great!

sorry it didnt work as i expected
it only make translation from default locale

danharrin:
Are you setting locale in middleware or something?

danharrin:
How about registering an event listener for `Illuminate\Foundation\Events\LocaleUpdated`

and then setting the config there?

aymanalareqi:
> How about registering an event listener for `Illuminate\Foundation\Events\LocaleUpdated`
> 
> and then setting the config there?

it work perfectly
thanks @danharrin

PR #4138, opened 2022-09-22
Fix Select/Multiselect relationship BelongsToThrough

wit3:
In the next days i'll investigate on entire package ( all component and table )
Now this PR is to fix Select/Multiselect component from Form using BelongsToThrough relationship.


danharrin:
Thanks

PR #4140, opened 2022-09-22
fix Type Error on return type

wit3:
Filament\Forms\Components\Select::getRelationship(): Return value must be of type Illuminate\Database\Eloquent\Relations\BelongsTo|Illuminate\Database\Eloquent\Relations\BelongsToMany|null, Znck\Eloquent\Relations\BelongsToThrough returned


danharrin:
Thanks!

PR #4166, opened 2022-09-23
Update Dutch translations

zepfietje:


dejury:
Why is `:count` not used?
And shouldn't it be "Er kunnen maar :count item(s) geselecteerd worden."

zepfietje:
Whoops, thanks for noticing, @dejury! Replaced the count when finding the best translation and forgot to put it back into place.

> And shouldn't it be `Er kunnen maar :count item(s) geselecteerd worden.`

The English translation reads "Only :count can be selected.", so I left out "items". Should be fine, right?

PR #4170, opened 2022-09-23
Fixes translation as reported in the comment section of PR #4166

dejury:
Fixes translation issue introduced in PR #4166

### DIFF
--- a/packages/forms/resources/lang/nl/components.php
+++ b/packages/forms/resources/lang/nl/components.php
@@ -211,7 +211,7 @@ return [
 
         'loading_message' => 'Laden...',
 
-        'max_items_message' => 'Er kunnen maar 3 geselecteerd worden.',
+        'max_items_message' => 'Er kunnen maar :count item(s) geselecteerd worden.',
 
         'no_search_results_message' => 'Er zijn geen resultaten voor je zoekopdracht.',
 


packages/forms/resources/lang/nl/components.php
```diff
@@ -211,7 +211,7 @@
 
         'loading_message' => 'Laden...',
 
-        'max_items_message' => 'Er kunnen maar 3 geselecteerd worden.',
+        'max_items_message' => 'Er kunnen maar :count item(s) geselecteerd worden.',
```
danharrin, line 214: https://laravel.com/docs/9.x/localization#pluralization

```suggestion
        'max_items_message' => 'Er kunnen maar :count item geselecteerd worden.|Er kunnen maar :count items geselecteerd worden.',
```

danharrin (changes requested):

packages/forms/resources/lang/nl/components.php
dejury: I don't think this is possible right now, since the maxItemsMessage is defined in the setup(). At this point we are still unsure how many items there might be. Or am I wrong?

Also we need to rewrite this part to use trans_choice.

danharrin: We actually have written a `trans_choice()` implementation in JavaScript, that's when this string gets evaluated... the number of max items never touches PHP

zepfietje: Since the English translation didn't have "items", I left it out in the Dutch translation as well.

zepfietje:
Since I couldn't push to your branch, @dejury, I've created a new PR. If you'd create a separate branch for the fix next time instead of the main (2.x) branch, maintainers can add to the PR.

Feel free to continue the discussion on adding "items" to the translation string here. At least the bug of the missing `:count` is fixed by https://github.com/filamentphp/filament/pull/4179.

dejury:
Great, dankjewel!

PR #4183, opened 2022-09-24
Merge extra attributes in layout components

ajnsn:
This will allow to define extra attributes on the new layout components, e.g. increase the spacing within the stack:

```php
Tables\Columns\Layout\Split::make([
    Tables\Columns\Layout\Stack::make([
        // ...
    ])
    ->alignCenter()
    ->extraAttributes([
        'class' => 'space-y-2',
    ]),
]),
```

Edit: Just saw the `getSpace()` afterwards - lol. But even though, this might be useful :)

### DIFF
--- a/packages/tables/resources/views/columns/layout/panel.blade.php
+++ b/packages/tables/resources/views/columns/layout/panel.blade.php
@@ -1,7 +1,9 @@
-<div @class([
-    'px-4 py-3 bg-gray-100 rounded-lg',
-    'dark:bg-gray-900' => config('forms.dark_mode'),
-])>
+<div
+    {{ $attributes->merge($getExtraAttributes())->class([
+        'px-4 py-3 bg-gray-100 rounded-lg',
+        'dark:bg-gray-900' => config('forms.dark_mode'),
+    ]) }}
+>
     <x-tables::columns.layout
         :components="$getComponents()"
         :record="$getRecord()"
--- a/packages/tables/resources/views/columns/layout/panel.blade.php
+++ b/packages/tables/resources/views/columns/layout/panel.blade.php
@@ -1,8 +1,10 @@
 <div
-    {{ $attributes->merge($getExtraAttributes())->class([
-        'px-4 py-3 bg-gray-100 rounded-lg',
-        'dark:bg-gray-900' => config('forms.dark_mode'),
-    ]) }}
+    {{ $attributes
+        ->merge($getExtraAttributes())
+        ->class([
+            'px-4 py-3 bg-gray-100 rounded-lg',
+            'dark:bg-gray-900' => config('forms.dark_mode'),
+        ]) }}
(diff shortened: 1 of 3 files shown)


packages/tables/resources/views/columns/layout/panel.blade.php
```diff
@@ -1,7 +1,9 @@
-<div @class([
-    'px-4 py-3 bg-gray-100 rounded-lg',
-    'dark:bg-gray-900' => config('forms.dark_mode'),
-])>
+<div
+    {{ $attributes->merge($getExtraAttributes())->class([
```
zepfietje, line 2: Would it be more readable if all chained methods were on their own line?

zepfietje (changes requested):

packages/tables/resources/views/columns/layout/panel.blade.php
ajnsn: Agree. I adopted the styling of the other "column" components, e.g. the one of the [TextInput](https://github.com/filamentphp/filament/blob/2.x/packages/tables/resources/views/columns/text-column.blade.php#L10-L39) - there's always a line feed after the div :)

zepfietje: Hmm, in other places I've formatted it like I suggested. So should be fine as well if we agree that's more readable.

ajnsn: Updated the PR, hopefully I catched it correctly

packages/tables/resources/views/columns/layout/panel.blade.php
```diff
@@ -1,7 +1,11 @@
...
-    'dark:bg-gray-900' => config('forms.dark_mode'),
-])>
+<div
+    {{ $attributes
+        ->merge($getExtraAttributes())
+        ->class([
+            'px-4 py-3 bg-gray-100 rounded-lg',
+            'dark:bg-gray-900' => config('forms.dark_mode'),
+        ]) }}
+>
```
zepfietje, line 8: Same thing for the other files.

```suggestion
<div
    {{
        $attributes
            ->merge($getExtraAttributes())
            ->class([
                'px-4 py-3 bg-gray-100 rounded-lg',
                'dark:bg-gray-900' => config('forms.dark_mode'),
            ])
    }}
>
```

zepfietje (changes requested):

packages/tables/resources/views/columns/layout/panel.blade.php
zepfietje: Not entirely, I've commented with a change request. 😄

danharrin:
By the way, `Stack` does have a `space(2)` method to accomplish the same thing. I've just added it to the docs in this PR.

PR #4193, opened 2022-09-25
Fix/IconButton method missing in GlobalSearch Action

Nuhel:
Added iconButton method GlobalSearch/Actions/Action , as the view file exist.


danharrin:
Thanks

PR #4195, opened 2022-09-25
fix: Duplicate select options

### LINKED ISSUE #2363: BelongsToManyMultiSelect displays already attached options in dropdown on initial load
### Package

filament/filament

### Package Version

v2.12.2

### Laravel Version

v9.11.0

### Livewire Version

_No response_

### PHP Version

PHP 8.1.4

### Bug description

When using the `BelongsToManyMultiSelect` component, already attached records are still displayed in the dropdown when the form is loaded initially.

With no records attached:
billede

Attaching some records (selected options is removed from dropdown):
billede

After having saved the new selection and then editing the list again:
billede

I would expect the dropdown to exclude the already attached records to shown in the dropdown.

This is how I configure the field:

```php
// File: RolesRelationsManager.php
{
    return $form
        ->schema([
            TextInput::make('label')->required(),
            TextInput::make('name')->required(),
            BelongsToManyMultiSelect::make('role_permissions')
                ->placeholder(__('Select permission(s) to be assigned to role'))
                ->preload()
                ->relationship(
                    'permissions',
                    'label',
                    fn (Builder $query, $livewire) => $query->whereBelongsTo($livewire->ownerRecord)
                )->columnSpan(2),
        ]);
}
```

This might be related to #1966.

### Steps to reproduce

Use a BelongsToManyMultiSelect to edit a belongsToMany relationship.

### Relevant log output

_No response_

danharrin:
Fixes #2363.

If this reintroduces other issues then I will revert it.


danharrin:
Reverted - https://github.com/filamentphp/filament/issues/2363#issuecomment-1259566113

PR #4199, opened 2022-09-25
fixing incorrect closing div location

dev3k:
it caused an issue for me in the dashboard widgets, if one of the grids is empty

Screen Shot 2022-09-26 at 12 01 46 AM


danharrin:
Thanks

PR #4210, opened 2022-09-26
fix: evaluate Closures passed to getIcon()

kamilst96:
This method wasn't wrapped in `$this->evaluate()`,
so a Closure passed to TextColumn wasn't correctly handled.

This fails before this PR:
```php
TextColumn->icon(fn() => 'heroicon-o-check')
```


danharrin:
thanks

PR #4215, opened 2022-09-26
feature: Hint actions

tanthammar:
I need a `slot` that is available to all fields, where I can place an `Action`.

My use case is to open a modal with a longer help text.

The only place I found, without a larger rewrite, was the `hint` position. Which also makes sense design wise...

I kept the `hint()` method and modified it to also accept an Action.
The action is auto-registered on the component.

Example (sorry for the Swedish)
image

```php
TextInput::make('iban')->hint(Action::make('ibanHelp')//...->icon('heroicon-o-question-mark-circle'))
Toggle::make('bankgiro')->hint(Action::make('bankgiroHelp')//...->icon('heroicon-o-question-mark-circle')),
```


packages/forms/resources/views/components/field-wrapper/index.blade.php
```diff
@@ -39,7 +39,16 @@
...
+                        @switch($hint)
+                            @case($hint instanceof \Filament\Forms\Components\Actions\Action)
+                            {{ $hint }}
+                            @break
+                            @case($hint instanceof \Illuminate\Support\HtmlString)
+                            {{ $hint }}
+                            @break
+                            @default
+                            {{ \Illuminate\Support\Str::of($hint)->markdown()->sanitizeHtml()->toHtmlString() }}
+                        @endswitch
```
zepfietje, line 51: Wouldn't it be cleaner to use `@if` and `@elseif` here instead?

packages/forms/src/Components/Concerns/HasHint.php
```diff
@@ -3,15 +3,16 @@
 namespace Filament\Forms\Components\Concerns;
 
 use Closure;
+use Filament\Forms\Components\Actions\Action;
 use Illuminate\Support\HtmlString;
 
 trait HasHint
 {
-    protected string | HtmlString | Closure | null $hint = null;
+    protected string | HtmlString | Closure | Action | null $hint = null;
```
zepfietje, line 11: I think it makes sense to move `Action` before `Closure` here (also in the other places).

```suggestion
    protected string | HtmlString | Action | Closure | null $hint = null;
```

packages/forms/src/Components/Concerns/HasHint.php
```diff
@@ -25,13 +26,23 @@ public function hintIcon(string | Closure | null $hintIcon): static
         return $this;
     }
 
-    public function getHint(): string | HtmlString | null
+    public function getHint(): string | HtmlString | Action | null
     {
-        return $this->evaluate($this->hint);
+        return $this->hint instanceof Action ? $this->evaluate($this->hint)?->component($this) : $this->evaluate($this->hint);
```
zepfietje, line 31: If the closure evaluates to an `Action` object as return value, `component($this)` is not set.

zepfietje (changes requested):

packages/forms/src/Components/Concerns/HasHint.php
tanthammar: It wasn't set before either, that is why I added the condition.

packages/forms/resources/views/components/field-wrapper/index.blade.php
tanthammar: I guess that is a matter of taste. Do you want me to change it? Will add a null case though, to break early.

tanthammar: Personally I don't like nested if, elsif, else

packages/forms/src/Components/Concerns/HasHint.php
zepfietje: But previously it wasn't possible to return an `Action` object, now it is.

tanthammar: Sorry, I must misunderstand you. Don't follow what is wrong. Can you update to what is required?

packages/forms/resources/views/components/field-wrapper/index.blade.php
tanthammar: and if not null before

tanthammar: This is no problem for me at all. I'll change it. It's just a matter of taste, what you and I find most readable.

zepfietje: Yeah, here I don't see why you'd use a `switch` statement if you're writing an entire condition in the `case`s anyway. Also, if-else-if-else is more readable here.

packages/forms/src/Components/Concerns/HasHint.php
zepfietje: `$this->evaluate($this->hint)` can return an `Action` object. Currently, only if the `hint` property directly contains an `Action` object (see the ternary statement), the `->component($this)` call is chained. It should also be chained if the return value of that `evaluate($this->hint)` call is of type `Action`.

tanthammar:
Hope I fixed it now.

packages/forms/src/Components/Concerns/HasHint.php
zepfietje: Exactly what I meant, thanks @tanthammar!

packages/forms/resources/views/components/field-wrapper/index.blade.php
```diff
@@ -39,7 +39,11 @@
 
                 @if ($hint || $hintIcon)
                     <x-forms::field-wrapper.hint :icon="$hintIcon">
-                        {{ filled($hint) ? ($hint instanceof \Illuminate\Support\HtmlString ? $hint : \Illuminate\Support\Str::of($hint)->markdown()->sanitizeHtml()->toHtmlString()) : null }}
+                        @if($hint instanceof \Filament\Forms\Components\Actions\Action || $hint instanceof \Illuminate\Support\HtmlString)
```
zepfietje, line 42:
```suggestion
                        @if ($hint instanceof \Illuminate\Support\HtmlString || $hint instanceof \Filament\Forms\Components\Actions\Action)
```

zepfietje (changes requested):

tanthammar:
@zepfietje Have I fixed the requested changes?

packages/forms/resources/views/components/field-wrapper/index.blade.php
```diff
@@ -39,7 +39,11 @@
 
                 @if ($hint || $hintIcon)
                     <x-forms::field-wrapper.hint :icon="$hintIcon">
-                        {{ filled($hint) ? ($hint instanceof \Illuminate\Support\HtmlString ? $hint : \Illuminate\Support\Str::of($hint)->markdown()->sanitizeHtml()->toHtmlString()) : null }}
+                        @if ($hint instanceof \Illuminate\Support\HtmlString || $hint instanceof \Filament\Forms\Components\Actions\Action)
+                            {{ $hint }}
+                        @else
+                            {{ \Illuminate\Support\Str::of($hint)->markdown()->sanitizeHtml()->toHtmlString() }}
```
zepfietje, line 45: We should probably also handle `$hint === null` to prevent the stringable stuff in that case for performance?

zepfietje (changes requested):

danharrin:
I'd say we should use a separate hintAction() so you can have a hint AND an action, like you can have a suffix AND a suffixAction() already

packages/forms/resources/views/components/field-wrapper/index.blade.php
tanthammar: It's handled in the preceding `@if ($hint || $hintIcon)`

tanthammar:
> I'd say we should use a separate hintAction() so you can have a hint AND an action, like you can have a suffix AND a suffixAction() already

ok, I'll refactor to hintAction().

packages/forms/resources/views/components/field-wrapper/index.blade.php
zepfietje: Not always, when `$hintIcon` is set and `$hint` is `null`, that condition passes and thus `$hint` gets passed to `Str::of()`.

tanthammar: Ok

tanthammar:
Refactored to hintAction

danharrin:
thank you very much @tanthammar

invaders-xx:
@tanthammar , many thanks for this PR. I am wondering if this PR can handle showing a tooltip when hovering a hint icon ?
If so, can you please help me how to implement that ?

Many thanks in advance for your help.

David

zepfietje:
@invaders-xx, have you tried configuring a tooltip on the action? https://filamentphp.com/docs/2.x/tables/actions#tooltips

ManojKiranA:
Hey guys can someone give me the example for this PR.

So far I have tried this

```

TextInput::make('input_with_hint')
                        ->hintAction(Action::make('test')->action('abort'))
                        ->label('Hint')
                        ->required(),
```
This is the error I got. 

image

Is there any problem in code or maybe I misunderstood the API ?

ping @danharrin @tanthammar

danharrin:
Add an `icon()` to trigger the action with

ManojKiranA:
works fine

PR #4234, opened 2022-09-27
Filter hidden modal actions

ajnsn:
With this one I am not 100% sure if it's the correct approach.
I noticed that table `modalActions()` are not respecting the hidden-status of an action, my current example is list of sent emails that have a resend option, depending on their status.

```php
Tables\Actions\Action::make('view')
    ->color('secondary')
    ->label('View')
    ->icon('heroicon-o-eye')
    ->action('resend')
    ->modalContent(fn (MailSend $record) => view('filament.modals.mail-send-modal', ['record' => $record]))
    ->modalActions(fn (MailSend $record): array => [
        Pages\Actions\Action::makeModalAction('submit')
            ->authorize($record->canBeResent())
            ->color('danger')
            ->label('Resend')
            ->submit(),
        Pages\Actions\Action::makeModalAction('cancel')
            ->color('secondary')
            ->label('Close')
            ->cancel()
    ]),
```

`authorize()` should hide the resend option but it doesn't :) 
The thing I am note sure with is the typing of the `array_filter´ callback. Maybe you can take over please :) Thanks!

### DIFF
--- a/packages/support/src/Actions/Concerns/CanOpenModal.php
+++ b/packages/support/src/Actions/Concerns/CanOpenModal.php
@@ -117,7 +117,9 @@ trait CanOpenModal
         }
 
         if ($this->modalActions !== null) {
-            return $this->evaluate($this->modalActions);
+            return $this->filterModalActions(
+                $this->evaluate($this->modalActions)
+            );
         }
 
         $actions = array_merge(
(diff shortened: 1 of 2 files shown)


danharrin:
The filtering should happen in `getModalActions()` I think

ajnsn:
Aye! I updated the PR and tested `modalActions()` (see above), `modalSubmitAction()` and `modalCancelAction()`. 

As a sidenote and it might be normal but the `authorize()` method does to receive the table-`$record`. I need to put the `modalActions` in the callback with the `$record` to access it + updated my example code above.

danharrin:
Thanks!

PR #4247, opened 2022-09-28
Feature: Add Html Support for Modal Heading and SubHeading

Nuhel:


zepfietje:
What is it exactly you're trying to achieve @Nuhel?
Wouldn't it be better to use `HtmlString` as is used in other places?

packages/support/src/Actions/Concerns/CanOpenModal.php
```diff
@@ -94,8 +94,9 @@ public function modalHeading(string | Closure | null $heading = null): static
         return $this;
     }
 
-    public function modalSubheading(string | Closure | null $subheading = null): static
+    public function modalSubheading(string | View | Htmlable | Closure | null $subheading = null): static
     {
+
```
zepfietje, line 99: Please undo these unrelated blank line changes.

zepfietje (changes requested):

Nuhel:
> What is it exactly you're trying to achieve @Nuhel? Wouldn't it be better to use `HtmlString` as is used in other places?

Yes It is better to use HtmlString to use in other places. 
Sometime client demands are fancy, we need these type of things, As my client want's a colorful sub-heading on a modal.
So I think it is good if filament supports it :)

zepfietje:
Have you tried styling the subheading as desired using the `filament-` CSS hook classes?

Nuhel:
> 

Yes, But with Html we can do much more things right?

zepfietje:
What specifically would you need that isn't currently possible?

Passing a view or HTML brings risks, so it's better to keep it simple if that suffices.

awcodes:
I’m not sure this is a good idea since the heading is getting rendered inside an h2. Injecting html here could cause accessibility issues. CSS and extraAttributes would probably be a better use.

Nuhel:
> What specifically would you need that isn't currently possible?
> 
> Passing a view or HTML brings risks, so it's better to keep it simple if that suffices.

Okay, I thought Using HtmlString object is fine with filament as HtmlString are being used already, If it is risky then , There is nothing to compromise.

Nuhel:
> I’m not sure this is a good idea since the heading is getting rendered inside an h2. Injecting html here could cause accessibility issues. CSS and extraAttributes would probably be a better use.

Yes that is what I am trying to achieve. Why we should be limited to h2? As with filament we can do much more. We can skip being render inside h2, if the heading is view or HtmlString right?

Btw I am not arguing, I have lot more to learn. I thought it will be good if filament supports it, That's why i made this pr.

Nuhel:
> Why would you use a different element than `h2` for the heading? To wrap it in something else? Semantically it makes sense to use `h2` and you could always use `span` tags inside the `h2` if needed. If there's no concrete use case, I'd hold off from going outside the `h2` tag.
> 
> As I said in my first comment, we should probably just add support for passing an instance of `\Illuminate\Support\HtmlString` to the heading. So if you could update this PR to add just that, it's all fine. 🙂


I am totally agree with you For heading we should always use h2. Actually i may not understand your first comment properly.

We can just add support for HtmlString than, I am updating my pr with this. 

Thank you.

danharrin:
Thanks

PR #4253, opened 2022-09-28
feature: Auto delete uploaded file

egyjs:
by using method, we can use  this Enable Auto Delete unwanted Files


zepfietje:
Could you elaborate on the motivation behind this PR, @egyjs?

zepfietje:
We've discussed this change internally and found that it leads to too many unintended side effects.

danharrin:
The proper way to achieve this is through a model observer

PR #4264, opened 2022-09-29
fix: Wizard steps indicator responsiveness

almo2k:
Hi team, 

Just working with the wizard and ran into this visual issue when having more than 4 steps

Screenshot 2022-09-29 at 10 17 30

I've added a count check then switch to grid if there are 4 or more steps

Screenshot 2022-09-29 at 10 16 46

I've tried to make the code more readable but I'm aware that It could/should be refactored inline.

There's probably more that could be done with the borders but I think it works really well with the two column grid layout

Kind regards

### DIFF
--- a/packages/forms/resources/views/components/wizard.blade.php
+++ b/packages/forms/resources/views/components/wizard.blade.php
@@ -83,20 +83,15 @@
     <ol
         {!! $getLabel() ? 'aria-label="' . $getLabel() . '"' : null !!}
         role="list"
-        @if (count($getChildComponentContainer()->getComponents()) >= 4)
+
             @class([
-            'border border-gray-300 shadow-sm bg-white rounded-xl overflow-hidden divide-y divide-gray-300 grid md:grid-cols-2 md:divide-y-0',
+            'border border-gray-300 shadow-sm bg-white rounded-xl overflow-hidden divide-y divide-gray-300 md:flex flex-wrap md:divide-y-0',
             'dark:bg-gray-800 dark:border-gray-700 dark:divide-gray-700' => config('forms.dark_mode'),
             ])
-        @else
-            @class([
-                'border border-gray-300 shadow-sm bg-white rounded-xl overflow-hidden divide-y divide-gray-300 md:flex md:divide-y-0',
-                'dark:bg-gray-800 dark:border-gray-700 dark:divide-gray-700' => config('forms.dark_mode'),
-            ])
-        @endif
+
     >
         @foreach ($getChildComponentContainer()->getComponents() as $step)
-            <li class="group relative overflow-hidden md:flex-1">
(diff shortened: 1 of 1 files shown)


zepfietje:
We should be able to achieve this using pure (Tailwind) CSS breakpoints, right?

Also, it looks like the step indicators need a bit more tweaking when wrapped to a new row.

Apart from that, thanks for bringing this up, @almo2k!

ziming:
Can we opt out of the switch to grid if more than 4 steps?

zepfietje:
I think we shouldn't necessarily tie this to > 4. Just need to make it responsive. Doesn't feel like there should be an opt-out for that, right?

almo2k:
I agree with the responsive approach and not tie to a specific number, I think I came at this from the wrong angle of php conditional rather than CSS as that's what worked for my use case.

almo2k:
I've just put together a possible workaround using just CSS, the complier needs updating to use w-1/2

ziming:
part of me think hiding the wizard steps headers on mobile is a more responsive approach

danharrin:
@filamentphp/design could you handle this PR please and decide on the best approach? Since it's entirely design related. You have my permission to merge it as well.

awcodes:
Setting the steps to always be 1/2 is not a good approach to this. If people have 3 steps and are forced to have it wrap on a screen where there's more than enough room to fit them in one row will be very upset.

I think it might be better to have a layout option for the steps that allows the user to move them to the side of the form and be vertical (which is typically how wizards with a lot of steps are done).

@zepfietje thoughts?

zepfietje:
Agree, @awcodes.

Think we should make the step indicators stack vertically on mobile: https://tailwindui.com/components/application-ui/navigation/steps#component-c84570c0de78ae9fc70cfb24a82abc87.

awcodes:
I agree with stacking them on mobile, but that falls apart if there are a lot of steps. Image have to scroll past 8-10 steps to start filling out the form, and do that 8-10 times.

Need to give it some more thought.

Nuhel:
What about pushing the form inside the steps, like a collapsible/accordion on mobile screen?
And yes it will be configurable, using method like snapInsideHeading or something else.

image

zepfietje:
I've looked into this once again.
I don't see this as a Filament bug honestly.
To me, wizards aren't designed to be used with such a large number of steps.

If this really is something that's required for more people, feel free to come up with a maintainable solution (like an option to always show the wizard steps vertically).

96fallenangel:
197332059-c1c0a441-d28d-4242-b75c-b1f13404a4b9
 How can I achieve this in the steps of the wizard?

zepfietje:
You would probably have to build that custom, @96fallenangel.

PR #4269, opened 2022-09-29
Currency formatting default to config value

ashleyhood:
Since `akaunting/laravel-money` supports the default currency being set in the key `default_currency` from the config file `config/money.php`, Filament should honour this when formatting state with the `money()` table helper.

This will keep the old behaviour of defaulting to 'usd' if the config key `default_currency` is not set.


danharrin:
Lovely, thanks.

PR #4283, opened 2022-09-29
feature: configure global search results limit in a resource

cameroncox:
added a configurable option to Resource to allow limiting the amount of results when getGlobalSearchResults is called.


danharrin:
thanks

PR #4307, opened 2022-10-01
Added Htmlable Support For TableHeader.

Nuhel:
Always making view, and pass it as table header is not something we want all time as we can use htmlstring in filament.

So I add support for this.

Also i see there may be issue with table headeraction, So what i have done , If the table header pass as argument is Illuminate\Support\HtmlString, I just added the action with the heading.

Here is an demo what it look like.
Screenshot 2022-10-01 at 11 19 16 AM

Screenshot 2022-10-01 at 11 31 38 AM

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -1,170 +1,171 @@
 @php
-use Filament\Tables\Actions\Position as ActionsPosition;
-use Filament\Tables\Filters\Layout as FiltersLayout;
+    use Filament\Tables\Actions\Position as ActionsPosition;
+    use Filament\Tables\Filters\Layout as FiltersLayout;
 
-$actions = $getActions();
-$actionsPosition = $getActionsPosition();
-$actionsColumnLabel = $getActionsColumnLabel();
-$columns = $getColumns();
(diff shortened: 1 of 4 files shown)


danharrin:
Your IDE has made hundreds of formatting changes, please revert them

Nuhel:
> Your IDE has made hundreds of formatting changes, please revert them

Yes, Sorry for that.
Have a look now :)

packages/tables/resources/views/index.blade.php
```diff
@@ -166,7 +166,18 @@ class="filament-tables-header-container"
             {!! ! $renderHeader ? 'x-cloak' : null !!}
         >
             @if ($header)
-                {{ $header }}
+                @if (gettype($header) == 'object' && get_class($header) == 'Illuminate\Support\HtmlString')
```
Z3d0X, line 169:
```suggestion
                @if (gettype($header) == 'object' && $header instanceof \Illuminate\Support\HtmlString)
```

danharrin:
We should actually support Htmlable in the `getTableHeading()`, not `getTableHeader()`. You want to keep the action buttons, but thats not the behaviour of `getTableHeading()`.

danharrin:
done

PR #4320, opened 2022-10-02
Using Laravel App Call in colors method,

Nuhel:
To be able to place callback function params in any order like everywhere else.

### DIFF
--- a/packages/tables/src/Columns/Concerns/HasColors.php
+++ b/packages/tables/src/Columns/Concerns/HasColors.php
@@ -26,7 +26,7 @@ trait HasColors
             foreach ($colors as $color => $condition) {
                 if (is_numeric($color)) {
                     $stateColor = $condition;
-                } elseif ($condition instanceof Closure && $condition($state, $record)) {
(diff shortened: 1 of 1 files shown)


zepfietje:
The closure passed to the `colors()` method is evaluated using Laravel's `app()->call()`: https://filamentphp.com/docs/2.x/forms/advanced#using-closure-customisation.
I think that should work for your use case, right?

danharrin:
No, @nuhel is right but we cant implement it until v3 as its a breaking change

zepfietje:
Wouldn't it currently be possible to resolve the parameters from the main closure you pass in?

And yeah, thought this was part of the `[3.x] Use $this->evaluate() everywhere` roadmap entry.

Nuhel:
> No, @Nuhel is right but we cant implement it until v3 as its a breaking change

Thank you :) .

zepfietje:
I'll leave this PR closed as we'll need to update it in a few other places as well for v3.

Thanks anyway, @Nuhel!

coolsam726:
I would like to revisit this, and submit a PR to support closures in Panel Colors.

zepfietje:
Feel free to work on that, @coolsam726!

PR #4323, opened 2022-10-02
feat(disabled dates) Allow to disable specific dates (Datetime field)

saifulapm:
Sometime need to disable specific dates like Booking/Schedule app. I think it will be good one. 

Anyway, Thanks all your hark working and making it awesome.


zepfietje:
Hey @saifulapm, thanks for opening a PR!

Did you accidentally close it?

PR #4325, opened 2022-10-02
Allow disabling date time picker dates

saifulapm:
Sometime need to disable specific dates like Booking/Schedule app. I think it will be good one.

Anyway, Thanks all your hark working and making it awesome.


saifulapm:
Also I am thinking to apply minDate/maxDate to Timepicker. Now Timepicker always start from 0.

PR #4331, opened 2022-10-03
Update Documentation for badge column

Nuhel:
As everywhere callback function are evaluated, we can access method params just using proper naming,

But badge column color and icon methods callback functions are not evaluated,  just accepting only one param in callback function will always give the state not the whole record. And it's bit confusing .


danharrin:
In the example, I would use the `$state` parameter instead of `$record->status` since the column is already called `status`

PR #4337, opened 2022-10-03
Make rows default sortable by multiple columns

### LINKED ISSUE #4068: Default sort tables by multiple columns
laravel-filament/filament#2626

dev:
For example:
```php
return $table
            ->columns([
                TextColumn::make('title')
                          ->label('Title'),

                BadgeColumn::make('is_active')
                           ->label('Status')
                           ->enum([
                               0 => 'Inactive',
                               1 => 'Active',
                           ])
                           ->colors([
                               'success',
                               'danger' => 0,
                           ])->sortable(),

                          ->label('Created at')
                          ->sortable()
                          ->dateTime('d/m/Y H:i'),
            ])
            ->actions([
                Tables\Actions\ViewAction::make(),
            ])->defaultSort('is_active-desc,created_at-asc');
```

This shows the active records first, ordered by creation date and then shows the inactive records ordered by creation date.

### DIFF
--- a/packages/tables/src/Concerns/CanSortRecords.php
+++ b/packages/tables/src/Concerns/CanSortRecords.php
@@ -74,6 +74,15 @@ trait CanSortRecords
         }
 
         if ($columnName === $this->getDefaultTableSortColumn()) {
+            if(\Str::contains($columnName, ',')) {
+                $columnNames = explode(',', $columnName);
+                foreach($columnNames as $columnName) {
+                    $direction = \Str::after($columnName, '-');
+                    $columnName = \Str::before($columnName, '-');
+                    $query->orderBy($columnName, $direction);
+                }
+                return $query;
+            }
             return $query->orderBy($columnName, $direction);
         }
 


zepfietje:
Thanks for the PR, @dev!

I'm not sure if this is the right API though.
What did you have in mind, @danharrin?

pxlrbt:
I think it would be better if we either use array syntax or a dedicated class:

```php
['is_active', 'DESC']
```

``php
Order/Sort::column('is_active')->desc()
```

Nuhel:
> I think it would be better if we either use array syntax or a dedicated class:
> 
> ```
> ['is_active', 'DESC']
> ```
> 
> ```
> new Sort('is_active')->asc()
> Sort::column('is_active')->desc()
> Order::column('is_active')->desc()
> ```

I agree with you. It will be easy to to maintain if there is sorting with relational column.
As filament table automatically handle relational column.

danharrin:
I don't feel comfortable mixing the `$tableSortColumn` and `$tableSortDirection` properties data together, and merging them would be a breaking change in v2. So I'm going to hold off on implementing this feature until v3. Thanks for the PR though.

PR #4346, opened 2022-10-04
Add `xs` and `xl` button sizes

intrepidws:
This PR introduces `xs` and `xl` sizes to actions. Note that I wasn't sure how you prefer to compile the assets in PRs, or whether that's just something that occurs in the build process.


intrepidws:
In general, I find that the buttons are larger than I would prefer and I think that the `xs` size is a great addition (especially in scenarios where there are several buttons in a row). Personally, I think the `xl` is overkill and too large - but it feels unbalanced to add `xs` without `xl`. While I think there could be an accessibility argument for keeping `xl`, I wouldn't be opposed to removing it since I think `lg` is already plenty large.

This is what an `xs` button would look like, next to an `sm` button and an `md` button:

Screen Shot 2022-10-04 at 11 45 44 AM

pxlrbt:
Yeah, the XS button looks unbalanced. Y-Padding > X-Padding.

intrepidws:
> Yeah, the XS button looks unbalanced. Y-Padding > X-Padding.

I actually think that's the case for all of the buttons in that screenshot, no? I assumed it was on purpose, I guess.

zepfietje:
The existing buttons can definitely use some tweaking (v3 maybe?) but the `xs` size stands out as unbalanced to me.

zepfietje:
But let's not dive into this too far now.

I'd like to approach this by first asking: "do we really need these sizes"?

intrepidws:
I vote for including `xs` but it's up to you guys. Let me know what you want to do.

pxlrbt:
Even bootstrap only comes with 3 sizes: https://getbootstrap.com/docs/5.0/components/buttons/#sizes

intrepidws:
Right, mirroring Tailwind was sort of my goal here, since Filament does that in so many other areas already.

I'm happy to rework the `xs` button padding/margin, perhaps one of you could walk me through what I should be doing in terms of compiling assets as part of the PR process?

zepfietje:
Looking at the Tailwind UI buttons, I think they have the following sizes:
- `xs`
- `sm`
- `md`
- `lg`
- `xl`

The current default Filament button size (`md`) is too small in terms of height (it's now closer to `sm`). I definitely think we should rework the button to include the above sizes and be consistent across the scale. Some might regard this as a breaking change UI-wise, so have to be careful here maybe. Still, I think we shouldn't hold off until v3 for these changes. What do you think, @awcodes @danharrin?

If we decide to continue, I'm happy to take over this PR if you like, @intrepidws.

pxlrbt:
You didn't ask for my opinion (😜), but I think we should include this with the v3 redesign.

intrepidws:
> If we decide to continue, I'm happy to take over this PR if you like, @intrepidws.

That's fine by me, whatever makes the most sense.

intrepidws:
One thing that might be good to clean up with a v3 redesign is the fact that actions have a default size of `sm` in tables but a default size of `md` on pages. That's a little confusing.

awcodes:
I'm ok with adding the sizes as options, I disagree that the md is too big, at least on desktop in the page and form actions area.

I would personally never want to use anything larger than sm in the tables though due then making rows unnecessarily tall. Which would get magnified if the default font size for the rows is decreased from 1rem.

danharrin:
I think the new options are good but the spacing does need fixing

intrepidws:
> I think the new options are good but the spacing does need fixing

@danharrin I'm happy to make some fixes for the spacing. Are you OK with doing this within v2 or should it wait for v3?

danharrin:
v2 is fine, we just cant change spacing on the existing sizes

intrepidws:
I have updated the styles for `xs` and `xl` to make them more consistent with the existing sizes. You can see samples of each in the screenshot below. As you can see, there is a bit of a jump between `md` and `lg`. In an ideal world, I would simply change `lg` to be `xl` and then slot another size in between, but that would be a breaking change. Perhaps that can be adjusted in v3?

Screen Shot 2022-10-04 at 8 30 49 PM

zepfietje:
So if we can't update the existing sizes in v2, the scale will look inconsistent. I think we should still make the new sizes match the ones in Tailwind UI. Probably `xs` and `sm` will be super similar then, but that's still better than making `xs` super small and having to change that in v3 as well.

intrepidws:
> So if we can't update the existing sizes in v2, the scale will look inconsistent.

The scale is already inconsistent, unfortunately. The middle three buttons are the existing sizes and, at least to my eye, the inconsistency occurs between the jump from `md` to `lg`. I don't believe that the addition of `xs` or `xl` adds to that inconsistency.

zepfietje:
We can surely add `xs` and `xl`. I'm just saying we should make those sizes look ok matching the Tailwind UI sizing instead of the current Filament sizes.

zepfietje:
I looked into this a little more. The current Filament button sizes (sm, md, lg) are closer to the Tailwind UI xs, sm and lg/xl sizes respectively.

In my opinion, we should hold off adding xs and xl to Filament in v2, since the inconsistency in the scale will be super confusing.

What we can do now, however, is adding CSS hook classes for all sizes you pass in like `filament-button-size-xs`. Would that solve your current case, @intrepidws? Let me know, then I'll add those.

Marking this PR as draft for v3.

zepfietje:
Closing this PR as v3 will require it to target a different branch.

I've also added this to our roadmap: https://github.com/orgs/filamentphp/projects/2/views/6?query=is%3Aopen+sort%3Aupdated-desc.

zepfietje:
@intrepidws, xs and xl size support has been added for v3 in https://github.com/filamentphp/filament/pull/5563.

intrepidws:
Love it @zepfietje thanks for all that you do.

zepfietje:
Glad you like it! Thanks for your support, @intrepidws. 🙏

PR #4353, opened 2022-10-04
Document recordSelectOptionsQuery() method

intrepidws:
This PR adds documentation for the `Document recordSelectOptionsQuery()` method, which was added in #3588

### DIFF
--- a/packages/admin/docs/02-resources/06-relation-managers.md
+++ b/packages/admin/docs/02-resources/06-relation-managers.md
@@ -395,6 +395,16 @@ In this example, `$action->getRecordSelect()` outputs the select field to pick t
 
 Please ensure that any pivot attributes are listed in the `withPivot()` method of the relationship *and* inverse relationship.
 
+### Scoping attachable models
+
+You may want to apply a scope to the models available for attaching in a relation manager:
+
+```php
+use Filament\Tables\Actions\AttachAction;
+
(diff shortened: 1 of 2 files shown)


packages/admin/docs/02-resources/06-relation-managers.md
```diff
@@ -395,6 +395,16 @@ In this example, `$action->getRecordSelect()` outputs the select field to pick t
...
 Please ensure that any pivot attributes are listed in the `withPivot()` method of the relationship *and* inverse relationship.
 
+### Scoping attachable models
+
+You may want to apply a scope to the models available for attaching in a relation manager:
+
+```php
+use Filament\Tables\Actions\AttachAction;
+
+AttachAction::make()->recordSelectOptionsQuery(fn ($query) => $query->whereBelongsTo(auth()->user())
```
danharrin, line 405: Please add the `recordSelectOptionsQuery()` method on a new line in both examples, and `$query` should be an instance of `Builder`, and that class needs importing too

intrepidws: All set Dan.

danharrin:
Maybe it's worth adding examples for AssociateAction too?

intrepidws:
> Maybe it's worth adding examples for AssociateAction too?

Good call, what do you think of the updates?

danharrin:
It doesnt apply to Detach or Dissociate

intrepidws:
Right, of course it doesn't - sorry. That's what I get for doing this late at night. This is fixed.

PR #4355, opened 2022-10-04
Docs: Document Action size() Method

intrepidws:
This documents the current possible sizes of `sm`, `md` and `lg`. These docs would need to be updated if #4346 gets merged (seems like this will probably be in v3 though).


intrepidws:
FYI, I made this a draft since it looks like #4346 will likely get merged and this PR will need to be updated to account for the new `xs` and `xl` sizes.

intrepidws:
As it now looks like #4346 will not be merged until v3, I have revived this PR.

PR #4365, opened 2022-10-05
Only show the select for paginating tables when there is more than 1 option

SebastiaanKloos:


danharrin:
Please move the check from the component to the parent component that renders it. Also `@if (` not `@if(`

SebastiaanKloos:
> Please move the check from the component to the parent component that renders it. Also `@if (` not `@if(`

Should be fixed now!

PR #4373, opened 2022-10-05
Docs: Note about referencing other fields from rules() method

intrepidws:


danharrin:
We actually suggest using `$get` for this which is a consistent API throughout Filament

`required(fn ($get) => filled($get('second_field')))`

intrepidws:
Right but that only works for rules that have a specific method, like `required`. If you need to use the `prohibited_if` rule, which doesn't have it's own method in Filament, then you need to use the `rules()` method, like this:

```php
TextInput::make('state')->rules(['prohibited_if:data.country,US'])
```

Perhaps the answer here is to PR missing rule methods instead?

danharrin:
I think `prohibited()` is one of the very few we are missing

intrepidws:
OK, I added that instead in PR #4376

PR #4375, opened 2022-10-05
Field Hint Tooltip Functionality

intrepidws:
When providing a hint or helper text, I often find that the text is not important enough to appear all the time and/or it is too long for the available space.

This PR allows users to make a form field hint appear as an icon rollover by specifying `->hintTooltip(true)`. When this is set, the text will only appear as a tooltip when the user hovers over the icon. If a specific icon is not provided via `->hintIcon()`, then the information circle icon will be used by default.

Screen Shot 2022-10-05 at 12 24 00 PM


danharrin:
I suggest using hint actions for this instead - #4215

intrepidws:
Hmmm interesting, I hadn't seen that before. Will that allow you to utilize a tooltip? A modal feels like overkill for quick hint text.

danharrin:
IMO a tooltip feels overkill for quick hint text too 😁 but yeah you could use `->tooltip()` on the action object when that is merged

intrepidws:
That makes sense to me, thanks for clarifying Dan.

PR #4376, opened 2022-10-05
Add prohibited validation rule

intrepidws:


Z3d0X:
This was intentionally left out in https://github.com/filamentphp/filament/pull/3890#issuecomment-1249214324

intrepidws:
Hmm... interesting. Is there a reason for that? It feels like a useful validation rule to me.

danharrin:
I don't really know why I added it to the list, but I've never had a use for it either

PR #4378, opened 2022-10-06
Font weights

martin-ro:
Add more options for font weights.


martin-ro (commented):
Add more options for font weights.

PR #4397, opened 2022-10-06
docs: use app.css instead of filament.css

eduardoarandah:
Instead of asking user to create a filament.css file, use app.css

app.css is standard in tailwind, so less questions to ask. 

also, user won't have to change much, since laravel breeze already reads app.css in `vite.config.js`


danharrin:
Since lots of people have a separate admin panel section to their app, we recommend this approach instead, as then the main app doesnt have all the admin panel styles. Thanks though.

PR #4406, opened 2022-10-07
fix: Escape inputs without closing modal

### LINKED ISSUE #4256: Pressing ESC on interactable components inside a modal should not close it
### Package

filament/forms

### Package Version

^2.0

### Laravel Version

9.32.0

### Livewire Version

^2.5

### PHP Version

8.1.6

### Problem description

If you have components like DateTimePicker or MultiSelect inside a modal and you press ESC on those components - the modal closes. 

This makes navigating with only a keyboard very cumbersome imho. 

### Expected behavior

I expect all the components that require an interaction to close the interaction instead before closing the modal. 

### Steps to reproduce

Create a DateTimePicker inside a modal (relation etc.)

### Reproduction repository

https://github.com/dissto/filament

### Relevant log output

_No response_

zepfietje:


zepfietje:
Not sure if there's any other fields that should be handled?

packages/forms/resources/views/components/date-time-picker.blade.php
```diff
@@ -19,6 +19,7 @@
             shouldCloseOnDateSelection: @js($shouldCloseOnDateSelection()),
             state: $wire.{{ $applyStateBindingModifiers('entangle(\'' . $getStatePath() . '\')') }},
         })"
+        x-on:keydown.esc="isOpen() && $event.stopPropagation()"
```
danharrin, line 22: does this work?

```suggestion
        x-on:keydown.esc.stop="isOpen()"
```

zepfietje: Hmm is that even an Alpine API that should work? Let me check.

danharrin: I think so

danharrin: wait no, it should only stop IF it's isOpen. so you're right

zepfietje: Doesn't work. The `stop` modifier is always applied then, no matter what's in the event handler.

PR #4407, opened 2022-10-07
fix: Empty state hides column search

### LINKED ISSUE #4224: Individual search column inputs disappear when empty state

danharrin:
Fixes #4224


Sicklou:
Thanks a lot !

Looks not like an easy one to fix.

mikebouwmans:
@danharrin Currently implementing individual column search using V3. Is it possible that searching with empty results removes the search input fields in the columns again? It's happening for me at least, not sure if I'm missing something or that it's expected behaviour for V3.

PR #4410, opened 2022-10-07
Fix code example to match text body in the documentation

emmadesilva:
Since the text reads "Or dynamically calculate the **color**" I assume the example is meant to show this.

I confirmed the updated example works.


danharrin:
Thanks!

PR #4411, opened 2022-10-07
Add button size class

zepfietje:
Not sure about the naming.
Shall we prefix the classes with their property: `-size-`?

### DIFF
--- a/packages/support/resources/views/components/button.blade.php
+++ b/packages/support/resources/views/components/button.blade.php
@@ -16,7 +16,7 @@
 
 @php
     $buttonClasses = array_merge([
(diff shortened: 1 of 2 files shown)


danharrin:
we prob need the same for icon button, link, etc.

also - maybe `button-sm` instead of `button-size-sm` ?

zepfietje:
> we prob need the same for icon button, link, etc.

Been thinking about that as well, but thought we'll add more and make it consistent in v3 anyway.

> also - maybe `button-sm` instead of `button-size-sm` ?

So that was exactly my dilemma. Won't we run into naming issues with other properties somewhere if we don't prefix it with the property name?

PR #4420, opened 2022-10-07
Add support for the before method in policy

lolen:
Currently there is no support for the `before` in policy method because checking if a method exists returns false.


danharrin:
Thanks

PR #4428, opened 2022-10-08
Allow customizing IconColumn size

emmadesilva:
## The new feature
Adds a new `->size()` method to the `IconColumn`. This is used customize the size of the icon using the `size()` method, where the default is `lg`:

```php
use Filament\Tables\Columns\IconColumn;

IconColumn::make('is_featured')
    ->options([
        // ...
    ])
    ->size('sm') // `sm`, `md` or `lg`
```

This PR also adds this to the documentation. I did not add tests, as I could not find a test covering this component.

## Feedback on what sizes?
I am open to a discussion on which sizes this should support. At the current state, the sizes match the ones found in [`icon-button.php#L32`](https://github.com/filamentphp/filament/blob/2.x/packages/support/resources/views/components/icon-button.blade.php#L32), however to ensure backwards compatibility this leads to the default size being `lg` which feels weird to me. The first commit offers a size set that feels more natural to me, where the default size is `md`. I'm also wondering if we should add `xs` and `xl` versions? I'd love some input on this!


caendesilva:
I also had an idea that it could support arbitrary integer values, but that might add too much complexity, as it would require a type check. Thoughts?

awcodes:
I like the idea of being able to pass a number but that will only work if there is a map from the integer to the tailwind classes, which may be more trouble than it's worth to maintain over time. Feel free to give it a go and see how it feels.

Let us know.

danharrin:
A number is probably risky though, the class still needs to be purged

awcodes:
On second thought using a number would require even more things to make it work like limiting it to certain numbers in case they don't exist in TW. 

Can you please add an xl too. 

Don't think there's any need for xs.

danharrin:
Sorry about the merge conflicts, we just deleted a docs page and renamed the existing one

caendesilva:
> On second thought using a number would require even more things to make it work like limiting it to certain numbers in case they don't exist in TW.
> 
> Can you please add an xl too.
> 
> Don't think there's any need for xs.

I'll put in an xl! As for allowing arbitrary numbers, the worst that can happen if a user uses an invalid number is that it won't work right? Is it our/Filaments responsibility to ensure that doesn't happen? 

> A number is probably risky though, the class still needs to be purged

Does it? I thought Tailwind uses JIT compiling. And I would guess that anyone who is at the stage where they'd need arbitrary numbers are likley already compiling assets on their end.


It's rather late where I live, so I'll take add `xl` tomorrow, as well as solve the merge conflicts. I may also play around with arbitrary numbers just to see how it feels in practise. Feel free to turn this PR into a draft if that is easier on your end. I do appreciate this discussion, as I was unsure what would fit in best with Filament!

danharrin:
A good compromise is allowing `sm` - `xl` and then if the user enters anything else, it gets treated as a class name that they compile themselves. You can use `match($size) { default => $size}` for this

awcodes:
Just make sure with the arbitrary numbers that there's a fall back. I'd rather it not work than break.

awcodes:
Keep in mind too that by arbitrary you can't do something like w-[10] that won't work. CSS doesn't know what 10 is.

That's why there has to be a map between the numbers and the tailwind sizing scale.

awcodes:
Also has to handle things like ->size(-1). 

We can't babysit people, but we can make the system stronger.

zepfietje:
Just zooming out a little after reading the good conversation you guys had above.
What's the motivation for changing the `IconColumn` size, @caendesilva?

emmadesilva:
> A good compromise is allowing `sm` - `xl` and then if the user enters anything else, it gets treated as a class string that they compile themselves. You can use `match ($size) { default => $size}` for this

That sounds like a good compromise to me. I'll try it out. 

> Just make sure with the arbitrary numbers that there's a fall back. I'd rather it not work than break.

Yeah I think anything invalid should add revert to the default behaviour.

> We can't babysit people, but we can make the system stronger.

Very true.

emmadesilva:
> Just zooming out a little after reading the good conversation you guys had above. What's the motivation for changing the `IconColumn` size, @caendesilva?

For me personally, I have an additional library of SVG icons that look HUGE in comparison to the built-in Heroicons. So I wanted an option that could help me normalize the sizes without having to edit hundreds of SVG files.

zepfietje:
> For me personally, I have an additional library of SVG icons that look HUGE in comparison to the built-in Heroicons. So I wanted an option that could help me normalize the sizes without having to edit hundreds of SVG files.

Thanks for the insight, @caendesilva.

There's probably not much use in customizing the IconColumn size on a per-use basis, as you'd like to keep things consistent and stick to the same size throughout your application, right? That's what you're trying to achieve here as well: making the two different icon libraries match in size. I wouldn't recommend mixing different icon sets to start with. But if you really have to, you could also add a custom class to icon columns with the 'huge' icons using `extraAttributes()` and target that CSS class in your own stylesheet.

I'm not saying we should definitely close this PR, but I'm being careful adding a feature if it isn't necessarily the required fix to start with. Icon sizing can be super opinionated, so it's hard to find a good sizing scale that matches everyone's desires. Hopefully this doesn't come across as negative, I'm just evaluating pros and cons. If we can settle on a good solution though, we might just add it. 🙂

emmadesilva:
> > For me personally, I have an additional library of SVG icons that look HUGE in comparison to the built-in Heroicons. So I wanted an option that could help me normalize the sizes without having to edit hundreds of SVG files.
> 
> Thanks for the insight, @caendesilva.
> 
> There's probably not much use in customizing the IconColumn size on a per-use basis, as you'd like to keep things consistent and stick to the same size throughout your application, right? That's what you're trying to achieve here as well: making the two different icon libraries match in size. I wouldn't recommend mixing different icon sets to start with. But if you really have to, you could also add a custom class to icon columns with the 'huge' icons using `extraAttributes()` and target that CSS class in your own stylesheet.
> 
> I'm not saying we should definitely close this PR, but I'm being careful adding a feature if it isn't necessarily the required fix to start with. Icon sizing can be super opinionated, so it's hard to find a good sizing scale that matches everyone's desires. Hopefully this doesn't come across as negative, I'm just evaluating pros and cons. If we can settle on a good solution though, we might just add it. 🙂

Thanks for the response! For some additional insight, the additional library are for file extension icons in a file manager. I think you definitely solved the "arbitrary sizing issue" we were discussing by using `extraAttributes()`. 

I'm going to mark this PR as ready for review, as I think the current state of it adds a nice simple feature with comparatively little complexity. I would of course understand if this is not something that fits for Filament, and I am absolutely open to further discussion and changes!

zepfietje:
I've made the implementation a little more robust. Should be good to go now! 🚀

Thanks @caendesilva.

emmadesilva:
> I've made the implementation a little more robust. Should be good to go now! 🚀
> 
> Thanks @caendesilva.

Looks beautiful, thank you!

zepfietje:
As a note: you can pass a custom size like `size('2xl')`, which adds a `filament-tables-icon-column-size-2xl` class which you can target with your own CSS. 😎

emmadesilva:
> As a note: you can pass a custom size like `size('2xl')`, which adds a `filament-tables-icon-column-size-2xl` class which you can target with your own CSS. 😎

That's awesome!

packages/tables/resources/views/columns/icon-column.blade.php
```diff
@@ -1,22 +1,42 @@
...
+            null => \Illuminate\Support\Arr::toCssClasses(['text-gray-700', 'dark:text-gray-200' => config('tables.dark_mode'),]),
+            default => $stateColor,
+        },
+        match ($size) {
+            'xs' => 'h-3 w-3',
+            'sm' => 'h-4 w-4',
+            'md' => 'h-5 w-5',
+            'lg' => 'h-6 w-6',
+            'xl' => 'h-7 w-7',
+            default => null,
```
danharrin, line 21: How about `default => $size` to support custom size classes @zepfietje

PR #4430, opened 2022-10-08
Cancel on edit/create page redirects to previous page.

flyingeek:
On Edit/Create pages, if you click cancel you are redirected to the index page. By using the HasBackUrl trait you will be redirected to the previous page.
Additionally the save, delete, forcedelete also get redirected. Finally, the breadcrumb first item is also linked to the same previous page.

This is working really well for me, but I think that this PR should be implemented in the core for the cancel button and to only add the trait for the save/delete/forcedelete and breadcrumb redirections.

I am by no way an expert in PHP/Laravel/Filament so double check ;-)

What do you think ?


danharrin:
I think this should probably be default behaviour if `url()->previous()` is set... not a trait. We can use this package - https://github.com/ralphjsmit/livewire-urls#previous-url

flyingeek:
> I think this should probably be default behaviour if `url()->previous()` is set... not a trait. We can use this package - https://github.com/ralphjsmit/livewire-urls#previous-url

No this package does not work, if you change the order of the page of the list page for example, it does not get recorded as this is done using a Livewire request.
url()->previous() should be set when going from list to edit or create unless referer is blocked.

danharrin:
Okay, use that then.

flyingeek:
> Okay, use that then.

Ok but if url()->previous() is set, overriding cancel is easy. But what about save, should we have a public method to get the snapback url ? and then let the user override getRedirectUrl with that value ? What about delete/forcedelete and the breadcrumbs ?

danharrin:
Thanks

juliomotol:
We've found a loop here where:

1. Go to List Page then click create.
2. Opens Create Page then proceeds to creating a new resource.
3. Gets redirected to Edit Page.
4. Clicking the "Cancel" redirects back to the Create Page
5. Clicking the "Cancel" again redirects back to the Edit Page
6. Loop found in # 4 and # 5

Shouldn't we expect that when after creating the resource, on the Edit Page, the cancel should redirect us back to the List Page?

Let me know your thoughts!

PR #4433, opened 2022-10-08
feature: Halt

### LINKED ISSUE #4380: Stop execution from lifecycle hooks
### Discussed in https://github.com/filamentphp/filament/discussions/4302

<sup>Originally posted by **jcgmalave** September 30, 2022</sup>
Is it possible when using the beforecreate or before save function, to perform a validation and if it is fulfilled, send a notification and stop the execution of the create or save until it is corrected? and return to the main resource</div>

Throw an exception to stop execution from `before` lifecycle hook or `before()` action method

danharrin:
Fixes #4380

### DIFF
--- a/packages/admin/.stubs.php
+++ b/packages/admin/.stubs.php
@@ -53,7 +53,7 @@ namespace Livewire\Testing {
 
         public function assertPageActionDoesNotHaveLabel(string $name, string $label): static {}
 
-        public function assertPageActionHeld(string $name): static {}
+        public function assertPageActionHalted(string $name): static {}
 
         public function assertHasPageActionErrors(array $keys = []): static {}
 
--- a/packages/admin/docs/02-resources/02-listing-records.md
+++ b/packages/admin/docs/02-resources/02-listing-records.md
@@ -247,6 +247,35 @@ CreateAction::make()
     })
 ```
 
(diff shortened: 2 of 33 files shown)


cobraprojects:
it works fine on `beforeCreate`

but `beforeSave` on edit page
it displays the notification but doesn't halt (it saves the record any way)

here is my code

```php
/**
 * @throws Halt
 */
protected function beforeSave(): void
{
    $options = collect($this->data['options'] ?? []);
    if ($options->count()) {
        $isTrue = $options->pluck('is_correct')->filter()->count();
        if ($isTrue != 1) {
            Notification::make()
                        ->title(__('You must select one answer as correct'))
                        ->danger()
                        ->send();

            $this->halt();
        }
    }
}
```


edit
---------------

this issue only happen while using `Repeater` with relationship

```php
Repeater::make('options')
              ->relationship()
              ->schema([
                  Grid::make()->schema([
                      TextInput::make('answer'),
                              
                      Toggle::make('is_correct')
                            ->translateLabel(),
                  ]),
              ])
```

danharrin:
Thats because repeater relationships are saved separately from the rest of the form, during the validation process. So you could try using beforeValidate() maybe for Edit

PR #4466, opened 2022-10-10
Add calculated values column to docs

electronick86:
Show the calculated values in the documentation. I don't know if it should reference this trick https://filamentphp.com/tricks/calculated-value-in-column


danharrin:
Thanks

jtomek:
Can you make this column searchable? I think it will throw an error unless you provide a calculated DB column, which makes this approach useless (for the searchable case).

electronick86:
In my opinion this column should not be searchable or sortable.

danharrin:
If you know what real DB columns to search / sort by, you can use `sortable(['one', 'two'])` or `searchable(['one', 'two'])`, which will use those columns for the DB query instead of your non-existant one.

PR #4470, opened 2022-10-11
Update chart-widget.blade.php

tommica:
Adds the pulse animation to the select dropdown to indicate that data loading is happening.


danharrin:
You can refactor this to use Livewire Loading States - https://laravel-livewire.com/docs/2.x/loading-states#toggling-classes

tommica:
@danharrin Wow, did not see that earlier, it is a really neat feature - I updated the code to use `wire:loading.class="animate-pulse"` instead

danharrin:
Thanks!

PR #4486, opened 2022-10-11
small avatar image, bg-primary background

egyjs:


zepfietje:
Hey @egyjs! I just saw the changes of this PR and found it doesn't include the color change for dark mode.

Also, I think it's not the best design choice to make the avatar so striking with the primary color.

We've reverted this PR, but we're open to discuss potential design changes related to the avatar for v3!

PR #4489, opened 2022-10-11
Revert "small avatar image, bg-primary background"

zepfietje:
Reverts filamentphp/filament#4486

The PR updated the avatar background color to a striking primary color.
In my opinion, this is bad design.

Also, the dark mode avatar color wasn't updated, which leads to inconsistency between dark and light mode.

I suggest to revert this change and discuss potential design changes related to the avatar for v3.


danharrin:
please note this on the original PR

PR #4495, opened 2022-10-12
Fix wrong method name used in docs

chengkangzai:
Hi filament teams,

just spot a wrong method name used in docs, it should be `reorderable()` instead of `orderable()`

PS: Love what filament team brings to the community, Thank you !


danharrin:
Thanks!

PR #4499, opened 2022-10-12
fix: only validate required keys in `getStateOnly()`

ryangjchandler:
Calling `getStateOnly()` on a component container still validates _all_ fields which isn't right. It should only validate the keys you're asking for.

There's now some duplication between `getStateOnly()` and `getState()` - not sure if we want to extract that into a separate method to be re-used. If we do, I don't know what it should be called.


danharrin:
I think that we could have `$only` and `$except` parameters for the main `getState()` method, and then just call that internally inside `getStateOnly()` and `getStateExcept()`

ryangjchandler:
Yeah, that could work.

Not sure how validation would work on the `except` though... I guess we'd need to do a `validateOnly()` and do something like `Arr::except(array_keys($this->getRawState()), $except)` or something...

danharrin:
Gonna close this as lots more work is required. For example, it needs to only save the relationships and call hooks of the specific fields, and also needs to work with `getStateExcept()`

PR #4514, opened 2022-10-13
Use Blade svg helper instead of components for icons

zepfietje:


packages/forms/resources/views/components/wizard.blade.php
```diff
@@ -132,17 +132,26 @@ class="w-10 h-10 flex items-center justify-center rounded-full"
...
-                                        x-cloak
-                                        x-bind:class="{
-                                            'text-gray-500 @if (config('forms.dark_mode')) dark:text-gray-400 @endif': getStepIndex(step) !== {{ $loop->index }},
-                                            'text-primary-500': getStepIndex(step) === {{ $loop->index }},
-                                        }"
-                                        class="w-5 h-5"
-                                    />
+                                @if ($icon = $step->getIcon())
+                                    @php
+                                        $inactiveIconClasses = \Illuminate\Support\Arr::toCssClasses([
```
zepfietje, line 140: This will be cleaned up once we remove the dark mode config check.

PR #4515, opened 2022-10-13
Add TextInput tel validation regex to config

oddvalue:
This PR adds the regex string in TextInput use for validating phone numbers to config so that it can be overridden.


zepfietje:
What about accepting the regex as an argument on the `tel()` method instead?

danharrin:
If the point of this PR is it override tel regex globally, it's probably best if you add a  `TextInput::telRegex(...)` static method and call it from a service provider instead...

PR #4516, opened 2022-10-13
Fix method name in edit documentation.

ghost:
Fixed the method name for the save notification.


danharrin:
We recently deprecated `getSavedNotificationMessage()` and replaced it with `getSavedNotificationTitle()`. Both still work, but you might need to `composer update` to get the new one.

ghost:
The latest tag does not have this deprecation yet: https://github.com/filamentphp/filament/blob/v2.16.31/packages/admin/src/Resources/Pages/EditRecord.php#L127

I see the new function in the main 2.x branch indeed.

danharrin:
ah ok 👌

PR #4517, opened 2022-10-13
Allow the tel regex to be overridden.

oddvalue:
Replaces: https://github.com/filamentphp/filament/pull/4515


danharrin:
Thanks!

PR #4526, opened 2022-10-13
Console UI improvement

shuvroroy:


danharrin:
This can be 2.x, right? Or is it L9 only?

shuvroroy:
> This can be 2.x, right? Or is it L9 only?

L9 only

PR #4556, opened 2022-10-15
Fix for BadgeColumn colours assigns colour to only last condition

naro-code-lab:
When different conditions are assigned to same colour, it assigns the colour only to last condition.

**Example **
```
colors([
  'warning' => 'processing',
  'warning' => 'scheduled',
  'warning' => 'escrowed',
])

 The warning colour will only be assigned to the last condition which is 'escrowed'.
```

This PR accepts assigning an array of conditions to a colour.
```
colors([
  'warning' => [
     'processing',
     'scheduled',
     'escrowed',
  ]
])
```

### DIFF
--- a/packages/tables/src/Columns/Concerns/HasColors.php
+++ b/packages/tables/src/Columns/Concerns/HasColors.php
@@ -28,6 +28,14 @@ trait HasColors
                     $stateColor = $condition;
                 } elseif ($condition instanceof Closure && $condition($state, $record)) {
                     $stateColor = $color;
+                } elseif (is_array($condition)) {
+                    $conditions = $condition;
+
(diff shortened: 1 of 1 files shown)


zepfietje:
Hey @naro-code-lab, thanks for creating a PR!

This would be a breaking change though, as currently the state is checked to equal the array you'd pass in.

Your desired behavior can be achieved by [passing a callback as a color's key value](https://filamentphp.com/docs/2.x/tables/columns/badge#customizing-the-color):
```php
BadgeColumn::make('status')
    ->colors([
        'warning' => static fn ($state): bool => in_array(
            $state,
            [
                'processing',
                'scheduled',
                'escrowed',
            ],
        ),
    ])
```

PR #4557, opened 2022-10-15
Allow controlling if a field is marked as required

zepfietje:
Sometimes it's better UX to not mark required fields with "*" and add an "optional" hint to non-required fields instead.

Default is still to mark required fields: https://www.nngroup.com/articles/required-fields/.

@danharrin, do you want me to add docs for this as well?


danharrin:
I feel like adding new attributes to the field wrapper is getting really messy and I don't see this as a priority feature. We should add this to v3 instead, where field wrappers have already been cleaned up.

Also yes, docs would be nice

zepfietje:
I don't think it's such a big issue to add this to v2 now and merge it into v3. But sure, I'll PR it to v3 instead.

zepfietje:
Done, @danharrin.

danharrin:
All instances should be `MarkedAsRequired` instead of `MarkedRequired` I think, with the setter being `markAsRequired()`

zepfietje:
Agree, was in doubt about that.

danharrin:
In fact, `shouldBe` -> `is`

zepfietje:
Ah, good one!

zepfietje:
This should be good to go now, @danharrin.

PR #4562, opened 2022-10-16
fix: Hide chart widget divider if it has no header

AAbosham:
Hide `<x-filament::hr />`  from chart-widget header if not set header or filters.


zepfietje (approved):

danharrin:
Thanks

PR #4568, opened 2022-10-17
Replaced load with loadMissing

iotron:
for multiple components, the load was called multiple times. Now, the $record load will be called once for any number of components on a page.


danharrin:
This has been reverted since the full load is deliberate, otherwise newly saved records are not fetched from the database

PR #4569, opened 2022-10-17
Replaced load with loadMissing to reduce queries

iotron:
Fixed tags relation loading multiple times for each component.


danharrin:
This has been reverted since the full load is deliberate, otherwise newly saved records are not fetched from the database

PR #4581, opened 2022-10-17
add new DE translations

RickDBCN:


zepfietje:
Thanks!

packages/forms/resources/lang/de/components.php
```diff
@@ -16,6 +20,10 @@
                 'label' => 'Löschen',
             ],
 
+            'move_item' => [
+                'label' => 'Verschieben',
+            ],
+
             'move_item_down' => [
                 'label' => 'Runter verschieben',
```
dissto, line 28: @zepfietje 

That is not a proper German translation. If anything, that should be "Nach unten verschieben".

packages/forms/resources/lang/de/components.php
```diff
@@ -204,6 +216,8 @@
 
         'loading_message' => 'Lädt...',
 
+        'max_items_message' => 'Es können nur :count ausgewählt werden.',
+
         'no_search_results_message' => 'Keine Optionen passen zu Ihrer Suche.',
```
dissto, line 221: This also sounds weird. This should be something like: "Die Suche ergab keine Treffer" or "Die Suche lieferte keine Ergebnisse". 🤔

packages/forms/resources/lang/de/components.php
zepfietje: Could you update these strings in a pull request?

RickDBCN:
Hi @dissto,
I fixed these in #4592 

thanks! :)

PR #4604, opened 2022-10-18
Fix double return typo in navigation docs

cheesegrits:
Fix for a simple typo in the admin panel navigation doc, a double return on the getNavigationGroup() method example.


zepfietje:
Thanks!

PR #4610, opened 2022-10-19
Clean up actions

zepfietje:
Fixes use of `config()` inside the actions package to use the package's config.

Other note, there's a reference of `filament/filament` config here: https://github.com/filamentphp/filament/blob/03e4d84063b26cd913563d1a10f362070dc1196d/packages/actions/resources/views/components/modal/actions.blade.php#L3
~~Shouldn't that be moved to the actions package config instead? Or probably even better, make it configurable on the action so we can move to `configureUsing()` instead of using the config.~~
I'm not sure how I feel about having that view inside the actions package when it's clearly used for `filament/filament`.

Also, what should the dusk selector be here?
https://github.com/filamentphp/filament/blob/8ef19eb5249a7e6f13067a45cb5cdfdc77e1b341/packages/forms/resources/views/components/actions/action.blade.php#L26

### DIFF
--- a/packages/actions/resources/views/button-action.blade.php
+++ b/packages/actions/resources/views/button-action.blade.php
@@ -3,7 +3,7 @@
     component="filament-actions::button"
     :outlined="$isOutlined()"
     :icon-position="$getIconPosition()"
-    class="filament-page-button-action"
+    class="filament-actions-button-action"
 >
     {{ $getLabel() }}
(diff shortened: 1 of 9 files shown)


danharrin:
> Other note, there's a reference of filament/filament config here

When we go through the config & clean it up, I'll sort this out

PR #4635, opened 2022-10-19
Added admin panel vs livewire docs for filter layout

cheesegrits:


zepfietje:
The admin panel docs for this feature already exist at https://filamentphp.com/docs/2.x/admin/resources/listing-records#displaying-filters-above-or-below-the-table-content.

Let me know if anything's missing. 🙂

cheesegrits:
@zepfietje yes, but that's the Admin Panel Resources manager docs. The docs I modified are the Table Builder docs, where the convention is to show you how to do things in a Table Builder either in the admin panel or stand alone.  As-is, those Filter layout docs only show how to do it standalone.

There are dozens of other examples in the Table Builder docs where it gives both ("If you are doing this in a Livewire component ... or if you are doing this in the Admin Panel ..."), for example at the top of the page I modified.

When referencing the docs, it really helps to have both flavors of code show in the standalone Form or Table docs, even if it duplicates some documentation in the Admin Panel docs.

I gather V3 will get rid of this, and it'll all become Admin Panel Flavored, which will make life a lot easier.  :)

zepfietje:
Personally I prefer linking to the admin panel docs from the table docs then. It's really hard to maintain the docs when features are documented in 3 places.

Could you show me where it's duplicated like you mentioned? Maybe we could add a note to the table docs, just unsure it should be duplicated like this.

And yes, I agree searching the docs across packages is hard sometimes. The unified API in v3 will make this a lot better.

PR #4659, opened 2022-10-20
Remove dark mode config

### LINKED ISSUE #4657: Remove dark mode config
In favor of Tailwind `darkMode: 'class'`.

zepfietje:


packages/support/resources/views/components/modal/actions.blade.php
```diff
@@ -1,6 +1,5 @@
 @props([
-    'alignment' => 'left',
-    'darkMode' => false,
+    'alignment' => config('filament.layout.actions.modal.actions.alignment') ?? config('filament-forms.components.actions.modal.actions.alignment') ?? config('filament-tables.layout.actions.modal.actions.alignment') ?? 'left',
```
zepfietje, line 2: We'll need to find a way for users to configure the support components instead of using the package ones. The config for these components should be the same across all packages anyway.

This is just ugly WIP code for now so I could at least move to using the support component in all places.

packages/app/docs/08-appearance.md
```diff
@@ -14,13 +14,7 @@ You may create a `resources/views/vendor/filament/components/brand.blade.php` fi
 
 ## Dark mode
 
-By default, Filament only includes a light theme. However, you may allow the user to switch to dark mode if they wish, using the `dark_mode` setting of the [configuration file](installation#publishing-configuration):
```
danharrin: This section needs reverting

packages/app/config/filament.php
```diff
@@ -160,16 +160,16 @@
     ],
 
     /*
-    |--------------------------------------------------------------------------
```
danharrin: spacing changes here?

PR #4661, opened 2022-10-21
Improve accessibility around the user menu avatar

marcusmoore:
This PR improves accessibility for the user menu in the top bar by wrapping the trigger, the user avatar, in a button that is labeled "user menu". This allows the user to tab to the user avatar and hit enter or space to open it, as well as closing it with enter or space. In addition, using an `aria-label` on the button allows screen readers to explain that the button is for the user menu (tested with macOS VoiceOver).

A nice bonus to making the trigger an actual button is that plugins like [vimium](https://github.com/philc/vimium/) can open it from the keyboard.


danharrin:
Hey Marcus! Appreciate the PR. Before this is merged, I will need to do some work to add a translation string for this new label though, unless you can.

marcusmoore:
Ah, I forgot that step @danharrin! I should be able to handle adding translation strings early next week 👍🏾

marcusmoore:
@danharrin I added translation strings that were run through Google Translate so I can't vouch for how accurate they are. Is that helpful or should I only contribute the ones that I know for sure are accurate?

danharrin:
Hey Marcus! Wow, thank you for so much effort. In the future, don't worry about Google Translate for languages you don't speak. We actually have a set of native-speaking translation managers who get notified when we add a new language string that needs translating. So new PRs only need the translation string added for English :)

Many thanks for your work though

marcusmoore:
@danharrin good to know! Thank you!

PR #4669, opened 2022-10-21
add pagination overview to mobile

cobraprojects:
add pagination overview on top of pagination

image

### DIFF
--- a/packages/tables/resources/views/components/pagination/index.blade.php
+++ b/packages/tables/resources/views/components/pagination/index.blade.php
@@ -29,7 +29,7 @@
             @endif
         </div>
 
-        <div class="flex justify-between items-center flex-1">
+        <div class="flex justify-between items-center flex-1 w-full">
             <div class="w-10">
                 @if ($paginator->hasPages() && (! $paginator->onFirstPage()))
                     <x-tables::icon-button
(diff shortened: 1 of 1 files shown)


zepfietje:
From a design perspective I think we shouldn't try to fit in the count. Feels really crowded. Also, look at how Tailwind UI handles it on mobile (they hide the count): https://tailwindui.com/components/application-ui/navigation/pagination.

I do sometimes find myself wanting to view the count on mobile though. Any other ideas?

cobraprojects:
Can we make it optional?
like we have
```php
protected function isTablePaginationEnabled(): bool 
{
    return false;
} 
```


we could do somthing like

```php
protected function showCountOnMobile(): bool 
{
    return true;
} 
```

we can make the default value `false`

or may be an option in the config file

zepfietje:
We could potentially make it opt-in yeah, but then we'd still need to find a proper design.

zepfietje:
@danharrin @awcodes, what do you think? There's no matching Tailwind UI design for this and I feel like it gets too crowded on mobile.

danharrin:
Yeah honestly I don't like it but I feel bad closing PRs lol

zepfietje:
There's no sense in keeping it open then. I can close it for you. 😅

No but seriously, I feel like it's probably better to hold off on this for now.

danharrin:
Thanks for making the decision for me. And thanks for the PR @cobraprojects

cobraprojects:
No worries, we all looking for the best

PR #4672, opened 2022-10-22
Show a failure message if (soft)delete / restore isn't successful

underdpt:
Hi,

When soft-deleting, deleting or restoring a record, a success notification is shown even if the action wasn't successful. This PR checks the model action result and shows a failure if it isn't successful.

### DIFF
--- a/packages/admin/src/Pages/Actions/DeleteAction.php
+++ b/packages/admin/src/Pages/Actions/DeleteAction.php
@@ -43,9 +43,13 @@ class DeleteAction extends Action
         });
 
         $this->action(function (): void {
-            $this->process(static fn (Model $record) => $record->delete());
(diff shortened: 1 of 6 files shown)


zepfietje:
Thanks for the PR, @underdpt!

I think table bulk actions should also be handled?

underdpt:
Yeah, I was thinking that too, but that has to be handled on a different way (by comparing the number of selected records vs the number of deleted / restored records) and also the failure message has to be more meaningful ("There was X records to delete, but some of them couldn't be processed successfully" or something similar). So I thought that should go in a different PR if this was accepted.

danharrin:
This is fine in it's current state. It's difficult to handle bulk actions where some records are deleted and others aren't, as two notifications with conflicting messages would be displayed which could be confusing to users.

PR #4675, opened 2022-10-22
Fix dropdown item label from truncating too early

ricki-io:
I saw that the items on the profile dropdown menu are being truncated even they still have space to grow.

Screenshot 2022-10-22 at 17 19 41

I made a quick fix for that. But I would recommend someone with more experience with this codebase, to confirm that this change is not breaking other dropdown menus.

Thanks.


danharrin:
thanks

PR #4681, opened 2022-10-22
Fix dark mode layout for sorting in table layout

StfnSl:
Fixed display bug for dark mode.
Missing German translation added for sorting

### DIFF
--- a/packages/tables/resources/lang/de/table.php
+++ b/packages/tables/resources/lang/de/table.php
@@ -148,4 +148,27 @@ return [
 
     ],
 
+    'sorting' => [
+
+        'fields' => [
+
+            'column' => [
+                'label' => 'Sortieren nach',
+            ],
+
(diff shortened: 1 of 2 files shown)


danharrin:
Please send before / after screenshots of these changes

StfnSl:
> Please send before / after screenshots of these changes

Sorry for the late reply, here are the screenshots

before:
Screenshot_20221025_194313

after:
Screenshot_20221025_194520

pxlrbt:
> Fixed display bug for dark mode.

Why does this change light mode styling then?

StfnSl:
I used the same css classes as when displaying the rows.

danharrin:
How about the direction select?

StfnSl:
> How about the direction select?

Yes, I had overlooked that, but it has now been made up and provided with the same css classes

PR #4687, opened 2022-10-23
Create custom type for callbacks

NathanaelGT:
This PR is a implementation from this proposal https://github.com/filamentphp/filament/discussions/2639#discussioncomment-3104926


packages/forms/src/Callbacks/Callback.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Filament\Forms\Callbacks;
```
danharrin: Potentially we should just use `Filament\Forms` instead. Thoughts?

packages/forms/src/Callbacks/Callback.php
```diff
@@ -0,0 +1,13 @@
...
+
+use Filament\Forms\Components\Component;
+
+abstract class Callback
+{
+    public function __construct(
```
danharrin: Also I think this Callback class abstraction is too far, we can just put this constructor within Get and Set individually

packages/forms/src/Callbacks/Callback.php
NathanaelGT: Honestly I'm not sure about the best namespace for this

danharrin:
Thanks

PR #4690, opened 2022-10-23
Only render visible widgets

### LINKED ISSUE #4678: Hidden widgets still take up visible space
### Package

filament/filament

### Package Version

v2.16.36

### Laravel Version

v9.36.4

### Livewire Version

v2.10.7

### PHP Version

8.1.10

### Problem description

When a widget is set in the `getHeaderWidgets` array, but the widget's `canView` method returns false, the `filament-widgets-container` component is still present in the DOM, while it is empty, it still takes up visible vertical space, leading to an extra gap due to the grid layout.

Here is a side by side comparison
image

### Expected behavior

The empty container should not have been added, or it should have been added in a way that does not increase the height.

### 1: Clone repo

```
git clone git@github.com:caendesilva/filament-bug-repro.git
cd filament-bug-repro
composer install
php artisan migrate --seed
```

### 2: Log in and view demo page

Visit http://127.0.0.1:8000/admin/, log in with email `test@example.com` and password `password`

Visit Filament page "[broken](http://127.0.0.1:8000/admin/broken)" and visually compare it to "[expected](http://127.0.0.1:8000/admin/expected)". The content is further down on the broken page, leading to a visual inconsistency. The broken page has a widget that is hidden, whereas the expected page has no widgets.

### Quick links to reproduction files

See this commit for a quick overview https://github.com/caendesilva/filament-bug-repro/commit/028f646882f89c753db4c2d4e7f47628d8bdd1cb

### Reproduction repository

### Relevant log output

_No response_

zepfietje:


emmadesilva (approved):
Changes applied to the reproduction repository. Confirmed to be working perfectly!

packages/admin/src/Pages/Page.php
```diff
@@ -174,6 +174,11 @@ protected function getHeaderWidgets(): array
         return [];
     }
 
+    protected function getVisibleHeaderWidgets(): array
+    {
+        return static::getVisibleWidgets($this->getFooterWidgets());
```
maartenpaauw: shouldn't `getFooterWidgets` be `getHeaderWidgets`?

emmadesilva: Without having written the code, I would guess so. I don't think there is any reason to add complexity to also add extra checks for footer widgets, as the empty space won't be noticed there.

maartenpaauw: I just mean; the `getVisibleHeaderWidgets` method is returning the filtered footer widgets.

emmadesilva: Whoops, my mind is half-focused on some other code, sorry. Yeah, you're right, while the fix works I don't think this should return `getFooterWidgets()` either.

zepfietje: Good catch @maartenpaauw. Copy-paste issue. 😅

packages/admin/src/Pages/Page.php
```diff
@@ -184,6 +189,16 @@ protected function getFooterWidgets(): array
         return [];
     }
 
+    protected function getVisibleFooterWidgets(): array
+    {
+        return static::getVisibleWidgets($this->getFooterWidgets());
+    }
+
+    protected static function getVisibleWidgets(array $widgets): array
```
danharrin:
```suggestion
    protected function filterVisibleWidgets(array $widgets): array
```

danharrin (changes requested):

packages/admin/src/Pages/Page.php
zepfietje: Thanks, the `get` naming already felt out of place to me.

emmadesilva (approved):
New naming looks great

PR #4708, opened 2022-10-24
Docs: Fix broken link on Tables: Columns: Getting Started page

intrepidws:


danharrin:
How about ../actions

intrepidws:
Thanks for the note. I was wondering if that would work, but didn't know for sure. That's done.

PR #4742, opened 2022-10-28
Add wire target to login button

howdu:


pxlrbt:
Can you explain what this PR improves?

howdu:
Loading state on login button

zepfietje:
This PR is not needed as the loading indicator target is already grabbed from the `form` attribute of the button.

PR #4743, opened 2022-10-28
add on chart click event

yemenpoint:
i added on chart click event to chart widget

### DIFF
--- a/packages/admin/resources/views/widgets/chart-widget.blade.php
+++ b/packages/admin/resources/views/widgets/chart-widget.blade.php
@@ -1,6 +1,7 @@
 @php
     $heading = $this->getHeading();
     $filters = $this->getFilters();
+    $chart_name  = \Illuminate\Support\Str::slug($heading, "_");
 @endphp
 
 <x-filament::widget class="filament-widgets-chart-widget">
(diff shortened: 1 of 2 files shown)


zepfietje:
It seems like this can be achieved using the current API with `getOptions()` already?

yemenpoint:
> It seems like this can be achieved using the current API with `getOptions()` already?

no . u must add the script

zepfietje:
At least it should be refactored to also work when using `getOptions()` then.

yemenpoint:
> At least it should be refactored to also work when using `getOptions()` then.
what is your suggestion ! 
1- should i add it while getting options then options will never be null ?
I suggest  leave it like this 
if you customize options you add 
```
[
        "onClick" => \Illuminate\Support\Str::slug($this->getHeading(), "_")."clickEvent"
]
```
to options? what you think?

yemenpoint:
> At least it should be refactored to also work when using `getOptions()` then.

onClick take function name it cant be string 
```
{'onClick' : function_name} this works 
{'onClick' : "function_name"} this doesnt work
```
and json_encode is encoding function name to string 

so adding this will not solve the problem 
```
data_set(static::$options, "onClick", \Illuminate\Support\Str::slug($this->getHeading(), "_") . "clickEvent");
```

###  **what is your suggestion ??? **

danharrin:
Since the `chart-widget.blade.php` file is only 100 lines long, you could always copy it into your own blade file and then set the widget's `$view` to be the new view path. Then you can add the function yourself to the file without it being in core :)

I don't want to merge this PR because it does introduce quite alot of technical debt and inconsistency with how we handle things in other parts of Filament.

Maybe we will find a better solution in the future if you explain why you need the click event?

Issue #4748, opened 2022-10-29
Component wildcard and subdirectories


Ashk2a:
### Package

filament/filament

### Package Version

v2.16.36

### Laravel Version

v9.37.0

### Livewire Version

v2.10.7

### PHP Version

PHP 8.1.10

### Problem description

I see [this recent PR](https://github.com/filamentphp/filament/pull/3512#issue-1339467293), it's working well except when components (resources/widgets/pages) are in a subdirectory.

### Expected behavior

Same behaviour even if components are in a subdirectory or not.

### Steps to reproduce

**config/filament.php**
```php
    'pages' => [
        'namespace' => 'App\\Modules\\*\\Filament\\Pages',
        'path' => app_path('Modules/*/Filament/Pages'),
        'register' => [
            Pages\Dashboard::class,
        ],
    ],

    /*
    |--------------------------------------------------------------------------
    | Resources
    |--------------------------------------------------------------------------
    |
    | This is the namespace and directory that Filament will automatically
    | register resources from. You may also register resources here.
    |
    */

    'resources' => [
        'namespace' => 'App\\Modules\\*\\Filament\\Resources',
        'path' => app_path('Modules/*/Filament/Resources'),
        'register' => [],
    ],

    /*
    |--------------------------------------------------------------------------
    | Widgets
    |--------------------------------------------------------------------------
    |
    | This is the namespace and directory that Filament will automatically
    | register dashboard widgets from. You may also register widgets here.
    |
    */

    'widgets' => [
        'namespace' => 'App\\Modules\\*\\Filament\\Widgets',
        'path' => app_path('Modules/*/Filament/Widgets'),
        'register' => [
            Widgets\AccountWidget::class,
            Widgets\FilamentInfoWidget::class,
        ],
    ],
```

Working : `app/Modules/User/Filament/Resources/UserResource.php`
Not working : `app/Modules/User/Filament/Resources/Admin/UserResource.php`

### Reproduction repository

https://github.com/Ashk2a/laravel-dev

### Relevant log output

```shell
FilamentServiceProvider::registerComponentsFromDirectory

L229 : the value of $variableNamespace is wrong when component is in a subdirectory. I tried to fix it but I didn't found a solution for the moment.

Some variable dump :

[
  "namespace" => "App\Modules\*\Filament\Resources"
  "file" => "/var/www/app/app/Modules/User/Filament/Resources/Admin/UserResource.php"
  "variableNamespace" => "User\Admin"
[...]

PR #4749, opened 2022-10-29
Fix components registration in subdirectory when using wildcard

Ashk2a:
When you have a wildcard configuration (for Resources/Pages/Widgets) and your components are in subdirectory, `$variableNamespace` returns a bad value and the components are not loaded.

The fix is not very elegant, I haven't found a simpler solution than this.

[Related issue](https://github.com/filamentphp/filament/issues/4748)


danharrin:
Thanks

PR #4750, opened 2022-10-29
Fix: Add id for dom diff

### LINKED ISSUE #4680: Select all button still shows after selecting when filtered
### Package

filament/tables

### Package Version

2.16.36

### Laravel Version

9.36.4

### Livewire Version

2.10.6

### PHP Version

8.1.9

### Problem description

When select all searching and filtering result don't disappear Select All action

image

### Expected behavior

search or filter table then select all items when click on `Select All #` don't disappear .

### Steps to reproduce

Demo app tables for ex: https://demo.filamentphp.com/shop/products

### Reproduction repository

### Relevant log output

_No response_

awcodes:
Adds an id to force a dom update when $allRecordsCount changes. This is a know issue with x-show and livewire variables.

#4680

### DIFF
--- a/packages/tables/resources/views/components/selection-indicator.blade.php
+++ b/packages/tables/resources/views/components/selection-indicator.blade.php
@@ -16,7 +16,7 @@
 
     <span @class(['dark:text-white' => config('tables.dark_mode')]) x-text="window.pluralize(@js(__('tables::table.selection_indicator.selected_count')), selectedRecords.length, { count: selectedRecords.length })"></span>
 
-    <span id="record-count-{{ $allRecordsCount }}" x-show="{{ $allRecordsCount }} !== selectedRecords.length">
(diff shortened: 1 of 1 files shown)


zepfietje:
Shouldn't `wire:key` be used instead?

awcodes:
This is an issue with alpine not livewire. A wire key won't force the the span to change the variable inside x-show.

awcodes:
Just ran a test, after Dan told me he prefers wire:key, and wire:key had no effect on this when using a blade variable inside an x-show directive.

Seems like setting an id is the only way to get this to update properly in alpine.

danharrin:
Okay understood, but could you please make sure the ID follows the same structure as other wire:keys on the page? Because also it needs to be unique for multiple tables on the same page

awcodes:
Done. I think that's correct. Let me know if it's not.

danharrin:
Thanks!

PR #4756, opened 2022-10-30
Add square column option

leandrocfe:


zepfietje:
@leandrocfe, isn't it already square by default?

leandrocfe:
> @leandrocfe, isn't it already square by default?
Hi @zepfietje ! Take a look:

Normal:
image

Square:
image

zepfietje:
Would you use both in the same project?

I feel like rounding the square image borders is an opinionated style you could add using CSS. After all, square is the default already. Thoughts?

leandrocfe:
> Would you use both in the same project?
> 
> I feel like rounding the square image borders is an opinionated style you could add using CSS. After all, square is the default already. Thoughts?

@zepfietje The idea would be a new style that you can apply easily. But I agree with you.

zepfietje:
Yeah I get the goal, but would you use both in the same project, @leandrocfe?

If not, I don't see why we should add this as it's basically just an opinionated style for the square image.

zepfietje:
@leandrocfe, I misunderstood this PR, sorry.
Thought the image is square already, but it's actually rectangular (not necessarily square).

We can add `square()` but it shouldn't have rounded corners.

packages/tables/src/Columns/ImageColumn.php
```diff
@@ -153,9 +170,14 @@ public function getWidth(): ?string
         return $width;
     }
 
-    public function isRounded(): bool
```
danharrin, line 156: Please reintroduce this method & deprecate. Call `$this->isCircular()` internally. Also update docs if there are any?

leandrocfe:
> @leandrocfe, I misunderstood this PR, sorry. Thought the image is square already, but it's actually rectangular (not necessarily square).
> 
> We can add `square()` but it shouldn't have rounded corners.

@zepfietje It makes sense. Do you think that shadow and rounded can be options as well?

feature

And the user can control other attributes with extra attributes... What is your opinion? I'm not sure...

zepfietje:
I think shadow and rounding are too opinionated and specific to have dedicated properties for. We could add `filament-tables-image-column-circular` and `filament-tables-image-column-square` CSS class hooks to make it easier to style though.

leandrocfe:
> I think shadow and rounding are too opinionated and specific to have dedicated properties for. We could add `filament-tables-image-column-circular` and `filament-tables-image-column-square` CSS class hooks to make it easier to style though.

Nice! I agree 👍

zepfietje:
@danharrin this should be ready to go! 🚀

danharrin:
Thank you very much @leandrocfe @zepfietje

PR #4764, opened 2022-10-31
Normalize wire:keys in table mount action forms

awcodes:
Remove record based identifier from wire key in mounted table actions


danharrin:
I can't see any immediate problems with this... lets see if it breaks anything

PR #4782, opened 2022-11-01
[2.x] feature: allow ignoring policies in authorization checks

ryangjchandler:
Introduces a couple of new methods `Resource::ignorePolicies(bool $condition = true)` and `Resource::shouldIgnorePolicies()`.

Right now, Filament is auto-wired to check policies by default and the only way to change that behaviour is to override the `Resource::can()` method.

That process is a little tedious on each resource, even if you abstract that logic into a trait. This PR makes it as simple as calling `MyResource::ignorePolicies()` in a service provider, you could even do it for all resources by calling it on the base `Resource` class instead.

There are valid usecases for this, e.g. your policies are all wired up for the front-end and you're installing Filament into an existing application (my personal use case), but you don't want to modify your existing policies.

### DIFF
--- a/packages/admin/src/Resources/RelationManagers/RelationManager.php
+++ b/packages/admin/src/Resources/RelationManagers/RelationManager.php
@@ -52,6 +52,8 @@ class RelationManager extends Component implements Tables\Contracts\HasRelations
 
     protected static ?string $title = null;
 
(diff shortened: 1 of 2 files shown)


danharrin:
Is it worth adding this for RMs too?

ryangjchandler:
Yeah good point!

PR #4784, opened 2022-11-01
Added $default to askRequired in CanValidateInput

cheesegrits:
I'm using this in a package with some fairly complex commands, and it just makes for a much better experience if I can let them call the command with no args, then walk them through each one with sane defaults on the ask() prompts.

Couldn't find a test that covered this, and not experienced enough with tests to write one from scratch.  But tested by hand with and without a default on askRequired() and it worked fine.  If there is an existing test, point me at it and I'll extend it to cover this.

### DIFF
--- a/packages/support/src/Commands/Concerns/CanValidateInput.php
+++ b/packages/support/src/Commands/Concerns/CanValidateInput.php
@@ -7,9 +7,9 @@ use Illuminate\Support\Facades\Validator;
 
 trait CanValidateInput
 {
-    protected function askRequired(string $question, string $field): string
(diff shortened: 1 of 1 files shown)


packages/support/src/Commands/Concerns/CanValidateInput.php
```diff
@@ -7,9 +7,9 @@
 
 trait CanValidateInput
 {
-    protected function askRequired(string $question, string $field): string
+    protected function askRequired(string $question, string $field, string $default = ''): string
```
danharrin:
```suggestion
    protected function askRequired(string $question, string $field, ?string $default = null): string
```

PR #4786, opened 2022-11-01
Wizard: Update query string based on step

SquidwardHill:
Boolean property on the Wizard Component to set the query string based on the current step.


vivere-dally:
It would also be great to have a method exposed in order to register a callback when to clear the query string.

PR #4787, opened 2022-11-01
Allow resource manager buttons to be navigable via keyboard

marcusmoore:
This PR allows the buttons that are clicked to display relationships on view resource pages to be navigated to via the keyboard.

View resource page

---

Currently, the button that represents the _selected_ relationship can be navigated by keyboard but the other buttons are not navigable so they cannot be selected using only keyboard shortcuts (meaning hitting tab multiple times until the button is highlighted and then hitting space or enter)

https://user-images.githubusercontent.com/1141514/199355375-d8b8f710-39e7-4bc9-a717-9e181fb531d8.mp4

With this PR inactive buttons are navigable and selectable via the keyboard (using space or enter)

https://user-images.githubusercontent.com/1141514/199355380-df5d0ce4-3971-47f6-a957-827d48623342.mp4

---

I believe this change helps screen readers as well. When I was testing the page a couple weeks ago with macOS' VoiceOver feature, it announced both buttons ("Pages" and "Sites" in this case) as selected which obviously is not the case. Unfortunately, I can't provide more details now because I accidentally changed some VoiceOver settings and it isn't reading the selected status for anything right now 😅

### DIFF
--- a/packages/admin/resources/views/components/resources/relation-managers/index.blade.php
+++ b/packages/admin/resources/views/components/resources/relation-managers/index.blade.php
@@ -29,9 +29,6 @@
                         wire:click="$set('activeRelationManager', {{ filled($tabKey) ? "'{$tabKey}'" : 'null' }})"
                         @if ($activeManager === $tabKey)
                             aria-selected
-                            tabindex="0"
-                        @else
-                            tabindex="-1"
                         @endif
                         role="tab"
                         type="button"


zepfietje:
Thanks for the PR, @marcusmoore!
The change looks good to me.

Just to be sure, is there any particular reason why non-active tabs had `tabindex="-1"` previously, @danharrin?

zepfietje (approved):

marcusmoore:
Thanks for the review @zepfietje 😄 

I wondered if I was missing _why_ `-1` was used but I think this is an ok change. Thanks for asking the clarifying question.

danharrin:
Yeah I don't know either 🤷 Thanks @marcusmoore

PR #4788, opened 2022-11-01
Wizard: Expose step value to querystring

SquidwardHill:
Signed-off-by: Sydney Hill <sydneyehill@gmail.com>


packages/forms/src/Components/Wizard.php
```diff
@@ -16,10 +16,14 @@ class Wizard extends Component
 
     protected bool | Closure $skippable = false;
 
+    protected bool | Closure $stepAsQuerystring = false;
+
     protected string | Htmlable | null $submitAction = null;
 
     public int | Closure $startStep = 1;
 
+    public ?int $currentStep = null;
```
danharrin: I think this is not being used?

packages/forms/src/Components/Wizard.php
```diff
@@ -16,10 +16,14 @@ class Wizard extends Component
 
     protected bool | Closure $skippable = false;
 
+    protected bool | Closure $stepAsQuerystring = false;
```
danharrin:
```suggestion
    protected bool | Closure $isStepInQueryString = false;
```

packages/forms/src/Components/Wizard.php
```diff
@@ -100,6 +104,13 @@ public function skippable(bool | Closure $condition = true): static
         return $this;
     }
 
+    public function setStepAsQuerystring(bool | Closure $condition = true): static
```
danharrin:
```suggestion
    public function queryStringStep(bool | Closure $condition = true): static
```

packages/forms/src/Components/Wizard.php
```diff
@@ -119,4 +130,9 @@ public function isSkippable(): bool
     {
         return $this->evaluate($this->skippable);
     }
+
+    public function isStepAsQuerystring(): bool
```
danharrin:
```suggestion
    public function isStepInQueryString(): bool
```

danharrin (changes requested):
You might need to rename other method and property references to match these

danharrin:
@zepfietje @awcodes wdyt about naming that I proposed? it's difficult

awcodes:
I think they are ok.

zepfietje:
I'd prefer something like `storeStepInQueryString()` with `shouldStoreStepInQueryString()`. Quite lengthy though.

Also, I'm wondering what's the purpose of storing the step in the query string if the step is not restored from the query string when loading the page anyway?

danharrin:
How about `persistStepInQueryString()`? And yes, it should be restored on load

zepfietje:
I like that. Think it matches the naming around table filters.

danharrin:
Okay, lets use that naming then @SquidwardHill 

`persistStepInQueryString()`
`$isStepPersistedInQueryString` / `isStepPersistedInQueryString()`

zepfietje:
`isPersisted` or `shouldPersist`, @danharrin?

danharrin:
I try and use `is` wherever it makes grammatical sense. `should` is a last resort

zepfietje:
Makes sense! Still waiting for your method naming docs. 😉

lhilton:
> Also, I'm wondering what's the purpose of storing the step in the query string if the step is not restored from the query string when loading the page anyway?

Speaking to this question, I can add some context: When we developed this PR, the decision was to make the minimal filament-side change possible and let the handling of that be on the consumer side. In other words, leave it to the end forms to determine the right way to respond to that querystring. I can see a case for going the other route, however this would require more substantial design in order to both support the change and allow developer overrides of that behavior.

zepfietje:
Thanks for clarifying, @lhilton! I think it's important to add the full functionality to Filament if we're going to add it.

lhilton:
@danharrin @zepfietje I'll coordinate with @SquidwardHill to move our restoration logic in as part of this PR then. Thank you for the feedback!

SquidwardHill:
Circling back to this on Monday.

zepfietje:
Any update on this PR @lhilton @SquidwardHill?

lhilton:
We have updated this PR with the requested changes (name changes) and we moved the logic to automatically restore step state if the feature is in use.

Additionally we found a need to modify the querystring key in use. This was previously hardcoded to `step`. It will now default to the value `step` but can be overridden when using the `persistStepInQueryString()` function as an optional second parameter.

danharrin:
Thanks!

PR #4792, opened 2022-11-02
Feature: Add expand/collapse all events to collapsible rows

oddvalue:
This PR adds some events to the collapsible table rows to allow implementing an expand/collapse
 all button via an action.

### DIFF
--- a/packages/tables/resources/lang/en/table.php
+++ b/packages/tables/resources/lang/en/table.php
@@ -87,6 +87,14 @@ return [
             'label' => 'Toggle columns',
         ],
 
+        'expand_all' => [
+            'label' => 'Expand all',
+        ],
+
+        'collapse_all' => [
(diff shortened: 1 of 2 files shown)


danharrin:
Thank you for the work, but I don't think this feature is going to be used by enough people to warrant the extra code complexity, and it doesn't really fit within our existing table layout, especially on mobile.

oddvalue:
> Thank you for the work, but I don't think this feature is going to be used by enough people to warrant the extra code complexity, and it doesn't really fit within our existing table layout, especially on mobile.

Completely understand that. Could I persuade you to at least include the events on the collapsible rows so that it could be implemented with an action?

danharrin:
Yeah sure, you can add event listeners if you wish

Edit this PR to remove everything else then mark it as ready for review

danharrin:
Thanks

PR #4795, opened 2022-11-02
Add aria label to sidebar expand and collapse buttons

marcusmoore:
This PR adds an aria-label to the sidebar expand/collapse button. 

Since the buttons is an SVG, screen readers are unable to tell users what the button does. With this change, screen readers (at least macOS' VoiceOver) will announce "expand sidebar" or "collapse sidebar" depending on the status of the side bar.

Current warning in Brave's Lighthouse tool:
Lighthouse warning showing button does not have an accessible name

### DIFF
--- a/packages/admin/resources/lang/en/layout.php
+++ b/packages/admin/resources/lang/en/layout.php
@@ -22,9 +22,11 @@ return [
             'label' => 'Sign out',
         ],
 
-        'sidebar' => [
-            'collapse' => 'Collapse sidebar',
-            'expand' => 'Expand sidebar',
+        'toggle_sidebar' => [
+            'label' => [
+                'collapse' => 'Collapse sidebar',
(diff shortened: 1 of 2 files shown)


danharrin:
Hey! Can we have that as a translation please in the `layout.php` file? Just English 😉

marcusmoore:
🤦🏾 as I hit submit I thought to myself "am I forgetting something?" and completely spaced on the translation. I'll add it in a bit!

danharrin:
Thanks mate!

zepfietje:
Shouldn't a similar label be added in `topbar.blade.php` for mobile?

danharrin:
Oh yeah, probably

zepfietje:
There you go: https://github.com/filamentphp/filament/pull/4803.

PR #4802, opened 2022-11-03
Use x syntax for Alpine

zepfietje:
@danharrin you must've missed this in https://github.com/filamentphp/filament/pull/4795.


marcusmoore:
Good catch. Thanks @zepfietje 😄

PR #4803, opened 2022-11-03
Add toggle sidebar label on mobile

zepfietje:
Not sure if we just use the same translations here or need to add new ones for open and close instead of expand and collapse. @danharrin, thoughts?


danharrin:
I think this is fine

PR #4808, opened 2022-11-03
Secondary color

danharrin:


packages/actions/src/Concerns/CanOpenModal.php
```diff
@@ -153,7 +153,7 @@ public function getModalSubmitAction(): ModalAction
             ->label($this->getModalButtonLabel())
             ->submit($this->getLivewireCallActionName())
             ->color(match ($color = $this->getColor()) {
-                'secondary' => 'primary',
```
zepfietje, line 156: What's the logic behind this, @danharrin?

danharrin: Actions that open modals have a submit button. That submit button usually matches the color of the action. Apart from gray actions, which have a primary submit button

PR #4809, opened 2022-11-03
Support Column text alignment of the text-start, text-end and text-justify

AAbosham:
We already can alignment text as left or right but for the multi-language with different direction it will be need condition like `->alignment(app()->getLocale() == 'ar' ? 'right' : 'left')`,
I think if we use the `text-start` and `text-end` align it can be handled without add any classes or condition for checking language direction.

I added `->alignStart()` , `->alignEnd()` and `->alignJustify()`


zepfietje:
Thanks for the PR, @AAbosham!

@danharrin and I have discussed this recently and want to move to start and end instead of left and right. Both APIs should still be supported, but we should document start and end. That does require some changes to this PR though (and probably add it to the other places as well). Dan and I can handle this if you like.

AAbosham:
Thank you for replying,I actually edit every text-right and text-left but I don't make PR wait for a response to this PR.

And also after [PR](https://github.com/laravel-filament/filament/commit/d625f40f10cd20675a489871512204eb6500379a)  we have some issue on dropdown item

image

So that i need to fix this.

I don't mind if you and @danharrin are going to move it or let me to make PR to fix this.
It is up to you to decide.

robertorinaldi-dev:
I noticed that the alignment doesn't work with IconColumn, I was looking at how to fix it but maybe after this change it would be easier

themahdavi:
> Thank you for replying,I actually edit every text-right and text-left but I don't make PR wait for a response to this PR.
> 
> And also after [PR](https://github.com/laravel-filament/filament/commit/d625f40f10cd20675a489871512204eb6500379a) we have some issue on dropdown item
> 
> image
> 
> So that i need to fix this.
> 
> I don't mind if you and @danharrin are going to move it or let me to make PR to fix this. It is up to you to decide.

same issue for my projects on Persian language
on RTL version User menu and all dropdowns like ActionGroup have issue.
image

danharrin:
Mark as ready for review when the `left` & `start` / `right` & `end` options are synonymous everywhere.

danharrin:
Thanks!

packages/tables/src/Columns/Concerns/HasAlignment.php
```diff
@@ -15,19 +15,34 @@ public function alignment(string | Closure | null $alignment): static
...
+    {
+        return $this->alignment(static fn (): ?string => $condition ? 'end' : null);
+    }
+
+    public function alignJustify(bool | Closure $condition = true): static
+    {
+        return $this->alignment(static fn (): ?string => $condition ? 'justify' : null);
+    }
+
+    public function alignLeft(bool | Closure $condition = true): static
```
zepfietje, line 38: @danharrin, shouldn't we mark the old methods as deprecated?

danharrin: Nah I think they're usable by people who dont ever support RTL

zepfietje: But still they're confusing for people only supporting RTL.

AAbosham: I think in some case we need this option, for ex mobile, tel or fax is starting from left for both RTL and LTR

zepfietje: Oh thanks for the insights, @AAbosham! Definitely have to keep those then.

danharrin: We'll need to fix the implementation then because I added them as aliases

AAbosham: I fixed that.

#4853

PR #4813, opened 2022-11-03
Copyable in TextColumn

moskoweb:
Allows you to copy the contents of a column of type TextColumn.

#4780


packages/tables/resources/views/columns/text-column.blade.php
```diff
@@ -42,6 +42,13 @@
         },
         'whitespace-normal' => $canWrap(),
     ]) }}
+    
+    @if ($isCopyable())
```
danharrin: I think you should be able to have an icon AND a copy button at once. Or, at least, I can imagine that we will get that request within a week 😂

moskoweb: I changed it to be set as default when activating, however, it can be customized if you want in a customized way.

packages/tables/src/Columns/TextColumn.php
```diff
@@ -16,6 +16,58 @@ class TextColumn extends Column
...
+    protected int | Closure | null $copyMessageDuration = null;
+
+    public function copyable(bool | Closure $condition = true): static
+    {
+        $this->isCopyable = $condition;
+        
+        if ($this->isCopyable) {
+            $this->icon = 'heroicon-o-clipboard';
+            $this->iconPosition = 'after';
+        }
```
moskoweb: Here

packages/tables/resources/views/columns/text-column.blade.php
danharrin: Yeah but people should be able to use an icon & the copy button. At the moment, they cant use an icon too if they use the copy button

moskoweb: I added an extra function that adds the copy icon and position at the end, for those who want a copy icon pattern.

ManojKiranA:
Hey @moskoweb have you removed     `->copyIcon()` method

danharrin:
I did that to keep it in line with the existing Color column copy functionality. You can add an icon() normally if you want

moskoweb:
> I did that to keep it in line with the existing Color column copy functionality. You can add an icon() normally if you want

You need to delete the copyIcon in the documentation so that it doesn't cause problems.

PR #4831, opened 2022-11-04
make global search case-insensitive, and allow the search to run on all locales

AbanoubNassem:
fixes https://github.com/filamentphp/filament/discussions/4828

### DIFF
--- a/packages/admin/src/Resources/Resource.php
+++ b/packages/admin/src/Resources/Resource.php
@@ -445,15 +445,14 @@ class Resource
             $query->when(
                 method_exists($model, 'isTranslatableAttribute') && $model->isTranslatableAttribute($searchAttribute),
                 function (Builder $query) use ($databaseConnection, $searchAttribute, $searchOperator, $searchQuery, $whereClause): Builder {
-                    $activeLocale = app()->getLocale();
 
                     $searchColumn = match ($databaseConnection->getDriverName()) {
-                        'pgsql' => "{$searchAttribute}->>'{$activeLocale}'",
-                        default => "json_extract({$searchAttribute}, \"$.{$activeLocale}\")",
+                        'pgsql' => "{$searchAttribute}::text",
+                        default => "json_extract({$searchAttribute}, '$')",
                     };
 
                     return $query->{"{$whereClause}Raw"}(
-                        "lower({$searchColumn}) {$searchOperator} ?",
+                        "lower({$searchColumn}) {$searchOperator} lower(?)",
                         "%{$searchQuery}%",
                     );
                 },
--- a/packages/tables/src/Columns/Concerns/InteractsWithTableQuery.php
+++ b/packages/tables/src/Columns/Concerns/InteractsWithTableQuery.php
@@ -87,14 +87,15 @@ trait InteractsWithTableQuery
             $query->when(
                 method_exists($model, 'isTranslatableAttribute') && $model->isTranslatableAttribute($searchColumnName),
                 function (Builder $query) use ($searchColumnName, $searchOperator, $search, $whereClause, $databaseConnection): Builder {
-
+                    $activeLocale = $this->getLivewire()->getActiveTableLocale() ?: app()->getLocale();
+                    
                     $searchColumn = match ($databaseConnection->getDriverName()) {
-                        'pgsql' => "{$searchColumnName}::text",
-                        default => "json_extract({$searchColumnName}, '$')",
+                        'pgsql' => "{$searchColumnName}->>'{$activeLocale}'",
+                         default => "json_extract({$searchColumnName}, \"$.{$activeLocale}\")",
                     };
 
                     return $query->{"{$whereClause}Raw"}(
-                        "lower({$searchColumn}) {$searchOperator} lower(?)",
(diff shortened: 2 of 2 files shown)


AbanoubNassem:
just noticed , it broke the local `Resource` search , working on fixing that

AbanoubNassem:
just noticed that the local `Resource` search , has same issue, working on fixing that as well.

**FIXED.**

danharrin:
I've reverted the fix for tables because they should have a LocaleSwitcher anyway

AbanoubNassem:
@danharrin 
> I've reverted the fix for tables because they should have a LocaleSwitcher anyway

we still have the same issues as there : https://github.com/filamentphp/filament/discussions/4828 , it's case sensitive as well. 

I believe, if we have a website with like 6 different locales , to keep switching the table locale to search, is not user friendly. Imagine a team of a company working together.

for individual/single owner websites, I see that they can switch the locale easily, because they know what they are doing , but in the scale of a company , this will not be the required behaviour.

I know this has its impact on performance , therefore if you are interested , I can work on implementing `scout` search capabilities, so we get the benefits of the indexed search.

danharrin:
Scout can be integrated like this - https://filamentphp.com/docs/2.x/tables/getting-started#searching-records-with-laravel-scout

danharrin:
I kinda agree now with the multi-locale search... what is the performance impact for searching 100,000 records?

AbanoubNassem:
> 

@danharrin 

with 110600 records,  app in debug mode with `debugbar` attached , not optimised app, on Macbook Pro 2016:

```
//current implementation 

1 character (a) : 709ms
2 character (se) : 822ms
5 character (volup) : 810ms
2 words (Amsterdam Padova) : 905ms

-----------------------------------
// with the one I provided

1 character (a) : 660μs
2 character (se) : 1.34s
5 character (volup) : 1.37s
2 words (Amsterdam Padova) : 1.61s
```

so form what I see , it's not a huge impact , it works fairly well actually. 

from a normal user perspective ,I can't notice the performance impact.

danharrin:
@zepfietje wdyt about these performance stats

zepfietje:
Hmm, 180% duration is not something we should ignore. What's causing the slowdown specifically? The multi-language search?

danharrin:
Yeah

zepfietje:
Hard one. How would you toggle between locales for the global search if it didn't search for all locales by default?

AbanoubNassem:
> Hard one. How would you toggle between locales for the global search if it didn't search for all locales by default?

@danharrin  @zepfietje 

I noticed something weird tho, global search , has no impact , performance is

```
1 character (a) : 294ms
2 character (se) : 369ms
5 character (volup) : 340ms
```

while I'm searching multiple tables , not one table like with resource-table search .

not really sure , why is that!?

**NOTE:** the performance stats , is not on global search it's on the resource-table search. the global search seems fine.

zepfietje:
Thanks for clarifying @AbanoubNassem!

Maybe @danharrin could explain, since I haven't looked too deep into this part.

As for my question, there's a toggle above the table then, right?

AbanoubNassem:
> Thanks for clarifying @AbanoubNassem!
> 
> Maybe @danharrin could explain, since I haven't looked too deep into this part.
> 
> As for my question, there's a toggle above the table then, right?

@zepfietje 

no there is not , it searches on the `$activeLocale = app()->getLocale();` , so it forces the app locale , no way to toggle/change it.

danharrin:
Thats only for global search as far as I remember? Tables search respects the LocaleSwitcher

AbanoubNassem:
> Thats only for global search as far as I remember? Tables search respects the LocaleSwitcher

it does respect the `LocaleSwitcher` but currently it lowers the column , not the input.

and on refresh I loses the LocaleSwitcher state, moreover I have 4 locales , to keep switching the locale to find the right locale which has the input is a bit cumbersome.

danharrin:
We could make a function to store the locale in the session?

AbanoubNassem:
> We could make a function to store the locale in the session?

hummm how about , if we make searching all locales optional? 
so by default it's false , but if you need it you can toggle it.

` public function applySearchConstraint(Builder $query, string $search, bool &$isFirst, bool $isIndividual = false, bool $searchAllLocales = false){}`

I think this would be a better approach, what do you think @danharrin  @zepfietje ?

danharrin:
You can do it already by just overriding that method on the Livewire class. Extract it into a trait in your project if you want?

AbanoubNassem:
> You can do it already by just overriding that method on the Livewire class. Extract it into a trait in your project if you want?

I can't , this is done on the `Column` -> `InteractsWithTableQuery`, I think you still confusing the `Column` search with `Global Search` , global search is working perfectly fine , it's the `Column` search now which has the issue.

danharrin:
Oh right, you can override it using `->searchable(query: fn (Builder $query) => ...)` then

AbanoubNassem:
> Oh right, you can override it using `->searchable(query: fn (Builder $query) => ...)` then

okey will do , thank you

zepfietje:
This thread needs no further action then?

danharrin:
i think so

PR #4833, opened 2022-11-04
add closure to table filters default state

husam-tariq:


danharrin:
It can already accept a closure, but the type is mixed since the default value could be a string, bool, int, float, whatever

husam-tariq:
> It can already accept a closure, but the type is mixed since the default value could be a string, bool, int, float, whatever

what if i change it to 

` protected Closure | null $defaultState = null;

    public function default(Closure | null $state ): static
    {
        $this->defaultState = $state;

        return $this;
    }
`

danharrin:
No, because then everything else is blocked as a default value. Please research how PHP union types work

PR #4844, opened 2022-11-05
Enhancement: default to action model in HasActions trait

awcodes:
This PR defaults the form model in actions to use the action's model before falling back to introspection to determine the model. This helps to mitigate problems where the introspection can result in the action form model being null.


danharrin:
Thanks

PR #4857, opened 2022-11-06
Expose Laravel Loop Object To Column

Nuhel:
Sometime we need to show the Serial Number/Iteration Number in table. I could not find any way to do this with filament table. (May be there is a way, if so i will close this PR)

This PR Expose Laravel Loop Object To Columns,

We can use it like below

```php

Tables\Columns\TextColumn::make('sl')->getStateUsing(function ($loop){
      return $loop->iteration;
}),
```

Also as it is laravel's foreach loop object, We can access every other property like
index, remaining, count etc.

### DIFF
--- a/packages/tables/src/Columns/Concerns/CanFormatState.php
+++ b/packages/tables/src/Columns/Concerns/CanFormatState.php
@@ -196,10 +196,10 @@ trait CanFormatState
         return $this->evaluate($this->timezone) ?? config('app.timezone');
     }
 
-    public function rowIndex(bool $fromZero = false): static
+    public function rowIndex(bool $isFromZero = false): static
     {
         $this->getStateUsing(static function (stdClass $rowLoop) use ($fromZero): ?string {
-            return $rowLoop->{ $fromZero ? 'index' : 'iteration'};
+            return (string) $rowLoop->{$isFromZero ? 'index' : 'iteration'};
         });
 
         return $this;
--- a/packages/tables/src/Columns/Concerns/CanFormatState.php
+++ b/packages/tables/src/Columns/Concerns/CanFormatState.php
@@ -11,6 +11,7 @@ use Illuminate\Support\Carbon;
 use Illuminate\Support\HtmlString;
 use Illuminate\Support\Str;
 use stdClass;
+
 trait CanFormatState
 {
     protected ?Closure $formatStateUsing = null;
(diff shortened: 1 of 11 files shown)


danharrin:
Instead of `loop`, please use `rowIndex` and start from 0

Nuhel:
> Instead of `loop`, please use `rowIndex` and start from 0

You mean instead of loop object just pass an int that starts from 0?

danharrin:
well at the moment you're passing in the column loop object and not the row loop object? i think? which is wrong right?

Nuhel:
> well at the moment you're passing in the column loop object and not the row loop object? i think? which is wrong right?

Yes, if we pass the loop object, we will have some extra options like remaining, count, depth, last etc, which will help if we need those.

But it's your call , i will update the PR if you think loop object is not a good fit.

danharrin:
I agree now that the loop object is fine, but please test your code. I don't think it works, as you're passing in the column loop and not the row loop. For clarity we can name it `rowLoop()` as well, but I think it still needs fixing

Nuhel:
> I agree now that the loop object is fine, but please test your code. I don't think it works, as you're passing in the column loop and not the row loop. For clarity we can name it `rowLoop()` as well, but I think it still needs fixing

Yes I was passing column loop -_- .

Test With normal Table Layout.
Screenshot 2022-11-07 at 7 19 55 PM

Test with Split and Stack
Screenshot 2022-11-07 at 7 19 26 PM

danharrin:
I've just refactored a bit, does it still work?

Nuhel:
> I've just refactored a bit, does it still work?

Not Working With Stack & Split.
Filament\Tables\Columns\Column::rowLoop(): Argument #1 ($loop) must be of type ?stdClass, Closure given,

Nuhel:
> I've just refactored a bit, does it still work?

{{ $layoutComponent->viewData(['recordKey' => $recordKey, 'rowLoop' => $rowLoop]) }}

rowLoop has to be passed in viewData.

danharrin:
@Nuhel how about passing it into the object instead and then using `$getRowLoop()` in the view? I've added the trait to the Component object

Nuhel:
> @Nuhel how about passing it into the object instead and then using `$getRowLoop()` in the view? I've added the trait to the Component object

May be i did not get you, can you elaborate please?

danharrin:
I've made the changes, test again please

Nuhel:
> I've made the changes, test again please

Yes working. 

Should I update documentation? And I am thinking to add below function to CanFormatState Trait

```php
public function rowIndex(bool $fromZero = false): static
{
    $this->getStateUsing(static function ($rowLoop) use($fromZero): ?string {
        return $rowLoop->{ $fromZero? "index":"iteration"};
    });

    return $this;
}
```

packages/tables/src/Columns/Concerns/CanFormatState.php
```diff
@@ -194,4 +195,13 @@ public function getTimezone(): string
     {
         return $this->evaluate($this->timezone) ?? config('app.timezone');
     }
+
+    public function rowIndex(bool $fromZero = false): static
```
danharrin: Please move this method to the TextInput class

PR #4863, opened 2022-11-07
added missing german translations

dabenzel:
Added missing german translations and fixed one mistake

### DIFF
--- a/packages/forms/resources/lang/de/components.php
+++ b/packages/forms/resources/lang/de/components.php
@@ -16,6 +16,10 @@ return [
                 'label' => 'Zu :label hinzufügen',
             ],
 
+            'create_item_between' => [
+                'label' => 'Einfügen',
+            ],
+
             'delete_item' => [
                 'label' => 'Löschen',
             ],
@@ -25,7 +29,7 @@ return [
             ],
 
             'move_item_down' => [
-                'label' => 'Nach unter verschieben',
+                'label' => 'Nach unten verschieben',
             ],
 
             'move_item_up' => [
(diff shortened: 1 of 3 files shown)


danharrin:
Thanks

packages/forms/resources/lang/de/components.php
```diff
@@ -52,6 +56,22 @@
 
     ],
 
+    'checkbox_list' => [
+
+        'buttons' => [
+
+            'deselect_all' => [
+                'label' => 'Alle deselektieren',
```
dissto, line 64: @dabenzel @zepfietje 

How about "Auswahl aufheben". 
I personally never heard someone say "deselektieren" in my life 😊

packages/forms/resources/lang/de/components.php
```diff
@@ -52,6 +56,22 @@
...
+            ],
+
+            'select_all' => [
+                'label' => 'Alle selektieren',
```
dissto, line 68: Same here "Alle/s auswählen" 🤔

packages/forms/resources/lang/de/components.php
dabenzel: Good idea, you can issue a PR for that if you want

packages/forms/resources/lang/de/components.php
dabenzel: s.o.

packages/forms/resources/lang/de/components.php
zepfietje: Should `select_all` also be updated then?

PR #4878, opened 2022-11-09
Allow add header & footer for custom content modal

AAbosham:
Sometime you may need to add custom content on footer of modal.

image


danharrin:
I think this should be simplified. Instead of using `modalContent()` again, we can introduce a new `modalFooter()` which is completely separate.

AAbosham:
What about modalContent? Will rename to modalHeader?

If modal have footer, user think its have header

danharrin:
`modalContent()` can stay as it is

danharrin:
Thanks

PR #4880, opened 2022-11-09
Fix: modal issues

awcodes:
Fixes modal state and DOM diffing issues by dispatching an event from the modal close which then unmounts the current action.


danharrin:
that's fixed it on my end!

awcodes:
This looks good to me and I like the implementation better.

Roxeen12:
I think there was an issue 
image

zepfietje:
How do you think it's related to this PR, @Roxeen12?

Would be best to open a new bug report so we can look into the issue.

danharrin:
Yeah we absolutely need a reproduction repo in an issue to be able to fix this

PR #4891, opened 2022-11-10
feat: add `registerViteTheme`

innocenzi:
This pull request is an attempt at providing a better solution for using a theme with Vite. 

A new `registerThemeUsingVite` is introduced: it accepts a single or a list of assets, as well as a build directory as the second parameter.

```php
Filament::registerViteTheme('resources/css/filament.css');
```

I think this is cleaner than the current recommended approach:

```php
Filament::registerTheme(
    app(Vite::class)('resources/css/filament.css'),
);
```

I would understand if you think this is superfluous though.

### DIFF
--- a/packages/admin/src/FilamentManager.php
+++ b/packages/admin/src/FilamentManager.php
@@ -16,6 +16,7 @@ use Illuminate\Contracts\Auth\Authenticatable;
 use Illuminate\Contracts\Auth\Guard;
 use Illuminate\Contracts\Support\Htmlable;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Foundation\Vite;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Event;
@@ -145,9 +146,9 @@ class FilamentManager
         $this->theme = $theme;
     }
 
-    public function registerViteTheme(string | array $theme, string $buildDirectory = null): void
+    public function registerViteTheme(string | array $theme, ?string $buildDirectory = null): void
     {
-        $this->theme = app(\Illuminate\Foundation\Vite::class)($theme, $buildDirectory);
+        $this->theme = app(Vite::class)($theme, $buildDirectory);
     }
 
     public function registerUserMenuItems(array $items): void
--- a/packages/admin/src/FilamentManager.php
+++ b/packages/admin/src/FilamentManager.php
@@ -48,10 +48,6 @@ class FilamentManager
 
     protected string | Htmlable | null $theme = null;
 
-    protected string | array | null $viteTheme = null;
-
-    protected ?string $viteBuildDirectory = null;
-
     protected array $userMenuItems = [];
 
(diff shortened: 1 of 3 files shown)


innocenzi:
> using VuePress' built-in support for Laravel Mix / Webpack Encore via Illuminate\Foundation\Vue::class(). This allows us to use webpack mix() or encore() functions inside our themes without having to manually add them into every single blade file we create!

Not quite riiight haha. Good bot though.

packages/admin/src/FilamentManager.php
```diff
@@ -145,6 +149,12 @@ public function registerTheme(string | Htmlable | null $theme): void
         $this->theme = $theme;
     }
 
+    public function registerThemeUsingVite(string | array $theme, string $buildDirectory = null): void
+    {
+        $this->viteTheme = $theme;
```
danharrin: New properties don't need introducing, we can just call `app(Vite::class)` and pass it to `$this->theme`

packages/admin/src/FilamentManager.php
```diff
@@ -145,6 +149,12 @@ public function registerTheme(string | Htmlable | null $theme): void
         $this->theme = $theme;
     }
 
+    public function registerThemeUsingVite(string | array $theme, string $buildDirectory = null): void
```
danharrin: We could name this just `registerViteTheme`

danharrin:
> Not quite riiight haha. Good bot though.

Don't be mean, it got excited at the end! (Yay!) 😂

innocenzi:
I applied the recommended changes — my initial thoughts for the added properties was to not call `Vite` inside the setter, but you're right, it's better this way 👍

danharrin:
Cool, might be worth updating the docs example too?

innocenzi:
Woops, definitely. Done

zepfietje:
I feel like we should also have a method for Laravel Mix if we add one for Vite...

packages/admin/src/FilamentManager.php
```diff
@@ -145,6 +145,11 @@ public function registerTheme(string | Htmlable | null $theme): void
         $this->theme = $theme;
     }
 
+    public function registerViteTheme(string | array $theme, string $buildDirectory = null): void
+    {
+        $this->theme = app(\Illuminate\Foundation\Vite::class)($theme, $buildDirectory);
```
zepfietje: Could you import the namespace at the top of the file?

zepfietje (changes requested):

innocenzi:
> I feel like we should also have a method for Laravel Mix if we add one for Vite...

I understand the feeling, but the syntax for Vite is more cumbersome than Mix's. Whereas there is the `mix` helper, there is no similar `vite` function -- you need to resolve Vite from the container before calling it, making the syntax quite bad in that situation, in my opinion.

zepfietje:
Yeah I agree, but still feels weird to have this new method since Vite is Laravel's default now.

PR #4893, opened 2022-11-10
Updated and added Bosnian translations

amsnvc:
Updated a few already existing translations and also added the missing ones :)


danharrin:
Thanks!

Issue #3916, opened 2022-09-09
Compatibility with blade-ui-kit/blade-heroicons 2.*


nocodelab:
### Package

filament/notifications

### Package Version

2.0

### Laravel Version

9.28.0

### Livewire Version

2.10.7

### PHP Version

8.1

### Problem description

The dependency to blade-ui-kit/blade-heroicons should be updated to 2.* to support the new heroicons version (2).

### Expected behavior

The installation should complete succesfully.

### Steps to reproduce

composer require blade-ui-kit/blade-heroicons:"^2.0"
composer require filament/notifications:"^2.0"

### Reproduction repository

https://github.com/blade-ui-kit/blade-heroicons

### Relevant log output

```shell
Problem 1
    - filament/notifications[v2.15.1, ..., 2.x-dev] require blade-ui-kit/blade-heroicons ^1.2 -> found blade-ui-kit/blade-heroicons[1.2.0, 1.2.1, 1.3.0, 1.3.1] but it conflicts with your root composer.json require (^2.0).
    - Root composer.json requires filament/notifications ^2.0 -> satisfiable by filament/notifications[v2.15.1, ..., 2.x-dev].

Use the option --with-all-dependencies (-W) to allow upgrades, downgrades and removals for packages currently locked to specific versions.

Installation failed, reverting ./composer.json and ./composer.lock to their original content.
```

ryangjchandler:
@nocodelab This won't have until version 3.x of Filament because it's considered a breaking change for existing applications using renamed / removed icons in the new version of Heroicons.

zepfietje:
Additionally, we can't upgrade to v2 of the Blade Heroicons package, since it dropped support for Laravel 8.

Closing this issue, as it's already being tracked on our v3 roadmap: https://github.com/orgs/filamentphp/projects/2/views/1.

danharrin:
Reopening this so people stop asking

zepfietje:
Done for v3: https://github.com/filamentphp/filament/issues/4490.

PR #4894, opened 2022-11-10
Upgrade blade-ui-kit/blade-heroicons

rolfvandekrol:
Filament uses blade-heroicons `^1.2`. This version is no longer updated and newer heroicons are not added to this package.

This PR upgrades blade-heroicons to `^2.0`. 

I'm not sure if this mergeable in the current state. Some icons are renamed from version 1.x to 2.x, which makes them not backwards compatible. Also version 2.x is not compatible with Laravel 8. 

On the other hand, being stuck on 1.x forever might not be the greatest path forward.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
     },
     "require-dev": {
         "akaunting/laravel-money": "^1.2|^2.0|^3.0",
-        "blade-ui-kit/blade-heroicons": "^1.2",
+        "blade-ui-kit/blade-heroicons": "^2.0",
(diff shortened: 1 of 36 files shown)


zepfietje:
Thanks for the effort, @rolfvandekrol, but we can't upgrade Heroicons in Filament v2.
This has all been done for v3 already.

Reference: https://github.com/filamentphp/filament/issues/3916.

howdu:
For anybody arriving here... you simple need to set the Heroicon version
```bash
composer require blade-ui-kit/blade-heroicons:'^1.2'
```

PR #4899, opened 2022-11-11
Add HasFontFamily in TextColumn

moskoweb:
Allows you to change the column font to Sans, Serif and Mono.


zepfietje:
You can already achieve this by [passing extra attributes to the column](https://filamentphp.com/docs/2.x/tables/columns/getting-started#custom-attributes).

I think we might close this PR. Thoughts, @danharrin?

moskoweb:
Changing color, size, alignment and weight can also add extra attributes, and it still has a method.

zepfietje:
Yeah I know. I'm discussing internally what's the best way forward.

pxlrbt:
I think we should have an easier way to add classes and reduce the helpers.

PR #4905, opened 2022-11-11
Action Group Button Type

moskoweb:


zepfietje:
The action group trigger button optionally having a label is something I've been tracking on the roadmap as a proposal for a while. I'm not sure about the best API for this, @danharrin?

danharrin:
Like all other actions, calling `button()` should change the `$this->view`. Not just use a conditional statement within the view like this.

moskoweb:
You will need to create a new view in both package tables and supports, right?

danharrin:
Yeah because the dark mode config is package-specific

moskoweb:
@danharrin I separated the views again, creating a view for the button and keeping the original one for the icon.

packages/tables/src/Actions/ActionGroup.php
```diff
@@ -21,4 +21,11 @@ public function getActions(): array
 
         return $actions;
     }
+
+    public function button(): static
+    {
+        $this->view('tables::actions.group-button');
+
+        return $this;
+    }
```
zepfietje, line 30: I believe we should also add the link variant of the action just like regular actions have button, link and icon button.

packages/support/resources/views/components/actions/group-button.blade.php
```diff
@@ -0,0 +1,38 @@
...
+            :dark-mode="$darkMode"
+            :icon="$icon"
+            :size="$size"
+            :tooltip="$tooltip"
+        >
+            <x-slot name="label">
+                {{ $label }}
+            </x-slot>
+        </x-filament-support::button>
+    </x-slot>
```
zepfietje, line 29: Wouldn't it be easier to use the existing `packages/support/resources/views/components/actions/group` component and only swap out the trigger for button/link/icon button?

zepfietje (changes requested):

moskoweb:
I had done that, a parameter that I changed, but I was told to make
separate files because of the dark theme.

Em sáb., 7 de jan. de 2023 17:27, Zep Fietje ***@***.***>
escreveu:

packages/support/resources/views/components/actions/group-button.blade.php
moskoweb: I had done that, a parameter that I changed, but I was told to make
separate files because of the dark theme.

moskoweb: @zepfietje do you have any suggestions on how to do it?  I had placed a Type property, and from there I changed the format, as I said @danharrin had commented about making separate files.

https://github.com/filamentphp/filament/pull/4905/commits/a8a8d1ff1a89293a213547dcc22fac4c2f9ff3f8

zepfietje: @danharrin

danharrin: I'll probably implement this for v3, action groups are much simpler as there is only 1 view.

zepfietje: Sounds good. Shall we close this PR then, @danharrin?

danharrin: Leave it open and ill use it when i write this feature

moskoweb: @danharrin When will version 3 come out, to count on this feature?

zepfietje: @moskoweb, we don't know yet, also depends on Livewire v3.

moskoweb: So this feature will be open until version 3, I will see to use my repository even then in the project, because I need this feature.

danharrin: You can use this in v2 without needing to use your fork

You can copy the new ActionGroup class and all the views into your app and use them like normal

Or keep the same ActionGroup that we use, copy the views into your app, and just pass ->view('your view') to the group when you need to change the button

danharrin:
I've implemented it this morning in v3. You can still use this in v2 if you copy the views and class into your project.

moskoweb:
Is this feature already possible? To have "label" in a group of buttons, instead of just "tooltip".

PR #4926, opened 2022-11-13
Fix: Validation error color for dark mode

### LINKED ISSUE #4917: Validation error text color
Sometimes, the validation error text colour is too dark for the background color of the layout component. For example, inside a wizard.

Maybe we should switch to using `danger` with an opacity?

Screenshot 2022-11-12 at 16 26 17

awcodes:
Makes validation easier to read in dark mode.

Screen Shot 2022-11-13 at 8 42 03 AM
Screen Shot 2022-11-13 at 8 41 54 AM


zepfietje:
I think the color comes across as a little too harsh in dark mode.

zepfietje:
Could you try the 300 shade, @awcodes? I believe I saw the Tailwind team use that somewhere. Just can't find it right now.

awcodes:
Don't like the 300 at all. I could live with the 400, but they both look washed out to me.

It's an error, shouldn't it be harsh to draw attention. ;)

300
Screen Shot 2022-11-13 at 9 14 43 AM

400
Screen Shot 2022-11-13 at 9 15 45 AM

danharrin:
I think 400 is a good compromise

danharrin:
Technically, even 300 doesnt pass some a11y standards...

Screenshot 2022-11-13 at 17 11 04

awcodes:
Yea, passing AAA is almost impossible without having a true black and white site with no styling. lol.

I'll go through and update this to 400. Have to go through and update the borders now though.

Stay tuned.

awcodes:
Think I got everything. Let me know if you see something I missed.

zepfietje:
Sorry, I was wrong about 300. Just came here to say we should try 400. Glad you found that worked already, @awcodes.

PR #4933, opened 2022-11-13
add group to notifications

husam-tariq:


zepfietje:
@husam-tariq could you explain your thoughts behind this PR a little, please?

husam-tariq:
> 

Hi Zep, I've added groups to notifications, to be more organized, like the Navigation Group 

image

As you can see in the attached image, the above notifications that do not contain a group will appear normally, but those with groups will appear as a group with the title of the group and can be collapsed.

I can now group notifications according to events, for example, to display all delete notifications isolated from creation notifications and other events.  All I have to do is add the group name to the notification so it belongs to a specific group.

zepfietje:
Thanks for clarifying @husam-tariq.

I've discussed this with @danharrin and we both don't really like the idea of grouping notifications like this. If using notifications consciously, I don't think there's a need to group creation/deletion notifications.

husam-tariq:
I'm building a projects management system and in my case I wanted to group the notifications of each project, so I mead this.

Even Facebook makes groups for notifications.

Screenshot_20221114-152440_Facebook

zepfietje:
The Facebook notifications are only grouped by time though, which really is different.

husam-tariq:
After all, it's a new feature if they don't use it. It will not affect anything.

57115202-78bad200-6d88-11e9-81d0-00acb57ca263

zepfietje:
> After all, it's a new feature if they don't use it. It will not affect anything.

Might not affect Filament users too much. Definitely does affect the codebase and maintainers. We sure do want to enable you to create basically anything with Filament. However, we also should make deliberate decisions on what to include in the framework or not. Maintainability of the project is affected by every single PR.

danharrin:
In this case, introducing the feature would actually be a breaking change, since it changes the structure of the `$notifications` array to be grouped

zepfietje:
I'm going to close this PR for now, since:

1. We're not convinced about the UX/need of the feature.
2. It's a breaking change.
3. The behavior should also work for non-database notifications.

We might revisit this request in the future though.
Thanks for your effort and thoughts, @husam-tariq.

Issue #4888, opened 2022-11-10
Nested modal not working


AAbosham:
### Package

Other

### Package Version

v2.16.42

### Laravel Version

v8.0

### Livewire Version

v2.0

### PHP Version

PHP 8.1.0

### Problem description

Nested modal not working after this [commit](https://github.com/filamentphp/filament/commit/61a7ae0b4dc03f2046dcd76432995bb9ef9b5a12)

### v2.16.42 modal not working

![image](https://user-images.githubusercontent.com/12083600/201082983-cb17d0fd-8bc3-4c06-beb1-6f1563832298.png)

### v2.16.41 modal working

![image](https://user-images.githubusercontent.com/12083600/201083103-f8829e5b-f1dc-4a08-84dc-8cecdedc6207.png)

### Expected behavior

Nested modal should open from inside modal

### Steps to reproduce

Open create create from inside modal

### Reproduction repository

https://github.com/AAbosham/filament-deme-nested-modal-isssue

danharrin:
How do we reproduce this in the reproduction repo? Looks like you've just cloned the demo.

danharrin:
If we don't have code & instructions to reproduce this with, the issue will be closed

AAbosham:
I only add this action on PostResource

```
 Tables\Actions\ViewAction::make('comments')
                    ->modalContent(fn ($record) => view('test', compact('record'))),
                    
```

![Screenshot 2022-11-10 at 15-31-56 Posts - Filament Demo](https://user-images.githubusercontent.com/12083600/201093015-a866ed17-e59c-4e71-b9d3-c2b7ca79e4c9.png)


![Screenshot 2022-11-10 at 15-32-54 Posts - Filament Demo](https://user-images.githubusercontent.com/12083600/201093033-ccea6b1e-78c4-407b-8fd3-4b94549f022e.png)

danharrin:
It's not possible to nest tables within table actions, since that would break W3 guidelines on nesting `<form>` elements:

![image](https://user-images.githubusercontent.com/41773797/201494080-88238774-5434-4656-a552-5026da44cf98.png)

In v3 we will aim to support this using Livewire v3's upcoming "teleport" feature.

PR #4936, opened 2022-11-14
Fix: Nested modal

AAbosham:
Hi about this issue #4888 `wire:key` will solved it

### DIFF
--- a/packages/support/resources/views/components/modal/index.blade.php
+++ b/packages/support/resources/views/components/modal/index.blade.php
@@ -54,7 +54,7 @@
     aria-modal="true"
     class="filament-modal {{ $displayClasses }}"
     wire:ignore.self
-    wire:key="{{ $id }}-modal"
+    wire:key="{{ $this->id }}-{{ $id }}-modal-container"
 >
     {{ $trigger }}
 
--- a/packages/support/resources/views/components/modal/index.blade.php
+++ b/packages/support/resources/views/components/modal/index.blade.php
@@ -54,6 +54,7 @@
     aria-modal="true"
     class="filament-modal {{ $displayClasses }}"
     wire:ignore.self
+    wire:key="{{ $id }}-modal"
 >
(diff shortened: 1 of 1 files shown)


zepfietje:
I'm pretty sure this will cause other issues. @danharrin?

danharrin:
Modals have `wire:key`s already - but passed through `$attributes`. Since the `wire:key` should be a different structure depending on the action.

And yes a generic wire:key based on the modal ID would reintroduce issues we've fixed relating to form components within the modals.

danharrin:
I guess we could add the Livewire ID to the key and it would be ok

packages/support/resources/views/components/modal/index.blade.php
```diff
@@ -54,6 +54,7 @@
     aria-modal="true"
     class="filament-modal {{ $displayClasses }}"
     wire:ignore.self
+    wire:key="{{ $id }}-modal"
```
danharrin:
```suggestion
    wire:key="{{ $this->id }}-{{ $id }}-modal-container"
```

danharrin: Please commit this and I will merge

packages/support/resources/views/components/modal/index.blade.php
```diff
@@ -54,6 +54,7 @@
     aria-modal="true"
     class="filament-modal {{ $displayClasses }}"
     wire:ignore.self
+    wire:key="{{ $this->id }}-{{ $id }}-modal-container"
```
zepfietje, line 57: @danharrin don't we use `.` instead of `-` in other places?

danharrin: yeah i fixed it the minute after i merged lol

zepfietje: Lol, didn't see that commit. Thanks.

PR #4958, opened 2022-11-15
Additional column assertions

titantwentyone:
Some additional assertions for a few items which could potentially be useful:

To test against `formatStateUsing()`
```
assertTableColumnFormattedStateSet(string $name, $value, $record)
assertTableColumnFormattedStateNotSet(string $name, $value, $record)
```

To test against `extraAttributes()`
```
assertTableColumnHasExtraAttributes(string $name, $value, $record)
assertTableColumnDoesNotHaveExtraAttributes(string $name, $value, $record)
```

To test against `description()`
```
assertTableColumnHasDescription(string $name, $description, $record, $position = 'below')
assertTableColumnDoesNotHaveDescription(string $name, $description, $record, $position = 'below')
```

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -127,16 +127,16 @@
         },
         {
             "name": "aws/aws-sdk-php",
-            "version": "3.242.1",
+            "version": "3.243.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/aws/aws-sdk-php.git",
-                "reference": "9bfd85f696fff6a9b7810f1361751ad33a9b23d1"
+                "reference": "c4ac1ff8f40f11fcf0d8e58b3d13838a44a8a582"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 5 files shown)


zepfietje:
Thanks for the PR! I do see how those assertions could be useful.

Quick note though, to be more consistent with Laravel's assertions, shouldn't we use `DoesntHave` instead of `DoesNotHave`? (Edit: I see the existing assertions also don't use the Laravel naming.)

titantwentyone:
I didn't even think about that if I'm honest and those other assertions with "DoesNot" are my handiwork! Happy to change these if you wish. I'll amend these for this PR if you'd like and create a new PR to amend the others. There are others in the admin package also.

If we are changing these - do we want to create an alias and deprecate the "DoesNot" versions or simply just replace them?

zepfietje:
We should definitely ensure backward compatibility, so marking as deprecated is the way to go.

danharrin:
I prefer `DoesNot`, `Doesnt` gets on my nerves as it misses an apostrophe

danharrin:
Thanks

PR #4961, opened 2022-11-16
Dispatch event

cheesegrits:
I'm not sure if I've done this right, like it maybe should be in a concern of it's own rather than sharing HasAction, and I'm confused by the fromCallable stuff on the getAction() method in HasAction, I tried doing it that way with getDispatch(), but was blowing up.  And my test case is a bit primitive, but I can't think of any other test, all I can test on the server side is that the x-on:click is there.  And should probably bake in json_encoding the getDispatchData() return.

Anyway, happy to fix stuff if I need to.  Just didn't want to invest too much time if you weren't interested in merging this feature.  In case you were wondering, this is the kind of thing I need it for, where I want to dispatch an event from a table row to a map.  It's probably possible with the existing mountTableAction() event through Livewire, but that's a round trip to the server and back, which seems a bit much when the event just needs to travel 2 inches up the page.  :)

https://www.screencast.com/t/o4vlQs3js

### DIFF
--- a/packages/support/src/Actions/Concerns/HasAction.php
+++ b/packages/support/src/Actions/Concerns/HasAction.php
@@ -8,6 +8,8 @@ trait HasAction
 {
     protected Closure | string | null $action = null;
 
+    protected Closure | string | null $dispatch = null;
+
     public function action(Closure | string | null $action): static
     {
         $this->action = $action;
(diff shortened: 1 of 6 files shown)


danharrin:
As discussed on Discord you can use `extraAttributes()` for this

PR #4966, opened 2022-11-16
Add filament classes to table header

arjendejong12:
Add some classes to table header elements for specific targetting.

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -199,12 +199,12 @@
             <div
                 x-show="@js($shouldRenderHeaderDiv = ($isReorderable || $isGlobalSearchVisible || $hasFiltersPopover || $isColumnToggleFormVisible)) || selectedRecords.length"
                 {!! ! $shouldRenderHeaderDiv ? 'x-cloak' : null !!}
-                class="flex items-center justify-between p-2 h-14"
+                class="filament-tables-header-toolbar flex items-center justify-between p-2 h-14"
                 x-bind:class="{
                     'gap-2': @js($isReorderable) || selectedRecords.length,
                 }"
             >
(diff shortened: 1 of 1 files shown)


zepfietje:
I don't really like the `left` and `right` naming. Any other options?

danharrin:
Yeah me either, it's not RTL-friendly

arjendejong12:
You're both right, but I don't have any suggestions as to what it should be sadly.
________________________________

danharrin:
I've added `filament-tables-header-toolbar`, you can then use `:nth-child` to select the first or second `<div>`

packages/tables/resources/views/index.blade.php
```diff
@@ -222,7 +222,7 @@ class="flex items-center justify-between p-2 h-14"
                 @if ($isGlobalSearchVisible || $hasFiltersPopover || $isColumnToggleFormVisible)
                     <div class="flex items-center justify-end w-full gap-2 md:max-w-md">
                         @if ($isGlobalSearchVisible)
-                            <div class="flex items-center justify-end flex-1">
+                            <div class="filament-tables-search-container flex items-center justify-end flex-1">
```
zepfietje, line 225: Maybe this is better?

```suggestion
                            <div class="filament-tables-header-search-container flex items-center justify-end flex-1">
```

danharrin: We already have `filament-tables-search-input` so i think this makes sense without `-header`

zepfietje: For v3 I'm looking to merge the different search inputs we have. Therefore it would make sense to prefix the container with `-header` as `filament-tables-search-input` will potentially be renamed to e.g. `filament-search-input`.

PR #4967, opened 2022-11-17
[3.x] feature: add back footer logo toggling

ryangjchandler:
Not sure if you want the config for this in the actual config file. Feels a bit excessive to _have_ to write custom CSS to hide the logo using the `filament-footer-logo` class.


zepfietje:
I agree actually removing the logo from the footer (instead of hiding with CSS) should be possible. Don't think the config file is the right place though. @danharrin thoughts?

zepfietje:
What if we remove the logo from the unbranded Filament and only render it from within a theme service provider that adds logos to the sidebar and footer (in addition to branded styling)?

danharrin:
I'm tempted to say we should just remove it completely

PR #4968, opened 2022-11-17
fix: exclude double tabindex in DateTimePicker

kamilst96:
When tabbing between form components, **DateTimePicker** will take two tabs to move onto the next component. This PR adds tabindex="-1" to button element that wraps the input.

### DIFF
--- a/packages/forms/resources/views/components/date-time-picker.blade.php
+++ b/packages/forms/resources/views/components/date-time-picker.blade.php
@@ -42,6 +42,7 @@
             aria-label="{{ $getPlaceholder() }}"
             dusk="filament.forms.{{ $getStatePath() }}.open"
             type="button"
+            tabindex="-1"
             @if ($isDisabled()) disabled @endif
             {{ $getExtraTriggerAttributeBag()->class([
                 'bg-white relative w-full border py-2 pl-3 pr-10 rtl:pl-10 rtl:pr-3 text-start cursor-default rounded-lg shadow-sm',


kamilst96:
I also have a dilemma concerning the **Select** component.

Currently it looks as if focus vanished during tabbing, since focus on **clear selection (X)** button is barely noticeable.
Excluding it from tabindex and binding **Del** key to clearing the field might be more intuitive.

My best guess would be `x-on:keydown.delete="state = null"`, which somewhat works, but I failed to find the **X** button's HTML in the blade file - is it generated by choices.js?

zepfietje:
> When tabbing between form components, **DateTimePicker** will take two tabs to move onto the next component. This PR adds tabindex="-1" to button element that wraps the input.

Is it still possible to open the date time picker panel from the keyboard?

zepfietje:
> My best guess would be `x-on:keydown.delete="state = null"`, which somewhat works, but I failed to find the **X** button's HTML in the blade file - is it generated by choices.js?

Yeah I believe it's added by Choices.js.

danharrin:
Without testing it, I assume this does break keyboard nav for the date picker. Maybe we need to remove the tabindex for the input instead?

kamilst96:
I decided to exclude button, because input has correct outline styling. I know, lazy thinking.
I tested on Chrome, excluding both input and button, and navigating the picker with keyboard works the same.

zepfietje:
Not sure if we're on the same page here. Dan and I are concerned that the date picker panel can no longer be opened using the keyboard if the button can't receive focus.

I haven't tested this PR, but like Dan mentioned, I feel like this breaks that behavior.

kamilst96:
Seems like language constraints on my part. Let's try this way:

**tabindex="-1" placed on input:**
Can be opened (Enter key): Yes
Can be navigated (arrow keys, tabbing within popover): Yes
Outline is properly styled: No

**tabindex="-1" placed on button:**
Can be opened (Enter key): Yes
Can be navigated (arrow keys, tabbing within popover): Yes
Outline is properly styled: Yes

I detected no difference in behavior on Chrome and Firefox.
I assume Safari behaves close to Chrome in this regard.

**Why exclude button instead of input?**
Clicking the field with a mouse pointer focuses input element.
I think keyboard should do the same for consistency.

If you decide to exclude input instead, then I'll need assistance with styling the outline.
Both elements need to have the same outline for consistency.

Button focus outline (first tab):


Input focus outline (second tab):

danharrin:
Thanks :)

PR #4969, opened 2022-11-17
[3.x] feature: add `filament-user-menu` class to user menu

ryangjchandler:


zepfietje:
Might this be used as a Blade component? If so, we should merge the class into `$attributes`.

PR #4984, opened 2022-11-18
[3.x] feature: allow non-LW auth routes to be used

ryangjchandler:
Overriding login/registration/password related components for a particular context is useful but it's limited to only Livewire components. This isn't very useful if your app has it's own auth flow already. The only way around it right now is to extend Filament's `Authenticate` middleware and overwrite the `redirectTo()` method.

These changes will allow somebody do this:

```php
$context
	->login(route('my-custom-login'))
```

Must simpler, and something that I tried to do initially.

### DIFF
--- a/packages/app/src/Context.php
+++ b/packages/app/src/Context.php
@@ -613,6 +613,10 @@ class Context
             return null;
         }
 
+        if (! class_exists($this->getLoginPage())) {
+            return $this->getLoginPage();
+        }
+
         return route("filament.{$this->getId()}.auth.login");
     }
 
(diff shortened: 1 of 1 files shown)


danharrin:
How about accepting any "route action". So a callable (invokable class, closure, [array, style])

Then we just pass it to the route. And you can redirect to your own login from within a simple closure. More flexible, right? We already do that for the billing portal feature.

ryangjchandler:
Ah, if that's what we're doing for billing, then yeah, let's do that. I'll update this in a minute!

danharrin:
You can rename `getLoginPage()` to `getLoginRouteAction()`. For the other auth pages too?

ryangjchandler:
Yeah, can update the other ones too.

PR #4989, opened 2022-11-18
[3.x] feature: add support for `readonly` input fields

ryangjchandler:
Only added to `TextInput` and `Textarea` since those are the only input types that semantically support `readonly`.

### DIFF
--- a/packages/forms/resources/views/components/text-input.blade.php
+++ b/packages/forms/resources/views/components/text-input.blade.php
@@ -62,7 +62,7 @@
                         'min' => (! $isConcealed) ? $getMinValue() : null,
                         'minlength' => (! $isConcealed) ? $getMinLength() : null,
                         'placeholder' => $getPlaceholder(),
-                        'readonly' => $isReadonly(),
+                        'readonly' => $isReadOnly(),
                         'required' => $isRequired() && (! $isConcealed),
                         'step' => $getStep(),
                         'type' => $hasMask ? 'text' : $getType(),
--- a/packages/forms/resources/views/components/text-input.blade.php
+++ b/packages/forms/resources/views/components/text-input.blade.php
@@ -62,6 +62,7 @@
                         'min' => (! $isConcealed) ? $getMinValue() : null,
                         'minlength' => (! $isConcealed) ? $getMinLength() : null,
                         'placeholder' => $getPlaceholder(),
+                        'readonly' => $isReadonly(),
                         'required' => $isRequired() && (! $isConcealed),
                         'step' => $getStep(),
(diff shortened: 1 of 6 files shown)


danharrin:
What are the real-life advantages over `disabled`?

ryangjchandler:
The field is still focusable and data is still dehydrated, unlike disabled fields (submitted in a regular form).

danharrin:
But if the data is dehydrated, its editable from the frontend? Is that desired?

danharrin:
Honestly I would just use `disabled()->dehydrated()` which makes the actual behaviour clearer IMO

ryangjchandler:
`readonly` is a real HTML attribute, so I think it's valid that we support it.

danharrin:
I agree, but I don't agree that the behaviour should be any different from `disabled()`. Could be unclear, right?

ryangjchandler:
I don't think so, if we have the attribute then the behaviour should match that of a regular HTML form no? Disabled forms don't get submitted in a regular HTML form, so it makes sense that we don't dehydrate them, but that's not the case for a readonly field.

danharrin:
Should readonly fields have disabled styling?

danharrin:
Also my brain is telling me the API should be `readOnly` instead of `readonly`, since its a hyphenated phrase in English

ryangjchandler:
The attribute isn't hyphenated though... and readonly fields have regular input styling. The main difference is that the value isn't mutable using the UI control.

danharrin:
The attribute isnt hyphenated, but our API doesn't correspond with HTML in other places (like `maxValue()` / `maxLength()`). Otherwise we'd end up like PHP's inconsistent API of string helpers that reflect C functions :D

danharrin:
I agree on the rest. Just the capitalisation now.

ryangjchandler:
Yeah fair, will sort capitalisation out shortly.

zepfietje:
> The attribute isnt hyphenated, but our API doesn't correspond with HTML in other places (like `maxValue()` / `maxLength()`). Otherwise we'd end up like PHP's inconsistent API of string helpers that reflect C functions :D

Shouldn't we rename `Textarea` to `TextArea` then? 😉

danharrin:
Following Mozilla as they say it's all 1 word when they use it in a sentence outside of code blocks - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea

Whereas in sentences, Mozilla use read-only https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly

zepfietje:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLTextAreaElement

Lol, inconsistent.

PR #4990, opened 2022-11-18
feat: Add saveStateUsing() method for editable columns

pxlrbt:
This PR adds the ability to override the default save behaviour for editable columns by providing a closure via `->saveStateUsing()`.

**Example from one of my apps:**

I have meters that have meter readings. I load them via leftJoin, so I need a special save routine.

```php
->saveStateUsing(function ($record, $columnName, $input) {
    $yearsAgo = str($columnName)->after('value_year_')->toInteger();

    MeterReading::query()
        ->where('meter_id', $record->id)
        ->whereBetween('created_at', [
            today()->startOfYear()->subYears($yearsAgo),
            today()->endOfYear()->subYears($yearsAgo),
        ])
        ->update(['value' => $input]);
})
```


packages/tables/src/Columns/Concerns/CanSaveState.php
```diff
@@ -0,0 +1,53 @@
...
+use Filament\Tables\Contracts\HasTable;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+use Illuminate\Support\Str;
+
+trait CanSaveState
+{
+    protected ?Closure $saveStateUsing = null;
+
+    public function bootCanSaveState()
```
danharrin: Where is this being called from?

packages/tables/src/Concerns/HasColumns.php
```diff
@@ -134,25 +130,12 @@ public function setColumnValue(string $column, string $record, $input): ?array
...
-            return null;
-        }
-
-        $record->setAttribute($columnName, $input);
-        $record->save();
+        $column->evaluate($column->getSaveStateUsing(), [
+            'columnName' => $columnName,
+            'record' => $record,
+            'table' => $this->getTable(),
+            'input' => $input,
```
danharrin: Maybe `$input` should be `$state`?

packages/tables/src/Concerns/HasColumns.php
```diff
@@ -134,25 +130,12 @@ public function setColumnValue(string $column, string $record, $input): ?array
...
-        if (! ($record instanceof Model)) {
-            return null;
-        }
-
-        $record->setAttribute($columnName, $input);
-        $record->save();
+        $column->evaluate($column->getSaveStateUsing(), [
+            'columnName' => $columnName,
+            'record' => $record,
+            'table' => $this->getTable(),
```
danharrin: Isn't `$table` already passed by default?

packages/tables/src/Columns/Concerns/CanSaveState.php
pxlrbt: Oh damn. I am getting to comfortable with Livewire booting these traits 🙈

packages/tables/src/Concerns/HasColumns.php
```diff
@@ -134,25 +130,12 @@ public function setColumnValue(string $column, string $record, $input): ?array
...
-
-        if (! ($record instanceof Model)) {
-            return null;
-        }
-
-        $record->setAttribute($columnName, $input);
-        $record->save();
+        $column->evaluate($column->getSaveStateUsing(), [
+            'columnName' => $columnName,
+            'record' => $record,
```
danharrin: Record should be passed by default anyway?

packages/tables/src/Concerns/HasColumns.php
```diff
@@ -134,25 +130,12 @@ public function setColumnValue(string $column, string $record, $input): ?array
...
-        }
-
-        if (! ($record instanceof Model)) {
-            return null;
-        }
-
-        $record->setAttribute($columnName, $input);
-        $record->save();
+        $column->evaluate($column->getSaveStateUsing(), [
+            'columnName' => $columnName,
```
danharrin: How about just using `$column->getName()`?

packages/tables/src/Columns/Concerns/CanSaveState.php
pxlrbt: What do you think is the best way? Renaming it to `setUp()` and importing it to `Column::setUp()` (if instance of Editable)?

pxlrbt:
@danharrin How would you solve the PHPStan issues? Column is of Type `Editable` but that doesn't ensure the methods on the trait. Should I add them to the interface?

packages/tables/src/Columns/Concerns/CanSaveState.php
danharrin: Just call `saveStateUsing()` directly from `setUp()`, I don't think it needs to be in the trait. Please test it too

danharrin:
Yeah add the methods you need from the trait to the interface, thats why i had to make it in the first place

packages/tables/src/Columns/Concerns/CanSaveState.php
pxlrbt: Hm, I'd like to keep it in the trait and keep the `setUp()` method clean. But we could also move it to `Column::setUp()` if you prefer.

pxlrbt:
I just changed this to draft because I am trying to solve the collision issues when typing during an update.

See the video. When typing to fast, the Livewire update might overwrite the new input. Josh Hanley helped me with this on Livewire Discord. Seems to work if we use Alpine's `x-data`:

**Before:**

```html
<div>
    <input
        value="{{ $value }}"
        x-on:change="
            response = await $wire.setColumnValue($event.target.value)
            if (response.state) value = response.state
        "
    />
</div>
```

**After:**

```html
<div x-data="{value: '{{ $value }}'}">
    <input
        x-model="value"
        x-on:change="
            response = await $wire.setColumnValue($event.target.value)
            if (response.state) value = response.state
        "
    />
</div>
```

`@entangle` also seems to work and keeps the input reactive if new data is coming from the server, but I think we can't use this as we have no `getStatePath()` on the editable column.

If you have a better idea or want to discuss this, hit me up on Discord. I still can't wrap my head around Livewire and Alpine internals 🙈 

https://user-images.githubusercontent.com/22632550/202820511-1b43d4fa-f859-4f86-951b-2de9ed092531.mov

danharrin:
Yeah the Alpine solution looks fine to me

pxlrbt:
Please don't mind those last commits. I had this in a project repo and PhpStorm pushed to both repos 🙈

danharrin:
What's left to do for this?

pxlrbt:
Need to remove my bad commits and check whether we need to transfer the Alpine solution to other editable fields, too. I try to finalize this by the end of the week.

pxlrbt:
@danharrin 

I think I am done with this.

- Introduce `saveStateUsing()` for editable columns
- Prevent flash of value on all editable columns by wrapping them in Alpine
- Use state returned from backend for `TextInputColumn` and `SelectColumn` (useful for TextInputColumn when formatting text, haven't found a use case for SelectColumn yet)
- Prevent double submits by blocking all editable columns from changes while loading. CheckboxColumn only blocked for mouse input. Wasn't sure what the best way to block keyboard input was.

packages/tables/resources/views/columns/checkbox-column.blade.php
```diff
@@ -1,16 +1,18 @@
 <div
-    x-data="{ error: undefined }"
+    x-data="{ error: undefined, state: @js($getState()), loading: false }"
```
danharrin: Please rename all `loading` Alpine property references to `isLoading` to indicate the boolean type

danharrin: Also format x-data properties on new lines with trailing commas please

packages/tables/src/Concerns/HasColumns.php
```diff
@@ -134,27 +129,14 @@ public function setColumnValue(string $column, string $record, $input): ?array
...
-        $record->setAttribute($columnName, $input);
-        $record->save();
+        $state = $column->evaluate($column->getSaveStateUsing(), [
+            'state' => $input,
+            'table' => $this->getCachedTable(),
+        ]);
 
-        return null;
+        return [
+            'state' => $state,
```
danharrin: Whats the reason to always return within this `state` key from this method?

danharrin: Cant we return the state without the array like previously?

packages/tables/src/Columns/SelectColumn.php
```diff
@@ -15,6 +15,9 @@ class SelectColumn extends Column implements Editable
     use Concerns\CanBeValidated {
         getRules as baseGetRules;
     }
+    use Concerns\CanSaveState {
```
danharrin: Please ensure that traits are alphabetical

packages/tables/src/Concerns/HasColumns.php
```diff
@@ -134,27 +129,14 @@ public function setColumnValue(string $column, string $record, $input): ?array
...
-            $columnName = (string) Str::of($columnName)->replace('.', '->');
-        }
-
-        if (! ($record instanceof Model)) {
-            return null;
-        }
-
-        $record->setAttribute($columnName, $input);
-        $record->save();
+        $state = $column->evaluate($column->getSaveStateUsing(), [
```
danharrin: I think that instead of doing this work in the HasColumns trait, we can just add a `$column->saveState($state)` method to the CanSaveState trait. Then we don't need the `getSaveStateUsing()` getter at all as we have access to `$this->saveStateUsing`.

packages/tables/src/Concerns/HasColumns.php
pxlrbt: We didn't return the sate at all previously. No sure why I used an array. Probably had something else in there and removed it.

pxlrbt:
@danharrin Not sure what's better: Have `saveStateUsing()` in the `Editable` interface to make sure we include that Trait or make it `protected` to ensure people don't accidentally use it

pxlrbt:
@danharrin I moved all the default `saveState` logic to that method.

danharrin:
Refactored this to use the existing `$livewire` accessible inside column objects, instead of passing in `$table`

PR #4991, opened 2022-11-18
Allow arbitrary depth dotted relations in TextColumn

cheesegrits:
Allow things like TextColumn::make('dailyLogs.pilots.name') or ('flightLegs.incidents.commander.name').

### DIFF
--- a/packages/tables/src/Columns/Concerns/CanFormatArrayState.php
+++ b/packages/tables/src/Columns/Concerns/CanFormatArrayState.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace Filament\Tables\Columns\Concerns;
+
+use Closure;
+use Illuminate\Support\HtmlString;
+
+trait CanFormatArrayState
+{
+    protected ?int $limit = null;
+
+    public function formatArrayStateUsing(?Closure $callback): static
+    {
+        $this->mutateArrayStateUsing = $callback;
+
+        return $this;
+    }
+
+    public function ul(): static
+    {
+        $this->mutateArrayStateUsing(function ($state) {
+            return new HtmlString(
+                '<ul><li>' . implode('</li><li>', $state) . '</li></ul>'
+            );
+        });
+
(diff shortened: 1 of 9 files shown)


cheesegrits:
*sigh* ignore the first commit message, apparently I just hit Commit with whatever was there from the last PR.

cheesegrits:
OK ... I refactored collectRelationValues() to make it cleaner and a little more obvious what it does:

Explode the dotted relations into an array and pass it to collectRelationsValues(), which shifts the first relation off the stack.  If only one remains on the stack (and hence the actual attribute), we're done, return the collection for that relation as an array.  If more than one, iterate through records for the relation, recursing, merging the returned arrays.

It doesn't need to check for relation types, as that doesn't matter.  Whatever the relation is, it will yield a collection, which we iterate and recurse on.  I've tested it with BelongsTo, HasOne, HasMany, BelongsToMany in arbitrary order up to a depth of 4.  Being relationship agnostic and recursive, it will handle literally any depth and combination of relations, including 3rd party ones like Staudenmeir's HasManyDeep.

And to wrap it up, I had to add the CanFormatArrayState concern (initially with ul(), ol() and grid()),  so I can render my distant relations nicely.  I have some with commas in them, so I can't use the TextColumn formatStateUsing.

I had to add a Comments model to the test models for the 'can can render text column with deep relationship' test.

Here's a video of it in action.  A HasMany => BelongsTo, and a BelongsTo => BelongsTo => BelongsTo => BelongsToMany.

```php
            TextColumn::make('flightLegs.incident.incident_number')
                ->ul()
                ->label('Incidents'),
              
            TextColumn::make('dailyLog.aircraft.aircraftType.missionCodes.name')
                ->grid(2,2),
```

https://www.screencast.com/t/XoYg1exa

cheesegrits:
Closing this one on v2, resubmitting on v3 with more changes.  Too many code changes for a v2 feature.

PR #4995, opened 2022-11-18
Decouple notification test helper from livewire

coleshirley:
As discussed in the filament discord: https://discord.com/channels/883083792112300104/883083792653381695/1043230796103811122

I first attempted to utilize a Trait that could be utilized in `Pest\Plugin::uses()` and in the `TestsNotifications` class but the `Livewire::mixin()` function is not compatible with a trait. So I opted to create a static function on the `Notification` class instead.

I didn't see any tests in the current test suite for the `assertNotified` test helper so I didn't add any of my own for the new helper methods. Happy to add them if needed


danharrin:
Thanks! I ended up removing the trait as I think having the mixin trait aside it probably would cause confusion. I've utilised the `assertNotified()` inside the Pest plugin.

PR #4998, opened 2022-11-19
Allow empty modal subheading

martin-ro:
Currently, when using `->requiresConfirmation()` for an action modal, removing the subheading is impossible.
You either have the `Are you sure you would like to do this?` subheading or define your own.

This PR allows setting the `->modalSubheading()` to an empty string or false to hide it.

The previous/default behavior should not be impacted by this PR.


martin-ro:
Can this and https://github.com/filamentphp/filament/pull/4999 be merged or are there any concerns?

danharrin:
Yes I had to refactor this for consistency - you can now call `->requiresConfirmation()->modalSubheading(null)`.

PR #5003, opened 2022-11-19
Select column option assertions

titantwentyone:
Two assertions to check options against a `SelectColumn`:

```
assertSelectColumnHasOptions(string $name, array $options, $record)
assertSelectColumnDoesNotHaveOptions(string $name, array $options, $record)
```

### DIFF
--- a/packages/tables/.stubs.php
+++ b/packages/tables/.stubs.php
@@ -121,6 +121,10 @@ namespace Livewire\Testing {
 
         public function assertTableColumnDoesNotHaveDescription(string $name, $description, $record, $position = 'below'): static {}
 
+        public function assertSelectColumnHasOptions(string $name, array $options, $record): static {}
+
+        public function assertSelectColumnDoesNotHaveOptions(string $name, array $options, $record): static {}
+
         public function sortTable(?string $name = null, ?string $direction = null): static {}
 
         public function searchTable(?string $search = null): static {}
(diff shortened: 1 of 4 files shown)


danharrin:
Confused why the `$record` param is required, if options are the same for the whole table?

danharrin:
Well I guess you don't need to pass a record actually...

Thanks

titantwentyone:
Thanks Dan. Since `options` can take a `Closure` it is possible to change the options based on some attribute of the `record`? Just tested this out with the following:

```
Tables\Columns\SelectColumn::make('dependent_options')
    ->options(fn($record) => $record->is_published ? ['published' => 'Published'] : ['unpublished' => 'Unpublished'])
```

```
$post_published = Post::factory()->create([
    'is_published' => true
]);

$post_unpublished = Post::factory()->create([
    'is_published' => false
]);

livewire(PostsTable::class)
    ->assertSelectColumnHasOptions('dependent_options', ['published' => 'Published'], $post_published)
    ->assertSelectColumnDoesNotHaveOptions('dependent_options', ['unpublished' => 'Unpublished'], $post_published)
    ->assertSelectColumnHasOptions('dependent_options', ['unpublished' => 'Unpublished'], $post_unpublished)
    ->assertSelectColumnDoesNotHaveOptions('dependent_options', ['published' => 'Published'], $post_unpublished);
```

PR #5011, opened 2022-11-20
Fix for non-existent method call for getting relation title

cheesegrits:


danharrin:
Just fixed this myself, also needed to pass in the owner record & page class

PR #5012, opened 2022-11-20
Fix: colors in preset to use tailwind 3.2 format for rgba color strings

awcodes:


packages/app/resources/css/theme.css
```diff
@@ -14,8 +14,8 @@
         background-image: radial-gradient(
             circle at top,
             theme('colors.primary.800'),
-            theme('colors.gray.800'),
```
danharrin: I believe this should remain as gray. Secondary could be a non-gray color

PR #5013, opened 2022-11-20
Fall back to default Tailwind fonts

zepfietje:


awcodes:
This doesn't work. The fallbacks have to be part of the --font-family custom property.

zepfietje:
Why so, @awcodes?
Works fine on my end.

awcodes:
When tailwind compiles this it will set the value of body font-family in preflight to be 'var() ... default fonts' which isn't a valid value for font-family in css. 

So to make it work the default fonts need to be appended to the base layout where --font-family is being set.

zepfietje:
Could you share how to reproduce the exact issue you're running into?

I'm getting this just fine:
```css
html {
    -webkit-text-size-adjust: 100%;
    font-feature-settings: normal;
    font-family: var(--font-family),ui-sans-serif,system-ui,-apple-system,BlinkMacSystemFont,"Segoe UI",Roboto,"Helvetica Neue",Arial,"Noto Sans",sans-serif,"Apple Color Emoji","Segoe UI Emoji","Segoe UI Symbol","Noto Color Emoji";
    line-height: 1.5;
    tab-size: 4;
}
```

Also, it correctly loads whatever font I configure.

awcodes:
Hmm. Didn't work when I first tried it. The font family showed an error for me in devtools saying it wasn't valid.

zepfietje:
Not seeing any errors. Tested in Safari and Chromium. Should be ready to merge I think, @danharrin.

awcodes:
I was on Firefox. But sure. Easy enough to revert if needed.

zepfietje:
Oh sorry, assumed you were on Chrome. Just tested on Firefox, works fine there as well.

zepfietje:
@danharrin, can you merge this PR?

PR #5023, opened 2022-11-21
Default group behavior

cheesegrits:
Fixes an issue where if you specify groups (so there's a dropdown grouping menu) and a default group, the page loads grouped by the default group (as one would expect) but it isn't selected on the dropdown, and you can't ever "ungroup" (by selecting the not-a-group "Group By" option).

This PR makes sure on boot that if the default group is also in the groups dropdown options, it gets selected.  And unselecting from the Group By dropdown will ungroup.

However ... I think there is a need for another option, like groupingRequired(), or perhaps a second bool arg to defaultGroup($groupId, $required), which forces the table to always be grouped.  So if not grouping required there's a "None" option on the dropdown, if required there isn't.

There is also a case to be made for adding the default group to groups, if groups are specified and default group isn't in them.

Thoughts?

### DIFF
--- a/packages/tables/resources/views/components/groups.blade.php
+++ b/packages/tables/resources/views/components/groups.blade.php
@@ -22,7 +22,7 @@
             </span>
 
             <x-filament::input.select wire:model="tableGrouping" class="w-full">
-                <option>-</option>
+                <option value="">-</option>
                 @foreach ($groups as $group)
                     <option value="{{ $group->getId() }}">{{ $group->getLabel() }}</option>
                 @endforeach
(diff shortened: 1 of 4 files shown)


danharrin:
I've just pushed ec85e5119f82643f4a25e5ce60881791a94b3922 that fixes this in an alternative way - if the default group is ALSO specified in the `groups()` array then it is deselectable, otherwise it is always applied. I think it handles all cases without the need for additional config options for required groups?

danharrin:
Let me know if it doesnt solve

PR #5032, opened 2022-11-21
Docs: added information on mountAction method to call page actions

mikrosmile:


zepfietje:
Do we want to document this approach, @danharrin?

Will have to update the wording and simplify it a bit.

mikrosmile:
> Do we want to document this approach, @danharrin?
> 
> Will have to update the wording and simplify it a bit.

I spent so many time figuring it :) really needed option to "call" page actions not just clicking a button itself ..

zepfietje:
You could achieve that by calling a Livewire component method from your action: `Action::make('delete')->action('delete')`. Then you can call the method from anywhere (also using listeners).

I think that's actually the preferred approach. Checking with @danharrin if this PR's approach is actually recommended.

ArtMin96:
I also spent a lot of time checking the code to find this. I had the following problem: I had to open the action modal and fill the form when I clicked on the button and passed the id of the record.

```html
<x-filament::button wire:click="functionName({{ $record->id }})"></x-filament::button>
```
```php
public function functionName(int $id)
{
    $record = Model::find($id);

    $this->mountAction('action_name');
    $this->getMountedActionForm()->fill($record->withoutRelations()->toArray());
}
```

mikrosmile:
> I also spent a lot of time checking the code to find this. I had the following problem: I had to open the action modal and fill the form when I clicked on the button and passed the id of the record.
> 
> 
> 
> ```html
> 
> <x-filament::button wire:click="functionName({{ $record->id }})"></x-filament::button>
> 
> ```
> 
> ```php
> 
> public function functionName(int $id)
> 
> {
> 
>     $record = Model::find($id);
> 
> 
> 
>     $this->mountAction('action_name');
> 
>     $this->getMountedActionForm()->fill($record->withoutRelations()->toArray());
> 
> }
> 
> ```

Does it works for you? I mean the ->fill part?

ArtMin96:
> > I also spent a lot of time checking the code to find this. I had the following problem: I had to open the action modal and fill the form when I clicked on the button and passed the id of the record.
> > ```
> > <x-filament::button wire:click="functionName({{ $record->id }})"></x-filament::button>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > public function functionName(int $id)
> > 
> > {
> > 
> >     $record = Model::find($id);
> > 
> > 
> > 
> >     $this->mountAction('action_name');
> > 
> >     $this->getMountedActionForm()->fill($record->withoutRelations()->toArray());
> > 
> > }
> > ```
> 
> Does it works for you? I mean the ->fill part?

Yes

danharrin:
Yeah I would really rather not add this to the docs as its an internal method that could be changed in the future. You're welcome to use it at your own discretion and thank you for the contribution

PR #5045, opened 2022-11-23
Dispatch events during the table reorder operations

lhilton:
We need to be able to run operations either before or after the table reordering operations run. This PR adds a before and after event to `Filament\Tables\Concerns\CanReorderRecords::reorderTable()` and passes the owning record as part of the event.

This mirrors a similar PR submitted against [spatie/eloquent-sortable](https://github.com/spatie/eloquent-sortable/pull/135) (in case that context is important for this PR).

### DIFF
--- a/packages/tables/src/Concerns/CanReorderRecords.php
+++ b/packages/tables/src/Concerns/CanReorderRecords.php
@@ -17,7 +17,7 @@ trait CanReorderRecords
             return;
         }
 
-        event(new TableSorting($this->ownerRecord));
+        event(new TableSorting($this->getOwnerRecord()));
 
         $orderColumn = $this->getTableReorderColumn();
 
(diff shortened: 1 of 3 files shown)


zepfietje:
Thanks for the PR, @lhilton!

I think we'd rather have hook methods to configure instead of dispatching events. What do you think, @danharrin?

danharrin:
Yeah, I feel like this should be on the table-level so you can hook for 1 table and not all tables at once. Are you able to handle that @lhilton? We just need new protected `beforeTableRecordSorted()` and `afterTableRecordSorted()` methods? Also, I don't think that passing the owner record is appropriate here, that's only available in relation managers...

danharrin:
Hey @lhilton just checking in to see if you're still intending to work on this?

PR #5052, opened 2022-11-24
Enable Customizing imageResizeUpscale

yob-yob:
This will allow devs to disable image upscaling when using the FilePond ImageResize Feature.

I need this because I was literally lost on why users uploading 3MB image are getting errors saying that the image they uploaded exceeds 10MB, and after 500 years (2 hours) of debugging I found out it's because of this.

Please accept this 😭 

Should I create tests for this be accepted?

### DIFF
--- a/packages/forms/src/Components/FileUpload.php
+++ b/packages/forms/src/Components/FileUpload.php
@@ -193,7 +193,7 @@ class FileUpload extends BaseFileUpload
         return $this->evaluate($this->imageResizeMode);
     }
 
-    public function getImageResizeUpscale(): ?bool
+    public function getImageResizeUpscale(): bool
     {
         return $this->evaluate($this->imageResizeUpscale);
     }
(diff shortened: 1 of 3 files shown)


danharrin:
I've checked the Filepond docs, should this be a bool instead of a string setting?

yob-yob:
sorry about that.. I'll change it to bool...

danharrin:
Have you tested this?

packages/forms/src/Components/FileUpload.php
```diff
@@ -184,6 +193,11 @@ public function getImageResizeMode(): ?string
         return $this->evaluate($this->imageResizeMode);
     }
 
+    public function getImageResizeUpscale(): ?string
```
danharrin: because this doesnt look right?

yob-yob: Sorry about that @danharrin... I change it to `bool`

PR #5075, opened 2022-11-26
3.x.column deep

cheesegrits:
Kind of a large PR, but was discussed in several meetings with Dan, who has committed to rewriting it to his liking, to implement the features it demonstrates.

Adds more comprehensive handling of distant relationships to Filament:

The ability to use distant relationships in columns, like TextColumn::make('posts.authors.name') or ('dailyLog.flightLegs.incidents.incident_name'), of any arbitrary depth or mix of relationship types (many to many, one to many, one to one, whatever), rendering the collected array of results using column methods like ul(), ol(), grid(), etc.

The ability to order/sort by an arbitrary length chain of one-to-one relationships, like 'comment.author.name'.

The ability to group by arbitrary length chains of one-to-one relationships.

The ability to use the Editable column feature on distant one-to-one chains.

The bulk of the documentation is in the main HasNestedRelationships trait, in ...

./packages/support/src/Concerns/HasNestedRelationships.php


danharrin:
Replaced by #5359.

PR #5079, opened 2022-11-27
Update html() with notice on formatting

damms005:


danharrin:
Can you explain why you would ever need to format HTML at the same time as the other formatters?

damms005:
I have a MySQL JSON column that stores an array of uploaded files. This `uploaded_files` column is [cast to array](https://laravel.com/docs/9.x/eloquent-mutators#array-and-json-casting). The requirement is to display the links to each of those uploaded files.

Before I started using Filament, I display these records on a paginated table by iterating through `Forms::all()`. For the `uploaded_files` column, I iterate through each file (recall now cast to array) and dynamically generate links to the file:

```
<a href="{url($form->uploaded_files[$index])}"> {{$form->uploaded_files[$index]}} </a>
```

Switching to Filament, for the array-cast `uploaded_files` column, from reading the docs I expected to use `formatStateUsing()` to dynamically generate those links like above and use `html()` to prevent raw HTML tags from being displayed.

This caused `Array to string` exception because `html()` internally uses `Str::of(...)` whose argument expects a string, but my column is cast to array.

The way I currently handle this is to check the implementation of `html()`, and then remove call to `html()` and then make `->formatStateUsing()` return `Str::of(implode('', $urls))->toHtmlString()`

pxlrbt:
We would need to add this to every function that uses `->formatStateUsing()` under the hood then.

damms005:
Alternatively, we can have a new section called `Formatters` and immediately under this section, we have this notice:

> **Warning**
> All the methods in this section are state formatters. You should only use only one of these formatters at a time on the same TextColumn. This is because they internally use the same `formatStateUsing()` API, which may cause one to overwrite the other.

Thoughts?

pxlrbt:
Would be a good solution if people would read docs from top to bottom 😅 But I doubt it.

danharrin:
I think people should reach for view/custom columns if they want to render HTML inside links etc, like @damms005. I believe that you would open your app up to XSS attacks if you rendered the URL etc without proper sanitisation.

PR #5083, opened 2022-11-28
Rules for each item in array fields

tanthammar:
It is a bit cumbersome to add validation for each item in array fields like Multiselect, CheckboxList and Tags fields.

This PR adds `->rulesForeachItem()` to any field that uses the provided `HasRulesForeachItem`. (Only applicable to fields of Array type.)

The PR adds the trait to `Select`, `CheckboxList` and `TagsInput`. 

### Example belongsToMany

```php
//Instead of this:
CheckboxList::make('teams')->label(__('fields.team'))
    ->relationship('teams', 'name')
    ->default([user()->current_team_id])
    ->options($allowedTeams = user()->ownedTeams()->pluck('name', 'id'))
    ->required()
    ->rules(['bail', 'array', function ($value, $fail) use ($allowedTeams) {
        foreach ($value as $id) {
            if (!in_array($id, $allowedTeams->keys()->toArray(), false)) {
                $fail("The selected record with id: $id, does not belong to you.");
        }
    }
}])

//You can:
CheckboxList::make('teams')->label(__('fields.team'))
    ->relationship('teams', 'name')
    ->default([user()->current_team_id])
    ->options($allowedTeams = user()->ownedTeams()->pluck('name', 'id'))
    ->required()
    ->rule('array')
    ->rulesForeachItem(Rule::in($allowedTeams->keys()->toArray()))

//Or if you would use a Select (multiSelect)
Select::make('teams')->label(__('fields.team'))
    ->multiple()
    ->relationship('teams', 'name')
    ->default([user()->current_team_id])
    ->options($allowedTeams = user()->ownedTeams()->pluck('name', 'id'))
    ->required()
    ->rule('array')
    ->rulesForeachItem(Rule::in($allowedTeams->keys()->toArray()))

```

The `rule()` or `rules()` methods applies the rules to the array itself whereas `rulesForeachItem()` applies to each item in the array. 
The previous example would generate rules for both:

```php

"data.teams" => array:2 [▼
    0 => "required"
    1 => "array"
  ]
  "data.teams.*" => array:1 [▼
    0 => "in:"2","3","4","5""
  ]
```

### Example TagsField

Sometimes you want to control that people don't add junk-tags to db. Without this PR it would be a long `foreach(...)` to cover for all these validation rules.

```php
TagsInput::make('tags')
    ->required()
    ->rule('array')
    ->rulesForeachItem(['bail', 'min:3', 'max:20', Rule::notIn(['A long list with banned words', 'Foo', 'Bar', 'S**t'])]),
```

I have yet to dig deep into Filament Spatie Tags, I guess it would also benefit from the `HasRulesForeachItem` Trait, if one wanted to apply the rules in the example?


tanthammar:
When/If this PR is merged, here's an example of macros for anyone to add to some ServiceProvider.

```php
TagsField::macro('ruleEachInOptions', fn(): static => $this->rulesForeachItem(['bail', fn($component): \Illuminate\Validation\Rules\In => Rule::in(array_keys($component->getOptions()))]));

Select::macro('ruleEachInOptions', fn(): static => $this->rulesForeachItem(['bail', fn($component): \Illuminate\Validation\Rules\In => Rule::in(array_keys($component->getOptions()))]));

CheckboxList::macro('ruleEachInOptions', fn(): static => $this->rulesForeachItem(['bail', fn($component): \Illuminate\Validation\Rules\In => Rule::in(array_keys($component->getOptions()))]));
```

danharrin:
I've renamed it to `nestedRecursiveRules()`

PR #5087, opened 2022-11-29
Include aggregated summarized columns in filter query

cheesegrits:
In getFilteredTableQuery(), in isGroupsOnly() mode, we need to include any columns that use relationship aggregating functions like sum(), that are also summarized, as these columns are based on computed attributes which won't otherwise get included in the base query, which blows up the summarizer methods.

I don't know what's going on with the formatting, btw.  I even set up a pint file watcher, but somehow it still changes a crapload of code formatting. Although perhaps the existing formatting was different from current project standards.

### DIFF
--- a/packages/tables/.stubs.php
+++ b/packages/tables/.stubs.php
@@ -1,296 +1,153 @@
 <?php
 
 namespace Livewire\Testing {
+
     use Illuminate\Support\Collection;
 
     class TestableLivewire
     {
-        public function mountTableAction(string $name, $record = null): static
-        {
-        }
-
-        public function setTableActionData(array $data): static
-        {
-        }
-
-        public function assertTableActionDataSet(array $data): static
-        {
-        }
-
-        public function callTableAction(string $name, $record = null, array $data = [], array $arguments = []): static
-        {
-        }
-
-        public function callTableColumnAction(string $name, $record = null): static
-        {
-        }
-
-        public function callMountedTableAction(array $arguments = []): static
-        {
-        }
-
-        public function assertTableActionExists(string $name): static
-        {
-        }
-
-        public function assertTableActionDoesNotExist(string $name): static
-        {
-        }
-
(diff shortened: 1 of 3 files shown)


zepfietje:
Pint doesn't have a rule for spacing around `|`. Will have to be manually formatted.

Edit: actually a rule does exist, looking into it now.
Edit 2: merge 3.x into this PR and formatting should get fixed by the bot. 😊
Edit 3: fixed it for you already.

cheesegrits:
> Pint doesn't have a rule for spacing around `|`. Will have to be manually formatted.
> 
> Edit: actually a rule does exist, looking into it now. Edit 2: merge 3.x into this PR and formatting should get fixed by the bot. 😊 Edit 3: fixed it for you already.

Tx.

BTW, I'm running pint locally with your rules, but it formatted that stubs file with empty function braces on new lines.  I had to turn pint off to fix that.

Can you check that?

zepfietje:
> BTW, I'm running pint locally with your rules, but it formatted that stubs file with empty function braces on new lines. I had to turn pint off to fix that.
> 
> Can you check that?

Hmm, that's weird. Running Pint locally should give the same result as CI. Are you running the actual CLI command or an IDE plugin?

cheesegrits:
> > BTW, I'm running pint locally with your rules, but it formatted that stubs file with empty function braces on new lines. I had to turn pint off to fix that.
> > Can you check that?
> 
> Hmm, that's weird. Running Pint locally should give the same result as CI. Are you running the actual CLI command or an IDE plugin?

I'm running the CLI command as a file watcher, so after closing / saving a file it runs that file through the command line.

That's how that stubs file got damaged.  It had gotten opened when I globally searched for a method name, then pint helpfully modified it when it got closed.

zepfietje:
Might it be that CI Pint doesn't format the stubs file?

PR #5097, opened 2022-11-29
fix: Non-inline radio button spacing

pxlrbt:
Non-inline radio buttons don't use `flex` and therefore `gap-2` has no effect.

**Before:**
image

**After:**
image

### DIFF
--- a/packages/forms/resources/views/components/radio.blade.php
+++ b/packages/forms/resources/views/components/radio.blade.php
@@ -24,7 +24,7 @@
                 :is-grid="! $isInline()"
                 direction="column"
                 :attributes="$attributes->merge($getExtraAttributes())->class([
-                    'filament-forms-radio-component flex flex-wrap gap-x-3 gap-y-2',
+                    'filament-forms-radio-component flex flex-wrap gap-3',
                     'flex-col' => ! $isInline(),
                 ])"
             >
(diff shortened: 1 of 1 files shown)


packages/forms/resources/views/components/radio.blade.php
```diff
@@ -26,7 +26,7 @@
                 :attributes="$attributes->merge($getExtraAttributes())->class([
                     'filament-forms-radio-component',
                     'flex flex-wrap gap-3' => $isInline(),
-                    'gap-2' => ! $isInline(),
+                    'space-y-2' => ! $isInline(),
```
zepfietje: What about using `flex flex-col`?

pxlrbt: Why introduce 2 classes if this can be solved by changing one?

zepfietje: `flex` can be moved to outside the conditionals as it's also used for inline. Makes it easier to maintain if we're using flex and gap instead of margins in half of the places.

pxlrbt: Should we increase the gap to `gap-3`? Then we could move that outside the conditional, too.

zepfietje: Not sure how that would be visually. Could you share a screenshot comparison?

zepfietje: Gap could also be moved outside conditional using `gap-x-3` and `gap-y-2`.

zepfietje: Maybe `flex-wrap` could also even be moved outside.

pxlrbt: **Before:**
image

**After:**
image

zepfietje (approved):

PR #5113, opened 2022-11-30
Improved style handling for summary rows

cheesegrits:
Summary columns need to have the same alignment as the column they summarize.  Most common example is right aligned numeric or money values.  if the column is right aligned, the summary needs to be as well.

I put this on the main summaries row template, so it gets added to the wrapper of any summary type.  Not sure if it needs to be here, or in the individual summary type templates, like summary/text.blade.php.  Let me know if you think it should be punted down in to the individual types.

[EDIT - see comment below, changed it so Summaries share header alignment and extra header attributes with parent column, but have their own extra attributes, also added row classes]

Before ...

After ...


cheesegrits:
PS ... I snuck adding a 'filament-tables-summary-row' class to the group summary rows in to this PR, so they can be styled separately from filament-tables-row.  Didn't seem worth opening a separate PR for it.

cheesegrits:
I've added in having the summaries inherit extraAttributes from their parent column.  Discussing best approach to take with Dan atm.  Will update if he decides he'd prefer Summarizer to have its own extraAttributes trait.

cheesegrits:
OK, I think this one is ready to go.  Spoke to Dan, changed it so Summarizer has it's own extraAttributes, which get pulled in to the individual summarizer templates.

For the summary header rows, it inherits its parent column's header wrapping and header attributes (as summaries don't have their own headers, it's just a repeat of the column header).

Each type of summary row now has it's own additional class (filament-tables-summary-header-row, filament-tables-summary-row, filament-tables-page-summary-row, filament-tables-total-summary-row) allowing for individual styling of each row type.

danharrin:
Thanks, I removed a text-center that had been added to all summary header columns, that wasn't deliberate right?

cheesegrits:
> Thanks, I removed a text-center that had been added to all summary header columns, that wasn't deliberate right?

Actually it was, because the main titles are centered.  So if you don't center the summaries as well, they get offset from the main titles.  And if you wrap the headers, the main titles are wrapped centered, but the summary titles are wrapped left aligned.

danharrin:
Can you provide screenshots?

cheesegrits:
> Can you provide screenshots?

The only styling we've done is colors, and applying wrapHeader().

danharrin:
Surely this should just inherit the column alignment config? Not sure why it would be fixed as text-center, because some columns are wide and the heading should be left aligned

cheesegrits:
The main headers are wrapped in a button, which all browsers (afaik) apply text-align=center to.  Chrome, Safari and Firefox do, anyway.

cheesegrits:
While we're on the subject ... having everything wrapped in a button in the headers seems to make extraHeaderAttributes less useful than it could be, because it applies the attributes to the th, not the button.  Although I'm not a CSS guy, so maybe there's a way of making those attributes apply to the button, I dunno.

danharrin:
But here, and in normal HTMl tables, headers are aligned with their content

Screenshot 2022-12-02 at 15 38 58

PR #5116, opened 2022-12-01
Feature: apply table filters by button

sadegh19b:
When we use table filters, this worked instantly after every change in filter fields and it's fine. but in some case we prefer to fill all filter fields and then click on button to apply filters. That's why I decided to sent this PR. I hope this is useful.

photo_2022-12-01_06-23-51

### Apply filters by button

To apply the table filters after click apply button, use the `shouldApplyTableFiltersByButton()` method: _(Should be used in list page of resource)_

```php
protected function shouldApplyTableFiltersByButton(): bool
{
    return true;
}
```


zepfietje:
I'm not sure about the UX on this one. How does it handle closing the filters panel before you've clicked the submit button?

sadegh19b:
> I'm not sure about the UX on this one. How does it handle closing the filters panel before you've clicked the submit button?

@zepfietje Closing the Filters panel works just like before, click the Filters trigger or click outside the filters panel to close.

zepfietje:
What happens to the filter state then? Is it cleared or do the inputs still contain the unapplied filters?

pxlrbt:
I like the idea for more complex filter requirements or SQL queries. Not 100% sure about closing the modal:

1) I think it should not do things I don't explicitly tell it. So applying the filters without clicking the buttons is unexpected.
2) On the other hand this is probably mostly just a mistake since users are already selecting filters. So I think it's better UX if it also applies on modal close. We could make this an option now or later and let the devs decide.

sadegh19b:
> What happens to the filter state then? Is it cleared or do the inputs still contain the unapplied filters?

@zepfietje 
If we select the filters and then close the filters panel, the states remain. If we want to apply the filters, we have to click on the apply button and we can also reset the states by clicking the reset filters button, even if we haven't clicked the apply button.

> On the other hand this is probably mostly just a mistake since users are already selecting filters. So I think it's better UX if it also applies on modal close. We could make this an option now or later and let the devs decide.

@pxlrbt 
I agree with you, this is a nice feature for better UX. I think we can work on this feature later on a separate PR and I also recommend that this feature is better if it is optional.

zepfietje:
I don't think we should merge this PR in its current state. Closing the filters dropdown and retaining its state while not applying the filters can lead to weird UX situations.

sadegh19b:
> I don't think we should merge this PR in its current state. Closing the filters dropdown and retaining its state while not applying the filters can lead to weird UX situations.

@zepfietje Do you mean to reset the filter fields after closing the filter panel? if so, I will work on it.

zepfietje:
That's the better behavior I think if you have an explicit save button. I'm not yet sure about the entire UX though.

packages/tables/resources/lang/it/table.php
```diff
@@ -93,6 +93,10 @@
 
         'buttons' => [
 
+            'apply' => [
+                'label' => 'Applicare filtri',
```
wit3:
```suggestion
                'label' => 'Applica filtri',
```

sadegh19b:
@zepfietje Hi, I finally found time to work on this PR. Now the filter fields are reset if you change them and do not press the apply button after closing the filter panel.

danharrin:
We've just discussed this internally and we're gonna hold off on merging this. We're not 100% sure about the implementation and the UX of such a feature, and we haven't had many requests for it.

Thanks for your work on it though! We will revisit later if we choose to implement it.

husam-tariq:
> When we use table filters, this worked instantly after every change in filter fields and it's fine. but in some case we prefer to fill all filter fields and then click on button to apply filters. That's why I decided to sent this PR. I hope this is useful.
> 
> photo_2022-12-01_06-23-51
> 
> ### Apply filters by button
> To apply the table filters after click apply button, use the `shouldApplyTableFiltersByButton()` method: _(Should be used in list page of resource)_
> 
> ```
> protected function shouldApplyTableFiltersByButton(): bool
> {
>     return true;
> }
> ```
Excellent work 👍🏻

yemenpoint:
@sadegh19b Big thanks 👍 
Awesome code from awesome developer 💯💯 🌹

PR #5122, opened 2022-12-01
Add Latvian translations for admin panel and tables

CreepPork:
Adds Latvian translations for admin panel and tables. Also fixes a missing translation key in support.


danharrin:
Thanks

PR #5129, opened 2022-12-01
Revamp database notifications UI

zepfietje:
- [ ] Fix re-rendering notifications after marking as read.


danharrin:
Is this ready for review?

zepfietje:
Sort of, the task in description still needs to be completed. I couldn't find the fix, so maybe you have an idea?

danharrin:
What specifically isnt working?

zepfietje:
When marking database notifications as read, the notification component doesn't re-render as read.

danharrin:
You need to get the notification object and pass `->unread(false)` to it, when it is marked

zepfietje:
So need to explicitly set it instead of getting the status from the notification object?

zepfietje:
@danharrin, I guess I'm blind, but I really can't see why the component does render but doesn't update the notification styling when marking all as read.

danharrin:
I don't know how to fix but its coming from `configureTransitions()` in the notification JS Alpine Component. You are probably caching the styles for that element somewhere so that they end up not updating? But I don't know how that all works, you probably do.

If you cant fix, maybe mount the notification Alpine onto a wrapper of the actual notification element? so that the styles arent retained

zepfietje:
Hmm, how did you find it's caused by the `configureTransitions()` function?

There's no caching there.
It only retrieves the styles to use the correct `display` value (since it could be different if someone provides their own notification Blade component).

danharrin:
> Hmm, how did you find it's caused by the configureTransitions() function?

I removed it from the JS and the classes updated. Its those Livewire hooks

zepfietje:
Oh `configureAnimations` you meant instead of `configureTransitions`. 😅

danharrin:
i think its configureTransitions... you can check

YasinAshorian (approved):
👌

zepfietje:
Closing for now.

PR #5155, opened 2022-12-02
Preserve table sort while reordering

Morry981:
Hi,
this is my first ever contribution to a github project so i hope i'm doing all correct.

I'm using in my websites your really good package and i found out that when reordering default sorting back always to ascending.
I really appreciate if this code i've changed can be accepted or improved for making my projects easier.
I tested in a simple situation ->defaultSort('position', 'desc') and isTablePaginationEnabledWhileReordering() = true

Let me know
Thanks!

PS. i made a private function to retrive the new position two time in the code
but i don't know if it's a good practice or the right spot for packages code

### DIFF
--- a/packages/tables/docs/02-getting-started.md
+++ b/packages/tables/docs/02-getting-started.md
@@ -479,6 +479,17 @@ protected function isTablePaginationEnabledWhileReordering(): bool
 }
 ```
 
+### Preserve order sort direction while reordering
+
+To sort always in ascending while reordering you can use:
+
+```php
+protected function preserveTableOrderWhileReordering(): bool
+{
+    return false;
+}
+```
+
 ## Polling content
 
 You may poll table content so that it refreshes at a set interval, using the `getTablePollingInterval()` method:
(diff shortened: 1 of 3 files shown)


danharrin:
Hi, thanks for the contribution. Can you give an example use case where this would be useful please?

Morry981:
Yes of course!

Immagine 2022-12-03 155414
immagine_2022-12-03_165909231

```
Resource class
in static table() method
with options
->reorderable('position')
->defaultSort('position', 'desc')
```

This is the simple scenario (i changed the data because i'm not sure i can already share it)
With my coworker we're building a page with a list of project made with the donations, newest on top (position desc).
To not confuse the client, who use the admin panel, we also ordered the table with position desc for a better preview
but when reordering the default sort back to asc, changing the preview data, so the client told us that was annoying.
We tried to speak about that but he prefer if the problem can be solved.
My coworker solved with a default asc position and on adding new record position = 1 and increase all other records by one position

immagine_2022-12-03_170728281
immagine_2022-12-03_170745675

Also when isTablePaginationEnabledWhileReordering() set to true
if sort order back to asc the current page data will be different

I don't like extra code and more query to db
and i think is more natural solving directly mantain same sort order
(that's why i set to true preserveTableOrderWhileReordering())
or at least having this new option

Hope it's readable english and not annoying images ;)

danharrin:
Hi, thanks for the PR but I've decided that the behaviour should be consistent: while reordering, the records are always sorted in the order.

PR #5156, opened 2022-12-03
Add ability to merge extra (alpine/input/trigger/img/header) attributes instead of overwriting them

ralphjsmit:
I noticed that when I used the `->extraAttributes()` function twice on a component (for example indirectly inside a macro or function + directly using it when defining the component), that the old values would be overwritten. I didn't really expected this and I looked for a way to allow merging those attributes.

This PR introduces a second optional parameter that allows the developer to state that new attributes should be merged with the existing attributes. The merge is done using the `ComponentAttributeBag`, so that we mirror the default Laravel behaviour when merging attributes (e.g. merge classes, but overwrite `type` attribute).

I set this `$merge = false` parameter to false by default, since it would theoretically be a breaking change. Personally I'd like the parameter to be `true` by default, but that might be an idea for V3. :)

The only caveat is when using merging with existing attributes that are determined by a closure. I know directly evaluated those closures to merge the contents. There might also be a way around it, by constructing a custom Closure that evaluates the already existing attributes later, but that would probably quite increase the complexity of the code.

Thanks!

### DIFF
--- a/packages/forms/src/Components/Concerns/HasExtraInputAttributes.php
+++ b/packages/forms/src/Components/Concerns/HasExtraInputAttributes.php
@@ -12,7 +12,7 @@ trait HasExtraInputAttributes
     public function extraInputAttributes(array | Closure $attributes, bool $merge = false): static
     {
         if ($merge) {
-            $attributes = array_merge($this->getExtraInputAttributes(), $attributes);
+            $attributes = $this->getExtraInputAttributeBag()->merge($this->evaluate($attributes))->getAttributes();
         }
 
         $this->extraInputAttributes = $attributes;
--- a/packages/forms/src/Components/Concerns/HasExtraInputAttributes.php
+++ b/packages/forms/src/Components/Concerns/HasExtraInputAttributes.php
@@ -9,7 +9,7 @@ trait HasExtraInputAttributes
 {
     protected array $extraInputAttributes = [];
 
-    public function extraInputAttributes(array|Closure $attributes, bool $merge = false): static
+    public function extraInputAttributes(array | Closure $attributes, bool $merge = false): static
     {
         if ($merge) {
             $this->extraInputAttributes[] = $attributes;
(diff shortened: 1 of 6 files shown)


zepfietje:
This would evaluate the `extraAttributes` value every time you'd pass a new value. Not sure if that's desirable performance-wise.

ralphjsmit:
Performance-wise I don't think it's a huge issue, since the number of persons that will be using multiple closures in here will be probably very low. In addition, the default behaviour is still not to merge. So I think it's better to support a feature than not to support it just because there might be some very rare situations.

In addition, if people have closures that take a long time, that's probably more of an issue on their side.

danharrin:
I don't think this would work properly with closures, since they cannot be evaluated within setters.

This should be refactored so that we can store multiple sets of attributes (an array of arrays/closures) inside `extraAttributes()`. Then `array_merge` them all inside the getter, at once.

ralphjsmit:
Thanks Dan, good solution. Will update the implementations to make use of that solution.

ralphjsmit:
This one is ready for review @danharrin.

I chose not to use `array_merge()`, but use a temporary ComponentAttributeBag. This, because the ComponentAttributeBag has special behaviour on the `merge()` method, for attributes like `class` etc.

ralphjsmit:
I also removed the `Closure` type from the protected properties. This _theoretically_ could be a breaking change, because in theory people can extend the form components and have the `protected array $attributes` still overriden like `protected array|Closure $attributes`. I can also add keep the `Closure` type and mark that as ready for deletion in V3.

packages/forms/src/Components/Concerns/HasExtraInputAttributes.php
```diff
@@ -7,18 +7,28 @@
...
+        $temporaryAttributeBag = new ComponentAttributeBag();
+
+        foreach ($this->extraInputAttributes as $extraInputAttributes) {
+            $temporaryAttributeBag = $temporaryAttributeBag->merge($this->evaluate($extraInputAttributes));
+        }
+
+        return $temporaryAttributeBag->getAttributes();
     }
 
     public function getExtraInputAttributeBag(): ComponentAttributeBag
```
danharrin, line 34: /wtd Refactor this so that getExtraInputAttributeBag() is responsible for merging attributes and getExtraInputAttributes() just extracts them from that bag

danharrin:
Thanks!

PR #5157, opened 2022-12-03
Fixed missing set state path in MorphToSelect

cheesegrits:
MorphToSelect currently blows up because of the missing state path.

### DIFF
--- a/packages/forms/src/Components/MorphToSelect.php
+++ b/packages/forms/src/Components/MorphToSelect.php
@@ -53,12 +53,13 @@ class MorphToSelect extends Component
         $relationship = $this->getRelationship();
         $typeColumn = $relationship->getMorphType();
         $keyColumn = $relationship->getForeignKeyName();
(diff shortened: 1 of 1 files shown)


cheesegrits:
Added fix for the morph type $get selector, needs to be 'thingable.thingable_type', not just 'thingable_type'.  Without this fix, second selector will never show on a new record.

cheesegrits:
Aaaand ... data isn't submitting, something messing up in getState ... it's returning the array, like ['reportable']['reportable_type'] and ['reportable']['reportable_id'] rather than shifting the individual items up into the main $data state path.

Working on it.

cheesegrits:
Hmmm.  And, when creating a new record, ONLY the dotted path exists for $get(), when editing a record both paths exist (dotted and not dotted) and when saving, only the non-dotted.  So this is beyond my pay grade, there's something funky going on deep in the guts.

danharrin:
As mentioned on Discord, please explain "blows up if there is no state path"

PR #5114, opened 2022-11-30
Update MakeUserCommand to make it inline fillable

alvleont:
This makes the artisan command make:filament-user inline fillable by using

Captura de pantalla 2022-11-30 a la(s) 4 49 00 p m

The PR also includes the --help options description and the options can be non-used inline: in that case, each option not used there will be prompted.

PR #5180, opened 2022-12-06
Add arguments to MakeUserCommand

bumbummen99:
Add arguments to `make:filament-user` command to simplify running from script. Will use arguments or ask for input as it already does.


danharrin:
Duplicate of #5114

PR #5185, opened 2022-12-06
fix: Brand name wrapping when expanding sidebar

pxlrbt:
I think the refactoring of the sidebar reintroduced this animation papercut:

https://user-images.githubusercontent.com/22632550/205878498-ca917015-6797-4ac5-9658-8da92c54cb87.mov

### DIFF
--- a/packages/admin/resources/views/components/brand.blade.php
+++ b/packages/admin/resources/views/components/brand.blade.php
@@ -1,6 +1,6 @@
 @if (filled($brand = config('filament.brand')))
     <div @class([
-        'filament-brand text-xl font-bold tracking-tight whitespace-nowrap',
+        'filament-brand text-xl font-bold tracking-tight',
         'dark:text-white' => config('filament.dark_mode'),
     ])>
         {{ $brand }}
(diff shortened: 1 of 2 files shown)


packages/admin/resources/views/components/brand.blade.php
```diff
@@ -1,6 +1,6 @@
 @if (filled($brand = config('filament.brand')))
     <div @class([
-        'filament-brand text-xl font-bold tracking-tight',
+        'filament-brand text-xl font-bold tracking-tight whitespace-nowrap',
```
zepfietje:
```suggestion
        'filament-brand whitespace-nowrap text-xl font-bold tracking-tight',
```

zepfietje (changes requested):

danharrin:
Do we need to support multi-line brand names though?

zepfietje:
> Do we need to support multi-line brand names though?

Not sure. Might as well use ellipsis.

pxlrbt:
Not sure. I think multiline brand names would look weird. People can use a SVG if they need it.

pxlrbt:
I think the last time I solved this by adding a transition:

https://user-images.githubusercontent.com/22632550/205897328-bb4ad752-4b71-4313-a33f-3aa9fbe068b6.mov

pxlrbt:
Updated this to use a transition instead of `whitespace-nowrap`

zepfietje:
Hmm feels like a dirty hack to me to be honest.

danharrin:
Better than breaking multi-line brands. Thanks

PR #5193, opened 2022-12-06
Update 10-text-input.md

bennettblack:
Adding verbiage to clarify that validation errors won't display on standalone tables unless alpine-tooltip is installed. 

I did not include the additional verbiage found [here](https://filamentphp.com/docs/2.x/tables/columns/getting-started#tooltips) regarding Tippy, since Tippy is a dependency of ryangjchandler/alpine-tooltip


danharrin:
Thanks

PR #5194, opened 2022-12-06
Default currency in money format

fabpl:
default_currency isn't into config file but in an environment variable : @see https://github.com/akaunting/laravel-money/blob/cbc66d1dc457c169f6081e0ae6c661b499dad301/src/helpers.php#L11

### DIFF
--- a/packages/tables/src/Columns/Concerns/CanFormatState.php
+++ b/packages/tables/src/Columns/Concerns/CanFormatState.php
@@ -135,7 +135,7 @@ trait CanFormatState
             }
 
             if (blank($currency)) {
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks

pxlrbt:
This really should be a config option. You should never use `env()` as it's all `null` on cached configs.
Should we introduce a `money` config or try to fix this upstream?

pxlrbt:
I just create a PR for the package:

https://github.com/akaunting/laravel-money/pull/72/files

danharrin:
fair, I did consider that also

PR #5204, opened 2022-12-07
Add isViewSafe() to notification class

JohnnyMaynne:
Add isViewSafe method to Notification class. So that when extends from base Notification class , you can override isViewSafe to use the view() method to add your own view to notification

https://discord.com/channels/883083792112300104/1001249817961451540/1050016360316555274

### DIFF
--- a/packages/notifications/src/Notification.php
+++ b/packages/notifications/src/Notification.php
@@ -62,12 +62,20 @@ class Notification extends ViewComponent implements Arrayable
             'icon' => $this->getIcon(),
             'iconColor' => $this->getIconColor(),
             'title' => $this->getTitle(),
+            'view' => $this->getView(),
         ];
     }
 
     public static function fromArray(array $data): static
     {
         $static = static::make($data['id'] ?? Str::random());
+
+        $view = $data['view'] ?? null;
+
+        if (filled($view) && ($static->getView() !== $view) && static::isViewSafe()) {
+            $static->view($data['view']);
+        }
+
         $static->actions(
             array_map(
                 fn (array $action): Action | ActionGroup => match (array_key_exists('actions', $action)) {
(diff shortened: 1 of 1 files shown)


packages/notifications/src/Notification.php
```diff
@@ -62,12 +62,20 @@ public function toArray(): array
...
         ];
     }
 
     public static function fromArray(array $data): static
     {
         $static = static::make($data['id'] ?? Str::random());
+
+        $view = $data['view'] ?? null;
+
+        if (filled($view) && ($static->getView() !== $view) && static::isViewSafe()) {
```
zepfietje, line 75:
```suggestion
        if (filled($view) && ($static->getView() !== $view) && static::isViewSafe($view)) {
```

packages/notifications/src/Notification.php
```diff
@@ -62,12 +62,20 @@ public function toArray(): array
...
+            $static->view($data['view']);
```
zepfietje, line 76:
```suggestion
            $static->view($view);
```

packages/notifications/src/Notification.php
```diff
@@ -86,6 +94,11 @@ public static function fromArray(array $data): static
         return $static;
     }
 
+    protected static function isViewSafe(): bool
+    {
+        return false;
+    }
+
```
zepfietje, line 101: (Will also need to add a type docblock here for 3.x, @danharrin.)

```suggestion
    protected static function isViewSafe(string $view): bool
    {
        return $view === 'notifications::notification';
    }

```

zepfietje (changes requested):

zepfietje:
@JohnnyMaynne, what's the reason behind closing this PR? It actually is something we should add I think.

PR #5214, opened 2022-12-09
[FIX]: fix apply callback order to relationship query

moh-abdullah-it:
this pull request apply order query if user order query to relationship callback query:

1- CheckboxList:
```php
CheckboxList::make('sites')->relationship('sites', 'name', fn (Builder $query) => $query->oldest('id'));
```

2- SelectFilter:
```php
SelectFilter::make('site')->relationship('sites', 'name', fn (Builder $query) => $query->oldest('id'));
```


danharrin:
Thanks

PR #5215, opened 2022-12-09
Feat: Add sidebar width methods to context provider

awcodes:
Adds `->sidebarWidth()` and `->collapsedSidebarWidth()` to context provider and Filament facade to allow setting sidebar widths without having to use a custom theme to do so.


packages/app/src/Context.php
```diff
@@ -779,6 +783,30 @@ public function getSuccessColor(): array
         return $this->successColor ?? Color::Green;
     }
 
+    public function sidebarWidth(string $sidebarWidth): static
+    {
+        $this->sidebarWidth = $sidebarWidth;
+
+        return $this;
+    }
```
zepfietje:
```suggestion
    public function sidebarWidth(string $width): static
    {
        $this->sidebarWidth = $width;

        return $this;
    }
```

packages/app/src/Context.php
```diff
@@ -779,6 +783,30 @@ public function getSuccessColor(): array
...
+    {
+        return $this->sidebarWidth;
+    }
+
+    public function collapsedSidebarWidth(string $collapsedSidebarWidth): static
+    {
+        $this->collapsedSidebarWidth = $collapsedSidebarWidth;
+
+        return $this;
+    }
```
zepfietje:
```suggestion
    public function collapsedSidebarWidth(string $width): static
    {
        $this->collapsedSidebarWidth = $width;

        return $this;
    }
```

packages/forms/src/Commands/MakeFormCommand.php
```diff
@@ -41,7 +41,8 @@ public function handle(): int
             ->map(fn ($segment) => Str::lower(Str::kebab($segment)))
             ->implode('.');
 
-        $model = (string) str($this->argument('model') ?? ($this->option('edit') ?
+        $model = (string) str($this->argument('model') ?? (
+            $this->option('edit') ?
             $this->askRequired('Model (e.g. `Product`)', 'model') :
             $this->ask('(Optional) Model (e.g. `Product`)')
         ))->replace('/', '\\');
```
zepfietje, line 48: This is quite unreadable to me now. 😅

zepfietje (changes requested):

packages/forms/src/Commands/MakeFormCommand.php
awcodes: I didn't do anything to this file. Why is it in this?

awcodes: Maybe this was added by github actions.

zepfietje: Yeah seems like Pint. Weird it only picked up the changes in this PR. Let's leave it like this for now.

zepfietje (approved):
Code looks good to me.

danharrin:
Thanks

PR #5219, opened 2022-12-10
Fix: dark mode gradient background

Z3d0X:
Before this fix:
Screenshot 2022-12-10 at 3 27 27 PM

After this fix:
Screenshot 2022-12-10 at 3 29 25 PM

Note: This issue is only present when dark mode is disabled from the config and `darkMode: 'class',` is missing from the tailwind config

### DIFF
--- a/packages/admin/resources/css/app.css
+++ b/packages/admin/resources/css/app.css
@@ -43,7 +43,7 @@
     }
 
     .filament-login-page form::before {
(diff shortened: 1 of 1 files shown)


zepfietje:
I don't think this is an actual bug as the dark mode key should be added to the Tailwind config as per the docs: https://filamentphp.com/docs/2.x/admin/appearance#building-themes.

PR #5232, opened 2022-12-13
stats overview card: add description icon position

kamilst96:
I want to place description icon before description text in StatsOverviewWidget.

Screenshot:

I extended description icon setter:
`descriptionIcon(?string $icon, ?string $position = null)`

Is it acceptable, or should it be like this?
`descriptionIcon(?string $icon)`
`descriptionIconPosition(?string $position)`

I am also not sure what to do about duplicated dynamic component.

### DIFF
--- a/packages/admin/resources/views/components/stats/card.blade.php
+++ b/packages/admin/resources/views/components/stats/card.blade.php
@@ -48,13 +48,13 @@
                     default => 'text-gray-600',
                 },
             ])>
-                @if ($descriptionIcon && $descriptionIconPosition == 'before')
+                @if ($descriptionIcon && ($descriptionIconPosition === 'before'))
                     <x-dynamic-component :component="$descriptionIcon" class="w-4 h-4" />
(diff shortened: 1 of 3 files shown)


danharrin:
Thanks

zepfietje:
Isn't it better to explicitly name `after`?

danharrin:
after by default. the property usually returns null

zepfietje:
For buttons we do use `before` as default instead of `null` though.

danharrin:
feel free to change it if you want

gldtn:
Sorry to post here guys, but this is the closest thing I found while searching, how can I set description icon position in filament 3? I tried passing to the stat ->descriptionIconPosition, tried something like `protected static ?string $descriptionIconPosition = 'before';`

but can't find any documentation on this

danharrin:
second param of descriptionIcon()

gldtn:
Thank you @danharrin!

PR #5234, opened 2022-12-13
Fix: Replace deprecated ${...} syntax in TestsForms.php

dbohn:
When testing our filament app with PHP 8.2, I noticed, that a warning was emitted by PHP, because of the usage of the now deprecated `${var}` syntax.
This MR swaps the position of the `$` character.
Other than that, I did not notice any errors or notices because of the update to PHP 8.2!

### DIFF
--- a/packages/forms/src/Testing/TestsForms.php
+++ b/packages/forms/src/Testing/TestsForms.php
@@ -153,7 +153,7 @@ class TestsForms
             Assert::assertInstanceOf(
                 Field::class,
                 $field,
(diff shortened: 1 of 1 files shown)


zepfietje:
Thanks for reporting, @dbohn!

I'm pretty sure this must've been a typo, since later on that line it does use the non-deprecated interpolation syntax.

danharrin:
Yeah, it was probably just me typoing with the JS syntax 😭 why does it have to be different...

PR #5242, opened 2022-12-13
Move uploaded file if disk remote matches

coleshirley:
This PR adds the ability for the user to have a uploaded file "moved" on save instead of read and written to its destination.

This significantly improves the performance of large file uploads to a remote S3 server since the files do not need to be downloaded and re-uploaded by the laravel application


danharrin:
Thanks

castellani8:
gj!

PR #5244, opened 2022-12-13
fix dateComparisonRule logical problem

ariaieboy:
According to the doc the dateComparisonRules such as `before` or `after`, etc should handle both date and another field comparison but right now it's not supporting date comparison unless you pass `true` as the second argument of these methods.


danharrin:
Thanks

PR #5258, opened 2022-12-14
Update 05-viewing-records.md

nchatzidakis:
Update wrong parameter in command.


zepfietje:
@nchatzidakis, what's wrong about the command? It should just be `--view`.

PR #5262, opened 2022-12-15
Fix: RichEditor debounce() not implemented

Capevace:
As a `Field`, the `debounce()` method can be called on the `RichEditor` field. 

However, no debounce was actually implemented anywhere in the blade view.

This PR adds debounce functionality to the `RichEditor` component. The implementation is similar to the `TextInput` one, but it listens to the `trix-change` event instead of `input` and sets the entangled `state` variable instead.

I tried to fix it using `$lazilyEntangledModifiers` of `$applyStateBindingModifiers()` too, but could not get it to work that way.

```php
// New debounce implementation
{!!
	$isDebounced()
		? "x-on:trix-change.debounce.{$getDebounce()}=\"state = \$event.target.value\""
		: 'x-on:trix-change="state = $event.target.value"'
!!}
```

### DIFF
--- a/packages/forms/resources/views/components/rich-editor.blade.php
+++ b/packages/forms/resources/views/components/rich-editor.blade.php
@@ -15,7 +15,11 @@
         x-data="richEditorFormComponent({
             state: $wire.{{ $applyStateBindingModifiers('entangle(\'' . $getStatePath() . '\')') }},
         })"
-        x-on:trix-change="state = $event.target.value"
+        {!!
+            $isDebounced()
+                ? "x-on:trix-change.debounce.{$getDebounce()}=\"state = \$event.target.value\""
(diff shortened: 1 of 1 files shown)


danharrin:
Have you tested this to ensure there is no data loss? I can imagine this could result in the end of the string being lost, if you type too quickly & the event listener is debounced. For example:

I type

`hel`

and the event listener runs

and then less than 200ms later I type

`lo`

but because its debounced, the event listener is skipped and the value is never saved to the state?

Have you tested this case?

Capevace:
Hey, thanks for having a look!

I investigated this a little further and found that data loss is indeed possible.
I have a messy fix working, but it's using inline state and functions to keep track of changes.

I'll have another look at this over the weekend, and will submit a cleaned up and tested implementation then.

PR #5270, opened 2022-12-16
Add morphToMany to reorderTable method

howdu:


danharrin:
As PHPStan says, I don't think this is correct? MorphToMany extends BelongsToMany, so the check is obsolete?

Please give a clearer indication as to what this does and we can reopen.

PR #5277, opened 2022-12-17
Add Closure type to the Tabs::tabs method parameter

rompetomp:
I noticed that the method `Tabs::make()->tabs()` only allows an array, while it passes the call through to `childComponents`, which can also allow a `Closure`.

If, for any reason, this was intentional, please disregard this PR.


danharrin:
Thanks

PR #5278, opened 2022-12-17
Fix: Global search hidden action

ArtMin96:
Global search action does not hide when hidden is true.

### DIFF
--- a/packages/admin/resources/views/components/global-search/actions/index.blade.php
+++ b/packages/admin/resources/views/components/global-search/actions/index.blade.php
@@ -4,8 +4,6 @@
 
 <div {{ $attributes->class('filament-global-search-actions mt-4 flex gap-3') }}>
     @foreach ($actions as $action)
-        @if(! $action->isHidden())
-            {{ $action }}
-        @endif
+        @unless ($action->isHidden())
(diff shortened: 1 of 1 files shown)


packages/admin/resources/views/components/global-search/actions/index.blade.php
```diff
@@ -4,6 +4,8 @@
 
 <div {{ $attributes->class('filament-global-search-actions mt-4 flex gap-3') }}>
     @foreach ($actions as $action)
-        {{ $action }}
+        @if(! $action->isHidden())
```
zepfietje:
```suggestion
        @unless ($action->isHidden())
```

packages/admin/resources/views/components/global-search/actions/index.blade.php
```diff
@@ -4,6 +4,8 @@
...
+            {{ $action }}
+        @endif
```
zepfietje:
```suggestion
        @endunless
```

zepfietje (changes requested):

ArtMin96:
Ready

danharrin:
Thanks

PR #5287, opened 2022-12-19
Add information on creating an action link for a custom page

SiCoUK:
I was reading both the admin and table builder documents on actions, and getting confused with the linking mechanics. The table builder uses the Laravel route() helper in actions and this document uses the resource.

I found the answer here: https://github.com/filamentphp/filament/discussions/2282 as I am not the only one to have missed the difference I figured it might be handy in the docs!

### DIFF
--- a/packages/admin/docs/02-resources/10-custom-pages.md
+++ b/packages/admin/docs/02-resources/10-custom-pages.md
@@ -29,3 +29,10 @@ To generate a URL for a resource route, you may call the static `getUrl()` metho
 ```php
 UserResource::getUrl('sort');
 ```
+
+To use the new page in an action link in a table call the static `getUrl()` method on the resource class in a callback:
+
+```php
+Tables\Actions\Action::make('sort')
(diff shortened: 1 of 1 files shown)


pxlrbt:
Not sure what new information this adds to the docs since the "linking mechanics" is described right above what you added (`UserResource::getUrl('sort')`). Can you explain what was confusing about that?

SiCoUK:
I guess at the time what I was trying to suggest is closing the loop as a new comer. I.e. this is how you create the page and this is how you make the action to get to the page. It's more about knowing where to look to put the pieces together I guess.

May be a link to actions or a note to see the actions docs. It's just one of those things that once you know you know!

zepfietje:
Hmm, yeah I'm not sure why this should be added to the custom page docs specifically.
The action docs clearly explain how to set a URL for an action.

Closing this PR, but feel free to add a subtle link to the action docs if you think that's really missing.

Edit: v3 will probably make this a lot easier to wrap your head around as it will unify the actions API in a new Filament Actions package.

PR #5296, opened 2022-12-20
Add tooltip documentation to `TextColumn` field

mattlibera:
The documentation for `TextInputColumn` contains the snippet about requiring the `@ryangjchandler/alpine-tooltip` package, but the `TextColumn` has the `copyable()` method, which also requires this package. Thus the same snippet should be referenced on the docs for that column as well.


pxlrbt (approved):

danharrin:
Thanks

PR #5299, opened 2022-12-21
Fix pre-collapsed nav group feature

oddvalue:
This PR fixes the `->collapsed()` feature on nav groups. 

The filament JS runs on the login page which will initialize the persisted `collapsedGroups` setting as an empty array causing `if (localStorage.getItem('collapsedGroups') === null)` to always come out as false.


pxlrbt (approved):

danharrin:
Thanks

PR #5300, opened 2022-12-21
Fix Repeater clone button hit collapsible event (bubbling effect from header)

pemudakoding:
This PR fixes bug on Repeater clone button that still hit header event on collapsible
image


danharrin:
Thanks

PR #5315, opened 2022-12-22
Feature / Tabs Component -  Active tab in browser's query string

stealeks:
Will store active tab in browser's query string and will read browser's query string to open tab on page load.

Same as in Wizard Component


danharrin:
Thanks

PR #5329, opened 2022-12-24
Feature: Page Actions renderHooks

ThijmenKort:


ThijmenKort:
This is wat it looks before:
image
This is wat it looks after:
image

pxlrbt:
The "Star" is just an action, right? So this should already be possible. Is there any other use case for non  action content?

ThijmenKort:
> The "Star" is just an action, right? So this should already be possible. Is there any other use case for non action content?

Yeah it is a action from a plugin. So how would you put this action on every single resource page with registering it once?
The hook was just to align it out nicely. Because if you use content.start it makes a new row above the header which looks horrible.

pxlrbt:
You can make your own base class and extend these. I think that would be a cleaner way for actions instead of render hooks. Maybe we could make it easier to swap out the default actions.

ThijmenKort:
> You can make your own base class and extend these. I think that would be a cleaner way for actions instead of render hooks. Maybe we could make it easier to swap out the default actions.

What i wanted to do first is show it on every page. But this would mean adding a hook to the header instead of page actions.
If you use content.start it would look like this:
image

My goal was to make this "star" action visible on every page next to the actions or on the dashboard page on the right side of the header. In my opinion this would be easier to do with renderHooks but if you have another way this is possible i would accept that.

zepfietje:
Wouldn't it be easiest/cleanest to provide the action from your plugin and manually register it on the pages you'd like to have it?

ThijmenKort:
> Wouldn't it be easiest/cleanest to provide the action from your plugin and manually register it on the pages you'd like to have it?

Like i said, i wanted it on every page which includes the dasbhoard. So no i don't think it would be easiest or cleanest to do it that way.

PR #5343, opened 2022-12-26
fix: Lowercase search column to allow searches inside json columns when attaching

wdarins:
When you search on attach modal for a column of type json even if using `*_ci` collation it does not find any values. To fix that I `LOWER($searchColumnName)` to first lowercase data inside json column.


joaovdiasb:
I believe this can cause problems in PostgreSQL for columns like uuid, bigint...

danharrin:
@joaovdiasb we do this in other places already and havent had issues reported yet

PR #5380, opened 2023-01-03
docs: add use keyword for DeleteAction class

lakuapik:
small addition for people who dont know where `DeleteAction` comes from.


danharrin:
It could be `Filament\Pages` or `Filament\Tables`, which is why I didn't add the import

lakuapik:
@danharrin ah i see

PR #5382, opened 2023-01-03
add js code

aymanalareqi:
This PR add js code to scroll into active side bar item when there is long list of sidebar items

### DIFF
--- a/packages/admin/resources/views/components/layouts/base.blade.php
+++ b/packages/admin/resources/views/components/layouts/base.blade.php
@@ -123,7 +123,17 @@
                 })
             </script>
         @endif
-
+        <script>
+            window.addEventListener('load', () => {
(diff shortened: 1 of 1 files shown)


YasinAshorian (approved):

zepfietje:
Thanks for the PR, @aymanalareqi!

@awcodes and I have discussed this change and decided to hold off for now.
Livewire v3 will bring a `@persist` Blade directive which we will possibly use to persist the sidebar between page loads.

Also, there's some additional things to take into account before merging this feature.
If Livewire and Filament v3 won't solve this, we might revisit the PR.

Thanks again!

Edit: you may actually use this PR's functionality as is by [loading a custom script in your Filament project](https://filamentphp.com/docs/2.x/admin/appearance#including-frontend-assets).

PR #5383, opened 2023-01-03
User Menu RTL aligment fix.

mustafa-online:
Fix for RTL aligment


mszabeh:
I think this is better:

`<x-slot name="trigger" class="ltr:ml-4 rtl:mr-4">`

mustafa-online:
@mszabeh Yes, you are right.

PR #5385, opened 2023-01-04
Add instruction to generate URL path from config

lloricode:


YasinAshorian (approved):

danharrin:
This isn't actually recommended, pages and resources have a `::getUrl()` method for this.

PR #5390, opened 2023-01-04
Section description compact

howdu:
Reduce font size of description, if Section is compacted. 

```php
Forms\Components\Section::make('Test')
    ->description('Test description')
    ->compact()
```


danharrin:
thanks!

PR #5393, opened 2023-01-05
bug: changed persisting null to alpine store to empty array.

byt3sage:


NoahNxT:
We also have this issue...
image

byt3sage:
> We also have this issue...
> 
> image
> 
> 

Clear your view cache! Ours was because we had published all the views.

NoahNxT:
After a couple hours of debugging we've found the issue. For some reason the **collapsedGroups** value was saved in localstorage as null...That's why it was crashing.

byt3sage:
Yeah - ours is because we had published the views in our app so the cache was nackered!

PR #5395, opened 2023-01-05
Update user-menu.blade.php

mustafa-online:
Add missing ltr


danharrin:
Thanks, but why? We don't have this anywhere else. Filament's CSS is LTR-first. Same as it is mobile-first.

mustafa-online:
I don't know but it seems not to work. the margin is on both sides.
Check the screen below.
Screenshot (1)

danharrin:
You're missing `rtl:ml-0`

PR #5423, opened 2023-01-08
Fully collapsible sidebar

### LINKED ISSUE #5420: [3.x] Fully collapsible sidebar doesn't work
Enable `sidebarFullyCollapsibleOnDesktop()` in the context. The sidebar is not fully collapsible. I think some of the boolean checks are wrong.

awcodes:
Resolves #5420 

I think there is possibly some more underlying logic issues though. It's possible to add both methods on the context at the same time, resulting in confusing results since both options depend on the same store variable.


danharrin:
Perfect, thank you!

PR #5428, opened 2023-01-09
Fix table alignment with action button

howdu:


danharrin:
Can you check my update to the PR still solves the issue?

Also, I'm confused about what the issue even is

howdu:
Yeah that's cleaner and fixes it.

The issue was when you have a column with alignment and action.
```php
IconColumn::make('test')
    ->alignCenter()
    ->action(function($record, $column) {

});
```

packages/tables/resources/views/components/columns/column.blade.php
```diff
@@ -42,7 +42,7 @@
         <a
             href="{{ $url ?: $recordUrl }}"
             {!! $shouldOpenUrlInNewTab ? 'target="_blank"' : null !!}
-            class="block"
+            class="inline-block"
```
zepfietje, line 45: @danharrin, this broke `recordUrl()` since the column is no longer full width and thus only the text part of the column can be clicked instead of the entire row.

howdu: The button can be made to fill the cell with the following css.
Screenshot 2023-01-25 at 11 09 41

danharrin: we already fixed

PR #5435, opened 2023-01-09
Feature/grid layout builder

alexbirtwell:
As mentioned on discord. I was looking to allow Grid and column span to work in a similar way to components and the Repeater field.

This PR introduced the Grid methods to the Builder field and also makes sure that you can use the columnSpan() options on the Block component. It means when using the builder for a page builder style template, the develop could introduce a width field in the block schema and call back to the columnSpan method on the Block so that the builder represents the visual grid layout.
Screenshot 2023-01-09 at 22 17 08


tonypartridger:
Ohhh me likely! I was wondering how I was going to implement this myself. I was contemplating a settings icon to the left for the colspan and then additional options which open in a modal.

alexbirtwell:
> Ohhh me likely! I was wondering how I was going to implement this myself. I was contemplating a settings icon to the left for the colspan and then additional options which open in a modal.

I do like the idea of having the settings modal dialog per block. to keep everything customisable it would be good if the settings modal worked like actions do in that you can customise the form schema of the modal if you want to. Not one for this PR I don't think but certainly one add once approved and in.

alexbirtwell:
> 

@TonyPartridgeR once in I have the dynamic resize working as per the screen shot with this code. Not overly pretty but Proof of concept. 
```

Builder\Block::make('heading')
            ->columnSpan(fn ($component) => $component->getChildComponents()[0]->getState() == 'full' ? 'full' : (int) $component->getChildComponents()[0]->getState())
            ->schema([
                Select::make('width')
                    ->options([
                        6 => '1/2',
                        'full' => 'full',
                        3 => '1/4',
                        4 => '1/3',
                        8 => '2/3',
                        9 => '3/4',
                    ])
                    ->default(6)
                    ->reactive(),
])
```

tonypartridger:
> > 
> 
> @TonyPartridgeR once in I have the dynamic resize working as per the screen shot with this code. Not overly pretty but Proof of concept.
> 
> ```
> 
> Builder\Block::make('heading')
>             ->columnSpan(fn ($component) => $component->getChildComponents()[0]->getState() == 'full' ? 'full' : (int) $component->getChildComponents()[0]->getState())
>             ->schema([
>                 Select::make('width')
>                     ->options([
>                         6 => '1/2',
>                         'full' => 'full',
>                         3 => '1/4',
>                         4 => '1/3',
>                         8 => '2/3',
>                         9 => '3/4',
>                     ])
>                     ->default(6)
>                     ->reactive(),
> ])
> ```

That is interesting, but I am not sure I fully gasp it, I would have though you would have in the heading an action() which then contains the form schema of the options for that block.

alexbirtwell:
> > > 
> > 
> > 
> > @TonyPartridgeR once in I have the dynamic resize working as per the screen shot with this code. Not overly pretty but Proof of concept.
> > ```
> > 
> > Builder\Block::make('heading')
> >             ->columnSpan(fn ($component) => $component->getChildComponents()[0]->getState() == 'full' ? 'full' : (int) $component->getChildComponents()[0]->getState())
> >             ->schema([
> >                 Select::make('width')
> >                     ->options([
> >                         6 => '1/2',
> >                         'full' => 'full',
> >                         3 => '1/4',
> >                         4 => '1/3',
> >                         8 => '2/3',
> >                         9 => '3/4',
> >                     ])
> >                     ->default(6)
> >                     ->reactive(),
> > ])
> > ```
> 
> That is interesting, but I am not sure I fully gasp it, I would have though you would have in the heading an action() which then contains the form schema of the options for that block.

That would be nice but I need to store the width value in the data json as I use that on the front end that this builder is "building" so to speak. Also keeps this PR clean for them to merge in this instance and just shows how you could go about using your mentioned action approach.

danharrin:
I've just pushed a refactor to this, which uses our existing grid column component. Does it still work ok for you?

Also, doesn't the "create item between blocks" button break in this PR? Since blocks aren't linear anymore... not sure if I like it

danharrin:
What do you think about the create item between button issue @awcodes @zepfietje ?

zepfietje:
Hmm, tricky.
I can't think of a nice way to handle this with the grid.
Not sure if the added complexity is worth it honestly.

awcodes:
That could get real bad real quick. In the screen shot above imagine an insert button below the first block. With this layout, would you expect the insert to be placed above the long bottom block or between the 1st and 2nd blocks?

zepfietje:
> That could get real bad real quick. In the screen shot above imagine an insert button below the first block. With this layout, would you expect the insert to be placed above the long bottom block or between the 1st and 2nd blocks?

And the layout after adding it will be determined by the block you choose from the picker triggered by the button. Basically impossible to get right.

danharrin:
and determined by the position of the current block horizontally...

danharrin:
Yeah this is too difficult to handle, the extra complexity is not worth it. Thanks for the PR though, Alex.

alexbirtwell:
> Yeah this is too difficult to handle, the extra complexity is not worth it. Thanks for the PR though, Alex.

Thanks anyway guys I've extended this to a custom component as I will be reusing this as a content builder on other projects as well

tonypartridger:
> > Yeah this is too difficult to handle, the extra complexity is not worth it. Thanks for the PR though, Alex.
> 
> Thanks anyway guys I've extended this to a custom component as I will be reusing this as a content builder on other projects as well

I think I'll have to do the same too!

PR #5455, opened 2023-01-12
Fix sidebar group collapse icon size

zepfietje:
Makes it match other uses of the icon.


danharrin:
Hmm, I think this looks too big now

zepfietje:
It's in line with the other places. It was inaccessibly small.

PR #5456, opened 2023-01-12
Fix table header sort icon direction

zepfietje:
We need a different alias for `asc` and `desc`.


zepfietje:
@danharrin, this should be ready to be merged. No big changes.

PR #5458, opened 2023-01-12
spatie translatable plugin - add support for Repeater and Builder components

cobraprojects:
for filament spatie translatable plugin


danharrin:
Thanks

PR #5475, opened 2023-01-15
Builder collapse add click to header

howdu:
Make builder header clickable so that it works the same as a collapsible Repeater


danharrin:
Thanks

PR #5476, opened 2023-01-15
Builder Containers reduce spacing between elements

howdu:
Size 12 is too big and not used anywhere else.


danharrin:
It's for the add button that sits between blocks. What does that look like at size 6?

howdu:
Updated to check if this create between is enabled.

PR #5478, opened 2023-01-15
Add color control to Repeater/Builder item create

howdu:


zepfietje:
@danharrin, can't we extract those to actions in v3?

danharrin:
Yeah I think so @zepfietje.

In which case, I will close this PR and extract to actions in v3.

Thanks for your contribution @howdu!

PR #5487, opened 2023-01-17
Fix Italian localization of dashboard title

xTalAlex:
In Italy, "Cruscotto" is used only for vehicles


zepfietje:
Thanks, @xTalAlex!

PR #5492, opened 2023-01-18
Callback support for NavigationItem badge

KaanTanis:
I needed callback for MenuItems defined via AppServiceProvider. I made a small contribution for this.

Now you can:
```php
NavigationItem::make('Test')
    ->setBadge(fn() => User::query()->count())
    ->setBadgeColor(fn() => User::query()->count() < 10 ? 'danger' : 'success')
    
// OR

NavigationItem::make('Test')
    ->setBadge(5)
    ->setBadgeColor('danger')
```

### DIFF
--- a/packages/admin/src/Navigation/NavigationItem.php
+++ b/packages/admin/src/Navigation/NavigationItem.php
@@ -3,7 +3,6 @@
 namespace Filament\Navigation;
 
 use Closure;
-use function PHPUnit\Framework\isInstanceOf;
 
 class NavigationItem
 {
@@ -104,7 +103,7 @@ class NavigationItem
         return $this;
     }
 
-    public function setBadge(string|Closure $callback): static
+    public function setBadge(string | Closure $callback): static
     {
         $this->badge = $callback instanceof Closure
             ? $callback()
@@ -118,7 +117,7 @@ class NavigationItem
         return $this->badge;
     }
 
-    public function setBadgeColor(string|Closure $callback): static
+    public function setBadgeColor(string | Closure $callback): static
     {
         $this->badgeColor = $callback instanceof Closure
             ? $callback()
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks, but I don't understand the value of this. It's identical to `->badge(User::query()->count())`

KaanTanis:
> Thanks, but I don't understand the value of this. It's identical to `->badge(User::query()->count())`

Ooh, I actually needed something else. This was just an example.
I can't think of any other way. That's why I did this :)

For my usage:
```php
->setBadge(function () {
    if (Cache::has('artists_count')) {
        return Cache::get('artists_count');
    } else {
        $count = User::query()->whereHas('roles', function (Builder $query) {
            $query->where('name', 'artist');
        })->count();
        Cache::put('artists_count', $count, now()->addMinutes(30));
        return $count;
    }
})
```

zepfietje:
We shouldn't introduce new setter methods here.
Though, updating the signature on the current `badge()` method would be a breaking change, @danharrin?

danharrin:
Sorry, but the example you posted can also be done without a callback like this. The callback is executed immediately, so there's no difference in just passing the result of your cache to the badge directly:

```php
->badge(value(function () {
    if (Cache::has('artists_count')) {
        return Cache::get('artists_count');
    } else {
        $count = User::query()->whereHas('roles', function (Builder $query) {
            $query->where('name', 'artist');
        })->count();
        Cache::put('artists_count', $count, now()->addMinutes(30));
        return $count;
    }
}))
```

austincarpenter:
@KaanTanis you can simplify this significantly to:
```php
Cache::remember(
    'artists_count',
    now()->addMinutes(30),
    fn () => User::whereRelation('roles', 'name', 'artist')->count()
);
```

PR #5497, opened 2023-01-18
 Default currency in money format

ashleyhood:
This currently doesn't work in production. When configs are cached then env() method will always return `null` and therefore the default option will always be selected.

https://laravel.com/docs/9.x/deployment#optimizing-configuration-loading

I have notified [akaunting/laravel-money](https://github.com/akaunting/laravel-money/issues/75) and suggested that they add a `default_currency` to their config.

### DIFF
--- a/packages/tables/src/Columns/Concerns/CanFormatState.php
+++ b/packages/tables/src/Columns/Concerns/CanFormatState.php
@@ -135,7 +135,7 @@ trait CanFormatState
             }
 
             if (blank($currency)) {
-                $currency = env('DEFAULT_CURRENCY', 'USD');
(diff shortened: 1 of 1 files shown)


danharrin:
Please reopen the PR when the config option gets merged into https://github.com/akaunting/laravel-money, but since it doesn't work right now I'm not going to merge it

For now, just pass in the currency manually

ashleyhood:
I understand why you would close this but it still is bad practice to use the `env()` function outside of config files.

If passing in the currency manually is the solution, then why not just remove the `env()` function altogether and default to 'USD'?

danharrin:
Because at least this way, it works on some machines. And we can replace it with the config option when available.

PR #5507, opened 2023-01-20
✨ Add `CanAssignColumnOrder` to `Filament\Forms\Components\Component`

juliomotol:
This PR aims to add column order assignment to Form components.

The parameter convention is the same as you would in `columnSpan()`.

```php
$component->columnOrder(1);     // order-1
$component->columnOrder([
    "sm" => null,               // `no order style applied`
    "md" => 1,                  // md:order-1
    "lg" => "last",             // lg:order-last
]);
$component->columnOrderFirst(); // order-first
$component->columnOrderLast();  // order-last
$component->columnOrderNone();  // order-none
```


juliomotol:
@zepfietje Any idea on when this would be released?

zepfietje:
After we've found the time to review the PR. 👍

danharrin:
@zepfietje @awcodes @pxlrbt serious question here... do we want to be introducing this many new Tailwind classes to every single Filament build?

danharrin:
TLDR - 90 new classes

danharrin:
Maybe there's a way to do this with inline styles and CSS variables... I bet we can cut this down to just 5 or 6 new classes. What do you think @juliomotol?

danharrin:
Also @juliomotol I would understand a feature for something like col-start and col-end classes... but what exactly is this feature useful for? Can you give us an example use case?

pxlrbt:
The fixed Tailwind classes aren't good anyway. What if someone has more than 12 columns? What if I need `-1`? A CSS var would be more flexible here. It's just a bit more noise in the codebase since you'd need a `<style>` tag for every column to make the breakpoints work, right?

zepfietje:
> Also @juliomotol I would understand a feature for something like col-start and col-end classes... but what exactly is this feature useful for? Can you give us an example use case?

I'm wondering if the same thing can be achieved using col-start and -end...

juliomotol:
> Maybe there's a way to do this with inline styles and CSS variables... I bet we can cut this down to just 5 or 6 new classes

Let me see what I can do.

> I would understand a feature for something like col-start and col-end classes... but what exactly is this feature useful for? Can you give us an example use case?

Say I have 3 cards in my form where the order in the desktop view is 1-2-3. Then for mobile, I want to display it as 1-3-2:

Untitled-2023-01-30-1550

zepfietje:
> Say I have 3 cards in my form where the order in the desktop view is 1-2-3. Then for mobile, I want to display it as 1-3-2:
> 
> Untitled-2023-01-30-1550

I think I've got a solution that might work for you:
Define the cards in the order `[1, 3, 2]` (I suppose 4 in the mobile design should be 2?).
Then set `sm:col-span-full` for card 1, `sm:col-start-1` for card 2, `sm:col-start-2` for card 3.

Could you try this and let us know if it works, @juliomotol?

awcodes:
IMHO this level of customization might be best handled in a stylesheet. Is there ever a situation where this wouldn't be specific to an individual app?

juliomotol:
> Could you try this and let us know if it works, @juliomotol?

Given the following schema:

```php
$form
    ->schema([
        Forms\Components\Section::make('Section 1')
            ->schema([
                Forms\Components\TextInput::make('Foo')
            ])
            ->columnSpanFull(),
        Forms\Components\Section::make('Section 3')
            ->schema([
                Forms\Components\TextInput::make('Foo')
            ])
            ->columnSpan(1),
        Forms\Components\Section::make('Section 2')
            ->schema([
                Forms\Components\TextInput::make('Foo')
            ])
            ->columnSpan(2),
    ])
    ->columns(3)
 ```

## Here is what it initially looks like:
image

## Now, adding a `grid-column-start` on the Section 2 and 3's grid column wrapper:
image
image
> Even putting `grid-column-start: 3` on Section 3 forces Section 2 to be on a new line.

## Then adding a `grid-column-end: 1` on the Section 3's grid column wrapper:
image
image
> There was a prompt on the CSS property saying "The **display: block** property prevents grid-column-end from having an effect"

**Comment by juliomotol** [comment]
> IMHO this level of customization might be best handled in a stylesheet. Is there ever a situation where this wouldn't be specific to an individual app?

The problem here is that the form component is wrapped in a `filament-support::grid.column` component within the `filament-forms::component-container`. I don't think you can style a parent element from a child (if my memory serves me right, haven't dealt with any major FE for a while now)

**Comment by zepfietje** [comment]
@juliomotol, your example can be fixed by using `grid-flow-dense` probably.

**Comment by awcodes** [comment]
@juliomotol you can just do this, then target the grid anyway you want with CSS.

```php
$form
    ->schema([
        Forms\Components\Grid::make(3)
            ->extraAttributes(['class' => 'custom-grid-layout'])
            ->schema([
                Forms\Components\Section::make('Section 1')
                    ->schema([
                        Forms\Components\TextInput::make('Foo')
                    ])
                    ->columnSpanFull(),
                Forms\Components\Section::make('Section 3')
                    ->schema([
                        Forms\Components\TextInput::make('Foo')
                    ])
                    ->columnSpan(1),
                Forms\Components\Section::make('Section 2')
                    ->schema([
                        Forms\Components\TextInput::make('Foo')
                    ])
                    ->columnSpan(2),
            ])
    ]);
```

zepfietje:
Solution I propose:

- Create a `columnStart()` API (https://twitter.com/zepfietje/status/1592126953632718850).
- Create a `dense()` API to control whether grid items should flow into empty cells.

@juliomotol, could you manually try this solution by adding `grid-flow-dense` to your grid?

juliomotol:
Well, that works.
image
image

IMO, this seems to need much more cognition as I'll need to count how many columns I have (which could vary for every breakpoints) and `columnSpan()` of every other components and only then I can base my `columnStart()` from that and I'll also need to make sure that I add a `dense()` on the grid wrapper. Where on the other hand, with `columnOrder()`, I could just count the component and its siblings and base the order from that.

Going back to the 1-2-3 desktop and 1-3-2 mobile example, if only want 2 and 3 to swap places on different breakpoints, I could just wrap them in a `Grid` and I could simply apply `columnOrderFirst()` or `columnOrderLast()` on any of them and forces 1 to always lead after them.

You might think that "What if I just place a `col-start-1` on Section 2 to force it after Section 3". I tried, and it still forces Section 2 on a new line.
image
image

Let me know your thoughts.

danharrin:
Hey, while I understand there is a legitimate use case for this, I feel like the feature has too much footprint for too few use cases, where custom CSS also works. Maintaining it would create struggle and increase the size of the CSS build dramatically. Thanks for the PR.

PR #5526, opened 2023-01-23
Fix `support::button` loading indicator when using action params

jacksleight:
The `<x-filament-support::button>` component adds a loading indicator if you're using `wire:click`, with the action copied over to `wire:target`. However if the action includes parameters (`wire:click="refresh(123)"`) the automatic `wire:target` breaks as that should just be the action name without the brackets and parameters.

This PR fixes that by setting `$loadingIndicatorTarget` to the value only up to the end of the action name.

### DIFF
--- a/packages/support/resources/views/components/button.blade.php
+++ b/packages/support/resources/views/components/button.blade.php
@@ -63,7 +63,7 @@
     $hasLoadingIndicator = filled($attributes->get('wire:target')) || filled($attributes->get('wire:click')) || (($type === 'submit') && filled($form));
 
     if ($hasLoadingIndicator) {
-        $loadingIndicatorTarget = html_entity_decode($attributes->get('wire:target', $attributes->get('wire:click', $form)), ENT_QUOTES);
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks

jackwh:
@jacksleight Unfortunately this causes pages with multiple action buttons to lose their loading target specificity.

On pages where `getActions()` defines several buttons, the value of `$loadingIndicatorTarget` gets set to `mountAction('action-x')`, `mountAction('action-y')`, `mountAction('action-z')` etc. — so this change causes every button to just target `mountAction`.

This affects actions which open modals or which return closures.

Screenshot 2023-02-04 at 17 08 52

jacksleight:
Ah whoops, this is obviously wrong. I'll submit a PR to revert.

Confused why the indicators aren't working here though, removing the params bit from `wire:target` was the only way to make them work. 🤔

jackwh:
@jacksleight no worries... I just put one in already, see here: https://github.com/filamentphp/filament/pull/5614

PR #5538, opened 2023-01-25
Get resource class from service container

howdu:
This allows for overwriting a resource registered in another package. E.g 
```php
$this->app->alias(\App\Filament\Resources\NavigationResource::class, \RyanChandler\FilamentNavigation\Filament\Resources\NavigationResource::class);
```


danharrin:
Thanks, but this change isn't gonna fully replace the resource. For example, navigation and routing will all use the original resource. I think it's the responsibility of the plugin author to allow their resources to be deregistered and overridden.

PR #5539, opened 2023-01-25
Account for unauthenticated users in action authorization

zepfietje:
`$user` can be `null` if user is unauthenticated.
Instead of throwing an exception, we now return `false` from `isAuthorized()`.


zepfietje:
Maybe we should actually handle this differently, @danharrin. Policies allow for methods where the user is nullable.

I think we should call `Gate::check()` instead of `$user->can()`?

PR #5544, opened 2023-01-25
Feature/sort by buttons

bernhardh:
As discussed, a couple of days ago on discord, this PR reintroduces optionally the up/down buttons for repeater and builder. By default, the buttons are disabled (not shown), so its a none breaking change. If you don't use it, you will not see any change.

To enable just change the config (`forms.components.repeater.up_and_down_buttons` and `forms.components.builder.up_and_down_buttons`).

When enabled, it will look like this:

image

Same goes for the builder. In the following example, i Have enabled collapsible and clonable as well, just to see multiple icons:

image


packages/forms/resources/views/components/repeater.blade.php
```diff
@@ -134,6 +134,9 @@ class="gap-6"
                                         'flex divide-x rtl:divide-x-reverse',
                                         'dark:divide-gray-700' => config('forms.dark_mode'),
                                     ])>
+
```
zepfietje, line 153:
```suggestion
```

zepfietje (changes requested):

packages/forms/src/helpers.php
```diff
@@ -10,7 +10,7 @@ function array_move_after(array $array, $keyToMoveAfter): array
         $indexToMoveAfter = array_search($keyToMoveAfter, $keys);
         $keyToMoveBefore = $keys[$indexToMoveAfter + 1] ?? null;
 
-        if (filled($keyToMoveBefore)) {
```
danharrin, line 13: Please explain this change

packages/forms/src/helpers.php
```diff
@@ -35,7 +35,7 @@ function array_move_before(array $array, $keyToMoveBefore): array
         $indexToMoveBefore = array_search($keyToMoveBefore, $keys);
         $keyToMoveAfter = $keys[$indexToMoveBefore - 1] ?? null;
 
-        if (filled($keyToMoveAfter)) {
```
danharrin, line 38: ...and also this

packages/forms/config/forms.php
```diff
@@ -26,6 +26,14 @@
 
         ],
 
+        'repeater' => [
```
danharrin: Please remove these config options, instead add `reorder ableWithButtons()` and `isReorderableWithButtons()` methods to the Repeater and Builder classes

packages/forms/resources/views/components/builder.blade.php
```diff
@@ -155,6 +155,8 @@
                                     'flex divide-x rtl:divide-x-reverse',
                                     'dark:divide-gray-700' => config('forms.dark_mode'),
                                 ])>
+                                    <x-forms::up-and-down-buttons component="builder" :is-item-movement-disabled="$isItemMovementDisabled" :loop="$loop" :state-path="$getStatePath()" :uuid="$uuid" />
```
danharrin: Please don't extract this Blade component, the code can be copied between each view which makes it easier to override the behaviour in the view for just 1 component

packages/forms/resources/views/components/repeater.blade.php
```diff
@@ -134,6 +134,8 @@ class="gap-6"
                                         'flex divide-x rtl:divide-x-reverse',
                                         'dark:divide-gray-700' => config('forms.dark_mode'),
                                     ])>
+                                        <x-forms::up-and-down-buttons component="repeater" :is-item-movement-disabled="$isItemMovementDisabled" :loop="$loop" :state-path="$getStatePath()" :uuid="$uuid" />
```
danharrin: See comment above

danharrin (changes requested):

packages/forms/src/helpers.php
bernhardh: Ah, forget it. I thouht it would be more readable.

packages/forms/src/helpers.php
bernhardh: Ah, forget it. I thouht it would be more readable.

packages/forms/config/forms.php
bernhardh: done

packages/forms/resources/views/components/builder.blade.php
bernhardh: Done. Just to understand it better for the next time: What is the benefit of not making it a component? This sorting behaviour logic is an atomic behaviour which can be easily isolated and it is already used by two individual components (repeater and builder). I don't get the benefit of copy+pasting code, especially if it has 40+ lines?

Making it a custom component, increases (in my eyes) the readability and most important: it makes it much more easy to overwrite the component in the app. For example if someone would change the icon of it, you will have to extend not only two files, but the whole blade template of repeater/builder.

Where is my mistake?

bernhardh:
All requested changes done. Thank you!

packages/forms/resources/views/components/builder.blade.php
danharrin: Its not a mistake, just my preference in this situation.

1) none of the other buttons are extracted yet, would prefer to do them all at once
2) in v3 we're extracting the buttons to actions anyway (probably), so this new view will just need to be removed again
3) as I said before, easier to override without having to look up a separate view

bernhardh:
Any chance, that this PR gets merged soon?

danharrin:
Thanks

PR #5550, opened 2023-01-26
Add ability to filter `HasManyThrough` relations

masterix21:
It adds the ability to create filters capable of interactions with `HasManyThrough` relations.


danharrin:
Thanks!

PR #5554, opened 2023-01-27
Fix docs access $this and protected property

howdu:


zepfietje:
What's the issue here, @howdu?
I believe this should work just fine on an admin page.

howdu:
Yeah standard admin page with Tables\Actions\DeleteAction. I got two errors:
> $this when not in object context.
> Cannot access protected property Filament\Tables\Actions\DeleteAction::$record

zepfietje:
Ah, probably because of the longhand closure.

Thanks!

danharrin:
Don't we need to inject `$record` into the closure?

zepfietje:
> Don't we need to inject `$record` into the closure?

Maybe that's cleaner yeah.

danharrin:
I don't agree with the change at all, this works with DeleteAction used on the Edit page. `$record` is for tables

danharrin:
the longhand closure makes no difference to `$this` scope

zepfietje:
But why is @howdu reporting errors then? 🤔

danharrin:
because he's using `Tables\Actions\DeleteAction` which is meant for tables, not page actions

zepfietje:
> because he's using `Tables\Actions\DeleteAction` which is meant for tables, not page actions

Lol, only see it now. 😅

howdu:
Sorry guys I thought this was a straight forward fix and don't mean to be annoying :)

When I specifically pass Pages DeleteAction: `before(function (\Filament\Pages\Actions\DeleteAction $action)`. [Example Repository](https://github.com/howdu/filament-tester/blob/delete-action/app/Filament/Resources/ProductResource.php#L52) I still get the error that it's the wrong: 
> App\Filament\Resources\ProductResource::App\Filament\Resources\{closure}(): Argument #1 ($action) must be of type Filament\Pages\Actions\DeleteAction, Filament\Tables\Actions\DeleteAction given, called in /filament-tester/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php on line 36

zepfietje:
@howdu, no worries.
Next time, try including some context (the error you ran into) in the PR description, that would make our lives a little easier.

As for your last error, make sure to use the right namespace with `DeleteAction::make()` as well.

PR #5556, opened 2023-01-27
AttachAction - use the specified related key names

Skullbock:
When using `BelongsToMany` relationships that have custom keys, the `AttachAction` needs to load the record using the `relatedKey` instead of the primary key of the record, because the id passed by the search query itself is the `relatedKey`, not the `primaryKey` of that record


zepfietje (approved):

danharrin:
Thanks

Skullbock:
Thanks for your great work on filament. It's really a lifesaver

PR #5564, opened 2023-01-28
Move callAfter hook outside of create another

howdu:
After event was only being called if you clicked 'Create & create another'


danharrin:
Breaking change and IMO it makes more sense how it was before. You can check `$arguments['another']` within `after()` if you want.

howdu:
@danharrin I think this is still an issue somebody else reported on discord and said that this change fixed it for them too:
https://discord.com/channels/883083792112300104/883085267383226478/1069821293475667968 

If `$arguments['another']` is false it will never reach the `callAfter`

PR #5567, opened 2023-01-28
update Kurdish translation

AlanD20:
Updated Kurdish translation and added other missing files and translation keys.


danharrin:
Thanks

PR #5570, opened 2023-01-29
Default table action configuration

zepfietje:


danharrin:
I don't understand the motivation... looping through every action every time to replace a simple `??` 🤔

zepfietje:
This PR is still very WIP and experimental, @danharrin. 😄

The purpose is the following:
Currently, actions in the table package are configured to be links and size `sm` by default.
Design-wise, the empty state actions should be `md` and button instead.
This PR aims to find a way to configure these default sizes when the action has no size set explicitly (`null`).

zepfietje:
Thoughts, @danharrin?

danharrin:
I think this work should be done in the setters which contains some similar logic like passing `->table()` I think.

Also you need to pass `link()` to the row actions.

zepfietje:
Yeah, but the issue with moving it to the setter would be that it's hard to override the default size and such, right?

danharrin:
By the time the setter is run, the action is set up, so the null check will still work.

danharrin:
I think that maybe instead of the null check, we introduce a new `defaultSize()` method and property which is then used inside `getSize()`. So that the logic doesn't leak out of the class

zepfietje:
Sounds interesting! Do you want to set up a POC in this PR?
I can work it out after that.

danharrin:
Replaced by #5792

PR #5581, opened 2023-01-30
hotfix: prevent the record gone after blast clicking action

kevariable:


pemudakoding:
I supporting this one, when the action is processing we need to disable it or otherwise the action still get execute if clickin' it twice (slow network). it impact another scenario as well like the thread starter when clickin it twice when the action processing with custom query it will make the record gone.

zepfietje:
Isn't this already handled by the underlying button component?

pemudakoding:
> Isn't this already handled by the underlying button component?

https://www.loom.com/share/326a9d9ab5324ee7bf7812525c4d94f1

you can see loom video above, i click the view twice, and when the first modal open you can see the action get loading again (hit again, affected by click twice before)

so in conclusion it's not handled yet by the button component

kevariable:
> > Isn't this already handled by the underlying button component?
> 
> https://www.loom.com/share/326a9d9ab5324ee7bf7812525c4d94f1
> 
> you can see loom video above, i click the view twice, and when the first modal open you can see the action get loading again (hit again, affected by click twice before)
> 
> so in conclusion it's not handled yet by the button component

Thanks mate to explain the incident

danharrin:
As Zep said, I don't think this changes anything. The button already has this attribute: https://github.com/filamentphp/filament/blob/2.x/packages/support/resources/views/components/button.blade.php#L79

PR #5582, opened 2023-01-30
Fix: modal height bug

### LINKED ISSUE #5228: Modal button is not visible, not clickable on  iPhone 6s, iOS: 15.6.1
### Package

filament/filament

### Package Version

v.2.16.55

### Laravel Version

v.9.33

### Livewire Version

v.2.10.7

### PHP Version

php 8.1

### Problem description

If the modal is open, the button at the bottom of the pop-up window cannot be clicked because the browser's control bar is above.
The error occurred on  iPhone 6s, iOS: 15.6.1, but it works fine on android. Cancel button is not showing, not clickable.
image0

### Expected behavior

We can click the button of the modal window.

### Steps to reproduce

Open the filament demo page on  iPhone 6s, iOS: 15.6.1 mobile phone (https://demo.filamentphp.com/shop/customers/1/edit?activeRelationManager=0), log in, open the customer menu. Edit any customer. Click the "New address" button.

### Reproduction repository

https://github.com/filamentphp/filament

### Relevant log output

_No response_

awcodes:
Resolves #5228 

Theoretically, this should fix the issue, but I have no way to test an iPhone6s.

### DIFF
--- a/packages/admin/resources/views/components/layouts/app.blade.php
+++ b/packages/admin/resources/views/components/layouts/app.blade.php
@@ -3,7 +3,7 @@
 ])
 
 <x-filament::layouts.base :title="$title">
-    <div class="filament-app-layout flex w-full min-h-screen overflow-x-clip">
+    <div class="filament-app-layout flex w-full min-h-full overflow-x-clip">
         <div
             x-data="{}"
             x-cloak
--- a/packages/admin/resources/views/components/layouts/app.blade.php
+++ b/packages/admin/resources/views/components/layouts/app.blade.php
@@ -3,7 +3,7 @@
 ])
 
 <x-filament::layouts.base :title="$title">
-    <div class="filament-app-layout flex w-full min-h-full overflow-x-clip">
+    <div class="filament-app-layout flex min-h-full w-full overflow-x-clip">
         <div
             x-data="{}"
             x-cloak
--- a/packages/admin/resources/views/components/layouts/base.blade.php
+++ b/packages/admin/resources/views/components/layouts/base.blade.php
@@ -6,7 +6,7 @@
 <html
     lang="{{ str_replace('_', '-', app()->getLocale()) }}"
     dir="{{ __('filament::layout.direction') ?? 'ltr' }}"
-    class="antialiased bg-gray-100 filament js-focus-visible h-full"
+    class="filament js-focus-visible h-full bg-gray-100 antialiased"
 >
     <head>
         {{ \Filament\Facades\Filament::renderHook('head.start') }}
--- a/packages/admin/resources/views/components/layouts/base.blade.php
+++ b/packages/admin/resources/views/components/layouts/base.blade.php
@@ -6,7 +6,7 @@
 <html
     lang="{{ str_replace('_', '-', app()->getLocale()) }}"
     dir="{{ __('filament::layout.direction') ?? 'ltr' }}"
-    class="antialiased bg-gray-100 filament js-focus-visible"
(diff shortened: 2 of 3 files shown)


zepfietje:
Haven't tested this, but the classes you added to `html` and `body` are exactly what I have in a non-Filament project of mine.

Should we still test this, or let @szommeragi pull in these changes?

If there's no other issues being introduced by this change, I think it looks ok.

awcodes:
Honestly, I don't know any way to test it. Everything else still functions normally as best I can tell in the demo with the changes.

zepfietje:
I can pull in the changes and run it on my iPhone.

awcodes:
Do you have an iPhone 6? I guess the issue would be the same on a 13 too though.

zepfietje:
I have a 12 mini, so small enough to reproduce. 😆

Been able to reproduce the issue, also with this PR.
So this doesn't fix it unfortunately.

Seems like it might have something to do with not being able to scroll when the modal is open?
When the browser tab bar is hidden (by first scrolling down), the modal shows just fine.
However, when showing the tab bar again, there's no way to hide it.

awcodes:
I wonder if the modal content just needs to set to overflow scroll, so the modal itself can't be larger than the available total height.

zepfietje:
Hmm, not sure if that'd fix it.
The modal content doesn't seem to overflow.

awcodes:
Think i got this @zepfietje.

awcodes:
Used valet share to ngrok to my phone and the modal itself was scrollable with the latest commit. But please check on your end too.

packages/admin/resources/views/components/layouts/base.blade.php
```diff
@@ -6,7 +6,7 @@
 <html
     lang="{{ str_replace('_', '-', app()->getLocale()) }}"
     dir="{{ __('filament::layout.direction') ?? 'ltr' }}"
-    class="antialiased bg-gray-100 filament js-focus-visible"
+    class="antialiased bg-gray-100 filament js-focus-visible h-full"
```
zepfietje:
```suggestion
    class="filament js-focus-visible h-full bg-gray-100 antialiased"
```

packages/admin/resources/views/components/layouts/app.blade.php
```diff
@@ -3,7 +3,7 @@
 ])
 
 <x-filament::layouts.base :title="$title">
-    <div class="filament-app-layout flex w-full min-h-screen overflow-x-clip">
+    <div class="filament-app-layout flex w-full min-h-full overflow-x-clip">
```
zepfietje:
```suggestion
    <div class="filament-app-layout flex min-h-full w-full overflow-x-clip">
```

packages/admin/resources/views/components/layouts/base.blade.php
```diff
@@ -75,7 +75,7 @@ class="antialiased bg-gray-100 filament js-focus-visible"
     </head>
 
     <body @class([
-        'filament-body bg-gray-100 text-gray-900',
+        'filament-body bg-gray-100 text-gray-900 min-h-full',
```
zepfietje:
```suggestion
        'filament-body min-h-full bg-gray-100 text-gray-900',
```

zepfietje (changes requested):
Just fixing the order.

zepfietje:
Confirmed this fixes the issue indeed.

Really great work on fixing this one, @awcodes! 💪

This can now be merged, @danharrin. 😄

zepfietje (approved):

danharrin:
Thanks so much @awcodes !

PR #5584, opened 2023-01-30
Add Chart Filter Loading Indicator

iAmKevinMcKee:
I added the loading indicator as discussed in Discord. A few things to note:

I added a public property and set it to `true`. Not sure if you want the default true or false. Also, I'm not sure if you want the long name I gave it. Of course you are free to edit whatever.

I played around with adding a `wire:loading.delay` but even with `wire:loading.delay.shortest` it felt janky. I think it is best as is and if someone doesn't want it, they can turn it off. Again, always up to you.

Finally, not sure if you are good with the svg being inline or if you want that loading indicator extracted as a component. If you want to do this, please let me know where you want that component to be and I can update.

If this is merged I'm happy to update docs, but wasn't sure what kind of changes you'd want to make before doing that.


zepfietje:
@iAmKevinMcKee, there already exists a loading indicator component being used in a lot of places (`<x-filament::loading-indicator>`.
Just search for it to find usages.
Could you update this PR to use that component? 😄

iAmKevinMcKee:
Got it. Made that update. Thanks for the feedback.

packages/widgets/resources/views/chart-widget.blade.php
```diff
@@ -23,17 +23,27 @@
...
-                        wire:model="filter"
-                        wire:loading.class="animate-pulse"
-                    >
-                        @foreach ($filters as $value => $label)
-                            <option value="{{ $value }}">
-                                {{ $label }}
-                            </option>
-                        @endforeach
-                    </select>
+                    <div class="flex items-center space-x-3">
```
zepfietje:
```suggestion
                    <div class="flex items-center gap-3">
```

packages/widgets/resources/views/chart-widget.blade.php
```diff
@@ -23,17 +23,27 @@
...
+                            wire:model="filter"
+                            wire:loading.class="animate-pulse"
+                        >
+                            @foreach ($filters as $value => $label)
+                                <option value="{{ $value }}">
+                                    {{ $label }}
+                                </option>
+                            @endforeach
+                        </select>
+                    </div>
```
zepfietje:
```suggestion
                        @if ($showFilterLoadingIndicator)
                            <x-filament::loading-indicator
                                x-cloak="x-cloak"
                                class="h-8 w-8 text-gray-500"
                                wire:loading
                                wire:target="filter"
                            />
                        @endif

                        <select
                            class="text-gray-900 border-gray-300 block h-10 transition duration-75 rounded-lg shadow-sm focus:border-primary-500 focus:ring-1 focus:ring-inset focus:ring-primary-500 dark:bg-gray-700 dark:border-gray-600 dark:text-gray-200 dark:focus:border-primary-500"
                            wire:model="filter"
                            wire:loading.class="animate-pulse"
                        >
                            @foreach ($filters as $value => $label)
                                <option value="{{ $value }}">
                                    {{ $label }}
                                </option>
                            @endforeach
                        </select>
                    </div>
```

zepfietje (approved):

packages/widgets/resources/views/chart-widget.blade.php
```diff
@@ -23,17 +23,28 @@
...
-                            <option value="{{ $value }}">
-                                {{ $label }}
-                            </option>
-                        @endforeach
-                    </select>
+                    <div class="flex items-center gap-3">
+                        @if ($showFilterLoadingIndicator)
+                            <x-filament::loading-indicator
+                                x-cloak="x-cloak"
+                                class="h-8 w-8 text-gray-500"
```
zepfietje, line 30: These size values look odd to me? (Haven't tested visually, but just from the code.)

zepfietje (changes requested):

packages/widgets/resources/views/chart-widget.blade.php
zepfietje: Also not sure about the color.

iAmKevinMcKee: Here's what it looks like: 
image

packages/widgets/src/ChartWidget.php
```diff
@@ -14,6 +14,8 @@ abstract class ChartWidget extends Widget
     protected ?array $cachedData = null;
 
     public string $dataChecksum;
+    
+    public bool $showFilterLoadingIndicator = true;
```
danharrin:
```suggestion
    public bool $hasFilterLoadingIndicator = true;
```

packages/widgets/resources/views/chart-widget.blade.php
```diff
@@ -23,17 +23,28 @@
...
-                        wire:loading.class="animate-pulse"
-                    >
-                        @foreach ($filters as $value => $label)
-                            <option value="{{ $value }}">
-                                {{ $label }}
-                            </option>
-                        @endforeach
-                    </select>
+                    <div class="flex items-center gap-3">
+                        @if ($showFilterLoadingIndicator)
```
danharrin:
```suggestion
                        @if ($hasFilterLoadingIndicator)
```

PR #5587, opened 2023-01-31
Fix extra request on non-modal action mount

zepfietje:


danharrin:
How about the other modals in this component?

zepfietje:
Yeah, they would need to be changed in a similar way.
However, I just wanted to be sure this change is a good fix before implementing it in all places.
Could you confirm? Then I'll finish the PR.

danharrin:
It looks fine, but have you tested by clicking on modal and non-modal actions sequentially to check that Alpine reregisters the new event listener?

zepfietje:
I've updated this PR to handle all modals.
From my testing everything works fine.

PR #5591, opened 2023-02-01
Always Show Pagination Overview

intrepidws:
When a resource has only one result, the pagination overview text (`Showing X to Y of Z results`)  is not shown. While it's obviously pretty easy to see that there is only one record, I find that the change in the UI is jarring. There really isn't a great reason not to show it IMO.

This PR shows the pagination overview regardless of how many results there are.

### DIFF
--- a/packages/tables/resources/lang/en/table.php
+++ b/packages/tables/resources/lang/en/table.php
@@ -27,7 +27,7 @@ return [
 
         'label' => 'Pagination Navigation',
 
-        'overview' => 'Showing :first to :last of :total results',
+        'overview' => '{1} Showing 1 result|[2,*] Showing :first to :last of :total results',
 
         'fields' => [
 
--- a/packages/tables/resources/lang/nl/table.php
+++ b/packages/tables/resources/lang/nl/table.php
@@ -27,7 +27,7 @@ return [
 
         'label' => 'Paginering navigatie',
 
-        'overview' => 'Toont :first tot :last van :total resultaten',
+        'overview' => '{1} Toont 1 resultaat|[2,*] Toont :first tot :last van :total resultaten',
 
         'fields' => [
 
--- a/packages/tables/resources/views/components/pagination/index.blade.php
+++ b/packages/tables/resources/views/components/pagination/index.blade.php
@@ -62,11 +62,15 @@
                     'pl-2 text-sm font-medium',
                     'dark:text-white' => config('tables.dark_mode'),
                 ])>
-                    {{ __('tables::table.pagination.overview', [
-                        'first' => $paginator->firstItem(),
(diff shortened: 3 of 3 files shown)


zepfietje:
We might want to update the string as it will now read "Showing 1 to 1 of 1 results"?

intrepidws:
Yea, it's certainly a bit awkward. What about just `Showing 1 result`?

zepfietje:
That would be better I guess.
Maybe that's the reason why it was hidden for 1 record in the first place? 😅

intrepidws:
If you're showing 10 records per page and there are 9 records available, the overview text shows but says `Showing 1 to 9 of 9 results`. So there's awkward sounding duplication there as well. In an ideal world, it would probably just say `Showing 9 results`.

While I'm willing to update this PR to account for the one page (`Showing 9 results`) and one record (`Showing 1 result`) scenarios, it would probably mean adding two new keys to every table language file, like this:

```
    'pagination' => [

        'label' => 'Pagination Navigation',

        'overview' => [
        
            'default' => 'Showing :first to :last of :total results',

            'one_page' => 'Showing :total results',

            'one_result' => 'Showing :total result',

        ],
```

I think that I could pretty easily figure out the wording for the one page scenario out for non-English languages. For instance, this:

`'default' => 'عرض :first إلى :last من :total النتائج',`

Would probably just become

`'one_page' => 'عرض :total النتائج',`

But the one result scenario would likely be implausible for many languages, since many are likely much more complicated than simply dropping the `s` off of `results`.

My vote would probably be to deal with the awkwardness of the rarely seen `Showing 1 to 1 of 1 results` and call it a day. I think slightly awkward language is better than a UI that changes unexpectedly.

zepfietje:
In my opinion, if we're going to make any changes to this, we should fix the translations first instead of showing an awkward string.

We probably don't even need new localization strings, since we can use [pluralization](https://laravel.com/docs/9.x/localization#pluralization).
Have a look and let me know if you can get this working in a nice way.

Mark the PR as ready when done so we can have another look. 😊

zepfietje:
I've updated this PR to use `Showing 1 result`.
It still uses `Showing 1 to 9 of 9 results` though, since changing that would require new translation keys, which are simply more of a hassle.

Still not too sure if this change is worth it, as this will now show `Showing 1 to 1 of 1 results` for locales that have not been updated to include this change.

intrepidws:
This feels like the right solution to me. As long as @danharrin is OK with it, I'll try and update more language files.

Thanks for getting to this @zepfietje, I've been massively sick for the last couple weeks.

danharrin:
Thanks, hope you're starting to feel better @intrepidws

PR #5603, opened 2023-02-02
Allow to override component container

eddierusinskas:
It would be useful if you could override ComponentContainer in forms with config as it would allow more flexibility to custom `$wire->` commands for custom Fields

### DIFF
--- a/packages/forms/config/forms.php
+++ b/packages/forms/config/forms.php
@@ -1,5 +1,6 @@
 <?php
 
+
 return [
 
     /*
@@ -13,6 +14,7 @@ return [
     */
 
     'components' => [
+        'container' => Filament\Forms\ComponentContainer::class,
 
         'actions' => [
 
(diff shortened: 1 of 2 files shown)


danharrin:
You should be able to just `app()->bind(ComponentContainer::class, YourOwnComponentContainer::class)`

darrena092:
I gave what you suggested a go @danharrin but it doesn't seem to successfully replace the `ComponentContainer` as far as being able to call methods from it in Alpine using `$wire` goes.

danharrin:
Component containers arent Livewire components

darrena092:
Okay, that being the case - how are the methods from the `SupportsSelectFields` trait being exposed to `selectFormComponent`? I'm a colleague of @eddierusinskas and our use case is a custom field that uses a similar approach to searching as the `Select` field.

danharrin:
The Livewire component (where you add HasForms) has an InteractsWithForms trait. That contains all the public Livewire methods that get called from form components.

So I would add a trait that you can reuse between Livewire components which adds methods accessible by `$wire` from your form component.

darrena092:
Thank you! I'll give that a go.

PR #5616, opened 2023-02-04
Fix: slideover enter animation

### LINKED ISSUE #4883: Modal slide-over enter animation not working
### Package

filament/filament

### Package Version

v2.16.42

### Laravel Version

v9.39.0

### Livewire Version

_No response_

### PHP Version

PHP 8.1.0

### Problem description

Modal slide-over enter animation only fades in opacity.

### Expected behavior

Modal should slide in from the right.

### Steps to reproduce

Create a slide-over table action.

### Reproduction repository

### Relevant log output

_No response_

awcodes:
Resolves #4883


awcodes:
Resolves #4883

packages/support/resources/views/components/modal/index.blade.php
```diff
@@ -88,33 +83,38 @@ class="filament-modal {{ $displayClasses }}"
...
             x-ref="modalContainer"
-            x-cloak
             {{ $attributes->class([
-                'relative w-full cursor-pointer pointer-events-none',
+                'relative w-full cursor-pointer pointer-events-none transition',
                 'my-auto p-4' => ! $slideOver,
             ]) }}
         >
             <div
+                x-data="{ show: false }"
```
zepfietje, line 93:
```suggestion
                x-data="{ isShown: false }"
```

packages/support/resources/views/components/modal/index.blade.php
```diff
@@ -88,33 +83,38 @@ class="filament-modal {{ $displayClasses }}"
...
+                x-show="show"
+                x-init="$nextTick(()=> {
+                    show = isOpen
+                    $watch('isOpen', () => show = isOpen)
+                })"
```
zepfietje:
```suggestion
                x-data="{ isShown: false }"
                x-init="$nextTick(()=> {
                    isShown = isOpen
                    $watch('isOpen', () => isShown = isOpen)
                })"
                x-show="isShown"
```

awcodes:
Looks good to me.

zepfietje:
Cool, I'll still pull these changes in locally and see if it's all fine visually. 👍

zepfietje:
Nice to see a smooth modal animation again! 😄

One issue still though is that sometimes the modal disappears before the hiding animation has completed. Thoughts on how we can fix that?

awcodes:
Different durations maybe?

zepfietje:
Hmm, that'd still be a partial fix probably.
We could also hide the modal and fire the closing event afterwards, maybe?

awcodes:
I guess the problem is that it's two animations that are tied by the same event and I'm not sure of a way to have transitions depend on each other directly. So, I really think it's about getting the timings right.

zepfietje:
Hmm, I tested this by increasing the transition durations, but it doesn't seem like what you suggested.
I think it's rather something to do with the modal content being removed before transition has finished.

awcodes:
Do you have a screen recording by any chance. Either I'm old and just can't see it or I just can't replicate it.

danharrin:
Thanks @awcodes, just gonna merge how it is and Zep can handle any other fixes as required :)

zepfietje:
Was about to suggest this, @danharrin!
The biggest issue has been fixed, so I'll see later if there's a way to polish even further.

PR #5620, opened 2023-02-06
Add decimal separator parameter for money mask

margarizaldi:


danharrin:
Breaking change, but thank you

margarizaldi:
Really? Hmmm... Never mind...

danharrin:
Yes, what if people are already using that method and expect the parameters in that old order?

margarizaldi:
OMG how can I forgot the order... Just because the docs give example with named params 🤣 Sorry...

PR #5629, opened 2023-02-07
Add deferred loading on table

Sicklou:
Hello,

This is the first time I make a PR on a feature, I'm aware it is not perfect at first try.
I know my naming choices are probably not good. Maybe I put some code at the wrong place.

So, this PR adds a way to defer the loading of data by exploiting the `wire:init` directive of livewire.

To make it work with a table resource, juste use the `deferLoading()` method : 
```php
{
    return $table
        ->columns([
            TextColumn::make('name')
        ])
        ->actions([
        ])
        ->bulkActions([
            Tables\Actions\DeleteBulkAction::make(),
        ])
        ->deferLoading();
}
```

To use it with the standalone package, `true` will defer loading, `false` will load data at the page loading.
```php
 public function isLoadingDeferred(): bool
 {
      return true;
 }
```


danharrin:
Hey! A few things...

- PHPStan failing
- Please rename methods and properties in the CanDeferLoading to include "table", so they don't get confused with other methods on the Livewire component. `$isTableLoaded`, `isTableLoadingDeferred()`, `isTableLoaded()`. `allowLoadingData()` can be removed as it does the same as `isTableLoaded()`.
- Instead of returning an empty collection, wrap the table contents in an `@if ($isTableLoaded)` instead.

Sicklou:
Thanks for the review.

I made the changes but if I wrap the table contents instead of returning an empty collection, the query run when the page loads, since the `getRecords` is called.
Which is what I want to avoid with this PR.

Do you see a way to do it ?

danharrin:
You can move the `$getRecords()` etc inside the `@if`

danharrin:
I think we need some sort of loading indicator when the records are not present. We should default the `$records` to be `null` instead and use `$records === null` and render the `<x-filament-support::loading-indicator />` like we do in other places

Sicklou:
Just added a loading indicator, it is nicer now

packages/tables/resources/views/index.blade.php
```diff
@@ -299,7 +304,11 @@ class="shrink-0"
             }"
         >
             @if ($content || $hasColumnsLayout)
-                @if (count($records))
+                @if (is_null($records))
+                <div class="flex items-center justify-center p-4">
+                    <x-filament-support::loading-indicator class="w-5 h-5" />
+                </div>
+                @elseif (count($records))
```
zepfietje: Could you fix indentation here, @Sicklou?

packages/tables/resources/views/index.blade.php
```diff
@@ -745,7 +755,15 @@ class="filament-table-header-cell-{{ \Illuminate\Support\Str::of($column->getNam
...
 
-                    @if (count($records))
+                    @if (is_null($records))
+                    <tr>
+                        <td colspan="{{ $columnsCount }}">
+                            <div class="flex items-center justify-center w-full p-4">
+                            <x-filament-support::loading-indicator class="w-5 h-5"/>
+                            </div>
+                        </td>
+                    </tr>
```
zepfietje: And here.

zepfietje (changes requested):

danharrin:
Thanks

Sicklou:
Maybe it would be nice to mention it in the documentation. But Im not sure where it can be added

danharrin:
I've added it in the v3 docs, I'm not sure where to put it in v2.

archilex:
Great PR @Sicklou! I’ve already implemented it in my resources. Very nice! 👏

I did notice that on mobile with lots of columns since the loading indicator is centered, you don’t see it. I’m wondering if changing it for `<x-tables::loading-cell>` would be a viable option so that you can see the loading animation on any size?

Potentially you could even get the default pagination count and render the entire table? Not sure. I haven’t looked into it as I’m on my phone. 

Thoughts?

archilex:
Just PRed the change to see if the group likes it. #5731

PR #5633, opened 2023-02-08
Add extra attributes to the section component

iAmKevinMcKee:
Used the same approach as other components that have the `HasExtraAttributes` and `HasExtraAlpineAttributes` traits.


danharrin:
Great, could you do the same for the infolists section Blade also?

PR #5637, opened 2023-02-09
$action->halt() Example update.

tonypartridger:
pass $action into the methods function, without it the code will not work.


packages/admin/docs/02-resources/06-deleting-records.md
```diff
@@ -120,7 +120,7 @@ use Filament\Notifications\Actions\Action;
 use Filament\Notifications\Notification;
 
 DeleteAction::make()
-    ->before(function () {
+    ->before(function ($action) {
```
danharrin: /wtd add a DeleteAction type for this variable

PR #5647, opened 2023-02-11
Fix: height issues on body with overflow-x-clip and min-h-full

awcodes:
This fixes a bug I came across with the min-h-full on the body causing issues when the body content is too short to reach the bottom of the window.

There is no know issues with Filament directly that this solves, but is causing issues with a plugin and could potentially create issues with other plugins.

### DIFF
--- a/packages/admin/resources/views/components/layouts/base.blade.php
+++ b/packages/admin/resources/views/components/layouts/base.blade.php
@@ -75,7 +75,7 @@
     </head>
 
     <body @class([
-        'filament-body min-h-full bg-gray-100 text-gray-900',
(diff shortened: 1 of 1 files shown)


danharrin:
Can you please give a little explanation of how this fix works?

awcodes:
It's a quirk with min-h-full. Without an explicit height on the element a browser doesn't calculate the full height of any children inside it with a height set to 100%. Adding the h-0 forces those children to calculate the actual height of the parent.

Without the fix:
Screen Shot 2023-02-11 at 1 45 41 PM

With the fix:
Screen Shot 2023-02-11 at 1 49 42 PM

danharrin:
Thanks!

PR #5649, opened 2023-02-11
Fix: Table TextColumn rowIndex pagination support

Z3d0X:
Current implementation of `->rowIndex()` on Table TextColumn does not take pagination into consideration.


danharrin:
We can't release breaking changes to `rowIndex()` but the docs can be updated to remove that method.

danharrin:
The correct `rowIndex()` is in v3

PR #5658, opened 2023-02-13
Fix: overflow issues on html, body with modals

awcodes:
This fixes bug which would have been created by #5647 while maintaining fix for original issue of modals not scrolling on mobile devices.


danharrin:
Thanks!

PR #5659, opened 2023-02-13
Add supporting media for relationship Models

fixmax:
It'd be helpful when you need to show media not only for editing a Model but also for the current Model's relationship.


danharrin:
Thanks for the PR! However...
- There's no way to set the relation
- There's no logic for saving the media to the correct model
- Its already possible using https://filamentphp.com/docs/2.x/forms/layout#saving-data-to-relationships

PR #5660, opened 2023-02-13
Update translations

zepfietje:


packages/tables/resources/lang/en/table.php
```diff
@@ -16,17 +16,17 @@
...
     'fields' => [
 
-        'search_query' => [
-            'label' => 'Search',
-            'placeholder' => 'Search',
-        ],
-
         'bulk_select_page' => [
-            'label' => 'Select / Deselect all items for bulk actions.',
+            'label' => 'Select/Deselect all items for bulk actions.',
```
danharrin:
```suggestion
            'label' => 'Select/deselect all items for bulk actions.',
```

zepfietje: The space should not be there.

packages/tables/resources/lang/en/table.php
```diff
@@ -16,17 +16,17 @@
...
         ],
 
         'bulk_select_record' => [
-            'label' => 'Select / Deselect item :key for bulk actions.',
+            'label' => 'Select/Deselect item :key for bulk actions.',
```
zepfietje:
```suggestion
            'label' => 'Select/deselect item :key for bulk actions.',
```

packages/tables/resources/lang/en/table.php
danharrin: hmm why?

zepfietje: https://www.grammarly.com/blog/slash/

"When a slash signifies alternatives between only two words, don’t use spaces before or after."

PR #5661, opened 2023-02-13
Switch from json_encode to castAsJson

archilex:
When running the test suite against a MySql database the CreateRecordTest fails due to json encoding the tags. This works when running againast Sqlite, but fails with MySql. 

Using the `castAsJson()` method works for both databases.


danharrin:
Thanks!

danharrin:
Thanks!

danharrin:
Hmm seems like this doesn't work with L8, reverting

archilex:
Ugh. sorry about that. I found this issue being discussed [here](https://github.com/laravel/framework/issues/43863).

PR #5667, opened 2023-02-14
Add decimalSeparator to money mask

RickDBCN:
Zep suggested we could add the decimalseparator in the function as a parameter at the end, so it won't be a breaking change.

For v3, any chance we could change the order?
```php
public function money(string $prefix = '$', string $thousandsSeparator = ',', string $decimalSeparator = '.', int $decimalPlaces = 2, bool $isSigned = true): static
    {

    }
```

### DIFF
--- a/packages/forms/src/Components/TextInput/Mask.php
+++ b/packages/forms/src/Components/TextInput/Mask.php
@@ -141,14 +141,14 @@ class Mask implements Jsonable
         return $this;
     }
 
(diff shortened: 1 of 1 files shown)


zepfietje:
Yeah, @RickDBCN asked me if this could be an option.

Let me know your thoughts, @danharrin.

danharrin:
I don't think that's the place for it to be honest, if you want to add this it should be in v3

zepfietje:
Yeah maybe better to only introduce in v3 instead of adding to v2 in an awkward place and changing it again for v3.

PR #5648, opened 2023-02-11
Fix/tables a11y

awcodes:
Resolves #4794

PR #5669, opened 2023-02-14
Update PT-BR translations for a11y

saade:
#5648 #5660

adds pt_BR translations for the new a11y translation keys.


zepfietje:
Thanks!

PR #5676, opened 2023-02-14
Fix outline-color transition

### LINKED ISSUE #5675: Outline-color transition flash due to Tailwind 3.2.5
### Package

filament/filament

### Package Version

v2.*

### Laravel Version

v.9

### Livewire Version

_No response_

### PHP Version

8.1

### Problem description

Tailwind version 3.2.5, which released February 8, added the `outline-color` property to the `transition` and `transition-colors` utilities. [release notes](https://github.com/tailwindlabs/tailwindcss/blob/master/CHANGELOG.md)

This causes a weird flash when clicking into an input field as the outline is now being transitioned. 

Note: this only occurs if you've themed filament (ie. changed the colors) as the default config is still using 3.2.4. 

A few ways this can be fixed until/if this is updated is:

1. Lock your tailwind version in package.json to 3.2.4, then npm update. 
2. Or add the following to your `filament.css` file

```css
@layer components {
    .transition {
        transition-property: color,background-color,border-color,text-decoration-color,fill,stroke,opacity,box-shadow,transform,filter,backdrop-filter,-webkit-backdrop-filter !important
    }
}
```

### Expected behavior

The input fields shouldn't flash an outline color.

### Steps to reproduce

1. Follow the documentation to theme filament. Following the documention you'll pull in your version of tailwind instead of the filament's, which as of February 8 is 3.2.6. 
2. Open your browser and click into an input field to see the outline-color transition.

### Reproduction repository

https://github.com/archilex/filament-demo/tree/outline-color-bug

### Relevant log output

_No response_

archilex:
This fixes the outline-color transition issue on inputs as a result of tailwind's 3.2.5 release when themeing Filament.

It applies the transition settings pre 3.2.5 which didn't include outline-color.

See issue #5675 

There may be a better way to solve this, like actually working with the outline-color instead of overwriting it, but I wanted to at least propose this as a solution.

### DIFF
--- a/packages/admin/resources/css/app.css
+++ b/packages/admin/resources/css/app.css
@@ -24,6 +24,10 @@
 }
 
 @layer components {
+    .transition {
+        transition-property: color,background-color,border-color,text-decoration-color,fill,stroke,opacity,box-shadow,transform,filter,backdrop-filter,-webkit-backdrop-filter !important
+    }
+
     .filament-login-page {
         @apply relative bg-no-repeat;
         background-image: radial-gradient(


archilex:
Just did some additional tests and another solution would be to add `outline-none` to the text-input, search-input, etc. in the core components. Not sure how many other files would need to be updated, but this would be another solution. If you prefer this way then I can submit a PR with those changes.

danharrin:
Please go with that `outline-none` fix, I don't feel comfortable overriding a core Tailwind class like this

danharrin:
I've downgraded Tailwind in the core until this is fixed. I hope their regression will be reverted soon

zepfietje:
Not sure if the regression will be reverted in Tailwind CSS though.
Linking back for reference: https://github.com/tailwindlabs/tailwindcss/pull/10385.

Will comment on that thread now.

archilex:
Sounds good and I agree. I’ll make the changes to the actual inputs. Much cleaner that way.

zepfietje:
@archilex, wouldn't this be a quite easy fix though?
A comment on that thread suggests that switching all occurrences of `focus:outline-none` to `outline-none` would do the job.
Would be easy to search for and replace that way.

What do you think?

archilex:
Yes, it shouldn’t be too hard to fix. In my quick testing adding outline-none was enough. I don’t believe there is a focus:outline-none on the input. I’ll check. Not sure what the consequences of replacing it would be as opposed to just adding outline-none. I’ll try to see.

zepfietje:
If replacing instead of adding works, I'd prefer that.
Let me know if you need any help.

archilex:
Yeah, no focus:outline-none on the inputs I just looked at. So we’ll have to just add outline-none. I won’t be able to get to it for an hour or so, so feel free to run with it if you need it before then. 

And good find with the Tailwind issue. Wish I would have seen that before spending an hours trying to figure out if I had messed up my configuration some how! 🤦‍♂️

zepfietje:
Only found that issue through your link to the Tailwind CSS changelog. 😂
I actually have to thank you for finding the issue, since I saw it in my projects and thought something was wrong with my browser. 😅

It's night here now, so feel free to work on this PR when you find the time.
Make sure to check all places where `focus:outline-none` is used.

archilex:
Sounds good. Will do

zepfietje:
@archilex, I'm working on a fix already. 👍

archilex:
Ok great!

zepfietje:
Replaced by https://github.com/filamentphp/filament/pull/5682.

Thanks for bringing this up and suggesting the fix, @archilex!

PR #5684, opened 2023-02-15
update french translations following #5648, #5660, #5591

ArnaudLier:


packages/tables/resources/lang/fr/table.php
```diff
@@ -16,6 +16,14 @@
 
     'fields' => [
 
+        'bulk_select_page' => [
+            'label' => 'Sélectionner/déselectionner tous les éléments pour les actions groupées.',
+        ],
+
+        'bulk_select_record' => [
+            'label' => 'Sélectionner/désélectionner l\'élément :key pour les actions groupées.',
```
zepfietje: Could you use `"` instead of `'` here to prevent having to escape in the string?

zepfietje (changes requested):

packages/tables/resources/lang/fr/table.php
ArnaudLier: Sure! Some other strings used `\` so I went with that.

zepfietje (approved):

zepfietje:
Perfect.
Thanks, @ArnaudLier!

PR #5685, opened 2023-02-15
Updates AR translations.

ahmed-abobaker:


packages/tables/resources/lang/ar/table.php
```diff
@@ -27,7 +35,7 @@
 
         'label' => 'التنقل بين الصفحات',
 
-        'overview' => 'عرض :first إلى :last من :total النتائج',
+        'overview' => '{1} عرض نتيجة واحدة|[3,10] عرض :first إلي :last من :total نتائج|[2,*] عرض :first إلي :last من :total نتيجة',
```
zepfietje, line 38: Could you clarify the use of `[3,10]`?

ahmed-abobaker: Yes, in Arabic the plural is different in several cases

zepfietje: Cool, learned something new today! 😀

I suppose `[*,2]` could be changed into `{2}`?
If so, could you update that and change the order so it's `{1}`, `{2}` and then `[*,2]`?
Also, how should a count of more than 10 be handled?

ahmed-abobaker: [3,10] is a different case, [2]  and [11,*] same case, so I put [3,10] in middle.
In first case will check {1} then if [3,10] then if 2 or more than 10 will do the last case because that i write [2,*] in the last case

zepfietje: You might mean `[2,*]` in code instead of `[*,2]` then? 😄

AAbosham: It's actually `[2,*]` but showing like this `[*,2]` from code editor

image

zepfietje: Huh, but it's also showing wrong on GitHub?

AAbosham: Yes, in all code editors when the Arabic text overlaps with another language or symbols. 
You can check by searching in the code or editor, that because Arabic is RTL.

AAbosham: Final result for this PR

[screencast-github.com-2023.02.15-12_11_51.webm](https://user-images.githubusercontent.com/12083600/218985060-8d2b2fcf-04e9-4e6e-8501-85f8f5b7747e.webm)

zepfietje: It's hard for me to check if everything works fine, but I'll assume it does if you say so.
Thanks for educating me about RTL text in editors, guys! 😄

zepfietje (approved):

PR #5688, opened 2023-02-15
Add a link to the closure customization page

megawubs:
To be able to use the `getOptionLabelFromRecordUsing` method it's good to understand how closures work in Filament.

### DIFF
--- a/packages/forms/docs/03-fields.md
+++ b/packages/forms/docs/03-fields.md
@@ -673,7 +673,7 @@ Select::make('authorId')
     ->getOptionLabelFromRecordUsing(fn (Model $record) => "{$record->first_name} {$record->last_name}")
 ```
 
-For more information on how colsure agruments work inside of Filament read [using closure customization](advanced#using-closure-customization).
+For more information on how closure arguments work in Filament, read [using closure customization](advanced#using-closure-customization).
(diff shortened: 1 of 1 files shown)


packages/forms/docs/03-fields.md
```diff
@@ -673,6 +673,8 @@ Select::make('authorId')
     ->getOptionLabelFromRecordUsing(fn (Model $record) => "{$record->first_name} {$record->last_name}")
 ```
 
+For more information on how colsure agruments work inside of Filament read [using closure customization](advanced#using-closure-customization).
```
zepfietje:
```suggestion
For more information on how closure arguments work in Filament, read [using closure customization](advanced#using-closure-customization).
```

zepfietje:
I'm actually not sure if we should add this.
There's a lot of places that would require adding this to be consistent which would also bloat the docs.

Same thing for https://github.com/filamentphp/filament/pull/5689/files.

danharrin:
I agree, `getOptionLabelFromRecordUsing()` doesn't actually use strict closure customisation either, we manually pass in the record based on the option that is being rendered. Usually the record is associated with the form itself, so this behaviour could be confusing.

PR #5691, opened 2023-02-15
standardize single quotes escapes in translations

ArnaudLier:
single quotes + `\`-escaping to double quotes in translation files


zepfietje:
Thanks a lot!

PR #5693, opened 2023-02-15
Accept Htmlable for table column label

Tabitha-art:
added to the property and setter as well

### DIFF
--- a/packages/tables/src/Columns/Concerns/HasLabel.php
+++ b/packages/tables/src/Columns/Concerns/HasLabel.php
@@ -4,14 +4,15 @@ namespace Filament\Tables\Columns\Concerns;
 
 use Closure;
 use Illuminate\Support\Str;
+use Illuminate\Contracts\Support\Htmlable;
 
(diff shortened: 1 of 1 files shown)


packages/tables/src/Columns/Concerns/HasLabel.php
```diff
@@ -4,14 +4,15 @@
 
 use Closure;
 use Illuminate\Support\Str;
+use Illuminate\Contracts\Support\Htmlable;
```
zepfietje:
```suggestion
use Illuminate\Contracts\Support\Htmlable;
use Illuminate\Support\Str;
```

zepfietje:
Doesn't this break the `shouldTranslateLabel` functionality, @Tabitha-art?

danharrin:
@zepfietje I don't think it breaks that, they just don't work together which isn't a bad thing. I don't even like `translateLabel()` lol.

zepfietje:
Me neither. 😂

PR #5704, opened 2023-02-16
Fix incorrect language tag for pagination overview

cheesegrits:


zepfietje:
Thanks for fixing, @cheesegrits. Didn't notice yet it's broken for me in prod! 😅

PR #5718, opened 2023-02-17
Add Progress Column

moskoweb:
I adapted the plugin created by @ryangjchandler to add the Progress column next to the Filament, I believe the column can be a standard feature of the Filament.

Credits: @ryangjchandler 
Plugin: https://github.com/ryangjchandler/filament-progress-column

### DIFF
--- a/packages/tables/docs/03-columns/14-progress.md
+++ b/packages/tables/docs/03-columns/14-progress.md
@@ -1,0 +1,57 @@
+---
+title: Progress column
+---
+
+Progress columns render a progress bar based on the values passed.
+
+```php
+use Filament\Tables\Columns\ProgressColumn;
+
+ProgressColumn::make('progress')
+```
+
+## Dynamic progress calculation
+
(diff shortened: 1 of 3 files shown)


zepfietje:
What is the benefit of having this in core instead of as a plugin, @moskoweb?
I feel like it's perfect as a plugin, but not necessarily a core feature.

Thoughts, @danharrin @ryangjchandler?

danharrin:
I agree with Zep 👍

moskoweb:
Just as there are other columns in the Filament, I see it as being advantageous to make the Filament more and more complete, without depending on plugins that may be without updates in the future.

danharrin:
The more features Filament has that not everyone uses, the larger the bundle and source size which is frankly uneconomical from a maintenance perspective or an efficiency perspective

ryangjchandler:
@moskoweb You can make pull requests to plugins too, not everything _has_ to be in core. If you think something is missing from my package, open a pull request.

zepfietje:
Oh and @ryangjchandler is a Filament core team member, so we'll make sure the package is updated. 😉

PR #5723, opened 2023-02-18
Fixes header cell sort icon spacing

archilex:
This fixes the extra margin-left spacing that was added in #5648 due to the additional sr-only span classes. 

With the addition of the spans, `space-x-1` is no longer viable. This removes the `space-x-1` and adds `ml-1` to the icon (and adjusts for RTL).

### DIFF
--- a/packages/tables/resources/views/components/header-cell.blade.php
+++ b/packages/tables/resources/views/components/header-cell.blade.php
@@ -14,7 +14,7 @@
         @endif
         type="button"
         @class([
-            'flex items-center w-full px-4 py-2 whitespace-nowrap space-x-1 rtl:space-x-reverse font-medium text-sm text-gray-600',
+            'flex items-center w-full px-4 py-2 whitespace-nowrap font-medium text-sm text-gray-600',
             'dark:text-gray-300' => config('tables.dark_mode'),
             'cursor-default' => ! $sortable,
             match ($alignment) {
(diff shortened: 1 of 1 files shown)


packages/tables/resources/views/components/header-cell.blade.php
```diff
@@ -14,7 +14,7 @@
         @endif
         type="button"
         @class([
-            'flex items-center w-full px-4 py-2 whitespace-nowrap space-x-1 rtl:space-x-reverse font-medium text-sm text-gray-600',
+            'flex items-center w-full px-4 py-2 whitespace-nowrap font-medium text-sm text-gray-600',
```
zepfietje, line 17: Wouldn't it be better/cleaner to use `gap-x-1` here instead? @danharrin @archilex @awcodes

danharrin: The point of the PR is that there are now more child elements of this button, so we can't use `space-` or `gap-`, we only need a left margin before the icon.

zepfietje: We can use gap, since flex gap doesn't add an additional gap for the `sr-only` items, while `space-` does as it simply adds a hard margin to all child elements.

danharrin: Okay, make the change if you want but I don't think the margin is objectively bad, just a different approach

zepfietje: Will create a PR.
To me `gap-x-1` is way more maintainable than `ml-1 rtl:ml-0 rtl:mr-1`.

zepfietje: https://github.com/filamentphp/filament/pull/5725

archilex: Agree @zepfietje and I learned something. Thanks.

zepfietje: Perfect, thank you for fixing the issue initially!

PR #5728, opened 2023-02-18
Adding ability to `emitSelf`, `emitUp` and `emitTo` for `CanEmitEvent` trait

titantwentyone:
PR to allow emitting to self, up and to from Notification Actions. Docs also updated.

I have also included this for the JS side of things. I could only test this anecdotally. Is there a test suite for JS in Filament?


danharrin:
Thanks!

PR #5731, opened 2023-02-18
Improves deferloading loading indicator for mobile

archilex:
The new deferLoading() method #5629  renders the loading-indicator, but on mobile where tables are usually wider than the device width, the indicator is off screen.  This swaps the the loading-indicator for the loading-cell animation so that it renders on mobile as well. 

I also went ahead and looped it 10 times as rendering a table usually show 10 rows. This can be removed of course if we prefer to only have it render only once. 

I didn't swap out the first indicator as it's for column layouts and those naturally wrap for mobile so I don't think it's a problem there.


zepfietje:
We definitely can't hardcode 10 rows.
Also, I think it's best to just show a single loading indicator just like when searching.

archilex:
I actually think it’d be better to extract this to a component so it’ll be easy to customize by overwriting the view. 

So the question is which should be the default?
1. loading-indicator (origianl)
2. loading-cell (just one)
3. loop of 10 loading-cells (this PR)

I’m inclined towards option 3, but you all tell me.

zepfietje:
Overriding views is strongly discouraged as a customization method.

In my opinion, there should only be a single loading indicator for deferred tables, as the amount of rows isn't know yet.

archilex:
Ok. So you’re saying option 2 above, right? Or just keep the original option? Or do the loading-cell but don’t extract to a component?

zepfietje:
I don't know from the top of my head how this is handled during search, but that exact behavior is what I would expect.

archilex:
How this is handled during while searching a table is interesting. When the page is loaded, and assuming there are records (which I think is a good asumption), searching will bring up the loading-cell animation for all the cells:

Screenshot 2023-02-18 at 8 55 53 PM

If the search returns results, then searching again has the same loading-cell animation.

However if the search returns no results, and you search again (or clear the search) then you get this spinner:

Screenshot 2023-02-18 at 8 56 16 PM

Both are different than the current implementation which is this loading-indicator:

Screenshot 2023-02-18 at 8 58 57 PM

This indicator is used in the global search bar (among other places I'm sure, but not in the tables).

All that brings me to think that we should use the loading-cell animation since it's the most common loading animation you see in the table. I continue to think it's a better UX on mobile since when you are loading the page for the first time you are expecting to see results and having a spinner off screen on mobile gives you no indication that something is happending. 

So here's a new commit, where I extracted this into it's own component with only one loading-cell animation row (removed the for loop).

It'll look like this:

Screenshot 2023-02-18 at 9 31 03 PM

And on mobile:

Screenshot 2023-02-18 at 9 31 43 PM

(I still think it looks the best with a loop of 10, but I understand and agree that hardcoding a number in there isn't great.)

zepfietje:
Ah, thanks for clarifying, @archilex!

The full row loading animation is there because there's already a certain number of rows being displayed before starting to search.
After all, I think the loading indicator in the second screenshots is still what we should use for deferred tables.

archilex:
Sounds good. Here's a new commit using a version of the empty-state indicator. I also replaced the column view loading indicator for this one so they match. 

Here's screenshots of what it looks like using the normal table view and if you're using the column view. If you're not using the column view, the indicator is still off screen on mobile, but at least it's consistent with the empty state indicator. (Colors are different since I've used a different color palette. 

Screenshot 2023-02-19 at 6 25 25 PM

Screenshot 2023-02-19 at 6 25 04 PM

And the placement of this defer-loading-indicator is exactly the same as the empty state icon so if you do happen to load a page that doesn't have any records there isn't any jumping around of the icons.

Let me know if there are other changes needed.

packages/tables/resources/views/components/defer-loading-indicator.php
```diff
@@ -0,0 +1,13 @@
+<div {{ $attributes->class([
+    'filament-tables-empty-state flex flex-1 flex-col items-center justify-center p-6 mx-auto space-y-6 text-center bg-white',
+    'dark:bg-gray-800' => config('tables.dark_mode'),
+]) }}>
+    <div @class([
+        'flex items-center justify-center w-16 h-16 text-primary-500 rounded-full bg-primary-50',
```
zepfietje: Dark background color should be handled.

packages/tables/resources/views/components/defer-loading-indicator.php
```diff
@@ -0,0 +1,13 @@
+<div {{ $attributes->class([
+    'filament-tables-empty-state flex flex-1 flex-col items-center justify-center p-6 mx-auto space-y-6 text-center bg-white',
```
zepfietje: `filament-tables-empty-state` should not be used here.

packages/tables/resources/views/components/defer-loading-indicator.php
```diff
@@ -0,0 +1,13 @@
...
+    'dark:bg-gray-800' => config('tables.dark_mode'),
+]) }}>
+    <div @class([
+        'flex items-center justify-center w-16 h-16 text-primary-500 rounded-full bg-primary-50',
+        'dark:bg-gray-700' => config('tables.dark_mode'),
+    ])>
+        <x-filament-support::loading-indicator
+            class="w-6 h-6"
+        />
```
zepfietje:
```suggestion
        <x-filament-support::loading-indicator class="w-6 h-6" />
```

zepfietje (changes requested):

zepfietje:
Looks much better already, @archilex, thanks!
We'll just have to clean up the code a bit.

packages/tables/resources/views/components/defer-loading-indicator.php
archilex: 👍 Should I change it for a `filament-tables-defer-loading-indicator` class? Or just remove it?

packages/tables/resources/views/components/defer-loading-indicator.php
archilex: This is the same way it’s handled in the empty state component. It looked fine to me in my testing. Am I missing something?

archilex: CA365913-7C1A-4BD3-A094-F7634F1081B8
38C06F65-474E-49F8-B14C-384589C420F7

packages/tables/resources/views/components/defer-loading-indicator.php
archilex: Went ahead and added the `filament-tables-defer-loading-indicator` hook.

packages/tables/resources/views/components/defer-loading-indicator.php
zepfietje: Looks good. Sorry, I missed the separate dark mode class the line below. In v3 they're all on a single line. 🤦‍♂️

packages/tables/resources/views/components/defer-loading-indicator.php
zepfietje: I'm not sure if defer-loading indicator is correct in general, since it's also being used for the search loading indicator, right?

archilex: It's closer to the empty-state component that includes the loading-indicator component inside it. I was thinking of naming it `defer-loading-state` to match `empty-state`, however the empty state component has 2 different states (loading or no records found), so I didn't feel calling `defer-loading-state` was a good match. But that would be another way to go.

But it's different than just the loading indicator because it has the bg colored circle around it.

zepfietje: But inside the table there's no other loading indicator being used, right?

archilex: And sorry such a small little loading-indicator became such an in-depth conversation when I'm sure there are more important features to be working on 🤦‍♂️

archilex: Correct, no other loading indicator now

danharrin:
I've moved it outside of the `<table>` now to fix the alignment

PR #5732, opened 2023-02-18
Fixing `assertNotified` to remove false positives and usage of `id` in equality

### LINKED ISSUE #5717: Possible issue with `assertNotified`
### Package

filament/filament

### Package Version

v2.17.2

### Laravel Version

v10.0.3

### Livewire Version

v2.11.2

### PHP Version

8.1.15

### Problem description

`assertNotified` seems to exhibit false positives and not asserting correctly.

### Expected behavior

For assertions to be correct.

### Steps to reproduce

Open the repo and run the tests for the package. I have added one test to indicate the issues in `tests/src/Admin/Pages/ActionTest.php` and added supporting actions in `tests/src/Admin/Fixtures/Pages/PageActions.php`. Code is:

In `ActionTest.php`:

```
it('can show a notification', function () {

    //this passes
    livewire(PageActions::class)
        ->callPageAction('shows_notification')
        ->assertNotified();

    //this passes but we have to 'send' the notification
    livewire(PageActions::class)
        ->callPageAction('shows_notification')
        ->assertNotified(
            Notification::make()
                ->title('A notification')
                ->success()
                ->send()
        );

    //this passes but a notification was never sent
    livewire(PageActions::class)
        ->callPageAction('does_not_show_notification')
        ->assertNotified(
            Notification::make()
                ->title('A notification')
                ->success()
                ->send()
        );

    //this fails - id is not the same
    livewire(PageActions::class)
        ->callPageAction('shows_notification')
        ->assertNotified(
            Notification::make()
                ->title('A notification')
                ->success()
        );
});
```

In PageActions.php:
```
...
Action::make('shows_notification')
    ->action(function() {
        Notification::make()
            ->title('A notification')
            ->success()
            ->send();
    }),
Action::make('does_not_show_notification')
...
```
I have stated that Laravel v10 is affected but this is also evident in v9.42.2.

### Reproduction repository

https://github.com/titantwentyone/filament/tree/issue/assert-notified

### Relevant log output

_No response_

titantwentyone:
Fixes #5717.

Two main things I found when digging into this. My tests are perhaps overkill but just wanted to be sure. Please trim if necessary! 

1. One of the issues that when comparing notifications, the `id` is used in that comparison. If an `id` is missing when an notification is created, then one is randomly generated, This happens when the notification is stored and retreived. As a result, the comparison invariably fails as most people won't set an `id` or don't even realise that an `id` can be set. The solution here was to just remove `id` when testing for equality between the arrays.

2. `assertNotified` doesn't mimic the removal of notifications when they are dispatched. So false positives can arise. Given the example:

```
Action::make('does_not_show_notification'),
Action::make('shows_notification_with_id')
    ->action(function() {
        Notification::make('notification_with_id')
            ->title('A notification')
            ->success()
            ->send();

...

it('should fail', function () {

    //this passes correctly
    livewire(PageActions::class)
        ->callPageAction('shows_notification_with_id')
        ->assertNotified(
            Notification::make('notification_with_id')
                ->title('A notification')
                ->success()
        );

    //this should fail
    livewire(PageActions::class)
        ->callPageAction('does_not_show_notification')
        ->assertNotified(
            Notification::make('notification_with_id')
                ->title('A notification')
                ->success()
        );
});
```
This test passes as the first assertion does not remove the notification from the session. The second assertion test just pulls the last notification from the session and uses that to test. I've added in a small amendment to remove the tested notification and write back to the session so that it is sync with the test.

### DIFF
--- a/packages/notifications/src/Notification.php
+++ b/packages/notifications/src/Notification.php
@@ -210,12 +210,12 @@ class Notification extends ViewComponent implements Arrayable
 
         Assert::assertIsArray($notifications->toArray());
 
-        if(is_string($notification)) {
-            $expectedNotification = $notifications->first(fn($n) => $n->title == $notification);
-        };
+        if (is_string($notification)) {
+            $expectedNotification = $notifications->first(fn ($n) => $n->title == $notification);
+        }
 
         if ($notification instanceof Notification) {
-            $expectedNotification = $notifications->first(fn($n, $key) => $n->id == $key);
+            $expectedNotification = $notifications->first(fn ($n, $key) => $n->id == $key);
         }
 
         if (blank($notification)) {
--- a/packages/notifications/src/Notification.php
+++ b/packages/notifications/src/Notification.php
@@ -211,11 +211,11 @@ class Notification extends ViewComponent implements Arrayable
         Assert::assertIsArray($notifications->toArray());
 
         if (is_string($notification)) {
-            $expectedNotification = $notifications->first(fn ($n) => $n->title == $notification);
+            $expectedNotification = $notifications->first(fn (Notification $notification): bool => $notification->title === $notification);
         }
 
         if ($notification instanceof Notification) {
-            $expectedNotification = $notifications->first(fn ($n, $key) => $n->id == $key);
+            $expectedNotification = $notifications->first(fn (Notification $notification, string $key): bool => $notification->id === $key);
         }
 
         if (blank($notification)) {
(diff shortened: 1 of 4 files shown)


titantwentyone:
Have converted to draft as looking for more elegant solution to clearing down notifications from with a test.

titantwentyone:
Last commit uses `Notifications` to handle the notifications in the session. Have also ensured that multiple assertions can be called and that notifications can be tested in any order. Feel that the majority of the tests are testing the actual assertion - perhaps need a different home for these?

danharrin:
Thanks!

PR #5739, opened 2023-02-20
feat/set notification view from ->view('view')

wychoong:
currently to customise notification view requires to configure in service provider and this set the view globally

this PR enable to set view from `Notification->view('view')`

```php
    public function notification()
    {
        Notification::make()
            ->title('default notification')
            ->success()
            ->send();
    }

    public function product_notification()
    {
        Notification::make()
            ->title('inline notification')
            ->body('product added to cart')
            ->view('notifications.product')
            ->viewData([
                'image' => 'https://images.unsplash.com/photo-1542291026-7eec264c27ff?ixlib=rb-4.0.3&ixid=MnwxMjA3fDB8MHxzZWFyY2h8NHx8cHJvZHVjdHN8ZW58MHx8MHx8&auto=format&fit=crop&w=500&q=60',
            ])
            ->send();
    }
```

image


zepfietje:
Thanks for contributing this PR, @wychoong!

Configuring a view which is different than the default requires a check to prevent security issues where a malicious user could set an arbitrary view. Have a look at how this is done in notification actions, which already support custom views: https://github.com/filamentphp/filament/blob/0192236a04a7ee64f3044cdce390fb5d1c346f89/packages/notifications/src/Actions/Action.php#L51

Mark this PR as ready for review when you're finished.
Also, let me know if you have any questions!

wychoong:
> Thanks for contributing this PR, @wychoong!
> 
> Configuring a view which is different than the default requires a check to prevent security issues where a malicious user could set an arbitrary view. Have a look at how this is done in notification actions, which already support custom views:
> 
> https://github.com/filamentphp/filament/blob/0192236a04a7ee64f3044cdce390fb5d1c346f89/packages/notifications/src/Actions/Action.php#L51
> 
> Mark this PR as ready for review when you're finished. Also, let me know if you have any questions!

It explicitly checks for `notifications::actions.` which wouldn’t work for this case
https://github.com/filamentphp/filament/blob/0192236a04a7ee64f3044cdce390fb5d1c346f89/packages/notifications/src/Actions/Action.php#L72

Or should it check for existence in resources folder?
```php
view()->exists(…)
```

I don’t see how is this dangerous looking at how ViewField or any other field currently is doing
https://filamentphp.com/docs/2.x/forms/fields#view
But I agree if to make it better I think need another method instead of `getViewData`

danharrin:
ViewField doesn't ever serialise to JSON... notifications do. So if someone sent a notification from the JS API or something, they shouldn't be able to set a view in your app to anything

zepfietje:
Exactly like Dan said.

> It explicitly checks for notifications::actions. which wouldn’t work for this case

You should then override the `isViewSafe` method inside your custom `Notification` class to accept the views you want.

wychoong:
> the JS API

Ah, forgot about this. 


> You should then override the `isViewSafe` method inside your custom `Notification` class to accept the views you want.

 I will likely change to check an array of allowed views, do you think I should submit the changes as PR?
- set valid views in config
- Check valid views
- rename getViewData to getNotificationViewData
- Add api to JS

zepfietje:
I think we shouldn't add new config options, as those will be removed in v3.

wychoong:
> I think we shouldn't add new config options, as those will be removed in v3.

i have updated it to use `configureUsing` instead of config option
```php
    Notification::configureUsing(function (Notification $notification): void {
            $notification->safeViews('notifications.product');
     });
```

danharrin:
Thanks, but weird PHPStan error. I will run it locally and double check.

wychoong:
thanks for merging this in! much appreciated

packages/notifications/src/Notification.php
```diff
@@ -86,6 +103,21 @@ public static function fromArray(array $data): static
         return $static;
     }
 
+    protected function isViewSafe(string $view): bool
+    {
+        return in_array($view, $this->safeViews, strict: true);
+    }
+
+    public function safeViews(string | array $safeViews): static
```
zepfietje, line 111: @danharrin, only now I see we're accepting string and array here. Since the name of the setter is `safeViews` (plural), wouldn't it make sense to only accept an array...? Would be breaking now I guess, but maybe for v3?

danharrin: I don't mind it as a DX thing TBH

zepfietje: Alright then.

PR #5742, opened 2023-02-20
Handle JSON columns in form select component relationships

### LINKED ISSUE #5740: Form Select component : using '->' to access nested json value in relationship method
### Package

filament/filament

### Package Version

v2.17.3

### Laravel Version

v9.52.0

### Livewire Version

v2.11.3

### PHP Version

PHP 8.0.2

### Problem description

When using the select component and passing a string containing the '->' as the second parameter in the relationship method, for example if we want to get the value of name from data json field
````
\Filament\Forms\Components\Select::make('select')->relationship('relation', 'data->name');

````
the following error is thrown
````
Undefined property: stdClass::$data->name
````

### Expected behavior

The select component should display the options based on the relationship and the $titleColumnName (being the value from json column in this case) defined in the second parameter of the relationship method.

### Steps to reproduce

Create a select component in a form with a relationship method call and set titleColumnName parameter to be a value from json column:
```
Select::make('select')
    ->relationship('relation', 'data->name')
```
Run the code and observe the error thrown.

### Reproduction repository

https://github.com/drhootch/filament-select-relationship-json-issue

### Relevant log output

_No response_

### PR DESCRIPTION
**Pull request by drhootch**
Adding alias to relationshipTitleColumnName inside pluck fixes the issue of not being able to get fields for json columns.
And won't break for non-json columns also.

### DIFF
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -474,6 +474,7 @@ class Select extends Field implements Contracts\HasNestedRecursiveValidationRule
             }
 
             $relationshipTitleColumnName = $component->getRelationshipTitleColumnName();
+
             return $relationshipQuery
                 ->pluck("{$relationshipTitleColumnName} as {$relationshipTitleColumnName}", $keyName)
                 ->toArray();
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -565,6 +565,7 @@ class Select extends Field implements Contracts\HasNestedRecursiveValidationRule
             }
 
             $relationshipTitleColumnName = $component->getRelationshipTitleColumnName();
+
             return $relationshipQuery
                 ->pluck("{$relationshipTitleColumnName} as {$relationshipTitleColumnName}", $relatedKeyName)
                 ->toArray();
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -433,7 +433,7 @@ class Select extends Field implements Contracts\HasNestedRecursiveValidationRule
             }
 
             return $relationshipQuery
-                ->pluck($component->getRelationshipTitleColumnName(), $keyName)
+                ->pluck($component->getRelationshipTitleColumnName() . ' as ' . $component->getRelationshipTitleColumnName(), $keyName)
                 ->toArray();
         });
 
@@ -472,7 +472,7 @@ class Select extends Field implements Contracts\HasNestedRecursiveValidationRule
             }
 
             return $relationshipQuery
-                ->pluck($component->getRelationshipTitleColumnName(), $keyName)
+                ->pluck($component->getRelationshipTitleColumnName() . ' as ' . $component->getRelationshipTitleColumnName(), $keyName)
                 ->toArray();
         });
 
@@ -561,7 +561,7 @@ class Select extends Field implements Contracts\HasNestedRecursiveValidationRule
             }
 
             return $relationshipQuery
-                ->pluck($component->getRelationshipTitleColumnName(), $relatedKeyName)
+                ->pluck($component->getRelationshipTitleColumnName() . ' as ' . $component->getRelationshipTitleColumnName(), $relatedKeyName)
                 ->toArray();
         });
 
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -433,7 +433,7 @@ class Select extends Field implements Contracts\HasNestedRecursiveValidationRule
             }
 
             return $relationshipQuery
-                ->pluck($component->getRelationshipTitleColumnName() . ' as ' . $component->getRelationshipTitleColumnName(), $keyName)
+                ->pluck("{$component->getRelationshipTitleColumnName()} as {$component->getRelationshipTitleColumnName()}", $keyName)
                 ->toArray();
         });
 
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -472,7 +472,7 @@ class Select extends Field implements Contracts\HasNestedRecursiveValidationRule
             }
 
             return $relationshipQuery
-                ->pluck($component->getRelationshipTitleColumnName() . ' as ' . $component->getRelationshipTitleColumnName(), $keyName)
+                ->pluck("{$component->getRelationshipTitleColumnName()} as {$component->getRelationshipTitleColumnName()}", $keyName)
                 ->toArray();
         });
 
@@ -561,7 +561,7 @@ class Select extends Field implements Contracts\HasNestedRecursiveValidationRule
             }
 
             return $relationshipQuery
-                ->pluck($component->getRelationshipTitleColumnName() . ' as ' . $component->getRelationshipTitleColumnName(), $relatedKeyName)
+                ->pluck("{$component->getRelationshipTitleColumnName()} as {$component->getRelationshipTitleColumnName()}", $relatedKeyName)
                 ->toArray();
         });
 
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -433,6 +433,7 @@ class Select extends Field implements Contracts\HasNestedRecursiveValidationRule
             }
 
             $relationshipTitleColumnName = $component->getRelationshipTitleColumnName();
+
             return $relationshipQuery
                 ->pluck("{$relationshipTitleColumnName} as {$relationshipTitleColumnName}", $keyName)
                 ->toArray();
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by drhootch** [comment]
The related issue can be found here https://github.com/filamentphp/filament/issues/5740

File: packages/forms/src/Components/Select.php [hunk 1]
```diff
@@ -433,7 +433,7 @@ public function relationship(string | Closure $relationshipName, string | Closur
             }
 
             return $relationshipQuery
-                ->pluck($component->getRelationshipTitleColumnName(), $keyName)
+                ->pluck($component->getRelationshipTitleColumnName() . ' as ' . $component->getRelationshipTitleColumnName(), $keyName)
```
zepfietje, line 436:
```suggestion
                ->pluck("{$component->getRelationshipTitleColumnName()} as {$component->getRelationshipTitleColumnName()}", $keyName)
```

drhootch:
```suggestion
                ->pluck("{$component->getRelationshipTitleColumnName()} as {$component->getRelationshipTitleColumnName()}", $keyName)
```

packages/forms/src/Components/Select.php
```diff
@@ -472,7 +472,7 @@ public function relationship(string | Closure $relationshipName, string | Closur
...
```
drhootch:
```suggestion
                ->pluck("{$component->getRelationshipTitleColumnName()} as {$component->getRelationshipTitleColumnName()}", $keyName)
```

packages/forms/src/Components/Select.php
```diff
@@ -561,7 +561,7 @@ public function relationship(string | Closure $relationshipName, string | Closur
             }
 
             return $relationshipQuery
-                ->pluck($component->getRelationshipTitleColumnName(), $relatedKeyName)
+                ->pluck($component->getRelationshipTitleColumnName() . ' as ' . $component->getRelationshipTitleColumnName(), $relatedKeyName)
```
drhootch:
```suggestion
                ->pluck("{$component->getRelationshipTitleColumnName()} as {$component->getRelationshipTitleColumnName()}", $relatedKeyName)
```

drhootch (commented):
parsing the variables within the string

danharrin:
Please save the title column name to a temporary variable so we don't call that method twice

packages/forms/src/Components/Select.php
```diff
@@ -433,7 +433,7 @@ public function relationship(string | Closure $relationshipName, string | Closur
             }
```
drhootch, line 434:
```suggestion
$relationshipTitleColumnName = $component->getRelationshipTitleColumnName();
```

packages/forms/src/Components/Select.php
```diff
@@ -433,7 +433,7 @@ public function relationship(string | Closure $relationshipName, string | Closur
...
 
             return $relationshipQuery
-                ->pluck($component->getRelationshipTitleColumnName(), $keyName)
+                ->pluck("{$component->getRelationshipTitleColumnName()} as {$component->getRelationshipTitleColumnName()}", $keyName)
```
drhootch, line 436:
```suggestion
                ->pluck("{$relationshipTitleColumnName} as {$relationshipTitleColumnName}", $keyName)
```

packages/forms/src/Components/Select.php
```diff
@@ -472,7 +472,7 @@ public function relationship(string | Closure $relationshipName, string | Closur
             }
 
             return $relationshipQuery
```
drhootch, line 474: $relationshipTitleColumnName = $component->getRelationshipTitleColumnName();

drhootch, line 478:
```suggestion
            $relationshipTitleColumnName = $component->getRelationshipTitleColumnName();
            return $relationshipQuery
```

packages/forms/src/Components/Select.php
```diff
@@ -433,7 +433,7 @@ public function relationship(string | Closure $relationshipName, string | Closur
...
```
drhootch, line 437:
```suggestion
            $relationshipTitleColumnName = $component->getRelationshipTitleColumnName();
            return $relationshipQuery
```

packages/forms/src/Components/Select.php
```diff
@@ -561,7 +561,7 @@ public function relationship(string | Closure $relationshipName, string | Closur
...
```
drhootch, line 569:
```suggestion
            $relationshipTitleColumnName = $component->getRelationshipTitleColumnName();
            return $relationshipQuery
```

packages/forms/src/Components/Select.php
drhootch:
```suggestion
                ->pluck("{$relationshipTitleColumnName} as {$relationshipTitleColumnName}", $keyName)
```

packages/forms/src/Components/Select.php
```diff
@@ -472,7 +472,7 @@ public function relationship(string | Closure $relationshipName, string | Closur
...
-                ->pluck($component->getRelationshipTitleColumnName(), $keyName)
+                ->pluck("{$component->getRelationshipTitleColumnName()} as {$component->getRelationshipTitleColumnName()}", $keyName)
```
drhootch:
```suggestion
                ->pluck("{$relationshipTitleColumnName} as {$relationshipTitleColumnName}", $keyName)
```

packages/forms/src/Components/Select.php
```diff
@@ -561,7 +561,7 @@ public function relationship(string | Closure $relationshipName, string | Closur
             }
 
             return $relationshipQuery
-                ->pluck($component->getRelationshipTitleColumnName(), $relatedKeyName)
+                ->pluck("{$component->getRelationshipTitleColumnName()} as {$component->getRelationshipTitleColumnName()}", $relatedKeyName)
```
drhootch:
```suggestion
                ->pluck("{$relationshipTitleColumnName} as {$relationshipTitleColumnName}", $relatedKeyName)
```

packages/forms/src/Components/Select.php
```diff
@@ -432,8 +432,9 @@ public function relationship(string | Closure $relationshipName, string | Closur
                     ->toArray();
             }
 
+            $relationshipTitleColumnName = $component->getRelationshipTitleColumnName();
```
zepfietje, line 435:
```suggestion
            $relationshipTitleColumnName = $component->getRelationshipTitleColumnName();

```

packages/forms/src/Components/Select.php
```diff
@@ -471,8 +473,9 @@ public function relationship(string | Closure $relationshipName, string | Closur
...
             return $relationshipQuery
```
zepfietje:
```suggestion
            $relationshipTitleColumnName = $component->getRelationshipTitleColumnName();

            return $relationshipQuery
```

packages/forms/src/Components/Select.php
```diff
@@ -560,8 +563,9 @@ public function relationship(string | Closure $relationshipName, string | Closur
                     ->toArray();
             }
 
+            $relationshipTitleColumnName = $component->getRelationshipTitleColumnName();
```
zepfietje, line 567:
```suggestion
            $relationshipTitleColumnName = $component->getRelationshipTitleColumnName();

```

danharrin:
Thanks

danharrin:
Unfortunately I had to revert these changes @drhootch, they broke quite a lot of Eloquent features such as casting (which is used by the spatie/translatable package to handle JSON fields).

If you want to use JSON, you must use `getOptionLabelFromRecordUsing(fn ($record) => $record->field['json'])`

drhootch:
> Unfortunately I had to revert these changes @drhootch, they broke quite a lot of Eloquent features such as casting (which is used by the spatie/translatable package to handle JSON fields).
> 
> If you want to use JSON, you must use `getOptionLabelFromRecordUsing(fn ($record) => $record->field['json'])`

@danharrin Indeed there is a conflict with spatie/translatable package (unfortunately),
But there is still hope to use Json columns with relationship without the need for getOptionLabelFromRecordUsing(),
So when using :
```php
\Filament\Forms\Components\Select::make('select')->relationship('relation', 'data->name');
```
the following error is thrown
```
Undefined property: stdClass::$data->name
```
Therefore an alias 'as' must be added to the relationshipTitleColumnName :
```php
Select::make('select')->relationship('relation', 'data->name as name');
```
But this wont do in current Select class because orderBy in Select is also using $component->getRelationshipTitleColumnName() which breaks with the following error :
```
SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'as `name` asc' at line 1 (Connection: mysql, SQL: select json_unquote(json_extract(`data`, '$."name"')) as `name`, `id` from `table` order by json_unquote(json_extract(`data`, '$."name"')) as `name` asc)
```
So to fix this there is a working solution with Forms/Components/Select without changing the code :
- By customizing the database query that retrieves options using the third parameter of the relationship() method:
```php
use Illuminate\Database\Eloquent\Builder;
 
Select::make('select')->relationship('relation', 'data->name as name', fn (Builder $query) => $query->orderBy('data->name'))
```

PR #5745, opened 2023-02-20
Adds note on testing with deferLoading()

archilex:
Resource list tests will fail if you are using deferLoading() as the livewire component hasnt loaded the records yet. To fix you need to call `->set('isTableLoaded', true)` before calling `assertCanSeeTableRecords()`. 

This PR adds a note to the documentation about this.

### DIFF
--- a/packages/tables/docs/07-testing.md
+++ b/packages/tables/docs/07-testing.md
@@ -33,7 +33,7 @@ it('can list posts', function () {
 
 > Note that if your table uses pagination, `assertCanSeeTableRecords()` will only check for records on the first page. To switch page, call `set('page', 2)`.
 
-> Note that if your table uses deferLoading(), you must call `set('isTableLoaded', true)` before `assertCanSeeTableRecords()` to force the the component to render the records.
+> Note that if your table uses `deferLoading()`, you should call `set('isTableLoaded', true)` before `assertCanSeeTableRecords()` to render the table records directly.
 
 ## Columns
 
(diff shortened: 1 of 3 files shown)


packages/tables/docs/07-testing.md
```diff
@@ -33,6 +33,8 @@ it('can list posts', function () {
 
 > Note that if your table uses pagination, `assertCanSeeTableRecords()` will only check for records on the first page. To switch page, call `set('page', 2)`.
 
+> Note that if your table uses deferLoading(), you must call `set('isTableLoaded', true)` before `assertCanSeeTableRecords()` to force the the component to render the records.
```
zepfietje:
```suggestion
> Note that if your table uses `deferLoading()`, you should call `set('isTableLoaded', true)` before `assertCanSeeTableRecords()` to render the table records directly.
```

danharrin:
How about adding a testing helper called `loadTable()` and then documenting that? It should just call `set()` under the hood.

I don't want to add `set()` to the docs, it's tied to the internal implementation of the feature too much.

zepfietje:
Agree, good idea, @danharrin!
Could you take this on, @archilex?

archilex:
Great idea. Will do

archilex:
Looking at some of your other test methods they seemed to call their setter methods instead of setting the property so I went with that here as well. Manually tested and works as expected.

packages/tables/docs/07-testing.md
```diff
@@ -33,6 +33,8 @@ it('can list posts', function () {
 
 > Note that if your table uses pagination, `assertCanSeeTableRecords()` will only check for records on the first page. To switch page, call `set('page', 2)`.
 
+> Note that if your table uses `deferLoading()`, you should call `loadTable()` before `assertCanSeeTableRecords()` to render the table records directly.
```
zepfietje:
```suggestion
> Note that if your table uses `deferLoading()`, you should call `loadTable()` before `assertCanSeeTableRecords()`.
```

zepfietje (approved):

danharrin:
Thanks

PR #5751, opened 2023-02-21
Update 02-listing-records.md

BrendanTWhite:
May I suggest mentioning that you need to add the `sort` to $fillable on the model, otherwise the sorting doesn't work. I know this, because I got stuck on this for ages! :-)

### DIFF
--- a/packages/admin/docs/02-resources/02-listing-records.md
+++ b/packages/admin/docs/02-resources/02-listing-records.md
@@ -198,7 +198,7 @@ public static function table(Table $table): Table
 }
 ```
 
-You will also need to add the `sort` property to the `$fillable` array on your `Model`.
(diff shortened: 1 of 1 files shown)


packages/admin/docs/02-resources/02-listing-records.md
```diff
@@ -198,6 +198,8 @@ public static function table(Table $table): Table
 }
 ```
 
+You will also need to add the `sort` property to the `$fillable` array on your `Model`.
```
danharrin:
```suggestion
If you're using mass assignment protection on your model, you will also need to add the `sort` attribute to the `$fillable` array there.
```

danharrin:
Thanks

PR #5780, opened 2023-02-24
TextInputColumn. Debounce numeric input

tanthammar:
Debounce input where type = number

If you have a numeric input and the value is changed via the inputs "arrow buttons" or the keyboard arrow up/down.
A network request is generated for every click/keystroke.

This PR tries to prevent that.

### DIFF
--- a/packages/tables/resources/views/columns/text-input-column.blade.php
+++ b/packages/tables/resources/views/columns/text-input-column.blade.php
@@ -13,6 +13,13 @@
         error: undefined,
         state: @js($state),
         isLoading: false,
+        update: async function(value) {
+            this.isLoading = true
+            this.response = await $wire.updateTableColumnState(@js($getName()), @js($recordKey), value)
+            this.error = response?.error ?? undefined
+            if (! this.error) this.state = response
+            this.isLoading = false
+        }
     }"
     x-init="
         Livewire.hook('message.processed', (component) => {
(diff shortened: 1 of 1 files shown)


danharrin:
We can't update the record on enter, that's a breaking change, so please remove that. We can keep the debounce for numeric inputs though

tanthammar:
Can you please explain how it is a breaking change? Are there other components triggered by keydown.enter from this input?
If I add .prevent.stop, will that solve it?

danharrin:
It's a breaking change as it changes how users interact with the system. I don't think it's clear that they need to press enter to save, and end-users may have been trained differently.

packages/tables/resources/views/columns/text-input-column.blade.php
```diff
@@ -46,13 +53,7 @@
...
         {!! ($placeholder = $getPlaceholder()) ? "placeholder=\"{$placeholder}\"" : null !!}
         {!! ($interval = $getStep()) ? "step=\"{$interval}\"" : null !!}
-        x-on:change="
-            isLoading = true
-            response = await $wire.updateTableColumnState(@js($getName()), @js($recordKey), $event.target.value)
-            error = response?.error ?? undefined
-            if (! error) state = response
-            isLoading = false
-        "
+        x-on:change{{ $getType() === 'number' ? '.debounce.750ms' : null }}="update($event.target.value)"
```
danharrin:
```suggestion
        x-on:change{{ $getType() === 'number' ? '.debounce.1s' : null }}="
            isLoading = true
            response = await $wire.updateTableColumnState(@js($getName()), @js($recordKey), $event.target.value)
            error = response?.error ?? undefined
            if (! error) state = response
            isLoading = false
        "
```

packages/tables/resources/views/columns/text-input-column.blade.php
```diff
@@ -13,6 +13,13 @@
         error: undefined,
         state: @js($state),
         isLoading: false,
+        update: async function(value) {
+            this.isLoading = true
+            this.response = await $wire.updateTableColumnState(@js($getName()), @js($recordKey), value)
+            this.error = response?.error ?? undefined
+            if (! this.error) this.state = response
+            this.isLoading = false
+        }
```
danharrin:
```suggestion
```

PR #5782, opened 2023-02-24
Create new filter layout: AboveContentCollapsable

lucasgiovanny:
Hey everyone! 👋🏼 

This PR creates a new layout for the filter table called `AboveContentCollapsable` to allow us to have the behavior below:

ezgif com-video-to-gif

Sometimes we have too many fields to be displayed above content immediately so it would be nice to the possibility to have a button to collapse it. 😄

I would love to have this appreciated. What do you think?

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -30,7 +30,8 @@
     $isLoaded = $isLoaded();
     $hasFilters = $isFilterable();
     $hasFiltersPopover = $hasFilters && ($getFiltersLayout() === FiltersLayout::Popover);
-    $hasFiltersAboveContent = $hasFilters && ($getFiltersLayout() === FiltersLayout::AboveContent);
+    $hasFiltersAboveContent = $hasFilters && ($getFiltersLayout() === FiltersLayout::AboveContent || FiltersLayout::AboveContentColapsable);
+    $hasFiltersAboveContentColapsable = $hasFilters && ($getFiltersLayout() === FiltersLayout::AboveContentColapsable);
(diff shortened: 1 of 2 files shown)


packages/tables/resources/views/index.blade.php
```diff
@@ -30,7 +30,8 @@
     $isLoaded = $isLoaded();
     $hasFilters = $isFilterable();
     $hasFiltersPopover = $hasFilters && ($getFiltersLayout() === FiltersLayout::Popover);
-    $hasFiltersAboveContent = $hasFilters && ($getFiltersLayout() === FiltersLayout::AboveContent);
+    $hasFiltersAboveContent = $hasFilters && ($getFiltersLayout() === FiltersLayout::AboveContent || $getFiltersLayout() === FiltersLayout::AboveContentColapsable);
```
danharrin: There's a typo in the constant name :)

lucasgiovanny: Oh, sorry about that 😄 Fixed!

danharrin:
I think we use "Collapsible" elsewhere instead of "Collapsable"

danharrin:
Thanks :)

jcgmalave:
How can I associate the AboveContentCollapsable property to the table filter, from the Resource? Is it from the main resource or from the list? Thanks in advanced.

lucasgiovanny:
> How can I associate the AboveContentCollapsable property to the table filter, from the Resource? Is it from the main resource or from the list? Thanks in advanced.

You need to declare the method below on your **ListPage**:

```php
protected function getTableFiltersLayout(): ?string
{
    return Layout::AboveContentCollapsible;
}
```

I will see to do a PR later today to update the docs on the website.

PR #5783, opened 2023-02-24
Update Tailwind config preset and stub

zepfietje:
Edit: I just realized we probably want a different stub for apps using `filament/app` and those without. Thoughts?

### DIFF
--- a/docs-assets/app/app/Providers/Filament/configure.php
+++ b/docs-assets/app/app/Providers/Filament/configure.php
@@ -5,3 +5,4 @@ use Filament\Context;
 return function (Context $context) {
     //
 };
+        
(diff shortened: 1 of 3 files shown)


danharrin:
I agree, as the CSS-var based colours don't apply outside of the app framework

danharrin:
The scaffolding isn't for building themes, it's for the individual packages. So I think how it is right now is fine

zepfietje:
But we don't offer a scaffold for building themes?

danharrin:
Not yet, that should probably be a separate `make:filament-theme` command

danharrin:
this is the install command, remember

zepfietje:
I know, but personally I often also need a Tailwind config for the non-Filament part of the app. Compile it all into a single stylesheet. But I see how this isn't needed if you're not customizing any of the Filament styling except colors and such for which PHP setters exist. Let's just leave this as is then.

danharrin:
And if we want to push people towards using the app framework for everything then...

zepfietje:
Agree. Marketing sites would be separate and out of scope for Filament core. 👍

PR #5789, opened 2023-02-25
Fix heroicon title attribute

ganyicz:
The value is being rendered as a literal string instead of compiling the php due to missing column

Screenshot 2023-02-25 at 12 10 02


danharrin:
Thanks

PR #5792, opened 2023-02-25
Table action default styling

danharrin:


packages/notifications/src/Actions/Action.php
```diff
@@ -35,7 +35,9 @@ protected function setUp(): void
     {
         parent::setUp();
 
-        $this->size('sm');
+        $this->defaultView(static::LINK_VIEW);
```
zepfietje, line 33: View property at top of this class can be removed?

zepfietje:
Rest looks good to me.

zepfietje:
Thanks, @danharrin!

PR #5794, opened 2023-02-25
Adds a new method for testing individual search columns

joaoprado:
Thank you for this great package. This PR introduces a new method to test individual column search when global search is turned off.

### DIFF
--- a/packages/tables/.stubs.php
+++ b/packages/tables/.stubs.php
@@ -129,7 +129,7 @@ namespace Livewire\Testing {
 
         public function searchTable(?string $search = null): static {}
 
-        public function searchTableColumns(array $search): static {}
+        public function searchTableColumns(array $searches): static {}
 
         public function filterTable(string $name, $data = null): static {}
 
--- a/packages/tables/.stubs.php
+++ b/packages/tables/.stubs.php
@@ -129,7 +129,7 @@ namespace Livewire\Testing {
 
         public function searchTable(?string $search = null): static {}
 
-        public function searchTableColumn(array|string $column, ?string $search = null): static {}
+        public function searchTableColumns(array $search): static {}
 
         public function filterTable(string $name, $data = null): static {}
 
--- a/packages/tables/.stubs.php
+++ b/packages/tables/.stubs.php
@@ -129,6 +129,8 @@ namespace Livewire\Testing {
 
         public function searchTable(?string $search = null): static {}
 
+        public function searchTableColumn(string $column, ?string $search = null): static {}
+
         public function filterTable(string $name, $data = null): static {}
 
         public function resetTableFilters(): static {}
--- a/packages/tables/.stubs.php
+++ b/packages/tables/.stubs.php
@@ -129,7 +129,7 @@ namespace Livewire\Testing {
 
         public function searchTable(?string $search = null): static {}
 
-        public function searchTableColumn(string $column, ?string $search = null): static {}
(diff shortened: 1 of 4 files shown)


packages/tables/src/Testing/TestsColumns.php
```diff
@@ -429,4 +429,13 @@ public function searchTable(): Closure
             return $this;
         };
     }
+
+    public function searchTableColumn(): Closure
+    {
+        return function (string $column, ?string $search = null): static {
+            $this->set('tableColumnSearchQueries', [$column => $search]);
```
zepfietje: Thanks for the PR, @joaoprado!

Maybe we should handle searching multiple columns as well?

joaoprado: That is a good idea, for multiple columns I'm thinking the first parameter should be an array and be passed as the `$this->set('tableColumnSearchQueries', $column)` parameter.

Do I need to run any validation with this approach? Like validating the array structure

zepfietje: We might rename this to be `searchTableColumns` instead and accept an array with keys being the column and value being the search string? No validation needed.

joaoprado: With that approach we are assuming the user is trying to search two columns with the same value and, at least in the use cases I've seen for individual column searching, that seems to not be the intention. It is used most of the time to search records with different values for each column.

I have added a new commit with the approach I mentioned and a [test case](https://github.com/filamentphp/filament/pull/5794/files#diff-4d5f6108cd32b292ec5aa83a683d89a2127590cabb4c97681c5e89f87a6eec88R88) to exemplify this.

zepfietje: No, passing an array allows searching multiple values:
```php
searchTableColumns([
    'name' => 'John',
    'city' => 'London',
])
```

zepfietje: Could you update this PR to that?

joaoprado: Oh ok, I missunderstood what you said before. So we will always want an array of keys and values even if it is just one column. I will send a new commit

packages/tables/.stubs.php
```diff
@@ -129,6 +129,8 @@ public function sortTable(?string $name = null, ?string $direction = null): stat
 
         public function searchTable(?string $search = null): static {}
 
+        public function searchTableColumns(array $search): static {}
```
danharrin:
```suggestion
        public function searchTableColumns(array $searches): static {}
```

packages/tables/src/Testing/TestsColumns.php
```diff
@@ -429,4 +429,13 @@ public function searchTable(): Closure
             return $this;
         };
     }
+
+    public function searchTableColumns(): Closure
+    {
+        return function (array $search): static {
```
danharrin:
```suggestion
        return function (array $searches): static {
```

packages/tables/src/Testing/TestsColumns.php
```diff
@@ -429,4 +429,13 @@ public function searchTable(): Closure
...
+            $this->set('tableColumnSearchQueries', $search);
```
danharrin:
```suggestion
            $this->set('tableColumnSearchQueries', $searches);
```

danharrin:
Thanks! Nice feature

PR #5803, opened 2023-02-27
fix richeditor ul & ol padding on rtl

aymanalareqi:
this PR fix issue on the following image
rtl richeditor padding issue


packages/forms/resources/css/components/rich-editor.css
```diff
@@ -22,3 +22,15 @@
 trix-toolbar .filament-forms-rich-editor-component-toolbar-button.trix-active {
     @apply bg-primary-600 text-white;
 }
+
+[dir='rtl'] .prose :where(ul):not(:where([class~='not-prose'] *)),
```
zepfietje, line 26: @danharrin, this CSS isn't scoped to the rich-editor in any way... sure this won't cause issues in other places?

danharrin: true

danharrin: should be fixed

zepfietje: @danharrin, could you revert this PR?
Then @aymanalareqi can attempt a new PR with a correct fix?

zepfietje: Never mind, I didn't read your "should be fixed" as "has been fixed". 😄

PR #5810, opened 2023-02-28
Make form actions sticky

JohnnyMaynne:
Adding sticky form actions if they don't fit on the page

### DIFF
--- a/packages/admin/config/filament.php
+++ b/packages/admin/config/filament.php
@@ -232,6 +232,7 @@ return [
         'forms' => [
             'actions' => [
                 'alignment' => 'left',
+                'sticky' => false
             ],
             'have_inline_labels' => false,
         ],
--- a/packages/admin/config/filament.php
+++ b/packages/admin/config/filament.php
@@ -232,7 +232,7 @@ return [
         'forms' => [
             'actions' => [
                 'alignment' => 'left',
-                'sticky' => false
+                'are_sticky' => false
             ],
             'have_inline_labels' => false,
         ],
--- a/packages/admin/resources/views/components/form/actions.blade.php
+++ b/packages/admin/resources/views/components/form/actions.blade.php
@@ -1,11 +1,32 @@
-@props([
-    'actions',
-    'fullWidth' => false,
-])
+@props(['actions', 'fullWidth' => false])
 
-<x-filament::pages.actions
-    :actions="$actions"
-    :alignment="config('filament.layout.forms.actions.alignment')"
-    :full-width="$fullWidth"
-    class="filament-form-actions"
-/>
+<div
+    x-data="{
+        sticky: false,
+        scrollPage() {
+            let documentHeight = document.body.scrollHeight;
+            let currentScroll = window.scrollY + window.innerHeight;
+            let modifier = 200;
+            if (currentScroll + modifier > documentHeight) {
+                this.sticky = false
+            } else {
+                this.sticky = true
+            }
+        }
+    }"
+    x-on:scroll.window="scrollPage"
+    x-init="scrollPage"
+    class="transform"
+    x-bind:class="{
+        'sticky left-0 bottom-0 mx-auto max-w-2xl -translate-y-4 rounded-md border border-gray-300 bg-white p-3 shadow-3xl transition-colors transition-shadow transition-transform': sticky,
+        'dark:border-gray-700 dark:bg-gray-800 dark:shadow-none': !!{{ config('filament.dark_mode') }} && sticky,
+        'translate-y-0': !sticky
+    }"
+>
+    <x-filament::pages.actions
+        :actions="$actions"
+        :alignment="config('filament.layout.forms.actions.alignment')"
+        :full-width="$fullWidth"
+        class="filament-form-actions"
+    />
+</div>
--- a/packages/admin/resources/views/components/form/actions.blade.php
+++ b/packages/admin/resources/views/components/form/actions.blade.php
@@ -4,7 +4,7 @@
 ])
 
 <div
-    @if(config('filament.layout.forms.actions.sticky'))
+    @if (config('filament.layout.forms.actions.are_sticky'))
         x-data="{
             sticky: false,
             scrollPage() {
--- a/packages/admin/resources/views/components/form/actions.blade.php
+++ b/packages/admin/resources/views/components/form/actions.blade.php
@@ -6,25 +6,24 @@
 <div
     @if (config('filament.layout.forms.actions.are_sticky'))
         x-data="{
-            sticky: false,
-            scrollPage() {
-                let documentHeight = document.body.scrollHeight;
-                let currentScroll = window.scrollY + window.innerHeight;
-                let modifier = 200;
-                if (currentScroll + modifier > documentHeight) {
-                    this.sticky = false
-                } else {
-                    this.sticky = true
-                }
-            }
+        
+            areSticky: false,
+            
+            evaluatePageScrollPosition: function() {
+                let documentHeight = document.body.scrollHeight
+                let currentScroll = window.scrollY + window.innerHeight
+                
+                this.areSticky = (currentScroll + 200) <= documentHeight
+            },
+            
         }"
-        x-on:scroll.window="scrollPage"
-        x-init="scrollPage"
+        x-init="evaluatePageScrollPosition"
+        x-on:scroll.window="evaluatePageScrollPosition"
         class="transform"
         x-bind:class="{
-            'sticky left-0 bottom-0 mx-auto max-w-2xl -translate-y-4 rounded-md border border-gray-300 bg-white p-3 shadow-3xl transition-colors transition-shadow transition-transform': sticky,
-            'dark:border-gray-700 dark:bg-gray-800 dark:shadow-none': !!{{ config('filament.dark_mode') }} && sticky,
-            'translate-y-0': !sticky
+            'sticky left-0 bottom-0 mx-auto max-w-2xl -translate-y-4 rounded-md border border-gray-300 bg-white p-3 shadow-3xl transition-colors transition-shadow transition-transform': areSticky,
+            'dark:border-gray-700 dark:bg-gray-800 dark:shadow-none': @js(config('filament.dark_mode')) && areSticky,
+            'translate-y-0': ! areSticky
         }"
     @endif
 >
--- a/packages/admin/resources/views/components/form/actions.blade.php
+++ b/packages/admin/resources/views/components/form/actions.blade.php
@@ -1,4 +1,7 @@
-@props(['actions', 'fullWidth' => false])
+@props([
+    'actions',
+    'fullWidth' => false,
+])
 
 <div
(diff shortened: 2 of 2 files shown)


zepfietje:
I'm not too sure if this should actually be the default.
Let's discuss it first before merging.

Thoughts, @danharrin @awcodes?

awcodes:
I'm ok with default, as long as there is a way to turn it off should that be what the user wants. Could treat it like the topbar where it could be disabled via CSS.

JohnnyMaynne:
@zepfietje @awcodes 
I've added a setting to the config file that disables sticky by default

packages/admin/resources/views/components/form/actions.blade.php
```diff
@@ -1,11 +1,34 @@
-@props([
-    'actions',
-    'fullWidth' => false,
-])
+@props(['actions', 'fullWidth' => false])
```
zepfietje:
```suggestion
@props([
    'actions',
    'fullWidth' => false,
])
```

zepfietje:
@JohnnyMaynne, can't this be achieved using a CSS-only solution by the way?

Also, mind sharing a screenshot/recording of how it works after this PR?
Allows me to review a little easier without having to pull in these changes locally.

awcodes:
@zepfietje you can make it sticky without this code, but this also changes the styling which can't be done through CSS alone since there's currently no way to target when an element is "stuck" to change it's appearance.

JohnnyMaynne:
@zepfietje @awcodes 

https://discord.com/channels/883083792112300104/883084832387760148/1080074348964098058

the whole point is that actions stick only when they are "hiding"

awcodes:
For reference @zepfietje I did a similar thing without changing the styling when it's stuck. @JohnnyMaynne approach just looks better too, IMO.

Screen Shot 2023-02-28 at 8 46 59 AM

zepfietje:
> @zepfietje you can make it sticky without this code, but this also changes the styling which can't be done through CSS alone since there's currently no way to target when an element is "stuck" to change it's appearance.

Yeah that's what I thought.
But then the question is, do we really need to change styling when it's sticky?

JohnnyMaynne:
If left transparent or not framed, it will be difficult to read from the page. The second point, it seems to me that such sticky elements should take up as little space as possible on the page, I would like to see the buttons in the center and the footer itself even less

zepfietje:
Thanks for pointing me to the video, @JohnnyMaynne.
Personally I think the approach @awcodes showed is a little less opinionated.

I'm not a big fan, but I think the solid full-width actions bar is still most common.

JohnnyMaynne:
@zepfietje as you say 😊

awcodes:
I like @JohnnyMaynne better than mine. 😅

ryangjchandler:
Would also be nice to enable it per-page instead of globally.

JohnnyMaynne:
> Would also be nice to enable it per-page instead of globally.

I also thought about it, but most likely the user should get used to ux, which moves from page to page

Wulfheart:
I have thought about it a lot recently and this would be a so huge UX improvement in my opinion.

howdu:
What's the plan for small screen sizes considering the header is already taking up space. 
IMG_6AEAF85D04B6-1

zepfietje:
Good point, @howdu. I think that would be another persistent bar at the bottom then...

danharrin:
Let me know when you're happy with this @zepfietje or @awcodes

awcodes:
I'm good with it. Zep might feel differently.

zepfietje:
I don't necessarily dislike @JohnnyMaynne's implementation, but I'd like to explore if a simple CSS-only solution could work too.

Also, I think we should make this behavior optional.
As @howdu pointed out, on mobile you'll lose a lot of space, so might prefer to not make it sticky there.
This would all be easier with a pure CSS solution.

Wulfheart:
I think especially such a hovering bar would be quite complicated using CSS only but tbh my CSS skills are quite limited.

andrewdwallo:
Does this apply to having multiple forms on one page? If so, how would someone be able to separate forms with their own separate notifications on save? If I have multiple forms that each have a different action on one page, how would this behave?

JohnnyMaynne:
> Относится ли это к наличию нескольких форм на одной странице? Если да, то как кто-то сможет разделить формы со своими отдельными уведомлениями при сохранении? Если у меня есть несколько форм, каждая из которых имеет свое действие на одной странице, как это будет себя вести?

This applies to the admin panel, the form generation package only generates inputs, not the form itself and its actions

JohnnyMaynne:
> Does this apply to having multiple forms on one page? If so, how would someone be able to separate forms with their own separate notifications on save? If I have multiple forms that each have a different action on one page, how would this behave?

This applies to the admin panel, the form generation package only generates inputs, not the form itself and its actions

JohnnyMaynne:
> I don't necessarily dislike @JohnnyMaynne's implementation, but I'd like to explore if a simple CSS-only solution could work too.
> 
> Also, I think we should make this behavior optional. As @howdu pointed out, on mobile you'll lose a lot of space, so might prefer to not make it sticky there. This would all be easier with a pure CSS solution.

If you don't need it in the core, you don't merge it, I was asked - I did it. In any case, I will use this behavior in my projects.

PS The setting that enables this behavior has been moved to the config

Wulfheart:
Tbh I would need this really much. This would have a huge UX improvement.

andrewdwallo:
> This applies to the admin panel, the form generation package only generates inputs, not the form itself and its actions

Okay sounds good just wanted to make sure in those cases. I'm totally for it then.

awcodes:
Seems like a regression. Will try to fix it tomorrow.

packages/admin/config/filament.php
```diff
@@ -232,6 +232,7 @@
         'forms' => [
             'actions' => [
                 'alignment' => 'left',
+                'sticky' => false
```
danharrin:
```suggestion
                'are_sticky' => false
```

packages/admin/resources/views/components/form/actions.blade.php
```diff
@@ -3,9 +3,35 @@
...
 ])
 
-<x-filament::pages.actions
-    :actions="$actions"
-    :alignment="config('filament.layout.forms.actions.alignment')"
-    :full-width="$fullWidth"
-    class="filament-form-actions"
-/>
+<div
+    @if(config('filament.layout.forms.actions.sticky'))
```
danharrin:
```suggestion
    @if (config('filament.layout.forms.actions.are_sticky'))
```

packages/admin/resources/views/components/form/actions.blade.php
```diff
@@ -3,9 +3,35 @@
...
+            }
+        }"
+        x-on:scroll.window="scrollPage"
+        x-init="scrollPage"
+        class="transform"
+        x-bind:class="{
+            'sticky left-0 bottom-0 mx-auto max-w-2xl -translate-y-4 rounded-md border border-gray-300 bg-white p-3 shadow-3xl transition-colors transition-shadow transition-transform': sticky,
+            'dark:border-gray-700 dark:bg-gray-800 dark:shadow-none': !!{{ config('filament.dark_mode') }} && sticky,
+            'translate-y-0': !sticky
+        }"
```
danharrin:
```suggestion
    @if (config('filament.layout.forms.actions.are_sticky'))
        x-data="{
        
            areSticky: false,
            
            evaluatePageScrollPosition: function() {
                let documentHeight = document.body.scrollHeight
                let currentScroll = window.scrollY + window.innerHeight
                
                this.areSticky = (currentScroll + 200) <= documentHeight
            },
            
        }"
        x-init="evaluatePageScrollPosition"
        x-on:scroll.window="evaluatePageScrollPosition"
        class="transform"
        x-bind:class="{
            'sticky left-0 bottom-0 mx-auto max-w-2xl -translate-y-4 rounded-md border border-gray-300 bg-white p-3 shadow-3xl transition-colors transition-shadow transition-transform': areSticky,
            'dark:border-gray-700 dark:bg-gray-800 dark:shadow-none': @js(config('filament.dark_mode')) && areSticky,
            'translate-y-0': ! areSticky
        }"
```

danharrin:
I updated the styling a bit and added a CSS hook class

archilex:
Playing around with the new release and see that on mobile the sticky bar still has the `-transalte-y-4` on it. IMO it looks a little weird being full-width but also separated at the bottom. Was this intentional? I can PR a fix if needed. 

Screenshot 2023-03-12 at 10 55 36 PM

Also, looking at the discussion above, wondering if you'd accept a PR to allow the full-width as a configurable option? Mayeb `is_full_width`. With the css hook I can just do it manually, but feel others might also prefer full-width as opposed to floating?

zepfietje:
Agree that should be fixed on mobile, @archilex.
Would be great if you could create a PR.

Like I've commented here before, I think the full width might actually be a better default.

danharrin:
If you don't like full width, use CSS to change it.

As for the -translate-y-4, it just needs an `md:-translate-y-4` probably

archilex:
Yup, easy fix. PR #5918 

Sounds good. Will use the css hook.

zepfietje:
> If you don't like full width, use CSS to change it.

@danharrin, the point is that full width is not the default currently.

danharrin:
it is, I made that change before I merged

archilex:
I think I introduced some confusion. I was referring to making it completly full-width like awcode's screenshot above. Right now it's a "floating" pill stuck to the bottom. 

That said, there actually doesn't seem to be a way to make it completely full width using the default filament settings since by default there's a max-w-7xl on the parent container which limits the width of the sticky-footer at larger sizes.

zepfietje:
I'm confused now. 😅
@danharrin says it is full width, while @archilex says it's not?

danharrin:
Ah right, I meant full page (content) width.

I think you can still make it really full width with CSS though?

archilex:
haha! Johnny mentioned at the beginning that he liked a floating pill that was just the width of the action buttons. That was his original PR I believe with it centered in the window and just the width of the buttons. 

Then there was discussion about it being "full-width", but really it's just the full width of the FORM container. That's what Dan did. He made the action bar the width of the form container (it sticks out a bit on the sides so it's more obvious it's there, but it's only the width of the form. 

I'm referring to it being the full width of the page container so it goes all the way to the sides, like Awcodes' screenshot.

awcodes:
@archilex i think my screenshot might just be tricking your eye. It's not full width in that one either. There's just no back ground around them so it just looks that way. The buttons are still bound by the max width container.

archilex:
@awcodes gotcha. Here's an example of what I thought you were showing with your screenshot:

Screenshot 2023-03-13 at 1 08 08 PM

awcodes:
Yea. I did that with the Sticky Header plugin. It's a lot more difficult to pull off. 😄

archilex:
Yeah seems like it! Maybe Sticky Footer plugin??? 🤣 Anyway, what we have now is great with this current implemtation is a great improvement! Thanks everyone.

PR #5818, opened 2023-03-01
Improve table grouping

zepfietje:
- Makes it match the existing design better.
- Add `titlePrefixedWithLabel` API.
- Make entire group title row clickable.


packages/tables/resources/views/index.blade.php
```diff
@@ -490,36 +490,48 @@ class="text-xs pl-2 pr-6 py-1 font-medium border-0 bg-gray-500/5 rounded-lg bord
                                 @endif
 
                                 <div @class([
```
zepfietje, line 492: @danharrin, could you check if this part still functions correctly?
I've only been able to check regular tables.

zepfietje: Checked. Looks good.

PR #5821, opened 2023-03-01
Prevents duplicate queries using the SelectFilter

masterix21:
Hi, dears!

Using the `SelectFilter` component and applying a filter, it executes the query two times in the same request.

The PR adds the ability to execute the query only once, caching the results in the `$this->options` property for subsequent usages.

### DIFF
--- a/packages/tables/src/Filters/Concerns/HasOptions.php
+++ b/packages/tables/src/Filters/Concerns/HasOptions.php
@@ -32,6 +32,8 @@ trait HasOptions
             $options = $options->toArray();
         }
 
+        $this->options = $options;
+
         return $options;
     }
 }


zepfietje:
Thanks for the PR, @masterix21.
This is a known 'issue', which will be fixed in v3.

For now, make sure to wrap your query in a closure `->options(fn () => ...)`.

masterix21:
@zepfietje, I'm sorry, but `->options(fn () => ...)` isn't a solution: the queries always remain two when the filter is applied. Try by yourself: this small PR fixes a bug before the V3 version that requires an _undefined_ Livewire v3.

zepfietje:
@danharrin, could you have a look?

danharrin:
In this case, please use the `spatie/once` package to cache this value. The fix has side effects - if the options are dynamic and change, they don't get updated again. v3 will provide the fix for this.

PR #5825, opened 2023-03-01
fix: Nested repeater actions

### LINKED ISSUE #5812: Actions of nested components in repeater not executed properly
### Package

filament/forms

### Package Version

2.17.8

### Laravel Version

10.1.5

### Livewire Version

2.12.1

### PHP Version

8.1.6

### Problem description

When a repeater has nested components, such as grids with some input components, and one or more of those input components have an action, the action is only fired for the last repeater item. For the other items, the loading indicator is shown when the action is clicked, a server request is executed with the correct action, however, nothing happens because the function `getMountedFormComponentAction` returns `null` instead of the desired action. 

I looked further into this and it seems because when `getMountedFormComponentAction` retrieves the listed of components by using `getFlatComponents`, it only returns the components of the last repeater item.

### Expected behavior

It might be that it is not desired to use grid or group components to nest components inside a repeater, but at least I thought of it as a way to properly organize the components inside the repeater item. If it indeed should be possible, then I expect the actions of the nested components to work, which they now don't.

### Steps to reproduce

- Create a form with a repeater component
- Inside the schema, put a grid or group component
- Put a input component, e.g. text input, inside the grid or group component
- Add a suffix action to the input component
- When having at least two repeater items, you will see that the actions of all repeater items but the last will not be executed properly

### Reproduction repository

https://github.com/damiandejong/filament-nested-repeater-components

### Relevant log output

_No response_

danharrin:
Fixes #5812.

### DIFF
--- a/packages/forms/src/Components/Concerns/Cloneable.php
+++ b/packages/forms/src/Components/Concerns/Cloneable.php
@@ -4,8 +4,18 @@ namespace Filament\Forms\Components\Concerns;
 
 trait Cloneable
 {
+    public function cloneChildComponents(): static
+    {
+        $components = $this->getChildComponentContainer()->getClone()->getComponents();
+
+        return $this->childComponents($components);
+    }
+
     public function getClone(): static
     {
-        return clone $this;
+        $clone = clone $this;
+        $clone->cloneChildComponents();
+
+        return $clone;
     }
 }


Saifallak:
Hello, i'm getting `Typed property Filament\Forms\Components\Component::$container must not be accessed before initialization` After this Change.

use case, i'm using helper function to create same field many times for each locale like the following

```php
/**
     * Returns Same Field With Multi Langs.
     *
     * @param Field $field
     * @return Field|Component
     */
    function translateFilamentField(Field $field): Field|Component
    {
        $locales = collect(config('app.available_locales'));
        return Group::make()
            ->columnSpanFull()
            ->columns($locales->count())
            ->schema(
                $locales->map(
                    fn (array $lang, string $key) => $field->getClone()
                        ->name($field->getName() . '.' . $key)
                        ->statePath($field->getName() . '.' . $key)
                        ->hint(__('admin.with_locale', ['locale' => $lang['native']]))
                        ->columnSpan(1),
                )->toArray(),
            );
    }
```

Using
```php
return $form
            ->columns(4)
            ->schema([
                                translateFilamentField(
                                        ->label(__('admin.name'))
                                        ->required()
                                        ->maxLength(191),
                                ),
            ]);
```

danharrin:
Maybe fixed in the next release? But really we can't protected against breaking changes in internal methods such as `getClone()`

Saifallak:
Will test tomorrow and get back to you.

Thanks, Im trying not use much of internal methods

PR #5829, opened 2023-03-02
Add row layout to checkbox list

wychoong:
while current CheckboxList can configure the columns with `->columns([..])`, the direction is always vertical, and this may not be ideal

image

this is introduced by https://github.com/filamentphp/filament/pull/3984 where the grid direction is set to be always `column`

and in the doc screenshot shows the `row` behaviour
image

This PR looks to introduce a `row()` method to configure it into a row direction

image
```php
Forms\Components\CheckboxList::make('workdays')
                        ->options(Carbon::getDays())
                        ->columns([
                            'default' => 2,
                            'lg' => 3,
                            'xl' => 4,
                        ])
                        ->row() // <- new method
```

- Radio component does not need this change as it can use `inline()`

### DIFF
--- a/packages/forms/resources/views/components/checkbox-list.blade.php
+++ b/packages/forms/resources/views/components/checkbox-list.blade.php
@@ -110,11 +110,11 @@
             :lg="$getColumns('lg')"
             :xl="$getColumns('xl')"
             :two-xl="$getColumns('2xl')"
-            :direction="$getDirection()"
+            :direction="$getGridDirection() ?? 'column'"
             :x-show="$isSearchable() ? 'visibleCheckboxListOptions.length' : null"
             :attributes="$attributes->class([
                 'filament-forms-checkbox-list-component gap-1',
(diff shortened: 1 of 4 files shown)


danharrin:
I think we should have the API `gridDirection('row')` for this. It sets the `$gridDirection` property. We need a `getGridDirection()` getter method. The default for the property is `null`, and then in the view we use `direction="$getGridDirection() ?? 'column'"`

wychoong:
> I think we should have the API `gridDirection('row')` for this. It sets the `$gridDirection` property. We need a `getGridDirection()` getter method. The default for the property is `null`, and then in the view we use `direction="$getGridDirection() ?? 'column'"`

Does `gridDirection` api with default param?

wychoong:
updated based on the comment. 

*Note that if `gridDirection('row')` no `columns([..])` set on the field, it will still shows as  single column

Issue #5639, opened 2023-02-09
Form not filled after changing language


icaliman:
### Package

filament/spatie-laravel-translatable-plugin

### Package Version

v2.16.69

### Laravel Version

v9.51.0

### Livewire Version

v2.11.2

### PHP Version

8.1.13

### Problem description

After changing language the form is not filled with data and Tabs are broken.

This happens when using `Tabs` in a form. 

https://user-images.githubusercontent.com/4498933/217831660-47a84733-7b11-4a37-b6f0-7e6b27126162.mov

### Expected behavior

After changing language the form must be filled and the tabs must work.

### Steps to reproduce

1. Create a page.
2. Add an Action block
3. Fill the form with some data
4. Change the language

### Reproduction repository

https://github.com/icaliman/filament-bug-translatable-tabs

danharrin:
Fixed by #5639

PR #5847, opened 2023-03-04
fix: tabs not updating in Builder #5639

alex552:


danharrin:
Hey, so how does this fix work? Why do we need an ID?

alex552:
Hi Dan,  forms::components.tabs doesn't have an id by default or a wire:key, that's why this bug was happening on refresh.

danharrin:
I don't think it should have an ID then, let's add a wire:key in the view

danharrin:
Thanks

PR #5848, opened 2023-03-04
SelectColumn class doesn't exist

bestiony:
I faced a problem following this doc. 
the "SelectColumn" class doesn't exist, instead, it's named "Select". you can clearly see that in the dir : "vendor\filament\forms\src\Components" after you install Filament


danharrin:
No? You're looking in the wrong directory. Tables have columns, not form components

PR #5865, opened 2023-03-06
Pass item component to isCollapsed

howdu:
Control which items are collapsed using `getRawState`
```php
Forms\Components\Repeater::make('test')
    ->collapsed(function (Forms\ComponentContainer $item): bool {
        $data = $item->getRawState();
        return $data['default'] ?? true;
    })
```


danharrin:
Thanks

PR #5868, opened 2023-03-06
Ensuring that notification actions from Notifications have full event information

titantwentyone:
This is a follow up on #5732. The original tests verified that `Action` was correctly parsing the event information (event name, data and it's direction) and generating the correct `wire:click` attribute for use in the HTML through `getLivewireMountAction`. What I missed was that when used with `Notifications`, the actions are serialized. I had put code in to unserialize the data into the array but not to put it in the array! My bad! Apologies. This PR should fix that issue and have added further tests to verify that the correct `emit` call is rendered and will be pulled out of `Notifications` correctly.

I kept the original tests in because if the `toArray` method in `Action` changes, this would trigger test failures.


danharrin:
Thanks

PR #5870, opened 2023-03-07
feat-v3/maxContentWidth

wychoong:
this PR to bring back maxContentWidth from v2

supported 
- context
- per page


packages/app/src/Pages/BasePage.php
```diff
@@ -52,6 +57,11 @@ public function getTitle(): string
                 ->title();
     }
 
+    protected function getMaxContentWidth(): ?string
```
danharrin:
```suggestion
    public function getMaxContentWidth(): ?string
```

packages/app/src/FilamentManager.php
```diff
@@ -264,6 +264,11 @@ public function getLogoutUrl(): string
         return $this->getCurrentContext()->getLogoutUrl();
     }
 
+     public function getMaxContentWidth(): ?string
+     {
+         return $this->getCurrentContext()->getMaxContentWidth();
+     }
+
```
danharrin:
```suggestion
    public function getMaxContentWidth(): ?string
    {
        return $this->getCurrentContext()->getMaxContentWidth();
    }

```

danharrin (changes requested):

PR #5875, opened 2023-03-07
Fix for breaking change to createOptionAction on Select field

cheesegrits:
Fixes an issue with a recent change which breaks createOptionAction() on Select fields if the Select is not based on a relationship


packages/forms/src/Components/Select.php
```diff
@@ -213,7 +213,9 @@ public function getCreateOptionAction(): ?Action
         }
 
         $action = Action::make($this->getCreateOptionActionName())
-            ->form(fn (Form $form) => $form->model($this->getRelationship()?->getModel()::class)->schema($actionFormSchema))
+            ->form(fn (Form $form) => $form->model(
+                $this->getRelationship() ? $this->getRelationship()->getModel()::class : null
```
```suggestion
                $this->getRelationship() ? $this->getRelationship()->getModel()::class : null,
```

PR #5880, opened 2023-03-08
fix: table loses its default sort and direction.

diegobas:
The default sorting field and direction are lost when a user clicks on the header several times removing the column sorting. This fix solves the issue.

### DIFF
--- a/packages/tables/src/Concerns/CanSortRecords.php
+++ b/packages/tables/src/Concerns/CanSortRecords.php
@@ -22,8 +22,8 @@ trait CanSortRecords
             $direction ??= 'asc';
         }
 
-        $this->tableSortColumn = $direction ? $column : null;
-        $this->tableSortDirection = $direction;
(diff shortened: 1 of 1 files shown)


diegobas:
I closed it by mistake, sorry.

danharrin:
Yes this is deliberate so there is still an option to remove the default sort if you want to. It's a breaking change if we modify this behaviour. You can still override the `sortTable()` method on your Livewire component / admin panel pages if you want :)

diegobas:
Ok, I understand your point of view. But if my default sorting column is not defined in the table columns array and I wanted to sort all my tables by that column, what can I do?

For example, I want all my tables sorted by desc ID when no column is the sorting one in a table. Is there any way to do it without overriding tableSort in every page?

danharrin:
You could extract it into a trait and then you don't have to copy the code everywhere. In Filament v3 the behaviour is fixed I think.

diegobas:
Ok. Thank you!

PR #5887, opened 2023-03-09
Make TenantSet event properties accessible

josefbehr:
Properties $tenant and $user are currently declared protected in the TenantSet event which prevents them from being accessed in listeners. This PR sets them as public (and readonly). I hope I'm not missing some reason for making them protected. If that's the case, let me know and delete this PR.

### DIFF
--- a/packages/app/src/Events/TenantSet.php
+++ b/packages/app/src/Events/TenantSet.php
@@ -17,18 +17,12 @@ class TenantSet
     ) {
     }
 
-    /**
-     * @return Model
-     */
     public function getTenant(): Model
     {
         return $this->tenant;
     }
 
-    /**
-     * @return HasTenants | Authenticatable | Model
-     */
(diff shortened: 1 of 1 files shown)


packages/app/src/Events/TenantSet.php
```diff
@@ -12,8 +12,8 @@ class TenantSet
     use Dispatchable;
 
     public function __construct(
-        protected Model $tenant,
-        protected Model | Authenticatable | HasTenants $user,
+        public readonly Model $tenant,
+        public readonly Model | Authenticatable | HasTenants $user,
```
zepfietje:
```suggestion
        public Model $tenant,
        public Model | Authenticatable | HasTenants $user,
```

zepfietje:
We can't make them `readonly` as that feature got introduced in PHP 8.1, while Filament requires minimum 8.0.

danharrin:
I think these should remain protected and we just add accessors to the class

PR #5888, opened 2023-03-09
Fixes alpine error with deferred loading and selection indicator

archilex:
When using the new deferLoading() method when tables are initially loaded `$allRecordsCount` is null which causes an alpine error in the `selection-indicator` component. This PR fixes the bug by only loading the `selection-indicator` after the data is loaded.

If you prefer you could also use `&& $allRecordsCount` instead of `&& $isLoaded`.


danharrin:
Thanks

PR #5896, opened 2023-03-09
add createOptionFormLabelUsing

vincent-tarrit:
This PR add the method createOptionFormLabelUsing to personnalize the heading in the Select.createOptionForm.

Thank you for your amazing work !

### DIFF
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -519,10 +519,10 @@ class Select extends Field implements Contracts\HasNestedRecursiveValidationRule
                 $relatedModels = $relationship->getResults();
 
                 $component->state(
-                // Cast the related keys to a string, otherwise JavaScript does not
-                // know how to handle deselection.
-                //
-                // https://github.com/filamentphp/filament/issues/1111
+                    // Cast the related keys to a string, otherwise JavaScript does not
+                    // know how to handle deselection.
+                    //
+                    // https://github.com/filamentphp/filament/issues/1111
                     $relatedModels
                         ->pluck($relationship->getRelatedKeyName())
                         ->map(static fn ($key): string => strval($key))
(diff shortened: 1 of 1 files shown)


packages/forms/src/Components/Select.php
```diff
@@ -505,10 +519,10 @@ public function relationship(string | Closure $relationshipName, string | Closur
...
 
                 $component->state(
-                    // Cast the related keys to a string, otherwise JavaScript does not
-                    // know how to handle deselection.
-                    //
-                    // https://github.com/filamentphp/filament/issues/1111
+                // Cast the related keys to a string, otherwise JavaScript does not
+                // know how to handle deselection.
+                //
+                // https://github.com/filamentphp/filament/issues/1111
```
zepfietje: Could you revert these changes?

vincent-tarrit: yes it's done sorry !

packages/forms/src/Components/Select.php
```diff
@@ -42,6 +42,8 @@ class Select extends Field implements Contracts\HasNestedRecursiveValidationRule
 
     protected ?Closure $createOptionUsing = null;
 
+    protected string | Closure | null $createOptionFormLabelUsing = null;
```
zepfietje: Shouldn't this be something like `createOptionModalHeading`?

vincent-tarrit: yes it could be, I edited it

zepfietje: The `Using` suffix isn't required, right? Since you could also just pass a string.

vincent-tarrit: That's right, I tried to copy how was structured the other field

packages/forms/src/Components/Select.php
```diff
@@ -234,6 +236,18 @@ public function getCreateOptionActionFormSchema(): ?array
         return $this->evaluate($this->createOptionActionFormSchema);
     }
 
+    public function createOptionModalHeading(string | Closure | null $callback): static
+    {
+        $this->createOptionModalHeading = $callback;
+
+        return $this;
+    }
```
```suggestion
    public function createOptionModalHeading(string | Closure | null $heading): static
    {
        $this->createOptionModalHeading = $heading;

        return $this;
    }
```

zepfietje:
Thanks, @vincent-tarrit!

PR #5897, opened 2023-03-10
Add support for relationships to Infolists

cheesegrits:
Adds support for arbitrary depth relationships to Infolists, so they work like table columns, like ...

```
                    TextEntry::make('flightLegs.dailyLog.aircraft.n_number')
                        ->label('Aircraft Flown')
                        ->distinctList()
                        ->listWithLineBreaks(),
```

### DIFF
--- a/packages/infolists/src/Components/Concerns/CanGetStateFromRelationships.php
+++ b/packages/infolists/src/Components/Concerns/CanGetStateFromRelationships.php
@@ -6,7 +6,7 @@ use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\Relation;
 use Illuminate\Support\Collection;
 
-trait InteractsWithTableQuery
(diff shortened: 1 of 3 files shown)


cheesegrits:
Occurred to me that you may not want to add that InteractsWithTableQuery dependency on the tables package, in which case I could extract what is needed from that and add it as a Concern in the Infolists package.

danharrin:
Yeah we can't leak code between packages, so its best if you copy the concern and only leave the methods that you actually use

danharrin:
Also PHPStan needs checking

cheesegrits:
OK, I think that fixes everything.  Extracted just what we need to a new InteractsWithTableQuery trait in the Infolists package, which also makes PHPStan happy now there is no longer unimplemented functions from the use of the table package one.

danharrin:
Thanks

PR #5898, opened 2023-03-10
Allow optional outlined create item buttons on builder and repeater

johncarter-:
Allow avoidance of this primary button overload:

image

To this:

image

### DIFF
--- a/packages/forms/resources/views/components/builder.blade.php
+++ b/packages/forms/resources/views/components/builder.blade.php
@@ -352,7 +352,10 @@
                 class="flex justify-center"
             >
                 <x-slot name="trigger">
-                    <x-forms::button size="sm" :outlined="$isCreateItemButtonOutlined()">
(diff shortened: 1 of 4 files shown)


danharrin:
I'm happy to have this as default instead of as a config option I think!

johncarter-:
Okay, I'll just set the button to outlined without the config methods. Sort this out soon.

danharrin:
Thanks!

margarizaldi:
Hi @johncarter- Sorry for this late question... Why just not letting the method to be there with the default 'outlined'?
Plus with even more minimalist button:

image

danharrin:
In v3 this is already built in. Not planning to add any more configuration methods atm that we will just end up deprecating.

margarizaldi:
> In v3 this is already built in. Not planning to add any more configuration methods atm that we will just end up deprecating.

Absolutely

PR #5900, opened 2023-03-10
add Select::getLabel() return types

braceyourself:
When passing an HtmlString to the select component label() method, the output is raw html. We need to add the Htmlable Contract return type to ensure the value is not cast to a string. Also added ' | null ' to match parent getLabel() method.


danharrin:
Thanks!

PR #5912, opened 2023-03-12
Add Octane documentation

haringsrob:
While working on a project I figured out that hooks etc are not working.

I solved it by registering a listener and decided to contribute my learnings.

### DIFF
--- a/packages/admin/docs/08-appearance.md
+++ b/packages/admin/docs/08-appearance.md
@@ -320,3 +320,49 @@ The available hooks are as follows:
 - `page.footer-widgets.end` - after page footer widgets
 - `page.actions.start` - before page actions
 - `page.actions.end` - after page actions
+
+## Laravel Octane
+
+When using Laravel octane, the filament serving and hooks will only fire on the first request.
+
+To work around this you can create an event listener to run the hooks in Octane:
+
+```php
+<?php
+
+namespace App\Listeners;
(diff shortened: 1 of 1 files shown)


danharrin:
I'm confused, because the Serving event is being dispatched via middleware. Why would Octane not run middleware on multiple requests? 🤔

haringsrob:
I believe it is because the service is scoped? 

```
        $this->app->scoped('filament', function (): FilamentManager {
            return new FilamentManager();
        });
```

There is no static caching inside that service, and octane does not reboot service providers so every request there is a new instance of FilamentManager, but it will be blank.

haringsrob:
Did a quick test and changed scoped to singleton and all works fine (so no workaround needed).

I do not know what the impact of it is.

danharrin:
Hm well I think it's probably best that we don't change it because it's probably got unintended consequences

haringsrob:
- [ ] Update the docs to use a middleware instead of providers to ensure a consistent outcome instead of a octane specific approach.

zepfietje:
Mark as ready for review when done, @haringsrob!

zepfietje:
@haringsrob, are you still able to finish this PR? 😄

zepfietje:
Closing due to inactivity.
Feel free to open a complete PR.

PR #5915, opened 2023-03-13
fix-v3/reduce nav jumping due to x-cloak

wychoong:


packages/app/resources/views/components/layouts/app/sidebar/group.blade.php
```diff
@@ -44,12 +44,11 @@ class="ml-1 shrink-0"
...
             @if ($collapsible)
                 <x-filament::icon
-                    name="heroicon-m-chevron-down"
+                    name="heroicon-m-chevron-up"
                     alias="app::sidebar.group.collapse"
                     size="h-5 w-5"
                     class="text-gray-600 transition dark:text-gray-300"
-                    x-bind:class="$store.sidebar.groupIsCollapsed(label) || '-rotate-180'"
-                    x-cloak="x-cloak"
+                    x-bind:class="!$store.sidebar.groupIsCollapsed(label) || 'rotate-180'"
```
wychoong, line 51: another alternative would be using class bind to toggle `invisible`, but this approach will have the all menu icons to be invisible on load while the suggested only have the collapsed menu icon to have rotate animation. 

```js
name="heroicon-m-chevron-up"
                    class="text-gray-600 transition dark:text-gray-300 invisible"
                    x-bind:class="{
                        'invisible': false,
                        'rotate-180': $store.sidebar.groupIsCollapsed(label),
                    }"

```

zepfietje: Haven't tested, but I think the current approach is ok.
Have recently done the same for table grouping (basically exact same bug).

danharrin: Are we ok to merge then?

zepfietje: Think so.

PR #5917, opened 2023-03-13
Fix bug when saving a reorderable field w/ custom table name

EldenChris:
When saving with a reorderable media library file field AND a custom table name for the Media class, it crashes because the fixed code does not take into account the custom media model from the config.


danharrin:
Thanks

PR #5920, opened 2023-03-13
add method to add unauthenticated tenant routes to context

josefbehr:
With tenancy enabled, the routes that are added to a context using ->routes() don't work (and probably shouldn't), but there might be cases where one wants to add custom unauthenticated routes to a tenant routes (e.g. a function to invite new users to a tenant). This PR adds that ability.


danharrin:
Thanks

PR #5925, opened 2023-03-13
Qualify the field name for sorting.

cheesegrits:
Protects against "ambiguous" errors when more than one table in the relationship chain has the same field name.

Not sure if my change is the most efficient way of doing this, but it's the only way I could figure out to do it.  Happy to change the code if anyone has a better way.

### DIFF
--- a/packages/tables/src/Grouping/Group.php
+++ b/packages/tables/src/Grouping/Group.php
@@ -224,7 +224,7 @@ class Group
             [];
 
         if (! count($relationships)) {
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks! Could you fix up PHPStan for me please? You probably want to use specific relationship types

cheesegrits:
@danharrin  Stan is happy now.

Sometimes Laravel's magic is a little too insidious.  There's a magic __call() method on the Relation class which forwards any unhandled methods on to the query, so I've gotten used to using things like $relation->getModel() when I really mean $relation->getQuery()->getModel().

PR #5927, opened 2023-03-13
Support subdomain tenancy

josefbehr:


danharrin:
Can you remind me why the route parameter needs to be configurable? want to make sure it doesn't introduce more complexity

josefbehr:
> Can you remind me why the route parameter needs to be configurable? want to make sure it doesn't introduce more complexity

It doesn't need to be. Just thought it was a nice addition while doing this. Maybe I was a bit over eager there. Can revert that, if you like.

danharrin:
Yes please, it does create some extra complexity that we can hold off on until it comes in useful

josefbehr:
> Yes please, it does create some extra complexity that we can hold off on until it comes in useful

Of course. It's been reverted, and I also extended the tenancy documentation for subdomain configuration

zepfietje:
@josefbehr, is this ready for review?
If so, please mark it as ready. 😊

danharrin:
Thanks, but we are actually now looking to implement a package such as `stancl/tenancy` or Spatie's tenancy to handle multi-tenancy. Would really rather not have to copy what they've done in those packages to handle paths/domains/multi-databases.

Thank you again for this contribution though

josefbehr:
I think that's a wise choice :) no hard feelings, thanks for the info

jeremiahsherrill:
Has there been any progress on implementing stancl or spatie ?

danharrin:
We don't need to implement anything, you can do it all on your end

jeremiahsherrill:
I thought with the current tenancy system you had in place that didn't support domain based logic that you were looking into these things.   nvm

PR #5930, opened 2023-03-14
Use attributesToArray() in getStateFromRecord()

cheesegrits:
Using getAttributes() doesn't pick up computed properties / accessors on the model record, so changed it to use attributesToArray().


cheesegrits:
What-the-diff appears to have lost its mind, I have no idea what that stuff about the missing semi-colon is all about.

PR #5938, opened 2023-03-15
Fix icon column alignment

### LINKED ISSUE #5937: IconColumn doesn't respect alignment
### Package

filament/filament

### Package Version

v2.17.15

### Laravel Version

v10.3.3

### Livewire Version

v2

### PHP Version

PHP 8.2

### Problem description

Using an IconColumn with ->alignRight() does align the heading to the right, but not the icon itself.
I guess this is due to the alignment class being text-right, which does not apply to the svg inside the cell.
Applying the classes 'flex' and 'justify-right' to the column results in the correct behaviour.

image

### Expected behavior

Heading of the column as well as content are aligned as specified.

### Steps to reproduce

Create a table with an icon column and apply alignRight() to the column.

### Reproduction repository

https://github.com/josefbehr/filament-v2-icon-column-alignment

### Relevant log output

_No response_

josefbehr:
Proposed fix for #5937

### DIFF
--- a/packages/tables/resources/views/components/columns/column.blade.php
+++ b/packages/tables/resources/views/components/columns/column.blade.php
@@ -76,7 +76,7 @@
     @else
         <div @class([
             'flex',
-            $alignmentClass
+            $alignmentClass,
         ])>
             {{ $slot }}
         </div>
--- a/packages/tables/resources/views/components/columns/column.blade.php
+++ b/packages/tables/resources/views/components/columns/column.blade.php
@@ -16,12 +16,12 @@
     $url = $column->getUrl();
 
     $alignmentClass = match ($alignment) {
-        'center' => 'text-center',
-        'end' => 'text-end',
-        'left' => 'text-left',
-        'right' => 'text-right',
-        'justify' => 'text-justify',
-        default => 'text-start',
+        'center' => 'text-center flex justify-center',
+        'end' => 'text-end flex justify-end',
+        'left' => 'text-left flex justify-start',
+        'right' => 'text-right flex justify-end',
+        'justify' => 'text-justify flex justify-between',
+        default => 'text-start flex justify-start',
     };
 
     $slot = $column->viewData(['recordKey' => $recordKey]);
--- a/packages/tables/resources/views/components/columns/column.blade.php
+++ b/packages/tables/resources/views/components/columns/column.blade.php
@@ -74,7 +74,10 @@
             {{ $slot }}
         </button>
     @else
-        <div @class([$alignmentClass])>
+        <div @class([
+            'flex',
+            $alignmentClass]
+        )>
             {{ $slot }}
         </div>
     @endif
--- a/packages/tables/resources/views/components/columns/column.blade.php
+++ b/packages/tables/resources/views/components/columns/column.blade.php
@@ -67,7 +67,7 @@
             wire:loading.class="cursor-wait opacity-70"
             type="button"
             @class([
-                'block w-full flex',
+                'w-full flex',
                 $alignmentClass,
             ])
         >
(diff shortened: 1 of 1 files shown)


packages/tables/resources/views/components/columns/column.blade.php
```diff
@@ -16,12 +16,12 @@
...
-        'left' => 'text-left',
-        'right' => 'text-right',
-        'justify' => 'text-justify',
-        default => 'text-start',
+        'center' => 'text-center flex justify-center',
+        'end' => 'text-end flex justify-end',
+        'left' => 'text-left flex justify-start',
+        'right' => 'text-right flex justify-end',
+        'justify' => 'text-justify flex justify-between',
+        default => 'text-start flex justify-start',
```
zepfietje: Could you move `flex` outside the match expression?

danharrin: Also maybe remove the text alignment classes if they dont work

danharrin: Or maybe the text alignment classes need to stay to ensure that the text column still works?

josefbehr: @danharrin I just tried it out. The text alignment classes are not needed, it works with just flex and justify classes. But: alignJustify does not work either way as far as I can see, since the text itself is wrapped again in a different div that does not extend to the whole column width. Should I try to fix that as well?

danharrin: Does flex still work if the text column has `wrap()`

josefbehr: > Does flex still work if the text column has `wrap()`

looks like it, yes:
image
image

danharrin: When the text is on multiple lines though

josefbehr: > When the text is on multiple lines though

Yeah, no. Good catch. I think it's best to leave the text alignment classes in.

packages/tables/resources/views/components/columns/column.blade.php
zepfietje, line 1: Doesn't the div at the end of the file also need the `flex` class?
Probably best to extract it.

zepfietje (changes requested):

packages/tables/resources/views/components/columns/column.blade.php
josefbehr: Ah yes, you're right, didn't see that one.
As for extraction, I'm not sure we gain anything by just putting `flex` into a variable, since the first two cases have additional classes `block w-full`, but that last one doesn't.

zepfietje: The `block` class can't be used in combination with `flex` as they both control the `display` property. I think `block` needs to be removed?

packages/tables/resources/views/components/columns/column.blade.php
```diff
@@ -39,7 +39,7 @@
             href="{{ $url ?: $recordUrl }}"
             {!! $shouldOpenUrlInNewTab ? 'target="_blank"' : null !!}
             @class([
-                'block w-full',
+                'block w-full flex',
```
zepfietje:
```suggestion
                'w-full flex',
```

packages/tables/resources/views/components/columns/column.blade.php
```diff
@@ -67,14 +67,17 @@
             wire:loading.class="cursor-wait opacity-70"
             type="button"
             @class([
-                'block w-full',
+                'block w-full flex',
```
zepfietje:
```suggestion
                'w-full flex',
```

zepfietje:
I've cleaned up.
Could you test if everything still works as expected, @josefbehr?

josefbehr:
> I've cleaned up. Could you test if everything still works as expected, @josefbehr?

Couldn't see any problem, seems to work fine.

zepfietje (approved):

PR #5939, opened 2023-03-15
Add icon to section class and template

josefbehr:
Add ability to add an icon to the section header using the `icon()` method as with e.g. tabs.
Looks like this (compact and non-compact):

image

I thought about adding an `iconSize()` method as well but decided against it and chose default size values that I think work well.

While at it, I also removed the superfluous `font-bold` from the heading classes.


danharrin:
Perfect, thanks

PR #5954, opened 2023-03-17
Feature/persist sort in session

eazyurk:
This PR adds persisting the sort in a user session. 

I also have a question regarding this PR, how should we handle `defaultTableSortColumn` and `defaultTableSortDirection`?

Right now they get set in the `mountInteractsWithTable` so they overwrite the `sortColumn` and `sortDirection` from the session.
I was thinking to move this:
```
        $this->tableSortColumn ??= $this->getDefaultTableSortColumn();
        $this->tableSortDirection ??= $this->getDefaultTableSortDirection();
```

Out of the `mountInteractsWithTable()` and put it in the `bootedInteractsWithTable()` above the part where the sort gets set by the session so the session can overwrite the default `sortColumn` and `sortDirection`.

### DIFF
--- a/packages/tables/docs/03-columns/01-getting-started.md
+++ b/packages/tables/docs/03-columns/01-getting-started.md
@@ -123,6 +123,17 @@ protected function getDefaultTableSortDirection(): ?string
 }
 ```
 
+### Persist search in session
+
+To persist the sort in the user's session, override the `shouldPersistTableSortInSession()` method on the Livewire component:
+
+```php
+protected function shouldPersistTableSortInSession(): bool
+{
+    return true;
+}
+```
+
 ## Searching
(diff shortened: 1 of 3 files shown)


danharrin:
Thanks

packages/tables/docs/03-columns/01-getting-started.md
```diff
@@ -123,6 +123,17 @@ protected function getDefaultTableSortDirection(): ?string
 }
 ```
 
+### Persist search in session
```
zepfietje, line 126: @danharrin, this heading is wrong.

danharrin: will fix

PR #5957, opened 2023-03-17
feature: allow form component actions to have arguments on mount

ryangjchandler:
Allows calling `$wire.mountFormComponentAction('state path', 'action', { arguments })` and retrieving via `$arguments` in the callback.

### DIFF
--- a/packages/forms/src/Concerns/HasFormComponentActions.php
+++ b/packages/forms/src/Concerns/HasFormComponentActions.php
@@ -16,6 +16,8 @@ trait HasFormComponentActions
 {
     public $mountedFormComponentAction = null;
 
+    public $mountedFormComponentActionArguments = [];
+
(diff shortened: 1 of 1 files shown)


packages/forms/src/Concerns/HasFormComponentActions.php
```diff
@@ -56,7 +58,10 @@ public function callMountedFormComponentAction(?string $arguments = null)
             return;
         }
 
-        $action->arguments($arguments ? json_decode($arguments, associative: true) : []);
+        $action->arguments(array_merge(
+            $this->mountedFormComponentActionArguments ?? [],
+            $arguments ? json_decode($arguments, associative: true) : []
```
danharrin:
```suggestion
            $arguments ? json_decode($arguments, associative: true) : [],
```

PR #5964, opened 2023-03-18
fix missing min and max values, if field is initially hidden

josefbehr:
Due to the wrapping in `! $isConcealed()` the values are not applied if a field is initially hidden via `visible()` or `hidden()` and won't get applied if the field becomes visible. I removed the conditional block and could not see any problems in my local testing.
If there's another solution, or this one does create problems, I'm happy to research further.


danharrin:
The reason why it's this way is because fields that are not visible in the browser are not focusable, and when the browser runs frontend validation, it results in console errors. Fields inside tabs and collapsible sections are also marked as "concealed".

In this case, I think that fields should only be concealed if `isHidden() === true` currently

josefbehr:
Then I guess the problem is that the values are not added as soon as the field is visible again. For me I noticed this with a hidden fieldset, which gets shown after a toggle is switched. If the fields are visible from the start, the validation works and I can't input values below minValue, but it doesn't if they start out hidden.
How do you think this can be solved?

danharrin:
Not sure, it sounds like a bug if the attributes aren't being added back when the field is visible

josefbehr:
@danharrin probably, also applies to sections apparently. If a section is collapsed, the attributes are also not applied. Any idea where the problem might lie?

danharrin:
Oh right, that's deliberate though. If a section is collapsed the browser can't autofocus it so there's a console error.

josefbehr:
@danharrin ah yeah, sorry, I get that, but when it's un-collapsed, the attributes still do not apply, meaning that e.g. min Value does not work as expected just as in initially-hidden fields.

danharrin:
Yeah well I'm not sure the best way of conditionally adding them... so that's why they're not there

josefbehr:
I see. I'd still consider this a bug then. Just dug into the code a bit, and I see how it comes about. `isConcealed()` evaluates to true as soon as the component is child of a component that can conceal its children, like section and repeater, no matter if they are currently concealed (hidden, collapsed, ...) or not.

Shouldn't it be possible to extend this to determine the current state of the parent component (or rather let the parent determine its `concealing` state), and when re-rendering the fields the attributes get added, because then `isConcealed()` evaluate to false?

danharrin:
I dont think so, because the state of a section / tabs is in JavaScript, not PHP

zepfietje:
What's the status of this PR?

danharrin:
This isn't the right solution, so if another one is found a new PR can be created

PR #5965, opened 2023-03-18
make required-asterisk stick to label

josefbehr:
Problem: When using inline-labels, it might occur that the asterisk denoting required fields wraps to a newline after the label. This does't look great:
image
Screenshot_20230318_022454

Putting `whitespace-nowrap` into the label classes is not a solutin imo as it leads to long labels not wrapping at all:
Screenshot_20230318_022555
Screenshot_20230318_022618

So I wrapped the `sup` tag with the asterisk inside a `span` with `whitespace-nowrap` and moved it directly behind the label text. It doesn't look pretty in the blade file, but it works and results in the asterisk sticking to the last label word without wrapping into its own line:
Screenshot_20230318_022740

Aside:
long words or too much label text are still not great, but I don't think there's much we can do
image
Screenshot_20230318_022358

### DIFF
--- a/packages/forms/resources/views/components/field-wrapper/label.blade.php
+++ b/packages/forms/resources/views/components/field-wrapper/label.blade.php
@@ -15,13 +15,12 @@
         'text-danger-700' => $error,
         'dark:text-danger-400' => $error && config('forms.dark_mode'),
     ])>
-        {{ $slot }}
-
-        @if ($required)
-            <sup @class([
-                'font-medium text-danger-700',
-                'dark:text-danger-400' => config('forms.dark_mode'),
-            ])>*</sup>
+        {{ $slot }}@if ($required)<span class="whitespace-nowrap">
+                <sup @class([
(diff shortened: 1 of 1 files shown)


danharrin:
I think this is a problem with the way you're using inline labels though. That design is only meant to be used when there is enough horizontal space for the labels to fill without needing to wrap. In your screenshot, it looks like there is very little space. On mobile, we already fall back to the normal style of label. What do you think about not merging this and rethinking your own app's form layout?

josefbehr:
For my app, I already changed the layout, but I still think having the asterisk stick to the last label word is a good idea as even with enough space (which in itself is hard enough given the many resolutions, browsers, and ways to use them) it might get wrapped on its own. But I have no ego in this, so if you don't think this is a worthwhile change, just close the PR and leave it as it is.

danharrin:
My only concern was that it might be easily reverted if someone fixes the formatting, so I added a comment

PR #5969, opened 2023-03-18
Fix sticky form actions with custom footer

joaovdiasb:
Improves the timing of the fixed footer appearing when you have a custom footer


danharrin:
Hmmm... not sure about this. Shouldn't we calculate how far the buttons are from the bottom of the page and use that value? Otherwise it won't respect relation managers or anything.

joaovdiasb:
> Hmmm... not sure about this. Shouldn't we calculate how far the buttons are from the bottom of the page and use that value? Otherwise it won't respect relation managers or anything.

Of course, I forgot about this scenario, I needed to create a reference element to check offsetTop because of sticky

danharrin:
Please use Alpine's `x-ref` feature instead of an ID :)

danharrin:
You could even get away with `$el` probably, then you don't need an extra element?

joaovdiasb:
> You could even get away with `$el` probably, then you don't need an extra element?

it's because with sticky the offset changes and the element keeps flashing when scrolling

zepfietje:
Are you able to finish this PR, @joaovdiasb? 🙂

joaovdiasb:
> Are you able to finish this PR, @joaovdiasb? 🙂

it is not possible to resolve referencing the buttons itself with `$el`, because it is applied `isSticky` class losing offsetTop, it works but causes a visual bug where the element keeps blinking when scrolling, any other ideas?

zepfietje:
Hmm not sure, but thanks for clarifying.
Could you explain why exactly the current approach is broken for custom footers?

joaovdiasb:
Current:
CleanShot 2023-04-10 at 15 56 04

New
CleanShot 2023-04-10 at 16 02 23

it's a better sticky approach only when needed, which is when the current view is above the form's buttons

zepfietje:
Hmm and how is this caused by a custom footer specifically?

zepfietje:
> Hmm and how is this caused by a custom footer specifically?

Could you clarify, @joaovdiasb? 🙂

danharrin:
I've reverted #5947 to fix this issue, that caused it for me. I don't know exactly what issue @oddvalue was experiencing as there was no issue report

PR #5977, opened 2023-03-20
fix/hide extra element in table when empty

wychoong:
currently when using column layout but without sortable columns, there is an extra element with `border-b`

image

**before**
image

**after**
image

example
```php
Tables\Columns\Layout\Stack::make([
])
```

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -312,9 +312,17 @@
         >
             @if (($content || $hasColumnsLayout) && ($records !== null) && count($records))
                 @if (($content || $hasColumnsLayout) && (! $isReordering))
+                    @php
+                        $sortableColumns = array_filter(
+                            $columns,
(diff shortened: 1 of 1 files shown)


packages/tables/resources/views/index.blade.php
```diff
@@ -312,9 +312,17 @@ class="shrink-0"
...
+                        $sortableColumns = array_filter(
+                            $columns,
+                            fn (\Filament\Tables\Columns\Column $column): bool => $column->isSortable(),
+                        );
+                    @endphp
+                    
                     <div @class([
                         'bg-gray-500/5 flex items-center gap-4 px-4 border-b',
                         'dark:border-gray-700' => config('tables.dark_mode'),
+                        'hidden' => !$isSelectionEnabled && !count($sortableColumns),
```
```suggestion
                        'hidden' => (! $isSelectionEnabled) && (! count($sortableColumns)),
```

PR #5981, opened 2023-03-20
fix: s3 bucket where file missing

phpsa:
if an image is missing from s3 bucket, the storage->exists() throws a `League\Flysystem\UnableToCheckFileExistence` exception, which prevents fixing the uploaded file etc.

this patch will allow that to return false/null where needed


danharrin:
Thanks

PR #5984, opened 2023-03-21
Adds key bindings to global search

archilex:
This PR adds a key_bindings configuration so that global search can be called from anywhere in the app.


danharrin:
I don't want to add anything else to the config in v2, since the config is gone in v3 anyway. I am going to add this feature to v3. Thank you very much :)

archilex:
Sounds good! 👍 Do you have any guidelines for what we should and shouldn’t PR for v2? Do you want help with PRs for V3? (Maybe PRing to v3 might be too much added work for you since it’s still in progress?) Thanks!

danharrin:
No more config options in v2 if possible 👍 We'd prefer no PRs to v3 right now as it's WIP, no-one should really be using it anyway.

archilex:
Sounds good! Loving working with Filament. Thanks! 👍👍

danharrin:
You're welcome! Glad you like it

PR #5985, opened 2023-03-21
Fix Placeholder translateLabel

ganyicz:
Unifies the Placeholder behavior with other fields - when fallback label is generated from field name, it is then translated when shouldTranslateLabel is enabled. Prior to this change, this didn't happen on Placeholder field.

In order to avoid duplication I abstracted this method into it's own trait.
Since the Placeholder doesn't extend Field, we cannot just inherit the behavior.

There was a slight difference in the getLabel implementation (lack of afterLast dot call) but I didn't find any evidence of this change being intentional. When this field was first added (before translateLabel existed) the implementation was identitcal: https://github.com/filamentphp/filament/blob/1134a770029d8052d012766a3607f5bc3abbaa03/packages/forms/src/Components/Placeholder.php

That one line then got lost between these two commits:

Screenshot 2023-03-21 at 07 33 38

Let me know what do you think


ganyicz:
I just noticed that the use of trait is not in line with other ones as it doesn't have it's own properties and relies on it being used on a class with getLabel() method - not sure about it

danharrin:
I've refactored it into the existing HasName trait instead of a new trait

danharrin:
Thanks

PR #5986, opened 2023-03-21
Refactor `Filament\locale_has_pluralization` to `Filament\Support\locale_has_pluralization`

iRaziul:
I used `filament/tables` separately and faced an issue while trying to use `DeleteBulkAction`. 😟

Here is the error message:

image

I've used the admin panel and there were no such issues.
I found that the function was inside `filament/filament` package and to fix the function was needed to move under `Filament\Support` namespace.

What I did?
- Moved the function from `Filament\locale_has_pluralization` to `Filament\Support\locale_has_pluralization`
- Update all the references

Everything is working fine now ✅


danharrin:
Thanks

PR #5990, opened 2023-03-21
Add 'empty.buttons.reset_column_searches.label' table.php

dennisvandalen:
Adds translation for empty column search label. When searching in a specific column this label is shown, this seemed like the best Dutch translation


danharrin:
Thanks

PR #5992, opened 2023-03-21
Allow hasManyDeep relationships in TagsColumn

archilex:
The TagsColumn currently doesn't work with HasManyDeep relationships. Filament already includes the `staudenmeir/belongsToThrough` package so I thought I try this PR to include the even more popular `hasManyDeep` package to allow these relationships to work with Tags. 

Tests are passing, but maybe there's another gotcha with implementing this?


danharrin:
Thanks

archilex:
Found the gotcha! Submitted a new PR to fix #6011

PR #5994, opened 2023-03-22
Notification - center notifications vertically

leandrocfe:
Hi! This pull request includes a new feature that enables users of Filament to easily center notifications vertically. Previously, this functionality was not available, but now with the addition of 'vertical_alignment' => 'center' in the filament config file, users can easily set their notifications to appear in the center of their screen.


packages/notifications/resources/views/notification.blade.php
```diff
@@ -13,6 +13,7 @@
             'center' => match (config('notifications.layout.alignment.vertical')) {
                 'top' => '-translate-y-12',
                 'bottom' => 'translate-y-12',
+                'center' => '',
```
danharrin:
```suggestion
                'center' => null,
```

PR #5998, opened 2023-03-22
Add documentation for the max value rule

asilvabe:


danharrin:
Thanks for the PR, but I'm not gonna merge it. `maxValue()` is already documented in the "Text input" section on the "Fields" page. This is because `maxValue()` is only available on the TextInput, not any other field. So we shouldn't add it to the main validation section as it doesn't work for all fields.

asilvabe:
Hello @danharrin !

Thanks for your answer. I understand that `maxValue()` is already documented in the appropriate section on the "Fields" page. I appreciate your explanation that it is only available on "TextInput" and not on all fields. Thanks again for taking the time to review my pull request.

Have a good day!

PR #6005, opened 2023-03-23
Table action hide disabled label

howdu:
This is the easiest change to allow for edit table action to disable label.
```php
\Filament\Tables\Actions\EditAction::make()
    ->disableLabel()
```

Not sure if it also needs to be added to the other table actions.


danharrin:
I think youre looking for `->iconButton()`

PR #6006, opened 2023-03-23
Add register listener

howdu:
Allow for singular listeners to be added and ability to overwrite default listeners

```php
Repeater::make()->registerListener('repeater::createItem', fn () => )
```


danharrin:
Thanks, but I don't see the value in this when array syntax also works. Also, I think this has a bug where it overwrites any existing listeners for the event.

In v3 the event system is being replaced and you wont need an array for this anyway.

howdu:
No worries I can do this in my extended class, but  that's exactly what I need to do is overwrite a default listener.

PR #6010, opened 2023-03-23
Call configure on ActionGroup after making it.

megawubs:
The `ActionGroup::make()` method was missing the call to `->configure()` inside the make method. This PR adds it.


danharrin:
Thanks

PR #6015, opened 2023-03-23
Update 08-appearance.md

bennettblack:
`container` works too!

### DIFF
--- a/packages/admin/docs/08-appearance.md
+++ b/packages/admin/docs/08-appearance.md
@@ -193,7 +193,7 @@ Filament exposes a configuration option that allows you to change the maximum co
 
 You must [publish the configuration](installation#publishing-configuration) in order to access this feature.
 
-In `config/filament.php`, set the `layouts.max_content_width` to any value between `xl` and `7xl`, or `full` for no max width:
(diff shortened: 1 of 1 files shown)


danharrin:
Does it?! Where do we do that in the view

bennettblack:
In `app.blade.php` it would _seem_ that passing `'container'` would result in a null return from the match statement...
```blade
@props([
    'maxContentWidth' => null,
])

{{-- ... --}}

<div @class([
    'filament-main-content flex-1 w-full px-4 mx-auto md:px-6 lg:px-8',
     match ($maxContentWidth ??= config('filament.layout.max_content_width')) {
          null, '7xl', '' => 'max-w-7xl',
          'xl' => 'max-w-xl',
          '2xl' => 'max-w-2xl',
          '3xl' => 'max-w-3xl',
          '4xl' => 'max-w-4xl',
          '5xl' => 'max-w-5xl',
          '6xl' => 'max-w-6xl',
          'full' => 'max-w-full',
          default => $maxContentWidth,
          },
     ])>
```

However.... changing the `max_content_width` config value to 'container' does reflect it in the rendered class list.

image

danharrin:
It's because of the `default` arm of the match statement. `container` is a custom class like any other, so we don't really need to document it as part of Filament. Thanks anyway

PR #6022, opened 2023-03-24
Fixes error display of nestedRecurrsiveValidationRules

archilex:
Currently, nested recurrsive validation rule errors never display since the logic to show them is using the null coalescing operator. Since the `$errors->first($statePath)` is always set (but evaluates to false) it never shows the nested rules. 

This fixes it by changing to the ternary operator.

### DIFF
--- a/packages/forms/resources/views/components/field-wrapper/index.blade.php
+++ b/packages/forms/resources/views/components/field-wrapper/index.blade.php
@@ -52,7 +52,7 @@
 
             @if ($errors->has($statePath) || ($hasNestedRecursiveValidationRules && $errors->has("{$statePath}.*")))
                 <x-forms::field-wrapper.error-message>
-                    {{ $errors->first($statePath) ? $errors->first($statePath) : ($hasNestedRecursiveValidationRules ? $errors->first("{$statePath}.*") : null) }}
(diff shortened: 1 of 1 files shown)


packages/forms/resources/views/components/field-wrapper/index.blade.php
```diff
@@ -52,7 +52,7 @@
 
             @if ($errors->has($statePath) || ($hasNestedRecursiveValidationRules && $errors->has("{$statePath}.*")))
                 <x-forms::field-wrapper.error-message>
-                    {{ $errors->first($statePath) ?? ($hasNestedRecursiveValidationRules ? $errors->first("{$statePath}.*") : null) }}
+                    {{ $errors->has($statePath) ? $errors->first($statePath) : ($hasNestedRecursiveValidationRules ? $errors->first("{$statePath}.*") : null) }}
```
danharrin:
```suggestion
                    {{ $errors->first($statePath) ?: ($hasNestedRecursiveValidationRules ? $errors->first("{$statePath}.*") : null) }}
```

archilex:
Would you like me to add this validation method to the documentation?

danharrin:
No thanks, it doesn't apply everywhere, I've yet to figure out where it fits

PR #6026, opened 2023-03-25
fix: resolves error regarding canvas already being used when updating

boumanb:
Resolves the following error:

Uncaught Error: Canvas is already in use. Chart with ID '0' must be destroyed before the canvas with ID '' can be reused.

Somehow the chart instance was not being kept correctly in `this.chart`.

I think this was being mentioned in https://github.com/filamentphp/filament/discussions/2552#discussioncomment-3559218 but hasn't been fixed.

### DIFF
--- a/packages/admin/resources/views/components/stats/card.blade.php
+++ b/packages/admin/resources/views/components/stats/card.blade.php
@@ -72,7 +72,7 @@
                 init: function () {
                     chart = Chart.getChart(this.$refs.canvas);
 
-                    chart !== undefined
+                    (chart !== undefined)
                         ? this.updateChart()
                         : this.initChart()
                 },
--- a/packages/admin/resources/views/components/stats/card.blade.php
+++ b/packages/admin/resources/views/components/stats/card.blade.php
@@ -66,17 +66,19 @@
         <div
             x-title="filament-stats-card-chart"
             x-data="{
-                chart: null,
-
                 labels: {{ json_encode(array_keys($chart)) }},
                 values: {{ json_encode(array_values($chart)) }},
 
                 init: function () {
-                    this.chart ? this.updateChart() : this.initChart()
+                    chart = Chart.getChart(this.$refs.canvas);
+
+                    chart !== undefined
+                        ? this.updateChart()
+                        : this.initChart()
                 },
 
                 initChart: function () {
-                    return this.chart = new Chart(this.$refs.canvas, {
+                    return chart = new Chart(this.$refs.canvas, {
                         type: 'line',
                         data: {
                             labels: this.labels,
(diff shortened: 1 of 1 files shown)


packages/admin/resources/views/components/stats/card.blade.php
```diff
@@ -117,9 +119,9 @@
                 },
 
                 updateChart: function () {
-                    this.chart.data.labels = this.labels
-                    this.chart.data.datasets[0].data = this.values
-                    this.chart.update()
+                    chart.data.labels = this.labels
```
danharrin, line 122: Does this actually still work? How does it access the chart variable without a property on the object?

boumanb: Yes it does work. I'm not able to explain why since I'm not that experienced with alpine. All I know is that it wasn't working when I started and it ended op working with the proposed changes. Sorry.

danharrin: Could you double check it works when polling is used with the chart? So the chart data changes data

boumanb: Yes it does. Maybe you could check it too.

danharrin:
Could you check `chart-widget.blade.php` to see if the same problem occurs, and if so ensure that it's consistent code with this PR?

boumanb:
> Could you check `chart-widget.blade.php` to see if the same problem occurs, and if so ensure that it's consistent code with this PR?

Just checked the functionality and `chart-widget.blade.php` works fine. The code, however, is not consistent, `packages/admin/resources/views/widgets/chart-widget.blade.php` leverages a emit (`updateChartData`) to update the chart. While `packages/admin/resources/views/components/stats/card.blade.php` does not. I'm not experienced enough with the Filament codebase to make both more consistent. All that matters to me (and others) is that both charts function correctly.

If you insist, I could try to make a proper PR with all the required refactoring to make both more consistent. This, however, is something I have to make time for to do it properly. While the proposed changes in this PR already fix the bug.

Thanks in advance.

danharrin:
Thanks

zepfietje:
@danharrin @boumanb, I'm pretty sure this PR broke charts.
I don't see how the chart variable can be accessed across functions if it's not stored in a property.

When toggling dark mode, the chart colors are no longer updated.

packages/admin/resources/views/components/stats/card.blade.php
zepfietje: I'm quite sure this doesn't work and broke charts actually...

danharrin:
Reverting.

boumanb:
> Reverting.

@danharrin @zepfietje reverting will break polling of the cards I guess (working as of `v2.17.20`). Attached video as proof of polling working and to verify that we're referencing the same chart.

https://user-images.githubusercontent.com/29407220/229120061-9aa4608e-f703-4ee2-8c57-b1443797725c.mov

danharrin:
Toggle dark mode and try again

boumanb:
> Toggle dark mode and try again

Just did. Chart not updating (broken as of `v2.17.21`). I'll try to make time so that I can create a proper PR.

https://user-images.githubusercontent.com/29407220/229122512-971af29f-930f-430d-a4ad-4e5258dd911d.mov

PR #6033, opened 2023-03-26
Adds dropdown action divider

archilex:
Since there's no ETA on Livewire v3 (and thus Filament v3) and also since some people who are currently using Filament v2 might not be able to update to v3 when it's out, I wanted to go ahead and add this feature to allow dropdown actions groups to have dividers. 

This  PR adds the `->dividerBefore()` and `dividerAfter()` methods to the BaseAction class.

Obvisouly, this is NOT the best way to implement this feature. Ideally there would be an entirly new `ListGroup` class or something like that and then the divider could be added through tailwinds `divide-y` class, BUT that would be a breaking change to v2 so this seems like the best way to add this. 

Hopefully v3 will support this natively and more elegantly, but for v2 users hopefully this helps.

### DIFF
--- a/packages/admin/resources/views/components/dropdown/list/divider.blade.php
+++ b/packages/admin/resources/views/components/dropdown/list/divider.blade.php
@@ -1,0 +1,1 @@
+<div {{ $attributes->class(['border-t border-gray-100 -mx-1 my-1 dark:border-gray-700']) }}></div>
--- a/packages/support/resources/views/components/actions/group.blade.php
+++ b/packages/support/resources/views/components/actions/group.blade.php
@@ -30,8 +30,14 @@
 
     <x-filament-support::dropdown.list>
         @foreach ($actions as $action)
+            @if ($action->hasDividerBefore())
+                <x-filament-support::dropdown.list.divider />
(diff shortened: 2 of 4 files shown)


zepfietje:
Thanks for the PR, @archilex.
However, like you said, the better way would be to support groups (dropdown list Blade component exists for this purpose).

I've discussed this before with @danharrin, and for now we concluded you could manually build your dropdown using Blade components. I do think though that it would be nice to revisit this and support setting groups in the actual `ActionGroup` object.

Closing this PR, but let's continue the discussion for a proper solution.

archilex:
Understandable @zepfietje. It isn't the most elegant solution. 

Actually, I was just digging through v3 source code and it looks like you all have this implemented for v3. Nice! Looks like you can nest ActionGroups as `Groupable`? Glancing at the changes I assume this is a breaking change, but maybe something similar could be implemented in v2? Should I take a look?

Or are you saying you all prefer to leave v2 as is?

danharrin:
I think the nested groups in v3 actually open nested dropdowns, but I could be wrong.

PR #6034, opened 2023-03-27
✨ Make a `CreateRecord` specific stub

juliomotol:
This PR aims to add a separate stub for `CreateRecord` generated by the `make:filament-resource` command.

Introducing a separate stub for `CreateRecord` allows us to customize the generated file without conflicting what the `make:filament-page --resource` command would generate.


danharrin:
Thanks, but I think the maintenance burden increase for us outweighs how useful this will actually be to people.

PR #6040, opened 2023-03-27
Allow testing fields via callbacks

ganyicz:
This PR adds the ability to test fields using callbacks like:

```php
it('has a title field', function () {
    livewire(CreatePost::class)
        ->assertFormFieldExists('title', function (TextInput $input) {
            return $input->isDisabled();
        });
});
```

This is inspired by Laravel testing helpers, eg:

```php
Mail::assertSent(OrderShipped::class, function (OrderShipped $mail) use ($user) {
    return $mail->hasTo($user->email);
});
```

The functionality is backwards compatible with the current testing helper where form can be passed in as second argument. In that case the callback needs to be passed in as thrid argument.

I also refactored `assertFormFieldIsDisabled` and `assertFormFieldIsEnabled` to use the new api

PR #6041, opened 2023-03-27
Add isReactive method

ganyicz:
I recently had to test specific configuration of a field including it's state bindings.
This would be helpful in combination with #6040:

```php
it('title field is reactive', function () {
    livewire(CreatePost::class)
        ->assertFormFieldExists('title', function (TextInput $input) {
            return $input->isReactive();
        });
});
```


danharrin:
Thanks

PR #6042, opened 2023-03-27
Allow for easier integration with Bouncer

ganyicz:
I'm currently integrating [Bouncer](https://github.com/JosephSilber/bouncer) into filament and I came across an obstacle with the fact that the existance of a policy and corresponding method is required for the ability to be checked at the gate.

This is confliting with how Bouncer works by default: the ability is checked even when the policy does not exist - this allows for super simple implementation of permissions just by granting the abilities and not having to worry about policies until you need them. Having to create policies in order to integrate with Filament breaks the DX.

My solution is to add a `Resource::alwaysCheckGate()` toggle which bypasses policies completely and checks everything at the gate when enabled. This can be then simply called inside a service provider by the user.

### DIFF
--- a/packages/admin/src/Resources/RelationManagers/RelationManager.php
+++ b/packages/admin/src/Resources/RelationManagers/RelationManager.php
@@ -6,6 +6,7 @@ use Closure;
 use Filament\Facades\Filament;
 use Filament\Http\Livewire\Concerns\CanNotify;
 use Filament\Resources\Form;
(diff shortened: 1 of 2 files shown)


danharrin:
Thanks

KnudH:
I think this change did not survive the migration from v2 to v3, right? Is there a different approach to make filament work together with Bouncer?

PR #6045, opened 2023-03-28
Fix "Illegal offset type" when query in mongodb

tungtm-unica:
this error should be handled when the query of [laravel mongodb](https://github.com/jenssegers/laravel-mongodb/blob/master/src/Query/Builder.php#L1033) package sets default query type as string

### DIFF
--- a/packages/tables/src/Columns/Concerns/InteractsWithTableQuery.php
+++ b/packages/tables/src/Columns/Concerns/InteractsWithTableQuery.php
@@ -87,6 +87,10 @@ trait InteractsWithTableQuery
 
         foreach ($this->getSearchColumns() as $searchColumnName) {
             $whereClause = $isFirst ? 'where' : 'orWhere';
+            $searchColumnName = match ($databaseConnection->getDriverName()) {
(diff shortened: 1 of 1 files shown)


danharrin:
Instead of this Mongo-specific override, is there a reason why we shouldn't always cast it to a string for all drivers?

Also what is Mongo returning if its not a string?

zepfietje:
I agree we need a little more context on this issue, @tungtm-unica.

zepfietje:
@tungtm-unica, could you clarify?

danharrin:
Closing due to inactivity. Please reopen if you clarify our questions.

f24aalam:
I am having the same issue while searching.

> Instead of this Mongo-specific override, is there a reason why we shouldn't always cast it to a string for all drivers?

Don't know about this

> Also what is Mongo returning if its not a string?

this is what I am getting 
```
Illuminate\Support\Stringable {#2196 ▼ // vendor/filament/tables/src/Columns/Concerns/InteractsWithTableQuery.php:97
  #value: "userName"
}

```

PR #6051, opened 2023-03-29
Fixed the collapsedGroups value check adding support for the json representation of the null value

LBreda:
In some cases the collapsedGroups value is the json stringified version of the null value, the string "null".

It seems related to discussion #5457

### DIFF
--- a/packages/admin/resources/views/components/layouts/app/sidebar/index.blade.php
+++ b/packages/admin/resources/views/components/layouts/app/sidebar/index.blade.php
@@ -87,7 +87,10 @@
         @endphp
 
         <script>
-            if (JSON.parse(localStorage.getItem('collapsedGroups')) === null || JSON.parse(localStorage.getItem('collapsedGroups')) === 'null') {
+            if (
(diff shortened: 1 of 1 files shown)


packages/admin/resources/views/components/layouts/app/sidebar/index.blade.php
```diff
@@ -87,7 +87,7 @@ class="filament-sidebar-close-button shrink-0 flex items-center justify-center w
         @endphp
 
         <script>
-            if (JSON.parse(localStorage.getItem('collapsedGroups')) === null) {
+            if (JSON.parse(localStorage.getItem('collapsedGroups')) === null || JSON.parse(localStorage.getItem('collapsedGroups')) === 'null') {
```
danharrin:
```suggestion
            if (
                (JSON.parse(localStorage.getItem('collapsedGroups')) === null) ||
                (JSON.parse(localStorage.getItem('collapsedGroups')) === 'null')
            ) {
```

danharrin:
Thanks

PR #6057, opened 2023-03-29
Fix test examples to use getRouteKey() for custom route key names in documentation

mnurullahsaglam:
This pull request addresses a bug in the test examples where the usage of **getKey()** would cause test failures when a custom route key name is specified in the model by overriding the **getRouteKeyName()** method. The **getKey()** method returns the primary key of the model, but in some cases, a custom field like a slug might be used as the route key name.

Changes include:

Replacing **getKey()** with **getRouteKey()** in the test examples to ensure compatibility with custom route key names in models, such as when using a slug or another field instead of the primary key.
Updating the testing documentation to reflect these changes and provide accurate guidance for users with custom route key names, explaining the difference between **getKey()** and **getRouteKey()**.
These changes will help users with custom route key names avoid test failures and improve the overall testing experience.

### DIFF
--- a/packages/admin/docs/10-testing.md
+++ b/packages/admin/docs/10-testing.md
@@ -191,7 +191,7 @@ it('can validate input', function () {
     $newData = Post::factory()->make();
 
     livewire(PostResource\Pages\EditPost::class, [
-        'record' => $post->getKey(),
+        'record' => $post->getRouteKey(),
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks

mnurullahsaglam:
Thank you

PR #6073, opened 2023-03-31
Fix assertNotified no longer working with multiple notifications

AmandoVledder:
Using`assertNotified` with a notification instance when having multiple notifications did no longer work because it was trying to find notifications by comparing the ID of the mountedNotification against its key in the send notifications array instead of comparing it with the ID of the given notification. 

And because notifications are keyed by ID it would always match on the first notification in the array.

### DIFF
--- a/packages/notifications/src/Notification.php
+++ b/packages/notifications/src/Notification.php
@@ -248,7 +248,7 @@ class Notification extends ViewComponent implements Arrayable
         }
 
         if ($notification instanceof Notification) {
-            $expectedNotification = $notifications->first(fn (Notification $mountedNotification, string $key): bool => $mountedNotification->id === $key);
+            $expectedNotification = $notifications->first(fn (Notification $mountedNotification): bool => $mountedNotification->id === $notification->id);
         }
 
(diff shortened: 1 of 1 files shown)


danharrin:
This is completely breaking the function in our test suite

AmandoVledder:
@danharrin ah yeah i see, its probably because those assertions are using Notification::make without a predefined ID. Do you have any suggestions for a possible different fix? Because now its not possible to assert multiple notifications have been send using notification instances.

Something like this used to work 

```php
test('can assert that multiple notifications are sent', function () {
    livewire(PageActions::class)
        ->callPageAction('two_notifications')
        ->assertNotified(
            Notification::make('first_notification')
                ->title('First notification')
                ->success()
        )
        ->assertNotified(
            Notification::make('second_notification')
                ->title('Second notification')
                ->success()
        )
    ;
});
```

danharrin:
Instead of comparing IDs, `->toArray()` each object, remove the `id` key from the array, and then check if all the entries in the expected array match the entires in the actual array?

PR #6083, opened 2023-03-31
fix: Delete media library files

### LINKED ISSUE #5971: SpatieMediaLibraryFileUpload file removal takes effect before saving model
### Package

filament/filament

### Package Version

v2.17.16

### Laravel Version

v10.3.3

### Livewire Version

v2.12.3

### PHP Version

PHP 8.2.1

### Problem description

When working with a single-file SpatieMediaLibraryFileUpload field, using the "x" button to remove the file takes effect immediately, rather than when the form is submitted. Additionally, hitting "cancel" does not undo the change; the image is permanently removed from the model. Someone previously mentioned this in https://github.com/filamentphp/filament/discussions/4698.

### Expected behavior

I'd expect the change to take effect only once "Save changes" is clicked, and that "cancel" would leave the model unchanged.

### Steps to reproduce

I've built a bare-bones repro at https://github.com/ceejayoz/filament-repro, with full install/repro steps in the readme.

1. Add a `SpatieMediaLibraryFileUpload::make('default')` to a Filament resource's form.
2. Create a record that has an image.
3. Open the edit view on that record.
4. Hit the "x" button by the uploaded image.
5. Note that the Media item has disappeared from the database immediately, without saving the form.

### Reproduction repository

https://github.com/ceejayoz/filament-repro

### Relevant log output

_No response_

danharrin:
Fixes #5971.

### DIFF
--- a/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
+++ b/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
@@ -4,6 +4,7 @@ namespace Filament\Forms\Components;
 
 use Closure;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
 use League\Flysystem\UnableToCheckFileExistence;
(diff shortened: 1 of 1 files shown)


ceejayoz:
Thank you!

PR #6092, opened 2023-04-01
fix: Nested BelongsTo layout component

### LINKED ISSUE #6039: Nested relationships not working on creation.
### Package

filament/filament

### Package Version

v2.17.18

### Laravel Version

v10.4.1

### Livewire Version

_No response_

### PHP Version

PHP 8.2.4

### Problem description

When trying to `create` a new model with 2 level nested relations only the first level gets saved.

The nested structure is as followed:

-> user
--> person
---> contact

All these relations are defined by `morphOne` relationships. 

### Expected behavior

After saving the new model only `user` and `person` exist but the expected behavior would be the have a `user` with a `person` and a `contact` belonging to the `person`.

In case of `edit` it works but not in case of `create`.

### Steps to reproduce

Download "Reproduction repository" and run

```
composer install
```

Edit .env file and the run

```
php artisan migrate
php artisan make:filament-resource User
```

Login in to Filament admin and try to create e new user.

### Reproduction repository

https://github.com/danielbehrendt/filament-test

### Relevant log output

_No response_

danharrin:
Fixes #6039.

### DIFF
--- a/packages/forms/src/Components/Concerns/BelongsToModel.php
+++ b/packages/forms/src/Components/Concerns/BelongsToModel.php
@@ -13,6 +13,8 @@ trait BelongsToModel
 
     protected ?Closure $saveRelationshipsUsing = null;
 
+    protected ?Closure $saveRelationshipsBeforeChildrenUsing = null;
+
     protected bool | Closure $shouldSaveRelationshipsWhenHidden = false;
 
     public function model(Model | string | Closure | null $model = null): static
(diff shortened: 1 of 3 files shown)


JonatasBSM:
is this fixed for filament 3 too? im currently having same problem with my create nested belongsTo form. it saves the first relation level, but nothing happens with the second one. it work as it should on edit

masihgh:
still ihave same issue in livewire 3

"php": "^8.2",
"filament/filament": "^3.2",
"laravel/framework": "^11.9",

PR #6100, opened 2023-04-02
Support route parameters on resource slug

linaspasv:
This small adjustment allows to build resource URLs with multiple named parameters, otherwise it results in a missing required parameter error when user navigates to the resource with route parameters defined in the slug.

```php
    {
        return $table
            ->columns([
                Tables\Columns\TextColumn::make('title'),
            ])
            ->actions([
                Tables\Actions\Action::make('categories')
                    ->url(fn (Menu $record) => CategoryResource::getUrl('index', ['menu' => $record->id]))
                    ->icon('heroicon-o-tag')
                    ->color('success'),
            ]);
    }
```

```php
class CategoryResource extends Resource
{
    protected static ?string $model = Category::class;

    protected static ?string $slug = 'menus/{menu}/categories';

    protected static bool $shouldRegisterNavigation = false;

    .........

    {
        return [
            'index' => Pages\ListCategories::route('/'),
            'create' => Pages\CreateCategory::route('/create'),
            'edit' => Pages\EditCategory::route('/{record}/edit'),
        ];
    }
````

### DIFF
--- a/packages/admin/src/Resources/Resource.php
+++ b/packages/admin/src/Resources/Resource.php
@@ -17,6 +17,7 @@ use Illuminate\Support\Facades\Gate;
 use Illuminate\Support\Facades\Route;
 use Illuminate\Support\Str;
 use Illuminate\Support\Traits\Macroable;
+use Route;
 
 class Resource
 {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/admin/src/Resources/Resource.php [hunk 1]
```diff
@@ -17,6 +17,7 @@
 use Illuminate\Support\Facades\Route;
 use Illuminate\Support\Str;
 use Illuminate\Support\Traits\Macroable;
+use Route;
```
zepfietje: I'm pretty sure this breaks.

zepfietje (changes requested):

packages/admin/src/Resources/Resource.php
linaspasv: You are right, route is double declared. I will update.

PR #6103, opened 2023-04-03
Add placeholder property to `TextColumn`s and `BadgeColumn`s

austincarpenter:
Sometimes, a placeholder value such as '–', 'None', 'Never', etc., is preferable to an otherwise empty table cell.

This PR allows text-based table columns whose state is empty (and not numeric) to use a placeholder string upon being formatted.

The reason for checking for a non-numeric state is so that `0` and `0.0` are not considered worthy of falling back to a placeholder, given that a zero in the context of other numbers (such as an aggregate column) should likely remain a 0.

### DIFF
--- a/packages/tables/docs/03-columns/02-text.md
+++ b/packages/tables/docs/03-columns/02-text.md
@@ -166,7 +166,7 @@ TextColumn::make('status')
     ->formatStateUsing(fn (string $state): string => __("statuses.{$state}"))
 ```
 
-## Adding a placeholder
+## Adding a placeholder if the cell is empty
 
 Sometimes you may want to display a placeholder if the cell's value is empty:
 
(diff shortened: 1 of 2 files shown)


danharrin:
Is this not the same as `default()`?

austincarpenter:
Similar but the key difference here is that `placeholder()` doesn't actually modify the state, just the formatting of it.

An example would be using the `date()` helper. Using `default()` causes a Carbon exception when 'Not paid' is parsed. By using `placeholder()`, the default is cosmetic, not affecting the state. In this case, `null` would be the expected state, but formatted right at the end of the chain to make the state more understandable to the end user.

```php
Tables\Columns\TextColumn::make('paid_at')
    ->label('Paid')
    ->date()
    ->default('Not paid') //Exception, alters state
    ->placeholder('Not paid') //No exception, state maintained
```

Because of nuances like this, `placeholder()` is far more easily applied globally using `TextColumn::configureUsing()`, for example.

This also ensures stricter types when using closure customisation with state injection, as `?Carbon $state` is preferable to `Carbon | string | null $state` where `string` is introduced with `default()` in the example.

PR #6110, opened 2023-04-03
Do not x-bind x-tooltip to translator string

tanthammar:
fix console error


danharrin:
Hmm... it's a Blade component so why does `:` syntax not work? It's consistent with everywhere else we use it

tanthammar:
I couldn't see you had `:` on any other  x-tooltip.raw. When I removed it, the error in console was gone.

tanthammar:
Closing, this was due to me merging my code to v3

PR #6117, opened 2023-04-04
Section heading with custom views

andresilvagomez:
Allow sections heading has custom views with his own behavior, not wrapped in h3

### DIFF
--- a/packages/forms/resources/views/components/section.blade.php
+++ b/packages/forms/resources/views/components/section.blade.php
@@ -56,6 +56,28 @@
             'filament-forms-section-header flex-1 space-y-1',
             'cursor-pointer' => $isCollapsible,
         ])>
+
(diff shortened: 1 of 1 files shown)


yelloalejo (approved):

JhonatanR89 (approved):

danharrin:
I don't understand what you're trying to do

danharrin:
This definitely doesn't work, with a normal heading there are 2 h3s and 2 icons?

andresilvagomez:
Screenshot 2023-04-07 at 8 44 51 AM

@danharrin for example, in that case I built a UI with a button, but if is wrapper in h2 the button will not work correctly.

Or really build a complete header.

danharrin:
This isn't the right solution, I think it's better if you create your own layout component instead of hacking the section to work how you want it to. You can copy the exact code from vendor so it is very easy.

PR #6129, opened 2023-04-05
Add more options for section aside

josefbehr:
* add method `asideCompact` to make the heading and description take up less width
* add method `asideOnRight` to move the aside part to the right of the form


danharrin:
Let's work out a solution where we combine existing methods like `->aside()->compact()` or `->aside()->formOnLeft()`

josefbehr:
> Let's work out a solution where we combine existing methods like `->aside()->compact()` or `->aside()->formOnLeft()`

I honestly don't think that's such a great idea. That would mean that `side()` would have to return some other class than the component class object it is part of, which - as far as I know - doesnt happen in filament anywhere else. That would also introduce an ambiguity and backwards-incompatibility for chaining `aside()` and `compact()` (which already is a method on the section component itself).

May you tell me what the reasoning here is? If you dont like the additional methods for some reason, we could move it into parameters for the `aside()` call, although I personally like to have specific methods that can be closure-customized.

danharrin:
Why does it need to return a different object? Return static and add all the methods to the Section class

josefbehr:
Sorry, I guess I misunderstood. You mean to not add aside-specific methods but have it behave differently, when e.g. `compact()` is applied to the section? Personally, I don't want the whole section to be compact, but rather just have a smaller aside.
As for `formOnLeft` I did not find that method in the `Section` class or its traits.

danharrin:
`formOnLeft` should be added. Probably as `formBefore()` instead as its more RTL friendly.

We can check both `compact()` and `aside()` together at once to combine the functionalities.

josefbehr:
> `formOnLeft` should be added. Probably as `formBefore()` instead as its more RTL friendly.

I can rename the `asideOnRight()` method ...

> We can check both `compact()` and `aside()` together at once to combine the functionalities.

But what if I want a smaller aside without having a compact section?

danharrin:
Small aside - compact() aside()
Normal aside - aside()
Small - compact()

josefbehr:
> Small aside - compact() aside() Normal aside - aside() Small - compact()

I pushed some changes. Is this how you imagined it to work?

packages/forms/src/Components/Section.php
```diff
@@ -111,4 +113,17 @@ public function getIcon(): ?string
     {
         return $this->evaluate($this->icon);
     }
+
+    public function formBefore(bool | Closure | null $formBefore = true): static
+    {
+        $this->formBefore = $formBefore;
```
danharrin:
```suggestion
    public function formBefore(bool | Closure | null $condition = true): static
    {
        $this->isFormBefore = $condition;
```

packages/forms/src/Components/Section.php
```diff
@@ -24,6 +24,8 @@ class Section extends Component implements Contracts\CanConcealComponents, Contr
 
     protected string | Closure | null $icon = null;
 
+    protected bool | Closure | null $formBefore = null;
```
danharrin:
```suggestion
    protected bool | Closure | null $isFormBefore = null;
```

packages/forms/src/Components/Section.php
```diff
@@ -111,4 +113,17 @@ public function getIcon(): ?string
...
+    public function formBefore(bool | Closure | null $condition = true): static
+    {
+        $this->isFormBefore = $condition;
+
+        return $this;
+    }
+
+    public function isFormBefore(): bool
+    {
+        return (bool) ($this->evaluate($this->formBefore) ?? false);
```
danharrin:
```suggestion
        return (bool) $this->evaluate($this-> isFormBefore);
```

packages/forms/src/Components/Section.php
```diff
@@ -24,6 +24,8 @@ class Section extends Component implements Contracts\CanConcealComponents, Contr
 
     protected string | Closure | null $icon = null;
 
+    protected bool | Closure | null $isFormBefore = null;
```
danharrin:
```suggestion
    protected bool | Closure $isFormBefore = false;
```

packages/forms/src/Components/Section.php
```diff
@@ -111,4 +113,17 @@ public function getIcon(): ?string
     {
         return $this->evaluate($this->icon);
     }
+
+    public function formBefore(bool | Closure | null $condition = true): static
```
danharrin:
```suggestion
    public function formBefore(bool | Closure $condition = true): static
```

danharrin:
Looks pretty good! Now passing this over to @zepfietje for design review & merge

zepfietje:
Great, will have a look next week!

packages/forms/src/Components/Section.php
```diff
@@ -111,4 +113,17 @@ public function getIcon(): ?string
...
+    public function formBefore(bool | Closure $condition = true): static
+    {
+        $this->isFormBefore = $condition;
+
+        return $this;
+    }
+
+    public function isFormBefore(): bool
+    {
+        return (bool) $this->evaluate($this-> isFormBefore);
```
danharrin:
```suggestion
        return (bool) $this->evaluate($this->isFormBefore);
```

zepfietje:
@danharrin @josefbehr, before I review this, what should I pay attention to?
In other words, what has been added in this PR that you want me to visually review? 🙂

danharrin:
@zepfietje compact aside layout, and aside layout with the form on the left of the headings

danharrin:
Thanks @josefbehr!

PR #6131, opened 2023-04-05
Improve composer.json snippet in README

JHWelch:
Right now the `composer.json` example on the README has two ellipses indicating that this is just an example. 

The ellipses are not legal JSON characters so they appear as a red error in the README. 

by swapping to `jsonc` and adding `//` these now appear with "comment" syntax highlighting and do not distract. 

image
updated to
CleanShot 2023-04-05 at 12 49 21@2x


zepfietje:
Thanks!

PR #6140, opened 2023-04-06
Add placeholder to search input in Select component

dmandrade:
Add option to set a placeholder in the search field providing a better ux when the select already has preloaded data.

image

image

### DIFF
--- a/packages/forms/resources/js/components/select.js
+++ b/packages/forms/resources/js/components/select.js
@@ -58,8 +58,8 @@ export default (Alpine) => {
                         removeItemButton: true,
                         renderChoiceLimit: optionsLimit,
                         searchFields: ['label'],
-                        searchResultLimit: optionsLimit,
                         searchPlaceholderValue: searchPlaceholder,
+                        searchResultLimit: optionsLimit,
                         shouldSort: false,
                     })
 
--- a/packages/forms/resources/js/components/select.js
+++ b/packages/forms/resources/js/components/select.js
@@ -50,7 +50,7 @@ export default (Alpine) => {
                             window.pluralize(maxItemsMessage, maxItemCount, {
                                 count: maxItemCount,
                             }),
-                        noChoicesText: null,
+                        noChoicesText: searchPrompt,
                         noResultsText: noSearchResultsMessage,
                         placeholderValue: placeholder,
                         position: position ?? 'auto',
--- a/packages/forms/resources/js/components/select.js
+++ b/packages/forms/resources/js/components/select.js
@@ -26,6 +26,7 @@ export default (Alpine) => {
             searchDebounce,
             searchingMessage,
             searchPrompt,
+            searchPlaceholder,
             state,
         }) => {
             return {
@@ -58,6 +59,7 @@ export default (Alpine) => {
                         renderChoiceLimit: optionsLimit,
                         searchFields: ['label'],
                         searchResultLimit: optionsLimit,
+                        searchPlaceholderValue: searchPlaceholder,
                         shouldSort: false,
                     })
 
--- a/packages/forms/resources/js/components/select.js
+++ b/packages/forms/resources/js/components/select.js
@@ -25,8 +25,8 @@ export default (Alpine) => {
             position,
             searchDebounce,
             searchingMessage,
-            searchPrompt,
             searchPlaceholder,
+            searchPrompt,
             state,
         }) => {
             return {
--- a/packages/forms/resources/js/components/select.js
+++ b/packages/forms/resources/js/components/select.js
@@ -25,7 +25,6 @@ export default (Alpine) => {
             position,
             searchDebounce,
             searchingMessage,
-            searchPlaceholder,
             searchPrompt,
             state,
         }) => {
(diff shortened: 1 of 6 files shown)


packages/forms/resources/js/components/select.js
```diff
@@ -58,6 +59,7 @@ export default (Alpine) => {
                         renderChoiceLimit: optionsLimit,
                         searchFields: ['label'],
                         searchResultLimit: optionsLimit,
+                        searchPlaceholderValue: searchPlaceholder,
```
zepfietje:
```suggestion
                        searchPlaceholderValue: searchPlaceholder,
                        searchResultLimit: optionsLimit,
```

packages/forms/resources/js/components/select.js
```diff
@@ -26,6 +26,7 @@ export default (Alpine) => {
             searchDebounce,
             searchingMessage,
             searchPrompt,
+            searchPlaceholder,
```
zepfietje:
```suggestion
            searchPlaceholder,
            searchPrompt,
```

packages/forms/resources/lang/en/components.php
```diff
@@ -250,6 +250,8 @@
 
         'search_prompt' => 'Start typing to search...',
 
+        'search_placeholder' => 'Search',
+
```
zepfietje:
```suggestion

        'search_placeholder' => 'Search',

        'search_prompt' => 'Start typing to search...',
```

packages/forms/resources/lang/en/components.php
```diff
@@ -248,8 +248,9 @@
 
         'searching_message' => 'Searching...',
 
-        'search_prompt' => 'Start typing to search...',
+        'search_placeholder' => 'Search',
 
+        'search_prompt' => 'Start typing to search...',
```
zepfietje:
```suggestion
        'search_prompt' => 'Start typing to search...',

```

packages/forms/resources/lang/pt_BR/components.php
```diff
@@ -249,6 +249,8 @@
         'searching_message' => 'Buscando...',
 
         'search_prompt' => 'Comece a digitar para pesquisar...',
+
+        'search_placeholder' => 'Buscar',
```
zepfietje:
```suggestion
        'search_placeholder' => 'Buscar',

        'search_prompt' => 'Comece a digitar para pesquisar...',

```

packages/forms/resources/views/components/select.blade.php
```diff
@@ -99,6 +99,7 @@
                         searchDebounce: @js($getSearchDebounce()),
                         searchingMessage: @js($getSearchingMessage()),
                         searchPrompt: @js($getSearchPrompt()),
+                        searchPlaceholder: @js($getSearchPlaceholder()),
```
zepfietje:
```suggestion
                        searchPlaceholder: @js($getSearchPlaceholder()),
                        searchPrompt: @js($getSearchPrompt()),
```

packages/forms/src/Components/Concerns/CanBeSearchable.php
```diff
@@ -17,6 +17,8 @@ trait CanBeSearchable
 
     protected string | Htmlable | Closure | null $searchPrompt = null;
 
+    protected string | Closure | null $searchPlaceholder = null;
```
zepfietje:
```suggestion
    protected string | Closure | null $searchPlaceholder = null;

    protected string | Htmlable | Closure | null $searchPrompt = null;
```

packages/forms/src/Components/Concerns/CanBeSearchable.php
```diff
@@ -52,6 +54,13 @@ public function searchPrompt(string | Htmlable | Closure | null $message): stati
         return $this;
     }
 
+    public function searchPlaceholder(string | Closure | null $message): static
+    {
+        $this->searchPlaceholder = $message;
+
+        return $this;
+    }
+
```
zepfietje: Could you move this above `searchPrompt` to ensure alphabetical sort, @dmandrade?

packages/forms/src/Components/Concerns/CanBeSearchable.php
```diff
@@ -62,6 +71,11 @@ public function getSearchPrompt(): string | Htmlable
         return $this->evaluate($this->searchPrompt) ?? __('forms::components.select.search_prompt');
     }
 
+    public function getSearchPlaceholder(): string | Htmlable
+    {
+        return $this->evaluate($this->searchPlaceholder) ?? __('forms::components.select.search_placeholder');
+    }
+
```
zepfietje: Move this too.

zepfietje (changes requested):
Mark as ready for review when done.

danharrin:
Why can't we use the existing `searchPrompt()`?

dmandrade:
> Why can't we use the existing `searchPrompt()`?

I thought about it but I believe that choicesjs will always display a message under the search field when there are no preloaded items, so I thought it best to add one more configuration option.

zepfietje:
Can we hide the prompt? Then just displaying the placeholder instead should be fine maybe?

dmandrade:
I can do that. 
do you think it would be a better ux just display the input with placeholder inside?

danharrin:
We shouldnt need to add any more methods to the class to do this

dmandrade:
New methods have been removed and the class is as it was before.
just passing the searchPrompt as placeholder and hiding the "has-no-choices" element when the field is a single selection.

danharrin:
Thanks

PR #6151, opened 2023-04-08
Fix dropdown button loading indicator to work with livewire event modifiers 

howdu:
Allow `wire:click.stop` events to work with dropdown.

### DIFF
--- a/packages/support/resources/views/components/icon-button.blade.php
+++ b/packages/support/resources/views/components/icon-button.blade.php
@@ -50,7 +50,7 @@
     $hasLoadingIndicator = filled($wireTarget) || ($type === 'submit' && filled($form));
 
     if ($hasLoadingIndicator) {
-        $loadingIndicatorTarget = html_entity_decode($wireTarget ?: $form), ENT_QUOTES);
+        $loadingIndicatorTarget = html_entity_decode($wireTarget ?: $form, ENT_QUOTES);
     }
 @endphp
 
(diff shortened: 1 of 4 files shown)


zepfietje:
Thanks for the PR, @howdu!
Is this behavior already supported on other components like buttons?

howdu:
No but I've applied this to the other components now too.

packages/support/resources/views/components/icon-button.blade.php
```diff
@@ -45,10 +45,12 @@
...
     ]);
 
-    $hasLoadingIndicator = filled($attributes->get('wire:target')) || filled($attributes->get('wire:click')) || (($type === 'submit') && filled($form));
+    $wireTarget = $attributes->whereStartsWith(['wire:target', 'wire:click'])->first();
+
+    $hasLoadingIndicator = filled($wireTarget) || ($type === 'submit' && filled($form));
 
     if ($hasLoadingIndicator) {
-        $loadingIndicatorTarget = html_entity_decode($attributes->get('wire:target', $attributes->get('wire:click', $form)), ENT_QUOTES);
+        $loadingIndicatorTarget = html_entity_decode($wireTarget ?: $form, ENT_QUOTES);
```
zepfietje, line 53: Stricter to use `??` perhaps?

```suggestion
        $loadingIndicatorTarget = html_entity_decode($wireTarget ?? $form, ENT_QUOTES);
```

howdu: I don't think so because the variable is always going to be set and covers empty string.

zepfietje: Will it be set to an empty string or `null`?

PR #6155, opened 2023-04-08
Section compact rounded corners

howdu:
Rebuild resources for rounded-t-lg/rounded-b-lg


danharrin:
I'll leave this one to you for merging @zepfietje

zepfietje:
@howdu, could you share screenshots of before and after this change?
Makes it easier for me to judge if this is something we should merge now. 😊

howdu:
Sure no problem. I just thought it looked better when it's consistent with the input roundedness especially when it's inside a fieldset/card with `rounded-xl`.

Screenshot 2023-04-08 at 15 56 35

zepfietje:
Thanks for the screenshots, @howdu.
I do understand your reasoning, but when not inside a fieldset, it would look inconsistent with sibling fieldsets and the regular size section..?

howdu:
With this only affecting the compact styling would that still be an issue? 

Happy to close this if it's adding too much complexity not a biggie.

Thanks for all your amazing work on Filament and considering this change.

zepfietje:
There's nothing holding people from using the normal or compact in whatever situation/combination. So I think this is too opinionated.

Probably best to just add some Tailwind classes in your project for this particular case where it looks better with `lg`.

PR #6166, opened 2023-04-10
Show key-value component delete button for single row

### LINKED ISSUE #6132: KeyValue field with on initial entry - no delete icon
### Package

filament/filament

### Package Version

v2.17.21

### Laravel Version

v10.6.2

### Livewire Version

v2.12.3

### PHP Version

PHP  8.1.2

### Problem description

Hi,
When a basic edit page initially loads, with the record having one entry in the KeyValue field, there is no delete button.

If I press "Add Row" in the KeyValue field, the delete icon appears for the new row and the original row.

If I edit a record with more than one initial entry in the KeyValue field, the delete icon is present for all rows.

https://www.loom.com/share/ce2aa3789e8b47efa66bff5303f601dd

### Expected behavior

Expect a delete icon even on that first row

### Steps to reproduce

Create a keyvalue field with one entry.

Note - the sample repo is not complete, eg, it won't update the model. The metadata value on the model is hard-coded to a one-entry array.

### Reproduction repository

https://github.com/ap1969/filament-keyvalue

### Relevant log output

_No response_

ap1969:
Allow delete icon even if there is only 1 key-value pair.

### DIFF
--- a/packages/forms/resources/views/components/key-value.blade.php
+++ b/packages/forms/resources/views/components/key-value.blade.php
@@ -48,7 +48,7 @@
                         @if (($canDeleteRows() || $isReorderable()) && $isEnabled())
                             <th
                                 scope="col"
-                                x-show="rows.length > 1"
+                                x-show="rows.length > 0"
                                 class="{{ ($canDeleteRows() && $isReorderable()) ? 'w-16' : 'w-12' }}"
                             >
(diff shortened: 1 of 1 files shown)


danharrin:
This didn't work because of the remaining x-show on the button cell... did you test this? Will merge with my fix but please test stuff in the future when submitting a PR

PR #6182, opened 2023-04-12
Improved handling of file-upload-started/finished events for FileUpload

linaspasv:
Before the improvement "Save changes" button was disabled too late and often resulted in user performing save action before file completed uploading.


danharrin:
Thanks!

PR #6191, opened 2023-04-13
Support `HtmlString` in repeater and builder `itemLabel`

bernhardh:
This PR adds an item label to the Builder. Its basically just a copy of the functionality already existing on the repeater and is used in the same way:

```php
Forms\Components\Builder::make("xxx")
    ->itemLabel(fn (array $state): ?string => $state['title'] ?? null)
```

This is very usefull, when sorting long lists, by collapsing:

image


danharrin:
I think `->itemLabel()` should be on the `Block` level instead of the `Builder` level, because then you can have a different label pattern for each block which is good for different data structures.

zepfietje:
Are you able to implement the requested changes, @bernhardh? 🙂

bernhardh:
Uh, sorry, had no time yet. Yes, I think I am able to do it. Will have a look at it, in the afternoon.

zepfietje:
Great, @bernhardh, thanks!

bernhardh:
@zepfietje 
I think I have changed it as suggested. You can now set the itemLabel on `Block`  level instead of globally for the whole `Builder`.

One minor thing I changed compared to the behavior of `Repeater`: You can also use/retour `HtmlString`'s:

```php
->itemLabel(fn (array $state): HtmlString => new HtmlString(
    $state['title'] ?? "<strong class='text-red-500'>TITLE MISSING!!!</strong>"
))
```

packages/forms/resources/views/components/builder.blade.php
```diff
@@ -133,8 +138,6 @@
                                         'dark:text-gray-400' => config('forms.dark_mode'),
                                     ])>
                                         @php
-                                            $block = $item->getParentComponent();
-
                                             $block->labelState($item->getRawState());
```
danharrin, line 141: Wait... I think this feature already exists using the existing `label()` method on the block.

bernhardh: Lol. You are totally right! I was so focused on the Builder level, that I never thought about the block level, because this is how it works on the Repeater. 

So my complete PR is useless. Only missing feature is the HtmlString instead of just plain string. Can/should I close this PR and create a new one to add HtmlString to both - Block label and Repeater itemLabel?

zepfietje: Sure!

danharrin: Or you can use this PR, it doesn't matter

bernhardh:
Closed in favor of #6321

PR #6195, opened 2023-04-13
Support dark mode system setting

bmckay959:
After a user selected "dark" or "light" mode, it was not possible to go back to "system" mode. This tries to fix that.


packages/app/resources/lang/en/layout.php
```diff
@@ -38,6 +38,10 @@
 
         ],
 
+        'system_mode' => [
+            'label' => 'Use system mode',
```
bmckay959: Not a huge fan of the verbiage here.

packages/app/resources/views/components/user-menu.blade.php
```diff
@@ -85,6 +101,10 @@
                 <x-filament::dropdown.list.item icon="heroicon-m-sun" x-show="theme === 'light'" x-on:click="close(); mode = 'manual'; theme = 'dark'">
                     {{ __('filament::layout.buttons.dark_mode.label') }}
                 </x-filament::dropdown.list.item>
+
+                <x-filament::dropdown.list.item icon="heroicon-m-cog" x-show="mode === 'manual'" x-on:click="close(); mode = 'auto';">
```
bmckay959: Not a huge fan of the UI here. Almost should be a dropdown with "Dark Mode", "Light Mode", "System Default" instead? This would also bring clarity to the verbiage and reduce from 2 buttons down to 1 dropdown?

zepfietje: Yeah nested dropdown would be better I think.
Could you take care of that, @bmckay959?
Dropdown component already supports it.

bmckay959: I'm actually not sure of best UI here. Let me mockup a few rough options to discuss.

bmckay959: Here is the current UI as it stands
Screenshot 2023-04-13 at 11 32 18 AM

zepfietje: Here's how the Tailwind CSS website handles it:
Screenshot 2023-04-13 at 19 50 16

If we don't like nesting, we could move it out of the user menu.

danharrin: I think it should remain in the user menu

danharrin: It should be one item that toggles between the three settings

zepfietje: And how would you make that UX clear? 😅

danharrin: Whatever you do, clicking the button should switch theme and sometimes toggle system mode.

If you are currently on dark mode in Filament and dark system, clicking it will set to  light mode.
If you are currently on dark mode in Filament but light system, clicking it will reset to system.
If you are currently on light mode in Filament but dark system, clicking it will reset to system.
If you are currently on light mode in Filament and light system, clicking it will set to dark mode.
If you are currently on light system mode in Filament, clicking it will set to dark mode.
If you are currently on dark system mode in Filament, clicking it will set to light mode.

zepfietje: Hmm, I see how that could work theoretically, but I think having explicit actions results in a better UX...

bmckay959: Here is a concept...could smooth out the design if you like the direction? I tend to agree with Zep on the explicit design here feeling more intuitive, but I'll convert what I currently have into what Dan suggested to give it a shot on how it feels.

Screenshot 2023-04-13 at 2 46 21 PM

zepfietje: Interesting idea, Ben!
I'm worried about mobile/responsiveness with that concept though?

bmckay959: I agree. I have it implemented the way Dan had suggested and clicked around a bit and it feels pretty good.

zepfietje: Really curious what that UX is like, because it still sounds complicated/confusing to me. 😅

bmckay959: I was curious of that too, but essentially, it will always be the opposite of what you currently have.

If you are currently manually toggling themes, the next click makes it auto.
If you are currently on dark, then next click makes it light.

| Current Mode | Current Theme | After Click Mode | After Click Theme |
| -------------- | --------------- | ---------------- | ------------------- |
| Auto | Light | Manual | Dark |
| Manual | Light | Auto | Dark |
| Auto | Dark | Manual | Light |
| Manual | Dark | Auto | Light |

zepfietje: But is it clear that it toggles to auto?

bmckay959: Just implemented this as an option and pushed.
 Only shows the "Follow System Theme" when mode === manual

Screenshot 2023-04-14 at 9 07 14 AM

bmckay959: Ok. So, this was a huge learning curve for me, but I think this is a good spot for you guys to review.

The `toggleTheme()` code was easy and self explainatory.

The issue came when a user was on manual or auto, but the system changed themes(Not based on a user click). This is why we have to listen for any theme/mode or systemMode change to handle appropriately.

I actually am really liking the single toggle idea after implementing.

Let me know your thoughts and what I need to change.

wychoong: > it will always be the opposite of what you currently have.

If it will always be the opposite, how is going back to system matters. It feels redundant

andrewdwallo: I don't think having System inside the option list should matter. The user would know whether Light or Dark is their System setting and whether the screen is currently dark or currently light. If we have the choice to choose between light and dark, then system shouldn't be involved here. Adding "System" to it just makes it more confusing in my opinion even if most companies like Tailwind/GitHub have it as an option.

zepfietje: I disagree. Having the system option allows for automatically matching the system theme when that changes based on time of day. Also good accessibility-wise.

andrewdwallo: I didn't know that your System theme can change automatically on time of day... I think it's fine I was just giving opinion.

bmckay959: I think we need @zepfietje  and @danharrin  to decide and get on the same page for how they want it and then I can make a pull request to match that. I see pros and cons to both directions and don't have a strong preference either way.

zepfietje: It's quite simple, we should just support light/dark/system.
I'm planning to review this PR as is so we can move forward. 👍

bmckay959: Ok. That's not how it is currently working. It is currently just dark/light with auto system toggle if theme != system theme.

I will work on this to get it to the 3 options.

zepfietje: Sorry, I didn't necessarily mean 3 options.
Your current implementation already supports light/dark/system.

No work required on your end now.
Will review first and then we can decide if we like it better with 3 options.

zepfietje: I just tested the PR in its current state and have to admit I find the behavior confusing.

If system is set to light mode, there's no way to fix the Filament theme to light instead of it following system. You'd have to toggle your system theme to dark, then toggle Filament theme to light, then toggle system theme back to light. All those steps are required to make sure that Filament stays in light mode if you manually toggle your system theme (I sometimes switch to dark in the evening) and want Filament to stay light.

In my opinion, we should not try to reinvent theme toggles here and simply stick to what's being used already in other places: dropdown with 3 options.

andrewdwallo: If this is going to be the decision, in my opinion, I think something like what Facebook does would be the best solution. The layout and placement of it is actually kind of similar to what Filament has now. But I think instead of having all of the options in the first initial click of the dropdown menu, something like this would look cool.

https://user-images.githubusercontent.com/104294090/234752993-7e0b23fa-e2fc-4bb6-a6af-53259d447ddb.mp4

andrewdwallo: A Multilevel/Nested Dropdown Menu (or whatever the name of this type of dropdown menu is) would be a nice addition as well for Filament V3 for the User Menu.

zepfietje: > A Multilevel/Nested Dropdown Menu (or whatever the name of this type of dropdown menu is) would be a nice addition as well for Filament V3 for the User Menu.

Yeah that's exactly the solution I would opt for. Little simpler than Facebook, but it does lead to a clearer UX compared to the "magic" system theme toggling with just 1 button.

andrewdwallo: Okay cool! Yeah... the Facebook version would be pretty hard to pull off but would be worth it. Would probably have to require making new dropdown components such as a submenu item with sliding transitions which displays a new dropdown menu list. I couldn't even think of how I would go about doing something like that haha. I'm sure you guys probably might.

zepfietje: I don't think it's needed though. A simple nested dropdown would suffice. 🙂

@bmckay959 @danharrin @awcodes could you share your thoughts?
Then we can move this PR forward.

danharrin: As long as the nested dropdown doesn't look weird that sounds good. It should probably open on the left of the main dropdown.

zepfietje: We could experiment with that.
Nesting without changing placement didn't look too bad when I tried, but positioning it left is very well possible.

awcodes: Just use the floating ui popup. it's already built in and will handle auto positioning based on available space.

zepfietje: The dropdown uses that under the hood. 🙂
I do think we should explicitly set the placement and offset for it to look good.

zepfietje:
@bmckay959, would you be able to rework this to a nested dropdown?
Let me know if you need help with any specifics. 🙂

awcodes:
After giving this some thought about why tailwind does it the way they do and the way it's implemented in the new docs site it makes more senses to have it as a separate component in the topbar. By making it a separate component it takes away any cognitive overhead. If the user selects light mode then it will always be light mode, and the same if they select dark. So selecting system deletes the storage for a theme setting thus defaulting to the users system preferences determined by the browser. I feel that treating it any other way with complicated logic and a drop-down menu really obfuscates the point and leads to more difficult maintenance.

zepfietje:
Nested dropdown or separate icon in the topbar would essentially be the same, right?
I think we're actually talking about the exact same implementation, @awcodes.

We can then decide to have it inside the user dropdown or outside, but that should be a matter of moving the component if we extract it.

andrewdwallo:
Yeah, tailwind choosing to put it in a different place is just a matter of design preference. I don't think that means it is better. Many more companies other than tailwind choose to not put it in the topbar and put it somewhere else. It would make since for tailwind because their website is essentially only for documentation. It isn't like a social media website or something like GitHub.

awcodes:
I actually implemented this as a panel/dropdown today and I honestly didn't like it. The panels were hard to see and got lost because they blended into the UI as a whole. Maybe the topbar drop downs needs a better visual separation in general.

bmckay959:
Ok. I didn't see @awcodes comment before I did this.

Honestly guys, I may have reached my "I'm new and don't know what I'm doing, wall". I think I have it working with the exception of a 'active' class on the theme that is currently selected.

I'm sure whatever Adam built is way more graceful than what I built, so no pride in ownership here if there is a better way.

bmckay959:
> I actually implemented this as a panel/dropdown today and I honestly didn't like it. The panels were hard to see and got lost because they blended into the UI as a whole. Maybe the topbar drop downs needs a better visual separation in general.

I agree on the separation. Even trying the dropdown.list.seperater doesn't look quite right because the `<hr>` doesn't go all the way across.

andrewdwallo:
In my opinion, the only way for this to look good is to make something similar to what Facebook did, which is a multilevel dropdown menu with sliding transitions, because a regular nested dropdown menu creates too tight of a space for the content to look separated, or just separate it from the dropdown menu which is what @awcodes is suggesting. Having it open to the left of the main dropdown menu doesn't really look modern or attractive in my opinion either, but it isn't up to me. I am just giving my perspective and trying to be helpful.

zepfietje:
> I agree on the separation. Even trying the dropdown.list.seperater doesn't look quite right because the `<hr>` doesn't go all the way across.

You don't need the separator; just use the dropdown list component and the groups will automatically be visually separated.

awcodes:
I'm so fascinated that a ternary option has had this much discussion. Lol.

howdu:
Another option not sure if you've discussed is adding it to the nav bottom/left. 
Screenshot 2023-05-08 at 12 54 49
When the menu's collapsed it could just show the selected icon.

awcodes:
Here's a crazy idea to simplify it. 🤷‍♂️

Screen Shot 2023-05-16 at 4 07 06 PM

zepfietje:
> Here's a crazy idea to simplify it. 🤷‍♂️
> 
> Screen Shot 2023-05-16 at 4 07 06 PM

Looks nice! I'm worried about non-technical users though...

awcodes:
Valid point, but I wonder how many non technical users wouldn't actually understand it these days.

zepfietje:
> Valid point, but I wonder how many non technical users wouldn't actually understand it these days.

Trust me: a lot.
I've had super obvious UI/UX in my projects that users simply didn't understand (certain generation mostly).

danharrin:
Tooltips?

awcodes:
Tooltips would be easy. Just don't really work on mobile.

danharrin:
I don't think it matters on mobile

awcodes:
So, do we want to go this direction?

danharrin:
Yeah! I think so

danharrin:
Thanks @bmckay959 @awcodes @zepfietje!

PR #6196, opened 2023-04-13
Add `Filament::isServingFilament()`

ralphjsmit:
This PR adds a method to the facade that will return a boolean based on whether the `ServingFilament` event was dispatched earlier in the current lifecycle, meaning that it will track whether we are on a Filament request or not.

```php
if (Filament::isServingFilament()) {
   //
}
```

It could be that the name `Filament::servingFilament(true)` is a bit confusing. In that case we could rename the methods to `setIsServingFilament()`/`isServingFilament()`.


packages/admin/src/FilamentManager.php
```diff
@@ -57,6 +57,8 @@ class FilamentManager
 
     protected array $renderHooks = [];
 
+    protected bool $servingFilament = false;
```
danharrin:
```suggestion
    protected bool $isServing = false;
```

packages/admin/src/FilamentManager.php
```diff
@@ -166,6 +168,11 @@ public function pushMeta(array $meta): void
         $this->meta = array_merge($this->meta, $meta);
     }
 
+    public function servingFilament(bool $servingFilament = true): void
```
danharrin:
```suggestion
    public function setServingStatus(bool $condition = true): void
```

packages/admin/src/FilamentManager.php
```diff
@@ -166,6 +168,11 @@ public function pushMeta(array $meta): void
...
+    {
+        $this->servingFilament = $servingFilament;
```
danharrin:
```suggestion
        $this->isServing = $condition;
```

packages/admin/src/FilamentManager.php
```diff
@@ -413,4 +420,9 @@ public function getMeta(): array
     {
         return array_unique($this->meta);
     }
+
+    public function isServingFilament(): bool
+    {
+        return $this->servingFilament;
```
danharrin:
```suggestion
        return $this->isServing;
```

packages/admin/src/FilamentManager.php
```diff
@@ -413,4 +420,9 @@ public function getMeta(): array
     {
         return array_unique($this->meta);
     }
+
+    public function isServingFilament(): bool
```
danharrin:
```suggestion
    public function isServing(): bool
```

packages/admin/src/FilamentServiceProvider.php
```diff
@@ -110,6 +110,10 @@ public function packageBooted(): void
             }
         }
 
+        Filament::serving(function () {
+            Filament::servingFilament();
```
danharrin:
```suggestion
            Filament::setServingStatus();
```

packages/admin/src/Facades/Filament.php
```diff
@@ -30,6 +30,7 @@
  * @method static array getUserMenuItems()
  * @method static string getUserName(Model | Authenticatable $user)
  * @method static array getWidgets()
+ * @method static bool isServingFilament()
```
danharrin:
```suggestion
 * @method static bool isServing()
```

packages/admin/src/Facades/Filament.php
```diff
@@ -47,6 +48,7 @@
  * @method static void registerWidgets(array $widgets)
  * @method static Htmlable renderHook(string $name)
  * @method static void serving(Closure $callback)
+ * @method static void servingFilament(bool $servingFilament = true)
```
danharrin:
```suggestion
 * @method static void setServingStatus(bool $condition = true)
```

danharrin:
Thanks

PR #6205, opened 2023-04-14
Allow usage of table alias in reorderable

linaspasv:
There is sometimes a need to use table alias in ``SELECT`` queries to overcome ambiguous column error when you have a more complex join query in your ``getTableQuery`` and both of the tables got a sort column with the same name. While for ``SELECT`` it perfectly worked before but unfortunately during the update Eloquent does not allow to fill with table alias. So PR takes care of this.

After the change it would be possible to clearly define table for the orderBy column and perform update on reordering action, e.g.

```php
    {
           ...
            ->actions([
                Tables\Actions\RestoreAction::make(),
                Tables\Actions\DeleteAction::make(),
            ])
            ->reorderable('products.order_to_display')
            ->defaultSort('products.order_to_display');
    }
```


danharrin:
Thanks

PR #6208, opened 2023-04-14
Optimisation for reorderTable action

linaspasv:
Reordering on larger tables are quite slow because we are iterating over each record and updating them individually. My PR solves the performance issues by introducing the CASE statement to update multiple records in a single query.

I see around 6-7x speed improvement for reorderable tables with 30 records, so there should be greater improvements with even larger tables. At the moment filamentphp executes around 60 queries (SELECT, UPDATE) for 30 records, and after the PR it should be reduced to a single query for the update.

### DIFF
--- a/packages/tables/src/Concerns/CanReorderRecords.php
+++ b/packages/tables/src/Concerns/CanReorderRecords.php
@@ -4,7 +4,6 @@ namespace Filament\Tables\Concerns;
 
 use Filament\Tables\Contracts\HasRelationshipTable;
 use Illuminate\Database\Eloquent\Relations\BelongsToMany;
-use DB;
 
 trait CanReorderRecords
 {
--- a/packages/tables/src/Concerns/CanReorderRecords.php
+++ b/packages/tables/src/Concerns/CanReorderRecords.php
@@ -4,6 +4,7 @@ namespace Filament\Tables\Concerns;
 
 use Filament\Tables\Contracts\HasRelationshipTable;
 use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+use DB;
 
 trait CanReorderRecords
 {
--- a/packages/tables/src/Concerns/CanReorderRecords.php
+++ b/packages/tables/src/Concerns/CanReorderRecords.php
@@ -31,11 +31,16 @@ trait CanReorderRecords
             return;
         }
 
-        foreach ($order as $index => $recordKey) {
-            $this->getTableRecord($recordKey)->update([
-                $orderColumn => $index + 1,
+        $model = new ($this->getTableModel());
+        $model
+            ->newModelQuery()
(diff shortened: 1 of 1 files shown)


ArtMin96:
Thanks for this. Here is more [optimization](https://www.yellowduck.be/posts/optimising-large-wherein-queries-in-laravel)
```php
->whereIntegerInRaw($model->getKeyName(), array_values($order))
```

linaspasv:
> Thanks for this. Here is more [optimization](https://www.yellowduck.be/posts/optimising-large-wherein-queries-in-laravel)
> 
> ```
> ->whereIntegerInRaw($model->getKeyName(), array_values($order))
> ```

Yeh, I guess it might work but we should not forget that some people might use UUID primaryKeys since Laravel 9. Also, we are talking about sorting hundreds or thousands of records to start seeing some performance improvements. I am for further optimization but I guess maintainers should make a call regards this. :-)

danharrin:
I don't understand what the advantage of only handling integers is... why would there be a Laravel method to do that but no generic `whereInRaw()`? weird...

linaspasv:
@danharrin by the way, I think we could also apply similar optimisation for the case with relationship just I could not recreate a test case, so just handled the regular one for now.

danharrin:
That's ok! No rush, just submit a PR whenever you want :)

wychoong:
This changes won’t emit model events during update right?

danharrin:
Thats probably a good thing though, there could be lots of models

linaspasv:
> This changes won’t emit model events during update right?

You are right, no model events are triggered during the process, otherwise it would be another performance concern. If you need to perform actions before / after sort event, then it would be best to have dedicated events for this.

wychoong:
Just highlighting there might be user depending on that event as this changes the behavior

linaspasv:
> 

In this case it would be better to release as a {minor} instead of {patch}, hopefully not many are affected. In general it's a bad practice to trigger observer events during this intensive operation but everyone got their own reasoning. :-)

danharrin:
Yeah there is a small chance that this could affect someone, but I would question the motivation for depending on a model event during a mass update, especially when it probably isn't useful anyway until all models have their new order assigned.

wychoong:
Perhaps for logging of record values changes. But I’m definitely go for the optimized way

linaspasv:
> 

Do we have any lifecycle hooks when the reordering operation has started / completed? Maybe it would be worth to add some documentation entries in "listing-records" to address this particular thing, in case, someone needs to perform any operation before or after the reordering?

danharrin:
We don't have any hooks for this at the moment

chosten:
This cause an issue for people using Laravel Scout. An "after reordering" hook would be really helpful.

PR #6216, opened 2023-04-15
Optimisation for file upload form component in spatie media library

linaspasv:
Right now when working with multiple files on a form there are excessive ``SELECT`` queries going into the database, e.g. if we have 10 files associated with a record it results into 11 ``SELECT`` queries (1 to load all the media data, and 10 queries for each file individually to fetch the same data), after PR it will load all the media once and fetch the rest needed data from the memory. 

There is no need to individually query database for each file because we already got all the data loaded in memory.


danharrin:
Thanks

PR #6217, opened 2023-04-15
Proposal: let columns apply table filters

josefbehr:
Since this trick (https://filamentphp.com/tricks/apply-filter-by-clicking-on-value-in-column) doesn't work due to escaping of the `extraAttributes` I wanted to propose another in-built solution, allowing to apply filters by clicking the column values.

I implemented the concern and applied it to the text and tags columns for demonstration.

With `filterName` on such a column the name of the filter that the value should be applied to can be specified (assuming single values for text and multiple for tags by default). If that's not enough, a `filterFunctionName` can be supplied, giving the name of a public component function that will be called on click instead.

Let me know if you think that is something that you'd consider adding the core, and if so, if you'd like me to change anything.

### DIFF
--- a/packages/tables/resources/views/columns/tags-column.blade.php
+++ b/packages/tables/resources/views/columns/tags-column.blade.php
@@ -29,9 +29,9 @@
             @if ($hasFilter)
                 @if (! blank($filterName))
                     wire:click="$set('tableFilters.tags.values', [ '{{ $tag }}' ])"
-            @endif
-            @if (! blank($filterFunctionName))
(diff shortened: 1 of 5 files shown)


danharrin:
I don't understand what the advantage of only handling integers is... why would there be a Laravel method to do that but no generic `whereInRaw()`? weird...

danharrin:
Whoops, just accidentally merged as I had 2 tabs open 😭 Will have to review this on the branch lol

danharrin:
See #6219

PR #6219, opened 2023-04-15
Table column filtering

danharrin:
See #6217

### DIFF
--- a/packages/tables/resources/views/columns/tags-column.blade.php
+++ b/packages/tables/resources/views/columns/tags-column.blade.php
@@ -1,5 +1,9 @@
 @php
     $state = $getTags();
+
+    $hasFilter = $hasFilter();
+    $filterName = $getFilterName();
+    $filterFunctionName = $getFilterFunctionName();
 @endphp
 
 <div {{ $attributes->merge($getExtraAttributes())->class([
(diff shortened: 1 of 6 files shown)


danharrin:
@josefbehr my problems with this is that it will only work with SelectFilter, as it assumes that a `value` or `values` field exists within the filter, which not all filters have. Also, there's no way to toggle it between `SelectFilter` `multiple()` or not. I don't think you should have to configure this with an extra method, it should just work. Do you have a solid solution for that, because I don't...

josefbehr:
@danharrin Yeah, I see the same problem. After some consideration I currently think the best option might be to just abstract this away into a way to call an arbitrary method on the component, as I added with the `filterFunctionName`. Renaming this to something more general would allow us to add a `wire:click` directive to columns, enabling devs to add whatever functionality they like to the click action, including applying filters without having to configure it or presuming a specific kind of filter. What do you think?

danharrin:
Well we do have `action()` functionality at the moment for columns which is similar... and you can use `extraAttributes()` to achieve something similar if you don't want a full action object. Will close this PR as other solutions would be completely different to the code implemented here. I'm not sure if it's worth adding any more helpers to achieve this, I don't think I'm convinced by current suggestions.

josefbehr:
Unfortunately, using `extraAttributes()` doesn't work, as the arguments get escaped. That was my inital reason to create this PR (I wrote as much in #6217). As for `action()`, that would seem like it could help to achieve this, but at least in my trials calling a method that sets a `$tableFilter` has no effect at all. Might be I'm missing something, will ask in the discord.

PR #6221, opened 2023-04-15
Fix all toggle column issues

### LINKED ISSUE #6215: Toggle column issues
### Package

filament/filament

### Package Version

v2

### Laravel Version

v9

### Livewire Version

v2

### PHP Version

8.1

### Problem description

Since I mentioned some bugs with the toggle columns in PR #6198 I thought it'd be helpful to have a minimal demo repo to make it easy to see and discuss the issues. Some of the issues that have been mentioned here and on discord:

1. Console error `cannot read properties of undefined (reading 'value')` (addresssed with PR #6198)
2. Console error  `cannot read properties of undefined (reading 'dispatchEvent') (address with PR #6198)
3. Only the first toggle works after filtering the page with the same type of column
4. Infinite loop on toggles on slow connections
5. Switching 1 toggle actually causes a different toggle to switch
6. Other issues??

### Expected behavior

Toggle columns should work

### Steps to reproduce

Clone the repo

```bash
composer install
```

Edit .env file as needed and then

```bash
php artisan key:generate
php artisan migrate:fresh --seed
```

Then head to the `User` resource.

## Steps to reproduce

After each test be sure to refresh your browser to clear out any caches and start with all the toggles **active**. This will give everyone the same starting point.

**Issue 1**
1. Open your console
2. Click on the `next page` button
3. Click on the `previous page` button
4. See the console error

**Issue 2**
1. Open your console
2. Open filters and select `Yes` from `Active`
3. Click on the first toggle button to make it `inactive` and thus remove it from the view.
4. See the console error

**Issue 3**
1. Open filters and select `Yes` from `Active`
2. Only the first toggle can be toggled
3. All other toggles cannot be turned off anymore. Note that two queries are being made. One to turn it off and one to turn it on.
4. If you refresh the page everything works again.

**Issue 4**
1. Turn your dev tools throttling to `fast 3g`
2. Refresh to page to load under `fast 3g`
3. Go down the list and toggle each of the toggles. 
4. Usually by the 5th or 6th toggle, the toggles will enter an infinite loop, each loop sending queries to the DB

**Issue 5**
I have seen this, but not sure how to reproduce at the moment

Hope this helps to get discussion going on how to fix this.

Thanks!

### Reproduction repository

https://github.com/archilex/filament-toggle-column-demo

### Relevant log output

_No response_

archilex:
I think I got it!

Like I had mentioned previously I think alpine's init watcher was causing the issue so I rewrote the code without using it. This version solves ALL the problems listed in Issue #6215!

1. No more console logs (dont even need the previous bug fix code.
2. Pagination works fine
3. Filtering and then toggling works fine, without multiple duplicate queries
4. No funny loop conditions. (If you toggle them quickly you can see them disable as they await the response). 

This also maintains all the same abilities as the previous toggle (I think):
1. The button will be disabled as the request is processed
5. If there is a validation error the tool tip will display the error and will not change states

However, I am unsure as to why there was a hidden input that was saving the state of the toggle. Was this in response to a different problem you were having with the toggle? If so, please check this code against whatever problem that was.

And let me know what stying changes you have.

Glad to have (hopefully) gotten this fixed!

### DIFF
--- a/packages/tables/resources/views/columns/toggle-column.blade.php
+++ b/packages/tables/resources/views/columns/toggle-column.blade.php
@@ -2,116 +2,116 @@
     $state = $getState();
 @endphp
 
-<div wire:key="{{ 'toggle-column-' . $recordKey . '-' . $getName() . '-' . json_encode($state) }}">
-<div
-    x-data="{
-        error: undefined,
-        state: @js((bool) $state),
-        isLoading: false,
-    }"
-    {{ $attributes->merge($getExtraAttributes())->class([
-        'filament-tables-toggle-column',
-    ]) }}
-    wire:ignore
->
-    <button
-        role="switch"
-        aria-checked="false"
-        x-bind:aria-checked="state.toString()"
-        x-on:click="
-            if (isLoading) return
-
-            isLoading = true
-            response = await $wire.updateTableColumnState(@js($getName()), @js($recordKey), ! state)
-            error = response?.error ?? undefined
-
-            if (error) {
-                return isLoading = false
-            }
-
-            state = ! state
-            isLoading = false
-        "
-        x-tooltip="error"
-        x-bind:class="{
-            'opacity-70 pointer-events-none': isLoading,
-            '{{ match ($getOnColor()) {
-                'danger' => 'bg-danger-500',
-                'secondary' => 'bg-gray-500',
-                'success' => 'bg-success-500',
-                'warning' => 'bg-warning-500',
-                default => 'bg-primary-600',
-            } }}': state,
-            '{{ match ($getOffColor()) {
-                'danger' => 'bg-danger-500',
-                'primary' => 'bg-primary-500',
-                'success' => 'bg-success-500',
-                'warning' => 'bg-warning-500',
-                default => 'bg-gray-200',
-            } }} @if (config('forms.dark_mode')) dark:bg-white/10 @endif': ! state,
+<div wire:key="{{ $this->id }}.table.record.{{ $recordKey }}.column.{{ $getName() }}.toggle-column.{{ json_encode($state) }}">
+    <div
+        x-data="{
+            error: undefined,
+            state: @js((bool) $state),
+            isLoading: false,
         }"
-        {!! $isDisabled() ? 'disabled' : null !!}
-        type="button"
-        class="relative inline-flex shrink-0 ml-4 h-6 w-11 border-2 border-transparent rounded-full cursor-pointer transition-colors ease-in-out duration-200 outline-none focus:ring-1 focus:ring-offset-1 focus:ring-primary-500 disabled:opacity-70 disabled:cursor-not-allowed disabled:pointer-events-none"
+        {{ $attributes->merge($getExtraAttributes())->class([
+            'filament-tables-toggle-column',
+        ]) }}
+        wire:ignore
     >
-        <span
-            class="pointer-events-none relative inline-block h-5 w-5 rounded-full bg-white shadow transform ring-0 ease-in-out transition duration-200"
+        <button
+            role="switch"
+            aria-checked="false"
+            x-bind:aria-checked="state.toString()"
+            x-on:click="
+                if (isLoading) return
+
+                isLoading = true
+                response = await $wire.updateTableColumnState(@js($getName()), @js($recordKey), ! state)
+                error = response?.error ?? undefined
+
+                if (error) {
+                    return isLoading = false
+                }
+
+                state = ! state
+                isLoading = false
+            "
+            x-tooltip="error"
             x-bind:class="{
-                'translate-x-5 rtl:-translate-x-5': state,
-                'translate-x-0': ! state,
+                'opacity-70 pointer-events-none': isLoading,
+                '{{ match ($getOnColor()) {
+                    'danger' => 'bg-danger-500',
+                    'secondary' => 'bg-gray-500',
+                    'success' => 'bg-success-500',
+                    'warning' => 'bg-warning-500',
+                    default => 'bg-primary-600',
+                } }}': state,
+                '{{ match ($getOffColor()) {
+                    'danger' => 'bg-danger-500',
+                    'primary' => 'bg-primary-500',
+                    'success' => 'bg-success-500',
+                    'warning' => 'bg-warning-500',
+                    default => 'bg-gray-200',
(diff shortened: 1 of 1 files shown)


awcodes:
Looks ok to me, I’d be curious though how it is handled by a screen reader. I’m sure it’s fine, just a hunch I have.

archilex:
@awcodes Ahhhh…that’s probably why there’s a hidden input isn’t it?! Makes sense. I feel we could add it back in and just set it in the same “onClick” method, right. I’ll try later.

awcodes:
I'm not sure it is a problem. Just a feeling. Seems like the right aria attributes are there. Would just want to physically test it first.

danharrin:
The reason why it was implemented with the hidden field is because if someone updates the property in an action modal, the table should update. I believe this is now broken again because of the PR.

danharrin:
For example, `is_enabled` is a toggle column on the table, and there is also a modal EditAction that can update `is_enabled`. When enabling inside the modal, the change should reflect to the column.

archilex:
Ok, I’ll check and report back

archilex:
Just a quick update. I did some quick testing and @danharrin you are right that this does not work with edit actions. Thanks for the heads up there. Thankfully, making it work with edit actions is easy, and most the other bugs continue to be solved as well. The only problem I’m having is when filtering and then toggling, it’s back to firing duplicate queries (as it also does currently). If I can solve that then I think we’ll be good to go. I’m traveling so it might be a few days until I can continue work on this, but I’m determined!

archilex:
Alright, I think I got it!! Long story short, basically all that needed to be done was to have Livewire completely rerender the entire toggle component on every state change. At the end of the day, it's a simple fix, but this took A LOT of rewrites to figure out.

In addition to solving the problems mentioned in the #6215, it also successfully changes state when edited by an action like Dan mentioned, again because Livewire is just swapping out the entire component.

The ONLY issue that this solution has is that when toggling from `on` to `off`, we lose the nice transition because the default state of the icons is off so when Livewire rerenders the component it's already `off`. The transition does occur when it's toggled on. I think losing the toggle off transition is well worth it. 

I'll also note that my PR doesn't have correct indenting on the new root `<div>`. This is because if you indent correctly it really messes up with the diffing and it seems like A LOT was changed. Once you approve this solution I can add another commit that indents properly.

archilex:
One more commit because I realized I needed to add the column name to `wire:key` to distinguish between multiple toggleColumns on the same record.

packages/tables/resources/views/columns/toggle-column.blade.php
```diff
@@ -2,53 +2,34 @@
     $state = $getState();
 @endphp
 
+<div wire:key="{{ 'toggle-column-' . $recordKey . '-' . $getName() . '-' . json_encode($state) }}">
```
zepfietje: Could you fix indentation, @archilex?

zepfietje (changes requested):

packages/tables/resources/views/columns/toggle-column.blade.php
archilex: Sure! Done. I also reformatted the`wire:key` to better follow Filament's existing style

zepfietje: Thanks!

packages/tables/resources/views/columns/toggle-column.blade.php
```diff
@@ -2,135 +2,116 @@
...
-            if (component.component.id !== @js($this->id)) {
-                return
-            }
-
-            if (! $refs.newState) {
-                return
-            }
-
-            let newState = $refs.newState.value === '1' ? true : false
+<div wire:key="{{ $this->id }}.table.record.{{ $recordKey }}.column.{{ $getName() }}.toggle-column.{{ json_encode($state) }}">
```
zepfietje: Not sure about having the state in the key.

archilex: You need the state in the key so that when it changes on the edit action it will rerender the toggle component.

archilex: Well, it rerenders each time it's toggled of course, but this allows it to work with the edit action, allowing us to get rid of `init` and `$watch` and the hidden input which was causing the issues. 

And that combined with the `wire:ignore` on the child div prevents the duplicate queries when filtering.

archilex: IMO this is a good fix for v2 since it solves all the current issues with toggle columns, however, I think this can be revisted again when livewire v3 ships. I still think there's some bug with alpine/livewire with the combination of `init` and `$watch` that's causing these dom diffing problems. Maybe those will be fixed with v3 and we can revisit and get the transition working again when toggling to `off`.

zepfietje: But the `wire:key` isn't unique if multiple rows have the same state value, right?

archilex: It is unique because of the $recordKey. Which is why I also added the $getName() incase a row might have multiple toggle columns.

archilex: I just pushed an update to the repo for #6215 so you can test this out. You'll have to copy and paste this code into that repo to try it out of course. 
The repo includes an edit action for the Resource so you can test that. It also has a rule that will fail validation commented out that you can uncomment to see that as well.

archilex: Pushed another update to repo for #6215 that overrides the toggle-column with this new code so you don't have to copy and paste anymore. Hope that helps 👍

zepfietje: Do we even need the state in the wire:key if it contains the record and column keys/names already?

archilex: Yes, you need the state in the key so that when the state changes on toggle it will rerender the toggle component. This allows us to get rid off all init and $watch and hidden input which were causing the issues.

danharrin:
This works great! However, we appear to have lost all transitions on the toggle itself...

danharrin:
Okay, think I've fixed it. Thank you

archilex:
Awesome! I had transitions working when turning on, just not off (I mentioned this in the comments with Zep above). Did your fix also fix the off transition? If so that’s great!

danharrin:
I didn't have transitions for either state. That is fixed now

PR #6228, opened 2023-04-16
Upgrade Tailwind CSS v3.3

datlechin:
This PR is upgrade to Tailwind CSS v3.3.

What changes?
- Upgrade `@tailwindcss/forms`, `@tailwindcss/typography`, `prettier-plugin-tailwindcss` and `tailwindcss` to the latest version.
- Use css variable without the `var()` keyword
- Change `ml-*` to `ms-*, `mr-*` to `me-*`, ... to make RTL simpler. Read more about this (https://tailwindcss.com/blog/tailwindcss-v3-3#simplified-rtl-support-with-logical-properties)


packages/admin/resources/views/components/user-avatar.blade.php
```diff
@@ -3,7 +3,7 @@
 ])
 
 <div
-    {{ $attributes->class([
+ l   {{ $attributes->class([
```
zepfietje:
```suggestion
    {{ $attributes->class([
```

dat-archielite: thx :sweat_smile:

zepfietje:
Thanks for the effort, @datlechin!
I really like these changes, though we probably can't merge this as it would be a breaking change.

danharrin:
Yeah it would break v2 for people outside the admin panel or with custom themes. So v3 only 👍

Sorry about that

datlechin:
@danharrin should it be created in `3.x` branch?

danharrin:
If you want to do that work, go for it. But if you don't want to, we will do it before v3 release anyway :)

mstfkhazaal:
> If you want to do that work, go for it. But if you don't want to, we will do it before v3 release anyway :)

I hope to add it as soon as possible for version 2

PR #6236, opened 2023-04-17
Support 950 color shade

zepfietje:
Not sure if we should require Tailwind CSS v3.3 for this.


danharrin:
We'll be upgrading to 3.3 anyway at some point before release

PR #6242, opened 2023-04-17
Map uuid to id from already loaded relationship in spatie media library

linaspasv:
When reorder save action happens there is an unnecessary database call executed to match uuid with id. There is no need to query database again because we already got all the data loaded in memory. Similar approach we did for #6216

### DIFF
--- a/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
+++ b/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
@@ -129,7 +129,11 @@ class SpatieMediaLibraryFileUpload extends FileUpload
 
             $mediaClass = config('media-library.media_model', Media::class);
 
-            $mappedIds = $mediaClass::query()->whereIn('uuid', $uuids)->pluck('id', 'uuid')->toArray();
+            $mappedIds = $component->getRecord()
+                ->getRelation('media')
+                ->whereIn('uuid', $uuids)
+                ->pluck('id', 'uuid')
+                ->toArray();
 
             $mediaClass::setNewOrder(array_merge(array_flip($uuids), $mappedIds));
 
--- a/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
+++ b/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
@@ -130,10 +130,10 @@ class SpatieMediaLibraryFileUpload extends FileUpload
             $mediaClass = config('media-library.media_model', Media::class);
 
             $mappedIds = $component->getRecord()
-                ->getRelation('media')
+                ->getRelationValue('media')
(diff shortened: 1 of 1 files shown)


ariaieboy:
This PR broked my application tests the error I get is 
```bash
Call to a member function determineOrderColumnName() on null

  at vendor/spatie/laravel-medialibrary/src/MediaCollections/Models/Concerns/IsSorted.php:39
     35▕     {
```
I've downgraded my filament `"filament/spatie-laravel-media-library-plugin": "2.17.26"` and my problem fixed.
cc: @danharrin

danharrin:
Okay will revert his @ariaieboy

linaspasv:
> ```shell
> determineOrderColumnName
> ```

Any idea why it causes you error from spatie/media-library? The outputs are identical before / after PR.

ariaieboy:
> > ```shell
> > determineOrderColumnName
> > ```
> 
> Any idea why it causes you error from spatie/media-library? The outputs are identical before / after PR.

the ordering is the problem if I remove `->orderable()` from the input the error goes away.
I've tested the version that does not contain this PR and everything was working ok.
The problem here is that you pass null as ID somehow.

linaspasv:
> > > ```shell
> > > determineOrderColumnName
> > > ```
> > 
> > 
> > Any idea why it causes you error from spatie/media-library? The outputs are identical before / after PR.
> 
> the ordering is the problem if I remove `->orderable()` from the input the error goes away. I've tested the version that does not contain this PR and everything was working ok. The problem here is that you pass null as ID somehow.

I wonder how your setup is different because it worked perfectly with orderable for me.

ariaieboy:
> > > > ```shell
> > > > determineOrderColumnName
> > > > ```
> > > 
> > > 
> > > Any idea why it causes you error from spatie/media-library? The outputs are identical before / after PR.
> > 
> > 
> > the ordering is the problem if I remove `->orderable()` from the input the error goes away. I've tested the version that does not contain this PR and everything was working ok. The problem here is that you pass null as ID somehow.
> 
> I wonder how your setup is different because it worked perfectly with orderable for me.

I don't know why or how but the lazyloaded relation returns an empty array while if you query the DB the relation exists it's either a bug in the relation resolver of filament or there is something wrong in my test setup

linaspasv:
@ariaieboy is there a way to share part of your test so it would be possible to recreate? I would love to test and see. There are other parts were we also did lazy loaded relations to optimize the orderable.

ariaieboy:
@linaspasv let me try to make a repo for this if i could recreate the situation

linaspasv:
> @linaspasv let me try to make a repo for this if i could recreate the situation
> 

Thanks

ariaieboy:
> > @linaspasv let me try to make a repo for this if i could recreate the situation
> 
> Thanks

Ok I've created a repo containing the problematic codes.
https://github.com/ariaieboy/issue-filament-reordering
clone the repo go to laravel folder.
install composer deps
if you have docker you can use sail right away by using `sail up`
then create a filament user using `make:filament-user` command and loginto filament and edit the user and add some pictures you will encounter the error.
image

PR #6251, opened 2023-04-18
Keeping a clean query string for pages with relation manager

linaspasv:
There is no need for activeRelationManager query string when displaying a single relation or the first one.


danharrin:
`0` isn't always the array key for the first tab. It needs to be retrieved from the relation managers array (it's the first array key). To do this, you will need to replace `$queryString` with `queryString()` so it can be dynamic.

zepfietje:
Are you able to handle this, @linaspasv?

linaspasv:
> 

I was very busy lately but will get back to this once got some free time. This could probably be closed for time being and I would open a new one once ready.

danharrin:
Alright, thanks for letting us know!

PR #6257, opened 2023-04-19
Revert "Map uuid to id from already loaded relationship in spatie media library"

danharrin:
Reverts filamentphp/filament#6242


ariaieboy:
@danharrin can you please tag this?

PR #6262, opened 2023-04-19
Document undocumented admin test helpers

JHWelch:
This PR adds all the remaining test helpers from `packages/admin/.stubs.php` to the documentation.

In the Stub file, I moved the two label assertions next to teach other like all the other paired assertions.

Will PR the other sections separately.


JHWelch:
Added another PR for Table assertions #6263

danharrin:
Thanks!

PR #6263, opened 2023-04-19
Document undocumented table test helpers

JHWelch:
This PR adds all the remaining test helpers from `packages/tables/.stubs.php` to the documentation.

In the Stub file, I moved the two label assertions next to teach other like all the other paired assertions.

This follows up on the work in #6262


danharrin:
Thank you!

PR #6265, opened 2023-04-19
Support extraAlpineAttributes on non-masked text input

leandrocfe:
Please note that the extraAlpineAttributes option is only available when using the mask option. However, with this feature, the position can be changed, allowing it to be used without a mask as well.

### DIFF
--- a/packages/forms/resources/views/components/text-input.blade.php
+++ b/packages/forms/resources/views/components/text-input.blade.php
@@ -52,7 +52,6 @@
                     {!! $isLazy() ? "x-on:blur=\"\$wire.\$refresh\"" : null !!}
                     {!! $isDebounced() ? "x-on:input.debounce.{$getDebounce()}=\"\$wire.\$refresh\"" : null !!}
                 @endunless
-                {{ $getExtraAlpineAttributeBag() }}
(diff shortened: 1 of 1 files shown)


zepfietje:
Thanks, @leandrocfe!
Doesn't it make sense to move it right above the line containing `$getExtraInputAttributeBag()` though?

leandrocfe:
> Thanks, @leandrocfe!
> Doesn't it make sense to move it right above the line containing `$getExtraInputAttributeBag()` though?

Hi Zep! I agree :)

danharrin:
Text inputs without a mask arent Alpine components, so why would we merge attributes?

There are other methods like extraInputAttributes() that are always applied

danharrin:
Hmm I've just seen that they are actually Alpine components because of validation, sorry for that confusion 🤔

I agree with Zep about positioning

zepfietje:
Mark as ready when you've applied the requested change, @leandrocfe! 🙂

leandrocfe:
> Mark as ready when you've applied the requested change, @leandrocfe! slightly_smiling_face

Hi @zepfietje, fixed.
Sorry for the delay.

zepfietje (approved):

PR #6269, opened 2023-04-20
Add classes to wizard component

arjendejong12:
I've had the need to target specific HTML elements in the wizard component, so I've added some filament classes to several elements. I'm not too sure about the class names though. Any suggestions for different names are welcome.


danharrin:
Thanks

PR #6278, opened 2023-04-21
Use Laravel's generated column key name instead of relationship qualified key name 

tommie1001:
When a model has an relation to itself the query wont build correctly, it now uses the qualifiedKeyName instead of the then generated laravel keyname of the query. 

You can easily reproduce this by creating an model that has an pivot with an relation to itself, then specifing the relationship and the inverse in the model, as in the relationmanager, and when creating the query it will use the keyname of the parent relationship instead of the generated `laravel_reserved_0` one

Becuase of it not using the laravel_reserved_0 column, it will return models that already have an relationship to the model and by that it produces an error when using unique on the pivot

### DIFF
--- a/packages/tables/src/Actions/AssociateAction.php
+++ b/packages/tables/src/Actions/AssociateAction.php
@@ -213,7 +213,7 @@ class AssociateAction extends Action
                             ->where($relationship->getQualifiedForeignKeyName(), $relationship->getParent()->getKey());
                     }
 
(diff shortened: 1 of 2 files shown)


danharrin:
Thanks. Is this only appropriate for `AttachAction` and not `AssociateAction`?

tommie1001:
Hi @danharrin ,

I have checked it and it also applies for the AssociateAction, for wich I also made an fix in this pr

danharrin:
Thanks!

PR #6280, opened 2023-04-21
Improve auth pages extensibility

zepfietje:


tests/src/App/Auth/LoginTest.php
```diff
@@ -82,21 +92,21 @@
 
 it('can validate `email` is required', function () {
     livewire(Login::class)
-        ->set('email', '')
+        ->fillForm(['email' => ''])
         ->call('authenticate')
         ->assertHasErrors(['email' => ['required']]);
```
danharrin: Tests are failing because these need to be `assertFormHasErrors()`. The Livewire helper doesn't account for the state path.

zepfietje: Of course, thanks.

PR #6284, opened 2023-04-21
Fixes markdown editor character overlapping in RTL

### LINKED ISSUE #3481: Markdown editor not show characters correctly on RTL

fsamapoor:
Fixes #3481 

I dug into the [git history leading to the addition of this utility](https://github.com/filamentphp/filament/commit/b09177506238b2820bb94c5ddb718b5aceabebe7), but I didn't get the reasoning honestly.
I have double-checked the proposed PR with @themahdavi and removing the rtl-specific utility does solve the issue.

### DIFF
--- a/packages/forms/resources/views/components/markdown-editor.blade.php
+++ b/packages/forms/resources/views/components/markdown-editor.blade.php
@@ -220,7 +220,7 @@
                         @if (! $isConcealed())
                             {!! $isRequired() ? 'required' : null !!}
                         @endif
-                        class="tracking-normal overflow-y-hidden font-mono block absolute bg-transparent top-0 text-sm left-0 z-1 w-full h-full min-h-full resize-none transition duration-75 rounded-lg shadow-sm outline-none focus:border-primary-500 focus:ring-1 focus:ring-inset focus:ring-primary-500 caret-black whitespace-pre-wrap rtl:whitespace-normal"
+                        class="tracking-normal overflow-y-hidden font-mono block absolute bg-transparent top-0 text-sm left-0 z-1 w-full h-full min-h-full resize-none transition duration-75 rounded-lg shadow-sm outline-none focus:border-primary-500 focus:ring-1 focus:ring-inset focus:ring-primary-500 caret-black whitespace-pre-wrap"
                         x-bind:class="{
                             'dark:caret-white dark:focus:border-primary-500': @js(config('forms.dark_mode')),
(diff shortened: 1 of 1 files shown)


danharrin:
The fix for #3481 (in #3526) was to add this class in the first place... I don't understand how removing it wouldn't reintroduce that issue.

themahdavi:
Hey @danharrin 
Like you I don't understand what happens!
When issued this problem code has `whitespace-pre-wrap` exists but problem present
Now this class work properly !

danharrin:
Have you definitely tried on Chrome?

themahdavi:
On Chrome, Firefox and Microsoft Edge tested

When `rtl:whitespace-normal` not removed on Firefox markdown editor crashed completely!

themahdavi:
this examples when `rtl:whitespace-normal` exists !
 
on Chrome : 
image


on Firefox:
image

themahdavi:
This PR : 

on Chrome :
image


on Firefox: 
image

fsamapoor:
@danharrin I might be wrong, but it seems to me that removing the utility from both elements did the trick. Removing just one of them did not fix the issue. Maybe that was the case before.

zepfietje:
> @danharrin I might be wrong, but it seems to me that removing the utility from both elements did the trick. Removing just one of them did not fix the issue. Maybe that was the case before.

But the fix in https://github.com/filamentphp/filament/pull/3526/files removed it from both elements actually..?

If you've both tested it, @fsamapoor @themahdavi, and you're sure it works fine now, we might as well merge it, @danharrin?

fsamapoor:
@zepfietje We did, and everything seemed fine. I just wasn't able to follow the commits step by step, so that I can argue if it was broken all along or if something has changed in the process.

PR #6288, opened 2023-04-22
Use actions for buttons and links

zepfietje:


danharrin:
Whats up with the tests?

zepfietje:
No idea, maybe some internal thing since we use `callAction()` instead of `call()` now as the buttons have been converted to actions.

PR #6303, opened 2023-04-24
Support dropdown max height

mackhankins:
This is a followup to [this discussion](https://github.com/filamentphp/filament/discussions/4333#discussioncomment-5666689).  @zepfietje you said you had some ideas about how to accomplish this, but here is a PR so we can discuss.


mackhankins:
The problem I see initially is with a declared height  like `h-[56rem]` you can run into extra height on short column lists.  Since that panel is used in a lot of places you need to be able to pass through a class.

packages/support/resources/views/components/dropdown/index.blade.php
```diff
@@ -40,7 +40,7 @@
             wire:key="{{ $attributes->get('wire:key') }}.panel"
         @endif
         @class([
-            'filament-dropdown-panel absolute z-10 w-full divide-y divide-gray-100 rounded-lg bg-white shadow-lg ring-1 ring-black/5 transition',
+            'filament-dropdown-panel absolute z-10 h-[56rem] overflow-y-scroll w-full divide-y divide-gray-100 rounded-lg bg-white shadow-lg ring-1 ring-black/5 transition',
```
zepfietje:
```suggestion
            'filament-dropdown-panel absolute z-10 max-h-[30rem] w-full divide-y divide-gray-100 overflow-y-auto rounded-lg bg-white shadow-lg ring-1 ring-black/5 transition',
```

zepfietje:
Thanks for putting together this PR, @mackhankins!

> The problem I see initially is with a declared height like `h-[56rem]` you can run into extra height on short column lists. Since that panel is used in a lot of places you need to be able to pass through a class.

The extra height can be fixed by using max-height instead.
I've updated this PR to use max-height with the value equal to the max dropdown height used in GitHub (e.g. issue labels dropdown) for now.

We should definitely make this configurable indeed (just like width already is).
Finally we should decide on what the default should be for the filters / columns toggle dropdowns.

mackhankins:
> The extra height can be fixed by using max-height instead.
> I've updated this PR to use max-height with the value equal to the max dropdown height used in GitHub (e.g. issue labels dropdown) for now.

Testing it now, clean.  I like the height it's currently set at it, but that's all subjective.  It works great for my purposes.

> We should definitely make this configurable indeed (just like width already is).
> Finally we should decide on what the default should be for the filters / columns toggle dropdowns.

This one might be a little out of my TALL knowledge, but regardless I'm going to dig around to learn how it works.

mackhankins:
Is this the type of configurable solution we're looking for or did you have some other ideas?

zepfietje:
> Is this the type of configurable solution we're looking for or did you have some other ideas?

I think it's best to accept the max-height as parameter in the dropdown component itself.
Then the toggleable columns and filters panel can be configured on their own.

No worries, totally understandable if it's new to you.
I'll work on this PR in a bit. 👍

zepfietje:
I'm not too sure yet about the approach/defaults.
The table filter and column toggle forms have a default max height now (not sure if we view this as a UI improvement or breaking change?).

As the default Tailwind CSS max height classes don't make sense for dropdown sizes, I've not added any other options to the match statement.
Passing nothing (so `null`) defaults to `max-h-[30rem]`.
Passing a custom value (e.g. `max-h-[40rem]`) simply uses that value.

Thoughts?

danharrin:
Adding a max height on all dropdowns is a breaking change, right? And I don't think it's a good default.

1) We should add the 30rem default inside the files that consume the dropdown instead of the dropdown itself.
2) Instead of passing Tailwind classes, we should use inline styles. In the dropdown view, we check for `$maxHeight` and add the `style` tag if it's there.

Thoughts @zepfietje?

zepfietje:
> Adding a max height on all dropdowns is a breaking change, right? And I don't think it's a good default.

Yeah I'm not sure if it's a good default, that's why I asked for your thoughts.
I do think that 30rem would be suitable for most cases though.

> 1. We should add the 30rem default inside the files that consume the dropdown instead of the dropdown itself.

That's fine by me as well.
Just implemented it inside the dropdown initially as we do the same for width.

> 2. Instead of passing Tailwind classes, we should use inline styles. In the dropdown view, we check for `$maxHeight` and add the `style` tag if it's there.

That could work too.
Though I stuck to Tailwind classes since they provide proper values that fit the design system.
Although in this case you'd probably not want to go smaller than 30rem and that's already out of the range of values provided by Tailwind.
So maybe in the end it is ok to skip the Tailwind values for max height altogether yeah.

danharrin:
Well you left the design system when you decided on arbitrary value classes 😛 I think it's ok to be explicit here to resize the dropdown contextually

zepfietje:
> Well you left the design system when you decided on arbitrary value classes 😛 I think it's ok to be explicit here to resize the dropdown contextually

Yeah that's what I meant. 😂

danharrin:
Thanks

PR #6315, opened 2023-04-25
the pagination in table was hidden

Whizboy-Arnold:


zepfietje:
Thanks for creating a PR, @Whizboy-Arnold!
This is not a bug though, as the extended pagination controls are only shown on desktop-size screens as the mobile design doesn't really allow for it.

Whizboy-Arnold:
am on desktop and yet it still doesn't show up. Or what am i doing wrong @zepfietje 
am using the stock tables as they come here is a sample on a part/



notworking

zepfietje:
It's probably because you only have a few table rows (less than the pagination records per page setting)?

Whizboy-Arnold:
@zepfietje it seem shten maybe the table is being displayed too thin by default, cant get the pagination to display.

sorry meant to send this

image


then i wanted to show in the official demo but  the demo does not show the tables or anything
https://demo.filamentphp.com/shop/products

image

zepfietje:
How many records are there in your first screenshot?
There's still 10, which is not more than the default records per page for pagination.

Regarding the demo, it's working just fine on my end.

Whizboy-Arnold:
@zepfietje  the initial screenshot of 'clients' was wrong i sent it wrongly. however the second screen where iit displays only 10 records i have more than 3000 records, i can manually show this when i remove the hide class on my browser.

image

----------
Hey just got both the demo and my work ok. 

oh yeap i have seen  it is working when i use a different browser instance, probably its an issue with my browser then  ok thenks

will contact in case i replicate the error

zepfietje:
Glad you got it working.
And yes, let us know if you've got a reproducible setup so we can fix it if it's an actual issue.

PR #6317, opened 2023-04-26
Add render hooks to relation managers

archilex:
This PR adds render hooks to the Relation managers.

(This is somewhat selfish as I would love to be able to render my plugin in relation managers as well, but hopefully other's can use this as well).


archilex:
Thanks!

PR #6321, opened 2023-04-26
Support `HtmlString` in repeater and builder `itemLabel`

bernhardh:
As discussed in PR #6191

### DIFF
--- a/packages/forms/src/Components/Repeater.php
+++ b/packages/forms/src/Components/Repeater.php
@@ -488,7 +488,7 @@ class Repeater extends Field implements Contracts\CanConcealComponents
         );
     }
 
-    public function getItemLabel(string $uuid): null | string | Htmlable
+    public function getItemLabel(string $uuid): string | Htmlable | null
     {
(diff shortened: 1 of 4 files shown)


danharrin:
If an `HtmlString` object is used, you shouldn't need to make changes like this to the view. This is just opening everything up to XSS attacks.

bernhardh:
Ok. I hope this was my last mistake ;)

packages/forms/src/Components/Repeater.php
```diff
@@ -488,7 +488,7 @@ public function getCachedExistingRecords(): Collection
         );
     }
 
-    public function getItemLabel(string $uuid): ?string
+    public function getItemLabel(string $uuid): null | string | Htmlable
```
zepfietje:
```suggestion
    public function getItemLabel(string $uuid): string | Htmlable | null
```

zepfietje (approved):

danharrin:
Thanks

PR #6326, opened 2023-04-26
Add support for copying badge column

suleymanozev:
Same as text column.


zepfietje:
Thanks for the PR, @suleymanozev!

v3 merges the badge and text columns, so should add support for this automatically.
Do you think we should merge this for v2 now or not, @danharrin?

danharrin:
Thanks @suleymanozev!

PR #6336, opened 2023-04-27
Fix Text Input Error Focus

andrewdwallo:
This pull request is to fix the text input error focus on Light and Dark mode. The color of the focus while in an "error state" was inconsistent on light and dark mode, and, in my opinion, keeping the focus and border color the same (which is red in an error state) seemed to be the most logical solution. 

These are all examples in an error state with focus.

### Light Old Input Error Focus State:
LightOldInputError

### Dark Old Input Error Focus State:
DarkOldInputError

### Light New Input Error Focus State:
LightNewInputError

### Dark New Input Error Focus State:
DarkNewInputError


danharrin:
Thanks

PR #6351, opened 2023-04-29
Decrease width of open sidebar on mobile

andrewdwallo:
The mobile width was a little too wide, making it a little difficult to click outside of it with your thumb. Also, the brand in the Sidebar's header had a wide clickable area, which could cause a user to be redirected to the application's `home_url`, so I made the clickable area match the text width of the brand/application name.

### Before: Around 85% of Screen Width
OldFilamentMobileUncollapsedSidebar

### After: 75% of Screen Width
Screenshot 2023-04-29 080408

### DIFF
--- a/packages/admin/resources/views/components/layouts/app/sidebar/index.blade.php
+++ b/packages/admin/resources/views/components/layouts/app/sidebar/index.blade.php
@@ -2,7 +2,7 @@
     x-data="{}"
     @if (config('filament.layout.sidebar.is_collapsible_on_desktop'))
         x-cloak
-        x-bind:class="$store.sidebar.isOpen ? 'filament-sidebar-open translate-x-0 max-w-[75%] lg:max-w-[var(--sidebar-width)]' : '-translate-x-full lg:translate-x-0 lg:max-w-[var(--collapsed-sidebar-width)] rtl:lg:-translate-x-0 rtl:translate-x-full'"
+        x-bind:class="$store.sidebar.isOpen ? 'filament-sidebar-open translate-x-0 max-w-[20em] lg:max-w-[var(--sidebar-width)]' : '-translate-x-full lg:translate-x-0 lg:max-w-[var(--collapsed-sidebar-width)] rtl:lg:-translate-x-0 rtl:translate-x-full'"
     @else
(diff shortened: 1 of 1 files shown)


danharrin:
Unfortunately we can't do the width change. I know it's just swapping Tailwind classes and technically not a breaking change, but there's no way I'm spending the next 2 months telling people to recompile their Tailwind custom themes because their sidebars are broken 😭 Painful

The other change seems good though.

andrewdwallo:
> Unfortunately we can't do the width change. I know it's just swapping Tailwind classes and technically not a breaking change, but there's no way I'm spending the next 2 months telling people to recompile their Tailwind custom themes because their sidebars are broken 😭 Painful
> 
> The other change seems good though.

Wait so which one can we do? Because the width of the whole sidebar only when open on mobile (not collapsed or open on desktop) was changed by only adjusting `max-w-[20em]` to `max-w-[16rem]`. I'm pretty sure `20em` isn't a thing either. Also, if you were reading based on what the "what-the-diff" bot said, the bot wasn't accurate with what changed. I'm not really sure what would be broken based on what was changed but I can understand any hesitation.

zepfietje:
Apart from the breaking change, I think the new sidebar would look too narrow on mobile.

andrewdwallo:
I mean I guess I was just trying to add a little more space so I could click outside of it on my phone. If it is a breaking change or the width is too narrow, all good!

danharrin:
@andrewdwallo both max width classes exist as Tailwind arbitrary classes, thats not the problem. The problem is that people who want to access the new class must recompile Tailwind, which will cause problems for those with custom themes who choose not to compile.

The fix for the brand not overflowing the sidebar is good.

andrewdwallo:
Okay sounds good, whatever is fine with me.

andrewdwallo:
Thanks for the info as well. I will keep that in mind.

andrewdwallo:
Since it is a breaking change do you want me to just remove the sidebar width change and keep the adjustment of the clickable area of the brand?

danharrin:
Yes please

danharrin:
Thanks

PR #6352, opened 2023-04-29
Add cell attributes to table columns

howdu:
Add attributes directly to the table cell `<td>` - `extraAttributes` is already added to a cell but it's nested 3 levels deep. 

Needed to change vertical alignment of particular column/cell.

### DIFF
--- a/packages/admin/resources/views/pages/actions/modal/actions/button-action.blade.php
+++ b/packages/admin/resources/views/pages/actions/modal/actions/button-action.blade.php
@@ -25,7 +25,7 @@
     :icon="$getIcon()"
     :icon-position="$getIconPosition()"
     :size="$getSize()"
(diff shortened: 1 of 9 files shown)


danharrin:
I think we can do the without modifying `cell.blade.php`. `<x-tables::cell {{ $column->getExtraCellAttributesBag() }}>`

howdu:
@danharrin without it you get error from ComponentAttributeBag
> trim(): Argument #1 ($string) must be of type string, array given

danharrin:
Sorry I mean `<x-tables::cell {{ $column->getExtraCellAttributeBag() }}>`

danharrin:
Ah I was wrong, its actually possible through `:attributes` and `\Filament\Support\prepare_inherited_attributes()`

danharrin:
Cleared up a few other places too

PR #6353, opened 2023-04-29
documentation: Add relationships for SelectColumn

Xoshbin:


zepfietje:
Thanks for the PR, @Xoshbin!
I'm not sure if this deserves its own section in the documentation though as it's not a dedicated relationship API or something.
Thoughts, @danharrin?

danharrin:
I agree @zepfietje, this sort of thing is more of an example and belongs on https://filamentphp.com/tricks

PR #6357, opened 2023-04-30
Fieldset optional legend

howdu:


danharrin:
I think fieldsets without legends break HTML guidelines. You can just use a Card instead.

howdu:
In our use case we've conditionally displaying a label so can't use a Card - it looks daft with empty string:
Screenshot 2023-04-30 at 11 34 43

Current web specs say the legend is optional
https://webplatform.github.io/docs/html/elements/fieldset/
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset

danharrin:
Well spotted, I didn't see that source

PR #6358, opened 2023-04-30
fix: consistent error state

danharrin:


awcodes:
This looks ok to me overall.

1. Not thrilled about the RichEditor not having focus styles in Firefox.
2. Native selects aren't showing border colors on error. 
3. Not sure, where we lost the arrow on native selects, but it's not showing

Screen Shot 2023-04-30 at 9 26 34 AM

danharrin:
The arrow has been gone for a while, do you mind looking into those issues for me? LMK if not and I will.

awcodes:
Pushed fixes for Select arrow and select the missing border color on native select errors.

RichEditor focus is fine, the issue is actually on the MarkdownEditor.

danharrin:
Thanks!

andrewdwallo:
I saw this and wanted to mention, are you sure that the error state is consistent in light mode? Because what is used right now, which is this in an error state (using the text-input component as an example):
```
'border-danger-600 ring-danger-600 dark:border-danger-400 dark:ring-danger-400': (@js($statePath) in $wire.__instance.serverMemo.errors),
```
Works only in dark mode, but the focus of the border and ring color in an error state, specifically only in light mode, stays the `primary` color, not red.

For the error state in light mode to stay red on focus, this had to be used instead or else the focus color of primary will show:
```
`border-danger-600 ring-danger-600 dark:border-danger-400 dark:ring-danger-400 focus:border-danger-500 focus:ring-danger-500`: (@js($statePath) in $wire.__instance.serverMemo.errors),
```

Maybe this is wanted, idk, but just making sure. Maybe it works in v3 on light mode somehow though.

PR #6376, opened 2023-05-02
Add stacking to images

archilex:
Per our discussion on Discord, this PR allows images to be stacked by extending the ImageColumn. 

I added all the additional methods at the end of `ImageColumn` to make it easier to review, but if you prefer to change the order let me know.

As for areas for improvement and future ideas:

1. **IMPLEMENTED:** _You will also see that I'm using the existing logic to only display images that return a path and then I add that to the $imageCount which I then use to calculate the remaining images. This works, however, if one or more images within the set `limit()` don't have an image, then the column could potentially not "fill up" the column with other available images._ 

_In other words, if the limit is set to 4, and say there are 8 potential `users.avatar`, if the users 3 & 4 dont have avatars, then the stack will only show 2 images and then it will say `+6` as the remaining. This is correct so it's not a bug, BUT ideally, we would continue to look through the remaining 4 images to see if there were 2 additional images so that we can show 4 images in the column._ 

_Not sure how big of a deal this is, but I didn't want to hold back on this initial PR to figure it out._

2. It might be nice to be able to add a z-index to the images on hover so they come to the front. 
3. It might also be neat to be able hover over the remaining count to show a window of the remaining images plus a field like name 
4. Right now when you hover over the row the row has it hover state color. The ring-color doesn't update to match this unfortunately. We'd have to add tailwind's `group` to the row class to enable this, but not sure if you want to do that or not.

Alright, let me know what changes need to be made!

### DIFF
--- a/packages/tables/resources/views/columns/image-column.blade.php
+++ b/packages/tables/resources/views/columns/image-column.blade.php
@@ -40,25 +40,23 @@
                     },
                 ])
             >
-                @foreach (array_slice($getImages(), 0, $getLimit()) as $image)
-                    @if ($path = $getStackedImagePath($image))
-                        @php
-                            $imageCount ++;
-                        @endphp
-                        <img
-                            src="{{ $path }}"
-                            style="
-                                {!! $height !== null ? "height: {$height};" : null !!}
-                                {!! $width !== null ? "width: {$width};" : null !!}
-                            "
-                            {{ $getExtraImgAttributeBag()->class([
-                                'max-w-none ring-white',
-                                'dark:ring-gray-800' => config('tables.dark_mode'),
-                                'rounded-full' => $isCircular(),
-                                $ring,
-                            ]) }}
-                        >
-                    @endif
+                @foreach ($getImagesWithPath() as $path)
+                    @php
+                        $imageCount ++;
+                    @endphp
+                    <img
+                        src="{{ $path }}"
+                        style="
+                            {!! $height !== null ? "height: {$height};" : null !!}
+                            {!! $width !== null ? "width: {$width};" : null !!}
+                        "
+                        {{ $getExtraImgAttributeBag()->class([
+                            'max-w-none ring-white',
+                            'dark:ring-gray-800' => config('tables.dark_mode'),
+                            'rounded-full' => $isCircular(),
+                            $ring,
+                        ]) }}
+                    >
                 @endforeach
 
                 @if ($shouldShowRemaining() && (! $shouldShowRemainingAfterStack()) && ($imageCount < count($getImages())))
--- a/packages/tables/resources/views/columns/image-column.blade.php
+++ b/packages/tables/resources/views/columns/image-column.blade.php
@@ -44,6 +44,7 @@
                     @php
                         $imageCount ++;
                     @endphp
+                    
                     <img
                         src="{{ $path }}"
                         style="
--- a/packages/tables/resources/views/columns/image-column.blade.php
+++ b/packages/tables/resources/views/columns/image-column.blade.php
@@ -41,7 +41,7 @@
                 ])
             >
                 @foreach (array_slice($getImages(), 0, $getLimit()) as $image)
-                    @if ($path = $getImagePath($image))
+                    @if ($path = $getStackedImagePath($image))
                         @php
                             $imageCount ++;
                         @endphp
--- a/packages/tables/resources/views/columns/image-column.blade.php
+++ b/packages/tables/resources/views/columns/image-column.blade.php
@@ -5,29 +5,120 @@
     @php
         $height = $getHeight();
         $width = $getWidth() ?? ($isCircular() || $isSquare() ? $height : null);
+        $overlap = $getOverlap() ?? 'sm';
+        $imageCount = 0;
+
+        $ring = match ($getRing()) {
+            0 => 'ring-0',
+            1 => 'ring-1',
+            2 => 'ring-2',
+            4 => 'ring-4',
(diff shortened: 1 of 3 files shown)


archilex:
I'm seeing an error with SpatieMediaLibraryImageColumn. Do I need to update that class to include the `$image = null` argument? Or is there a better way to fix that? Would this be a breaking change for other packages that extend the ImageColumn?

archilex:
I updated the logic for pending feature 1 to now filter images without paths so it will now show the available images up to `limit`, assuming they exist.

archilex:
Just realized I need to rework this a bit. I've only been working with square pictures and the css is going to stretch things on non square pictures.

archilex:
Ok fixed the cropping. However, I'm unsure why there is/was a wrapping div with overflow-hidden. I don't think that object-cover/object-center require the wrapping div do they? Maybe it's a fallback for browser compatibility?

packages/tables/src/Columns/ImageColumn.php
```diff
@@ -219,4 +203,168 @@ public function getExtraImgAttributeBag(): ComponentAttributeBag
...
+    public function showRemainingAfterStack(bool | Closure $condition = true): static
+    {
+        $this->shouldShowRemainingAfterStack = $condition;
+
+        return $this;
+    }
+
+    public function shouldShowRemaining(): bool
+    {
+        return $this->evaluate($this->shouldShowRemaining);
```
danharrin:
```suggestion
        return (bool) $this->evaluate($this->shouldShowRemaining);
```

packages/tables/src/Columns/ImageColumn.php
```diff
@@ -219,4 +203,168 @@ public function getExtraImgAttributeBag(): ComponentAttributeBag
...
+    }
+
+    public function shouldShowRemainingAfterStack(): bool
+    {
+        return $this->evaluate($this->shouldShowRemainingAfterStack);
```
danharrin:
```suggestion
        return (bool) $this->evaluate($this->shouldShowRemainingAfterStack);
```

packages/tables/src/Columns/ImageColumn.php
```diff
@@ -219,4 +203,168 @@ public function getExtraImgAttributeBag(): ComponentAttributeBag
...
+    public function stacked(bool | Closure $condition = true): static
+    {
+        $this->isStacked = $condition;
+
+        return $this;
+    }
+
+    public function isStacked(): bool
+    {
+        return $this->evaluate($this->isStacked);
```
danharrin:
```suggestion
        return (bool) $this->evaluate($this->isStacked);
```

packages/tables/resources/views/columns/image-column.blade.php
```diff
@@ -5,29 +5,119 @@
...
+                        default => '-space-x-1',
+                    },
+                ])
             >
-       @endif
-    </div>
+                @foreach ($getImagesWithPath() as $path)
+                    @php
+                        $imageCount ++;
+                    @endphp
```
danharrin, line 49:
```suggestion
                    @endphp
                    
```

danharrin:
Here's some initial feedback from me:

- This should target 3.x. I really do not want to have to merge all of these changes in by hand when the time comes, and also v3 includes much better handling of column relationships and JSON (retrieving column values from nested relationships etc), so it should provide a better baseline.
- We don't need `separator()`, we should support a JSON array of image paths by default, and then optionally a relationship. The first example in the docs should be with a JSON array, relationship support can be implied from the other examples / the existing "Column Relationships" docs.
- There should only be 1 call to `getImages()` in the view, so the logic isn't executed more than once. Preferably, we return the paths directly from that method, so we don't need `getImagesWithPath()`.
- Support will need to be added to the SpatieMediaLibraryImageColumn so that it is able to return more than one image.
- When multiple images are in the column but it isn't `stacked()`, what happens? We should probably still show multiple images in that case, but not stack them.
- We recently introduced `defaultImageUrl()` on the `ImageColumn`, those changes need merging in here. For a column with multiple images, we probably don't need to show a default image at all. But for a normal column that only displays 1 image, it should render the default. I'm open to other thoughts on this.

archilex:
That all sounds good and I'll work on those changes for v3. However, since there's no ETA on v3, and I already had a functioning stacked image column plug-in ready to go, are you opposed to me just releasing that now to benefit the community. Might also help those that won't be able update to v3 right away.

danharrin:
You can do whatever you wish - I just don't want to have to maintain it across two versions

archilex:
@danharrin Here's the next update per your comments:

- I've reworked the code and the docs as you suggested regarding supporting a JSON array by default and removing `separator()`. Relationships work as well: `ImageColumn::make('images.path')`
- I've fixed the multiple calls to getImages()
- I've merged support for defaultImageUrl(). This works really nicely in that if you are working with multiple images and a model doesn't have any images at all, it'll can show this default image.
- Regarding not stacking images if `stacked()` isn't used, I've update the blade conditional to check if there are multiple images or just a single image instead of whether or not `stacked()` is used. If it's an array then it will display multiple images with a `space-x-1` gap between them. If you add `stacked()` then they will stack on top with a default of `-space-x-1`. You can further customize this by using `overlap()` to tell how much overlap you want. 
- The only thing not included in PR is support for SpatieMediaLibraryImageColumn. I thought it'd be better to get this part nailed down and then apply this to Spatie. After you give your feedback on these changes I'll get started on that.

danharrin:
Thanks!

PR #6379, opened 2023-05-02
Upgrade to esbuild v0.17.x

tanthammar:
As requested @danharrin 
esbuild 0.17.x has a new api.


tanthammar:
Only one compiling error, but that is from the Trix src:
(but non-breaking, only a warning)

image

zepfietje:
Thanks for the PR, @tanthammar!
Looks like the compiled JS is no longer minimized though?

danharrin:
Just tested this, all form components are broken for me, similarly to what happened when I attempted this. I tested on all browsers with the same behaviour.

tanthammar:
Let's check on Wednesday in my repo and yours.

PR #6381, opened 2023-05-02
Respect disableForm(false) for ViewAction

linaspasv:
The following PR allows to override disableForm setting on a ViewAction.

```php
Tables\Actions\ViewAction::make()->disableForm(false)
```

### DIFF
--- a/packages/admin/src/Resources/Pages/ListRecords.php
+++ b/packages/admin/src/Resources/Pages/ListRecords.php
@@ -279,9 +279,9 @@ class ListRecords extends Page implements Tables\Contracts\HasTable
             ->authorize(fn (Model $record): bool => static::getResource()::canRestore($record));
     }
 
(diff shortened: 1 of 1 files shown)


zepfietje:
Could you elaborate on the thinking behind this PR, @linaspasv?
Seems like you'd simply want to use the `EditAction` instead?

linaspasv:
We had a discussion with Dan https://discord.com/channels/883083792112300104/1102651142925721742 , and I was offered for my use-case to use ``ViewAction::make()->disableForm(false)`` that we found later did not work. I could of course create a custom Action by copying most of the internals of ViewAction but it made sense to just allow to override and respect the ``disableForm(false)`` property on ViewAction if needed.

zepfietje:
Alright, handing this over to @danharrin then.

PR #6383, opened 2023-05-02
Fix mountedFormComponentActionsData state path name

cheesegrits:
The state path in HasFormComponentActions was using the wrong name, mountedFormComponentActionData instead of mountedFormComponentActionsData, from the recent nested actions changes.  This caused a fatal error when using things like an add option form on a select.


cheesegrits:
Just to be clear (the what-the-diff summary suggests this is just for readability), the problem is that it doesn't match the actual property name, so data_set() freaks out when it tries to actually set the data ...

https://github.com/filamentphp/filament/blob/3.x/packages/forms/src/Concerns/HasFormComponentActions.php#L31

PR #6387, opened 2023-05-03
Add German auth translation

pxlrbt:


packages/app/resources/lang/de/pages/auth/register.php
```diff
@@ -0,0 +1,47 @@
...
+            'label' => 'E-Mail-Adresse',
+        ],
+
+        'name' => [
+            'label' => 'Name',
+        ],
+
+        'password' => [
+            'label' => 'Passwort',
+            'validation_attribute' => 'password',
```
zepfietje: Also, shouldn't these be capitalized, @pxlrbt?
```suggestion
            'validation_attribute' => 'passwort',
```

packages/app/resources/lang/de/pages/auth/register.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+return [
+
+    'title' => 'Registrierung',
```
zepfietje: Wouldn't you want this to match the heading?

pxlrbt: They don't match in english either, but the title is now following the same convention as on Login.

packages/app/resources/lang/de/pages/auth/register.php
pxlrbt: Those didn't make sense to me because a `min` rule kicked in which doesn't use the attribute translation. Fixed it now

packages/app/resources/lang/de/pages/auth/register.php
zepfietje: Yeah it's a little different in English.
Dutch translations I've added today are similar to how you've approached it for German now.
Looks good to me.

PR #6388, opened 2023-05-03
Get form state before rate limiter so the form is validated

pxlrbt:
Currently the rate limiter kicks in before the form is validated so you have to wait a whole minute if some input is invalid. In case of password validation this is really frustrating.

### DIFF
--- a/packages/app/src/Pages/Auth/EmailVerification/EmailVerificationPrompt.php
+++ b/packages/app/src/Pages/Auth/EmailVerification/EmailVerificationPrompt.php
@@ -43,7 +43,7 @@ class EmailVerificationPrompt extends CardPage
             ->label(__('filament::pages/auth/email-verification/email-verification-prompt.buttons.resend_notification.label') . '.')
             ->action(function (): void {
                 try {
-                    $this->rateLimit(1);
+                    $this->rateLimit(2);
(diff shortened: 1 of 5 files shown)


zepfietje:
Isn't that the whole point of rate limiting login attempts?

pxlrbt:
Okay, probably the login one was too much, but I don't want to rate limit users on registration if they haven't entered a valid password, right? 😅

zepfietje:
I think all those rate limiters where correct originally.
The problem with registration now is that users can check if an account exists for an email without being rate limited.

pxlrbt:
Hm, we should at least increase the limit to 5 then. Otherwise you enter a password, it says "please enter at least X digits" you enter a new password and are rate limited 🤔

zepfietje:
Agree. Maybe also change the decay to 30 seconds instead?

danharrin:
I agree that it was correct before.

zepfietje:
What about the number of tries though, @danharrin?

danharrin:
I think it should depend on the page. I think we can safely increase it to 2 on the registration page, but not on login.

zepfietje:
But then you'd have to wait an entire minute when you accidentally type 1 character of your email address wrong..?

danharrin:
Okay lets do 2 for all forms then

PR #6392, opened 2023-05-03
Set validation attribute on fields

pxlrbt:
`Password::min(4)` is in the default rules so there should be a translation for that, too.

### DIFF
--- a/packages/app/src/Pages/Auth/Register.php
+++ b/packages/app/src/Pages/Auth/Register.php
@@ -100,7 +100,8 @@ class Register extends CardPage
                     ->required()
                     ->rule(Password::default())
                     ->dehydrateStateUsing(fn ($state) => Hash::make($state))
-                    ->same('passwordConfirmation'),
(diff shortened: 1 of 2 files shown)


zepfietje (approved):
Looks ready to be merged to me, @danharrin.

packages/app/src/Pages/Auth/Register.php
```diff
@@ -146,6 +146,7 @@ protected function getUserModel(): string
     protected function getMessages(): array
     {
         return [
+            'password.min' => __('validation.min.string', ['attribute' => __('filament::pages/auth/register.fields.password.validation_attribute')]),
```
danharrin: Shouldn't these be `data.password.min`? Or, can we not just set the `validationAttribute()` on the field?

zepfietje: Oh maybe yes actually, haven't tested.

zepfietje:
Reworked this PR, @danharrin. ✅

PR #6404, opened 2023-05-04
Support loading multiple relation counts

wit3:
this allow to load multiple count relations

Utilis for example when in custom column you would access to multiple relations count

like this
```php
<?php

namespace App\Tables\Columns;

use Filament\Tables\Columns\Column;

class CustomColumn extends Column
{
    protected string $view = 'tables.columns.custom-column';

    public static function make(string $name): static
    {
        $static = app(static::class, ['name' => $name]);
        $static->configure();
        $static->counts(['comments', 'likes']);
        return $static;
    }
}

```

### DIFF
--- a/packages/tables/src/Columns/Concerns/CanAggregateRelatedModels.php
+++ b/packages/tables/src/Columns/Concerns/CanAggregateRelatedModels.php
@@ -82,7 +82,7 @@ trait CanAggregateRelatedModels
         return $this->evaluate($this->relationshipToAvg);
     }
 
-    public function getRelationshipToCount(): array | string | null
+    public function getRelationshipToCount(): string | array | null
     {
         return $this->evaluate($this->relationshipToCount);
     }
--- a/packages/tables/src/Columns/Concerns/CanAggregateRelatedModels.php
+++ b/packages/tables/src/Columns/Concerns/CanAggregateRelatedModels.php
@@ -12,7 +12,7 @@ trait CanAggregateRelatedModels
 
     protected string | array | Closure | null $relationshipsToCount = null;
 
-    protected string | Closure | null $relationshipsToExistenceCheck = null;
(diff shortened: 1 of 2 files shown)


packages/tables/src/Columns/Concerns/CanAggregateRelatedModels.php
```diff
@@ -34,7 +34,7 @@ public function avg(string | Closure | null $relationship, string | Closure | nu
         return $this;
     }
 
-    public function counts(string | Closure | null $relationship): static
+    public function counts(string | Closure | null | array $relationship): static
```
zepfietje:
```suggestion
    public function counts(string | array | Closure | null $relationship): static
```

packages/tables/src/Columns/Concerns/CanAggregateRelatedModels.php
```diff
@@ -10,7 +10,7 @@ trait CanAggregateRelatedModels
 
     protected string | Closure | null $relationshipToAvg = null;
 
-    protected string | Closure | null $relationshipToCount = null;
+    protected string | Closure | null | array $relationshipToCount = null;
```
zepfietje:
```suggestion
    protected string | array | Closure | null $relationshipToCount = null;
```

packages/tables/src/Columns/Concerns/CanAggregateRelatedModels.php
```diff
@@ -82,7 +82,7 @@ public function getRelationshipToAvg(): ?string
         return $this->evaluate($this->relationshipToAvg);
     }
 
-    public function getRelationshipToCount(): ?string
+    public function getRelationshipToCount(): array | string | null
```
zepfietje:
```suggestion
    public function getRelationshipToCount(): string | array | null
```

danharrin:
Thanks

wit3:
Thanks guys

PR #6409, opened 2023-05-04
Fix DateTimePicker evaluations

saade:
Datetime picker accepts closures for the `hasDate`, `hasSeconds` and `hasTime`, but they're are not being evaluated. This PR fixes the behavior calling `$this->evaluate()` to evaluate the closures.


saade:
The calendar breaks when dynamically changing these properties, is this the "expected" behavior and the cause to the current non-evaluation approach?

If its a known `can't fix` issue, should we remove the Closure typing?

image

danharrin:
Thanks

PR #6410, opened 2023-05-04
Lighten cell description in dark mode

howdu:
Before
Screenshot 2023-05-04 at 18 21 07

After
Screenshot 2023-05-04 at 18 21 54

### DIFF
--- a/packages/tables/resources/views/columns/text-column.blade.php
+++ b/packages/tables/resources/views/columns/text-column.blade.php
@@ -56,7 +56,7 @@
     @if (filled($descriptionAbove))
         <div @class([
             'text-sm text-gray-500',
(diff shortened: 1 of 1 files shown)


packages/tables/resources/views/columns/text-column.blade.php
```diff
@@ -54,7 +54,7 @@
     ]) }}
 >
     @if (filled($descriptionAbove))
-        <div class="text-sm text-gray-500">
+        <div class="text-sm text-gray-500 dark:text-gray-400">
```
zepfietje: Please add a check for the tables dark mode config.

zepfietje (changes requested):

PR #6414, opened 2023-05-04
Add x-cloak to collapsible nav groups

howdu:
While alpine is loading all nav groups are visible and if this exceeds the window height you get a nasty scrollbar jump.

### DIFF
--- a/packages/admin/resources/views/components/layouts/app/sidebar/group.blade.php
+++ b/packages/admin/resources/views/components/layouts/app/sidebar/group.blade.php
@@ -46,8 +46,11 @@
     @endif
 
     <ul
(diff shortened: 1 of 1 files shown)


danharrin:
The scrollbar jump only affects Windows. However, on all machines, this `x-cloak` will cause a jump in all grouped sidebar items when navigating between pages, since they will be hidden before Alpine gets initialised.

@zepfietje @awcodes LMK if you disagree with this decision and we can reopen

awcodes:
Would style="display: none" fix the jump. It does the same thing as x-cloak but doesn't have to wait on the stylesheet to be parsed?

danharrin:
It would still cause a jump before Alpine is loaded though, which is my problem

howdu:
Yeah I see what you mean - if only we had `wire:navigate` now there wouldn't be a need for a full redirect after first load.

PR #6407, opened 2023-05-04
Do not send unnecessary headers when pretending to be a file

linaspasv:
This PR enhances the handling of Set-Cookie headers, which Laravel typically includes in every request. By preventing these headers, files can more effectively simulate being the requested assets. Additionally, this change has a positive impact on CloudFlare caching since CloudFlare bypasses all requests containing Set-Cookie headers, including static assets. This update ensures better caching behavior and improves overall performance.

PR #6426, opened 2023-05-05
Do not send unnecessary headers when pretending to be a file

linaspasv:
PR #6407 improvement with fixes for HTTP/1.1 users.

### DIFF
--- a/packages/admin/src/Http/Controllers/AssetController.php
+++ b/packages/admin/src/Http/Controllers/AssetController.php
@@ -74,11 +74,17 @@ class AssetController
             ]);
         }
 
-        return response()->file($path, [
-            'Content-Type' => $contentType,
-            'Expires' => $this->getHttpDate($expires),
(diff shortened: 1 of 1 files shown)


TFSThiagoBR98:
@danharrin This breaks Octane/Swoole:
Swoole\ExitException, the exit() call cause an error in Swoole

danharrin:
Reverting.

linaspasv:
Oh, I bet this is due ``exit()`` at the end of the method, the simple ``return`` with the added headers for HTTP 1.1 should work for all (including Octane/Swoole). :-)

```php
return response()
            ->file($path, [
                'Accept-Ranges' => 'bytes',
                'Content-Length' => filesize($path),
                'Content-Type' => $contentType,
                'Expires' => $this->getHttpDate($expires),
                'Cache-Control' => $cacheControl,
                'Last-Modified' => $this->getHttpDate($lastModified),
            ])
            ->send();
```

PR #6428, opened 2023-05-05
Refactor action name logic

zepfietje:


danharrin:
I think we should support with AND without the Action suffix

zepfietje:
This PR doesn't change anything, just refactors to make the code more readable, @danharrin.
Both with and without Action suffix are supported still.

danharrin:
As discussed on Discord

PR #6432, opened 2023-05-06
Add ability to search values in select fields

archilex:
When working on PR #6431 I saw that `select.js` only searches `labels`. However it can sometimes be useful to search the `value` instead/as well, especially when you're trying to avoid `getSearchResultsUsing()`. This PR adds that ability through new `shouldSearchInValues()` and `shouldSearchInLabels()` methods.

Background:
I needed to create my own version of an icon picker where the `value` is the icon name and the `label` is the html that renders the icon. The only way to make this searchable currently is to add `getSearchResultsUsing()` which is slow since it has to load the files again. By NOT using `getSearchResultsUsing()` you can take advantage of `choice.js` searching which is blazing fast. However to do that you need to be able to search the `value`. 

SLOW using `getSearchResultsUsing()`

https://user-images.githubusercontent.com/6097099/236599169-7fa5aa72-ce9e-47cd-9725-84500b2327f5.mov

FAST just using `searchable()`

https://user-images.githubusercontent.com/6097099/236599198-8543fcb0-625c-49a9-bd6f-2ff0828ab2dc.mov

### DIFF
--- a/packages/forms/resources/js/components/select.js
+++ b/packages/forms/resources/js/components/select.js
@@ -26,6 +26,7 @@ export default (Alpine) => {
             searchDebounce,
             searchingMessage,
             searchPrompt,
+            searchFields,
             state,
         }) => {
             return {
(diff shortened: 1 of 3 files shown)


danharrin:
Thanks

PR #6439, opened 2023-05-06
Fix PostgreSQL data type incompatibility issue in Notification query

a21ns1g4ts:
The issue was caused by a data type incompatibility between postgres and the given query. Laravel defined the 'data' column as 'text' type.  It should be noted that there may be reasons why the 'data' column is defined as 'text' type, but regardless, this problem can be circumvented by using the 'like' operator to search for a specific string within the 'data' column. This allows for the retrieval of records that correspond to the 'format' key with the value 'filament', regardless of the data type defined in the 'data' column."  

To resolve the issue, this PR updates the query to use appropriate syntax for each database driver. For PostgreSQL, the query now uses '%"format":"filament"%' using the 'like' operator. . For all other drivers, the query continues to searches for the key 'format' using the '->' operator

This fix ensures that notifications are queried correctly regardless of the underlying database driver, allowing for consistent behavior across different database setups.

### DIFF
--- a/packages/notifications/src/Http/Livewire/Notifications.php
+++ b/packages/notifications/src/Http/Livewire/Notifications.php
@@ -83,9 +83,9 @@ class Notifications extends Component
     public function getDatabaseNotificationsQuery(): Builder | Relation
     {
         return match ($this->getUser()->notifications()->getConnection()->getDriverName()) {
-             /** @phpstan-ignore-next-line */
+            /** @phpstan-ignore-next-line */
             'pgsql' => $this->getUser()->notifications()->where('data', 'like', '%"format":"filament"%'),
-             /** @phpstan-ignore-next-line */
+            /** @phpstan-ignore-next-line */
             default => $this->getUser()->notifications()->where('data->format', 'filament'),
         };
     }
--- a/packages/notifications/src/Http/Livewire/Notifications.php
+++ b/packages/notifications/src/Http/Livewire/Notifications.php
@@ -55,7 +55,7 @@ class Notifications extends Component
             $this->notifications->forget($id);
         }
 
-        if (! $this->hasDatabaseNotifications()) {
+        if (!$this->hasDatabaseNotifications()) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


mackhankins:
Thank You!  Hoping this gets approved soon.

packages/notifications/src/Http/Livewire/Notifications.php
```diff
@@ -83,7 +83,12 @@ public function getDatabaseNotifications(): DatabaseNotificationCollection
     public function getDatabaseNotificationsQuery(): Builder | Relation
     {
         /** @phpstan-ignore-next-line */
-        return $this->getUser()->notifications()->where('data->format', 'filament');
+        return match ($this->getUser()->notifications()->getConnection()->getDriverName()) {
```
zepfietje: Could you extract `$this->getUser()->notifications()` to `$query` just above the match expression?

zepfietje (changes requested):

packages/notifications/src/Http/Livewire/Notifications.php
zepfietje: Mark as ready for review when done please.

packages/notifications/src/Http/Livewire/Notifications.php
```diff
@@ -83,7 +83,12 @@ public function getDatabaseNotifications(): DatabaseNotificationCollection
     public function getDatabaseNotificationsQuery(): Builder | Relation
     {
         /** @phpstan-ignore-next-line */
-        return $this->getUser()->notifications()->where('data->format', 'filament');
+        $query = $this->getUser()->notifications();
+
+        return match ($query->getConnection()->getDriverName()) {
+            'pgsql' => $query->notifications()->where('data', 'like', '%"format":"filament"%'),
+            default => $query->notifications()->where('data->format', 'filament'),
```
zepfietje:
```suggestion
            'pgsql' => $query->where('data', 'like', '%"format":"filament"%'),
            default => $query->where('data->format', 'filament'),
```

a21ns1g4ts (commented):
Good. I was unsleeping at last night. :sleepy:

danharrin:
To be honest, I don't like the use of `like` to get a JSON property. We have this note in the docs about just using a `json()` column, does this not work?

Screenshot 2023-05-08 at 12 24 23

a21ns1g4ts:
@danharrin You are right. It is much easier to change the type to JSON, but currently this field stores other things that are not JSON. Anyway, we can simply make it easier to extend this behavior.

danharrin:
The field shouldn't store anything apart from JSON if it should work with Filament notifications

a21ns1g4ts:
What is the best way to extend this functionality without modifying the package code ?

danharrin:
Extend the `Notifications` class, override `getDatabaseNotificationsQuery()` and use that as the LW component instead.

But I would urge you to not store plain text in that column. JSON is the only thing that will work.

PR #6445, opened 2023-05-07
Add ability to translate suggestions for input tags

howdu:
Allow for array with key + label in suggestions.

```php
Forms\Components\TagsInput::make('')->suggestions(fn () => [['test' => __('test')]])
```


danharrin:
I don't understand, whats the point? 🤔

howdu:
Show a translated label which may be different to the value that's saved.

danharrin:
I don't think that's not how `<datalist>` works IIRC

PR #6468, opened 2023-05-10
Update Asset Manager registration

dvanscott:
This update uses `instanceof` in the `match` statement of the `AssetManager` registration method. By using this check, it means that as long as a bound asset class extends the Filament version, it should work as expected.

I tested and it seems to work for my use case, but am definitely open to feedback on ways to make it better.

Sorry for the extra theming classes from my other PR. Made a mistake with my branching.


dvanscott:
So one thing I'm noticing now that could be a problem... the theme asset class extends from the CSS asset class, which means that themes will get incorrectly classified as a CSS asset instead of a theme asset. Thoughts on the best way to address that?

danharrin:
> So one thing I'm noticing now that could be a problem... the theme asset class extends from the CSS asset class, which means that themes will get incorrectly classified as a CSS asset instead of a theme asset. Thoughts on the best way to address that?

We just move the Theme check first so that it will have priority over a generic Css object

PR #6474, opened 2023-05-10
Setting time in DateTimePicker init() after defining the watchers

### LINKED ISSUE #6463: Datetimepicker issue on second item update
### Package

filament/forms

### Package Version

v2.17.3

### Laravel Version

v10.9

### Livewire Version

v2.12.3

### PHP Version

PHP 8.1

### Problem description

The time in the datetime turns to zero after the second edit of records in admin panel using the default DateTimePicker.

### Expected behavior

After edit the second item the time should be kept.

### Steps to reproduce

1. Create two a resource with a form that has DateTimePicker as show below:
```
                Forms\Components\DateTimePicker::make('from'),
```

2. Add two records to the list with datetime.
3. Open the first record in edit mode and press save.
4. Open the second record in edit mode and press save. (You will notice the time turns to 00:00).

Steps To Reproduce
 

### Reproduction repository

https://github.com/malzariey/datepickerissue

### Relevant log output

_No response_

malzariey:
PR Summary
Solving issue #6463 by defining setting the time after defining the watchers.

### DIFF
--- a/packages/forms/resources/js/components/date-time-picker.js
+++ b/packages/forms/resources/js/components/date-time-picker.js
@@ -72,10 +72,6 @@ export default (Alpine) => {
                         date = null
                     }
 
-                    this.hour = date?.hour() ?? 0
-                    this.minute = date?.minute() ?? 0
-                    this.second = date?.second() ?? 0
-
                     this.setDisplayText()
                     this.setMonths()
                     this.setDayLabels()
@@ -249,6 +245,10 @@ export default (Alpine) => {
 
                         this.setDisplayText()
                     })
+
+                    this.hour = date?.hour() ?? 0
+                    this.minute = date?.minute() ?? 0
+                    this.second = date?.second() ?? 0
                 },
 
                 clearState: function () {


danharrin:
Hey, could you please explain how changing the order fixes the problem? I would like to understand

zepfietje:
Was about to ask the same, as technically it shouldn't make any difference I think.

malzariey:
> Hey, could you please explain how changing the order fixes the problem? I would like to understand

In The first run, first init() of the library, everything is defined well. But once the second init() (Different Record and not opening the picker) I dont know why but the watches get triggered with the previous time (hour, minutes and seconds watcher) then triggers again and zero the time. So resetting the state on each init() solved the problem.

malzariey:
> > Hey, could you please explain how changing the order fixes the problem? I would like to understand
> 
> In The first run, first init() of the library, everything is defined well. But once the second init() (Different Record and not opening the picker) I dont know why but the watches get triggered with the previous time (hour, minutes and seconds watcher) then triggers again and zero the time. So resetting the state on each init() solved the problem.

This method will just make sure that watchers does not reset the time variables incorrectly.

danharrin:
I bet this is an Alpine bug, watchers should get destroyed when you destroy the root element of the component

zepfietje:
So how do we go forward with this PR?

danharrin:
I don't know... I have a funny feeling this is going to have unintended consequences

danharrin:
Let's see how it goes this week

danharrin:
Reverting: https://discord.com/channels/883083792112300104/1106970245928853535

PR #6479, opened 2023-05-10
Fix: Hidden Inputs taking up space in grid

awcodes:
Before:

Screen Shot 2023-05-10 at 10 09 58 AM

After:

Screen Shot 2023-05-10 at 10 10 07 AM


zepfietje:
Shouldn't we make sure to pass `true` for `$hidden` instead?
I don't know where it's passed from though, so this PR might be just fine.

awcodes:
We can't do it as hidden true, because it takes it out of the form submission. I tried it. lol. That's why it's set in the HiddenInput component as ->columnSpan('hidden') instead of ->hidden()

->hidden() also prevents the component from being rendered at all.

danharrin:
Thanks

PR #6494, opened 2023-05-11
Add ability to change datepicker icon

howdu:


packages/forms/docs/03-fields.md
```diff
@@ -1142,10 +1142,16 @@ To disable specific dates:
 use Filament\Forms\Components\DateTimePicker;
 
 DateTimePicker::make('date')
-    ->label('Appointment date')
-    ->minDate(now())
-    ->maxDate(Carbon::now()->addDays(30))
-    ->disabledDates(['2022-10-02', '2022-10-05', '2022-10-15'])
+    ->icon('heroicon-o-calendar')
```
zepfietje: This snippet should remain unchanged?

packages/forms/resources/views/components/date-time-picker.blade.php
```diff
@@ -1,3 +1,6 @@
+@php
+    $icon = $getIcon();
+@endphp
```
zepfietje, line 3:
```suggestion
@endphp

```

packages/forms/resources/views/components/date-time-picker.blade.php
```diff
@@ -73,14 +78,25 @@
...
-                    <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M8 7V3m8 4V3m-9 8h10M5 21h14a2 2 0 002-2V7a2 2 0 00-2-2H5a2 2 0 00-2 2v12a2 2 0 002 2z" />
-                </svg>
-            </span>
+            @if ($icon)
+                <span class="absolute inset-y-0 right-0 flex items-center pr-2 pointer-events-none rtl:right-auto rtl:left-0 rtl:pl-2">
+                    @if ($icon === 'default')
+                        <svg @class([
+                            'w-5 h-5 text-gray-400',
+                            'dark:text-gray-400' => config('forms.dark_mode'),
+                        ]) xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" stroke="currentColor">
```
zepfietje: Please put each attribute on their own line.

```suggestion
                        <svg @class([
                            'w-5 h-5 text-gray-400',
                            'dark:text-gray-400' => config('forms.dark_mode'),
                        ]) xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" stroke="currentColor">
```

packages/forms/resources/views/components/date-time-picker.blade.php
```diff
@@ -73,14 +78,25 @@
...
+                            <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M8 7V3m8 4V3m-9 8h10M5 21h14a2 2 0 002-2V7a2 2 0 00-2-2H5a2 2 0 00-2 2v12a2 2 0 002 2z" />
+                        </svg>
+                    @else
+                        <x-dynamic-component
+                                :component="$icon"
+                            @class([
+                                'w-5 h-5 text-gray-400',
+                                'dark:text-gray-400' => config('forms.dark_mode'),
+                            ]) />
```
zepfietje: Fix formatting.

packages/forms/src/Components/DateTimePicker.php
```diff
@@ -38,6 +38,8 @@ class DateTimePicker extends Field
 
     protected string | Closure | null $timezone = null;
 
+    protected string | Closure | null $icon = 'default';
```
zepfietje: `null` should be default.
Disabling should be done by accepting a bool `false`.

zepfietje (changes requested):

packages/forms/resources/views/components/date-time-picker.blade.php
```diff
@@ -58,6 +62,8 @@
                 'dark:border-danger-400' => $errors->has($getStatePath()) && config('forms.dark_mode'),
                 'opacity-70' => $isDisabled(),
                 'dark:text-gray-300' => $isDisabled() && config('forms.dark_mode'),
+                'px-3' => ! $icon,
+                'pl-3 pr-10 rtl:pl-10 rtl:pr-3' => $icon || $icon === null,
```
zepfietje:
```suggestion
                $icon === false
                    ? 'px-3'
                    : 'pl-3 pr-10 rtl:pl-10 rtl:pr-3',
```

packages/forms/docs/03-fields.md
```diff
@@ -1148,6 +1148,15 @@ DateTimePicker::make('date')
     ->disabledDates(['2022-10-02', '2022-10-05', '2022-10-15'])
 ```
 
+The calendar icon can be changed, or disabled by passing `icon(false)`
```
zepfietje: Maybe add a separate example below for disabling.
Also add `:` to line before example.

packages/forms/src/Components/DateTimePicker.php
```diff
@@ -38,6 +38,8 @@ class DateTimePicker extends Field
 
     protected string | Closure | null $timezone = null;
 
+    protected string | Closure | null $icon = null;
```
zepfietje: Should still accept `bool`?

howdu: When you pass false `->icon(false)` then `$this->evaluate($this->icon)` returns empty string. 

I was trying to keep it simple without needing to add a `disableIcon` and simple blade if conditions.

zepfietje: Hmm really? Evaluating `false` should still evaluate to `false`.
Also, the approach I mentioned is how we handle such cases in other places too.
Don't need a separate `disableIcon` API or prop.

Edit: the empty string is the result of casting `false` to string.

howdu: Sorry yeah that's what was throwing me - updated now

packages/forms/src/Components/DateTimePicker.php
```diff
@@ -38,6 +38,8 @@ class DateTimePicker extends Field
 
     protected string | Closure | null $timezone = null;
 
+    protected bool | string | Closure | null $icon = null;
```
zepfietje: `bool` should come after `string` in this case.

howdu: No worries I appreciate your thoroughness

zepfietje: Glad! 😄

packages/forms/docs/03-fields.md
```diff
@@ -1148,6 +1148,24 @@ DateTimePicker::make('date')
     ->disabledDates(['2022-10-02', '2022-10-05', '2022-10-15'])
 ```
 
+The calendar icon can be changed:
```
zepfietje:
```suggestion
You may change the calendar icon using the `icon()` method:
```

packages/forms/docs/03-fields.md
```diff
@@ -1148,6 +1148,24 @@ DateTimePicker::make('date')
...
+You may change the calendar icon using the `icon()` method:
+
+```php
+use Filament\Forms\Components\DateTimePicker;
+
+DateTimePicker::make('date')
+    ->icon('heroicon-o-calendar')
+```
+
+Or you may pass `false` to hide the icon:
```
zepfietje:
```suggestion
Alternatively, you may remove the icon altogether by passing `false`:
```

packages/forms/resources/views/components/date-time-picker.blade.php
```diff
@@ -73,14 +80,32 @@
...
 
-            <span class="absolute inset-y-0 right-0 flex items-center pr-2 pointer-events-none rtl:right-auto rtl:left-0 rtl:pl-2">
-                <svg @class([
-                    'w-5 h-5 text-gray-400',
-                    'dark:text-gray-400' => config('forms.dark_mode'),
-                ]) xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" stroke="currentColor">
-                    <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M8 7V3m8 4V3m-9 8h10M5 21h14a2 2 0 002-2V7a2 2 0 00-2-2H5a2 2 0 00-2 2v12a2 2 0 002 2z" />
-                </svg>
-            </span>
+            @if ($icon || $icon === null)
```
zepfietje:
```suggestion
            @if ($icon !== false)
```

packages/forms/resources/views/components/date-time-picker.blade.php
```diff
@@ -73,14 +80,32 @@
...
+                    @else
+                        <x-dynamic-component
+                                :component="$icon"
+                            @class([
+                                'w-5 h-5 text-gray-400 hover:text-gray-800 focus:text-primary-600',
+                                'dark:text-gray-400 dark:hover:text-gray-200 dark:focus:text-primary-600' => config('forms.dark_mode'),
+                            ])
+                        />
+                    @endif
+                </span>
```
zepfietje:
```suggestion
                <span class="absolute inset-y-0 right-0 flex items-center pr-2 cursor-pointer pointer-events-none rtl:right-auto rtl:left-0 rtl:pl-2">
                    <x-dynamic-component
                        :component="$icon ?? 'heroicon-s-calendar'"
                        @class([
                            'w-5 h-5 text-gray-400 hover:text-gray-800 focus:text-primary-600',
                            'dark:text-gray-400 dark:hover:text-gray-200 dark:focus:text-primary-600' => config('forms.dark_mode'),
                        ])
                    />
                </span>
```

packages/forms/resources/views/components/date-time-picker.blade.php
```diff
@@ -73,14 +80,17 @@
...
+            @if ($icon !== false)
+                <span class="absolute inset-y-0 right-0 flex items-center pr-2 cursor-pointer pointer-events-none rtl:right-auto rtl:left-0 rtl:pl-2">
+                    <x-dynamic-component
+                        :component="$icon ?? 'heroicon-s-calendar'"
+                        @class([
+                            'w-5 h-5 text-gray-400 hover:text-gray-800 focus:text-primary-600',
+                            'dark:text-gray-400 dark:hover:text-gray-200 dark:focus:text-primary-600' => config('forms.dark_mode'),
+                        ])
+                    />
+                </span>
```
zepfietje: I think we could just use the heroicon instead of the svg we had here.
Could you try refactoring to add the `span` classes onto the `x-dynamic-component` so we can get rid of that wrapping `span` @howdu?

Feel free to share a final screenshot to make sure it all looks good.

packages/forms/src/Components/DateTimePicker.php
```diff
@@ -38,6 +38,8 @@ class DateTimePicker extends Field
 
     protected string | Closure | null $timezone = null;
 
+    protected string | bool | Closure | null $icon = null;
```
zepfietje: Could you move the property up to ensure alphabetical order?

packages/forms/resources/views/components/date-time-picker.blade.php
howdu: @zepfietje I think you made these changes in the end - please let me know if not.

zepfietje: I did refactor to use dynamic component, but the wrapping span should still be removed.
Could you handle that and move the classes on span to `@class`?

howdu: I think it's best to keep this separate due the absolute positioning + padding. 

Noticed you've changed the icon to be 'solid' just checking that was intended.

Before
Screenshot 2023-05-18 at 11 08 44

After
Screenshot 2023-05-18 at 11 07 53

zepfietje: Why would the classes have to be on `span` though?

packages/forms/resources/views/components/date-time-picker.blade.php
```diff
@@ -73,14 +80,17 @@
...
-                    'w-5 h-5 text-gray-400',
-                    'dark:text-gray-400' => config('forms.dark_mode'),
-                ]) xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" stroke="currentColor">
-                    <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M8 7V3m8 4V3m-9 8h10M5 21h14a2 2 0 002-2V7a2 2 0 00-2-2H5a2 2 0 00-2 2v12a2 2 0 002 2z" />
-                </svg>
-            </span>
+            @if ($icon !== false)
+                <span class="absolute inset-y-0 right-0 flex items-center pr-2 cursor-pointer pointer-events-none rtl:right-auto rtl:left-0 rtl:pl-2">
+                    <x-dynamic-component
+                        :component="$icon ?? 'heroicon-s-calendar'"
```
zepfietje:
```suggestion
                        :component="$icon ?? 'heroicon-o-calendar'"
```

packages/forms/resources/views/components/date-time-picker.blade.php
zepfietje: @howdu, I've changed the icon back to outline as it would be "breaking".
Will have to update it to be solid/mini in v3.

howdu: > Why would the classes have to be on `span` though?

I tried it but because there's padding being applied you needed to incorporate this into the width/height.

zepfietje: Shouldn't the padding just be swapped out for margin then?

howdu: Yeah I guess but I always found it not great to mix absolute positioning with margin. Modern browsers handle this ok now but they didn't used to.  

By removing the wrapping span element the CSS will need to be rebuilt to include top: 50% and negative transform because you can’t use flex box to centre the icon. Checking if you think it's still worth it for the one element?

zepfietje: Well, let's keep it as is then.
Thanks for clarifying, @howdu.

packages/forms/src/Components/DateTimePicker.php
zepfietje: Could you handle this one still, @howdu?

packages/forms/resources/views/components/date-time-picker.blade.php
```diff
@@ -58,6 +62,9 @@
                 'dark:border-danger-400' => $errors->has($getStatePath()) && config('forms.dark_mode'),
                 'opacity-70' => $isDisabled(),
                 'dark:text-gray-300' => $isDisabled() && config('forms.dark_mode'),
+                $icon === false
+                    ? 'px-3'
+                    : 'pl-3 pr-10 rtl:pl-10 rtl:pr-3',
```
danharrin:
```suggestion
                'px-3' => $icon === false,
                'pl-3 pr-10 rtl:pl-10 rtl:pr-3' => $icon !== false,
```

PR #6502, opened 2023-05-13
Added Mongolian (mn) locale

tuvshinjargal:
Added Mongolian translations (admin, forms, support, tables).


danharrin:
Thanks

zepfietje:
Shouldn't we also add the locale to e.g. Filepond?

tuvshinjargal:
> Shouldn't we also add the locale to e.g. Filepond?

Yes. I will create an another PR in later.

PR #6507, opened 2023-05-14
Adds the possibility of having '0' as a prefix or postfix for different form inputs

fsamapoor:

This PR adds the possibility of having `'0'` as a prefix or postfix for three form inputs:
- text input
- select input
- color-picker input

I tried to add `0` as a prefix to the phone number input with `->prefix('0')` and I noticed it is not being rendered, since it is being evaluated as a falsy parameter for the blade's `@if ($label = $getPrefixLabel())` directive.

before prefix fixation:
before
after prefix fixation:
after

I don't know if this is the most elegant way to do so or if it is really needed to support both prefixes and postfixes of all three inputs.
I just did every six of them to make it consistent.

Feel free to cherry-pick, make changes in any shape or form, or even close the PR. I just need `'0'` as a prefix to the text input and I did open the PR to document the desired result, instead of opening an issue.

### DIFF
--- a/packages/forms/resources/views/components/color-picker.blade.php
+++ b/packages/forms/resources/views/components/color-picker.blade.php
@@ -102,7 +102,7 @@
             </div>
         </div>
 
-        @if ($label = $getSuffixLabel())
+        @if (filled($label = $getSuffixLabel()) && ($getSuffixLabel() !== false))
             <span @class($affixLabelClasses)>
                 {{ $label }}
             </span>
--- a/packages/forms/resources/views/components/color-picker.blade.php
+++ b/packages/forms/resources/views/components/color-picker.blade.php
@@ -29,7 +29,7 @@
             <x-dynamic-component :component="$icon" class="w-5 h-5" />
         @endif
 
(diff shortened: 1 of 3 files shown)


fsamapoor:
Hey @danharrin, thank you for merging the PR.
One little gotcha. I noticed you removed the second condition, which I want to elaborate on.

`@if (filled($label = $getPrefixLabel()) && ($getPrefixLabel() !== false))`

The thing is, `filled(false)` returns `true`. So basically providing a closure as a prefix/postfix with the return type of `false`, makes the blade condition `true` and that's why I added the second check.

Without the second condition and for the sake of simplicity, having `->prefix(fn () => false)` will cause the template to render an empty `<span>`, which will force a visual effect as well:

Screenshot 2023-05-17 at 08-41-07 Create Customer - Rozalie
Screenshot from 2023-05-17 08-43-05

danharrin:
Thats fine, I dont see why anyone would use false. Null or an empty string is good

fsamapoor:
Very well then.

PR #6508, opened 2023-05-15
Preserve persisted sort order with defaultSort() set

lupinitylabs:
### The problem

When using the `->defaultSort()` method on the table builder while using session persisted sort via `shouldPersistTableSortInSession()`, the default sort overrides any persisted sort order on any new render of the table.

### Expected behavior

`defaultSort()` should only be applied if the `$tableSortColumn` is blank AND there is no persisted table sort info in the session.

### Steps to reproduce

Add a `->defaultSort('name')` for any table, and on your ListRecords class, define 
```php
    protected function shouldPersistTableSortInSession(): bool
    {
        return true;
    }
```
The table should now be sorted by `name`. Click on any other column to change the sorting, e.g. to `created_at`. Reload the table from the navigation. The expected result should be a table sorted by `created_at`, but the actual sorting is by `name`.

### Explanation

This behavior has always puzzled me, as it effectively isn't possible to use sort persistence on a table with a default sorting.

The previous code always applied the default sort column and order on component mount, and since persisted sort settings are always applied if no sort column is set, it is never recalled on tables with a default sort order.

This PR suggests a change to mitigate this by setting the default sort column and direction only if the sort column is empty after applying both the user choice and persisted sort settings from the session. 

### Defaulting to default sort order

Because we are now applying the default sort column every time the component is refreshed and the sort column is null, and not only once on mount, the sort behavior changes:

#### Previous behavior

> clicked-column asc
> clicked-column desc
> null null

#### New behavior

> clicked-column asc
> clicked-column desc
> default-column default-sort

I would like to discuss whether it feels more logical to default to no sort at all (table order) if a default order is set for the table. In order to keep this behavior, it would be possible to move the code pulling the persisted sort settings from booted to mounted. 

However, I would personally expect the default sort order to be applied whenever the table would otherwise be unsorted.

What do you think?

### DIFF
--- a/packages/tables/src/Concerns/CanSortRecords.php
+++ b/packages/tables/src/Concerns/CanSortRecords.php
@@ -69,13 +69,14 @@ trait CanSortRecords
             return $query->orderBy($this->getTableReorderColumn());
         }
 
-        $sortColumn = $this->tableSortColumn;
+        $sortColumn = $this->tableSortColumn ?? $this->getDefaultTableSortColumn();
 
         if (! $sortColumn) {
             return $query;
         }
 
-        $sortDirection = $this->tableSortDirection === 'desc' ? 'desc' : 'asc';
+        $sortDirection = $this->tableSortDirection ?? $this->getDefaultTableSortDirection();
+        $sortDirection = $sortDirection === 'desc' ? 'desc' : 'asc';
 
         $column = $this->getCachedTableColumn($sortColumn);
 
--- a/packages/tables/src/Concerns/InteractsWithTable.php
+++ b/packages/tables/src/Concerns/InteractsWithTable.php
@@ -132,7 +132,9 @@ trait InteractsWithTable
 
             $this->tableSortColumn = $sort['column'] ?? null;
             $this->tableSortDirection = $sort['direction'] ?? null;
-        } elseif ($shouldPersistTableSortInSession) {
+        }
+
+        if ($shouldPersistTableSortInSession) {
             session()->put(
                 $sortSessionKey,
                 [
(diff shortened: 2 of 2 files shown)


lupinitylabs:
It could also be discussed to actually move the code for recalling the persisted sort settings to the mount hook instead of moving the defaults into the booted hook, as the persistence should only come into play when the component is reloaded (mounted).

danharrin:
Just refactored this a bit @lupinitylabs, does it still work as you think it should? Should keep the query string cleaner

lupinitylabs:
Hey @danharrin, sorry I didn't have time to get back at this, but from what I have seen the refactoring most likely just reverted my changes. It is important that the default sort order stays unset until the persisted settings are pulled in from the session. So my guess would be that the persistence code should have moved to the CanSortRecords trait as well.

I'll try the new 2.x and tell you whether it still works or not 🙏

danharrin:
The public properties should remain null when the default sort is applied. We shouldn't have anything in the query string when loading a table with a default sort.

lupinitylabs:
Nice, I'll give it a try right away!

lupinitylabs:
I've tried with the current `2.x-dev`, and unfortunately the default sort still overrides any persisted sort. ☹️

danharrin:
I just tested this myself, using `defaultSort('updated_at')`. I disabled query string persistence, and enabled session persistence. What is wrong with this behaviour?

https://github.com/filamentphp/filament/assets/41773797/bc61df99-2c68-4de3-b745-49edb0ce3862

lupinitylabs:
Yes, it works as long as you have the sort order in your query string - so reloading will keep the order. Try leaving the Category resource to another resource and returning to the Category resource afterwards. You will see that the table will then be ordered by `updated_at` again.

danharrin:
No - in my example, I disabled the query string, so I didn't need to leave every time, and it was ok?

PR #6532, opened 2023-05-17
Check if relation is already eagerloaded

Skullbock:
The table query builder in filament is smart enough to detect if a column is interacting with a relation.
In this case, it adds the eager loading automatically to the query.

If the query itself (set via `getTableQuery`) already sets the eager loading with some custom parameters, like a callback (`->with(['relation' => fn($q) => $q->where('foo', 'bar')])`) filament overrides it with its own eager loading.

This PR adds the check that the eager load is present or not before adding the new eager loading setting


danharrin:
Thanks

Skullbock:
Thanks to you for everything you do with filament and the speed of the reaction to the pr

PR #6537, opened 2023-05-17
Pass data to view as second param

howdu:
Consistent with how Laravel's view method works and needing to know about `viewData()`.

ViewField->make()->view('filament.forms.components.range-slider', ['custom_data' => CustomData::get()])

### DIFF
--- a/packages/support/src/Components/ViewComponent.php
+++ b/packages/support/src/Components/ViewComponent.php
@@ -32,9 +32,13 @@ abstract class ViewComponent extends Component implements Htmlable
 
     protected static array $methodCache = [];
 
-    public function view(string $view): static
+    public function view(string $view, array $viewData = []): static
     {
         $this->view = $view;
+
+        if ($viewData) {
+            $this->viewData = $viewData;
+        }
 
         return $this;
     }


packages/support/src/Components/ViewComponent.php
```diff
@@ -32,10 +32,14 @@ abstract class ViewComponent extends Component implements Htmlable
 
     protected static array $methodCache = [];
 
-    public function view(string $view): static
+    public function view(string $view, array $viewData = []): static
     {
         $this->view = $view;
 
+        if ($viewData) {
```
zepfietje, line 39: Isn't this check always going to be true if the default value is `[]`?

howdu: Empty array is false in PHP

zepfietje: Oh true, lol. Dan fixed this already after merging. 👍

zepfietje: Would be an issue if you call `viewData` before `view` though, @danharrin.

PR #6546, opened 2023-05-18
Remove unused method

zepfietje:
I'm not 100% sure, so please have a good look, @danharrin.
To me it seemed like this method isn't used (anymore).

### DIFF
--- a/packages/app/src/Resources/Pages/CreateRecord.php
+++ b/packages/app/src/Resources/Pages/CreateRecord.php
@@ -264,9 +264,4 @@ class CreateRecord extends Page
     {
         return static::$canCreateAnother;
     }
-
-    public static function disableCreateAnother(): void
-    {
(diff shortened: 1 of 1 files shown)


danharrin:
It's part of our public API, allowing you to disable the button across all create pages at once (since we dont have config anymore)

zepfietje:
Ah totally missed this.
Though we may still want to rename `disableCreateAnother` to `createAnother`, just like in the other places?

danharrin:
Sure, that's not bad. If it's in the v2 API, the old method will need to be deprecated too

zepfietje:
Ah I see there's already a method `createAnother()` on the class that calls the `create()` method... Thoughts, @danharrin?

danharrin:
Ah well we'll have to leave it as-is

Issue #6492, opened 2023-05-11
unable to delete bulk records when sorted with on relation table property (count)


henk-n:
### Package

filament/filament

### Package Version

v2.17.38

### Laravel Version

10.10.0

### PHP Version

8.2.2

### Problem description

When sorting on a related count column in the table, you get an error message when you want to delete the selected rows in the bulk action. 

response: (example of the git repo)
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'clothing_count' in 'order clause'

### Expected behavior

that the selected rows are deleted without error message

### Steps to reproduce

clone repo
configure a mysql database
migrate and seed
start laravel
go to the agencies resource
select one or more rows
bulk action, choose delete selected
sql error

### Reproduction repository

https://github.com/henk-n/sortissue/

danharrin:
Fixed by #6554

PR #6554, opened 2023-05-18
added applyEagerLoading and applyRelationshipAggregates to get selected table records

alperenersoy:
hey,

when we don't apply relationship aggregates, selected records are being hydrated and passed to actions without aggregates. while adding that, I thought `applyEagerLoading` is also needed here since we have it for the table query.

I think pipeline failure is not related to the changes. So keeping the PR open.


danharrin:
Thanks

henk-n:
Applied the fix manually to check #6492 is solved, and that is the case.

PR #6558, opened 2023-05-19
Update 05-actions.md

fetzi:
Add information about bulk action callback parameter name.

### DIFF
--- a/packages/tables/docs/05-actions.md
+++ b/packages/tables/docs/05-actions.md
@@ -81,7 +81,7 @@ BulkAction::make('delete')
     ->action(fn (Collection $records) => $records->each->delete())
 ```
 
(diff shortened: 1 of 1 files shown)


packages/tables/docs/05-actions.md
```diff
@@ -81,6 +81,8 @@ BulkAction::make('delete')
     ->action(fn (Collection $records) => $records->each->delete())
 ```
 
+> Please note that the parameter name `$records` in the `action` callback function is required, any other name won't work here.
```
danharrin:
```suggestion
> Please note that Filament uses the parameter name `$records` in the function to inject the collection. Any other parameter name will resolve from the container instead.
```

danharrin:
Thanks

PR #6561, opened 2023-05-19
Improve display of multi-line brand name

howdu:
Before
Screenshot 2023-05-19 at 12 22 31

After
Screenshot 2023-05-19 at 12 23 56

### DIFF
--- a/packages/admin/resources/views/components/brand.blade.php
+++ b/packages/admin/resources/views/components/brand.blade.php
@@ -1,6 +1,6 @@
 @if (filled($brand = config('filament.brand')))
     <div @class([
-        'filament-brand text-xl font-bold tracking-tight',
+        'filament-brand text-xl font-bold leading-none tracking-tight',
         'dark:text-white' => config('filament.dark_mode'),
     ])>
         {{ $brand }}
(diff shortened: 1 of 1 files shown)


danharrin:
To be honest, I don't like it. Looks to squashed to me. Handing this over to @filamentphp/design

howdu:
> To be honest, I don't like it. Looks to squashed to me. Handing this over to @filamentphp/design

Maybe it needs `line-height:1.1` if it's too tight, but reason for the change was because the gap between the lines was greater than the gap above/below which looks bad.

awcodes:
1.2 line height looks ok to me, but still feels crowded.
Screen Shot 2023-05-19 at 10 03 20 AM

I think this is really getting into the territory where this use case would be best handled by the existing blade logo overrides. Then the dev can tweak any way they need to. 

Personally, not thinking this really needs to be something core needs to worry about.

@danharrin if you do want to handle it I would go with an explicit 1.2 line height.

danharrin:
@awcodes is `leading-5` ok? (1.25)

awcodes:
I'd say that's fine too.

packages/admin/resources/views/components/brand.blade.php
```diff
@@ -1,6 +1,6 @@
 @if (filled($brand = config('filament.brand')))
     <div @class([
-        'filament-brand text-xl font-bold tracking-tight',
+        'filament-brand text-xl font-bold leading-none tracking-tight',
```
danharrin:
```suggestion
        'filament-brand text-xl font-bold leading-5 tracking-tight',
```

Issue #1553, opened 2022-02-10
Money Input Mask Empty When Zero


intrepidws:
### Package

filament/filament

### Package Version

v2.10.6

### Laravel Version

v8.83.0

### Livewire Version

v2.10.1

### PHP Version

PHP 8.0.13

### Bug description

When using the `money` input mask on a TextInput field with a value of zero - only `$` is displayed instead of `$0.00`.

### Steps to reproduce

My form schema has the following:

```
Forms\Components\TextInput::make('amount')
    ->mask(fn (Forms\Components\TextInput\Mask $mask) => $mask->money('$', ',', 2)),
```

For a field with a value of zero, this displays only `$` when I would expect that it should show `$0.00`.

ryangjchandler:
@intrepidws Is this still a problem since https://github.com/laravel-filament/filament/pull/1533 has been merged in?

danharrin:
Hm, I think this is just an iMask bug. We pass `->lazyPlaceholder(false)` which should fix this, but it doesn't.

intrepidws:
> @intrepidws Is this still a problem since #1533 has been merged in?

@ryangjchandler Yes I tested this on v2.10.6, which had #1533 merged.

Just to be certain, I upgraded to the recently released v2.10.7 and can confirm the problem still exists. I'll look into filing a bug report with iMask.

intrepidws:
@danharrin I'm sure that I've simplified the implementation too much or missed something obvious but I just made a CodePen with iMask that seems to work fine with a zero value.

https://codepen.io/intrepidws/pen/zYPzWdK

ryangjchandler:
I wonder if it's something to do with the configuration that the `Mask::money` method creates. Perhaps the normalise zeroes?

Will have to debug & dig into it at some point this weekend. @danharrin

intrepidws:
That pen should (at least to the best that I can tell) utilize all of the same iMask config that the `Mask::money` method creates.

danharrin:
Could you maybe try converting it to Alpine (using our code in text-input.js) and see if that still works?

intrepidws:
So I'm a little out of my element here, but from what I can [these lines](https://github.com/laravel-filament/filament/blob/d93c9fe3b8b26074eeca28a3e934eba803ccac26/packages/forms/resources/js/components/text-input.js#L18-L20) are meant to assign the value when there is one. But there are two issues here:

[...]

PR #6565, opened 2023-05-19
fix zero value not showing when using mask on text input

ariaieboy:
Alright, I've made a simple change that will fix this annoying issue https://github.com/filamentphp/filament/issues/1553.

The condition of the if will count as false since JS count number `0` as false and because of that the value of the input will not update and you can not see value 0 in the input.


danharrin:
Thanks

PR #6569, opened 2023-05-19
Fixes divide-y line in TagsInput when no tags are present

archilex:
I kept `dark:divide-gray-600` together with `divide-y`, but it could stay with the other classes.

### DIFF
--- a/packages/forms/resources/views/components/tags-input.blade.php
+++ b/packages/forms/resources/views/components/tags-input.blade.php
@@ -24,11 +24,14 @@
         }}
     >
         <div x-show="state.length || @js(! $isDisabled)">
-            <div @class([
-                'block w-full transition duration-75 divide-y rounded-lg shadow-sm sm:text-sm border overflow-hidden focus-within:ring-1 dark:divide-gray-600',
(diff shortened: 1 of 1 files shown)


packages/forms/resources/views/components/tags-input.blade.php
```diff
@@ -24,11 +24,14 @@
         }}
     >
         <div x-show="state.length || @js(! $isDisabled)">
-            <div @class([
-                'block w-full transition duration-75 divide-y rounded-lg shadow-sm sm:text-sm border overflow-hidden focus-within:ring-1 dark:divide-gray-600',
-                'border-gray-300 focus-within:border-primary-500 focus-within:ring-primary-500 dark:border-gray-600 dark:focus-within:border-primary-500' => ! $errors->has($statePath),
-                'border-danger-600 ring-danger-600 dark:border-danger-400 dark:ring-danger-400' => $errors->has($statePath),
-            ])>
+            <div 
+                x-bind:class="{ 'divide-y dark:divide-gray-600': state?.length }"
```
zepfietje: Could you try adding a top border to the tags div instead?
Would require no JS.

zepfietje (changes requested):

packages/forms/resources/views/components/tags-input.blade.php
archilex: Done!

danharrin:
Thanks

PR #6599, opened 2023-05-23
Update French translations

agencetwogether:
After last update v2.17.43, I update french translations to support new Underline button in RichText field.


zepfietje:
Thanks!

PR #6601, opened 2023-05-23
esbuild watch and auto-publish filament assets

tanthammar:
Please try this again.

- support --watch
- Fixing process.env.NODE_ENV
- plus auto publish filament assets in both watch and build.


danharrin:
Please remove the `php artisan filament:assets`, that is dependant on where your app is located

PR #6602, opened 2023-05-23
Navigation item additional methods

ousid:
Based on the discussion in https://github.com/filamentphp/filament/discussions/6566, this PR add the ability to show/hide the custom navigation items, by setting `visible` or `hidden` methods to the custom item.

### Example
```php
use Filament\Navigation\NavigationItem;

    Filament::registerNavigationItems([
        NavigationItem::make('Analytics')
            ->visible(fn (): bool => false)
            // or
            ->hidden(fn (): bool => true),
    ]);
});
```


packages/admin/resources/views/components/layouts/app/sidebar/item.blade.php
```diff
@@ -4,11 +4,18 @@
     'badge' => null,
     'badgeColor' => null,
     'icon',
+    'visible' => null,
+    'hidden' => null,
```
danharrin: I don't think any views need to be changed to accommodate this. If a navigation item is hidden, then it should not be in the array of items in the first place

packages/admin/src/Navigation/NavigationItem.php
```diff
@@ -26,6 +26,10 @@ class NavigationItem
 
     protected string | Closure | null $url = null;
 
+    protected ?bool $isHidden = null;
+
+    protected ?bool $isVisible = null;
```
danharrin:
```suggestion
    protected bool $isHidden = null;

    protected bool $isVisible = true;
```

packages/admin/src/Navigation/NavigationItem.php
```diff
@@ -60,6 +64,24 @@ public function icon(string $icon): static
...
+    }
+
+    public function hidden(bool | Closure $condition = null): static
+    {
+        $this->isHidden = is_callable($condition)
+                                ? $condition()
+                                : $condition;
+
+        return $this;
+    }
```
danharrin:
```suggestion
    public function visible(bool | Closure $condition = true): static
    {
        $this->isVisible = value($condition);

        return $this;
    }

    public function hidden(bool | Closure $condition = true): static
    {
        $this->isHidden = value($condition);

        return $this;
    }
```

packages/admin/src/Navigation/NavigationItem.php
```diff
@@ -123,6 +145,16 @@ public function getIcon(): string
         return $this->icon;
     }
 
+    public function getVisible(): ?bool
+    {
+        return $this->isVisible;
```
danharrin:
```suggestion
        return ! $this->isHidden();
```

packages/admin/src/Navigation/NavigationItem.php
```diff
@@ -123,6 +145,16 @@ public function getIcon(): string
...
+    }
+
+    public function getHidden(): ?bool
+    {
+        return $this->isHidden;
```
danharrin:
```suggestion
        if ($this->evaluate($this->isHidden)) {
            return true;
        }

        return ! $this->evaluate($this->isVisible);
```

packages/admin/src/Navigation/NavigationItem.php
```diff
@@ -123,6 +145,16 @@ public function getIcon(): string
         return $this->icon;
     }
 
+    public function getVisible(): ?bool
+    {
+        return $this->isVisible;
+    }
+
+    public function getHidden(): ?bool
```
danharrin:
```suggestion
    public function getHidden(): bool
```

packages/admin/src/Navigation/NavigationItem.php
```diff
@@ -123,6 +145,16 @@ public function getIcon(): string
         return $this->icon;
     }
 
+    public function getVisible(): ?bool
```
danharrin:
```suggestion
    public function getVisible(): bool
```

packages/admin/resources/views/components/layouts/app/sidebar/item.blade.php
danharrin: So these changes should be removed and replaced with changes in the code that returns the navigation structure. Test should also be updated

danharrin (changes requested):

danharrin:
Please mark this as `Ready for review` when the refactor is complete as requested :)

ousid:
Hi, Dan. 

Sure thing, I'll update the PR As soon as I can :)

packages/admin/resources/views/components/layouts/app/sidebar/group.blade.php
```diff
@@ -64,17 +64,19 @@ class="ml-1 h-3 w-3 flex-shrink-0"
...
-                    :icon="$item->getIcon()"
-                    :active-icon="$item->getActiveIcon()"
-                    :url="$item->getUrl()"
-                    :badge="$item->getBadge()"
-                    :badgeColor="$item->getBadgeColor()"
-                    :shouldOpenUrlInNewTab="$item->shouldOpenUrlInNewTab()"
-                >
-                    {{ $item->getLabel() }}
-                </x-filament::layouts.app.sidebar.item>
+                @if ($item->getVisible() || ! $item->getHidden())
```
danharrin:
```suggestion
                @if ($item->getVisible())
```

danharrin (changes requested):

danharrin:
Thanks

PR #6630, opened 2023-05-29
Fix global search panel position on mobile

ishmam99:
Adding sm:right-0 for mobile view search container

image
here we can see the search results are hidden on the left side,,, if we do the change it looks like this 
image

### DIFF
--- a/packages/admin/resources/views/components/global-search/results-container.blade.php
+++ b/packages/admin/resources/views/components/global-search/results-container.blade.php
@@ -8,7 +8,7 @@
     x-on:keydown.escape.window="isOpen = false"
     x-on:click.away="isOpen = false"
     x-on:open-global-search-results.window="isOpen = true"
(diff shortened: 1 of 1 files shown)


awcodes:
Looks fine to me, but does adding 'teleport' to the floating ui not force it to stay on screen?

zepfietje:
> Looks fine to me, but does adding 'teleport' to the floating ui not force it to stay on screen?

It's not using Floating UI, is it?
If it would, then we could add `shift` and/or `teleport` probably yeah.

danharrin:
This could be refactored to use Floating UI

awcodes:
@ishmam99 can you convert this to using floating ui. You can use the the support drop-down component as a reference for implementation.?

danharrin:
Since implementing Floating UI is a very different task, and you may choose not to take on the work, please open a new PR

PR #6652, opened 2023-05-31
Support `Htmlable` record titles

bernhardh:
TLDR: This PR only adds `HtmlString` as a return type to the `getTitle` and `getRecordTitle` method.

Reason:

I have a hotels resource, where some of the hotels have HTML inside their name, like `Hotel XY <sup>Superior</sup>` (Hotel XY <sup>Superior</sup>).  On edit pages and inside the breadcrumb, this looks obviously uggly, since its rendered htmlencoded. After PR I can:

```php

  public static function getRecordTitle(?Model $record): string | HtmlString | null
  {
        return new HtmlString($record->name);
  }
```

### DIFF
--- a/packages/admin/src/Pages/Page.php
+++ b/packages/admin/src/Pages/Page.php
@@ -13,6 +13,7 @@ use Illuminate\Contracts\Support\Htmlable;
 use Illuminate\Contracts\View\View;
 use Illuminate\Support\Facades\Route;
 use Illuminate\Support\Str;
+use Illuminate\Support\HtmlString;
 use Illuminate\Validation\ValidationException;
 use Livewire\Component;
 
@@ -223,7 +224,7 @@ class Page extends Component implements Forms\Contracts\HasForms, RendersFormCom
         return $this->subheading;
     }
 
-    protected function getTitle(): string
+    protected function getTitle(): string | HtmlString
     {
         return static::$title ?? (string) Str::of(class_basename(static::class))
             ->kebab()
--- a/packages/admin/src/Pages/Page.php
+++ b/packages/admin/src/Pages/Page.php
@@ -13,7 +13,6 @@ use Illuminate\Contracts\Support\Htmlable;
 use Illuminate\Contracts\View\View;
 use Illuminate\Support\Facades\Route;
 use Illuminate\Support\Str;
-use Illuminate\Support\HtmlString;
 use Illuminate\Validation\ValidationException;
 use Livewire\Component;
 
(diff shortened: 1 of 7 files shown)


zepfietje:
@danharrin, shouldn't all `HtmlString` type declarations be updated to `Htmlable` (also in rest of codebase)?

danharrin:
Correct, yes

zepfietje:
> Correct, yes

Can we do this in v2 or would it be breaking somehow?
There's just a couple that still need changing.

bernhardh:
Should I change it to `Htmlable` in my PR?

danharrin:
We can probably change all the HtmlString types to Htmlable without breaking in v2, yes

zepfietje:
Could you handle this and mark as ready for review when done, @bernhardh? 😊

bernhardh:
@zepfietje Done and thank you both for your great work! How to mark it as ready for review? I can't add labels to issues?

zepfietje:
Thanks, @bernhardh!
LGTM, handing over to @danharrin for merge.

> How to mark it as ready for review?

For drafts there's a button at the bottom of the PR where you'd usually have the merge button. I've marked it as ready now. 👍

juliomotol:
Not a breaking change but ever since we upgraded our Filament version, we encountered this in PHPStan with level 8:

```
Binary operation "." between Illuminate\Contracts\Support\Htmlable|string and ' ' results in an error.
 ```
 
 This is where we try to contcat `getTitle()` with another string.

**Comment by zepfietje** [comment]
Maybe cast it to a string first, @juliomotol?

**Comment by juliomotol** [comment]
Tried that, and it says:

```
Cannot cast Illuminate\Contracts\Support\Htmlable|string to string.
```

zepfietje:
Right, then you could check if it's `Htmlable` and call `toHtml()`.

juliomotol:
To add more context in this situation, we're not concatenating on `getTitle()` for display but rather for it to be saved to a database (more specifically to generate a description for an activity log). For now, we'll omit using `getTitle()` and use our own logic instead.

bernhardh:
@zepfietje wheren't these changes merged to v3? I am trying to update to v3, but they are once again not there. For example `packages/admin/src/Resources/Pages/Concerns/InteractsWithRecord.php` is again just returning `string` instead of `Htmlable|string`

zepfietje:
I think so. CC @danharrin.

bernhardh:
Ah, forget it. I am talking about the breadcrumb and this didn't work in v2 either ;). 

# V2
image

# V3
image


Will have a closer look and then maybe submit a new PR

PR #6655, opened 2023-05-31
Prevent livewire poll updates from overwriting TextInputColumn alpine state during editing.

jamesalvarez:
In a table I can set it to livewire poll (say every 5 seconds) using ->poll('5s').  This works nicely but if I have a TextInputColumn, and I start editing it, when the poll happens it erases what I have typed back to the value on the db.  
 
The issue is that alpine only updates the db on the change event, and not the input event.  It's possible to switch to the input event and update the db with every character typed, but this undesirable for a number of reasons.

PR sets an isEditing boolean to the component's alpine state, and ignores the livewire updates when it is true.  The isEditing boolean is set on focus and blur.

Tagging @pxlrbt as requested from discussion on discord.

### DIFF
--- a/packages/tables/resources/views/columns/text-input-column.blade.php
+++ b/packages/tables/resources/views/columns/text-input-column.blade.php
@@ -13,10 +13,15 @@
         error: undefined,
         state: @js($state),
         isLoading: false,
+        isEditing: false,
     }"
     x-init="
         Livewire.hook('message.processed', (component) => {
             if (component.component.id !== @js($this->id)) {
+                return
+            }
+
+            if (isEditing) {
                 return
             }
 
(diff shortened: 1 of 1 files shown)


pxlrbt:
I just searched for the last PR that changed it and it seems like Dan did work on this in the meantime to make these reactive again and thereby the "do not update" behaviour was lost.

Can you please check what currently happens when you have a `SelectColumn` opened for selection, while the polling happens? Do we need to apply this for that column too?

jamesalvarez:
It is less of a problem but the highlighting does indeed switch back to the previously selected item in the dropdown in Firefox, but not in Chrome.  

https://github.com/filamentphp/filament/assets/6019020/f4f192ed-c8db-40dc-b7ef-9d4acc2690bf

pxlrbt:
Can this be fixed similarly?

danharrin:
Hey @jamesalvarez please mark this as Ready for Review when this has been implemented for the SelectColumn too

danharrin:
Ok, just gonna merge this, if you want the fix for SelectColumn too then you will need to open a new PR. Thanks.

PR #6672, opened 2023-06-01
Fix TagsInput border when disabled

archilex:
Just realized that the border fix I implemented with #6569 didn't take into account when it is `disabled()`. This PR only shows the border when it's not disabled.

### DIFF
--- a/packages/forms/resources/views/components/tags-input.blade.php
+++ b/packages/forms/resources/views/components/tags-input.blade.php
@@ -80,7 +80,10 @@
                         x-if="state?.length"
                         x-cloak
                     >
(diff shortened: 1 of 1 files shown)


zepfietje:
Why should the border be hidden though when disabled, @archilex?

archilex:
@zepfietje If you want to show tags, but don't want to allow other tags to be entered then disabling only shows the tags and not the input box above. But that means that you get the border at the top of the tags field if you don't hide it when it's disabled. I should have included a screenshot:

Screenshot 2023-06-02 at 12 51 13 AM

zepfietje:
Screenshot make it obvious right away, thanks!

PR #6688, opened 2023-06-04
Add modal cancel action label method

archilex:
There is a method to set the label of the submit button, but not the cancel button. This PR adds that method. 

While you could set the label globally in the lang file, there may be instances where your modal allows you to perform actions that don't need to be submitted or cancelled. In those cases being able to change the "cancel" button to just say "close" would be better. 

And while there is a "X" button and you can click outside to close, some users might prefer a big obvious "close" button.


danharrin:
Thanks

PR #6692, opened 2023-06-05
Fix issue with page scroll up to the top after change value of any Fields or Create,Save button Inside Tab Component.

ibrahimBougaoua:
# This issuse appear when i use tab.

( Select Field,Create,Save,... ) whene changing or editing any data in the input field the page view goes up to the start of the page . i want to prevent the page view from changing after changing the data entry on the input field.

**Solution** : I added wire:ignore.self to the parent div of **tab** component,becouse wire:ignore.self  only ignores the DOM element it is attached to,and all children element and fields work perfectly.

### DIFF
--- a/packages/forms/resources/views/components/tabs.blade.php
+++ b/packages/forms/resources/views/components/tabs.blade.php
@@ -1,4 +1,4 @@
-<div
+<div wire:ignore.self
     x-data="{
 
         tab: null,
--- a/packages/forms/resources/views/components/tabs.blade.php
+++ b/packages/forms/resources/views/components/tabs.blade.php
@@ -1,4 +1,4 @@
-<div wire:ignore.self
+<div
     x-data="{
 
         tab: null,
@@ -33,6 +33,7 @@
     ]) }}
     {{ $getExtraAlpineAttributeBag() }}
     wire:key="{{ $this->id }}.{{ $getStatePath() }}.{{ \Filament\Forms\Components\Tabs::class }}.container"
+    wire:ignore.self
 >
     <input
         type="hidden"


packages/forms/resources/views/components/tabs.blade.php
```diff
@@ -1,4 +1,4 @@
-<div
+<div wire:ignore.self
```
zepfietje: Please put this on its own line after `wire:key`. 🙂

zepfietje (changes requested):
Mark as ready for review when done!

packages/forms/resources/views/components/tabs.blade.php
ibrahimBougaoua: > Please put this on its own line after `wire:key`. 🙂

it's done

zepfietje: Yes, I noticed, thanks!

ibrahimBougaoua: Why this issue appear " Some checks were not successful : fix-code-style / fix-code-style (pull_request) ".

zepfietje: Unrelated to this PR.

ibrahimBougaoua: okey,thanks.

danharrin:
Thanks

rafayrty:
The following issue also occurs on the wizard component of the filamentphp project by adding wire:ignore.self to wizard.blade.php this issue can be resolved for it as well. @danharrin

danharrin:
Can you PR it then?

rafayrty:
yes i will add a PR.

PR #6696, opened 2023-06-05
Allow action group trigger to be full button

oddvalue:

### DIFF
--- a/packages/admin/docs/02-resources/02-listing-records.md
+++ b/packages/admin/docs/02-resources/02-listing-records.md
@@ -134,6 +134,31 @@ public static function table(Table $table): Table
 }
 ```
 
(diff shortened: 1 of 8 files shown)


danharrin:
Hi, thanks but this has been done in v3 already with a very different implementation, and I would really rather not maintain 2 differently built versions of it. Hope you understand.

oddvalue:
@danharrin yep, no worries, totally understand. Would you consider merging if I just ported the implementation from v3?

danharrin:
I feel like it would be quite difficult and add quite alot of complexity, since also we restructured internally in v3 a lot to make it much easier to implement the feature. For example, there is now only 1 group view. For that reason, I dont think its really worth it

PR #6697, opened 2023-06-05
extract footer logo to its own view

saade:

This makes easister to customize the footer logo without the need to use render hooks. This PR does not affect the current behavior.


danharrin:
I'm going to implement this in v3, because ideally we would move this to `<x-filament::footer.logo>`, but then for consistency we would move the main view to `footer/index.blade.php`, and that would be a breaking change.

danharrin:
Actually in v3 ATM we dont even have a footer logo

PR #6706, opened 2023-06-06
fix: custom navigation does not respect policies

MostafaHassan1:
Current behavior: When using the custom navigation builder any resource registered will still appear even if the viewAny policy function is returning false 

Changes: Added canViewAny function to the resource's getNavigationItems to fix this issue


danharrin:
This is a breaking change, you just need need to check `::shouldRegisterNavigation()` before merging them into your navigation

PR #6707, opened 2023-06-06
Support starting on last step of Wizard

john-inkesta:
By using array `.at()`, you can use `startOnStep(0)` to start on the last step of the Wizard. This is useful in Workflow scenario's where the Wizard is growing dynamically at every "step" in the workflow.


danharrin:
Thanks

PR #6717, opened 2023-06-07
Fix Display required label in Safari

bnzo:
For some weirdly reason, when you have a required field label containing an hyphen character, the label is displayed with a carrier return in Safari:

Screenshot 2023-06-07 at 16 02 54

This quick fix will prevent this UI behaviour:

Screenshot 2023-06-07 at 16 03 04

### DIFF
--- a/packages/forms/resources/views/components/field-wrapper/label.blade.php
+++ b/packages/forms/resources/views/components/field-wrapper/label.blade.php
@@ -21,10 +21,10 @@
     >
         {{-- Deliberately poor formatting to ensure that the asterisk sticks to the final word in the label. --}}
         {{ $slot }}@if ($required)<sup
-                    @class([
-                        'text-danger-700 font-medium whitespace-nowrap',
-                        'dark:text-danger-400' => config('forms.dark_mode'),
-                    ])>
+                @class([
+                    'text-danger-700 font-medium whitespace-nowrap',
+                    'dark:text-danger-400' => config('forms.dark_mode'),
+                ])>
                 *
             </sup>
         @endif
--- a/packages/forms/resources/views/components/field-wrapper/label.blade.php
+++ b/packages/forms/resources/views/components/field-wrapper/label.blade.php
@@ -6,7 +6,8 @@
 ])
 
 <label
-    {{ $attributes->class(['filament-forms-field-wrapper-label inline-flex items-center space-x-3 rtl:space-x-reverse']) }}>
+    {{ $attributes->class(['filament-forms-field-wrapper-label inline-flex items-center space-x-3 rtl:space-x-reverse']) }}
+    >
     {{ $prefix }}
 
     <span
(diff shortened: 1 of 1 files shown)


packages/forms/resources/views/components/field-wrapper/label.blade.php
```diff
@@ -20,7 +20,8 @@
         ])
     >
         {{-- Deliberately poor formatting to ensure that the asterisk sticks to the final word in the label. --}}
-        {{ $slot }}@if ($required)<span class="whitespace-nowrap">
+        {{ $slot }}@if ($required)
+            <span class="whitespace-nowrap">
```
zepfietje: Could you try removing this wrapping `span` and moving `<sup>` onto the same line as `@if`? I recently noticed the asterisk doesn't stick to the final word anymore.

zepfietje (changes requested):

bnzo:
I made basic label tests and it seems displayed correctly, without the span, I kept the `whitespace-nowrap` into the `<sup>` tag

zepfietje:
Thanks!
Could you try if it still works without the `whitespace-nowrap` class? I believe it did when I tested.

Also the indentation needs to be reduced a little.

bnzo:
It works with `whitespace-nowrap` class attribute indeed. But if the label is the exact size of the input, the required `*` is displayed with a carrier return. This is a rare case but that is why I kept it. Indentation is fixed, I try to respect it and not follow my linter 👍🏼

packages/forms/resources/views/components/field-wrapper/label.blade.php
```diff
@@ -20,16 +20,14 @@
...
-                    ])
-                >
-                    *
-                </sup>
-            </span>
+        {{ $slot }}@if ($required)<sup
+                 @class([
+                     'text-danger-700 font-medium whitespace-nowrap',
+                     'dark:text-danger-400' => config('forms.dark_mode'),
+                 ])
```
zepfietje:
```suggestion
             @class([
                 'text-danger-700 font-medium whitespace-nowrap',
                 'dark:text-danger-400' => config('forms.dark_mode'),
             ])
```

packages/forms/resources/views/components/field-wrapper/label.blade.php
```diff
@@ -20,16 +20,14 @@
...
-                    ])
-                >
-                    *
-                </sup>
-            </span>
+        {{ $slot }}@if ($required)<sup
+             @class([
+                 'text-danger-700 font-medium whitespace-nowrap',
+                 'dark:text-danger-400' => config('forms.dark_mode'),
+             ])
```
zepfietje:
```suggestion
                @class([
                    'text-danger-700 font-medium whitespace-nowrap',
                    'dark:text-danger-400' => config('forms.dark_mode'),
                ])
```

zepfietje:
Ah I see why that class is still required indeed.
Thanks for having a look, @bnzo!

This should be ready for testing/merging, @danharrin.

elmergustavo:
Good, thanks 💚

PR #6727, opened 2023-06-08
[Feature] Auto hide parent component when children are hidden

Jehizkia:
TLDR: Adds a new feature to the CanBeHiddenConcern which hides the parent component if all child components are hidden

# Description
This feature addresses an issue where an empty example component would be shown when all children of a Section component were hidden. For example when the `autoHideIfChildrenHidden()` function is called, the Section component will automatically check if any of its children are hidden. If all children are hidden, the Section component will be hidden as well. If any children are visible, the Section component will be shown along with its children.

Without autoHideIfChildrenHidden():
Screenshot 2023-06-08 at 16 40 20

### Usage

```php
         Forms\Components\Section::make('Pricing')
                            ->schema([
                                Forms\Components\TextInput::make('price')
                                    ->numeric()
                                    ->hidden(() => // some condition)
                                    ->rules(['regex:/^\d{1,6}(\.\d{0,2})?$/'])
                                    ->required(),

                                Forms\Components\TextInput::make('old_price')
                                    ->label('Compare at price')
                                    ->hidden(() => // a different condition)
                                    ->numeric()
                                    ->rules(['regex:/^\d{1,6}(\.\d{0,2})?$/'])
                                    ->required(),


... a lot more components with conditions



                            ])
                            ->hideIfChildrenHidden()
                            ->columns(2),
```

### Why?
I encountered a problem where my ﻿hidden functions were becoming very large for my ﻿Section components, leading to a lot of duplicate code.

### DIFF
--- a/packages/forms/src/Components/Concerns/CanBeHidden.php
+++ b/packages/forms/src/Components/Concerns/CanBeHidden.php
@@ -3,6 +3,7 @@
 namespace Filament\Forms\Components\Concerns;
 
 use Closure;
+use Filament\Forms\Components\Component;
 use Filament\Forms\Contracts\HasForms;
 use Illuminate\Support\Arr;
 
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks

PR #6729, opened 2023-06-09
reset all (from x button) fixed

### LINKED ISSUE #6722: problem after reset all with toggle filter
### Package

filament/filament

### Package Version

latest

### Laravel Version

latest

### Livewire Version

latest

### PHP Version

8.1

### Problem description

When reset the filter (reset all) then error comes like below video

[Screencast from 06-08-2023 12:11:40 PM.webm](https://github.com/filamentphp/filament/assets/28250303/0bd60fbf-1424-42a5-9647-8db32b07437e)

This happen inside `vendor\filament\tables\src\Concerns\HasFilters.php` in `removeTableFilters()`
If replace this `$this->updatedTableFilters();` with `$this->resetTableFiltersForm();` then this problem is gone

### Expected behavior

Not clear all toggle filter

### Steps to reproduce

clone, install, migrate, seed
goto users.. and edit one or two user (mark it as admin)
Filter by country, not toggle and then reset all (in remove all filter `X` button) inside dropdown it's working

### Reproduction repository

https://github.com/Valpuia/filament-filter-bug

### Relevant log output

```shell
output in video
```

valpuia:
No new function

Before:
When clicking reset all (X) button, then it remove all filter and only show toggle filter (but not visible what it filter)
[Screencast from 06-09-2023 09:21:54 AM.webm](https://github.com/filamentphp/filament/assets/28250303/4f40105c-fc82-44cc-ae96-f6f68fb30544)

After:
[Screencast from 06-09-2023 09:22:18 AM.webm](https://github.com/filamentphp/filament/assets/28250303/6b5f79f5-37da-413e-bed9-84747c49dbe0)

This fixed #6722

### DIFF
--- a/packages/tables/src/Concerns/HasFilters.php
+++ b/packages/tables/src/Concerns/HasFilters.php
@@ -164,7 +164,7 @@ trait HasFilters
         return $name::getDefaultName();
     }
 
-    protected function getTableFiltersFormColumns(): int|array
+    protected function getTableFiltersFormColumns(): int | array
     {
         return match ($this->getTableFiltersLayout()) {
             Layout::AboveContent, Layout::BelowContent => [
--- a/packages/tables/src/Concerns/HasFilters.php
+++ b/packages/tables/src/Concerns/HasFilters.php
@@ -105,11 +105,12 @@ trait HasFilters
             $field->state(match (true) {
                 is_array($state) => [],
                 $state === true => false,
+                $state === false => false,
                 default => null,
             });
         }
 
-        $this->resetTableFiltersForm();
+        $this->updatedTableFilters();
     }
 
     public function resetTableFiltersForm(): void
@@ -163,7 +164,7 @@ trait HasFilters
         return $name::getDefaultName();
     }
 
-    protected function getTableFiltersFormColumns(): int | array
+    protected function getTableFiltersFormColumns(): int|array
     {
         return match ($this->getTableFiltersLayout()) {
             Layout::AboveContent, Layout::BelowContent => [
--- a/packages/tables/src/Concerns/HasFilters.php
+++ b/packages/tables/src/Concerns/HasFilters.php
@@ -109,7 +109,7 @@ trait HasFilters
             });
         }
 
-        $this->updatedTableFilters();
+        $this->resetTableFiltersForm();
     }
 
     public function resetTableFiltersForm(): void


archilex:
@Valpuia The reason there are two separate methods `removeTableFilters` and `resetTableFiltersForm` is that they do two separate things. `removeTableFilters` removes all the filters _independent_ of the default values, while `resetTableFilter` resets everything to the defaults. 

The reason the difference is important is because in your example, if you use the `default()` method on the toggle Filter and then test your PR, when you click the Remove Table Filters button, it doesn't actually remove the filter since its default state is “on”.

To fix the problem you discovered, try adding this to the `removeTableFilters` method:

```php
$field->state(match (true) {
      is_array($state) => [],
      $state === true => false,
      $state === false => false, // <- add this line
      default => null,
  });
```

This fixes the actual issue which is that the match function isn't picking up on the false state and instead sets it to null which causes the issues you found. 

And I'm glad you found this bug since it also affects some of my apps. No client has brought it to my attention yet, but I'm glad I can fix it before anyone notices! 🙏

valpuia:
> @Valpuia The reason there are two separate methods `removeTableFilters` and `resetTableFiltersForm` is that they do two separate things. `removeTableFilters` removes all the filters _independent_ of the default values, while `resetTableFilter` resets everything to the defaults.
> 
> The reason the difference is important is because in your example, if you use the `default()` method on the toggle Filter and then test your PR, when you click the Remove Table Filters button, it doesn't actually remove the filter since its default state is “on”.
> 
> To fix the problem you discovered, try adding this to the `removeTableFilters` method:
> 
> ```
> $field->state(match (true) {
>       is_array($state) => [],
>       $state === true => false,
>       $state === false => false, // <- add this line
>       default => null,
>   });
> ```
> 
> This fixes the actual issue which is that the match function isn't picking up on the false state and instead sets it to null which causes the issues you found.
> 
> And I'm glad you found this bug since it also affects some of my apps. No client has brought it to my attention yet, but I'm glad I can fix it before anyone notices! pray

Yes, now I understand the logic a little bit, and changes my PR according to your suggestions..

danharrin:
I'm 99% sure this would break ternary filters, can you check those please?

archilex:
I think you're totally right. We'll have to figure out a different solution to the problem. Maybe checking that it isn't an instance of TernaryFilter? Or is an instance of Filter? Not sure...on my phone.

valpuia:
I check again and this doesn't break anything if i recall..

But one thing I notice is the ternary itself is not working (it cannot filter) but this filter clear is not the problem..

[Ternary Filter](https://filamentphp.com/docs/2.x/tables/filters#ternary-filters) cannot filter boolean.. but clearing (with my PR) doesn't break anything..

Let me know if I'm missing something..

archilex:
I can confirm as well that this doesn't appear to break Ternary Filters when removing Filters. I tested with a normal boolean ternary filter, a nullable ternary filter, as well as using the Deleted records filter (which is just a ternay filter that has true/false/blank queries already set up).

I agree Dan that it seems like this SHOULD break ternary filters. Since ternary filters can have a false state it seems that by adding the `false` match, you'd think that when the ternary filter was false it wouldn't clear that filter, but everything does clear properly. Weird. 

Is there another combination of ternary filters we could test against? Or something else you are thinking about?

And @Valpuia regarding ternary filters not working, I'm not seeing that problem. Ternary filters are working for me. Maybe try casting your column to boolean?

valpuia:
`Tables\Filters\TernaryFilter::make('is_admin')` this filter is not working in User (model/resource), and yes I casts as boolean also.. but this same filter is working in another model/resource (in my other project :smile: )

[Screencast from 06-12-2023 10:37:54 AM.webm](https://github.com/filamentphp/filament/assets/28250303/15a08352-01a9-43ea-bf25-461b1242e45f)

valpuia:
Hi, I got email about `fix-code-style, Attempt #2: All jobs have failed` what should I do?

zepfietje:
> Hi, I got email about `fix-code-style, Attempt #2: All jobs have failed` what should I do?

Nothing. Unrelated to the code changes in this PR.

danharrin:
Thanks

PR #6737, opened 2023-06-10
Add solid background color to icon button

howdu:
Add `solid` option to control the background color of icon button.

Screenshot 2023-06-10 at 08 31 12

Screenshot 2023-06-10 at 08 31 00

```html
<x-filament::icon-button size="sm" color="primary" icon="heroicon-s-fire" solid>
<x-filament::icon-button size="sm" color="secondary" icon="heroicon-s-bell" solid>
<x-filament::icon-button size="sm" color="gray" icon="heroicon-s-bell" solid>
<x-filament::icon-button size="sm" color="success" icon="heroicon-s-check" solid>
<x-filament::icon-button size="sm" color="warning" icon="heroicon-s-bell" solid>
<x-filament::icon-button size="sm" color="danger" icon="heroicon-s-ban" solid>
<x-filament::icon-button size="sm" color="primary" icon="heroicon-s-fire">
<x-filament::icon-button size="sm" color="danger" icon="heroicon-s-trash">
```

### DIFF
--- a/packages/support/resources/views/components/icon-button.blade.php
+++ b/packages/support/resources/views/components/icon-button.blade.php
@@ -52,9 +52,11 @@
         'w-6 h-6' => $size === 'lg',
     ]);
 
-    $indicatorClasses = \Illuminate\Support\Arr::toCssClasses(array_merge([
-        'filament-icon-button-indicator absolute rounded-full text-xs inline-block w-4 h-4 -top-0.5 -right-0.5'
-        ], $solid ? [
+    $indicatorClasses = \Illuminate\Support\Arr::toCssClasses(array_merge(
+        [
+            'filament-icon-button-indicator absolute rounded-full text-xs inline-block w-4 h-4 -top-0.5 -right-0.5',
+        ],
+        $solid ? [
             'text-white',
             'bg-danger-600' => $color === 'danger',
             'bg-gray-500' => $color === 'secondary',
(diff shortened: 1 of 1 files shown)


howdu (commented):
Added filament-icon-button-size-{$size}

danharrin:
Thanks, but personally I don't think this should be an option in our design system - wdyt @zepfietje @awcodes ?

zepfietje:
> Thanks, but personally I don't think this should be an option in our design system - wdyt @zepfietje @awcodes ?

I think I agree.
It complicates things quite a bit (as can be seen from the screenshots with indicator) and I don't really see many cases where this is really needed.

Could you elaborate on use cases, @howdu?

awcodes:
Personally, I'm kinda torn on this one. I can see the need in some use cases where the UI could be more clear that the icon is a button and not just decorative. I think though with the proper filament classes ie. filament-icon-button-danger, etc that it could be managed reasonably via css alone.

danharrin:
Also, it introduces inconsistency whereby some buttons are `solid` by default and some arent, and if we fixed that, I would always have to pass `solid(false)` because I don't like how these look

howdu:
> Could you elaborate on use cases, @howdu?

I found it useful to make the icon button appear more visually clickable and differentiated from the collapse.
Screenshot 2023-06-12 at 07 01 34

zepfietje:
> I found it useful to make the icon button appear more visually clickable and differentiated from the collapse. Screenshot 2023-06-12 at 07 01 34

Okay, thanks for elaborating.

I think it's best to handle such one-of case using CSS like @awcodes said.

Adding the solid option would complicate the design system, and I think you'd usually either use solid or not solid icon buttons.

Closing this PR now, but feel free to bring up any issues you run into when customizing with CSS.

PR #6741, opened 2023-06-11
Implement prependFilters() and appendFilters() to Table

miguilimzero:

Filament Admin Panel `Table` class has methods to append and prepend `actions`, `bulkActions` and `headerActions` but does not have for `filters`. 

This pull request implements the `prependFilters()` and `appendFilters()` methods.

### Use Case

I want to create a base resource class that has defined filters and actions and do the following table construction in the final resource class:

```php
return $table
    ->appendFilters([
        //
    ])
    ->appendActions([
        //
    ])
    ->appendBulkActions([
        //
    ]);
```
However, I can't append filters at the moment so I'm limited only to:

```php
return $table
    ->appendActions([
        //
    ])
    ->appendBulkActions([
        //
    ]);
```

### DIFF
--- a/packages/admin/src/Resources/Table.php
+++ b/packages/admin/src/Resources/Table.php
@@ -131,6 +131,13 @@ class Table
         return $this;
     }
 
+    public function prependFilters(array $filters): static
+    {
+        $this->filters = array_merge($filters, $this->filters);
+
+        return $this;
+    }
+
     public function prependHeaderActions(array $actions): static
     {
         $this->headerActions = array_merge($actions, $this->headerActions);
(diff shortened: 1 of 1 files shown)


danharrin:
Hey, methods like `appendActions()` and `appendBulkActions()` are deprecated, so we won't be adding any more. Sorry. Since Filament `v2.13.0`, we haven't recommended using "default" actions and bulk actions, the user defines them all at once using the `actions()` and `bulkActions()` methods.

miguilimzero:
@danharrin Do you think sending a PR adding the @ deprecated message to these methods would be a good one?

danharrin:
If you want, but I don't think it's incredibly important in this case 👍

PR #6758, opened 2023-06-13
Fix select placeholder not showing

### LINKED ISSUE #6736: Select's placeholder disappears when updated by another select
### Package

filament/filament

### Package Version

v2

### Laravel Version

v10.13.5

### Livewire Version

_No response_

### PHP Version

8.1.14

### Problem description

The "Select an option" placeholder disappears completely when we reset it with the `$set('myComponent', null);` method in `afterStateUpdated()`

This problem comes particularly from the `searchable()` method present on the select.

https://github.com/filamentphp/filament/assets/9550130/9f5024ad-06a0-46c0-990f-5fe4763c6f68

### Expected behavior

After the component is reset by `$set('myComponent', null);`, I would like the component to return to the original value of the placeholder: "Select an option" as it does when we click on the cross in the select. For now, it remains worthless.

### Steps to reproduce

- Create a form on a resource
- Add the following code:
```
Select::make('test')
    ->options([
        'Option #1',
        'Option #2',
    ])
    ->reactive()
    ->afterStateUpdated(function(Closure $set) {
        $set('test-updated', null);
    }),
    
Select::make('test-updated')
    ->searchable()
    ->options([
        'Option Test #1',
        'Option Test #2',
    ]),
```
- Observe the value of the placeholder of the select 'Test Updated' when you change the value of the select 'Test'

OR 

You can reproduce the attached deposit on the form of the 'User' resource

### Reproduction repository

https://github.com/antoinemy/filament_bug_disappear_placeholder

### Relevant log output

_No response_

archilex:

This PR should fix issue #6736 

It seems the problem lies in watch method. `this.refreshPlaceholder()` is called at the beginnging of the watch, but then `this.select.clearStore()` is called which removes the placeholder. Adding `this.refreshPlaceholder()` after `this.select.clearStore()` fixes this issue.

You can test this by just removing `clearStore()` and the placeholder will show. I'm sure `clearStore` is needed so the easiest option seems to be to refresh the placholder after calling `clearStore()`.

Potentially we could get rid of the first `refreshPlaceholder`, but with the early return when the state is still being updated, maybe it needs to stay?

### DIFF
--- a/packages/forms/resources/js/components/select.js
+++ b/packages/forms/resources/js/components/select.js
@@ -157,6 +157,8 @@ export default (Alpine) => {
 
                         this.select.clearStore()
 
+                        this.refreshPlaceholder()
+
                         this.setChoices(choices)
 
                         if (![null, undefined, ''].includes(this.state)) {


antoinemy:
how can I upgrade this fonctionality? 
`composer update` & `php artisan filament:upgrade` didn't change anything? 
sorry I'm beginner with github and this community...

danharrin:
Wait for a new release to be made, this has not been released yet

PR #6794, opened 2023-06-16
Update Resource.php

xcalabuig:
global search not works properly when the searchable field is a json, and also it does not respect the database collation.
I mean when you have a spatie-translatable field (so it works as a json field) it ignores the database.php collation config.
(my desire case is to search words ignoring accents)

todo : how to apply this at the individual search query of each resource?

 Current result:
image
Expecting result:
image


danharrin:
Search the entire `/packages` for something distinctive like `'pgsql' => 'ilike'` and it should help you find the other instances where we do similar queries.

Also, PHPStan is failing.

PR #6800, opened 2023-06-17
Add render hooks to table header toolbar

archilex:
This PR adds several render hooks to the table header toolbar. In short it adds renderhooks:
- before and after the left hand side group of actions (reorder, bulk, group)
- before and after the right hand side group of filters (search, filter, toggle)
- one "in the middle" between those two. This renderhook isn't meant to be in the middle, but when search, filters, and toggle columns are turned off there wouldn't be a way to render a hook there, so adding this renderhook allows someone to take advantage of the `justify-between` on the parent and then render something on the far right.

Since renderhooks are only available in the Panel, i've done a check to ensure it is the panel. There might be a better way to do this, but what occured to me is to see if the `filament` helper function exists.

Of course, another approach would be to actually add the ability to call renderhooks in the tablebuilder, but that would require a much larger discussion.

Finally, I think you could make an argument for at least one additional render hook between the search and the filters/toggle column icons. That way if someone wanted to add an icon before those two they could.

Let me know and I can do that.

Dan please work your naming magic!!!

### DIFF
--- a/packages/panels/docs/09-configuration.md
+++ b/packages/panels/docs/09-configuration.md
@@ -120,6 +120,11 @@ The available hooks are as follows:
 - `resource.pages.list-records.table.end` - after the resource table
 - `resource.relation-manager.start` - before the relation manager table
 - `resource.relation-manager.end` - after the relation manager table
+- `table.header-toolbar.actions.start` - before a table's header toolbar actions
+- `table.header-toolbar.actions.end` - after a table's header toolbar actions
+- `table.header-toolbar` - between a table's header toolbar actions and filters
+- `table.header-toolbar.filters.start` - before a table's header toolbar filters
+- `table.header-toolbar.filters.end` - after a table's header toolbar filters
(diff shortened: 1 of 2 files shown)


wychoong:
Understand this is a small change, but ultimately what is the use case? To render html or add action buttons?

archilex:
@wychoong correct, to be able to render additional action buttons, legends, html, etc. in that area. It's from a discussion that occurred in discord.

MACscr:
I really want this in the table builder too. I was part of that discussion on discord. =)

archilex:
We probably need to hold off on this as I discovered that this, along with the other RM render hooks, have an issue when used with GroupedRelationManagers. The discussion is here on Discord: https://discord.com/channels/883083792112300104/1122228435318734899

Trying to figure out a fix.

danharrin:
Beta will make render hooks available outside of the admin panel

PR #6801, opened 2023-06-17
Feat: `CanBeCopied` copyText

Z3d0X:

### DIFF
--- a/packages/tables/resources/views/columns/badge-column.blade.php
+++ b/packages/tables/resources/views/columns/badge-column.blade.php
@@ -52,7 +52,7 @@
             <span
                 @if ($isCopyable)
                     x-on:click="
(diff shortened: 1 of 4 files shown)


danharrin:
How about calling this feature "copyable state" (`copyableState()`, `getCopyableState()`), and also refactoring so that if `$this->copyableState` is null, we return `$this->getState()` from that `getCopyableState()` method. Then we can get rid of the `??` in the views.

Z3d0X:
Can return type on `getCopyableState(): ?string`? since we are now doing  `?? $this->getState()`. Do u think we should remove the return type?

danharrin:
The state could be null so I think the return type is fine

danharrin:
Thanks

PR #6807, opened 2023-06-18
Add host and port to broadcasting

francoism90:

I don't think documentation has to be changed for this, if so, I'm happy to sync. :)

The reason for adding wsHost + wsPort/wssPort, is when someone uses a Nginx proxy for sockets (like we do).
It could also be someone using the docker container host and listening on a different url.


danharrin:
Thanks

PR #6808, opened 2023-06-19
Add CSS hook class to user menu

valpuia:

In my use case, I want to display `theme-toggle` between search and user-menu, I can add using render hook.. but I want to hide the toggle inside user-menu. or display the toggle if screen = xx,

This PR add one class `filament-theme-toggle` in `div`. it helps when trying to add custom css


ArtMin96:
Hi. Can you move that div inside the if statement?

valpuia:
> Hi. Can you move that div inside the if statement?

I only add  `class="filament-theme-toggle"` the 'div' is as it is.. not sure how inside "if statement" will impact

PR #6809, opened 2023-06-19
Added support for before policy filter

linaspasv:
This PR checks before policy filter and triggers (if available). See more at: https://laravel.com/docs/10.x/authorization#policy-filters

```php

use App\Models\User;
use Illuminate\Auth\Access\HandlesAuthorization;

class CompanyPolicy
{
    use HandlesAuthorization;

    public function viewAny(): bool
    {
        return false;
    }

    public function before(User $user, $ability): ?bool
    {
        // check for any condition and return bool / or null to pass authorization further the chain

        return null;
    }
}
```

### DIFF
--- a/packages/admin/src/Resources/Resource.php
+++ b/packages/admin/src/Resources/Resource.php
@@ -135,6 +135,10 @@ class Resource
             return true;
         }
 
+        if (method_exists($policy, 'before') && is_bool($condition = $policy->before($user, $action))) {
+            return $condition;
+        }
+
         if (! method_exists($policy, $action)) {
             return true;
         }


linaspasv:
Hm, on the other hand I see that global before / after filters are not getting triggered with or without the PR.

What is the purpose of this code?

```php
        if (! method_exists($policy, $action)) {
            return true;
        }
```

Maybe if policy exists we should just allow to perform a check and default to false, instead of trying to check for existing method. This could be a breaking change for some but would more closely resemble on how Laravel works with Policies. :-)

danharrin:
Thanks!

packages/admin/src/Resources/Resource.php
```diff
@@ -135,6 +135,10 @@ public static function can(string $action, ?Model $record = null): bool
             return true;
         }
 
+        if (method_exists($policy, 'before') && is_bool($condition = $policy->before($user, $action))) {
```
zepfietje, line 138: What if the `before` method returns an authorization response instead of a boolean, @danharrin @linaspasv?

danharrin: We don't support those inside the method either ATM, so this is consistent for now

zepfietje: Don't support in what way? I bet it still manages authorization correctly, but just doesn't throw the same status code again?

This is different though, since if `before` is **not** authorized, by returning a specific status code, we now actually **do** authorize since the `if` condition evaluates to `false` because of `is_bool`.

danharrin: Reverting until 3.x

PR #6810, opened 2023-06-19
Add slot to selection indicator

howdu:
Useful when you need to add buttons inside the selection indicator.

Screenshot 2023-06-19 at 09 36 47

Wrapped small screen
Screenshot 2023-06-19 at 09 49 15


danharrin:
I think this should be a named slot like `end` as I don't think it's immediately clear that the content of the indicator should be aligned to the right? I'd expect it to be at the start.

So `{{ $end }}` and `@props(['end' => null])`

howdu:
No worries I've moved the standard `$slot` to the beginning.

howdu:
Also added ability to change the property name `selectedRecords`. Needed when you have mutiple selection-indicators on one page.

PR #6824, opened 2023-06-20
fix get tags with no types.

Saifallak:
trying to get tags with no type specified (means get all with type and without), expected to get all tags but with this, it searches with no-tags.


Saifallak:
@danharrin can you help with those phpstan errors?

danharrin:
The current behaviour is correct IMO and this would be a breaking change. Extend the SpatieTagsColumn in your own project and override this behaviour if you wish.

Saifallak:
maybe add `'*'` to pass to `->type()` ? to do this behaviour ?
because i think the normal is to get all tags
`->type` is there for filtering.


anyway, thanks for your efforts, i will extend it in my project for now.

Saifallak:
hi @danharrin i saw this PR #7374 and it applied to form, if it was accepted for v3, maybe this PR can be accepted for V3 too?

PR #6837, opened 2023-06-22
Fix deprecated string interpolation

joelbutcher:
Fixes the following deprecation warning in production:

```
PHP Deprecated:  Using ${expr} (variable variables) in strings is deprecated, use {${expr}} instead in /app/vendor/filament/support/src/Concerns/EvaluatesClosures.php on line 95
```

- [ ] ~New functionality has been documented or existing documentation has been updated to reflect changes.~
- [ ] ~Visual changes are explained in the PR description using a screenshot/recording of before and after.~


joelbutcher:
Not sure why phpstan is complaining here – I've not touched that area of code

PR #6848, opened 2023-06-24
Remove secondary color

zepfietje:

Secondary color is optional, so no need to keep the key since we support adding custom named colors now anyway.


zepfietje:
This is ready to be merged, @danharrin.

danharrin:
I need to think about the BC consequences before I do

zepfietje:
Sure.

danharrin:
Since we have a rector for this, I think we're ok?

zepfietje:
> Since we have a rector for this, I think we're ok?

Yeah, the new `secondary` color concept was only something we added in v3.

PR #6852, opened 2023-06-26
Execute custom function after the table reorder

Morry981:
Hi, i'm creating this PR for my friend's request described [here](https://github.com/filamentphp/filament/discussions/6619#discussion-5235550).

I simply added a second not required parameter to `reorderable` function.
In the function you can now, in our case for example, execute something that is not executed because events are not fired.

Let me know for any questions and fixes.

PS. i tried to write it also as string parameter but i don't know well how it works... so if needed please fix and add it
```php
public function getReorderFunction(): ?Closure
{
    $function = $this->reorderFunction;

    if (is_string($function)) {
        $function = Closure::fromCallable(class_exists($function) ? $function : [$this, $function]);
    }

    return $function;
}
```

### DIFF
--- a/packages/admin/src/Resources/Pages/ListRecords.php
+++ b/packages/admin/src/Resources/Pages/ListRecords.php
@@ -395,6 +395,11 @@ class ListRecords extends Page implements Tables\Contracts\HasTable
         return filled($this->getTableReorderColumn()) && static::getResource()::canReorder();
     }
 
+    protected function getTableReorderFunction(): ?Closure
+    {
+        return $this->getResourceTable()->getReorderFunction();
+    }
+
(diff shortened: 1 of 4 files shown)


danharrin:
Hi, we are no longer merging large new features into 2.x, so I will implement this in 3.x. Thanks!

danharrin:
Actually - it's already quite easy to do this in your own app by overriding a method, so I don't think I need to make it a feature:

```php
public function reorderTable(array $order): void
{
    parent::reorderTable($order);

    // Do your thing
}
```

Morry981:
Oh right, i didn't think of that solution.
Thanks! Added the override in Pages/ListModel.php file and it work perfectly fine!

PR #6855, opened 2023-06-27
[upg3] Update versions in README badges

Jubeki:


zepfietje:
Thanks for bringing this up!

I've committed this change to 3.x as this PR was incomplete and targeting the wrong branch.

PR #6878, opened 2023-06-29
Update MakeUserCommand.php

arkoe:
Remove stupid dot at end (ctrl + click from terminal includes the dot, causing 404)

image


zepfietje:
Is this really a Filament issue though? Not sure if all terminals have this bug.

PR #6886, opened 2023-06-29
BugFix: Broken Mask Options

JacobBennett:
In commit 5deaa867494df811aa5d8e2c6b82f7f32e396608 when using Prettier to update the `text-input.blade.php`, a bug was introduced.

When using the mask option on a TextInput, the field will fail to apply the mask and will throw this error in the console. Re-introduction of the parens will fix this bug.

Errant Commit:
https://github.com/filamentphp/filament/commit/5deaa867494df811aa5d8e2c6b82f7f32e396608#r120141338

Error in Console:
image


danharrin:
Thanks

PR #6891, opened 2023-06-30
Navigation Item - Add new method to check Gate for visibility

salahhusa9:

Old:
```php
use Filament\Navigation\NavigationItem;
 
NavigationItem::make('Analytics')
    ->visible(auth()->user()->can('view-analytics'))
    // or
    ->hidden(! auth()->user()->can('view-analytics'));
```
New:
```php
use Filament\Navigation\NavigationItem;
 
NavigationItem::make('Analytics')
    ->visibleIfCan('view-analytics')
    // or
    ->hiddenIfCan('view-analytics');
```

### DIFF
--- a/packages/admin/src/Navigation/NavigationItem.php
+++ b/packages/admin/src/Navigation/NavigationItem.php
@@ -73,9 +73,23 @@ class NavigationItem
         return $this;
     }
 
+    public function visibleIfCan(string $ability): static
+    {
+        $this->visible(fn () => auth()->user()?->can($ability));
+
+        return $this;
+    }
+
     public function hidden(bool | Closure $condition = true): static
     {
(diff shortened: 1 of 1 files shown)


danharrin:
Thank you, but I'm not going to merge this. I don't think the authorization logic should be defined on the navigation item, it should be defined on the page class etc, otherwise it's dangerous because then people will forget to actually authorise their pages too.

salahhusa9:
I didn't add a new delegation logic (it already exists) what I did is add a function whose purpose is a script shortcut

danharrin:
Yeah, but adding such a function would encourage users to do this, and they shouldnt. They should add authorization to their page and then call it from their nav item.

PR #6901, opened 2023-07-02
Add scrollableContent to modal

howdu:
Added `scrollContent` to allow for the inner modal content to be scrollable so that it works the same as screen/slideOver but uses the variable modal widths.

### DIFF
--- a/packages/support/resources/views/components/modal/index.blade.php
+++ b/packages/support/resources/views/components/modal/index.blade.php
@@ -13,6 +13,7 @@
     'hrComponent' => 'filament-support::hr',
     'id' => null,
     'openEventName' => 'open-modal',
+    'scrollContent' => false,
     'slideOver' => false,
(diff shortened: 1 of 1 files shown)


zepfietje:
@danharrin, you didn't port this over to 3.x, did you?
If you did, we don't need it because of the new sticky footer actions api we added there.

danharrin:
I did, but I can revert it

zepfietje:
I'd say revert it.
Unless you see any reason it needs to be in 3.x too?
We redid the entire scrolling stuff IIRC.

Also mark this new API as deprecated on 2.x right away then. 😅

danharrin:
I reverted only in v3

zepfietje:
> I reverted only in v3

Yeah, but as the API will not exist in 3.x, might as well mark it as deprecated in 2.x now.

PR #6914, opened 2023-07-04
Global search result scroll bar visibility

valpuia:

Before: When search anything we can see scroll bar by default (also in empty result)
before

After: Only when search result is scrollable, the scroll bar visible
after


danharrin:
Thanks

PR #6938, opened 2023-07-06
fix: Remove viewAny permission checks on create, edit & view

jeanremy:

When we want to restrict access to only a view page, Filament checks for the `viewAny` permission on the Resource. Let's say we have a user, and we only want to show him his page and not the list page. We can use Laravel policies to check that:
```
class UserPolicy
{
    public function viewAny(User $user): bool
    {
        return false;
    }
    
    public function view(User $user, User $other): bool
    {
        return $user->id === $other->id;
    }
```
With this policy, the `viewAny`is checked before the `view`, leading to a 403 instead of the view page.

This PR aims at correcting this issue, if it is a real one. I really appreciate your input on this behavior.

### DIFF
--- a/packages/admin/src/Resources/Pages/CreateRecord.php
+++ b/packages/admin/src/Resources/Pages/CreateRecord.php
@@ -43,8 +43,6 @@ class CreateRecord extends Page implements HasFormActions
 
     protected function authorizeAccess(): void
     {
-        static::authorizeResourceAccess();
-
         abort_unless(static::getResource()::canCreate(), 403);
     }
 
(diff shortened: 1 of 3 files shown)


zepfietje:
I'm pretty sure the current behavior is by design.

@danharrin, reopen if not.

danharrin:
Yeah this is a deliberate decision. You're thinking of "view any" more like "view all", when it is not

jeanremy:
Thank you for your answers ! I really appreciate.

To get a better understanding of the issue, why is the `viewAny` verification for, if the `view` is checked just after ?
And would it be possible to give access to a view or edit page, but not to a list ?

Thanks for your inputs !

danharrin:
Let me guess, you're making a profile page? You should not use a resource for that - you can create a custom Profile page (make:filament-page) and then use the form builder to add the form to it. No need for a resource that covers both use cases

jeanremy:
That's indeed my use case. Thanks for the advice and for taking the time to answer !

PR #6949, opened 2023-07-06
Feat: Allow style attributes in html sanitizer config

awcodes:
Adds style as allowed attributes to the html sanitizer config to make it work with the new color system when using 'Htmlables'

### DIFF
--- a/packages/support/config/html-sanitizer.php
+++ b/packages/support/config/html-sanitizer.php
@@ -21,11 +21,11 @@ return [
      */
     'tags' => [
         'abbr' => [
-            'allowed_attributes' => ['class'],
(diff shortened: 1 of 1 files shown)


danharrin:
Refactored to use a different library for sanitising, how is it?

awcodes:
I thought you did too, but when I pulled the lasted 3.x down it still had the other one. Maybe something just got messed up in local files. Sorry for the PR. I'll double check the new one in a couple hours.

danharrin:
I did it in upg3 but this morning ported it over to 3.x

awcodes:
Ah ok. Will let you know. Thanks.

PR #6960, opened 2023-07-08
Update 07-relation-managers.md

VasasA:
Attaching with pivot attributes: Instead of using $action->getRecordSelect(), you can create a custom select field.

### DIFF
--- a/packages/admin/docs/02-resources/07-relation-managers.md
+++ b/packages/admin/docs/02-resources/07-relation-managers.md
@@ -467,6 +467,22 @@ AttachAction::make()
 
 In this example, `$action->getRecordSelect()` outputs the select field to pick the record to attach. The `role` text input is then saved to the pivot table's `role` column.
 
+Instead of using `$action->getRecordSelect()`, you can create a custom select field. The parameter of the `make()` must be `'recordId'`:
+```php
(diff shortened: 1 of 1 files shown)


danharrin:
You could just chain config onto `$action->getRecordSelect()->label('Test')`? Then you don't need to worry about the field name.

zepfietje:
I don't think we need a separate section for this, as above-mentioned approach should be quite obvious.

Reopen if you disagree, @danharrin.

VasasA:
@danharrin  @zepfietje 

This is weird: I have tested: the `$action->getRecordSelect()->label('Test')` doesn't work.
These work: `->options()`, `->required()`, `->searchable()`

danharrin:
What about it doesnt work?

VasasA:
@danharrin 
The label isn't displayed. The select field hasn't got label.

danharrin:
Ah, try `->disableLabel(false)`

VasasA:
👍  It works, thank You very much!
Should we mention this solution in the documentation?
(The `disableLabel` does not give results when searching in the documentation of filamentphp.com.)

danharrin:
`disableLabel()` could be added to the Fields documentation aside the main label docs

PR #7036, opened 2023-07-18
Add chunk uploads support to FileUpload

sawirricardo:
Chunk Uploads can be useful when the end users are required to upload files bigger than the server's capacity limit.

This PR adds chunk upload functionality, using FilePond's chunking configurations

Behind the scenes, in `uploadUsing()` `file-upload.blade.php`, the Javascript will handle slicing the file and based on the configration options, and use the chunk size. Then, it will use Livewire's `uploadMultiple` and call `processChunks()`, which does the heavy lifting to compile the file into one piece.

- [v] Changes have been thoroughly tested to not break existing functionality.
- [v] New functionality has been documented or existing documentation has been updated to reflect changes.
- [v] Visual changes are explained in the PR description using a screenshot/recording of before and after.

It's v3 compatible. Thoughts?


ralphjsmit:
Love this PR!

diana-mujoiu:
It's very useful

danharrin:
Please retarget this at the 3.x branch and test it with Livewire 3.

sawirricardo:
just find out that this will result in error if livewire is using S3 to store temporary file upload

zepfietje:
Could you elaborate on the exact problem, @sawirricardo?

sawirricardo:
I've digged into Livewire's core and found out that Livewire's S3 doesn't support multiple file upload.

https://github.com/livewire/livewire/blob/main/src/Features/SupportFileUploads/S3DoesntSupportMultipleFileUploads.php

So this PR effectively only works with local driver set on livewire. (can be documented)

However, I also think this should also be addressed in Livewire's to 'natively' upload the file using s3's multipart upload.

Thoughts?

On a side note, I'm still experimenting again, but have other works to do first. WIll get back on this. Happy to hear other suggestions.

sawirricardo:
Update: Making a PR into Livewire's (https://github.com/livewire/livewire/pull/5959). Still a draft, hopefully can move the workloads to Livewire's

alex-mottier:
Love this PR ! I'm developping corporate applications and sometimes the CSV files are way bigger than some MB. Chunking thoses files will simplify my work a lot !

danharrin:
Hi, please reopen this when Livewire is ready for it 👍

peter-mw:
Any update for support of chunked uploads ?

zerkxubas:
Any update for support of chunk uploads ?

ralphjsmit:
If you need chunked uploads, I added it specifically to my Upload plugin as a custom-built implementation: https://filamentphp.com/plugins/ralphjsmit-upload-pro. Works with all type of disks, fro S3 disks it even uses Multipart upload instead. If you have any questions always feel free to let me know.

zerkxubas:
@ralphjsmit Thanks for the information, But the pricing for your package is too much higher even for a single project.

ralphjsmit:
@zerkxubas No problem, it's not a simple thing to implement securely though. I do offer some PPP discount if you're in such a country (email me then).

PR #7040, opened 2023-07-18
Add note to documentation regarding route order of pages

adamlast:
If the custom page route is added at the bottom of `getPages()` then it is likely that the standard `view` route will match but give a 404 message

- [Y] Changes have been thoroughly tested to not break existing functionality.
- [Y] New functionality has been documented or existing documentation has been updated to reflect changes.
- [Y] Visual changes are explained in the PR description using a screenshot/recording of before and after.

### DIFF
--- a/packages/admin/docs/02-resources/10-custom-pages.md
+++ b/packages/admin/docs/02-resources/10-custom-pages.md
@@ -21,7 +21,8 @@ public static function getPages(): array
     ];
 }
 ```
-Note: Ensure that the new route you define in `getPages()` is defined before any routes with wildcards that will match your new url.
+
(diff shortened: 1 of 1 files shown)


zepfietje:
Thanks for the PR, @adamlast!
I'm not sure if we should add this to our docs though, as it's just a Laravel thing.

adamlast:
> Thanks for the PR, @adamlast! I'm not sure if we should add this to our docs though, as it's just a Laravel thing. What do you think, @danharrin?

I thought it would be useful as putting the new route at the bottom of the array returned by `getPages()` is default behaviour for many people, it caught me out today, and it has caught others out before: [https://github.com/filamentphp/filament/discussions/6197](https://github.com/filamentphp/filament/discussions/6197)

PR #7044, opened 2023-07-19
Add passthrough for `mountedFormComponentActionArguments`

juliomotol:

This PR aims to resolve an issue in Form Component Actions where the `arguments()` is not persisted in subsequent requests.

The proposed solution implements a passthrough to the livewire component's `$mountedFormComponentActionArguments`.

### DIFF
--- a/packages/forms/src/Components/Actions/Concerns/BelongsToComponent.php
+++ b/packages/forms/src/Components/Actions/Concerns/BelongsToComponent.php
@@ -25,4 +25,9 @@ trait BelongsToComponent
     {
         return $this->getComponent()->getLivewire();
     }
+
+    public function getArguments(): array
+    {
+        return $this->getLivewire()->mountedFormComponentActionArguments;
+    }
 }
--- a/packages/forms/src/Components/Actions/Concerns/BelongsToComponent.php
+++ b/packages/forms/src/Components/Actions/Concerns/BelongsToComponent.php
@@ -28,6 +28,6 @@ trait BelongsToComponent
 
     public function getArguments(): array
     {
-        return $this->getLivewire()->mountedFormComponentActionArguments;
+        return array_merge($this->getLivewire()->mountedFormComponentActionArguments, parent::getArguments());
     }
 }


juliomotol:
I noticed that only form components has this behaviour where the initial arguments are set upon calling the mounted action.

Should I apply #5957 on `Filament\Pages\Concerns\HasActions`, `Filament\Tables\Concerns\HasActions`, and `Filament\Tables\Concerns\HasBulkActions` on this PR or on a separate one?

danharrin:
Please provide more context about the issue because I don't think this is the correct fix, what were you trying to do when you hit this?

juliomotol:
We built a custom form component for a "Tree" structured list that resembles a `Reapeater`. But instead of showing the item's field, we show it in a modal through the use of `Actions`.

### `Tree` component
```php
class Tree extends Field
{
    {
        parent::setUp();

        $this->registerActions([TreeFormAction::make()]);

        $this->registerListeners([
            'tree::createItem' => [
                function (self $component, string $statePath): void {
                    if ( ! Str::startsWith($statePath, $component->getStatePath())) {
                        return;
                    }

                    $component->getLivewire()->mountFormComponentAction(
                        $component->getStatePath(),
                        'tree-form',
                        ['activeTreeStatePath' => "{$statePath}." . (string) Str::uuid()]
                    );
                },
            ],
            'tree::editItem' => [
                function (self $component, string $statePath): void {
                    if ( ! Str::startsWith($statePath, $component->getStatePath())) {
                        return;
                    }

                    $component->getLivewire()->mountFormComponentAction(
                        $component->getStatePath(),
                        'tree-form',
                        ['activeTreeStatePath' => $statePath]
                    );
                },
            ],
        ]);
    }
}
```

### `TreeFormAction`
```php
class TreeFormAction extends Action
{
    public static function getDefaultName(): ?string
    {
        return 'tree-form';
    }

    {
        parent::setUp();

        $this->modalHeading(function ($component, self $action) {
            return ($activeTreeState = $action->getActiveTreeState())
                ? trans('Edit :label', ['label' => $component->getItemLabel($activeTreeState)])
                : trans('Add :name', ['name' => (string) Str::of($component->getName())->headline()->singular()]);
        });

        $this->mountUsing(function (self $action, ComponentContainer $form) {
            $form->fill($action->getActiveTreeState());
        });

        $this->form(function ($component) {
            return $component->getChildComponents();
        });

        $this->action(function ($livewire, self $action, array $data) {
            data_set(
                $livewire,
                $action->getActiveTreeStatePath(),
                array_merge($action->getActiveTreeState(), $data)
            );
        });
    }

    protected function getActiveTreeStatePath(): string
    {
        return $this->getArguments()['activeTreeStatePath'];
    }

    protected function getActiveTreeState(): array
    {
        return data_get($this->getLivewire(), $this->getActiveTreeStatePath()) ?? [];
    }
}
```

> These are not the entire snippet of their respective classes, I only included important bits.

And finally, we're consuming this by doing:
```php
Tree::make('nodes')
    ->itemLabel(fn (array $state) => $state['label'] ?? null)
    ->schema([
        // modal form schema...
    ])
```

With the snippets out of the way. Lets go through the step by step process. 

1. User clicks the "Add item" button and dispatches a `tree::createItem`.
2. Our listener will mount the `tree-form` action to with an argument of the item's state path. We call this as the active tree state path.
    > This state path is suffixed with a `uuid` 
3. Then the modal will be shown to the user.

Now, the issue occurs when the `Tree`'s `schema()` consist a `reactive()` field. When making a livewire call, The `TreeFormAction` no longer has the context of the active tree item state path which was provided to it earlier. We are using this argument to compose the `modalHeading()`.

After a bit of reading, I figured that the initial argument is being stored in the `$mountedFormComponentActionArguments` livewire property. So I figured I could use that to my advantage.

danharrin:
Okay, this isn't the correct fix though. Please open a minimal reproduction repository with an issue. Your code example contains too much detail, please simplify it to be very basic.

juliomotol:
> Your code example contains too much detail, please simplify it to be very basic.

In a gist, since we are dynamically mounting the form action (through the use of `mountFormComponentAction()`) we want the argument passed during mounting to persist to any subsequent request since we have a `reactive()` field in the action's form. Calling `getArguments()` during mounting works fine, but always returns empty in any subsequent requests and will only be available again when calling the mounted action. Its already being persisted in `$mountedFormComponentActionArguments` but is not being reinstated in the mounted action.

PR #7099, opened 2023-07-21
Fix table grouping headers

sumardi:

Before:
Screenshot 2023-07-21 at 7 53 04 PM

After:
Screenshot 2023-07-21 at 7 52 24 PM

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -553,17 +553,18 @@
                                     ])
                                 >
                                     @php
-                                        $tag = $group->isCollapsible() ? 'button' : 'div';
+                                        $isCollapsible = $group->isCollapsible();
+                                        $tag = $isCollapsible ? 'button' : 'div';
                                     @endphp
 
                                     <{{ $tag }}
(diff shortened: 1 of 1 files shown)


zepfietje:
I'm failing to see how this change fixes the issue.
Could you elaborate?

sumardi:
> I'm failing to see how this change fixes the issue. Could you elaborate?

I can't think a better explanation. Maybe there is better approach. But this approach avoiding livewire v3 injecting morph markers, maybe because it is variable not public method.

sumardi:
https://github.com/livewire/livewire/discussions/5844

zepfietje:
Thanks for clarifying, @sumardi!

Ugh, seems like the Livewire regex doesn't match the method calls while it does match variables..?

@danharrin, what should we do here?

danharrin:
We need to make a failing test for Livewire otherwise it wont be fixed

zepfietje:
Agree, can you take that on, since you've done it before, @danharrin?

danharrin:
livewire/livewire#5972

PR #7124, opened 2023-07-21
Update 14-upgrade-guide.md

mrezauli:
fix to latest available release number to Filament upgrade tool.


danharrin:
Hmm, I don't think this is right? Why should we do this?

mrezauli:
I was under the impression that those instructions were intended solely for the beta release. However, after reading your comment, I now understand that those lines are meant for the final release.

PR #7128, opened 2023-07-21
fix: autocomplete/autofill styling

andrewdwallo:
This commit adds the `rounded-lg` class to the input field to ensure that autofilled inputs maintain the same border radius
as non-autofilled inputs, improving visual consistency.

**Before (autofill radius overlaps input radius):**
Screenshot 2023-07-21 at 7 20 31 PM

**After:**
Screenshot 2023-07-21 at 7 26 03 PM

### DIFF
--- a/packages/support/resources/views/components/input/index.blade.php
+++ b/packages/support/resources/views/components/input/index.blade.php
@@ -6,7 +6,7 @@
 <input
     {{
         $attributes->class([
-            'fi-input block w-full border-none bg-transparent py-1.5 text-base text-gray-950 outline-none transition duration-75 placeholder:text-gray-400 focus:ring-0 disabled:text-gray-500 disabled:[-webkit-text-fill-color:theme(colors.gray.500)] disabled:placeholder:[-webkit-text-fill-color:theme(colors.gray.400)] dark:text-white dark:placeholder:text-gray-500 dark:disabled:text-gray-400 dark:disabled:[-webkit-text-fill-color:theme(colors.gray.400)] dark:disabled:placeholder:[-webkit-text-fill-color:theme(colors.gray.500)] sm:text-sm sm:leading-6',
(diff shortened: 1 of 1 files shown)


zepfietje:
Hmm, thanks for bringing this up.
We can't just add the rounding to the input though, as the input should not be rounded when used with inline prefixes for example.

Is there a way to prevent the autofill input background overflowing the parent?

Oh seems that Chrome just adds a background color, so we could just add `overflow-hidden` to the wrapping affixes component. However, we can't just do that for all fields as e.g. selects should be able to overflow the parent.

zepfietje:
I've added this to just the text input component: https://github.com/filamentphp/filament/commit/2805e5a7ceec4b00d6a5fb17b1f7d1db43f1a128.

Could you verify that this fixes the issue, @andrewdwallo?

andrewdwallo:
@zepfietje Apologies. I'm out of town without my computer. I tried other things but not "overflow-hidden" but if you notice that it looks fine with autofill then I would say all good 👍 It happens in all browsers though not just chrome...

zepfietje:
I didn't test it, so please have a look when you get back to your computer. 😄

andrewdwallo:
Alright sounds good.

andrewdwallo:
Got back to my computer, looks good now!

zepfietje:
Perfect, thanks for checking!

andrewdwallo:
@zepfietje This needs to be added for the blade components as well:
```
<x-filament::input.wrapper>
    <x-filament::input
        type="text"
        wire:model="name"
    />
</x-filament::input.wrapper>
```

zepfietje:
Hmm, how would we check if you're passing just a regular input though? The issue is that for example select fields (with Choices.js) should be able to overflow the input wrapper. That's the reason I added it to the text input field. Ideas?

andrewdwallo:
Not really sure. I guess just either let the user know through documentation or let them figure it out...

PR #7138, opened 2023-07-22
Fix issue to edit record with combined relation manager tabs

sumardi:

Edit records with `hasCombinedRelationManagerTabsWithContent()` set to `true` throws `Argument #1 ($manager) must be of type Filament\Resources\RelationManagers\RelationManagerConfiguration|string, null given`


zepfietje:
This needs to be fixed differently, by not making the call to that method for the content tab.
Handing this over to @danharrin.

PR #7154, opened 2023-07-23
Add `items-center` to user menu component

akunbeben:

Update user-menu trigger button

Before: 
image
Testing visually using render hook
image

After: 
image

### DIFF
--- a/packages/panels/resources/views/components/user-menu.blade.php
+++ b/packages/panels/resources/views/components/user-menu.blade.php
@@ -15,7 +15,7 @@
 {{ \Filament\Support\Facades\FilamentView::renderHook('user-menu.start') }}
 
 <x-filament::dropdown placement="bottom-end" class="fi-user-menu">
-    <x-slot name="trigger">
+    <x-slot name="trigger" class="flex items-center">
         <button
             aria-label="{{ __('filament::layout.actions.open_user_menu.label') }}"
             type="button"


zepfietje:
Could you explain the exact problem and which render hook you're using?

zepfietje:
I'll reopen this once the issue is clearer. 🙂

akunbeben:
I was playing around on v3, tried the tenancy feature and then I noticed that the tenant menu disappeared when I use the top navigation `->topNavigation(true)`. So, I created a custom tenant menu and put it beside the user menu. 

I see all the component on the topbar is centered and I noticed that the user menu trigger button is not centered, then open this PR 😃 

```php
->renderHook(
    'user-menu.start',
    fn (): string => $panel->hasTopNavigation() ? Blade::render('@include(\'components.custom-tenant-menu\')') : '',
)
```

zepfietje:
Thanks for clarifying.
I'm confused though, as adding `class="flex items-center"` to the user menu trigger shouldn't change how its sibling is aligned vertically?

In fact, its parent (`packages/panels/resources/views/components/topbar/index.blade.php`) is already `flex` and center-aligns its children.

akunbeben:
Yes, you are right. I think it's because the trigger button was wrapped by another div in the `dropdown > trigger slot`. So, that's why the topbar `flex items-center` is only aligned the dropdown div, not the actual trigger button.

For example, the tenant menu is wrapping its trigger button inside the `flex items-center`
https://github.com/filamentphp/filament/blob/5d10b326d5794fc2f376c087acbdb35877990517/packages/panels/resources/views/components/tenant-menu.blade.php#L31

But, it's okay. I still can use custom component though 🙂

zepfietje:
So it appears to be an issue when the trigger is not `flex`.
I've just added `flex` to the dropdown trigger slot, so no `items-center` is needed, and no fixes on specific components.

Took a while for me to understand what was going on here.
Thanks for bringing this up and your patience. 🙏

By the way, if you want to verify this is fixed, pull in the latest changes from the 3.x branch to test.

akunbeben:
Thanks, Zep! updating the composer now🚀

PR #7161, opened 2023-07-23
fix(docs): fixing links in docs

sooand:
I've come across some broken links in the docs and tried to fix them, but don't know if it's caused by the import to the actual docs on beta.filamentphp.com app. Since it seems to need relative paths on e.g. "04-pages.md" but not on "06-navigation.md"

I hope this can at least show as a guide over which links needs some attention, as they either do nothing, or send the user to the https://beta.filamentphp.com/docs/3.x/panels/installation/ page


zepfietje:
Thanks for bringing this up, @sooand.
We're aware of this issue and are working on a proper fix. 🙂

PR #7184, opened 2023-07-24
Fix: Select heights and User menu alignment

awcodes:

Before
Screen Shot 2023-07-24 at 11 39 59 AM
Screen Shot 2023-07-24 at 11 40 04 AM

After
Screen Shot 2023-07-24 at 11 34 38 AM
Screen Shot 2023-07-24 at 11 34 47 AM


zepfietje:
@awcodes, could you check if the user menu alignment is also fixed without your change?
Not sure what version of 3.x you worked upon, but I'd probably fixed it earlier today by adding `flex` to the dropdown trigger.

packages/support/resources/views/components/input/select.blade.php
```diff
@@ -21,7 +21,7 @@
     @endif
     {{
         $attributes->class([
-            'fi-select-input block w-full border-none bg-transparent py-1.5 pe-8 text-base transition duration-75 focus:ring-0 disabled:text-gray-500 disabled:[-webkit-text-fill-color:theme(colors.gray.500)] dark:disabled:text-gray-400 dark:disabled:[-webkit-text-fill-color:theme(colors.gray.400)] sm:text-sm sm:leading-6',
+            'fi-select-input block w-full border-none bg-transparent py-2 pe-8 text-base transition duration-75 focus:ring-0 disabled:text-gray-500 disabled:[-webkit-text-fill-color:theme(colors.gray.500)] dark:disabled:text-gray-400 dark:disabled:[-webkit-text-fill-color:theme(colors.gray.400)] sm:text-sm sm:leading-6',
```
zepfietje, line 24: @awcodes, what does this fix exactly?
I'm pretty sure it should be `py-1.5`.

awcodes: Selects don't have a line height so the padding on them always has to be more that on a standard input. The screenshots show the difference.

awcodes:
I'm on beta9 and the user menu wasn't aligned. The issue is the button, not the trigger. The button is respecting the line height without being explicitly set to something other than inline, which is the default for buttons.

awcodes:
Selects are ok, was a theme issue.

on beta9 the user menu is still not aligned and the fi-dropdown-trigger doesn't have a flex class. But if you've got in an outstanding PR, feel free to close this. :)

packages/support/resources/views/components/input/select.blade.php
zepfietje: Hmm, could you share the height in `px` after this change?

zepfietje: Oh just saw your other comment.

zepfietje:
I pushed my commit earlier today, but 8 minutes after Dan tagged the latest alpha. 😂
https://github.com/filamentphp/filament/commit/8fc75aa1181da8a1ad6923e8d6fbcced42edb8d7

zepfietje:
Sorry to close your PR, but glad all these issues are actually resolved already.
As always, thanks for bringing these up. 😄

PR #7189, opened 2023-07-24
Handle password reset

4m1n0s:

Handle user password reset cases like INVALID_USER, RESET_THROTTLED

### DIFF
--- a/packages/panels/src/Pages/Auth/PasswordReset/RequestPasswordReset.php
+++ b/packages/panels/src/Pages/Auth/PasswordReset/RequestPasswordReset.php
@@ -80,21 +80,19 @@ class RequestPasswordReset extends SimplePage
             },
         );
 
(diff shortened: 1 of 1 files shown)


packages/panels/src/Pages/Auth/PasswordReset/RequestPasswordReset.php
```diff
@@ -80,21 +80,19 @@ function (CanResetPassword $user, string $token): void {
...
+        } else {
             Notification::make()
                 ->title(__($status))
                 ->danger()
                 ->send();
-
-            return;
         }
 
         $this->form->fill();
```
zepfietje, line 95: Hasn't the behavior now changed since we're calling `fill()` even when throttled, @danharrin?

danharrin: good catch

PR #7194, opened 2023-07-25
add new method `hasPlugin`

atmonshi:
Adding `hasPlugin` make it easier to check if a plugin is enabled in the current panel to set routes or other usages from the plugin service provider.

    _currently no docs for internal methods, so I don't know where to add it to._
    _no visual changes._

### DIFF
--- a/packages/panels/src/Facades/Filament.php
+++ b/packages/panels/src/Facades/Filament.php
@@ -54,6 +54,7 @@ use Illuminate\Support\Facades\Facade;
  * @method static Panel getPanel(?string $id = null)
  * @method static array<string, Panel> getPanels()
  * @method static Plugin getPlugin(string $id)
+ * @method static bool hasPlugin(string $id)
  * @method static string | null getProfileUrl(array $parameters = [])
  * @method static string | null getRegistrationUrl(array $parameters = [])
  * @method static string | null getRequestPasswordResetUrl(array $parameters = [])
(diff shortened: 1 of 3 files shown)


zepfietje:
This can probably already be achieved by accessing the plugin config using `filament('unique-plugin-id')`?

https://beta.filamentphp.com/docs/3.x/panels/plugins#configuring-plugins-per-panel

Closing this, unless I'm wrong. 😄

atmonshi:
hi @zepfietje  :)

the `filament('zeus-sky')` will return a plugin object but If the plugin is not enabled in the panel, it will throw an exception

`Plugin [zeus-sky] is not registered for panel [admin].`

because of this

```
public function getPlugin(string $id): Plugin
{
    return $this->getPlugins()[$id] ?? throw new Exception("Plugin [{$id}] is not registered for panel [{$this->getId()}].");
}
```

the idea of `hasPlugin` is to check if the plugin is enabled (the user added `->plugin(SkyPlugin::make())` to the panel) or not without throwing an exception.

or change the `getPlugin` method to return null or something? but I guess that will break things not sure. 

not a big change, as I am currently doing something like
```php
$filament = app('filament');

if ($filament->getCurrentPanel() !== null && array_key_exists('zeus-sky', $filament->getCurrentPanel()->getPlugins())) {

```

but maybe will help other plugin devs too.

zepfietje:
Fair point!

PR #7197, opened 2023-07-25
Re-introduce wire:navigate (SPA mode)

danharrin:
Anyone who wants to see SPA mode in Filament should test this PR. Currently, the `wire:navigate` feature is not stable enough to use. Let's track any issues here to ensure they are fixed before we merge it.

## Todo

- [x] Add `wire:navigate` to all links, including actions.
- [x] Remove duplicate Livewire loading indicator.


zepfietje:
@danharrin, could you handle adding `wire:navigate` to actions in this PR?
I'll close the issue for it: https://github.com/filamentphp/filament/issues/7083.

lbitoy:
@danharrin  is testing still needed on this branch or can we merge?

danharrin:
Please test it and let me know how it goes

emilianotisato:
@danharrin I suggest 2 changes:

1- for the body script loading the echo broadcast, we can add `data-navigate-once` to the script tag https://livewire.laravel.com/docs/navigate#scripts-in-the-body-are-re-evaluated

2- The missing pice to get the feel of an spa is adding the `@persist` directive (https://livewire.laravel.com/docs/navigate#persisting-elements-across-page-visits)
Of the top of my head this should be added in siderbar and top bar.

Do you want me to update this PR with those changes?

danharrin:
We cannot use persist AFAIK since then the sidebar active state / breadcrumbs wouldnt change between visits?

faizananwerali:
+1

zepfietje:
Instead of "+1'ing", please test this PR and provide actual feedback on what issues you did or didn't encounter.

atmonshi:
Update: cloned locally and updated the branch from 3.x and all good 👍

~~how i can install this branch, it conflicts with every other filament plugin :)~~

~~did `dev-spa as 3.0` and `composer update -W`~~

atmonshi:
doing some testing....

## 1- ~~what to use for~~
`document.addEventListener('DOMContentLoaded')` ~~not working~~

figure it out by using both:
```
document.addEventListener('livewire:navigated', scrollToSection);
document.addEventListener('DOMContentLoaded', scrollToSection);
```



## 2- dual loading indicator:

https://github.com/filamentphp/filament/assets/1952412/3fb5b2e8-2573-4b82-a333-daccb9454955


## 3- console errors with alpain, or maybe LW:
```
Uncaught SyntaxError: Identifier 'collapsedGroups' has already been declared
    at swapCurrentPageWithNewHtml
```

Screenshot 2023-08-25 at 12 52 27 PM


##  Q: the `wire:navigate` is only for sidebar nav? any plans to add it to other links like (create action)?

ArtMin96:
When the **form package** is used on pages that work with wire:navigate, there are errors in the console when you go to another page and return with the back button of the browser.

https://github.com/filamentphp/filament/assets/29732308/a5a482d9-9a25-41d8-ae89-0d8e67ce17cb

zepfietje:
@atmonshi, I'm aware of the duplicate loading indicator, which IMO should be fixed in Livewire. So maybe you could create a PR there?

And yes, we should add this to actions and a couple other places too, as I mentioned in my initial comment on this PR.

mentamarindos:
hey guys how can I test this feature? I couldn't install this `spa ` branch using composer, any chance you could give me a guide to follow? Thank you in advance 🤗

zepfietje:
@mentamarindos, thanks for helping test this feature!
Please check out the contribution guide in the README: https://github.com/filamentphp/filament#contributing.

mentamarindos:
Ok, I've been testing it. I love that it gives that blazingly fast effect when loading pages 🚀 I haven't encountered any other issues apart from the console error that has already been mentioned, it's a Livewire bug. we'll need to wait for a fix before resuming testing. Nothing critical for now, though, it's working!

akunbeben:
I've been tested this SPA mode, I noticed most of the errors is come from Livewire itself. 

In case people doesn't know.

If we take a look at the some of the errors, most of them is about redeclare existing variables. Yes, livewire is rewrite the page. But, those variables are still in the memory.

For example: If I change this collapsedGroups variable with `var` keyword instead of `let`. It will work navigating between pages with no errors. As we know, that `var` keyword can be redeclared.
https://github.com/filamentphp/filament/blob/f8e986091f7954a676f8a2ed0ee92d5b7f522ede/packages/panels/resources/views/components/sidebar/index.blade.php#L124

This is only one example. In Filament, there is lot more.

danharrin:
@ArtMin96 your video isn't loading for me, can you explain what it shows?

> For the body script loading the echo broadcast, we can add data-navigate-once to the script tag

Done

> Uncaught SyntaxError: Identifier 'collapsedGroups' has already been declared

Will report this to LW, I think it's a bug

danharrin:
This will be in the next release, marked in the docs as "experimental". This should allow people to test it easier.

Now, instead of being enabled by default, you chain `->spa()` onto a panel.

ArtMin96:
@danharrin you can here [https://streamable.com/p4nhoy](https://streamable.com/p4nhoy)

danharrin:
Which form components are you using that trigger this

ArtMin96:
DatePicker, TimePicker, Select, Toggle, TextInput, and 2 custom fields. I closed the custom fields to ensure the problem does not come from them.
image
image

Also
image
image

I have done the following steps
1. `php artisan filament:upgrade`
2. `php artisan optimize:clear`
3. I don't have anything published in the resource directory
4. I have cleared the cache of the browser
5. 
image

danharrin:
I can't reproduce this inside the panel builder, so maybe its something to do with your app?

ArtMin96:
It is not in the panel. It is in the form package. I'm not using panel

danharrin:
I know it's not in the panel, but if it was a problem with the forms package why can I not reproduce it inside the panel? Since that also uses the forms package

Please try reproducing this in a new Laravel project, and submit that as a reproduction repository in an issue. Please **do not** submit your existing project. Please create a minimal example

francoism90:
I tried using wire:navigate in my own projects and also more lazy, however both were very buggy and unstable. For example, it would lose track of components, and causes JS to be parsed incorrectly.

Is Filament using tests? Because it's possible to test Livewire components for breaking.

Don't get me wrong, I really like Livewire V3, however something's don't feel ready yet, like modelable not working, etc.

danharrin:
It's an optional feature now, as I mentioned above. So doesn't really matter if its unstable.

ziming:
Found another situation where spa mode break

So i have a situation where i have a resource table action that has 2 actions. download media & excel, both of them fail as long as spa mode is on. When spa mode is on it will simply show a page with gibberish text instead of triggering the download

I also tried ->openUrlInNewTab() which fail as well if spa mode is on (gibberish text on a new tab instead of starting download). So perhaps there need to be a way to disable spa for specific action urls, especially if the link is suppose to be a download (be it excel, pdf, or other file types download)

```php
            ->actions([
                EditAction::make(),
                ActionGroup::make([
                    Action::make('excel')
                        ->label('Excel')
                        ->requiresConfirmation()
                        //->openUrlInNewTab() as long as spa mode is on, open in new tab will also fail
                        ->url(function (Model $record): string {
                            // move this download method to this class next time
                            return action([ExcelController::class, 'generateExcel'], ['id' => $record->id]);
                        }),

                   Action::make('download_media')
                        ->label('D/L Media')
                        ->url(function (Model $record): string {
                            // returns a spatie media library MediaStream zip file
                            return action([MediaController::class, 'download'], ['id' => $record->id]);
                        })
                        ->requiresConfirmation()
                        ->openUrlInNewTab(),
                ]),

            ])
```

binaryfire:
SPA mode is causing similar issues with Octane (Swoole) as in https://github.com/filamentphp/filament/issues/7969. One of our apps uses Spatie's tenancy package for subdomain multitenancy. If I enable `->spa()` and visit `tenant1.domain.com` followed by `tenant2.domain.com`, tenant 2's navigation sidebar links usually point to tenant 1. 

Octane boots once and stays in memory, which means any data cached in memory is persistent and shared across requests. So statics, global variables and singletons can cause problems. This is a good article, in case it helps: https://toluwaniadewale.com/blog/laravel-octane-with-great-performance-comes-great-responsibility

binaryfire:
Sorry my bad, the issue is unrelated to SPA mode. It's another occurrence of https://github.com/filamentphp/filament/issues/7969. I'll re-post there.

tareq-alqadi:
I had problem with most of the components of Filament such as tables, forms and even the header.
image
image

I faced  these problems when I create a new Laravel project, then downloaded Filament, and just added the user resource and added spa.
These problems appear after i click a wire navigated links

danharrin:
1) If you encounter `wire:navigate` issues that work when you don't have it enabled, please do your best to submit the bug report to Livewire (through a failing test pull request), and not Filament. The Filament team does not have enough time in the day to report every one of your bugs for you.
2) If you think it's a problem with Filament and not Livewire, please open a Filament issue with a reproduction repository, or open a PR with your proposed fix.

PR #7214, opened 2023-07-26
[3.x] add missing fa translations for table

mokhosh:
This just adds the new "en" keys to the "fa" lang file and translates the new items.


zepfietje:
Thanks!

PR #7217, opened 2023-07-26
Add new API to force light theme

sawirricardo:
Before, when we disable `->darkMode(false)`, filament will let the system default to be used for theme.

This may be unintentional as some dev are required to still 'force' the users to use `light` theme.

This PR lets the developers to force the users to use `light` theme.

### DIFF
--- a/packages/panels/resources/views/components/layout/base.blade.php
+++ b/packages/panels/resources/views/components/layout/base.blade.php
@@ -78,6 +78,15 @@
             </script>
         @endif
 
+        @if (filament()->hasLightModeForced())
+            <script>
+                document.addEventListener('alpine:init', () => {
+                    localStorage.setItem('theme', 'light')
+                    document.documentElement.classList.remove('dark')
+                })
+            </script>
+        @endif
+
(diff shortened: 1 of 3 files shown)


zepfietje:
This new API still allows for switching to a different theme, right?

sawirricardo:
> This new API still allows for switching to a different theme, right?

Correct. What it does is only force the users to "switch" to light theme on each page load.

[EDIT]
Let's say, the devs enable something like this
```
$panel->darkMode()->forceLightMode();
```

What the user will experience:
1. Open the app, with the theme 'swithced' to light.
2. Open the Theme Switcher. Change to dark mode.
3. Open other page. The user will see that the app 'switched' back to light theme.

danharrin:
`darkMode(false)` should force light theme, not system mode. That can be fixed in a new PR.

sawirricardo:
@danharrin I also thought like this first, but, maybe there are develoeprs who want to use the `system` theme, no?

danharrin:
That would create a poor user experience for those who don't get to control which theme they use.

sawirricardo:
Noted, created the new PR here https://github.com/filamentphp/filament/pull/7224

PR #7222, opened 2023-07-26
Fix phpdoc for table pagination options

erikgaal:
PHPStan will complain when using this method according to the docs:

```
Parameter #1 $options of method Filament\Tables\Table::paginationPageOptions() expects array<int, string>|Closure|null, array{25, 'all'} given. 
```


danharrin:
Cheers

PR #7225, opened 2023-07-26
Add conditionable trait to `Panel` class

Summonshr:
- use Conditionable Trait.

use case - 

```php
$panel->when(!app()->runningInConsole(), function ($panel) {
    return $panel
        ->viteTheme('resources/css/filament/admin/theme.css');
})
```

### DIFF
--- a/packages/panels/src/Panel.php
+++ b/packages/panels/src/Panel.php
@@ -33,7 +33,8 @@ class Panel extends Component
     use Panel\Concerns\HasTheme;
     use Panel\Concerns\HasTopNavigation;
     use Panel\Concerns\HasUserMenu;
-
+    use Illuminate\Support\Traits\Conditionable;
+    
     protected bool $isDefault = false;
 
     protected ?Closure $bootUsing = null;
--- a/packages/panels/src/Panel.php
+++ b/packages/panels/src/Panel.php
@@ -6,6 +6,7 @@ use Closure;
 use Filament\Support\Components\Component;
 use Filament\Support\Facades\FilamentColor;
 use Filament\Support\Facades\FilamentIcon;
+use Illuminate\Support\Traits\Conditionable;
 
 class Panel extends Component
 {
@@ -33,7 +34,7 @@ class Panel extends Component
     use Panel\Concerns\HasTheme;
     use Panel\Concerns\HasTopNavigation;
     use Panel\Concerns\HasUserMenu;
-    use \Illuminate\Support\Traits\Conditionable;
+    use Conditionable;
     
     protected bool $isDefault = false;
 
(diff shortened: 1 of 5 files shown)


packages/panels/src/Panel.php
```diff
@@ -33,7 +33,8 @@ class Panel extends Component
     use Panel\Concerns\HasTheme;
     use Panel\Concerns\HasTopNavigation;
     use Panel\Concerns\HasUserMenu;
-
+    use \Illuminate\Support\Traits\Conditionable;
```
zepfietje: Please import the namespace.

packages/panels/src/Panel.php
```diff
@@ -33,7 +34,7 @@ class Panel extends Component
     use Panel\Concerns\HasTheme;
     use Panel\Concerns\HasTopNavigation;
     use Panel\Concerns\HasUserMenu;
-    use \Illuminate\Support\Traits\Conditionable;
+    use Conditionable;
```
zepfietje: Move it to the top to be sorted alphabetically please. 🙂

packages/panels/src/Panel.php
Summonshr: Imported.

zepfietje:
Maybe we should add `Conditionable` to `Filament\Support\Components` instead, @danharrin?

Summonshr:
> Maybe we should add `Conditionable` to `Filament\Support\Components` instead, @danharrin?

I tried this actually. Not sure what did not work.

Summonshr:
> Maybe we should add `Conditionable` to `Filament\Support\Components` instead, @danharrin?

It conflicts with CanBeHidden Traits.

`Declaration of Filament\Forms\Components\Concerns\CanBeHidden::when(Closure|bool $condition = true): static must be compatible with Filament\Support\Components\Component::when($value = null, ?callable $callback = null, ?callable $default = null)`

zepfietje:
Ah, I see. Our `when()` method is an alias of `visible()`, so we might remove `when()` in favor of adding the `Conditionable` trait. This would be a breaking change though, so would have to be implemented before releasing v3 stable, @danharrin.

danharrin:
I'm ok with that, we can remove the old `when()` methods

danharrin:
Mark this as ready when you've done that @Summonshr

Summonshr:
> Mark this as ready when you've done that @Summonshr

Done

packages/panels/src/Panel.php
```diff
@@ -33,7 +33,7 @@ class Panel extends Component
     use Panel\Concerns\HasTheme;
     use Panel\Concerns\HasTopNavigation;
     use Panel\Concerns\HasUserMenu;
-
+
```
zepfietje:
```suggestion

```

zepfietje:
Thanks

PR #7238, opened 2023-07-27
Pass state into repeater conditional methods

thienthanit04:

### Preventing the user from deleting specific item

You may prevent the user from deleting specific item from the repeater using the `itemDeletable()` method:
This method accepts a closure that receive the current item's data in a `$state` variable.
You must return the boolean value to determine that item deletable or not.

```PHP
use Filament\Forms\Components\Repeater;

Repeater::make('members')
    ->schema([
        // ...
    ])
    ->itemDeletable(fn ($state): bool => !$state['is_default'])
```
Above example means that attribute `is_default` is true will prevent user delete item.

image

### DIFF
--- a/packages/forms/docs/03-fields/12-repeater.md
+++ b/packages/forms/docs/03-fields/12-repeater.md
@@ -111,6 +111,23 @@ Repeater::make('members')
     ->deletable(false)
 ```
 
+### Preventing the user from deleting specific item
+
+You may prevent the user from deleting specific item from the repeater using the `itemDeletable()` method:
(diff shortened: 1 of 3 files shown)


zepfietje:
`deletable()` is what you're looking for.

danharrin:
I don't think `deletable()` supports passing the state. But I agree that we should use that existing method instead of introducing a new one. Mark this as ready when you've done that @thienthanit04

danharrin:
Hey @thienthanit04 feel free to reopen this PR on the same branch if you want to work on this.

PR #7244, opened 2023-07-27
Added support for login conditions

billmn:
This PR allows to customize login conditions (like `status => active`) and prevent an error if `remember` component was removed from `form` array.

### DIFF
--- a/packages/panels/src/Pages/Auth/Login.php
+++ b/packages/panels/src/Pages/Auth/Login.php
@@ -65,10 +65,7 @@ class Login extends SimplePage
 
         $data = $this->form->getState();
 
-        if (! Filament::auth()->attempt([
-            'email' => $data['email'],
-            'password' => $data['password'],
-        ], $data['remember'])) {
+        if (! Filament::auth()->attempt($this->loginConditions($data), $data['remember'] ?? false)) {
             throw ValidationException::withMessages([
                 'data.email' => __('filament::pages/auth/login.messages.failed'),
             ]);
@@ -153,5 +150,13 @@ class Login extends SimplePage
     protected function hasFullWidthFormActions(): bool
     {
         return true;
+    }
+
+    protected function loginConditions(array $data): array
+    {
+        return [
+            'email' => $data['email'],
+            'password' => $data['password'],
+        ];
     }
 }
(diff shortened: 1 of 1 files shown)


zepfietje:
Could you elaborate on how this new API would work?
I don't think it's the right approach to be honest.

billmn:
@zepfietje Suppose I want to allow access only to active users.

I could extend the `Filament\Pages\Auth\Login` class and define any additional conditions.

Like this:
```php
<?php

namespace App\Livewire;

use Filament\Pages\Auth\Login;

class MyLogin extends Login
{
    protected function loginConditions(array $data): array
    {
        return [
            'email' => $data['email'],
            'password' => $data['password'],
            'active' => 1,
        ];

        // Or with complex query conditions
        return [
            'email' => $data['email'],
            'password' => $data['password'],
            fn ($query) => $query->has('activeSubscription'),
        ];
    }
}
```

zepfietje:
I think it's better to override the `authenticate()` method if you're going to have custom authentication logic. Let's see what @danharrin thinks though.

billmn:
Specifying additional conditions at login is a fairly common operation, I personally believe it is better to consider a dedicated method rather than overriding the entire`authenticate` method.

In any case I await your feedback.

danharrin:
I don't mind this kind of abstraction that just introduces more methods to override, I just don't like it when other classes then get involved with extra configuration. I'm absolutely fine with this.

packages/panels/src/Pages/Auth/Login.php
```diff
@@ -154,4 +151,12 @@ protected function hasFullWidthFormActions(): bool
     {
         return true;
     }
+
+    protected function loginConditions(array $data): array
```
danharrin:
```suggestion
    protected function getCredentialsFromFormData(array $data): array
```

packages/panels/src/Pages/Auth/Login.php
```diff
@@ -65,10 +65,7 @@ public function authenticate(): ?LoginResponse
 
         $data = $this->form->getState();
 
-        if (! Filament::auth()->attempt([
-            'email' => $data['email'],
-            'password' => $data['password'],
-        ], $data['remember'])) {
+        if (! Filament::auth()->attempt($this->loginConditions($data), $data['remember'] ?? false)) {
```
danharrin:
```suggestion
        if (! Filament::auth()->attempt($this->getCredentialsFromFormData($data), $data['remember'] ?? false)) {
```

PR #7250, opened 2023-07-27
Allow native DatePicker/DateTimePicker to open onfocus

sumardi:

For consistency between native and non-native, this PR will allow native datetime-local to open onfocus.

Without this PR, this also works:
```php
Forms\Components\DateTimePicker::make('start_date')
    ->extraAlpineAttributes(['onfocus' => 'this.showPicker?.()'])
```

Browser compatibility: https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/showPicker#browser_compatibility

### DIFF
--- a/packages/forms/resources/views/components/date-time-picker.blade.php
+++ b/packages/forms/resources/views/components/date-time-picker.blade.php
@@ -43,6 +43,7 @@
                             'max' => (! $isConcealed) ? $getMaxDate() : null,
                             'min' => (! $isConcealed) ? $getMinDate() : null,
                             'placeholder' => $getPlaceholder(),
+                            'onfocus' => 'this.showPicker?.()',
                             'readonly' => $isReadOnly(),
                             'required' => $isRequired() && (! $isConcealed),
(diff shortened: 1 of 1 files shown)


zepfietje:
I'm not sure if we should change this default behavior.
Could you explain the need, @sumardi?

sumardi:
https://github.com/filamentphp/filament/assets/315603/2d900584-c5af-41bd-a486-9ac48e22c552

Attached screen record for comparison, for now we need to click on the calendar icon to open the native picker.

zepfietje:
Hmm, right.
I was testing on Safari and there it opens the panel when you click the input.
If you focus the input using the keyboard, it doesn't open the panel.
What I found though, is that you can open the panel by pressing `space` when the input is focused.
This works across both Chromium and Safari.

zepfietje:
I'm going to close this PR for now, since messing with default browser behavior often only causes problems. When more people brings it up as an issue, I'll reconsider merging this PR.

PR #7258, opened 2023-07-27
Fix hints not wrapping correctly when using inlineLabel()

Capevace:
Back in V2, hints were displayed below the label when using `inlineLabel()` on a component.

As of 3.0.0-beta17 (and in 3.x branch), hints are instead placed inline, which looks very broken when having longer hints.

Behaviour in V2:
Screenshot 2023-07-27 at 23 16 19

Broken behaviour in V3:
Screenshot 2023-07-27 at 23 07 59

This PR fixes the issue by taking the hint into account in the classes:
```php
    $hasError = $errors->has($statePath) || ($hasNestedRecursiveValidationRules && $errors->has("{$statePath}.*"));
    $hasHint = filled($hint) || $hintIcon || count($hintActions);
@endphp

...
<div
        @class([
	        'flex',
	        'items-center justify-between gap-x-3' => !$hasInlineLabel,
	        'sm:pt-1.5' => $hasInlineLabel && !$hasHint,
	        'flex-col gap-y-1' => $hasInlineLabel && $hasHint
        ])
```

This is how it looks after the fix:
Screenshot 2023-07-27 at 23 31 40

I'm assuming the `sm:pt-1.5` was already there, in order to vertically center the label when using `inlineLabel`? This looks weird when the hints is placed below, so it's only included when there's no hint:
Screenshot 2023-07-27 at 23 54 55

Let me know if there's anything I can improve, as this is my first Filament PR. :)

### DIFF
--- a/packages/forms/resources/views/components/field-wrapper/index.blade.php
+++ b/packages/forms/resources/views/components/field-wrapper/index.blade.php
@@ -42,6 +42,7 @@
     );
 
     $hasError = $errors->has($statePath) || ($hasNestedRecursiveValidationRules && $errors->has("{$statePath}.*"));
+    $hasHint = filled($hint) || $hintIcon || count($hintActions);
 @endphp
 
 <div {{ $attributes->class(['fi-fo-field-wrp']) }}>
@@ -57,11 +58,13 @@
             'sm:grid-cols-3 sm:items-start sm:gap-x-4' => $hasInlineLabel,
         ])
     >
-        @if (($label && (! $labelSrOnly)) || $labelPrefix || $labelSuffix || filled($hint) || $hintIcon || count($hintActions))
+        @if (($label && (! $labelSrOnly)) || $labelPrefix || $labelSuffix || $hasHint)
             <div
                 @class([
-                    'flex items-center justify-between gap-x-3',
-                    'sm:pt-1.5' => $hasInlineLabel,
+                    'flex',
+                    'items-center justify-between gap-x-3' => !$hasInlineLabel,
+                    'sm:pt-1.5' => $hasInlineLabel && !$hasHint,
+                    'flex-col gap-y-1' => $hasInlineLabel && $hasHint
                 ])
             >
                 @if ($label && (! $labelSrOnly))
@@ -82,7 +85,7 @@
                     {{ $labelSuffix }}
                 @endif
 
-                @if (filled($hint) || $hintIcon || count($hintActions))
+                @if ($hasHint)
                     <x-filament-forms::field-wrapper.hint
                         :actions="$hintActions"
                         :color="$hintColor"


zepfietje:
Hmm, hints aren't designed for long text.
Since they're placed on the same line as the label by default, I've moved them there for inline labels too.

An example of a hint would be an `Optional` hint next to the label to explicitly indicate that the field is optional. Or what you're using as helper text, `Öffentlich sichtbar`, would also be a hint.

Instead, in your screenshot you're using the hint as what should be the helper text.
The helper text always gets placed below the input, but we might add an option to place it below the label instead. This would make sense as an option for both inline and non-inline fields.

I'm going to close this PR as it's not a bug, but we can continue discussion regarding the above here if you like.

Capevace:
Hmm fair enough i never considered the semantics behind `hint()` and `helperText()` much, especially since it's been like that in V2.

However I do feel like putting the description / helper text under the label feels much more natural than putting it under the input itself, so I'd love to help make that an option, any API ideas? I don't personally see too many situations where I'd wanna put a hint on the same line as the label in a 1/3||2/3 layout.

But I guess that's what `->fieldWrapperView()` is for then. 😂

danharrin:
To be honest @zepfietje I agree with @Capevace here, I think it looks bad inline when you are using the inline label. "hint" as a feature does not suggest any position, it is purely semantic, so I think we should just make it look as neat as possible rather than trying to make it consistent across layouts

danharrin:
Example:

Screenshot 2023-07-28 at 08 46 27

I feel like those hints just belong below the label instead...

zepfietje:
> I feel like those hints just belong below the label instead...

I don't necessarily think so, but we might try.
Also, changing the field to use `inlineLabel()` doesn't change the meaning of the hint.
In my opinion, the hint should fit on the same line as the label just fine.
So even if we're moving the hint to below the label, my original comment still holds true.

zepfietje:
Created an issue for implementing the option to show helper text below label: https://github.com/filamentphp/filament/issues/7265.

danharrin:
I feel like `inlineLabel()` has and should represent a different field wrapper rather than just the setting to make the label inline. This is why they were separate views in the first place. We shouldn't cower away from other tweaks to make the UI look more polished just because they appear out of scope of the original feature, they are essentially side effects

zepfietje:
Oh the new hint position has nothing to do with the merge of the field wrapper views. I had already moved the hint when I hadn't merged the views. But as I said, I'll look into it and add a new option for having the helper text display below the label.

danharrin:
I don't understand why the helper text can't stay below the field, that sounds more inconsistent than just moving the hint 🤔

zepfietje:
Because they're different things semantically.

Capevace:
I've thought some more on this, and I do agree that the text below the label should be `helperText` if semantics are strictly considered. 

However I am wondering where the hint would be placed in that case. Putting it on the same line as the label seems like it would clutter the space potentially. 

From a package user perspective, I feel like simply swapping the two when using inline labels may be the smoothest option and might look cleaner to a user. Put long-form help texts on the side where there's room and place the short hint under the input, clearing space to focus on the input itself. 

Maybe having an option like `->inlineLabel(swapHintWithHelper: false)` or `->inlineLabel(inlineHints: true)` can make this relatively easy to opt-out of as well?

Just my two cents :)

epertinez:
I think both helperText and hint should have the ability to go under the label. I can imagine cases in which you need a message BEFORE the input field AND AFTER the input field. If both methods can be placed under the field, you can use both in those cases.

zepfietje:
@epertinez, https://github.com/filamentphp/filament/issues/7265.

PR #7273, opened 2023-07-29
widgets[feature]: separates chart-widget backgroundColor and borderColor

bezhanSalleh:

This PR separates the backgroundColor and borderColor and uses the colors defined by filamnet's panel theme and makes the border use the solid color defined while for background a shade of it just like v2. Doing so also fixes the current responsiveness of the chart as well.

Current:
https://github.com/filamentphp/filament/assets/10007504/5ab5c6d3-8658-4ec9-8c20-e0c1bbbf80ff

PR:
https://github.com/filamentphp/filament/assets/10007504/4f10249b-8039-45a2-a848-c9ac59e8564b


zepfietje:
Hmm, we had removed the background color on purpose.
Does this add it back to all charts by default?

bezhanSalleh:
> Hmm, we had removed the background color on purpose. Does this add it back to all charts by default?

yeah, for all widget charts. I did ask dan in the support channel before making the PR.
Having same border color and background color makes the charts kinda seem funky and ***it forces plugin developers to hardcode them*** and given that filament has themeing and not being able to use it, makes it kinda counter intuitive.

zepfietje:
@danharrin @bezhanSalleh this changed the fill of the data points:

Screenshot 2023-07-29 at 10 56 32

zepfietje:
Looks weird in dark theme as it doesn't cover the line:

Screenshot 2023-07-29 at 10 57 31

zepfietje:
No big issue though.
I see you used the background color of the stats overview widget card chart so that's consistent at least. 👍
Thanks, @bezhanSalleh!

danharrin:
Yeah I saw. Do you prefer them with a solid fill? I think I do, but I don't really mind

zepfietje:
I think the solid fill looks better in the way we use it, especially with this dark mode bug. But let's just leave it as is for now.

danharrin:
its ok I'll fix it

PR #7282, opened 2023-07-29
Allow us to go a little deeper in relations for Spatie Media Library Table Column

Keoghan:
This is a bug fix for the Spatie Media Library Table Column. Previously there were errors if attempting to use a media collection nested on a relation of the main table record.

There are no documentation or visual changes to add.


danharrin:
Thanks!

PR #7285, opened 2023-07-29
Fix search input element

sumardi:

Fix incorrect use of `<label for=''>`.


zepfietje:
Good catch, thanks!

PR #7287, opened 2023-07-29
Fix table content overflow

sumardi:

Possible fix for https://github.com/filamentphp/filament/discussions/7283


zepfietje:
This issue was on my radar, but thanks for taking the time to fix it already! 😄

PR #7297, opened 2023-07-30
Fix deselect records after completion

sumardi:

Fix issue with table records not deselect after completion.


danharrin:
Thanks

PR #7299, opened 2023-07-30
fix notification classes in docs

tvdr:

Just noticed while tinkering with notifications that the css classes differ


zepfietje:
Thanks, but we've decided to no longer document the exact classes since it's hard to keep up-to-date. I've removed the list from the docs. Thanks again for bringing this up!

PR #7303, opened 2023-07-30
adds `gap-4` for to adjust global search's render hooks

bezhanSalleh:

**Current:**
LTR:
Screenshot 2023-07-30 at 9 07 10 PM
RTL:
Screenshot 2023-07-30 at 9 14 33 PM

now if we try to handle it ourselves then depending on whether its start or end
Screenshot 2023-07-30 at 9 12 13 PM
Screenshot 2023-07-30 at 9 17 00 PM

but add the `gap-4` like topbar has would fix all of it.
Screenshot 2023-07-30 at 9 18 44 PM
Screenshot 2023-07-30 at 9 18 33 PM

### DIFF
--- a/packages/panels/resources/views/components/global-search/index.blade.php
+++ b/packages/panels/resources/views/components/global-search/index.blade.php
@@ -1,4 +1,6 @@
-<div class="fi-global-search flex items-center gap-4">
+{{ \Filament\Support\Facades\FilamentView::renderHook('panels::global-search.before') }}
+
+<div class="fi-global-search flex items-center">
     {{ \Filament\Support\Facades\FilamentView::renderHook('panels::global-search.start') }}
 
     <div class="relative">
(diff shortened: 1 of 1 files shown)


zepfietje:
Thanks for bringing this up, @bezhanSalleh!

I didn't add a gap to the component to give users full control, but I agree it makes sense to have it. However, maybe we should just move the render hooks out of the user menu component and into `topbar/index.blade.php`?

bezhanSalleh:
> Thanks for bringing this up, @bezhanSalleh!
> 
> 
> 
> I didn't add a gap to the component to give users full control, but I agree it makes sense to have it. However, maybe we should just move the render hooks out of the user menu component and into `topbar/index.blade.php`?
> 
> 
> 
> What do you think, @danharrin?

Np. Moving them would work fine as well.

danharrin:
The render hooks are fine as they are `start` and `end` instead of `before` and `after`

danharrin:
We should add before and after, outside of this component, instead of adding gap

zepfietje:
Replace `start` and `end` you mean, or add the others?

danharrin:
Add more

PR #7305, opened 2023-07-30
Filter tabs render hooks

saade:

Add four new render hooks to filter tabs.

`panels::resource.pages.list-records.filter-tabs.before` - Before the tabs
`panels::resource.pages.list-records.filter-tabs.start` - Start of the tabs (before the first tab)
`panels::resource.pages.list-records.filter-tabs.end` - End of the tabs (after the last tab)
`panels::resource.pages.list-records.filter-tabs.after` - After the tabs

image

image

<details>
  <summary>ps: If we define only one render hook before/ after the tabs, the whole group (filter tabs + rendered hook) will be centered on the page, causing a visual misalignment when only looking at the filter tabs. I can't find a good way around this, maybe @zepfietje  can have a look :)</summary>
  
  image
</details>

### DIFF
--- a/packages/panels/resources/views/resources/pages/list-records.blade.php
+++ b/packages/panels/resources/views/resources/pages/list-records.blade.php
@@ -6,27 +6,37 @@
 >
     <div class="flex flex-col gap-y-4">
         @if (count($tabs = $this->getTabs()))
-            <div
-                class="mx-auto max-w-full rounded-xl bg-white p-2 shadow-sm ring-1 ring-gray-950/5 dark:bg-gray-900 dark:ring-white/10"
-            >
-                <x-filament::tabs class="self-center">
-                    @foreach ($tabs as $tabKey => $tab)
-                        @php
-                            $activeTab = strval($activeTab);
-                            $tabKey = strval($tabKey);
-                        @endphp
+            <div class="flex max-w-full mx-auto">
+                {{ \Filament\Support\Facades\FilamentView::renderHook('panels::resource.pages.list-records.filter-tabs.before', scopes: $this->getRenderHookScopes()) }}
 
-                        <x-filament::tabs.item
-                            :active="$activeTab === $tabKey"
-                            :badge="$tab->getBadge()"
-                            :icon="$tab->getIcon()"
-                            :icon-position="$tab->getIconPosition()"
-                            :wire:click="'$set(\'activeTab\', ' . (filled($tabKey) ? ('\'' . $tabKey . '\'') : 'null') . ')'"
-                        >
-                            {{ $tab->getLabel() ?? $this->generateTabLabel($tabKey) }}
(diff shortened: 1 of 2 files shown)


danharrin:
Lets replace `filter-tabs` with just `tabs`

saade:
> Lets replace `filter-tabs` with just `tabs`

done

packages/support/docs/06-render-hooks.md
```diff
@@ -51,6 +51,10 @@ FilamentView::registerRenderHook(
 - `panels::page.header-widgets.after` - After the page header widgets, also [can be scoped](#scoping-render-hooks) to the page or resource class
 - `panels::page.header-widgets.before` - Before the page header widgets, also [can be scoped](#scoping-render-hooks) to the page or resource class
 - `panels::page.start` - Start of the page content container, also [can be scoped](#scoping-render-hooks) to the page or resource class
+- `panels::resource.pages.list-records.tabs.after` - After the resource filter tabs, also [can be scoped](#scoping-render-hooks) to the page or resource class
+- `panels::resource.pages.list-records.tabs.before` - Before the resource filter tabs, also [can be scoped](#scoping-render-hooks) to the page or resource class
+- `panels::resource.pages.list-records.tabs.end` - The end of the filter tabs (after the tab), also [can be scoped](#scoping-render-hooks) to the page or resource class
```
danharrin:
```suggestion
- `panels::resource.pages.list-records.tabs.end` - The end of the filter tabs (after the last tab), also [can be scoped](#scoping-render-hooks) to the page or resource class
```

packages/panels/resources/views/resources/pages/list-records.blade.php
```diff
@@ -6,27 +6,37 @@
...
-                            :badge="$tab->getBadge()"
-                            :icon="$tab->getIcon()"
-                            :icon-position="$tab->getIconPosition()"
-                            :wire:click="'$set(\'activeTab\', ' . (filled($tabKey) ? ('\'' . $tabKey . '\'') : 'null') . ')'"
-                        >
-                            {{ $tab->getLabel() ?? $this->generateTabLabel($tabKey) }}
-                        </x-filament::tabs.item>
-                    @endforeach
-                </x-filament::tabs>
+            <div class="flex flex-col max-w-full gap-4 mx-auto md:flex-row">
```
zepfietje, line 9: Hmm, why is it flex-col on mobile, @saade @danharrin?

danharrin: Probably so the stuff in the render hooks gets added above the tabs instead of below

saade: Because of the before and after content. On mobile, the tabs should be w-full hence no space for the before and after hooks to render in the same place it was on the desktop.

PR #7315, opened 2023-07-31
Introduce a simple repeater

ryangjchandler:
This PR introduces the initial piece of functionality from my "simple repeater" plugin into core.

A simple repeater can be created using the `->simple()` modifier on a regular `Repeater` object. Instead of calling `->schema()`, you instead call the `field()` method and provide a single `Field` object.

Right now, the UI isn't as minimal as the UI from my plugin, but I'm hoping that one of the @filamentphp/design guys can help out with that!


zepfietje:
Thanks, @ryangjchandler!

Regarding the `field()` method, doesn't it make sense to keep using `schema()` to allow passing two fields in a horizontal grid for example?
I'm pretty sure there's cases where you'd want to use a simple repeater with a couple of fields instead of just one.

Let me know if there's some thinking behind this that I'm missing though!

wychoong:
> Thanks, @ryangjchandler!
> 
> Regarding the `field()` method, doesn't it make sense to keep using `schema()` to allow passing two fields in a horizontal grid for example? I'm pretty sure there's cases where you'd want to use a simple repeater with a couple of fields instead of just one.
> 
> Let me know if there's some thinking behind this that I'm missing though!

simple repeater is to store as simple array
```php
["abc", "def", "ghi"]
# instead of
[{"field_1": "abc"}, {"field_1": "def"}, {"field_1": "ghi"}]
```

since now its a `->simple()` api in Repeater, I would suggest it should still able to use `->schema([])` but it will only take the first field component 
`->field()` can still remain, and possibly `->simple(isSimple: true, field: TextInput::make('item'))`

awcodes:
Wouldn't having a different underlying data structure break the app if it's ever changed in the future?

For example:
Simple repeater is used to handle an array of emails. Great it all works. 

Now 6 months later boss says we also need to have aliases for then. 

So, now I go in and remove ->simple() and switch to ->schema(). Now that repeater isn't loading any data because it doesn't know how to parse the data. Which means to fix it I'd have to run a script against the production db to update all the records that use that repeater.

wychoong:
> Wouldn't having a different underlying data structure break the app if it's ever changed in the future?

With the same scenario of adding a new field but instead of using json column, a new table and `->relationship` is used, does it still auto populate?

even if it does, I think when a change of data structure a script to migrate data is required

ziming:
Maybe can have a ->simpleLegacy() to keep it as single array Since sometime we could be passing that field to other libraries that expect single item array

zepfietje:
I've locked conversation on this PR to prevent a long thread of discussion as we'd still need to discuss the approach with the team internally.

ryangjchandler:
@zepfietje Any chance @filamentphp/design can take a look at this soon, or provide some pointers for design changes?

zepfietje:
Apart from the design changes, I think we should first decide on the final API for this?
I think we need 2 separate APIs: one to use a simple design, and one to change the storage format.

Thoughts?

ryangjchandler:
I don't think there needs to be any changes to the storage format. The whole point of the `simple()` API is that it explicitly stores values in a flat array, without nesting.

zepfietje:
There is definitely a use case for having the simple design with multiple fields though.
That's when you'd not want a flat array.

ryangjchandler:
I can't think of a use case myself, this PR is only focused on implementing the same behaviour found in my simple-repeater package.

awcodes:
My concern is that if the same component can store data in different ways then it'll create breaking issues should the user ever need to change it. To go from simple back to normal would break the field since the data couldn't be read back out and parsed correctly. 

I'm 100% for a simple repeater, but if it's underlying data structure isn't the same as the repeater, I think it should be a separate component.

ryangjchandler:
I think I'm going to close this PR, maintain a separate package still but make it open-source for Filament v3.

zepfietje:
I think it should be fine as long as the data structure is explicitly set using a separate API. No need to maintain a separate package if you ask me, @ryangjchandler. Just need to find the right approach and then I'm happy to take on the design part from there.

zepfietje:
Would be good to get @danharrin's thoughts on this.

danharrin:
I am happy with the PR being merged as it is, the design just needs to be fixed

danharrin:
Thanks Ryan

PR #7343, opened 2023-08-01
fix new global search render hooks(before,after)

bezhanSalleh:

They can be added either inside the `@if (filament()->isGlobalSearchEnabled())` check or outside. 
I prefer them to be outside so even if the global search is not enabled these hooks can be used. Just a thought


zepfietje:
Thanks!

PR #7347, opened 2023-08-01
Bump Rector in upgrade script to 0.17

duyphuongn:
This pull requests increases Rector version constraint in upgrade script (`packages/upgrade`) from `^0.15` to `^0.17` (latest at the time of this PR).

Test in `packages/upgrade/` directory:

```shell
cd filament/packages/upgrade/
./vendor/bin/rector process src --dry-run --config src/rector.php
                                                                                                                        
 [OK] Rector is done!                                                                                                   
                                                                                                                        
```

### DIFF
--- a/packages/upgrade/composer.json
+++ b/packages/upgrade/composer.json
@@ -10,7 +10,7 @@
     "require": {
         "php": "^8.1",
         "nunomaduro/termwind": "^1.0",
-        "rector/rector": "^0.15"
+        "rector/rector": "^0.17"
     },
     "autoload": {
         "psr-4": {


danharrin:
Please test the upgrade with an actual project, I have a feeling there is a reason why its locked to that version. What breaking changes are there?

ziming:
I dont think there is an actual reason for "locking" to that particular version actually

it is just how composer works

if a package version is below 1.0

composer will not bump the middle number in composer.lock when u composer update it

so u have to manually bump it in composer.json then composer update when the package is below 1.0

duyphuongn:
> Please test the upgrade with an actual project, I have a feeling there is a reason why its locked to that version. What breaking changes are there?

There isn't any breaking change with existing configured rules. With `rector.php` being in specific `packages/upgrade`, it affects only files within that dir and was OK with [0.17.7](https://github.com/rectorphp/rector/releases/tag/0.17.7) checks (as explained in PR).

PR #7361, opened 2023-08-01
Fix globally configuring filters layout

yassine-kessal:

**Fix: overwriting filters layout when it's not configured in component/resource table builder**

This issue overwrites the filter layout when nothing is configured in the component/resource table builder.

For example: when we try to set a global filter layout in a service provider using Table::configureUsing, it gets overwritten.

### DIFF
--- a/packages/tables/src/Table/Concerns/HasFilters.php
+++ b/packages/tables/src/Table/Concerns/HasFilters.php
@@ -51,8 +51,10 @@ trait HasFilters
             $this->filters[$filter->getName()] = $filter;
         }
 
-        $this->filtersLayout($layout);
-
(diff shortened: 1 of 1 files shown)


packages/tables/src/Table/Concerns/HasFilters.php
```diff
@@ -51,8 +51,10 @@ public function filters(array $filters, FiltersLayout | string | Closure | null
             $this->filters[$filter->getName()] = $filter;
         }
 
-        $this->filtersLayout($layout);
-
+        if($layout !== null) {
```
danharrin:
```suggestion
        if ($layout) {
```

PR #7374, opened 2023-08-02
Return suggestions when the $type is not set on SpatieTagsInput

sten:
If there is no type set on the SpatieTagsInput component and tags records have types, the suggestions do not work. This is because the suggestions query adds a condition that a type should be NULL if it is not set. By removing this condition, the suggestions always return all types if no type is set. I believe this is the correct behaviour.

### DIFF
--- a/packages/spatie-laravel-tags-plugin/src/Forms/Components/SpatieTagsInput.php
+++ b/packages/spatie-laravel-tags-plugin/src/Forms/Components/SpatieTagsInput.php
@@ -66,8 +66,7 @@ class SpatieTagsInput extends TagsInput
         return $tagClass::query()
             ->when(
                 filled($type),
-                fn (Builder $query) => $query->where('type', $type),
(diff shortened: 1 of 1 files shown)


zepfietje:
I believe this was done on purpose, but let's see what @danharrin thinks. Also, the `when()` call would no longer be needed.

danharrin:
If I'm going to accept this, there needs to be a way to return to the old behaviour. I don't know what an API for that would look like.

danharrin:
Please mark this as ready for review when that is done

danharrin:
Gonna merge this in quickly since it's kinda "breaking" and needs to be in an early v3 release before too many people depend on the "broken" behaviour. But it would be really appreciated if you PRed a way to make `type(null)` return to the old behaviour.

sten:
@danharrin thanks a lot! I will make a pr after my holidays.

PR #7384, opened 2023-08-02
Add select action to forms

lukas-frey:

This PR allows to add affix/suffix SelectActions to Form Fields. This might be useful in cases such as a currency selector for a given field, instead of having to use a full field for the currency, for example:

[Watch the video](https://github.com/filamentphp/filament/assets/10926334/5ef05af4-b033-4a40-8eed-f78e60316b8f)

On the right is the currency field you had to create before, on the left is  text field with a suffix select action for the currency.

I had to modify the SelectAction in the /actions package though, to include the proper state path. The Select Action will act as a  standalone sibling form field to the component it was added to.


zepfietje:
Thanks for creating this PR, @LukasFreyCZ!
Make sure to test it with `prefix`, `suffix`, `inlinePrefix`, `inlineSuffix`, additional affix actions, combinations of affix actions/icons/labels. Currently all of these combinations are supported in the design, so we should ensure this new feature still supports that.

packages/forms/docs/06-actions.md
```diff
@@ -224,6 +224,24 @@ Actions::make([
 
 <AutoScreenshot name="forms/layout/actions/anonymous/vertically-aligned-end" alt="Anonymous actions vertically aligned to the end" version="3.x" />
 
+## Form component Select action
+You may also use a SelectAction in your component's affix/suffix actions.
```
zepfietje, line 229:
```suggestion

You may also use a SelectAction in your component's affix/suffix actions.
```

lukas-frey:
> Thanks for creating this PR, @LukasFreyCZ! Make sure to test it with `prefix`, `suffix`, `inlinePrefix`, `inlineSuffix`, additional affix actions, combinations of affix actions/icons/labels. Currently all of these combinations are supported in the design, so we should ensure this new feature still supports that.

Hi! So I tested a few combinations and recorded a video of it. I forgot to remove the box shadow, but other than that all seems to work:


https://github.com/filamentphp/filament/assets/10926334/966c9ef3-7a67-430f-9fe8-4a637f664a0d


One thing I noticed is that when I set `inlineSuffix` or `inlinePrefix` when there's any action set, there's still the separator between the input and the actions. But that happens even with the default actions, not only with the new SelectAction.

mstfkhazaal:
How can test this PR?

packages/actions/resources/views/select-action.blade.php
```diff
@@ -12,7 +12,7 @@
         <x-filament::input.select
             :disabled="$isDisabled"
             :id="$id"
-            :wire:model.live="$getName()"
+            :attributes="$getExtraAttributeBag()"
```
zepfietje: Please wrap in `\Filament\Support\prepare_inherited_attributes()`.

packages/forms/resources/css/components/select-action.css
```diff
@@ -0,0 +1,3 @@
+.fi-fo-field-wrp .fi-ac-select-action .fi-input-wrapper {
+    @apply bg-transparent shadow-none ring-0;
+}
```
zepfietje: What's this needed for?

zepfietje (commented):
Thanks, @LukasFreyCZ.
Could you have a look at these couple comments?

packages/forms/resources/css/components/select-action.css
lukas-frey: > What's this needed for?

The SelectAction normally looks like a button. But it doesn't look good when as suffix / prefix action as a button. Neither did it look good in the hint. This just removes the background and outline.

All other prefix / suffix actions also don't have these styles applied by default.

lukas-frey:
> Thanks, @LukasFreyCZ. Could you have a look at these couple comments?

Yes, will change what is needed as soon as I can.

packages/forms/resources/css/components/select-action.css
zepfietje: Hmm, I think it's better to remove the wrapper instead, as we're nesting wrappers now...

mstfkhazaal:
Any update?

zepfietje:
It's been weekend, @mstfkhazaal; please don't put pressure on others (even if unintentional). 🙂

packages/forms/resources/css/components/select-action.css
lukas-frey: Oh, yeah. That works much better!

lukas-frey:
Ready for review. :) Also noticed that when the parent component is disabled, the select action would still work. That should be fixed now too.

packages/actions/resources/views/select-action.blade.php
```diff
@@ -1,28 +1,31 @@
 @php
     $id = $getId();
     $isDisabled = $isDisabled();
+    $attributes = $getExtraAttributeBag();
```
zepfietje, line 4: This is only used once, so there's no need to assign it to a variable. :)

zepfietje (changes requested):

packages/actions/resources/views/select-action.blade.php
lukas-frey: Oh, I thought that was a filament convention to assign them at the top. :D 

I will change it. However, I just realised that removing the input wrapper will not work.

The reason it was added as CSS is, that it uses the same `select-action` view as any other select action, to prevent duplicity. Now even the regular SelectActions (such as the LocaleSwitcher) won't have the input wrapper.

So either we need to revert back to the CSS, or add the input wrapper conditionally.

zepfietje: > Oh, I thought that was a filament convention to assign them at the top. :D

Generally only to prevent a getter from being called more than once.

> I will change it. However, I just realised that removing the input wrapper will not work.
> 
> The reason it was added as CSS is, that it uses the same `select-action` view as any other select action, to prevent duplicity. Now even the regular SelectActions (such as the LocaleSwitcher) won't have the input wrapper.
> 
> So either we need to revert back to the CSS, or add the input wrapper conditionally.

Hmm, the wrapper also handles the disabled state (background color).
What was the CSS for again?

lukas-frey: > Hmm, the wrapper also handles the disabled state (background color). What was the CSS for again?

If there's an input wrapper, it adds a background and a border to make it look like a standalone input. That works great when using as a Header Action (for example the Locale Switch).

However, since I added the SelectAction as a suffix action to form fields, they already have a background and a border.

So I had to remove it, which I did using the CSS from my first commit. That can also be done by simply removing the input wrapper, but the thing is:

the blade view is the same for both select actions, whether they are used as a header action or as a prefix/suffix action.

Maybe you have a better idea, but what comes to my mind currently is either using the CSS again or creating a method, say `hasInputWrapper()` which will return false for the form field SelectAction and then conditionally add/remove the input wrapper in the blade view.

zepfietje: Thanks for clarifying, would have to think about this.

zepfietje:
@danharrin, I feel like maybe this shouldn't be a select action after all, rather be a custom field (maybe as a plugin). The approach in this PR doesn't allow for non-live updating of the select value. I'd rather discuss this now than merge this in and run into limitations later. What do you think?

lukas-frey:
> @danharrin, I feel like maybe this shouldn't be a select action after all, rather be a custom field (maybe as a plugin). The approach in this PR doesn't allow for non-live updating of the select value. I'd rather discuss this now than merge this in and run into limitations later. What do you think?

The SelectAction never had the option for non-live updating. It is hardcoded there, as far as I remember it didn't even have the extra attributes bag.

I kept everything to the defaults like it was, the only thing I did was create a Action that can be added as a suffix action and made the required changes to make it look good (ie remove that button-feel).

zepfietje:
Yeah I know, @LukasFreyCZ, and that's kind of the point. I feel like this select shouldn't be a select action, as the goal of a select action is different.

lukas-frey:
> Yeah I know, @LukasFreyCZ, and that's kind of the point. I feel like this select shouldn't be a select action, as the goal of a select action is different.

Well, maybe if there was a way to use **any** form field as a suffix / prefix. That would actually be even better. That way you could not only use a Select there, but even a `ColorPicker` or so. 

That'd require bigger changes though, as there would need to be some kind of `compact` version of each form field, without   the surrounding labels, hint, helper texts and so on.

zepfietje:
So that's why I think this would potentially be better as a plugin. But let's wait to see what Dan thinks.

danharrin:
Since this can be completely added in a plugin, then yeah I think that would be the best idea. None of the core files actually get changed here in a way that you can't do in a plugin. There is a reason that we don't document SelectAction, it's because its not really an action, it's just something we use for the translatable package internally. Thanks though @LukasFreyCZ, you should just be able to 1:1 copy all of this into a plugin without too much work

Issue #7093, opened 2023-07-21
Livewire overrides `disabled` attribute


zepfietje:
`wire:loading.attr="disabled"` overrides conditional disabled state (found in at least checkbox list and radio components).

maharsh-mangal:
@zepfietje  hello, can i open a pr on this? i was going to make a branch and make one but i dont have permission. Can you direct me how to make contributions please if im missing something, thanks

zepfietje:
Check out our contribution guide: https://github.com/filamentphp/filament#contributing.

maharsh-mangal:
@zepfietje Please review my PR, thanks :)

sumardi:
Do we still need `wire:loading.attr="disabled"` in form components? According to [the docs](https://livewire.laravel.com/docs/actions#disabling-inputs-while-a-form-is-being-submitted),

> Livewire automatically disables the submit button and all form inputs inside the <form> element while a wire:submit action is being processed. This ensures that a form isn't accidentally submitted twice.

saade:
@sumardi not on fields I guess, but I think we need it in the submit button because other events can trigger a disabled state, like uploading a file through filepond for example

but good catch

sumardi:
> @sumardi not on fields I guess, but I think we need it in the submit button because other events can trigger a disabled state, like uploading a file through filepond for example
> 
> but good catch

filepond use alpine `x-bind:disabled`, not an issue if we remove `wire:loading.attr="disabled"`. Because Livewire v3 automatically add `disabled` to submit button and `readonly` to form inputs.

saade:
> > @sumardi not on fields I guess, but I think we need it in the submit button because other events can trigger a disabled state, like uploading a file through filepond for example
> > but good catch
> 
> filepond use alpine `x-bind:disabled`, not an issue if we remove `wire:loading.attr="disabled"`. Because Livewire v3 automatically add `disabled` to submit button and `readonly` to form inputs.

it uses x-bind:disabled only on the input IIRC, I think it does not apply to the submit button. The button should be disabled while uploading a file

zepfietje:
There's a lot more cases than just fields inside forms. E.g. also text input columns on tables.

sakanjo:
This bug is related to livewire v3, specifically [wire-loading.js](https://github.com/livewire/livewire/blob/main/js/directives/wire-loading.js#L10),

it should be fixed when the [PR](https://github.com/livewire/livewire/pull/6353) is merged.

[...]

PR #7391, opened 2023-08-02
fix for bug #7093 Livewire overrides disable in input components

maharsh-mangal:
This PR includes the fixes for the issue #7093, Livewire `wire:loading.attr = "disabled"` overrides disabled which is used dynamically which results in the toggle components on the view user screen become toggleable. 


maharsh-mangal:
committed with wrong account

PR #7393, opened 2023-08-02
Fix/issue 7093

maharsh-mangal:
This PR includes the fixes for the issue https://github.com/filamentphp/filament/issues/7093, Livewire wire:loading.attr = "disabled" overrides disabled which is used dynamically which results in the toggle components on the view user screen become toggleable.


zepfietje:
Hey thanks for attempting to fix that issue!
However, `readonly` and `disabled` really are different HTML attributes semantically.
The issue is probably a bug in Livewire, which requires a different fix.

PR #7398, opened 2023-08-02
Multiple domains per panel

pxlrbt:
Currently `->domain()` supports only a single domain. For stuff like multi-tenancy it's useful to provide an array. E.g. showing a panel to tenants only. 

With this PR `->domain()` will accept an array of domain.

Need feedback for:
- Should we introduce `->domains()` instead of allowing an array here. But that would clutter the class.
- I renamed `getDomain()` to `getDomains()` to reflect the return value. The method is public so this could already be a breaking change, although we only use it for web.php

### DIFF
--- a/packages/panels/routes/web.php
+++ b/packages/panels/routes/web.php
@@ -6,6 +6,7 @@ use Filament\Http\Controllers\Auth\LogoutController;
 use Filament\Http\Controllers\RedirectToHomeController;
 use Filament\Http\Controllers\RedirectToTenantController;
 use Illuminate\Support\Facades\Route;
+use Illuminate\Support\Arr;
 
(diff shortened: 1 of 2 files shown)


damms005:
A note on breaking change: since it will cause an API change, perhaps we should now not limit the parameters to `string` and `array` types only; we should also accept a closure that returns `string | array`. 

This will help in realtime evaluation of domains

martin-ro:
v3.0.9 causing `Method Filament\Panel::getDomain does not exist.`. Is this related?

aeq-dev:
After upgrading to this release v3.0.9, I'm getting this error : 
`Method Filament\Panel::getDomain does not exist.`

Issue #7129, opened 2023-07-21
Query string includes all available filters causing an error on refresh


archilex:
### Package

filament/filament

### Package Version

v3

### Laravel Version

v9.19

### Livewire Version

v3

### PHP Version

PHP 8.1

### Problem description

In v3 when you select a filter, the query string will then populate with **all** the available filters, setting the others to `null`. While this pollutes the query string, the bigger issue is that with certain types of filters if you then refresh the page you will get an error as `null` is not an acceptable value.

In v2, only the active filter would be present in the query string.

### Expected behavior

The query string should only include active filters

### Steps to reproduce

1. On the users page of the demo app filter by just one of the two dates, leaving the other date field alone.
2. The query string will populate with the unfiltered date setting it to `null`: `http://demov3.test/users?tableFilters[created_at][created_from]=2023-07-01&tableFilters[created_at][created_until]=null`
3. Refresh the page
4. Error: `Could not parse 'null': Failed to parse time string (null) at position 0 (n): The timezone could not be found in the database`

### Reproduction repository

https://github.com/archilex/demo

archilex:
You can pull down the repo, but you can also see this on the filament v3 demo as well.

zepfietje:
> You can pull down the repo, but you can also see this on the filament v3 demo as well.

Thanks, I've confirmed this on the demo indeed.

> In v2, only the active filter would be present in the query string.

I'm pretty sure this is caused by the change in how Livewire handles query parameters by default in v3.

archilex:
I agree. I've tried some work arounds but it seems to be a livewire issue. 

Probably need to add a failing test so they can work on it. I would, but I'm slammed for the next week.

zepfietje:
Is it really a Livewire bug instead of designed behavior though?

ariaieboy:
> Is it really a Livewire bug instead of designed behavior though?

where can we access to the V3 demo?

zepfietje:
> where can we access to the V3 demo?

Pull it in locally and use the 3.x branch.

archilex:
I don't think this is by design as it seems like LW3's new intention is to NOT show query string parameters since they got rid of `exclude` and now you have to use `keep` to have them appear in the url.

ariaieboy:
It's a bug in the livewire JS part if you set the history parameter to true it will not change the empty input to null. 
```php
class Test extends Component
{
    #[Url(history: true,keep: false)]
    public string $search;

    public function render()
    {
[...]

PR #7407, opened 2023-08-02
Fix filter query string

sumardi:

Possible fix for #7129.

### DIFF
--- a/packages/tables/src/Concerns/HasFilters.php
+++ b/packages/tables/src/Concerns/HasFilters.php
@@ -65,7 +65,7 @@ trait HasFilters
             $field->state($filterResetState[$fieldName] ?? match (true) {
                 is_array($state) => [],
                 is_bool($state) => false,
-                default => null,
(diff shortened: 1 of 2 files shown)


archilex:
Unfortunately, this does not appear to fix #7129. Not sure if it solves a different problem, but the main problem with #7129 is when there are more than one filter (like two dates filters), setting one adds the others to the query string with null values, which then can causes errors on a page refresh.

sumardi:
> Unfortunately, this does not appear to fix #7129. Not sure if it solves a different problem, but the main problem with #7129 is when there are more than one filter (like two dates filters), setting one adds the others to the query string with null values, which then can causes errors on a page refresh.

Erm, you're right. This PR only solves the issue for multiple `SelectFilter` and `TernaryFilter`. Filters with custom form components still have the issue. Form component default state need to be set as empty string `""`, not `null`. Unless Livewire v3 accept query param `=null` as empty.

PR #7412, opened 2023-08-02
Improve default state handling for boolean values

andrewdwallo:

This commit modifies the `default()` method to correctly handle boolean values, enhancing usability and user experience. Prior to this change, passing a boolean value as a default state could cause unexpected behavior due to the JavaScript interpretation of booleans. This could lead to confusion for users expecting boolean values to be applied correctly.

Now, boolean values are explicitly converted to string '1' (for true) or '0' (for false), ensuring proper handling. This provides a more intuitive and predictable functionality for components using default boolean states, aligning with user expectations in the UI. The change thus enhances user experience.

Beforehand, applying a boolean value to the default state could cause unexpected issues for some fields - mostly the Select Field and fields with the `options([])` array:

```php
Forms\Components\Select::make('is_active')
          ->boolean('Yes', 'No')
          ->default('1'),
Forms\Components\Select::make('enabled')
          ->options([
                   '1' => 'Yes',
                   '0' => 'No',
           ])
           ->default(1),
Forms\Components\Select::make('is_company')
           ->boolean('Yes', 'No')
           ->default(true),
Forms\Components\Select::make('is_person')
           ->options([
                    true => 'Yes',
                    false => 'No',
           ])
           ->default(true),
Forms\Components\Select::make('entity')
           ->options([
                    'company' => 'Company',
                    'person' => 'Person',
          ])
          ->default('company'),
```

Screenshot 2023-08-02 at 6 45 47 PM

Now a user can use all boolean accepted values instead of just an int or a string passed into the `default()` method:

Screenshot 2023-08-02 at 6 50 12 PM

### DIFF
--- a/packages/forms/src/Components/Concerns/HasState.php
+++ b/packages/forms/src/Components/Concerns/HasState.php
@@ -85,6 +85,10 @@ trait HasState
 
     public function default(mixed $state): static
     {
+        if (is_bool($state)) {
+            $state = $state ? '1' : '0';
+        }
+
         $this->defaultState = $state;
         $this->hasDefaultState = true;
 
--- a/packages/forms/src/Components/Concerns/HasState.php
+++ b/packages/forms/src/Components/Concerns/HasState.php
@@ -85,10 +85,6 @@ trait HasState
 
     public function default(mixed $state): static
     {
-        if (is_bool($state)) {
-            $state = $state ? '1' : '0';
-        }
-
         $this->defaultState = $state;
         $this->hasDefaultState = true;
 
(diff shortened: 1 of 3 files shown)


andrewdwallo:
I also wanted to mention,... and I looked into this for a while but couldn't figure out what the problem was, but for boolean values in the EditRecord page, the saved record won't display.

All of the saved values here are equal to true.
Screenshot 2023-08-02 at 6 55 02 PM

zepfietje:
Isn't the actual issue in the Blade view instead? I doubt this is the right fix.

andrewdwallo:
@zepfietje yes it is in the blade view and it can be fixed as long as you don't want to use strict comparison. But also this just handles it whenever needed without having to think about it and whenever users are building custom fields they can use strict comparison without issues happening.

Issue #7324, opened 2023-07-31
Table Action triggered before form


mrgla55:
### Package

filament/tables

### Package Version

v3.0.0-beta25

### Laravel Version

v10.16.1

### Livewire Version

v3.0.0-beta.5

### PHP Version

PHP 8.2.7

### Problem description

When defining a closure action within the Table Action definition, it is called correctly after submitting the form. But if I reference the function with a string - it calls the action before the form is shown.

Result : Unable to resolve dependency [Parameter #1 [ <required> array $data ]] 

### Expected behavior

The form should show before the action is called. The data of the form passed to the action function

### Steps to reproduce

(The commented variations work correctly)

Filament\Tables\Actions\Action::make('Payment')
                ->button()
                // ->action(fn(Attendance $attendance, array $data) => dd($data))
                // ->action(function(Attendance $attendance, array $data) {
                //     dd($data);
                // })
                ->action('payment')
                ->form([
                    TextInput::make('amount')
                        ->numeric()
                        ->required(),

    public function payment(Attendance $attendance, array $data)
    {
        dd($data);
    }

### Reproduction repository

https://github.com/mrgla55/filament-testing

danharrin:
Hey, where in the docs does it say this is supported? Because it isn't, as far as I know

mrgla55:
Ok. Just checked the docs, its always Closures, but the action function is defined to accept string

    public function action(Closure | string | null $action): static

I had used the string previously in v2 to separate the action logic from the action definition

danharrin:
Okay, we have actually changed it in v3 so that the string gets fed directly into the `wire:click`, which is something you couldn't do before. And that explains why the modal doesn't open, because we can't intercept it. To do what you want to accomplish, try `->action(Closure::fromCallable([$this, 'payment']))`

mrgla55:
thanks @danharrin, planning on moving the logic into the model or something similar. Great job on v3 - fills a huge list of great enhancements

PR #7416, opened 2023-08-03
Add more changes to upgrade guides: masks, panel configuration, action namespaces, action function calls

nikspyratos:
👋 Thought I'd PR some undocumented changes I've had to make on my V2->V3 app.

- Numeric masks are gone, use Alpine instead
- Action namespace move
- Action function calls no longer work (and were not officially supported)


zepfietje:
Thanks for taking the effort to document these changes!

There's a lot more APIs that have been marked as deprecated but still work until they're removed in a next major version. We don't document these now.

Regarding passing a method name to the `action()` method of an action, that should still work if I'm correct.

nikspyratos:
Deprecations: For sure, that's just what I picked up along the way. Can probably remove in favour of a more comprehensive doc addition later.

> Regarding passing a method name to the action() method of an action, that should still work if I'm correct.

From my understanding of #7324, even if it it does work, it's a bit janky. I had that issue happen after upgrading in my row-level actions, and it was only fixed by moving the logic into a closure.

zepfietje:
> From my understanding of #7324, even if it it does work, it's a bit janky. I had that issue happen after upgrading in my row-level actions, and it was only fixed by moving the logic into a closure.

The use case in that issue really is an edge case though. The method name API is still officially supported and it's quite obvious that the behavior in that issue occurs. It's not what the dev expected, but it totally makes sense that that is what's happening. So it's designed behavior. Shouldn't be documented as deprecated in the docs.

nikspyratos:
Not sure I understand. The issue boiled down to `->action('someFunctionName')` not working because the function ends up being run when the action modal loads, not on submission. The exact same thing happened to me. Dan explicitly mentions in that issue that that way of doing this isn't supported, and no workaround to get it working other than the `Closure::fromCallable` way was suggested. To me it really doesn't sound like this works at all in V3.

As for edge case, kind of. That's why I put it in the medium impact section. Actions can have some lengthy logic in them, and in some codebases it's preferred to move that logic out to another function. All of those would break in this scenario.

zepfietje:
It's really only ever an issue when using the method name API in combination with an action form. Maybe worth to add a note to its docs section, but definitely shouldn't be marked unsupported as it works as intended. Dan's suggested workaround is only to fix that specific case.

nikspyratos:
@zepfietje Fair. I've updated the PR to specifically reference using forms, and removed the deprecation section.

packages/forms/docs/10-upgrade-guide.md
```diff
@@ -56,6 +56,10 @@ php artisan vendor:publish --tag=filament-config --force
 rm config/forms.php
 ```
 
+#### Panel configuration is now in each AppPanelProvider
```
danharrin: Belongs in the panels upgrade guide, high impact section. "AppPanelProvider" can just be "panel provider"

packages/forms/docs/10-upgrade-guide.md
```diff
@@ -101,3 +105,44 @@ An easy way to upgrade your code quickly is to find and replace:
 
 - `Closure $get` to `\Filament\Forms\Get $get`
 - `Closure $set` to `\Filament\Forms\Set $set`
+
+#### TextInput numeric masks are replaced with Alpine.js masks
```
danharrin: Not just numeric masks, the entire mask object has been replaced. Please use `` for code snippets.

packages/forms/docs/10-upgrade-guide.md
```diff
@@ -101,3 +105,44 @@ An easy way to upgrade your code quickly is to find and replace:
...
+JS))
+```
+If you were using input masks for money related fields, Alpine already has [builtin helpers](https://alpinejs.dev/plugins/mask#money-inputs) for it:
+```
+->mask(RawJs::make(<<<'JS'
+    $money($input, '.', ',', 2)
+JS))
+```
+
+#### Action execution with forms
```
danharrin: Doesn't really belong in the forms upgrade guide, should probably be in a new actions upgrade guide file that you'll need to create.

danharrin (changes requested):

danharrin:
Also merge conflicts to resolve. Please mark this as ready to review when completed. Thanks.

packages/forms/docs/10-upgrade-guide.md
nikspyratos: Ah I see my confusion here. I will edit this.

IMO would useful to have one upgrade page instead of having to chase down multiple, or at least link them to each other. While Filament supports modular usage, structuring the docs around that instead of an "ordinary" use case involving most of the packages makes it confusing to navigate.

packages/forms/docs/10-upgrade-guide.md
nikspyratos: >Please use `` for code snippets.

Three backticks is the proper markdown method for code snippets, and is used throughout the existing docs, so I'm not sure if you meant something else here.

packages/forms/docs/10-upgrade-guide.md
danharrin: Yeah, the problem is that people who are not using the panel builder at all will then have to skip past all the stuff that isn't relevant to them

packages/forms/docs/10-upgrade-guide.md
danharrin: I mean `TextInput` here instead of TextInput, for example

nikspyratos:
Just to double check - are there any extra steps needed to get new files loaded in the docs? Had a quick search in the monorepo and didn't find anything.

danharrin:
I do the build process for you once the docs are ready

packages/actions/docs/10-upgrade-guide.md
```diff
@@ -0,0 +1,71 @@
+---
+title: Upgrading from v2.x
+---
+
+> If you see anything missing from this guide, please do not hesitate to [make a pull request](https://github.com/filamentphp/filament/edit/3.x/packages/panels/docs/14-upgrade-guide.md) to our repository! Any help is appreciated!
```
danharrin:
```suggestion
> If you see anything missing from this guide, please do not hesitate to [make a pull request](https://github.com/filamentphp/filament/edit/3.x/packages/actions/docs/14-upgrade-guide.md) to our repository! Any help is appreciated!
```

packages/actions/docs/10-upgrade-guide.md
```diff
@@ -0,0 +1,71 @@
...
+
+```bash
+composer require filament/upgrade:"^3.0-stable" -W --dev
+
+vendor/bin/filament-v3
+```
+
+Make sure to carefully follow the instructions, and review the changes made by the script. You may need to make some manual changes to your code afterwards, but the script should handle most of the repetitive work for you.
+
+A new `app/Providers/Filament/*PanelProvider.php` file will be created, and the configuration from your old `config/filament.php` file should be copied. Since this is a [Laravel service provider](https://laravel.com/docs/providers), it needs to be registered in `config/app.php`. Filament will attempt to do this for you, but if you get an error while trying to access your panel then this process has probably failed. You can manually register the service provider by adding it to the `providers` array.
```
danharrin:
```suggestion
```

packages/actions/docs/10-upgrade-guide.md
```diff
@@ -0,0 +1,71 @@
...
+
+```json
+"minimum-stability": "dev",
+```
+
+### Low impact changes
+
+#### Action execution with forms
+
+In v2, you were able to externalise action logic to a function, like so:
```
danharrin:
```suggestion
In v2, if you passed a string to the `action()` function, and used a modal, the method with that name on the class would be run when the modal was submitted:

```

packages/actions/docs/10-upgrade-guide.md
```diff
@@ -0,0 +1,71 @@
...
+```php
+ Action::make('import_data')
+    ->action('importData')
+    ->form([
+        FileUpload::make('file'),
+    ])
+```
+In v3, if your Action uses a form like the above example, you'll need to change how your action logic is called. Otherwise, the function will be called when the action modal attempts to load, which may cause errors.
```
danharrin:
```suggestion

In v3, passing a string to the `action()` hard-wires it to run that action when the trigger button is clicked, so it does not open a modal.
```

packages/actions/docs/10-upgrade-guide.md
```diff
@@ -0,0 +1,71 @@
...
+```php
+ Action::make('import_data')
+    ->action('importData')
+    ->form([
+        FileUpload::make('file'),
+    ])
+```
+In v3, if your Action uses a form like the above example, you'll need to change how your action logic is called. Otherwise, the function will be called when the action modal attempts to load, which may cause errors. 
+
+It is recommended to place your logic in a closure instead. See the [Action documentation](https://filamentphp.com/docs/3.x/actions/overview) for more information.
```
danharrin:
```suggestion
It is recommended to place your logic in a closure function instead. See the [documentation](overview) for more information.
```

packages/actions/docs/10-upgrade-guide.md
```diff
@@ -0,0 +1,71 @@
...
+
+If you absolutely need the action logic to be in a separate function, you can use `Closure::fromCallable()`:
```
danharrin:
```suggestion
One easy alternative is using `Closure::fromCallable()`:

```

packages/forms/docs/10-upgrade-guide.md
```diff
@@ -102,6 +102,27 @@ An easy way to upgrade your code quickly is to find and replace:
 - `Closure $get` to `\Filament\Forms\Get $get`
 - `Closure $set` to `\Filament\Forms\Set $set`
 
+#### `TextInput` masks are replaced with Alpine.js masks
+
+Filament v2 had a fluent closure mask syntax for managing more complex input masks. In v3 you can use Alpine.js masking instead. So instead of
```
danharrin:
```suggestion
Filament v2 had a fluent mask object syntax for managing input masks. In v3 you can use Alpine.js' masking syntax instead. Please see the [input masking documentation](fields/text-input#input-masking) for more information.
```

packages/forms/docs/10-upgrade-guide.md
```diff
@@ -102,6 +102,27 @@ An easy way to upgrade your code quickly is to find and replace:
...
+->mask(RawJs::make(<<<'JS'
+    $input.startsWith('34') || $input.startsWith('37') ? '9999 999999 99999' : '9999 9999 9999 9999'
+JS))
+```
+If you were using input masks for money related fields, Alpine already has [builtin helpers](https://alpinejs.dev/plugins/mask#money-inputs) for it:
+```php
+->mask(RawJs::make(<<<'JS'
+    $money($input, '.', ',', 2)
+JS))
+```
```
danharrin:
```suggestion
```

packages/panels/docs/14-upgrade-guide.md
```diff
@@ -64,6 +64,8 @@ A new `app/Providers/Filament/AdminPanelProvider.php` file will be created, read
 
 Most configuration transfer is very self-explanatory, but if you get stuck please refer to the [configuration documentation](configuration).
 
+For configuration done via `Filament::serving` (for theme customisation, navigation groups, user menu items, etc), this will need to be ported to the new panel provider. Consult the [Configuration](https://filamentphp.com/docs/3.x/panels/configuration), [Navigation](https://filamentphp.com/docs/3.x/panels/navigation) and [Themes](https://filamentphp.com/docs/3.x/panels/themes) documentation for the new syntax.
```
danharrin:
```suggestion
This will especially affect configuration done via the `Filament::serving()` method, which was used for theme customization, navigation and menu registration. Consult the [configuration](configuration), [navigation](navigation) and [themes](themes) documentation sections.
```

danharrin (changes requested):

packages/forms/docs/10-upgrade-guide.md
```diff
@@ -102,6 +102,10 @@ An easy way to upgrade your code quickly is to find and replace:
 - `Closure $get` to `\Filament\Forms\Get $get`
 - `Closure $set` to `\Filament\Forms\Set $set`
 
+#### `TextInput` masks are replaced with Alpine.js masks
```
danharrin:
```suggestion
#### `TextInput` masks now use Alpine.js' masking package
```

danharrin:
Thanks @nikspyratos!

PR #7425, opened 2023-08-03
Add tenant related exceptions and update middleware

damms005:
- [x] New functionality has been documented or existing documentation [has been updated to reflect changes](https://github.com/filamentphp/filament/pull/7458).

Currently, an HTTP 404 error is used in place of the implementation errors that these new exceptions seek to address.

NonTenantModelException and TenantAuthorizationException make it easy to debug issues related to why tenancy may not work.

They each provide contextual information about specific tenancy implementation error and provides helpful note on how to fix them.

### DIFF
--- a/packages/panels/src/Exceptions/NonTenantModelException.php
+++ b/packages/panels/src/Exceptions/NonTenantModelException.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Filament\Exceptions;
+
+use Exception;
+use Throwable;
+
+class NonTenantModelException extends Exception
+{
+    final public function __construct(string $message = '', int $code = 0, ?Throwable $previous = null)
+    {
+        parent::__construct($message, $code, $previous);
+    }
+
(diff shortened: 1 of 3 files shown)


danharrin:
Hmm, while extra information is good, I don't want these to have to be exceptions that devs then need to catch. Otherwise, they will end up being reported as errors. I think this is the correct behaviour already, similar to any other 404 that you might experience when a page doesn't exist. Feel free to push back on this if I'm wrong.

damms005:
I understand that this is how you feel the API should be. It was a paper cut for me; which was why I offered the PR. I got a 404 error in dev and I was unsure about the part of the multi-tenancy flow that caused it.

As tenancy in Filament has more than one paths to a 404 error, being explicit would have saved me some dev time trying to figure out the cause of same.

danharrin:
I totally understand that, but I don't think that throwing an exception, which would then get logged in production, is the correct approach here

Issue #7401, opened 2023-08-02
TextEntry::make()->placeholder('-'); not working for nullable columns


CodeWithDennis:
### Package

filament/filament

### Package Version

v3.0.3

### Laravel Version

v10.10

### PHP Version

PHP 8.2

### Problem description

```PHP
TextEntry::make('name')->placeholder('-'),
```

When the column is `null` it won't trigger the placeholder

### Expected behavior

The placeholder should show in this case: `-`

### Steps to reproduce

Add a `TextEntry::make('column_name_here')->placeholder('-')` with a column that currently has a `null` value.

ashleyhood:
@zepfietje I can confirm that this is an issue.

The problem is in the view when looping through the data:

https://github.com/filamentphp/filament/blob/41897f7f56035a714d8753cf7ea775eb1d5b4709/packages/tables/resources/views/columns/text-column.blade.php#L64

Since the array will be empty, it will skip the `foreach` loop.

A possible solution is after:

https://github.com/filamentphp/filament/blob/41897f7f56035a714d8753cf7ea775eb1d5b4709/packages/tables/resources/views/columns/text-column.blade.php#L38

add the following code:

```PHP
if (blank($arrayState) && filled($placeholder)) {
    $arrayState = \Illuminate\Support\Arr::prepend($arrayState, '');
}
```

A work around is to use `->default` on the TextColumn instead of `->placeholder`.

danharrin:
@ashleyhood can you make a PR then please?

PR #7439, opened 2023-08-03
fix: text column placeholder

### LINKED ISSUE #7823: placeholder() not working on Relationship / TextColumn::make('relationship.name')
### Package

filament/filament

### Package Version

v3.0-stable

### Laravel Version

10.10

### Livewire Version

_No response_

### PHP Version

8.2.7

### Problem description

TextColumn::make('**relationship**.column')
    ->placeholder('NOT ASSIGNED YET')
    
Doesn't work when it is using in relationship 

however it works when it is without relationship
TextColumn::make('column')
    ->placeholder('NOT ASSIGNED YET')

### Expected behavior

it should display  "Never"

### Steps to reproduce

 Try using a placeholder('Never') on a relationship column and make the relationship null
TextColumn::make('**relationship**.column')
    ->placeholder('Never')
    
image
image

### Reproduction repository

https://github.com/abdulmejidshemsu/filament-relationship

### Relevant log output

_No response_

ashleyhood:
This solves #7401


danharrin:
Hey, where does `$placeholder` come from here? Have you tested this?

ashleyhood:
I have tested this on my local dev machine.

The `$placeholder` comes from:

https://github.com/filamentphp/filament/blob/3e8455b12685463284aec9616802ad2c338a717a/packages/tables/src/Columns/Concerns/CanFormatState.php#L32

This is why it is wrapped in `filled()` to make sure it only does this if a placeholder is set and the state is `null`.

sumardi:
The issue is about `TextEntry`. You should add this in `text-entry.blade.php` as well. And can we prepend `null` instead of empty string `""`. It is better if we want to format state, for example 

```php
->formatStateUsing(fn ($state) => $state?->format('d F Y'))
```

danharrin:
@ashleyhood I'm sorry I don't understand, a protected property should not be accessible in the view like that. Are you on the latest version?

ashleyhood:
I'm on Filament v3.0.7 and Livewire v3.0.0-beta.7.

That is really weird, because it does read the `$placeholder ` and only changes the `$arrayStates` if it is filled!?!

How would I track down where this is being set if it is not in the `CanFormatState` trait? 

Would be helpful to have `livewire/wiretap` :stuck_out_tongue_winking_eye:

ashleyhood:
On second thoughts, isn't this okay to access the `$placeholder` property because it is being set in the blade `@php` directive and therefore being rendered before sending the request?

ashleyhood:
My bad :confused: 

This is not working as it is returning:

```PHP
Filament\Tables\Columns\TextColumn::placeholder(Closure|string|null $placeholder): static
```

Which will always be true.

We could only check for `blank($arrayState)` or we make the `$placeholder` accessible from a getter

```PHP
public function getPlaceholder(): string | Closure | null
{
    return $this->placeholder;
}

and then check if it is `filled`.

**Comment by ashleyhood** [comment]
This will also need to be add to the InfoLists `text-entry.blade.php` as mentioned by @sumardi above.

Should that be in a separate PR or could it be added to this PR?

**Comment by danharrin** [comment]
Yeah we can make a `getPlaceholder()`. And please make the changes to infolists here instead of a new PR.

File: packages/infolists/resources/views/components/text-entry.blade.php [hunk 1]
```diff
@@ -37,6 +37,10 @@
         }
 
         $arrayState = \Illuminate\Support\Arr::wrap($arrayState);
+
+        if (blank($arrayState) && filled($getPlaceholder())) {
+            $arrayState = \Illuminate\Support\Arr::prepend($arrayState, '');
```
danharrin: Doesn't this need to prepend the actual placeholder as state instead of an empty string? Also, can't it just reassign `$arrayState` completely into a new array since its already blank, instead of prepending

ashleyhood: This needs to be an empty string because when we call `formatState ()` on line 69 it expects a `blank()` state:

https://github.com/filamentphp/filament/blob/64811c8be6a1c7cb55bf1eb5fc76f69c66fb3c6c/packages/tables/src/Columns/Concerns/CanFormatState.php#L226

You are right about just assigning a new array and that makes sense.

danharrin: Lets use `null` instead of an empty string please

danharrin: Potentially, we don't want the placeholder to run through `formatState()` though 🤔 It might need different styling, for example. Maybe we should use an if statement in the view to decide whether or not to render a placeholder. And then remove the placeholder application from the `formatState()` method

ashleyhood: I think that looking at how the placeholder is formatted might be for another PR?

This PR was about trying to address how a placeholder works as intended only if the state is an empty string but not if the state is `null`.

ashleyhood:
I have removed the Arr::wrap that was evaluating `null` values as an empty array. This keeps the same results for an empty string and allows a `null` value to give same results as an empty string.

ashleyhood:
This also brings up the fundamental difference between `$defaultState` and `$placeholder` as there seems to be an overlap. For instance, if the state is `null` and both `$defaultState` and `$placeholder` are set then `$defaultState` will be used but if the state is an empty string then `$placeholder` is used.
Why would an empty string be replaced by the `$placeholder`? Wouldn't the expectation be the same as the `$defaultState`?

danharrin:
`$placeholder` should probably have different styling and is not simply a "formatting" thing, whereas `$defaultState` acts as if it came from the database

ashleyhood:
> `$placeholder` should probably have different styling and is not simply a "formatting" thing, whereas `$defaultState` acts as if it came from the database

I see your point. Would you be happy to merge this fix and then another PR could address the issue of having a placeholder that accepted it's own styling.

DanielSpravtsev:
Any news about state of this PR? Maybe help needed? @danharrin

danharrin:
Work needs to be done on it, but by the time I have time to write instructions I might as well do it myself

danharrin:
Alright this should be fixed up now for all columns and entries, they have a differently-styled placeholder

faizananwerali:
Not working now for me in Infotlist with relationship. It was working before. The last working version was v3.0.42, after that, it has not been fixed. tried every single version to find the last working version.

```php
TextEntry::make('profile.father_name')
    ->label('Father Name')
    ->placeholder('N/A'),
TextEntry::make('profile.father_occupation')
    ->label('Father Occupation/Profession')
    ->placeholder('N/A'),
TextEntry::make('profile.grand_father_name')
    ->label('Grand Father Name')
    ->placeholder('N/A'),
TextEntry::make('profile.mother_name')
    ->label("Mother's Name")
    ->placeholder('N/A'),
TextEntry::make('profile.mother_from')
    ->label("Mother's Home Town")
    ->placeholder('N/A'),
TextEntry::make('profile.mother_occupation')
    ->label("Mother's Occupation/Profession")
    ->placeholder('N/A'),
TextEntry::make('profile.guardian_phone_number')
    ->label('Guardian Phone Number')
    ->placeholder('N/A'),
TextEntry::make('profile.brothers')
    ->label('No. of Brothers')
    ->placeholder('N/A'),
TextEntry::make('profile.sisters')
    ->label('No. of Sisters')
    ->placeholder('N/A'),
TextEntry::make('profile.brothers_married')
    ->label('No. of Brothers Married')
    ->placeholder('N/A'),
TextEntry::make('profile.sisters_married')
    ->label('No. of Sisters Married')
    ->placeholder('N/A'),
```

This is how it looks.

Screenshot 2023-09-12 at 6 12 54 PM


Right now I using `formatStateUsing` to resolve this temporarily.

```php
TextEntry::make('profile.father_name')
    ->label('Father Name')
    ->formatStateUsing(fn (User $record): string => $record->profile->father_name ?? 'N/A')
    ->placeholder('N/A'),
TextEntry::make('profile.father_occupation')
    ->label('Father Occupation/Profession')
    ->formatStateUsing(fn (User $record): string => $record->profile->father_occupation ?? 'N/A')
    ->placeholder('N/A'),
TextEntry::make('profile.grand_father_name')
    ->label('Grand Father Name')
    ->formatStateUsing(fn (User $record): string => $record->profile->grand_father_name ?? 'N/A')
    ->placeholder('N/A'),
TextEntry::make('profile.mother_name')
    ->label("Mother's Name")
    ->formatStateUsing(fn (User $record): string => $record->profile->mother_name ?? 'N/A')
    ->placeholder('N/A'),
TextEntry::make('profile.mother_from')
    ->label("Mother's Home Town")
    ->formatStateUsing(fn (User $record): string => $record->profile->mother_from ?? 'N/A')
    ->placeholder('N/A'),
TextEntry::make('profile.mother_occupation')
    ->label("Mother's Occupation/Profession")
    ->formatStateUsing(fn (User $record): string => $record->profile->mother_occupation ?? 'N/A')
    ->placeholder('N/A'),
TextEntry::make('profile.guardian_phone_number')
    ->label('Guardian Phone Number')
    ->formatStateUsing(fn (User $record): string => $record->profile->guardian_phone_number ?? 'N/A')
    ->placeholder('N/A'),
TextEntry::make('profile.brothers')
    ->label('No. of Brothers')
    ->formatStateUsing(fn (User $record): string => $record->profile->brothers ?? 'N/A')
    ->placeholder('N/A'),
TextEntry::make('profile.sisters')
    ->label('No. of Sisters')
    ->formatStateUsing(fn (User $record): string => $record->profile->sisters ?? 'N/A')
    ->placeholder('N/A'),
TextEntry::make('profile.brothers_married')
    ->label('No. of Brothers Married')
    ->formatStateUsing(fn (User $record): string => $record->profile->brothers_married ?? 'N/A')
    ->placeholder('N/A'),
TextEntry::make('profile.sisters_married')
    ->label('No. of Sisters Married')
    ->formatStateUsing(fn (User $record): string => $record->profile->sisters_married ?? 'N/A')
    ->placeholder('N/A'),
```


This is how now it looks.

Screenshot 2023-09-12 at 6 14 04 PM

 
 `$this->record` is the `User` Model. I've set the record this way.
 
 ```php
    public function infolist(Infolist $infolist): Infolist
    {
        return $infolist
            ->record($this->record)
            ->schema([
                Section::make(__('Family Info'))
                    ->heading('')
                    ->schema([
                        Grid::make(3)
                            ->schema([
                                // ....
                            ])
                    ])
        ]);
    }
 ```

danharrin:
Please open an issue as I cannot reproduce it

PR #7457, opened 2023-08-04
Update 08-appearance.md

OliverTe:
Updating import path from app.css to index.css on Filament 3


zepfietje:
Thanks, though I think that section on theming needs more updates. Also not sure if this instance should be `index.css` or `theme.css`, but will leave it to Dan.

danharrin:
You're PRing this to Filament 2? I don't understand

PR #7461, opened 2023-08-04
Add prefix and suffix for tag

sawirricardo:

This will help folks who need to add prefix, but don't want to type it manually.

image

### DIFF
--- a/packages/forms/src/Components/TagsInput.php
+++ b/packages/forms/src/Components/TagsInput.php
@@ -71,14 +71,14 @@ class TagsInput extends Field implements Contracts\HasNestedRecursiveValidationR
         $this->placeholder(__('filament-forms::components.tags_input.placeholder'));
     }
 
-    public function tagPrefix(string | Closure $prefix): static
+    public function tagPrefix(string | Closure | null $prefix): static
     {
         $this->tagPrefix = $prefix;
 
         return $this;
     }
 
-    public function tagSuffix(string | Closure $suffix): static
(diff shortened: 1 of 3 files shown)


danharrin:
This should probably be `tagPrefix()` and `tagSuffix()` rather than the other way round. Also for the properties and getter methods. Also, there shouldn't be any default `''` for the setter methods, its a required parameter

packages/forms/src/Components/TagsInput.php
```diff
@@ -67,6 +71,20 @@ protected function setUp(): void
         $this->placeholder(__('filament-forms::components.tags_input.placeholder'));
     }
 
+    public function tagPrefix(string | Closure $prefix): static
```
danharrin:
```suggestion
    public function tagPrefix(string | Closure | null $prefix): static
```

packages/forms/src/Components/TagsInput.php
```diff
@@ -67,6 +71,20 @@ protected function setUp(): void
...
+    {
+        $this->tagPrefix = $prefix;
+
+        return $this;
+    }
+
+    public function tagSuffix(string | Closure $suffix): static
```
danharrin:
```suggestion
    public function tagSuffix(string | Closure | null $suffix): static
```

danharrin:
Can you add some basic docs to `tags-input.md` please?

sawirricardo:
Updated the docs. Do you think the explanation is clear enough?

danharrin:
Thanks

PR #7478, opened 2023-08-05
Add Page public method to call model static method

killjin:

A method that can extend page components in the model.


danharrin:
Nononononono, major security issue there, I'm sorry. That means any user would be able to call any model function...

Issue #7377, opened 2023-08-02
Unable to find component: [app.filament.pages.tenancy.register]


mstfkhazaal:
### Package

filament/filament

### Package Version

v3.0.2

### Laravel Version

v10.0.0

### Livewire Version

v3.0.0

### PHP Version

8.2.4 

### Problem description

I Edit Register Tenant but when i add FileUpload i get this error
https://github.com/filamentphp/filament/assets/47450090/8174369a-1134-42e5-9e6d-b304254ac9c3

[this is file](https://github.com/mstfkhazaal/onyx/blob/master/app/Filament/Pages/Tenancy/RegisterEntity.php)

I have

### Expected behavior

The code works without any problems

### Steps to reproduce

I searched for a similar problem and couldn't find it

### Reproduction repository

https://github.com/mstfkhazaal/onyx

telkins:
This is happening to at least two other people: https://discord.com/channels/883083792112300104/1136210890987929682

In my case, it's happening with 3.0.7.

danharrin:
@telkins please provide a reproduction repository

telkins:
> @telkins please provide a reproduction repository

Thx.  I will try to get one set up asap.

mstfkhazaal:
I find the solution for this
Because the the page is not in folder of panel 
Move the file to folder panel and it work

In my case a move Register page from App\Filament\Pages\Tenancy
To
App\Filament\App\Pages\Tenancy

App is name of my panel

Is not a bug because the panel access to specific folder

This note should be added to the documentation

telkins:
> @telkins please provide a reproduction repository

@danharrin Here is a reproduction repository: https://github.com/telkins/tenancy-bug-demo

Steps to reproduce:

- migrate
- create filament user
- log in
- point browser to http://localhost/employers
- enter name for new organization
- you should see the error when you submit the register organization form

**Please note that I have not yet tried @mstfkhazaal 's suggestion...yet.  I'll update once I do.**

mstfkhazaal:
> > @telkins please provide a reproduction repository
> 
> @danharrin Here is a reproduction repository: https://github.com/telkins/tenancy-bug-demo
> 
> Steps to reproduce:
> 
> * migrate
> * create filament user
> * log in
> * point browser to http://localhost/employers
> * enter name for new organization
> * you should see the error when you submit the register organization form
> 
> **Please note that I have not yet tried @mstfkhazaal 's suggestion...yet. I'll update once I do.**

[...]

PR #7481, opened 2023-08-05
Add to documentation: Move File to panel directory

mstfkhazaal:
https://github.com/filamentphp/filament/issues/7377

### DIFF
--- a/packages/panels/docs/10-tenancy.md
+++ b/packages/panels/docs/10-tenancy.md
@@ -60,6 +60,19 @@ You'll also want users to be able to [register new teams](#adding-a-tenant-regis
 
 ## Adding a tenant registration page
 
+Note: If you have multiple panels, change the page directory to the panel directory. 
+
+For example, if you have two panels (Admin & Employers), the structure is:
+```
+.
++-- Filament
+|   +-- Pages
+|   |   +-- ....
+|   +-- Employers
+|   |   +-- RegisterEmployers.php
+|   |   +-- EditEmployersProfile.php
+```
+
 A registration page will allow users to create a new tenant.
 
 When visiting your app after logging in, users will be redirected to this page if they don't already have a tenant.


danharrin:
I feel like this is not the right place, and really quite basic information that you should be able to understand if you're using multiple panels. The multi-panel docs explain that pages have their own directory and namespace, and the tenancy docs explain that this file can be put anywhere as long as it is registered correctly.

bilogic:
I registered it as `->discoverPages(in: app_path('Filament/A/B/C/Pages'), for: 'App\\Filament\\A\\B\\C\\Pages')`, the page was able to load.

BUT I still faced `Unable to find component: [app.filament.a.b.c.pages.example]` only when `/livewire/update` was called to see if there are notifications.

mstfkhazaal:
> I registered it as `->discoverPages(in: app_path('Filament/A/B/C/Pages'), for: 'App\\Filament\\A\\B\\C\\Pages')`, the page was able to load.
> 
> BUT I still faced `Unable to find component: [app.filament.a.b.c.pages.example]` only when `/livewire/update` was called to see if there are notifications.

Try `php artisan optimize`
Or `php artisan filament:optimize`
Or `composer dump-autoload`

bilogic:
Already did that, however, the problem behaves a bit differently with and without `optimize`. 

Before I go further, my page has a `canAccess` method to check on the user's role and team using spatie/permission. (I'm also using the Teams feature)

So, if I optimize it, my page is able to load, however, it quickly shows a notification `You do not have permission to access` followed by redirecting to `/`

I added a log in my `canAccess` to dump out values. During calls to my page, `getPermissionsTeamId()` returns a numeric value, however during calls to `/livewire/update`, `getPermissionsTeamId()` returns `false`.

Maybe I should check with spatie.

PR #7504, opened 2023-08-06
Feature/fix docs

thethunderturner:

When trying to use the activeTab in the blade file, you will get the following error:
**Undefined variable $activeTab**
image

Adding the attribute in the class of the Page fixes the error.

Also I added some documentation about the tabs being interactive.


danharrin:
I'm sorry, not gonna add this. `$activeTab` is just an example, not part of the component, and I would really rather not explain how to use Livewire in these docs.

PR #7505, opened 2023-08-06
Fix disabled problem for Radio, Checkbox, Checkbox List & Toggle

margarizaldi:
This commit wants to fix **Radio**, **Checkbox**, **Checkbox List** & **Toggle** that can't be disabled.


danharrin:
Isn't this a Livewire bug in v3? Should we really have to implement hacks in the Filament codebase like this? 🤔

danharrin:
Would be great if you just made a failing test PR to Livewire instead

zepfietje:
Closing this as there's more places that need to be handled.
Agree with Dan that this should be fixed in Livewire.

PR #7516, opened 2023-08-07
fix some Farsi translations

mszabeh:


mokhosh (commented):
Good job, but some of the changes don't seem quite right. Like the description key in table.php seems to have changed the meaning.

And some one-word breaks which should technically have "nim fasele" now have full space which again changes the meaning incorrectly.

mszabeh:
Hi @mokhosh , Thanks for feedback.
I try to fix

PR #7522, opened 2023-08-07
Set default pagination page option to null

howdu:

Set the default `defaultPaginationPageOption` to null otherwise, it was never reaching the last part of this:
```php
public function getDefaultPaginationPageOption(): int | string | null
{
    return $this->evaluate($this->defaultPaginationPageOption) ?? Arr::first($this->getPaginationPageOptions());
}
```

Use case example:
```php
$table->paginated([5, 10, 25, 50, 100]);
```


danharrin:
Thanks

PR #7525, opened 2023-08-07
Make tenant parameter optional

andrewdwallo:

- Modified the route prefix definition to tolerate the absence of a tenant parameter.
- Beneficial for initial user interactions or scenarios where tenant-specific data is not yet established but users need access to certain functionalities such as when creating a new tenant through the tenant registration page.

Note: This change doesn't make the tenant parameter truly optional. Instead, it prevents errors in specific contexts within the panel where a tenant hasn't been defined yet.

All pages within a panel that has tenancy has the tenantID attached to the url. This should not be the case and should be fixed. In the realm of multi-tenancy web applications, the delineation between tenant-specific and global contexts is crucial for intuitive and functional design. While certain elements are naturally scoped within the confines of a specific tenant – such as data records, analytics, or settings – others transcend this boundary. Take, for instance, a user's profile page. A user's personal details, preferences, and account settings are inherently user-centric, rather than tenant-specific. Attaching a tenantID to a profile page URL (which is what occurs when creating a custom profile page) would misleadingly suggest that a user has distinct profiles for each tenant, which is typically not the case. Moreover, it complicates the URL structure, potentially leading to confusion or errors when users navigate or share links. By avoiding tenantID attachment for such universal functionalities, developers can maintain a cleaner, more logical URL structure while ensuring that the application's architecture accurately represents the user's experience and expectations.

This currently means that you cannot have a custom profile page within a panel that has tenancy without errors. If that is the case then it needs to be documented or fixed. If I need to submit an issue for this I will, but I know this is the case for many users.

Before:
Screenshot 2023-08-07 at 8 03 05 AM

After:
Screenshot 2023-08-07 at 8 03 37 AM

### DIFF
--- a/packages/support/src/Commands/InstallCommand.php
+++ b/packages/support/src/Commands/InstallCommand.php
@@ -2,6 +2,7 @@
 
 namespace Filament\Support\Commands;
 
+use Filament\PanelProvider;
 use Filament\Support\Commands\Concerns\CanManipulateFiles;
 use Illuminate\Console\Command;
 use Illuminate\Filesystem\Filesystem;
@@ -20,7 +21,9 @@ class InstallCommand extends Command
     public function __invoke(): int
     {
         if ($this->option('panels')) {
-            $this->installAdminPanel();
+            if (! $this->installAdminPanel()) {
+                return static::FAILURE;
+            }
         }
 
         if ($this->option('scaffold')) {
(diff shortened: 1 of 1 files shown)


andrewdwallo:
This is closed now but i still think this should be a very high priority in my opinion. At least make it to where a custom profile page doesn't have the tenant ID attached to the url.. I tried my best but I honestly couldn't even exactly find how the tenant ID was being set. Even where I thought it would be, changing the code did absolutely nothing...

danharrin:
Why did you close it if it needs implementing? And why would you need custom profile page routing if you can override the built-in one which doesnt scope to the tenant?

andrewdwallo:
@danharrin Because why would I want the profile page to be outside the panel? It's not supposed to look like or be a login or register page.

andrewdwallo:
I unintentionally closed this pull request by force-pushing to the branch—my mistake. While it didn't address the actual issue, I didn't find closing it concerning. Currently, I've resolved this by creating a separate panel for user profiles, thus avoiding tenant scoping. It works, but I'm uncertain if it's the best solution, considering that the user would then have to switch between different panels - making the logic more complex. The decision on whether or not to allow exceptions for tenancy within certain pages lies with you and Filament. If you prefer not to have exceptions or do not want users to decide themselves (by making it configurable), that's fine, and we'll explore other alternatives as needed.

devhoussam1998:
reopen this it's important

PR #7526, opened 2023-08-07
fix: EditProfile - debounce password change update requests 

DanielSpravtsev:

Changed TextInput::make('password')->live() to debounce() for reducing update requests to the server


packages/panels/src/Pages/Auth/EditProfile.php
```diff
@@ -220,7 +220,7 @@ protected function getPasswordFormComponent(): Component
             ->autocomplete('new-password')
             ->dehydrated(fn ($state): bool => filled($state))
             ->dehydrateStateUsing(fn ($state): string => Hash::make($state))
-            ->live()
+            ->debounce()
```
danharrin:
```suggestion
            ->live(debounce: 500)
```

PR #7527, opened 2023-08-07
Feature: Register all colors variants by default

KennedyTedesco:
I use a ton of color variants in badges, table columns, etc. So, I thought about registering these by default. Maybe it's useful for others? I completely understand if you guys only want to register the most important ones (like primary, danger, etc.). At the very least, could we have a `getColors()` method in the `Color` class? This way, users can easily register those colors in their applications, like this:

```
FilamentColor::register([
    ...[
        'red' => Color::Red,
        'gray' => Color::Zinc,
        'info' => Color::Sky,
        'primary' => Color::Blue,
        'success' => Color::Green,
        'warning' => Color::Amber,
    ],
    ...Color::getColors(),
]);
```

Regarding the `getColors()` method in the `Color` class, I considered using `Reflection` to retrieve all the class constants. However, I'm wondering if it's worth it. Since these colors are unlikely to change frequently over time, it seems like unnecessary overhead in this specific case. But I'm open to making changes if necessary.

What you guys think?

Ps: This change preserves the user's ability to override colors using `FilamentColor::register([...])`.

### DIFF
--- a/packages/support/src/Colors/Color.php
+++ b/packages/support/src/Colors/Color.php
@@ -375,4 +375,35 @@ class Color
 
         return $colors;
     }
+
+    /**
+     * @return array<string, array{50: string, 100: string, 200: string, 300: string, 400: string, 500: string, 600: string, 700: string, 800: string, 900: string, 950: string}>
+     */
+    public static function getColors(): array
+    {
+        return [
+            'slate' => static::Slate,
+            'gray' => static::Gray,
+            'zinc' => static::Zinc,
+            'neutral' => static::Neutral,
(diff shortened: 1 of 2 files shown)


danharrin:
This will significantly increase the size of the HTML on every page, as each color then needs to be output with CSS variables. Feel free to do a separate PR of a `Color::all()` method 👍

KennedyTedesco:
@danharrin Oh, I didn't know about that. I've just started exploring Filament's core recently. I'll definitely do the other PR. Thanks for explaining it to me.

PR #7533, opened 2023-08-07
[bug] [2.x] fix upload progress indicator

coleshirley:
This PR fixes the filepond upload progress indicator not working. This is generally not noticable for small files since uploading is very fast but for large files there is no indication of progress.

image

The livewire `$wire.upload()` progress callback is not directly compatible with the filepond progress callback. This small callback function makes the two compatible


danharrin:
Thanks!

PR #7543, opened 2023-08-08
Fix: Sidebar collapsible button style

killjin:

Fix: Sidebar collapsible button style
image

### DIFF
--- a/packages/panels/resources/views/components/sidebar/group.blade.php
+++ b/packages/panels/resources/views/components/sidebar/group.blade.php
@@ -42,13 +42,14 @@
             </span>
 
             @if ($collapsible)
-                <x-filament::icon-button
+                <x-filament::link
                     color="gray"
+                    tag="button"
                     icon="heroicon-m-chevron-up"
                     icon-alias="panels::sidebar.group.collapse-button"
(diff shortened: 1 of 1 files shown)


danharrin:
Sorry, this isn't right. This is not the job of a link. If you have design improvements to the icon button, please be specific about them.

killjin:
I don't think there should be a yellow border, it can be a gray background. Similar to github.

image

danharrin:
I think I agree that the focus border on Chrome needs some work @zepfietje. I think using a grey background like in v2 is good.

killjin:
good, hover:bg-gray-xx  focus:bg-gray-xx

zepfietje:
I switched away from just a gray background to using a ring on purpose for accessibility reasons. The right fix here is making sure Chromium doesn't keep the focus styling after the button has been clicked. Haven't looked into how though, so let me know if you do.

zepfietje:
Okay seems as simple as replacing `focus:` with `focus-visible:`: https://tailwindcss.com/docs/hover-focus-and-other-states#focus-visible.

zepfietje:
Tried this and ran into a lot of edge cases. Maybe look into replacing focus ring with the native CSS outline.

PR #7566, opened 2023-08-08
Updating docs to fix reference to the max content width settings

glennjacobs:
This update looks to fix the outdated documentation for `maxContentWidth()`.

Filament v2 had this available on the Page, but in v3 it is moved to the Panel settings.

### DIFF
--- a/packages/panels/docs/04-pages.md
+++ b/packages/panels/docs/04-pages.md
@@ -292,14 +292,3 @@ public function getFooter(): ?View
 ```
 
 This example assumes you have a Blade view at `resources/views/filament/settings/custom-footer.blade.php`.
-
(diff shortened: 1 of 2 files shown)


danharrin:
`getMaxContentWidth` does actually still exist on the page though, and it works right?

glennjacobs:
It didn't work for me when I used it (going from memory).

glennjacobs:
Assuming it should still be available per-page also, I can have a look, but won't be for over a week or so now as I'm on holiday.

glennjacobs:
@danharrin I checked, and it is working on the latest version. So I've amended this PR accordingly.

danharrin:
Thanks

PR #7570, opened 2023-08-08
Feature: Add item order to Repeater itemLabel

andcarpi:
Useful for creating countable labels with functions on Repeater itemLabel() - Code and Docs updated

```php
Repeater::make('Items')
    ->schema([
        Forms\Components\TextInput::make('copies')
            ->required()
            ->label('Quantity')
    ])
    ->itemLabel(fn(array $state, int $order): ?string => 'Package ' . $order ?? null),
```

- Simple code change to Repeater.php that enables countable labels
- Modified the docs (but didn't generated the screenshot)
- No tests created, but no problems with my project

image

- [ No ] Changes have been thoroughly tested to not break existing functionality.

### DIFF
--- a/packages/panels/src/Panel/Concerns/HasComponents.php
+++ b/packages/panels/src/Panel/Concerns/HasComponents.php
@@ -313,13 +313,9 @@ trait HasComponents
                 $variableNamespace = (string) str($variableNamespace)->before('\\');
             }
 
-            if ($variableNamespace === null) {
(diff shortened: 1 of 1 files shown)


zepfietje:
Similar API already exists for `Builder`, so I think we should make the two match: https://filamentphp.com/docs/3.x/forms/fields/builder#numbering-builder-items.

PR #7575, opened 2023-08-09
Support showing current wizard step heading instead of all steps

josefbehr:

before:
image

after:
image

### DIFF
--- a/packages/forms/docs/04-layout/05-wizard.md
+++ b/packages/forms/docs/04-layout/05-wizard.md
@@ -202,3 +202,15 @@ Wizard::make([
         fn (Action $action) => $action->label('Next step'),
     )
 ```
+
+## Hiding the header
+
+The whole header of the wizard can be hidden by using `hiddenHeader()` analog to `hiddenLabel()`
+
+```php
+use Filament\Forms\Components\Wizard;
+
+Wizard::make([
+    // ...
+])->hiddenHeader()
+```
--- a/packages/forms/docs/04-layout/05-wizard.md
+++ b/packages/forms/docs/04-layout/05-wizard.md
@@ -205,12 +205,12 @@ Wizard::make([
 
 ## Hiding the header
 
-The whole header of the wizard can be hidden by using `hiddenHeader()` analog to `hiddenLabel()`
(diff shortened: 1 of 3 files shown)


josefbehr:
Added optional parameter `stepLabelAsHeading` which allows to show the current step's label as a heading for the form components.

This is what it looks like:
image

without icon:
image

zepfietje:
@josefbehr, could you please share screenshots of what this looks like for uncontained wizards?

Also, how does this work when used with #7576.

josefbehr:
@zepfietje currently, it would look like this:
image

zepfietje:
Thanks, @josefbehr.
Unfortunately that looks really bad then.

I'm going to close this PR as it essentially contains two features (hiding the steps and using the current step label), doesn't properly support all existing APIs, requires a lot of reworking, and complicates the wizard component to a level that feels outside the scope of the wizard.

josefbehr:
@zepfietje Thanks for the feedback.

Regarding the bad look, that could be amended, just wanted to show you how it would look like with the current implementation.
As for the other points, I'd be glad to adapt the PR to satisfy all requirements. In my opinion, at least a wizard without the header would be a great addition, and four likes on the PR suggests to me a wider interest in this.
If you are finite on this decision, I respect that, as I understand that you will have to deal with this on the long run, but if you're open to discussing this further, I'm happy to find a solution.

zepfietje:
I do understand the use case of hiding the steps. However, I'm still not too sure of the complete solution, as (also when showing just the current step label) it's very similar to a section with buttons.

josefbehr:
@zepfietje Great. I see your point. I guess it is visually similar to a section with a button. But I'd argue that in function it's not. I don't think this multi-step form process is easily replicated using sections, so I think there's distinct use-cases here.

I think it's best to handle the possibility of showing the step label as a separate issue and enhancement as not to muddy the waters here. So let's exclude that from the discussion for now.

zepfietje:
If @danharrin agrees, you can create a new PR which just adds support for hiding the steps.

danharrin:
Fine, hiding the steps can be an option

PR #7576, opened 2023-08-09
Add compact wizard styling

josefbehr:

before: 
Screenshot_20230809_030647

after:
image

### DIFF
--- a/packages/forms/docs/04-layout/05-wizard.md
+++ b/packages/forms/docs/04-layout/05-wizard.md
@@ -202,3 +202,15 @@ Wizard::make([
         fn (Action $action) => $action->label('Next step'),
     )
 ```
+
+## Compact header
+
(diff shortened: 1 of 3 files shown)


zepfietje:
@josefbehr are you using both the regular and the compact wizard in the same project?

I'm not too keen on merging this in as it complicates the design and it has to be supported for every single feature we add, redesign we do, or themes we create.

Also, in my opinion, descriptions should still work fine in all modes.

josefbehr:
@zepfietje yeah, I get that, no worries. If you don't want to support this, I can just maintain it for myself. On the other hand, it might be useful for others as well, as the default wizard's header is quite massive in my opinion and takes up a lot of space in confined areas. That's where I'm using it actually. In addition to the normal layout, I'm using the compacted one in a place where the wizard is used for a multi-step form in a multi-column layout.

zepfietje:
Thanks for clarifying, @josefbehr!
I'd probably advise to make design changes to not have the wizard in a small parent component. Not just because of the wizard size, but also because of your overall application UX.

Going to hold off on merging this for now.

PR #7583, opened 2023-08-09
Support hidden label in fieldset

josefbehr:

before: 
image

after:
image


josefbehr:
I know that the label can be hidden by just setting an empty string as the label, but I think it should respect the `hiddenLabel()` method as well, which also enables dynamically showing or hiding the label.

zepfietje:
Thanks for the PR, @josefbehr.
I've updated the implementation.
However, I think we should potentially still output the label, but add `sr-only`.

danharrin:
We should add a `label-hidden` attribute to the fieldset Blade component that handles `sr-only`

zepfietje:
> We should add a `label-hidden` attribute to the fieldset Blade component that handles `sr-only`

Agree. Mark as ready for review when you've handled that, @josefbehr. 🙂

josefbehr:
done

PR #7586, opened 2023-08-09
Fix fileUpload button tooltip, Fix translate

killjin:

@js directive not working in @foreach? Change to Illuminate\Support\Js::from that's ok.

### DIFF
--- a/packages/forms/resources/lang/zh_CN/components.php
+++ b/packages/forms/resources/lang/zh_CN/components.php
@@ -6,24 +6,36 @@ return [
 
         'actions' => [
 
+            'clone' => [
+                'label' => '克隆',
+            ],
+
             'add' => [
-                'label' => '添加 :label',
+                'label' => '添加到 :label',
+            ],
+
+            'add_between' => [
+                'label' => '在块之间插入',
             ],
 
             'delete' => [
                 'label' => '删除',
             ],
 
+            'reorder' => [
+                'label' => '移动',
+            ],
+
             'move_down' => [
-                'label' => '下移',
+                'label' => '向下移动',
             ],
 
             'move_up' => [
-                'label' => '上移',
+                'label' => '向上移动',
             ],
 
             'collapse' => [
-                'label' => '收起',
+                'label' => '折叠',
             ],
 
             'expand' => [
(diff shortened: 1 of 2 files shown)


zepfietje:
What are the `toHtml()` calls needed for?

Also, are you sure `{{ }}` works here in Blade component attributes? It doesn't always work IIRC.

killjin:
@js is equivalent to from()->toHtml()
I test @js doesn't work here, {{}} works.

danharrin:
I've made some changes, does it still work?

killjin:
image
image

It won't work after modification

danharrin:
Can you look into why that is? We should avoid `{{}}` inside Blade components if possible, hence why I concatenated it

killjin:
According to my observation, @js doesn't work in @foreach loop, it may be a bug of blade.

In this file, {{ }} is used very frequently, and now the content of $action['label'] must be output, but @js does not work, and the php function Js::from can only be used instead.

killjin:
```
<?php
//vendor/laravel/framework/src/Illuminate/View/Compilers/Concerns/CompilesJs.php

namespace Illuminate\View\Compilers\Concerns;

use Illuminate\Support\Js;

trait CompilesJs
{
    /**
     * Compile the "@js" directive into valid PHP.
     *
     * @param  string  $expression
     * @return string
     */
    protected function compileJs(string $expression)
    {
        return sprintf(
            "<?php echo \%s::from(%s)->toHtml() ?>",
            Js::class, $this->stripParentheses($expression)
        );
    }
}
```

@js is a short name for {{ Illuminate\Support\Js::from()->toHtml() }} , why can't it be used?

killjin:
Sorry for not seeing the : before the code due to my carelessness, this works.

```
:x-tooltip="'{
        content: ' . \Illuminate\Support\Js::from($action['label']) . ',
        theme: $store.theme,
    }'"
```

killjin:
Correct my mistake, @js works in @foreach, but it doesn't work in blade component attribute..

It's work:
```
@php
$vv = [
    'a' => ['label' => 'labela'], 'b' => ['label' => 'labelb'], 'c' => ['label' => 'labelc']
];
@endphp

@foreach ($vv as $v)
<a
    x-tooltip="{
        content: @js($v['label']),
        theme: $store.theme,
    }">vv</a>
@endforeach
```

Put it inside <x-filament::button and it won't work:
```
<x-filament::button 
    x-tooltip="{
        content: @js($v['label']),
        theme: $store.theme,
    }">vv</x-filament::button>
```

danharrin:
So the current implementation in this PR works, right?

killjin:
Yes, This PR is fine and works fine.

PR #7602, opened 2023-08-09
Apply column search when overriding applySearchToTableQuery

tanthammar:
In the documentation for Laravel Scout one needs to apply individual column searches to the Table query like in the parent version of the method.

### DIFF
--- a/packages/tables/docs/10-advanced.md
+++ b/packages/tables/docs/10-advanced.md
@@ -193,6 +193,8 @@ use Illuminate\Database\Eloquent\Builder;
 
 protected function applySearchToTableQuery(Builder $query): Builder
 {
+    $this->applyColumnSearchesToTableQuery($query);
+    
     if (filled($search = $this->getTableSearch())) {
         $query->whereIn('id', Post::search($search)->keys());
     }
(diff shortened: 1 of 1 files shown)


nessimabadi:
Hi @tanthammar and @danharrin, I was having trouble with this and found an edge case in this implementation.
If using global table search instead of individual column search the code added to the docs will not work.

I had to add the following code for it to work.
` $this->applyGlobalSearchToTableQuery($query);`

Could you confirm that I'm not wrong and I'll create a PR, or should I create a PR.

Thanks

danharrin:
The whole point of the `whereIn()` bit with Scout that it replaces the global search, so it sounds like your Scout index isnt working? The default implementation of this method is

```php
protected function applySearchToTableQuery(Builder $query): Builder
{
    $this->applyColumnSearchesToTableQuery($query);
    $this->applyGlobalSearchToTableQuery($query);

    return $query;
}
```

tanthammar:
@nessimabadi 

Don't know if this helps, but here is an example from my codebase;

```php

use App\Models\Booking;

protected function applySearchToTableQuery(Builder $query): Builder
{
    if (filled($searchQuery = $this->getTableSearch())) {
        $query->whereIn('id', Booking::search($searchQuery)->select('id'));
    }

    return $query;
}

```

PR #7610, opened 2023-08-09
Image optimisation and resizing on SpatieMediaLibraryFileUpload

joshembling:
I'm submitting this PR to add image optimization and resizing functionality to the v2 SpatieMediaLibraryFileUpload plugin.

### The issue at hand:
Currently, when you upload an image and use conversions with `media-library`, the original file is saved with its corresponding versions provided on your model. 

This is fine if you have a lot of disk space, but what if you'd rather reduce the file sizes before uploading - especially in the case where you know you'll never need to save 10 x 10MB images the user has uploaded.

#### optimize()

A new `optimize()` method has been added.

- Converts the binary image to an `InterventionImage` object
- Optimises it by reducing its file size when passing in a `(string)` image extension to convert to

E.g. I want to convert my image to 'webp': 

```php
use Filament\Forms\Components\SpatieMediaLibraryFileUpload;

SpatieMediaLibraryFileUpload::make('attachment')
    ->image()
    ->optimize('webp'),
```

Uploaded image is now in webp format.

#### resize()

A new `resize()` method has also been added. 
- Gets the `InterventionImage` object and resizes it by taking in an `(int)` percentage to reduce the image dimensions
- Maintains aspect ratio

E.g. I'd like to reduce my image (1280px x 720px) by 50%:
```php
use Filament\Forms\Components\SpatieMediaLibraryFileUpload;

SpatieMediaLibraryFileUpload::make('attachment')
    ->image()
    ->resize(50),
```
Uploaded image size is 640px x 360px.

As mentioned, these chained methods will replace the original file(s) uploaded. I.e. the versions saved to the disk are the converted images only.

### DIFF
--- a/packages/spatie-laravel-media-library-plugin/docs/02-form-components.md
+++ b/packages/spatie-laravel-media-library-plugin/docs/02-form-components.md
@@ -97,3 +97,27 @@ SpatieMediaLibraryFileUpload::make('attachments')
         'thumb' => ['orientation' => '90'],
     ]),
 ```
+
+## Optimizing images
+
+Before uploading your image, you may choose to optimize it by converting to your chosen format. The file saved to your disk will be the converted version only.
+
+```php
+use Filament\Forms\Components\SpatieMediaLibraryFileUpload;
+
+SpatieMediaLibraryFileUpload::make('attachment')
+    ->image()
+    ->optimize('webp'),
+```
+
+### Resizing images
+
+You may also want to resize an image by passing in a percentage you would like to reduce the image by. This will also maintain aspect ratio.
+
+```php
+use Filament\Forms\Components\SpatieMediaLibraryFileUpload;
+
+SpatieMediaLibraryFileUpload::make('attachment')
+    ->image()
+    ->resize(50),
+```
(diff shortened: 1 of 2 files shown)


danharrin:
@awcodes how useful do you think this is? Also, do you think it is worth porting it to the base file upload as well as the Spatie one, since these aren't really Spatie features, right?

awcodes:
I can see some value to this, but I think it would be more relevant to the FileUpload than to Spatie. Although I understand the use case described, personally, I feel that any image manipulation base on Spatie's package would be best handled in the way that Spatie's package intends for it to be done.

I'm also a firm believer that these kinds of image manipulations should be modified before upload or with validation, like if you don't want 10M images then set a validation rule so the user has to modify it before it ever reaches the server. Although that is not always the case when dealing with users, so I understand the point of view.

I do have some concerns with the API, though. Should you choose to move forward with this PR I'm happy to add some notes to the code.

wychoong:
would this be better if the transformation is handled by Filepond before upload to server https://pqina.nl/filepond/docs/api/plugins/image-transform/

danharrin:
Yeah I think I agree. @wychoong I think we already have those features baked into the file upload. Also, with the new image editor, everything can be changed before upload as well. Thanks though @joshembling.

PR #7611, opened 2023-08-09
Add tooltip to hint icon

howdu:

Useful when combined with a question mark icon.
Screenshot 2023-08-09 at 18 19 02

```php
->hintIcon('heroicon-o-question-mark-circle', hintIconTooltip: 'Helper text only visible when user requests it'),
```

### DIFF
--- a/packages/forms/src/Components/Concerns/HasHint.php
+++ b/packages/forms/src/Components/Concerns/HasHint.php
@@ -29,6 +29,8 @@ trait HasHint
 
     protected string | Closure | null $hintIcon = null;
 
+    protected string | Closure | null $hintIconTooltip = null;
+
     public function hint(string | Htmlable | Closure | null $hint): static
     {
         $this->hint = $hint;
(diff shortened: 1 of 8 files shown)


packages/forms/src/Components/Concerns/HasHint.php
```diff
@@ -46,9 +48,10 @@ public function hintColor(string | array | Closure | null $color): static
...
     }
 
-    public function hintIcon(string | Closure | null $hintIcon): static
+    public function hintIcon(string | Closure | null $hintIcon, string | Closure | null $hintIconTooltip = null): static
     {
         $this->hintIcon = $hintIcon;
+        $this->hintIconTooltip = $hintIconTooltip;
 
         return $this;
     }
```
danharrin:
```suggestion
    public function hintIcon(string | Closure | null $icon, string | Closure | null $tooltip = null): static
    {
        $this->hintIcon = $icon;
        $this->hintIconTooltip = $tooltip;

        return $this;
    }
```

danharrin:
Can we get a short section in the docs about this please @howdu? Probably with a separate subheading under the hint icon one

danharrin:
Thanks, this is nice

zepfietje:
This implementation doesn't use the current theme for the tooltip.
I've fixed all tooltip instances when working on the v3 redesign.
Opened an issue for this: https://github.com/filamentphp/filament/issues/7640.

Could you have a look at how the theme is passed to the tooltip in all other places tooltips are used and apply that here in a new PR, @howdu?

PR #7612, opened 2023-08-09
Update MakeThemeCommand.php

ijpatricio:
Fix "Using NPM" instead of "Using Node.js"

This PR is very simple. As image below, the output told me it was using Node v9, instead of NPM V9.

This just changes that output, to "Using NPM v..."

Let me know if allowing usage of `yarn` and `pnpm` is seen with good eyes.
People that do that will know their way, but it's a convenience that's good to have.
It can be autodetected (based on lockfile search), deafulting to NPM.

What do you think? I'll gladly to that PR.

image


danharrin:
Sure, I would accept `--flags` for the other providers if you want to implement that :)

PR #7636, opened 2023-08-10
Fix double borders when using grid table

### LINKED ISSUE #7632: Custom table row borders
### Package

filament/tables

### Package Version

v3.0.13

### Laravel Version

v10

### Livewire Version

_No response_

### PHP Version

8.2

### Problem description

If using `Split` and `Stack` within a table, the dividing border get rendered twice, currently also visible within the Demo's authors list:
https://demo.filamentphp.com/blog/authors

The primary cause seems to be [L503](https://github.com/filamentphp/filament/blob/3.x/packages/tables/resources/views/index.blade.php#L503)). 

There's also a double border between the header and the first row if you're not using at least one sortable column. Origin seems to be the divide-Class in [L381](https://github.com/filamentphp/filament/blob/3.x/packages/tables/resources/views/index.blade.php#L381)

As I do not know all the possbile combinations, I am quite stucked how to solve via a PR without breaking any other layout approach. Just wanted to bring this to your attention. If you have an idea how to solve this, I am happy to provide a PR. Thank you!

### Expected behavior

No double borders :)

### Steps to reproduce

Add `Split` and `Stack` columns to any table layout.
Test

### Reproduction repository

https://github.com/ajnsn/filament-demo-borders

### Relevant log output

_No response_

ajnsn:
Fixes #7632 

At least on Chrome, Safari and Firefox (latest versions) on Mac this looks good to me :) Edge on Win should work the same as Chrome?

Below is a before/after comparison
- Grid table with bulk-actions/sorting (before / after)
- Grid table without bulk actions or sorting (before / after)
- Default table (not affected)


ajnsn:
fyi: The second commit-diff looks terrible but just moves the `@if ($isSelectionEnabled || count($sortableColumns))` around the whole div, just instead of applying the `hidden` attribute to the class.

danharrin:
@zepfietje this seems like the logical solution that we implemented previously, can you remember what was wrong with it?

zepfietje:
Like Dan said in the issue you created, this implementation is on purpose.
Unfortunately not all browsers appear to render `0.5px` borders nicely, which leads to the visual inconsistency you reported, @ajnsn.

However, the reason behind needing a half pixel border, is that in dark mode the border is transparent white. The Tailwind CSS `divide-` classes only apply a top border to elements, which results in an indistinguishable border in dark mode when hovering over or selecting an adjacent row. This isn't an issue with regular tables, because their internal borders actually follow a similar approach already.

So, unfortunately this PR doesn't fix the issue and only brings back the broken behavior we weren't able to fix in any other way. Feel free to come up with a solution if you can think of any, given the problem description I outlined above. 🙂

danharrin:
@zepfietje can we change the border color on hover to make it more distinguishable without it looking weird?

zepfietje:
I think I tried those approaches, and I'm not sure that would yield a good result.
However, that may be something @ajnsn might want to try.

danharrin:
Yeah, let's see if we can get it looking "normal" visually by with a slight change in border color on hover

zepfietje:
Note, you would have to change the border color when the previous sibling is selected/hovered. Tricky, but maybe possible.

ajnsn:
Ah, interesting. I am not using dark mode.

One easy solution could be to just omit the bottom border when not using dark mode?
`dark:[&>*:not(:last-child)]:border-b-[0.5px]`:

Line 503
```diff
- '[&>*:not(:first-child)]:border-t-[0.5px] [&>*:not(:last-child)]:border-b-[0.5px] [&>*]:border-gray-200 dark:[&>*]:border-white/5' => ! $contentGrid,
+ '[&>*:not(:first-child)]:border-t-[0.5px] dark:[&>*:not(:last-child)]:border-b-[0.5px] [&>*]:border-gray-200 dark:[&>*]:border-white/5' => ! $contentGrid,
```
At least for me now both look pretty :) 
Let me know if that is a proper solution for this issue.

zepfietje:
That wouldn't work, as some browser do correctly render half pixels. But yeah, Dan's suggestion would entail having only a top border. I'd say, give it a try and create a new PR with screenshots of light and dark mode when static, hovering, hovering sibling, selected, sibling selected, selected and hovering, sibling selected and hovering.

ajnsn:
Can you please provide the browser and version you are referring to? As for me this works on Chromium and Wekbit, which should be the majority of browsers. And I feel that you already have concrete ideas how the solution should look like and you are also deeper in this Filament stuff for sure. So I tried two times, now let's see if some other guys have ideas here.

In general I am now not a huge fan of the current approach as its difficult to maintain and it prefers dark mode over light mode. Light mode in v3 is having so little contrasts that users  won't even recognize any thing except they see this huge border issue. This immediately was reported by about 50 users whereas no one tells us about hovering effects.

danharrin:
50 users reported a 2px border? 😅 I feel sorry for you dealing with those people day-to-day

ajnsn:
They rated the overall "relaunch" (we migrated to v3) and yes, many said that the tables look broken. They were not 50 but many :) 
One other thing is the low contrast in comparison to v2. We changed the body color to bg-gray-100 and the immediate feedback was a lot better :)

danharrin:
> In general I am now not a huge fan of the current approach as its difficult to maintain and it prefers dark mode over light mode

I don't really understand how we prefer one over the other, they are both supported equally, and I can guarantee we would have a different bug report filed if the borders disappear on hover in dark mode

zepfietje:
> Can you please provide the browser and version you are referring to? As for me this works on Chromium and Wekbit, which should be the majority of browsers.

I'm not seeing the issue on the latest versions of Safari, Chromium and Firefox on the latest version of macOS. So perhaps you could provide some more details on your environment to make sure the issue can actually be reproduced?

> And I feel that you already have concrete ideas how the solution should look like and you are also deeper in this Filament stuff for sure. So I tried two times, now let's see if some other guys have ideas here.

Did you try the approach Dan and I suggested above?

> In general I am now not a huge fan of the current approach as its difficult to maintain and it prefers dark mode over light mode.

Light and dark mode are treated equally, which is the reason the current fix is in place, as without it dark mode would have a degraded UX. Also, the current fix is a lot easier to maintain than the alternative approaches we outlined above.

> Light mode in v3 is having so little contrasts that users won't even recognize any thing except they see this huge border issue. This immediately was reported by about 50 users whereas no one tells us about hovering effects.

If your users don't happen to use dark mode, that doesn't mean we should degrade the UX of all users using dark mode with Filament.

ajnsn:
For sure its a matter of audience. Afaik our users don't even know that dark mode exists. They are business users during daytime. What I meant is that users will immediately recognize the 2px bug rather than they see an missing border - but that might stay my assumption.

And I do strongly believe that this is not browser than a screen issue. No non-retina screen can render 0,5px. So this solution is in general questionable.

Thank you both for your time investigating this issue.

danharrin:
I have an idea. `flex flex-col gap-px bg-BORDERCOLOR` @zepfietje?

danharrin:
essentially a fake border using flex gap

danharrin:
Reopening this so someone at least has a go at that solution

packages/tables/resources/views/index.blade.php
```diff
@@ -500,7 +497,7 @@ class="flex gap-x-3 py-3"
                         @class([
                             'gap-4 p-4 sm:px-6' => $contentGrid,
                             'pt-0' => $contentGrid && $this->getTableGrouping(),
-                            '[&>*:not(:first-child)]:border-t-[0.5px] [&>*:not(:last-child)]:border-b-[0.5px] [&>*]:border-gray-200 dark:[&>*]:border-white/5' => ! $contentGrid,
+                            'divide-y divide-gray-200 dark:divide-white/5' => ! $contentGrid,
```
```suggestion
                            'gap-px bg-gray-200 dark:bg-white/5' => ! $contentGrid,
```

zepfietje:
> I have an idea. `flex flex-col gap-px bg-BORDERCOLOR` @zepfietje?

That won't solve the issue where hovering a row in dark mode will make the border disappear.
A different (hacky) solution that came to mind is faking the border using `before`. However, that would require adding `0.5px` padding, which will again render weirdly?

danharrin:
Why will hovering a row in dark mode make it disappear? Because it is the same color or something? can we make it slightly lighter across all tables?

zepfietje:
> Why will hovering a row in dark mode make it disappear?

Because the hover background color is the same as the border, and the border lives outside the background. In regular tables the border overlaps with the background, creating the effect we need since we're using `white/`

> Because it is the same color or something? can we make it slightly lighter across all tables?

I tried using a less transparent white, but it didn't work nicely.

Perhaps my suggested solution with `before:` isn't that bad.

ajnsn:
I tried the gap-px approach with the bg-classes. This also needs bg-classes on the rows, e.g. `bg-white dark:bg-gray-900` as otherwise the rows will have have the bg from the grid bg-classes and the gap will not produce the intended border. If you then change the row's dark:hover to something other than a transparent white (e.g. dark:hover:bg-black/40), you'll have the desired effect. Example:

CleanShot 2023-08-11 at 15 59 51

This is now just the inverse look yes and needs fine tuning but iMho this could be also optimized by removing the bg-white-transparencies at other places. Maybe instead we use fixed color-shades.

Note that I also found at least one other occurence of the 0,5px approach. And there's also one in the pagination.

zepfietje:
Dark mode really needs the transparent white shades.

ajnsn:
Currently the transparency are stacking. You could make the same without it, e.g. use this for the gap:
`gap-px bg-gray-200 dark:bg-[#2F2F32]`

and this for the row:
`hover:bg-gray-50 dark:hover:bg-[#242427]`

This will make the gap without hover look a little bit brighter but will look the same on hover.

zepfietje:
Unfortunately that's not going to work well, as you should be able to change the background color without having to manually calculate all the layering colors again.

ajnsn:
Aha, but stacking transparencies will not solve this problem. You could also define e.g. a bg-gray-25 or something, just do not use opacity on the colors. Otherwise the white shades will not be visible at all. 

As I said, I do not know the whole context - just expierenced this transparency stacking a few times and this was always the easiest solution. 

Other approach is to use a darker hover color. Will also work (see a few posts above).

ajnsn:
fyi: I updated the PR to follow Dan's idea. The hover effect and both "borders" are active in dark mode. The result is barely visible with the current colors but it works and is just a matter of hover/gap color. 
In light mode the double border is gone.

danharrin:
How about adding a background color to the container behind the one with the gap, so that the opacity colors still work correctly and nothing needs recalculating

danharrin:
> fyi: The second commit-diff looks terrible but just moves the @if ($isSelectionEnabled || count($sortableColumns)) around the whole div, just instead of applying the hidden attribute to the class.

Don't understand why you did this so I'm reverting it, pretty sure it is hidden with CSS for a reason

danharrin:
Well it's not perfect on dark mode when you have table rows selected, but it's much better than the double border. Thanks @ajnsn!

ajnsn:
Hey @danharrin 

nope - the only kudos goes to your whole team and this great "framework". And excuse me if if I did not recognize the whole context (thanks now to @zepfietje for explaining all the relations and dependencies) and haven't expressed myself well enough! I'll try better next time.

> Don't understand why you did this so I'm reverting it, pretty sure it is hidden with CSS for a reason

There's is still one double border when you do not have any sorting or bulk actions. This is why I put the if-clause around that part that was formerly just `hidden` by the class-directive.  

Let's see if we can bring the original hover behaviour back at some point!

zepfietje:
Don't get me wrong, but how is this implementation any better than removing our `0.5px` fix altogether?

Haven't thought of any edge cases, but I think we should add the border using `before` instead (https://github.com/filamentphp/filament/pull/7636). Then high resolution screens will render everything just fine, while screens which aren't able to render the before at `-0.5px` top will have the current (broken) dark mode behavior.

zepfietje:
> There's is still one double border when you do not have any sorting or bulk actions. This is why I put the if-clause around that part that was formerly just `hidden` by the class-directive.

Hmm, so now all screens will have this bug?

ajnsn:
Hey @zepfietje 

yes and no. This is something that was already there not not introduced by this PR. 

It's coming from the "divide-y divide-gray-200" on the "fi-ta-ctn" (container) level. I could not find any reason why the whole bulk- and filter-div is hidden via the class, so the possible solution would be to have a if-clause around it. 

Perhaps we can have a following PR that restores the original hover behavior in dark mode while taking screen resolution into account. For now, I think this implementation is better than double borders.

PR #7655, opened 2023-08-11
[Feature] Add optional route tenant prefix

x7ryan:
Currently a multitenant configuration adds the tenant id or slug directly to the URL after the panel path. While there is nothing wrong with this behavior some, like myself, would prefer to apply a prefix to the tenant id or slug.

Currently a panel with a path of "admin" on the domain "example.com" that has a default multitenant config has a URL structure that looks like this: "https://example.com/admin/{tenant_id}"

This PR adds a new `->tenantRoutePrefix()` method you can chain on in the panel definition that will accept a string of what to prefix the tenant id or slug with. For example if you use `->tenantRoutePrefix('team')` that URL structure now becomes: "https://example.com/admin/team/{tenant_id}".

I had considered adding the prefix such as "team" in this example to the panel path. For example using `->path('admin/team')` instead of `->path('admin')` but I found this to not be desirable because that became part of the path for non-tenant related routes in the panel. For example "https://example.com/admin/login" would become "https://example.com/admin/team/login". Perhaps worse still "https://example.com/admin/profile" would become "https://example.com/admin/team/profile" which was real confusing because the profile was for the user not the tenant so having team on that route made no sense.

This was a fairly minor change, that is completely optional opt-in only feature and does not affect backwards compatibility and is not a breaking change.

### DIFF
--- a/packages/panels/docs/10-tenancy.md
+++ b/packages/panels/docs/10-tenancy.md
@@ -446,25 +446,6 @@ public function panel(Panel $panel): Panel
 }
 ```
 
-## Configuring the tenant route prefix
-
-By default the URL structure will put the tenant ID or slug immediately after the panel path. If you wish to prefix it with another URL segment, use the `tenantRoutePrefix()` method:
-
-```php
-use Filament\Panel;
-
-public function panel(Panel $panel): Panel
-{
-    return $panel
-        // ...
-        ->path('admin'}
-        ->tenant(Team::class)
-        ->tenantRoutePrefix('team');
-}
-```
-
-Before, the URL structure was `/admin/1` for tenant 1. Now, it is `/admin/team/1`.
-
 ## Configuring the name attribute
 
 By default, Filament will use Filament\Resources\Pages\CreateRecord;use the `name` attribute of the tenant to display its name in the app. To change this, you can implement the `HasName` contract:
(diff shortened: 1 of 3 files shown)


danharrin:
Duplicate of #7525?

packages/panels/src/Panel/Concerns/HasTenancy.php
```diff
@@ -57,6 +59,13 @@ public function tenant(?string $model, ?string $slugAttribute = null, ?string $o
         return $this;
     }
 
+    public function tenantRoutePrefix(string $prefix): static
```
danharrin:
```suggestion
    public function tenantRoutePrefix(?string $prefix): static
```

packages/panels/docs/10-tenancy.md
```diff
@@ -446,6 +446,25 @@ public function panel(Panel $panel): Panel
 }
 ```
 
+## Configuring the tenant route prefix
+
+By default the URL structure will put the tenant id or slug immediately after the panel path. If you wish to prefix the tenant id or slug with a value to tenant routes you may do so by using the `tenantRoutePrefix()` method.
```
danharrin:
```suggestion
By default the URL structure will put the tenant ID or slug immediately after the panel path. If you wish to prefix it with another URL segment, use the `tenantRoutePrefix()` method:
```

packages/panels/docs/10-tenancy.md
```diff
@@ -446,6 +446,25 @@ public function panel(Panel $panel): Panel
...
+{
+    return $panel
+        // ...
+        ->path('admin'}
+        ->tenant(Team::class)
+        ->tenantRoutePrefix('team');
+}
+```
+
+Without this the URL structure will be `https://example.com/admin/1` for tenant id 1, but with the route prefix it will become `https://example.com/admin/team/1`.
```
danharrin:
```suggestion
Before, the URL structure was `/admin/1` for tenant 1. Now, it is `/admin/team/1`.
```

danharrin:
Thanks

PR #7656, opened 2023-08-11
[Docs] Update docs on navagation to named routes

x7ryan:
Most of the examples on the docs page for navigation showed using the Laravel `route()` helper directly. However I noticed when doing this I always get a "Route Not Defined" Exception. Calling the route helper in a closure solves this.

This PR updates the examples to ensure they are all shown in closure form. I optionally added a note warning the helper should only be used in a closure, however in case you do not wish to include that it is a separate commit that can be easily removed.

### DIFF
--- a/packages/panels/docs/06-navigation.md
+++ b/packages/panels/docs/06-navigation.md
@@ -202,6 +202,8 @@ public function panel(Panel $panel): Panel
 }
 ```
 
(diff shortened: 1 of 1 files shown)


danharrin:
The cause of the error will be fixed in the next version. Also, I'm refactoring all examples to use Page::getUrl()

PR #7657, opened 2023-08-11
[Docs] Correct example showing the record passed in as param

x7ryan:
When following this example I discovered I could not call `$this->record` from that context. After a quick check I discovered the `$record` must be passed in as a parameter into the closure.

**NOTE:** There are other examples on this docs page of `$this->record` that I suspect may also need to be updated but I have not tested that so I am not updating them, unless someone can confirm that is also the case there. I am making this a draft until I can get confirmation on this.

### DIFF
--- a/packages/actions/docs/04-modals.md
+++ b/packages/actions/docs/04-modals.md
@@ -38,9 +38,9 @@ Action::make('updateAuthor')
             ->options(User::query()->pluck('name', 'id'))
             ->required(),
     ])
-    ->action(function (array $data): void {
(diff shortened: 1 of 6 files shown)


danharrin:
Hmm I think you're right and we should move all examples to `$record`. But `$this->record` does still work in some places.

x7ryan:
Yeah I wasn't sure. But I can go ahead and just switch all the examples to `$record`. Personally that's probably how I'd actually do it instead of having to figure out/remember where `$this` does and does not work.

PR #7658, opened 2023-08-11
[Feature] Allow disabling tenant menu

x7ryan:
This PR allows the tenant menu to be entirely hidden for a panel otherwise using multitenancy. I have found this is something I'm wanting to do where I am using tenancy for automatic scoping of resources by setting the User as the tenant. In this case there is no need to have a tenant menu because the setup makes it impossible for a user to belong to multiple tenants. 

Another possible (perhaps more likely) use case would be using multitenancy to share resources between multiple users, but where each user by design will only ever belong to a single tenant.

### DIFF
--- a/packages/panels/docs/10-tenancy.md
+++ b/packages/panels/docs/10-tenancy.md
@@ -356,6 +356,21 @@ MenuItem::make()
     ->hidden(fn (): bool => ! auth()->user()->can('manage-team'))
 ```
 
+### Conditionally hiding the entire tenant menu
+
+Finally, you can hide the entire tenant menu all togther by using:
+
+```php
+use Filament\Panel;
+
+public function panel(Panel $panel): Panel
+{
+    return $panel
+        // ...
+        ->hideTenantMenu();
+}
+```
+
 ## Setting up avatars
 
 Out of the box, Filament uses [ui-avatars.com](https://ui-avatars.com) to generate avatars based on a user's name. However, if you user model has an `avatar_url` attribute, that will be used instead. To customize how Filament gets a user's avatar URL, you can implement the `HasAvatar` contract:
--- a/packages/panels/docs/10-tenancy.md
+++ b/packages/panels/docs/10-tenancy.md
@@ -371,6 +371,8 @@ public function panel(Panel $panel): Panel
 }
 ```
 
+This may be desirable in simple multitenant setups where a user can only belong to one tenant and therefore there is no need for a menu to switch tenants.
+
 ## Setting up avatars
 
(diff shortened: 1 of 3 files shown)


danharrin:
The only real benefit of our tenancy feature is so you can have switching like this and a different URL scheme... if you're using user-based tenancy you might as well just apply some global scopes and not use the feature? LMK if I'm wrong.

x7ryan:
I'm not saying I have the best use case for this but I only need the scoping on one panel not all which is why I didn't go with a global scope. There probably is still a better way but it is working the menu is just superfluous in my case.

And I can imagine other cases where you may decide for one reason or another to limit a user to a single tenant but maybe use tenancy to share resources between users. Or maybe you want to opt to use your own tenant switcher. For example recreating the Jetstream style switcher in the user menu.

The only viable workaround I can think of right now is to echo out the opening of an html comment before and closing of an html comment after using render hooks which seems kinda silly.

danharrin:
Add the global scope in a middleware and only apply it to 1 panel

x7ryan:
I never though you could configure a global query scope in Middleware but I guess that makes sense. Still dosent address the other possible use cases so I'm sure eventually someone is going to come along wanting to remove the menu for one of those other reasons.

danharrin:
If you have a user belonging to 1 team, for example, you still wouldn't want the team ID in the URL. Again, you would use a global scope there. As I said, the *only* reason you should use the Filament tenancy feature is if you have multiple different tenants that you want to switch between.

x7ryan:
First off I'd disagree that you'd always want that. But I tend to generally avoid statements like always or never because there's often exceptions. I don't have a specific example off the top of my head but I can imagine somewhere someone will want the URL id without the need to have a switcher.

Secondly as I also mentioned it gives more flexibility if someone prefers to use their own tenant switcher. I for example am not a fan of it (maybe it's the avatar but it feels too bulky for my taste) or it's location (it's too up front for my liking) and could see myself preferring to put a list of tenants in the users menu like Jetstream does.with something like this someone (maybe myself in another project) could just turn it off to make their own switcher. 

If the work has already been done I don't see why not provide the option. Especially since there is no reasonable workaround. Just my 2 cents. 

I'm not trying to be difficult just trying to advocate for what I think is best. Because I personally value non opinionated approaches that give the developer optionally if/when for example they decide they don't want the switcher for whatever reason instead of assuming there no legitimate reason for it. So do with that feedback what you want.

danharrin:
Okay, I do understand. I will hold off on merging this though until someone specifically asks for it to be removed from a many-to-many tenant setup. It can always be hidden or customized with CSS if you don't like the look of it.

binaryfire:
@danharrin I was looking for a way to do this and stumbled across this PR. Just adding my vote :) I'd also find this useful for the reasons Ryan mentioned. I'm hiding it via CSS atm but loading unnecessary HTML isn't ideal.

PR #7688, opened 2023-08-11
fix: case insensitive global search for related translatable attributes

miagg:

This will enhance global search to allow for case insensitive search on related translatable attributes. 
It will use a `whereRaw` query in order to append the database collation and allow searching for non-english languages.

**To demonstrate:**
Let's say we have a resource that we need to search for the name of the product.
```php
    public static function getGloballySearchableAttributes(): array
    {
        return ['id', 'product.name'];
    }
```

Previously, searching for a product without specifing the exact case and accents would not return any results (e.g. `μαλαγουζια` instead of `Μαλαγουζιά`

before

After applying the fix, I can search without being strict with case or accents.

after

### DIFF
--- a/packages/admin/src/Resources/Resource.php
+++ b/packages/admin/src/Resources/Resource.php
@@ -6,8 +6,6 @@ use Closure;
 use Filament\Facades\Filament;
 use Filament\GlobalSearch\GlobalSearchResult;
 use Filament\Navigation\NavigationItem;
-use function Filament\Support\get_model_label;
-use function Filament\Support\locale_has_pluralization;
 use Illuminate\Contracts\Support\Htmlable;
 use Illuminate\Database\Connection;
 use Illuminate\Database\Eloquent\Builder;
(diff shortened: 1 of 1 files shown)


packages/admin/src/Resources/Resource.php
```diff
@@ -472,29 +477,29 @@ protected static function applyGlobalSearchAttributeConstraint(Builder $query, a
...
+                function (Builder $query) use ($databaseConnection, $searchAttribute, $searchOperator, $searchQuery, $whereClause, $collation): Builder {
                     $searchColumn = match ($databaseConnection->getDriverName()) {
                         'pgsql' => "{$searchAttribute}::text",
-                        default => "json_extract({$searchAttribute}, '$')",
+                        default => "json_extract({$searchAttribute}, '$') {$collation}",
                     };
 
                     return $query->{"{$whereClause}Raw"}(
-                        "lower({$searchColumn}) {$searchOperator} ?",
+                        "lower({$searchColumn}) {$searchOperator} lower(?)",
```
danharrin: We cannot use lower(?) here since it has performance implications, is this really necessary?

danharrin (changes requested):

packages/admin/src/Resources/Resource.php
miagg: No, it can safely be omitted.

danharrin:
Thanks for the PR!

I've decided to merge it into v3, in #8784. I've made some changes:
- I was afraid that this could be a breaking change for some people, so I've made it so that it only reads `search_collation` instead of `collation`, so this is an opt-in feature now.
- I have applied it to every single search query throughout Filament, not just the global search

PR #7689, opened 2023-08-12
Fix duplicate renderhooks in relation managers

archilex:
LW has a [bug](https://github.com/livewire/livewire/pull/6327) where nested components that use `lazy` will call the boot method twice. This PR turns off lazy loading until the bug is fix.

Additional info:
[Filter sets](https://filamentphp.com/plugins/kenneth-sese-filter-sets) uses the boot method to render itself in Relation Managers using renderhooks. The LW bug causes the boot method to be called twice meaning the renderhook gets duplicated. 

I acknowledge that this is a selfish PR that the team is welcome to close. While it would be helpful, it probably doesn't affect most people, and lazy loading RM's is probably more helpful. I mostly wanted to reference it here so people were aware.

### DIFF
--- a/packages/panels/resources/views/components/resources/relation-managers.blade.php
+++ b/packages/panels/resources/views/components/resources/relation-managers.blade.php
@@ -70,7 +70,7 @@
             class="flex flex-col gap-y-4"
         >
             @php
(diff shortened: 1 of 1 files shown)


danharrin:
It's reasonable, thanks

PR #7692, opened 2023-08-12
Fix nested infolist action record injection

HanakJakub:
I was just trying out repeatable entries with nested action and I couldn't get the correct model in badge action. I was expecting to have `Role` model injected but `User` model is injected instead.

Currently Infolist action is getting record from infolist but changing that to infolistComponent fixed the issue.
 
```php
        return $infolist
            ->record($this->user)
            ->schema([
                RepeatableEntry::make('roles')
                    ->schema([
                        TextEntry::make('name')
                            ->state(fn (Role $record) => $record->type->name) // Injects Role model as expected
                            ->badge()
                            ->hiddenLabel()
                                ->action(
                                    Action::make('remove allergy')
                                        ->requiresConfirmation()
                                        ->action(function ($record) {
                                            // This injects User model instead of Role model
                                            dump($record); // User{}
                                        })
                                )
                    ])
            ]);
```

### DIFF
--- a/packages/infolists/src/Components/Actions/Action.php
+++ b/packages/infolists/src/Components/Actions/Action.php
@@ -45,8 +45,9 @@ class Action extends MountableAction
     protected function resolveDefaultClosureDependencyForEvaluationByName(string $parameterName): array
     {
         return match ($parameterName) {
-            'record' => [$this->getRecord()],
-            'infolist' => [$this->getInfolist()],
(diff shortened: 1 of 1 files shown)


danharrin:
Nice catch. Thanks!

zepfietje:
I think this solves the issue I found and we talked about recently, @danharrin.
Thanks, @HanakJakub!

PR #7708, opened 2023-08-12
assume parsed datetime is in supplied timezone (fixes #7414)

### LINKED ISSUE #7414: DateTimePicker selected value in filter, increasing with each page refresh, if timezone is configured for DateTimePicker
### Package

filament/filament

### Package Version

v3.0.5

### Laravel Version

v10.17.1

### Livewire Version

v3.0.0-beta.6

### PHP Version

PHP 8.2.8

### Problem description

I'm trying to show date-time columns and date-time pickers according to specific timezone which are saved in "users" table. I added **timezone** field to users table (by default it is set to Europe/Berlin).
I configured timezone for DateTimePicker in create/edit form and. for TextColumn, see code example below:
```php
    {
        return $form
            ->schema([
                Forms\Components\DateTimePicker::make('published_at')
                    ->native(false)
                    ->timezone(
                        auth()->user()->timezone
                    ),
            ])
            ->columns(1);
    }
```
```php
    {
        return $table
            ->defaultSort('id', 'desc')
            ->columns([
                Tables\Columns\TextColumn::make('id'),
                Tables\Columns\TextColumn::make('title'),
                Tables\Columns\TextColumn::make('published_at')
                    ->dateTime()
                    ->timezone(
                        auth()->user()->timezone
                    ),
            ]);
    }
```
And here everything works as expected, the value DateTimePicker is saved according to timezone which I set in this component, and in DB **publshed_at** is saved in default application timezone (UTC) as I expected. For the TextColumn same, date is shown correctly according to authenticated user timezone.

However, when I added a filter with two date-time pickers (Published Date From and Published Date To) and configured for both same authenticated user timezone as I did on create/edit page. But when I picked a date in this filter (at least for on date-time picker it doesn't matter which one), I see a filter indicator with selected date-time everything is fine, but when I refresh a page in the browser my selected date-time in the filter is increasing by configured timezone value (increased by 2 hours, because of Europe/Berling timezone). And this value is increasing with each page refreshing.
Here is my configuration for filters:
```php
    {
        return $table
            ->defaultSort('id', 'desc')
            ->columns([
                Tables\Columns\TextColumn::make('id'),
                Tables\Columns\TextColumn::make('title'),
                Tables\Columns\TextColumn::make('published_at')
                    ->dateTime()
                    ->timezone(
                        auth()->user()->timezone
                    ),
            ])
            ->filters([
                Tables\Filters\Filter::make('published_at')
                    ->label('Published Date')
                    ->form([
                        Forms\Components\DateTimePicker::make('from')
                            ->label('Published Date From')
                            ->timezone(
                                auth()->user()->timezone
                            )
                            ->native(false),
                        Forms\Components\DateTimePicker::make('to')
                            ->label('Published Date To')
                            ->timezone(
                                auth()->user()->timezone
                            )
                            ->native(false),
                    ])
                        $query
                            ->when($data['from'], function (Builder $q, string $from) {
                                return $q->where('published_at', '>=', $from);
                            })
                            ->when($data['to'], function (Builder $q, string $to) {
                                return $q->where('published_at', '<=', $to);
                            });
                    })
                    ->indicateUsing(function (array $data) {
                        $indicatorLabels = collect();

                        if ($data['from']) {
                            $indicatorLabels->add('from ' . $data['from']);
                        }

                        if ($data['to']) {
                            $indicatorLabels->add('to ' . $data['to']);
                        }

                        if ($indicatorLabels->isEmpty()) {
                            return null;
                        }

                        return 'Published date ' . $indicatorLabels->implode(' ');
                    })
            ]);
    }
```

https://github.com/filamentphp/filament/assets/9802540/ae7b20be-a70f-464a-95c2-f753215ff215

Not sure if it's a bug or I missed something.

### Expected behavior

The selected value from DateTimePicker in a filter should be the same after refreshing the page (if DateTimePicker has configured custom timezone).

### Steps to reproduce

# Filament Demo
### Demo for reproducing issue with timezone in filters
Steps to reproduce:
1. Clone this repository:
```bash 
git clone https://github.com/bokoch/filament-timezone-issue.git
```
2. Copy .env.example to .env:
```bash 
cp .env.example .env
```
3. Start docker containers:
```bash 
docker-compose up -d
```
4. Install composer dependencies:
```bash 
composer install
```
5. Run migrations:
```bash 
./vendor/bin/sail artisan migrate
```
6. Run seeders:
```bash 
./vendor/bin/sail artisan db:seed
```
7. Go to http://laravel.test/admin and login in 
with credentials:
Email: test@example.com
Password: password
8. Go to http://laravel.test/admin/posts
9. On the posts page open filters menu and pick date for
"**Published Date From**" or/and "**Published Date To**"
10. After picked values will be applied, you will see them in browser
url string and filter indicator. After that refresh the browser page with applied filters
and you will see that picked filter date is increasing
by configured timezone with each page refresh.

### Reproduction repository

https://github.com/bokoch/filament-timezone-issue/

### Relevant log output

_No response_

josefbehr:
Fixes #7414 by providing the timezone to `Carbon::parse()` if the state is not a DateTime/Carbon object. This prevents the auto-incrementing of selected times in filters, but assumes that times set as string are in the provided timezone, which might not always be true (as it might, and is in the case of filters, not be true that times are in application timezone).
This change might break stuff for some users, but will also fix DateTime filters for all those that use different timezones in the db and frontend.

- [see above, change does not break any core functionality] Changes have been thoroughly tested to not break existing functionality.
- [if accepted, I will add documentation for the DateTimePicker] New functionality has been documented or existing documentation has been updated to reflect changes.
- [does not apply] Visual changes are explained in the PR description using a screenshot/recording of before and after.

### DIFF
--- a/packages/forms/src/Components/DateTimePicker.php
+++ b/packages/forms/src/Components/DateTimePicker.php
@@ -84,10 +84,10 @@ class DateTimePicker extends Field implements Contracts\HasAffixActions
 
             if (! $state instanceof CarbonInterface) {
                 try {
-                    $state = Carbon::createFromFormat($component->getFormat(), $state);
+                    $state = Carbon::createFromFormat($component->getFormat(), $state, $component->getTimezone());
                 } catch (InvalidFormatException $exception) {
                     try {
-                        $state = Carbon::parse($state);
+                        $state = Carbon::parse($state, $component->getTimezone());
                     } catch (InvalidFormatException $exception) {
                         $component->state(null);
 


danharrin:
Please give more details about what will break for some users

josefbehr:
> Please give more details about what will break for some users

The only scenario where this would break things is if someone supplied an initial state to the component as a string and also set a timezone on the component. Currently, in that case, the timezone would not be applied to the parsed Carbon object and the supplied time would be taken as is. After this change, in those cases, the supplied datetime-string would be shown in the set timezone. I can't imagine anyone is using it like this though.

danharrin:
I've just reverted this. It was a mistake to merge it. I did not properly understand the issue, and now the date time picker is exhibiting wrong behaviour.

This PR makes it so that all date pickers in normal resource forms do not work correctly with the `timezone()` method. The timezone is not correctly applied when the form is loaded, so the time is not transformed correctly.

The `timezone()` method is only appropriate if you have an existing date in the database that you want to format in a timezone that is more appropriate for the user. And then it gets converted back to the system timezone for saving. It is not meant to be used for date filters.

The reason for this is that the `timezone()` method largely depends on the "hydration / dehydration" process of the field. When using it with a filter, the dehydration process never happens. `$data` is a raw representation of the current date in the filter. If you want to filter by a different timezone, you should apply that transformation within the actual `query()` of the filter. For example, `whereDate('created_at', Carbon::parse($data['date'])->setTimezone('timezone here'))`.

PR #7709, opened 2023-08-13
[Enhancement] Allow rounded corners for square image in table and infolist

josefbehr:

before: 
image

after:
image

I think this looks better and fits with the overall rounded-corner style of Filament. Nothing breaks, completely backwards-compatible, just a small change for better visual styling.

### DIFF
--- a/packages/infolists/docs/03-entries/04-image.md
+++ b/packages/infolists/docs/03-entries/04-image.md
@@ -68,6 +68,16 @@ ImageEntry::make('author.avatar')
     ->square()
 ```
 
+For rounded corners set parameter `rounded` to `true`:
+
+```php
+use Filament\Infolists\Components\ImageEntry;
+
+ImageEntry::make('author.avatar')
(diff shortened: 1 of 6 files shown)


howdu:
This is great @josefbehr but I noticed you have a typo in the class name 'rouneded-xl'.

I think 'rounded-xl' is too much for repeated images in a list/table, I would prefer just `rounded`.

josefbehr:
@howdu thanks. Corrected the typo and changed it to just `rounded` for the image-column.

zepfietje:
Thanks for the PR, @josefbehr.

I do see the use case.
However, the rounding is something you'd probably want to apply application-wide.
I prefer not to introduce an API for this, and suggest adding custom CSS to control the square image rounding.
Potentially in a future Filament version we could ship a slightly rounded square image by default (to make it match the color column for example).
Happy to do that already if @danharrin thinks that's not a breaking change UI-wise.

PR #7743, opened 2023-08-14
Livewire component

njxqlus:
- [ х ] Changes have been thoroughly tested to not break existing functionality.
- [ х ] New functionality has been documented or existing documentation has been updated to reflect changes.

As discussed in #7735


wychoong:
Few questions
- does the component can dehydrate the state, if no, should it be mentioned in doc
- the new livewire component should allow passing class `livewire::make(App\Livewire\Components\Foo::class, […])`

atmonshi:
Better to throw an exception if certain methods are not available as expected with other components.
I like the idea BTW :)

njxqlus:
@wychoong 
> does the component can dehydrate the state, if no, should it be mentioned in doc

The component is a **layout** component (like tabs). I'm not sure but it cannot be dehydrated because it's not a _field_ especially in an infolist.

> the new livewire component should allow passing class livewire::make(App\Livewire\Components\Foo::class, […])

Yes, you can pass a component like this: 
```php
\Filament\Infolists\Components\Livewire::make(\App\Livewire\Foo::class, ['foo' => 'bar'])
```

@atmonshi 

> Better to throw an exception if certain methods are not available as expected with other components

The same methods from `\Filament\Infolists\Components\View` are available like `->hidden()` except `->childComponents()` but I think it's in the concept because, for example, `\Filament\Infolists\Components\ColorEntry` also has this method but doesn't throw an exception

Thanks for getting attention to my PR!

njxqlus:
Make passing _record_ optional as a component data param. Current usage:
```php
use Filament\Forms\Components\Livewire;

Forms\Components\Livewire::make('foo', ['record' => $infolist->getRecord(), 'bar' => 'qux'])
```

```php
// Livewire component
class Foo extends Component
{
    public function mount(array $componentData = []): void
    {
        $this->data = $componentData;       
    }
```

wychoong:
> The component is a **layout** component

I see, that makes a lot of sense now.

> Yes, you can pass a component like this:
> 
> ```
> \Filament\Infolists\Components\Livewire::make(\App\Livewire\Foo::class, ['foo' => 'bar'])
> ```

not LW expert but this feels unsafe, as LW doc never (I might be wrong) show examples of using `@livewire(\App\Livewire\Foo::class)`

would suggest to have some checking if the name passed in is a livewire component, and also resolve the name 
example
```php
public static function make($component, $data){
    // .... check if is valid livewire component
    $this->componentName = $component::getName();
    // ...
}

// in blade
@livewire($componentName, $data)
```

njxqlus:
@wychoong 

> not LW expert but this feels unsafe, as LW doc never (I might be wrong) show examples of using `@livewire(\App\Livewire\Foo::class)`

I don't use it like this. Current usage in blade.php:
```php
@livewire($component, compact('componentData'))
```

rahmanramsi:
Can you add lazy load support too?
Something like :
```php
Livewire::make(\App\Livewire\Foo::class, ['foo' => 'bar'])->lazy()
```

njxqlus:
> Can you add lazy load support too? 

I'm using method **view** from `Filament\Forms\Components`. My component is just a shortcut for `\Filament\Forms\Components\View`. So if the View component supports `->lazy()` my component might support it also.

rahmanramsi:
> I'm using method **view** from `Filament\Forms\Components`. My component is just a shortcut for `\Filament\Forms\Components\View`. So if the View component supports `->lazy()` my component might support it also.

I mean this implementation [https://livewire.laravel.com/docs/lazy](https://livewire.laravel.com/docs/lazy)

njxqlus:
> I mean this implementation https://livewire.laravel.com/docs/lazy

Oh sorry, I didn't understand. 

Now I'm using blade directive `@livewire()` to call a component. Not sure that it works with lazy. I will have a look

rahmanramsi:
> Now I'm using blade directive `@livewire()` to call a component. Not sure that it works with lazy. I will have a look

I think this will work `@livewire($component, ['lazy' => true])`

njxqlus:
Update usage of component.
```php
\Filament\Forms\Components\Livewire::make(\App\Livewire\Foo::class)
    ->lazy()
    ->withRecord()
    ->componentData(['bar' => 'form'])
```

@rahmanramsi now it's supports `->lazy()`

wychoong:
How it goes with if https://github.com/livewire/livewire/pull/5780 get merged some day?

njxqlus:
@wychoong 
> How it goes with if [livewire/livewire#5780](https://github.com/livewire/livewire/pull/5780) get merged some day?

This component is part of a form or an infolist so it can't be a full-page livewire component. Despite that, I call livewire component in _native_ way via `@livewire()`blade directive which shouldn't cause any problems.

danharrin:

danharrin:
Thanks!

Nicholasbell03:
@njxqlus Thanks for your work on this. I wanted to find something out I am struggling with. I am wanting to use this to add a livewire component essentially as a custom step in my Form Wizard. The issue I'm facing is getting the livewire component to essentially be part of the form flow? i.e. the component has form awareness. So in the livewire component I can perform logic with the current state of the form if that makes sense? 

I'm not sure if you can help, I see that you state in the docs that when being used for a new record creation it can be hidden, but I'm wanting to use this as part of the form wizard flow for new and existing records. 

just thought I'd reach out as I've been battling with this for a little while now.

njxqlus:
@Nicholasbell03 I'm not sure about what you want to achieve. Wizard is just a layout so you can use it with any components inside.

From docs:
> Since all Filament forms are built on top of [Livewire](https://livewire.laravel.com/), the form can adapt dynamically to user input, even after it has been initially rendered. Developers can use [parameter injection](https://filamentphp.com/docs/3.x/forms/advanced#form-component-utility-injection) to access many utilities in real time and build dynamic forms based on user input. The [lifecycle](https://filamentphp.com/docs/3.x/forms/advanced#field-lifecycle) of fields is open to extension using hook functions to define custom functionality for each field. This allows developers to build complex forms with ease.

Nicholasbell03:
@njxqlus thank you for your response.
Sorry I know I didn't explain my issue very well. 

Perhaps an example. 
I have a filament form 
- It includes 2 fields (native filament functionality)
- Then the 3rd field is a custom Livewire component 
- I want the livewire component to compute the sum of the first two fields, in realtime, so as changes are made they reflect in the livewire component. 

So the livewire component needs the current raw state of the parent filament form at all times if that makes sense? This is a really basic example, but I think it expresses the same problem I have. 

I've found a way to do this with event dispatch and a listener inside the livewire component, but it seems like a hacky approach. I'm not sure if I am missing something really obvious here and just can't see the wood from the trees. 

Many thanks again for the feature you added and taking the time to respond.

juliancc:
@Nicholasbell03, did you figure this out? I'm struggling with my custom Livewire form field component and believe it might be the same issue you had. I want the Livewire component to maintain its state just like a regular Filament field.

For example, I have an input field inside my custom Livewire component that I want to attach to the form's state.

The way I'm doing it right now (just like you mentioned) feels like a hack. I have to dispatch events from my component, then create a listener in the form's Livewire component to store whatever value is in the custom component. This way, I can save it with the beforeRegister() hook.

Nicholasbell03:
@juliancc it sounds like what you're doing is somewhat the same as what I was trying to achieve, except the opposite way around. i.e. you're wanting the filament form to receive updates from the livewire component. To be honest, I haven't yet had the need to build a custom livewire component that accepts input as part of the form data. 
I suspect this has been done a lot though as it would be a very common use case. Have you looked on the Discord group, I'd be surprised if this matter hasn't been raised. 
Sorry I can't be of more help. 
To answer the question about my problem, I ended up using event dispatching and listening as I never found an alternative. I wasn't particularly happy with it, but couldn't figure out another way.

PR #7747, opened 2023-08-14
[3.x] add closure customization to `transformOptionsForJs`

saade:

I have a need to use JSON as Select values. While this is possible if we override the `getOptionLabelUsing`, the js options won't reflect that change since the values for select are transformed in a separate function. This PR adds closure customization to the `transformOptionsForJs` method.

## Before this PR:
```php
\Filament\Forms\Components\Select::make('option_id')
  ->options([
    1 => ['id' => 1, 'title' => 'One', 'deleted' => true],
    2 => ['id' => 2, 'title' => 'Two', 'deleted' => false],
  ])
  ->getOptionLabelUsing(static function (\Filament\Forms\Components\Select $component, $value): ?string {
    $options = $component->getOptions();
    return $options[$value]['title'];
  })
  ->dehydrateStateUsing(function (\Filament\Forms\Components\Select $component, $state) {
    return $component->getOptions()[$state];
  });
```
Without closure customization the options presented in the UI are wrong.

image

## After this PR:

```php
->transformOptionsForJsUsing(static function (array $options): array {
  return collect($options)
    ->map(fn (array $option) => ['label' => $option['title'], 'value' => strval($option['id'])])
    ->values()
    ->all();
})
```

Allowing closure customization allows me to format the options for JS and use JSON as values.

image


saade:
@danharrin Hi mate, would you mind reviewing this? It does not break any existing behavior as it's just a closure customization.

Sorry for the ping btw, just needing this real bad :/

danharrin:
No worries

PR #7754, opened 2023-08-14
Add file detection options

killjin:

image

This option speeds up the form access speed. Trying to modify js is unsuccessful. If it is 0 here, it is better to show that the file size is unavailable.


killjin:
Is there a way to return 0 or -1 or false, hide the file size display, or display as Size not available, for example:

filepond Labels: labelFileSizeNotAvailable
DEFAULT VALUE: 'Size not available'
Label used when no file size information was received

packages/forms/src/Components/BaseFileUpload.php
```diff
@@ -79,7 +81,9 @@ protected function setUp(): void
             $files = collect(Arr::wrap($state))
                 ->filter(static function (string $file) use ($component): bool {
                     try {
-                        return blank($file) || $component->getDisk()->exists($file);
```
killjin, line 82: here should be 
```php
filled($file) && $component->getDisk()->exists($file);
```

Nuhel:
```php
 $FileDetection  = $component->shouldFileExistDetection();
```
 
 Variable name should started with lowercase latter.

danharrin:

danharrin:
Thanks

PR #7770, opened 2023-08-15
only discover PHP files for discoverResources/Pages/Widgets

wivaku:
When discovering Resources/Pages/Widgets: only look for files with php extension (which was the case in v2).  
Added optional parameter to continue to use to original v3 behaviour: discover all files.

Use case: easily exclude files by renaming them (to e.g. MyResource.php.bak). Normally discoverResources etc would include those files which would fail composer autoload.

Current PR is set to exclude non-PHP files by default ($all=false), which mimics the v2 behaviour.  
Feel free to inverse that default ($all=true), which makes the PR backwards compatible with original v3.

https://github.com/filamentphp/filament/discussions/7340

### DIFF
--- a/packages/panels/src/Panel/Concerns/HasComponents.php
+++ b/packages/panels/src/Panel/Concerns/HasComponents.php
@@ -149,7 +149,7 @@ trait HasComponents
         return $widget;
     }
 
-    public function discoverPages(string $in, string $for): static
+    public function discoverPages(string $in, string $for, bool $all=false): static
     {
         $this->pageDirectories[] = $in;
         $this->pageNamespaces[] = $for;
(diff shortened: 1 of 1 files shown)


wychoong:
not sure if v2 only discover `.php` files, but I think this is a reasonable change that to exclude not `.php` files without the need to have the option to `$all = true`

danharrin:
I don't think we need the `$all` option at all? It can always exclude non-PHP files

wivaku:
If `$all` is not needed at all (and I agree) then it's probably easiest to deal with it the same way as in v2: check if the class exists.

https://github.com/filamentphp/filament/blob/9177b9c3fda17409d1c8ea9169805171756c1e9c/packages/admin/src/FilamentServiceProvider.php#L156-L164

Add PR using that approach: https://github.com/filamentphp/filament/pull/7780

danharrin:
Closing in favour of #7780

PR #7774, opened 2023-08-15
Improve profile page actions layout

ryangjchandler:

Changes the layout of the default "Profile" page. Original design had an action link which looked out of place.

New design in screenshot below.

CleanShot 2023-08-15 at 11 43 22@2x

Also introduced a new `getBackFormAction()` method for consistency with the `getSaveFormAction()` method.
Marked `backAction()` method as deprecated, pointing to new method as replacement.

Added `Alignment::Between` case and support for that inside of the actions component.


zepfietje:
Personally, I like right-aligned actions, but since we're left-aligning them in our design system, I think it's best if we do so for these pages too.

Makes them match modals and looks ok if you change the "Back" label to "Cancel" (just like we have for modals and resource edit pages).

danharrin:
@zepfietje what needs to be done here to get it ready for merge?

zepfietje:
> @zepfietje what needs to be done here to get it ready for merge?

1. Change `Back` to `Cancel`.
2. Align buttons to start, so save button comes before cancel.
3. Make sure buttons are **not** full-width.

Request my review when those changes have been implemented (by whoever is working on it now 😅) and I'll merge it. 😄

ryangjchandler:
@zepfietje 
CleanShot 2023-08-17 at 10 08 28@2x

packages/panels/src/Pages/Auth/EditProfile.php
```diff
@@ -269,7 +275,12 @@ protected function getSaveFormAction(): Action
 
     protected function hasFullWidthFormActions(): bool
     {
-        return true;
+        return false;
+    }
+
+    public function getFormActionsAlignment(): string
```
danharrin:
```suggestion
    public function getFormActionsAlignment(): string | Alignment
```

danharrin: Might need to update a few other places

packages/actions/resources/views/components/actions.blade.php
```diff
@@ -27,6 +27,7 @@
                     match ($alignment) {
                         Alignment::Center, 'center' => 'justify-center',
                         Alignment::End, Alignment::Right, 'end', 'right' => 'flex-row-reverse',
+                        Alignment::Between, 'between' => 'justify-between',
```
zepfietje, line 30: We no longer need this, right?

packages/panels/resources/lang/en/pages/auth/edit-profile.php
```diff
@@ -43,7 +43,7 @@
     'actions' => [
 
         'back' => [
```
zepfietje: Not sure if we can somehow change the `back` key to `cancel` in a non-breaking way?

danharrin: We can just find and replace

zepfietje: Wouldn't that break applications where people override this translation string?

danharrin: We haven't been considering that as a breaking change up until this point

zepfietje: Fine.

PR #7817, opened 2023-08-16
Add sticky header support to aside section

lukas-frey:

This adds the possibility to make the description of an aside section sticky. If aside or sticky is set to false, it will behave like it did until now. Only if aside AND sticky are both true, it will stick to the top.

https://github.com/filamentphp/filament/assets/10926334/26786dd3-0b62-4bac-a4c8-f78ed60375f2


packages/forms/src/Components/Section.php
```diff
@@ -21,6 +22,7 @@ class Section extends Component implements Contracts\CanConcealComponents, Contr
     use HasHeading;
     use HasIcon;
     use HasIconColor;
+    use CanBeStickied;
```
zepfietje: Should be `CanBeSticky`.

zepfietje (changes requested):
It's really only the header that's sticky, right?
I think that should be reflected in the API.
With the current `sticky()` API I would expect sections to actually be sticky as a whole.

lukas-frey:
> It's really only the header that's sticky, right? I think that should be reflected in the API. With the current `sticky()` API I would expect sections to actually be sticky as a whole.

Yes, only the header. So would you suggest something like `stickyHeader()` or `withStickyHeader()` instead?

zepfietje:
`stickyHeader()` should work fine I think. Have a look at what's the exact API for sticky modal footers; we could use similar naming.

lukas-frey:
Ready for review. The only thing I'm unsure about is whether the `CanBeSticky` name for the trait is still valid.

Since it defines whether the header is sticky, maybe something like `HasStickyHeader` or `CanHaveStickyHeader` or something is suited better?

zepfietje:
I think `CanHaveStickyHeader` should be good. Though let's see what @danharrin thinks.

zepfietje:
Also I'm thinking that technically you'd expect the header to be sticky even on mobile or when not `aside`.

lukas-frey:
> Also I'm thinking that technically you'd expect the header to be sticky even on mobile or when not `aside`.

Yeah, but design-wise it doesn't make much sense imo, when there's a long description.

danharrin:

zepfietje:
@danharrin, I think not just the `aside` section should support this. Let's try adding it to the regular section as well.

danharrin:
I'm honestly not sure we should support this. It relies on the fact that the section headers are sticky `-20` from the top of the screen, which only really works in the admin panel where the topbar is that height. In another app, it might be a different distance.

I think this feature could be implemented quite easily in CSS within any app:

```php
Section::make('Heading')->extraAttributes(['class' => 'fi-section-sticky-header'])
```

```css
.fi-section-sticky-header .fi-section {
    @apply md:relative;
}

.fi-section-sticky-header .fi-section-header {
    @apply md:sticky md:start-0 md:top-20;
}
```

PR #7818, opened 2023-08-16
Added instructions for PostgreSQL for searchable()

timsinakiran:
Apparently, the current searchable() implementation does not work with PosgreSQL. Though a better fix might be needed later, this might help with the ones struggling right now.

### DIFF
--- a/packages/tables/docs/03-columns/01-getting-started.md
+++ b/packages/tables/docs/03-columns/01-getting-started.md
@@ -175,6 +175,8 @@ TextColumn::make('full_name')
     })
 ```
 
+If you are using PostgreSQL, check `forceSearchCaseInsensitive()` which is explained in later.
+
 ### Searching individually
 
 You can choose to enable a per-column search input field using the `isIndividual` parameter:
@@ -240,6 +242,7 @@ TextColumn::make('name')
     ->searchable()
     ->forceSearchCaseInsensitive()
 ```
+For PostgreSQL works slighly different that other databases, `forceSearchCaseInsensitive()` is required for search to work properly with PostgreSQL.
 
 ## Column actions and URLs
 


danharrin:
Apparently it does work if you use a case-insensitive database collation though? If this didn't work out of the box in Postgres then we wouldn't have this as the default behaviour, we would instead handle it how we did it in v2. Instructing all Postgres users to force insensitivity is reintroducing those previous performance issues that we fixed. How can we handle this better in the code to ensure people don't need to force?

timsinakiran:
@danharrin 

The recommended alternative to using 'ilike' in postgres is by using lower() in the search column, which is what the forceCaseInsensitive() does.
`function (EloquentBuilder $query) use ($search, $searchColumn, $whereClause): EloquentBuilder {
                    $caseAwareSearchColumn = $this->isSearchForcedCaseInsensitive() ?
                        new Expression("lower({$searchColumn})") :
                        $searchColumn;

                    return $query->when(
                        $this->queriesRelationships($query->getModel()),
                        fn (EloquentBuilder $query): EloquentBuilder => $query->{"{$whereClause}Relation"}(
                            $this->getRelationshipName(),
                            $caseAwareSearchColumn,
                            'like',
                            "%{$search}%",
                        ),
                        fn (EloquentBuilder $query): EloquentBuilder => $query->{$whereClause}(
                            $caseAwareSearchColumn,
                            'like',
                            "%{$search}%",
                        ),
                    );
                },`
                
This already has better performance than using 'ilike' which has been demonstrated many times. Here is my output from 100 tests in my local db that is about 1.1GB (using Postgres).

select id,name from users where lower(name) like lower('Kiran Timsina');
  id   |     name      
-------+---------------
 XXXXX | Kiran Timsina
 XXXXX | Kiran Timsina
    XX | Kiran Timsina
     X | Kiran Timsina
(4 rows)

Time: 39.311 ms

select id,name from users where name ilike lower('Kiran Timsina');
  id   |     name      
-------+---------------
 XXXXX | Kiran Timsina
 XXXXX | Kiran Timsina
    XX | Kiran Timsina
     X | Kiran Timsina
(4 rows)

Time: 56.157 ms

Asking postgres users to use force insensitivity is the only apparent solution for the search to work, which is still way performant than using 'like'. This could also be done through a config by forcing insensitivity globally for postgres users. I will make a PR if that's a better approach instead.

Thank you

chrisreedio:
I can confirm with the default Filament setup out of the box on Postgres 13, my search is majorly glitched. 

I hadn't dug into it yet but was going to have to investigate before we take any of our new Filament sites live.

danharrin:
I'm leaning towards adding a default for `isSearchForcedCaseInsensitive()` on Postgres only to be true

danharrin:
Have you read this @timsinakiran @chrisreedio https://www.cybertec-postgresql.com/en/case-insensitive-pattern-matching-in-postgresql/

timsinakiran:
Thank you for pointing that @danharrin 

I tried but couldn't get any performance improvement over using lower().

Upon digger deeper, I realize the issue arises because we convert the query string to lower case and try to use 'like' for querying it. If we do not convert the query string case, the search function without glitch. I tested this by forcing the value of $search to be back to what I had entered in InteractsWithTableQuery's applySearchConstraint(), and the search works without glitch.

timsinakiran:
As said above, here is the line that's causing the issue:

// CanSearchRecords.php
protected function applyGlobalSearchToTableQuery(Builder $query): Builder
    {
        $search = trim(Str::lower($this->getTableSearch()));
        ...
        
I'm happy to make a PR for this

danharrin:
So the issue is with `Str::lower()`? I'm not sure I understand. I know that at least on MySQL, if we don't do that, then it is not case insensitive by default.

timsinakiran:
Postgres has slightly different way to handle this. 'like' is a case sensitive pattern matching operator while 'ilike' is case insensitive pattern matching operator.

Using an index in the search column (the job of the system designer/backend devs, not ours) providing a blazing fast query using 'like' paired with lower(). 

Now since in our approach, we are forcing the query string to be lower, and using 'like' without forcing the column value to lower(), we will not get the search result in postgres. We should either use 'ilike' or use lower(search_column) if we are lowering the search string. 

Alternately, we can avoid lowering the search string for postgres which will perform search, but not case insensitive unless we force it to do using the current implementation. We can put this info in docs for postgres user that for case insensitive search, we must force it.

danharrin:
In the article I linked, it mentions using a case-insensitive ICU collation. Is that an option here? It feels weird to *only* make the string lowercase for MySQL 🤔

danharrin:
Okay I've just tested with MySQL, and removing the `Str::lower()` doesn't appear to have an impact on functionality, which is good

danharrin:
Opened #7849

timsinakiran:
Cheers!

PR #7824, opened 2023-08-16
fix: Null relationship select

### LINKED ISSUE #7822: Laravel Select component relationship bug: Null value not properly handled when using relationship() method
### Package

filament/forms

### Package Version

v3.0-stable

### Laravel Version

10.10

### Livewire Version

_No response_

### PHP Version

8.2.7

### Problem description

There is a possible bug with the relationship on Select component.

The bug is that if you use a relationship on Select to select one value and save it, it works correctly. However, if you want to make the value null, you can select the "Select an option" option to select a null value. However, the component does not allow you to make it null, and it will save the old value.

This bug only occurs when you use the relationship() method. If you use the option() method instead, it works correctly.

The issue is happening after the $this->dehydrateState($state) method in the HasState concerns. This method is responsible for converting the state of the component into a serializable format. For some reason, it is not properly handling the null value when it is used with the relationship() method.

image

I tried to replicate locally
image

### Expected behavior

It should allow to make the select null or EMPTY instead of saving the old value

### Steps to reproduce

1. Go to https://demo.filamentphp.com/shop/products/1/edit
2. Make brand value null
image
3.Click Save

### Reproduction repository

### Relevant log output

_No response_

danharrin:
Fixes #7822.

### DIFF
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -988,6 +988,8 @@ class Select extends Field implements Contracts\HasAffixActions, Contracts\HasNe
             $relationship = $component->getRelationship();
 
             if (! $relationship instanceof BelongsToMany) {
+                $relationship->associate($state);
+                
                 return;
             }
 


AAbosham:
Hi @danharrin ,

After this fix introduced a new bug for saving a singular relationship.

image


```
Forms\Components\Section::make('Address')
                ->relationship('address')
                ->schema([
                    Forms\Components\Grid::make()
                        ->schema([
                            Forms\Components\Select::make('region_id')
                                ->relationship('region', 'name')
                                ->preload()
                                ->reactive(),

                            Forms\Components\Select::make('city_id')
                                ->relationship(
                                    'city',
                                    'name',
                                    fn (Get $get, $query) =>
                                    $query->where('region_id', $get('region_id'))
                                )
                                ->reactive()
                                ->searchable(),

                            Forms\Components\Select::make('district_id')
                                ->relationship(
                                    'district',
                                    'name_ar',
                                    fn (Get $get, $query) =>
                                    $query->where(
                                        'city_id',
                                        $get('city_id')
                                    )
                                )
                                ->searchable()
                        ])->columns(3),

                    Forms\Components\Grid::make()
                        ->schema([
                            Forms\Components\TextInput::make('street')
                                ->maxLength(255),

                            Forms\Components\TextInput::make('building_no')
                                ->numeric(),
                        ])->columns(2),

                  
                ]),
```

After creating a new record, an error occurred

image

PR #7833, opened 2023-08-17
Fix filter indicator wrapping

archilex:
Currently when there are multiple filters, the indicators expand past the container. The pushes the "X" close button off screen, but also means you can scroll the table into the navigation. This PR adds `flex-wrap` to the indicator `div` to fix this. 

Before
Screenshot 2023-08-16 at 10 12 51 PM (2)

After
Screenshot 2023-08-16 at 10 13 32 PM (2)

While this fixes the main problem, visually you'll see that "Active Filters" is now split onto two lines. There are other options we can consider:

1. We could also add `whitespace-nowrap` to "Active filters" so that it remains on one line:
Screenshot 2023-08-16 at 10 15 02 PM

2. Another option could be to find a way to move the indicators below "Active filters" and the "X" when the width of the indicators is larger than the space available between "Active filters" and the "X". This is actually what happens on mobile. 
Screenshot 2023-08-16 at 10 28 44 PM

If you remove the `sm:flex-row` and `sm:items-center` you can accomplish this effect, BUT it has the downside that now even just one filter forces two rows, no matter the width. This look worse IMO.
Screenshot 2023-08-16 at 10 37 04 PM

I don't think container queries works here since we need to change the parent container depending on the width of the children (indicators).

Another option would be to use JS/Alpine to keep track of the width of the indicators and conditionally apply classes to the parent so that when the indicators are wider it can move down to a new line like mobile. 

Let me know how you'd like to proceed and I'll gladly dig into it.

### DIFF
--- a/packages/tables/resources/views/components/filters/indicators.blade.php
+++ b/packages/tables/resources/views/components/filters/indicators.blade.php
@@ -5,9 +5,9 @@
 <div
     {{ $attributes->class(['fi-ta-filter-indicators flex items-start justify-between gap-x-3 bg-gray-50 px-3 py-1.5 dark:bg-white/5 sm:px-6']) }}
 >
-    <div class="flex flex-col gap-x-3 gap-y-1 sm:flex-row sm:items-center">
+    <div class="flex flex-col gap-x-3 gap-y-1 sm:flex-row">
         <span
-            class="text-sm font-medium leading-6 text-gray-700 dark:text-gray-200"
+            class="text-sm font-medium leading-6 text-gray-700 whitespace-nowrap dark:text-gray-200"
         >
             {{ __('filament-tables::table.filters.indicator') }}
         </span>
--- a/packages/tables/resources/views/components/filters/indicators.blade.php
+++ b/packages/tables/resources/views/components/filters/indicators.blade.php
@@ -12,7 +12,7 @@
             {{ __('filament-tables::table.filters.indicator') }}
         </span>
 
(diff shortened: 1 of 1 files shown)


danharrin:
We should add `whitespace-nowrap` to the label, and then for mobile *only* we should force it so that the label is on its own line. So a combination of both options IMO :)

archilex:
I might not have been clear, or maybe I'm misunderstanding, but mobile already forces the label to one line above the filter indicators so that part is done. Sounds like you're good with option 1 to handle everything wider than mobile.

archilex:
With this commit

Mobile:
Screenshot 2023-08-17 at 4 28 05 AM

Desktop:
Screenshot 2023-08-17 at 4 28 14 AM

danharrin:
Noice

zepfietje:
Ideally the badges would be below the label when they wrap, but I think that's hard to achieve while keeping them on the same line when not wrapping. So looks good now. Thanks guys.

archilex:
@zepfietje Yes, I agree that it looks best to have them below when they need to wrap. I looked into `@container` paired with `:has` which gets close, but it's not perfect (you'd have to do it by the number of elements, not the overall width of the children) and there are too many edge cases. 

I think JS is currently the only way to accomplish this. Do you want me to explore that? Or is this good enough?

zepfietje:
Yeah, thanks for exploring, @archilex.
I think the current approach is good enough.
Generally prefer to not add too much JS for these things. 😄

PR #7837, opened 2023-08-17
[3.x] Move ku to ckb with complete translation

AlanD20:

Hey there!

This PR moves existing `ku` translation to `ckb` that includes full complete Central Kurdish translation.

**Why?**
The Kurdish language has a variety of dialects, leading to inconsistencies in many packages. The `ku` locale, for instance, might represent Central Kurdish translation in one package, but in another package, it could be Northern or Southern Kurdish. These dialect differences are significant and could result in unintentional selection of the locale.

Therefore, this pull request proposes a consistent [CLDR](https://cldr.unicode.org/index) locale code by moving the existing `ku` locale to `ckb` locale, which represents Central Kurdish translation, into its own designated locale code.

Some packages support `ckb` for Central Kurdish translation such as [Carbon](https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Lang/ckb.php) which is very intuitive to easily sync the translation once it's changed within Laravel's locale, but unfortunately some JS packages haven't merged the already opened migration from `ku` to `ckb` locale for Central Kurdish translations. Therefore, this PR will use the existing `ku` translation in `dayjs` and exports it as `ckb` translation. If in the future these locales merged successfully, we can also update the locale packages for Filament as well.

After [FilePond merges `ckb` locale](https://github.com/pqina/filepond/pull/927), we can also update supported locales in Filament.

Thank you!

**References:**

- [Picking the right language identifier by Unicode CLDR](https://cldr.unicode.org/index/cldr-spec/picking-the-right-language-code)
- [CKB in Glottolog](https://glottolog.org/resource/languoid/id/cent1972)
- [CKB in Ethnologue](https://www.ethnologue.com/language/ckb/)
- [CKB in ISO 639-3](https://iso639-3.sil.org/code/ckb)


danharrin:
Is this a breaking change for people currently using `ku` though?

danharrin:
I think so, so we probably need to copy the translations over rather than move them completely

AlanD20:
Yes. One possible solution we could do is that, we have to sync both `ku` and `ckb` translations, till someone else in the future reserves `ku` locale to another Kurdish translation dialect.

Another thing is that, I'm not sure what `ku` locale is actually reserved for since if we follow [ISO 639-3](https://iso639-3.sil.org) for consistency, each Kurdish dialect has its own identifier.

ISO 639-3 - [ckb](https://iso639-3.sil.org/code/ckb) -> Central Kurdish
ISO 639-3 - [kmr](https://iso639-3.sil.org/code/kmr) -> Northern Kurdish
ISO 639-3 - [sdh](https://iso639-3.sil.org/code/sdh) -> Southern Kurdish

Let me know if we should copy the entire `ckb` translation to `ku` translation for the purpose of not breaking existing translation.

danharrin:
Yes please, they need to be kept in sync until further notice

AlanD20:
Done. Thank you!

danharrin:
Thanks

PR #7852, opened 2023-08-17
Update documentation for naming actions

eazyurk:
Added a new section "Naming Actions" in overview documentation for Filament package. It clarifies allowable characters for action names.


eazyurk:
Created this PR because I ran into a bug that I had a action with a dash in it's name and my modal wouldn't open. Spent way too long debugging why it didn't work and thought it would be a good idea to add this in the docs

danharrin:
Thanks, but this is essentially just telling people that PHP methods don't allow `-` in their names, so I don't think it's very relevant to Filament actions.

Actions should allow `-` in their names if they are, for example, table actions. The only actions that don't allow it in their names are actions mounted to a Livewire component method.

PR #7866, opened 2023-08-17
Fix bulk action records should be not loaded before actually executed

bangnokia:

"This will prevent the database query from being executed to retrieve records for the action if the action doesn't require using the selected records. For instance, when dealing with a selection of thousands of records, we could utilize the `Table $table` injection into the action closure to manage the query. This approach helps in avoiding the loading of thousands of models, which could potentially lead to reaching the memory limit."

### DIFF
--- a/packages/tables/src/Table/Concerns/HasBulkActions.php
+++ b/packages/tables/src/Table/Concerns/HasBulkActions.php
@@ -113,7 +113,7 @@ trait HasBulkActions
     public function getBulkAction(string $name): ?BulkAction
     {
         $action = $this->getFlatBulkActions()[$name] ?? null;
-        $action?->records($this->getLivewire()->getSelectedTableRecords());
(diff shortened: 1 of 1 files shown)


packages/tables/src/Table/Concerns/HasBulkActions.php
```diff
@@ -113,7 +113,7 @@ public function getFlatBulkActions(): array
     public function getBulkAction(string $name): ?BulkAction
     {
         $action = $this->getFlatBulkActions()[$name] ?? null;
-        $action?->records($this->getLivewire()->getSelectedTableRecords());
+        $action?->records(fn() => $this->getLivewire()->getSelectedTableRecords());
```
danharrin:
```suggestion
        $action?->records($this->getLivewire()->getSelectedTableRecords(...));
```

PR #7882, opened 2023-08-18
Add togglable collapsed class

howdu:

Add the `fi-collapsed` class if the element is collapsed. 

After adding a background color to the section header also needed to adjust the rounding when collapsed. 
Screenshot 2023-08-18 at 13 15 14

```css
.fi-section > header {
    @apply rounded-t-xl bg-gray-50 dark:bg-gray-700;
}

.fi-section.fi-collapsed > header {
    @apply rounded-b-xl;
}
```

### DIFF
--- a/packages/forms/resources/views/components/builder.blade.php
+++ b/packages/forms/resources/views/components/builder.blade.php
@@ -61,6 +61,7 @@
                         x-data="{
                             isCollapsed: @js($isCollapsed($item)),
                         }"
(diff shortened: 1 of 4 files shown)


zepfietje:
Honestly, I think the rounding fix needs to be applied in core?

zepfietje:
Also `x-bind:class` should be moved down to be the final attribute on elements, just like we do elsewhere and with `class`.

Could you handle both in a new PR, @howdu?

Issue #7809, opened 2023-08-16
Customized routes are inaccessible


StringKe:
### Package

filament/filament

### Package Version

3.0.23

### Laravel Version

10.19.0

### Livewire Version

3.0.0-beta.9

### PHP Version

8.2.8

### Problem description

Customized routes are inaccessible

https://github.com/filamentphp/filament/blob/3.x/packages/panels/routes/web.php#L129-L131

Accessible before manually adjusting routes to tenant routes

Following the rules of laravel routing, maybe the order should be adjusted here?

![router](https://github.com/filamentphp/filament/assets/31089228/40b76644-ed1d-4c78-88dd-cd71e1f5701c)

### Expected behavior

Unable to access route

### Steps to reproduce

```php
<?php

namespace App\Providers\Filament;

use App\Filament\Team\Pages\Tenancy\BindTeam;
use App\Filament\Team\Pages\Tenancy\RegistryTeam;
use App\Models\Team;
use Filament\Http\Middleware\Authenticate;
use Filament\Http\Middleware\DisableBladeIconComponents;
use Filament\Http\Middleware\DispatchServingFilamentEvent;
use Filament\Pages;
use Filament\Panel;
use Filament\PanelProvider;
use Filament\Support\Colors\Color;
use Filament\Widgets;
use Illuminate\Cookie\Middleware\AddQueuedCookiesToResponse;
use Illuminate\Cookie\Middleware\EncryptCookies;
use Illuminate\Foundation\Http\Middleware\VerifyCsrfToken;
use Illuminate\Routing\Middleware\SubstituteBindings;
use Illuminate\Session\Middleware\AuthenticateSession;
use Illuminate\Session\Middleware\StartSession;
use Illuminate\Support\Facades\Route;
use Illuminate\View\Middleware\ShareErrorsFromSession;

class TeamPanelProvider extends PanelProvider
{
    public function panel(Panel $panel): Panel
    {
        return $panel->default()
            ->id('portal')
            ->path('portal')
            ->authGuard('web')
            ->login()
            ->registration()
            ->profile()
            ->emailVerification()
            ->tenant(Team::class, slugAttribute: 'slug')
            ->tenantRegistration(RegistryTeam::class)
            ->tenantRoutes(function () {
                // add to /portal/{teamSlug}/*
                Route::get('/ccc', BindTeam::class)
                    ->name('ccc');
                Route::get('/ddd', BindTeam::class)
                    ->name('ddd');
            })
[...]

PR #7891, opened 2023-08-18
Prioritize custom routes over the defaults

margarizaldi:

Move down all custom routes after the default routes to enable route overriding.

### DIFF
--- a/packages/panels/routes/web.php
+++ b/packages/panels/routes/web.php
@@ -23,10 +23,6 @@ Route::name('filament.')
                     ->name("{$panelId}.")
                     ->prefix($panel->getPath())
                     ->group(function () use ($panel, $hasTenancy, $tenantRoutePrefix, $tenantSlugAttribute) {
-                        if ($routes = $panel->getRoutes()) {
-                            $routes($panel);
-                        }
-
                         Route::name('auth.')->group(function () use ($panel) {
                             if ($panel->hasLogin()) {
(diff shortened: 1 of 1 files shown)


danharrin:
Isn't this the reverse of #7810 which fixed an issue?

danharrin:
Please investigate that issue and provide more information if you would like me to merge this

margarizaldi:
> Please investigate that issue and provide more information if you would like me to merge this

I've checked and tested the problem of #7809 and now I can see the dilemmatic situations since the "global" routes (login, logout, register, etc.) are placed at the same level with the tenant slug[^1].

However, the fix #7810 came with a drawback that avoided us to override any default routes. The best option I can think so far is stick to the current version, close this PR, and enable customizations for all global routes.

[^1]: To be honest, I'm not the fan of this route structure, it requires us to list some "reserved words" if the tenant slug is a meaningful string (yeah, we're building a SaaS for customers, so we would never expose the tenant id in the url). If possible I will move my tenant routes to a sub path like `/u/{tenant}/path`.

danharrin:
Please suggest how you want to achieve customisation of global routes, as you mention

margarizaldi:
> Please suggest how you want to achieve customisation of global routes, as you mention

Basically I want to be able to override the default routes so we need to put the custom routes at the end. But then it creates another issue making tenant routes being inaccessible. So since this is not critical to my current projects, I won't think too much for that currently.

PR #7900, opened 2023-08-19
Increase font size for lg/xl links

howdu:

Increased font size for lg/xl size for buttons and links. 

The reason for the change is that I needed a large link and apart from 'xs' all size options are the same size. 
```blade
<x-filament::link size="lg">
 Large link
</x-filament::link>
``` 

Applied the same change to the buttons for consistency when stacked alongside.

Before
Screenshot 2023-08-19 at 08 57 34

After
Screenshot 2023-08-19 at 08 56 01

### DIFF
--- a/packages/support/docs/09-blade-components/02-link.md
+++ b/packages/support/docs/09-blade-components/02-link.md
@@ -27,9 +27,13 @@ By default, a link's underlying HTML tag is `<a>`. You can change it to be a `<b
 
 ## Setting the size of a link
 
-By default, the size of a link is "medium". You can make it "small", "large", "extra large" or "extra extra large" by using the `size` attribute:
+By default, the size of a link is "medium". You can make it "extra small", "small", "large" or "extra large" by using the `size` attribute:
 
 ```blade
+<x-filament::link size="xs">
+    New user
+</x-filament::link>
+
 <x-filament::link size="sm">
     New user
 </x-filament::link>
@@ -39,10 +43,6 @@ By default, the size of a link is "medium". You can make it "small", "large", "e
 </x-filament::link>
 
 <x-filament::link size="xl">
-    New user
-</x-filament::link>
-
-<x-filament::link size="2xl">
     New user
 </x-filament::link>
 ```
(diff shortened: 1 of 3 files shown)


danharrin:
I agree with the link text sizing, but not the button sizing. Zep worked very hard to get these balanced.

zepfietje:
Yeah so the tricky thing is that the links sizes (and their icon size) are aligned with the button sizes to make sure actions of both types look good when placed next to each other at the same size...

howdu:
@zepfietje I had a feeling it might not be so straight forward :) How are you testing this scenario?

I've just noticed when testing the icon sizing button vs link, for size lg+xl they are different sizes
- Link has `h-6 w-6` but button svg has `h-5 w-5`

Screenshot 2023-08-19 at 11 05 53

```php
protected function getActions(): array
    {
        return [
            Actions\Action::make('test')
                ->label('New product')
                ->link()
                ->icon('heroicon-o-plus-circle')
                ->size('lg'),
            Actions\Action::make('test2')
                ->label('New product')
                ->icon('heroicon-o-plus-circle')
                ->size('lg'),
        ];
}
```

Maybe this code is redundant?
https://github.com/filamentphp/filament/blob/16393c6bc60b5af08ee6109d9b158c2f7390ff2e/packages/support/resources/views/components/link.blade.php#L25-L28

After removing that code and with my font increase changes they line up ok. 
Screenshot 2023-08-19 at 11 02 07

Screenshot 2023-08-19 at 11 00 09

zepfietje:
> How are you testing this scenario?

Manually, by creating actions of all sizes and displaying them next to each other.

> I've just noticed when testing the icon sizing button vs link, for size lg+xl they are different sizes
> 
> * Link has `h-6 w-6` but button svg has `h-5 w-5`

Ah, good catch! That's something I missed during a final button design refactor.

> Maybe this code is redundant?

Yeah, cleaned it up in https://github.com/filamentphp/filament/commit/3e8ef7d60aa852d007cd17ce3f4264dd3bab8185.

zepfietje:
Could you share more details about the specific case where you expect the link font size to be larger, @howdu?

howdu:
> Could you share more details about the specific case where you expect the link font size to be larger, @howdu?

When you need to link a heading that's larger font size than other links.

Screenshot 2023-08-22 at 15 07 02

zepfietje:
Hmm, I'm not too sure about that case.
If one of those links would be a button, you'd expect the button and link to be the same font-size, because it would look weird otherwise.

I'm going to close this PR for now, but feel free to provide additional context on what component that heading is part of for example.

Issue #7827, opened 2023-08-16
Action button wrong background when disabled


ilpav12:
### Package

filament/actions

### Package Version

v3.0.21

### Laravel Version

v10.18

### PHP Version

PHP 8.2

### Problem description

When applying the `->disabled()` function to one of the default Actions or a custom CreateAction the resulting button cannot be clicked and does not trigger the pointer cursor as expected, but it still present the default attributes. In fact, the opacity stays full and when hovering the background color still changes.
With table actions this behavior is working correctly as shown here:
![image](https://github.com/filamentphp/filament/assets/35311720/01717228-b86c-4494-8b0f-8664dc11ea0f)

### Expected behavior

The button opacity should be set to 70% with no background color change on hover.

### Steps to reproduce

On any ListRecords page add `->disabled()` to the default CreateAction

### Reproduction repository

Easy to reproduce

PR #7905, opened 2023-08-19
Fixing button disabled styling

ilpav12:

Fix for issue #7827.
This is a naive fix to the styling that adds classes containing `:hover` only when the button component is not disabled and adds `opacity-70` class otherwise.

### DIFF
--- a/packages/support/resources/views/components/button/index.blade.php
+++ b/packages/support/resources/views/components/button/index.blade.php
@@ -63,20 +63,32 @@
                 [
                     'fi-btn-outlined ring-1',
                     match ($color) {
-                        'gray' => 'text-gray-950 ring-gray-300 hover:bg-gray-400/10 focus:ring-gray-400/40 dark:text-white dark:ring-gray-700',
-                        default => 'text-custom-600 ring-custom-600 hover:bg-custom-400/10 dark:text-custom-400 dark:ring-custom-500',
+                        'gray' => [
+                            'text-gray-950 ring-gray-300 focus:ring-gray-400/40 dark:text-white dark:ring-gray-700',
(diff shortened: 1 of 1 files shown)


packages/support/resources/views/components/button/index.blade.php
```diff
@@ -63,20 +63,32 @@
                 [
                     'fi-btn-outlined ring-1',
                     match ($color) {
```
danharrin:
```suggestion
                    ...match ($color) {
```

zepfietje:
Can we revert this PR, @danharrin?
I think this is caused by the Livewire issue that has recently been fixed, as the button already had disabled styling defined at the top.

danharrin:
@zepfietje can you please test and verify that is the case before I revert this

zepfietje:
I don't know if @ilpav12 encountered the issue when using a button or link, as he didn't provide a reproduction repository (proves you should always provide one, even for this "simple" bug). So I can't test his issue.

This PR did fix the issue for links, that wouldn't have been fixed by Livewire.
I did clean up the changes from this PR though, as it could be achieved in a much simpler way: https://github.com/filamentphp/filament/commit/15e7a0f427f137c32e9420a703eb6d642d4c88e7.

No further action needed.

zepfietje:
@danharrin, this PR also broke the outlined button hover background.
I've fixed this in https://github.com/filamentphp/filament/commit/eb0116f7dee0f50d22b12d4df9cca76eac618b18, so you might have to tag a new release.

ilpav12:
@zepfietje thank you for the fix and for the time to improve my naive solution

PR #7916, opened 2023-08-19
Fix: overflow in select filter indicator with a large number of options

qrasa1:
If there are multiple options selected in a select filter, the filter indicator will overflow as seen here:
Before
After removing `shrink-0` and `whitespace-nowrap` from the badge component it will look like this:
After

I don't know if this will have a large effect on the style where the badge component is used, but the demo app looked fine.
Maybe you guys have a different/better approach to this.

### DIFF
--- a/packages/support/resources/views/components/badge.blade.php
+++ b/packages/support/resources/views/components/badge.blade.php
@@ -25,7 +25,7 @@
     {{
         $attributes
             ->class([
-                'fi-badge flex shrink-0 items-center justify-center gap-x-1 whitespace-nowrap rounded-md  text-xs font-medium ring-1 ring-inset',
+                'fi-badge flex items-center justify-center gap-x-1 rounded-md text-xs font-medium ring-1 ring-inset',
                 match ($size) {
                     'xs' => 'px-0.5 min-w-[theme(spacing.4)] tracking-tighter',
                     'sm' => 'px-1.5 min-w-[theme(spacing.5)] py-0.5 tracking-tight',


zepfietje:
@danharrin, `shrink-0` was added recently to fix another issue.

danharrin:
@zepfietje what issue? I only think it needs to be applied for icons

zepfietje:
Nope, not just for icons: https://github.com/filamentphp/filament/pull/7799. @danharrin

danharrin:
I dont understand why flexbox would shrink padding...

zepfietje:
I don't know, but it does apparently. Is it really required to remove it to fix the issue in this PR?

Also, I feel like the badge content should be truncated instead of wrapped.

danharrin:
I feel like truncating would not be a good approach in this case. @qrasa1 why was shrink removed in the first place?

qrasa1:
@danharrin I tried keeping it, but nothing changed.

zepfietje:
> I feel like truncating would not be a good approach in this case.

I think it would be, since the indicator is like a preview of the filter. If there's an enormous amount of tags, you can always view all tags by opening the filter instead. In my opinion, a badge should basically never grow vertically as it's essentially an inline element,

zepfietje:
> > I feel like truncating would not be a good approach in this case.
> 
> I think it would be, since the indicator is like a preview of the filter. If there's an enormous amount of tags, you can always view all tags by opening the filter instead. In my opinion, a badge should basically never grow vertically as it's essentially an inline element.

zepfietje:
> I feel like truncating would not be a good approach in this case.

I think it would be, since the indicator is like a preview of the filter. If there's an enormous amount of tags, you can always view all tags by opening the filter instead. In my opinion, a badge should basically never grow vertically as it's essentially an inline element.

danharrin:
Ok I agree, truncate with ... is a good approach then

PR #7917, opened 2023-08-19
Update 02-sending-notifications.md

WatheqAlshowaiter:
changes docs:

I don't know how to change the icon as follows

```
<AutoScreenshot name="notifications/icon" alt="Notification with icon" version="3.x" />
```

I only upload the right image for the docs..


qrasa1:
The image is correct according to the code since the `->success()` method is used, the icon gets overwritten by a ✅ one. Removing this method will solve the problem.

danharrin:
We need the AutoScreenshot, the screenshot should be fixed within /docs-assets and not replaced

PR #7934, opened 2023-08-20
[Docs] Add example of injecting $get in validation closure rules doc.

cheesegrits:
This gets asked often enough in the Discord, probably worth adding it as an example

### DIFF
--- a/packages/forms/docs/05-validation.md
+++ b/packages/forms/docs/05-validation.md
@@ -481,6 +481,20 @@ TextInput::make('slug')->rules([
 ])
 ```
 
+You may also inject a $get callable to your closure rules, if you need to reference other fields on your form:
+
+```php
+TextInput::make('slug')->rules([
+    function (Get $get) {
+        return function (string $attribute, $value, Closure $fail) use ($get) {
+            if ($get('other_field') === 'foo' && $value !== 'bar') {
+                $fail("The {$attribute} is invalid.");
+            }
+        };
+    },
+])
+```
+
 ## Validation attributes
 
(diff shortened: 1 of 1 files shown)


packages/forms/docs/05-validation.md
```diff
@@ -481,6 +481,20 @@ TextInput::make('slug')->rules([
...
+TextInput::make('slug')->rules([
+    function (Get $get) {
+        return function (string $attribute, $value, Closure $fail) use ($get) {
+            if ($get('other_field') === 'foo' && $value !== 'bar') {
+                $fail("The {$attribute} is invalid.");
+            }
+        };
+    },
+])
+```
```
danharrin:
```suggestion
You may inject utilities like `$get` into your custom rules, for example if you need to reference other field values in your form:

```php
use Closure;
use Filament\Forms\Get;

TextInput::make('slug')->rules([
    fn (Get $get): Closure => function (string $attribute, $value, Closure $fail) use ($get) {
        if ($get('other_field') === 'foo' && $value !== 'bar') {
            $fail("The {$attribute} is invalid.");
        }
    },
])
```
```

File: packages/forms/docs/05-validation.md [hunk 2]
```diff
@@ -481,6 +481,20 @@ TextInput::make('slug')->rules([
 ])
 ```
 
+You may inject utilities like `$get` into your custom rules, for example if you need to reference other field values in your form:
```
danharrin:
```suggestion
You may [inject utilities](advanced#form-component-utility-injection) like [`$get`](advanced#injecting-the-state-of-another-field) into your custom rules, for example if you need to reference other field values in your form:
```

PR #7941, opened 2023-08-21
Apply rounding to collapsible header

howdu:

Changes requested from @zepfietje 
https://github.com/filamentphp/filament/pull/7882#issuecomment-1684902907

Also updated the builder + repeater to use a `<header>` tag.

### DIFF
--- a/packages/forms/resources/views/components/builder.blade.php
+++ b/packages/forms/resources/views/components/builder.blade.php
@@ -92,7 +92,7 @@
                     >
                         @if ($isReorderableWithDragAndDrop || $isReorderableWithButtons || $hasBlockLabels || $isCloneable || $isDeletable || $isCollapsible)
                             <header
-                                class="flex items-center gap-x-3 px-4 py-2 rounded-t-xl"
+                                class="flex items-center gap-x-3 rounded-t-xl px-4 py-2"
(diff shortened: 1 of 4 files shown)


zepfietje:
Thanks for following up, @howdu!

Can the rounding be achieved by adding an `overflow-hidden` class instead? Would make it easier to maintain and apply different rounding to the box as a whole.

> Also updated the builder + repeater to use a `<header>` tag.

Can you revert this change please? The `<header>` tag is only valid semantically inside a sectioning element. No worries, I had to look this up during the redesign too. 😄

howdu:
@zepfietje I see your point maybe best to revert all of this apart from the attribute reordering? 

Adding `overflow-hidden` could prevent absolute positioned elements from overflowing. Tested with dropdown at the bottom and Choices.js cleverly knows it's going to be hidden so works differently but it's not great UX.

https://github.com/filamentphp/filament/assets/533658/a9aad42d-ac45-4419-aed2-a563f7e1d4a1

zepfietje:
Ah yeah that's the reason. Maybe just apply `overflow-hidden` when collapsed?

danharrin:
Okay if you're planning to revert all of this, you might as well submit a new PR with that `overflow-hidden` only for collapsed sections, as Zep says 👍

PR #7943, opened 2023-08-21
docs: fix filling existing data assertion

trovster:
Use `assertSet()` instead of `assertFormSet()` as the latter does not work with infolist data.


danharrin:
This is wrong though, the test is for the edit page, and you should use assertFormSet() for forms

PR #7948, opened 2023-08-21
docs: fix filling existing data assertion when viewing

trovster:
This fixes the mismatch of `assertSet()` in the code example.


danharrin:
This is wrong though, it's now broken for view forms and infolists don't have a property for data. assertSet() doesn't properly work in either case. I think you're just trying to find infolist testing features that don't exist yet

trovster:
Well, there is a mismatch between the introduction text and the code example.

> To check that the form is filled with the correct data from the database, you may `assertSet()` that the data in the form matches that of the record:

`assertFormSet()` does not work for Infolists but it works with `assertSet()`.

danharrin:
Yeah well the docs are inconsistent there

But I see no reason that assertSet() should work. It checks public properties. Infolists don't use public properties.

trovster:
I have just checked my code and I have `assertSee()` for my Infolist test!

danharrin:
assertSee is not assertSet though, but that does make sense it would work. You could PR in a note about infolists and assertSee if you want

PR #7953, opened 2023-08-21
FIX: php > 8.1 str_replace deprecation in HasComponents panels concern

nathandaly:
This fixes a > PHP 8.1 deprecation where null isn't allowed as 2nd parameter in the replace function.

ErrorException:
```
str_replace(): Passing null to parameter #2 ($replace) of type array|string is deprecated in /var/www/vhost/cms/vendor/laravel/framework/src/Illuminate/Support/Str.php on line 976
```

xdebug
stack error

- [ ] ~~New functionality has been documented or existing documentation has been updated to reflect changes.~~  `Not applicable`

### DIFF
--- a/packages/panels/src/Panel/Concerns/HasComponents.php
+++ b/packages/panels/src/Panel/Concerns/HasComponents.php
@@ -313,7 +313,7 @@ trait HasComponents
                 $variableNamespace = (string) str($variableNamespace)->before('\\');
             }
 
-            if (is_null($variableNamespace)) {
+            if ($variableNamespace === null) {
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


packages/panels/src/Panel/Concerns/HasComponents.php
```diff
@@ -313,6 +313,10 @@ protected function discoverComponents(string $baseClass, array &$register, ?stri
                 $variableNamespace = (string) str($variableNamespace)->before('\\');
             }
 
+            if (is_null($variableNamespace)) {
```
danharrin:
```suggestion
            if ($variableNamespace === null) {
```

PR #7956, opened 2023-08-21
Add `filament-panels` directory to content array in tailwind config

thecrazybob:

I noticed CSS for Tailwind classes applied inside the logo file needed to be generated. So, I added the directory where the [logo file is supposed to be stored](https://filamentphp.com/docs/3.x/panels/themes#changing-the-logo) in the content array for the custom theme stub (since I am using one).

I tested it with the default theme, and the issue still exists. Could you help me add it to the default theme as well?


danharrin:
We cannot add it for the default theme as it is pre-compiled. Also, I am not a huge fan of adding this directory to the content since it kinda encourages people to publish other views. I would much rather add docs that say you should add the single file to your Tailwind config if you override the logo with custom classes.

PR #7962, opened 2023-08-21
Do not serialize string value of state for select column

### LINKED ISSUE #7470: `int` backed `enum` throws during render in `SelectColumn` for `Table` since the enum value can't be converted to `string`
### Package

filament/filament

### Package Version

v3.0.7

### Laravel Version

10.17.1

### Livewire Version

v3.0.0-beta.7

### PHP Version

8.2.8

## Issue
When using an enum that is backed by int and used in a table, there is code in `select-column.blade.php` that tries to stringify the enum value, but int enums do not support `__toString()` unfortunately, so we might need to check explicitly for this case and manually convert the enum's value to string.

See code below, exception fires when visiting: `admin/users`

## Exception
`Object of class App\Enums\UserProviderType could not be converted to string`

CleanShot 2023-08-04 at 14 14 33

CleanShot 2023-08-04 at 14 13 40

## Code
`UserProviderType.php`:
```php
namespace App\Enums;

use Filament\Support\Contracts\HasLabel;

enum UserProviderType: int implements HasLabel
{
    case APPLE = 1;
    case GOOGLE = 2;

    public function getLabel(): ?string
    {
        return match($this) {
            self::APPLE => 'Apple',
            self::GOOGLE => 'Google'
        };
    }
}
```

`UserResource.php`:
```php
    {
        return $table
            ->columns([
                Tables\Columns\SelectColumn::make('provider_type')
                                    ->label('Provider')
                                    ->options(UserProviderType::class)
                                    ->sortable(),
            ]);
}
```

`User.php:
```php

protected $casts = [
        'email_verified_at' => 'datetime',
        'provider_type' => UserProviderType::class,
    ];
```

`select-column.blade.php`:
```php
<input
        type="hidden"
        value="{{ str($state)->replace('"', '\\"') }}"
        x-ref="newState"
    />
```

### Expected behavior

Since int backed enums can easily be translated to string, we can use reflection or other mechanism to detect and provide the string value?

`select-column.blade.php`
```php
@php
    $canSelectPlaceholder = $canSelectPlaceholder();
    $isDisabled = $isDisabled();
    $rawState = $getState();
    
    try {
        $state = str($rawState);
    } catch (\Throwable $e) {
        $reflectedEnum = new ReflectionEnum($rawState);
        $reflectedBackingType = $reflectedEnum->getBackingType();

        if ($reflectedBackingType === 'int') {
            $state = str($rawState->getValue());
        } else {
           // re-throw
        }
    }
@endphp
```

Oddly enough, it works on the edit page, just not on the select column in the table. The enum is hooked up roughly the same way, but it's the `Select` component, not the `SelectColumn`:

```php
Select::make('provider_type')
                    ->label('Provider')
                    ->options(UserProviderType::class)
                    ->searchable(),
```

So maybe bringing some code over from `Select` into `SelectColumn` :) 

### Steps to reproduce

1. Create int backed enum
2. Hook up to form table select as direct class usage
3. Visit url to invoke table render
4. Exception fires

### Reproduction repository

https://github.com/bradleybernard/FilamentEnumBug

### Relevant log output

_No response_

bradleybernard:

I don't have an expansive app that uses Filament today, so hard to test this one!

No docs changes needed for bug fix.

No visual changes needed for bug fix.

Issue: https://github.com/filamentphp/filament/issues/7470
Background on changes: https://github.com/filamentphp/filament/commit/1a77740db7f6b16046af2e74e4718688aa3acd3c

### DIFF
--- a/packages/tables/resources/views/columns/select-column.blade.php
+++ b/packages/tables/resources/views/columns/select-column.blade.php
@@ -71,10 +71,9 @@
             x-on:change="
                 isLoading = true
 
-                let newState = $event.target.value;
-                if (newState === '') {
-                    newState = null;
-                }
+                let newState = $event.target.value !== '' ?
+                    $event.target.value :
+                    null
 
                 const response = await $wire.updateTableColumnState(
                     name,
--- a/packages/tables/resources/views/columns/select-column.blade.php
+++ b/packages/tables/resources/views/columns/select-column.blade.php
@@ -76,14 +76,12 @@
             x-on:change="
                 isLoading = true
 
-                let newState = $event.target.value !== '' ?
-                    $event.target.value :
-                    null
-
                 const response = await $wire.updateTableColumnState(
                     name,
                     recordKey,
-                    newState,
+                    $event.target.value !== '' ?
+                        $event.target.value :
+                        null,
                 )
 
                 error = response?.error ?? undefined
--- a/packages/tables/resources/views/columns/select-column.blade.php
+++ b/packages/tables/resources/views/columns/select-column.blade.php
@@ -79,9 +79,7 @@
                 const response = await $wire.updateTableColumnState(
                     name,
                     recordKey,
-                    $event.target.value !== '' ?
-                        $event.target.value :
-                        null,
+                    $event.target.value,
                 )
 
                 error = response?.error ?? undefined
--- a/packages/tables/resources/views/columns/select-column.blade.php
+++ b/packages/tables/resources/views/columns/select-column.blade.php
@@ -48,7 +48,7 @@
 >
     <input
         type="hidden"
-        value="{{ str($state)->replace('"', '\\"') }}"
+        value="{{ $state }}"
         x-ref="newState"
     />
 
--- a/packages/tables/resources/views/columns/select-column.blade.php
+++ b/packages/tables/resources/views/columns/select-column.blade.php
@@ -1,7 +1,12 @@
 @php
     $canSelectPlaceholder = $canSelectPlaceholder();
     $isDisabled = $isDisabled();
+
     $state = $getState();
+    if ($state instanceof \BackedEnum) {
+        $state = $state->val;
+    }
+    $state = (string) str(strval($state))->replace('"', '\\"');
 @endphp
 
 <div
@@ -14,7 +19,7 @@
 
         recordKey: @js($recordKey),
 
-        state: @js($state ?? ''),
+        state: @js($state),
     }"
     x-init="
         Livewire.hook('commit', ({ component, commit, succeed, fail, respond }) => {
--- a/packages/tables/resources/views/columns/select-column.blade.php
+++ b/packages/tables/resources/views/columns/select-column.blade.php
@@ -6,7 +6,7 @@
     if ($state instanceof \BackedEnum) {
         $state = $state->value;
     }
-    $state = (string) str(strval($state))->replace('"', '\\"');
+    $state = strval($state);
 @endphp
 
 <div
@@ -53,7 +53,7 @@
 >
     <input
         type="hidden"
-        value="{{ $state }}"
+        value="{{ str($state)->replace('"', '\\"') }}"
         x-ref="newState"
     />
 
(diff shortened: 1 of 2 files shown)


bradleybernard:
Argh, this doesn't work with choosing the default: "Select an option", since we try to set the value on a model that has a cast to backed enum, but the value sent is an empty string ""... hmmmmmm


https://github.com/filamentphp/filament/assets/1431995/08011466-f3ad-4c3f-a924-f6df3844ea8c

packages/tables/resources/views/columns/select-column.blade.php
```diff
@@ -48,7 +48,7 @@
 >
     <input
         type="hidden"
-        value="{{ str($state)->replace('"', '\\"') }}"
+        value="{{ $state }}"
```
danharrin:
```suggestion
        value="{{ str(strval($state))->replace('"', '\\"') }}"
```

danharrin: How does this perform?

bradleybernard: Tried that out, did not work unfortunately. 

I had to add a bit more logic to the on change to satisfy the following:

`select-column.blade.php`
```php
const response = await $wire.updateTableColumnState(
```

which ultimately calls `CanUpdateState`
```php
$record->setAttribute($columnName, $state);
```

which fails in Laravel core from:

```php
if ($this->isEnumCastable($key)) {
    $this->setEnumCastableAttribute($key, $value);

    return $this;
}
```

because we were setting `""` to the enum value and it was trying to cast that to the backed enum. 
Instead we'll have to send `null` as a value.

Before:
```php
$record->setAttribute($columnName, $state /* "" */);
```

After:
```php
$record->setAttribute($columnName, $state /* null */);
```

Let me know what you think! This fixes it for me :)

bradleybernard:
Now should all work, fixed setting a Model's attribute that is casted to a backed enum. See comment: https://github.com/filamentphp/filament/pull/7962#discussion_r1300617144

Video demo:
https://github.com/filamentphp/filament/assets/1431995/daedd143-c7b4-4eee-bfcc-d14e8148c5cc

bradleybernard:
Need to double check we are okay with dropping the str replace that was added 6 months ago for a reactivity fix: https://github.com/filamentphp/filament/commit/1a77740db7f6b16046af2e74e4718688aa3acd3c

cc @danharrin

danharrin:
The str replace needs to be there if the value contains `"`, which needs to be escaped otherwise it interrupts the `"`s on the attribute `value=""`

gotzsys:
I am having a similar problem using a string type enumerator.

```
<?php

namespace App\Enums;

use Filament\Support\Contracts\HasLabel;

enum WaitlistStatus: string implements HasLabel
{
  case WAITING = 'waiting';
  case CALLED = 'called';

  public function getLabel(): ?string
  {
    return match ($this) {
      self::WAITING => 'Na fila',
      self::CALLED => 'Chamado',
    };
  }
}
```
```
@php
    $canSelectPlaceholder = $canSelectPlaceholder();
    $isDisabled = $isDisabled();
    $state = $getState();
    dd($state);
@endphp
```
```
App\Enums\WaitlistStatus {#3167 ▼ // vendor/filament/tables/src/../resources/views/columns/select-column.blade.php
  +name: "WAITING"
  +value: "waiting"
}
```

Returns the error:
Object of class App\Enums\WaitlistStatus could not be converted to string

Making this change works... but obviously, the solution must be more complex.
image

drbyte:
I've just run into this issue as well. 
@bradleybernard 's proposed changes solve it for me. 
In my case the enum is an int, so no concerns about escaping quotation marks.

The "fix" I came up with before finding this PR and corresponding issue was kinda heavy, and probably has limitations ... it simply inspects for a contract-abiding enum right at the top of the file:

```diff
@php
    $canSelectPlaceholder = $canSelectPlaceholder();
    $isDisabled = $isDisabled();
    $state = $getState();
+   if (is_a($state, \Filament\Support\Contracts\HasLabel::class)) $state = $state->value;
@endphp
```

danharrin:
Thanks

bradleybernard:
Thanks Dan!! Looks good :)

PR #1997, opened 2022-03-24
Fix file upload not loading correctly in secondary tabs and collapsed sections

jacksleight:
Fixes https://github.com/laravel-filament/filament/issues/1967

Replaces `x-show` with a custom set of classes that hide the tabs without using `display: none`. This is the workaround proposed in the FilePond issue (https://github.com/pqina/filepond/issues/819#issuecomment-1077313749).

Issue #1967, opened 2022-03-22
FileUpload - Delete file button not working


martinmildner:
### Package

filament/forms

### Package Version

v2.10.33

### Laravel Version

v8.x

### Livewire Version

v2.10.4

### PHP Version

PHP 8.1

### Bug description

Currently, uploaded files/images can no longer be deleted when using **FileUpload::make()**. FilePond's Delete button is not clickable anymore. Using **->removeUploadedFileButtonPosition('bottom')** works fine, if the field is high enough (e.g. with an image). **->panelLayout(null)** also seems to fix the problem (thanks for the hint to **saade**).

### Steps to reproduce

Create a FileUpload field:

`FileUpload::make('image')
->disk('public')
->directory('xxx/images')
->image() 
`

Upload an image. Save. Try to delete the file.
The problem was confirmed by another user on Discord.

martinmildner:
@jacksleight yes, correct - it's on a secondary tab. Working correctly on the first tab.

jacksleight:
FYI: https://github.com/pqina/filepond/issues/819

PR #7997, opened 2023-08-22
Added animated collapsable sections

thethunderturner:

## Description
As the title suggests I added a feature where the user can add a `->animated()` getter and the section would collapse/expand in 200ms time

https://github.com/filamentphp/filament/assets/64212185/d22e23b5-edd0-4cb9-ae65-0b0c0e1cbe85

## How?
This was mainly done by utilizing x-show from alpineJS. I had a look on how the other getters were implemented and I followed the same structure.

### DIFF
--- a/packages/support/resources/views/components/section.blade.php
+++ b/packages/support/resources/views/components/section.blade.php
@@ -135,12 +135,6 @@
     @endif
 
     <div
-        @if ($hasAnimation)
-            x-cloak
-            x-show="isCollapsed"
-            x-collapse.duration.200ms
-        @endif
-
         @if ($collapsible)
             x-bind:aria-expanded="(! isCollapsed).toString()"
             @if ($collapsed)
@@ -157,10 +151,15 @@
             'border-t border-gray-200 dark:border-white/10' => $hasHeader && (! $aside),
             'md:order-first' => $contentBefore,
         ])
+        @if ($hasAnimation)
+            x-cloak
+            x-show="!isCollapsed"
+            x-collapse.duration.200ms
+        @endif
     >
         <div
             @class([
-                // 'fi-section-content',
+                'fi-section-content',
                 'rounded-xl bg-white shadow-sm ring-1 ring-gray-950/5 dark:bg-gray-900 dark:ring-white/10' => $aside,
                 match ($compact) {
                     true => 'p-4',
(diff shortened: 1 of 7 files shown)


demobiel:
would be nice if the the animation time could be set, something like ->animated(300) which will make the animation time 300ms

danharrin:
> would be nice if the the animation time could be set, something like ->animated(300) which will make the animation time 300ms

I disagree, that would be a job for a "theme"

demobiel:
true... I am just not into themes yet :)

zepfietje:
I don't think we need a separate API to control animation to be honest.
Also, I'm not too sure if this improves or degrades the UX.

packages/support/resources/views/components/section.blade.php
```diff
@@ -139,14 +140,22 @@ class="-m-2"
             @if ($collapsed)
                 x-cloak
             @endif
-            x-bind:class="{ 'invisible h-0 border-none': isCollapsed }"
+            @if (!$hasAnimation){
+                x-bind:class="{ 'invisible h-0 border-none': isCollapsed }"
```
zepfietje, line 144: There's a reason we use this approach, right @danharrin?
Won't this break animated collapsible sections?

danharrin: Break in what way?

zepfietje: Possibly we didn't use `x-show` for a reason to do with form components state? Since `x-show` is used for this new animated collapsible section, I'm afraid that must break something?

danharrin: Can we animate the height property instead maybe

thethunderturner: > Can we animate the height property instead maybe

Well, technically you can do it with `max-height`, but in order to do so you need to set a hardcoded value. I didn't find any issues with `x-show` though and its used in the groups.php so why not go with `x-show`? At least I couldn't find anything that would break.

zepfietje: I don't know the exact reason, but there is a reason for the current approach.

danharrin:
Reintroduces #1967, which was fixed in #1997 by the invisible class

thethunderturner:
@danharrin Can you explain why it was closed? How does the FileUpload relate to this?

danharrin:
Please read that issue, and the PR

PR #7999, opened 2023-08-22
Fix issue with null filter values in query string on refresh

### LINKED ISSUE #7129: Query string includes all available filters causing an error on refresh

archilex:

In light of Caleb's [comment](https://github.com/livewire/livewire/pull/6188#issuecomment-1686492498) that he won't be able to address this in Livewire until later, this PR implements the work done by @KennedyTedesco and @jvitasek into Filament until LW can be fixed. 

Test suite is passing and I have confirmed this fixes the issue with selectFilters, trashedFilter, ternaryFilter, and filters with forms with textFields on refresh. It also works when filters have `->default()` values.

Closes #7129

### DIFF
--- a/packages/panels/src/Resources/Pages/ListRecords.php
+++ b/packages/panels/src/Resources/Pages/ListRecords.php
@@ -71,6 +71,10 @@ class ListRecords extends Page implements Forms\Contracts\HasForms, Tables\Contr
         ) {
             $this->activeTab = array_key_first($tabs);
         }
+
+        if ($this->tableFilters) {
+            $this->replaceNullValues($this->tableFilters);
+        }
     }
 
     public function getBreadcrumb(): ?string
@@ -348,5 +352,20 @@ class ListRecords extends Page implements Forms\Contracts\HasForms, Tables\Contr
         return (string) str($key)
             ->replace(['_', '-'], ' ')
             ->ucfirst();
+    }
+
+    protected function replaceNullValues(array &$data): void
+    {
+        foreach ($data as &$value) {
+            if (is_array($value)) {
+                $this->replaceNullValues($value);
+            } elseif ($value === 'null') {
+                $value = null;
+            } elseif ($value === 'false') {
+                $value = false;
+            } elseif ($value === 'true') {
+                $value = true;
+            }
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


archilex:
Hold on...I just realized this only fixes the issue for Panels, and not with the stand alone Table Builder.

archilex:
Fixed for Table Builder

Note: `->replaceNullValues()` needs to be before `$this->getTableFiltersForm()->fill($this->tableFilters)`.

KennedyTedesco:
I've been using this fix for a couple of days now. The query string still looks awful, with `null`, `false` etc, but at least dont breaks the filtering system.

LGTM, until LW fixes this in a more robust way.

Another option for the method name is: `normalizeBooleanAndNullValues()`

danharrin:
Can we unset null values from the array maybe

KennedyTedesco:
Yes, that's another option. I didn't test it, but I believe it would work just fine.

KennedyTedesco:
"Null" values can be unset, but "true" values need to be normalized to true. So, I think we can simply normalize everything, like this pull request does, streamlining the process into a single flow.

archilex:
I agree with @KennedyTedesco. This solution, until LW is hopefully updated, seems to be the most streamlined. I also updated the method name as he suggested since `normalizeBooleanAndNullValues()` is a better indicator of the actual function.

PR #8003, opened 2023-08-22
Fix console error with select.js when closing modals

archilex:
When a searchable select field is used inside a modal, when the modal is closed the following console error is thrown: `TypeError: Cannot read properties of null (reading 'clearStore')`. It seems Alpine/LW's watcher isn't being destroyed causing this issue. This PR adds an early return to prevent this.


danharrin:
This is an underlying issue with Livewire that I need to pinpoint and report

jdion84:
this still happens

jdion84:
@danharrin

danharrin:
I know. If you want, submit a failing test PR to Livewire. I have not had time yet.

PR #8005, opened 2023-08-23
add the ability to exclude items in a repeater before saving

meditto:
## Description

This pull request introduces new functionality that enhances the repeater component. It adds the capability to exclude specific items by utilizing the `mutateRelationshipDataBeforeCreateUsing` method.

### DIFF
--- a/packages/forms/src/Components/Repeater.php
+++ b/packages/forms/src/Components/Repeater.php
@@ -734,9 +734,8 @@ class Repeater extends Field implements Contracts\CanConcealComponents
 
                 $itemData = $component->mutateRelationshipDataBeforeCreate($itemData);
 
-                // This is helpful in case we want to ignore an item; we can simply return an empty array,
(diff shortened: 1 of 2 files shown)


danharrin:
I think returning `null` is better for this

meditto:
I added `null` as a valid return type to ignore the item

packages/forms/src/Components/Repeater.php
```diff
@@ -895,7 +900,7 @@ public function mutateRelationshipDataBeforeCreateUsing(?Closure $callback): sta
      * @param  array<array<string, mixed>>  $data
      * @return array<array<string, mixed>>
```
danharrin:
```suggestion
     * @return array<array<string, mixed>> | null
```

danharrin:
Thanks

meditto:
Thank you @danharrin

PR #8008, opened 2023-08-23
bugfix: add fillFormWithDataAndCallHooks so it is the same as in Edit…

demobiel:
…Record Page

I am working on the filament translatable plugin, and in order to get the TenantProfilePage translatable, I think this change is needed. 

However I am not sure if this should become a Trait or not... I was a bit afraid by the "never call this method, but copy it" :)

### DIFF
--- a/packages/panels/src/Pages/Tenancy/EditTenantProfile.php
+++ b/packages/panels/src/Pages/Tenancy/EditTenantProfile.php
@@ -80,6 +80,16 @@ abstract class EditTenantProfile extends Page
     {
         $data = $this->tenant->attributesToArray();
 
+        $this->fillFormWithDataAndCallHooks($data);
+    }
+
+    /**
+     * @internal Never override or call this method. If you completely override `fillForm()`, copy the contents of this method into your override.
+     *
+     * @param  array<string, mixed>  $data
+     */
+    protected function fillFormWithDataAndCallHooks(array $data): void
+    {
         $this->callHook('beforeFill');
 
         $data = $this->mutateFormDataBeforeFill($data);


danharrin:
The only reason you'd want this method extracted is if you wanted to override it, and we specifically tell you not to do that as you will expose your app to breaking changes, hence the message 😉

danharrin:
The reason why it isn't done for the EditTenantProfile is because we don't support that page to be translatable yet. Parts of the plugin require a resource to be registered

demobiel:
I understand :) but as I am trying to get the spatie translatable plugin to work with the EditTenantProfile. 

I am trying to create a new Translatable concern for this plugin for the EditTenantProfileClass, based on the Translatable concern for the EditRecord class.

The EditRecord class uses the Trait "HasTranslatableFormWithExistingRecordData", which looks to me could be used for the EditTenantProfile trait I am trying to create (as the tenant profile page also has a translatable form with existing record data).

And the only reason (I think) that stops me from using this Trait is this missing function "fillFormWithDataAndCallHooks" in the EditTenantProfile page.

This is also linked to discussion: #8007, where indeed I am trying to create this new feature :)

danharrin:
We also need to remove references to the resource's defined locales, right?

danharrin:
Also need docs on how to use this

We should probably create a separate Translatable trait for EditTenantProfile instead of reusing EditRecord's one

demobiel:
Hi Dan

I will delete this pull request, as indeed as I have to create a separate Translatable Trait I won't need this extra function anymore.

Regarding the translatable trait for EditTenantProfile -> I will create a separate PR for this once it is finished (I honestly have no idea how much work this is :) )

However I have a more profound question on the translatable plugin : why is it so tight to resources at the moment, and is not all the logic based on the forms plugin? Wouldn't this make it more flexible (for example custom pages / livewire components etc).

An example is that the translatableAttributes are asked from the active record at the moment, however this could be rewritten to the form component I think. Making a method that analyses the form including and subcomponents with relations. But maybe I am underestimating the work I am proposing (to myself :) )

danharrin:
It's tied to resources so you dont have to define the locales more than once per resource

PR #8012, opened 2023-08-23
Allow JS asset to register with type=module

ernoferrari:
This pull request allows to register js asset with type="module" attribute for script tag.

### DIFF
--- a/packages/support/src/Assets/Js.php
+++ b/packages/support/src/Assets/Js.php
@@ -13,6 +13,8 @@ class Js extends Asset
 
     protected bool $isCore = false;
 
+    protected bool $isModule = false;
+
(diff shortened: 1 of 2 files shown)


danharrin:
How about `type('module')` instead? And then it is more flexible and you can pass any string

ernoferrari:
As it is specified here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type

If you pass to type attribute some other value than module or JavaScript MIME type, then script content is required. Now it's not possible to set content to script tag so other values are not possible anyway.

If I rename module method to type, then I should add content method as well for clarity. What do you think about that?

danharrin:
Fair enough, thanks!

PR #8018, opened 2023-08-23
Truncate long badge label

qrasa1:
I wanted to make the indicator truncate instead of wrapping as discussed in the last PR regarding text indicator overflowing when multiple options are selected.

Why did I use the `grid` class instead of the `flex` class or something else? Well, it turns out that if you have more than one `span` parent with the `flex` class, the `truncate` class won't work (text will overflow).

This will work: 
```html
<div class="flex">
   <span class="truncate">
       *long text here*
   </span>
</div>
``` 
This will not work:
```html
<div class="flex">
  <div class="flex">
     <span class="truncate">
         *long text here*
     </span>
  </div>
</div>
``` 
You may obviously have a better solution than this.

Here are the after images:
Computer:
Screenshot_2023-08-23-09-59-03-534_com android chrome
Mobile:
Screenshot_2023-08-23-10-00-20-561_com android chrome

### DIFF
--- a/packages/support/resources/views/components/badge.blade.php
+++ b/packages/support/resources/views/components/badge.blade.php
@@ -57,17 +57,9 @@
         <x-filament::icon :icon="$icon" :class="$iconClasses" />
     @endif
 
-    @if ($truncated)
-        <div class="grid">
-            <span class="truncate">
-                {{ $slot }}
-            </span>
-        </div>
-    @else
-        <span>
-            {{ $slot }}
-        </span>
-    @endif
+    <span class="truncate">
+        {{ $slot }}
+    </span>
 
     @if ($isDeletable)
         <button
--- a/packages/support/resources/views/components/badge.blade.php
+++ b/packages/support/resources/views/components/badge.blade.php
@@ -27,7 +27,6 @@
         $attributes
             ->class([
                 'fi-badge flex items-center justify-center gap-x-1 rounded-md text-xs font-medium ring-1 ring-inset',
-                'shrink-0 whitespace-nowrap' => ! $truncated,
                 match ($size) {
                     'xs' => 'px-0.5 min-w-[theme(spacing.4)] tracking-tighter',
                     'sm' => 'px-1.5 min-w-[theme(spacing.5)] py-0.5 tracking-tight',
--- a/packages/support/resources/views/components/badge.blade.php
+++ b/packages/support/resources/views/components/badge.blade.php
@@ -5,6 +5,7 @@
     'icon' => null,
     'iconPosition' => 'before',
     'size' => 'md',
+    'truncated' => false,
 ])
 
 @php
@@ -26,6 +27,7 @@
         $attributes
             ->class([
                 'fi-badge flex items-center justify-center gap-x-1 rounded-md text-xs font-medium ring-1 ring-inset',
+                'shrink-0 whitespace-nowrap' => !$truncated,
                 match ($size) {
                     'xs' => 'px-0.5 min-w-[theme(spacing.4)] tracking-tighter',
                     'sm' => 'px-1.5 min-w-[theme(spacing.5)] py-0.5 tracking-tight',
@@ -55,9 +57,17 @@
         <x-filament::icon :icon="$icon" :class="$iconClasses" />
     @endif
 
-    <span>
-        {{ $slot }}
-    </span>
+    @if($truncated)
+        <div class="grid">
+            <span class="truncate">
+                {{ $slot }}
+            </span>
+        </div>
+    @else
+        <span>
+            {{ $slot }}
+        </span>
+    @endif
 
     @if ($isDeletable)
         <button
--- a/packages/support/resources/views/components/badge.blade.php
+++ b/packages/support/resources/views/components/badge.blade.php
@@ -25,7 +25,7 @@
     {{
         $attributes
             ->class([
-                'fi-badge flex items-center justify-center gap-x-1 rounded-md text-xs font-medium ring-1 ring-inset',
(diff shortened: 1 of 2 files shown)


packages/support/resources/views/components/badge.blade.php
```diff
@@ -55,9 +57,17 @@
...
+        <div class="grid">
+            <span class="truncate">
+                {{ $slot }}
+            </span>
+        </div>
+    @else
+        <span>
+            {{ $slot }}
+        </span>
+    @endif
```
zepfietje:
```suggestion
    <span class="truncate">
        {{ $slot }}
    </span>
```

packages/tables/resources/views/components/filters/indicators.blade.php
```diff
@@ -14,7 +14,7 @@ class="whitespace-nowrap text-sm font-medium leading-6 text-gray-700 dark:text-g
 
         <div class="flex flex-wrap gap-1.5">
             @foreach ($indicators as $wireClickHandler => $label)
-                <x-filament::badge>
+                <x-filament::badge truncated>
```
zepfietje:
```suggestion
                <x-filament::badge>
```

packages/support/resources/views/components/badge.blade.php
```diff
@@ -26,6 +27,7 @@
         $attributes
             ->class([
                 'fi-badge flex items-center justify-center gap-x-1 rounded-md text-xs font-medium ring-1 ring-inset',
+                'shrink-0 whitespace-nowrap' => ! $truncated,
```
zepfietje:
```suggestion
```

packages/support/resources/views/components/badge.blade.php
```diff
@@ -5,6 +5,7 @@
     'icon' => null,
     'iconPosition' => 'before',
     'size' => 'md',
+    'truncated' => false,
```
zepfietje:
```suggestion
```

packages/support/resources/views/components/badge.blade.php
qrasa1: You'd think that this will work, except it won't because this component is used inside other components that have the `flex` class

zepfietje:
No need to have a separate API for this IMO.
Have a look at the simplified implementation if you like, @qrasa1. 🙂
Anyway, thanks for creating the PR!

qrasa1:
@zepfietje good job, but I made it like that for a reason, take a look at this image, idk if this is the desired behavior:
Screenshot_2023-08-24-10-09-04-410_com android chrome

zepfietje:
Hmm, is that on a regular table?

qrasa1:
The screenshot was taken from the orders table within the demo app

zepfietje:
Fixed it in https://github.com/filamentphp/filament/commit/de2825e6157bcb09082aa2a3459a75d0065cc3c1. Thanks.

zepfietje:
Okay seems like this is still broken in button badges (e.g. the database notifications trigger). Opened an issue for it: https://github.com/filamentphp/filament/issues/8038.

Feel free to let me know if you happen to know a way to fix it already, @qrasa1. 😄

qrasa1:
The only fix I could think of is adding the `truncated` prop back:

```blade
@props([
    'color' => 'primary',
    'closeIcon' => null,
    'deleteButton' => null,
    'icon' => null,
    'iconPosition' => 'before',
    'size' => 'md',
    'truncated' => false,
])
``` 
```blade
    $attributes
            ->class([
                'fi-badge flex items-center justify-center gap-x-1 rounded-md text-xs font-medium ring-1 ring-inset',
                'shrink-0' => ! $truncated,
``` 
```blade
    <span class="grid">
        <span @class(['truncate' => $truncated])>
            {{ $slot }}
        </span>
    </span>
``` 
`indicators.blade.php`:
```blade
  <x-filament::badge truncated>
``` 
Here are before and after images:
Mobile
Mobile after

zepfietje:
I'd fixed the issue in your screenshot already in the commit I linked to above.
Remaining issue is button badges, but I'll fix it.

zepfietje:
https://github.com/filamentphp/filament/commit/b60c120771f80d59ef44ab00262ce64204c9b2a5

qrasa1:
It still not fixed on mobile though, you could see this by just making your browsers screen smaller on the orders table from the demo app. 
It can be fixed by removing `max-` from `flex max-w-max` on line `98` in `filament/packages/tables/resources/views/columns/text-column.blade.php`. 
Idk if this will have some side effects.
Another thing is that the badge will truncate in where it does not have enough space. Like in blog/posts/1/edit for example:
Tagbadge
Idk if this should be considered an issue, because badges shouldn't be that long anyways.

zepfietje:
Yeah I just noticed that it's still broken on mobile.
I applied your fix `w-max` as I think it shouldn't break anything.

Regarding the tags input, that's fine as we decided tags should be truncated when there's not enough space.

All good now (hopefully).
Thanks again for your help!

faizananwerali:
I think this should be optional because all my badges text are broken. By default it should not be enabled, only when it is needed.
 
Screenshot 2023-08-27 at 5 17 36 PM

faizananwerali:
Does anyone know how to not use truncate for badges without overwriting this view file?

qrasa1:
If you still have not figured out a way:

If you are using the panel builder you should first follow this: https://filamentphp.com/docs/3.x/panels/themes#creating-a-custom-theme
After that you can add this to `theme.css` .if you are not using the panel builder add it to `resources/css/app.css`: 
```css
.fi-badge span > span {
overflow: visible;
} 
``` /
If this does not fix it, you can try other css properties.

zepfietje:
@faizananwerali, please provide the exact Filament version you're using as I'm not sure if my final fix has been released yet.
I'm pretty sure your issue should be fixed on 3.x already.

PR #8019, opened 2023-08-23
Bug: Global Search Crash when searched in Id

timsinakiran:
Global search crashes if 'id' is one of the search criteria. This worked fine in V2, and this bug is introduced as we are comparing lower() for case insensitive search.

Before this fix:
The search crashes with an SQL error:
function lower(bigint) does not exist. This is tested with PostGres.

After this fix:
The search works normally and returns the result normally.

### DIFF
--- a/packages/panels/src/Resources/Resource.php
+++ b/packages/panels/src/Resources/Resource.php
@@ -653,7 +653,7 @@ abstract class Resource
                     };
 
                     $caseAwareSearchColumn = $isForcedCaseInsensitive ?
-                        new Expression("lower({$searchColumn})") :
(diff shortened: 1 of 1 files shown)


danharrin:
I don't believe you tested this. `$searchColumn` is a string column name, not an integer. You should just use `id::text` as the column name or something.

timsinakiran:
My bad. I am actually manually copying and pasting here in git to make PR (still trying to figure out the ways). I was able to fix the issue by checking the name $searchColumn === 'id' instead of is_int($searchColumn). But that fixes the issue only if if the primary key 'id' is of type int, and not other.

danharrin:
So have you tried `id::text` as the column name

timsinakiran:
Yes, using id:text works in getGloballySearchableAttributes().

PR #8020, opened 2023-08-23
Bug: Global Search Crash when searching with ID

timsinakiran:
The previous PR was a mistake.

This resolves the issue of searching in non-string field such as ID according to the way you have suggested.

### DIFF
--- a/packages/panels/src/Resources/Resource.php
+++ b/packages/panels/src/Resources/Resource.php
@@ -678,7 +678,12 @@ abstract class Resource
                             "%{$search}%",
                         );
                     },
-                    function ($query) use ($isForcedCaseInsensitive, $whereClause, $searchAttribute, $search) {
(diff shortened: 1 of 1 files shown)


danharrin:
We deliberately don't do this for all columns as it can cause performance issues, the user can do this if they wish when they define the search columns

timsinakiran:
Can I asd this in documentation then?

danharrin:
Good idea

timsinakiran:
@danharrin Though we can do this in global search, it will still not work in normal table search. Right now, almost all table searches are broken in a case where data has numbers are unique identifiers such as phone numbers where search is required. Also it would be tedious to manually cast every int to text everywhere we are searching even if we develop a way to cast in Table columns.

I recommend a type check in the core and cast to text if it's integer (and if the driver is postgres). My previous solution only checked driver , but if we can add a type check for int|integer, I believe this issue will be resolved without any performance hit.

PR #8026, opened 2023-08-23
Fix support for BelongsToThrough relationships in Select/SelectFilter

dchatry:
See examples below.
No need for docs changes, works as documented.
No visual changes.

This PR fixes support for [staudenmeir/belongs-to-through](https://github.com/staudenmeir/belongs-to-through) relationships which used to be (at least partially) supported in V2.

Changes in `packages/forms/src/Components/Select.php` handle the proper key retrieval for this type of relationship for select population and searching.

A `distinct()` was added to the relationship query in `packages/tables/src/Filters/SelectFilter.php` to de-duplicate values used to build indicator labels.

Before:
image

After:
image

### DIFF
--- a/packages/tables/src/Filters/SelectFilter.php
+++ b/packages/tables/src/Filters/SelectFilter.php
@@ -49,6 +49,7 @@ class SelectFilter extends BaseFilter
                     $relationshipQuery = $filter->getRelationshipQuery();
 
                     $labels = $relationshipQuery
+                        ->distinct()
                         ->whereKey($state['values'])
                         ->pluck($relationshipQuery->qualifyColumn($filter->getRelationshipTitleAttribute()))
                         ->all();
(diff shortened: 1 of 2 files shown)


packages/tables/src/Filters/SelectFilter.php
```diff
@@ -49,6 +49,7 @@ protected function setUp(): void
                     $relationshipQuery = $filter->getRelationshipQuery();
 
                     $labels = $relationshipQuery
+                        ->distinct()
```
danharrin: This is going to reintroduce a problem we fixed before, we probably need to check the relationship type first

dchatry: Thanks for the feedback, I added a check before applying the distinct().

danharrin:
Thanks!

PR #8028, opened 2023-08-23
Automatically detect boolean cast attribute in icon column and entry

### LINKED ISSUE #7108: Automatically detect boolean cast attribute in icon column and entry

joshbonnick:
This PR resolves #7108.

Column being cast to a `boolean` is decided by checking the `$casts` array for `bool` or `boolean` - Laravel docs only mention `boolean`, however `bool` is also checked when casting a Model's attributes. If the column is not cast to a `boolean` it also checks if the state is actually a boolean, thus supporting a database column with `true`/`false` without a cast.

The tests this PR adds are slightly more verbose than their surrounding tests as I could not find a method to call and assert on a specific column's method. 

Also to note, the default value for the `isBoolean` property has been set to `null`. This is to check whether the value has already been set by the user before using the cast. Ensuring the user's choice is the priority over the cast.

- ~~Visual changes are explained in the PR description using a screenshot/recording of before and after.~~


joshbonnick:
The updated `isBoolean` method of `Filament\Infolists\Components\IconEntry` is untested, if you could point me in the right direction to add tests for Infolists, then I can add it.

danharrin:

danharrin:
Thanks! I have removed the state check, since this could have some performance impact by resolving the state multiple times per cell, and we do try to avoid that. I've also added in some caching to ensure that the casts are not checked more than once per column.

PR #8030, opened 2023-08-23
Fix multi-tenancy HasManyThrough record creation

### LINKED ISSUE #8029: Multi-Tenancy HasManyThrough Tenant Relationship
### Package

filament/filament

### Package Version

v3.0.23

### Laravel Version

v10.19.0

### Livewire Version

v3.0.0-beta.9

### PHP Version

PHP 8.2.9

### Problem description

When using Multi-Tenancy with a Resource where the Tenant has a HasManyThrough relationship to the Resource, the following error message is encountered upon trying to create a new record for the given Resource:

> Call to undefined method Illuminate\Database\Eloquent\Relations\HasManyThrough::save()

This error is originating from the following method: https://github.com/filamentphp/filament/blob/24fe227d68da7db6981c569926b55aadbe5f1a46/packages/panels/src/Resources/Pages/CreateRecord.php#L184

This method is attempting to call the `save` method on a HasManyThrough relationship, which is not supported by the Laravel framework.

This seems similar to the problem described in #7513 

### Expected behavior

In a Multi-Tenant environment, I should be able to create records for Resources which the Tenant has a HasManyThrough relationship to.

### Steps to reproduce

In a multi-tenant environment:

1. Create a Resource where the Tenant has a HasManyThrough relationship to the Resource
2. Attempt to create a new record for the resource
3. Observe error

### Reproduction repository

https://github.com/kylekanderson/filament-tenancy-hasmanythrough

### Relevant log output

_No response_

kylekanderson:

Fixes #8029

### DIFF
--- a/packages/panels/src/Resources/Pages/CreateRecord.php
+++ b/packages/panels/src/Resources/Pages/CreateRecord.php
@@ -11,6 +11,7 @@ use Filament\Pages\Concerns\InteractsWithFormActions;
 use Filament\Support\Exceptions\Halt;
 use Illuminate\Contracts\Support\Htmlable;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\HasManyThrough;
 use Illuminate\Support\Str;
 
 /**
@@ -183,7 +184,14 @@ class CreateRecord extends Page
 
     protected function associateRecordWithTenant(Model $record, Model $tenant): Model
     {
-        return static::getResource()::getTenantRelationship($tenant)->save($record);
+        $relationship = static::getResource()::getTenantRelationship($tenant);
+
+        if (!$relationship instanceof HasManyThrough) {
+            return $relationship->save($record);
+        }
+        
+        $record->save();
+        return $record;
     }
 
     /**
--- a/packages/panels/src/Resources/Pages/CreateRecord.php
+++ b/packages/panels/src/Resources/Pages/CreateRecord.php
@@ -186,12 +186,12 @@ class CreateRecord extends Page
     {
         $relationship = static::getResource()::getTenantRelationship($tenant);
 
-        if (!$relationship instanceof HasManyThrough) {
-            return $relationship->save($record);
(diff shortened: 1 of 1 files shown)


packages/panels/src/Resources/Pages/CreateRecord.php
```diff
@@ -183,7 +184,14 @@ protected function handleRecordCreation(array $data): Model
...
     {
-        return static::getResource()::getTenantRelationship($tenant)->save($record);
+        $relationship = static::getResource()::getTenantRelationship($tenant);
+
+        if (!$relationship instanceof HasManyThrough) {
+            return $relationship->save($record);
+        }
+        
+        $record->save();
+        return $record;
```
daveyvdgit: Would it be better to switch up the condinitionals and to return `$record->save()`:

```suggestion
        $relationship = static::getResource()::getTenantRelationship($tenant);

        if ($relationship instanceof HasManyThrough) {
            return $record->save();
        }

        return $relationship->save($record);
```

I think this would improve readability.

With a shorthand:

```suggestion
        $relationship = static::getResource()::getTenantRelationship($tenant);
        
        return ($relationship instanceof HasManyThrough) ? $record->save() : $relationship->save($record);
```

To Lower the lines of code and to improve readability.

kylekanderson: @Daveyvdweide - thanks for having a look. I appreciate your suggestions. And agreed, inverting the conditionals improves readability. I pushed up a change that incorporates that. Looking through the rest of the codebase, it seems that traditional if/else conditionals are typically used in place of ternary operators, so I stuck with that to best align with the rest of the code.

I wasn't able to directly commit your proposed change however, because your code was setup to return `$record->save()` (which will return a boolean) rather than `$record` (an object) in the case where `$relationship` is an instance of `HasManyThrough`.

Thanks again for the review - much appreciated!

danharrin:
Thanks

rubenlopezgea-at-coodex:
This fix passed away with Filament 4.
The new approach leads to the same issue that existed

PR #8065, opened 2023-08-25
[3.x-Notifications] Added authGuard variable to use custom auth guard provider

xreal:
Added a static property $authGuard to the Notifications Livewire Classes. This allows for custom Auth Guards, for example i use the default laravel auth guard for the frontend but a different one for the backend (where i use the notification filament package).


faizananwerali:
You're breaking changes, I've used multi-guard and it is working fine for me. Also, no doc file has been updated so those who have used multi-guard can follow changes.

danharrin:
This definitely isn't breaking as far as I can see @faizananwerali

PR #8071, opened 2023-08-25
2.x minor documentation change: substitute attribute name in custom validation rule message

judgej:

Just a non-functional minor documentation change that may save people a little time and frustration. I can do this for V3 once I've tested it  there. I'm using the documented technique on V2 and it works for me.

### DIFF
--- a/packages/forms/docs/05-validation.md
+++ b/packages/forms/docs/05-validation.md
@@ -472,7 +472,7 @@ TextInput::make('slug')->rules([
     function () {
         return function (string $attribute, $value, Closure $fail) {
             if ($value === 'foo') {
-                $fail("The :attribute is invalid.");
+                $fail('The :attribute is invalid.');
             }
         };
     },
--- a/packages/forms/docs/05-validation.md
+++ b/packages/forms/docs/05-validation.md
@@ -472,7 +472,7 @@ TextInput::make('slug')->rules([
     function () {
         return function (string $attribute, $value, Closure $fail) {
             if ($value === 'foo') {
-                $fail("The {$attribute} is invalid.");
+                $fail("The :attribute is invalid.");
             }
         };
     },


judgej:
An example of how I am using it:

```php
            // Password field in a form.
            Password::make('current_password')
                ->label('Current Password')
                ->validationAttribute('Current Password')
                ->password()
                ->required()
                ->minLength(1)
                ->maxLength(100)
                ->rules([
                    // Custom rule confirms checks if the user has entered their existing password correctly.
                    function () {
                        return function (string $attribute, $value, Closure $fail) {
                            if (! Hash::check($value, auth()->user()->password)) {
                                // Using {$attribute} here shows "current_password", lcase with an underscore
                                $fail("The :attribute field does not match the current password on your account.");
                            }
                        };
                    },
                ])
                ->autocomplete('new-password')
                ->lazy(),
```

image

(`Phpsa\FilamentPasswordReveal\Password` extends `TextField` and just adds some reveal functionality.)

judgej:
Not convinced the what-the-diff bot has added anything useful here, appart from a bunch of sales spiel.

packages/forms/docs/05-validation.md
```diff
@@ -472,7 +472,7 @@ TextInput::make('slug')->rules([
     function () {
         return function (string $attribute, $value, Closure $fail) {
             if ($value === 'foo') {
-                $fail("The {$attribute} is invalid.");
+                $fail("The :attribute is invalid.");
```
danharrin:
```suggestion
                $fail('The :attribute is invalid.');
```

judgej:
Yes, it would make sense for the message could then use single quotes instead of double for this. Saw that as a "suggested change" via email, but no idea where it is on this site.

danharrin:
I committed it before I merged the PR

Screenshot 2023-08-29 at 22 41 56

PR #8076, opened 2023-08-25
Fix issue where a collapsed section content was still filling space

lamberttraccard:

## Before
CleanShot 2023-08-25 at 21 31 20@2x

## After
CleanShot 2023-08-25 at 21 32 11@2x

### DIFF
--- a/packages/panels/src/Resources/Pages/CreateRecord.php
+++ b/packages/panels/src/Resources/Pages/CreateRecord.php
@@ -283,14 +283,14 @@ class CreateRecord extends Page
     {
         $resource = static::getResource();
 
-        if ($resource::hasPage('edit') && $resource::canEdit($this->getRecord())) {
-            return $resource::getUrl('edit', ['record' => $this->getRecord(), ...$this->getRedirectUrlParameters()]);
-        }
-
         if ($resource::hasPage('view') && $resource::canView($this->getRecord())) {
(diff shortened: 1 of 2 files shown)


zepfietje:
Thanks for the PR, @lamberttraccard.
However, I do think there's a reason for this approach.
I didn't implement it initially so turning to @danharrin again (this is what we talked about on that other PR last week too).

danharrin:
The `h-0` should cover it. There is a reason why it's invisible but I can't remember

lamberttraccard:
I can't tell you either 😅. Just that in the current state something is off. 🙂

lamberttraccard:
Maybe the invisible was for a transition effect, but making it hidden doesn't seem to change anything, expect the current issue

danharrin:
Note to future self: don't merge until you find the reason it was invisible

danharrin:
Found the reason @lamberttraccard @zepfietje - #1967, to do with using a FileUpload inside it

Implemented in #1997

zepfietje:
Right, and we can't use `x-if` instead of `x-show` because we'd lose field state?

danharrin:
I haven't tried it, but I feel like it wouldnt be very performant

lamberttraccard:
Thanks, I will have a look this weekend if I can find a workaround and why the problem occur sometimes

danharrin:
We're going to need an explanation as to why `invisible h-0` doesn't work, and which browsers can reproduce this on 👍 Once you reattempt a different fix, you can submit a new PR

PR #8080, opened 2023-08-26
Fix placeholder for Infolist TextEntry

miguilimzero:

The `placeholder()` is currently not working for `Infolists/TextEntry`. 

This happens because the `Arr::wrap()` function by Laravel converts `null` to an empty array `[]`, instead of wrapping `null` to `[null]`. With an empty array, there is no element for the foreach and therefore the placeholder is not being used.

This pull request fixes this by adding an if statement handling the null case.


danharrin:
Dup #7439

PR #8093, opened 2023-08-27
Validate MakePage command page name input.

Nuhel:

Last day while i was creating a page, I mistakenly provided a name that contained white space.
Which generated class with the white space.

### DIFF
--- a/packages/panels/src/Commands/MakePageCommand.php
+++ b/packages/panels/src/Commands/MakePageCommand.php
@@ -23,21 +23,25 @@ class MakePageCommand extends Command
     public function handle(): int
     {
 
-        $page = (string) str(
+        $page = (string)str(
             $this->argument('name') ??
             text(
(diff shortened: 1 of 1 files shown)


Nuhel:
@danharrin 

Check this, if u like this approach, i will merge this with other commands to. 

We can also use below regex to accept name with namespace 

>  ^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*(\\[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)*$

danharrin:
Is there no built-in validation rule in Laravel we can use? Like `alpha_num`

danharrin:
Or does that not allow `/` or something

danharrin:
Hmm, I think it's okay if we leave out validation for now, thanks though

PR #8096, opened 2023-08-27
Update simple.blade.php

Jaydee-31:
Fixed footer rendering at the side of simple page


zepfietje:
Please don't remove the PR description template and provide screenshots as explained there.
I'm pretty sure this completely breaks the current footer design, so it really needs explanation of the issue and fix.

PR #8102, opened 2023-08-27
Support nested navigation groups

Nuhel:

This is under review.

image


zepfietje:
As @danharrin commented on the discussion, the indentation is off.
The `Report` group is at the same level as the `My site` item, right?

Nuhel:
> As @danharrin commented on the discussion, the indentation is off. The `Report` group is at the same level as the `My site` item, right?

Yes.

danharrin:
So how do we fix that visually? This was the first problem we encountered when implementing this feature in v3.

Nuhel:
> So how do we fix that visually? This was the first problem we encountered when implementing this feature in v3.

I think the tree tail wasn't matching, so i did some tweak.

Before
image


After

Bullet in group label middle
image

Bullet in group label top
image

danharrin:
The collapsible group arrows on the right need to line up, at the moment, they have more and more right padding

Nuhel:
> The collapsible group arrows on the right need to line up, at the moment, they have more and more right padding

image

zepfietje:
Left padding is still inconsistent.

Nuhel:
> Left padding is still inconsistent.

image

zepfietje:
How does this work with highlighting the current page for deeply nested items?

glennjacobs:
For reference, here's an example of how we might look to use this functionality. This is from our old Livewire admin panel.

It might give some ideas for the UI. 

image

Nuhel:
> How does this work with highlighting the current page for deeply nested items?

image


Like this?

incon:
cp-editing-v4

Side menu example from statamic.

zepfietje:
Thanks for providing your example, @glennjacobs!

First thing that came to mind is that it looks much like the Statamic menu (which @danharrin and I discussed when exploring the v3 nested navigation design).
Thanks for confirming my thoughts, @incon! 😄

Honestly I think the kind of menu you guys provided could work.
I'm not a fan of the nested design by @Nuhel.
Nothing personal, but it just has too many edge cases when mixing groups and items with and without icons, and highlighting nested items which aren't fullwidth etc.

So let's go for the nested design @glennjacobs and @incon posted above.
Essentially, this means all navigation items are still horizontally aligned and fullwidth.
We may have to limit making the first-level group collapsible when the nested one is too?
Also, when setting icons on all items and (nested) groups, the UX could be confusing.
Eventually it's up to the devs implementing it in their projects though, but let's run through these edge cases thoroughly once the new nested navigation design has been implemented.

That said, could you rework this PR to match our new direction, @Nuhel?

Nuhel:
> Thanks for providing your example, @glennjacobs!
> 
> First thing that came to mind is that it looks much like the Statamic menu (which @danharrin and I discussed when exploring the v3 nested navigation design). Thanks for confirming my thoughts, @incon! 😄
> 
> Honestly I think the kind of menu you guys provided could work. I'm not a fan of the nested design by @Nuhel. Nothing personal, but it just has too many edge cases when mixing groups and items with and without icons, and highlighting nested items which aren't fullwidth etc.
> 
> So let's go for the nested design @glennjacobs and @incon posted above. Essentially, this means all navigation items are still horizontally aligned and fullwidth. We may have to limit making the first-level group collapsible when the nested one is too? Also, when setting icons on all items and (nested) groups, the UX could be confusing. Eventually it's up to the devs implementing it in their projects though, but let's run through these edge cases thoroughly once the new nested navigation design has been implemented.
> 
> That said, could you rework this PR to match our new direction, @Nuhel?

Yes the design provided by @glennjacobs and @incon looks much more clean. 
I thought ,i was limited with the current design pattern, that's why i did not think to redesign the whole thing.

Allow me some time to rework this pr. @zepfietje

danharrin:
I just want to pipe in here, I'd actually like to go down the route of using `$navigationSubGroup` / `getNavigationSubGroup()` on resources and pages INSTEAD of allowing real group nesting (unlimited nesting), and then forcing this no-icon design on any sub-grouped items. Nesting would be limited to 3 levels, and then no further.

killjin:
image
relatively compact

The three-level menu can meet most needs.

zepfietje:
Please read the entire discussion above, @killjin.
We've come to the conclusion that your design doesn't work well.

binaryfire:
Here's an example already implemented in a Filament project: https://docs.larafa.st/installation. Very clean and adds a lot to the UX.

Repo: https://github.com/better-futures-studio/docs-panel

zepfietje:
Looks like the existing navigation groups, @binaryfire?

binaryfire:
@zepfietje Sorry, must have been hallucinating 🥴 Could have sworn I saw nested items.

meeftah:
any update on this?

trollfalgar:
Any update?

zepfietje:
@glennjacobs @incon @Nuhel @danharrin, I just realized the proposed design is much like GitHub's settings: https://github.com/settings.

Essentially, the first-level groups are blocks of navigation items, while nested groups function much like navigation items, but are collapsible.

saade:
> @glennjacobs @incon @Nuhel @danharrin, I just realized the proposed design is much like GitHub's settings: https://github.com/settings.
> 
> Essentially, the first-level groups are blocks of navigation items, while nested groups function much like navigation items, but are collapsible.

I agree with this, should the new api be something like?

```php
NavigationItem::make()
    ->....
    ->items([
        NavigationItem::make()
            ->...
    ])
```

I can help with this one if you desire

danharrin:
We're going with a simple $navigationSubGroup and limiting it to 1 level deep

saade:
> We're going with a simple $navigationSubGroup and limiting it to 1 level deep

Sweet! Should cover most of the cases, some questions:

1. `NavigationSubGroup` can have a url, or only be collapsible? And it can have an icon if its inside a `NavigationGroup` that has an icon?
2. A `NavigationItem` that's inside a `NavigationSubGroup` can have icons? Or should it be only a label? (like Github)

Are you guys already working on this rn or need help?

zepfietje:
I'll take over this PR to ensure all edge cases are covered. 👍

danharrin:
Replaced by #9737

PR #8103, opened 2023-08-27
Support overriding table filters

wivaku:
Use case: have Resource table with filters, 
and in RelationManager reuse the table, but without / with different filters.

E.g.

```
#UserResource/RelationManagers/PostsRelationManager.php

{
    return PostResource::table($table)->filters(); // reused PostResource table, but without filters.
}
```

### DIFF
--- a/packages/tables/src/Table/Concerns/HasActions.php
+++ b/packages/tables/src/Table/Concerns/HasActions.php
@@ -34,6 +34,21 @@ trait HasActions
      */
     public function actions(array | ActionGroup $actions, ActionsPosition | string | Closure | null $position = null): static
     {
+        $this->actions = [];
+        $this->pushActions($actions);
+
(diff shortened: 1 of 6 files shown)


wivaku:
related to: https://filamentphp.com/docs/3.x/panels/resources/relation-managers#overriding-shared-configuration-on-the-relation-manager

zepfietje:
`filters([])` should work. Correct me if I'm wrong and I'll reopen this PR.
Edit: sorry, I was confused by the changes in this PR. Will leave this to Dan to review and comment on the implementation.

wivaku:
Indeed, the current implementation allows you to add filters, but not to overwrite them.

danharrin:
I've fixed the default behaviour so it overwrites the same as v2, but you can also `->pushFilters()`

PR #8105, opened 2023-08-27
Update 03-select.md

sakanjo:
`titleAttribute` doesn't have a default value which will trigger `Expected 3 arguments. Found 2.`,

we can edit [Select.php#L674](https://github.com/filamentphp/filament/blob/3.x/packages/forms/src/Components/Select.php#L674) `titleAttribute` parameter to be null by default but it's just a bad idea.


sakanjo:
You're lying what-the-diff, the correct summary is:

## PR Summary

* Set titleAttribute value to null on the documentation page.

danharrin:
I've given it a default value now

PR #8109, opened 2023-08-27
Add getQuery function

sakanjo:

Pros:

1. I'm working on a big project where i've many auth models `user`, `admin`, `patient`, `doctor`, `nurse` and many more, instead of defining the inverse relation for each model this PR makes it seamless to work with relationships.

2. Doesn't override lines `314-316` which will break multi-tenancy.

Cons:

1. On edit page it breaks and need to specify qualified key name `protected static ?string $recordRouteKeyName = 'analyses.id';` and this problem exists even when using with `getEloquentQuery`.

---

For hasMany relations

Before:
```php
// Patient -> hasMany -> Prescription
public static function getEloquentQuery(): Builder
{
    return parent::getEloquentQuery()
      ->whereBelongsTo(auth('patient')->id());
}
```

After
```php
// Cleaner
public static function getQuery(): Builder {
    return auth('patient)->user()
      ->prescriptions()
      ->getQuery();
}
```

---

For hasManyThrough relations

Before:
```php
// Patient -> hasMany -> Analyses -> through -> AnalysisRequest
public static function getEloquentQuery(): Builder
{
    return parent::getEloquentQuery()
      ->whereRelation('analysis_request', 'patient_id', auth('patient')->id());
}
```

After
```php
// Cleaner
public static function getQuery(): Builder {
    return auth('patient')->user()
        ->analyses()
        ->getQuery();
}
```

### DIFF
--- a/packages/panels/src/Resources/Resource.php
+++ b/packages/panels/src/Resources/Resource.php
@@ -303,7 +303,8 @@ abstract class Resource
         return static::$breadcrumb ?? Str::headline(static::getPluralModelLabel());
     }
 
-    public static function getQuery(): Builder {
+    public static function getQuery(): Builder
+    {
         return static::getModel()::query();
     }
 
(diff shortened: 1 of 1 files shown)


danharrin:
Unfortunately I don't think there is a good name for this method, so I suggest you just add the tenancy scope to your override

PR #8118, opened 2023-08-28
Document update, Actions\ActionGroup::placement does not exist. It should be changed to 'dropdownPlacement'.

tanshiqi:
Replace `placement` with `dropdownPlacement`.


zepfietje:
Thanks!

PR #8128, opened 2023-08-28
Shades registration

zepfietje:

---

## Todo

- [ ] Move `get_color_css_variables` calls as close as possible to element containing the `custom-x` CSS classes. This may remove the need for conditional shade registration.


zepfietje:
I'm still cleaning up the aliases (see PR description todo list), but I think it'd be good if you could have an initial look at the implementation and provide any feedback, @danharrin.

packages/forms/resources/views/components/field-wrapper/hint.blade.php
```diff
@@ -16,9 +15,11 @@
                 },
             ])
             ->style([
-                \Filament\Support\get_color_css_variables($color, shades: [400, 500, 600]),
-            ])
-    }}
+                \Filament\Support\get_color_css_variables(
+                    $color,
+                    shades: [400, 500, 600]),
+                    alias: 'forms::components.field-wrapper.hint'
```
danharrin:
```suggestion
                    alias: 'forms::components.field-wrapper.hint',
```

packages/support/src/Colors/ColorManager.php
```diff
@@ -11,6 +11,21 @@ class ColorManager
      */
     protected array $colors = [];
 
+    /**
+     * @var array<int>
+     */
+    protected array $overridingShades = [];
```
danharrin, line 17: Maybe use `overridden` instead of `overriding` everywhere

danharrin (changes requested):

packages/support/src/Colors/ColorManager.php
zepfietje: `overridden` would be confusing as it signifies the shades that will be overridden instead of the shades that override. Thoughts?

zepfietje:
@danharrin, I've removed most aliases (except `badge`) to prevent potential breaking changes by moving around aliases, as I want to move them as close as possible to the actual element containing the `custom-` color classes.

I just need `badge` for now, and will add others as I need them (in PRs).

packages/support/src/Colors/ColorManager.php
danharrin: Ok ok fair

danharrin:
Technically, I'm fine with BCs until we add documentation, so even if you kept the rest of the hooks, it doesnt matter really

zepfietje:
Yeah, though since I wanted to update them anyway, I chose to remove them for now. 👍

zepfietje:
CI passing now. Ready for final review, @danharrin.

PR #8130, opened 2023-08-28
Truncate parameter for badge in text column

margarizaldi:

### Add `truncate` parameter to `badge()` method.
I'm not sure why badge should use truncate class as it will be very hard to see the content if it is truncated, and unfortunately there's no tooltip to help. In the other side, I think the reason behind using badge is simply "for a better appearance", then truncated text of badge is really nonsense for that reason. At this point, I really want to make truncate class is not used by default.

However, this change keeps the default badge to use truncate class so it won't change the previous behavior.

Before:
image
_Wait, also check the icons become smaller there, we can fix it later._

After (with `truncate: false` parameter):
```php

TextColumn::make('status')
    ->badge(truncate: false),
// ...
```
image

### DIFF
--- a/packages/support/resources/views/components/badge.blade.php
+++ b/packages/support/resources/views/components/badge.blade.php
@@ -5,6 +5,7 @@
     'icon' => null,
     'iconPosition' => 'before',
     'size' => 'md',
+    'canTruncateBadge' => true,
 ])
 
 @php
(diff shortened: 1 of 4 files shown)


zepfietje:
No need for an API to control this.
Badges being truncated in table columns was a regression which has been fixed on 3.x already. Please let us know if you run into any issues still.

PR #8133, opened 2023-08-28
add configuration support for Page title without Action name as prefix

wivaku:
Triggered by https://github.com/filamentphp/filament/discussions/8108.

It is possible to manually override a specific Page title to only show the record title (e.g. "John Doe") instead of getting the Action name prefixed (e.g. "View John Doe").

This PR adds configuration support so that the Action name is not shown in the title of any View / Edit page.
```
    return $panel
        // ...
        ->titleAction(showAction: false);
```

### DIFF
--- a/packages/panels/docs/09-configuration.md
+++ b/packages/panels/docs/09-configuration.md
@@ -115,7 +115,7 @@ public function panel(Panel $panel): Panel
 
 ## Hiding the action in title for View / Edit pages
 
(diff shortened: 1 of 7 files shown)


danharrin:
Sorry, but I'm not a fan. You could use a trait or an abstract class to do this if you want.

wivaku:
Wanted to give something back to the community, but up to you, you're the boss. ;-)

The reason for it was: for _all_ my projects I prefer this behaviour for _all_ my Resources.
Therefore I feel config is the best place, instead of adding trait to each Resource in each project.

PR #8134, opened 2023-08-28
Prevent hidden navigation items from being rendered

### LINKED ISSUE #8117: Navigation item hidden in sidebar but not hidden in top bar navigation
### Package

filament/filament

### Package Version

3.0-stable

### Laravel Version

^10.10

### Livewire Version

3.12.3

### PHP Version

8.1.6

### Problem description

I use hidden function to hide custom navigation items, its work in sidebar mode but not work in top navigation bar

### Expected behavior

the navigation item shouldn't show either as in sidebar

### Steps to reproduce

1. add some custom navigation and merged into one group in panel provider configuration
2. add hidden method to the custom navigation item
3. the navigation item hidden in sidebar
4. add method top navigation in panel configuration
5. the navigation item still appear in top navigation

### Reproduction repository

https://github.com/mohdfuady/filament-top-navigation

### Relevant log output

_No response_

sakanjo:
This PR adds `$item->isVisible()` to the top navigation menu items making sure it's visible when it should be.

### DIFF
--- a/packages/panels/src/Panel/Concerns/HasNavigation.php
+++ b/packages/panels/src/Panel/Concerns/HasNavigation.php
@@ -158,6 +158,7 @@ trait HasNavigation
 
                 return $sort;
             })
+            ->filter(fn (NavigationGroup $group): bool => $group->getItems()->some->isVisible())
             ->all();
     }
 
--- a/packages/panels/src/Panel/Concerns/HasNavigation.php
+++ b/packages/panels/src/Panel/Concerns/HasNavigation.php
@@ -158,7 +158,7 @@ trait HasNavigation
 
                 return $sort;
             })
-            ->filter(fn (NavigationGroup $group): bool => $group->getItems()->some->isVisible())
+            ->filter(fn (NavigationGroup $group): bool => collect($group->getItems())->some->isVisible())
             ->all();
     }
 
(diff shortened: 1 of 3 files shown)


sakanjo:
@zepfietje while this PR fixes #8117, there's another point to take care of and that's when there are not items

image

Is it a normal behavior ?

zepfietje:
When there's no visible items in a group, the entire group should be hidden. Could you double-check the sidebar to make sure that's also the current behavior there?

sakanjo:
That should do it!

packages/panels/src/Panel/Concerns/HasNavigation.php
```diff
@@ -158,6 +158,7 @@ public function getNavigation(): array
 
                 return $sort;
             })
+            ->filter(fn (NavigationGroup $group): bool => collect($group->getItems())->some->isVisible())
```
zepfietje: Not sure if it has any performance impact, but it might be better to filter first, before all items get sorted?

sakanjo: It does, thanks for the catch.

danharrin:
Do we need both checks, the filter and the @if? Doesn't the filter cover both? And can we share this code with the sidebar somehow?

sakanjo:
Yup no need for the `@if` directive after changing the ordering.

danharrin:
Thanks

PR #8139, opened 2023-08-28
Make the home redirect route optional

bpudenz:
This PR keeps existing functionality, but allows for the ability to skip the creation of redirect routes similar to other routes.

This is primarily helpful on secondary panels that have a dynamic element to them. Since the service provider that establishes the panel routes runs before any existing route files, a dynamic panel path takes precedence and cannot be overridden.

### DIFF
--- a/packages/panels/routes/web.php
+++ b/packages/panels/routes/web.php
@@ -13,6 +13,7 @@ Route::name('filament.')
             /** @var \Filament\Panel $panel */
             $panelId = $panel->getId();
             $hasTenancy = $panel->hasTenancy();
+            $needsHomeRedirectUrl = $panel->getNeedsHomeRedirectUrl();
             $tenantRoutePrefix = $panel->getTenantRoutePrefix();
             $tenantSlugAttribute = $panel->getTenantSlugAttribute();
             $domains = $panel->getDomains();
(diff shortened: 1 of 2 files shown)


danharrin:
I would advise registering routes in the panel provider

```php
->routes(function () {
   // Route::get()
})
```

Which happens before built-in routes are registered

cpbret:
I hadn't seen that in documentation yet, and that'd work for overwriting specific routes, but in my case, the path being dynamic doesn't allow it to fall through when there is a trailing slash.

On all other paths, it falls through, because {abc}/defined and {abc}/defined-other do not interfere. {abc}/ is the problem

bpudenz:
I hadn't seen that in documentation yet, and that'd work for overwriting specific routes, but in my case, the path being dynamic doesn't allow it to fall through when there is a trailing slash.

On all other paths, it falls through, because {abc}/defined and {abc}/defined-other do not interfere. {abc}/ is the problem

danharrin:
Understood, I just feel like having config options to disable each specific route would become bloat our codebase, and after merging this PR it would be quite difficult to justify dening further PRs that achieve that. Thoughts?

bpudenz:
I understand bloat. For now, I've got a composer script that comments it out and all is well. Thanks Dan

danharrin:
I have an idea, use DI to replace the controller with your own

bpudenz:
If I were overwriting it with my own code, that would be great - but this is conflicting with clockwork and ajax livewire calls. 

I have an AppPanel pointing to path('/') and another ContextPanel that has a dynamic path('{context}/{id}') that is limited by middleware. The conflict is with the dynamic panel that catches when there is a trailing "/". 

I understand this is not an issue many others will have. It's just a small configuration that fixes dynamic routing. As I mentioned, my scripting hack is fine with me and I understand your feedback. Hopefully this makes more sense though.

danharrin:
Hmm, maybe you should use the tenancy feature on your ContextPanel instead of a dynamic path? The "tenant" here is the context

PR #8144, opened 2023-08-28
Provide configuration to SpatieTagsInput to allow original behaviour of suggestions for tags with types

sten:
@danharrin as promised in https://github.com/filamentphp/filament/pull/7374 I added an extra configuration flag to the SpatieTagsInput to be able to overwrite the old default behaviour with the new implementation.

Do you think the naming of the config variable `disableTypeCheckingForSuggestions` is clear?

### DIFF
--- a/packages/spatie-laravel-tags-plugin/src/Forms/Components/SpatieTagsInput.php
+++ b/packages/spatie-laravel-tags-plugin/src/Forms/Components/SpatieTagsInput.php
@@ -27,7 +27,7 @@ class SpatieTagsInput extends TagsInput
             $record->load('tags');
 
             if ($component->isAnyTagTypeAllowed()) {
-                $tags = $record->tags;
+                $tags = $record->getRelationValue('tags');
             } else {
                 $tags = $record->tagsWithType($type);
             }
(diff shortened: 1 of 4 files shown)


danharrin:
How about this API

`SpatieTagsInput::make()` - suggestions from all types
`SpatieTagsInput::make()->type('blog')` - suggestions only from `blog` type
`SpatieTagsInput::make()->type(null)` - suggestions only from `null` type

sten:
@danharrin I implemented your suggested API. Because `$type` is set to `null` on object creation, I added another var to be able to check if `$type` was forcefully set to `null`.

danharrin:
Maybe we could set the default of that prop to be `new AllTypes()` where AllTypes is some empty object that represents that state. You can put classes like this in `Filament\SpatieLaravelTagsPlugin\AllTypes` or something. Then we can do a simple `instanceof AllTypes` check

danharrin:
Also, I think this should be applied to more places than just suggestions.

The actual tags that get loaded into both SpatieTagsInput and SpatieTagsColumn should be scoped to the type(), with identical behaviour. Wdyt?

sten:
@danharrin  Thank you for the feedback! I added a new type as you suggested to be able to set the input, infolist and column to allow any type. 

I had to add a function to sync the tags to the model to avoid creating duplicate tags without a type if the type already existed, because the HasTags trait did not have this functionality.

Let me know what you think.

danharrin:
Thanks @sten, this is exactly what we need. You interpreted my requests really well

PR #8167, opened 2023-08-30
Provide a configuration property to determine if remove uploaded file button is shown 

lucasdotvin:
The idea here is simply to provide a way to hide the uploaded file button in Filepond through their constructor option `allowRemove`.

Developers that wish their file component does not show a remove button (in cases where it is not allowed to change the file, for instance) can call `removable(false)` to do so.

### DIFF
--- a/packages/forms/resources/views/components/file-upload.blade.php
+++ b/packages/forms/resources/views/components/file-upload.blade.php
@@ -22,7 +22,6 @@
         ax-load-src="{{ \Filament\Support\Facades\FilamentAsset::getAlpineComponentSrc('file-upload', 'filament/forms') }}"
         x-data="fileUploadFormComponent({
                     acceptedFileTypes: @js($getAcceptedFileTypes()),
-                    allowRemove: @js($isRemovable()),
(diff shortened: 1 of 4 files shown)


packages/forms/resources/views/components/file-upload.blade.php
```diff
@@ -22,6 +22,7 @@
         ax-load-src="{{ \Filament\Support\Facades\FilamentAsset::getAlpineComponentSrc('file-upload', 'filament/forms') }}"
         x-data="fileUploadFormComponent({
                     acceptedFileTypes: @js($getAcceptedFileTypes()),
+                    allowRemove: @js($isRemovable()),
```
danharrin: Please pass this in as isRemovable and map to allowRemove in the JS file

danharrin (changes requested):
Not sure if this should be isRemovable, we use isDeletable everywhere else

lucasdotvin:
Well pointed! I'll rename the property and map it in the JS file.

danharrin:
Thanks

PR #8176, opened 2023-08-30
Fixed widget infinite loading

### LINKED ISSUE #8175: Chart filters are loading infinitely
### Package

filament/filament

### Package Version

v3.0.37

### Laravel Version

v10.0.21

### Livewire Version

_No response_

### PHP Version

8.2.7

### Problem description

When a filter is created for a widget - it is loading non-stop:

image

image

This indicator never goes away

### Expected behavior

The loading indicator should disappear once the filter is done loading

### Steps to reproduce

Log in to the application using:
admin@admin.com / password

On the dashboard, you will see the issue.

---

The chart was created from the documentation example without any other modifications:
https://filamentphp.com/docs/3.x/widgets/charts#filtering-chart-data

### Reproduction repository

https://github.com/ModestasV/filamentissue-chart-loading-infinite

### Relevant log output

_No response_

sakanjo:
Fixes #8175.

This PR basically changes `flex` to `hidden`.

This issue is related to livewire's styling in [FrontendAssets.php#L88](https://github.com/livewire/livewire/blob/main/src/Mechanisms/FrontendAssets/FrontendAssets.php#L88) which doesn't include `wire:loading.delay.flex` so instead set hidden by yourself.

Q: why didn't you post this issue to livewire
A:
- Basically i think because it's gonna make livewire's styles long since you can't predict the ordering of delay and the modifiers of loading directives as the examples shows on [Livewire docs](https://livewire.laravel.com/docs/loading) it doesn't dictate that you can use delay and display value.
- Search for `.delay.` in all filament files, the only place that defines `delay` and display property is in the committed file.

A better solution is to remove `delay` from `wire:loading.delay.flex` as the livewire docs doesn't dictate that you can do that and you got to remove `delay` from [wrapper.blade.php#L141](https://github.com/sakanjo/filament/blob/c35fb5a6433e47f2f5b90f8834282fc37fe9ad5a/packages/support/resources/views/components/input/wrapper.blade.php#L141) but this is not my choice it's the author's choice and i respect that.

### DIFF
--- a/packages/support/resources/views/components/input/wrapper.blade.php
+++ b/packages/support/resources/views/components/input/wrapper.blade.php
@@ -89,7 +89,7 @@
                 wire:key="{{ \Illuminate\Support\Str::random() }}" {{-- Makes sure the loading indicator gets hidden again. --}}
             @endif
             @class([
-                'flex items-center gap-x-3 ps-3',
+                'hidden items-center gap-x-3 ps-3',
                 'pe-1' => $inlinePrefix && filled($prefix),
                 'pe-2' => $inlinePrefix && blank($prefix),
                 'border-e border-gray-200 pe-3 ps-3 dark:border-white/10' => ! $inlinePrefix,


danharrin:
Thanks

tvdr:
Hi, this commit hides the magnifying glass icon from the column individual search inputs.

Is this intentional? Should we update something in the column config to fix this on our side?

(attached image shows "flex" wrapper on the left which was the original and "hidden" on the right)

image

saulens22:
This commit hides ALL prefix icons. It breaks all search icons (global, table, column searches).

I think `hidden` should be added conditionally, as `wire:loading.delay.flex` is also conditional.

Possible solution:
```blade
@class([
...
'hidden' => ! $hasPrefix,
...
])

sakanjo:
A better way is to define it globally in css, or use tailwind css arbitrary values to check if it has the attribute `wire:loading.delay.flex` to prevent conditional attributes in php.

PR #8177, opened 2023-08-30
Update textarea component for window events

### LINKED ISSUE #7620: Textarea autosize doesn't work on page load & after resource save
### Package

filament/filament

### Package Version

v3.0.15

### Laravel Version

v10.18.0

### Livewire Version

v3.0-beta

### PHP Version

PHP 8.1.7

### Problem description

Textarea [auto sizing](https://filamentphp.com/docs/3.x/forms/fields/textarea#autosizing-the-textarea) works only when you write some text in the field.
If you click save amount of rows return to default. If you reload page result is the same.
Screenshot from 2023-08-10 00-46-42
Screenshot from 2023-08-10 00-46-29

### Expected behavior

Textarea with autosize attr has size of text in it as in Filament v2

### Steps to reproduce

1. Add long text to Textarea with autosize attr
2. Press save - check, reload page - check

### Reproduction repository

https://github.com/anatolyeltsov/filament/tree/textarea-autosize

### Relevant log output

_No response_

ilpav12:
Fix for issue #7620.
Added 'load' and 'resize' window event listeners to the textarea component. These changes are to ensure that, when using autosize, the textarea height is correct when the window is loaded or resized. This improves the responsiveness of the textarea component in different window sizes.

This is are the examples of how the textarea appears before the fix:
1. when typing and firing the 'input' event that calls the render() function that correctly updates the height 
image
2. when the page is reloaded the textarea goes back to the default height
image
3. when changing windows width the textarea does not update
image
image
After the fix the height is always correct in both cases.

### DIFF
--- a/packages/forms/resources/views/components/textarea.blade.php
+++ b/packages/forms/resources/views/components/textarea.blade.php
@@ -13,6 +13,8 @@
             ax-load-src="{{ \Filament\Support\Facades\FilamentAsset::getAlpineComponentSrc('textarea', 'filament/forms') }}"
             x-data="textareaFormComponent({ initialHeight: @js($initialHeight) })"
             x-ignore
+            x-on:load.window="render()"
+            x-on:resize.window="render()"
             x-on:input="render()"
             style="height: {{ $initialHeight }}rem"
             {{ $getExtraAlpineAttributeBag() }}


danharrin:
Closes #7620

danharrin:
Thanks!

packages/forms/resources/views/components/textarea.blade.php
```diff
@@ -13,6 +13,8 @@
             ax-load-src="{{ \Filament\Support\Facades\FilamentAsset::getAlpineComponentSrc('textarea', 'filament/forms') }}"
             x-data="textareaFormComponent({ initialHeight: @js($initialHeight) })"
             x-ignore
+            x-on:load.window="render()"
```
zepfietje, line 16: Shouldn't we just use `x-init` instead?

qrasa1:
As @zepfietje suggested `x-init` can indeed be used, one problem is that after saving the text area returns to default height. This can be fixed by adding `wire:ignore`. Plus idk why `x-on:resize.window="render()"` was added since it only works when you resize the browser's window.

packages/forms/resources/views/components/textarea.blade.php
ilpav12: I confirm that `x-init="render()"` has the same behavior as `x-on:load.window="render()"`

ilpav12:
@qrasa1 you are correct `wire:ignore` fixes height reset on save. I added the event on browser resize for the times like when you have the browser in full screen and then you split in half to copy some data from another window. I think this is a common behavior for single monitor users.

packages/forms/resources/views/components/textarea.blade.php
zepfietje: Should probably even be moved into `init()` in the JS file.

qrasa1: This won't work because the `init()` method can only be used in `x-data` or `Alpine.data()` as stated [here](https://alpinejs.dev/directives/init#auto-evaluate-init-method)

There's still one problem where if the `Textarea` is inside a modal, the auto size will work the first time, but not the second time. I think that's due to `x-init` only running once.

One possible but not optimal fix is to dispatch an event in `packages/support/resources/views/components/modal/index.blade.php` inside the `open` method like so:
`setTimeout(() => $wire.dispatch('modal-opened'), 500)` and then listen to it in `packages/forms/resources/views/components/textarea.blade.php` like so: `x-on:modal-opened.window="render()"`

I should note that there's already an `modal-opened` event being fired inside the `open` method:
```javascript
this.$refs.modalContainer.dispatchEvent(
                new CustomEvent('modal-opened', { id: '{{ $id }}' }),
            )
``` 
But it did not work, even after putting it inside `setTimeout`

zepfietje: > This won't work because the init() method can only be used in x-data or Alpine.data() as stated [here](https://alpinejs.dev/directives/init#auto-evaluate-init-method)

It should work; we use that pattern in many other components.

qrasa1: I did not know that you could use object syntax with `init`. It works btw

PR #8181, opened 2023-08-30
[v2.x-forms] Fixes `CheckboxList`'s `->afterStateUpdated()` lifecycle hook

bezhanSalleh:

Dispatching change event inside the loop creates race conditions and fires the event as many times as the number of options available which makes the `state` inside the `->afterStateUpdated()` lifecycle hook inconsistent.

This PR fixes the state inconsistency without firing events by setting the state directly in one go.

Current Behavior:
before

Expected Behavior:
after

### DIFF
--- a/packages/forms/resources/views/components/checkbox-list.blade.php
+++ b/packages/forms/resources/views/components/checkbox-list.blade.php
@@ -155,7 +155,7 @@
                     wire:key="{{ $this->id }}.{{ $getStatePath() }}.{{ $field::class }}.options.{{ $optionValue }}"
                 >
                     <label
-                        class="flex items-center space-x-3 filament-forms-checkbox-list-component-option-label rtl:space-x-reverse"
+                        class="filament-forms-checkbox-list-component-option-label flex items-center space-x-3 rtl:space-x-reverse"
                         @if ($isSearchable())
                             x-show="
                                 $el.querySelector('.filament-forms-checkbox-list-component-option-label-text')
--- a/packages/forms/resources/views/components/checkbox-list.blade.php
+++ b/packages/forms/resources/views/components/checkbox-list.blade.php
@@ -58,14 +58,22 @@
             },
 
             toggleAllCheckboxes: function () {
+                updatedStates = []
                 state = ! this.areAllCheckboxesChecked
 
                 this.visibleCheckboxListOptions.forEach((checkboxLabel) => {
                     checkbox = checkboxLabel.querySelector('input[type=checkbox]')
 
                     checkbox.checked = state
-                    checkbox.dispatchEvent(new Event('change'))
+
+                    if (state) {
+                        updatedStates.push(checkbox.value);
+                    }
+
+                    this.checkIfAllCheckboxesAreChecked()
                 })
+
+                @this.set(@js($getStatePath()), updatedStates);
 
                 this.areAllCheckboxesChecked = state
             },
@@ -147,7 +155,7 @@
                     wire:key="{{ $this->id }}.{{ $getStatePath() }}.{{ $field::class }}.options.{{ $optionValue }}"
                 >
                     <label
-                        class="filament-forms-checkbox-list-component-option-label flex items-center space-x-3 rtl:space-x-reverse"
+                        class="flex items-center space-x-3 filament-forms-checkbox-list-component-option-label rtl:space-x-reverse"
                         @if ($isSearchable())
                             x-show="
                                 $el.querySelector('.filament-forms-checkbox-list-component-option-label-text')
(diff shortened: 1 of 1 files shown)


packages/forms/resources/views/components/checkbox-list.blade.php
```diff
@@ -157,9 +165,6 @@ class="filament-forms-checkbox-list-component-option-label flex items-center spa
                         @endif
                     >
                         <input
-                            @if ($isBulkToggleable())
```
danharrin, line 160: We still need this I think, so if the user toggles all manually it shows "Deselect all"

packages/forms/resources/views/components/checkbox-list.blade.php
```diff
@@ -58,15 +58,23 @@
             },
 
             toggleAllCheckboxes: function () {
+                updatedStates = []
```
danharrin: We can probably call this `updatedState`

packages/forms/resources/views/components/checkbox-list.blade.php
```diff
@@ -58,15 +58,23 @@
...
                     checkbox = checkboxLabel.querySelector('input[type=checkbox]')
 
                     checkbox.checked = state
-                    checkbox.dispatchEvent(new Event('change'))
+
+                    if (state) {
+                        updatedStates.push(checkbox.value);
+                    }
+
+                    this.checkIfAllCheckboxesAreChecked()
```
danharrin: We probably should move this outside of the loop, so it only happens once afterwards?

packages/forms/resources/views/components/checkbox-list.blade.php
```diff
@@ -58,15 +58,23 @@
...
                 })
 
+                @this.set(@js($getStatePath()), updatedStates);
```
danharrin: Let's use `$wire` here please for consistency

danharrin (changes requested):
Thanks. Could you please also do this for v3 so we maintain consistency?

packages/forms/resources/views/components/checkbox-list.blade.php
bezhanSalleh: I have tested it, since we do `this.checkIfAllCheckboxesAreChecked()` we do not need it anymore on the input.

packages/forms/resources/views/components/checkbox-list.blade.php
bezhanSalleh: Done

packages/forms/resources/views/components/checkbox-list.blade.php
bezhanSalleh: done

packages/forms/resources/views/components/checkbox-list.blade.php
bezhanSalleh: done

packages/forms/resources/views/components/checkbox-list.blade.php
danharrin: If you individually select each checkbox instead of "Select all", does it switch to "Deselect all" when you check the last one? Because I dont understand when that check would happen, `toggleAllCheckboxes` isnt run in this case

danharrin (changes requested):
Please ensure that code styling is consistent, with no `;` at the end of lines

bezhanSalleh:
> Thanks. Could you please also do this for v3 so we maintain consistency?

as i mentioned on the discord support group these changes does not fix the bug for v3.
i hoped that it would but it doesn't. in v3 when bulktoggleable is one, and when you select all, the first element is removed meaning the option with index 0 is removed from the state and when you deselect all the first element is still present in the state. So i'm outta option for v3 don't know where the state gets mutated this way. but if you still want these changes i will do it anyway so let me know.

packages/forms/resources/views/components/checkbox-list.blade.php
bezhanSalleh: yes it does, because of the `$nextTick()`

danharrin:
Maybe we should move from individual `wire:model` to one giant `$entangle`, that might solve some problems in v3

packages/forms/resources/views/components/checkbox-list.blade.php
danharrin: but init() is only run once, not after every checkbox is individually checked, right? 🤔 sorry for pushing back on this

bezhanSalleh:
> Maybe we should move from individual `wire:model` to one giant `$entangle`, that might solve some problems in v3

i didn't try entangle, was waiting for your PR to get merged in livewire first.
but maybe it should be a gaint `$entangle()`, because currently the `afterStateUpdated()` lifecylce isn't working as it should.

packages/forms/resources/views/components/checkbox-list.blade.php
bezhanSalleh: no worries, and you are right!... well i checked and it works fine... i got nothing 🙃
weird

PR #8182, opened 2023-08-30
Fix: drag animation on mobile

qrasa1:
I have noticed that the row ghost doesn't follow the touch when moving it on mobile:

https://github.com/filamentphp/filament/assets/68746821/587a9c15-43cf-475d-a103-b8948b77882b

After removing `transition duration-75` it works fine: 

https://github.com/filamentphp/filament/assets/68746821/54deda4d-160c-4213-848b-5b9e7f34680e

Obviously `transition` and `duration-75` were there for a reason, but I couldn't see a difference.
I also change the cursor when hovering over the icon-button to `cursor-move` instead of a pointer.

Feel free to fix it the way you think is best!

### DIFF
--- a/packages/tables/resources/views/components/row.blade.php
+++ b/packages/tables/resources/views/components/row.blade.php
@@ -23,7 +23,7 @@
     @endif
     {{
         $attributes->class([
-            'fi-ta-row transition duration-0 md:duration-75',
+            'fi-ta-row transition duration-0 [@media(hover:hover)]:duration-75',
             'hover:bg-gray-50 dark:hover:bg-white/5' => $recordAction || $recordUrl,
             $stripedClasses => $striped,
         ])
--- a/packages/tables/resources/views/components/row.blade.php
+++ b/packages/tables/resources/views/components/row.blade.php
@@ -23,7 +23,7 @@
     @endif
     {{
         $attributes->class([
-            'fi-ta-row duration-0 md:duration-75',
+            'fi-ta-row transition duration-0 md:duration-75',
             'hover:bg-gray-50 dark:hover:bg-white/5' => $recordAction || $recordUrl,
             $stripedClasses => $striped,
         ])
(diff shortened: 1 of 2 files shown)


qrasa1:
I should have noted that setting the `transition-duration` to `0s` fixes the problem and so something like `transition duration-0 md:duration-75` should be possible which I just added. I did not do that in the first commit, because that did not work for me as the `duration-0` class did not register, maybe because I was using the demo app, but I tested it in a different project and it worked fine.

packages/tables/resources/views/components/row.blade.php
```diff
@@ -23,7 +23,7 @@
     @endif
     {{
         $attributes->class([
-            'fi-ta-row transition duration-75',
+            'fi-ta-row transition duration-0 md:duration-75',
```
danharrin: How about `[@media(hover:hover)]:duration-75` so it only affects non-touch devices? Also, do we still need `duration-0`?

danharrin (changes requested):

packages/tables/resources/views/components/row.blade.php
qrasa1: Unfortunately yes, otherwise the default duration from `transition` will be applied

zepfietje:
Honestly, I think a better fix would be in your Tailwind config:

```js
export default {
    future: {
        hoverOnlyWhenSupported: true,
    },
};

```

@danharrin @qrasa1

packages/tables/resources/views/components/row.blade.php
zepfietje: IMO we should apply transition with the media query then.

zepfietje: Applied a couple fixes here, @danharrin @qrasa1: https://github.com/filamentphp/filament/commit/baf1cb6962976243d94816dada80a5b935d1a143.

PR #8193, opened 2023-08-31
Further improvements on textarea

ilpav12:
Changed `x-on:load.window` to `x-init` and added `wire:ignore` after discussion on PR #8177 


packages/forms/resources/views/components/textarea.blade.php
```diff
@@ -13,9 +13,10 @@
             ax-load-src="{{ \Filament\Support\Facades\FilamentAsset::getAlpineComponentSrc('textarea', 'filament/forms') }}"
             x-data="textareaFormComponent({ initialHeight: @js($initialHeight) })"
             x-ignore
-            x-on:load.window="render()"
+            x-init="render()"
             x-on:input="render()"
             x-on:resize.window="render()"
+            wire:ignore
```
zepfietje: Doesn't this have negative side effects?
I think this breaks error state?

packages/forms/resources/views/components/textarea.blade.php
```diff
@@ -13,9 +13,10 @@
             ax-load-src="{{ \Filament\Support\Facades\FilamentAsset::getAlpineComponentSrc('textarea', 'filament/forms') }}"
             x-data="textareaFormComponent({ initialHeight: @js($initialHeight) })"
             x-ignore
-            x-on:load.window="render()"
+            x-init="render()"
```
zepfietje: Could you move this to `init()` inside the JS file maybe?

ilpav12: Done in commit https://github.com/filamentphp/filament/pull/8193/commits/ef67b59ea81097a7345ba20d356c1e83a5a8d702

packages/forms/resources/views/components/textarea.blade.php
ilpav12: @qrasa1 suggested this. I am not aware of what the side effects are

qrasa1: I mean... it shouldn't? I'm not a Livewire expert by any means, but if it did have some negative side effects, wouldn't that be mentioned in the [docs](https://laravel-livewire.com/docs/2.x/alpine-js#ignoring-dom-changes)?

Edit: I think you mean that the errors will stay on the page after saving? It shouldn't happen since the errors aren't added inside the `<textarea>`. I did test the required error and it disappeared after saving.

danharrin: It will break conditionally disabling the field, we can't use wire:ignore here

zepfietje: Also the error styling wouldn't be rendered.

qrasa1: I see... here is one possible workaround:
`textarea.blade.php` (note the `h-full` near the end):
```blade
    <div
        @if ($shouldAutosize = $shouldAutosize())
            ax-load
            x-ignore
            wire:ignore.self
            x-on:input="render()"
            x-on:resize.window="render()"
            style="height: {{ $initialHeight }}rem"
        @endif
    >
        <textarea
            @if($shouldAutosize)
                {{ $getExtraAlpineAttributeBag() }}
            @endif
            {{
                $attributes
                    ->merge([
                        'autocomplete' => $getAutocomplete(),
                        'autofocus' => $isAutofocused(),
                        'cols' => $getCols(),
                        'disabled' => $isDisabled(),
                        'id' => $getId(),
                        'maxlength' => (! $isConcealed) ? $getMaxLength() : null,
                        'minlength' => (! $isConcealed) ? $getMinLength() : null,
                        'placeholder' => $getPlaceholder(),
                        'readonly' => $isReadOnly(),
                        'required' => $isRequired() && (! $isConcealed),
                        'rows' => $rows,
                        $applyStateBindingModifiers('wire:model') => $statePath,
                    ], escape: false)
                    ->merge($getExtraAttributes(), escape: false)
                    ->merge($getExtraInputAttributes(), escape: false)
                    ->class([
                        'fi-fo-textarea block w-full h-full rounded-lg border-none bg-white px-3 py-1.5 text-base text-gray-950 shadow-sm outline-none ring-1 transition duration-75 placeholder:text-gray-400 focus:ring-2 disabled:bg-gray-50 disabled:text-gray-500 disabled:[-webkit-text-fill-color:theme(colors.gray.500)] disabled:placeholder:[-webkit-text-fill-color:theme(colors.gray.400)] dark:bg-white/5 dark:text-white dark:placeholder:text-gray-500 dark:disabled:bg-transparent dark:disabled:text-gray-400 dark:disabled:[-webkit-text-fill-color:theme(colors.gray.400)] dark:disabled:placeholder:[-webkit-text-fill-color:theme(colors.gray.500)] sm:text-sm sm:leading-6',
                        'ring-gray-950/10 focus:ring-primary-600 dark:ring-white/20 dark:focus:ring-primary-500 dark:disabled:ring-white/10' => ! $errors->has($statePath),
                        'ring-danger-600 focus:ring-danger-600 dark:ring-danger-500 dark:focus:ring-danger-500' => $errors->has($statePath),
                    ])
            }}
        ></textarea>
    </div>
``` 
`textarea.js`:
```javascript
export default function textareaFormComponent({ initialHeight }) {
    return {
        init: function () {
          setTimeout(() => this.render())
        },
        render: function () {
            if (this.$el.children[0].scrollHeight > 0) {
                this.$el.style.height = initialHeight + 'rem'
                this.$el.style.height = this.$el.children[0].scrollHeight + 'px'
            }
        },
    }
}
``` 
I thought that the `$getExtraAlpineAttributeBag()` belongs to the text area so I added there.

ilpav12: Ok thank you for the explanation of the drawbacks. Is there any event we can listen for to call the `render()` when you save the edit form?

ilpav12: One solution I have found is to add `wire:key="{{ rand() }}"` which forces Livewire to destroy the old components and to create a new one doing again the `init()`. This does not fix the problem with the modal on second opening.

danharrin: We should not do that, its bad practice to use a random key

danharrin:
Please remove the random key, we shouldnt do that

ilpav12:
Ok do you have any suggestion on how to fix the problem?

zepfietje:
Just a note: GitHub doesn't even seem to render auto-growing textareas at the full content height on page load, only after you start typing. Makes me think what's actually the best UX in this case.

ilpav12:
Ok interesting point. I think that the possibility to immediately have a glance to the full content of a textarea without manually expanding each of them is very convenient, but if you think it is unnecessary we can rollback all the changes and update only on input.

qrasa1:
We could use something like [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver#examples) .
While this could work, it's ultimately up to you to decide whether it's worth it or not. Using markdown editor with no toolbar buttons might be a good alternative.

ilpav12:
Little update: I have added an intersection observer as suggest by @qrasa1 on https://github.com/filamentphp/filament/pull/8193/commits/285b38bf4577275b376e882b85dba9f5b704e446. This indeed fixes the behavior with modal. I am still looking for a fix on a full page save.

ilpav12:
Summarizing the updates as of now, the following cases are fixed:
- initial page load (in dedicated `init()` function instead of  `x-on:load.window` as it is live now)
- browser resize (with` x-on:resize.window="render()"`) 
- modal opening (with `x-intersect.once="render()"`).

Instead, in case of page refresh with Livewire the height is reset. The possible solution are:
- `wire:ignore` but it breaks the conditional disabling
- `wire:key="{{ rand() }}"` but it not a best practice
- listen to a global Livewire page refresh event but I did not manage to find one.

danharrin:
Thanks

PR #8203, opened 2023-08-31
updade pt-BR translations

jeffsmor:


packages/actions/resources/lang/pt_BR/create.php
leandrocfe, line 1: ### Criar :label
Ok

**"Novo"** is an adjective and depends on the gender of the noun. We can say **"Novo"** or **"Nova"** + Noun, depending on whether the noun is masculine or feminine.

**"Criar"** is a verb, and we can use it with all nouns. **Personally, I agree with "Criar"**.

packages/forms/resources/lang/pt_BR/components.php
```diff
@@ -298,7 +298,7 @@
 
                 'actions' => [
                     'link' => 'Link',
-                    'unlink' => 'Unlink',
```
leandrocfe, line 301: Ok

packages/forms/resources/lang/pt_BR/components.php
```diff
@@ -394,7 +394,7 @@
     ],
 
     'tags_input' => [
-        'placeholder' => 'Novo rótulo',
```
leandrocfe, line 397: **Personally, I would keep "Novo rótulo"**.

leandrocfe (changes requested):

packages/actions/resources/lang/pt_BR/create.php
saade: I agree with "Criar" too

packages/forms/resources/lang/pt_BR/components.php
```diff
@@ -298,7 +298,7 @@
 
                 'actions' => [
                     'link' => 'Link',
```
saade, line 300:
```suggestion
                    'link' => 'Inserir Link',
```

packages/forms/resources/lang/pt_BR/components.php
saade: Ok

packages/forms/resources/lang/pt_BR/components.php
saade: I would go for "Rótulo" ou "Etiqueta", yeah

danharrin:
@jeffsmor please address this feedback from our translation managers

packages/actions/resources/lang/pt_BR/create.php
jeffsmor: "Criar" or "Adicionar", here we definitely need a verb...

packages/forms/resources/lang/pt_BR/components.php
jeffsmor: 'Remover link' ?

packages/forms/resources/lang/pt_BR/components.php
jeffsmor: 'Tag' is used by Wordpress so I suggested it, 'Etiqueta' sounds good in my opinion...

packages/actions/resources/lang/pt_BR/create.php
saade: "Criar" fits the best because we're creating something new, whereas "Adicionar" sounds like we're adding something that already exists.

packages/forms/resources/lang/pt_BR/components.php
saade: Yes.

PR #8230, opened 2023-09-02
[Plugin] MediaLibrary: Honor order_column sort order when media is reorderable

drbyte:
I've tested this by putting the `sortBy()` both above and below the `filter()` call. I feel like it'll be slightly more performant "after" the filtering.

I explored using methods in the `Spatie\MediaLibrary\MediaCollections\Models\Concerns\IsSorted` trait, but that applies to the model, not the collection, and the Filament plugin code is querying the collection data using its own bespoke relationship query.
So I think this hardcoded approach is best.

I suppose another approach could be to offer another helper method like `ordered()` to `ImageColumn`, but perhaps that's a bigger design decision to explore separately.

- [N/A] New functionality has been documented or existing documentation has been updated to reflect changes.
- [N/A] Visual changes are explained in the PR description using a screenshot/recording of before and after.


danharrin:
Thanks!

PR #8233, opened 2023-09-02
Handle user login attemp on unauthorized panel

underdpt:
Hello,

When a user tries to login on a panel for which he doesn't have permissions, he's shown a 403 error but the user remains logged in.

This PR verifies that the user has access to the panel during the login process, and if the user isn't allowed to access the panel, it logouts the user and show the credentials mismatch validation error.

- [x ] Changes have been thoroughly tested to not break existing functionality.
There's no documentation about the login process.

There's no visual changes on this PR


packages/panels/src/Pages/Auth/Login.php
```diff
@@ -75,6 +75,14 @@ public function authenticate(): ?LoginResponse
             ]);
         }
 
+        if (! Filament::auth()->user()->canAccessPanel(Filament::getCurrentPanel())) {
+            Filament::auth()->logout();
+
+            throw ValidationException::withMessages([
```
danharrin: Maybe let's extract this throw into its own method `protected function throwFailureValidationException(): void` and then reuse it in both places?

danharrin: Maybe let's extract this throw into its own method `protected function throwFailureValidationException(): void` and then reuse it in both places?

danharrin:
Also PHPstan appears to be failing? And you should add a test for this.

packages/panels/src/Pages/Auth/Login.php
underdpt: Done. Returning `never` instead of `void` as we're on PHP8.1.

underdpt:
Hi,

I've submitted all changes.

I don't know why all checks are failing, they all work on my local environment, and it seems related to CI:

```
Error: Can't use 'tar -xzf' extract archive file: /home/runner/work/_actions/_temp_42897cc5-320e-41f8-9537-4c6ce62b604e/bf82fa85-cb66-4e2c-b2af-f15cf547fbe4.tar.gz. return code: 2.
```

wychoong:
please review your PR, seems like there are some codes duplicated during merge

underdpt:
@wychoong thanks, i'm sorry that got leaked to the PR!

PR #8243, opened 2023-09-02
Fix grouping direction being reset when called programatically

archilex:
When setting tableGrouping and tableGroupingDirection programatically, the current implementation was overriding the direction, resetting it to `asc`. This also meant a separate request was made. This PR adds an early return if the grouping and direction are set.

### DIFF
--- a/packages/tables/resources/views/components/groups.blade.php
+++ b/packages/tables/resources/views/components/groups.blade.php
@@ -15,6 +15,10 @@
     }"
     x-init="
         $watch('group', function (newGroup, oldGroup) {
+            if (newGroup && direction) {
+                return
+            }
+            
             if (! newGroup) {
                 direction = null
 


danharrin:
Not sure about this, when changing the column IMO the direction should be reset

archilex:
It still does reset when changing.

danharrin:
How so? Both the newGroup and direction should be set when changing groups, right?

archilex:
I just got back home to test this and you are right that it does not change back to `asc` when selecting a new column. However, it doesn't currently do that right now either. Currently, when changing the group it keeps whatever direction you have selected. The `oldGroup` early return prevents it from resetting back to `asc` as you were thinking it does.

Now when you remove the group and then select a new group, then it does reset back to `asc`. This PR does that too, which is what I was thinking of.

So basically, this PR keeps all current behavior, but just allows the direction to also be set initially without being overriden. This will be important if in the future you allow `defaultGroup()` to also include the sort direction, since with the current implementation it would be overriden without this PR.

If you like, I can try to implement this in a way that also resets the direction when changing the column.

danharrin:
No worries

PR #8258, opened 2023-09-04
Tables Text-Column: allow Descriptions to wrap

drbyte:
In a Table TextColumn, whether in a `Resource` or in a `RelationManager`, the `wrap()` directive is ineffective if the `->description()` value is the part that actually needs to wrap.

## Example code:

```php
    {
        return $table
            ->columns([
                    ->description(fn (Category $record): string => $record->description ?? '')
                    ->wrap(),
            ])
//...
```

## Screenshots: Resource Table
Before - (notice how the dropdown gets squished because of the lack of wrapping classes):
Screen Shot 2023-09-03 at 10 10 50 PM

After:
Screen Shot 2023-09-03 at 10 11 24 PM

## Screenshots: RelationManager

In a `RelationManager`, horizontal scrolling is required to see table actions:

Screen Shot 2023-09-03 at 9 57 40 PM

Screen Shot 2023-09-03 at 9 57 50 PM

After this change, the wrapping appears as expected:

Screen Shot 2023-09-03 at 10 00 04 PM

    - [x] TO BE CLEAR: I HAVE tested exactly the scenarios above. I've not explored further.
- [n/a] New functionality has been documented or existing documentation has been updated to reflect changes.

### DIFF
--- a/packages/tables/resources/views/columns/text-column.blade.php
+++ b/packages/tables/resources/views/columns/text-column.blade.php
@@ -50,7 +50,11 @@
     }}
 >
     @if (filled($descriptionAbove))
-        <p class="text-sm text-gray-500 dark:text-gray-400">
+        <p
+            @class([
+               'text-sm text-gray-500 dark:text-gray-400',
+               'whitespace-normal' => $canWrap,
+           ])>
             {{ $descriptionAbove }}
         </p>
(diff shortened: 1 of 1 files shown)


danharrin:
Cheers

PR #8281, opened 2023-09-05
Added docblocks

bilogic:
https://github.com/filamentphp/filament/discussions/8257

Docblocks to explain the utility of each method, meant for IDEs to aid new users of this library and also lessen the need to memorize

### DIFF
--- a/packages/forms/src/Components/BaseFileUpload.php
+++ b/packages/forms/src/Components/BaseFileUpload.php
@@ -364,6 +364,9 @@ class BaseFileUpload extends Field
         return $this;
     }
 
+    /**
+     * Specify the maximum number of files that can uploaded
+     */
     public function maxFiles(int | Closure | null $count): static
     {
         $this->maxFiles = $count;
@@ -371,6 +374,9 @@ class BaseFileUpload extends Field
         return $this;
     }
 
+    /**
+     * Specify the minimum number of files that must be uploaded
+     */
     public function minFiles(int | Closure | null $count): static
     {
         $this->minFiles = $count;
@@ -378,6 +384,9 @@ class BaseFileUpload extends Field
         return $this;
     }
 
+    /**
+     * Adds the `multiple` HTML attribute to allow uploading multiple files at once
+     */
     public function multiple(bool | Closure $condition = true): static
     {
         $this->isMultiple = $condition;


packages/forms/src/Components/BaseFileUpload.php
danharrin, line 1: Can you please explain the ambiguity of these method names, as I think they cannot be confused with any other behaviour? I'm trying to make sure we aren't documenting code that is already self-documenting

bilogic:
Hi Dan,

My why slightly differ from yours. I wrote those to help confirm what I think it actually does in a short sentence or two. I just recalled 2 other ambiguous case

### Case 1
`navigationGroups` and `navigationItems` in `HasNavigation.php`

Navigation menus can usually be visualized as tree. It is not clear to me here how `navigationGroups` and `navigationItems` work together to form a tree

### Case 2

https://github.com/filamentphp/filament/blob/a3a4cf2a66e26c2553613cff1ecdecd77f7352f5/packages/panels/src/Panel.php#L49-L54

- As I understood it, each panel has a root url, e.g. `/admin`, so what does `default` do?
- I jumped from the panel's default to the code above, but it does not tell me much, only that `$condition` is stored in `$isDefault`
- Then I have to pull out google, or search the code to be certain for sure, e.g. what happens if I don't specify it etc?


1. I'm really just trying to share my experience as a new user of filament and found that my VSCode IDE could not help much in self-exploring the code to confirm my thoughts
2. Diffferent projects have different quality of naming, even with good naming, as a new user, I have no idea where the benchmark is
3. These days, I prefer not to make too many assumptions what 3rd party methods do simply based on their name as it has backfired several times in the past
4. Now because our why differs, please feel free to reject PR

Thank you.

margarizaldi:
@bilogic
- First, please just answer Dan's question "Can you please explain the ambiguity of these method names?"
- If you really want to explain everything, why there are only 3 among hundreds of methods there?
- You don't have to memorize all the methods, vs code autocompletions and peeks will help us as long as we "know" how to read them. Arguments, return type, namespace, etc are very descriptive. 
For example: `after()` method can be in `...\Concerns\CanBeValidated::after` so we know it is a validation rule related to datetime: image
while another `after()` is in `...\Concerns\HasLifecycleHooks::after` which is meant to be something executed after another execution:
image

- About default(), navigation groups, etc, you better to read the documentation, dig into the source code, or simply throw your questions to discord instead of googling.

bilogic:
@margarizaldi I explained my approach in https://github.com/filamentphp/filament/discussions/8257

danharrin:
I agree that the structure of the navigation, along with the `default()` method, could be explained better. However, I don't think that the code is the best place to do this. Since these are userland API methods, they should be documented on our website along with the others. Also, the explanations would probably be quite long for comments in code, and would not be Googleable as our docs are. You are more than welcome to improve our documentation for that through PRs 👍 Hope you understand

bilogic:
image

@danharrin sure, how about having just URL links to the documentation? It gets turned into a clickable link, at least in VSCode.

danharrin:
Fine, we can add documentation links to various methods. I have no issue with that.

bilogic:
@danharrin 

Sorry to be discussing this again, but I have new insights:

1. Adding links is complicated because the URL links have version in them, e.g. v2, v3
2. If we were to link to the documentation, every new version will incur high overheads
3. I don't think documention links is a good idea, unless there is a permalink to the latest documentation

### New example `->homeUrl(...)` 
5. Another case study here `->homeUrl(...)` and it might sound like a bit of a rant here, please bear with me:
6. 2 weeks ago, there was a suggested solution here https://github.com/filamentphp/filament/discussions/8252#discussioncomment-6899975
7. I feel lucky now about not following it. Looking at Patrick's profile, he probably used filament for a much longer time than me
8. I actually gave up already, because previously, I searched for `->getUrl(`, which yielded too many results
9. Today, I decided to search the code base for `$panel->getUrl` to see where I could get to, and saw `RedirectToHomeController` which sounded interesting, but did not lead far in terms of a solution
10. In the `HasRoutes` traits, I stumbled upon the internal variable `$homeUrl` at the top of the file and decided to pursue it because of the common word `home`, eventually, it led to me finding `function homeUrl()`
11. Like it or not, `homeUrl()` has 0 search results in google and the filament documentation
12. I detailed the above to show the amount of luck and effort required because alot of the methods are not self explanatory to a new user
13. Perhaps the question I would ask is, is there enough manpower to structure good documentation? Or should we just start with the low hanging fruit of documenting the methods first? Someone might eventually leverage on that to structure good documentation

For your consideration please. Thank you.

danharrin:
You can remove version from URLs and it redirects to the latest one

bilogic:
Ok, let's start with links thanks

danharrin:
The `homeUrl()` issue is just a lack of documentation in the first place, it should be added to our website by someone

PR #8292, opened 2023-09-05
Improve Laravel Scout example

francoism90:

This fixes https://github.com/filamentphp/filament/discussions/7697, as the result order was being parsed incorrectly.
It turns out this wasn't a Laravel issue, but I think a change in the MySQL/MariaDB parsing, and one may use `FIND_IN_SET` as a workaround.

The search will now reflect the actual Laravel Scout result order + the user is allowed to overrule the order of columns.

Please let me know if I need to make any adjustments/improvements. :)


danharrin:
Since this isn't a cross-DB solution (doesn't work on Postgres, right?) I don't think we should add it to the docs. However, I think it would be great as an Googleable "trick" article on our website, if you want to PR it to filamentphp/filamentphp.com?

PR #8305, opened 2023-09-06
Fixed where clause for MariaDB compatibility

kaktusss:


danharrin:
Not gonna merge this unless you've tested with MySQL, Postgres, SQL Server, SQlite... and benchmarked performance against the old version. It's a critical query in the database notifications feature

kaktusss:
Mmmh, I first thought the fix was easy because the -> operator is just an alias/shortcut for JSON_EXTRACT in MySQL
(https://dev.mysql.com/doc/refman/8.0/en/json-search-functions.html#operator_json-column-path) so there would be no performance difference. Also, a similar PR was accepted in the past
https://github.com/filamentphp/filament/pull/2524

But it is true that json_extract() is not available on Postgres.

So it needs a condition similar to this:
https://github.com/filamentphp/filament/blob/6e6d21e8670e65e09dada16a7222c320605178ec/packages/tables/src/Columns/Concerns/InteractsWithTableQuery.php#L103C21-L103C21

Anyway in the current state, would be useful to mention in the docs that MariaDB is not fully supported.

danharrin:
Can we tell if a connection is MariaDB or a different database based on the Connection object from Laravel?

kaktusss:
Unfortunately, $query->getConnection()->getDriverName() also returns 'mysql', but $query->getConnection()->isMaria() works.
The method is from Illuminate\Database\MysqlConnection so I added an additional condition to avoid calling isMaria() on PostgresConnection, SQLiteConnection or SqlServerConnection.

packages/notifications/src/Livewire/DatabaseNotifications.php
```diff
@@ -84,8 +84,16 @@ public function isPaginated(): bool
...
     public function getNotificationsQuery(): Builder | Relation
     {
-        /** @phpstan-ignore-next-line */
-        return $this->getUser()->notifications()->where('data->format', 'filament');
+        $query = $this->getUser()->notifications();
+        $databaseConnection = $query->getQuery()->getConnection();
+
+        if (method_exists($databaseConnection, 'isMaria') && $databaseConnection->isMaria()) {
+            /** @phpstan-ignore-next-line */
+            return $query->whereRaw('JSON_UNQUOTE(JSON_EXTRACT(data,\'$.format\')) = "filament"');
```
danharrin:
```suggestion
            return $query->whereRaw('json_unquote(json_extract(data, \'$.format\')) = "filament"');
```

danharrin:
Looks good. Thanks

wychoong:
I’m not sure am I missing something here. I definitely been using `where(‘column->key’, ‘abc’)` and it works. 

And eloquent have this exact handling for mysql/mariadb here https://github.com/illuminate/database/blob/c29a5c3719c9ac7e0909d54cdcf13aa22cc666b4/Query/Grammars/MySqlGrammar.php#L366C40-L366C40

are we changing this for the sake of changing or @kaktusss there is an actual error you hitting?

kaktusss:
@wychoong
MariaDB doesn't support the json arrow search notation:
https://laravel.com/docs/10.x/queries#json-where-clauses
https://jira.mariadb.org/browse/MDEV-13594

The query doesn't trigger an error on my MariaDB instance, it just doesn't find any result because it can't extract with the -> operator.

wychoong:
> The query doesn't trigger an error on my MariaDB instance, it just doesn't find any result because it can't extract with the -> operator.

are you sure?

> MariaDB doesn't support the json arrow search notation:
> https://laravel.com/docs/10.x/queries#json-where-clauses
> https://jira.mariadb.org/browse/MDEV-13594

I'm aware of the incompatibilities or mariadb and mysql you linked before. While the Laravel Doc you referenced didn't explicitly mentioned mariadb, it's common to assume it is compatible with it unless specifically mentioned its not. And that's the whole point of Eloquent, to handle database abstraction as much as it can, as you can see in the link I referred earlier

https://github.com/illuminate/database/blob/c29a5c3719c9ac7e0909d54cdcf13aa22cc666b4/Query/Grammars/MySqlGrammar.php#L366-L371
```php
    protected function wrapJsonSelector($value)
    {
        [$field, $path] = $this->wrapJsonFieldAndPath($value);


        return 'json_unquote(json_extract('.$field.$path.'))';
    }
```
https://github.com/illuminate/database/blob/c29a5c3719c9ac7e0909d54cdcf13aa22cc666b4/Concerns/CompilesJsonPaths.php#L15-L24
```php
    protected function wrapJsonFieldAndPath($column)
    {
        $parts = explode('->', $column, 2);


        $field = $this->wrap($parts[0]);


        $path = count($parts) > 1 ? ', '.$this->wrapJsonPath($parts[1], '->') : '';


        return [$field, $path];
    }
```


for the sake of testing, i ran this in tinker

```php
> use Filament\Notifications\Notification;                                 
> use Illuminate\Notifications\DatabaseNotification;                       
> DatabaseNotification::count();                                           
= 0

> $user = User::first();  

> Notification::make()->title('Saved successfully')->sendToDatabase($user);

> DatabaseNotification::count();                                           
= 1

> $query = $user->notifications();                                         

> $databaseConnection = $query->getQuery()->getConnection();               

> $databaseConnection->isMaria()                                           
= true

> $query->where('data->format', 'filament')->count()                       
= 1

> DatabaseNotification::count();                                                                                                                                  
= 1
```
and it works fine on v3.0.39
image


So again, is there an error in your app that this actually fixes and provide a repo to replicate? If not I did suggest @danharrin to revert this as it is not actually fixing anything

kaktusss:
@wychoong
Thanks for the detailed answer. I could reproduce my issue in tinker with the code you provided (0 result).
I had a look at the raw SQL query and you're totally right, it has been transformed to
``json_unquote(json_extract(`data`, '$."format"')) = 'filament'``
By removing the double quotes from "format", results were then returned.

Then it became obvious that I should try a more recent version of MariaDB (I'm using Laragon). Should have been more obvious before actually...
Updated to 10.6 to have the same version as the Ploi.io deployed version and that's it, now everything works as expected.
I'm sorry for the waste of time and thanks again for clarifying.
As you said, these changes should be reverted.

danharrin:
Okay I will revert

francoism90:
This PR is really weird, sorry.

I'm using MariaDB for years together with Laravel, and I can use something like this just fine:
```php
Book::query()
    ->where('options->color', 'green')
```
I think something else is wrong, I think the cast is missing (`json` or `array`).
If you need more control, something like https://github.com/spatie/laravel-schemaless-attributes would be interesting.

But the same thing remains, I haven't seen any issues with JSON column parsing using MariaDB. If it would, I think you better of reporting this to Laravel itself.

@kaktusss I think the Laravel documentation uses MySQL also for MariaDB. MariaDB has JSON-column support: https://mariadb.com/kb/en/json-data-type/

tonypartridge:
> 

Please see the thread, it is going to be reverted due to it being a legacy MariaDB issue.

francoism90:
@danharrin Any news about the revert? :)
Should I create a ticket for this?

I would like to upgrade Filament, but I'm afraid this breaks notifications, as I'm using the latest MariaDB.

Many thanks

danharrin:
It was done days ago

wychoong:
> @danharrin Any news about the revert? :) Should I create a ticket for this?
> 
> I would like to upgrade Filament, but I'm afraid this breaks notifications, as I'm using the latest MariaDB.
> 
> Many thanks

its already done https://github.com/filamentphp/filament/commit/bda777ed11eb9f82a04a4414aaa0a69cc1b81881

francoism90:
Ah sorry, missed that one!

Thanks for the fast fix. :)

PR #8311, opened 2023-09-06
Remove unused arguments in RM static methods

archilex:
Unless I'm missing something, these unused arguments don't need to be here, and in fact they are preventing these methods from being called statically from outside the class.

### DIFF
--- a/packages/panels/src/Resources/RelationManagers/RelationManager.php
+++ b/packages/panels/src/Resources/RelationManagers/RelationManager.php
@@ -304,17 +304,17 @@ class RelationManager extends Component implements Actions\Contracts\HasActions,
         return $table;
     }
 
-    public static function getIcon(Model $ownerRecord, string $pageClass): ?string
+    public static function getIcon(): ?string
     {
         return static::$icon;
     }
 
-    public static function getIconPosition(Model $ownerRecord, string $pageClass): IconPosition
+    public static function getIconPosition(): IconPosition
     {
         return static::$iconPosition;
     }
 
-    public static function getBadge(Model $ownerRecord, string $pageClass): ?string
+    public static function getBadge(): ?string
     {
         return static::$badge;
     }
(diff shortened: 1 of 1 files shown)


danharrin:
The arguments are passed from the view on render. While they are not used in the default implementations of the class, they are used by users who want to override and customize those features based on the owner record / page class

archilex:
Gotcha...sorry about that.

danharrin:
also, breaking change if you overrode them, even if you didnt use the params :)

archilex:
Oops...
Since this method changed from v2 is there a better way to access the title from outside this class?

danharrin:
What are you trying to do? You can always just pass the params in

archilex:
Ok that works...I was just trying not to instantiate another class, but that's fine. Using this: `$static::getTitle(new Customer(), '')`. Look good to you?

archilex:
Nope, that doesn't work if you override the RM title with `->heading()`. Not sure how to call `getTable()->getHeading()` outside this class.

danharrin:
If you explain what you're trying to do though, I can help. And no, you wont be able to access anything about the table object

archilex:
In Filter Sets, when a user saves a view, the namespace of the resource or relation manager gets saved along with it. Separately there is a resource where a user can manage all of their saved views. In this table, to help distinguish between all of the user’s views, I include the name of the resource and, if it’s tied to a RM, I also include the name of the RM. Before I could call getTitle() and get the title of the RM. Now I can’t. I could just format the saved namespace, but ideally I would be able to get the actual title since that could be different than the namespace or model name. 

This isn’t a problem with Resources since resources have several static methods I can call to get the name/label/etc. 

But, it’s not a big deal anymore, v3 of my plugin changes a lot anyway, so I just implemented this by formatting the namespace. But I am curious if there is a way to do this.

danharrin:
Maybe ask the user for the name before it saves? And then remember that name for the future. And generate a good default from the class name.

archilex:
Yeah thanks. I’ve already done something similar. And as of v3 of my plugin all the management will now happen in the actual resource/RM itself. Big improvement over v2. So where I was trying to do this will now just be for Admins and therefore less critical to have perfect naming.

Anyway, thanks for the responses and sorry for the bad initial PR! 🤦‍♂️

PR #8323, opened 2023-09-07
Hint to run migration - Update 01-installation.md

ajaxray:
If a user is trying the installation guide with a new laravel installation, there is a chance that the migration has not been executed yet. But it's necessary for the next step "Create a user".
image

So, giving the user a hint to run migration before trying and failing to create a user.


danharrin:
Thanks, but I don't think people should be using Filament if they're entirely new to Laravel. If they forget to do this and can't work out what the error says, they will probably also forget `cp .env.example .env`, `composer install`, etc, and we can't go ahead and add every Laravel installation instruction to the docs.

ajaxray:
Got your point @danharrin. 
Thanks for the clarification. 👍🏼

PR #8343, opened 2023-09-08
[Spatie Media Library] fallback to original file when conversion does not exist

rahmanramsi:

This PR will fix the broken image when the conversion file is not exist yet.


rahmanramsi:
Do we need to allow multiple conversion? So the user can decide to fallback to an other conversion until to the original file.

danharrin:
Thanks!

PR #8383, opened 2023-09-10
fix: Disabled toggle

### LINKED ISSUE #7370: Toggle doesn't honor `disabled()`
### Package

filament/filament

### Package Version

v3.0.2

### Laravel Version

v10.17

### Livewire Version

v3-beta

### PHP Version

8.2

### Problem description

[Screencast from 08-02-2023 10:33:38 AM.webm](https://github.com/filamentphp/filament/assets/28250303/38a0b8cb-93ff-41f2-88a7-98bf0d1c099b)
As we can see from vide, toggle is not 'disabled' in view page (modal)

### Expected behavior

Cannot toggleable

### Steps to reproduce

Clone
Install
migrate --seed
Check user resource

### Reproduction repository

https://github.com/valpuia/v3-datetimepicker-bug

### Relevant log output

_No response_

danharrin:
Fixes #7370.

### DIFF
--- a/packages/forms/resources/views/components/toggle.blade.php
+++ b/packages/forms/resources/views/components/toggle.blade.php
@@ -42,6 +42,7 @@
                         'role' => 'switch',
                         'type' => 'button',
                         'wire:loading.attr' => 'disabled',
+                        'wire:target' => $statePath,
                     ], escape: false)
                     ->merge($getExtraAttributes(), escape: false)
                     ->merge($getExtraAlpineAttributes(), escape: false)


sakanjo:
@danharrin Thanks for the PR, but that doesn't fix the problem, right now it only disable the toggle if the property value is being changed but not when saving or updating which is a lot worse.

I've tried it before but it's not gonna help.

sakanjo:
I think we should re-open the issue and revert the PR.

danharrin:
None of the other fields are disabled onsave though, why should this one be?

sakanjo:
the checkbox input is, replace `Toggle` with `Checkbox`.

danharrin:
I don't understand why it needs to be? Thats much less of an issue than Toggle totally ignoring disabled() when its passed manually IMO

sakanjo:
You are the author :smile:.

danharrin:
Yeah but why do you think it needs disabling on save, what use case am I missing?

sakanjo:
It's just a UX issue, since the all the elements on the form page gets disabled when fetching except the non-native input elements, like the button and the multi-select etc...

sakanjo:
Livewire just doesn't handle non-native elements [supportDisablingFormsDuringRequest.js#L16](https://github.com/livewire/livewire/blob/main/js/features/supportDisablingFormsDuringRequest.js#L16).

PR #8392, opened 2023-09-10
Split table search string by runs of white space

judgej:

Discussion on this at https://github.com/filamentphp/filament/discussions/8325 in particular https://github.com/filamentphp/filament/discussions/8325#discussioncomment-6957655

### DIFF
--- a/packages/tables/src/Concerns/CanSearchRecords.php
+++ b/packages/tables/src/Concerns/CanSearchRecords.php
@@ -149,7 +149,7 @@ trait CanSearchRecords
      */
     public function getTableSearch()
     {
-        return filled($this->tableSearch) ? trim($this->tableSearch) : null;
+        return filled($this->tableSearch) ? trim(strval($this->tableSearch)) : null;
     }
 
     public function hasTableSearch(): bool
--- a/packages/tables/src/Concerns/CanSearchRecords.php
+++ b/packages/tables/src/Concerns/CanSearchRecords.php
@@ -149,7 +149,7 @@ trait CanSearchRecords
      */
     public function getTableSearch()
     {
-        return $this->tableSearch === null ? null : trim($this->tableSearch);
+        return filled($this->tableSearch) ? trim($this->tableSearch) : null;
     }
 
     public function hasTableSearch(): bool
--- a/packages/tables/src/Concerns/CanSearchRecords.php
+++ b/packages/tables/src/Concerns/CanSearchRecords.php
@@ -113,7 +113,7 @@ trait CanSearchRecords
 
     protected function applyGlobalSearchToTableQuery(Builder $query): Builder
     {
-        $search = trim($this->getTableSearch());
+        $search = $this->getTableSearch();
 
         if (blank($search)) {
             return $query;
@@ -149,7 +149,7 @@ trait CanSearchRecords
      */
     public function getTableSearch()
     {
-        return $this->tableSearch;
+        return $this->tableSearch === null ? null : trim($this->tableSearch);
     }
 
     public function hasTableSearch(): bool
--- a/packages/tables/src/Concerns/CanSearchRecords.php
+++ b/packages/tables/src/Concerns/CanSearchRecords.php
@@ -104,7 +104,7 @@ trait CanSearchRecords
     }
 
     /**
-     * @return array<integer, string>
+     * @return array<string>
      */
     protected function extractTableSearchWords(string $search): array
     {
--- a/packages/tables/src/Concerns/CanSearchRecords.php
+++ b/packages/tables/src/Concerns/CanSearchRecords.php
@@ -144,10 +144,7 @@ trait CanSearchRecords
         return $query;
     }
 
-    /**
-     * @return ?string
-     */
-    public function getTableSearch()
+    public function getTableSearch(): ?string
     {
         return filled($this->tableSearch) ? trim(strval($this->tableSearch)) : null;
     }
@@ -256,7 +253,7 @@ trait CanSearchRecords
             // Nested array keys are flattened into `dot.syntax`.
             $searches[
                 implode('.', array_slice($path, 0, $iterator->getDepth() + 1))
-            ] = trim($value);
+            ] = trim(strval($value));
         }
 
         return $searches;
--- a/packages/tables/src/Concerns/CanSearchRecords.php
+++ b/packages/tables/src/Concerns/CanSearchRecords.php
@@ -103,15 +103,20 @@ trait CanSearchRecords
         return $query;
     }
 
+    protected function searchWords(string $search): array
+    {
+        return explode(' ', preg_replace('/\s+/', ' ', trim($search)));
+    }
+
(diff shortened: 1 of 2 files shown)


packages/tables/src/Concerns/CanSearchRecords.php
```diff
@@ -103,15 +103,23 @@ protected function applyColumnSearchesToTableQuery(Builder $query): Builder
         return $query;
     }
 
+    /**
+     * @return array<integer, string>
+     */
+    protected function searchWords(string $search): array
+    {
+        return explode(' ', preg_replace('/\s+/', ' ', trim($search)));
```
danharrin: IMO the trim should happen in `getTableSearch()`, leaving only the `preg_replace()` here

packages/tables/src/Concerns/CanSearchRecords.php
```diff
@@ -103,15 +103,23 @@ protected function applyColumnSearchesToTableQuery(Builder $query): Builder
         return $query;
     }
 
+    /**
+     * @return array<integer, string>
+     */
+    protected function searchWords(string $search): array
```
danharrin: Please name this `extractTableSearchWords()` and also use it within the "individual search" query for columns, so that also uses this behaviour

danharrin (changes requested):

judgej:
Thanks. If I don't get the changes sorted tonight, I'll push them tomorrow (Monday).

packages/tables/src/Concerns/CanSearchRecords.php
judgej: Should I leave the trim here too? This method is dealing with white space, and if the string is wrapped in white space, then it is not handling it appropriately. `' hello world '` would give us `['', 'hello', 'world' ,'']`, four words, two of which are empty, which feels kind of strange.

danharrin: The trim should be in `getTableSearch()`, which is passed to `$search`, so the result is the same. There's no need to repeat it.

judgej:
Ignore last commit - misread.

packages/tables/src/Concerns/CanSearchRecords.php
```diff
@@ -141,7 +149,7 @@ protected function applyGlobalSearchToTableQuery(Builder $query): Builder
      */
     public function getTableSearch()
     {
-        return $this->tableSearch;
+        return $this->tableSearch === null ? null : trim($this->tableSearch);
```
danharrin:
```suggestion
        return filled($this->tableSearch) ? trim($this->tableSearch) : null;
```

packages/tables/src/Concerns/CanSearchRecords.php
judgej: The trim is now is `getTableSearch()`, preserving a null if present.

packages/tables/src/Concerns/CanSearchRecords.php
judgej: Renamed the method and extended its use.

Not sure on the review - do I "resolve conversation" when I believe I've resolved, or do you do that when happy with the changes?

packages/tables/src/Concerns/CanSearchRecords.php
judgej: Done that, with a test. An empty string will be returned now as a `null`. It cannot return an empty string. Also a string containing only whitespace.

tests/src/Tables/Concerns/CanSearchRecordsTest.php
```diff
@@ -0,0 +1,56 @@
...
+    assertCount(2, $trait->extractTableSearchWords("testy \t \n \r  test"));
+
+    // Leading and trailing whitespace are extracted as empty words.
+    assertCount(3, $trait->extractTableSearchWords('  test  '));
+    assertCount(3, $trait->extractTableSearchWords(" \t test \n "));
+    assertCount(4, $trait->extractTableSearchWords('  testy test  '));
+    assertCount(4, $trait->extractTableSearchWords(" \t testy test \n "));
+    assertCount(5, $trait->extractTableSearchWords("   \t testy  \t  tasty  \n  test \r   "));
+
+    $this->assertSame(['', 'test', ''], $trait->extractTableSearchWords('   test   '));
```
danharrin: Please explain this behaviour, why should words be created out of whitespace, when they're just `''`?

judgej: That's the consequence of not trimming in the `extractTableSearchWords()` method being tested. The test just shows it explicitely. It's not desired behaviour IMO.

judgej: An empty string would also return `['']` which is probably also undesirable. Better would be `[]` (i.e. no words).

danharrin: Yes, I don't believe that there should ever be empty strings generated like this, the regex should match **one or more** whitespace characters maybe?

judgej:
My original thinking was, the query string is what it is - spaces, characters, null (at the start), mixed letter case. That doesn't get changed in the process, since any party of that could be important to someone's search rules.

The two functions that do SQL searching then needs a list of words. The `extractTableSearchWords()` function would return an array of words from the *raw* search text. It would handle all the parsing - trimming, white space, etc. The methods using `extractTableSearchWords()` would not need to care what that raw string looks like, whether it has white space around it or not. That's why I had the `trim()` in there.

The search string does not need to be trimmed before checking if it is `filled()` because `filled()` will treat a string of zero or more whitespace characters as unfilled.

danharrin:
Ok, for now I've removed the tests about untrimmed whitespace, since that will never be passed to `extractWords()` in the first place 👍

Thanks

packages/tables/src/Concerns/CanSearchRecords.php
```diff
@@ -103,15 +103,23 @@ protected function applyColumnSearchesToTableQuery(Builder $query): Builder
         return $query;
     }
 
+    /**
+     * @return array<integer, string>
```
danharrin:
```suggestion
     * @return array<string>
```

packages/tables/src/Concerns/CanSearchRecords.php
```diff
@@ -141,7 +149,7 @@ protected function applyGlobalSearchToTableQuery(Builder $query): Builder
      */
     public function getTableSearch()
     {
-        return $this->tableSearch;
+        return filled($this->tableSearch) ? trim($this->tableSearch) : null;
```
danharrin:
```suggestion
        return filled($this->tableSearch) ? trim(strval($this->tableSearch)) : null;
```

JohnRoux:
This word splitting is really nice for small applications and tables. But it does come with a downside from an indexing perspective.

Is there room for having both, defaulting to the nice word splitting, but having the option to disable it easily on the Page level?

I put a bit more info in this discussion item: https://github.com/filamentphp/filament/discussions/13466

PR #8406, opened 2023-09-11
Fix type of tag isn't handled correctly

kerbylav:
Type of tag wasn't handled correctly because of missing assignment grouping in condition. Because of that newly new tags of default type were created instead of using exiting ones.

### DIFF
--- a/packages/spatie-laravel-tags-plugin/src/Forms/Components/SpatieTagsInput.php
+++ b/packages/spatie-laravel-tags-plugin/src/Forms/Components/SpatieTagsInput.php
@@ -40,7 +40,7 @@ class SpatieTagsInput extends TagsInput
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


packages/spatie-laravel-tags-plugin/src/Forms/Components/SpatieTagsInput.php
```diff
@@ -40,7 +40,7 @@ protected function setUp(): void
                 return;
             }
 
-            if ($type = $component->getType() && ! $component->isAnyTagTypeAllowed()) {
+            if (($type = $component->getType()) && ! $component->isAnyTagTypeAllowed()) {
```
danharrin:
```suggestion
            if (($type = $component->getType()) && (! $component->isAnyTagTypeAllowed())) {
```

danharrin:
Thanks

PR #8417, opened 2023-09-11
[Draft]: Add Label as fallback when `navigationSort` is not set

alexanderkroneis:
This PR introduces a fallback for sorting navigation items when `navigationSort` is not set. It'll use `getLabel()` of the navigation item to provide a alphabetical sorting.

Bildschirmfoto 2023-09-11 um 22 14 41

Bildschirmfoto 2023-09-11 um 22 16 56

- [x]  New functionality has been documented or existing documentation has been updated to reflect changes.
- [x]  Visual changes are explained in the PR description using a screenshot/recording of before and after.

### DIFF
--- a/packages/panels/src/Navigation/NavigationItem.php
+++ b/packages/panels/src/Navigation/NavigationItem.php
@@ -178,9 +178,9 @@ class NavigationItem extends Component
         return $this->evaluate($this->label);
     }
 
-    public function getSort(): int
+    public function getSort(): ?int
     {
-        return $this->evaluate($this->sort) ?? -1;
+        return $this->evaluate($this->sort) ?? null;
     }
 
     public function getUrl(): ?string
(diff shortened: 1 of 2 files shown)


alexanderkroneis:
Feel free to discuss my current approach. I started with Filament last week, so there might be a better approach than this.

danharrin:
What is the default order currently? I thought it was the label TBH

danharrin:
Actually, after some consideration, I think this is a breaking change that we cannot make. People might be relying on the old order.

alexanderkroneis:
I didn't find any hints on that question within the code (pretty sure there is tho) but as far as I can tell from UI: It's ordered by the English label, which can differ from the translated one.

alexanderkroneis:
> Actually, after some consideration, I think this is a breaking change that we cannot make. People might be relying on the old order.

That's why it's still drafted, I thought about adding a flag at the `AdminPanelProvider`.

```php
->useNavigationSort(NavigationSort::TranslatedLabel)
```

and other values could be `NavigationSort::Sort` or `NavigationSort::Default`. `Default` could be the default then, so nothing of the known behavior would change.

danharrin:
To be honest, people just need to add sorts to their nav items, that would solve it 😛 Also, if I were a user, I wouldn't want things in the UI moving around depending on what language I am using, you know?

alexanderkroneis:
Yeah, I see that point. Thank you for discussion. :)

PR #8425, opened 2023-09-12
fix: Spatie tags sync

### LINKED ISSUE #8411: Array to string conversion when using spatie tags
### Package

filament/filament

### Package Version

v3.0.47

### Laravel Version

v10.10

### Livewire Version

v3.0.2

### PHP Version

PHP 8.10

### Problem description

When you try to add tags using the spatieTagsInput to any resource there's a string to array conversion error. This worked fine on 3.0.40. Updated a project today and it didn't work anymore.

### Expected behavior

A tag to be assigned to the resource.

### Steps to reproduce

Try to add a tag to the user resource in the example project.

### Reproduction repository

https://github.com/Tim-Frensch/filament-tags-bug

### Relevant log output

_No response_

danharrin:
Fixes #8411.

### DIFF
--- a/packages/spatie-laravel-tags-plugin/src/Forms/Components/SpatieTagsInput.php
+++ b/packages/spatie-laravel-tags-plugin/src/Forms/Components/SpatieTagsInput.php
@@ -83,7 +83,7 @@ class SpatieTagsInput extends TagsInput
             return $tag;
         })->flatten();
 
-        $record->tags()->sync($tags);
+        $record->tags()->sync($tags->pluck('id'));
     }
 
     public function type(string | Closure | AllTagTypes | null $type): static


sairiz:
I tried the fix. Apparently it only works when the model doesnt have any tags.

If it already has, and new tags were to be saved, it will throw SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry
If it already has and some of the tags removed and were to be saved, it will throw SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry
If it already has and all of the tags removed and were to be saved, no error thrown but the tags doesnt get removed.

danharrin:
I can't reproduce that behaviour, please create a reproduction repo @sairiz

sairiz:
I am using custom tag model and override the tags method according to official spatie documentation 

```
public function tags(): MorphToMany
{
    return $this
        ->morphToMany(self::getTagClassName(), 'taggable', 'taggables', null, 'tag_id')
        ->orderBy('order_column');
}

```
It differ from their current method in the HasTags trait

```
public function tags(): MorphToMany
    {
        return $this
            ->morphToMany(self::getTagClassName(), $this->getTaggableMorphName())
            ->using($this->getPivotModelClassName())
            ->ordered();
    }
```

The current method chain  has using($this->getPivotModelClassName())

I made some adjustment and its all good but you might want to put in documentation for those who use their own tag model and follow the official instruction.

danharrin:
What adjustment did you need to make

sairiz:
In my particular case, i am using 2 seperate databases. 
The tags & taggables tables is on the main database.
The model i want to associate tags with is in secondary database,
so for that i need to create custom Tag model which extends Spatie\Tags\Tag

```
    // This class model is in the secondary database 

    use HasTags;

    public static function getTagClassName(): string
    {
        return App\Models\Tag::class;
    }

    public function tags(): MorphToMany
    {
        return $this
            ->morphToMany(self::getTagClassName(), $this->getTaggableMorphName(), 'main_database.taggables')
            ->using($this->getPivotModelClassName())
            ->ordered();
    }
```

I need to explicitly append the main_database name to the table name otherwise it will look for taggables table in the secondary database where the model is located. 

However in most use case (single database), i believe that will be unnecessary and the tags function did not have to be overridden unless the custom setup involve custom taggables table and model as well.

Issue #7591, opened 2023-08-09
Translatable resource with media bug


anaseqal:
### Package

filament/spatie-laravel-translatable-plugin

### Package Version

v3.0.13

### Laravel Version

v10.18.0

### Livewire Version

v3.0.0-beta.8

### PHP Version

PHP 8.2.8

### Problem description

Editing translatable resource image throw exception: 

> Call to a member function hasGeneratedConversion() on null

Even if no conversion defined, there is still an error:

> Call to a member function getAttributeValue() on null

### Expected behavior

Edit image(s) without throwing errors

### Steps to reproduce

1. create a new customer and upload an image.
2. edit the customer and change the image then click save changes

![image](https://github.com/filamentphp/filament/assets/5034308/d358b091-6dc8-4b89-890c-7888d6933f6f)

### Reproduction repository

https://github.com/anaseqal/filament3-media-translation

AlexanderFalkenberg:
Having the same issue.

AxelDeneu:
Same here

elnurxf:
One more here. Stuck for hours to figure it out. :/

Saifallak:
Same here

pieinn:
Same issue 
![issue](https://github.com/filamentphp/filament/assets/29000827/530e3f7a-efda-48d7-8295-529a950a97ec)

awcodes:
Figured out the issue, not sure of how to resolve it though, @danharrin leaving this here for you to see what I'm seeing.

The issue is in the Translatable Plugin at Resources\Pages\EditRecord\Concerns\Translatable class.

When looping through the locales to save the record data is getting out of sync.

data in `save()` for 'en' loop:
![Screenshot 2023-08-26 at 2 59 26 PM](https://github.com/filamentphp/filament/assets/3596800/28d94a3e-c3a1-412d-b22e-f845cb4c38cf)

record in `handleRecordUpdate()`:
![Screenshot 2023-08-26 at 2 59 33 PM](https://github.com/filamentphp/filament/assets/3596800/ddecab6e-e16f-4db0-8c7c-f1f2beaa1433)

All good to this point.

data in `save()` for 'es' loop:
![Screenshot 2023-08-26 at 2 59 37 PM](https://github.com/filamentphp/filament/assets/3596800/ddd70f5d-a2a2-409b-a49d-862844de54e9)

record in `handleRecordUpdate()`:
[...]

PR #8432, opened 2023-09-12
spatie-laravel-translatable-plugin - Fix media bug

robertorinaldi-dev:
This should solve https://github.com/filamentphp/filament/issues/7591.
Or at least it works for me 😄

### DIFF
--- a/packages/spatie-laravel-translatable-plugin/src/Resources/Pages/EditRecord/Concerns/Translatable.php
+++ b/packages/spatie-laravel-translatable-plugin/src/Resources/Pages/EditRecord/Concerns/Translatable.php
@@ -30,13 +30,18 @@ trait Translatable
         $originalActiveLocale = $this->activeLocale;
 
         try {
-            foreach ($this->getTranslatableLocales() as $locale) {
+            /** @internal Read the DocBlock above the following method. */
+            $this->validateFormAndUpdateRecordAndCallHooks();
+
+            $otherTranslatableLocales = Arr::except($this->getTranslatableLocales(), $originalActiveLocale);
+
+            foreach ($otherTranslatableLocales as $locale) {
                 $this->setActiveLocale($locale);
 
                 $nonTranslatableData = Arr::except(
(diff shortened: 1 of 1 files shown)


atmonshi:
thank you

After a quick test, it seems to work for only the first language
If I edit, then change the language, then change the uploaded photo it won't save.

robertorinaldi-dev:
You are right.
I made a patch that should fix it.
I don't like it very much but it seems to work. Unfortunately I don't have time to test it well right now.

robertorinaldi-dev:
Basically, it always proceeds first with the currently active language and then continues with the others (copying the non-translatable fields from the first)

atmonshi:
so far so good, will get more testers soon to confirm :)

robertorinaldi-dev:
I don't know if it can create problems if there are other particular fields, like Repeater.
I also need to test it for the case where the media field is also translatable (in my case it is not)

danharrin:
Can you please explain how this fixes the problem @robertorinaldi-dev ?

robertorinaldi-dev:
Before the change $nonTranslatableData was taken in the initial state, which in the case of a media was a TemporaryUploadedFile and then it was being applied to all languages; this way when saving the second language (or third, fourth, ...) it resulted in what @awcodes noted [here](https://github.com/filamentphp/filament/issues/7591#issuecomment-1694483651), that is, it was trying to save the already saved file again and didn't find it anymore (it was deleted when saving the first language).
After the change, first it proceeds with saving the current language as it is, then it gets $nonTranslatableData form the language it just saved (therefore without TemporaryUploadedFile) and applies it to all other languages; this way it doesn't try to save the same file again.

I hope I explained myself well

danharrin:
Thanks, and this isn't an issue with Create?

robertorinaldi-dev:
How it works in creation is slightly different, so no, it doesn't seem to cause any problems to me

danharrin:
Thanks mate, quality work as always

PR #8435, opened 2023-09-12
Update return type of getRelationshipToSum method

gasner:
Changed the return type of getRelationshipToSum function in CanAggregateRelatedModels class. Previously, the function could only return string or null, Now it can return string, array, or null. This change gives the option to set additional query constraints on the sum query.


danharrin:
1) Please test your PRs as this doesn't work, `$this->relationshipToSum` is never set to an array in the first place.
2) I think we should change the syntax, instead of allowing an array we should add a third param to `sum()` as a Closure to scope the relationship by.

danharrin (approved):

PR #8440, opened 2023-09-12
chore: update public asset paths

SupianIDz:

For consistent placement of assets from vendors and have discussed #8415 

before :
Screenshot 2023-09-12 at 12 02 25

after :
Screenshot 2023-09-13 at 04 19 48

Screenshot 2023-09-13 at 04 21 11

### DIFF
--- a/packages/support/src/Assets/AlpineComponent.php
+++ b/packages/support/src/Assets/AlpineComponent.php
@@ -11,7 +11,7 @@ class AlpineComponent extends Asset
 
     public function getRelativePublicPath(): string
     {
-        return "js/{$this->getPackage()}/components/{$this->getId()}.js";
+        return "vendor/{$this->getPackage()}/components/js/{$this->getId()}.js";
     }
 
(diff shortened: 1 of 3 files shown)


danharrin:
This is a breaking change. It needs to be a configuration option in config/filament.php if you want to see this.

SupianIDz:
> This is a breaking change. It needs to be a configuration option in config/filament.php if you want to see this.

it will update automatically when the user does `composer update` or `composer dump` or `artisan filament:upgrade`

So there is no need for additional config just for that path.

even as if nothing had happened,
Screenshot 2023-09-13 at 07 14 40

danharrin:
And it will leave a ton of dead files in their /public directory. No thanks.

SupianIDz:
If I add config as you mean, will it be merged?

To be honest, it's very annoying when files from packages that shouldn't need to be done anything get mixed up with asset files from user development.

danharrin:
Yup, I will merge config

PR #8445, opened 2023-09-13
Explicitly name emptyStateActions when creating relation manager

cheesegrits:
If the emptyStateActions are not named, they collide with the headerActions.  So when the user sets (say) ->after() on the headerActions CreateAction ... it mysteriously doesn't work, because the (unnamed) CreateAction from emptyStateActions overwrites it.


danharrin:
As discussed on Discord, this is on my radar

PR #8447, opened 2023-09-13
Fix pagination overlap

### LINKED ISSUE #7597: Pagination overlaps page selector on mobile
image

Hasnayeen:
Currently in smaller screen size the per page select and page navigation overlap like shown in before screenshot. This PR fix the pagination overlap issue.

Before:
image

After:
image


danharrin:
Fixes #7597

danharrin:
Probably not perfect as the page selector is now not centered, but it's better than the overlap

zepfietje:
> Probably not perfect as the page selector is now not centered, but it's better than the overlap

Yeah I really don't like this to be honest. Reopening the original issue to track this and implement a proper fix using container queries.

Issue #7969, opened 2023-08-21
Filament Sidebar not working correctly with Multi-tenancy + Octane server


miguilimzero:
### Package

filament/filament

### Package Version

v3.0.30

### Laravel Version

v10.19.0

### Livewire Version

v3.0.0-beta.10

### PHP Version

PHP 8.2.9

### Problem description

There are currently two different problems depending on which Octane adapter you're now using:

### Swoole
If you have more than one tenancy and switch between them, the sidebar link will not update accordingly. Example:

- I'm on team ID 1
- Now I switched to team ID 2
- The post resource sidebar link will still point to team ID 1

This happens randomly. Sometimes it updates the links correctly, sometimes not. Most of time the bug will appear.

### Roadrunner
If you keep refreshing the page, the sidebar resources will appear and disappear randomly.

**I could only reproduce at the beginning. I don't know if it was a fault on my side. However, the Swoole issue is easily reproducible.**

### Expected behavior

Everything works seamlessly.

### Steps to reproduce

- Download the example repository
- php artisan migrate
- Set the octane runner you want to test using `OCTANE_SERVER` (Swoole)
- Start the local server with `php artisan octane:start`
- Register your user
- Create team 1 and then create team 2

### Reproduction repository

https://github.com/miguilimzero/filament-tenant-octane-bug-example

miguilimzero:
One thing that I noticed while debugging this issue is that the `mountNavigation()` method from `Panel\Concerns\HasNavigation` is called multiple times using **Swoole**, while with the default web server or Roadrunner, the method is called one time in one request.

- 4 times in one request when the sidebar link is updated correctly.
- 3 times in one request when the sidebar link points to the wrong team ID.

binaryfire:
Just ran into this as well. I'm using Swoole. Here's a video demonstrating the problem: https://www.youtube.com/watch?v=1r7deK__IJI

I've pushed a new version of the example repo to https://github.com/binaryfire/filament-tenant-octane-bug. The packages have been updated and a missing relationship has been fixed. You should be able to just clone it and run `php artisan octane:start` to reproduce. Just make sure you have the Swoole extension installed.

Any chance this could get bumped to high priority? I'm happy to help with testing. Not sure how much help I'll be on the debugging front though.

binaryfire:
Just tested an app that uses Spatie's tenancy package for subdomain multitenancy. The same issue occurs with that too (with Filament's tenancy disabled).

[...]

PR #8448, opened 2023-09-13
Fix Octane tenant navigation

binaryfire:
This fixes the issues described in https://github.com/filamentphp/filament/issues/7969

It's a bit hacky. We're resetting state for [`$isNavigationMounted`](https://github.com/filamentphp/filament/blob/95ad3aa5e777381d447b3604a01bb64346e226b8/packages/panels/src/Panel/Concerns/HasNavigation.php#L13), [`$navigationGroups`](https://github.com/filamentphp/filament/blob/95ad3aa5e777381d447b3604a01bb64346e226b8/packages/panels/src/Panel/Concerns/HasNavigation.php#L18) and [`$navigationItems`](https://github.com/filamentphp/filament/blob/95ad3aa5e777381d447b3604a01bb64346e226b8/packages/panels/src/Panel/Concerns/HasNavigation.php#L23) for Octane users at the end of the request. It'd be better to refactor so they're not used statically, but it gets things working for now.

### DIFF
--- a/packages/panels/src/Panel.php
+++ b/packages/panels/src/Panel.php
@@ -7,6 +7,7 @@ use Filament\Support\Components\Component;
 use Filament\Support\Facades\FilamentColor;
 use Filament\Support\Facades\FilamentIcon;
 use Filament\Support\Facades\FilamentView;
+use Illuminate\Support\Facades\Event;
 
 class Panel extends Component
 {
@@ -70,6 +71,14 @@ class Panel extends Component
 
         foreach ($this->plugins as $plugin) {
             $plugin->boot($this);
+        }
+
+        if (class_exists(\Laravel\Octane\Events\RequestTerminated::class)) {
+            Event::listen(\Laravel\Octane\Events\RequestTerminated::class, function () {
+                $this->isNavigationMounted = false;
+                $this->navigationGroups = [];
+                $this->navigationItems = [];
+            });
         }
 
         if ($callback = $this->bootUsing) {
--- a/packages/panels/src/Panel.php
+++ b/packages/panels/src/Panel.php
@@ -73,16 +73,16 @@ class Panel extends Component
             $plugin->boot($this);
         }
 
+        if ($callback = $this->bootUsing) {
+            $callback($this);
+        }
+
         if (class_exists(\Laravel\Octane\Events\RequestTerminated::class)) {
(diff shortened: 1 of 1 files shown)


binaryfire:
Update: This breaks [Registering custom navigation items](https://filamentphp.com/docs/3.x/panels/navigation#registering-custom-navigation-items). The additional items are only visible on the first request, then they disappear.

danharrin:
I don't actually understand why this needs to happen. Panel objects are bound to the container through the FilamentManger object. FilamentManager objects are "scoped" bindings, so they should get refreshed between Octane requests. Am I wrong here?

binaryfire:
@danharrin You're probably right. I think the underlying issue is that the navigation array is being persisted. Maybe this helps demonstrate...

If I make this change to `HasNavigation`:

```
    public function getNavigation(): array
    {
        if ($this->navigationBuilder instanceof Closure) {
            return $this->buildNavigation();
        }

        $this->mountNavigation();
        // if (! $this->isNavigationMounted) {
        //     $this->mountNavigation();
        // }
        
        ....

```

And then visit a bunch of different tenants' dashboards, I get this:

image

binaryfire:
The urls of those nav items are a mixed bag of different tenants' urls. And each request adds more.

I'm testing with the Spatie package atm, not Filament's tenancy. But the same thing is happening with both

danharrin:
I believe you that it's happening, but without understanding why the same object is being persisted between requests, I am hesistant to merge any PRs that "fix it"

binaryfire:
@danharrin For sure. As I mentioned this breaks other things so it shouldn't be merged.

I've hit the limits of my debugging skills unfortunately. All I can say is that based on other stuff I've had to debug with Octane in the past, it's probably related to a static property / method somewhere.

I think static properties and methods retain their state across requests until the Octane workers are restarted. My understanding is that scoping FilamentManager objects would work for instance variables, but static properties / methods are not part of that scope. They're class-level and live as long as the class is loaded in memory (the lifetime of the worker).

binaryfire:
@danharrin From the Laravel docs: https://laravel.com/docs/10.x/octane#managing-memory-leaks

danharrin:
We do not use static properties to handle state, anything we store in a static property can be persisted between requests. As far as I know, we do not merge into static arrays.

PR #8458, opened 2023-09-13
Change canView to non-static

iAmKevinMcKee:
The canView method is not really useful as a static method because it doesn't have access to any of the information it might need to determine if the widget should be viewable or not.

Updating to non-static will allow you to override this method with some logic based on the properties in the class.


iAmKevinMcKee:
Closing this, as I've looked through the code and it seems this static `canView` is a common approach. Maybe we need to add a public `shouldRender` or something. I will see if I can PR that.

PR #8461, opened 2023-09-13
Support additional repeater and builder actions

yassine-kessal:
This PR adds the possibility of setting custom actions in the form's reapter field (other than predefined add, clone, etc. actions), using the repeater's registerActions method, example :

```php
Repeater::make('fields')
      ->collapsible()
      ->reorderableWithButtons()
      ->itemLabel(fn (array $state): ?string => $state['name'] ?? null)
      ->columns(2)
      ->registerActions([
          Action::make('field-settings')
              ->icon('heroicon-o-cog')
              ->iconButton(),
          Action::make('copy')
              ->icon('heroicon-o-clipboard')
              ->iconButton(),
      ]);
```

**Preview :** 
image


yassine-kessal:
Oupps sorry for commits sync on this branch, let me know if you want me to recreate the PR.

yassine-kessal:
A little commit to add this feature for Builder form component and be more consistent.

tonypartridger:
Thoughts on adding support to builder blocks too?

danharrin:
This PR covers repeater and builder header actions.

tonypartridger:
> This PR covers repeater and builder header actions.

It sure does, but unless I am missing something it doesn't cover Builder 'Block' actions.

danharrin:
I do not understand what a block action is, compared to a header action.

tonypartridger:
> I do not understand what a block action is, compared to a header action.

No problem, please see attached.
Screenshot 2023-12-11 at 09 23 44

danharrin:
its a header action

tonypartridger:
> its a header action

It's a header action of the builder block type yep. This PR only supports header actions on the repeater and the builder header actions not the builder block header actions as far as I can tell. 

I am asking what your thoughts are on adding it to the Builder Block Header Actions too.

danharrin:
in your screenshot, they both look like `extraItemActions()` that are possible in this PR

tonypartridger:
See attached the Builder\Block does not support extraItemActions. This is what I was suggesting we expand support to.

Screenshot 2023-12-11 at 09 54 30

danharrin:
So are you asking for block-specific actions that are not added to every block type at once? Because they would look the same as the current solution

PR #8479, opened 2023-09-14
Update 05-viewing-records.md

indigoram89:
infolist method is not static


danharrin:
It is on the resource class

PR #8493, opened 2023-09-14
fix: Empty state actions overriding

### LINKED ISSUE #7649: Adding form fields to AttachAction does not work
### Package

filament/tables

### Package Version

v3.0.16

### Laravel Version

v10.18.0

### Livewire Version

v3.0.0-beta.7

### PHP Version

PHP 8.2.8

### Problem description

When adding an AttachAction to a table inside a RelationManager, any custom fields added to it using `->form()` do not show up. Below are two images and the resulting form:

image

image

A link to the relation manager that is misbehaving: https://github.com/bram-pkg/filament-issue-7649/blob/main/app/Filament/Admin/Resources/TenantResource/RelationManagers/UsersRelationManager.php#L43-L49

Note that I have confirmed with the documentation that this is the expected syntax.

### Expected behavior

It's expected that the select dropdown shows up underneath the user selection dropdown.

### Steps to reproduce

Create an AttachAction inside a RelationManager and add a custom field to the form with code like:

```php
Tables\Actions\AttachAction::make()
    ->preloadRecordSelect()
    ->form(fn (Tables\Actions\AttachAction $action): array => [
        $action->getRecordSelect(),
        Forms\Components\TextInput::make('role')->required(),
    ])
```
                   

### Reproduction repository

https://github.com/bram-pkg/filament-issue-7649

### Relevant log output

_No response_

danharrin:
Fixes #7649.

On top of preventing empty state actions from overriding header actions, this completely removes empty state actions from the stubs of resources and relation managers. This isn't a breaking change, only new resources and relation managers will be affected. The reason why I made this decision, was due to the confusion surrounding *which* `CreateAction`, `AttachAction`, `AssociateAction` to `modify()`, for beginners. There's two occurrences of the "same" action, and they don't understand which one to touch. This should clear that up. Advanced users can just add these to their stubs.

### DIFF
--- a/packages/panels/src/Commands/MakeRelationManagerCommand.php
+++ b/packages/panels/src/Commands/MakeRelationManagerCommand.php
@@ -101,19 +101,19 @@ class MakeRelationManagerCommand extends Command
             return static::INVALID;
         }
 
-        $tableHeaderAndEmptyStateActions = [];
+        $tableHeaderActions = [];
 
-        $tableHeaderAndEmptyStateActions[] = 'Tables\Actions\CreateAction::make(),';
+        $tableHeaderActions[] = 'Tables\Actions\CreateAction::make(),';
 
         if ($this->option('associate')) {
-            $tableHeaderAndEmptyStateActions[] = 'Tables\Actions\AssociateAction::make(),';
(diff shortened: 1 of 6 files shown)


bram-pkg:
Cheers Dan, thank you for this!

PR #8494, opened 2023-09-14
feature: Disable read only relation managers for entire panel at once

danharrin:
Seems like everyone hates that feature (apart from the people who constantly suggested it in v2 😳) sooooooo


ralphjsmit:
Just wanted to note I got this request yesterday from a user who requested relation managers on view pages to be readonly, because it was "confusing that the form above was readonly and the relations editable". So not everyone hates it :) 💛

ziming:
I like it too

PR #8504, opened 2023-09-14
Check for index page existence before generating breadcrumb

victorlap:

Currently the `getBreadcrumbs` assumes that every resource has an index page. I'm currently creating an app where this is not the case. This PR adds a check to see if the 'index' page actually exists before adding the breadcrumb, in line with the other breadcrumbs.

I think there is no need to change the documentation or update a screenshot since nothing changes for existing apps. Currently an exception occurs when calling the `$resource::getUrl()` method.

### DIFF
--- a/packages/panels/src/Resources/Pages/Concerns/InteractsWithRecord.php
+++ b/packages/panels/src/Resources/Pages/Concerns/InteractsWithRecord.php
@@ -47,9 +47,11 @@ trait InteractsWithRecord
     {
         $resource = static::getResource();
 
-        $breadcrumbs = [
-            $resource::getUrl() => $resource::getBreadcrumb(),
-        ];
+        $breadcrumbs = [];
+
+        if ($resource::hasPage('index')) {
+            $breadcrumbs[$resource::getUrl()] = $resource::getBreadcrumb();
+        }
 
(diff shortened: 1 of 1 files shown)


danharrin:
Hey, whats the use case of a resource without the index page?

victorlap:
Currently I link to this resource from a relation manager. The concept of the index page is not relevant for my app, but on the 'view' page I show some history for this record. Other than this I had to make no changes for the functionality to work.

danharrin:
In this situation, I do not suggest using resources. Resources are for full CRUDs. Use a custom page, and the form builder.

victorlap:
Hmm, disagree here since it's a very minor change and the page will need to look like the other resource pages. Duplicating this to it's own page makes it harder to for example use the relationmanager.

zepfietje:
How would the breadcrumbs work when there's no index page for a resource though?
It wouldn't be clear that the page is scoped by the resource when you leave out the first breadcrumb, right?

danharrin:
So many features of a resource rely on an index page, it just doesn't make sense for a resource not to be able to "list" IMO

I'm more than happy to make it easier to add a relation manager to a custom page, that is reasonable

PR #8511, opened 2023-09-15
Workaround for apparent LW3/Alpine bug in section

cheesegrits:
This seems to be a regression / bug in LW3 / Alpine, which crept in to Filament between 3.0.31 and 3.0.32, when LW3 came out of Beta.  At which point 'collapsible' on Sections started breaking, with use of x-bind:class in more than one sub-element complaining about isCollapsed not being defined.

https://app.screencast.com/dUChDq6YKWgrE

https://discord.com/channels/883083792112300104/1151593778637983856

I confirmed the broken behavior.  It seems that simply moving the x-data where isCollapsed is defined outside of the Blade conditional fixes the problem.  I don't understand why, but it does.  I ran across several Guthub issue conversations on Alpine that solved similar issues this way.

With this change, collapsible sections work again.


markdynamo6:
I may have found a related issue: https://github.com/filamentphp/filament/discussions/12266

PR #8514, opened 2023-09-15
Default AboveContentCollapsible to hidden to avoid flickering

linaspasv:
Hello, there is a short duration flickering when freshly opening or refreshing list page where table uses ``FiltersLayout::AboveContentCollapsible``. My PR will make the initial loaded state of the filter hidden.

```php

use Filament\Resources\Pages\ListRecords as BaseListRecords;
use Filament\Tables\Enums\FiltersLayout;
use Filament\Tables\Table;

class ListRecords extends BaseListRecords
{
    {
        return parent::table($table)
            ->filtersLayout(FiltersLayout::AboveContentCollapsible);
    }
}

```

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -167,7 +167,7 @@
                     <x-filament-tables::filters
                         :form="$getFiltersForm()"
                         x-show="areFiltersOpen"
-                        style="display: none;"
+                        x-cloak
                         @class([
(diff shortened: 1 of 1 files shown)


danharrin:
Maybe replace with `x-cloak`?

linaspasv:
> Maybe replace with `x-cloak`?

Thank you for pointing out, changed. (not very familiar with alpine.js yet) :-)

PR #8518, opened 2023-09-15
Fix darkmode text color

Osknor:
The optional table heading did not look right in dark:mode so added tailwind classes to be in line with section heading.

Fieldset label also lacked text color classes, added the same ones there.

- [no] Changes have been thoroughly tested to not break existing functionality.
- [no] New functionality has been documented or existing documentation has been updated to reflect changes.
- [yes] Visual changes are explained in the PR description using a screenshot/recording of before and after.

table-before
table-after
field-before
field-after

### DIFF
--- a/packages/support/resources/views/components/fieldset.blade.php
+++ b/packages/support/resources/views/components/fieldset.blade.php
@@ -13,7 +13,7 @@
     @if (filled($label))
         <legend
             @class([
-                '-ms-2 px-2 text-sm font-medium leading-6',
+                '-ms-2 px-2 text-sm font-medium leading-6 text-gray-950 dark:text-white',
                 'sr-only' => $labelHidden,
             ])
         >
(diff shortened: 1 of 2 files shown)


danharrin:
Isn't this fixed by having sensible default text colors set on the `<body>`?

Osknor:
While that sounds totally valid and a solution, it was my naive experience that I flipped on dark mode found that the other filament components (that I had happen to use) translated over beautifully. So seemed in line to add this pattern of classes in the two places it seemed to have been overlooked.
I haven't looked closely enough to figure out if there was some reasoning why they were excluded here or if there were other components where they also logically could be added.

danharrin:
This would be caused by any element that has text but no text color set, as in the panel builder we do have a default color. Anyway, I don't think this hurts too much.

zepfietje:
This fix is fine indeed. I've added explicit text color to most elements, and probably just forgot these two.

PR #8525, opened 2023-09-15
Feature: add root class to panel

ghost:
This feature allows a developer to add a root class to the `<body>` so that pages can be more easily styled based on this root class. A default implementation has been added called `routeRootClass()` which slugifies the routename into a css class with lowercases and dashes always prefixed with `fi-`, like so:

- Route `filament.admin.auth.login` to `fi-filament-admin-auth-login`.
- Route `SomeCustom.Route.name` gets converted to `fi-some-custom-route-name`.

This class will directly be inserted into the `<body>` tag in `layout.base`, so it becomes:
```html
<body class="fi-body fi-filament-admin-auth-login ...">
```

The developer can now target this specific class to style only the login page, like creating a specific styling for a component that could be used on other pages as well:
```css
.fi-filament-admin-auth-login {
    .fi-simple-main {
        background: #1F3342 !important;
        color: #FFFFFF;
    }
}
```

## Usage
```php
{
    {
        return $panel
        // Use default class made from the route name.
        ->routeRootClass()
        ->rootClass(function () {
            // Custom callback for your own implementation.
        });
    }
}
```

## Possible issues
- There might be CSS collisions with simple route names like `login` which gets parsed to `fi-login` which could already be in use. Since this feature is opt-in and customizable I don't see an immediate issue.

## Checks

### DIFF
--- a/packages/panels/resources/views/components/layout/base.blade.php
+++ b/packages/panels/resources/views/components/layout/base.blade.php
@@ -90,7 +90,7 @@
     </head>
 
     <body
-        class="fi-body min-h-screen bg-gray-50 font-normal text-gray-950 antialiased dark:bg-gray-950 dark:text-white"
+        class="fi-body {{ filament()->getRootClass() }} min-h-screen bg-gray-50 font-normal text-gray-950 antialiased dark:bg-gray-950 dark:text-white"
     >
         {{ \Filament\Support\Facades\FilamentView::renderHook('panels::body.start') }}
 
--- a/packages/panels/src/FilamentManager.php
+++ b/packages/panels/src/FilamentManager.php
@@ -198,6 +198,11 @@ class FilamentManager
         return $this->getCurrentPanel()->getModelResource($model);
     }
 
+    public function getRootClass(): ?string
+    {
+        return $this->getCurrentPanel()->getRootClass();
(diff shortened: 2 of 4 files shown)


danharrin:
I don't know how much I like this to be honest, it feels a little weird. Like, resource pages already have these classes to customize them with:

Screenshot 2023-09-16 at 09 03 34

I'd be ok with adding this hard-coded class to the login page, if you want to PR that:

Screenshot 2023-09-16 at 09 04 21

But leaking the route name as a CSS class doesn't right to me, and also, I expect this doesn't work very well with `wire:navigate` (not sure if it swaps out the layout or not)

ghost:
@danharrin I've had a look at this but it wasn't an option because adding a class in that spot adds it to:
```
body > div > div > main > div
```
Which is scoped too deep. I'd like to style CSS classes that are a few levels higher which I cant target using CSS that way.

The same issue also applies to `fi-resource-*` classes which also prohibits you from styling `body > div` since the `fi-resource` class is too deep (however for my purposes I only needed to style the login page).

I don't see a direct issue with sharing a slugified route name in HTML code. I would be ok if the `routeRootClass()` would be removed so developers require their own implementation by default.

PR #8542, opened 2023-09-16
Fix truncated tab item badges

sakanjo:

Before:
Before

After:
After

### DIFF
--- a/packages/support/resources/views/components/badge.blade.php
+++ b/packages/support/resources/views/components/badge.blade.php
@@ -56,7 +56,7 @@
     @endif
 
     <span class="grid">
-        <span class="truncate">
+        <span>
             {{ $slot }}
         </span>
     </span>


zepfietje:
The `truncate` class is needed: https://github.com/filamentphp/filament/pull/8018.

Potentially a `w-max` class should be added to the tab item badge.
Could you update this PR, verify the fix, check if no other usages of badges break, and then mark this PR as ready for review, @sakanjo?

Edit: are these tabs in your screenshot by the way? Doesn't look like Filament's design.

sakanjo:
`w-max` is the way I guess, or maybe we should add `$attributes->mergeClass` to delete the `truncate` class when it's used in the tab badge, I'm not 100% sure.

It's filament's design, just a project for a customer.

zepfietje:
I tried applying `w-max` but it didn't solve it in this case.
The issue with the original approach in linked PR was that the badge would shrink and truncate the label when it shouldn't (in regular table columns for example).

Let me know if you can look into this and find a fix.
Else, we may need to rework the original truncate approach for the badge.

zepfietje:
Lol, I was probably still asleep yesterday; hadn't correctly linked to my local Filament repo.

Fixed in https://github.com/filamentphp/filament/commit/225337c6655a6bcc345b748cbf3824de92f08e8e.

PR #8544, opened 2023-09-16
Fixed docs

thethunderturner:

Currently by using this function in my custom Page:
```php
protected function getHeaderWidgetsColumns(): int | array
{
    return // ...;
}
```
as stated by the docs:
image

Will give the following error:
image

This is because `getHeaderWidgetsColumns()` in the Page is public and not protected. Therefore, should both functions be protected or both public?

### DIFF
--- a/packages/panels/src/Pages/Page.php
+++ b/packages/panels/src/Pages/Page.php
@@ -159,7 +159,7 @@ abstract class Page extends BasePage
     /**
      * @return int | string | array<string, int | string | null>
      */
-    protected function getHeaderWidgetsColumns(): int | string | array
+    public function getHeaderWidgetsColumns(): int | string | array
     {
         return 2;
     }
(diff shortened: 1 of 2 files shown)


danharrin:
Should be public, the docs are wrong

thethunderturner:
> Should be public, the docs are wrong

Okk, I will fix the PR  docs then.

packages/panels/src/Pages/Page.php
```diff
@@ -159,7 +159,7 @@ public function getVisibleHeaderWidgets(): array
     /**
      * @return int | string | array<string, int | string | null>
      */
-    public function getHeaderWidgetsColumns(): int | string | array
+    protected function getHeaderWidgetsColumns(): int | string | array
```
danharrin:
```suggestion
    public function getHeaderWidgetsColumns(): int | string | array
```

PR #8553, opened 2023-09-17
Fix: Respect isVisible() in NavigationItem

mabdelfattah:
It is already well documented, but for some reason was not working/implemented.

- [] Visual changes are explained in the PR description using a screenshot/recording of before and after.

### DIFF
--- a/packages/panels/resources/views/components/sidebar/group.blade.php
+++ b/packages/panels/resources/views/components/sidebar/group.blade.php
@@ -63,20 +63,22 @@
         class="fi-sidebar-group-items flex flex-col gap-y-1"
     >
         @foreach ($items as $item)
-            <x-filament-panels::sidebar.item
-                :active-icon="$item->getActiveIcon()"
-                :active="$item->isActive()"
-                :badge-color="$item->getBadgeColor()"
-                :badge="$item->getBadge()"
-                :first="$loop->first"
-                :grouped="filled($label)"
-                :icon="$item->getIcon()"
-                :last="$loop->last"
-                :url="$item->getUrl()"
-                :should-open-url-in-new-tab="$item->shouldOpenUrlInNewTab()"
-            >
-                {{ $item->getLabel() }}
-            </x-filament-panels::sidebar.item>
+            @if($item->isVisible())
+                <x-filament-panels::sidebar.item
+                    :active-icon="$item->getActiveIcon()"
+                    :active="$item->isActive()"
+                    :badge-color="$item->getBadgeColor()"
+                    :badge="$item->getBadge()"
+                    :first="$loop->first"
+                    :grouped="filled($label)"
+                    :icon="$item->getIcon()"
+                    :last="$loop->last"
+                    :url="$item->getUrl()"
+                    :should-open-url-in-new-tab="$item->shouldOpenUrlInNewTab()"
+                >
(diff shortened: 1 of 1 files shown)


danharrin:
Here's where we do it already:

https://github.com/filamentphp/filament/blob/3.x/packages/panels/src/Panel/Concerns/HasNavigation.php#L103

Doing this in the view is wrong, it will cause empty groups to appear

danharrin:
I believe the navigation stuff is actually all tested, so you can add a failing test if you want

mabdelfattah:
> Here's where we do it already:
> 
> https://github.com/filamentphp/filament/blob/3.x/packages/panels/src/Panel/Concerns/HasNavigation.php#L103
> 
> Doing this in the view is wrong, it will cause empty groups to appear

Thank you, for sharing this. The problem persists with:
```php
if ($this->navigationBuilder instanceof Closure) {
    return $this->buildNavigation();
}
```
https://github.com/filamentphp/filament/blob/b86fa17c0fe990ace6ed474bdc339918501235b7/packages/panels/src/Panel/Concerns/HasNavigation.php#L94

danharrin:
Ah ok, you can fix the `buildNavigation()` method instead

mabdelfattah:
Or should I fix it in `NavigationBuilder::getNavigation()`?

I can replace:
```php
NavigationGroup::make()->items($items),
```

with
```
NavigationGroup::make()->items(
    collect($items)
        ->filter(fn (NavigationItem $item): bool => $item->isVisible())
         ->sortBy(fn (NavigationItem $item): int => $item->getSort())
         ->all()
)
```

But, I'm uncertain whether this is what you prefer or not.


https://github.com/filamentphp/filament/blob/b86fa17c0fe990ace6ed474bdc339918501235b7/packages/panels/src/Navigation/NavigationBuilder.php#L75C17-L75C56

danharrin:
It's not just the ungrouped items that don't hide, right?

mabdelfattah:
> It's not just the ungrouped items that don't hide, right?

I'm not certain, did not try the grouped ones. My perspective is doing it on `NavigationBuilder::getNavigation()` to make sure the items are filtered and sorted before passing them to the groups. So, it should work on both grouped/ungrouped.

danharrin:
`NavigationGroup::make()->items($items),` is only handling ungrouped items there, the other groups are handled under that line

mabdelfattah:
So, should we move it to a lower level to `NavigationBuilder::items()`, and replace
```php
$this->items = [
            ...$this->items,
            ...$items,
        ];
```

with:
```php
$this->items = collect([
            ...$this->items,
            ...$items,
        ])->filter(fn(NavigationItem $item): bool => $item->isVisible())
            ->sortBy(fn(NavigationItem $item): int => $item->getSort())
            ->all();

```

https://github.com/filamentphp/filament/blob/b86fa17c0fe990ace6ed474bdc339918501235b7/packages/panels/src/Navigation/NavigationBuilder.php#L35

adampatterson:
> It's not just the ungrouped items that don't hide, right?

I know this is closed and in v3, but I am experiencing similar problems in v2. 

I was going to post this in the discord channel but thought I'd look around before I did so.

I have linked a gist with a couple of scenarios.

https://gist.github.com/adampatterson/75ec08f5938e0d26b851dd743726781c

PR #8557, opened 2023-09-17
Fix live predefined actions for repeater and builder field form

yassine-kessal:

Add a fix to refresh state when a predefined repeater field actions are triggered. When we add, remove, clone, reorder, move up/down an item the state will be refreshed

It will be useful if we want to update another field and need to know if an item was added, removed, ... etc.

Let me know if you want me to add a method in the repeater class to enable/disable this.


danharrin:
Please also do this for the Builder, so its consistent

yassine-kessal:
Hi @danharrin,

It's done :grinning:, i also update the title of the PR.

PR #8558, opened 2023-09-17
Add `brandLogo` API

ousid:

## Description

This __PR__ introduces the new `brandLogo` feature to Filament, enhancing the customization options for panel logos. With this feature, developers can easily set a custom logo path for their Filament panel, providing greater flexibility and branding possibilities.

## Changes Made

- Added the `brandLogo` method to the Filament panel configuration.
- Documented the usage of the `brandLogo` method in the project's documentation.

## Usage

   ```php
use Filament\Panel;

{
    return $panel
        // ...
        ->brandLogo('/path/to/your/logo.png');
}
```

## Related Resources
closes #8526

### CODE REVIEW

**Comment by mabdelfattah** [comment]
I like the idea, but I think it is better to also support the user to pass both alt, and class. So, the syntax could be something similar to:
```
use Filament\Panel;

{
 return $panel
     // ...
     ->brandLogo(src: '/path/to/your/logo.png', class: 'h-10', alt: 'visit home page of my brand');
}
```

Let me know if you want me to do these changes.

danharrin:
> I like the idea, but I think it is better to also support the user to pass both alt, and class. So, the syntax could be something similar to:
> 
> ```
> use Filament\Panel;
> 
> public function panel(Panel $panel): Panel
> {
>  return $panel
>      // ...
>      ->brandLogo(src: '/path/to/your/logo.png', class: 'h-10', alt: 'visit home page of my brand');
> }
> ```
> 
> Let me know if you want me to do these changes.

I disagree, HTML classes should be kept out of PHP code. Also, the alt text is already customisable with `brandName()`.

ousid:
> I like the idea, but I think it is better to also support the user to pass both alt, and class. So, the syntax could be something similar to:
> 
> ```
> use Filament\Panel;
> 
> public function panel(Panel $panel): Panel
> {
>  return $panel
>      // ...
>      ->brandLogo(src: '/path/to/your/logo.png', class: 'h-10', alt: 'visit home page of my brand');
> }
> ```
> 
> Let me know if you want me to do these changes.

I Agree with Dan. Besides, you have the ability to extend the `logo` component, and make it as you like to be.

Thank you for the suggestion @mabdelfattah

zepfietje:
> Besides, you have the ability to extend the `logo` component, and make it as you like to be.

Publishing vendor views is not recommended.
Styling the element by targeting the CSS hook class is the way to go.

zepfietje:
Please post a screenshot of the final UI here when done, so I can review it before merging.

danharrin:
> > Besides, you have the ability to extend the `logo` component, and make it as you like to be.
> 
> Publishing vendor views is not recommended. Styling the element by targeting the CSS hook class is the way to go.

This logo.blade.php is the exception, publishing that view is currently the way to change the logo

zepfietje:
> This logo.blade.php is the exception, publishing that view is currently the way to change the logo

Ah right, fair enough.
Though I think we'd change that advice after introducing this API?

danharrin:
Yes, the docs have been changed, but this is still not a harmful view to override, since it has no parameters

aliloubm:
Hi,

How to Switch Logo in Dark Mode ??

PR #8564, opened 2023-09-18
Fix type

noxoua:


zepfietje:
I believe `string` should still be supported, right, @danharrin?

Edit: or maybe not, lol. 🤔

danharrin:
> I believe `string` should still be supported, right, @danharrin?
> 
> Edit: or maybe not, lol. 🤔

This setting was first introduced as an enum, there was never a string version

zepfietje:
Yeah that's what got me confused haha.

PR #8591, opened 2023-09-19
Tables: BulkActions - Fix getSelectedTableRecords query to include filters.

JustinElst:
This way the thrashed records will be included in bulk actions if the TrashedFilter is applied.

### DIFF
--- a/packages/tables/src/Concerns/HasBulkActions.php
+++ b/packages/tables/src/Concerns/HasBulkActions.php
@@ -312,6 +312,8 @@ trait HasBulkActions
                 $column->applyRelationshipAggregates($query);
             }
 
+            $this->filterTableQuery($query);
+
             return $this->cachedSelectedTableRecords = $query->get();
         }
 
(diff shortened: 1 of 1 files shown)


danharrin:
This needs to only apply when `shouldDeselectAllRecordsWhenFiltered()` is true

JustinElst:
Thanks! I'll be looking into it, and updating the request.

JustinElst:
After your suggestion it still works as expected. There is no way that I can oversee and test all use cases. So thanks again (and your great work). Hopefully this will not break a thing..

packages/tables/src/Concerns/HasBulkActions.php
```diff
@@ -312,6 +312,10 @@ public function getSelectedTableRecords(): EloquentCollection
                 $column->applyRelationshipAggregates($query);
             }
 
+            if ($table->shouldDeselectAllRecordsWhenFiltered()) {
```
danharrin, line 315:
```suggestion
            if ($table->shouldDeselectAllRecordsWhenFiltered()) {
```

danharrin:
Thanks

PR #8596, opened 2023-09-19
Update 05-viewing-records.md

HelgeSverre:
If you don't pass `->record($this->record)` you get the following error:

```
Infolist has no [record()] or [state()] set.
```

 given the context of the page (a ViewRecord page), this should be added so people don't run Gto this problem, which is not easy to debug if you are following the "guide" for the first time.

Also in `ViewRecord` pages the `infolist()` should not be static

```
Cannot make non-static method ViewRecord->infolist(infolist: \Filament\Infolists\Infolist) static
```


danharrin:
> define an `infolist()` method on the resource class

Defining the method on the resource class instead of the page fixes these issues

PR #8607, opened 2023-09-20
Allow panels to be used for unauthenticated users.

AbdelElrafa:
The only thing stopping a Panel from being used as a public Panel without user authentication is the `user-menu` being required in the top bar. This adds an auth check to only render the `user-menu` component when a user is authenticated.

I saw your [comment](https://github.com/filamentphp/filament/discussions/6119#discussioncomment-5528346) that Panels aren't meant to be used for unauthenticated users, however I think this small change as well as the user removing the Authenticate middleware will allow the panel to be used for unauthenticated users. This will make the panels much more flexible and allow them to be used to build application experiences for unauthenticated users. My use case is to build a docs panel to ship as a plugin in a package. 

### DIFF
--- a/packages/panels/resources/views/components/topbar/index.blade.php
+++ b/packages/panels/resources/views/components/topbar/index.blade.php
@@ -131,7 +131,9 @@
                 @livewire(Filament\Livewire\DatabaseNotifications::class, ['lazy' => true])
             @endif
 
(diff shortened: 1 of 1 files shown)


CraftLogan:
I like the idea of using Filament Panels for anything, so thumbs up from me:)

PR #8620, opened 2023-09-20
Translatable plugin docs: making a model translatable

AmooAti:
Hi, 

I've added a section on making models translatable for those unfamiliar with `spatie/laravel-translatable`.

### DIFF
--- a/packages/spatie-laravel-translatable-plugin/README.md
+++ b/packages/spatie-laravel-translatable-plugin/README.md
@@ -43,27 +43,11 @@ public function panel(Panel $panel): Panel
 
 ## Make your model translatable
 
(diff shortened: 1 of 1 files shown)


danharrin:
Can we not, instead, link to Spatie's documentation? I'd rather not copy it

AmooAti:
That's a better way to do it. 
Anything that users know they need to do the spatie setup and then continue with filament. 
One of my co-workers, who didn't know about spatie itself, was stuck for 2 hours to set up this plugin. :D.
I'll do it ASAP.

AmooAti:
@danharrin 
I've changed it.

PR #8623, opened 2023-09-20
Ignore collapsible sidebar CSS when top navigation is enabled.

andrewdwallo:

Previous to this PR, the CSS for the collapsible sidebar would still be applied when a User has top navigation enabled. This resulted in the following.

**Before:**
Screenshot 2023-09-20 at 8 53 02 AM

The CSS applied for a collapsible sidebar should be ignored when a User has enabled top navigation. Beforehand, using `->sidebarFullyCollapsibleOnDesktop()` would not result in any issues, but using `->sidebarCollapsibleOnDesktop()` would result in issues. This PR improves this. 

**After:**
Screenshot 2023-09-20 at 8 53 25 AM

### DIFF
--- a/packages/panels/resources/views/components/sidebar/index.blade.php
+++ b/packages/panels/resources/views/components/sidebar/index.blade.php
@@ -10,7 +10,7 @@
 {{-- format-ignore-start --}}
 <aside
     x-data="{}"
-    @if (filament()->isSidebarCollapsibleOnDesktop() && ! filament()->hasTopNavigation())
+    @if (filament()->isSidebarCollapsibleOnDesktop() && (! filament()->hasTopNavigation()))
         x-cloak
         x-bind:class="
             $store.sidebar.isOpen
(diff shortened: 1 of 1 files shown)


danharrin:
I mean, no-one should use these settings together, they make no sense, but whatever

andrewdwallo:
That is understandable and I agree. But I think it would be good to prevent the issue out of negligence. I only came across the problem when I had dynamic settings where a user would enable top navigation while a collapsible sidebar was a fixed setting in my application.

danharrin:
If I were mean, I'd throw an exception when both are used together

andrewdwallo:
Well good thing you aren't mean 😃

zepfietje:
@danharrin, go for it and throw an exception, please.
Honestly, I didn't check for these cases as you'd never use both of them anyway.
Having to account for each and every combination only clutters the code and makes it harder to maintain in my opinion.

PR #8624, opened 2023-09-20
Constraint order values

sakanjo:
Setting `tableGroupingDirection` to any value will raise an exception, this PR makes sure it doesn't.

### DIFF
--- a/packages/tables/src/Concerns/CanGroupRecords.php
+++ b/packages/tables/src/Concerns/CanGroupRecords.php
@@ -44,6 +44,8 @@ trait CanGroupRecords
             return $query;
         }
 
-        return $group->orderQuery($query, $this->tableGroupingDirection ?? 'asc');
+        $orderDirection = $this->tableGroupingDirection === 'desc' ? 'desc' : 'asc';
+
+        return $group->orderQuery($query, $orderDirection);
     }
 }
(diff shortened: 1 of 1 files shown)


DariusIII:
This causes errors now when grouping is used. I have updated to v3.0.55 and started receiving error below.

`Filament\Tables\Grouping\Group::orderQuery(): Argument #2 ($direction) must be of type string, null given, called in /var/www/html/vendor/filament/tables/src/Concerns/CanGroupRecords.php on line 56`

I had to remove defaultGrouping() from table. Also tried with groups(), but on every group change same error pops out, until i click on ordering button once.

sakanjo:
@DariusIII i do encounter the same issue due to passing `null` to [Group.php#L306](https://github.com/filamentphp/filament/blob/3.x/packages/tables/src/Grouping/Group.php#L306),

@danharrin my commit should do the job as I've copied it from [CanSortRecords.php#L91](https://github.com/filamentphp/filament/blob/3.x/packages/tables/src/Concerns/CanSortRecords.php#L91), Can we re-open the PR ?

danharrin:
It was fixed in a release shortly after, please upgrade

DariusIII:
I have not noticed a new release after .55. I have updated to .56 and all is back to normal. Thanks!

PR #8626, opened 2023-09-20
Remove toggle column start margin

sakanjo:
Before:

Before

After:

After


zepfietje:
The toggle column didn't make it in the design review before releasing 3.0.
So thanks for catching and fixing this already, @sakanjo!

PR #8633, opened 2023-09-21
Add `wire:key` in forms select component

austincarpenter:

It would seem that the default behaviour with dependent selects has changed from Livewire v2 to v3. It may have been possible to get away without `wire:key` on `select` `option`s previously (perhaps handled by JS) however, now without them, existing functionality of Filament v2 projects breaks.

https://github.com/livewire/livewire/discussions/6882

Have added `wire:key` to `option` elements in the select component in the Forms namespace.

### DIFF
--- a/packages/forms/resources/views/components/select.blade.php
+++ b/packages/forms/resources/views/components/select.blade.php
@@ -61,6 +61,7 @@
                                 <option
                                     @disabled($isOptionDisabled($groupedValue, $groupedLabel))
                                     value="{{ $groupedValue }}"
+                                    wire:key="{{ $groupedValue }}"
                                 >
                                     @if ($isHtmlAllowed)
                                         {!! $groupedLabel !!}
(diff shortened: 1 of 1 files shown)


packages/forms/resources/views/components/select.blade.php
```diff
@@ -74,6 +75,7 @@ class="fi-fo-select"
                         <option
                             @disabled($isOptionDisabled($value, $label))
                             value="{{ $value }}"
+                            wire:key="{{ $value }}"
```
zepfietje, line 78: Pretty sure this would have to be scoped by the component state path.
Will leave this to @danharrin for a proper review.

danharrin:
@austincarpenter to be honest, I don't think we should have to do this. IMO, you should submit a failing test to Livewire, since this is a very very common use case that I think should be handled out of the box.

austincarpenter:
Agreed 😀 https://github.com/livewire/livewire/pull/6889

danharrin:
Ok, closing this then, thanks

PR #8640, opened 2023-09-21
Fix/make form action label translatable and add German translation for button and notification

michaelklopf:
- ~New functionality has been documented or existing documentation has been updated to reflect changes.~ not necessary
- ~Visual changes are explained in the PR description using a screenshot/recording of before and after.~ not necessary

We added the German translation for the save button and also add a function to make the button configurable similar to `getSavedNotificationTitle` above.

### DIFF
--- a/packages/panels/resources/lang/de/pages/tenancy/edit-tenant-profile.php
+++ b/packages/panels/resources/lang/de/pages/tenancy/edit-tenant-profile.php
@@ -1,0 +1,25 @@
+<?php
+
+return [
+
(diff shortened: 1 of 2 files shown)


danharrin:
Please remove `getSaveFormActionLabel`, you can just use

```php
protected function getSaveFormAction(): Action
{
    return parent::getSaveFormAction()
        ->label('Your label');
}
```

michaelklopf:
Didn't think of that. I removed it.

PR #8662, opened 2023-09-22
Fix broken cross-links in plugin docs

drbyte:
Without the preceding `/` the links were "relative", thus firing some fallback redirects where were unhelpful.


drbyte:
For the Alpine Component link correction in the panels doc file, please double-check that I've chosen the correct one. The prior link pointed to a non-existent destination.

zepfietje:
Thanks!

PR #8664, opened 2023-09-22
Update Panel.php for Laravel Octane

haringsrob:
See https://github.com/filamentphp/filament/issues/7969

### DIFF
--- a/packages/panels/src/Panel.php
+++ b/packages/panels/src/Panel.php
@@ -79,6 +79,16 @@ class Panel extends Component
         if ($callback = $this->bootUsing) {
             $callback($this);
         }
+
+        // In a Laravel Octane instance we have to make sure we rebuild the navigation on every
+        // request since it may require the current requests state. This is especially the case
+        // for multi-tenancy.
+        if (class_exists(\Laravel\Octane\Events\RequestReceived::class)) {
+            Event::listen(\Laravel\Octane\Events\RequestReceived::class, function () {
+                $this->navigationItems = [];
+                $this->isNavigationMounted = false;
+            });
+        }
     }
 
     public function bootUsing(?Closure $callback): static


danharrin:
Please explain why this fixes the problem, otherwise I don't really want to merge it. This is a scoped binding, the object should not be shared between requests, so I don't understand why it needs to be reset.

haringsrob:
Hey @danharrin afaik, service providers are only booted once with octane. So there will only be 1 panel instance (for each panel). It is not being rebuilt on a new request.

danharrin:
The panels are stored in a FilamentManager object, which is a scoped binding, which gets refreshed on every request

haringsrob:
As per `https://laravel.com/docs/10.x/octane#dependency-injection-and-octane`

```
Since Octane boots your application once and keeps it in memory while serving requests, there are a few caveats you should consider while building your application. For example, the register and boot methods of your application's service providers will only be executed once when the request worker initially boots. On subsequent requests, the same application instance will be reused.
```

This means, that it will only register the panel providers once. Not for every request. It is strange then that this does work at all given that the filamentManager is scoped?

If the panels would be in config, and on init the manager registers them it would be different I guess?

------

Edit: I have tried various other strategies to resolve this but none of it seems to work.

------

So in the end, I believe, even if there is only 1 manager, there is also only one instance of the panel objects there. They may need some context flushed on every request so it takes into account the new tenant if any.

danharrin:
Okay, I'm going to have to investigate deeper myself as to why the panel objects are persisted then, hope you understand the importance of that.

haringsrob:
Hey Dan, no issue. I can patch it for now using this pr. I hope you can find maybe a better/more generic solution.

binaryfire:
Hey @haringsrob

I tried the same thing in this PR: https://github.com/filamentphp/filament/pull/8448. It does fix resource navigation links but it breaks [custom navigation items](https://filamentphp.com/docs/3.x/panels/navigation#registering-custom-navigation-items). They disappear after the first request. If you figure out a way to solve that, please let me know! And hopefully Dan can figure out a more permanent solution at some stage.

linaspasv:
@danharrin I can confirm there is a problem with navigation not updating between requests when octane is being used. I was trying to understand  why https://github.com/stechstudio/filament-impersonate plugin is not working, as I got policies for different roles that hide some of menu items for different users but with octane navigation is never updated. The provided PR seems to fix the problem but I am not sure if there are no other side effects.

linaspasv:
After some analysis, it's likely the main cause of the issue is that ``PanelProvider`` register method is called on a first request when running with octane only. Panel configuration is loaded but the problem is that Panel is never re-rendered again and stuck with the ``isNavigationMounted = true`` and the code below never gets executed...

```php
        foreach ($this->getPages() as $page) {
            $page::registerNavigationItems();
        }

        foreach ($this->getResources() as $resource) {
            $resource::registerNavigationItems();
        }
```

I am not sure about the elegant solution but this looks like a design flaw that is incompatible with the octane.

mtbossa:
Any updates? As @binaryfire said, this fixes the tenant route issue, however, it breaks custom navigation items. I've been trying to find a fix, but would need some help.

mateusgalasso:
I've been having problems with Octane and the filament sidebar too. My problem is access, some users can see some menus and others cannot, when a user logs out of the machine and another logs in we have the problem of a user with less access, seeing all the menus that they could not see. Without Octane the problem does not happen

Temporary solution is to use customizable navigation items

```php
 NavigationItem::make('Business Unit')
                    ->url('/admin/business-units')
                    ->icon('heroicon-m-building-storefront')
                    ->visible(fn (): bool => auth()->user()->hasPermissionTo('view_any_business::unit')),
```

danharrin:
Replaced by #9875, I have written notes to detail the solution

PR #8671, opened 2023-09-22
Support guest users

Quadrubo:
This PR introduces the possiblity of guests accessing your panels. Users don't have to be authenticated to visit the panel but authentication is still possible.

If the PR is wanted I will:

- [ ] Add missing translations
- [ ] Add documentation

## Problems

- [x] Email verification
- [ ] Language string duplication
- [ ] Menu looks bad, any alternative design ideas?

### Email verification

~~`emailVerification(isRequired: false)` has to be set because the underlying Middleware `\Illuminate\Auth\Middleware\EnsureEmailIsVerified` is not meant for allowing unauthenticated users.~~

~~The middleware should not be applied when the current user is `null` and the panel allows guests. Any tips on where to implement this?~~

This now applies a seperate middleware incase `allowGuests` is true which lets unauthenticated users through.

### Language string duplication

Language strings are currently being duplicated as the strings for "Sign in" and "Sign up" already exist. Should I use the existing strings from the `Login` / `Register` page or should the strings be duplicated to the other files.

## Screenshots

- widget when unauthenticated & guest menu
grafik

- guest menu
grafik

- guest menu on simple pages
grafik

### DIFF
--- a/packages/panels/resources/lang/de/layout.php
+++ b/packages/panels/resources/lang/de/layout.php
@@ -6,6 +6,10 @@ return [
 
     'actions' => [
 
+        'login' => [
+            'label' => 'Anmelden',
+        ],
+
         'logout' => [
             'label' => 'Abmelden',
         ],
@@ -14,8 +18,16 @@ return [
             'label' => 'Benachrichtigungen öffnen',
         ],
 
+        'open_guest_menu' => [
+            'label' => 'Gastmenü',
+        ],
+
         'open_user_menu' => [
             'label' => 'Benutzermenü',
+        ],
+
+        'register' => [
+            'label' => 'Registrieren',
         ],
 
         'sidebar' => [
--- a/packages/panels/resources/lang/de/widgets/account-widget.php
+++ b/packages/panels/resources/lang/de/widgets/account-widget.php
@@ -4,8 +4,16 @@ return [
 
     'actions' => [
 
+        'login' => [
+            'label' => 'Anmelden',
+        ],
+
         'logout' => [
             'label' => 'Abmelden',
+        ],
+
+        'register' => [
+            'label' => 'Registrieren',
         ],
 
     ],
--- a/packages/panels/resources/lang/en/layout.php
+++ b/packages/panels/resources/lang/en/layout.php
@@ -22,7 +22,7 @@ return [
             'label' => 'Open notifications',
         ],
 
-        'open_user_menu' => [
+        'open_guest_menu' => [
             'label' => 'Guest menu',
         ],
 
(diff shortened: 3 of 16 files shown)


wychoong:
why would you want this feature? as panel is for managing resources which should be only allowed to access after authentication? are you trying to reuse navigation builder or something else?

Quadrubo:
> why would you want this feature? as panel is for managing resources which should be only allowed to access after authentication? are you trying to reuse navigation builder or something else?

As seen in #4400 many users including myself also build their entire app within filament. I think needing users to look at your app without registering is a pretty basic requirement.

wychoong:
> > why would you want this feature? as panel is for managing resources which should be only allowed to access after authentication? are you trying to reuse navigation builder or something else?
> 
> As seen in #4400 many users including myself also build their entire app within filament. I think needing users to look at your app without registering is a pretty basic requirement.

doing a quick search there is only 1 discussion opened related to Guest usage, so it may not be as common as you thought using admin panel without login. 

table & form builder, action, notification already can support using most of the features admin panel provides in livewire app. And just looking at this PR, there are a lot of if-else here and there which may adds complexity to support this. so if you can share more about the usage maybe there is a better way to implement it, specifically what is stopping you to implement this as a plugin package?

ousid:
I'm intrigued! Having read the preceding discussion, I still find myself pondering. Why would anyone allow users to access the dashboard without requiring authentication? Are there any real-life examples of this feature being actively used?

As mentioned by Dan in the aforementioned discussion:

> They enjoy the admin panel's CRUD features and prebuilt authentication and layout. Their users have access to the dashboard...

From my perspective, the admin panel was created for a specific purpose, which is to manage the application's resources and authenticate users.

Considering that you're hiding all the features from guest users, such as the menu, profile, etc., and only offering the Login and Register options, I question whether the majority of users would actually utilize this feature.

In reality, the majority of users tend to interact with the frontend, which utilizes the guest layout, to showcase the features of their applications. Therefore, the key question here is: Is there a compelling reason for this particular feature?

Quadrubo:
I mean it's called a Panel Builder now. It's not an admin panel anymore. I don't know why panels shouldn't provide a way for unauthenticated users to view resources.

I'm using the panel builder for my front facing application and need users to be able to see it. I know that I could more or less recreate the functionality with the existing packages but it's easier to do so using the panel.

This also provides the added benefit of using panel specific plugins like, for example, the panel switch which I need in my application. Authenticated users will just be able to do more stuff than unauthenticated users which only need to view the resources. With this implementation I don't have to duplicate my application logic when the panels plugin provides basically everything I need in the exact layout I want.

I think deciding if this functionality is within the scope of filament or not should be up to the maintainers.

Quadrubo:
Closued in favor of #8746 and a custom plugin.

Quadrubo:
#8908 is my new attempt at implementing this. Wanted to notify you guys and hope I could address your concerns.

PR #8677, opened 2023-09-22
Update 04-pages.md

babalolajnr:
When the example used in the documentation is used. It works for all other pages except the Edit page. This is a workaround I found and I think it's worth documenting to prevent headache for users that might encounter the same issue in the future.


danharrin:
Thanks, but I am not going to merge this, as since resources use models, and models have policy classes, this logic belongs in a policy class, rather than on the page itself.

PR #8682, opened 2023-09-22
Fix console errors when adding filters to grouped records

archilex:
This maybe another LW/$watch issue, but when you are using grouped records and then apply a filter, the console explodes with errors since the original elements are no longer present. You can see the errors in the demo app on the [Order page](https://demo.filamentphp.com/shop/orders?tableGrouping=created_at&tableGroupingDirection=asc) and then clicking on the `Trashed Filter` and selecting `Only deleted records`. 

This PR just returns the empty array when the record is no longer in the dom.

### DIFF
--- a/packages/tables/resources/js/components/table.js
+++ b/packages/tables/resources/js/components/table.js
@@ -61,8 +61,12 @@ export default function table() {
             this.isLoading = false
         },
 
-        getRecordsInGroupOnPage: function (group) {
+        getRecordsInGroupOnPage: function (group) {            
             const keys = []
+
+            if (this.$root === undefined) {
+                return keys
+            }
 
             for (let checkbox of this.$root.getElementsByClassName(
                 'fi-ta-record-checkbox',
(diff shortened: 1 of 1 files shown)


packages/tables/resources/js/components/table.js
```diff
@@ -61,9 +61,13 @@ export default function table() {
             this.isLoading = false
         },
 
-        getRecordsInGroupOnPage: function (group) {
+        getRecordsInGroupOnPage: function (group) {
```
valpuia: Why extra space?

archilex: Figured code style actions would have taken care of that.

zepfietje: Unfortunately we can't run code style CI and commit to forks.
But yeah, it would run after merge.

danharrin:
Thanks, but I'd rather not merge Livewire fixes into Filament, over time it will just pollute the codebase. Which is why I haven't fixed the watchers in the form fields yet. Luckily, these console errors do not affect the function of the application, as far as I know.

PR #8740, opened 2023-09-26
Introduce a way to set the priority between view and edit for the record url 

lamberttraccard:

This PR introduce a way to set the priority between the view and edit action for the record url.

There are cases when we don't necessarily want to default to the view page when the user has the view and edit privilege.


lamberttraccard:
hmm test are failing, can't we use laravel Arr helpers inside the codebase ?

danharrin:
This would be a breaking change for some applications that expect the behaviour even if they have no action. You can fix this in your own code by setting a `recordUrl()` on the table.

lamberttraccard:
Hey @danharrin, I understand your concern, but it can be quite cumbersome to set the recordUrl() on each table. Would you be open to adding a new specific method, such as actionRecordPriority, or something similar? This way, it wouldn’t be a breaking change, and it would allow users to define which action should take precedence. Let me know your thoughts!

danharrin:
I don't think I would be open to that, no. You could add a macro to the Table object that does the getUrl() for you

PR #8746, opened 2023-09-27
Support custom email verification middleware

Quadrubo:
In #8607 first steps have been made to allow unauthenticated users access to the panel. However it still wasn't possible to have email verification forced for authenticated users while still allowing unauthenticated users.

This PR provides the possibility to change the email verification middleware with the optional parameter `middlewareName`. One could then implement custom logic for letting unauthenticated users through.

### DIFF
--- a/packages/panels/src/Panel/Concerns/HasAuth.php
+++ b/packages/panels/src/Panel/Concerns/HasAuth.php
@@ -18,6 +18,8 @@ use Illuminate\Support\Facades\URL;
 
 trait HasAuth
 {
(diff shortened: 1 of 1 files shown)


ziming:
I support this PR because we renamed the email verified middleware to email.verified on our site to more clearly indicate it is email verification (as we also have mobile phone verification too)

As a side note, maybe filament can support a banner kind of email verified too, like show a jetstream like banner at top of panel to to encourage user to verify email with a link to click rather than hard enforce email verification

```blade
<x-banner message="Please verify your email by clicking here" href="/email/verify" />
```

PR #8754, opened 2023-09-27
Fix an SQL error for Postgres on JSON field.

sanis:
We are using Postgres database with the Filamentphp. However, after I login into the admin area I'm receiving an error caused by Json field (see screenshot).
image

After a change query should be formatted in the correct way and error isn't thrown anymore:
image

EDIT: I have accidentally taken screenshots of "different" queries, but codebase is still the same. One is the count of the same query results, another fetches results instead. However the difference I guess is pretty clear here.

### DIFF
--- a/packages/notifications/src/Livewire/DatabaseNotifications.php
+++ b/packages/notifications/src/Livewire/DatabaseNotifications.php
@@ -85,7 +85,7 @@ class DatabaseNotifications extends Component
     public function getNotificationsQuery(): Builder | Relation
     {
         /** @phpstan-ignore-next-line */
-        return $this->getUser()->notifications()->where('data->format', 'filament');
+        return $this->getUser()->notifications()->whereJsonContains('data', ['format' => 'filament']);
     }
 
     public function getUnreadNotificationsQuery(): Builder | Relation


wychoong:
please provide info for your postgres version and have you tested this change on mysql/mariadb?

sanis:
Yes I did test it, attaching the query which happens after the change on MYSQL:
image


I'm using postgres 16.

wychoong:
this seems to be setup issue as by default Laravel's notification table is defined as `->text('data')`, and from the [doc](https://laravel.com/docs/10.x/notifications#database-prerequisites) it is obviously intended to use with json functions and query builder should have handle `data->format` correctly. 

so to me it seems not filament or laravel bug, and more to your DB config. check this
https://laracasts.com/discuss/channels/laravel/undefined-function-7-error-operator-does-not-exist-text-unknown

sanis:
Honestly, I don't think that everyone will check database field type. 

Of course changing it to JSON solves the problem, however, how many users will know that, who will use default filament and laravel out of the box?

Having additional migration or casting the data to JSON should be a solution which would work for most of the cases.

zepfietje:
> Of course changing it to JSON solves the problem, however, how many users will know that, who will use default filament and laravel out of the box?

The docs contain a note about this: https://filamentphp.com/docs/3.x/notifications/database-notifications#overview.

wychoong:
> how many users will know that, who will use default filament and laravel out of the box

maybe submit a PR to laravel for the change to migration then

PR #8756, opened 2023-09-27
Documenting Middleware-based Authentication for Restricting Panel Access

ccc-developer:
This PR extends the Filament documentation to detail the implementation of authentication middleware, enabling restricted access to specific panels based on user roles or attributes.

### DIFF
--- a/packages/panels/docs/08-users.md
+++ b/packages/panels/docs/08-users.md
@@ -219,6 +219,72 @@ public function panel(Panel $panel): Panel
 }
 ```
 
+### Limiting Panel Access Using Middleware:
+
+1. **Extend the User Model**:
+
+   Introduce a flag, such as `is_owner` or `is_admin`, to your User model. This flag can differentiate between user types or roles.
+
+   ```php
+   protected $fillable = [
+       ...,
+       'is_owner',
+   ];
+   ```
+
+2. **Middleware Creation**:
+
+   Utilize the Artisan CLI to craft a new middleware:
+
+   ```bash
+   php artisan make:middleware VerifyIsOwner
+   ```
+
+3. **Define Middleware Logic**:
+
+   Inside the generated middleware file, set up the access logic based on the user's attribute:
+
+   ```php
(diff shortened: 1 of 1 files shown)


ccc-developer:
This is my first pull request, so happy to learn if I need to do something differently in the future. I'm also new to FilamentPHP v3 and struggled creating multiple panels and limiting access. After figuring out what to do I decided to document it for future me (and others). I hope this suggestion is helpful and can be incorporated into the documentation in some way. Thanks! Donnell

wychoong:
hey, maybe you missed out this part of the documentation https://filamentphp.com/docs/3.x/panels/installation#deploying-to-production
which does the similar thing. 

if you want panel specific control, you can use `$panel->getId()` to identify the current panel and add your logic
else this Pr would be better suit as tips/tricks

ccc-developer:
Hi @wychoong, are suggesting something like this instead of a custom middleware that my PR suggests?

`public function canAccessPanel(Panel $panel): bool
    {
        if ($panel->getId() === 'owner'){
            return $this->is_owner === 1;
        } return true;
    }`

wychoong:
> Hi @wychoong, are suggesting something like this instead of a custom middleware that my PR suggests?
> 
> `public function canAccessPanel(Panel $panel): bool { if ($panel->getId() === 'owner'){ return $this->is_owner === 1; } return true; }`

that should do the same as per your PR doc. 

I'm not suggesting which way are better because that is up to the dev (you). I'm just saying that what you wrote would be better fit as tip/tricks instead of in the main doc https://github.com/filamentphp/filamentphp.com#submitting-an-article-to-the-community-section

ccc-developer:
@wychoong I'm happy to write up my PR as an article for the tips/tricks section. Do I close, delete, orremove this PR? Or do the mods do that for me?
---
(Though I do think the section of Panels could include more details on how to limit access to panels either through Middleware or via the canAccessPanel function, FWIW). :-)

zepfietje:
As @wychoong said, the `canAccessPanel()` API is designed for this.

PR #8759, opened 2023-09-27
Fix link Blade component docs

RuliLG:
Hey!
Just a quick, small typo to make the docs a bit better. In this case, since we are in the link component page, I think the title should state how to use it as a button instead of as a link, since it's already a link.

Keep up the good work! :-)


zepfietje:
Good catch, probably a copy-paste error from the button component.
Thanks, @RuliLG! 🙏

PR #8766, opened 2023-09-28
Allow html in table empty state description

binaryfire:
Small change that allows HTML in table empty state descriptions eg:

image

It's fine to not escape this since it's not derived from user input.


binaryfire:
Closing since `->emptyStateDescription(new HtmlString('html goes here'))` works too. Thanks @awcodes

PR #8790, opened 2023-09-29
Fix error with morphing toggled relation manager tabs

### LINKED ISSUE #8788: Issue when toggling a resources relationmanager
### Package

filament/filament

### Package Version

v3.0.62

### Laravel Version

v10.25.2

### Livewire Version

v3.0.5

### PHP Version

PHP 8.2.5

### Problem description

Having two relationmanagers on a resource, one with canViewForRecord(), controlled by an  $ownerRecord attribute, e.g. a boolen toggle.

switch off the toggle,save record -> all relationmanagers disappear with js error : "Uncaught Snapshot missing on Livewire component with id: cdUvjw37Lpwhz4xgoRYE"

switch on the toggle, save record -> tablist nav item is shown, but relationmanager tabpanel has empty livewire component, also js error: "Uncaught Snapshot missing on Livewire component with id: JBybPURsFHmsJocGGswn"

### Expected behavior

tabbed relationmanagers to appear when controller attribute is set, or remove tabbed manager and show single relationmanager when attribute is unset

### Steps to reproduce

add Toggle to a form of a resource with two relationmanagers

```
public static function getRelations(): array
{
        return [
            RelationManagers\AddressesRelationManager::class,
            RelationManagers\PaymentsRelationManager::class,
        ];
}
```

```
Forms\Components\Toggle::make('has_payments')
```

define canViewForRecord method on one of the managers

```
public static function canViewForRecord(Model $ownerRecord, string $pageClass): bool
    {
        return $ownerRecord->has_payments;
    }
```

### Reproduction repository

https://github.com/tvdr/filamentdemo/tree/tabbed_relationmanager_fix

### Relevant log output

_No response_

tvdr:
This PR fixes the bug described in #8788
Defining a key to the element will help livewire/alpine to morph the dom as expected.

After hiding a currently active manager, the activeRelationManager is set to the first item.

Tested with all possible relationmanager configuration what i could think of (with groups,  with hasCombinedRelationManagerTabsWithContent, with associative array keys,with more than 2 relationmanagers .. )

some inline comment describing the activemanager reset case

### DIFF
--- a/packages/panels/resources/views/components/resources/relation-managers.blade.php
+++ b/packages/panels/resources/views/components/resources/relation-managers.blade.php
@@ -18,14 +18,6 @@
 
             return $manager;
         };
-
-        //when previously active manager is hidden, reset tabs
-        //but not when hasCombinedRelationManagerTabsWithContent and active tab is the form
-        if ($activeManager !== "" && !isset($managers[$activeManager])){
-            $activeManager = strval(array_key_first($managers));
-            $this->activeRelationManager = strval(array_key_first($managers));
-        }
-
     @endphp
 
     @if ((count($managers) > 1) || $content)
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks. I feel like it's kinda weird to be setting public properties in Livewire from the view though. Can we use a hook or something, maybe `rendering()`, to make sure that `$activeRelationManager` is valid?

We have a HasRelationManagers trait, so you could add a `renderingHasRelationManagers()` method to it that should be called at the correct time.

Since this will require a rewrite, I will close this PR and you can reopen it when the changes have been made. Thanks!

danharrin:
Actually gonna reopen this so I can merge the `wire:key`, but a separate PR can be used for the  `renderingHasRelationManagers()` thing

PR #8791, opened 2023-09-29
Add dropdown indicator to topbar items

ilpav12:
A dropdown indicator has been added to the topbar items, providing a more intuitive interface for users. This changes the UX to better indicate when a topbar item can be interacted with as a dropdown. A conditional is added to check if the item URL exists, if it doesn't the dropdown indicator is shown.

Before the update:
image
image

After the update:
image
image

### DIFF
--- a/packages/panels/resources/views/components/topbar/item.blade.php
+++ b/packages/panels/resources/views/components/topbar/item.blade.php
@@ -52,7 +52,7 @@
             </x-filament::badge>
         @endif
 
-        @if(! $url)
+        @if (! $url)
             <x-filament::icon
                 icon="heroicon-m-chevron-down"
                 @class([
--- a/packages/panels/resources/views/components/topbar/item.blade.php
+++ b/packages/panels/resources/views/components/topbar/item.blade.php
@@ -55,8 +55,9 @@
         @if (! $url)
             <x-filament::icon
                 icon="heroicon-m-chevron-down"
+                icon-alias="panels::topbar.item.toggle-button"
                 @class([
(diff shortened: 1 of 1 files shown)


zepfietje:
Not sure if we should add this, as it takes up more space in the navigation menu.
Just commenting now to make sure Dan and I discuss this before potentially merging.

packages/panels/resources/views/components/topbar/item.blade.php
```diff
@@ -50,5 +51,16 @@
                 {{ $badge }}
             </x-filament::badge>
         @endif
+
+        @if(! $url)
```
danharrin:
```suggestion
        @if (! $url)
```

packages/panels/resources/views/components/topbar/item.blade.php
```diff
@@ -28,6 +28,7 @@
             'flex items-center justify-center gap-x-2 rounded-lg px-3 py-2 text-sm font-semibold outline-none transition duration-75 hover:bg-gray-50 focus:bg-gray-50 dark:hover:bg-white/5 dark:focus:bg-white/5',
             'text-gray-700 dark:text-gray-200' => ! $active,
             'bg-gray-50 text-primary-600 dark:bg-white/5 dark:text-primary-400' => $active,
+            'flex items-center gap-x-1' => ! $url,
```
zepfietje: There's already `flex`, `items-center` and `gap-x-2` classes on the element on line 28.

zepfietje (changes requested):

PR #8792, opened 2023-09-30
Fix autofocus issue due to hidden main container

### LINKED ISSUE #8783: Autofocus issue with form fields
### Package

filament/forms

### Package Version

v3.0.62

### Laravel Version

10.25.1

### Livewire Version

_No response_

### PHP Version

8.2.4

### Problem description

I set the autofocus, it works in the first case. But when I refresh the page, it doesn't work anymore.

### Expected behavior

No autofocus on form field after multiple page refreshes.

### Steps to reproduce

Use the autofocus option.

### Reproduction repository

https://github.com/gergo85/filament-test-autofocus

### Relevant log output

_No response_

tvdr:

This PR is a possible solution to fix the bug mentioned in #8783

fi-main-ctn had a hidden class, hence it is not a [focusable area](https://html.spec.whatwg.org/multipage/interaction.html#focusable-area)

with opacity, the container won't flicker on pageload, but the contents are focusable

### DIFF
--- a/packages/panels/resources/views/components/layout/index.blade.php
+++ b/packages/panels/resources/views/components/layout/index.blade.php
@@ -21,21 +21,21 @@
                 x-bind:class="{
                     'fi-main-ctn-sidebar-open': $store.sidebar.isOpen,
                 }"
-                x-bind:style="'display: flex'" {{-- Mimics `x-cloak`, as using `x-cloak` causes visual issues with chart widgets --}}
+                x-bind:style="'display: flex; opacity:1;'" {{-- Mimics `x-cloak`, as using `x-cloak` causes visual issues with chart widgets --}}
             @elseif (filament()->isSidebarFullyCollapsibleOnDesktop())
                 x-data="{}"
                 x-bind:class="{
                     'fi-main-ctn-sidebar-open': $store.sidebar.isOpen,
                 }"
-                x-bind:style="'display: flex'" {{-- Mimics `x-cloak`, as using `x-cloak` causes visual issues with chart widgets --}}
+                x-bind:style="'display: flex; opacity:1;'" {{-- Mimics `x-cloak`, as using `x-cloak` causes visual issues with chart widgets --}}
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks

andrewdwallo:
@danharrin This causes the entire `fi-main-ctn` to quickly disappear and reappear (flicker) on page reloads. I don't think this is the best solution for the original issue mentioned here. I am surprised I just noticed this actually because it is pretty obvious.

andrewdwallo:
Here are videos from before this PR and after.

### Before PR

https://github.com/filamentphp/filament/assets/104294090/2885b34f-aa4a-4a19-bb46-2d7f2c0044a5

### After PR

https://github.com/filamentphp/filament/assets/104294090/ba3a140a-28a1-47cc-972f-98fce8f747b4

danharrin:
What is the alternative? The autofocus thing is an actual bug, whereas this is more of an inconvenience IMO. Would appreciate @zepfietje's input to see if he agrees with me or not.

zepfietje:
I think I need more context on why the main container content gets hidden in the first place, @danharrin.

andrewdwallo:
I definitely agree that the autofocus issue should be a priority, but I also believe that everything should be done to fix this "inconvenience"... It just doesn't look appealing, in my opinion, based on the video I showed so I figured that it would be best to mention it.

andrewdwallo:
Based on the comments in the code, I wanted to ask, what "visual issues" occur with chart widgets when using `x-cloak`?

danharrin:
The view has comments, it replaces x-cloak which has issues with Chart widgets

zepfietje:
But why do we need x-cloak on the main content?

andrewdwallo:
Maybe I am wrong but it seems that `x-cloak` was needed because without it, this happens.

https://github.com/filamentphp/filament/assets/104294090/620bae82-357d-4fc4-848e-eb0924bd59af

And I suppose, since `x-cloak` causes issues with chart widgets, using `hidden transition-all` was used instead, but then this caused issues with autofocus on form fields.

danharrin:
So essentially, the content is currently being hidden when it is flashing in

PR #8810, opened 2023-10-01
Ignore UUID `id` fields in form generation

paulohenriquesg:
When the table has a UUID `id` primary field, the form is generated with that field as a text input. 

It should be ignored in the same way the auto increments are.

### DIFF
--- a/packages/forms/src/Commands/Concerns/CanGenerateForms.php
+++ b/packages/forms/src/Commands/Concerns/CanGenerateForms.php
@@ -32,8 +32,7 @@ trait CanGenerateForms
             $columnName = $column->getName();
 
             if (str($columnName)->is([
-                'id',
-                'uuid',
+                app($model)->getKeyName(),
                 'created_at',
                 'deleted_at',
                 'updated_at',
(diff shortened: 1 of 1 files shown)


saade:
Not everyone changes its primary auto-incrementing column to a uuid, most people instead, uses a separated column `uuid` to expose it to the frontend, but keeps using the original `id` column for relationships and internal reference.

I think you should check for wether the column has the name `uuid` and/or the type `UUID` too

danharrin:
@saade simplified this by just using the model's primary key name for now

francoism90:
@danharrin Couldn't you use ` getRouteKey()` instead?

This simply uses the value used for routing, with a fallback to getKey() (if I'm not mistaken).

I use this already to overrule the Livewire logic: https://github.com/francoism90/hub-api/blob/main/src/Support/Livewire/ModelSynth.php

I really dislike exposing the model ID, so it would be great to force the usage of the route-key instead.

danharrin:
This isn't exposing the ID, this is just generating the form? You can still change it. IMO it doesn't feel right to use the route key, it isn't anything related to routing.

PR #8831, opened 2023-10-02
Configurable wire loading delay

### LINKED ISSUE #8408: UI feels sluggish on high latency connections because of fixed `wire:loading.delay`
### Package

filament/filament

### Package Version

3.0.46

### Laravel Version

10.2.6

### Livewire Version

3.0.1

### PHP Version

8.2

### Problem description

@danharrin First of all, thanks for your hard work on v3!

The default value of `wire:loading.delay` is 200ms which is ok for users close to the server, since most requests complete before that. However users on the other side of the globe have an additional 300-400ms latency on top of each request. We're in Australia and are developing a resource-intensive app which needs dedicated servers. The only affordable ones are in the US / Europe which means we have to factor that latency into how we build our UIs. Have a look at the [ping times from Sydney](https://wondernetwork.com/pings/Sydney) and you'll see what I mean.

The result is that we get loading indicators for every request Filament makes (which is great) but they don't appear for 200ms (not so great). It makes the whole app feel very sluggish. Our current VueJS dashboard feels snappy by comparison because it displays indicators immediately (even though it takes about the same amount of time to return data).

### Expected behavior

It'd be great if we could configure this behavior in `config/filament.php`. Livewire has [multiple values for delays](https://laravel-livewire.com/docs/2.x/loading-states#delaying-loading). So maybe:

- `'wire_loading_delay_disable' => true` Removes `wire:loading.delay` from components

- `'wire_loading_delay' => 'shortest'` Sets it to `wire:loading.delay.shortest`
- `'wire_loading_delay' => 'short'` Sets it to `wire:loading.delay.short`
.... etc

And then do something like this to make it reusable?

```
    public function getWireLoadingDelayAttribute(): string
    {
        if (config('filament.wire_loading_delay_disable')) {
            return '';
        }

        $delayValue = config('filament.wire_loading_delay', '');
        return empty($delayValue) ? 'wire:loading.delay' : 'wire:loading.delay.' . $delayValue;
    }

```

### Steps to reproduce

If testing from US or the UK, just launch any Filament app on a VPS in Australia then click around.

### Reproduction repository

https://github.com

### Relevant log output

_No response_

binaryfire:
Adds configurable `wire:loading.delay` times as discussed in https://github.com/filamentphp/filament/issues/8408.

This PR depends on my Livewire PR, which was just accepted: https://github.com/livewire/livewire/pull/6953. So this shouldn't be merged until the next Livewire release.

The difference this makes to UX on high latency connections can't be understated. Setting this to 'none' on a European server makes the app feel super responsive from Australia.

I wasn't sure how much detail to go into in the config file. I only mentioned 'none' and 'default', but the full list of allowed values is:

`none`   0ms
`shortest`   50ms
`shorter`   100ms
`short`   150ms
`default`   200ms
`long`   300ms
`longer`   500ms
`longest`   1000ms

The new values have been [added to the docs](https://github.com/livewire/livewire/pull/6953/commits/d0e555058501b8cbc469dcb658f18fb89c70885e#diff-425b56b15147fe67a97afc921ece67e715bb6f6100017abf329d288359520dd2) so it might be worth including a link to them in the comments.

The PR is backwards compatible because `default` is used if the config entry isn't set


packages/support/resources/views/components/button/index.blade.php
```diff
@@ -221,8 +221,14 @@
...
-                    :wire:target="$loadingIndicatorTarget"
+                    :attributes="
+                        \Filament\Support\prepare_inherited_attributes(
+                            new \Illuminate\View\ComponentAttributeBag([
+                                'wire:loading.delay.' . config('filament.wire_loading_delay', 'default') => '',
+                                'wire:target' => $loadingIndicatorTarget,
+                            ])
+                        )
+                    "
                     :class="$iconClasses"
```
zepfietje: If you're using `:attributes`, please make sure to use `->class()` there instead of passing it separately as `:class`.

zepfietje (changes requested):

packages/support/resources/views/components/button/index.blade.php
binaryfire: Hey @zepfietje. That was my initial approach but I changed it to be consistent with this: https://github.com/filamentphp/filament/blob/865084713c11cb1f079a46f7e950438337520b38/packages/support/resources/views/components/input/wrapper.blade.php#L118.

No problems, will update it.

zepfietje: Oh that case was probably missed by someone during a refactor or so.
Could you update that one too?

binaryfire: @zepfietje Updated

zepfietje: Thanks

danharrin:
Unfortunately there are quite a few places missed by this PR - try searching the codebase for `wire:loading.remove.delay` and `wire:loading.delay.class.remove`. Will merge when this is all consistent :)

binaryfire:
Hey @danharrin, I think that's all of them now.

danharrin:
Thanks

packages/support/resources/views/components/badge.blade.php
```diff
@@ -114,19 +114,29 @@
...
             <x-filament::icon
-                :alias="$iconAlias"
-                :icon="$icon"
-                :wire:loading.remove.delay="$hasLoadingIndicator"
-                :wire:target="$hasLoadingIndicator ? $loadingIndicatorTarget : null"
-                :class="$iconClasses"
+                :attributes="
+                    \Filament\Support\prepare_inherited_attributes(
+                        new \Illuminate\View\ComponentAttributeBag([
+                            'alias' => $iconAlias,
```
zepfietje, line 120: @danharrin, these can just be passed as attributes like previously, right?
What's the reason these changed?

danharrin: I just made it consistent with what the rest of the PR did, I think this is slightly more readable when all the attributes are in the same place

zepfietje: Fair enough. Now it's inconsistent with the rest of the codebase though.
Personally I prefer the old way, as it was clear which attributes really needed to be passed as `:attributes`, which makes it easier to see when `:attributes` is no longer needed and attributes can be passed separately once one is no longer needed.

PR #8832, opened 2023-10-02
Update install instructions to use ^3.0 instead of ^3.0-stable

drbyte:
I'm seeing lots of posts on Discord where people are trying to use "latest" features, but they're not available even after running composer update.
It's because they've put `^3.0-stable` in their composer.json instead of just `^3.0`.

### DIFF
--- a/packages/actions/docs/01-installation.md
+++ b/packages/actions/docs/01-installation.md
@@ -20,7 +20,7 @@ Filament requires the following to run:
 Require the Actions package using Composer:
 
 ```bash
-composer require filament/actions:"^3.0-stable" -W
+composer require filament/actions:"^3.0" -W
 ```
 
 ## New Laravel projects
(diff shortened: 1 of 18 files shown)


drbyte:
I'm not sure if you had a script already to update all the places. If so, feel free to close this and use that.

I touched all the `.md` files I could see that contained it,

AND I updated the upgrade script as well .... please review that to be sure I didn't miss something else important in your build processes...

danharrin:
Hmm, all the "latest features" should be on `3.x-stable` though...

The reason why we have this in place, is because some 3.x releases (like 3.x-alphas and betas) did not have the updated version constraints for Livewire, Laravel and PHP, so people who installed using just `^3.0` would be able to install it on incompatible projects. That causes confusion, as they are then locked to very old v3 version. I'd like to keep this in place to prevent that confusion.

drbyte:
@danharrin would it be safe to update all the `^3.0-stable` to `^3.1` now? 

Today I someone with `"3.0-stable"` (missing the `^`), so they were stuck on 3.0.0 .... I feel like maybe the `^` won't be so easily missed if they're looking at something shorter like `^3.1` instead of fixating on the word "stable" and missing the `^`.

Just my thoughts. I defer to your choice!

danharrin:
Since the major deps like Livewire and Laravel are locked in all 3.1 releases, I think we can do that, yes.

PR #8836, opened 2023-10-02
Update MakeWidgetCommand to work with latest versions of `laravel/prompt`

### LINKED ISSUE #8835: `MakeWidgetCommand` not working with latest versions of `laravel/prompts`
### Package

filament/widgets

### Package Version

v3.0.63

### Laravel Version

v10.25.2

### Livewire Version

v3.0.5

### PHP Version

8.2

### Problem description

Not sure if it's a [recent change](https://github.com/laravel/prompts/pull/84/files) in the `laravel/prompts` package or something else, but the `make:filament-widget` commands doesn't work if you select the default option on the "_Where would you like to create this?_" step and always returns a __Require__ validation error.

image

Changing the code in 

```filament/widgets/src/Commands/MakeWidgetCommand.php:86```

To 
```php 
'App\Livewire' => '[App\\Livewire] alongside other Livewire components',
```
(or any other truthy value for the key) seems to solve the issue.

### Expected behavior

You should be able to create widgets by selecting the only given option on the "_Where would you like to create this?_" step of the command

### Steps to reproduce

1. Use the latest versions of laravel and filament
2. Try running the `php artisan make:filament-widget Test` command
3. Don't create the widget inside a resource (step 1 of the prompt)
4. Select `App\Livewire` in the 2nd prompt
5. 2 files should be created. Instead, you'll get a validation error

Unfortunately, I could only partially reproduce this

Running this under Windows 10 and Powershell, it seems to work as expected, but the prompt looks different (attached screenshot).
image

### Reproduction repository

https://github.com/sprtk-ches/laravel-filament

### Relevant log output

_No response_

sprtk-ches:
I'm not sure if this is the correct/preferred solution for this issue, but it seems to work without any side effects. The problem is that an empty value is considered invalid and thus `laravel/prompts` returns a validation error.

By changing the key to a _truthy_, we pass the validation. However, I'm not sure if `App\Livewire` could be a potential panel name (I'm not familiar enough). Maybe a different key (that is guaranteed to never be a panel) could be better?

Fixes #8835 

### DIFF
--- a/packages/widgets/src/Commands/MakeWidgetCommand.php
+++ b/packages/widgets/src/Commands/MakeWidgetCommand.php
@@ -83,7 +83,7 @@ class MakeWidgetCommand extends Command
                             fn (Panel $panel): string => "The [{$panel->getId()}] panel",
                             $panels,
                         ),
-                        '' => '[App\\Livewire] alongside other Livewire components',
+                        'App\Livewire' => '[App\\Livewire] alongside other Livewire components',
                     ]),
                 )] ?? null;
             }


sprtk-ches:
There's a much better solution for this [here](https://github.com/filamentphp/filament/pull/8840)

PR #8840, opened 2023-10-03
Fix widget make command

### LINKED ISSUE #8835: `MakeWidgetCommand` not working with latest versions of `laravel/prompts`

cheesegrits:

A recent change to select() in Prompts is breaking the usage in make-widget, whereby a select option cannot be empty, and required(false) no longer works for select().  The fix for this is to name the additional option in the make-widget panel selection.

There is also another "bug", whereby the make fails if the user's Livewire namespace is not App/Livewire, for example if they have updated from v2 to v3 and elected not to move from App/Http/Livewire.  The fix for this is to use config('livewire.class_namespace') for deriving the path and namespace.

### DIFF
--- a/packages/widgets/src/Commands/MakeWidgetCommand.php
+++ b/packages/widgets/src/Commands/MakeWidgetCommand.php
@@ -97,7 +97,7 @@ class MakeWidgetCommand extends Command
 
         if (! $panel) {
             $namespace = config('livewire.class_namespace');
-            $path = app_path(Str::of($namespace)->after('App\\')->replace('\\', '/')->toString());
+            $path = app_path((string) str($namespace)->after('App\\')->replace('\\', '/'));
         } elseif ($resource === null) {
             $widgetDirectories = $panel->getWidgetDirectories();
             $widgetNamespaces = $panel->getWidgetNamespaces();
--- a/packages/widgets/src/Commands/MakeWidgetCommand.php
+++ b/packages/widgets/src/Commands/MakeWidgetCommand.php
@@ -74,6 +74,7 @@ class MakeWidgetCommand extends Command
 
             if (! $panel) {
                 $panels = Filament::getPanels();
+                $namespace = config('livewire.class_namespace');
 
                 /** @var ?Panel $panel */
                 $panel = $panels[select(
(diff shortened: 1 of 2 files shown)


sprtk-ches:
Closes #8835

packages/widgets/src/Commands/MakeWidgetCommand.php
```diff
@@ -95,8 +96,8 @@ public function handle(): int
         $resourceNamespace = null;
 
         if (! $panel) {
-            $path = app_path('Livewire/');
-            $namespace = 'App\\Livewire';
+            $namespace = config('livewire.class_namespace');
+            $path = app_path(Str::of($namespace)->after('App\\')->replace('\\', '/')->toString());
```
danharrin:
```suggestion
            $path = app_path((string) str($namespace)->after('App\\')->replace('\\', '/'));
```

PR #8845, opened 2023-10-03
Simple repeater [Object Object] fix

valpuia:

Before:
Screenshot from 2023-10-03 16-54-56

After:
Screenshot from 2023-10-03 16-55-26

I have check a lot the use case of what I changed. I think this docs https://github.com/filamentphp/filament/blob/3.x/packages/forms/docs/03-fields/12-repeater.md?plain=1#L424 is the reason, but I never get like below data (from docs) in my database if I use simple repeater
```php
[
    'invitations' => [
        'dan@filamentphp.com',
        'ryan@filamentphp.com',
    ],
],
```
All I got is below, I stored in json column and casts as array
```php
[
    'dan@filamentphp.com',
    'ryan@filamentphp.com',
]
```
I think how it display in this docs is not actually how it store, suggest to remove if I am correct.

I may have missed others, if yes please tell me and I'll look more

### DIFF
--- a/packages/actions/resources/lang/pt_PT/associate.php
+++ b/packages/actions/resources/lang/pt_PT/associate.php
@@ -1,0 +1,45 @@
+<?php
+
+return [
+
+    'single' => [
+
+        'label' => 'Associar',
+
+        'modal' => [
+
+            'heading' => 'Associar :label',
+
+            'fields' => [
+
+                'record_id' => [
+                    'label' => 'Registo',
+                ],
+
+            ],
+
+            'actions' => [
+
+                'associate' => [
+                    'label' => 'Associar',
+                ],
+
+                'associate_another' => [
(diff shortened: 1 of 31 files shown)


danharrin:
I don't think this is correct. In our example, `invitations` in the object represents the column in the database which stores the array - or the result of `getState()` on the form.

danharrin:
Received more info - this is only caused by defaultItems()

valpuia:
Yes, my changes cause error on edit, don't know why it's working (in my system) on yesterday :roll_eyes: 
I've check more and more, but cannot figure out how to solve this.. I'll just submit an issue (if none is report) for ref and close this.

jdion84:
the reason your fix didn't work properly is because there are 2 places you need to fix inside the Repeater class, and you only did it in 1 place.

line 102 and line 543

PR #8864, opened 2023-10-03
Support getting media temporary URL based on user specified conversion name

mansoorkhan96:
This PR adds support to get media temporary URL based on user specified conversion name. 

### Problem:
Currently, the `SpatieMediaLibraryFileUpload` generates temporary URL for private files without using the conversion when use has specified.

Following example would load the actual media instead of specified conversion.

```php
SpatieMediaLibraryFileUpload::make('image')
      ->disk('local')
      ->visibility('private')
      ->conversion('thumb')
```

### Solution:
This PR fixes it by passing the conversion name to `$media?->getTemporaryUrl()`, only when user has specified the conversion name and the conversion the generated conversion is available.

### Checklist:


danharrin:
Thanks

mansoorkhan96:
> Thanks

Thanks to you @danharrin, actually and all the team members for all the hard work you guys put into Filament.

Much Love.

PR #8892, opened 2023-10-04
Fix Markdown editor rtl support

### LINKED ISSUE #8863: MarkdownEditor typing pointer on RTL not work correctly and has problem with direction
### Package

filament/filament

### Package Version

v3.0.63

### Laravel Version

v10.26.2

### Livewire Version

v3.0.0

### PHP Version

PHP 8.2.11

### Problem description

back to issue #8307 

image

This issue happens on RTL and 2 problem :

- Direction of pointer always on LTR
- Behavior of pointer when using on RTL not correct and shown not good. 

### Expected behavior

- Direction change depends on which locale type . for RTL : pointer from right to left
- Behavior of pointer good as LTR version . show in correct position

### Steps to reproduce

just clone and run composer install.
then need database (mysql) with name : filament_markdown_rtl_issue

last thing to do : php artisan migrate --seed

on users resource we have bio MarkdownEditor 

### Reproduction repository

https://github.com/themahdavi/filament-markdown-rtl-issue

### Relevant log output

_No response_

awcodes:
- [x ] Changes have been thoroughly tested to not break existing functionality.
- [x ] New functionality has been documented or existing documentation has been updated to reflect changes.
- [x ] Visual changes are explained in the PR description using a screenshot/recording of before and after.

Fixes #8863 

before

after


zepfietje:
Thanks, @awcodes!

themahdavi:
Thanks Direction fixed but behavior of pointer not work correctly like on LTR mode
I think this issue from CodeMirror because on filament v2 markdowneditor doesn't have this issue.

for example when typing words between ** or **** for italic or bold or use list items pointer of current state not correct position and behavior is wired .

image
In this picture can see cursor pointer on end of italic but when typing characters type in italic !

zepfietje:
Are you able to look into this, @themahdavi?
As I'm not an RTL user, it's very hard for me to know what the right behavior should be like.

awcodes:
Hmm. The fix simply implemented the direction per the easymde docs for setting the code mirror option.

themahdavi:
@zepfietje  > this video show whats happens to pointer when typing and using toolbar options
 
CleanShot 2023-10-06 at 14 57 00

this behavior not shown on filament v2 MarkdownEditor

awcodes:
I will do some more digging, but this could be a bug in EasyMDE.

themahdavi:
Yes @awcodes , I think so EasyMDE not compatible with RTL languages

awcodes:
I can confirm this is a bug in EasyMDE. Issue is happening outside of Filament too.

Screenshot 2023-10-09 at 10 35 31 AM

PR #8898, opened 2023-10-05
Fix hydration of boolean states in Select component

andrewdwallo:
### Description:
Updated the Select component to correctly hydrate Boolean states. By converting them to integer representations (1 for true, 0 for false), we avoid showing duplicate options in the Select list for both the Create and Edit form.

### Context
When using the Select component, I noticed that Boolean states weren't hydrating correctly, leading to duplicate options being displayed which were sometimes either `0, 1, true` or `0, 1, false`. This was only happening for "Edit" forms and not "Create" forms.

### Changes
- Modified the method handling Boolean state hydration in the Select component.

Previously:
```php
public function getDefaultState(): mixed
{
    $state = parent::getDefaultState();

    if (is_bool($state)) {
        return $state ? 1 : 0;
    }

    return $state;
}
```

Updated to:
```php
public function hydrateDefaultState(?array &$hydratedDefaultState): void
{
    parent::hydrateDefaultState($hydratedDefaultState);

    $state = $this->getState();

    if (is_bool($state)) {
        $this->state($state ? 1 : 0);
    }
}
```

The revised method ensures consistent behavior by always hydrating Boolean states to their integer representations. This rectifies the issue of the select options list potentially showing duplicate options due to ambiguous Boolean states.

### Screenshots

**Before:**
Screenshot 2023-10-03 at 11 49 23 PM
Screenshot 2023-10-03 at 11 49 38 PM

**After:**
Screenshot 2023-10-03 at 11 51 49 PM
Screenshot 2023-10-03 at 11 51 41 PM


danharrin:
Thanks

PR #8907, opened 2023-10-05
Force value attribute in Hidden Form Component

jals-es:
Until now, when you added a Hidden field with Reactive to a form and passed it a default value, it created the input with the default value in the value attribute.

The problem came when you modified that value and livewire rendered the component again, it removes the value attribute.

With this change, the value attribute persists.


danharrin:
`wire:model` should handle this instead, no?

jals-es:
Yes when you just want to deal with it in php. If you need to deal with it in javascript you have a problem when the value disappears.

Initially, the value attribute is created when you pass something to it through the ->default() function, but if you have it reactive, when you modify the value it disappears.

danharrin:
To modify it from JavaScript, you should use the `$wire.foo = 'bar'` syntax in Alpine.js. Then it will be properly persisted.

jals-es:
@danharrin 
That is true, but sometimes it is more convenient and faster to be able to take the html element and make it an element.value, especially in certain processes where you cannot afford to do an await.

I believe that this change does not affect anyone who uses it through $wire, and in turn we give the option of using the fastest value with javascript natively without having to go through livewire.

> To modify it from JavaScript, you should use the `$wire.foo = 'bar'` syntax in Alpine.js. Then it will be properly persisted.

danharrin:
It's just not good practice to set the `value` when `wire:model` is used, it can create state that becomes out of sync with the backend. Livewire is able to overwrite the value of that attribute at any time.

PR #8908, opened 2023-10-05
Allow user menu for unauthenticated users

Quadrubo:
This is the second attempt of introducing an easy way for unauthenticated users to use the panel. #8671 had a lot of unnecessary backend code and convoluted if conditions. This PR takes a different approach by only modifying the views.

Some checks for authentication could also be removed because the user menu and the account widget are now fully functional without an authenticated user. Because the user menu can now also be displayed on simple pages, as a side effect, switching from dark to light mode is now also supported before the user is authenticated.

If this PR is welcome and should be merged I will:
- [ ] Add missing translations
- [ ] Add documentation

## Screenshots

- widget when unauthenticated & user menu
269899848-2166f7d2-9e4e-46af-b00a-4b90653fca5c

- user menu when unauthenticated
269900206-17087c58-2664-494b-adb1-19a904d32740

- user menu on simple pages
269900753-e5d30e6f-c002-4ae9-ad1e-f41fc7daefdd


wychoong:
Sorry, I still think this is more suitable for a package/plugin to provide the widget and render hook for the guest menu

Quadrubo:
> Sorry, I still think this is more suitable for a package/plugin to provide the widget 

Filament provides the account widget and, in my opinion, should not crash on the dashboard when the `Authenticate` middleware is removed. Therefor I've opted to modify the widget to make it compatible.

> and render hook for the guest menu

In this PR there is no seperate guest menu, displaying no menu whenever the `Authenticate` middleware is removed looks out of place. There is no option to sign in or sign up anywhere. This modifies nothing for everyone who applies their Authenticate middleware so I don't see why the process should be made mor complicated for users who'd want this feature.
When solving this with a render hook, the whole menu hast to be duplicated like I originally did in #8671 which I don't think is a nice solution. Keeping up with changes in the "real" user menu would not be very maintanable.

danharrin:
I agree that Filament should not crash when removing the authenticate middleware. I am more than happy to merge a PR that fixes that.

I am also happy to merge in any extra render hooks that would be required to create a guest menu in a plugin or in your own app.

However, I do not like the overhead that this "default" guest menu - agreed with @wychoong. Sorry. 

In the future, I am more than happy to validate PR ideas before you spend time on them, to avoid any wasted time.

danharrin:
And with a render hook, you don't really need to keep up with any changes to the normal user menu, your menu would be entirely managed by your codebase, using the `x-filament::dropdown` component

Quadrubo:
Okay thank you, its not any extra work for me if I can reuse this for the plugin. Will work on a PR for the crash and the render hooks in a few weeks.

danharrin:
Sounds good. We want to make Filament as extensible as possible with plugins, and render hooks are a great way to achieve that.

PR #8916, opened 2023-10-05
Enhance drag-and-drop sorting UX

ijpatricio:

The new feature improves the UX, by adding an animation to the movement, and also by fading the future position of the chosen element.

Before:
before

After:
after


ijpatricio:
There are some console errors, but they were already present.
Related to choices.js wanting to re-declare a component.

Let me know what you think?

packages/forms/resources/js/index.js
```diff
@@ -4,3 +4,4 @@ import '../css/components/markdown-editor.css'
 import '../css/components/rich-editor.css'
 import '../css/components/select.css'
 import '../css/components/tags-input.css'
+import '../css/components/sortable.css'
```
zepfietje:
```suggestion
import '../css/components/sortable.css'
import '../css/components/tags-input.css'
```

packages/support/resources/js/sortable.js
```diff
@@ -8,6 +8,10 @@ export default (Alpine) => {
             draggable: '[x-sortable-item]',
             handle: '[x-sortable-handle]',
             dataIdAttr: 'x-sortable-item',
+            animation: 300,
+            // Class name for the drop placeholder
+            // Applying styles on packages/forms/resources/css/components/sortable.css
+            ghostClass: "sortable-ghost",
```
zepfietje: We probably need to change this into a Filamenty class.
Not sure yet what though.

packages/support/resources/js/sortable.js
```diff
@@ -8,6 +8,10 @@ export default (Alpine) => {
             draggable: '[x-sortable-item]',
             handle: '[x-sortable-handle]',
             dataIdAttr: 'x-sortable-item',
+            animation: 300,
```
zepfietje: Have you tried different values, like 200, to make it feel a little snappier?

zepfietje (changes requested):
Wooh, nice one, @ijpatricio!

Such a simple addition that improves the UX a lot, I like it.
Thanks for adding the clear screen recordings.
Could you add a couple more for other places where sortable is used (sortable table rows I think?)?

I've left some comments to make sure the UX is as optimal as it can get when we merge this. 😊

packages/support/resources/js/sortable.js
ijpatricio: ehheeh sure 😄

packages/support/resources/js/sortable.js
ijpatricio: Yeah, I played a little.
I enjoy where it's at, 200.

What do you think of keeping a default, but also customizable?

zepfietje: It's currently at 300, not 200. 😄
Since this will also be used for ordering table rows (right?), I figured 300 might be a little too long.

Hmm, just having a sensible default should work I think.
How would you allow users to customize this?

ijpatricio: eheheh my typo!

So, I went ahead and implemented the custom animation time. I followed the naming sortable has - as in "duration" and not "durationTime".

Let me know what you think. 😄 

image

ijpatricio: Could we have a GitPod setup?

It would come in handy for trying PR's in 1 click. Similarly to the demo https://www.youtube.com/watch?v=IZLcKZzoqW8

zepfietje: GitHub Codespaces should probably do the same thing for you?

packages/forms/src/Components/Repeater.php
```diff
@@ -32,6 +32,8 @@ class Repeater extends Field implements Contracts\CanConcealComponents
 
     protected bool | Closure $isReorderable = true;
 
+    protected int | Closure $sortableAnimation = 300;
```
zepfietje: Naming should probably be changed to be component-specific like `reorderAnimationDuration` or similar. Should be added to all places where sortable is used. Let's wait on @danharrin's opinion before spending more time on this.

zepfietje (changes requested):

packages/forms/src/Components/Repeater.php
danharrin: `reorderAnimationDuration` sounds good

zepfietje: Cool, could you implement this for all places where sortable is used, @ijpatricio?

ijpatricio: Yeps, will do!

packages/support/resources/js/sortable.js
ijpatricio: I tried it, but it's not as flexible. I don't need one, I thought you (or the team) could benefit from tinkering with the UI, with a single click to visit the PR.

I could help by bringing that up.

Does it make sense for you? (whether one or the other)

ijpatricio:
Hey @zepfietje 

https://github.com/filamentphp/filament/pull/8916#discussion_r1362555844
As discussed before, I went ahead and updated every `x-sortable` usage to be able to change the animation duration.

- the public api is now `reorderAnimationDuration`
- default value is 300ms, but can be changed

Let me know what you think?

If all seems good, I will then manually test an example on every location impacted by what the PR changes.

Thanks!

zepfietje:
Thanks, @ijpatricio.
Requesting a review from @danharrin for the API.
I'll do a final review to make sure the default duration makes sense in all places.

packages/forms/src/Components/KeyValue.php
```diff
@@ -314,6 +316,18 @@ public function reorderable(bool | Closure $condition = true): static
         return $this;
     }
 
+    public function reorderAnimationDuration(int | Closure $animation = 300): static
```
danharrin:
```suggestion
    public function reorderAnimationDuration(int | Closure $animation): static
```

packages/forms/src/Components/KeyValue.php
```diff
@@ -314,6 +316,18 @@ public function reorderable(bool | Closure $condition = true): static
...
+    {
+        $this->reorderAnimationDuration = $animation;
+
+        return $this;
+    }
+
+    public function getReorderAnimationDuration()
+    {
+        return (string) $this->evaluate($this->reorderAnimationDuration);
```
danharrin:
```suggestion
    public function getReorderAnimationDuration(): int
    {
        return $this->evaluate($this->reorderAnimationDuration);
```

packages/forms/src/Components/Repeater.php
```diff
@@ -606,6 +608,18 @@ public function disableItemMovement(bool | Closure $condition = true): static
         return $this;
     }
 
+    public function reorderAnimationDuration(int | Closure $animation = 300): static
```
danharrin:
```suggestion
    public function reorderAnimationDuration(int | Closure $animation): static
```

packages/forms/src/Components/Repeater.php
```diff
@@ -606,6 +608,18 @@ public function disableItemMovement(bool | Closure $condition = true): static
...
+    {
+        $this->reorderAnimationDuration = $animation;
+
+        return $this;
+    }
+
+    public function getReorderAnimationDuration()
+    {
+        return (string) $this->evaluate($this->reorderAnimationDuration);
```
danharrin:
```suggestion
    public function getReorderAnimationDuration(): int
    {
        return $this->evaluate($this->reorderAnimationDuration);
```

packages/support/resources/js/sortable.js
```diff
@@ -8,6 +8,10 @@ export default (Alpine) => {
             draggable: '[x-sortable-item]',
             handle: '[x-sortable-handle]',
             dataIdAttr: 'x-sortable-item',
+            animation: parseInt(el.dataset?.sortableAnimationDuration) || 300,
+            // Class name for the drop placeholder
+            // Applying styles on packages/forms/resources/css/components/sortable.css
+            ghostClass: "sortable-ghost",
```
danharrin:
```suggestion
            ghostClass: 'fi-sortable-ghost',
```

packages/tables/src/Table/Concerns/CanReorderRecords.php
```diff
@@ -31,6 +33,18 @@ public function reorderable(string | Closure | null $column = null, bool | Closu
         return $this;
     }
 
+    public function reorderAnimationDuration(int | Closure $animation = 300): static
```
danharrin:
```suggestion
    public function reorderAnimationDuration(int | Closure $animation): static
```

packages/tables/src/Table/Concerns/CanReorderRecords.php
```diff
@@ -31,6 +33,18 @@ public function reorderable(string | Closure | null $column = null, bool | Closu
...
+    {
+        $this->reorderAnimationDuration = $animation;
+
+        return $this;
+    }
+
+    public function getReorderAnimationDuration()
+    {
+        return (string) $this->evaluate($this->reorderAnimationDuration);
```
danharrin:
```suggestion
    public function getReorderAnimationDuration(): int
    {
        return $this->evaluate($this->reorderAnimationDuration);
```

packages/forms/resources/css/components/sortable.css
```diff
@@ -0,0 +1,3 @@
+.sortable-ghost {
```
danharrin:
```suggestion
.fi-sortable-ghost {
```

danharrin (approved):
Doesn't seem to have an effect on reorderable tables, but we can still merge it and when Sortable.js fixes it, we will get that update

ijpatricio:
What a pleasure to give back in the form of improving the source code!
Thank you, @zepfietje and @danharrin  ❤️

binaryfire:
Hi @ijpatricio. Any idea if there’s a way of getting this working with tables in v4? It’d make the UX much nicer.

ijpatricio:
> Hi @ijpatricio! Any idea if there’s a way of getting this working with tables in v4? It’d make the UX much nicer.

Hi @binaryfire I haven't used them myself yet, but the odds are very positive :)

binaryfire:
> > Hi @ijpatricio! Any idea if there’s a way of getting this working with tables in v4? It’d make the UX much nicer.
> 
> Hi @binaryfire I haven't used them myself yet, but the odds are very positive :)

@ijpatricio Great! Would you mind taking a look at the v4 codebase to see if you can get it working for tables? I haven't been able to figure out the problem.

PR #8919, opened 2023-10-06
Fix: Get disk name from registered collection inside Model.

mansoorkhan96:
This PR adds support to optionally get disk name from the collection that is defined inside the model.

## Problem:

`SpatieMediaLibraryFileUpload` field does not pick the disk name from the registered collections inside the model.

```php
// App\Models\Attachment.php
class Attachment extends Model implements HasMedia
{
    use InteractsWithMedia;

    public function registerMediaCollections(): void
    {
        $this->addMediaCollection('file')
            ->useDisk('attachments_disk') // We have disk specified here for this collection.
            ->singleFile();
    }
}
```

**This would not save the file to `attachments_disk` unless we explicity pass the disk name in `disk()` method**
```php
SpatieMediaLibraryFileUpload::make('file')->collection('file'),
```

## Solution

This pull request overrides the `getDiskName` method of `FileUpload` and adds code to optionally load the disk name from the collection defined in model. The first preference is still given to `disk()` so users can specify disk on field itself.

## Checklist

### DIFF
--- a/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
+++ b/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
@@ -9,6 +9,7 @@ use League\Flysystem\UnableToCheckFileExistence;
 use Livewire\Features\SupportFileUploads\TemporaryUploadedFile;
 use Spatie\MediaLibrary\HasMedia;
 use Spatie\MediaLibrary\MediaCollections\FileAdder;
+use Spatie\MediaLibrary\MediaCollections\MediaCollection;
 use Spatie\MediaLibrary\MediaCollections\Models\Media;
 use Throwable;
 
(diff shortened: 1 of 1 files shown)


mansoorkhan96:
Hi, I am open to any code reviews or suggestions like cleaning up something or doing it in any other way.

Thanks.

packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
```diff
@@ -227,6 +228,23 @@ public function deleteAbandonedFiles(): void
...
+        if ($diskName = $this->evaluate($this->diskName)) {
+            return $diskName;
+        }
+
+        /** @var Model&HasMedia $model */
+        $model = $this->getModelInstance();
+
+        $diskNameFromRegisteredConversions = $model
+            ->getRegisteredMediaCollections()
+            ->filter(fn (MediaCollection $collection) => $collection->name === $this->getCollection())
```
danharrin:
```suggestion
            ->filter(fn (MediaCollection $collection): bool => $collection->name === $this->getCollection())
```

packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
```diff
@@ -227,6 +228,23 @@ public function deleteAbandonedFiles(): void
...
+            return $diskName;
+        }
+
+        /** @var Model&HasMedia $model */
+        $model = $this->getModelInstance();
+
+        $diskNameFromRegisteredConversions = $model
+            ->getRegisteredMediaCollections()
+            ->filter(fn (MediaCollection $collection): bool => $collection->name === $this->getCollection())
+            ->value('diskName');
```
danharrin:
```suggestion
            ->first()
            ?->diskName;
```

danharrin:
Thanks

PR #8929, opened 2023-10-06
Register page.header render hook

mohamedsabil83:
This PR adds a new render hook between the title and actions in the page header.
A use case example is adding simple filter tabs in the same line with the title and actions instead of taking an extra space between the header and the page content.


danharrin:
I think the more appropriate name is `panels::page.header.actions.before`. Let's also add `panels::page.header.actions.after`

PR #8945, opened 2023-10-07
Custom DatabaseNotifications Livewire

rupadana:
This PR add new parameter to databaseNotification in HasNotification trait and modify some topbar/index.blade.php

- [x]  Changes have been thoroughly tested to not break existing functionality.

### DIFF
--- a/packages/panels/resources/views/components/topbar/index.blade.php
+++ b/packages/panels/resources/views/components/topbar/index.blade.php
@@ -1,141 +1,102 @@
 @props([
-    'navigation',
+'navigation',
 ])
 
-<div
-    {{
-        $attributes->class([
-            'fi-topbar sticky top-0 z-20 overflow-x-clip',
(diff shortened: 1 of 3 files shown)


wychoong:
for context: https://discord.com/channels/883083792112300104/1159889039739076708

while this is a quick and easy way to allow dev to customize, but it feels very un-filament. 
let's say the intention is to allow customization for notification modal "clear all, mark all" buttons, would it be better with some api to pass in the FilamentAction component, then you can put your code in `->action()`

zepfietje:
I'm not entirely sure what this PR is meant to achieve.
I've read the Discord thread and it seems like adding an `actions()`-like API for the notification modal header actions is a better approach.

We'd need more context on the exact use case of customizing the actions.
Going to close this PR as it includes way too many unrelated (formatting) changes too.

PR #8957, opened 2023-10-08
add translation to breadcrumbs for the plural resource name

ronnytorresmtz:
apply translation to the breadcrumbs resource name

### DIFF
--- a/packages/panels/src/Resources/Pages/Page.php
+++ b/packages/panels/src/Resources/Pages/Page.php
@@ -114,7 +114,7 @@ abstract class Page extends BasePage
         $breadcrumb = $this->getBreadcrumb();
 
         return [
-            $resource::getUrl() => $resource::getBreadcrumb(),
+            $resource::getUrl() => __($resource::getBreadcrumb()),
             ...(filled($breadcrumb) ? [$breadcrumb] : []),
         ];
     }


zepfietje:
What's the issue here exactly?
The label returned from `getBreadcrumb` should be translated already.

ronnytorresmtz:
For example if the application displays "Contacts >Edit" when apply translate as you mentioned that the getBredcrub translate the reality is that I got the action translate like "Contracts/Editar" but not the plural resource name.

I have this for example I used "hola" just to explicit the issue:

```
     public function getBreadcrumb(): ?string
      {
          return __('Hi');
      }
      
  
      public function getModelLabel(): ?string
      {
          return __('Contacts');
          
      }
  
      public function getTitle(): string | Htmlable
      {
          return __('Contacts');
      }
```
Te result is "**Contacts** > hola" insted of  "**Contactos** > hola"..... the plural resource name is not translated

withoutTrans

withTrans

zepfietje:
You probably want to override `getPluralModelLabel`.

ronnytorresmtz:
Thank you Zep, I tested and suppose to work as you mentioned but it did not.

withTrans2

result

Regards,

zepfietje:
Make sure to pass the plural, since now you're still letting Laravel handle the pluralization, which only works well for English.

ronnytorresmtz:
I not sure what do you mean because its a direct translation word I made the pluralization in the translation,....

The only way I could make work was modifing this (on purple color). (vendor/filament/filament/scr/Resources/Pages/Page.php)

codefix

Thank you Zep

PR #8983, opened 2023-10-09
Add alias for color shades in button component

dvanscott:

Noticed this one when chatting with Dan about the other PRs I submitted.

### DIFF
--- a/packages/actions/resources/views/components/group.blade.php
+++ b/packages/actions/resources/views/components/group.blade.php
@@ -9,6 +9,7 @@
     'group' => null,
     'icon' => null,
     'iconButton' => false,
+    'iconSize' => null,
     'label' => null,
     'link' => false,
     'size' => null,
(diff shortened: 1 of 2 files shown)


zepfietje:
@dvanscott @danharrin, I didn't add this yet as I was planning to move the color styles closer to the actual element (label, icon, etc.). This will impact the shades alias as well, as `button` no longer covers all then. Would be a breaking change, wouldn't it?

Let's discuss and update the approach before tagging a new release.

danharrin:
This is currently on the outer element. @dvanscott as long as you `addShades()` instead of replacing them, you should be ok, as child elements will inherit the new shades.

zepfietje:
That's right. 👍

dvanscott:
@danharrin @zepfietje sounds good. Is it temporary that I'll need to use `addShades` instead of `replaceShades` or will that always be the case?

zepfietje:
Advised until I've refactored this part. Though generally `addShades` is a little safer than replacing.

PR #8985, opened 2023-10-09
Ability to send arguments to specific nested action in tests

dommer1:
To use this feature, simply pass the arguments parameter to the mountAction() method when mounting a nested action. The arguments parameter should be a array that key contains the name of the action you want to pass the arguments and value is the array of already mention arguments.

For example, the following action uses nested action in footer.

```PHP
Action::make('edit')
    ->action(function (array $arguments) {
        // $arguments['payload'] contain 'foo'
    })
    ->extraModalFooterActions(fn (): array => [
        Action::make('duplicate')
            ->action(function (array $arguments) {
                // $arguments['payload'] contain 'bar'
            }),
    ]),
```

Mounting this action in a test would look like this:

```PHP
livewire(Actions::class)
    ->mountAction('edit.duplicate', arguments: [
        'edit' => ['payload' => 'foo'],
        'duplicate' => ['payload' => 'bar'],
    ])
```

### DIFF
--- a/packages/actions/src/Testing/TestsActions.php
+++ b/packages/actions/src/Testing/TestsActions.php
@@ -29,11 +29,13 @@ class TestsActions
             /** @phpstan-ignore-next-line */
             $name = $this->parseNestedActionName($name);
 
-            foreach ($name as $actionNestingIndex => $actionName) {
+            foreach ($name as $actionName) {
                 $this->call(
                     'mountAction',
                     $actionName,
-                    (count($name) === $actionNestingIndex + 1) ? $arguments : [],
+                    count($name) > 1
+                        ? $arguments[$actionName] ?? []
+                        : $arguments,
                 );
             }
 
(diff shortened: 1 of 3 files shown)


danharrin:
Isn't this a breaking change if people arent passing in the arguments with action name keys?

packages/actions/src/Testing/TestsActions.php
```diff
@@ -29,11 +29,13 @@ public function mountAction(): Closure
...
 
-            foreach ($name as $actionNestingIndex => $actionName) {
+            foreach ($name as $actionName) {
                 $this->call(
                     'mountAction',
                     $actionName,
-                    (count($name) === $actionNestingIndex + 1) ? $arguments : [],
+                    count($name) > 1
+                        ? $arguments[$actionName] ?? []
+                        : $arguments,
```
danharrin:
```suggestion
                    count($name) > 1 ?
                        $arguments[$actionName] ?? $arguments :
                        $arguments,
```

danharrin:
Alright I'm gonna pass in the arguments array if there is no array key, instead of an empty one, which shouldn't be a BC

danharrin:
Made another change so that it only passes in the arguments if it's the first action in the chain, even if there is no array key that matches the name.

PR #8990, opened 2023-10-09
Remove unwanted spaces in link

### LINKED ISSUE #8352: Unwanted space between link Blade component and punctuation
### Package

filament/filament

### Package Version

v3.0.41

### Laravel Version

10.22.0

### Livewire Version

v3.0.2

### PHP Version

8.2.10

### Problem description

When using the Filament link component a space in the output HTML is added, which looks very clunky when followed by punctuation. I suspect this is due to a newline in the output?

I am not going to make a reproduction repository for this since it can be reproduced with a single line of code so I don't think I would save anyone's time by creating one.

By the way, a `php artisan filament:diagnose` command to print all the versions for the bug report would be really nice IMO. What do you think? Should I make an issue/PR?

### Expected behavior

A link gets rendered like this:

There will be a space after this link !

When this is expected:

There will be a space after this link!

laravel test_

### Steps to reproduce

Use the following Blade snippet:

```blade
<p>
    There will be a space after this <x-filament::link>link</x-filament::link>!
</p>
```

### Reproduction repository

### Relevant log output

_No response_

andcarpi:
Resolves #8352 in a good way i think
also leaves a good way to resolve other unwanted space in any other view, if found

- extended blade and added the `@trim` directive (remove spaces before and after the directive when finding it)
- remove unwanted spaces in link.blade.php using the new directive

### DIFF
--- a/packages/support/resources/views/components/link.blade.php
+++ b/packages/support/resources/views/components/link.blade.php
@@ -144,6 +144,7 @@
             </div>
         @endif
     </a>
+    @trim
 @elseif ($tag === 'button')
     <button
         @if ($keyBindings || $hasTooltip)
@@ -222,4 +223,5 @@
             </div>
         @endif
     </button>
+    @trim
 @endif
--- a/packages/support/src/SupportServiceProvider.php
+++ b/packages/support/src/SupportServiceProvider.php
@@ -97,6 +97,10 @@ class SupportServiceProvider extends PackageServiceProvider
             return "<?php echo \Filament\Support\Facades\FilamentAsset::renderStyles({$expression}) ?>";
         });
 
+        Blade::extend(function($view) {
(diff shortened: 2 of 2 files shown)


zepfietje:
Interesting approach, @andcarpi!
Could `@trim` just be moved to the bottom of the file?

andcarpi:
unfortunately, no... if you do that, the unwanted space reappear
the problem with blade is how it read and compile the @, it sometimes need a space/nl before it to be recognized as a syntax, and when compiling it does not get rid of that extra spacing... so the `@trim` needs to go inside the `@if` statements...

zepfietje:
Alright, thanks for explaining. 👍

danharrin:
Cool, thanks

emmadesilva:
Novel approach! Very neat!!

PR #8992, opened 2023-10-09
Check for recordTitle in AttachAction $getOptions

### LINKED ISSUE #7581: AttachAction $titleAttribute vs. recordTitle()
### Discussed in https://github.com/filamentphp/filament/discussions/7568

<sup>Originally posted by **ajnsn** August  8, 2023</sup>
In v2 it was possible to have a $recordTitleAttribute on the RelationManager together with a custom recordTitle() on the AttachAction.

In v3 there's this if-clause (currently on line [268](https://github.com/filamentphp/filament/blob/3.x/packages/tables/src/Actions/AttachAction.php#L268)) which prefers the  $titleAttribute to the mapping with getRecordTitle() (currently on line [278](https://github.com/filamentphp/filament/blob/3.x/packages/tables/src/Actions/AttachAction.php#L278)) 

Took me a while to find out and it seems to be unwanted, at least from my side. The RelationManager's should be the fallback, not the priority.  This specifically prevents the ability to show HTML labels with the record select. Or is there any other way to achieve this in v3?

Thanks</div>

cheesegrits:

At some point, the ability to use a custom recordTitle() on AttachAction got broken.  The problem being, at the point it checks $titleAttribute in the line I've changed in this PR, it's always going to be set, regardless.  So the alternate return that maps $this->getRecordTitle() never gets called.  So this ...

```php
                Tables\Actions\AttachAction::make()
                    ->recordSelectSearchColumns(['name', 'email'])
                    ->recordTitle(function ($record) {
                        return sprintf('%s - %s', $record->name, $record->email);
                    })
```

... never called recordTitle().

I'm not 100% sure it's kosher to just directly check if $this->recordTitle has a value like this, but there's no simple accessor that doesn't do a bunch of evaluation that isn't necessary.  At this point we just need to know if there is one, not what it is.


cheesegrits:
Ah, I never even saw your issue.  Bad Hugh, search issues more thoroughly.

Thanks for the complete fix and merge.

danharrin:
Thanks to you!

PR #8998, opened 2023-10-10
Ability to omit currency symbol in `format_money()` and `money()` helpers

coolsam726:
This PR adds the `omitSymbol` parameter to the core `format_money()` helper as well as the `money()` methods of Table TextColumns and Infolist TextEntries. This adds the flexibility of formatting monetary values without necessarily prepending or appending the currency symbol, which may be preferred by some of the developers.

As an illustration, instead of having a monetary table column displaying values as **KES 98,000.00**, with this addition you can now omit the symbol and have the same column displaying the values as **98,000.00** which is cleaner in some occasions.

### DIFF
--- a/packages/infolists/docs/03-entries/02-text.md
+++ b/packages/infolists/docs/03-entries/02-text.md
@@ -82,6 +82,15 @@ TextEntry::make('price')
     ->money('EUR')
 ```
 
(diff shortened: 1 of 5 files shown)


danharrin:
Thanks, but I'm not going to merge this. On top of the technical complexity introduced by overriding the currency formatting, adding new arguments to existing methods for extra settings can get very messy, and the order of arguments eventually becomes ambiguous.

I'd suggest creating a macro in your own project / plugin to allow you to do this.

PR #9006, opened 2023-10-10
Fix \Illuminate\Database\QueryException when deleting records with foreign keys

ghost:
This PR addresses the issue where a `\Illuminate\Database\QueryException` is thrown when attempting to delete or force delete a record with foreign keys in the database.

\Illuminate\Database\QueryException

The normal behavior of the system would previously throw an exception when attempting to delete a record with foreign keys. Now, the system gracefully handles this situation and provides a user-friendly message to inform the user that the record cannot be deleted.

### DIFF
--- a/packages/actions/src/DeleteAction.php
+++ b/packages/actions/src/DeleteAction.php
@@ -45,15 +45,20 @@ class DeleteAction extends Action
         });
 
         $this->action(function (): void {
-            $result = $this->process(static fn (Model $record) => $record->delete());
+            try {
+                $result = $this->process(static fn (Model $record) => $record->delete());
 
-            if (! $result) {
+                if (! $result) {
+                    $this->failure();
+
+                    return;
+                }
+
+                $this->success();
+            } catch (\Illuminate\Database\QueryException $exception) {
+
                 $this->failure();
-
-                return;
             }
-
-            $this->success();
         });
     }
 }
--- a/packages/actions/src/ForceDeleteAction.php
+++ b/packages/actions/src/ForceDeleteAction.php
@@ -33,15 +33,21 @@ class ForceDeleteAction extends Action
         $this->modalIcon('heroicon-o-trash');
 
         $this->action(function (): void {
-            $result = $this->process(static fn (Model $record) => $record->forceDelete());
+            try {
 
-            if (! $result) {
+                $result = $this->process(static fn (Model $record) => $record->forceDelete());
+
+                if (! $result) {
+                    $this->failure();
+
+                    return;
+                }
+
+                $this->success();
+            } catch (\Illuminate\Database\QueryException $exception) {
+
                 $this->failure();
-
-                return;
             }
-
-            $this->success();
         });
 
         $this->visible(static function (Model $record): bool {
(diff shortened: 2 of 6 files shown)


danharrin:
Thanks, but I don't think I am going to merge

1) Query exceptions are there to indicate that something went wrong. IMO, we should prevent exceptions in the first place, which you can do by disabling / hiding actions that should not be clicked. For example `->hidden(fn ($record) => $record->relationship->exists())`
2) By catching all query exceptions, you are silencing many other issues that could occur, like a SQL syntax error. There is no way to properly catch foreign key exceptions only.

ghost:
@danharrin Thank you for your feedback. In PostgreSQL Or MySQL, we can target these specific error codes for foreign key violations:

- **23000 | Integrity Constraint Violation**
- **23001 | Restrict Violation**
- **23502 | Not Null Violation**
- **23503 | Foreign Key Violation**
- **23505 | Unique Violation**
- **23514 | Check Violation**
- **23P01 | Exclusion Violation**

Ref: https://www.postgresql.org/docs/current/errcodes-appendix.html

For MySQL:

- **Error Number 1451 | ER_ROW_IS_REFERENCED_2**
... etc

Ref: https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html#error_er_row_is_referenced_2

Just an example below to illustrate how we can handle a foreign key violation error:

```php
try {
    // Our code here.
} catch (\Illuminate\Database\QueryException $e) {
    // For Example, we can use an array of error codes:
    if ($e->getCode() == 23503) {
        // Handle the foreign key violation error appropriately.
        // Display a user-friendly message.
        return;
    }
    // For other errors, we can rethrow the exception.
    throw $e;
}
```

if you want i can rewrite my RP and improve it to catch only errors related to foreign key violation error, what do you think?

danharrin:
Hm but what I said about preventing the exceptions in the first place is probably best, I dont think we should be silencing exceptions

ghost:
I 100% agree with you that we should not be silencing exceptions entirely. Our approach will involve catching and handling exceptions selectively. Specifically, we will catch and handle exceptions when a user is trying to delete a record that has related data elsewhere in the database. In such cases, we will display a meaningful message to inform the user that they cannot delete the record due to its usage in other places.

For other exceptions, such as SQL syntax errors or any unforeseen issues, we will not catch them but instead re-throw the errors back to the user, as you mentioned in point 2. This way, we maintain transparency and provide users with clear feedback about why they cannot perform certain actions while ensuring that other unexpected errors are appropriately raised and not hidden.

Or If we need a way to handle these errors in a specific action, we can consider introducing a new method called onError(Exception $e). Currently, there isn't a built-in way to manage this error scenario. It's important to note that, in my case, we don't want to hide the delete button; we want to inform the user that the data is used elsewhere and cannot be deleted. This new method would allow us to handle such cases effectively.

what do you think? @danharrin

danharrin:
I understand that we can selectively silence, but I believe that we should never get the DB into a situation where it sends an error, it's our job as developers to prevent the DB from wasting its time doing operations that will fail. What is your opposition to doing PHP checks beforehand?

If you want, you can override this code in your own app to catch the exception

PR #9024, opened 2023-10-11
Improve brand logo

zepfietje:


packages/panels/src/FilamentManager.php
```diff
@@ -84,11 +84,16 @@ public function getBrandName(): string | Htmlable
         return $this->getCurrentPanel()->getBrandName();
     }
 
-    public function getBrandLogo(): ?string
+    public function getBrandLogo(): string | Htmlable | null
     {
         return $this->getCurrentPanel()->getBrandLogo();
     }
 
+    public function getBrandLogoHeight(): string | int | null
```
danharrin:
```suggestion
    public function getBrandLogoHeight(): ?string
```

packages/panels/src/Facades/Filament.php
```diff
@@ -28,7 +28,8 @@
  * @method static string getAuthGuard()
  * @method static string | null getAuthPasswordBroker()
  * @method static string | Htmlable getBrandName()
- * @method static string | null getBrandLogo()
+ * @method static string | Htmlable | null getBrandLogo()
+ * @method static string | int | null getBrandLogoHeight()
```
danharrin:
```suggestion
 * @method static string | null getBrandLogoHeight()
```

danharrin (changes requested):
Once you add some basic docs for `brandLogoHeight` this should be ok

packages/panels/resources/views/components/logo.blade.php
```diff
@@ -1,19 +1,38 @@
 @php
     $brandName = filament()->getBrandName();
-    $brandLogo = filament()->getBrandLogo()
+    $brandLogo = filament()->getBrandLogo();
+    $brandLogoHeight = filament()->getBrandLogoHeight() ?? '1.25rem';
```
danharrin, line 4: IMO this should be `1.5` or `1.75`

zepfietje: Yeah I wasn't too sure about 1.25 vs 1.5. However, 1.25 looked slightly better with the Filament logo.

danharrin: I think we should lean towards 1.5, so I am going to swap it

Issue #7703, opened 2023-08-12
Remove and Edit buttons overlap on FileUpload field with both avatar() and imageEditor()


kitbs:
### Package

filament/forms

### Package Version

v3.0.19

### Laravel Version

v10.18.0

### Livewire Version

v3.0.0-beta.7

### PHP Version

8.2.8

### Problem description

When attempting to use avatar() and imageEditor() on a FileUpload field, the X (remove) icon is overlapped by the pencil (edit) icon, so only the pencil is clickable.

### Avatar field without editor

```php
FileUpload::make('avatar')
    ->avatar()
```
![image](https://github.com/filamentphp/filament/assets/4569320/1fad13ac-ee76-4d7c-9b5d-446055c18d43)

### Avatar field with editor

```php
FileUpload::make('avatar')
    ->avatar()
    ->imageEditor()
```

![image](https://github.com/filamentphp/filament/assets/4569320/ef0407b9-fba8-491d-9c39-7d1f8e7c1c0b)

### Detail of overlapping buttons
![image](https://github.com/filamentphp/filament/assets/4569320/b138f9e3-57e6-41cc-b4b0-fcac019ca0b3)

### Expected behavior

Although there is not much room for both buttons to be displayed in the avatar circle, it should be possible to both remove the avatar and also edit it using the image editor. 

### Steps to reproduce

1. Add the following field to a resource's form:

```php
FileUpload::make('avatar')
    ->avatar()
    ->imageEditor()
```

2. Navigate to the resource's edit page

### Reproduction repository

https://github.com/kitbs/filament-7703

danharrin:
Unfortunately we cannot do anything about this, it's how Filepond is designed. You cannot use the avatar layout with the editor

kitbs:
Hi @danharrin

The Filepond demo shows the remove and edit buttons side by side: https://pqina.nl/filepond/#profile-picture-demo

![image](https://github.com/filamentphp/filament/assets/4569320/28d61acd-ed13-4b0b-b633-a931c047442c)

This seems to be achieved with the option `styleButtonRemoveItemPosition: 'left bottom'`

danharrin:
Okay go on then, you can PR it if you want :) Didn't know this layout existed

Issue #8841, opened 2023-10-03
korean lang video upload issue


initred:
### Package

filament/forms

### Package Version

v3.0.63

### Laravel Version

v10.25.2

### PHP Version

8.2

### Problem description

Hello, I enjoy laravel in Korea, and I am currently using filament to build a manager site.

I don't know why, but some video file names including Korean language are not uploaded.

If you rename the file to English and upload it with the special symbols in question, the function of uploading the file works normally.

### Expected behavior

Files must be uploaded normally, regardless of language or file name length.

### Steps to reproduce

I'll list some file names that have raised issues.

1. movie-1.mp4 => it's ok
2. 마음의소리_1회_(가장의권위두개머리가몇개야)_(더빙판)_NonDRM_[HD] (편집본).mp4 => Not working File Upload

### Reproduction repository

https://github.com/initred/filamentphp-korean-lang-video-upload-issue

### Relevant log output

```shell
no log.
```

zepfietje:
I suspect this is because the filename contains a period `.`.

initred:
@zepfietje I removed the period (.) from the file and tried again, but there is still a problem.
And I uploaded the file to the Reproduction repository.

I would appreciate it if you could reactivate this issue.

wychoong:
@initred  can you test if it is working with plain Livewire upload? https://livewire.laravel.com/docs/uploads
also test with filename: 마음의소리123.mp4

initred:
@wychoong I will test it and reply when I have time.

zepfietje:
Yeah thanks for commenting, @wychoong.
Would be great to know if this is a Filament or Livewire issue, @initred.

initred:
@wychoong @zepfietje

I tested different names on the same file with livewire.

마음의소리_1회_(가장의권위두개머리가몇개야)_(더빙판)_NonDRM_[HD] (편집본).mp4

The above files export the logs below.

```
[...]

PR #9056, opened 2023-10-12
File upload improvements

tanthammar:
File upload fixes:

**If original file name is not preserved**
- fix: prevents Livewire temporary file names longer than 255 chars, uses Str::ulid() instead.

**If original file name is preserved**
- new: explanatory file name with incrementing versioning
   - example: `file-foo.png` becomes `file-foo-edit-1.png`, `file-foo-edit-2.png`
   - covers previously saved file with custom naming, containing a version,
       -  like `file-foo-edit-1-suffix.png` becomes `file-foo-edit-2-suffix.png`

**Other**
- fix: catches filepond add file error, previously hidden behind the editor modal.
- new: confirmation to warn when editing svg files which completely destroys the image quality.
- new: method to disable/enable editing of svg files, disabled by default
- fix: #7703 
- fix: filepond upload button sometimes hidden when editing multiple times

**Example of cropped svg file**

Before:
image

After:
image


tanthammar:
There are still CSS issues with multiple file uploads when cropping multiple times. The button disappears.
Will work on that tomorrow.

tanthammar:
> There are still CSS issues with multiple file uploads when cropping multiple times. The button disappears. Will work on that tomorrow.

done

wychoong:
imo filename length 255 chars should be fixed in LW, as it will encode it and will failed in LW before reaching filament 

see https://github.com/filamentphp/filament/issues/8841#issuecomment-1751199793

wychoong:
Not too sure about the confirmation prompt for svg. Can the quality issue be solved by converting the svg into image before editing?

https://github.com/fengyuanchen/cropperjs/issues/733
https://github.com/pqina/filepond-plugin-image-preview/issues/5

tanthammar:
@wychoong My impression is that the link you provided for the svg issues are unrelated. One is a about a Firefox bug and the other is resolved. 
FilePond works very nice with svgs.
It is the html canvas to blob that is destructive.

Do you have another solution for svgs?

wychoong:
I’ll will try pull it down to see if there is another way

wychoong:
Attic open livestream

wychoong:
couldnt replicate the destructive svg you facing

I'm using latest filament and LW
tested with this svg file
map

in the editor I tried 
- cropping without zooming
- zoom in without resize cropper
booth scenario the file saved is in good quality
tested on Edge,Firefox,Safari

saved file
c1RZ7qH6AruCdtXeDD72Vbec0HiUCM-metabWFwLnN2Zw==-

wychoong:
The issue I found when testing is on Create form, the file edited is not saved on the record, but both original and edited files are stored on filesystem, separate issue that I might open an issue later

tanthammar:
Thank you @wychoong , will look into it.

tanthammar:
**Changes**

1. SVG editing is enabled by default.
2. Maintain `mb-0` on the `filepond--root` element.
3. Remove the alert on input errors that are hidden behind the modal. Simply close the modal. The error will be displayed by Filepond or as a regular validation message.
4. Incremental version naming has changed from `edit-#` to `v#`.
    - Example: `file-foo-v1.png`, `file-foo-v2.png`, and so on.
    - Custom naming with suffixes is supported, such as `file-foo-v#-suffix.png`.


**Regarding SVG Files**
After conducting investigations, testing, and reviewing the source code of multiple SVG files, I have concluded that an SVG file must have `width` and `height` attributes with defined units in points to achieve an acceptable quality SVG-to-PNG conversion.

**Criteria that must be met:**
- An SVG file must include width and height attributes.
- The width and height must have defined units.
- For optimal results, the units should be in pt (points).
- The width and height must match the values in the viewBox attribute (which they often don't).

If these criteria are not met, the conversion can be highly destructive.

**Solution**
Process all SVG files before passing them to the image editor by adding or updating the width and height in points. Utilize the viewBox attribute to set the dimensions. (Note that viewBox may have different casings; this PR includes case-insensitive matching.)

**SVG confirmation and disabled alert**
I would like to retain this features, even if it means creating more translations.

**Scenario**
Consider a user who uploads an image in November. Next summer, they want to crop it. The Filepond input field doesn't display the full filename, making it impossible to discern the file extension. Furthermore, many of my users are not well-versed in the distinctions between various file formats. Without a warning before cropping the image, the harm is done.

Even if this PR optimizes SVGs for the best possible resolution, the conversion remains destructive. The infinite scalability is lost.

**Benefits of Keeping the Feature**

- **Prevent Unintentional Conversions:** By offering an option to disable SVG editing or displaying a confirmation warning, we empower users to make informed decisions. This prevents unintentional conversions and safeguards the valuable features of SVG files.
- **Enhanced User Experience:** The solution enhances the user experience by providing transparency and control over the editing process. Users can confidently edit their images without the fear of accidental data loss. Additionally, the alert displayed when editing of SVG files is disabled is necessary to inform the user why nothing happens when they click the editor button.

wychoong:
Question
- can the alert/confirmation be disabled
- If the viewbox from svg doesn’t fulfill your checking I think it will still load into editor and ends up destructive? Is it possible to render with a fixed width and get the height to “patch” the svg

tanthammar:
Changes:
Both the alert and the confirmation are disabled by default.

tanthammar:
> Question
> 
> * can the alert/confirmation be disabled
> * If the viewbox from svg doesn’t fulfill your checking I think it will still load into editor and ends up destructive? Is it possible to render with a fixed width and get the height to “patch” the svg

@wychoong 
I have yet to find any SVG file without the viewBox attribute. I only found that the casing could be different.
Don't really know which width/height to use as a fallback, because it would stretch/distort the image if the ratio is incorrect.
Do you have a suggestion?


(The alert/confirmation are disabled now by default.)

packages/forms/resources/css/components/file-upload.css
```diff
@@ -5,7 +5,7 @@
 @import 'filepond-plugin-media-preview/dist/filepond-plugin-media-preview.css';
 
 .filepond--root {
-    @apply mb-0 overflow-hidden;
+    margin-bottom: 0;
```
danharrin:
```suggestion
    @apply mb-0;
```

sinnbeck:
It seems that this PR might have made some unwanted changes to the upload field that makes it extend below, overlaying other elements

image

This means it isn't possible to click the button anymore

danharrin:
Unfortunately, the overflow-hidden was also causing issues too, can you try to find a different solution rather than adding it back?

sinnbeck:
I will see if I can find a way to get it all working but I am just on my way on vacation so it might be a while sadly

sinnbeck:
It seems that filepond adds a translate3d that drops it below the input. Think a source dive in filepond is needed.

image

sinnbeck:
Adding this css style as an override seems to fix it, but it might be worth diving deeper into the filepond code. I will dig into it more when I get back

```css
.filepond--list-scroller {
    top: -100% !important;
}

sinnbeck:
In case anyone else has the same issue. It seems that the admin panel works with this (it still overflows but the buttons work) so I dug into why. It is all down to if the button has the class `relative`. So just add that to your buttons/element and it should work

danharrin:
Is it worth PRing the fix for `top`?

sinnbeck:
I had to remove that again as it caused other problems sadly. Ensuring overlapping element has positon: relative seems to be the best way of fixing it

illia-sapryga:
I am having same issue

illia-sapryga:
@danharrin can you please look into this issue ? I tried many thing and it breaks something in all cases.

danharrin:
I do not fix issues without reproduction repositories. Also, it sounds like a Filepond issue rather than a Filament issue.

PR #9058, opened 2023-10-12
Fix $hydratedDefaultState being populated accidentally by Select

Magnesium38:

A Select component [overrides the base `hydrateDefaultState`](https://github.com/filamentphp/filament/blob/1d5bec4b8dce8d0f4e289a929612d65d04a8673b/packages/forms/src/Components/Select.php#L1179-L1192) with it's own that handles hydrating of boolean states (#8898). Crucially, it sets the $hydratedDefaultState to be the new state, regardless of what the `$hydratedDefaultState` variable currently is. From looking at how the parent method works, `$hydratedDefaultState` is only changed if it is not null.

This PR changes the override to be consistent with the parent functionally. I did look through the repo for any relevant tests that I could add a test case proving the bug, but I didn't see any relevant ones. If I missed them, please let me know and I'll do my best to write a test for this too.

As for how this bug actually manifests itself, the problem arises when the `$hydratedDefaultState` has been turned into an array accidentally when it was meant to stay null and then another Component tries to hydrate the default state. It ends up [overwriting whatever state it did have with null](https://github.com/filamentphp/filament/blob/1d5bec4b8dce8d0f4e289a929612d65d04a8673b/packages/forms/src/Components/Concerns/HasState.php#L215-L219).

The Form that I'm working with is minimally defined as follows:

```php
Forms\Components\Select::make('is_visible'),
Forms\Components\TextInput::make('name'),
```

Both columns exist in the database and are non-nullable. Additionally, the model has `is_visible` cast to a boolean. If the cast does not exist, the bug doesn't trigger. The presence of the `->boolean()` modifier on the Select component is irrelevant with how the code currently exists.

This results in a form that is unable to populate the name field, even though it's using that same name field to render the title just above it.

image

If you instead swap the order of the columns, it renders the name correctly.

```php
Forms\Components\TextInput::make('name'),
Forms\Components\Select::make('is_visible'),
```

image

------

I have tested this as much as I can locally, but I don't have a robust Filament application that I could use to dogfood the change. Again, the change is mostly based off my understanding of what the parent does and how Select's implementation differs. With that in mind, I wouldn't be suprised if there should be further checks to see if the `$hydratedDefaultState` already has the `$statePath`, again to match the parent. However, since I have been unable to find a bug that manifests itself as a result of that additional check not existing, I went with the most minimal change to fix the bug that I was able to find.

### DIFF
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -1185,7 +1185,9 @@ class Select extends Field implements Contracts\HasAffixActions, Contracts\HasNe
 
             $this->state($state);
 
-            Arr::set($hydratedDefaultState, $this->getStatePath(), $state);
+            if (is_array($hydratedDefaultState)) {
+                Arr::set($hydratedDefaultState, $this->getStatePath(), $state);
+            }
         }
     }
 }


danharrin:
Thanks!

PR #9059, opened 2023-10-12
Explain form's reorderable $condition parameter

drbyte:
I went source-diving to find out whether I could conditionally enable reordering, 
and found that it can be done, but it was not found in the docs.

### DIFF
--- a/packages/tables/docs/10-advanced.md
+++ b/packages/tables/docs/10-advanced.md
@@ -132,6 +132,18 @@ public function table(Table $table): Table
 }
 ```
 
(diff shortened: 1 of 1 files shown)


packages/tables/docs/10-advanced.md
```diff
@@ -132,6 +132,18 @@ public function table(Table $table): Table
 }
 ```
 
+The `reorderable()` method also accepts, as a second parameter, a boolean condition to control whether the reorderable feature will be enabled:
```
zepfietje:
```suggestion
The `reorderable()` method also accepts a boolean condition as its second parameter, allowing you to conditionally enable reordering:
```

danharrin:
Cheers

PR #9072, opened 2023-10-13
[Docs] Clarify that actions should have the same name as the Livewire method

sprtk-ches:
While trying to add an action to a Livewire component, I noticed that if the function name and the action name didn't match, the action didn't work properly.

Even though the button was rendered, the action 

1. would not show a confirmation message
2. would not actually execute the `action` method/closure

That's why I added the note that the names should match.

### DIFF
--- a/packages/actions/docs/06-adding-an-action-to-a-livewire-component.md
+++ b/packages/actions/docs/06-adding-an-action-to-a-livewire-component.md
@@ -65,7 +65,7 @@ class ManagePost extends Component implements HasForms, HasActions
     
     public function deleteAction(): Action
     {
-        return Action::make('delete')
+        return Action::make('deleteAction')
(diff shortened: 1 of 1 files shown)


danharrin:
Actually, I think we also allow `Action` as the suffix of the method name as well. Have you tried this exact example?

zepfietje:
`deleteAction()` as the method name and `delete` without `Action` suffix as the action name should just work, right, @danharrin? I believe we made it work like this on purpose.

danharrin:
Yeah, I think this is correct

sprtk-ches:
Yeah sorry about this. I remember trying that specific example and it didn't work (hence the issue) but maybe it was something else I missed. Or just a brainfart ¯\\\_(ツ)\_/¯

Anyway, sorry for the confusion.

PR #9084, opened 2023-10-13
Wrap all expand-concealing-component event code in $nextTick

cheesegrits:

It seems that in LW3, the expand-concealing-component runs before the DOM is mutated, so the "scroll to first error" was broken - the querySelectorAll can't find any data-validation-error because they aren't there yet..  Wrapping all the x-on:expand-concealing-component code in $nextTick({...}) fixes this.

NOTE - what this doesn't fix is that if errors are not in a concealable component, no scrolling happens.  But that's another issue, which I'll raise separately.  I think the whole form needs to be wrapped in some scroll-to-first-error event code, regardless of whether it has concealable components or not.


danharrin:
As far as I remember, I don't think we've ever had scrolling to validation errors outside of concealed components like this.

Such a shame that we have to implement hacks like this, as I personally believe this Livewire behaviour is a bug.

PR #9090, opened 2023-10-14
Fix to support nested relationship for select field

margarizaldi:

Previously
```php
Select::make('foo')->relationship('bar.foo')
```
incorrectly called `Model::bar.foo() ...`

Though it almost doesn't make sense to use nested relationship in form, but this is important for filter.

### DIFF
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -1085,13 +1085,26 @@ class Select extends Field implements Contracts\HasAffixActions, Contracts\HasNe
 
     public function getRelationship(): BelongsTo | BelongsToMany | \Znck\Eloquent\Relations\BelongsToThrough | null
     {
-        $name = $this->getRelationshipName();
-
-        if (blank($name)) {
+        if (blank($this->getRelationshipName())) {
             return null;
         }
 
-        return $this->getModelInstance()->{$name}();
+        $record = $this->getModelInstance();
+
+        $relationship = null;
+
+        foreach (explode('.', $this->getRelationshipName()) as $nestedRelationshipName) {
+            if (! $record->isRelation($nestedRelationshipName)) {
+                $relationship = null;
+
+                break;
+            }
+
+            $relationship = $record->{$nestedRelationshipName}();
+            $record = $relationship->getRelated();
+        }
+
+        return $relationship;
     }
 
     public function getRelationshipName(): ?string


CodeWithDennis:
When using the options method like below:

```php
SelectFilter::make('myRelation.type') 
 ->label(__('Type'))
 ->multiple()
 ->preload()
 ->options(Type::class),
```

It doesn't show these options but instead it shows the available values based on database records. In the examples below i have one record with 'mail' value.

Expected:
Scherm­afbeelding 2023-10-15 om 16 06 59

What actually is happening:

Scherm­afbeelding 2023-10-15 om 16 07 15



@margarizaldi

danharrin:
Are you sure this was caused by the release today and not before?

margarizaldi:
@CodeWithDennis I couldn't replicate the issue in my project. It works either using:
```php
SelectFilter::make('firstLevel.secondLevel.column_name');
```
or using more concise, `relationship` method:
```php
SelectFilter::make('whatever')
    ->relationship('firstLevel.secondLevel', 'column_name');
```
also for simple relationship:
```php
SelectFilter::make('relationshipName.column_name');
SelectFilter::make('whatever');
    ->relationship('relationshipName', 'column_name');
```

Can you identify and show the query produced by the filter?

CodeWithDennis:
I think i misunderstood the feature, i thought it would allow me to filter on a deeper level `relationshipName.type` but its only for populating the Select field.

My bad.. sorry!

@margarizaldi

margarizaldi:
> I think i misunderstood the feature, i thought it would allow me to filter on a deeper level `relationshipName.type`

Yes it does. It is intended to filter nested relationships. The "filter" actually has that feature, but the Select component didn't supported it before.

CodeWithDennis:
Is this intended?

Lead.php
```PHP
public function quoteRequest(): BelongsTo
{
    return $this->belongsTo(QuoteRequest::class);
}
```

QuoteRequest.php
```php
protected $casts = [
        'status' => Status::class,
        'type' => Type::class,
 ];
```
    
    
https://github.com/filamentphp/filament/assets/23448484/e5ca6aea-fa5e-4939-9190-7a50a38655f4

Maybe its a completely different issue tho so ignore me if thats the case

margarizaldi:
@CodeWithDennis I think you should check your Type class. Is that enum?

CodeWithDennis:
> @CodeWithDennis I think you should check your Type class. Is that enum?

 Yea its an enum

margarizaldi:
@CodeWithDennis Well...
1. Options() will be ignored if it is a relationship, so you don't need to put there.
2. Always worth to check the query, using debugbar or anything you use.
3. You might want to try without enum cast, if it works, check the enum, if not, check your relationship, or raise an issue.
4. Looks like you don't have such nested or deep relationship. So I think the problem isn't relevant to this PR.

CodeWithDennis:
Sure ill do some more research and create a proper issue

margarizaldi:
@CodeWithDennis Always happy to see your feedback.

PR #9091, opened 2023-10-14
Fix UI for hidden affixes action

sumardi:

Small UI fix for hidden suffix/prefix action

```php
...
->suffixActions([
      Action::make('unlinkAccount')
          ->icon('heroicon-m-minus-circle')
          ->color('danger')
          ->tooltip(__('Unlink account'))
          ->visible(fn (): bool => auth()->user()->is_admin)
...
```

Before:
Screenshot 2023-10-14 at 3 50 29 PM

After:
Screenshot 2023-10-14 at 3 51 21 PM

### DIFF
--- a/packages/support/resources/views/components/input/wrapper.blade.php
+++ b/packages/support/resources/views/components/input/wrapper.blade.php
@@ -16,8 +16,8 @@
 ])
 
 @php
-    $hasPrefix = count($prefixActions) || $prefixIcon || filled($prefix);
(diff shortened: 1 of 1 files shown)


zepfietje:
Shouldn't they be filtered before being passed to the wrapper instead?

danharrin:
I think this is fine, there might be cases where `getPrefixActions()` should return hidden ones

PR #9102, opened 2023-10-14
Use route-key

francoism90:

### DIFF
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -993,7 +993,7 @@ class Select extends Field implements Contracts\HasAffixActions, Contracts\HasNe
 
             $form->model($record)->saveRelationships();
 
-            return $record->getKey();
+            return $record->getRouteKey();
         });
 
         $this->fillEditOptionActionFormUsing(static function (Select $component): ?array {
--- a/packages/notifications/src/Events/DatabaseNotificationsSent.php
+++ b/packages/notifications/src/Events/DatabaseNotificationsSent.php
@@ -31,7 +31,7 @@ class DatabaseNotificationsSent implements ShouldBroadcast
 
         $userClass = str_replace('\\', '.', $this->user::class);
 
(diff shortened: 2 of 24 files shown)


francoism90:
This is a draft, because maybe I'm missing something.

It seems Filament is using `$record->getKey()`, which is fine, however this exposes the model ID.
This can be a problem, as someone could check out the number of users, or do other 'fun' things, like trying to gain access to other models, e.g. /users/1/edit, /users/2/edit, .. instead of actually having to use a slug, prefixed_id, uuid, etc.

It should be compatible with current users, as it will fallback to the `getKey()` or `$key` of the model.

I'm also seeing `$model->getKeyName()`, you could also use `$model->getRouteKeyName()` for this.

Maybe I'm being wrong here, if that's the case, I'm happy to close this PR. :)

danharrin:
This is a huge breaking change, and not one that I'm looking to implement, route keys IMO are for routing only.

francoism90:
@danharrin I understand, however since Livewire is also front-end focussed, it actually welcomed to use route-keys.

danharrin:
Yeah, well I don't understand that decision either. Using the route key exposes the values used in routes, so surely that makes other URLs easier to guess with the same UUID? I just don't think its a huge problem that Filament should solve, you should have authorisation logic for all operations anyway.

Also, everything in Filament can be overriden, so you can implement these changes in your app without creating breaking changes for others

francoism90:
@danharrin The problem is having routes like this:
- /users/1
- /users/2
- ..

This also happens when someone performs a search/select, as the keys inside that one are `1`, `2`, etc. as well. For some models this isn't a problem, but for users this can expose the number of active users/models, and it also makes scraping a lot easier.

I'm already overruling the IDs as much as possible, also waiting for this to be merged: https://github.com/livewire/livewire/pull/6598

I really like Livewire, but exposing too many database details, is kinda unwanted.

Furthermore, I fully understand your decision to not merge this, it's a big change, and maybe this can be overruled already.

PR #9133, opened 2023-10-16
Add Notifications Livewire Component to Layout in Docs

iAmKevinMcKee:
Looking through the docs I noticed there is no indication to the user they need to add the notifications component. Not sure if this is how you want to do it, but I think this is important to be in the docs.


wychoong:
There is a different livewire component to be included depending which notification type you using. It’s in their own section

danharrin:
I think this is a good default

PR #9137, opened 2023-10-17
Add prompt to ask for type of widget when no type is passed in command

iAmKevinMcKee:
Opening this to see if you are interested in this type of update to the `make:filament-widget` command. I find myself constantly making widgets and deleting them because I don't pass in `-C` or `-S`. This would resolve that problem.

The implementation may not be great (this is my first time using Prompts) so happy to make changes if needed. I was hoping the prompt could update the `->option('stats-overview')` instead of creating a new `$chartType` variable on the class, but I couldn't figure out how to do that.

Also wondering if this might cause some issues around this code: 

```php
if (! $this->option('force') && $this->checkForCollision([
            $path,
            ...($this->option('stats-overview') || $this->option('chart')) ? [] : [$viewPath],
        ])) {
            return static::INVALID;
        }
```


danharrin:
@iAmKevinMcKee refactored a bit, does it still work as intended?

PR #9140, opened 2023-10-17
[3.x] Add new `Markdown` helper class

ryangjchandler:

This introduces a new `Markdown` helper class in the support package that will let you write Markdown in all of the places where we currently accept `Htmlable`, e.g. `helperText()`, `hint()`, etc.

Right now it uses block-level parsing, so will produce `<p>Markdown goes here</p>` for example, but I do wonder if it makes more sense to use inline-level parsing with the `Str::inlineMarkdown()` method instead.


danharrin:
Thanks!

Issue #8616, opened 2023-09-20
Repeater with `BelongsToMany` relationship doesn't work as expected


sprtk-ches:
### Package

filament/forms

### Package Version

v3.0.53

### Laravel Version

v10.24.0

### Livewire Version

v3.0.5

### PHP Version

8.2

### Problem description

Following the [documentation ](https://filamentphp.com/docs/3.x/forms/fields/repeater#integrating-with-a-belongstomany-eloquent-relationship) for the `Repeater` type doesn't work as expected.

1. Not all related fields are shown. In the linked repo example, we have a user with 2 roles but only one is shown
2. Any edits to the roles usually lead to some sort of error, depending on the action (eg trying to remove an existing role might give a different error than trying to add a new one)

I have tried changing the name of the repeater and other suggested solutions from [here (discord link)](https://discord.com/channels/883083792112300104/1151806053164523591) but nothing seemed to work.

### Expected behavior

Using the `Repeater` I should be able to manage `BelongsToMany` relationships

### Steps to reproduce

Using the linked repo

1. Build the app 
2. Run the migrations and serve
3. Navigate to the homepage
4. You should have at least one user with 2 roles but only one will appear
5. Try editing the roles

Note that creating new relationships seems to work for the 1st time. After that, we get back to the original issue where not all records appear in the form.

### Reproduction repository

https://github.com/sprtk-ches/laravel-filament

MarkKhromov:
@sprtk-ches, hi, friend! I had the same problem. While researching the Filament code, I discovered that it uses a model key to display the elements in the repeater. Try adding something like this to your pivot model:

```
protected $primaryKey = 'custom_id';
public function getCustomIdAttribute(): string {
    return $this->model1_id. '_' . $this->model2_id;
}
```

This code will fix the problem with displaying one entry, but it will not be possible to change it. I'm dealing with the change right now.

MarkKhromov:
@sprtk-ches after studying the Filament code in more detail, I came to the conclusion that the easiest way would be to add an identifier to your Pivot, for example, some kind of UUID. This will solve almost all problems with changing, adding and deleting records. I'll do some more digging to see if I can fix this on the Filament side.

sprtk-ches:
@MarkKhromov Thanks for the workaround 👍 ! I'll leave the issue open and let the filament team decide how they want to handle this.

danharrin:
[...]

PR #9143, opened 2023-10-17
Added a message about the need to use an identifier.

MarkKhromov:
For issue #8616


packages/forms/docs/03-fields/12-repeater.md
```diff
@@ -264,6 +264,8 @@ public function orderProducts(): HasMany
 
 If you don't already have an `OrderProduct` pivot model, you should create that, with inverse relationships to `Order` and `Product`:
 
+> Important! Each of your pivot models must have an ID in the database table.
```
danharrin:
```suggestion
> Please ensure that your pivot model has a primary key column, like `id`, to allow Filament to keep track of which repeater items have been created, updated and deleted.
```

danharrin:
Thanks!

PR #9149, opened 2023-10-17
Fix text entry and column alignment

wychoong:

when using textEntry with badge/copyable and alignment it is not aligned

```php
Infolists\Components\TextEntry::make('status')
      ->state(['abc', '123'])
      ->alignEnd()
      ->badge(),
Infolists\Components\TextEntry::make('reference')
      ->alignEnd()
      ->icon('heroicon-o-clipboard')
      ->iconPosition(IconPosition::After)
      ->copyable(),
```

**before**
image

**after**
image


danharrin:
Does this need to be done for TextColumn too?

wychoong:
> Does this need to be done for TextColumn too?

I did a simple test with TextColumn chained with badge/copyable and alignEnd, it is fine

image

wychoong:
after this change the UI kinda breaks a bit when using `bulleted()->badge()` together

image

not too sure how to solve this, so figured i did mention it here for the UI team to have a look

packages/infolists/resources/views/components/text-entry.blade.php
```diff
@@ -56,8 +58,17 @@
             >
                 <{{ $isListWithLineBreaks ? 'ul' : 'div' }}
                     @class([
+                        'flex flex-wrap items-center' => ! $isListWithLineBreaks,
                         'list-inside list-disc' => $isBulleted(),
-                        'flex flex-wrap items-center gap-1.5' => $isBadge,
+                        'gap-1.5' => $isBadge,
+                        match ($alignment) {
```
zepfietje, line 75: Let's make `Alignment` a backed enum to simplify these match arms.

zepfietje (changes requested):

zepfietje:
> I did a simple test with TextColumn chained with badge/copyable and alignEnd, it is fine

It probably only seems fine since the label is right aligned and the label is narrower than the row cell contents. I checked, and the text column doesn't use the alignment, so it still needs to be handled too.

zepfietje:
> after this change the UI kinda breaks a bit when using `bulleted()->badge()` together

Did this work fine previously?

wychoong:
> > after this change the UI kinda breaks a bit when using `bulleted()->badge()` together
> 
> Did this work fine previously?

I would say it's less ugly 😂 
image

zepfietje:
> I would say it's less ugly 😂

Lol, though old and new both look bad. 😂
I think we don't really support bulleted lists with badges (I mean, when do you really use that?).
So it's fine for now.
Let's look into this if someone ever reports this as a bug.

zepfietje:
Could you mark as ready when you've worked on my other change requests, @wychoong?

wychoong:
i can work on the backed enum, but i not sure what to do with TextColumn

packages/infolists/resources/views/components/text-entry.blade.php
wychoong: @zepfietje  i worked on changing to backed enum, it seems to be breaking change
 
this is because even is backed, `Alignment::End != 'end'`
```php
// example with ActionSize::ExtraLarge dump
Filament\Support\Enums\ActionSize {
  +name: "ExtraLarge"
  +value: "xl"
}
```

as can seen in the codebase, there are multiple places you still using enum and string for the match
image


another way is to resolve the value back to enum 
```php
    public function getAlignment(): Alignment|string|null
    {
        $alignment = $this->evaluate($this->alignment);

        if (is_string($alignment)) {
            return Alignment::tryFrom($alignment);
        }

        return $alignment;
    }
```

Let me know if I'm understand you request wrong, or how you want to proceed with this

zepfietje:
Text column probably just needs the same treatment as text entry?
They're pretty similar components after all.

packages/infolists/resources/views/components/text-entry.blade.php
wychoong: i will leave this PR as draft waiting for your #9167

wychoong:
> Text column probably just needs the same treatment as text entry? They're pretty similar components after all.

I have given the TextColumn another good look
the table cells are wrapped in a column wrapper and the alignment is done here
https://github.com/filamentphp/filament/blob/9ef262bebbb5cbd4984df3bdd88d5633c4bff30b/packages/tables/resources/views/components/columns/column.blade.php#L19-L29

wychoong:
@zepfietje applied the same to TextColumn

table column with `->badge()->alignJustify()`
image

TextEntry
image

packages/infolists/resources/views/components/text-entry.blade.php
zepfietje: Yeah sorry, was about to say I'm handling this in linked PR.
Requires quite a big refactor. 🙂

zepfietje:
> @zepfietje applied the same to TextColumn

Thanks, @wychoong!

zepfietje:
https://github.com/filamentphp/filament/pull/9167 is ready for review now, so when that gets merged, you can finalize this PR and request a review from me when you're done, @wychoong. 🙂

wychoong:
@zepfietje its ready for review.
please note that in the last commit I added some missing alignments as it is getting error
```php
Tables\Columns\TextColumn::make('title')
      ->alignJustify()
```
this is because in the blade it is matching the enum and those not matched are return directly `match($alignment) ... default => $alignment` and gets `Object of class Filament\Support\Enums\Alignment could not be converted to string`

packages/infolists/resources/views/components/entry-wrapper/index.blade.php
```diff
@@ -108,6 +108,7 @@
                     }"
                 @endif
                 @class([
+                    'flex',
```
zepfietje: What's this `flex` needed for? 🙂

packages/infolists/resources/views/components/text-entry.blade.php
```diff
@@ -56,8 +62,18 @@
...
                     @class([
+                        'flex flex-wrap items-center' => ! $isListWithLineBreaks,
+                        'gap-1.5' => $isBadge,
                         'list-inside list-disc' => $isBulleted(),
-                        'flex flex-wrap items-center gap-1.5' => $isBadge,
+                        match ($alignment) {
+                            Alignment::Start => 'justify-start',
+                            Alignment::Center => 'justify-center',
+                            Alignment::End => 'justify-end',
+                            Alignment::Justify => 'justify-between',
```
zepfietje: This should be `Alignment::Between`, and probably just `null` for `Alignment::Justify`?

packages/tables/resources/views/columns/layout/stack.blade.php
```diff
@@ -15,9 +15,10 @@
             ->class([
                 'flex flex-col',
                 match ($alignment) {
-                    Alignment::Start => 'items-start',
+                    Alignment::Start, Alignment::Left => 'items-start',
                     Alignment::Center => 'items-center',
                     Alignment::End, Alignment::Right => 'items-end',
+                    Alignment::Justify => null,
```
zepfietje: `Alignment::Between` should probably also be handled.
Alternatively, we could decide to not support custom classes for alignment and just do `default => null`?

zepfietje (changes requested):
Thanks, @wychoong!

I left a couple of comments.

packages/infolists/resources/views/components/entry-wrapper/index.blade.php
wychoong: weird, in my initial PR it was required to get the text to align, will remove

packages/infolists/resources/views/components/text-entry.blade.php
wychoong: this is tricky. as the HasAlignment Trait doesn't have a setter for `Alignment::Between`. so you can't configure with `->alignBetween`.

packages/tables/resources/views/columns/layout/stack.blade.php
wychoong: i think that would be a breaking change as some might do `->align('justify-between')` and include that php file in vite? as currently you can't `->alignBetween`

however I do agree to just do a `default => null`

packages/infolists/resources/views/components/text-entry.blade.php
wychoong: https://github.com/filamentphp/filament/blob/642a78e613665ab6f5c6fb1c10b56ef18df02f1f/packages/tables/resources/views/components/columns/column.blade.php#L26-L34

😬

danharrin:
Do we need to re-review?

wychoong:
> Do we need to re-review?

pending @zepfietje to decide what to do with the usage of `Alignment::Justify` and `Alignment::Between` in blade

packages/infolists/resources/views/components/text-entry.blade.php
zepfietje: There's an `alignment()` method, right?
That would allow people to pass an enum instance.

packages/tables/resources/views/columns/layout/stack.blade.php
zepfietje: Fair, that could be breaking indeed.
Right, @danharrin?

danharrin: Yeah, breaking

packages/infolists/resources/views/components/text-entry.blade.php
wychoong: should `alignBetween` be added? or its on purpose not to include

zepfietje: Isn't it the same as `justify` in this case?

wychoong: what I meant is in this trait https://github.com/filamentphp/filament/blob/3.x/packages/support/src/Concerns/HasAlignment.php there is no `alignBetween` method

> Isn't it the same as `justify` in this case?

then in this `match` can be updated to `Alignment::Between, Alignment::Justify => 'justify-between text-justify', `


or `alignBetween` method is purposely not provided?

zepfietje: Yeah that's what I meant about the match expression.
I think adding an `alignBetween` method makes sense, @danharrin?

danharrin: If it ensures consistency, sure

glennjacobs:
@danharrin @zepfietje could I be cheeky and give this a little bump, as we really need it? :-)

zepfietje:
@glennjacobs, it's on my list of PRs to review and merge.

zepfietje:
Alright, this PR required a lot more work to fix all cases.
Thanks for doing the groundwork though, @wychoong!

zepfietje (approved):

DariusIII:
With this change ->wrap() does not wrap the text anymore. In 3.1.20 it works fine.

This is on 3.1.21:
image

This is with 3.1.20
image

```php
TextColumn::make('notes')
                    ->color(function ($record) {
                        if ($record->status === 'Enabled') {
                            return 'success';
                        } elseif ($record->status === 'Disabled') {
                            return 'danger';
                        } elseif ($record->status === 'Waiting for file') {
                            return 'warning';
                        } elseif ($record->status === 'Waiting') {
                            return 'error';
                        }

                        return 'info';
                    })
                    ->html()->toggleable(isToggledHiddenByDefault: false)->wrap()->label('Notes'),
```

danharrin:
fixed in the next version

PR #9151, opened 2023-10-17
load the table when it's visible

akunbeben:

When we use wire:navigate and use the Filament table, it will lead us to this error. Loading the table only when it's visible will fix it.
image


akunbeben:
https://github.com/filamentphp/filament/assets/46495960/96a24293-96d3-4515-a115-218a7d0f330c

PR #9153, opened 2023-10-18
Use data_get in applyFiltersToTableQuery to support dotted relations

cheesegrits:

Using data_get in applyFiltersToTableQuery() allows the use of dotted syntax in things like SelectFilter and TernaryFilter.

Without this PR, related filters have to be done like this:

```php
                SelectFilter::make('agency_id')
                    ->label('Agency')
                    ->relationship('agency', 'name')
                    ->searchable(),
                    
                TernaryFilter::make('all-users')
                    ->relationship('channelType', 'is_all_users')
                    ->label('All Users'),
```

After the PR, they can be done like this:

```php
                SelectFilter::make('agency.name')
                    ->label('Agency')
                    ->searchable(),

                TernaryFilter::make('channelType.is_all_users')
                    ->label('All Users'),
```

Is it strictly necessary?  Probably not.  But as the columns themselves are built using the dotted notation, seems like it makes sense to be able to use the same dotted notation for the filters.  And afaict, it's literally only that data_get() which is needed to allow it.  And there's no bc issues because anything that worked with $data[$filter->getName()] will still work with data_get($data, $filter->getName()).

### DIFF
--- a/packages/tables/src/Concerns/HasFilters.php
+++ b/packages/tables/src/Concerns/HasFilters.php
@@ -107,7 +107,7 @@ trait HasFilters
         foreach ($this->getTable()->getFilters() as $filter) {
             $filter->applyToBaseQuery(
                 $query,
-                $data[$filter->getName()] ?? [],
+                data_get($data, $filter->getName()) ?? [],
             );
         }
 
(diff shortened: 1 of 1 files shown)


danharrin:
Hey, not gonna merge this. The reason is because it fundamentally changes the data structure of the filter form. Currently, filters with dot notation will store their data in an array key on the top level of the filter form. This allows the data of filters to be easily accessible at the array key with the same name as the filter.

This change would encourage people to use dot notation, which nests the data deeper into the array, and could cause issues, such as `Filter::make('author')` and `Filter::make('author.name')` being used together and creating an overlapping dataset where the second filter's data is nested within the first's.

PR #9162, opened 2023-10-18
Fix pagination responsiveness

### LINKED ISSUE #7597: Pagination overlaps page selector on mobile

zepfietje:

Closes #7597.

- Uses container queries when supported by the client.
- Uses screen breakpoints only when container queries aren't supported.
- When pagination items become very wide, they don't overlap with the select field, rather grow and push the field left (this is the `grid-cols-3` to `grid-cols-[1fr_auto_1fr]`) change.

I'm sure there's still edge cases where the pagination overview text (`Showing x of x results`) wraps to a new line, but we can always tweak the breakpoints if needed. Also, we may consider setting `onEachSide(0)` instead of the default of 1 to show `... currentPage ...` instead of `... currentPage-1 currentPage currentPage+1 ...`. It seems like Laravel always keeps the items to a minimum of 7 though, and I'd rather not mess with the framework defaults.

Anyway, this should fix most issues that have been reported. 😊


jals-es:
I think this change broke pagination styles.

I update from 3.0.73 to 3.0.81 and now the pagination of my tables appears like this.

Captura de Pantalla 2023-10-20 a las 12 45 49

PR #9164, opened 2023-10-18
Evaluate the conditions in the alignment shorthand methods

### LINKED ISSUE #9152: Closure passed to alignment methods are never executed
### Package

filament/filament

### Package Version

v3.0.79

### Laravel Version

v10.28.0

### Livewire Version

v3.0.9

### PHP Version

PHP 8.2.10

### Problem description

If you pass a closure as `$condition` to e.g. `->alignEnd(...)` it is never evaluated. 

```php
TextEntry::make('created_at')
      ->alignEnd(static fn (): bool => false) // <-- evaluates to true
      ->since(),
```

### Expected behavior

The entry should not align to end because the given `$condition` returns false.

### Steps to reproduce

Pass a closure which returns false to e.g. `->alignEnd(...)` and it will still align e.g. the text entry to the end.

This code will reproduce it:

```php
<?php

use Filament\Infolists\Components\TextEntry;

$textEntry = TextEntry::make('foo');

$textEntry->alignEnd(static fn (): bool => false);

$textEntry->getAlignment(); // returns Filament\Support\Enums\Alignment::End but expected to be NULL.

```

### Reproduction repository

https://github.com/maartenpaauw/filament

### Relevant log output

_No response_

ashleyshenton:
Fixes #9152 

Currently if you pass a closure it isn't evaluated. Because the closure itself is always truthy, this leads to it always setting the alignment. This PR evaluates the conditions using the Laravel `value` helper. I checked and saw this used in other places throughout the codebase so assume it is okay to use here too.

I noticed there is an `evaluate` method used in `getAlignment` but it seems overkill for this.

Here is an output of the same code block shown in the issue.

image

### DIFF
--- a/packages/support/src/Concerns/HasAlignment.php
+++ b/packages/support/src/Concerns/HasAlignment.php
@@ -18,32 +18,32 @@ trait HasAlignment
 
     public function alignStart(bool | Closure $condition = true): static
     {
-        return $this->alignment(static fn (): ?Alignment => $condition ? Alignment::Start : null);
+        return $this->alignment(static fn (): ?Alignment => value($condition) ? Alignment::Start : null);
     }
 
     public function alignCenter(bool | Closure $condition = true): static
     {
-        return $this->alignment(static fn (): ?Alignment => $condition ? Alignment::Center : null);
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks

PR #9167, opened 2023-10-18
Improve actions spacing

zepfietje:


zepfietje:
My goal here was/is to use the same actions component everywhere and handle spacing including negative margins on icon button actions.

Unfortunately, this gets tricky in places where we use actions inside dropdowns as triggers, or use a wrapping `<span>` to apply `x-on:click`. An alternative solution would be to handle the negative margin on icon buttons inside the actual icon button Blade component. Would be much easier, but has the downside of always making icon buttons take up less space than they actually are sized.

What do you think would be the best approach, @danharrin?
Let me know if this is unclear and you need more info. 🙂

zepfietje:
Decided on a different approach: adding the negative spacing to the icon button component itself.

zepfietje:
Not sure why the test is failing. Maybe something went wrong when merging in 3.x?

PR #9169, opened 2023-10-18
compact section content

wychoong:

currently when using Section components in Infolist or Form, its only the section container padding is adjusted
this PR aims to make the content in the Section to be tighter as well

Infolist section with inline label TextEntry
image

Form with placeholders
image

### DIFF
--- a/packages/forms/resources/views/component-container.blade.php
+++ b/packages/forms/resources/views/component-container.blade.php
@@ -2,6 +2,7 @@
     $parent = $container->getParentComponent();
     $isCompact = $parent instanceof \Filament\Forms\Components\Section && $parent->isCompact();
 @endphp
+
 <x-filament::grid
     :default="$getColumns('default')"
     :sm="$getColumns('sm')"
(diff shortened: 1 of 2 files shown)


zepfietje:
I don't think we should do this, since the spacing of form fields will be odd.

wychoong:
> I don't think we should do this, since the spacing of form fields will be odd.

i think that can be tweaked. I tried `gap-x-6 gap-y-3` instead of `gap-3`

form
image

infolist
image

hope you can reconsider this

zepfietje:
Have you tried a long list of many form fields?

wychoong:
> Have you tried a long list of many form fields?

I think it looks ok, if to avoid breaking the current UX, it can be introduced as `->compactBody()` api
image

image

my use case is I'm creating an infolist with sidebar, its purely info (with minimal custom entry that might have input)
image

so if it can be compacted it will actually looks better without too much white spaces. That being said if the concerns are for form fields UI/UX I'm ok to revert for Form section to get Infolist section compact going

also one of the thing need to look into is when nesting a Group/Grid in the section

zepfietje:
I'm not a big fan of the compact forms, since spacing inside form fields like checkbox lists is larger than spacing between all fields, which looks weird and causes UX issues.

If you could provide a clear side-by-side comparison of the current sidebar design and the design you'd like to achieve, we may reconsider this.

wychoong:
> If you could provide a clear side-by-side comparison of the current sidebar design and the design you'd like to achieve, we may reconsider this.

image

ideally the RepeatableEntry child components should be compact too

zepfietje:
Honestly, the right screenshot really needs more space to breathe. 😅

PR #9170, opened 2023-10-18
Update ckb and ku translations

RawanD201:
Hello, this PR updates the current CKB/KU translation and makes a few corrections.


zepfietje:
Thanks

Issue #9177, opened 2023-10-19
Summarize Query Bug on PostgreSQL


chrisreedio:
### Package

filament/filament

### Package Version

v3.0.80

### Laravel Version

v10.28.0

### Livewire Version

v3.0.10

### PHP Version

v8.2.11

### Problem description

When using a custom Summarizer on PostgreSQL, I get the following QueryException:

`SQLSTATE[42601]: Syntax error: 7 ERROR: zero-length delimited identifier at or near """"` 

The generated query is:

```sql
SELECT
  sum(price * quantity) AS aggregate
FROM
  (
    SELECT
      *
    FROM
      "invoice_items"
  ) AS ""
```

You can see that the last line of the generated query is `) AS ""`.

**Possible Solution Identified?**

If this pans out, I can submit a PR for this fix.

I looked at 
https://github.com/filamentphp/filament/blob/62cf9339776b6ba517a510774bfe89feb3bd14fe/packages/tables/src/Columns/Summarizers/Summarizer.php#L108

and 

https://github.com/filamentphp/filament/blob/62cf9339776b6ba517a510774bfe89feb3bd14fe/packages/tables/src/Concerns/CanSummarizeRecords.php#L70

When I took the line from `CanSummarizeRecords` and replaced L108 in `Summarizer` with it, the problem was resolved. 

It seems on PostgreSQL we can't have the second argument of that function be null (the default).

### Broken - Summarizer.php - L108

```php
$query = DB::table($query->toBase());
```

### Works - CanSummarizeRecords.php - L70

```php
$query = DB::table($query->toBase(), $query->getModel()->getTable());
```

### Expected behavior

It would display the proper summary such as this example from another project.
![chrome_f2ReTePhkA](https://github.com/filamentphp/filament/assets/77644584/f5cec6a2-5c7d-4f32-9ee9-8451831e8f63)

### Steps to reproduce

I have a Resource/Model with a quantity and price field/column.

I want to multiply them together for each line in the table to build out the 'total price'.

```php
Tables\Columns\TextColumn::make('price')
	->money()
	->summarize(
		Summarizer::make()
			->label('Total')
			->money()
			->using(fn (\Illuminate\Database\Query\Builder $query) => $query->sum(DB::raw('price * quantity')))
	),
```

### Reproduction repository

https://github.com/chrisreedio/filament-summarize-delim-bug/blob/main/app/Filament/Resources/InvoiceItemResource.php#L74

danharrin:
[...]

PR #9178, opened 2023-10-19
Fixes Summarize Query Bug on PostgreSQL #9177

chrisreedio:
This fixes the bug I documented in #9177 .

## Summary

When using a custom Summarizer on PostgreSQL, I get the following QueryException:

SQLSTATE[42601]: Syntax error: 7 ERROR: zero-length delimited identifier at or near """"

with the following code: 

```php
Tables\Columns\TextColumn::make('price')
	->money()
	->summarize(
		Summarizer::make()
			->label('Total')
			->money()
			->using(fn (\Illuminate\Database\Query\Builder $query) => $query->sum(DB::raw('price * quantity')))
	),
```

Before the fix:

image

After the fix:

image

Copied existing code from:
https://github.com/filamentphp/filament/blob/62cf9339776b6ba517a510774bfe89feb3bd14fe/packages/tables/src/Concerns/CanSummarizeRecords.php#L70

The second parameter being null on the Summarizer generates the faulty `) AS ""` at the end of the query.

### DIFF
--- a/packages/tables/src/Columns/Summarizers/Summarizer.php
+++ b/packages/tables/src/Columns/Summarizers/Summarizer.php
@@ -105,7 +105,7 @@ class Summarizer extends ViewComponent
                 );
         }
 
-        $query = DB::table($query->toBase());
+        $query = DB::table($query->toBase(), $query->getModel()->getTable());
 
         if ($this->hasQueryModification()) {
             $query = $this->evaluate($this->modifyQueryUsing, [


chrisreedio:
So that total still isn't correct but the SQL doesn't fail now. 

Perhaps I have an error in the way I'm trying to calculate the total?

danharrin:
Thanks!

PR #9183, opened 2023-10-19
Update 01-getting-started.md

kombiahrk:
static keyword is missing


zepfietje:
The infolist method is only static on resources.

PR #9187, opened 2023-10-19
Prevent GitHub from opening when installing with --no-interaction

### LINKED ISSUE #8880: Installing FilamentPHP with --no-interaction opens the GitHub webpage
### Package

filament/filament

### Package Version

v3.0.67

### Laravel Version

10.26.2

### Livewire Version

v3.0.5

### PHP Version

8.2.11

### Problem description

When running `php artisan filament:install --panels --no-interaction`, the https://github.com/filamentphp/filament page opens in the browser.

### Expected behavior

The page should not be opened in the browser when running without interaction.

This could probably be fixed by changing the default value of the "Would you like to show some love by starring the Filament repo" prompt to false when running without the interactive flag.

### Steps to reproduce

Setup:

```bash
laravel new filament-bug-repro --no-interaction
cd filament-bug-repro 
composer require filament/filament:"^3.0-stable" -W --no-interaction
```

Repro:
```bash
php artisan filament:install --panels --no-interaction
```

NOTE: Since the bug does not require a repro, I just posted my standard one. All you need to do is to run the repro command above.

### Reproduction repository

### Relevant log output

_No response_

ashleyshenton:
Fixes #8880 

I know this issue was assigned to the person who raised it, so I hope you don't mind me raising a PR. The simplest solution in my opinion is to set the confirmation default to false when the no-interaction option is passed.

I've tested this for normal installation (without `--no-interaction`) and you are still prompted with the default being 'yes'.
When running with `--no-interaction` or `-n`, the rest of the script runs as before but the browser is no longer opened with the filament GitHub page.

### DIFF
--- a/packages/support/src/Commands/InstallCommand.php
+++ b/packages/support/src/Commands/InstallCommand.php
@@ -37,7 +37,7 @@ class InstallCommand extends Command
 
         if (confirm(
             label: 'All done! Would you like to show some love by starring the Filament repo on GitHub?',
-            default: true,
+            default: ! $this->option('no-interaction'),
         )) {
             if (PHP_OS_FAMILY === 'Darwin') {
                 exec('open https://github.com/filamentphp/filament');
--- a/packages/support/src/Commands/InstallCommand.php
+++ b/packages/support/src/Commands/InstallCommand.php
@@ -35,9 +35,20 @@ class InstallCommand extends Command
 
         $this->installUpgradeCommand();
 
+        $this->askToStar();
+
+        return static::SUCCESS;
+    }
+
+    protected function askToStar(): void
+    {
+        if ($this->option('no-interaction')) {
+            return;
+        }
+
         if (confirm(
             label: 'All done! Would you like to show some love by starring the Filament repo on GitHub?',
-            default: ! $this->option('no-interaction'),
+            default: true,
         )) {
             if (PHP_OS_FAMILY === 'Darwin') {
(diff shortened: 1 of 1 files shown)


packages/support/src/Commands/InstallCommand.php
```diff
@@ -37,7 +37,7 @@ public function __invoke(): int
 
         if (confirm(
             label: 'All done! Would you like to show some love by starring the Filament repo on GitHub?',
-            default: true,
+            default: ! $this->option('no-interaction'),
```
danharrin: Instead, maybe we should early return before this even happens, so we don't need to ask the question in the first place

danharrin (changes requested):

danharrin:
Perfect, thanks. You're smashing these! It's really appreciated.

packages/support/src/Commands/InstallCommand.php
ashleyshenton: I've extracted a method in which it returns early if the `no-interaction` option is passed. I've tested it and it has the same desired effect.

emmadesilva:
Thanks! I was going to try my hand at this today, but don't mind you doing it at all! Been super busy myself, so haven't been able to code much.

zepfietje:
Just came here to say thank you, @ashleyshenton.
Really appreciate people like you jumping in and helping us out getting the issue count down. 🫶

PR #9192, opened 2023-10-19
Update Table Builder 01-installation.md

valpuia:

I've install Table builder on my existing project and the notification cannot work, I've struggle a lot. I create a fresh application and notice that in the layout components `@livewire('notifications')` is added. I've added this one in my existing and it works, so I thought this need to be instructed as well because notification is also installed along with table.

I didn't check other builder but may need this also..

### DIFF
--- a/packages/tables/docs/01-installation.md
+++ b/packages/tables/docs/01-installation.md
@@ -142,6 +142,8 @@ Create a new `resources/views/components/layouts/app.blade.php` layout file for 
     <body class="antialiased">
         {{ $slot }}
 
+        @livewire('notifications')
+
         @filamentScripts
         @vite('resources/js/app.js')
     </body>


danharrin:
The notifications installation is not required for the tables to work, so I would rather not bloat the layout just yet. I think it is probably quite self-explanatory that the notifications package needs installing if you want to receive notifications.

valpuia:
Screenshot from 2023-10-20 09-03-54
If I run `composer require filament/tables:"^3.0-stable" -W` then I got like image, notification is pre-installed along with others. But in order to use I need to put  `@livewire('notifications')` in layout. Normally dev (like me) just copy-paste from docs that's why

>the notifications package needs installing if you want to receive notifications.

I didn't check for receiving notifications rn, but what I check is showing "Updated successfully" on top right side (don't know whether it's same or not :rofl: )

PR #9198, opened 2023-10-19
Properly fixes decimal issues for money fields

incapable:
Instead of simply dividing without bcdiv, this introduces an optional scaling variable that fixes the decimal display for money fields.
It has the added benefits that the scale can be overridden to display whatever scale you want.

### DIFF
--- a/packages/support/src/helpers.php
+++ b/packages/support/src/helpers.php
@@ -12,12 +12,12 @@ use Illuminate\View\ComponentAttributeBag;
 use NumberFormatter;
 
 if (! function_exists('Filament\Support\format_money')) {
-    function format_money(float | int $money, string $currency, int $divideBy = 0): string
+    function format_money(float | int $money, string $currency, int $divideBy = 0, int $scale = 2): string
     {
         $formatter = new NumberFormatter(app()->getLocale(), NumberFormatter::CURRENCY);
 
         if ($divideBy) {
-            $money /= $divideBy;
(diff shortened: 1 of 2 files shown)


danharrin:
1) What issue are we fixing using `bcdiv()`? As far as I'm aware it only really matters when you're looking for high precision, but currency never has more than a few decimal places anyway.
2) "the scale can be overridden to display whatever scale you want" as far as I know, that is incorrect, since PHP handles formatting the number to the correct number of decimal places, based on the currency. So the scale doesn't impact display at all.

I would accept `bcdiv()` if you fetched the number of decimal places from the currency dynamically. But I don't believe that is possible in native PHP. Hence why I used `/=`.

incapable:
1. The only real reason is being able to set the scale, other than that it doesn't matter.
2. The point is that this way, the scale can be set to whatever is needed.

It's a very small thing, the `/=` fix would already be fine, but as `bcdiv()` was used before, this felt a bit like a regression.

I agree that retrieving at least the default value from the given currency would be better, but that'd probably need a pretty extensive library somewhere with currencies. That seems overkill.

danharrin:
I still don't understand though, the scale is lost when the money is formatted. PHP overrides the scale with the scale for that currency.

incapable:
Ah I see, I missed your point.
If PHP overrides the scale in the formatter, then this is indeed entirely pointless.
I'll close this pull request :)

danharrin:
Alright, thanks anyway

PR #9200, opened 2023-10-19
Added @auth check to navigation

WebKenth:

I was playing around with panels and found that if you wanted to have an Authentication Less panel, it would not work with the default theme, as it has a dependency on having a user in the `vendor/filament/filament/resources/views/widgets/account-widget.blade.php` 

This causes an exception ( `Filament\FilamentManager::getUserAvatarUrl(): Argument #1 ($user) must be of type Illuminate\Database\Eloquent\Model|Illuminate\Contracts\Auth\Authenticatable, null given` ) and rather than having to deep dive into creating my own theme just to extend this simple widget i figured a PR would be handy for anyone else experiencing the same :) 

Before:
image

After:
image

Easy to replicate

Simply do a `php artisan make:filament-panel` and remove the AuthMiddleware part of the panel configuration (`app/Providers/Filament/<panel>PanelProvider.php`
```php
            ->authMiddleware([
                Authenticate::class,
            ]);
```

Visit your panel and there you see the exception

With the change from this PR you can now make Front facing panels for Public view if needed

### DIFF
--- a/packages/panels/resources/views/widgets/account-widget.blade.php
+++ b/packages/panels/resources/views/widgets/account-widget.blade.php
@@ -4,6 +4,7 @@
 
 <x-filament-widgets::widget class="fi-account-widget">
     <x-filament::section>
+        @auth
         <div class="flex items-center gap-x-3">
             <x-filament-panels::avatar.user size="lg" :user="$user" />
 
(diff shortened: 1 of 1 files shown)


zepfietje:
Honestly, I don't think this is the right approach.
You should either simply remove this widget from your panel, display it conditionally, or come up with a new design for the unauthenticated state of the widget.

Feel free to attempt a new PR if your approach requires changes to Filament core. 🙂

PR #9204, opened 2023-10-19
Fix form fields ax-load issue on spa-mode

akunbeben:

As per #9151 fixes the table ax-load issue on spa mode, the same issue also happened in some form fields. 
This PR will fix it.

https://github.com/filamentphp/filament/assets/46495960/acc6b039-3aaa-47a2-8e50-4abf1cf7f671


akunbeben:
DateTimePicker still has the issue with the displayText etc. I'm working on it

zepfietje:
Thanks for working on this, @akunbeben!
Could you mark this PR as ready for review when you've implemented all pending changes?

ArtMin96:
I have the same issue. Thanks @akunbeben

akunbeben:
> Thanks for working on this, @akunbeben! Could you mark this PR as ready for review when you've implemented all pending changes?

Yes, I will. 
I'll continue to work on it later today

ArtMin96:
Any news for this PR?

akunbeben:
> DateTimePicker still has the issue with the displayText etc. I'm working on it

Sorry for the delay. After a while, I finally recognized that the problem was when we async load the component, it is not really asynchronous because the livewire.js expects the props in the first place. 

Simply "wire:ignore" along with ax-load="visible" solves the problem. 

Manually tested.

danharrin:
Please revert all unnecessary changes where you are moving large code blocks like this

New imports can be added at the top of the existing code blocks, they don't need to be split

Screenshot 2023-11-15 at 10 03 19

akunbeben:
Changing it, Sorry

packages/forms/resources/views/components/date-time-picker.blade.php
```diff
@@ -57,7 +59,12 @@
         @else
             <div
                 x-ignore
-                ax-load
+                @if (FilamentView::hasSpaMode())
+                    wire:ignore
```
danharrin: Is this wire:ignore deliberate? I think it might break some dynamic features

danharrin: Removed it

ArtMin96:
The problems of Date and Time pickers have not been fixed. Adding `wire:ignore` fixed the issue and it's not breaking the dynamic feature (tested locally).
```blade
@if (FilamentView::hasSpaMode())
    ax-load="visible"
    wire:ignore
@else
    ax-load
@endif
```
image

danharrin:
wire:ignore would break dynamic min/max dates and disabled dates

PR #9205, opened 2023-10-19
Update Swedish translations

adevade:
- Sync from all English lang files, to get all missing values
- In places where the original has changed since last PR, I tried to match it

At first I changed the translation from _Notis_ to _Avisering_, but then i found #9046 so I changed it back again. I agree with @tanthammar with using _Notis_ instead. It's not as formal.

Also, I first changed the translation for _Record_ from _Rad_ to _Post_, but I reverted that as well. What do you think about that change, @tanthammar? I think neither feel on point, but I can't come up with something better...

### DIFF
--- a/packages/actions/resources/lang/sv/associate.php
+++ b/packages/actions/resources/lang/sv/associate.php
@@ -4,16 +4,16 @@ return [
 
     'single' => [
 
-        'label' => 'Associera',
+        'label' => 'Koppla',
 
         'modal' => [
 
-            'heading' => 'Associera :label',
+            'heading' => 'Koppla :label',
 
             'fields' => [
 
                 'record_id' => [
-                    'label' => 'Rad',
+                    'label' => 'Objekt',
                 ],
 
             ],
@@ -21,11 +21,11 @@ return [
             'actions' => [
 
                 'associate' => [
-                    'label' => 'Associera',
+                    'label' => 'Koppla',
                 ],
 
                 'associate_another' => [
-                    'label' => 'Associera & associera en till',
+                    'label' => 'Koppla & koppla en till',
                 ],
 
             ],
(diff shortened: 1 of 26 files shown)


zepfietje:
Please let us know once you've decided on the final changes, @adevade @tanthammar. 😊

tanthammar:
@adevade Andreas? Let's have a call? I do believe these translations have some issues with tempus form. Faster to discuss IRL. You're on discord?

tanthammar:
In a SaaS context where clarity is important, and to avoid confusion with the term "post," we could consider using "objekt" in Swedish to refer to the record. 
For "active record ID," a translation could then be "aktivt objekt ID" This term could help maintain a clear distinction, reducing the likelihood of misunderstanding in a CMS environment or similar applications.

Opinions?

"Rad" feels awkward...

adevade:
> You're on discord?

@tanthammar Sure, I'm on the Filament server! Same username over there! 👍

adevade:
@tanthammar 

> In a SaaS context where clarity is important, and to avoid confusion with the term "post," we could consider using "objekt" in Swedish to refer to the record. For "active record ID," a translation could then be "aktivt objekt ID" This term could help maintain a clear distinction, reducing the likelihood of misunderstanding in a CMS environment or similar applications.

Yeah, `objekt` could work. It could feel kinda wrong for abstract models, like `Rating` or `Direction`, but that might not matter 😅

> "Rad" feels awkward...

Yeah, I'm not really a fan of `rad` either. And you'd have to know that it's referring to a row in the DB or table for it to make sense really. I think I PRed that, but only to get rid of `post`. I go back and forth apparently 😅

zepfietje:
Mark this as ready for review when it's final, @adevade. 🙂

PR #9250, opened 2023-10-22
Fix tags input duplicate tags on rehydration

s-7777:

This fixes a bug in the Tags Input component where duplicate tags were created during rehydration. The issue arose from the immediate initialization of the Alpine.js component's state in the x-data attribute, leading to conflicts with Livewire's rehydration.

The solution delays the Alpine.js component initialization until after the DOM update by using x-init.

### DIFF
--- a/packages/forms/resources/views/components/tags-input.blade.php
+++ b/packages/forms/resources/views/components/tags-input.blade.php
@@ -10,12 +10,10 @@
     <div
         ax-load
         ax-load-src="{{ \Filament\Support\Facades\FilamentAsset::getAlpineComponentSrc('tags-input', 'filament/forms') }}"
-        x-init="function() {
-            return tagsInputFormComponent({
-                state: $wire.{{ $applyStateBindingModifiers("\$entangle('{$statePath}')") }},
-                splitKeys: @js($splitKeys),
-            })
-        }"
+        x-init="() => tagsInputFormComponent({
+            state: $wire.{{ $applyStateBindingModifiers("\$entangle('{$statePath}')") }},
+            splitKeys: @js($splitKeys),
+        })"
         x-ignore
         {{
             $attributes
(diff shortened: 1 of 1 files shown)


packages/forms/resources/views/components/tags-input.blade.php
```diff
@@ -10,10 +10,12 @@
     <div
         ax-load
         ax-load-src="{{ \Filament\Support\Facades\FilamentAsset::getAlpineComponentSrc('tags-input', 'filament/forms') }}"
-        x-data="tagsInputFormComponent({
-                    state: $wire.{{ $applyStateBindingModifiers("\$entangle('{$statePath}')") }},
-                    splitKeys: @js($splitKeys),
-                })"
+        x-init="function() {
```
zepfietje: This can be refactored to `() => {}` right?

danharrin: I think even `() => (tagsInputFormComponent(...))` which replaces the `return` IIRC

zepfietje: Oh yeah, I saw multiple lines while it's just a single function call lol.

packages/forms/resources/views/components/tags-input.blade.php
```diff
@@ -10,10 +10,12 @@
...
+            return tagsInputFormComponent({
+                state: $wire.{{ $applyStateBindingModifiers("\$entangle('{$statePath}')") }},
+                splitKeys: @js($splitKeys),
+            })
+        }"
```
zepfietje:
```suggestion
        x-init="() => tagsInputFormComponent({
            state: $wire.{{ $applyStateBindingModifiers("\$entangle('{$statePath}')") }},
            splitKeys: @js($splitKeys),
        })"
```

danharrin:
Also, how does Alpine know this is a component without an x-data ?

zepfietje:
I was going to comment exactly that, @danharrin.

s-7777:
Yeah, this is clearly breaking now -- I was testing a few different things yesterday to get to the root cause and I thought this was it, but it's clearly not right. Going to close this out and take a proper look again.

zepfietje:
Maybe we just need to add a `wire:ignore` attribute, @s-7777?
Could you try and test if that fixes the issue for you?
Making sure nothing breaks of course. 😄

PR #9252, opened 2023-10-23
Add docs on disabling lazy loading in relation managers

archilex:
RM's are lazy loaded by default. This causes an issue with pagination in Livewire per this [discussion on discord](https://discord.com/channels/883083792112300104/1164244431440982026) and this [PR](https://github.com/livewire/livewire/pull/6760) in LW. 

You can disable lazy loading in RMs by passing `'lazy' => false` to the properties. This PR adds this to the docs. 


zepfietje:
Maybe it makes sense to add the same API as widgets have (an `$isLazy` property)?

archilex:
@zepfietje It would keep things consistent. However, AFAIK you can't pass properties to widgets like you can for RM. This means there could be a conflict if a dev were to pass both the property but also use the API. I would just need to know which should take precedence.

zepfietje:
Widgets do allow passing arguments to their `make()` method.
There was actually an issue or PR the other day caused by a conflict like you described I believe.

Let's leave this up to @danharrin to decide.

PR #9259, opened 2023-10-23
feature: Filter modal

danharrin:

Screenshot 2023-10-23 at 11 28 31

Screenshot 2023-10-23 at 11 28 19


zepfietje:
@danharrin I feel like close should be "Apply", though that would be confusing since the filters are automatically applied... What do you think?

danharrin:
You're correct, it would be confusing. When we introduce a deferred filter feature, then we can introduce that instead.

zepfietje:
Sounds good!

imperiumbee:
Amazing! 
@danharrin Could you tell when are deferred filters expected?

danharrin:
We don't have a date, you can PR it but previous PRs have failed due to unclear UX, so be prepared for a long review process

imperiumbee:
What about #5116 ? Logic seems clear, or not?
there can add a confirmation on  the modal closing, apply filters or not

danharrin:
Where would we store the old filter values to restore them to?

PR #9283, opened 2023-10-24
fix summerize count

l3aro:

image

As you can see, this page has 4 green ticks and 6 x marks, but filament miscount to 4x4 ticks and 6x6 marks
This PR is to fix this by looping the values that need to be counted only once

image

### DIFF
--- a/packages/tables/src/Columns/Summarizers/Count.php
+++ b/packages/tables/src/Columns/Summarizers/Count.php
@@ -37,7 +37,7 @@ class Count extends Summarizer
 
         $state = [];
 
-        foreach ($query->clone()->pluck($attribute)->uniqueStrict() as $value) {
+        foreach ($query->clone()->distinct()->pluck($attribute) as $value) {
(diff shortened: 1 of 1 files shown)


packages/tables/src/Columns/Summarizers/Count.php
```diff
@@ -37,7 +37,7 @@ public function summarize(Builder $query, string $attribute): int | float | arra
 
         $state = [];
 
-        foreach ($query->clone()->pluck($attribute) as $value) {
+        foreach ($query->clone()->pluck($attribute)->uniqueStrict() as $value) {
```
danharrin: Maybe this would work better instead, on the database level? I haven't tested it, can you please?

```suggestion
        foreach ($query->clone()->distinct()->pluck($attribute) as $value) {
```

l3aro: that's the good idea, its work as expected

danharrin:
Thanks

PR #9287, opened 2023-10-24
[3.x] Ask for panel ID when installing with `--panels`

ryangjchandler:

This changes how the `filament:install --panels` command behaves slightly and asks for the `id` of the panel, the same way the `make:filament-panel` command does.

When used from the `filament:install` command, it will still default to `admin` and generate the same code but it makes it easier for the developer to use a custom name for the panel by default, without having to generate another panel or manually modify the code it would generate previously.

> NOTE: It still maintains the same behaviour with regards to target directories, where resources/pages/widgets etc are registered from, nothing new there.

### DIFF
--- a/packages/panels/src/Commands/MakePanelCommand.php
+++ b/packages/panels/src/Commands/MakePanelCommand.php
@@ -2,6 +2,7 @@
 
 namespace Filament\Commands;
 
+use Filament\Support\Commands\Concerns\CanGeneratePanels;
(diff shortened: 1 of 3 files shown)


valpuia:
Sorry for late comment but this PR is not working with custom ID. `admin` is working fine but others will throw below error

Screenshot from 2023-11-04 09-37-49

danharrin:
This should be fixed in 5d3705f1a994d38f1093566bfba4b43beae60e6b

PR #9291, opened 2023-10-25
feature: Allow table searches to include phrases when wrapped in doub…

SignetPlanet:
When searching in tables the search previously broke each word separately on spaces between them. However sometimes that generates an unexpected large number of results, especially when search is enabled for multiple columns. By adding the ability to not break words up on spaces **when** they are within double quotes, we can have better search results.  So a search for `"Jane Doe"` will do a single search for the entire phrase `Jane Doe`. You can also mix them, so if in search bar you type `treatment "Jane Doe"` the system would search for both `treatment` and `Jane Doe`.

The old system used `explode()` which returned an array containing an empty string `['']` if the search string was an empty string. The new method uses `str_getcsv()` which returns an array containing null `[null]` so to ensure it was compatible, I wrapped so if the search string was empty it just returned `['']` without running the `str_getcsv()`. This should make it so it does not break any existing functionality. I added tests to confirm, which I ran before and after making the change from `explode()` to `str_getcsv()`. Then I also added tests that the string is not split on spaces within the double quotes. `" "`. When testing in a laravel installation, it worked and it even let me search for a single space `" "` now and it finds records that contain at least one space.

I did not update documentation as I didn't see any documentation about how the search is divided up already on each space and so I was unsure if and where you would want this.

There are no visual changes, you just wrap phrases in double quotes `"` in the table search input field. If you type in "Jane Doe" you will see a single search tag appear with quotes `Search: "Jane Doe"`. 

If you would like documentation updated, please let me know where and I'll add it.

### DIFF
--- a/packages/tables/src/Concerns/CanSearchRecords.php
+++ b/packages/tables/src/Concerns/CanSearchRecords.php
@@ -108,7 +108,10 @@ trait CanSearchRecords
      */
     protected function extractTableSearchWords(string $search): array
     {
-        return $search === '' ? [''] : str_getcsv(preg_replace('/\s+/', ' ', $search), ' ');
+        return array_filter(
+            str_getcsv(preg_replace('/\s+/', ' ', $search), ' '),
+            fn ($word): bool => filled($word),
+        );
     }
 
     protected function applyGlobalSearchToTableQuery(Builder $query): Builder
--- a/packages/tables/src/Concerns/CanSearchRecords.php
+++ b/packages/tables/src/Concerns/CanSearchRecords.php
@@ -108,7 +108,7 @@ trait CanSearchRecords
      */
     protected function extractTableSearchWords(string $search): array
     {
-        return explode(' ', preg_replace('/\s+/', ' ', $search));
+        return $search === '' ? [''] : str_getcsv(preg_replace('/\s+/', ' ', $search), ' ');
     }
 
     protected function applyGlobalSearchToTableQuery(Builder $query): Builder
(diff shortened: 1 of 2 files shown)


packages/tables/src/Concerns/CanSearchRecords.php
```diff
@@ -108,7 +108,7 @@ protected function applyColumnSearchesToTableQuery(Builder $query): Builder
      */
     protected function extractTableSearchWords(string $search): array
     {
-        return explode(' ', preg_replace('/\s+/', ' ', $search));
+        return $search === '' ? [''] : str_getcsv(preg_replace('/\s+/', ' ', $search), ' ');
     }
```
SignetPlanet, line 115: `str_getcsv()` returns `[null]` if `$search` is empty string, so it was wrapped so if empty it returns `['']` bypassing the need to run `str_getcsv()`

danharrin: Is there a reason we should ever return `['']`? IMO, if there are no search terms, we should return `[]`

Maybe we need to filter the array before returning, and only allow terms that are `filled()`

SignetPlanet: > Is there a reason we should ever return `['']`? IMO, if there are no search terms, we should return `[]`
> 
> Maybe we need to filter the array before returning, and only allow terms that are `filled()`

Not sure to be honest, I didn't have the time to dive in deep enough to see if it could ever be passed an empty string. Regardless however, if this method gets re-used elsewhere, would they know to ensure it is not passed an empty string. I guess if we want to be safe, we could have it throw an exception if it was passed an empty string.

danharrin: I've refactored it to filter the array of empty strings

PR #9293, opened 2023-10-25
New table filters indicator API

zepfietje:

May need some more docs still, but let's go for a review first.

### DIFF
--- a/packages/tables/src/Filters/Indicator.php
+++ b/packages/tables/src/Filters/Indicator.php
@@ -25,7 +25,7 @@ class Indicator extends Component
 
     public static function make(string | Closure | null $label = null): static
     {
-        return new static($label);
+        return app(static::class, ['label' => $label]);
     }
 
     public function label(string $label): static
--- a/packages/tables/src/Filters/Indicator.php
+++ b/packages/tables/src/Filters/Indicator.php
@@ -1,0 +1,78 @@
+<?php
+
+namespace Filament\Tables\Filters;
+
+use Closure;
+use Filament\Support\Components\Component;
+use Filament\Support\Concerns\HasColor;
+
+class Indicator extends Component
+{
+    use HasColor;
+
+    protected bool | Closure $isRemovable = true;
+
+    protected string | Closure | null $label = null;
+
+    protected string | Closure | null $removeField = null;
+
+    protected string | Closure | null $removeLivewireClickHandler = null;
+
+    public function __construct(string $label)
+    {
+        $this->label($label);
+    }
+
+    public static function make(string | Closure | null $label = null): static
+    {
+        return new static($label);
+    }
+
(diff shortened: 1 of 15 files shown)


packages/tables/src/Filters/SelectFilter.php
```diff
@@ -39,7 +39,7 @@ protected function setUp(): void
                 __('filament-tables::table.filters.select.placeholder'),
         );
 
-        $this->indicateUsing(function (SelectFilter $filter, array $state): array {
+        $this->indicators(function (SelectFilter $filter, array $state): array {
```
s-7777: @danharrin you mentioned that we can indicate on selected items and remove them individually, do you think we should make that the default here to make that functionality more discoverable?

danharrin: At the moment, each default filter only has one indicator for simplicity and consistency

zepfietje: Yeah I think it makes sense to keep it simple by default. Especially since it's relatively easy to set up separate indicators for each value.

danharrin (changes requested):
I think I am missing the point of `indicators()`, as you would never want to pass an array in without a closure. This is because passing an array would mean you cant access the state, so the indicators would always be active regardless of whether or not the filter was applied.

If you agree, let's remove `indicators()` and just go back to `indicateUsing()`

zepfietje:
I agree with the fact that passing an array to `indicators()` doesn't make sense indeed.
Personally I still prefer `indicators()` over `indicateUsing()` when you're passing a closure though.

Why do you like `indicateUsing()` more?

danharrin:
Our `*Using` pattern denotes places where only a closure can be passed, whereas `indicators()` for sure suggests you can use an array.

We could even keep `indicators()` as an alias, but I think `indicateUsing()` should be documented and used internally.

zepfietje:
Fair point I guess.

packages/tables/src/Filters/Indicator.php
```diff
@@ -0,0 +1,78 @@
...
+
+    protected bool | Closure $isRemovable = true;
+
+    protected string | Closure | null $label = null;
+
+    protected string | Closure | null $removeField = null;
+
+    protected string | Closure | null $removeLivewireClickHandler = null;
+
+    public function __construct(string $label)
```
qrasa1, line 21: Am I missing something, or shouldn't the type be changed from `string` to `string | Closure | null`? The same applies to `label()`, including the evaluation of `$label` in `getLabel()`.

danharrin: Good point, fixed that on the 3.x branch

zepfietje: Yup, thanks for noticing and fixing.

PR #9294, opened 2023-10-25
Optimize spatie tags table column query

margarizaldi:

To eliminate n+1 queries.

Before:
image

After:
No more queries.


margarizaldi:
I believe the phpstan failure is safe to ignore. Should I put ignore-next-line there?

PR #9299, opened 2023-10-25
Fix to allow state to be maintained with custom notifications

### LINKED ISSUE #9268: Custom notifications do not persist state
### Package

filament/notifications

### Package Version

v3.0.83

### Laravel Version

v10.28.0

### Livewire Version

v3.0.10

### PHP Version

v8.2.11

### Problem description

Using custom notifications as detailed in the docs, the state of a custom property within a custom notification is not maintained.

### Expected behavior

The state to be maintained.

### Steps to reproduce

Repo included. The notification on the Settings page dumps the state of the notification which includes the default "size" property and not the one specified. The test at tests/Feature/SettingsTest.php also shows that the state is not maintained.

Related, the test shows that the "actual" result is correct when the "expected" is incorrect. I think this is down to the assertion rather than the `Notification`.

I started a discussion at https://github.com/filamentphp/filament/discussions/8719 which is related and indicates the possible culprit.

### Reproduction repository

https://github.com/titantwentyone/filament-custom-notification-issue

### Relevant log output

_No response_

titantwentyone:

This is a fix for #9268.

Functionality is already documented but docs do not reflect what should be possible. I'm not sure what happened here. It's either a case that it's changed and because there are were no tests for custom notifications, this was missed or something is missing from the docs.

I couldn't see the purpose of the container binding given in the docs also. The tests and my own app run this amend fine without it. Maybe from when the functionality originally worked? Any insight would be appreciated!

As this approach is not tied to one single binding, this could be used to show different custom notifications rather than rely on just one overridden custom class.

### DIFF
--- a/packages/notifications/src/Collection.php
+++ b/packages/notifications/src/Collection.php
@@ -29,7 +29,7 @@ class Collection extends BaseCollection implements Wireable
     public static function fromLivewire($value): static
     {
         return app(static::class, ['items' => $value])->transform(
-            fn (array $notification): Notification => Notification::fromArray($notification),
+            fn (array $notification): Notification => $notification['type']::fromArray($notification),
         );
     }
 }
--- a/packages/notifications/src/Livewire/Notifications.php
+++ b/packages/notifications/src/Livewire/Notifications.php
@@ -35,7 +35,8 @@ class Notifications extends Component
     public function pullNotificationsFromSession(): void
     {
         foreach (session()->pull('filament.notifications') ?? [] as $notification) {
-            $notification = Notification::fromArray($notification);
+            $notification_class = $notification['type'];
+            $notification = $notification_class::fromArray($notification);
 
             $this->pushNotification($notification);
         }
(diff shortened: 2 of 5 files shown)


zepfietje:
We can't do this as it would be a major security issue.
This change allows users to specify pass any class to the back end which would then be instantiated using `fromArray()`.

titantwentyone:
How about encrypting/decrypting?
```
...
$static->type(Crypt::encryptString($data['type']));
...
```
and
```
...
'type' => Crypt::decryptString($this->getType())
...

danharrin:
We can just check if the class `is_subclass_of()` the base Notification class

danharrin:
The container binding should work to be honest, as we use the container to instantiate new Notification classes. Let's just remove that from the docs for now.

danharrin:
@titantwentyone can you have a look at #9311 please and let me know if it fixes it?

titantwentyone:
@danharrin Just rolled PR #9311 from @wychoong into my example repo in issue #9268 and all checks out. I think that PR is more akin to how it should (and probably originally did) work rather than mine so please close this in favour of that if you're happy.

danharrin:
Thanks!

PR #9311, opened 2023-10-26
Fix custom notification object

### LINKED ISSUE #9268: Custom notifications do not persist state

wychoong:

closes #9268 

when recreating the notification object from array, `Notification` class resolved the correct `CustomNotification` but the `CustomNotification::fromArray` is not called again. this PR fixes that by checking if the notification object resolved from container is custom and its called from `Notification` class to ensure its only called once and prevent infinite loop

another approach would be in livewire\Notifications to resolved the notification object but that would have multiple useless instance created and more code changes that I not able to test

### DIFF
--- a/packages/notifications/src/Notification.php
+++ b/packages/notifications/src/Notification.php
@@ -89,8 +89,14 @@ class Notification extends ViewComponent implements Arrayable
     {
         $static = static::make($data['id'] ?? Str::random());
 
-        if (is_subclass_of($static::class, self::class) && get_called_class() === self::class) {
-            $static = $static::fromArray($data);
+        // If the container constructs an instance of child class
+        // instead of the current class, we should run `fromArray()`
+        // on the child class instead.
+        if (
+            ($static::class !== self::class) &&
+            (get_called_class() === self::class)
+        ) {
+            return $static::fromArray($data);
         }
 
         $static->actions(
--- a/packages/notifications/src/Notification.php
+++ b/packages/notifications/src/Notification.php
@@ -89,7 +89,7 @@ class Notification extends ViewComponent implements Arrayable
     {
         $static = static::make($data['id'] ?? Str::random());
 
-        if (is_subclass_of($static, self::class) && get_called_class() === self::class) {
+        if (is_subclass_of($static::class, self::class) && get_called_class() === self::class) {
             $static = $static::fromArray($data);
         }
 
--- a/packages/notifications/src/Notification.php
+++ b/packages/notifications/src/Notification.php
@@ -88,6 +88,11 @@ class Notification extends ViewComponent implements Arrayable
     public static function fromArray(array $data): static
     {
         $static = static::make($data['id'] ?? Str::random());
+
+        if (is_subclass_of($static, self::class) && get_called_class() === self::class) {
+            $static = $static::fromArray($data);
+        }
+
         $static->actions(
             array_map(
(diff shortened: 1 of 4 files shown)


danharrin:
Duplicate of #9299?

wychoong:
> Duplicate of #9299?

Fixed the same issue without exposing unnecessary info

danharrin:
I don't think I understand how it works?

wychoong:
> I don't think I understand how it works?

in this file
https://github.com/filamentphp/filament/blob/73d6c1c42f10407455c689902e93171fe4086912/packages/notifications/src/Livewire/Notifications.php#L35-L42

`Notification::fromArray($notification)` is `Filament\Notifications\Notification`

here it will return the `CustomNotification` from the `make` method
https://github.com/filamentphp/filament/blob/73d6c1c42f10407455c689902e93171fe4086912/packages/notifications/src/Notification.php#L88-L91

and it continues to configure the object with the array data, thus it never gets to call to `CustomNotification::fromArray`, this can be verified by `dd` in `CustomNotification::fromArray`

with this fix
```php
if (is_subclass_of($static::class, self::class) && get_called_class() === self::class) {
    $static = $static::fromArray($data);
}
```
it checks if the `$static` notification class is extended from the base `Notification`.
And also make sure it is called from `Notification`. This is because we need to call `parent::fromArray` in the `CustomNotification` to ensure the rest of properties are set. 
Without these checks it will be infinite loops as `base->custom->base->custom->...`

packages/notifications/src/Notification.php
```diff
@@ -88,6 +88,11 @@ public function toArray(): array
     public static function fromArray(array $data): static
     {
         $static = static::make($data['id'] ?? Str::random());
```
danharrin, line 90: Shouldn't the `make()` method instantiate a new CustomNotification from the container, instead? I think thats the original idea behind binding it to the container

wychoong: it does get the CustomNotification, this is not wrong, but as it only gets it here, the `CustomNotification::fromArray` is not called at all

wychoong: in `Livewire\Notifications`
```php
// before fix
Notification::fromArray
   > `make()` return `CustomNotification`
   > set properties to `CustomNotification` with Notification's definition
   > return to lw

$this->push into collection
```

```php
// after fix
Notification::fromArray
   > `make()` return `CustomNotification`
   > call `CustomNotification::fromArray`
        > call `parent:fromArray`
        > run `CustomNotification` logic
   > return to lw

$this->push into collection
```

maybe this is clearer?

danharrin:
Thanks!

PR #9362, opened 2023-10-29
Fix table groups only extra column

### LINKED ISSUE #9297: groupsOnly while summarizing table with sum mismatch in table ui
### Package

filament/filament

### Package Version

v3.0.83

### Laravel Version

v10.29.0

### Livewire Version

v3.0.8

### PHP Version

PHP 8.1

### Problem description

While using method to groupOnly in resource table column are not showing properly
Summarizing column is offset by one column.
Table code shown below
image

```
    {
        return $table
            ->columns([
                Tables\Columns\TextColumn::make('wastetype.name')
                    ->sortable(),
                Tables\Columns\TextColumn::make('opening_balance')
                    ->summarize(Sum::make()),
                Tables\Columns\TextColumn::make('generation_quantity')
                    ->summarize(Sum::make()),
                Tables\Columns\TextColumn::make('disposal_quantity')
                    ->summarize(Sum::make()),
                Tables\Columns\TextColumn::make('closing_balance')
                    ->summarize(Sum::make()),
            ])
            ->defaultSort('date', 'desc')
            ->defaultGroup('wastetype.name')
            ->groupsOnly()
            ->filters([
                Tables\Filters\SelectFilter::make('wastetype_id')
                    ->multiple()
                    ->label('Wastetype')
                    ->options(fn(): array => \App\Models\Wastetype::where('status', TRUE)->pluck('name', 'id'))
                    ->relationship('wastetype', 'name', fn(Builder $query) => $query->where('status', TRUE)),
                Tables\Filters\Filter::make('date')
                    ->form([
                        Forms\Components\DatePicker::make('date_from'),
                        Forms\Components\DatePicker::make('date_until'),
                    ])
                        return $query
                            ->when(
                                $data['date_from'],
                                fn(Builder $query, $date): Builder => $query->whereDate('date', '>=', $date),
                            )
                            ->when(
                                $data['date_until'],
                                fn(Builder $query, $date): Builder => $query->whereDate('date', '<=', $date),
                            );
                    }),
            ], layout: FiltersLayout::AboveContent)
            ->actions([
            ])
            ->bulkActions([
            ]);
    }
```

### Expected behavior

It should only display sum value and we don't require label Sum for every table cell as we are using groupsOnly method
Please help to get me below output.
Below is image for expected behavior
image

### Steps to reproduce

Install the project and seed data open url /app/reports/wastes to get the output mentioned in problem.

### Reproduction repository

https://github.com/dk009dk/filament-grouponly-table-ui
- Install
- Migrate and seed
- /admin
- Username: admin@admin.com Password: password
- Open /admin/balances

### Relevant log output

_No response_

wychoong:
fixes #9297 

when using summary with groups only renders an empty column
this is because when enabled `->groupsOnly`, a heading is added to the summary row, so it pushes the columns to the right. and the empty column is from the grouped column


packages/tables/resources/views/components/summary/row.blade.php
```diff
@@ -34,7 +39,7 @@
 
     @if ($extraHeadingColumn || $groupsOnly)
         <x-filament-tables::cell
-            class="text-sm font-medium text-gray-950 dark:text-white"
+            class="px-3 text-sm font-medium text-gray-950 dark:text-white"
```
danharrin: Can you send some screenshots of the design change here please?

packages/tables/resources/views/components/summary/index.blade.php
```diff
@@ -16,6 +17,9 @@
     use Filament\Tables\Enums\ActionsPosition;
     use Filament\Tables\Enums\RecordCheckboxPosition;
 
+    if ($groupsOnly && $groupColumn) {
+        $columns = collect($columns)->reject(fn ($column) => $column->getName() == $groupColumn)->toArray();
```
danharrin:
```suggestion
        $columns = collect($columns)
            ->reject(fn (Column $column): bool => $column->getName() === $groupColumn)
            ->all();
```

packages/tables/resources/views/components/summary/row.blade.php
```diff
@@ -16,6 +17,10 @@
     use Filament\Support\Enums\Alignment;
     use Filament\Tables\Enums\ActionsPosition;
     use Filament\Tables\Enums\RecordCheckboxPosition;
+
+    if ($groupsOnly && $groupColumn) {
+        $columns = collect($columns)->reject(fn ($column) => $column->getName() == $groupColumn)->toArray();
```
danharrin:
```suggestion
        $columns = collect($columns)
            ->reject(fn (Column $column): bool => $column->getName() === $groupColumn)
            ->all();
```

packages/tables/resources/views/components/summary/index.blade.php
```diff
@@ -13,9 +14,15 @@
...
     use Filament\Support\Enums\Alignment;
+    use Filament\Tables\Columns\Column;
     use Filament\Tables\Enums\ActionsPosition;
     use Filament\Tables\Enums\RecordCheckboxPosition;
 
+    if ($groupsOnly && $groupColumn) {
+        $columns = collect($columns)
+            ->reject(fn (Column $column): bool => $column->getName() === $groupColumn)
+            ->all();
+    }
```
danharrin, line 25:
```suggestion
    }
    
```

packages/tables/resources/views/components/summary/row.blade.php
zepfietje: Agree we need to be careful here.
If additional spacing is needed, it should probably be added to a wrapper around the heading instead, as `px-3` collides with the horizontal spacing classes on the cell component that depend on screen size.

wychoong: so while the main goal of this is to solve the empty column, I also noticed that the select all checkbox is still visible while I believe it should not when using `groupOnly`

so after the first fix on the empty column, it looks great and well
image

however the checkbox seems wrong thus I removed it. now here comes with the styling issue
image

the issue is with that the heading content is not padded
image
image
image

so this padding is a quick and easy fix
image

zepfietje: To me it seems like the `Type` heading has too much padding actually?

wychoong: no actually I believe is that the empty `td` which are meant for checkboxes is still in the table 
image

of course the ideal one probably is to remove that checkbox column, but I think that requires too much of changes for this PR that I'm comfortable to make

zepfietje: Yeah that's the fix indeed.

wychoong: well it turns out not that many changes that I imagined
image

packages/tables/resources/views/components/summary/row.blade.php
```diff
@@ -36,7 +44,9 @@
         <x-filament-tables::cell
             class="text-sm font-medium text-gray-950 dark:text-white"
         >
-            {{ $heading }}
+            <span class="grid gap-y-1 px-3 py-4">
```
zepfietje: Are all of these classes really needed?

packages/tables/resources/views/index.blade.php
```diff
@@ -55,7 +55,7 @@
     $isReordering = $isReordering();
     $isColumnSearchVisible = $isSearchableByColumn();
     $isGlobalSearchVisible = $isSearchable();
-    $isSelectionEnabled = $isSelectionEnabled();
+    $isSelectionEnabled = ! $isGroupsOnly && $isSelectionEnabled();
```
zepfietje:
```suggestion
    $isSelectionEnabled = $isSelectionEnabled() && (! $isGroupsOnly);
```

packages/tables/resources/views/index.blade.php
```diff
@@ -1069,6 +1070,7 @@ class="fi-ta-record-checkbox"
                                 :actions="count($actions)"
                                 :actions-position="$actionsPosition"
                                 :columns="$columns"
+                                :group-column="$group?->getColumn()"
```
zepfietje, line 1073: Do we actually need the `?` here?

packages/tables/resources/views/components/summary/row.blade.php
wychoong: without the classes 
image

I'm just copying the styling from text-column, else it can be just a `px-3` or `ps-3`

packages/tables/resources/views/index.blade.php
wychoong: it should? line 857, 858 is using the ?

packages/tables/resources/views/components/summary/row.blade.php
zepfietje: Alright, we don't need the grid and gap classes though.

packages/tables/resources/views/index.blade.php
zepfietje: Hmm, in most other places we don't. 🤔

wychoong: don't think it matters much tbh

maytham553-elite:
i have same issue , in relation manger with last update !!

PR #9364, opened 2023-10-29
feature: Import action

danharrin:


pxlrbt:
Shouldn't this be a separate Filament package? Especially since it introduces two migrations that not everybody will need?

danharrin:
The migrations need to be published, so it doesn't actually have a footprint on your app.

The reason why I don't want it as a plugin, is because there are no real dependencies apart from league/csv which everyone has already, and also we'd probably need two separate plugins, one for filament/actions and eventually one for filament/tables to handle the table version of this action, if that makes sense? So I dont mind it being in core, its quite a common use case

packages/actions/docs/07-prebuilt-actions/08-import.md
```diff
@@ -0,0 +1,20 @@
...
+use App\Filament\Imports\ProductImporter;
+use Filament\Actions\ImportAction;
+
+ImportAction::make()
+    ->importer(ProductImporter::class)
+```
+
+The "importer" class needs to be created to tell Filament how to import each row of the CSV.
+
+## Creating an importer
```
zepfietje, line 29: This section still needs to be written? 🙂

packages/actions/resources/lang/en/import.php
```diff
@@ -0,0 +1,77 @@
...
+
+    'modal' => [
+
+        'heading' => 'Import :label',
+
+        'form' => [
+
+            'file' => [
+                'label' => 'File',
+                'placeholder' => 'Upload a .csv file',
```
zepfietje: You probably used ".csv" on purpose?
Just bringing this up to discuss, since it's a CSV file and .csv is the extension.

```suggestion
                'placeholder' => 'Upload a CSV file',
```

packages/actions/resources/lang/en/import.php
```diff
@@ -0,0 +1,77 @@
...
+            ],
+
+            'columns' => [
+                'label' => 'Columns',
+                'placeholder' => 'Select a column',
```
zepfietje, line 20: Shouldn't the placeholder also be plural?

packages/actions/resources/lang/en/import.php
```diff
@@ -0,0 +1,77 @@
...
+            ],
+
+        ],
+
+        'actions' => [
+
+            'download_example' => [
+                'label' => 'Download an example .csv file.',
```
zepfietje: Might be nice to keep it concise?

```suggestion
                'label' => 'Download example CSV file.',
```

packages/actions/resources/lang/en/import.php
```diff
@@ -0,0 +1,77 @@
...
+            ],
+
+            'import' => [
+                'label' => 'Start import',
```
zepfietje: Does this work as well?

```suggestion
                'label' => 'Import',
```

packages/actions/resources/lang/en/import.php
```diff
@@ -0,0 +1,77 @@
...
+    'notifications' => [
+
+        'completed' => [
+
+            'title' => 'Import completed',
+
+            'actions' => [
+
+                'download_failed_rows_csv' => [
+                    'label' => 'Download information about the failed row|Download information about the failed rows',
```
zepfietje, line 48: Haven't checked the UI yet but these labels are quite long?

packages/actions/resources/lang/en/import.php
```diff
@@ -0,0 +1,77 @@
...
+                ],
+
+            ],
+
+        ],
+
+        'max_rows' => [
+            'title' => 'That file is too large to import',
```
zepfietje:
```suggestion
            'title' => 'Uploaded CSV file is too large',
```

packages/actions/resources/lang/en/import.php
```diff
@@ -0,0 +1,77 @@
...
+            'body' => 'You may not import more than :count row at once.',
```
zepfietje: This translation will need to handle singular/plural still.

packages/actions/resources/lang/en/import.php
danharrin: No, the label is used on the fieldset that wraps the columns, whereas the placeholder is used on each column map individually

packages/actions/src/Commands/Concerns/CanGenerateImporterColumns.php
```diff
@@ -0,0 +1,158 @@
...
+                '*_token',
+            ])) {
+                continue;
+            }
+
+            $columnData = [];
+
+            if (in_array($columnName, [
+                'sku',
+                'uuid',
```
zepfietje, line 47: `id` column should also be uppercase.

packages/actions/resources/lang/en/import.php
danharrin: Its ok, the button has its own line

packages/actions/resources/lang/en/import.php
danharrin: We use trans_choice(), we just dont need it in English

packages/actions/src/Commands/Concerns/CanGenerateImporterColumns.php
danharrin: We exclude the primary key from generation I think, as its usually readonly

packages/actions/src/ImportAction.php
```diff
@@ -0,0 +1,373 @@
...
+
+        $this->label(fn (ImportAction $action): string => __('filament-actions::import.label', ['label' => $action->getPluralModelLabel()]));
+
+        $this->modalHeading(fn (ImportAction $action): string => __('filament-actions::import.modal.heading', ['label' => $action->getPluralModelLabel()]));
+
+        $this->modalDescription(fn (ImportAction $action): Htmlable => new HtmlString('<p>' . $action->getModalAction('downloadExample')->toHtml() . '</p>'));
+
+        $this->modalSubmitActionLabel(__('filament-actions::import.modal.actions.import.label'));
+
+        $this->groupedIcon('heroicon-m-arrow-up-tray');
```
zepfietje: Could you please add icon aliases for new icons? 🙂

packages/actions/src/ImportAction.php
```diff
@@ -0,0 +1,373 @@
...
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->label(fn (ImportAction $action): string => __('filament-actions::import.label', ['label' => $action->getPluralModelLabel()]));
+
+        $this->modalHeading(fn (ImportAction $action): string => __('filament-actions::import.modal.heading', ['label' => $action->getPluralModelLabel()]));
+
+        $this->modalDescription(fn (ImportAction $action): Htmlable => new HtmlString('<p>' . $action->getModalAction('downloadExample')->toHtml() . '</p>'));
```
zepfietje: Why's this wrapped in `p` tags?

packages/actions/src/Imports/Http/Controllers/DownloadImportFailureCsv.php
```diff
@@ -0,0 +1,46 @@
...
+use Filament\Actions\Imports\Models\Import;
+use League\Csv\Writer;
+use SplTempFileObject;
+use Symfony\Component\HttpFoundation\StreamedResponse;
+
+class DownloadImportFailureCsv
+{
+    public function __invoke(Import $import): StreamedResponse
+    {
+        abort_unless($import->user->is(auth()->user()), 403);
```
zepfietje, line 15: Might make sense to add more fine-grained control?
Not sure if that's the case, but can any user access the download route, since it's not scoped to a single panel?

packages/actions/resources/lang/en/import.php
zepfietje: Alright, maybe just remove `the` from the label then?

packages/actions/resources/lang/en/import.php
zepfietje: Ah right, thanks for clarifying.

packages/actions/resources/lang/en/import.php
zepfietje: But if the max is 1 row?

packages/actions/src/Commands/Concerns/CanGenerateImporterColumns.php
zepfietje: Fair.

packages/actions/src/ImportAction.php
danharrin: You're gonna hate it but its to add some spacing, otherwise the button is too close to the heading. You can have a look when you pull it down if you want

packages/actions/src/Imports/Http/Controllers/DownloadImportFailureCsv.php
danharrin: 1) This URL is only exposed when the "import completed" notification happens
2) The "import completed" notification only gets sent to `$import->user`

You can bind a new controller to the container if you want to replace this

packages/actions/resources/lang/en/import.php
danharrin: `You may not import more than 1 row at once.`

packages/actions/src/Imports/Http/Controllers/DownloadImportFailureCsv.php
zepfietje: Still a security issue though, even though the route is only exposed temporarily.
Can't we just check `$import->user` here?

packages/actions/resources/lang/en/import.php
zepfietje: Sorry, I meant more than 1.
Will it display `You may not import more than 3 row at once.`?

packages/actions/resources/lang/en/import.php
danharrin: Hm I think this reads better

packages/actions/resources/lang/en/import.php
danharrin: Fixed

packages/actions/resources/lang/en/import.php
zepfietje: But length..? 😅
Could you post a screenshot of what the UI looks like?

packages/actions/resources/lang/en/import.php
zepfietje: Perfect.

packages/actions/resources/lang/en/import.php
danharrin: It fits on one line in the database notifications slideover

SebastiaanKloos:
Import and export should be part of the core imo. Really cool API @danharrin. I like it 🚀

packages/actions/src/Imports/Http/Controllers/DownloadImportFailureCsv.php
pxlrbt: Why not use a signed temporary URL for this?

danharrin: What's the security issue?

We should not use a temporary URL as we cannot guarantee when the user is going to download the CSV, and the URL needs to be stored in the database notifications database.

drbyte:
Wonderful addition!

packages/actions/resources/lang/en/import.php
zepfietje: Alright.

packages/actions/src/Imports/Http/Controllers/DownloadImportFailureCsv.php
zepfietje: Sorry, I was just being blind, lol.
No issues here.

packages/actions/src/ImportAction.php
zepfietje: I checked, and the HTML was completely messed up.
`<p>` tags can't be nested, since that's invalid HTML.

I've removed the wrapping tag and the spacing looks good to me.
Consistent with other places like database notifications modal header actions.

zepfietje (changes requested):
Looks good to me, apart from the `Creating an importer` docs section that's still missing.
Also, the icon alias should be added to the docs, @danharrin.

packages/actions/routes/web.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+use Filament\Actions\Imports\Http\Controllers\DownloadImportFailureCsv;
+use Illuminate\Support\Facades\Route;
+
+Route::get('/filament/imports/{import}/failed-rows/download', DownloadImportFailureCsv::class)
+    ->name('filament.imports.failed-rows.download')
+    ->middleware(['web', 'auth']);
```
Saifallak, line 8: shouldn't this use current panel auth? not just auth?

danharrin: Nope, this is not a panels feature, this is part of filament/actions. We don't have access to the panel middleware

PR #9373, opened 2023-10-30
Hide pagination options if nr. of records is less than the smallest per page select option

ralphjsmit:
I got a request from a client who found it confusing that the per page select dropdown options were still visible, even if a table had only one or a few records. It was not "clean and logical":

Paginator

If there are not many records and the nr. of records is smaller than the , the dropdown to select a nr. of options per page in principle doesn't change anything. 

This PR proposes to remove the dropdown if the nr. of records is smaller than the smallest page option, thus the dropdown having no effect in any situation.

After:

Table

This PR is only about hiding the dropdown. Things like the nr. of records etc are all not touched.

### DIFF
--- a/packages/support/resources/views/components/pagination/index.blade.php
+++ b/packages/support/resources/views/components/pagination/index.blade.php
@@ -7,7 +7,7 @@
 @php
     $isRtl = __('filament-panels::layout.direction') === 'rtl';
     $isSimple = ! $paginator instanceof \Illuminate\Pagination\LengthAwarePaginator;
-	
+
     $smallestPageOption = collect($pageOptions)
         ->filter(fn (int $option) => $option > 0)
         ->sort()
--- a/packages/support/resources/views/components/pagination/index.blade.php
+++ b/packages/support/resources/views/components/pagination/index.blade.php
@@ -9,7 +9,7 @@
     $isSimple = ! $paginator instanceof \Illuminate\Pagination\LengthAwarePaginator;
 
     $smallestPageOption = collect($pageOptions)
-        ->filter(fn (int | string $option) => is_numeric($option)  && $option > 0)
+        ->filter(fn (int | string $option) => is_numeric($option) && $option > 0)
         ->sort()
         ->first();
 @endphp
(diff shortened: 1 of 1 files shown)


packages/support/resources/views/components/pagination/index.blade.php
```diff
@@ -49,7 +54,7 @@ class="fi-pagination-overview text-sm font-medium text-gray-700 dark:text-gray-2
         </span>
     @endif
 
-    @if (count($pageOptions) > 1)
+    @if (count($pageOptions) > 1 && ($isSimple || ($smallestPageOption || $paginator->total() > $smallestPageOption)))
```
danharrin:
```suggestion
    @if ((count($pageOptions) > 1) && ($isSimple || $smallestPageOption || ($paginator->total() > $smallestPageOption)))
```

packages/support/resources/views/components/pagination/index.blade.php
```diff
@@ -7,6 +7,11 @@
 @php
     $isRtl = __('filament-panels::layout.direction') === 'rtl';
     $isSimple = ! $paginator instanceof \Illuminate\Pagination\LengthAwarePaginator;
+
+    $smallestPageOption = collect($pageOptions)
+        ->filter(fn (int | string $option) => is_numeric($option) && $option > 0)
```
danharrin:
```suggestion
        ->filter(fn (int | string $option): bool => is_numeric($option) && ($option > 0))
```

zepfietje:
I've reverted this and the other related PR since there's a couple cases that still need to be handled before being production-ready.

Please request a review from me for UI changes like these in the future, @danharrin. 🙂
There's a good reason why I said the following in Discord 😅:
> Feel free to create a PR *so we can do a proper review to make sure nothing breaks*

Anyway, feel free to attempt a new PR which addresses the following issues, @ralphjsmit:

- Per page selector should be hidden when using simple pagination and the total number of records equals or is smaller than the smallest per page select option.
- The overview label (`showing x of x...`) is always hidden on mobile, which causes an empty table footer if pagination renders nothing.

PR #9387, opened 2023-10-31
Reduce vertical space between sidebar navigation groups

valpuia:

I notice that when we use `$navigationGroup` there is a lot of space between the groups and this PR aims to reduce that space (check settings and test group)
Before:
Screenshot from 2023-10-31 16-16-35
After:
Screenshot from 2023-10-31 16-16-45


zepfietje:
Thanks for bringing this up, @valpuia!
I noticed a couple other things that needed fixing, so I pushed them to this PR.

PR #9394, opened 2023-10-31
[3.x] Support presigned upload URLs in FileUpload

saade:

When using a s3-compatible disk, we don't need to upload the file to the server and then to s3, files can be directly uploaded to the disk, saving unnecessary server processing.

Also, when working on an serverless environment (AWS Lambda using Laravel Vapor for example), there's a hard limit of 6MB for the payload of the request (including files).

This PR adds support presigned uploads using this new api:

```php
FileUpload::make('image')
    ->s3(diskName: 's3', presigned: true)
```

The presigned upload is optional and the new api can be used only as a shortcut to `->disk()` and `->visibility()` if desired.

### DIFF
--- a/packages/forms/src/Components/BaseFileUpload.php
+++ b/packages/forms/src/Components/BaseFileUpload.php
@@ -47,6 +47,8 @@ class BaseFileUpload extends Field
 
     protected int | Closure | null $minFiles = null;
 
+    protected bool | Closure $presignedUpload = false;
+
     protected bool | Closure $shouldPreserveFilenames = false;
 
     protected bool | Closure $shouldMoveFiles = false;
@@ -227,6 +229,15 @@ class BaseFileUpload extends Field
     public function directory(string | Closure | null $directory): static
     {
         $this->directory = $directory;
+
+        return $this;
+    }
+
+    public function s3(string | Closure $diskName = 's3', bool | Closure $presigned = false): static
+    {
+        $this->diskName = $diskName;
+        $this->presignedUpload = $presigned;
+        $this->visibility = 'private';
 
         return $this;
     }
@@ -451,6 +462,11 @@ class BaseFileUpload extends Field
         return (bool) $this->evaluate($this->isOpenable);
     }
 
+    public function isPresigned(): bool
+    {
+        return (bool) $this->evaluate($this->presignedUpload);
+    }
+
     public function isPreviewable(): bool
     {
         return (bool) $this->evaluate($this->isPreviewable);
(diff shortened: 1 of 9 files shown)


packages/forms/src/Components/BaseFileUpload.php
```diff
@@ -231,6 +233,15 @@ public function directory(string | Closure | null $directory): static
         return $this;
     }
 
+    public function s3(string | Closure $diskName = 's3', bool | Closure $presigned = false): static
```
wychoong, line 236: will it throw an error is this api is used on a not supported driver?

saade: Yes, Filesystem handles that already

wychoong: that would happen after an upload is attempted or when the form is rendered?

saade: Upload attempted

saade:
@danharrin should i target 3.1 on this too?

danharrin:
Its ok, this is fine.

Question, I thought Livewire did this already for S3 disks?

saade:
> Its ok, this is fine.
> 
> Question, I thought Livewire did this already for S3 disks?

AFAIK the file is sent to the server then streamed to s3. Presigned uploads should always be client -> bucket, if the file is sent to the server, then your server is processing it.

This PR implements client -> bucket so there's no server involved (besides generating the url)

saade:
My real world probem now is that i need to cap the max file size to 5mb as i'm running on Vapor. I've implemented the same solution into my Flutter app, the file upload process is Flutter -> Bucket to avoid the limits.

danharrin:
Try setting the `temporary_file_upload` config value to s3? https://livewire.laravel.com/docs/uploads#uploading-directly-to-amazon-s3

> If you wish to bypass your application server and instead store Livewire's temporary uploads in an S3 bucket, you can configure that behavior in your application's config/livewire.php configuration file. First, set livewire.temporary_file_upload.disk to s3 (or another custom disk that uses the s3 driver):

danharrin:
pretty sure that's gonna fix it, so I'm gonna close this, but if it doesn't then we can reopen this

saade:
@danharrin Sorry for revisiting this after nearly two years. I’ve recently needed to support large file uploads in my app, and they’re all failing.

Reviewing the code again, without this PR it uses `$wire.upload()`, which sends the entire file to the `/livewire/upload-file` endpoint. Checking the network tab, I can also confirm that the payload size matches the file size.

In my approach, we’re not using $wire.upload; instead, the files are uploaded directly from the browser to S3.

Would you kindly reconsider this? Thank you!

saade:
After years of chasing this issue, I finally found the culprit.

I was using [League’s Filesystem Path Prefixing](https://flysystem.thephpleague.com/docs/adapter/path-prefixing) with a scoped driver to handle tenant folders. That setup was the root cause of the problem.

Switching to a plain s3 disk fixed everything. Since Laravel now supports a `root => ''` configuration for disks, we can simplify things by removing the scoped driver entirely and relying on the built-in disk configuration.

PR #9427, opened 2023-11-02
Load `az` locale from FilePond

nasimic:

### DIFF
--- a/packages/forms/resources/js/components/file-upload.js
+++ b/packages/forms/resources/js/components/file-upload.js
@@ -664,6 +664,7 @@ export default function fileUploadFormComponent({
 }
 
 import ar from 'filepond/locale/ar-ar'
+import az from 'filepond/locale/az-az'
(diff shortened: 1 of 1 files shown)


zepfietje:
Do we still need to wait for a new version of FilePond to be tagged?

nasimic:
@zepfietje I guess we don't. Looks like FilePond not tagging offen 😅

zepfietje:
Yeah that's why I'm wondering. 😅
We can't merge this PR if it's not available in FilePond yet, right?

nasimic:
@zepfietje They are only pushing changes to master branch over 2 years. 
Would be great if they release a tag, but seems not going to happen. I guess there is no polite way to force update for end users 😄

zepfietje:

danharrin:
So isnt this PR broken without Filepond tagging? It hasnt been compiled

zepfietje:
Yeah I think so.

PR #9445, opened 2023-11-03
Set Inter font family to file upload

urufudev:
The Inter font was set in the file upload field because it currently displays Segoe UI. This takes advantage of the fact that tailwindcss already has Inter font configured and is being used.

Before: 
Captura de pantalla 2023-11-03 104558

After:
Captura de pantalla 2023-11-03 104937


zepfietje:
Thanks for spotting this, @urufudev!
Could you rework this PR to use the font family CSS variable that Filament uses internally?

Please mark as ready for review when done.

danharrin:
Please recompile the assets with `npm run build` and test it

urufudev:
> Please recompile the assets with `npm run build` and test it

I was able to test it with 'npm run build,' and it was working correctly. Should I also upload the changes to 'packages/forms/dist/index.css'

danharrin:
Yes, please upload the compiled files

packages/forms/resources/css/components/file-upload.css
```diff
@@ -5,6 +5,7 @@
 @import 'filepond-plugin-media-preview/dist/filepond-plugin-media-preview.css';
 
 .filepond--root {
+    font-family: var(--font-family);
```
zepfietje, line 8: @danharrin @urufudev, can't we just use `font-sans` in `@apply` instead?

danharrin: These files are not compiled with Tailwind.

zepfietje: Hmm fair, but then the font family won't work outside Filament panels...

danharrin: It will fall back to the Filepond default, or they can set the CSS variable. Or they can add the class to their own project.

Or we can try wrapping Filepond in an `!font-sans` in the HTML and hope that `!important` overrides this?

zepfietje: If that doesn't work, we could use `[&_.filepond--root]-font-sans`?

zepfietje: Just like I've done here: https://github.com/filamentphp/filament/blob/932007dbdd36e6b7765eb525df0c448561229a2e/packages/forms/resources/views/components/select.blade.php#L137

danharrin: Sure, can you change that then please?

zepfietje: Actually, these files will be compiled with Tailwind, @danharrin?
There's literally a lot of `@apply` statements in here. 😂
I think we should simply be able to use `font-sans`.
Correct me if I'm wrong though.

danharrin: The end user's Tailwind config is not used, it's our own

zepfietje: Ah of course.

zepfietje: Fixed by https://github.com/filamentphp/filament/commit/5ad3677b4b7ba24702f9b3ba3c0851d7e466fb10.

PR #9457, opened 2023-11-04
Now allowing defaultSort()'s  argument to be a Closure.

telkins:
The `Filament\Tables\Table\Concerns\CanSortRecords` trait's `defaultSort()` method has the following signature:
```php
public function defaultSort(string | Closure | null $column, string | Closure | null $direction = 'asc'): static
```
The `$direction` argument, if not `null`, is passed through `Str::lower()` and then assigned to `$this->defaultSortDirection`, which is a nullable string.

**There is at least one problem**: if a Closure is passed, then `Str::lower()` chokes on it.

image

I decided to do the following:

- pass `$direction` through `evaluate()` right away and allow the rest of the code/types to remain as they are currently
- remove the "last minute" call to `evaluate()` in `getDefaultSortDirection()`

All existing tests pass.  No new tests were added.

### DIFF
--- a/packages/tables/src/Table/Concerns/CanSortRecords.php
+++ b/packages/tables/src/Table/Concerns/CanSortRecords.php
@@ -10,7 +10,7 @@ trait CanSortRecords
 {
     protected ?string $defaultSortColumn = null;
 
-    protected string | Closure | null $defaultSortDirection = null;
+    protected ?string $defaultSortDirection = null;
 
     protected ?Closure $defaultSortQuery = null;
 
@@ -22,6 +22,12 @@ trait CanSortRecords
             $this->defaultSortQuery = $column;
         } else {
             $this->defaultSortColumn = $column;
+        }
+
+        $direction = $this->evaluate($direction);
+
+        if ($direction !== null) {
+            return Str::lower($direction);
         }
 
         $this->defaultSortDirection = $direction;
@@ -62,7 +68,7 @@ trait CanSortRecords
 
     public function getDefaultSortDirection(): ?string
     {
-        return Str::lower($this->evaluate($this->defaultSortDirection));
+        return $this->defaultSortDirection;
     }
 
     public function getDefaultSortQuery(): ?Closure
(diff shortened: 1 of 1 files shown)


telkins:
@awcodes FYI....

There were at least two ways to do this, depending on when/where `evaluate()` should be called on whatever the user passed in.  I opted to go one way first, keeping `evaluate()` to be called late....but then decided to go with the "cleaner" solution, which was to move `evaluate()` to the beginning of the process.

I have mixed feelings, as there may be advantages to having `evaluate()` called later vs earlier.  🤔

packages/tables/src/Table/Concerns/CanSortRecords.php
```diff
@@ -24,6 +24,8 @@ public function defaultSort(string | Closure | null $column, string | Closure |
             $this->defaultSortColumn = $column;
         }
 
+        $direction = $this->evaluate($direction);
```
danharrin: Evaluation should always happen in getters, it should not be evaluated here.

danharrin (changes requested):
Can you please explain why you need to pass a function in?

packages/tables/src/Table/Concerns/CanSortRecords.php
danharrin: ive changed it

telkins:
> Can you please explain why you need to pass a function in?

Sure.  I have a filter that filters from today forward or from today backward, depending.

I wasn't able to find a way to sort in the filter.  The `orderBy()` doesn't seem to be respected.  So, I found that I could pass a function in for `defaultSort()`'s `$direction` argument, determine the filter state, and then sort accordingly.

If there's a better way, I'd love to know about it.

In any case, this was a bug worth fixing, I thought.  🤓

packages/tables/src/Table/Concerns/CanSortRecords.php
telkins: > Evaluation should always happen in getters, it should not be evaluated here.

I had done that first, as you may have seen.  I noticed that this was normally in the getters.  But, I changed my mind thinking you might prefer to keep the changes smaller.

telkins: > ive changed it

Thx.  You did what I had done on my first iteration.  Makes me feel good that I had the right solution...before I changed it.  😅

PR #9460, opened 2023-11-04
Allow dropdowns to be fixed

archilex:
Currently the dropdown panel is positioned with `absolute` for `x-float`. However when using a dropdown inside of a dropdown, this causes the nested dropdown panel to get positioned based on a parent container instead of the viewport. This causes the position to be off in certain instances:

Example 1: The nested panel flips up even though there is viewport space below, becuase the table "parent" is smaller than the panel. 
Screenshot_2023-11-04_at_10 20 52_AM

This is even a bigger issue when the panel is taller than the viewport as the navbar covers it up.
image

Example 2: The nested panel shifts to the left because it's tied with the table. This pushes the panel offscreen causing scrolling.
Screenshot 2023-11-04 at 10 15 55 AM

This PR allows a dropdown to be positioned with `fixed` instead of `absolute`. The default continues to be `absolute` to avoid breaking changes. This is one of the recommended strategies by floating-ui: https://floating-ui.com/docs/misc#clipping

With fixed the above issues are resolved:

Screenshot_2023-11-04_at_10 37 24_AM
Screenshot_2023-11-04_at_10 37 41_AM

### DIFF
--- a/packages/support/resources/views/components/dropdown/index.blade.php
+++ b/packages/support/resources/views/components/dropdown/index.blade.php
@@ -6,7 +6,7 @@
     'teleport' => false,
     'trigger' => null,
     'width' => null,
-    'fixed' => true,
+    'fixed' => false,
 ])
 
 <div
--- a/packages/support/resources/views/components/dropdown/index.blade.php
+++ b/packages/support/resources/views/components/dropdown/index.blade.php
@@ -6,6 +6,7 @@
     'teleport' => false,
     'trigger' => null,
     'width' => null,
+    'fixed' => true,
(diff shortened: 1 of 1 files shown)


zepfietje:
Thanks for bringing this up, @archilex.
Isn't this what `teleport` is for though?
Did the issues you outlined occur with or without `teleport` on the dropdown?

archilex:
@zepfietje Teleport works perfectly and fixes both of the above problems. Didn't know about that. Thanks!

zepfietje:
Cool!

PR #2486, opened 2022-05-20
Add support to HasOneThrough in filters

masterix21:
The PR adds the ability to filter the `HasOneThrough` relationship.

PR #9467, opened 2023-11-05
Restore using HasOneThrough for filters

skeemer:
All of the tests are passing.

I'm not sure that supporting different relationship types is documented anywhere. There is no reference to `\Znck\Eloquent\Relations\BelongsToThrough` in the docs.

No visual changes.

As my primary concern is restoring functionality that appears to have been added twice already in #2486 and #5550, I've focused my testing on getting that working.

However, I did notice one area that might need additional work to fully support HasOneThrough, but I'm not really sure how to work on it. The anonymous function starts on line 811, called by `$this->loadStateFromRelationshipsUsing()`.


danharrin:
I don't think `loadStateFromRelationshipsUsing()` is used for filters, but I've added a check in anyway

PR #9469, opened 2023-11-05
Use wire:navigate for create and edit page redirects when SPA mode is enabled

binaryfire:
Adds a property to enable `navigate: true` for create and edit page redirects. [See these Livewire docs](https://livewire.laravel.com/docs/navigate#redirects) for more info.

This should be configurable independently of SPA mode for these pages because we sometimes need to redirect to external URLs (and external URLs don't work when `navigate:` is set to `true`).

Another option could be adding logic to check if the supplied URL is internal or external when SPA mode is enabled but that will add some performance overhead. Not sure if it's worth it, what do you reckon? Maybe better to just allow toggling it with a property.

### DIFF
--- a/packages/panels/src/Resources/Pages/CreateRecord.php
+++ b/packages/panels/src/Resources/Pages/CreateRecord.php
@@ -2,17 +2,18 @@
 
 namespace Filament\Resources\Pages;
 
-use Filament\Actions\Action;
-use Filament\Actions\ActionGroup;
-use Filament\Facades\Filament;
 use Filament\Forms\Form;
-use Filament\Notifications\Notification;
-use Filament\Pages\Concerns\InteractsWithFormActions;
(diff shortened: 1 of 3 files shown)


binaryfire:
There are other places in the codebase where this can be set automatically based on SPA mode, eg. the user and tenant profile pages. I'll try and make a separate PR for that sometime soon.

binaryfire:
Just pushed another update which removes the property and makes this automatic. The overhead is probably fine- it's just a string evaluation.

danharrin (commented):
Just refactored this to use our existing "app URL detection" logic - `str($url)->startsWith(request()->root())`, which handles situations where the app URL mismatches, potentially in multi-tenant environments.

Does this all still work ok?

binaryfire:
@danharrin Yep, all good!

PR #9470, opened 2023-11-05
Fix: added missing old  param on afterStateUpdated for upload field

### LINKED ISSUE #9185: $old not avaiable for file uploads
### Package

filament/forms

### Package Version

v3.0.80

### Laravel Version

v10.28.0

### Livewire Version

_No response_

### PHP Version

PHP 8.2

### Problem description

According to these docs there it is possible to inject `$old` into `afterStateUpdated()` to get the previous "state"

https://filamentphp.com/docs/3.x/forms/advanced#field-updates

However when trying to do this with a FileUpload I get the following error

>An attempt was made to evaluate a closure for [Filament\Forms\Components\FileUpload], but [$old] was unresolvable.

### Expected behavior

I would expect to somehow know what files are uploaded earlier so I know which ones are new

### Steps to reproduce

Add this to a file upload form
```
->afterStateUpdated(function (?array $state, ?array $old) {
      dd($state, $old);
 })
```

### Reproduction repository

https://github.com/sinnbeck

### Relevant log output

_No response_

asefsoft:
As mentioned in issue #9185, for upload field the `old` param is not injected in `afterStateUpdated` callback. I have added `old` param for that function in this way:

The problem with the upload field is that the `$statePath` contains three parts instead of the usual two found in other fields. It appears like this: `data.{field_name}.{some_random_uuid}`, for example, `data.myfile.f0f5612c-a7fd-4333-a0c9-c1b2cc4a58da`. This unique structure complicates tracking the state of the actual field (`data.myfile`).

To address this, I made a fix in the `updatingInteractsWithForms` function. I added logic to check the number of parts in the `statePath`. If it has three parts, I remove the extra part from the `statePath` before saving the state of that field in the `oldFormState` array. Additionally, I check if there is any uploaded data for that field using `$this->getFormUploadedFiles($newStatePath)`.

Since I'm relatively new to Filament core, I would appreciate feedback on whether this is the correct way to resolve the issue. If not, I'm eager to learn and make necessary adjustments.

### DIFF
--- a/packages/forms/src/Components/BaseFileUpload.php
+++ b/packages/forms/src/Components/BaseFileUpload.php
@@ -195,7 +195,7 @@ class BaseFileUpload extends Field
 
             $this->evaluate($callback, [
                 'state' => $this->isMultiple() ? $state : Arr::first($state ?? []),
-                'old'   => $this->getOldState(),
+                'old' => $this->isMultiple() ? $this->getOldState() : Arr::first($this->getOldState() ?? []),
             ]);
         }
 
--- a/packages/forms/src/Components/BaseFileUpload.php
+++ b/packages/forms/src/Components/BaseFileUpload.php
@@ -195,6 +195,7 @@ class BaseFileUpload extends Field
 
             $this->evaluate($callback, [
                 'state' => $this->isMultiple() ? $state : Arr::first($state ?? []),
+                'old'   => $this->getOldState(),
             ]);
         }
 
--- a/packages/forms/src/Concerns/InteractsWithForms.php
+++ b/packages/forms/src/Concerns/InteractsWithForms.php
@@ -229,6 +229,19 @@ trait InteractsWithForms
 
     public function updatingInteractsWithForms(string $statePath): void
     {
+        // is it UploadField with 3 part state?
+        $parts = explode(".", $statePath);
+
+        if(count($parts) > 2) {
+            array_pop($parts);
+
+            $newStatePath = implode(".", $parts);
+
+            $uploadedFiles = $this->getFormUploadedFiles($newStatePath);
+
+            $statePath = blank($uploadedFiles) ? $statePath : $newStatePath;
+        }
+
         $this->oldFormState[$statePath] = data_get($this, $statePath);
     }
 
--- a/packages/forms/src/Concerns/InteractsWithForms.php
+++ b/packages/forms/src/Concerns/InteractsWithForms.php
@@ -229,25 +229,14 @@ trait InteractsWithForms
 
     public function updatingInteractsWithForms(string $statePath): void
     {
-        // is it UploadField with 3 part state?
-        $parts = explode(".", $statePath);
-
-        if(count($parts) > 2) {
-            array_pop($parts);
-
-            $newStatePath = implode(".", $parts);
-
-            $uploadedFiles = $this->getFormUploadedFiles($newStatePath);
-
-            $statePath = blank($uploadedFiles) ? $statePath : $newStatePath;
-        }
+        $statePath = (string) str($statePath)->before('.');
 
         $this->oldFormState[$statePath] = data_get($this, $statePath);
     }
 
     public function getOldFormState(string $statePath): mixed
     {
-        return $this->oldFormState[$statePath] ?? null;
+        return data_get($this->oldFormState, $statePath);
     }
 
     public function updatedInteractsWithForms(string $statePath): void


packages/forms/src/Components/BaseFileUpload.php
```diff
@@ -195,6 +195,7 @@ public function callAfterStateUpdated(): static
 
             $this->evaluate($callback, [
                 'state' => $this->isMultiple() ? $state : Arr::first($state ?? []),
```
asefsoft: Yes I saw that, I have tested for multiple files and single file upload and for single file upload; the old value is always `null`. it is bcoz of nature of filament or maybe filepond that for single mode you first have to click on X button and remove old file then after that you can add other file and therefor there is never an OLD state and old state is always null for single upload.

It's worth mentioning that file upload field in this context works completely different than simple fields and it calls `callAfterStateUpdated` after you choose files not after you saved the form. 
also in multiple mode it calls `callAfterStateUpdated` for each file you choose separately. if you select 4 files then it will call `callAfterStateUpdated` 4 times and each time with new file added to both $old and new $state.

packages/forms/src/Concerns/InteractsWithForms.php
```diff
@@ -229,6 +229,19 @@ public function getActiveFormsLocale(): ?string
 
     public function updatingInteractsWithForms(string $statePath): void
     {
+        // is it UploadField with 3 part state?
```
asefsoft: in `updatingInteractsWithForms `the index would be `data.myfile.f0f5612c-a7fd-4333-a0c9-c1b2cc4a58da` so is saves something like this in `oldFormState` array:

`$this->oldFormState['data.myfile.f0f5612c-a7fd-4333-a0c9-c1b2cc4a58da'] = data_get($this, 'data.myfile.f0f5612c-a7fd-4333-a0c9-c1b2cc4a58da');`

which would be

`$this->oldFormState['data.myfile.f0f5612c-a7fd-4333-a0c9-c1b2cc4a58da'] = null;`

bcos of that index. we need to correct the index when it save date into $this->oldFormState array.

to become:
`$this->oldFormState['data.myfile'] = [SOME DATA];`

danharrin: Other fields can have 3 part state paths, for example repeaters, builders, any JSON nesting. Also file uploads that are mounted to a Livewire property only have 2 parts. We just cannot rely on this.

danharrin: The solution should be implemented somewhere in the BaseFileUpload class, you probably need to override `getOldState()`. Be aware that if it's `$this->isMultiple()`, we need to return an array of uploads, not just the first.

danharrin: How about just `$statePath = (string) str($statePath)->before('.');`

Which will then always store the most state that you could possibly need from the form, even if we don't need all of it

packages/forms/src/Components/BaseFileUpload.php
danharrin, line 197: You see here how we decide whether or not the upload is multiple, and if it isnt then call `Arr::first()`? I think it's similar for inside `getOldState()`

packages/forms/src/Concerns/InteractsWithForms.php
danharrin: It would replace all the code you added here

asefsoft: That will return just first part :) 
`data.myfile.f0f5612c-a7fd-4333-a0c9-c1b2cc4a58da` => `data`

maybe something like this you prefer
`        $statePath = (string) str($statePath)->explode('.')->slice(0,2)->join('.');`

Yes that is cleaner but in conclusion you approve that the best place to handle this issue is in `updatingInteractsWithForms` right?

and also plz help me to test this type of field you mentioned:

>  file uploads that are mounted to a Livewire property

how can create such field?

asefsoft: Oh! I think you mean we always keep all data of form! I did not read it carefully!

You mean it will be always something like this:

$this->oldFormState['data'] = data_get($this, 'data');

asefsoft: Good, I need to test those other 3 parts to see how they works in this context. but for those 2 part upload field you mentioned => if they are 2 parts then there should not be any problem and they have correct `statePath` so they will save old state correctly as usual. is not true? I will test that too, but plz tell me how I should create such field? =>

> > file uploads that are mounted to a Livewire property

but about `getOldState()` => I dont think that can help bcos root of issue is that in the first place old state is not correctly saved in the `$this->oldFormState`  bcos of that 3 part incorrect index. so we should hook on whenever we are saving into `$this->oldFormState` not when we are getting from it. 
agree?

danharrin: What do you mean it is not correctly saved?

danharrin: Yes

asefsoft: Ok, accordingly we should change `getOldFormState` to use `data_get` instead of getting it directly:

```
    public function getOldFormState(string $statePath): mixed
    {
        return data_get($this->oldFormState, $statePath);
    }
```

also I tested this approach with a fileupload field inside a repeater and all of them are getting their old and new state correctly.

asefsoft: @danharrin as we've discussed the new state path always will be `data`, would you recommend that we do it like this 
```
    public function updatingInteractsWithForms(string $statePath): void
    {
        $this->oldFormState['data'] = data_get($this, 'data');
    }
```
or like this?
```
    public function updatingInteractsWithForms(string $statePath): void
    {
        $statePath = (string) str($statePath)->before('.');

        $this->oldFormState[$statePath] = data_get($this, $statePath);
    }
```

first one would make `$statePath` as an unused parameter and second one looks like doing extra and useless job 😄

danharrin: Second one, forms are not always mounted to `data`, they can have many different state paths

asefsoft:
Ok I made new changes base on our discuses.

asefsoft:
@danharrin, could you please let me know if everything is in order with this pull request, or if there are any other factors to consider?

packages/forms/src/Components/BaseFileUpload.php
```diff
@@ -195,6 +195,7 @@ public function callAfterStateUpdated(): static
...
+                'old'   => $this->getOldState(),
```
danharrin:
```suggestion
                'old' => $this->isMultiple() ? $this->getOldState() : Arr::first($this->getOldState() ?? []),
```

danharrin:
Does it all work as expected now?

asefsoft:
Yes, I tested it in some different scenarios, and everything was fine.

PR #9475, opened 2023-11-05
Add performant table loading indicators using CSS (instead of Livewire / JS)

binaryfire:
A UI tweak to improve the perceived responsiveness of tables.


danharrin:
Can you please add a `$table->loadingOverlay()` (`$this->hasLoadingOverlay`, `$this->hasLoadingOverlay()`) configuration option to enable this, disabled by default. Also some documentation.

Some people like the fact that you can interact with the table while it is loaded.

danharrin:
Also, please merge in `3.1` into your branch.

binaryfire:
@danharrin Sure. Just checking out the codebase for the best way to implement this. A new trait? If you can point me to an existing trait / whatever that follows the approach you're talking about, I'll use that as a reference.

danharrin:
Like `CanDeferLoading`

binaryfire:
@danharrin Updated. Although I'm not sure if the `Filament\Tables\Concerns\HasLoadingOverlay` trait is necessary? Things work fine with just the `Filament\Tables\Table\Concerns\HasLoadingOverlay` trait.

Do you want me to extract the overlay to a separate `support/resources/views/components/loading-overlay.blade.php` component? It's only used in 2 places (the table and grid views) so might be overkill.

zepfietje:
@binaryfire, could you provide a screen recording showing this change?
I'm not too sure about it yet for a couple reasons.
One being the fact that this may get in the way of other loading indicators/states if we decide to reintroduce any in the future (if performance issues allow us to...).

binaryfire:
@zepfietje No problems. This is something I'm really missing after migrating from a Vue SPA. On medium / slow connections (and even fast high-latency connections) there's no feedback that anything is happening after changing something on a table. It makes the whole app feel unresponsive. 

Here's a couple of videos showing the overlay on a throttled connection.

Table: https://www.youtube.com/watch?v=WrMOxbz85vU
Grid: https://www.youtube.com/watch?v=AvXsCn15iq8

All the other datatables packages I've used have something along these lines.

binaryfire:
Works nicely for search too

zepfietje:
Thanks, @binaryfire.

Would this be more performant than the per-row loading skeleton we had in v2, @danharrin @pxlrbt?
I'm ok with reintroducing a default loading state, as long as it doesn't cause performance issues again.
Just want to make sure we've thought this through before merging.
From your videos, the first thing that comes to mind is we need the overlay to have a pulse animation. Otherwise it's still just a static layer which doesn't add much feedback.

binaryfire:
@zepfietje Yeah it's definitely is more performant. I intentionally went for an overlay rather than a per-row option for that reason. You're only changing a single class rather than a class-per-row. The overhead with this should be almost zero.

zepfietje:
@binaryfire, would it somehow be possible to trigger the loading state on a per-row basis based on a CSS class on the parent for example? I would assume performance to be better then compared to what we had in v2.

binaryfire:
@zepfietje Is there any benefit to per-row indicators though? All those actions (searching, sorting, pagination) modify the entire table. I'm not sure what the extra overhead / complexity would bring, and showing multiple indicators while a table is loading could look a bit strange IMHO.

Regarding adding a circle or pulse- that'd be easy. We can just center it in the overlay and perhaps make it optional via a named argument.

zepfietje:
Loading skeleton with rows provides a better UX in my opinion.
Also, that'd allow for a nice loading state without too many visual jumps for deferred table loading.
Not saying it's the final solution, but we should definitely explore it in this PR.

binaryfire:
Ok no worries. Maybe someone else can offer some insights. I couldn't think of any performant per-row options tbh.

Adding / removing classes per-row is going to need JS diffing so it'll incur a performance penalty. And I don't think you can add divs to `<tr>` elements, so the only way to add individual row overlays that could be made visible by toggling a single class would be by adding a div to each cell. That'd create different performance issues because of the increase in dom size.

zepfietje:
We could target the rows from the parent table element with Tailwind using something like `[&_tr]:bg-red-500`?

Then we'd only need to toggle that class on the table root.

binaryfire:
I'm not familiar with that syntax. Is that Sass? 

Something like this might work:

```
<style>
  .loading-state {
      // Default styles
  }
  
  .loading-active .loading-state {
    // When the loading-active class is present, change the loading-state styles
  }
</style>

  <table wire:loading.class="loading-active"`>
      <tbody>
          <tr class="loading-state">
              <!-- Table cells -->
          </tr>
      </tbody>
  </table>

```

But there's no overlay / placeholder, so you'd have to manipulate the `<tr>` styles via `.loading-active .loading-state`

Another option could be adding a pseudo-element to each _cell_ as an overlay. Pretty sure that would work with `<td>`'s (as opposed to `<tr>`'s):

```
@keyframes pulse {
  0% {
    opacity: 0.5;
  }
  50% {
    opacity: 1;
  }
  100% {
    opacity: 0.5;
  }
}

.loading-state td::after {
  content: '';
  display: none;
  position: absolute;
  top: 10%;  /* inset from top edge */
  left: 10%; /* inset from left edge */
  right: 10%; /* inset from right edge */
  bottom: 10%; /* inset from bottom edge */
  background: #E0E0E0; /* Grey background for skeleton loader */
  z-index: 1;
  border-radius: 10px; /* rounded edges */
  animation: pulse 1.5s ease-in-out infinite; /* Add pulse animation */
}

.loading-active .loading-state td::after {
  display: block; /*  show pseudo-element overlay when .loading-active present on table */
}

.loading-active .loading-state td > * {
  visibility: hidden; /* hide contents of underlying cell when .loading-active present on table */
}
```

This is just pseudo-code, I'd have to play around to see if it would work. It means that instead of a bar across each row, we'd have smaller bars across each cell. But you're relying on the browser's CSS engine to render the changes which should be way faster than JS. There's no dom modifications going on and it's GPU-accelerated.

I can give it a shot if you'd be happy with that UI outcome.

zepfietje:
Yeah let's just give it a try, the cell loading skeleton, @binaryfire.
The syntax is *just* Tailwind CSS by the way: https://tailwindcss.com/docs/adding-custom-styles#arbitrary-variants.

binaryfire:
Should we remove the ->loadingOverlay() option if we get the cell option working and just make it default?

binaryfire:
Works pretty well. Haven't tested this with loads of records but it's pure CSS so it shouldn't have any noticeable performance impact: https://www.youtube.com/watch?v=M9pTeiAzthw

The same pseudo-element is applied to all cells so I had to find generic "one-size-fits-all" padding settings. It means the loading skeletons won't line up perfectly with the column headers, but they get pretty close.

I haven't pushed an update for this yet- let me know what you think and I'll tidy it up and push tomorrow.

danharrin:
We removed the loading skeleton in v3, since the performance sending all of those elements over the wire, and diffing them in the browser, was staggering. I'd rather not re-introduce that please.

binaryfire:
@danharrin Hey Dan, have you taken a look at the latest update? It's not doing any of that. This is a pure CSS solution without any of those performance issues.

binaryfire:
@danharrin Actually sorry, I haven't pushed it yet :) Will do it tomorrow and you can have a look

pxlrbt:
@zepfietje  @danharrin We could still introduce some form of per-row indicator via pure CSS which would circumvent the performance issue.  Something like this. Maybe not `blur()` because it's requires quite some performance (for CSS) but we could achieve something similar. Only issue 

```css
.is-loading tr.fi-ta-row::after {
    content: ' ';
    width: 100%;
    height: 100%;
    z-index: 99;
    position: absolute;
    top: 0;
    left: 0;
    backdrop-filter: blur(2px);
}
```

binaryfire:
@pxlrbt @danharrin @zepfietje That's what I've done (CSS indicators). 

Just pushed the changes. They're inline in `table.blade.php` atm and should be extracted to a css file but it's enough to give you an idea.

Video of the result: https://www.youtube.com/watch?v=M9pTeiAzthw

It's disabled and can be enabled with `->loadingOverlay()`. Although I don't see the harm in enabling them by default since there won't be any performance issues.

pxlrbt:
I don't think we really need a method to enable it. If it's working performance wise, it should be fine. I think we should introduce a new class `.fi-ta-cell-content` though that wraps the content itself so we can target that with the loading state. The `top: 35%` solution sounds hacky as it doesn't really account for image and icon columns and probably also others.

binaryfire:
@pxlrbt Yeah good call. I'll make those changes tomorrow.

binaryfire:
Updates:

- Removed `HasLoadingOverlay` configuration traits
- Wrapped action and column cell content in `.fi-ta-cell-content` instead of targeting all `<td>` elements
- Tested with loads of records, didn't notice any difference in performance 🎉

Wrapping specific content was a good idea. The indicators line up perfectly with the column headers now and the checkboxes are no longer affected (they turn grey during loading anyway).

Video: https://www.youtube.com/watch?v=5rMZurkIGjw

The CSS needs to be extracted from `table.blade.php` but not sure where to put it. There's currently no css dir or Tailwind config file in the tables package.

Haven't played around with the grid view yet, will do that soon.

packages/tables/resources/views/components/table.blade.php
```diff
@@ -4,8 +4,60 @@
     'reorderable' => false,
 ])
 
+{{-- TODO: Extract to css file --}}
+<style>
+    @keyframes pulse {
```
zepfietje: Can use Tailwind's `animate-pulse`?

packages/tables/resources/views/components/table.blade.php
```diff
@@ -4,8 +4,60 @@
...
+    }
+    
+    .loading-active .fi-ta-cell-content > * {
+      visibility: hidden;
+    }
+</style>
+
 <table
     {{ $attributes->class(['fi-ta-table w-full table-auto divide-y divide-gray-200 text-start dark:divide-white/5']) }}
+    wire:loading.delay.{{ config('filament.livewire_loading_delay', 'default') }}.class="loading-active"
```
zepfietje:
```suggestion
    wire:loading.delay.{{ config('filament.livewire_loading_delay', 'default') }}.class="fi-loading"
```

packages/tables/resources/views/components/table.blade.php
```diff
@@ -32,4 +84,4 @@ class="divide-y divide-gray-200 whitespace-nowrap dark:divide-white/5"
             </tr>
         </tfoot>
     @endif
-</table>
+</table>
```
zepfietje:
```suggestion
</table>

```

packages/tables/resources/views/index.blade.php
```diff
@@ -986,14 +986,16 @@ class="fi-ta-record-checkbox"
...
-                                            <x-filament-tables::columns.column
-                                                :column="$column"
-                                                :is-click-disabled="$column->isClickDisabled() || $isReordering"
-                                                :record="$record"
-                                                :record-action="$recordAction"
-                                                :record-key="$recordKey"
-                                                :record-url="$recordUrl"
-                                            />
+                                            <div class="fi-ta-cell-content">
+                                                <x-filament-tables::columns.column
```
zepfietje: Please fix formatting and bring back the final newline. 🙂

packages/tables/resources/views/components/table.blade.php
```diff
@@ -4,8 +4,60 @@
...
+        }
+    }
+
+    .fi-ta-cell-content {
+        position: relative;
+        min-height: 1em;
+    }    
+
+    .fi-ta-cell-content::after {
+        content: '';
```
zepfietje: Let's clean up all of this using `@apply` and Tailwind classes.

zepfietje:
Nice work, @binaryfire.
Really like the direction.

zepfietje:
By the way, I think we should also look into column specific loading indicators (like a different skeleton for icon columns for example).

pxlrbt:
@zepfietje Don't think we need different skeletons if we have the `.fi-ta-cell-content` class that wraps the actual content without any padding etc.

zepfietje:
Yeah maybe we won't use it eventually, but it's nice to explore it I think.
Like having circular images actually show circular skeletons.

binaryfire:
@zepfietje @pxlrbt Hey guys. Added logic for the grid layout: https://www.youtube.com/watch?v=3vBDq2chgCA

I've taken this as far as I can- hopefully one of you can tweak it. The rest is a UI/UX CSS thing which isn't my strength. Pseudo-elements cover the entire size of the element they're `::after` so you need to know exactly what to wrap with `.fi-ta-cell-content` to make things look right. I'm just pulling at threads now eg. I removed the padding from the overlay and it looks fine on grid but now it's taking up the full height of table rows.

Summary so far:

1. Added `.fi-loading` to [table](https://github.com/filamentphp/filament/blob/945f7d4318cc1d876b8e2702aeced0e054c30407/packages/tables/resources/views/components/table.blade.php#L9) and [grid](https://github.com/filamentphp/filament/blob/945f7d4318cc1d876b8e2702aeced0e054c30407/packages/support/resources/views/components/grid/index.blade.php#L52) parent elements when `wire:loading`

2. CSS is currently in [container.blade.php](https://github.com/filamentphp/filament/blob/945f7d4318cc1d876b8e2702aeced0e054c30407/packages/tables/resources/views/components/container.blade.php#L1) and needs to be extract and refactored to use `@apply`

3. Added `fi-ta-cell-content` in various places to swap out content for the loader when loading. I tried to get as close to the actual content as I could, but still needs tweaking

zepfietje:
Cool, @binaryfire!
We could also handle the loading skeleton on a per-column basis using Tailwind CSS classes like `[.fi-table.fi-loading_&]:after`.
I think I prefer that, since it keeps styling in the component view files, and allows for easier customization per component. Not sure how much bigger the DOM would get though.

binaryfire:
No worries! This was one of the main things I've been missing since moving from Vue :) Glad we've found a solution.

Any CSS-based approach should be fine performance-wise. As long as you're not adding/remove classes dynamically with Livewire or Alpine (other than 1 or 2 on the parent elements) it'll be fast.

danharrin (changes requested):
I have some concerns about this. At the moment, all the loading indicators are different heights, and it looks kinda weird. Also, there is no space between indicators so they stick together.

My proposal:
- Indicators are placed on the cell element itself, not the cell content. This will allow all indicators to have a fixed height.
- Indicators are inset from the edge, probably by `p-1` or `p-2`, to add space between them
- For advanced table layouts, which don't use cells, the entire record should have a loading indicator, not each component in the layout. So one indicator for each row or each card in the grid.

Does this make sense @binaryfire?

binaryfire:
Hey @danharrin. Yeah the styling isn't right in its current state. See my earlier comment for context: https://github.com/filamentphp/filament/pull/9475#issuecomment-1798304828

Your suggestion sounds good- there are probably many other ways to approach it too. I'm not great with CSS and UI (personally I'd be happy with the original simple overlay lol). I suggest asking @zepfietje or @pxlrbt. Someone with good CSS chops should be able to make it look right.

packages/support/resources/views/components/grid/index.blade.php
```diff
@@ -48,6 +48,7 @@
                 },
             )
     }}
+    wire:loading.delay.{{ config('filament.livewire_loading_delay', 'default') }}.class="fi-loading"
```
zepfietje, line 51: This should be moved up to come before `$attributes`.

packages/tables/resources/css/index.css
```diff
@@ -0,0 +1,40 @@
+@keyframes fi-ta-loading-pulse {
```
zepfietje, line 1: This will still need to be refactored to Tailwind CSS classes with `@apply`.

packages/tables/resources/views/components/actions.blade.php
```diff
@@ -42,7 +42,9 @@ function ($action) use ($record): bool {
         ])
     }}
 >
-    @foreach ($actions as $action)
-        {{ $action }}
-    @endforeach
+    <div class="fi-ta-cell-content">
```
zepfietje, line 45: Not sure about this.

packages/tables/resources/views/components/table.blade.php
```diff
@@ -6,6 +6,7 @@
 
 <table
     {{ $attributes->class(['fi-ta-table w-full table-auto divide-y divide-gray-200 text-start dark:divide-white/5']) }}
+    wire:loading.delay.{{ config('filament.livewire_loading_delay', 'default') }}.class="fi-loading"
```
zepfietje:
```suggestion
    wire:loading.delay.{{ config('filament.livewire_loading_delay', 'default') }}.class="fi-loading"
    {{ $attributes->class(['fi-ta-table w-full table-auto divide-y divide-gray-200 text-start dark:divide-white/5']) }}
```

packages/tables/resources/views/index.blade.php
```diff
@@ -258,7 +258,7 @@ class="fi-ta-header-toolbar flex items-center justify-between gap-x-4 px-4 py-3
                 wire:poll.{{ $pollingInterval }}
             @endif
             @class([
-                'fi-ta-content relative divide-y divide-gray-200 overflow-x-auto dark:divide-white/10 dark:border-t-white/10',
+                'fi-ta-content relative divide-y divide-gray-200 overflow-x-auto dark:divide-white/10 dark:border-t-white/10 [&_.fi-ta-cell-content::after]:bg-gray-300 dark:[&_.fi-ta-cell-content::after]:bg-gray-700',
```
zepfietje, line 261: We should either keep this styling in the Blade view, or move everything to the CSS file.

packages/tables/resources/views/components/actions.blade.php
danharrin: This adds the loading indicator to the actions cell

packages/tables/resources/views/index.blade.php
danharrin: We cannot move this to the CSS file as it needs to be compiled with the user's Tailwind config

pxlrbt: We can still use `@apply`, right?

danharrin: Not for colors, this CSS file is compiled and shipped, not imported into the user's own CSS file

pxlrbt: Oh, is this a separate CSS file? Is is because of support for standalone tables?

danharrin: Yeah

binaryfire:
Are there still plans to implement some version of this? If not, I think the original optional overlay would be a good alternative? Waiting for tables to change on high latency connections with no feedback feels like the app has frozen.

danharrin:
Whoops, I didn't mean for it to close

danharrin:
I don't think I can reopen it, can you? Target 3.x

binaryfire:
"Allow edits by maintainers" got unchecked when it was closed. I just rechecked it - can you re-open it now?

Can't see a way of re-opening it on my end.

binaryfire:
@danharrin Or maybe this: https://github.com/gitbucket/gitbucket/issues/378#issuecomment-789511309

danharrin:
I think you need to make a new PR from the same branch

PR #9481, opened 2023-11-05
making default widgets not lazy by default

MrPunyapal:
Added $isLazy property to AccountWidget and FilamentInfoWidget and make it false.

default widgets do not seem to be lazy loaded as they have very minimal information and no need for an extra request for loading them...


danharrin:
Good point

PR #9483, opened 2023-11-05
Update import "example" doc text

drbyte:
A couple words were missing from this paragraph.


drbyte:
Dan, I also wondered about the last line of the doc, which says this:

> The current record (if it exists yet) is accessible in `$this->record`, and the `[options](#using-import-options)` using `$this->options`.

I found myself having to "think" about what the word "options" in `[options]` meant. It took a moment, but then it made sense that it was the "import action form options". 
How do you feel about changing the `[options]` link to `[import form options]`?
I'm fine with whatever you decide ... just wanted to point it out.

Thanks for all the work that went into this import feature! And its documentation.

danharrin:
Yeah `import form options` sounds great, thanks!

drbyte:
Updated.

PR #9495, opened 2023-11-06
add HasLifecycleHooks for CanUpdateState

noxoua:
- [x]  Changes have been thoroughly tested to not break existing functionality.
- [x]  New functionality has been documented or existing documentation has been updated to reflect changes.

___

Added `HasLifecycleHooks` for `CanUpdateState`.

This allows you to add hooks `before` and `after` for the following columns:

- CheckboxColumn
- SelectColumn
- TextInputColumn
- ToggleColumn

Example:

```php
TextInputColumn::make()
    ->beforeStateUpdated(function ($record, $state) {
        // Runs before the state is saved to the database.
    })
    ->afterStateUpdated(function ($record, $state) {
        // Runs after the state is saved to the database.
    })
```


packages/tables/src/Columns/Concerns/HasLifecycleHooks.php
danharrin, line 1: Please rename these to `afterStateUpdated()`, `callAfterStateUpdated()` etc

Also, please add them directly to the CanUpdateState trait, we don't need a new trait if we're only using them in one place

packages/tables/docs/03-columns/08-text-input.md
```diff
@@ -35,3 +35,17 @@ use Filament\Tables\Columns\TextInputColumn;
 
 TextInputColumn::make('background_color')->type('color')
 ```
+
+## Lifecycle hooks
```
danharrin, line 39: Please also add the relevant documentation for other editable columns

danharrin: Would also be useful to see how the `$record` can be injected. And the `$state`.

packages/tables/src/Columns/Concerns/CanUpdateState.php
```diff
@@ -30,6 +32,8 @@ public function updateState(mixed $state): mixed
             ]);
         }
```
danharrin, line 55: IMO, the hooks should still be run, even if you're using `updateStateUsing()`

packages/tables/src/Columns/Concerns/CanUpdateState.php
```diff
@@ -30,6 +32,8 @@ public function updateState(mixed $state): mixed
...
 
+        $this->callBefore();
```
danharrin: It's probably worth passing in `$this->evaluate(..., ['state' => $state])`, also for after()

packages/tables/src/Columns/Concerns/HasLifecycleHooks.php
noxoua: done ✅

packages/tables/src/Columns/Concerns/CanUpdateState.php
noxoua: done ✅

packages/tables/src/Columns/Concerns/CanUpdateState.php
noxoua: done ✅

PR #9501, opened 2023-11-07
Add A image lightbox in imageColumn

lam0819:

In this pull request,
Our goal is to introduce a basic image lightbox functionality without incorporating any extra libraries.
For a more sophisticated image lightbox effect or if additional libraries are desired, the use of an additional filament plugin is recommended.
Upon clicking an image within the imageColumn, a lightbox popup will be displayed.
image

### DIFF
--- a/packages/panels/docs/06-navigation.md
+++ b/packages/panels/docs/06-navigation.md
@@ -175,6 +175,19 @@ public function panel(Panel $panel): Panel
 }
 ```
 
+By default, when you collapse the sidebar on desktop, the navigation icons still show but without text. You can collapse the sidebar and show the text using the `collapsibleNavigationWithText()` method:
+
+```php
+use Filament\Panel;
+
+public function panel(Panel $panel): Panel
+{
+    return $panel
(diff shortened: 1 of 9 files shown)


zepfietje:
Honestly, I think this would be better suited as a plugin.
Also, you could simply attach an action to the column already and have it open a modal that shows the image.

Going to close this, but will notify @danharrin for a final decision.

wychoong:
> could simply attach an action to the column already and have it open a modal

would like to see this lightbox implementation in the core, a modal with static content (eg: an image or announcement) without a server request

lam0819:
ya, my target is simple lightbox in image. for other attachments, eg pdf , i think need additional plugin.

lam0819:
@wychoong 

Since it will not merge into core function, i make a package to do this now. 
Keep base on ImageColumn and TextColumn.
Support Image , PDF, Word, PPTX too.

Enjoy 

https://github.com/solutionforest/Filament-SimpleLightBox

PR #9504, opened 2023-11-07
Fix inline label for placeholder field

margarizaldi:

```php
Placeholder::make('Info')
    ->inlineLabel()
    ->content('Lorem ipsum dolor sit amet')
```

Before:
image

After:
image

> Currently, for plain string content, it is up to the developers how to align the vertical position (since it is much easier than manually setting up inline label). Let me know if there's a suggestion.

### DIFF
--- a/packages/forms/resources/views/components/placeholder.blade.php
+++ b/packages/forms/resources/views/components/placeholder.blade.php
@@ -3,6 +3,7 @@
     :id="$getId()"
     :label="$getLabel()"
     :label-sr-only="$isLabelHidden()"
+    :has-inline-label="$hasInlineLabel()"
     :helper-text="$getHelperText()"
     :hint="$getHint()"
     :hint-actions="$getHintActions()"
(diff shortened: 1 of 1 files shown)


danharrin:
Can you please fix the vertical padding, its a little out of line in inline mode

packages/forms/resources/views/components/placeholder.blade.php
```diff
@@ -15,7 +15,10 @@
         {{
             $attributes
                 ->merge($getExtraAttributes(), escape: false)
-                ->class(['fi-fo-placeholder sm:text-sm'])
+                ->class([
+                    'fi-fo-placeholder sm:text-sm',
+                    'py-1.5' => $hasInlineLabel() && !$getContent() instanceof \Illuminate\Contracts\Support\Htmlable
```
margarizaldi: Also prevent py-1.5 for Hmtlable content because sometimes we want to use border or different background for the content.

zepfietje:
I'm not sure if fixing this on the entry level is the right approach.

danharrin:
I think it is, because other entries like Image and Color don't need the padding above

zepfietje:
Fixed in https://github.com/filamentphp/filament/pull/9892/commits/e4993aefc64e08a22ed3eeeacba1444899fc1e1e as part of https://github.com/filamentphp/filament/pull/9892.

PR #9514, opened 2023-11-07
add luxon adapter to Chart.js

### LINKED ISSUE #9359: Chart widget - Time Cartesian Axis not working
### Package

filament/widgets

### Package Version

v3.0.87

### Laravel Version

v10.29.0

### Livewire Version

v3.0.10

### PHP Version

PHP 8.2.12

### Problem description

I tried using a chart widget of type line with this function

```
    protected function getOptions(): ?array {
        return [
            'scales' => [
                'x' => [
                    'display' => true,
                    'type' => 'time',
                ]
            ],
        ];
    }
```

This results in showing an empty chart and throwing an error in browser console

grafik

relevant info:
#6088
#8350

One solution would be to make it optional to add a adapter into this line:
https://github.com/filamentphp/filament/blob/7c0f4048e8042b91965aecf3e61dbe1b3ff06a3a/packages/widgets/resources/js/components/chart.js#L2

see https://www.chartjs.org/docs/latest/axes/cartesian/time.html#date-adapters for details

### Expected behavior

Error should not be thrown and chart should be displayed

### Steps to reproduce

Log into admin panel and navigate to Dashboard

### Reproduction repository

https://github.com/manhofmann/filament-time-adapter

### Relevant log output

_No response_

manhofmann:

This PR closes #9359


danharrin:
How about we just install the adapter by default? Which is the smallest adapter we can use?

manhofmann:
> How about we just install the adapter by default? Which is the smallest adapter we can use?

I thought about this as well. I don’t know which adapter to chose to be honest. What would the criteria be? Only the smallest one? 

Personally the luxon one worked well for me. 

What if someone wants to have a different adapter because of some missing feature? Making customizable avoids this issue. Looking at the Issues and Discussions it seems to be not used by many developers. So enabling it if needed makes sense.

danharrin:
All I know is that the customisation process here isn't strong enough to document, and having a small adapter included by default probably wouldnt hurt. I bet the adapter has other use cases too?

manhofmann:
I think the other use cases are about formatting e.g. Dates as far as I know, if I understand your question correctly. However I believe other adapters might support this as well but I’m not sure.

danharrin:
Lets install the adapter you want to use, and evaluate the size after

manhofmann:
With the luxon adapter another 64kb have been added in the compiled chart.js file:

without adapter: 193kb
with adapter: 257kb

wychoong:
do you mind to test with dayjs adapter as well, to compare the size, as filament already using it
afaik the difference is timezones, and I think that can be handled in php before passing to the chart object

danharrin:
64kb is pretty small, so I'm happy with that.

The Day.js adapter is not official, so I'm not sure I trust it to be maintained as well. Also, I don't think we'd be able to avoid importing Day.js twice, as some pages may have a date picker, some may have a chart, there's no sharing there

PR #9527, opened 2023-11-08
Add more missing icon aliases

zepfietje:

---

Couple of things for discussion:

- [x] Prebuilt actions aliases are now different from other actions (`-action` suffix vs. no suffix). Do we want to update prebuilt actions' aliases in a backward compatible way?
- [x] For themes, I'd like to change the icons for panels page stubs, resource stubs, and settings page stub. Are we okay with moving these icons from the stubs to the extended classes so we can use icon aliases?
- [x] `danger()`, `info()`, `success()`, and `warning()` methods in `Notification.js` don't use an alias yet. How do we want to add them? Changing the underlying API for these methods to a `$status` property?


danharrin:
I think keeping the `-action` suffix is ok, inline with the class name

For the stubs, we can resolve the icon upon generation and pass it to the stub

Same for the notifications, resolve the icon and then pass it to the notification icon property

packages/panels/resources/views/components/tenant-menu.blade.php
```diff
@@ -91,9 +91,10 @@ class="ms-auto h-5 w-5 shrink-0 text-gray-400 transition duration-75 group-hover
                 <x-filament::dropdown.list.item
                     :color="$profileItem?->getColor()"
                     :href="$profileItemUrl ?? filament()->getTenantProfileUrl()"
-                    :target="($profileItem?->shouldOpenUrlInNewTab() ?? false) ? '_blank' : null"
                     :icon="$profileItem?->getIcon() ?? 'heroicon-m-cog-6-tooth'"
+                    icon-alias="panels::tenant-menu.profile-button"
```
danharrin: I think this would break being able to set an icon on the MenuItem instance, as getIcon() would be ignored

Maybe, instead of using this attribute, we need to resolve the icon after ?? and use it as the default

zepfietje:
> I think keeping the `-action` suffix is ok, inline with the class name

Yeah, though the translations don't use the suffix. Thoughts?

> For the stubs, we can resolve the icon upon generation and pass it to the stub

How would that work with per-panel themes? Also, we should use a different icon when using top navigation, just like already happens for the other pages.

> Same for the notifications, resolve the icon and then pass it to the notification icon property

How exactly? There's currently no way to pass props to the `Notification` JS class?

packages/panels/resources/views/components/tenant-menu.blade.php
zepfietje: Oh, I just copied this over from the user menu, but I think you're right.

danharrin:
> Yeah, though the translations don't use the suffix. Thoughts?

I don't think it's worth making and handling it as a BC, for such a minor inconsistency

> How would that work with per-panel themes? Also, we should use a different icon when using top navigation, just like already happens for the other pages.

Let's actually leave it. Since the code is in the user's own codebase, I don't think it should matter which icon gets added. They can easily change it. I doubt anyone uses the default icons anyway, they're just a placeholder.

> How exactly? There's currently no way to pass props to the Notification JS class?

Fine, let's introduce a status then. But we still need the icon prop for overrides

zepfietje:
> I don't think it's worth making and handling it as a BC, for such a minor inconsistency

Fair. Let's keep using the suffix for actual action classes for now to be consistently inconsistent lol.

> Let's actually leave it. Since the code is in the user's own codebase, I don't think it should matter which icon gets added. They can easily change it. I doubt anyone uses the default icons anyway, they're just a placeholder.

Alright, just thought it'd be a nice to ensure people use the right icon type (e.g. mini vs. outline for Minimal Theme). But agree it's not worth it.

> Fine, let's introduce a status then. But we still need the icon prop for overrides

Yeah, exactly.

PR #9530, opened 2023-11-08
Update 02-getting-started.md

bartduisters:
Rephrase `pets` to `cats, dogs, or rabbits` as is done the first time in the guide.

### DIFF
--- a/packages/panels/docs/02-getting-started.md
+++ b/packages/panels/docs/02-getting-started.md
@@ -76,7 +76,7 @@ public function boot(): void
 
 ### Setting up relationships between models
 
-Let's set up relationships between the models. For our system, pet owners can own multiple pets (patients), and patients can have many treatments:
+Let's set up relationships between the models. For our system, pet owners can own multiple patients (cats, dogs, or rabbits), and patients can have many treatments:
 
 ```php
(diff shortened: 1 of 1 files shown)


zepfietje:
I think the current docs read a little better.
Feel free to clarify if I'm missing something though.

bartduisters:
It's not a big issue, but in the first section of that page there is this sentence:

It will support adding new patients **(cats, dogs, or rabbits)**, assigning them to an owner, and recording which treatments they received.

Which implies to me that a patient will always be either a cat, dog or rabbit.

Later on there is once `pets (patients)` and once `patients (pets)`. It is clear to me that pets and patients are used as synonyms, but it also made me think that now all of a sudden a patient could be anything, not just a dog, cat or rabbit.

To have it consistent I suggested the change.


Otherwise my suggestion is to keep the order consistent and swap `pets (patients)` to `patients (pets)` as well.

If none of the suggestions are applied, that's fine as well. Could just be me that took it too literal.

zepfietje:
I think Dan wrote the docs like this initially, since it refers to patients explicitly when talking about the model, while the introduction is more of an explanation of the business logic.

PR #9556, opened 2023-11-09
Support Closures in the registration of Filament Panel Colors

coolsam726:

This PR adds support for using closures in each of the colors passed as an array to the `FilamentColor::register()` method. This makes it possible to evaluate theme colors during the application's runtime, which is currently only possible using a custom Middleware. A good use case for this is in a multi-tenant app where tenants should be able to set their own theme colors as part of the tenant configuration.

With this change, the following color registration will be possible from any service provider's boot method:

```php
use Filament\Support\Colors\Color;
use Filament\Support\Facades\FilamentColor;

FilamentColor::register([
    'primary' => fn() => tenant()?->primary_color ?: Color::Indigo,
    'info' => fn() => Filament::getTenant()?->info_color ?: Color::Amber,
]);
```

Additionally, it will also be possible to use closures in the `$panel->colors()` method as follows:

```php
use Filament\Panel;
use Filament\Support\Colors\Color;

{
    return $panel
        // ...
        ->colors([
            'primary' => fn() => Filament::getTenant()?->danger_color ?: Color::Indigo, // Evaluate the color during runtime
            'info' => fn() => settings()->info_color ?: Color::Amber,
        ]);
}
```


zepfietje (changes requested):
Let's only accept a closure as the top-level argument, instead of per color. Offers the same flexibility and should be more performant.

coolsam726:
> Let's only accept a closure as the top-level argument, instead of per color. Offers the same flexibility and should be more performant.

That was my initial consideration.My only worry was: Will this still allow plugins to override only a single color without necessarily setting all the six colors at once especially via `$panel->colors()`? I have done tests and confirm this is still possible. Submitting the changes shortly.

danharrin:
Please run Pint

coolsam726:
> Please run Pint

Done.

PR #9557, opened 2023-11-09
Add Documentation about resource declaration within pages.

tonypartridge:

### DIFF
--- a/packages/panels/docs/14-upgrade-guide.md
+++ b/packages/panels/docs/14-upgrade-guide.md
@@ -184,22 +184,6 @@ In v2, you can customize the logo of the admin panel using a `/resources/views/v
 
 Filament v3 has a new universal plugin system that breaches the constraints of the admin panel. Learn how to build v3 plugins [here](plugins).
 
-#### Pages
-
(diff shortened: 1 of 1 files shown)


danharrin:
Hmm, can you explain a bit more? I don't think we changed anything here? Were you using Filament/Pages/Page before in v2?

danharrin:
I'm going to close this until there is more clarity :)

tonypartridge:
Sorry Dan, there was a discord issue; see:

https://discord.com/channels/883083792112300104/1171832702765371553

It looks like when custom pages may have been created and the definition is not declared. But I've asked in the chat for the whole start so we can see where they were extending from.

tonypartridge:
So yeah it was extending:

use Filament\Resources\Pages\Page;

danharrin:
That class in v2 also required `$resource`

tonypartridge:
Thanks Dan, I've checked and in prod we have the attached. Which works flawlessly with filamentphp 2.17.52 

image

But requires the resource classification on V3. I've noticed a few posts about this now on Discord too.

danharrin:
I think this only applies to pages that don't render, and redirect in `mount()`, but fair enough

PR #9563, opened 2023-11-09
Allow infolist modal to infolist action

damiandejong:

With the current version of Filament, and the situation below, the action does not execute:
```
use Filament\Infolists\Components\TextEntry;
use Filament\Infolists\Components\Actions\Action;
use Filament\Resources\Resource;
use Filament\Infolists\Infolist;

class ProjectResource as Resource
{
     // ...

     public static function infolist(Infolist $infolist): Infolist
     {
          return $infolist
               ->schema([
                    TextEntry::make('description')
                         ->suffixAction(Action::make('revisions')
                              ->infolist(fn ($infolist, $record) => $infolist
                                   ->record($record)
                                   ->schema([...]))
               ])
     }

     // ...
}
```

Perhaps there is a grounded reason for why it is not possible to have an action inside an infolist with a modal that displays another infolist, but I guess then `Filament\Infolists\Components\Actions\Action` should have the `infolist()` function.

The reason it was not working, is that the `getInfolistName()` function inside `Filament\Infolists\Components\Actions\Action` is not set, and therefore, `getInfolist()` will return `null`.

### DIFF
--- a/packages/infolists/src/Components/Actions/Action.php
+++ b/packages/infolists/src/Components/Actions/Action.php
@@ -79,4 +79,9 @@ class Action extends MountableAction
 
         return $action->component($this->getInfolistComponent());
     }
+
+    public function getInfolistName(): string
+    {
+        return 'mountedInfolistActionInfoList';
+    }
 }
--- a/packages/infolists/src/Components/Actions/Action.php
+++ b/packages/infolists/src/Components/Actions/Action.php
@@ -82,6 +82,6 @@ class Action extends MountableAction
 
     public function getInfolistName(): string
     {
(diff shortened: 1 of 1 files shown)


danharrin:
But `mountedInfolistActionInfoList()` isnt a method that is defined, check for usages of that string in the codebase and I can't find any? Did you test this?

damiandejong:
Sorry, I made a typo in the name... But with that being fixed, adding `getInfolistName()` does indeed solve my problem. If you want, I can setup a demo repo to show it's not working in the current version of Filament.

danharrin:
Thanks

PR #9565, opened 2023-11-09
Added the icon position to the action group

SebastiaanKloos:

The action group trigger was missing the icon position. This PR fixed that.


zepfietje:
Icon position is not passed there as not all action group types support it (e.g. badge).
I expected it to work with `$attributes`, but it probably doesn't because a prop exists for it?

zepfietje:
Could you check if the same problem occurs for badge and badge color on action groups?

PR #9574, opened 2023-11-10
Add smaller avatar sizes

noxoua:

The size `xs` is perfectly suitable for the select component.
This will make it easier to add other sizes without having to build a css.

____

Screenshot 2023-11-10 at 11 16 01

____

```php
Select::make('user')
    ->native(false)
    ->allowHtml()
    ->options(function () {
        return User::get()
            ->map(fn ($user) => [
                'value' => $user->id,
                'label' => Blade::render(
                    '<x-filament::avatar
                        src="' . filament()->getUserAvatarUrl($user) . '"
                        size="xs"
                        class="inline mr-2"
                    /> ' . $user->name
                )
            ])
            ->pluck('label', 'value');
    });
```

### DIFF
--- a/packages/support/docs/09-blade-components/02-avatar.md
+++ b/packages/support/docs/09-blade-components/02-avatar.md
@@ -27,7 +27,26 @@ Avatars are fully rounded by default, but you may make them square by setting th
 
 ## Setting the size of an avatar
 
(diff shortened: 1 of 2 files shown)


zepfietje:
Thanks, added small size in https://github.com/filamentphp/filament/commit/6e1089e49f197a41f59d408c227ef9f1808d53e9.

PR #9622, opened 2023-11-13
Fix record key route resolver in resources

a21ns1g4ts:


danharrin:
There are many other places where getUrl() is used, and I don't think it's reasonable to fix this in every one of those places.

Instead, you can edit `getPages()` and add the key name to the parameter: `{record:slug}` etc

PR #9665, opened 2023-11-15
Add panel colors getter

binaryfire:
I've been trying to sync the styling of some stand-alone pages with a panel. Getting the colors has been challenging. This will allow us to get the panel then call `$panel->getColors()` to retrieve the colors (like we already can with the font, brand logo etc).


flolanger:
Thank you!! ❤️

PR #9713, opened 2023-11-17
Added groupedActions() to table

amirfallahii:
adds a new method groupedActions() to allow grouping actions together in the table


zepfietje:
You can simply use the existing actions API with the action group object already.
This PR would only introduce inconsistencies.

PR #9728, opened 2023-11-18
Add 2xl icon column size

binaryfire:
Sometimes you want larger icons for things like categories when using a card layout.


danharrin (changes requested):
Please do the same for IconEntrySize and the infolists docs

binaryfire:
Done

PR #9730, opened 2023-11-18
Hide group selection when needed

martinmildner:
This pull request adds the ->groupSelectionVisible method to exclude the Group Selection from the table output. This makes sense especially when only one group or ->defaultGroup() is used.

Example:

```php
use Filament\Tables\Table;

{
    return $table
	->defaultGroup('status');
        ->groupSelectionVisible(false);
}
```

Screenshot:
Screenshot-2023-11-18 at 12 50 10@2x

### DIFF
--- a/packages/tables/docs/08-grouping.md
+++ b/packages/tables/docs/08-grouping.md
@@ -333,9 +333,7 @@ use Filament\Tables\Table;
 public function table(Table $table): Table
 {
     return $table
-        ->groups([
-            // ...
-        ])
+		->defaultGroup('status');
         ->groupSelectionVisible(false);
 }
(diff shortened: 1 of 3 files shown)


zepfietje:
Shouldn't this be done automatically instead of adding a new API?

martinmildner:
I don't think so. Even if there's just one group, someone might want to offer the user the option to choose between ascending and descending order. Don't you think?

Screenshot-2023-11-18 at 13 04 19@2x

zepfietje:
Yeah right, forgot about the sort direction.

martinmildner:
Thanks, Dan!!

PR #9737, opened 2023-11-18
feature: Navigation sub groups

danharrin:
When a parent item is active, the sub items will be visible.

image

All items that have a parent item do not have an icon

Except when the sidebar is partially collapsed:

https://github.com/filamentphp/filament/assets/41773797/b8dd4ef9-31e8-4c74-b541-7a6296d925d7

We will not be going further than 3 navigation levels. This is the absolute limit we are prepared to support UI-wise in Filament.


danharrin:
Please note that this does not cover top navigation, as there is extra complexity surrounding the fact that this is a "parent item" feature rather than another level of navigation groups, and there isn't an easy way to make the parent item clickable as well as being able to open the child item dropdown. This can be revisited at another time.

zepfietje (changes requested):
Quick first thing I noticed from the screenshot.
The child items line and dot should start with a dot from the first item. Currently it starts with a line.

danharrin:
IMO it should not do that, if the first child has a dot then it doesn't look like a child list anymore, it looks like it's at the top level of the parent

zepfietje:
Hmm, fair point, but I still don't like it.

danharrin:
As I said before, I'm just copying Shopify, this is basically how they do it 🤷

zepfietje:
Yeah I noticed, though I prefer to not look at just 1 platform.
Having at least 3 gives us a more solid base.

glennjacobs:
I am very much in favour of this and I think the design works OK. The solution is nice and simple (K.I.S.S.).

And of course, this would help us with our little Filament build ;-)

@zepfietje is it the UI design or the UX you're not sure on? If it's just the look of it, I'm happy to suggest a few ideas using Figma.

zepfietje:
@glennjacobs, it's probably just the dot in front of nested items I'm unsure about. Probably we should also disallow having icons for nested items, as there's no visual hierarchy if you have icons for the parent and child items.

Apart from that, I've discussed the UX quite thoroughly with Dan and it turns out to basically be the same design as I suggested on that other PR.

Will review this PR, implement any fixes, and will let you know if there's something I'd love to have your thoughts on, @glennjacobs. 😊

Appreciate your input.

glennjacobs:
I quite like the way Statamic do it. Might be worth checking out.

zepfietje:
Yeah that's essentially the same thing, without icons for nested items.
GitHub does the same.

danharrin:
As mentioned on Discord, no need to prevent icons from being added as they are used for the collapsed sidebar

saade:
@zepfietje not related to this PR, but noticed that the the icons are not horizontally centered within collapsed sidebar. Worth going through this when reviewing this PR

image

zepfietje:
@saade, is that a regression?
Please open a separate issue so we can have a look at the specifics (which OS/browser it occurs on, since I haven't run into it myself).

zepfietje (changes requested):
Currently, the (also new in 3.1) sub-navigation feature doesn't support navigation subgroups. Maybe we need to move the sub-navigation to dropdowns instead of a select on mobile?

saade:
> @saade, is that a regression? Please open a separate issue so we can have a look at the specifics (which OS/browser it occurs on, since I haven't run into it myself).

I don't think its a regression, its currently live in 3.x. I've just pasted here because of the design review you'll go through on this PR.

Your sidebar looks ok?

danharrin:
Now supports subnavigation

zepfietje (approved):

PR #9742, opened 2023-11-19
add slots to input wrapper component

marvinosswald:
This is supposed to reduce duplicate code when developing custom fields.
With this change `x-filament::input.wrapper` can accept two additional slots:

- `x-slot:suffixSlot`
- `x-slot:prefixSlot`

for example this change enables development (without duplicating all the input-wrapper code) of a custom field to suffix a TextField with a select input.

text_field_with_select

My implementation allows for affix actions, labels and icons alongside the slot - there might be a discussion to only have one or the other.

### DIFF
--- a/packages/support/resources/views/components/input/wrapper.blade.php
+++ b/packages/support/resources/views/components/input/wrapper.blade.php
@@ -9,11 +9,13 @@
     'prefixIcon' => null,
     'prefixIconColor' => 'gray',
     'prefixIconAlias' => null,
+    'prefixSlot' => null,
     'suffix' => null,
     'suffixActions' => [],
     'suffixIcon' => null,
     'suffixIconColor' => 'gray',
     'suffixIconAlias' => null,
+    'suffixSlot' => null,
     'valid' => true,
 ])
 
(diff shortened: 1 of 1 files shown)


danharrin:
Please use the `Filament\Support\is_slot_empty` helper to perform the checks

marvinosswald:
> Please use the `Filament\Support\is_slot_empty` helper to perform the checks

thanks for looking into my pr, didn't know about the helper, it's changed 👍

danharrin:
I've re-implemented this feature as `start` and `end`, which is more consistent with our other components. I hope this is still useful? I think this is more flexible and will prevent us from needing to add other slots. I do realise that this is probably slightly less convenient to use than the previous implementation due to the fact that the slots are outside of the prefix and suffix containers, but you an copy the class from those containers if you need them.

Alternatively, you could use the existing `prefix` and `suffix` props as slots, but they would get wrapped in divs.

danharrin (approved):

zepfietje:
@danharrin, isn't the existing `prefix` and `suffix` slot exactly what @marvinosswald is looking for?

I'd prefer to not add any additional slots, as it's a hell to maintain.

danharrin:
This is to avoid the wrapper elements for the prefix and suffix, these slots are not increasing maintenance burden as long as they remain at the start and end of the element IMO

zepfietje:
But what's the use for these slots without the wrapper elements?
The current input wrapper is already a nightmare to maintain with all possible options for prefixes and suffixes, which will only get worse if devs expect things to look good when using these new slots.

Honestly, I don't see the use. 🤔

zepfietje:
@marvinosswald, what are you running into when using the existing `suffix` slot?

marvinosswald:
@zepfietje to be honest i've just tried the suffix slot, works fine for my purpose. I've changed my Field around quite a bit since i wrote the PR but can't find the reason why i this didn't work originally. I'm sorry i wasted everyones time on this one. I might have sprung to conclusions to quickly when opening this PR.

zepfietje:
Thanks for checking, @marvinosswald.
I've removed the new slots in https://github.com/filamentphp/filament/commit/d788225d9ceb6c7dfc36239b93aa62b896c40a18.

PR #9755, opened 2023-11-20
improved handling of rare morph relationship scenarios

ihebbouzid:
Refactor save function to enhance support for uncommon scenarios in morph relationships

In certain rare scenarios involving multiple morph relationships with the same model, specifically when dealing with customized queries, the existing save function needed modification to ensure robust handling. For instance, consider the case where a model, such as 'Product,' has a morph-to-many relationship with categories, but with distinct query conditions for different category subsets, like 'Gaming Categories' and 'Tech Categories.' This commit addresses these complexities, providing a more versatile and resilient solution for managing diverse morph relationships within the application.


danharrin:
Hi, in an effort to avoid the complexity here, I would suggest that you instead use `Select::make()->saveRelationshipsUsing()` and write this code in your own application. I am not confident enough in why it is written the way it is to take on the responsibility of maintaining it right now.

PR #9763, opened 2023-11-20
Select setting the correct data if canSelectPlaceholder is false

schmeits:
Added additional test

If you create a form with a select on it, and you set the canSelectPlaceholder(false) the first item of the select is selected, but the data is still null. This PR fixes the data to reflect the item being shown on your screen. Normally, you would have to set the default value in your Form component fetching the options again and selecting the first key as default value.

### DIFF
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -115,7 +115,16 @@ class Select extends Field implements Contracts\HasAffixActions, Contracts\HasNe
     {
         parent::setUp();
 
-        $this->default(static fn (Select $component): ?array => $component->isMultiple() ? [] : null);
+        $this->default(static function (Select $component): int|array|null {
+            if ($component->isMultiple()) {
+                return [];
+            }
+            if (! $component->canSelectPlaceholder())
+            {
+                return collect($component->getOptions())->keys()->first();
+            }
+            return null;
+        });
 
         $this->afterStateHydrated(static function (Select $component, $state): void {
             if (! $component->isMultiple()) {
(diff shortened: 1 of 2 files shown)


danharrin:
I'm not sure that this is the desired behaviour. I feel like it could result in forms being filled out inaccurately.

Maybe the user doesn't notice that a field isn't filled out, because it has a first option selected. At the moment, an error will occur if they don't manually select an option. With this change, no error will occur, and the form will submit with potentially the wrong information.

I think it's probably the user's responsibility to implement this default value themselves, if they feel like it improves the UX of their own application. But having it there by default is probably a breaking change.

What do you think?

danharrin:
The alternative is to make `canSelectPlaceholder()` simply disable the placeholder option, rather than hide it altogether. That will still force the user to select the option, but the first option in the select is not in the placeholder by default. I'm open to thoughts on that as well.

schmeits:
Hi Dan, 

I get your point from a UX perspective, but It's also the default functionality of an HTML select element. If I create a select in a form without a placeholder item as the first item in the select (same as using the `canSelectPlaceholder(false)`) then the JS value of that select is also the value of the first element.

My use case is that the user must select one of 3 options, for example. Default, I want to select the first item of the select without giving the user the possibility to unselect it. The select is a reactive field triggering another section in the form with a dynamic form portion.

Poh It's a tough one, maybe make it optional some way to set the data value to the first item in the option list?

schmeits:
My current solution is partially duplicating my code 

```
      ->options(fn () => Model::pluck('title', 'id'))
      ->default(fn () => Model::first()->id)

```

danharrin:
However, the current implementation in this PR is also not efficient, it fetches all options just to get the first one. So `Model::all()` twice, once for the default and once for the options. This could be accomplished better if the user writes the `default()` themselves.

I think it's best if we leave this up to the user, I think there are too many unanswered questions, and passing a `default()` is quite easy.

PR #9778, opened 2023-11-20
Adjusting rules checking user input against a list of allowed values

mariusticha:
## template

  -> indeed it does change the validation functionality 
  -> no changes needed
  -> no visual changes

## context

https://discord.com/channels/883083792112300104/1174699150420476014

after taking a look into the implementation of filaments fluent rule api's, i recognized there are some more rules that have the same behaviour like the `->in()` rule: 

```
->notIn()
->startsWith()
->doesntStartWith()
->endsWith()
->doesntEndWith()
```

please note that i've adjusted all of these rules in this PR. 

## background

my idea was that all of these rules should have the exact same behaviour like passing their laravel's equivalent into a custom rule, e.g.

``` php
// user input: 'foo'
->in([])               // fails
->rule(Rule::in([])    // fails
```

whereas the current implementation works like this:

``` php
// user input: 'foo'
->in([])               // passes, since it's conditionally skipped
->rule(Rule::in([])    // fails
```

### conditionally skipping

laravel's Rule::in() also fails if I allow `null` while the input is anything other than `null` or `''`. to gain the same behavior in filament's fluent api but at the same time add the ability to conditionally skip the validation, I thought the easiest way is to add a second parameter inspired by most of the other rules. 

i've specified my idea of the whole behaviour in `tests/src/Forms/ValidationTest.php`

### DIFF
--- a/packages/forms/src/Components/Concerns/CanBeValidated.php
+++ b/packages/forms/src/Components/Concerns/CanBeValidated.php
@@ -75,7 +75,7 @@ trait CanBeValidated
     /**
      * @param  array<scalar> | Arrayable | string | Closure  $values
      */
-    public function doesntStartWith(array | Arrayable | string | Closure $values): static
+    public function doesntStartWith(array | Arrayable | string | Closure $values, bool | Closure $condition = true): static
     {
         $this->rule(static function (Field $component) use ($values) {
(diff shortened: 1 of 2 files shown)


danharrin (approved):

danharrin:
Thanks

PR #9792, opened 2023-11-21
Update CreateRecord/.../Translatable.php

almooradi-dev:
# Error Link
https://flareapp.io/share/95JwJerm

# Description of the problem (solved)
When using `mutateFormDataBeforeCreate` and we have values for different languages, the error appears.

# Steps to reproduce the solved problem
1. Create a resource with translation
2. Add translatable columns to the model
3. Add some data using `mutateFormDataBeforeCreate`
4. Create new record (must fill more than one language data, so the error appears)
=> The error will appears

### DIFF
--- a/packages/spatie-laravel-translatable-plugin/src/Resources/Pages/CreateRecord/Concerns/Translatable.php
+++ b/packages/spatie-laravel-translatable-plugin/src/Resources/Pages/CreateRecord/Concerns/Translatable.php
@@ -114,11 +114,11 @@ trait Translatable
 
         $translatableAttributes = $record->getTranslatableAttributes();
 
-        // Fill non-translatable data
-        $record->fill(Arr::except(Arr::first($data), $translatableAttributes));
-
-        // Fill translatable data
         foreach ($data as $locale => $localeData) {
+            if ($locale === $this->activeLocale) {
+                $record->fill(Arr::except($localeData, $translatableAttributes));
+            }
+            
             $localeData = Arr::only(
                 $localeData,
                 app(static::getModel())->getTranslatableAttributes(),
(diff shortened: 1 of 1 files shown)


danharrin (changes requested):
I don't think this makes much sense, it's just doing the same as line 123-126?

AmooAti:
I don't get the point of checking active locale because always `$localeData` must contain translatable attributes.
I think we can remove the if statement and everything will be fine.

almooradi-dev:
> I don't think this makes much sense, it's just doing the same as line 123-126?

When we use `mutateFormDataBeforeCreate`, the edited/added data are not added to a specific locale, it's added to all locales, as we can see on line 98:
```
$data[$locale] = $this->mutateFormDataBeforeCreate($localeData);
```

So after `mutateFormDataBeforeCreate` the `$data` will be:
```
$data = [
    'en' => [
        'name' => 'value',
        'age' => 'value',
        'some_translatable_field' => 'value',

        // 👇 Added/Edited field (not transalatable field) from "mutateFormDataBeforeCreate"
        // ❗ This field must be in one locale, because it's not translatable
        'added' => 'value'
    ],
    'ar' => [
        'name' => 'value',
        'age' => 'value',

        // 👇 Added/Edited field (not transalatable field) from "mutateFormDataBeforeCreate"
        // ❗ This field must be in one locale, because it's not translatable
        'added' => 'value'
    ],
]
```


So, on line 127-129, we are passing `added` field to `$record->setTranslation($key, $locale, $value);`
```
// STEPS:
1️⃣ $record->setTranslation($key, $locale, $value);
2️⃣ $record->setTranslation('added', 'en', 'value');
3️⃣ ⚠️ ERROR
```

So, we need to reomove the if statement to remove non-translatable fields/data.



I hope this clarify the issue.

almooradi-dev:
I checked my commit now, the if-statement must be removed, I don't know why it's not, sorry for that, my bad
I will not make sense it the if-statement is still 😅, sorry.

I made another commit to the file removing the if-statement

PR #9802, opened 2023-11-22
Fix nested Repeater Eager Loading

### LINKED ISSUE #9776: Eager Loading Bug - 2...n nested Repeater
### Package

filament/filament

### Package Version

v3.0.99

### Laravel Version

v10.32.1

### Livewire Version

_No response_

### PHP Version

PHP 8.*

### Problem description

Eager loading for **2...n nested repeaters** does not work or only works in the first level:
If you have several repeaters among each other, eager loading only works for the first repeater.

This code:
```php
return $form->schema([
            Forms\Components\TextInput::make('name')->required(),
            Forms\Components\Repeater::make('states')
                ->itemLabel(fn($state) => $state['name'])
                ->relationship('states')->schema([
                    Forms\Components\TextInput::make('name')->required(),
                    Forms\Components\Repeater::make('cities')
                        ->itemLabel(fn($state) => $state['name'])
                        ->relationship('cities')->schema([
                            Forms\Components\TextInput::make('name')->required(),
                            Forms\Components\Repeater::make('streets')
                                ->itemLabel(fn($state) => $state['name'])
                                ->relationship('streets')->schema([
                                    Forms\Components\TextInput::make('name')->required(),
                                    Forms\Components\Repeater::make('houses')
                                        ->itemLabel(fn($state) => $state['name'])
                                        ->relationship('houses')->schema([
                                            Forms\Components\TextInput::make('name')->required(),
                                        ]),
                                ]),
                        ]),
                ]),
        ]);
```
with 3 subordinate elements each requires 123 queries although
```php
public static function getEloquentQuery():Builder{
    return parent::getEloquentQuery()->with('states.cities.streets.houses');
}
```
is used. The repeaters ignore the eager loading although it was done in advance.
It doesn't matter whether I do it within the resource or directly in the model:
```php
$with = ['relationship'];
```
image

### Expected behavior

When using
```php
public static function getEloquentQuery():Builder
{
    return parent::getEloquentQuery()->with('states.cities.streets.houses');
}
```
the relations should not be loaded individually but together in a 5 queries but l got 123 for this Resource.

If I enter the following in repeater line 1040, it works with 5 queries:
```php
<?php

namespace Filament\Forms\Components;
class Repeater extends Field implements Contracts\CanConcealComponents
{
    $relatedKeyName = $relationship->getRelated()->getKeyName();
    // Line 1039

    if(!is_null($relationship->getParent()->{$this->statePath})){
        return $this->cachedExistingRecords = $relationship->getParent()->{$this->statePath}->mapWithKeys(
            fn (Model $item): array => ["record-{$item[$relatedKeyName]}" => $item],
        );
    }
    
    //Line 1041
    return $this->cachedExistingRecords = $relationshipQuery->get()->mapWithKeys(
        fn (Model $item): array => ["record-{$item[$relatedKeyName]}" => $item],
    );
}
```
image

### Steps to reproduce

Set repeater 2...n nested inside each other eager define 
```php
$with = ['relation.subrelation'];
```
or 
```php
->with('relation')
```
or download the example repo migrate & seed.

_I was able to reproduce the same bug with V2. Exactly the same procedure._

### Reproduction repository

https://github.com/cyquer/filament-repeater-bug

### Relevant log output

_No response_

john-wink:
Checks if the relationship is already eagerly loaded, returning itself to avoid querying the database again and causing an N+1 query problem.

Issue/Bug Description: https://github.com/filamentphp/filament/issues/9776

### DIFF
--- a/packages/forms/src/Components/Repeater.php
+++ b/packages/forms/src/Components/Repeater.php
@@ -1031,6 +1031,14 @@ class Repeater extends Field implements Contracts\CanConcealComponents
             return $this->cachedExistingRecords = $this->getRecord()->{$relationshipName}->mapWithKeys(
                 fn (Model $item): array => ["record-{$item[$relatedKeyName]}" => $item],
             );
+        } elseif (
+            $this->getModelInstance()->relationLoaded($relationshipName) &&
+            (! $this->modifyRelationshipQueryUsing) &&
+            filled($orderColumn)
+        ){
+            return $this->cachedExistingRecords = $this->getRecord()->{$relationshipName}->sortBy($orderColumn)->mapWithKeys(
+                fn (Model $item): array => ["record-{$item[$relatedKeyName]}" => $item],
+            );
         }
         
         $relationshipQuery = $relationship->getQuery();
--- a/packages/forms/src/Components/Repeater.php
+++ b/packages/forms/src/Components/Repeater.php
@@ -1024,21 +1024,14 @@ class Repeater extends Field implements Contracts\CanConcealComponents
 
         if (
             $this->getModelInstance()->relationLoaded($relationshipName) &&
-            (! $this->modifyRelationshipQueryUsing) &&
-            blank($orderColumn)
+            (! $this->modifyRelationshipQueryUsing)
             
         ) {
-            return $this->cachedExistingRecords = $this->getRecord()->{$relationshipName}->mapWithKeys(
-                fn (Model $item): array => ["record-{$item[$relatedKeyName]}" => $item],
-            );
-        } elseif (
-            $this->getModelInstance()->relationLoaded($relationshipName) &&
-            (! $this->modifyRelationshipQueryUsing) &&
-            filled($orderColumn)
-        ){
-            return $this->cachedExistingRecords = $this->getRecord()->{$relationshipName}->sortBy($orderColumn)->mapWithKeys(
-                fn (Model $item): array => ["record-{$item[$relatedKeyName]}" => $item],
-            );
+            return $this->cachedExistingRecords = $this->getRecord()->getRelationValue($relationshipName)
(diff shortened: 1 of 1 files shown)


john-wink:
Local `composer test`:
image

danharrin:
Your previous implementation would have broken `modifyRelationshipQueryUsing()` and `orderColumn()`. I've added in checks to ensure that this is not used when either of those is set. If you want to use them with eager loading, you must add these clauses to the eager loading itself. Does it still work ok for your use case?

john-wink:
> Your previous implementation would have broken `modifyRelationshipQueryUsing()` and `orderColumn()`. I've added in checks to ensure that this is not used when either of those is set. If you want to use them with eager loading, you must add these clauses to the eager loading itself. Does it still work ok for your use case?

would sortBy fix the orderColumn Problem?
```php
elseif (
    $this->getModelInstance()->relationLoaded($relationshipName) &&
    (! $this->modifyRelationshipQueryUsing) &&
    filled($orderColumn)
){
    return $this->cachedExistingRecords = $this->getRecord()->{$relationshipName}->sortBy($orderColumn)->mapWithKeys(
        fn (Model $item): array => ["record-{$item[$relatedKeyName]}" => $item],
    );
}
```

That would take the eager loading into account & still work, wouldn't it?

danharrin:
It would fix it, but I don't know why you would want to do that, when you can chain it onto the `with()` anyway? It's more performant if done on the query instead of the collection

john-wink:
In my case it does not work. 
If I use ->orderColumn() I can sort & save it.
If I use ->reorderable() I can sort it but when I save it, it resets itself. and also a reload of the page shows the status before sorting.

However, I have applied orderBy('sort') directly to the relationship

PR #9813, opened 2023-11-22
Add Support for Cursor Pagination

stefanzweifel:

This pull request introduces support for [Cursor Pagination](https://laravel.com/docs/10.x/pagination#cursor-pagination) in the Table component and paginator.

Since `cursorPaginate()` generates a `CursorPaginate` object, I've updated type hints throughout the codebase.(I hope this doesn't count as a breaking change)

The `<x-filament::pagination />` Blade component has been updated as well to check if the given Paginator has a `getCursorName`-method. (Similar to how Livewire itself [handles this](https://github.com/livewire/livewire/blob/95083c0b61c1910ad28c4628e0dfa9c227eb1c41/src/Features/SupportPagination/views/simple-tailwind.blade.php#L39-L47))

The "wire-click-action" now adapts based on the method's availability. Previously, I duplicated HTML markup, but I simplified this in f9433b2d679613fe04d30bb20acfd15b34881d69. With this change, only one set of "Next"/"Previous" buttons needs to be maintained.

A user can now use `$query->cursorPaginate()` in a `ListRecords` component like this:

```php
protected function paginateTableQuery(Builder $query): Paginator | CursorPaginator
{
    return $query->cursorPaginate($this->getTableRecordsPerPage() == 'all' ? $query->count() : $this->getTableRecordsPerPage());
}
```

Screenshot 2023-11-22 at 20 22 57

---

I've also updated the documentation to include `cursorPaginate()` alongside the `simplePaginate()` examples.

Unfortunately, I couldn't find any tests covering pagination functionality, so I've omitted this step. I'm more than willing to contribute tests if someone can direct me to the appropriate location in the test suite.

### DIFF
--- a/packages/support/docs/09-blade-components/02-pagination.md
+++ b/packages/support/docs/09-blade-components/02-pagination.md
@@ -28,12 +28,13 @@ class ListUsers extends Component
 <x-filament::pagination :paginator="$users" />
 ```
 
-Alternatively, you can use simple pagination, which will just render a "previous" and "next" button:
+Alternatively, you can use simple pagination or cursor pagination, which will just render a "previous" and "next" button:
 
 ```php
 use App\Models\User;
 
 User::query()->simplePaginate(10)
+User::query()->cursorPaginate(10)
 ```
 
 ## Allowing the user to customize the number of items per page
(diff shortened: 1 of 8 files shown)


danharrin:
Thanks!

PR #9825, opened 2023-11-23
Added red color to notifications number badge

jvitasek:
Hey, I'm thinking about something like this.

If we had some type of configuration so that users could customize the color of their badge when no notifications are present, and also when there are notifications.

### DIFF
--- a/packages/panels/resources/views/components/topbar/database-notifications-trigger.blade.php
+++ b/packages/panels/resources/views/components/topbar/database-notifications-trigger.blade.php
@@ -1,6 +1,7 @@
 <x-filament::icon-button
     :badge="$unreadNotificationsCount"
     color="gray"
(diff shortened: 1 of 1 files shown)


danharrin:
I think we would only accept it if it is customisable, but I'm really not sure if it's useful enough?

jvitasek:
> I think we would only accept it if it is customisable, but I'm really not sure if it's useful enough?

So the reason I'm thinking about this is that multiple clients reported they "miss" new notifications. I thought about doing this over JS but maybe someone would find the extra customization useful?

AFAIK right now you can only override the actual notification list, not the default trigger.

danharrin:
How about publishing the `database-notifications-trigger.blade.php` view into your app, and then you can modify it? I think this is probably too far to introduce config values for since we haven't had this request before, but if I do get this request again then I will remember this PR and see if we should merge it. Thank you anyway

danharrin:
Let me know if you have trouble publishing this view

PR #9836, opened 2023-11-24
fix #8788 - reset activerelationmanager when selected is hidden

### LINKED ISSUE #8788: Issue when toggling a resources relationmanager

tvdr:
This PR fixes the bug described in https://github.com/filamentphp/filament/issues/8788 considering @danharrin 's [comment](https://github.com/filamentphp/filament/pull/8790#issuecomment-1763314753)  

After hiding a currently active relation manager, the activeRelationManager is set to the first item and viewdata is updated before rendering.

### DIFF
--- a/packages/panels/src/Resources/Pages/Concerns/HasRelationManagers.php
+++ b/packages/panels/src/Resources/Pages/Concerns/HasRelationManagers.php
@@ -45,25 +45,19 @@ trait HasRelationManagers
         return $manager;
     }
 
-    public function mountHasRelationManagers(): void
+    public function renderingHasRelationManagers(): void
     {
         $managers = $this->getRelationManagers();
 
-        if (array_key_exists($this->activeRelationManager, $managers) || $this->hasCombinedRelationManagerTabsWithContent()) {
+        if (array_key_exists($this->activeRelationManager, $managers)) {
             return;
         }
 
-        $this->activeRelationManager = array_key_first($this->getRelationManagers()) ?? null;
-    }
-
-    public function renderingHasRelationManagers(View $view): void
-    {
-        $managers = $this->getRelationManagers();
-
-        if (!array_key_exists($this->activeRelationManager, $managers) && !$this->hasCombinedRelationManagerTabsWithContent()) {
-            $this->activeRelationManager = array_key_first($this->getRelationManagers()) ?? null;
-            $view->with('activeRelationManager', $this->activeRelationManager);
+        if ($this->hasCombinedRelationManagerTabsWithContent()) {
+            return;
         }
+
+        $this->activeRelationManager = array_key_first($managers);
     }
 
     public function hasCombinedRelationManagerTabsWithContent(): bool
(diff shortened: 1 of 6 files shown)


packages/panels/src/Resources/Pages/Concerns/HasRelationManagers.php
```diff
@@ -55,6 +56,16 @@ public function mountHasRelationManagers(): void
         $this->activeRelationManager = array_key_first($this->getRelationManagers()) ?? null;
     }
 
+    public function renderingHasRelationManagers(View $view): void
+    {
+        $managers = $this->getRelationManagers();
+
+        if (!array_key_exists($this->activeRelationManager, $managers) && !$this->hasCombinedRelationManagerTabsWithContent()) {
+            $this->activeRelationManager = array_key_first($this->getRelationManagers()) ?? null;
+            $view->with('activeRelationManager', $this->activeRelationManager);
```
danharrin: Is this line needed? I think that all public properties are exposed to the view by default

tvdr: i'm not sure if i fully understand how the full lifecycle here works, but without setting the View-s data here, only after a following update request will the active tab set to the first item.

while trying to figure this out, i found that in the edit-record.blade.php there is a difference between $this->activeRelationManager and $activeRelationManager, (accessing the property via $this return the desired value without setting it on the viewdata) 
but can't find anything about this in the docs. i'm not sure if this is a bad pattern to rely on.

so passing :active-manager="$this->activeRelationManager to  <x-filament-panels::resources.relation-managers works, but i don't want to submit something that is not acceptable.

danharrin: I've switched to `$this->activeRelationManager`, that should be fine

PR #9852, opened 2023-11-25
Fix sidebar header ring overflow-x

4m1n0s:

Before:
image

After:

image


zepfietje (changes requested):
Thanks for reporting this, @4m1n0s.
The use of box-shadow/ring is by design though, to match the positioning of content under the topbar, as the ring doesn't take up any space while the border does.

Could you rework this PR to just prevent overflow-x on the sidebar header? Similar approach is used on the topbar, though it may require a wrapping element.

4m1n0s:
@zepfietje in this case i think only a wrapping div like that is enough?

```html
<div class="overflow-x-clip">
    <header>.....</header>
</div>
```

zepfietje:
I don't know, @4m1n0s.
Please test it and commit if it works. :)

zepfietje (approved):

zepfietje:
Thanks!

PR #9866, opened 2023-11-25
fix: registration evaluation of HasAuth trait

marcotas:
The getRegistrationRouteAction method is now able to evaluate the closure or value given to it.


danharrin:
The closure is meant to be a closure that gets executed when the route is hit, like `Route::get('/', function () {})`

Evaluating like this is too early

PR #9868, opened 2023-11-25
[Docs - Select] Order of "multiple()" & "relationship()"

Ahmant:
Mention that it's necessary to define `multiple()` before `relationship()` in the `Select` field
``` 
// ❌
Select::make('users')
    ->multiple()

// ✅
Select::make('users')
    ->multiple()
```

Because `dehydrated()` is related to `multiple()`
```
// vendor\filament\forms\src\Components\Select.php - Line 984
$this->dehydrated(fn (Select $component): bool => ! $component->isMultiple());
```

### DIFF
--- a/packages/forms/docs/03-fields/03-select.md
+++ b/packages/forms/docs/03-fields/03-select.md
@@ -159,6 +159,7 @@ Select::make('technologies')
     ->multiple()
     ->relationship(titleAttribute: 'name')
 ```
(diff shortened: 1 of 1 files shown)


danharrin:
I don't think this is right. The `isMultiple()` check is part of a closure function that is not run at that point, it is run later.

Ahmant:
Yeah, you are right.
The issue was from the `disabled()`, mentioned in this PR [#9869](https://github.com/filamentphp/filament/pull/9869), not the `multiple()`.

Sorry for that, my bad.

PR #9869, opened 2023-11-25
[Docs - Field] Order of "disabled()" & "relationship()"

Ahmant:
Mention that it's necessary to define `disabled()` before `relationship()`.
```
// ❌
SomeField::make('users')
    ->disabled()

// ✅
SomeField::make('users')
    ->disabled()
```

Because `dehydrated()` in `relationship()` will be overrided by the one in `disabled()`
```
// vendor\filament\forms\src\Components\Concerns\CanBeDisabled.php - Line 17
$this->dehydrated(fn (Component $component): bool => ! $component->evaluate($condition));
```

### DIFF
--- a/packages/forms/docs/03-fields/01-getting-started.md
+++ b/packages/forms/docs/03-fields/01-getting-started.md
@@ -238,7 +238,7 @@ Toggle::make('is_admin')
     ->dehydrated()
 ```
 
-> When using `disabled()` with `relationship()`, ensure that `disabled()` is defined before `relationship()`. This ensures that the `dehydrated()` function, called within `relationship()`, is not overridden by the one called in `disabled(...)`.
+> When using `disabled()` with `relationship()`, ensure that `disabled()` is called before `relationship()`. This ensures that the `dehydrated()` call from within `relationship()` is not overridden by call from `disabled()`.
 
 > If you choose to dehydrate the field, a skilled user could still edit the field's value by manipulating Livewire's JavaScript.
 
--- a/packages/forms/docs/03-fields/01-getting-started.md
+++ b/packages/forms/docs/03-fields/01-getting-started.md
@@ -237,7 +237,8 @@ Toggle::make('is_admin')
     ->disabled()
     ->dehydrated()
 ```
-> When using `disabled($someCondition)` with `relationship()`, ensure that `disabled(...)` is defined before `relationship()`. This ensures that the `dehydrated()` function, called within `relationship()`, is not overridden by the one called in `disabled(...)`.
+
+> When using `disabled()` with `relationship()`, ensure that `disabled()` is defined before `relationship()`. This ensures that the `dehydrated()` function, called within `relationship()`, is not overridden by the one called in `disabled(...)`.
 
 > If you choose to dehydrate the field, a skilled user could still edit the field's value by manipulating Livewire's JavaScript.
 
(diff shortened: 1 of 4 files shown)


packages/forms/docs/03-fields/01-getting-started.md
```diff
@@ -237,6 +237,7 @@ Toggle::make('is_admin')
     ->disabled()
     ->dehydrated()
 ```
+> When using `disabled($someCondition)` with `relationship()`, ensure that `disabled(...)` is defined before `relationship()`. This ensures that the `dehydrated()` function, called within `relationship()`, is not overridden by the one called in `disabled(...)`.
```
danharrin:
```suggestion

> When using `disabled()` with `relationship()`, ensure that `disabled()` is defined before `relationship()`. This ensures that the `dehydrated()` function, called within `relationship()`, is not overridden by the one called in `disabled(...)`.
```

danharrin:
Not all fields have `relationship()`, and not all `relationship()` call `dehydrated()`, so I'm not sure if this belongs on the Getting Started page. Should we add it multiple times on more specific pages?

Ahmant:
I see your point.

I this situation we only have 3 components with `relationship()`, and in all these 3 components the `dehydrated()` function is called, so adding that note 3 times will not be bad.
- CheckboxList
    - Path: forms\src\Components\CheckboxList.php
    - Line: 194
- Repeater
    - Path: forms\src\Components\Repeater.php
    - Line: 914
- Select
    - Path: forms\src\Components\Select.php
    - Line: 984


And note that the order will make an issue only when `$someCondition == true` in `disabled($someCondition)`, because in this case `dehydrated` will become `false`:
```
// filament\forms\src\Components\Concerns\CanBeDisabled.php
public function disabled(bool | Closure $condition = true): static
{
    $this->isDisabled = $condition;
 👉 $this->dehydrated(fn (Component $component): bool => ! $component->evaluate($condition));

    return $this;
}
```


Is this right or I'm missing something?

Ahmant:
I'm unsure if it's worth adding this to the documentation, but it took me around 4 hours maybe to figure out. After upgrading from `v2` to `v3`, I encountered an error when saving my record, which was caused by this incorrect order.


Strangely, it worked in `v2` but not in `v3` 🤷‍♂️.

danharrin:
Let's add the note to those 3 separate pages then, instead of Getting Started

In v2, we didn't disable dehydration for disabled fields by default, thats why the issue wasn't there

packages/forms/docs/03-fields/01-getting-started.md
```diff
@@ -238,6 +238,8 @@ Toggle::make('is_admin')
     ->dehydrated()
 ```
 
+> When using `disabled()` with `relationship()`, ensure that `disabled()` is defined before `relationship()`. This ensures that the `dehydrated()` function, called within `relationship()`, is not overridden by the one called in `disabled(...)`.
```
danharrin:
```suggestion
> When using `disabled()` with `relationship()`, ensure that `disabled()` is called before `relationship()`. This ensures that the `dehydrated()` call from within `relationship()` is not overridden by call from `disabled()`.
```

Ahmant:
Ok 👌, I will add them, and push them for review

Ahmant:
I pushed the changes.

But I made a mistake in my previous statement.
> And note that the order will make an issue only when `$someCondition == true` in `disabled($someCondition)`, because in this case `dehydrated` will become `false`:


The `dehydrated` in the `relationship()` must be `false` (it is by default), so the order will make an issue when `$someCondition == false` in `disabled($someCondition)`, because in this case `dehydrated` will become `true`:

PR #9873, opened 2023-11-25
Support sub-navigation position end

glennjacobs:
Allows sub-navigation to align-right.

LEFT
image

RIGHT
image


packages/panels/resources/views/components/sidebar/subnavigation.blade.php
danharrin, line 1: Please call this `sub-navigation`

packages/panels/src/Pages/Concerns/HasSubNavigation.php
danharrin, line 1: We always try to avoid using avoid using "right" in our APIs, as it's not RTL friendly. Let's call this "end" instead.

`$isSubNavigationAlignedEnd`, `isSubNavigationAlignedEnd()`

danharrin (changes requested):

packages/panels/src/Pages/Concerns/HasSubNavigation.php
glennjacobs: Good idea. I was struggling with the naming on this one.

glennjacobs:
@danharrin do you think we need the option to set this at the resource level also?

danharrin:
Yes, I think so

glennjacobs:
> Yes, I think so

Done and added docs.

danharrin:
Please use `$isSubNavigationAlignedEnd`, `isSubNavigationAlignedEnd()` in methods and properties

Including `$isSubNavigationAlignedEnd` in the view `$isSubNavigationAlignedEnd = $this->isSubNavigationAlignedEnd();`

zepfietje:
@danharrin @glennjacobs wouldn't it be nicer to have an alignment property instead (start/end)?
Feels a little nicer to me than a boolean.

danharrin:
Yeah could be a good idea, we can use the alignment enum

zepfietje:
> Yeah could be a good idea, we can use the alignment enum

Yup, could you handle this, @glennjacobs?

glennjacobs:
But then it would be left and right, which didn't feel right to me considering RTL. I did consider it initially.

glennjacobs:
Oh, add start&end to the enum. Ok.

glennjacobs:
@zepfietje @danharrin all done.

packages/panels/resources/views/components/page/index.blade.php
```diff
@@ -132,6 +84,12 @@ class="fi-page-footer-widgets"
 
                 {{ \Filament\Support\Facades\FilamentView::renderHook('panels::page.footer-widgets.after', scopes: $this->getRenderHookScopes()) }}
             </div>
+
+            @if ($subNavigation && $subNavigationAlignment === Alignment::End)
+                <x-filament-panels::sidebar.sub-navigation
+                    :subNavigation="$subNavigation"
```
zepfietje: Could you make sure to use kebab case for HTML attributes, @glennjacobs?

```suggestion
                    :sub-navigation="$subNavigation"
```

zepfietje (changes requested):

packages/panels/resources/views/components/page/index.blade.php
glennjacobs: Done

zepfietje:
I'll take over this PR to merge it with the changes I've just made.

glennjacobs:
> I'll take over this PR to merge it with the changes I've just made.

No worries. I'll keep an eye out so I understand.

glennjacobs:
@zepfietje I have noticed one issue which is when the navigation is aligned right, the mobile menu is then presented at the bottom of the screen. Yikes.

zepfietje:
Will take care of that, @glennjacobs.

zepfietje:
Alright, all done.
Thanks, @glennjacobs!

PR #7392, opened 2023-08-02
Support persisting collapsible section state

DarpanKulkarni:
Added an option that allow to persist the sections collapse state to the browser's LocalStorage. This enhancement leverages the Alpine.js Persist plugin, which is now included with Livewire 3.

The sections state can be persisted using `->persistCollapseState()`.

The localStorage keys use the component's id, which is generated using the section heading. If, for some reason, two or more sections have the same headings, the `->id()` can be used to avoid conflicts.

The changes have been throughly tested. However, I have identified one minor issue related to flash of content during the page reload when the section is persisted and expanded. This is because the collapse toggle behaviour is managed by Alpine.js on the client side, and it's not feasible to conditionally add x-cloak.  I think this is minor issue an can be ignored. I will continue investigating potential solutions to address this issue, any suggestions or feedback from the team are welcomed and highly appreciated.

Thank you for your time and consideration.

PR #9874, opened 2023-11-25
Saving the Active Tab in the LocalStorage

AlexisSerneels:
It allows to persist the Active Tab in the session rather than the Query String.
By simply adding ```->persistTabInSession()``` on the ```Tabs```field configuration.

**Why doing so?**
There is a current bug where the active tab resets to the first tab when switching locales using the `LocaleSwitcher` action
It happens because the query string is not sent  along with the Livewire`update` request, causing the active tab to always revert back to the first tab. As we do not have access to `update` request from the Tabs Component to include the Query Strings, we rely on the session instead.

The session name is dynamically generated based on the Filament Page Type, the associated Model's database table, and the Model's Primary Key. This ensures uniqueness and prevents conflicts between different sessions.

There is no visual change.


danharrin (changes requested):
Instead of using the backend session store, let's use the local storage and Alpine's `$persist()`, identically to #7392

Should be much simpler

AlexisSerneels:
I agree. Thank you for the tip.
Changes have been pushed.

danharrin (changes requested):
Please also apply this for the Infolists Tabs component

danharrin:
Also worth copying the wording in the docs for the Section persistCollapsed()

AlexisSerneels:
I've:
- Added support for Infolist.
- Enabled Closure resolution.
- Allowed customization of the local storage name.
- Updated the documentation to ensure consistent wording.

AlexisSerneels:
I've:

Added support for Infolist.
Enabled Closure resolution.
Allowed customization of the local storage name.
Updated the documentation to ensure consistent wording.

AlexisSerneels (commented):
As requested, I've:

- Added support for Infolist.
- Enabled Closure resolution.
- Allowed customization of the local storage name.
- Updated the documentation to ensure consistent wording.

AlexisSerneels:
I've fixed the imports, thereby resolving the failing tests.

PR #9882, opened 2023-11-26
Add hex color validation rule

binaryfire:
Add the new hex color validation rule

### DIFF
--- a/packages/forms/docs/05-validation.md
+++ b/packages/forms/docs/05-validation.md
@@ -220,6 +220,14 @@ The field value must be greater than or equal to another. [See the Laravel docum
 Field::make('newNumber')->gte('oldNumber')
 ```
 
+### Hex color
+
+The field value must be a valid color in hexadecimal format. [See the Laravel documentation.](https://laravel.com/docs/validation#rule-hex-color)
+
+```php
+Field::make('color')->hexColor()
+```
+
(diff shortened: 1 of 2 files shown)


danharrin (approved):

danharrin:
Is it worth adding validation by default to ColorPicker?

zepfietje:
Default validation for color picker sounds good to me.

danharrin:
Might mean we have to add custom rules for RGB etc

binaryfire:
@danharrin The native color picker can handle hex, RGB and HSL: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color

Is it possible we might expand the field at some point to allow showing this color mode selector? Might be better to avoid making it check for hex by default.

binaryfire:
@danharrin 
image

danharrin:
You can already do it with TextInput::make()->type('color')

binaryfire:
Oh right, I didn't realise the ColorPicker already [accepts other formats](https://filamentphp.com/docs/3.x/forms/fields/color-picker#setting-the-color-format).

Only `hex_color` was added to Laravel core so yeah, it would require custom rules. Wouldn't that be better in a separate PR? This one is just to keep the Filament rules up to date with core.

PR #9892, opened 2023-11-26
Fix inline fields

zepfietje:

This PR started with a fix for just the radio component (description below), but I've found more related things to fix regarding form field inline label alignment.

The `inline()` API of the radio button form component currently also sets the label to be inline. This is unexpected, as the dedicated `inlineLabel()` API should control this. This PR updates the radio component to support non-inline label with inline radios. Visually this may be regarded as a breaking change, but to me this is a bug since the current result is unexpected.

### Before

Screenshot 2023-11-26 at 19 51 56

### After

Screenshot 2023-11-26 at 19 52 28

### DIFF
--- a/packages/forms/resources/views/components/builder.blade.php
+++ b/packages/forms/resources/views/components/builder.blade.php
@@ -1,30 +1,30 @@
+@php
+    $containers = $getChildComponentContainers();
+    $blockPickerColumns = $getBlockPickerColumns();
+    $blockPickerWidth = $getBlockPickerWidth();
+    $blocks = $getBlocks();
+
+    $addAction = $getAction($getAddActionName());
+    $addBetweenAction = $getAction($getAddBetweenActionName());
+    $cloneAction = $getAction($getCloneActionName());
+    $collapseAllAction = $getAction($getCollapseAllActionName());
+    $expandAllAction = $getAction($getExpandAllActionName());
+    $deleteAction = $getAction($getDeleteActionName());
+    $moveDownAction = $getAction($getMoveDownActionName());
+    $moveUpAction = $getAction($getMoveUpActionName());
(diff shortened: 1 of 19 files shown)


danharrin:
I'm concerned as the `inline()` method for checkboxes does add them inline with the label, hence why it was like this before. I think this is too much of a breaking change tbh

zepfietje:
> I'm concerned as the `inline()` method for checkboxes does add them inline with the label, hence why it was like this before. I think this is too much of a breaking change tbh

For single checkboxes it makes sense as is, but checkbox lists (which are the checkbox counterpart of radios) don't have the `inline()` API.

I understand your concern regarding the change being breaking. Though we should then add a temporary API to allow for the right behavior introduced by this PR, because the new behavior is in fact what it should have been all along (also check Tailwind UI for example: https://tailwindui.com/components/application-ui/forms/radio-groups#component-f77fa2476964716cd375c934954229ae). By the way, the radio `inline()` API was inconsistent already, as for toggles and checkboxes it displayed the inputs in front of the label, while radios displayed after the label (like with `inlineLabel()`).

danharrin:
Removed the BC by calling `inlineLabel()` from `inline()` conditionally. To revert that, you just to `inline()->inlineLabel(false)` which I think makes sense.

zepfietje:
Yup, good workaround. Where do we document this change so we can remove it in v4, @danharrin?

danharrin:
I don't think we should remove it in v4 IMO, this behaviour is useful as it allows you to put a radio next to a checkbox/toggle without consuming extra height

zepfietje:
Well, it should be enabled explicitly. Just making the radios inline shouldn't make the label inline. Using `->inline()->inlineLabel()` should be used explicitly.

PR #9895, opened 2023-11-26
Validate field comparison rules conditionally

mariusticha:

---

we want to add the feature to conditionally validate the `same()` rule, like a lot of other rules using the extra parameter `bool | Closure $condition`. 

noticing that this rule gets passed into the `fieldComparisonRule()` we've updated all rules using this method too: 

```php
same()
different()
gt()
gte()
lt()
lte()
```


danharrin:
`different(fn () => $condition ? 'field name' : null)`, returning null disables the rule

PR #9896, opened 2023-11-26
Add check if field must be accepted

mariusticha:

--- 

while testing a form's field configuration we've noticed that a method is missing to check if a field must be accepted (analog to e.g. `isRequired()`, `isHidden()` or `isDisabled()`:

```php
// form
Toggle::make('foo')
    ->required(),
Toggle::make('bar')
    ->disabled(),
Toggle::make('baz')
    ->accepted(),
                    
// tests
livewire(Test::class)
    ->assertFormFieldExists('foo', fn (Toggle $field) => $field->isRequired())
    ->assertFormFieldExists('bar', fn (Toggle $field) => $field->isDisabled())
    ->assertFormFieldExists('baz', fn (Toggle $field) => $field->mustBeAccepted()) // <-- missing


danharrin:
There are too many rules to do this for every one of them, I suggest calling `getValidationRules()` and then using `in_array()` to check if the rule is present

Issue #9838, opened 2023-11-24
Modal button order (primary action vs cancel action) is not consistent across action types


ryanmortier:
### Package

filament/filament

### Package Version

v3

### Laravel Version

v10

### Livewire Version

v3

### PHP Version

v8.2

### Problem description
edit: I've managed to work-around the issue with the following:

```php
Tables\Actions\DeleteAction::make()
    ->modalSubmitAction(fn (StaticAction $action) =>
        $action->extraAttributes(['class' => 'order-first'])),
```

In a create modal, the button order is primary -> secondary
In an edit modal, the button order is primary -> secondary
_**In a delete confirmation modal, the button order is secondary -> primary**_

I believe this to be a UX issue and could cause a user to misclick an action if not paying attention.

Also, I don't think I can fix this in CSS because fixing it for one modal, reverses it on the other.

Best described in screenshots:

Create:
![CleanShot 2023-11-23 at 21 45 14](https://github.com/filamentphp/filament/assets/2053960/05e5303e-5860-4dbd-b1e5-fcfecfe1ccfa)

Edit:
![CleanShot 2023-11-23 at 21 45 26](https://github.com/filamentphp/filament/assets/2053960/cce0ad43-f24b-4dba-909f-52cd5d7b7f45)

Delete:
![CleanShot 2023-11-23 at 21 45 37](https://github.com/filamentphp/filament/assets/2053960/f89af1b0-b5d3-43c3-af8d-5aa921635e7a)

### Expected behavior

Primary button orders should be consistent across all modals for UX.

### Steps to reproduce

Open create/edit/delete modals in the Filament demo app.

### Reproduction repository

https://github.com/filamentphp/demo

danharrin:
This is to do with if the actions are full width or not. You can observe the same in Tailwind UI:

<img width="910" alt="Screenshot 2023-11-27 at 23 13 49" src="https://github.com/filamentphp/filament/assets/41773797/f32e595e-1911-4eb5-827e-28eb8f0e00a5">

<img width="910" alt="Screenshot 2023-11-27 at 23 13 43" src="https://github.com/filamentphp/filament/assets/41773797/d6ba0e68-ed94-465d-b54a-58a3a278475b">

If you align the actions to the right, they swap position, like Tailwind UI:

<img width="910" alt="Screenshot 2023-11-27 at 23 14 27" src="https://github.com/filamentphp/filament/assets/41773797/daa399a9-86d8-4a17-8e2a-898dc93cc2ee">

PR #9905, opened 2023-11-27
Update manage-issue.yml

saade:
Url can be null as defined in https://github.com/filamentphp/filament/blob/9f2f63a611594dedc97130cf5a08eebb38ba7e07/.github/workflows/manage-issue.yml#L34

Because of that, its causing errors while evaluating NO_REPRO_URL further down in the code, see action log:
https://github.com/filamentphp/filament/actions/runs/7005326321

This PR fixes the evaluation by adding a check wether `url` is not falsy.

Also fixes the checks for repro urls under `filamentphp` org to avoid urls like:
https://github.com/filamentphp/filament/issues/9838

```js
const url = 'https://github.com/filamentphp/demo'

// before
((! url.includes('https://github.com/')) || url.includes('https://github.com/filamentphp/') || url === 'https://github.com/filamentphp') && (! url.includes('https://github.com/filamentphp/demo')) // false

// after
((! url) || (! url.includes('https://github.com/')) || (url.includes('https://github.com/filamentphp'))) // true
```

### DIFF
--- a/.github/workflows/manage-issue.yml
+++ b/.github/workflows/manage-issue.yml
@@ -158,7 +158,7 @@ jobs:
                 Hey @${user}! We're sorry to hear that you've hit this issue. 💛
             `.trim()
 
-            const NO_REPRO_URL = ((! url) || (! url.includes('https://github.com/')) || (url.includes('https://github.com/filamentphp')))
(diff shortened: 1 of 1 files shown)


danharrin:
We want to allow `filamentphp/demo` but deny `filamentphp/filament`

PR #9910, opened 2023-11-27
fixing translate error in radio component

eduardo-de-carvalho-couto:
Radio translation implementation.

### DIFF
--- a/packages/forms/src/Components/Radio.php
+++ b/packages/forms/src/Components/Radio.php
@@ -31,7 +31,7 @@ class Radio extends Field
         parent::setUp();
     }
 
-    public function boolean(string $trueLabel = null, string $falseLabel = null): static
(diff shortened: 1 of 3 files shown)


danharrin (changes requested):
These are translations for the select, please copy them for the radio component too

packages/forms/src/Components/Radio.php
```diff
@@ -31,11 +31,11 @@ protected function setUp(): void
         parent::setUp();
     }
 
-    public function boolean(string $trueLabel = 'Yes', string $falseLabel = 'No'): static
+    public function boolean(string $trueLabel = null, string $falseLabel = null): static
```
danharrin:
```suggestion
    public function boolean(?string $trueLabel = null, ?string $falseLabel = null): static
```

PR #9913, opened 2023-11-27
Update max content width options

ilpav12:
Updated the maximum content width options and the corresponding Filament's documentation. Added `Zero`, `None`, `MinContent`, `MaxContent`, `FitContent` to the MaxWidth Enum options and made adjustments in the system to match these changes. This enhancement provides more width options to accommodate varying screen sizes and improve the user interface layout on larger screens.

### DIFF
--- a/packages/panels/docs/09-configuration.md
+++ b/packages/panels/docs/09-configuration.md
@@ -100,7 +100,7 @@ public function panel(Panel $panel): Panel
 
 ## Customizing the maximum content width
 
-By default, Filament will restrict the width of the content on the page, so it doesn't become too wide on large screens. To change this, you may override the `getMaxContentWidth()` method. Options correspond to [Tailwind's max-width scale](https://tailwindcss.com/docs/max-width). The options are `ExtraSmall`, `Small`, `Medium`, `Large`, `ExtraLarge`, `TwoExtraLarge`, `ThreeExtraLarge`, `FourExtraLarge`, `FiveExtraLarge`, `SixExtraLarge`, `SevenExtraLarge`, `Full`, `MinContent`, `MaxContent`, `FitContent`,  `Prose`, `ScreenSmall`, `ScreenMedium`, `ScreenLarge`, `ScreenExtraLarge` and `ScreenTwoExtraLarge`. The default is `SevenExtraLarge`:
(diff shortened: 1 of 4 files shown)


ilpav12:
I am not sure if moving the tailwind classes strings from the blade to the Enum will stop tailwind from the generation of those in the css.

danharrin:
They should not be in the enum for that reason

danharrin:
What is the use of the extra options that you cant accomplish with existing options?

ilpav12:
`max-w-fit` for example it is not available on the Filament css so I wanted to provide all the possible max width options available out of the box.

danharrin:
Right, but what is that actually useful for in this context?

ilpav12:
I have some tables that have a width between 7xl and full and with max-w-fit I can get the perfect width even if it changes slightly between pages. I don't know if this replies to your question

ilpav12:
I moved back the classes in the blade file. Now the only difference is the added options.

danharrin:
What would be the point of Zero and None?

ilpav12:
In this case I don't think there is a useful use case but I wanted to add all the options for completeness. Before in the Enum there was an additional case `Screen` that doesn't exist in the tailwind classes.

danharrin:
Can we remove the ones that have no practical use please 😁

ilpav12:
Done: [Removed Zero and None options](https://github.com/filamentphp/filament/pull/9913/commits/2f39ebda5218a76a52ea2ceb61a51f23941766d0)

packages/panels/docs/09-configuration.md
```diff
@@ -100,7 +100,7 @@ public function panel(Panel $panel): Panel
...
+By default, Filament will restrict the width of the content on the page, so it doesn't become too wide on large screens. To change this, you may override the `getMaxContentWidth()` method. Options correspond to [Tailwind's max-width scale](https://tailwindcss.com/docs/max-width). The options are `ExtraSmall`, `Small`, `Medium`, `Large`, `ExtraLarge`, `TwoExtraLarge`, `ThreeExtraLarge`, `FourExtraLarge`, `FiveExtraLarge`, `SixExtraLarge`, `SevenExtraLarge`, `Full`, `MinContent`, `MaxContent`, `FitContent`,  `Prose`, `ScreenSmall`, `ScreenMedium`, `ScreenLarge`, `ScreenExtraLarge` and `ScreenTwoExtraLarge`. The default is `SevenExtraLarge`:
```
danharrin:
```suggestion
By default, Filament will restrict the width of the content on the page, so it doesn't become too wide on large screens. To change this, you may use the `maxContentWidth()` method. Options correspond to [Tailwind's max-width scale](https://tailwindcss.com/docs/max-width). The options are `ExtraSmall`, `Small`, `Medium`, `Large`, `ExtraLarge`, `TwoExtraLarge`, `ThreeExtraLarge`, `FourExtraLarge`, `FiveExtraLarge`, `SixExtraLarge`, `SevenExtraLarge`, `Full`, `MinContent`, `MaxContent`, `FitContent`,  `Prose`, `ScreenSmall`, `ScreenMedium`, `ScreenLarge`, `ScreenExtraLarge` and `ScreenTwoExtraLarge`. The default is `SevenExtraLarge`:
```

packages/support/src/Enums/MaxWidth.php
```diff
@@ -45,4 +45,6 @@ enum MaxWidth
     case ScreenExtraLarge;
 
     case ScreenTwoExtraLarge;
+
+    case Screen;
```
ilpav12, line 49: Why is screen back? Have I missed a spot where it is used because it is not a valid tailwind class

danharrin: It's used in the modal to apply `fixed inset-0`. Please search the codebase before removing stuff

PR #9931, opened 2023-11-28
Add maxItems per block for Builder blocks

binaryfire:
This adds a `->maxItems()` method for each block in Builder fields. Blocks are removed from the block picker after the max number has been added.

Video: https://www.youtube.com/watch?v=hMTXZ_0FcMI

Where should I add the docs for this? Under "Adding items"?


packages/forms/src/Components/Builder.php
```diff
@@ -831,6 +831,23 @@ public function isBlockLabelTruncated(): bool
...
+    
+        return array_filter($this->getBlocks(), function ($block) use ($state) {
+            /** @var Block $block */
+            $count = collect($state)->filter(function ($item) use ($block) {
+                return $item['type'] === $block->getName();
+            })->count();
+
+            return is_null($block->getMaxItems()) || $count < $block->getMaxItems();
+        });
+    }
```
danharrin: If there is no maximum, please prevent calculating the number of existing blocks

packages/forms/src/Components/Builder.php
```diff
@@ -831,6 +831,23 @@ public function isBlockLabelTruncated(): bool
         return (bool) $this->evaluate($this->isBlockLabelTruncated);
     }
 
+    /**
+     * @return Component[]
```
danharrin:
```suggestion
     * @return array<Block>
```

binaryfire: @danharrin The method is returning `Component` instead of Block:

```
841    Method Filament\Forms\Components\Builder::getBlockPickerBlocks() should return array<Filament\Forms\Components\Builder\Block> but returns
         array<Filament\Forms\Components\Component>

```
So I used `@return array<Component>` instead

danharrin: It actually does return blocks though, there are ways we can convince PHPstan of that with a `@var` block

PR #9935, opened 2023-11-28
Docs: Add `titleAttribute` to Select example

ahinkle:
This PR adds `titleAttribute` to the documentation select example as `titleAttribute` is required.

Screenshot 2023-11-28 at 4 15 26 PM


danharrin:
Let's make that parameter optional, the point of that example is that you shouldn't need to pass the title attribute if you define how the query is searched and ordered etc

danharrin:
Actually I've just checked, that parameter is optional in v3

Issue #9933, opened 2023-11-28
Column grow() only works within a Split or Stack


nmalinoski:
### Package

filament/filament

### Package Version

3.0.103

### Laravel Version

10.34.1

### Livewire Version

3.2.1

### PHP Version

8.1.25

### Problem description

When adding columns to a table for a model resource, all columns take up an equal percentage of the table width, even when adding `grow(false)` to the TextColumn.

### Expected behavior

I expected the columns with `grow(false)` to shrink to their minimum width, as described in the manual. This does work if I wrap these columns in a Split, but Splits don't support labels; so I have to choose between sensible column widths and column headings?

### Steps to reproduce

1. Create a new project with Laravel 10.x and Filament 3.x
1. Follow Filament 3.x Panels install documentation to install filament panels and create a new user account
1. Create an ExampleModel model, db migration, and seeder, with 3 columns; run migration and seed some sample data
1. Create a control panel and index table for ExampleModel (`php artisan make:filament-resource ExampleModel --generate`)
1. Log into the admin panel and select Example Models to view the table to see that the columns are all about the same width
1. Add to the `table` function in `app/Filament/Resources/ExampleModelResource.php`:
 ```
    public static function table(Table $table): Table
    {
        return $table
            ->columns([
                //
                TextColumn::make('col1')
                    ->grow(false),
                TextColumn::make('col2'),
                IconColumn::make('col3')
                    ->grow(false),
            ])
...
    }
```
7. Reload the page to see that col1 and col3 columns have _not_ shrunk to their minimum width with col2 taking up the remaining space

### Reproduction repository

https://github.com/nmalinoski/filament-grow-column-not-working

nmalinoski:
This change has the effect I'm looking for: https://github.com/filamentphp/filament/compare/3.x...nmalinoski:filament:9933-apply-column-grow-to-html-column

danharrin:
If you want to make a PR then, go for it. Does flex-1 really make a difference in tables though?

nmalinoski:
> If you want to make a PR then, go for it. Does flex-1 really make a difference in tables though?

Looks like it doesn't. I dropped it from the added classes. I'll submit a PR shortly.

PR #9936, opened 2023-11-29
9933 apply column grow to html column

nmalinoski:
Support use of column `grow` on HTML columns, instead of just within Splits and Stacks.
https://github.com/filamentphp/filament/issues/9933


zepfietje:
Cleaned this up in https://github.com/filamentphp/filament/commit/291a641d8bb4d49f0ac7c018cfc700ed3bff4c6b.

PR #9974, opened 2023-11-30
Make action group dropdown offset customizable

mwikala:
No visual changes, as we keep the default dropdown offset of 8 and just allow that to be changed.

`HasDropdown` now has the `dropdownOffset` and `getDropdownOffset` methods. And inside the Action Group blade component `group.blade.php` we pass that prop to the underlying dropdown which already supports an `offset` prop that wasn't being utilised. 

### DIFF
--- a/packages/actions/docs/05-grouping-actions.md
+++ b/packages/actions/docs/05-grouping-actions.md
@@ -91,3 +91,14 @@ ActionGroup::make([
 ])
     ->maxHeight('400px')
 ```
+
+### Controlling the dropdown offset
+
+You may control the offset of the dropdown using the `dropdownOffset()` method, by default the offset is set to `8`.
+
+```php
+ActionGroup::make([
+    // Array of actions
+])
+    ->dropdownOffset(16)
+```
--- a/packages/actions/docs/05-grouping-actions.md
+++ b/packages/actions/docs/05-grouping-actions.md
@@ -92,7 +92,7 @@ ActionGroup::make([
     ->maxHeight('400px')
 ```
 
-### Controlling the dropdown offset
+## Controlling the dropdown offset
 
 You may control the offset of the dropdown using the `dropdownOffset()` method, by default the offset is set to `8`.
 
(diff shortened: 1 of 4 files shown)


danharrin (changes requested):
Can we keep all the `dropdown*` properties in alphabetical order please? :)

packages/actions/resources/views/components/group.blade.php
```diff
@@ -7,6 +7,7 @@
     'dropdownMaxHeight' => null,
     'dropdownPlacement' => null,
     'dropdownWidth' => null,
+    'dropdownOffset' => 8,
```
zepfietje: Please remove the default here, so we have it just in a single place (the actual dropdown Blade component).

zepfietje (changes requested):

packages/actions/src/Concerns/HasDropdown.php
```diff
@@ -13,8 +13,11 @@ trait HasDropdown
 
     protected string | Closure | null $dropdownMaxHeight = null;
 
+    protected int | Closure | null $dropdownOffset = 8;
```
danharrin: Can this be null by default?

danharrin (changes requested):

packages/actions/src/Concerns/HasDropdown.php
mwikala: It could, I kept it here so that the default would be clearly visible in User land without having to dig to the underlying blade component.

What do you think? Make it null here too and leave the default in the blade component?

zepfietje: Yup, the latter, as per my previous comment.

packages/actions/docs/05-grouping-actions.md
```diff
@@ -91,3 +91,14 @@ ActionGroup::make([
 ])
     ->maxHeight('400px')
 ```
+
+### Controlling the dropdown offset
```
danharrin:
```suggestion
## Controlling the dropdown offset
```

PR #9995, opened 2023-12-01
Add option to remove default constraint operators'

tiagof:
- [n.a.] Visual changes are explained in the PR description using a screenshot/recording of before and after.

### DIFF
--- a/packages/tables/docs/04-filters/04-query-builder.md
+++ b/packages/tables/docs/04-filters/04-query-builder.md
@@ -352,19 +352,43 @@ TextConstraint::make('author')
 Each constraint type has a set of default operators, which you can customize by using the `operators()`method:
 
 ```php
+use Filament\Tables\Filters\QueryBuilder\Constraints\Operators\IsFilledOperator;
 use Filament\Tables\Filters\QueryBuilder\Constraints\TextConstraint;
-use Tables\Filters\QueryBuilder\Constraints\TextConstraint\Operators\EqualsOperator;
 
 TextConstraint::make('author')
     ->relationship(name: 'author', titleAttribute: 'name')
     ->operators([
-        EqualsOperator::make() // or a custom operator
-    ], keepDefault: true)
+        IsFilledOperator::make(),
+    ])
 ```
 
-This will add the `EqualsOperator` to the existing/default operators.
+This will remove all operators, and register the `EqualsOperator`.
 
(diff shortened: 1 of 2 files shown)


danharrin:
Instead of introducing a new argument, how about a new method `pushOperators()`

danharrin:
Then we use `pushOperators()` from within `operators()` after the array is reset

tiagof:
I totally agree but in that case the behavior of the operators() method is changed and could impact others. 
Are you ok with that?

EDIT: on second thought, wouldn't it be cleaner to have just one the "operator adding" method, with optional parameters to modify its behaviour?
For example, I like that the position of `table->filters()` is modified by the optional 2nd parameter `layout`.
On the other hand, it looks less clean (to me at least) when we have `listWithLineBreaks()` and `bulleted()` when we could have the `bulleted()` "modification" as an optional parameter.

Anyway, up to you to decide which is more aligned with the DX. Happy to implement wither way!

danharrin:
Let's just have 1 `pushOperators()` method that appends to the end of the array

Refactoring `operators()` should not involve breaking changes

tiagof:
But `operators()`already appends to the end of the array, that's the "problem". (and maybe an unplanned behaviour?)
I'm trying to add the possibility to "replace" the array without breaking changes.

Maybe a new `replaceOperators()`method then?

danharrin:
`operators()` should do the same thing, and replace all operators.

`pushOperators()` should add operators onto the end of the array.

There is no breaking change, I was just suggesting a refactor that we can reuse `pushOperators()` inside `operators()` once the array is cleared out

tiagof:
> `operators()` should do the same thing, and replace all operators.

This would be a breaking change since right now `operators` is appending to the end of the array.

I get your point to split the functionality into 2 methods, no issue here. 

To avoid breaking changes we either add the parameter as proposed by the current commits in this PR
or
we add a new replaceOperators() method like:

```php
public function replaceOperators(array $operators): self
{
    $this->operators = [];

    return $this->operators($operators); // this is actually pushing operators, not replacing
}
```

(Sorry if there's something loss in translation here)

danharrin:
Sorry, I am the one confused here. I thought the existing behaviour was opposite.

danharrin:
I'm actually going to make this breaking change, as I consider it a bug. The API is currently inconsistent with other areas of Filament like `actions()` and `pushActions()`

PR #10007, opened 2023-12-01
replace flip with autoPlacement in x-float

ariaieboy:
Ok, there is a weird bug with the floating UI when we are in RTL mode.

https://github.com/filamentphp/filament/assets/15873972/48b86d5c-e410-41d7-bd80-a86e2b01f0c7

For example, here we have a date filter in our resource table and when we click on the date picker the floatingui library keeps changing the component position indefinitely.
and that's because of the `.flip` middleware. 
Since `autoPlacement` is kind of have similar functionality and does not have the same bug in RTL mode I changed the `flip` option with `autoPlacement`.

This bug originally reported here :https://github.com/ariaieboy/filament-jalali-datetimepicker/issues/16


themahdavi:
Hello,
thanks to @ariaieboy for providing a solution to that issue.

I believe @awcodes can be very productive in this case because of his make the x-float package.

danharrin:
Yes, I am not going to merge this unless @awcodes approves, he is the expert here

awcodes:
I can't think of why rtl would affect this. If autoPlacement works, that's fine. But the point of using flip vs autoPlacement is that flip constrains it to the same axis. Personally, from a UX perspective, I wouldn't want the calendar to potentially pop up on the left or the right of the field.

danharrin:
How about only enabling autoPlacement in RTL mode then? So others arent affected

ariaieboy:
we can set the https://floating-ui.com/docs/autoplacement#allowedplacements and disable left and right alignment completely 
By default auto placement align the object to the side that has the most available space.

danharrin:
Sounds good

awcodes:
works for me too. thanks.

ariaieboy:
@awcodes @danharrin 
Should I disable left and right alignment in all 3 components or just for the date time picker?

awcodes:
I'll let Dan make that call, but my opinion is that if the bug only appears in the Date time picker then only apply it there.

danharrin:
All 3 probably?

ariaieboy:
@awcodes can you check it now ?

packages/forms/resources/views/components/color-picker.blade.php
```diff
@@ -91,7 +91,7 @@ class="h-5 w-5 rounded-full"
                 wire:ignore.self
                 wire:key="{{ $this->getId() }}.{{ $statePath }}.{{ $field::class }}.panel"
                 x-cloak
-                x-float.placement.bottom-start.offset.flip.shift="{ offset: 8 }"
+                x-float.placement.bottom-start.offset.autoPlacement.shift="{ offset: 8 }"
```
danharrin:
```suggestion
                x-float.placement.bottom-start.offset.autoPlacement.shift="{
                        offset: 8,
                        autoPlacement: { allowedPlacements: ['top', 'bottom'] },
                    }"
```

packages/support/resources/views/components/dropdown/index.blade.php
```diff
@@ -37,7 +37,7 @@
 
     <div
         x-cloak
-        x-float{{ $placement ? ".placement.{$placement}" : '' }}.flip{{ $shift ? '.shift' : '' }}{{ $teleport ? '.teleport' : '' }}{{ $offset ? '.offset' : '' }}="{ offset: {{ $offset }} }"
+        x-float{{ $placement ? ".placement.{$placement}" : '' }}.autoPlacement{{ $shift ? '.shift' : '' }}{{ $teleport ? '.teleport' : '' }}{{ $offset ? '.offset' : '' }}="{ offset: {{ $offset }},autoPlacement:{allowedPlacements: ['top', 'bottom']} }"
```
danharrin:
```suggestion
        x-float{{ $placement ? ".placement.{$placement}" : '' }}.autoPlacement{{ $shift ? '.shift' : '' }}{{ $teleport ? '.teleport' : '' }}{{ $offset ? '.offset' : '' }}="{
                        offset: {{ $offset }},
                        autoPlacement: { allowedPlacements: ['top', 'bottom'] },
                    }"
```

packages/forms/resources/views/components/color-picker.blade.php
danharrin: Autoplacement missing here, right?

ariaieboy: I didn't set it for the color picker cause I think for color picker it can go to the sides if needed.
but for dropdowns or date pickers, it is better to keep it at the top or bottom of the field. 

if you think we should add this for the color picker I can commit your suggestion.

zepfietje: Let's keep things consistent for all components.

zepfietje:
@danharrin @ariaieboy, I've read a little more about `autoPlacement` vs `flip` and found that this PR doesn't keep the original behavior.

https://floating-ui.com/docs/autoPlacement#conflict-with-flip

`flip()` uses a fallback “no space” strategy. Ensures the preferred placement is kept unless there is no space left.

`autoPlacement()` uses a primary “most space” strategy. Always chooses the placement with the most space available.

I think this results in the configured placement being ignored completely now. Honestly, I think we should therefore revert this PR.

danharrin:
Okay, if it's a BC with configured placement I will revert

PR #10008, opened 2023-12-01
Make date time picker columns grow when label is too wide

ariaieboy:
before in fa lang
image

after in fa lang
image

after in en lang
image


danharrin:
Pretty sure there is a reason we did it like that before

zepfietje:
> Pretty sure there is a reason we did it like that before

Yeah, probably for min width of the date buttons.

danharrin:
Is the min width the `spacing.7` bit?

zepfietje:
Yeah I suppose.

zepfietje (changes requested):
Please make it so that the default width is still used, and the buttons only grow when the label is too wide, @ariaieboy.

packages/forms/resources/views/components/date-time-picker.blade.php
```diff
@@ -155,7 +155,7 @@
                         'fi-fo-date-time-picker-panel absolute z-10 rounded-lg bg-white p-4 shadow-lg ring-1 ring-gray-950/5 dark:bg-gray-900 dark:ring-white/10',
                     ])
                 >
-                    <div class="grid gap-y-3">
+                    <div class="gap-y-3">
```
danharrin: Gap only works with grid or flex elements, right?

ariaieboy: I've only changed line 197 what is this codes 🤔

danharrin: Please check the full git diff

zepfietje:
Tested the change and it seems to keep the original behavior when labels aren't too wide.
Thanks, @ariaieboy!

zepfietje (approved):

PR #10009, opened 2023-12-01
add should_translate_labels_by_default config

RoduanKD:
- [ ] ~Visual changes are explained in the PR description using a screenshot/recording of before and after.~ (no visual changes)

### DIFF
--- a/packages/support/config/filament.php
+++ b/packages/support/config/filament.php
@@ -73,10 +73,10 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Translate Labels By Deafult
+    | Translate Labels By Default
     |--------------------------------------------------------------------------
     |
-    | This determines if labels should be translate by default.
+    | This determines if labels should be translatable by default.
     |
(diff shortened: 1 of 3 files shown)


packages/support/config/filament.php
```diff
@@ -71,4 +71,17 @@
 
     'livewire_loading_delay' => 'default',
 
+    /*
+    |--------------------------------------------------------------------------
+    | Translate Labels By Deafult
+    |--------------------------------------------------------------------------
+    |
+    | This determines if labels should be translate by default.
```
WaleedAlrashed: translat**able**

danharrin:
We highly discourage using `translateLabel()` as it encourages incredibly bad translation practice. It ignores the fact that many locales require translations to be contextual, not just a 1:1 replacement with English.

Adding this feature would encourage people to use `translateLabel()`, and we really don't want to do that.

Also, it is already possible using `Filament\Forms\Components\Field::configureUsing(fn ($field) => $field->translateLabel())`

PR #10010, opened 2023-12-01
Fix layout break in notification for long words

kainiklas:

If the notification text is too long the layout of the notification items breaks, e.g., icons disapear. See screenshot.
notification-before

The fix breaks words and therefore fixes the layout. See screenshot.
notification-fixed

### DIFF
--- a/packages/notifications/resources/views/components/body.blade.php
+++ b/packages/notifications/resources/views/components/body.blade.php
@@ -1,5 +1,5 @@
 <p
-    {{ $attributes->class(['fi-no-notification-body text-sm text-gray-500 dark:text-gray-400 break-all']) }}
+    {{ $attributes->class(['fi-no-notification-body text-sm text-gray-500 dark:text-gray-400 break-words overflow-hidden']) }}
 >
     {{ $slot }}
 </p>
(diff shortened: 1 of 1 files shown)


packages/notifications/resources/views/components/body.blade.php
```diff
@@ -1,5 +1,5 @@
 <p
-    {{ $attributes->class(['fi-no-notification-body text-sm text-gray-500 dark:text-gray-400']) }}
+    {{ $attributes->class(['fi-no-notification-body text-sm text-gray-500 dark:text-gray-400 break-all']) }}
```
zepfietje: Shouldn't we use `break-words` instead?
https://tailwindcss.com/docs/word-break#break-words

zepfietje (changes requested):

packages/notifications/resources/views/components/body.blade.php
kainiklas: In theory `break-words` should be the better alternative. But long words need to be forced with `break-all`. My example didn't work with `break-words`.

zepfietje: `break-all` also breaks text when it shouldn't, so we can't use that.

kainiklas: Agree. This does not look nice. What about using `break-words overflow-hidden`? This solves the problem as well.

zepfietje: Is `overflow-hidden` really required for it to work?

kainiklas: As far as I can tell: Yes. See screenshots.

Without `overflow-hidden`:

without_overflow_hidden

With `overflow-hidden`:
with_overflow_hidden

zepfietje: Alright, thanks.

zepfietje (approved):

PR #10043, opened 2023-12-03
fix: remove failed at check

maartenpaauw:
Using CSV import action in combination with strict models (`Model::shouldBeStrict(! $this->app->isProduction());`) fails, because the Import model does not have a `failed_at` attribute.

After removing this line, the success notification later on is being send. I'm not sure if there needs to be another way to check if the import failed?


danharrin:
Huh, not sure where that's left over from, thanks

PR #10087, opened 2023-12-05
Update infoList documentation

jcsoriano:
In `Filament\Resources\Pages\ViewRecord.php` line 219, `infolist` is not a static method. Copying from the docs causes the IDE to complain that it is 'not compatible' with the method in ViewRecord

image


zepfietje:
The method is static in fact on resource classes.

PR #10103, opened 2023-12-05
Fixed 'Is Blank' operator for isFilledOperator

Justin991q:

I noticed when I used the new query builder with the following code and selected 'Is Blank' as operator, it did not select all the blank columns.

```php
QueryBuilder\Constraints\TextConstraint::make('embed_title')->nullable(),
```

**My data looks like this:**
image

**Previous:**
```sql
select
  count(*) as aggregate
from
  `users`
where
  (
    (
      `users`.`embed_title` is null
      and `users`.`embed_title` = ''
    )
  )
  and `users`.`deleted_at` is null
  ```

image

**After this change:**
```sql
select
  count(*) as aggregate
from
  `users`
where
  (
    (
      `users`.`embed_title` is null
      or `users`.`embed_title` = ''
    )
  )
  and `users`.`deleted_at` is null
```

image


danharrin:
Good spot!

PR #10113, opened 2023-12-06
Enables autofocusing in wizard on page load

maltebaer:
Hey folks,

I noticed that autofocusing an input does not work on the first/default step of a wizard.

I'm not sure if that is a bug or intentionally omitted, due to some reason I'm not aware of.

If it was disabled on purpose, feel free to close this PR. Otherwise I'm happy to tackle the three points below (where necessary).


danharrin:
Can you explain why the fields inside the wizard arent autofocused in the same way that ones outside the wizard are by default?

maltebaer:
Honestly, I do not really know why this happens. And apparently it does work from time to time.

I tried to reproduce the bug in the filament demo. At least on my machine it does not work as expected.

You can check it out here: https://github.com/filamentphp/demo/compare/main...maltebaer:filament-demo:main

zepfietje:
@danharrin, I'm not sure if this is a good idea when you have other fields outside the wizard on the same page? I guess autofocus on those fields would be broken by the wizard field, or is that fine?

danharrin:
I thought that too, before I realised that only one field on a page should be autofocused anyway

zepfietje:
Yeah exactly, should be fine then.

PR #10131, opened 2023-12-06
use empty string as fallback when calling $record->getAttributeValue()

wivaku:
`$record->getAttributeValue()` requires a string according to Laravel's code: https://github.com/laravel/framework/blob/7d26bfb23fbc2d82adb095d4c6c28fe7f2e4bc4b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L481-L490

However: `getRecordTitleAttribute` returns an optional string. This results in an error for `getAttributeValue()` when `getRecordTitleAttribute()` returned null.

Therefore: use empty string as fallback when calling `$record->getAttributeValue()`

### DIFF
--- a/packages/tables/src/Table/Concerns/HasRecords.php
+++ b/packages/tables/src/Table/Concerns/HasRecords.php
@@ -112,7 +112,7 @@ trait HasRecords
                 Model::class => $record,
                 $record::class => $record,
             ],
-        ) ?? $record->getAttributeValue($this->getRecordTitleAttribute() ?? '') ?? $this->getModelLabel();
+        ) ?? (($titleAttribute = $this->getRecordTitleAttribute()) ? $record->getAttributeValue($titleAttribute) : null) ?? $this->getModelLabel();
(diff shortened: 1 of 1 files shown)


packages/tables/src/Table/Concerns/HasRecords.php
```diff
@@ -112,7 +112,7 @@ public function getRecordTitle(Model $record): string
                 Model::class => $record,
                 $record::class => $record,
             ],
-        ) ?? $record->getAttributeValue($this->getRecordTitleAttribute()) ?? $this->getModelLabel();
+        ) ?? $record->getAttributeValue($this->getRecordTitleAttribute() ?? '') ?? $this->getModelLabel();
```
danharrin:
```suggestion
        ) ?? (($titleAttribute = $this->getRecordTitleAttribute()) ? $record->getAttributeValue($titleAttribute) : null) ?? $this->getModelLabel();
```

PR #10134, opened 2023-12-06
Rename navigationParentItem variable for clarity

binaryfire:
Minor thing, but as someone who digs through the codebase this confused me. `$group` implies the parent item can only be a group but in reality it can be either a group or an item.


binaryfire:
I think the tests failed for an unrelated reason. Packagist has been having issues today- maybe Monolog wasn't downloaded properly.

PR #10138, opened 2023-12-06
Add doc note about dashboard sort orders

drbyte:
Inspired from one of Dan's comments on Discord.

### DIFF
--- a/packages/panels/docs/05-dashboard.md
+++ b/packages/panels/docs/05-dashboard.md
@@ -273,3 +273,11 @@ You may also customize the title of the dashboard by overriding the `$title` pro
 ```php
 protected static ?string $title = 'Finance dashboard';
 ```
+
+The primary dashboard shown to a user is the first one they have access to (controlled by `canView()`, see above), according to the defined sort order.
+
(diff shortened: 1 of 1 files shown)


drbyte:
Should more also be added like `shouldRegisterNavigation()`?

packages/panels/docs/05-dashboard.md
```diff
@@ -273,3 +273,11 @@ You may also customize the title of the dashboard by overriding the `$title` pro
 ```php
 protected static ?string $title = 'Finance dashboard';
 ```
+
+The primary dashboard shown to a user is the first one they have access to (controlled by `canView()`, see above), according to the defined sort order.
+
+The default sort order for dashboards is `-2`. You can control the sort order of custom dashboards with `$navigationSort`:
+
+```php
+    protected static ?int $navigationSort = 15;
```
danharrin:
```suggestion
protected static ?int $navigationSort = 15;
```

danharrin:
I think `shouldRegisterNavigation()` can be implied based on the navigation docs

PR #10143, opened 2023-12-06
fix: Modal content called multiple times

### LINKED ISSUE #8763: Action button with modal execute multiple duplicated calls
### Package

filament/filament

### Package Version

3.0.62

### Laravel Version

10.25.1  

### Livewire Version

3.0.5

### PHP Version

8.1.23

### Problem description

when passing action arguments to modal content view it become empty

a behaviour that is observed is `Action->modalContent` is being called 4 times and only the first call contain the arguments
```php
[2023-09-28 09:56:16] local.INFO: add {"product":1} 
[2023-09-28 09:56:16] local.INFO: add  
[2023-09-28 09:56:16] local.INFO: add  
[2023-09-28 09:56:16] local.INFO: add  
```

### Expected behavior

only call one time and the arguments are passed to the view

### Steps to reproduce

clone the repo and go to "Action Test" page

### Reproduction repository

https://github.com/wychoong/filament-test/tree/passing-arguments-to-action-modal-view

### Relevant log output

_No response_

danharrin:
Fixes #8763.

### DIFF
--- a/packages/actions/src/Concerns/CanOpenModal.php
+++ b/packages/actions/src/Concerns/CanOpenModal.php
@@ -508,6 +508,16 @@ trait CanOpenModal
         return $this->evaluate($this->modalContentFooter);
     }
 
+    public function hasModalContent(): bool
+    {
+        return $this->modalContent !== null;
+    }
+
+    public function hasModalContentFooter(): bool
+    {
+        return $this->modalContentFooter !== null;
+    }
+
     public function getCustomModalHeading(): string | Htmlable | null
     {
         return $this->evaluate($this->modalHeading);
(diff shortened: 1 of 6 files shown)


damiandejong:
Hi,

This fix breaks the ability to call a modal conditionally. In my application, I have an action that depends on the record's status whether it can execute immediately or whether it requires a confirmation modal. I set this up by passing the record into the closure method of eg. `modalDescription()` and will return something in case the modal is needed and `null` in case the action can be executed right away. In this situation, it is desireable to have the closure evaluated which `hasModalDescription()` now doesn't.

Is there currently another way to achieve the above? Otherwise, might it be an idea to have a dedicated function like `shouldOpenModal()` on the action that can overrule `mountedActionShouldOpenModal()` in case the content is filled?

danharrin:
@damiandejong https://filamentphp.com/docs/3.x/actions/modals#conditionally-hiding-the-modal

damiandejong:
Awesome, that's exactly what I meant. Sorry that I missed it already exists. Thanks for replying and all the good work!

danharrin:
I've reverted the description change

PR #10151, opened 2023-12-07
Allow unlabelled navigation groups

binaryfire:
Currently, standalone (ungrouped) nav items are always at the top of the menu: https://github.com/filamentphp/filament/blob/b8882ecaed8a9c00b76c4f25fb7c1e033fc5ef99/packages/panels/src/Panel/Concerns/HasNavigation.php#L150
```
            ->sortBy(function (NavigationGroup $group, ?string $groupIndex): int {
                if (blank($group->getLabel())) {
                    return -1;
                }
```
And groups are always displayed underneath. 

This limits the possibilities for navigation customization. For example, you can't mix it up like this:
image

The order in the above example is:

```
Home (item)
Store (group)
Apps (group)
Finances (item)
Marketing (group)
Feedback (item)
Resolution center (item)
```

This PR adds a `->hideLabel()` method for groups which allows groups of top level items to be positioned anywhere in the menu without looking like a group. It would be cleaner to refactor the overall sorting concept but I think it would be a BC. This was the best workaround I could think of.

With this PR, this is now possible:

image

### DIFF
--- a/packages/panels/src/Navigation/NavigationGroup.php
+++ b/packages/panels/src/Navigation/NavigationGroup.php
@@ -22,6 +22,8 @@ class NavigationGroup extends Component
 
     protected string | Closure | null $label = null;
 
+    protected bool | Closure | null $hideLabel = null;
+
     final public function __construct(string | Closure | null $label = null)
     {
         $this->label($label);
@@ -75,6 +77,13 @@ class NavigationGroup extends Component
         return $this;
     }
 
+    public function hideLabel(bool | Closure | null $condition = true): static
+    {
+        $this->hideLabel = $condition;
+
+        return $this;
+    }
+
     public function getIcon(): ?string
     {
         $icon = $this->evaluate($this->icon);
@@ -97,6 +106,11 @@ class NavigationGroup extends Component
     public function getLabel(): ?string
     {
         return $this->evaluate($this->label);
+    }
+
+    public function hasHiddenLabel(): bool
+    {
+        return (bool) $this->evaluate($this->hideLabel);
     }
 
     public function isCollapsed(): bool
--- a/packages/panels/src/Navigation/NavigationGroup.php
+++ b/packages/panels/src/Navigation/NavigationGroup.php
@@ -22,7 +22,7 @@ class NavigationGroup extends Component
 
     protected string | Closure | null $label = null;
 
-    protected bool | Closure | null $hideLabel = null;
+    protected bool | Closure $isLabelHidden = false;
 
     final public function __construct(string | Closure | null $label = null)
     {
@@ -77,9 +77,9 @@ class NavigationGroup extends Component
         return $this;
     }
 
-    public function hideLabel(bool | Closure | null $condition = true): static
+    public function hiddenLabel(bool | Closure $condition = true): static
     {
-        $this->hideLabel = $condition;
+        $this->isLabelHidden = $condition;
 
         return $this;
     }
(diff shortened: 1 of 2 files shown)


danharrin:
1) Please follow the naming convention set out for methods and properties using the `hiddenLabel()` feature of form fields, instead of `hideLabel()`.
2) Please check how this works with top navigation.

binaryfire:
@danharrin Naming fixed. I haven't used top navigation much but looks like the same improvement applies? This is the same example with `->topNavigation()` enabled:
image

Now we have `Item | Group (dropdown) | Item | Item`. I'm assuming that wasn't possible before for the same reason - i.e. standalone items could only be positioned before (to the left of) dropdowns

binaryfire:
I think the cleanest solution would be to not put standalone items into nameless groups at all, add a  `->sort` method to `NavigationGroup` and then use the sort int to order all root items in the final nav array regardless of type. Putting items into a group and then removing the label just to reposition them feels a bit hacky. But pretty sure that would mean a BC?

zepfietje:
Internally, separate items are already grouped into a group.

binaryfire:
@zepfietje Yeah that's what this PR takes advantage of. Currently it only happens for the one group at the top. Which means there's no way to put separate items anywhere else (eg. between or under groups).

The reason that top group has no dropdown container is because the group has no name. This PR just hijacks that logic by removing the name from any group with `hiddenLabel()`. Which then lets you put more groups of separate items anywhere.

As I mentioned in https://github.com/filamentphp/filament/pull/10151#issuecomment-1845158479 it might be cleaner to just remove the internal grouping of separate items and add a `->sort()` to `NavigationGroup`. Then you could treat all root navigation items (groups and separate items) the same way and sort them by their sort value. But that refactor would probably be a breaking change.

zepfietje:
I think separate items would still need to be grouped for the way they're rendered in groups in Blade.

binaryfire:
@zepfietje Ok cool :) That's what this is currently doing. 

Also I've been been playing around with top nav some more and would definitely want a way to put separate items between / to the right of dropdowns. This solves that too.

zepfietje:
@binaryfire, honestly I'm confused by what this PR tries to achieve.

It's currently already possible to create the navigation in your second screenshot:
```php
->navigation(
    fn (NavigationBuilder $navigation): NavigationBuilder => $navigation->groups([
        NavigationGroup::make()
            ->items([
                NavigationItem::make('Dashboard')
                    ->icon('heroicon-o-home'),
            ]),
        NavigationGroup::make('Users')
            ->items([
                NavigationItem::make('Users')
                    ->icon('heroicon-o-users'),
                NavigationItem::make('Roles')
                    ->icon('heroicon-o-lock-closed'),
            ]),
        NavigationGroup::make()
            ->items([
                NavigationItem::make('Help')
                    ->icon('heroicon-o-lifebuoy'),
                NavigationItem::make('Settings')
                    ->icon('heroicon-o-cog-8-tooth'),
            ]),
    ]),
)
```
Screenshot 2023-12-13 at 11 26 18

binaryfire:
@zepfietje No worries. Was trying to find a simpler way of doing this than custom navigation. Custom navigation means you need to recreate all resource link logic (urls, active states etc.) manually:

 ```
                       NavigationGroup::make()
                            ->items([
                                NavigationItem::make('Settings')
                                    ->url(route('filament.admin.resources.settings.index'), shouldOpenInNewTab: false)
                                    ->icon('heroicon-o-cog')
                                    ->isActiveWhen(fn (): bool => request()->routeIs('filament.admin.resources.settings.index')),
                        ]),
                        // Other groups / resources....
```

That's a lot of boilerplate just to have individual items in places other than at the top (sidebar) or on the left (topbar), which is a common requirement. But hopefully the alternative group type you mentioned in https://github.com/filamentphp/filament/pull/10164#issuecomment-1845900247 will address this.

PR #10164, opened 2023-12-07
Add compact sidebar navigation

binaryfire:
In most admin panels, standalone items and dropdown items look the same and are evenly spaced. Visually a "group" is just a collapsible item with children. For example:
image

This PR adds a `->compactSidebarNavigation()` method which makes the spacing between all items in the sidebar the same regardless of whether it's a group of standalone item.

With `->compactSidebarNavigation()`:
image
image

Without `->compactSidebarNavigation()` (the current / default setting):
image
image

### DIFF
--- a/packages/panels/docs/06-navigation.md
+++ b/packages/panels/docs/06-navigation.md
@@ -187,6 +187,21 @@ public function panel(Panel $panel): Panel
 }
 ```
 
+### Using compact sidebar navigation
+
+By default, navigation groups have extra top and bottom padding. You may disable this and make all navigation items evenly spaced using the `compactSidebarNavigation()` method:
+
+```php
+use Filament\Panel;
+
+public function panel(Panel $panel): Panel
+{
+    return $panel
+        // ...
+        ->compactSidebarNavigation();
+}
+```
+
 ## Registering custom navigation items
 
 To register new navigation items, you can use the [configuration](configuration):
--- a/packages/panels/docs/06-navigation.md
+++ b/packages/panels/docs/06-navigation.md
@@ -189,7 +189,7 @@ public function panel(Panel $panel): Panel
 
 ### Using compact sidebar navigation
 
-By default, navigation groups have extra top and bottom padding. You may disable this and make all navigation items evenly spaced using the `compactSidebarNavigation()` method:
+By default, navigation groups have extra top and bottom padding. You may remove this and make all navigation items evenly spaced using the `compactSidebarNavigation()` method:
 
 ```php
 use Filament\Panel;
--- a/packages/panels/resources/views/components/sidebar/index.blade.php
+++ b/packages/panels/resources/views/components/sidebar/index.blade.php
@@ -113,7 +113,7 @@
         @endif
 
         @if (filament()->hasNavigation())
-            <ul class="-mx-2 flex flex-col gap-y-7">
+            <ul class="-mx-2 flex flex-col {{ filament()->hasCompactSidebarNavigation() ? 'gap-y-1' : 'gap-y-7' }}">
                 @foreach ($navigation as $group)
                     <x-filament-panels::sidebar.group
                         :collapsible="$group->isCollapsible()"
--- a/packages/panels/src/FilamentManager.php
+++ b/packages/panels/src/FilamentManager.php
@@ -616,6 +616,11 @@ class FilamentManager
         return $this->getCurrentPanel()->isSidebarFullyCollapsibleOnDesktop();
     }
 
+    public function hasCompactSidebarNavigation(): bool
+    {
+        return $this->getCurrentPanel()->hasCompactSidebarNavigation();
+    }
+
     public function mountNavigation(): void
     {
         $this->getCurrentPanel()->mountNavigation();
--- a/packages/panels/src/Panel/Concerns/HasSidebar.php
+++ b/packages/panels/src/Panel/Concerns/HasSidebar.php
@@ -14,6 +14,8 @@ trait HasSidebar
 
     protected bool $hasCollapsibleNavigationGroups = true;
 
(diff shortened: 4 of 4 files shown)


binaryfire:
This PR will be especially useful when combined with https://github.com/filamentphp/filament/pull/10151. It'll allow the creation of navigation layouts like in the example screenshot above.

PS. Do you guys think the extra space around groups should just be removed? It looks strange to me. I've haven't seen any other panels which do it:

https://flowbite.com/blocks/application/sidenav/#default-side-navigation
https://pro.chakra-ui.com/components/application/sidebars

zepfietje:
@binaryfire, the feature you're looking for is [nested navigation items](https://filamentphp.com/docs/3.x/panels/navigation#grouping-navigation-items-under-other-items).

Maybe the actual issue is that you want the parent item to not have a link itself?
Closing this PR, but we can discuss further improvements to the nested navigation items.

binaryfire:
@zepfietje Zep this is different  :) Nested navigation items isn't great for this use case. There's no up/down chevron to indicate whether there are child items and there's no animation to slide open the menu.

zepfietje:
That's what we should add then, @binaryfire.
Having child items without the parent having a link is what we need.
Trust me, we've researched the nav design a lot.

binaryfire:
@zepfietje There are other issues too. Using nesting rather than groups means that clicking the parent will navigate to the parent resource. A lot of the time you just want the parent to be a placeholder that opens the dropdown. Doesn't all this mean we'd just be re-inventing groups in another place?

zepfietje:
I agree, @binaryfire. That's why I've suggested that exact API (nesting groups instead of items) to @danharrin when he was working on it. He preferred to have parent items with links, so that's what was released in 3.1.0. What I'm suggesting (and I've been thinking about before 3.1.0 release already) is that we add support for using a group instead of a parent item. This can be done in a backward compatible way.

We wouldn't be reinventing groups, as there's going to be 2 types of groups essentially. The top-level groups which are collapsible by default (which shouldn't be collapsible by default in my opinion, but that would be breaking), and the groups that look exactly like and are positioned inline with items. Does that clear up any confusion, @binaryfire?

zepfietje:
So after suggested changes, you could still have groups with spacing between them, and one of those groups could then look like the screenshot you provided in this PR description.

binaryfire:
@zepfietje My brain currently has groups in subgroups nested under parents which have dropdowns inside other groups which are compact that have parents with links (sometimes).

So I need sleep and coffee before I reply.

binaryfire:
@zepfietje You said "you could still have groups with spacing between them". Did you mean "without"? If so then yep that's great! And sorry, I didn't properly read your comment about the parent not having a link. Yes that's what I was looking for.

So to confirm, this new group type would:

- Look exactly like an item (can have an icon, will be positioned inline with items and use the same top/bottom padding)
- Won't have a parent link and will use the same open/close chevron + animation as normal groups
- Will be able to have children with icons if the parent doesn't, and vice versa (like normal groups)
- Unlike normal groups, will be able to be ordered anywhere relative to individual items (will it have a `->sort()` method?)

Have I understood correctly? And would these groups allow nesting too? The example screenshot only has 1 level of nesting but what if I wanted to add `Gold` and `Silver` to that layout under `Marketing` -> `Affiliates`. Will that be possible? That's still only 2 levels of nesting.

zepfietje:
> You said "you could still have groups with spacing between them". Did you mean "without"?

No, I meant the top-level groups would still exist like in the current design, but you simply wouldn't notice that in the nav structure of your screenshot, as all items (and nested groups once we support them) would be part of a single top-level group.

> So to confirm, this new group type would:
> 
> * Look exactly like an item (can have an icon, will be positioned inline with items and use the same top/bottom padding)

Yes.

> * Won't have a parent link and will use the same open/close chevron + animation as normal groups

Yes.

> * Will be able to have children with icons if the parent doesn't, and vice versa (like normal groups)

Not too sure about that.

> * Unlike normal groups, will be able to be ordered anywhere relative to individual items (will it have a `->sort()` method?)

That's something for you and Dan to figure out. 😄

> Have I understood correctly? And would these groups allow nesting too? The example screenshot only has 1 level of nesting but what if I wanted to add `Gold` and `Silver` to that layout under `Marketing` -> `Affiliates`. Will that be possible? That's still only 2 levels of nesting.

Nope, that would be 3 levels of nesting, since the top-level group still exists (as I mentioned at the top of this comment).

binaryfire:
@zepfietje Sorry bear with me, I'm still confused 😂

> No, I meant the top-level groups would still exist like in the current design, but you simply wouldn't notice that in the nav structure of your screenshot, as all items (and nested groups once we support them) would be part of a single top-level group.

So you mean the single top level group is a technical / internal group rather than a visual one? Because visually speaking the Store, Apps and Marketing groups aren't nested under anything:
image


> Nope, that would be 3 levels of nesting, since the top-level group still exists (as I mentioned at the top of this comment).

Sorry I meant 2 levels of nesting under the parent, so 3 in total as per Dan's screenshot in the subgroups PR:

image

Isn't that the same level of nesting as this? j(ust ignore the vertical line on the left of the Marketing parent group - I know that wouldn't be there in the Filament UI):
image

binaryfire:
Here's a more accurate mockup of what I imagine the nested Marketing group would look like if recreated in Filament:

image

3 levels of nesting just like Dan's screenshot. In which case the only visual difference to a normal group is that there's no extra padding on top/bottom so it looks like a normal item when closed.

zepfietje:
> So you mean the single top level group is a technical / internal group rather than a visual one? Because visually speaking the Store, Apps and Marketing groups aren't nested under anything

Yeah they are nested in that single first-level group if I'm right.

> Isn't that the same level of nesting as this?

No, it's not, because of the first-level group that I'm talking about.

PR #10173, opened 2023-12-08
Show or Hide the tenant menu

kombiahrk:


danharrin:
Please explain the use case here

kombiahrk:
**when Multi tenant features is enabled it will show menu on dashabord** 

if it is only one tenant(team) no need for that menu to switch the tenant and also even when there are two or more tenants(teams) if we dont want to show the tenant menus to user means we can hide 

**with Teanant Menu**
with tenant menu

**without Teanant Menu**
without tenant menu

PR #10176, opened 2023-12-08
Add ability to specify ->delimiter(char) for CSV separator

drbyte:
Ref: [League CSV docs](https://csv.thephpleague.com/9.0/connections/controls/#the-delimiter-character)

Ref: [Filament Discord discussion](https://discord.com/channels/883083792112300104/1181204479140429905/1181204479140429905)

### DIFF
--- a/packages/actions/docs/07-prebuilt-actions/08-import.md
+++ b/packages/actions/docs/07-prebuilt-actions/08-import.md
@@ -495,6 +495,17 @@ ImportAction::make()
 
 If you are encountering memory issues when importing large CSV files, you may wish to reduce the chunk size.
 
+## Changing the field delimiter
+
+The default delimiter for CSV is the comma (`,`).
+If your import will use a different delimiter, you may call the `delimiter()` method on the action. This also works in conjunction with `example()` when writing a file. NOTE: You must specify only a single character, else an exception will be thrown:
+
+```php
+ImportAction::make()
+    ->importer(ProductImporter::class)
+    ->delimiter(';')
+```
+
 ## Customizing the import job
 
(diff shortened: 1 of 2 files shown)


packages/actions/src/Concerns/CanImportRecords.php
```diff
@@ -385,6 +413,11 @@ public function getMaxRows(): ?int
         return $this->evaluate($this->maxRows);
     }
 
+    public function getDelimiter(): ?string
+    {
+        return $this->evaluate($this->csvDelimiter);
```
drbyte, line 418: I had considered including some sort of validation here (ie: if something other than the allowed 1-char string gets set), but decided that letting LeagueCSV throw an exception makes best sense: the dev should discover it in testing since the delimiter is set in code.

danharrin: Agree

packages/actions/docs/07-prebuilt-actions/08-import.md
```diff
@@ -495,6 +495,17 @@ ImportAction::make()
...
+## Changing the field delimiter
+
+The default delimiter for CSV is the comma (`,`).
+If your import will use a different delimiter, you may call the `delimiter()` method on the action. This also works in conjunction with `example()` when writing a file. NOTE: You must specify only a single character, else an exception will be thrown:
+
+```php
+ImportAction::make()
+    ->importer(ProductImporter::class)
+    ->delimiter(';')
+```
```
danharrin:
```suggestion
The default delimiter for CSVs is the comma (`,`).
If your import should use a different delimiter, you may call the `delimiter()` method on the action, passing a new one:

```php
ImportAction::make()
    ->importer(ProductImporter::class)
    ->delimiter(';')
```

You can only specify a single character, otherwise an exception will be thrown.
```

danharrin:
Thanks

PR #10186, opened 2023-12-08
Extending the CanImportRecords trait by delimiter guessing

OccTherapist:
Most of the time, the delimiter in CSV files can be guessed by the ammount of appearance in the file.


danharrin:
Please change this so that the reader is optionally passed to `getCsvDelimiter()`, which also handles the guessing. Instead of checking twice each time.

packages/actions/src/Concerns/CanImportRecords.php
```diff
@@ -341,7 +358,7 @@ public function getUploadedFileStream(TemporaryUploadedFile $file)
         ]));
     }
 
-    public static function getDefaultName(): ?string
+    public static function getfaultName(): ?string
```
danharrin: Also error here

danharrin (changes requested):

OccTherapist:
I edited my commit.

OccTherapist:
Thank you very much for the good refactoring.

PR #10197, opened 2023-12-08
fix require confirmation

AyoobMH:

### More Info
Even when we pass `false` to `requiresConfirmation` a modal still pops up, which is not an expected behaviour, what do you think?

Screenshot 2023-12-09 at 12 38 00 AM


danharrin:
fixed in a different way

PR #10212, opened 2023-12-09
only show DB notifications icon when user has notifications

wivaku:
Previously the DB notifications icon was always shown, even if there are no notifications: icon with badge (0). The icon should only be shown to indicate action is expected from the user.

With this change the notifications icon is only shown if the user has notifications, either read or unread. 

When pressing "mark all as read" the icon remains, with badge (0). The user can still access old notifications.

When pressing "clear" all notifications disappear, as well as the icon (on the next refresh). That's as expected, as there are no notifications to access.

## Description

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/panels/resources/views/components/topbar/index.blade.php
+++ b/packages/panels/resources/views/components/topbar/index.blade.php
@@ -127,7 +127,7 @@
             {{ \Filament\Support\Facades\FilamentView::renderHook('panels::global-search.after') }}
 
             @if (filament()->auth()->check())
-                @if (filament()->hasDatabaseNotifications())
+                @if (filament()->hasDatabaseNotifications() && count(auth()->user()->notifications))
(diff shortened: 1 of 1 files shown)


danharrin:
Hm no, its a breaking change, plus we have a dedicated empty state to tell the user they have no notifications

wivaku:
agreed @danharrin about the breaking change.
What do you mean with the dedicated empty state?
I do believe it makes sense to -have the option to- only show the icon if there are actual notifications.That way the icon draws attention (only) when there are notifications. 

Maybe add flag in panel config? E.g. ->hasDatabaseNotifications(hideWhenNone: true) or ->hideDatabaseNotificationsWhenNone() ?

danharrin:
In the next version `databaseNotifications(fn (): bool => count(auth()->user()->notifications))`

PR #10233, opened 2023-12-11
Correct the sort order for reorderable Spatie Media Library Files when displaying in table or infolist

bomshteyn:
## Description

Currently when using `SpatieMediaLibraryFileUpload::make()->reorderable()` if you have 3 items will get the following sort order indicated by their `order_column`:

1.  ->   3
2. ->   2
3.  ->   1

When using `SpatieMediaLibraryImageColumn::make()->stacked()` I would expect to see the top item from the form first, but its currently using `sortBy` instead of `sortByDesc` resulting in last item being shown first.

## Code style

- [v] `composer cs` command has been run.

## Testing

- [v] Changes have been tested manually 

## Documentation

no new functionality has been added.


danharrin:
The form should save them in the correct order, we should not need to fix the column or infolist. The fix is the wrong way round.

Does `->appendFiles()` on the file upload fix the order saving problem?

bomshteyn:
@danharrin `->appendFiles()` does indeed fix it. Shouldn't that be default though?

danharrin:
It's a workaround for a FilePond bug. I'd rather not litter the codebase with defaults that might not be relevant when Filepond fixes the issue. It is mentioned quite front and center in the `reorderable()` docs: https://filamentphp.com/docs/3.x/forms/fields/file-upload#reordering-files

PR #10238, opened 2023-12-11
Fix tooltip with incorrect property for duration

howdu:
Replace timeout with duration. 

Screenshot 2023-12-11 at 11 06 29


zepfietje:
This breaks tooltips, as they now use the copy message duration for the transition duration. It seems like copy message duration can't be supported by Tippy.js then?

PR #10248, opened 2023-12-11
Correction of the gap property in the navigation group

bzy107:
## Description

I noticed that the spacing between navigation groups, which was fixed in #9387, has reverted.

I reverted it with this fix, but is there a specific reason for this? 
If there is no issue, please merge this pull request and reapply the fix from #9387.

Modified Commits: 
スクリーンショット 2023-12-12 2 28 57

before:
スクリーンショット 2023-12-12 2 26 52

after:
スクリーンショット 2023-12-12 2 26 25

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/panels/resources/views/components/sidebar/index.blade.php
+++ b/packages/panels/resources/views/components/sidebar/index.blade.php
@@ -113,7 +113,7 @@
         @endif
 
         @if (filament()->hasNavigation())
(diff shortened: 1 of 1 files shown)


danharrin:
We had complaints that it was too small, so we reverted it, and introduced "parent navigation items" that have gap-1 spacing

PR #10271, opened 2023-12-12
Dispatch reorder table event

martinmildner:
## Description

This pull request adds the functionality for a "reorder-table" event to be dispatched when reordering a table.

### DIFF
--- a/packages/tables/docs/10-advanced.md
+++ b/packages/tables/docs/10-advanced.md
@@ -160,6 +160,22 @@ public function table(Table $table): Table
 }
 ```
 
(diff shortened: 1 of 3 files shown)


danharrin:
Using events for this sort of thing doesn't really fit the rest of the Filament API. If anything, it'd be a little closure method that gets executed instead.

For this, you can just override `reorderTable()` on the List page / Relation Manager, call `parent::reorderTable()`, and then execute your logic.

danharrin:
Also, registering a listener on a resource definitely wouldnt work

PR #10284, opened 2023-12-13
Fix: Hint icon tooltip

bert-w:
## Description

The tooltips for the hint icons did not work (underused feature maybe? Because it seems like it has been like this for a while).

After:

image

Before: No tooltip :)

## Code style

## Testing

## Documentation


danharrin:
Underused on the placeholder component probably, but it works elsewhere

PR #10288, opened 2023-12-13
Add and revise missing Japanese translations

bzy107:
Add Japanese missing translations and revise Japanese translations


bzy107:
I will cancel the pull request because it includes unexpected file modifications.

PR #10289, opened 2023-12-13
Panels: Add new Panel::route() helper for generating routes

ryangjchandler:
## Description

You can register routes for a panel using the `$panel->routes(function () { ... })` method, but generating route URIs is tedious because the names are prefixed with `filament.{$panel->getId()}`. 

This new `$panel->route(...)` helper has the same signature as Laravel's own `route()` helper but makes it easier to generate the URLs themselves.

## Code style

## Testing

## Documentation


danharrin:
@ryangjchandler just one last thing before merge, can you go through and refactor all appropriate `route()` references to use this please?

ryangjchandler:
@danharrin I've updated the appropriate places to use the new helper. I've only done the ones inside of `Panel` related methods since we have access to the panel itself and there's no risk of breaking changes.

PR #10297, opened 2023-12-14
Add CSS classes to sidebar based on label

ralphjsmit:
## Description

I was looking for a way to apply some custom CSS to a certain item in the navigation to make it stand out more (an onboarding page). I noticed there wasn't really a good way to target a specific navigation item using CSS classes.

This PR automatically sluggifies the navigation item label and adds that as a class. It will generate classes like this:

```
fi-sidebar-item-onboarding
fi-sidebar-item-tasks
...
```

Thanks!


danharrin:
I don't think we should merge this. The classes would then be changing based on the current locale, in translated apps.

ralphjsmit:
That's true indeed, would you be open to adding an `extraAttributes()` method?

danharrin:
Sure

PR #10327, opened 2023-12-15
Fix modal cancel action

mariusticha:
### current behaviour

I'm writing an Action and I wanted the cancel button to be shown conditionally: 

```php
Action::make('delete')
    ->form([
        TextInput::make('input')
    ])
    ->modalCancelAction(fn () => $condition)
    ->action(fn () => null);
```

if `$condition` is `false` everything works as expected and the submit button is hidden, but if `$condition` is `true` i get this error:

> Filament\Actions\MountableAction::getModalCancelAction(): Return value must be of type ?Filament\Actions\StaticAction, true returned

### expected behaviour

I expected the method to return the default cancel action.

--- 

--> no visual changes

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/actions/src/Concerns/CanOpenModal.php
+++ b/packages/actions/src/Concerns/CanOpenModal.php
@@ -436,6 +436,10 @@ trait CanOpenModal
                 default => $color,
             });
 
+        if ($this->evaluate($this->modalSubmitAction, ['action' => $action]) === true) {
+            return $action;
+        }
+
         if ($this->modalSubmitAction !== null) {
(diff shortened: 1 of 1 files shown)


danharrin:
`->modalCancelAction(fn ($action) => $condition ? $action : null)`

danharrin:
or `->modalCancelAction(fn ($action) => $action->visible($condition))`

mariusticha:
thanks for your answers, i already wrote a similar solution in my current project.

i've just wondered that a method that allows an input of type bool crashes when i pass `true`.

PR #10340, opened 2023-12-16
Improvement of translation management

AlexisSerneels:
This Pull Request improves the current logic of `filamentphp/spatie-laravel-translatable-plugin` in three ways:

1. Currently, the fallback value is returned for missing translations. Additionally, this fallback value is saved as a translation during the update process. This results as having the model translated in all languages, as it was purposefully translated. But it's not. When I save my entry with French selected via the language selector, I expect to only edit the French version, not to inject the fallback translation into all other missing translations.
This current behavior also prevents the development of a "missing translations" system, since the fallback value is saved.  

2. Furthermore, the form is validated when changing languages via the language selector. Again, wanting to change languages does not mean "wanting to save." I might want to check what translations I have in Italian to implement them in Spanish, for example. Currently, this is not possible as Filament attempts to validate upon language change.  

3.  Considering the above two points, I believe the language selector should only be an action for **displaying** the form in the desired language, and only impacting the selected language.

This logic is implemented in Statamic CMS and many others.

With this PR:
- The field is no longer populated with the fallback value. If there is no translation, the field will remain empty, as it should.
- Validation is no longer executed during a language change. It will always occur upon form submission though.

### DIFF
--- a/packages/spatie-laravel-translatable-plugin/src/Resources/Pages/Concerns/HasTranslatableFormWithExistingRecordData.php
+++ b/packages/spatie-laravel-translatable-plugin/src/Resources/Pages/Concerns/HasTranslatableFormWithExistingRecordData.php
@@ -20,7 +20,7 @@ trait HasTranslatableFormWithExistingRecordData
             $translatedData = [];
 
             foreach ($translatableAttributes as $attribute) {
-                $translatedData[$attribute] = $record->getTranslation($attribute, $locale, false);
+                $translatedData[$attribute] = $record->getTranslation($attribute, $locale, useFallbackLocale: false);
             }
 
             if ($locale !== $this->activeLocale) {
--- a/packages/spatie-laravel-translatable-plugin/src/Resources/Pages/Concerns/HasTranslatableFormWithExistingRecordData.php
+++ b/packages/spatie-laravel-translatable-plugin/src/Resources/Pages/Concerns/HasTranslatableFormWithExistingRecordData.php
@@ -20,7 +20,7 @@ trait HasTranslatableFormWithExistingRecordData
             $translatedData = [];
 
             foreach ($translatableAttributes as $attribute) {
-                $translatedData[$attribute] = $record->getTranslation($attribute, $locale);
+                $translatedData[$attribute] = $record->getTranslation($attribute, $locale, false);
             }
 
             if ($locale !== $this->activeLocale) {
--- a/packages/spatie-laravel-translatable-plugin/src/Resources/Pages/CreateRecord/Concerns/Translatable.php
+++ b/packages/spatie-laravel-translatable-plugin/src/Resources/Pages/CreateRecord/Concerns/Translatable.php
@@ -78,6 +78,8 @@ trait Translatable
             return;
         }
 
+        $this->resetValidation();
+
         $translatableAttributes = static::getResource()::getTranslatableAttributes();
 
         $this->otherLocaleData[$this->oldActiveLocale] = Arr::only($this->data, $translatableAttributes);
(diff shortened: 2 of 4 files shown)


danharrin:
The fallback value fix is valid, thanks for that.

As for the validation, it does need happen at some point, otherwise invalid locales will not be saved at all and changes could be lost. I've just made some changes, so that any invalid locales that already exist in the database show validation errors on save if they fail. Can you give this PR a good testing to make sure this doesn't result in unexpected behaviour or break anything?

atmonshi:
thank you @AlexisSerneels 
did some testing following this flow:

edit a page
switch lang
add title
saving dont do anything
switching to the next lang, add title, save wont trigger
unless I switch to the first lang and save it works

I think it expected from the user that is save the form no matter which active lang

(follow up): saving does not work at all from other lang only from the first one

https://github.com/filamentphp/filament/assets/1952412/53b9b75f-f702-4fb5-9b91-ea6b2347109b

AlexisSerneels:
@danharrin You are right about the need for validation; I had missed that. Your implementation works great!

I took the liberty of re-adding the `resetValidation()`on `updatedActiveLocale()` so that error messages for one language would no longer be displayed for another, which one might want to leave the translation empty, for example.

@atmonshi I agree with your report. I think it refers to my first version which had the validation removed. @danharrin has fixed it. Can I ask you to pull the latest version of the PR and retest?

AlexisSerneels:
I can open another PR if needed, but there is one thing that might need improvement:

- The active locale is not persisted, so a page refresh reverts it back to English. Moreover, if I have a language selected on the `ViewPage`, that active locale is not carried over to the `EditPage` and the user has to select the correct locale again.

Might it be possible to use $persist or the session to store the active locale?

danharrin:
Hm, I dont understand in what situation resetting the validation would make a difference?

danharrin:
We can just put the active locale in the URL for now, thats the easiest approach

AlexisSerneels:
@danharrin 
> Hm, I dont understand in what situation resetting the validation would make a difference?

Here is a video of the current situation. You can see the error message is wrongfully persisted.

https://github.com/filamentphp/filament/assets/287688/236de3d6-891c-46c8-89a9-3983f5b46bce

And here is with the validation reset, which clears the message.

https://github.com/filamentphp/filament/assets/287688/6ea57378-e353-464e-bebb-ef478776b7bd

atmonshi:
the edit situation is fixed, however, the creation have the same issue?!!
click new
change lang
click save

unless I fill the first lang
I think this related to resetting the validation as Alexis mentioned

https://github.com/filamentphp/filament/assets/1952412/6a48f576-81d4-4ab1-af7f-62dbfabe6ca4



> The active locale is not persisted

I agree with dan, a url parameter it better than session, when the user can open multiple taps and will be hard to keep track them and end up storing so much data

AlexisSerneels:
@atmonshi I could not reproduce your issue though. Please see below.

https://github.com/filamentphp/filament/assets/287688/4a6acbe9-a61d-489b-8220-0a49d6c8fd4f

Can you manually comment this line and retry if it work without?: https://github.com/filamentphp/filament/pull/10340/commits/289ff3f493adbbd3464de1457944d4aa13e487cc

And what is your fallback locale in `app.php`?

Anyway, I'm willing not push the "validation" changes if it's too sensitive. I'll recheck and open a new PR if needed later.
The main issue was the fallback values being saved. This one was fixed ;)

atmonshi:
ya I found out the reason, I have 
```
public $translatable = ['title', 'subtitle'];
```
but the `subtitle` is hidden input (I hide it for testing)
sorry for that :)

now everything working as expected thank you again :)

AlexisSerneels:
> now everything working as expected thank you again :)

Glad you could find the unrelated reason!

@danharrin I let you decide regarding: https://github.com/filamentphp/filament/pull/10340#issuecomment-1859221796
All works for me and my tests.

danharrin:
Thanks

PR #10368, opened 2023-12-18
Fix to enable `data-` attributes to be parsed correctly

titantwentyone:
## Description

Currently the following results in two attributes being output on the target element. One being incorrect:
```
TextInput('something')
   ->extraAttributes(['data-foo' => 'bar']);

//Results in...
<div...
   datafoo='bar'
   data-foo='bar'
>
```

PR includes tests for basic html attributes, alpine attributes and data attributes.

Before PR:
image

After PR:
image

### DIFF
--- a/tests/src/Support/helpersTest.php
+++ b/tests/src/Support/helpersTest.php
@@ -16,7 +16,7 @@ it('will prepare attributes', function () {
     ]);
 });
 
-it('will prepare alpine attributes', function () {
+it('will prepare Alpine attributes', function () {
     $bag = new ComponentAttributeBag([
         'x-data' => '{foo:bar}',
     ]);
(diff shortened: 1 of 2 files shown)


tests/src/Support/helpersTest.php
```diff
@@ -0,0 +1,41 @@
...
+    ]);
+
+    $attributes = prepare_inherited_attributes($bag);
+
+    expect($attributes->getAttributes())->toBe([
+        'style' => 'color:red',
+    ]);
+});
+
+it('will prepare alpine attributes', function () {
```
danharrin:
```suggestion
it('will prepare Alpine attributes', function () {
```

danharrin:
Thanks

PR #10372, opened 2023-12-18
Allow individual logout for each of the panels.

soymgomez:
## Description

Since version 3.x allows to manage several panels from the same installation, the lines to destroy the user session when logging out do not make sense.

Because when logging out from one of the panels is done automatically from the rest.

## Code style

## Testing

## Documentation


danharrin:
I think the reason why the session gets invalidated is so that someone cannot use that cookie to log in again after the user logs out. Laravel does recommend it in their documentation.

If you know a method of regenerating the session only after all authentication guards are logged out, feel free to PR that. Or, if you know a way of regenerating the session and keeping the other auth guards logged in.

PR #10408, opened 2023-12-20
Allow individual logout for each of the panels v2

soymgomez:
## Description

Since version 3.x allows to manage several panels from the same installation, the lines to destroy the user session when logging out do not make sense.

Because when logging out from one of the panels is done automatically from the rest.

This PR is related with #10372 with a logic to detect if the user has more than one logedin panels

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/panels/src/Http/Controllers/Auth/LogoutController.php
+++ b/packages/panels/src/Http/Controllers/Auth/LogoutController.php
@@ -11,13 +11,14 @@ class LogoutController
     {
         Filament::auth()->logout();
 
-        $count = collect(session()->all())->filter(function ($value, $key) {
-            return strpos($key, 'login_') === 0;
-        })->count();
+        $logins = array_filter(
+            array_keys(session()->all()),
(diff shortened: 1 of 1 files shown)


danharrin:
I can't seem to reproduce a scenario where `login_` items are added to the session, how does this work?

soymgomez:
> I can't seem to reproduce a scenario where `login_` items are added to the session, how does this work?

It happened to us in a new installation of filament adding a second panel(https://filamentphp.com/docs/3.x/panels/configuration#creating-a-new-panel) and setting a diferent guard. 

With this one you can login in both panels at the same time but logging out in one of them destroys the whole session.

tonypartridge:
Are you adding a custom session driver / middleware for the login_ session key?

danharrin (changes requested):
Please can you send a simple reproduction repository that works with this setup, so I can test it?

danharrin:
Just tested this locally and it doesn't seem to work at all 🤷‍♂️

PR #10433, opened 2023-12-22
Add fix-overflowing-images support to Table ImageColumn

drbyte:
## Description

Extends #10415 "Fix overflowing images" on infolists (from Filament v3.1.25) by (optionally) adding `flex-wrap` to table image column as well.

While not enabled by default, this PR adds the `->wrap(true)` option to enable the wrapping.

When a table is displayed as a normal table, one will probably want to use limit() instead of wrapping, else images will stack vertically. However, when a table is displayed in a grid, using `wrap(true)` will pleasantly display the images, instead of overflowing out of bounds.

Doesn't interfere with existing `stacked`/`overlay`/`limit`/etc features.

**Grid Table:**

Before:
Screen Shot 2023-12-21 at 1 39 37 PM

After adding `wrap()`:
Screen Shot 2023-12-21 at 1 39 15 PM

**Normal Table:**
Before:
Screen Shot 2023-12-21 at 1 54 27 PM

After adding `wrap()` (obviously not everyone will want multiple lines, so they can still use `limit()` instead)
Screen Shot 2023-12-21 at 1 55 10 PM

Caveat:
If one adds `wrap()` to a normal row-based table, the images will stack:
Screen Shot 2023-12-21 at 3 09 52 PM

## Code style

- [x] Make sure code style follows the rest of the codebase.

## Testing

## Documentation

- [x] Documentation updates included

### DIFF
--- a/packages/tables/docs/03-columns/03-icon.md
+++ b/packages/tables/docs/03-columns/03-icon.md
@@ -99,3 +99,17 @@ IconColumn::make('is_featured')
 ```
 
 <AutoScreenshot name="tables/columns/icon/boolean-color" alt="Icon column to display a boolean with custom colors" version="3.x" />
+
+## Wrapping the icons
+
+When displaying multiple icons, they can be set to wrap if they can't fit on one line, using `wrap()`:
+
+```php
+use Filament\Tables\Columns\IconColumn;
+
+IconColumn::make('icon')
+    ->wrap()
+```
+
+Note: the "width" for wrapping is affected by the column label, so you may need to use a shorter or hidden label to wrap more tightly.
+
--- a/packages/tables/docs/03-columns/03-icon.md
+++ b/packages/tables/docs/03-columns/03-icon.md
@@ -100,7 +100,7 @@ IconColumn::make('is_featured')
 
 <AutoScreenshot name="tables/columns/icon/boolean-color" alt="Icon column to display a boolean with custom colors" version="3.x" />
 
-## Wrapping the icons
+## Wrapping multiple icons
 
 When displaying multiple icons, they can be set to wrap if they can't fit on one line, using `wrap()`:
 
--- a/packages/tables/docs/03-columns/04-image.md
+++ b/packages/tables/docs/03-columns/04-image.md
@@ -131,7 +131,7 @@ ImageColumn::make('colleagues.avatar')
 
 ## Wrapping the images
 
-When displaying the table as a Grid, images can be set to wrap if they can't fit on one line, using `wrap()`:
+Images can be set to wrap if they can't fit on one line, by setting `wrap()`:
 
 ```php
 use Filament\Tables\Columns\ImageColumn;
--- a/packages/tables/docs/03-columns/04-image.md
+++ b/packages/tables/docs/03-columns/04-image.md
@@ -129,7 +129,7 @@ ImageColumn::make('colleagues.avatar')
     ->overlap(2)
 ```
 
-## Wrapping the images
+## Wrapping multiple images
 
 Images can be set to wrap if they can't fit on one line, by setting `wrap()`:
 
--- a/packages/tables/docs/03-columns/04-image.md
+++ b/packages/tables/docs/03-columns/04-image.md
@@ -129,6 +129,19 @@ ImageColumn::make('colleagues.avatar')
     ->overlap(2)
 ```
 
+## Wrapping the images
+
+When displaying the table as a Grid, images can be set to wrap if they can't fit on one line, using `wrap()`:
+
+```php
+use Filament\Tables\Columns\ImageColumn;
+
+ImageColumn::make('colleagues.avatar')
+    ->circular()
+    ->stacked()
+    ->wrap()
+```
+
 ## Setting a limit
 
 You may limit the maximum number of images you want to display by passing `limit()`:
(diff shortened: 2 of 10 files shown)


packages/tables/resources/views/columns/image-column.blade.php
```diff
@@ -55,6 +56,10 @@
             <div
                 @class([
                     'flex',
+                    match ($wrap) {
+                        true => 'flex-wrap',
+                        default => '',
+                    },
```
danharrin:
```suggestion
                    'flex-wrap' => $canWrap,
```

danharrin (changes requested):
Please extract `wrap()` / `$canWrap()` / `canWrap()` from TextColumn into a `CanWrap` trait, then use across both classes.

packages/tables/resources/views/columns/image-column.blade.php
```diff
@@ -7,6 +7,7 @@
     $isStacked = $isStacked();
     $overlap = $isStacked ? ($getOverlap() ?? 2) : null;
     $ring = $isStacked ? ($getRing() ?? 2) : null;
+    $wrap = $getWrap();
```
danharrin:
```suggestion
    $canWrap = $canWrap();
```

danharrin (commented):
Maybe it's worth adding `wrap()` to color/icon column too?

drbyte:
> Maybe it's worth adding `wrap()` to color/icon column too?

At first glance, they seem to already have `flex-wrap` on them:
https://github.com/filamentphp/filament/blob/3.x/packages/tables/resources/views/columns/color-column.blade.php#L6

I guess we could make it optional ...

drbyte:
> Please extract `wrap()` / `$canWrap()` / `canWrap()` from TextColumn into a `CanWrap` trait, then use across both classes.

Agreed. Working on it ...

drbyte:
~One problem I foresee: the default for wrapping is `false` ... but in all cases except the non-Grid table it was previously hardcoded to `true`.~

~To change the default to `true` is possible .... I guess then we'd just change the documentation to say "set `wrap(false)` to disable it if undesired"?~

For Tables, leaving `wrap` set to `false` by default makes sense.

drbyte:
Icon/Color columns with `wrap()`:
Screen Shot 2023-12-24 at 12 42 53 AM

Icon/Color columns with `wrap(false)` (which is the default):
Screen Shot 2023-12-24 at 12 44 02 AM

(Screenshot note: the quantity and values of icons and colors were randomly generated, which is why they differ between screenshots, due to screen refresh. The layout is what's intended to be portrayed.)

Observation, pre-existent before this PR: When the Icon or Colour column has a title that's wide enough to permit multiple items, if `wrap()` is enabled, it will wrap based on that width, so for "perfect" wrapping which keeps only one icon or color per row, one must use a very short or blank column name/title.
Note added to docs about this.

drbyte:
**EDIT: This part has been removed:**

Infolist Icon/Color with `wrap(false)`:
Screen Shot 2023-12-24 at 1 26 03 AM

with `wrap(true)` (which is also the default):
Screen Shot 2023-12-24 at 1 27 25 AM

danharrin (changes requested):
Thanks for your work on this! Sorry to be a pain, but I don't think we should actually have a configurable `wrap(false)` for infolists. It does really break the design and no-one should ever want to really use that. Can you remove that option for infolists please? The rest looks great, thanks.

drbyte:
> Sorry to be a pain, but I don't think we should actually have a configurable wrap(false) for infolists. It does really break the design and no-one should ever want to really use that.

Yes, I agree. The more I've thought about it, the more icky it feels!

Reverted.

packages/tables/docs/03-columns/04-image.md
```diff
@@ -129,6 +129,19 @@ ImageColumn::make('colleagues.avatar')
...
+
+```php
+use Filament\Tables\Columns\ImageColumn;
+
+ImageColumn::make('colleagues.avatar')
+    ->circular()
+    ->stacked()
+    ->wrap()
+```
+
```
danharrin, line 144:
```suggestion

Note: the "width" for wrapping is affected by the column label, so you may need to use a shorter or hidden label to wrap more tightly.

```

packages/tables/docs/03-columns/04-image.md
```diff
@@ -129,6 +129,21 @@ ImageColumn::make('colleagues.avatar')
     ->overlap(2)
 ```
 
+## Wrapping the images
+
+When displaying the table as a Grid, images can be set to wrap if they can't fit on one line, using `wrap()`:
```
danharrin:
```suggestion
Images can be set to wrap if they can't fit on one line, by setting `wrap()`:
```

packages/tables/docs/03-columns/03-icon.md
```diff
@@ -99,3 +99,17 @@ IconColumn::make('is_featured')
 ```
 
 <AutoScreenshot name="tables/columns/icon/boolean-color" alt="Icon column to display a boolean with custom colors" version="3.x" />
+
+## Wrapping the icons
```
danharrin:
```suggestion
## Wrapping multiple icons
```

packages/tables/docs/03-columns/04-image.md
```diff
@@ -129,6 +129,21 @@ ImageColumn::make('colleagues.avatar')
     ->overlap(2)
 ```
 
+## Wrapping the images
```
danharrin:
```suggestion
## Wrapping multiple images
```

danharrin:
Thanks for your contributions and fast iteration @drbyte!

PR #10434, opened 2023-12-22
Utility injection in CanBeValidated@rules

Carnicero90:
## Description

This pr aims to allow utility injection in CanBeValidated@rules, since I need it for a couple of edge cases (I need to make the same db query for a bunch of custom rules, so that by using utility injection on every single rule, that would be repeated). It is working well in my tests, if it pass your further scrutiny let me know if I should somehow mention this feature in the docs as well ([here](https://filamentphp.com/docs/3.x/forms/validation#custom-rules)).
Thanks.

## Code style

- [x ] `composer cs` command has been run.

## Testing

- [ x] Changes have been tested.

## Documentation

### DIFF
--- a/packages/forms/src/Components/Concerns/CanBeValidated.php
+++ b/packages/forms/src/Components/Concerns/CanBeValidated.php
@@ -420,7 +420,7 @@ trait CanBeValidated
     /**
      * @param  string | array<mixed>  $rules
      */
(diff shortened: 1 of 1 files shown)


danharrin:
This would not work for many utilities, as it's too early in the lifecycle. We do avoid evaluating in setter methods.

You can pass multiple closures to an array already, for evaluation later: `[fn ($get) => ...]`

PR #10451, opened 2023-12-23
Document guidance for guest-access in a panel

drbyte:
## Description

This is a frequently-asked topic. I posted a similar answer on Discord, so contributing it for consideration as part of core docs.

### DIFF
--- a/packages/panels/docs/08-users.md
+++ b/packages/panels/docs/08-users.md
@@ -264,3 +264,26 @@ public function panel(Panel $panel): Panel
         ->authPasswordBroker('users');
 }
 ```
+
+## Guest Access
+
+By default Filament expects to work with Authenticated Users.
+
+To work with Guest users, you need to avoid using components which expect a User (such as profiles, avatars), and skip the built-in Authentication middleware.
+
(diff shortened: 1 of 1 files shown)


packages/panels/docs/08-users.md
```diff
@@ -264,3 +264,15 @@ public function panel(Panel $panel): Panel
...
 }
 ```
+
+## Setting up guest access to a panel
+
+By default, Filament expects to work with authenticated users only. To allow guests to access a panel, you need to avoid using components which expect a signed-in user (such as profiles, avatars), and remove the built-in Authentication middleware:
+
+- Remove the default `Authenticate::class` from the `authMiddleware()` array in the panel configuration.
+- Remove `->login()` and any other [authentication features](#authentication-features) from the panel.
+- Remove the default `AccountWidget` from the `widgets()` array, because reads the current user's data.
```
drbyte, line 274: @danharrin should "because reads the current" have the word "it" inserted? ie: "because it..."

danharrin: sorry, fixing

PR #10452, opened 2023-12-23
Crosslink docs for Table Widgets

drbyte:
## Description

I find myself repeatedly clicking on the Widgets package documentation looking for Widget documentation for Pages, Resources, Tables, Dashboards ... and then I have to go hunting. Cross-linking them will hopefully make it less painful for others who are looking for the same.

(Sorry for the filename renumbering, but it felt really strange having `02-Stats`, `03-Charts`, `04-Adding A Widget To A Blade View`, and then `05-Tables` ... so I renumbered 04 to 05.)


danharrin:
I know this is annoying, but there's actually a reason we don't have a page for this. It's because we don't want to encourage users to use the table widget if they are not using the panel builder. Since at that point, they can just use a normal Livewire component that renders a table. Table widgets are a little buried into the panel builder docs for this reason, they don't really have any unique features that don't exist in the table builder

drbyte:
Okay, yes, I can appreciate that.

But then ... why not just "say that" in the docs directly, so the intention is clear.

And perhaps less verbose than my proposal, a simple link saying "Are you looking for Panel Widgets instead? See [link]"

Again, my rationale: the main nav in the Filament Docs has a nice big "Widgets" button ... which has nothing to do with the widgets one is using in Panels, which is what Filament's focus is about.  So, the disconnect just brings confusion IMO.

Sorry, I'm not trying to belabour the point. I accept whatever you decide to do with it.

danharrin:
Ok fair, I've added a note here - 225e02377d29fa1fd728fa9c933951f4c2202fdd

PR #10478, opened 2023-12-24
Center text on logo header

thethunderturner:
## Description

This PR adds a class to the child of the sidebar header, where the logo is. Initially the logo is off center, meaning there is more space below the logo, than above it. This fixes it

https://github.com/filamentphp/filament/assets/64212185/8ce21d1b-0797-4020-a8cf-e31c2c1c3bdf

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/panels/resources/views/components/sidebar/index.blade.php
+++ b/packages/panels/resources/views/components/sidebar/index.blade.php
@@ -44,6 +44,7 @@
             class="fi-sidebar-header flex h-16 items-center bg-white px-6 ring-1 ring-gray-950/5 dark:bg-gray-900 dark:ring-white/10 lg:shadow-sm"
         >
             <div
+                class="flex"
                 @if (filament()->isSidebarCollapsibleOnDesktop())
                     x-show="$store.sidebar.isOpen"
                     x-transition:enter="lg:transition lg:delay-100"


zepfietje:
@thethunderturner, could you share some more details about the cause of the spacing issue?

thethunderturner:
> @thethunderturner, could you share some more details about the cause of the spacing issue?

Hello! Im not so sure what is going on but I think its an issue of the parent div. Atm Im using a custom view to load the brand logo:

```php
<div class="flex justify-center items-center gap-2">
    <div>
        <a href="https://example.com/" target="_blank">
            <img class="hover:scale-125 h-3rem fi-logo transition duration-300 ease-in-out" src="{{ asset('logo/logo.svg') }}" alt="Test packages">
        </a>
    </div>
    <div class="text-xl font-bold text-gray-950 dark:text-white">
        Example Text
    </div>
</div>
```

As you can see from the video, the logo isnt really centered because the div sorounding it (the view in this case) is not centered in the header. This is from my understanding.

zepfietje:
Could you please check using the inspector?

zepfietje:
The logo is nested in an `<a>` tag, so you can't have another link inside there. Have a look at the inspector and you'll see that in the DOM.

PR #10488, opened 2023-12-25
Unique over the records of current tenant only

bdsumon4u:
## Description

_What if I want a record to be unique only over the records of the current tenant, not across the entire table?_

image

Here, __name__ is unique within the current __business__ (**tenant**), but different __business__ can have the same __name__.

## Code style

## Testing

- [x] Changes have been visually tested.

## Documentation


danharrin:
The form builder can be installed outside of the panel builder. Having a feature like this would cause issues for people who use it like that.

danharrin:
This can be implemented in userland using the `modifyRuleUsing` parameter of `unique()`

PR #10512, opened 2023-12-26
- Remove inlineLabel() call from inline() in Radio; Issue #9994

AbdullahFaqeir:
## Description

- Remove inlineLabel() call from inline() in Radio; Issue #9994


danharrin:
This is a v4 issue since it is a breaking change.

AbdullahFaqeir:
> This is a v4 issue since it is a breaking change.

Should I PR it to the v4 branch?

danharrin:
It's ok, we can handle it ourselves when the time comes, thanks :)

PR #10516, opened 2023-12-26
feature: Unsaved changes alert

### LINKED ISSUE #9992: Wait, don't leave!
Alert the user if they have unsaved changes in a form.

danharrin:
Closes #9992.


pxlrbt:
I like the simplicity of the solution. But it might affect loading time especially on larger forms when we send duplicate data, right?

Should probably be opt-in then?

pxlrbt:
Another idea: Instead of sending the full data, couldn't we generate a md5 hash or similar and just compare the hash? Would be significantly less data for large forms.

danharrin:
If you want to work on the MD5 hash version you can, but this is a good first pass I think

binaryfire:
@danharrin Have you considered using local storage to track the dirty state? That's how I do it for my pages. It's more performant and works with SPA mode. I have this at the bottom of my views:

```
<script>
    document.addEventListener('DOMContentLoaded', function () {
        localStorage.removeItem('has-unsaved-filament-form');
    });
</script>

@script
    <script>
        $wire.on('form-updated', function () {
            localStorage.setItem('has-unsaved-filament-form', 'true');
        });

        $wire.on('form-saved', function () {
            localStorage.removeItem('has-unsaved-filament-form');
        });        

        window.addEventListener('beforeunload', function (e) {
            if (localStorage.getItem('has-unsaved-filament-form')) {
                e.preventDefault();
                e.returnValue = '';
            }
        });      
    </script>
@endscript
```

danharrin:
Having to dispatch a browser event (from the backend?) after every field is changed would be refreshing the component tons, right?

binaryfire:
@danharrin Yeah sorry, that particular example was from a builder (fully live) page. But the general idea of using local storage could still work. Perhaps you could store a hash of the initial data in LS and compare that to the new hash on the way out?

danharrin:
Is your concern that the old data gets sent back and forth with each request?

binaryfire:
@danharrin That's part of it, but that wouldn't be an issue if it was a hash as @pxlrbt suggested. My bigger concern is it not working with SPA mode. That's a fairly major issue IMHO. If there's an alternative solution which lets it work in both modes I think it's worth exploring.

danharrin:
I do not understand how your suggestion supports SPA mode? The issue is that when navigating, you cannot cancel the request yet

binaryfire:
Oh right, the issue is with using `beforeunload`. Gotcha. Livewire has a `wire:navigating` event that fires _before_ pages are navigated away from: https://livewire.laravel.com/docs/navigate#javascript-hooks. Couldn't that serve the same purpose for SPA mode?

danharrin:
Nope, because even if you catch the event, there is currently no way to stop the navigation from happening

danharrin:
@pxlrbt @binaryfire I've implemented an MD5 hash solution in a56997b423d0db4ae5f1a1b3ff99f5e61fb02fa9

bilogic:
> Nope, because even if you catch the event, there is currently no way to stop the navigation from happening

I'm curious if LW is aware of this limitation? I think it could be worth informing them since reminding a user to save can be important in some cases.

binaryfire:
Hi @bilogic. This has been solved and PR'd: https://github.com/filamentphp/filament/pull/12363

bilogic:
ok thanks!

PR #10532, opened 2023-12-28
Missing mount function in the infolist example

snarexx:
There is a missing mount function in the example. It took me an hour to find the solution, and I don't want others to waste their time on this.

## Description

## Code style

## Testing

## Documentation


danharrin:
You do not need a `mount()` function, since Livewire will bind the property using its name from the route / component props. We do assume that people have used Livewire before trying to add infolists/forms/tables to their components.

PR #10554, opened 2023-12-29
Fix refreshFormData with Enum

fouteox:
## Description

If you use refreshFormData on values that are cast as Enums in the Model, the enum will be loaded instead of its value, which will cause a problem when submitting the form.

My PR fixes this problem by iterating through the values to refresh and taking ->value in the case of an Enum.

The solution is certainly not perfect for Filament's framework.

I'll let the experts correct this properly :)

## Code style

## Testing

See my previous comment for this.

### DIFF
--- a/packages/panels/src/Resources/Pages/EditRecord.php
+++ b/packages/panels/src/Resources/Pages/EditRecord.php
@@ -111,7 +111,7 @@ class EditRecord extends Page
         $updatedData = $this->getRecord()->only($attributes);
 
         $transformedData = array_map(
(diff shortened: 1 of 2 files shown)


danharrin:
When the form is first loaded, the value is an enum instead of a string. Right?

fouteox:
Nope, the first time the component is loaded with the form, it is indeed a string, even if the field is of type Enum.

The issue only occurs when using the refreshFormData function with a casted Enum field.

danharrin:
Thanks

PR #10567, opened 2023-12-30
Fix chartWidget property must not be accessed error

archilex:
Using a chartWidget with `InteractsWithPageTable` throws a `Typed property Filament\Widgets\ChartWidget::$dataChecksum must not be accessed before initialization` error. 

Setting a default value to the property (seems to) fix this (This is one of those livewire things that I don't fully understand). 

Another option that works is:

`public ?string $dataChecksum = null;`

### DIFF
--- a/packages/widgets/src/ChartWidget.php
+++ b/packages/widgets/src/ChartWidget.php
@@ -16,7 +16,7 @@ abstract class ChartWidget extends Widget
     protected ?array $cachedData = null;
 
     #[Locked]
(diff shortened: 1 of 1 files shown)


CodeWithDennis:
Im not sure if related but a similar error also happens when you have a lot of widgets on a dashboard. @danharrin @archilex 

```
Typed property Filament\Widgets\TableWidget::$table must not be accessed before initialization
```

PR #10577, opened 2023-12-30
Update 07-enums.md

malle-pietje:
Remove unneeded return type array

## Description

Removes the IMHO unneeded return type array from the enum example.

## Code style

N.a.

## Testing

N.a.

## Documentation

N.a.


malle-pietje:
Minor but can confuse a new PHP developer

danharrin:
I'm not going to do this, as technically an array can be returned if you want a custom palette. Could confuse people who know what they're trying to do 😁

malle-pietje:
Okay, understood. Hadn’t yet seen that use case.

PR #10587, opened 2023-12-31
Support tags color in `TagsInput`

WilgnerFSDev:
## Description

This PR introduces two new feature for TagsInput component, the `->color()` method (HasColor Trait) to change the color of the tags/badges and the method `->inlineInput()` to make the input inline, after the badges/tags.

**Before:**
tagsinput_before

**After:**
tagsinput_after

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/forms/resources/views/components/tags-input.blade.php
+++ b/packages/forms/resources/views/components/tags-input.blade.php
@@ -6,6 +6,7 @@
     $isDisabled = $isDisabled();
     $isReorderable = $isReorderable();
     $statePath = $getStatePath();
+    $isInlineInput = $getInlineInput();
 @endphp
 
 <x-dynamic-component
--- a/packages/forms/resources/views/components/tags-input.blade.php
+++ b/packages/forms/resources/views/components/tags-input.blade.php
@@ -47,7 +47,7 @@
             x-ignore
             {{ $getExtraAlpineAttributeBag() }}
         >
-            @if(!$isInlineInput)
+            @if (! $isInlineInput)
                 <x-filament::input
                     autocomplete="off"
                     :autofocus="$isAutofocused()"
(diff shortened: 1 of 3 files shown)


danharrin:
I'm leaning towards saying no to the inline method, but I'll let Zep have the final call

packages/forms/resources/views/components/tags-input.blade.php
```diff
@@ -97,6 +104,7 @@ class="hidden"
                                     @class([
                                         'cursor-move' => $isReorderable,
                                     ])
+                                    :color="$color"
```
zepfietje: Move this up to be the first attribute.

packages/forms/resources/views/components/tags-input.blade.php
```diff
@@ -6,6 +6,8 @@
     $isDisabled = $isDisabled();
     $isReorderable = $isReorderable();
     $statePath = $getStatePath();
+    $isInlineInput = $getInlineInput();
+    $color = $getColor();
```
zepfietje: Sort this alphabetically.

zepfietje (changes requested):
Thanks for this PR, @WilgnerFSDev!

Could you revert the "inline" changes?
I think there's more edge cases to it that need to be handled, so let's hold off on that part.

@danharrin, regarding the color API. Wouldn't it be better to support per-tag colors as we do on text columns and entries?

danharrin:
The tags are stored in Alpine on the frontend so we should not support per-tag colors ATM.

WilgnerFSDev:
> Thanks for this PR, @WilgnerFSDev!
> 
> Could you revert the "inline" changes? I think there's more edge cases to it that need to be handled, so let's hold off on that part.
> 
> @danharrin, regarding the color API. Wouldn't it be better to support per-tag colors as we do on text columns and entries?

Sure, is done.
So for now, I will create a custom field to meet the client's requirement because the TagsInput (default) in the repeater takes up too much space.

Thank you for your time @danharrin and @zepfietje

PR #10590, opened 2024-01-01
Fix: standardizes pt_br strings

fcno:
## Description

In some places the term 'Visualizar' is used and in others 'Mostrar'. Both are synonymous, but in Brazilian Portuguese, the term 'Visualizar' makes more sense, or at least, is more used in web applications.

It would be like using View and Show in English. There is no completely wrong choice, but if you choose a one, you must keep it.

## Code style

Didn't affect any style.

## Testing

## Documentation

There is no new functionality or changes on existing functionality.

### DIFF
--- a/packages/panels/resources/lang/pt_BR/resources/pages/view-record.php
+++ b/packages/panels/resources/lang/pt_BR/resources/pages/view-record.php
@@ -2,14 +2,14 @@
 
 return [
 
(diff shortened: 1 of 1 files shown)


danharrin:
Pending review by our pt_BR managers.

saade (approved):

PR #10613, opened 2024-01-02
Support: Add a default 'self' binding that resolves to the current component instance

Voltra:
## Description

This PR provides a default closure dependency name `'self'` that can be used to resolve to `$this` component instance.

The most important use case is components that use the closure dependency system, but don't provide any custom binding. Sometimes we would like to use the current instance, but can't.

For instance there is, AFAIK, currently no way for someone to get the current `Filament\Resources\Components\Tab` instance in its `badge` method to easily reuse the query modification function there.

With this PR, you'd be able to write something like this:

```php
namespace App\Filament\Resources\PlaceResource\Pages;

use App\Enums\PlaceStatus;
use App\Filament\Resources\PlaceResource;
use App\Models\Place;
use Filament\Resources\Components\Tab;

class ListPlaces extends ListRecords
{
    public function getTabs(): array
    {
        return [
            Tab::make('All'),
            Tab::make('Active')
                ->modifyQueryUsing(fn(Builder $query) => $query->whereIn('status', [
                    PlaceStatus::CONSIDERING,
                    PlaceStatus::ACTIVELY_LOOKING,
                    PlaceStatus::OK,
                ]))
                ->badge(fn(Tab $self) => $self->modifyQuery(Place::query())->count()),
        ];
    }
}
```

## Code style

## Testing

## Documentation

> Will edit the docs website's pages related to parameters resolution

### DIFF
--- a/packages/support/src/Concerns/EvaluatesClosures.php
+++ b/packages/support/src/Concerns/EvaluatesClosures.php
@@ -100,6 +100,10 @@ trait EvaluatesClosures
      */
     protected function resolveDefaultClosureDependencyForEvaluationByName(string $parameterName): array
     {
+        if ($parameterName == 'self') {
+            return [$this];
+        }
+
         return [];
     }
 


danharrin:
Each component can have an `$evaluationIdentifier` set to do this. Most components have this. For example, form components have `$component`. This class just needs its `$evaluationIdentifier` set to `'tab'` instead. Please submit another PR for that if you want to see it.

Voltra:
I wouldn't mind doing so, but this would in all likelihood mean a lot of PRs across all packages that provide components where this would fix the issue once and for all "by default" (and for everyone, including 3rd-party packages) whilst allowing each component to still provide another name (or an override) if needed

danharrin:
I understand, but this is the pattern we have chosen right now. Also, `self` is a PHP keyword and could be confused with `$this` in a non-static function

PR #10614, opened 2024-01-02
Fix when linking from a panel with tenancy to one without

x7ryan:
When linking from a panel with multitenancy to one without multitenancy it was adding "?tenant=(id)" to the URL, even if I explicitly passed null as the tenant parameter. 

As a temporary workaround I am using route(getRouteName()) which is what getUrl is doing anyway.

This PR provides a solution by only adding the tenancy parameter if you are linking within the same panel, or if the panel you are linking to also has multitenancy.


danharrin:
Thanks

PR #10616, opened 2024-01-03
Explain how to join vertical nav items with a bar

drbyte:
I didn't see this addressed elsewhere in the docs.

Is this the best spot for it? I suppose it could go in the Navigation Groups area. 
I'm just not sure where people will "look for" this to be explained, and what words they will use when searching, so that we give them something to find with the search.

### DIFF
--- a/packages/panels/docs/06-navigation.md
+++ b/packages/panels/docs/06-navigation.md
@@ -31,7 +31,7 @@ To customize a navigation item's [icon](https://blade-ui-kit.com/blade-icons?set
 protected static ?string $navigationIcon = 'heroicon-o-document-text';
 ```
 
-If you set `$navigationIcon = null` on multiple resources within the same navigation Group, those items will be joined with a vertical bar below the Group name.
+If you set `$navigationIcon = null` on all items within the same navigation group, those items will be joined with a vertical bar below the Group name.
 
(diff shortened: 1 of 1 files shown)


drbyte:
Example from the Demo site:
Screen Shot 2024-01-02 at 10 43 37 PM

zepfietje:
It's not just for group items, also for nested items of a parent which has an icon and children without.

danharrin:
But that one isn't a choice, so I don't think there is much to explain in the docs.

zepfietje:
> But that one isn't a choice, so I don't think there is much to explain in the docs.

Fair.

packages/panels/docs/06-navigation.md
```diff
@@ -31,6 +31,8 @@ To customize a navigation item's [icon](https://blade-ui-kit.com/blade-icons?set
 protected static ?string $navigationIcon = 'heroicon-o-document-text';
 ```
 
+If you set `$navigationIcon = null` on multiple resources within the same navigation Group, those items will be joined with a vertical bar below the Group name.
```
danharrin:
```suggestion
If you set `$navigationIcon = null` on all items within the same navigation group, those items will be joined with a vertical bar below the Group name.
```

PR #10623, opened 2024-01-03
Rework toggle buttons

### LINKED ISSUE #10003: Support checkbox list buttons and button group
Follow-up of https://github.com/filamentphp/filament/pull/9860.

Introduce new `ToggleButtons` component.

zepfietje:
## Description

## Code style

## Testing

## Documentation


zepfietje:
@danharrin there's still a minor design issue when selecting multiple grouped buttons (the dividers), but it's tricky to solve and ok to merge in it's current state I think:

Screenshot 2024-01-03 at 12 39 00

Don't have time to update the docs, so it'd be great if you could handle that.

Also, `CheckboxList` currently has relationship support, but I'll leave that up to you if we want to extract that and add it to `ToggleButtons` as well.

Oh, and also the new icon aliases should be added to the docs still.

danharrin:
I don't think we will add relationship support, I don't think this is the correct UI for it. Same with normal radio, which doesn't have it either.

gavinhewitt:
> I don't think we will add relationship support, I don't think this is the correct UI for it

I was kinda suprised that relationship is not available on ToggleButtons. I feel they are the same as CheckboxList, with the exception that the checkbox itself is hidden. Guess I'll have to hack something together then :)

danharrin:
You can work something out with loadStateFromRelationshipsUsing() and saveRelationshipsUsing() if you need that desperately, same as Radio

PR #10629, opened 2024-01-03
Add readOnly method documentation with security note

emmadesilva:
I find it unintuitive that the readonly method does not call `->dehydrated(false)` automatically, so this adds a note to remind the developer to call it manually.

## Description

- [x] Adding a code comment has no change on functionality.

## Code style

- [x] `composer cs` command has not been run as I made the edit using the web UI.

## Testing

## Documentation

### DIFF
--- a/packages/forms/docs/03-fields/02-text-input.md
+++ b/packages/forms/docs/03-fields/02-text-input.md
@@ -196,6 +196,23 @@ TextInput::make('amount')
     ->numeric()
 ```
 
+## Making the field read-only
+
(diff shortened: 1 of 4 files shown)


danharrin:
In which case would you use this instead of disabled() ?

emmadesilva:
> In which case would you use this instead of disabled() ?

In case I want a field that cannot be edited, but can still be selected and interacted with. For example, readOnly is still keyboard accessible and focusable. The disabled option also dims the input.

Disabled:
image

Readonly:
image
image

danharrin:
I think it's probably better if this was added to the docs instead of a docblock, its more discoverable as I feel like not enough people source dive if something "works" just fine

PR #10649, opened 2024-01-04
Update table column affix helpers to support Htmlables

emmadesilva:
## Description

Updates the `TableColumn` `->prefix()` and  `->suffix()` helpers support `Htmlables` like `HtmlStrings`.

```php
  ->suffix(function (Project $record): ?HtmlString {
      return $record->isActive()
        ? new HtmlString('<span class="text-green-500">Active</span>') 
        : null;
  })
```

Fixes https://github.com/filamentphp/filament/discussions/9729 and https://www.answeroverflow.com/m/1154022222118256700

image

image

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/tables/src/Columns/Concerns/CanFormatState.php
+++ b/packages/tables/src/Columns/Concerns/CanFormatState.php
@@ -286,12 +286,12 @@ trait CanFormatState
         return $this->evaluate($this->isHtml) || $this->isMarkdown();
     }
 
-    public function getPrefix(): null | string | Htmlable
+    public function getPrefix(): string | Htmlable | null
     {
         return $this->evaluate($this->prefix);
     }
 
-    public function getSuffix(): null | string | Htmlable
+    public function getSuffix(): string | Htmlable | null
     {
         return $this->evaluate($this->suffix);
     }
--- a/packages/tables/src/Columns/Concerns/CanFormatState.php
+++ b/packages/tables/src/Columns/Concerns/CanFormatState.php
@@ -8,9 +8,7 @@ use Filament\Tables\Columns\TextColumn;
 use Filament\Tables\Table;
 use Illuminate\Contracts\Support\Htmlable;
 use Illuminate\Support\Carbon;
-use Illuminate\Support\HtmlString;
 use Illuminate\Support\Str;
-use Illuminate\Support\Stringable;
 
 use function Filament\Support\format_money;
 use function Filament\Support\format_number;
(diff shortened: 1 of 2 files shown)


packages/tables/src/Columns/Concerns/CanFormatState.php
```diff
@@ -227,10 +228,18 @@ public function formatState(mixed $state): mixed
...
+                $this->isHtml = true;
+            }
         }
 
         if (filled($suffix = $this->getSuffix())) {
             $state = $state . $suffix;
+
+            if ($suffix instanceof Htmlable) {
+              $this->isHtml = true;
+            }
```
emmadesilva: Is this the best way to do this?

emmadesilva:
I'm not super well versed in PHPStan. Are the test failures my fault?

danharrin:
The problem with this is that if the prefix/suffix are Htmlable, but the main content isn't, the main content is still open to XSS attacks as if it was

So when you set `isHtml`, you will also need to htmlencode the state if it is not Htmlable.

danharrin (changes requested):
The problem with this is that if the prefix/suffix are Htmlable, but the main content isn't, the main content is still open to XSS attacks as if it was

So when you set isHtml, you will also need to htmlencode the state if it is not Htmlable.

emmadesilva:
> The problem with this is that if the prefix/suffix are Htmlable, but the main content isn't, the main content is still open to XSS attacks as if it was
> 
> So when you set isHtml, you will also need to htmlencode the state if it is not Htmlable.

Okay, how should we handle this? I think that the main content state should also support Htmlable, since the TextColumn state also supports it.

danharrin:
the main content state is the TextColumn state I was referring to

emmadesilva:
> the main content state is the TextColumn state I was referring to

The TextColumn state is Htmlable, it's where I got the code for this PR from.

For example, here's the workaround I'm currently using to get a HTML suffix:

```php
    ->formatStateUsing(function (string $state, Project $record): HtmlString {
        $suffix = $record->isActive() ? view('components.filament.support.active-project-badge')->render() : '';

        return new HtmlString('<span class="flex">'.$state.$suffix.'</span>');
    })
```

danharrin:
I know, but using an Htmlable prefix/suffix should not also make the main content accept HTML. Its a security issue

emmadesilva:
> I know, but using an Htmlable prefix/suffix should not also make the main content accept HTML. Its a security issue

Ah, now I get what you mean. My implementation has a side effect. You're more well versed in how these internals work, what do you think is a better way of doing this?

danharrin:
When you set `$isHtml = true` for the prefix/suffix, you will also need to `e()` the main content state.

PR #10662, opened 2024-01-05
Eager load grouped by relations

### LINKED ISSUE #10580: Grouping rows doesn't eager load the relationship with dot notation
### Package

filament/filament

### Package Version

v3.1.32

### Laravel Version

v10.39.0

### Livewire Version

v3.3.3

### PHP Version

PHP 8.2.8

### Problem description

When [grouping rows by a relationship attribute](https://filamentphp.com/docs/3.x/tables/grouping#grouping-by-a-relationship-attribute) on a table, a query is performed for every entry in the table.

Only happens if you don't already have the same field in the table itself. If you add it, the relationship is eager loaded and only one query is performed.

### Expected behavior

I guess it should eager load the relationship, as it does in normal columns on the table.

### Steps to reproduce

1. Access the Posts resource in the reproduction repository.
2. Check in Debugbar the queries and see how a query is performed for each post to get the category.
3. Modify the example and add a 'category.name' column.
4. Check in Debugbar again, where now only one query is done to get all the categories.

### Reproduction repository

https://github.com/RixzZ/filament-issue-01

### Relevant log output

_No response_

omaresmael:
## Description

This pr eager loads the table relations when the relation is used in [grouping by](https://filamentphp.com/docs/3.x/tables/grouping#grouping-by-a-relationship-attribute), it fixes this reported [issue](https://github.com/filamentphp/filament/issues/10580)

```php
    {
        return $table
            ->defaultGroup('category.name')
            ->columns([
                Tables\Columns\TextColumn::make('title')
                    ->searchable(),
            ])
    }
```

Queries Before this pr
image

Queries after 
image

let me know please what you think, and thank you guys for creating such a brilliant project ❤️ 🚀

### DIFF
--- a/packages/tables/src/Concerns/CanGroupRecords.php
+++ b/packages/tables/src/Concerns/CanGroupRecords.php
@@ -53,6 +53,8 @@ trait CanGroupRecords
             return $query;
         }
 
+        $group->applyEagerLoading($query);
+
         return $group->orderQuery($query, $this->getTableGroupingDirection() ?? 'asc');
     }
 }
--- a/packages/tables/src/Concerns/CanGroupRecords.php
+++ b/packages/tables/src/Concerns/CanGroupRecords.php
@@ -55,6 +55,8 @@ trait CanGroupRecords
 
         $group->applyEagerLoading($query);
 
-        return $group->orderQuery($query, $this->getTableGroupingDirection() ?? 'asc');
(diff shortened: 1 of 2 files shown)


danharrin (changes requested):
Please copy the method from that trait into the Group class, we do not need the whole trait.

PR #10677, opened 2024-01-06
Update base Widget class to support HtmlString placeholder() return types

emmadesilva:
Adds a union type so that `HtmlStrings` can be returned from overridden `placeholder()` methods in custom `Widget` classes, making it easier to customize without having to create a view.

```php
public function placeholder(): HtmlString
{
    return new HtmlString('<div>Example</div>');
}
```

## Description

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/widgets/src/Widget.php
+++ b/packages/widgets/src/Widget.php
@@ -2,6 +2,7 @@
 
 namespace Filament\Widgets;
 
+use Illuminate\Support\HtmlString;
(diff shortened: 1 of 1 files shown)


danharrin:
Hmm, I'm going to say no to this one. I do not think it is good practice to write large amounts of HTML in PHP classes, so I want to discourage it. Placeholders often require a lot of HTML.

PR #10695, opened 2024-01-08
add reverseBoolean method to Radio field

ziming:
## Description

Add reverseBoolean() method to Radio field

Often a radio field is used for declarations in the form of:

"I have never drunk drive before"

- Yes
- No

However the database column, is in positive form of "has_drunk_drive_before" which make sense since writing code we often prefer it to be in positive form

the purpose of ->reverseBoolean() is to support this use case.


danharrin:
Thanks, but I'm not going to add this one. It is quite easily achievable without adding a new method, and it's a little confusing.

Issue #4319, opened 2022-10-02
SlideOver appearance animation


JohnnyMaynne:
### Package

filament/filament

### Package Version

v2.16.19

### Laravel Version

v9.33

### Livewire Version

v2.10.7

### PHP Version

v8.1.3

### Problem description

Slideover spawn animation starts from the left

Video:  https://www.dropbox.com/s/qxw0s3saeziq5gq/111.mov?dl=0

### Expected behavior

Slideover spawn animation starts from the right

### Steps to reproduce

Add to page action slideOver() methode

### Reproduction repository

https://github.com/not-link

danharrin:
I can't reproduce this myself, and you've given no reproduction repository or code examples.

andrewdwallo:
I can confirm that this does happen sometimes. I do not know why. But I did change this in the modal:
```js
x-show="isShown"
x-transition:enter="duration-300"
x-transition:leave="duration-300"
@if ($width === MaxWidth::Screen)
@elseif ($slideOver)
       x-transition:enter-start="translate-x-full rtl:-translate-x-full"
       x-transition:enter-end="translate-x-0"
       x-transition:leave-start="translate-x-0"
       x-transition:leave-end="translate-x-full rtl:-translate-x-full"
@else
```

To this:
```js
x-show="isShown"
x-transition:enter-start="duration-300"
x-transition:enter-end="duration-300"
x-transition:leave-start="duration-300"
x-transition:leave-end="duration-300"
@if ($width === MaxWidth::Screen)
@elseif ($slideOver)
       x-transition:enter-start="translate-x-full rtl:-translate-x-full"
       x-transition:enter-end="translate-x-0"
       x-transition:leave-start="translate-x-0"
       x-transition:leave-end="translate-x-full rtl:-translate-x-full"
@else
```
Which is pretty much the same right? But it causes the issue to go way.

danharrin:
Can you make a PR please @andrewdwallo?

andrewdwallo:
Yeah sure! I am not on the computer right now but I will first thing tomorrow.

andrewdwallo:
@danharrin I noticed the issue occurred specifically when having an href element inside the modal, if that tells you anything of what  might be causing it. But here is more info in the meantime: https://discord.com/channels/883083792112300104/1138244989495824495/1193704489044168866

PR #10713, opened 2024-01-09
fix: Stabilize slideover modal to prevent jitteriness on open

andrewdwallo:
This pull request addresses an issue with the slideover modal where it exhibited jittery and shaky behavior upon opening. This would be more pronounced based on the browser you were using and whether it utilized hardware-acceleration. The goal was to ensure a smooth, stable transition when the modal is activated. This issue was also mentioned here #4319, in which that video is what it looks like on Chrome. The video below is what it looks like using Firefox.

### Key Changes
**Transition Property Adjustment:** Modified the transition properties of the slideover modal. Previously, the `translate-x-full` property was used, which caused the modal to slide more than necessary, leading to an unstable appearance. This has been adjusted to a more appropriate value that maintains the visual integrity of the sliding effect without causing instability.

**CSS Refinement and Removal of `will-change`:** Reviewed and refined the CSS associated with the modal to eliminate any conflicting styles or properties that could contribute to the jitteriness. Importantly, the `will-change: transform` property was removed from the outer div of the modal. This change was made because the `transform` CSS properties were not being utilized on this element, and `will-change: transform` does not affect the children of an element. This adjustment helps ensure that browser rendering optimizations are correctly applied where actually needed.

**Cross-Browser Consistency:** Ensured that the changes provide a consistent experience across different browsers and devices.

These changes result in a more fluid user experience when interacting with the slideover modal, enhancing the overall usability and visual appeal of the interface.

### Before

[SlideoverBefore](https://github.com/filamentphp/filament/assets/104294090/06b0c94c-1cf8-4e55-be31-a6e8918c21cb)

### After

[SlideoverAfter](https://github.com/filamentphp/filament/assets/104294090/9b475913-8d6e-4465-9d4a-0360bc0c69ab)

## Testing

### DIFF
--- a/packages/support/resources/views/components/modal/index.blade.php
+++ b/packages/support/resources/views/components/modal/index.blade.php
@@ -115,7 +115,6 @@
                 'fi-modal-close-overlay fixed inset-0 bg-gray-950/50 dark:bg-gray-950/75',
                 'cursor-pointer' => $closeByClickingAway,
             ])
-            style="will-change: transform"
         ></div>
 
         <div
@@ -143,14 +142,14 @@
                     x-on:keydown.window.escape="close()"
                 @endif
                 x-show="isShown"
-                x-transition:enter="duration-300"
-                x-transition:leave="duration-300"
+                x-transition:enter="transition ease-out duration-300"
+                x-transition:leave="transition ease-in duration-300"
                 @if ($width === MaxWidth::Screen)
                 @elseif ($slideOver)
-                    x-transition:enter-start="translate-x-full rtl:-translate-x-full"
+                    x-transition:enter-start="translate-x-1/2 rtl:-translate-x-1/2"
                     x-transition:enter-end="translate-x-0"
                     x-transition:leave-start="translate-x-0"
-                    x-transition:leave-end="translate-x-full rtl:-translate-x-full"
+                    x-transition:leave-end="translate-x-1/2 rtl:-translate-x-1/2"
                 @else
                     x-transition:enter-start="scale-95"
                     x-transition:enter-end="scale-100"


danharrin:
Thank you very much for the effort put into the explanation, that really helps.

danharrin:
So, I've just tested this locally and the results are kinda opposite for me:

Before:

https://github.com/filamentphp/filament/assets/41773797/4ee28e89-4ddf-4f7e-821d-19183508d60c

After:

https://github.com/filamentphp/filament/assets/41773797/682025e7-3de5-4d04-b404-2ec94691a239

In my opinion, "before" looks smoother, as "after" flickers out before the exit animation finishes.

I am using Arc on MacOS.

I am going to temporarily revert this PR, so you can see if you can replicate this behaviour. And then we can discuss if can make any changes to improve the experience on Arc/other Chromium based browsers with hardware acceleration enabled.

andrewdwallo:
Um I mean did you not see the video I had from before the changes? It was really bad. The videos you showed are almost identical. Go on Edge, turn off hardware acceleration, and then tell me what it looks like from using the code before the changes of this commit. Some people with better systems may not even notice the jittery behavior like you, but I can guarantee you that it is there for many many people.

andrewdwallo:
As long as we don't use `translate-x-full` for the transition then it should be fine for those with less performant systems. Maybe something closer to it such as `translate-x-3/4` would be better common ground.

andrewdwallo:
And if you really want to, we could just put `backface-visibility: hidden` on the div, which does remove the jittery behavior but it would still slide out a little more than needed on some browsers.

danharrin:
Can you show me what `backface-visibility: hidden` looks like in Chrome using hardware acceleration?

danharrin:
> Um I mean did you not see the video I had from before the changes? It was really bad. The videos you showed are almost identical.

I did, and yes mine are nearly identical. That doesn't mean I am open to compromises for the most commonly used browser configuration.

andrewdwallo:
Yes I will show you what backface-visibility applied looks like on Chrome. The power went out cause we have bad storms but I will get back to you on that.

danharrin:
Stay safe

sprtk-ches:
FWIW, I also get the same behaviour as the one described in the original issue. It's not always, but a lot of the time the slide-overs behave erratically - they will go halfway to the screen then bounce back etc.

I've seen this on w10/w11 + firefox/chrome. As I said, it doesn't happen all the time but, when it does, it's very off-putting

andrewdwallo:
@danharrin Okay I am back. I just wanted to mention that the original video was using hardware acceleration with Firefox. I only mentioned Edge because it is much worse and more noticeable with it.

andrewdwallo:
@danharrin This is what it looks like in Chrome (hardware-acceleration enabled) on MacOS.

### Before Commit

https://github.com/filamentphp/filament/assets/104294090/8dfb60b5-1661-4bc8-911e-659a5588cf3c

### After Commit

https://github.com/filamentphp/filament/assets/104294090/77070a92-2897-4b77-8704-437e89f846a6

andrewdwallo:
This is with Edge (hardware-acceleration enabled).

### Before Commit

[](https://github.com/filamentphp/filament/assets/104294090/4b1521db-8dff-402b-a891-af62e461bd73)

### After Commit

[](https://github.com/filamentphp/filament/assets/104294090/514c0b15-c5d2-4898-b7d9-155c474a3b85)

andrewdwallo:
@danharrin I justed tested the `backface-visibility: hidden` with the code from before this commit on Chrome, and it has no positive effect. It only seems to fix the jittery issue on Firefox and Edge.

andrewdwallo:
> So, I've just tested this locally and the results are kinda opposite for me:
>
> In my opinion, "before" looks smoother, as "after" flickers out before the exit animation finishes.
> 
> I am using Arc on MacOS.
> 
> I am going to temporarily revert this PR, so you can see if you can replicate this behaviour. And then we can discuss if can make any changes to improve the experience on Arc/other Chromium based browsers with hardware acceleration enabled.

If you want the exit animation to be smoother as before, we can find common ground using this code:
```blade
x-transition:enter="duration-300"
x-transition:leave="duration-300"
@if ($width === MaxWidth::Screen)
@elseif ($slideOver)
    x-transition:enter-start="translate-x-2/3 rtl:-translate-x-2/3"
    x-transition:enter-end="translate-x-0"
    x-transition:leave-start="translate-x-0"
    x-transition:leave-end="translate-x-full rtl:-translate-x-full"
@else
```

This will look smoother on all browsers and will fix the jittery issue and the modal sliding out too far for the enter animation.


[](https://github.com/filamentphp/filament/assets/104294090/1da767de-a5db-42ce-a512-556814d45730)

Just get back to me whenever you can. If you think the above looks good then I will update the commit.

danharrin:
It looks good in the video, but if you open a PR I can test it 👍

andrewdwallo:
Oh so I guess I can't push to this PR anymore? I have to make another separate one?

PR #10714, opened 2024-01-09
Docs for relation manager: Add missing "?"

SebastianSchoeps:
When copying the example for "Customizing the relation manager title" the following error occurs:

```
Type of ...\RelationManagers\ContributionsRelationManager::$title must be ?string (as in class Filament\Resources\RelationManagers\RelationManager)
```

This just adds the missing "?" before `string` in the docs.


danharrin:
Thanks

PR #10724, opened 2024-01-09
Docs: Add quick reference for Blade icon usage

emmadesilva:
## Abstract

Adds a line users can quickly copy to use the Blade icon component

```blade
<x-filament::icon icon="heroicon-o-check-circle" />
```

## Description

- [x] No code changes

## Code style

- [x] No code changes

## Testing

- [x] No code changes

## Documentation


danharrin:
It could be confusing why we use this internally, since normal Blade icon output works just fine. Hence why it is not documented. It only applies when you are using an icon alias.

emmadesilva:
> It could be confusing why we use this internally, since normal Blade icon output works just fine. Hence why it is not documented. It only applies when you are using an icon alias.

Ah okay, understood!

PR #10735, opened 2024-01-10
fix: Stabilize slideover modal

andrewdwallo:
This is an update (part 2) to stabilize the slideover modal to prevent jittery behavior on open. The original commit was reverted. This commit is to find common ground on what would be best as previously discussed. The original commit and all relevant information: #10713

### DIFF
--- a/packages/support/resources/views/components/modal/index.blade.php
+++ b/packages/support/resources/views/components/modal/index.blade.php
@@ -115,7 +115,6 @@
                 'fi-modal-close-overlay fixed inset-0 bg-gray-950/50 dark:bg-gray-950/75',
                 'cursor-pointer' => $closeByClickingAway,
             ])
-            style="will-change: transform"
         ></div>
 
         <div
@@ -147,7 +146,7 @@
                 x-transition:leave="duration-300"
                 @if ($width === MaxWidth::Screen)
                 @elseif ($slideOver)
-                    x-transition:enter-start="translate-x-full rtl:-translate-x-full"
+                    x-transition:enter-start="translate-x-2/3 rtl:-translate-x-2/3"
                     x-transition:enter-end="translate-x-0"
                     x-transition:leave-start="translate-x-0"
                     x-transition:leave-end="translate-x-full rtl:-translate-x-full"


binaryfire:
Thanks for working on this! I've been noticing this issue as well.

packages/support/resources/views/components/modal/index.blade.php
```diff
@@ -147,7 +146,7 @@
                 x-transition:leave="duration-300"
                 @if ($width === MaxWidth::Screen)
                 @elseif ($slideOver)
-                    x-transition:enter-start="translate-x-full rtl:-translate-x-full"
+                    x-transition:enter-start="translate-x-2/3 rtl:-translate-x-2/3"
```
zepfietje, line 149: Could you elaborate on how this fixes the issue, @andrewdwallo?
Also, does adding `will-change` here make any difference?

andrewdwallo: Hi @zepfietje,

The change from `translate-x-full` to `translate-x-2/3` reduces the off-screen movement of the modal. This adjustment is particularly beneficial in browsers not utilizing hardware acceleration, as CPU rendering can exacerbate jitteriness. Even with hardware acceleration, this modification contributes to a smoother transition, as the issue was still present.

Regarding the removal of `will-change: transform`, it was eliminated for several reasons:
1. **No Direct Transformation:** It was applied to the outer div, which didn't undergo direct transformations. `will-change` is intended to optimize elements that experience transformations. Its use where no transformations occur can mislead the browser's optimization strategies.
2. **Non-Application to Child Elements:** This property does not inherently apply to an element's children. The actual transformations were on the modal's children, not the outer div, rendering the use of `will-change: transform` on the outer div ineffective.
3. **Performance Considerations:** `will-change` should be used sparingly, as a last resort, because it hints to the browser to prepare for changes. Overuse or misuse can lead to resource mismanagement. Removing it from elements where it's not beneficial ensures more efficient resource use.
4. **Inappropriate Static Use:** The property was originally statically applied. Best practices recommend applying `will-change` dynamically before an element undergoes a transformation or transition, and removing it afterward. This approach prevents the browser from holding onto unnecessary optimizations and aligns with efficient resource management.
5. **Other Benefits:** Eliminating unnecessary use of `will-change` also helps in preventing side effects like creating new layers that could complicate the rendering process, thus maintaining a leaner and more efficient rendering pipeline.

In summary, these adjustments, both in the transition properties and the removal of `will-change: transform`, align with best practices for performant animations and transitions. They ensure the browser's optimizations are effectively and appropriately used.

I hope this clarifies the modifications made in the pull request.

andrewdwallo: Also, even if `will-change: transform` was applied to the correct element, there would most likely be no benefit. I tested it myself and saw no improvement, so it can depend, but it is best to not use it at all. Using `will-change: transform` on elements with transformations is often unnecessary as modern browsers already apply optimal rendering practices. Overriding this with `will-change` can be costly, potentially leading to unnecessary resource allocation.

danharrin:
I think the `will-change: transform` needs to stay, for #6102

andrewdwallo:
@danharrin Adding `will-change: transform` here did not fix that issue. Actually, that issue is still present currently. I just tested it. `will-change: transform` here is not affecting that in any manner. I can try to find out the cause of #6102 and #10926 and try to fix it though if you would like?

andrewdwallo:
I found the fix to #6102 and #10926. I will make a separate PR here shortly.

danharrin:
Thanks!

zepfietje (changes requested):
Let me know once this is ready for final review, @danharrin @andrewdwallo.

Probably want to add an opacity translation as well, since otherwise the slide-over modal would just appear and hide with the horizontal animation. Or is it already present?

andrewdwallo:
I'm not 100% sure, but I think specifically for the slideover, adding an opacity transition (fading in and out), will make it lose that sliding effect and instead will appear as if it’s simply fading into view more..

andrewdwallo:
I will check tomorrow though after I get some sleep. 👍

andrewdwallo:
I think it is good. I also did want to mention that removing the opacity from `x-transition.duration.300ms.opacity` has a pretty major effect on what is actually occurring. I think somehow the transform scale and origin is appearing in the center of the page and then quickly positions in its place to produce the slideover effect. Not sure if this is intended behavior though...

andrewdwallo:
Hey guys, I was hoping for some feedback about this PR. Is there a reason why this is being delayed from being merged? I don't really see any negative consequences from the changes made in this PR, and I actually see only positive aspects of merging this considering that many users have also pointed out this issue.. Is there something else that needs to be worked on first before this? Is it an underlying alpine issue or something else?

zepfietje:
I'm looking into this now, @andrewdwallo.
The reason for it not being merged yet is simply that I have limited time to work on Filament (I'd love to be able to spend more time on it...).

Will keep you updated with my findings.

zepfietje:
I've fixed the modal animation issue in https://github.com/filamentphp/filament/commit/66f9f7d772e1197462aebadd166c8e38038e40ba.
It was caused by overflow being hidden on a wrapping element.

PR #10742, opened 2024-01-10
Added ability to customize default panel route slugs

boris-glumpler:
## Description

Currently the default slugs, like auth and billing, are hardcoded and cannot be changed. For the majority of apps this is probably fine, but for apps in languages other than English this can be a bit of an annoyance. This PR fixes this by making them configurable.

## Code style

## Testing

## Documentation


packages/panels/src/Panel/Concerns/HasAuth.php
danharrin, line 1: Please add new methods to set the slugs, instead of adding more parameters to the existing methods, which are getting too long

danharrin (changes requested):

packages/panels/src/Panel/Concerns/HasAuth.php
boris-glumpler: I've created dedicated methods for each slug.

devhoussam1998: you forgot `->profile()`

boris-glumpler: Nope, the`EditProfile` component registers its own route, so you could for example override the `getSlug()` method to set a custom slug.

pxlrbt: @danharrin Should we maybe move to a general `->slugs()`,`->urls()` or similar method to keep this cleaner? With the route name as the key?

```php
->slugs([
  'register' => '...',
  'reset' => '...',
])

boris-glumpler: Or just implement `HasRoutes` on the hardcoded route components. Would make it consistent. It could potentially be a bit confusing when there are two ways to adjust the slugs.

danharrin: No, I don't like the idea of an array of slugs with no autocomplete

danharrin:
Thanks

boris-glumpler:
Pleasure was all mine!

devhoussam1998:
> Pleasure was all mine!

you forgot the add slug for profile?

boris-glumpler:
> you forgot the add slug for profile?

Nope, see my answer to your earlier remark.

PR #10747, opened 2024-01-10
Allow Placeholder to render HTML content

4m1n0s:
This PR add possibility to Placeholder component to render HTML content.


danharrin:
You just need to return an HtmlString() object from the content

PR #10766, opened 2024-01-11
Prepend options to select improving UX

kainiklas:
## Description

In a select field with hundreds of options you want to give the user e.g., the top 5 common/ popular options at the top of the selection list. This PR adds the ability to prepend options to the list. 

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/forms/docs/03-fields/03-select.md
+++ b/packages/forms/docs/03-fields/03-select.md
@@ -532,6 +532,20 @@ Select::make('author_id')
 
 Ensure that you are not lowering the debounce too much, as this may cause the select to become slow and unresponsive due to a high number of network requests to retrieve options from server.
 
+## Prepend options
+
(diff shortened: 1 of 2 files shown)


danharrin:
Thanks for the contribution. However, I am going to reject it this time.

There are some unknowns here and I think we would have to introduce excessive complexity to handle them:
- Search results would not include the extra options, and in which cases would we want them to?
- `getOptionLabelUsing()` etc methods would not include the extra options, which could actually break the UI and show underlying option values instead of their corresponding labels.

Since prepending options can be accomplished in userland, by chaining `->options()` after `->relationship()`, I don't feel like we need to insert this backdoor and introduce the complexity. If this was not possible otherwise, then I would consider it.

kainiklas:
Thanks for the response. Even better if it can be done with the methods we have already build-in.

PR #10795, opened 2024-01-14
Addition of 'whereCustomProperties' to Spatie media library file upload

chrispage1:
## Description

Hey Dan!

Spatie Media Library within Filament has an inherent issue in that it's not easily achievable to use within repeaters/builders. Within a repeater field schema, currently there is no way of determining what files go where, until now.

This addition, adds more flexibility to the plugin by adding a `whereCustomProperties` method, leveraging the pre-existing `custom_properties` attribute and allowing you to scope down according to the attributes value.

This allows the user to easily retrieve certain sets of media that contain the given attributes. By not being too verbose, it still gives the user the flexibility to achieve what they want, by using this methods in a variety of ways.

Here is an example use case - 

```php
Forms\Components\Builder::make('content')
    ->blocks([
        Forms\Components\Builder\Block::make('gallery')
            ->schema([
                Forms\Components\Hidden::make('gallery_id')
                    ->default(fn() => Str::random(12)),
                Forms\Components\SpatieMediaLibraryFileUpload::make('gallery')
                    ->whereCustomProperties(fn(Forms\Get $get) => ['gallery_id' => $get('gallery_id')])
                    ->customProperties(fn(Forms\Get $get) => ['gallery_id' => $get('gallery_id')])
                    ->multiple()
            ]),
    ]),
```

With the code above, each block will be assigned a random 'gallery_id', which can then be used to scope down the media that has been added to it.

Once this has been applied, it's easy to loop through media on the front end and filter by the generated gallery id (for this particular example).

This provides a solution to various cases raised, e.g. https://www.answeroverflow.com/m/1103751391115227288

## Documentation

I've added a block to the plugins documentation


danharrin (changes requested):
Instead of limiting just to custom properties, let's just implement a generic `->filterMedia()` method that accepts a function that returns true or false based on the `$media`? If you evaluate it, you'll be able to access `$get()`

And please move it to after the `whereNotIn()` statement

chrispage1:
Sure, that makes sense and gives a lot more flexibility. I'll work something up this morning 👌

chrispage1:
Morning @danharrin, it's a cold one today!

What do you think to the updated implementation?

One potential problem I have identified, but it is out of the scope of this actual commit is that when removing a block that contains a spatie file upload, the detachMedia is not executed (understandably). Can you see any way around this other than a manual user implementation, e.g. adding an observer and cleaning up removed media ?

Chris.

danharrin:
Hey, for that you could try

```php
->deleteAction(fn ($action) => $action->before(function (array $arguments, Builder $component): void {
    $data = $component->getState()[$arguments['item']];

    // delete the media
}))
```

danharrin:
Thanks Chris!

chrispage1:
Thanks Dan! We've only been using Filament for a couple of weeks but I can see it being a great asset for us. I look forward to contributing on it more!

PR #10799, opened 2024-01-14
Add panel css selector

binaryfire:
In many cases it makes sense to use a single theme for multiple panels. This PR adds a `fi-panel-panel-id` class to each panel, allowing us to use panel-specific CSS in themes.

I also removed the `@stack('styles')` which I added in [this PR](https://github.com/filamentphp/filament/pull/9455). It doesn't work properly with SPA mode because of persistence issues, so it's just dead code atm.

### DIFF
--- a/packages/panels/resources/views/components/layout/base.blade.php
+++ b/packages/panels/resources/views/components/layout/base.blade.php
@@ -62,8 +62,6 @@
             }
         </style>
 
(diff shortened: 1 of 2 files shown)


danharrin:
Removed the breaking change, we can't ship that now, most people are not using SPA mode anyway

PR #10808, opened 2024-01-15
Adds ability to access selected records in a TableAction (not BulkAction)

tiagof:
## Description

Adds ability to access selected records in a TableAction (not BulkAction).
This can be useful when we want to act on multiple records (the selected ones) based on another record (the one with the TableAction).
Example: Copy all/some properties from record A to records B, C, D, ...

Have a look at the changes in the README for a more concrete and complete example.

## Code style

## Testing

## Documentation


packages/tables/src/Actions/Action.php
```diff
@@ -22,6 +22,11 @@ public function getLivewireCallMountedActionName(): string
         return 'callMountedTableAction';
     }
 
+    public function getAlpineClickHandler(): ?string
+    {
+        return "mountTableAction('{$this->getName()}')";
```
danharrin: I think we really need to avoid sending the selected records to the backend if we don't need them, so do we need to add in a condition here to make sure that is the case?

tiagof: Makes sense yes.
Any suggestion on how to do it implicitly? (is it even possible?)
I can only think of having the user explicitly saying that a particular action needs access to the selected records.
Example:
```
Action::make()...->requiresAccessToSelectedRecords()
```

Is this what was on your mind?

danharrin: Yeah, something like that

tiagof: Done!

danharrin:
Thanks

tiagof:
Thank you! 🙏

PR #10820, opened 2024-01-15
add `extraAttributes` in Navigation Group

atmonshi:
## Description

Adding support for `extraAttributes`, in NavigationGroup. I need to pass some CSS classes to specific navigation group 

example usage:
```php
return $panel
->navigationGroups([
        NavigationGroup::make()
            ->label('App')
            ->extraAttributes(['class'=>'highlighted']),
        ])
```

## Code style

## Testing

## Documentation


atmonshi:
running `composer cs` cases weird issues and some code deleted! so I revert it

packages/panels/resources/views/components/sidebar/index.blade.php
danharrin, line 1: Please pass the attributes to topnav groups too

danharrin: And anywhere else we use the group component, if we do

atmonshi: the sidebar navigation (in cluster and sub navigation) dose use the `sidebar.group` but there is no way (I know of) to configure the group from the provider `navigationGroups`

danharrin: Let's do it anyway, I think it is possible just requires a bit of hacking

atmonshi:
all done 👍

packages/panels/resources/views/components/page/sub-navigation/sidebar.blade.php
```diff
@@ -8,6 +8,7 @@
 >
     @foreach ($navigation as $navigationGroup)
         <x-filament-panels::sidebar.group
+            :attributes="\Filament\Support\prepare_inherited_attributes($navigationGroup->getExtraAttributeBag())"
```
zepfietje: Please move `:attributes` to be the last one in the list of attributes (like we do elsewhere).

zepfietje (changes requested):

danharrin:
I've changed my mind, the attributes that apply to a topnav dropdown would probably not be useful to apply to a sidebar group, which is used on mobile devices. Let's replace `extraAttributes()` with `extraSidebarAttributes()` and `extraTopbarAttributes()`?

atmonshi:
hey @danharrin .

np, should I create a new `trait` for `extraSidebarAttributes` and `extraTopbarAttributes`?
or add the methods in the `NavigationGroup` class

danharrin:
Please create new traits :)

atmonshi:
done, hope this what you have in mind! let me know.
also if can you double check the docs `06-navigation.md` if needs improvement :)

danharrin:
Thanks

PR #10847, opened 2024-01-16
fix(exports) update prepare csv job to account for order by

WayneHarris:
## Description

If a descending ```defaultSort``` is set on a resource table, large exports over 2000 records will fail at 1999 records due to the use of ```chunkById``` which does not account for the order direction of the query.

This PR updates the [PrepareCsvExport](https://github.com/filamentphp/filament/blob/3.x/packages/actions/src/Exports/Jobs/PrepareCsvExport.php) to check the base query for any ```$orders``` and direction of the first order.

After identifying if the direction is descending, the pass $descending param is passed to the ```orderedChunkById``` method instead of using ```chunkById```

- [] Visual changes (if any) are shown using screenshots/recordings of before and after.

## Code style

## Testing

## Documentation


WayneHarris:
@danharrin, let me know if you confirm the issue. I'm still looking at other options on how to resolve this, but I've resolved the conflicts for now.

danharrin:
Looks good to me, thanks

PR #10869, opened 2024-01-17
i18: Uzbek language translations changed

encoderuz:
Some words changed.


zepfietje (changes requested):
Please consistently use single quotes instead of double quotes around translation strings.

PR #10875, opened 2024-01-17
Support Alpine Click Handler via modifier

awcodes:
## Description

Adds support for calling `alpineClickHandler()` directly on an Action instead of having to override `getAlpineClickHandler()`

```php
Action::make('test')
    ->alpineClickHandler(fn () => alert('Hello World'))
```


zepfietje:
I've coined this idea multiple times in the past, but IIRC Dan wasn't a big fan of it.
I do agree it's needed though, but I'm not sure about the naming.

awcodes:
He told me the other day he would accept a PR for it with that name.

PR #10876, opened 2024-01-17
Fixed bug in PrepareCsvExport for models with custom primary keys

nathanpelton:
## Description

When the PrepareCsvExport model tries to create the query to export the records it assumes that the id in chunkById is 'id' instead of using $keyName.

## Code style

## Testing

## Documentation

- [N/A] Documentation is up-to-date.

### DIFF
--- a/packages/actions/src/Exports/Jobs/PrepareCsvExport.php
+++ b/packages/actions/src/Exports/Jobs/PrepareCsvExport.php
@@ -103,14 +103,16 @@ class PrepareCsvExport implements ShouldQueue
             return;
         }
 
+        $qualifiedKeyName = $query->getModel()->getQualifiedKeyName();
+
         $query->toBase()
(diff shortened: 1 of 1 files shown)


leogiouldasis:
@nathanpelton 
Hello, this change breaks the export functionality at least in my implementation.
I get `The chunkById operation was aborted because the [discounts.id] column is not present in the query result.`

The model is Discounts and the table is discounts, id exists in discounts. 
If i manually override the chunkById alias and pass id without discounts it works fine.
DB is a psql

`$query->toBase()
            ->select([$qualifiedKeyName])
            ->chunkById(
                $this->chunkSize * 10,
                fn (Collection $records) => $dispatchRecords(
                    Arr::pluck($records->all(), $keyName),
                ),
                'id',
            );`

danharrin:
My fault there, I'll patch it in the next release

PR #10877, opened 2024-01-17
Import Unicode character fix

mxm1070:
Per discussion.

### DIFF
--- a/packages/actions/src/Imports/Jobs/ImportCsv.php
+++ b/packages/actions/src/Imports/Jobs/ImportCsv.php
@@ -72,7 +72,7 @@ class ImportCsv implements ShouldQueue
             $rows = unserialize(base64_decode($this->rows));
         }
 
-        foreach ($rows ?? $this->rows as $row) {
+        foreach (($rows ?? $this->rows) as $row) {
             try {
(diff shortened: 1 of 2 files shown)


packages/actions/src/Imports/Jobs/ImportCsv.php
```diff
@@ -68,7 +68,11 @@ public function handle(): void
         $processedRows = 0;
         $successfulRows = 0;
 
-        foreach ($this->rows as $row) {
+        if( ! is_array($this->rows)) {
```
danharrin:
```suggestion
        if (! is_array($this->rows)) {
```

packages/actions/src/Imports/Jobs/ImportCsv.php
```diff
@@ -68,7 +68,11 @@ public function handle(): void
         $processedRows = 0;
         $successfulRows = 0;
 
-        foreach ($this->rows as $row) {
+        if (! is_array($this->rows)) {
+            $rows = unserialize(base64_decode($this->rows));
+        }
+
+        foreach ($rows ?? $this->rows as $row) {
```
danharrin:
```suggestion
        foreach (($rows ?? $this->rows) as $row) {
```

Issue #10835, opened 2024-01-16
Exporter can't export relationships data


hanaexcl:
### Package

filament/filament

### Package Version

v3.2.2

### Laravel Version

v10.10

### PHP Version

v3.3.5

### Problem description

Exporter can't export relationships data, It will export success, but missing relationships data.

### Expected behavior

Export relationships data.

### Steps to reproduce

Create an exporter that includes relationships data.

```php
public static function getColumns(): array
{
    return [
        ExportColumn::make('id'),
        ExportColumn::make('customer.name'), //Missing
    ];
}
```

### Reproduction repository

https://github.com/hanaexcl/filament-export-test

maartenpaauw:
Having the same exact problem. Data from the model directly is working fine, but data from a relationship is not working.

ekrist1:
Same here. Seems to be a bug with JSON columns as well, it's printed out as ARRAY.

wannesmatthys:
Same issue here

NicholasMartens:
Same here

akaflipper:
Same this side ... despite columns being defined in exporter class, on bulk action a modal pops up with column choices to be exported, then on export ignores column definitions in getColumns() in  Exporter::class ... which are defined as make('relationship.name')

pvdptje:
I have located the issue, but I'm not able to provide a fix yet.

in Concerns\CanExportRecords.php on line 129:

``` $columnMap = collect($data['columnMap'])
                ->filter(fn (array $column): bool => $column['isEnabled'] ?? false)
                ->mapWithKeys(fn (array $column, string $columnName): array => [$columnName => $column['label']])
                ->all();
```

The filter() method removes the relation from the columnmap.
The reason is because a relation is a level deeper in the array. Like this:

![image](https://github.com/filamentphp/filament/assets/3110215/ec415a81-bb89-4b37-86ab-5ff2e93f21da)

The ```$column['isEnabled'] ``` will always return false for a relation.

pvdptje:
This fixes the issue for me:

```         
$columnMap = collect($data['columnMap'])
         ->flatMap(function (array $column, string $columnName) {
             // If it's a simple column
             if (isset($column['isEnabled']) && $column['isEnabled']) {
                 return [$columnName => $column['label']];
             }
     
             // If it's a relation
             return collect($column)->filter(function ($subColumn) {
                 return is_array($subColumn) && ($subColumn['isEnabled'] ?? false);
[...]

PR #10898, opened 2024-01-18
Update CanExportRecords.php

pvdptje:
## Description

Possible fix for: https://github.com/filamentphp/filament/issues/10835


danharrin:
I think #10899 might be slightly closer, but thanks

maartenpaauw:
@pvdptje I'd try out your solution. It works nice for a relationship 1 level deep, but if I want to go deeper (`order.line.price`) it does not work. I did just open a pull request which handles deeper relationships.

PR #10905, opened 2024-01-18
Added italian translation for export.php

Rattone:


Rattone:
@danharrin faster releaser in history

danharrin:
😂 thanks

PR #10927, opened 2024-01-19
Prevent "outlined" button border from being partially hidden.

bennettblack:
## Issue
- Action button borders defined with the `->outlined()` class are partially hidden due to hidden overflow on header.

image

## After Change

image

## Section Snippet
- Note: this does not include the schema.
```php
return Section::make('Notes')
    ->description('Contact Notes')
    ->icon('heroicon-o-pencil')
    ->headerActions([
        Action::make('Add Note')
            ->label('Add Note')
            ->color('primary-blue')
            ->size(ActionSize::ExtraSmall)
            ->outlined()
            ->modalDescription('Create a new note for this contact.')
            ->icon('heroicon-o-plus')
            ->form([
                Textarea::make('content')
                    ->placeholder('A note about this contact...')
                    ->label('')
                    ->rows(3)
                    ->rules(['required', 'string', 'max:500'])
            ])
            ->action(function ($record, $data) {
                $record->notes()->save(new Note([
                    'content' => $data['content'],
                ]));
            }),
    ])
```

### DIFF
--- a/packages/support/resources/views/components/section/index.blade.php
+++ b/packages/support/resources/views/components/section/index.blade.php
@@ -66,7 +66,7 @@
                 } => ! $aside,
             ])
         >
-            <div class="flex items-center gap-3 overflow-hidden">
+            <div class="flex items-center gap-3">
                 @if ($hasIcon)
                     <x-filament::icon
                         :icon="$icon"
--- a/packages/support/resources/views/components/section/index.blade.php
+++ b/packages/support/resources/views/components/section/index.blade.php
@@ -66,7 +66,7 @@
                 } => ! $aside,
             ])
         >
-            <div class="flex items-center gap-3">
+            <div class="flex items-center gap-3 overflow-hidden px-px">
                 @if ($hasIcon)
                     <x-filament::icon
                         :icon="$icon"


danharrin (changes requested):
Hmm, I am very hesitant to do this, as there must be a reason why this class was added in the first place. Maybe Zep can give some more context.

Maybe we should just add an `px-px` to the container instead?

bennettblack:
I've updated the PR - adding `px-px` also fixes this issue.

danharrin:
Question - does it still work if you also add `-mx-px` on top of `px-px`?

packages/support/resources/views/components/section/index.blade.php
```diff
@@ -66,7 +66,7 @@
                 } => ! $aside,
             ])
         >
-            <div class="flex items-center gap-3 overflow-hidden">
+            <div class="flex items-center gap-3 overflow-hidden px-px">
```
zepfietje, line 69: Did you guys mean `p-px`?
I don't think that's a nice fix though.
Have you checked git blame to see why overflow-hidden is here in the first place, @bennettblack?

zepfietje (changes requested):

packages/support/resources/views/components/section/index.blade.php
bennettblack: @zepfietje [Commit Here](https://github.com/filamentphp/filament/commit/36a1aa72abe7878293ad4dece334b4b79dab9f37) Seems that it addresses a previous UI issue.

danharrin: overflow-hidden was in the markup before that commit

zepfietje: Yeah, this was only a recent rework by Dan.

zepfietje: I've used git blame to go back to the creation of the section component, and I can't find the reason `overflow-hidden` is there. I'm just going to remove it, as it causes other issues and I don't currently see why it's needed.

zepfietje:
https://github.com/filamentphp/filament/commit/86f5196982775fd258257aa12921512c2ee5074c

For v4, buttons will be redesigned and the border of outlined buttons will probably be internal instead of the current outside ring. That should fix potential similar issues.

PR #10956, opened 2024-01-21
fix: Focus issue with revealable Password suffix action in input wrapper

andrewdwallo:
### Summary
This pull request addresses an issue where clicking the password reveal suffix action in an input caused the "focus ring" on the input wrapper to flicker or disappear momentarily.

### Problem
The focus ring on the input wrapper was lost when the password reveal button was clicked. This behavior created a visual inconsistency and could be confusing for users, as the focus state seemed to toggle rapidly.

### Solution
Implemented `x-on:pointerdown.prevent` on the password reveal button/action. This change prevents the default behavior associated with the `pointerdown` event without blocking the subsequent click event. As a result, the focus remains consistently on the input wrapper, providing a smoother and more intuitive user interaction.

### Before

https://github.com/filamentphp/filament/assets/104294090/557656ee-bb21-48a1-b1bc-10d2269520cd

### After

https://github.com/filamentphp/filament/assets/104294090/1d628628-e670-4190-80d3-947c0a509c34

### Testing
- The changes have been tested across multiple browsers and devices. The focus ring now remains stable and visible around the input wrapper when the reveal password button/action is clicked, enhancing the user experience and interface consistency.

### Additional Notes
- The solution maintains accessibility standards and does not interfere with the primary function of the reveal password action.

### DIFF
--- a/packages/forms/resources/views/components/text-input.blade.php
+++ b/packages/forms/resources/views/components/text-input.blade.php
@@ -86,6 +86,7 @@
                         $applyStateBindingModifiers('wire:model') => $statePath,
                         'x-bind:type' => $isPasswordRevealable ? 'isPasswordRevealed ? \'text\' : \'password\'' : null,
                         'x-mask' . ($mask instanceof \Filament\Support\RawJs ? ':dynamic' : '') => filled($mask) ? $mask : null,
+                        'class' => $isPasswordRevealable ? 'fi-input-password-revealable' : null,
                     ], escape: false)
             "
         />
--- a/packages/forms/src/Components/TextInput/Actions/HidePasswordAction.php
+++ b/packages/forms/src/Components/TextInput/Actions/HidePasswordAction.php
@@ -25,6 +25,7 @@ class HidePasswordAction extends Action
         $this->extraAttributes([
             'x-cloak' => true,
             'x-show' => 'isPasswordRevealed',
+            'x-on:pointerdown.prevent' => '',
         ]);
 
         $this->alpineClickHandler('isPasswordRevealed = false');
--- a/packages/forms/src/Components/TextInput/Actions/HidePasswordAction.php
+++ b/packages/forms/src/Components/TextInput/Actions/HidePasswordAction.php
@@ -25,7 +25,6 @@ class HidePasswordAction extends Action
         $this->extraAttributes([
             'x-cloak' => true,
             'x-show' => 'isPasswordRevealed',
-            'x-on:pointerdown.prevent' => '',
         ]);
 
         $this->alpineClickHandler('isPasswordRevealed = false');
--- a/packages/forms/src/Components/TextInput/Actions/ShowPasswordAction.php
+++ b/packages/forms/src/Components/TextInput/Actions/ShowPasswordAction.php
@@ -24,6 +24,7 @@ class ShowPasswordAction extends Action
 
         $this->extraAttributes([
             'x-show' => '! isPasswordRevealed',
+            'x-on:pointerdown.prevent' => '',
         ]);
 
         $this->alpineClickHandler('isPasswordRevealed = true');
--- a/packages/forms/src/Components/TextInput/Actions/ShowPasswordAction.php
+++ b/packages/forms/src/Components/TextInput/Actions/ShowPasswordAction.php
@@ -24,7 +24,6 @@ class ShowPasswordAction extends Action
 
         $this->extraAttributes([
             'x-show' => '! isPasswordRevealed',
-            'x-on:pointerdown.prevent' => '',
         ]);
 
         $this->alpineClickHandler('isPasswordRevealed = true');
--- a/packages/support/resources/views/components/input/wrapper.blade.php
+++ b/packages/support/resources/views/components/input/wrapper.blade.php
@@ -100,6 +100,7 @@
 >
     @if ($hasPrefix || $hasLoadingIndicator)
         <div
+            x-on:mousedown.prevent {{-- Prevents input wrapper from losing focus when clicking on the prefix. --}}
             @if (! $hasPrefix)
                 wire:loading.delay.{{ config('filament.livewire_loading_delay', 'default') }}.flex
                 wire:target="{{ $loadingIndicatorTarget }}"
@@ -178,6 +179,7 @@
 
     @if ($hasSuffix)
         <div
+            x-on:mousedown.prevent {{-- Prevents input wrapper from losing focus when clicking on the suffix. --}}
             @class([
                 'flex items-center gap-x-3 pe-3',
                 'ps-1' => $inlineSuffix && filled($suffix),


andrewdwallo:
Also, it is worth mentioning that the Edge browser will also display an eye icon to show/hide the password by default, in which case there will be 2 eye icons side by side. It doesn't look good. The `::-ms-reveal` css pseudoclass can't be targeted using inline styles so I left that  alone for ya'll to decide on what should be done for that.

danharrin:
I think the issue is on all icon buttons used in the suffix/prefix of a field. Is there a good way we can apply it globally?

Please add the `-ms-reveal` override to `panels/resources/css/index.css`

andrewdwallo:
Yes it can be applied globally. Do you want it for all icons and icon actions for affixes in the input? Or just the icon affix actions? Because just clicking on a suffix icon's area also causes the input to lose focus, for example.. even though it is not a button.

andrewdwallo:
Also do you want the `::-ms-reveal` override to apply only when the input is revealable, or just in general?

danharrin:
I think this is only an issue with `<button>` elements flickering, icons don't have the same styling?

Ideally, the CSS would only be applied when the input is revealable, so you might need to introduce a new class on the text input.

andrewdwallo:
@danharrin I'd like to clarify the behavior concerning both interactive (like buttons) and non-interactive elements (like icons) within the suffix area of the input field (for prefix/suffix).

For the suffix action button, the issue was that clicking it caused the focus of the input wrapper to flicker or rapidly gain and lose focus. This behavior was creating a visual inconsistency and could potentially confuse users. By applying `x-on:pointerdown.prevent` to the suffix action area, we've successfully fixed this issue, ensuring that the focus remains stable and does not flicker when the button is clicked.

Example code to be used for just fixing this:

`filament/packages/support/resources/views/components/input/wrapper.blade.php`
```blade
@if ($hasSuffix)
     <div
          @class([
              'flex items-center gap-x-3 pe-3',
              'ps-1' => $inlineSuffix && filled($suffix),
              'ps-2' => $inlineSuffix && blank($suffix),
              'border-s border-gray-200 ps-3 dark:border-white/10' => ! $inlineSuffix,
          ])
      >
          @if (filled($suffix))
              <span class="{{ $labelClasses }}">
                  {{ $suffix }}
              </span>
          @endif

          @if ($suffixIcon)
              <x-filament::icon
                  :alias="$suffixIconAlias"
                  :icon="$suffixIcon"
                  :class="$getIconClasses($suffixIconColor)"
                  :style="$getIconStyles($suffixIconColor)"
              />
          @endif

          @if (count($suffixActions))
              <div x-on:pointerdown.prevent class="{{ $actionsClasses }}"> // HERE
                  @foreach ($suffixActions as $suffixAction)
                      {{ $suffixAction }}
                  @endforeach
              </div>
          @endif
    </div>
@endif
```

For non-interactive elements like icons within the suffix area, the issue was slightly different from that of interactive buttons. Clicking on these icon areas didn't cause flickering, but rather led to the input wrapper losing its focus entirely. This behavior is akin to clicking on a non-interactive part of the page, such as the page body, which results in the input field losing focus. By applying `x-on:pointerdown.prevent` to the entire suffix container, we ensure that clicking anywhere within this area — whether on an interactive button or a non-interactive icon — does not cause the input wrapper to lose its focus.

`filament/packages/support/resources/views/components/input/wrapper.blade.php`
```blade
@if ($hasSuffix)
     <div x-on:pointerdown.prevent
          @class([
              'flex items-center gap-x-3 pe-3',
              'ps-1' => $inlineSuffix && filled($suffix),
              'ps-2' => $inlineSuffix && blank($suffix),
              'border-s border-gray-200 ps-3 dark:border-white/10' => ! $inlineSuffix,
          ])
      >
          @if (filled($suffix))
              <span class="{{ $labelClasses }}">
                  {{ $suffix }}
              </span>
          @endif

          @if ($suffixIcon)
              <x-filament::icon
                  :alias="$suffixIconAlias"
                  :icon="$suffixIcon"
                  :class="$getIconClasses($suffixIconColor)"
                  :style="$getIconStyles($suffixIconColor)"
              />
          @endif

          @if (count($suffixActions))
              <div class="{{ $actionsClasses }}"> // HERE
                  @foreach ($suffixActions as $suffixAction)
                      {{ $suffixAction }}
                  @endforeach
              </div>
          @endif
    </div>
@endif
```

andrewdwallo:
@danharrin Here are video examples of what I mean regarding suffix icons as well. Obviously this isn't that big of a deal, but it would be good to maintain consistency.

### Applying `x-on:pointerdown.prevent` to only the suffix action container

[](https://github.com/filamentphp/filament/assets/104294090/724d6b26-c233-4513-a28c-75da0fdb0e75)

### Applying `x-on:pointerdown.prevent` to the entire suffix container (yes I am clicking here 😂)
[](https://github.com/filamentphp/filament/assets/104294090/d36444d7-47bd-4705-821d-78c62105edcc)

danharrin:
My only concern is with action buttons, will this ever interfere with `Action::make()->action(fn () => ...)` or `Action::make()->url()`, I just want to be sure that clicking the button will still run the function / open the URL

andrewdwallo:
It won't prevent actions from being carried out. I tested it with a modal. I haven't tested a link inside it yet though. I will do that and let you know.

danharrin:
Can you also double check links in SPA mode?

andrewdwallo:
Okay so `x-on:pointerdown.prevent` works for everything except for specifically urls in SPA mode. Clicking on it does not trigger a redirect, etc. I tested this and we could instead use `x-on:mousedown.prevent` which will work for everything while still fixing the issue. We could as well use `x-on:mousedown.prevent x-on:touchstart.prevent` to also prevent the issue from happening on touch devices. What do you think?

danharrin:
`x-on:mousedown.prevent x-on:touch start.prevent` sounds good to me! Thanks for checking SPA mode, I was supicious.

andrewdwallo:
Okay great. Good catch with SPA mode. Didn't think about that. Will update this soon.

andrewdwallo:
@danharrin Okay so for CSS class to remove the Edge browser `::-ms-reveal`, I have this:

`forms/resources/views/components/text-input.blade.php`
```blade
<x-filament::input
            :attributes="
                \Filament\Support\prepare_inherited_attributes($getExtraInputAttributeBag())
                    ->merge($extraAlpineAttributes, escape: false)
                    ->merge([
                        'autocapitalize' => $getAutocapitalize(),
                        'autocomplete' => $getAutocomplete(),
                        'autofocus' => $isAutofocused(),
                        'disabled' => $isDisabled,
                        'id' => $id,
                        'inlinePrefix' => $isPrefixInline && (count($prefixActions) || $prefixIcon || filled($prefixLabel)),
                        'inlineSuffix' => $isSuffixInline && (count($suffixActions) || $suffixIcon || filled($suffixLabel)),
                        'inputmode' => $getInputMode(),
                        'list' => $datalistOptions ? $id . '-list' : null,
                        'max' => (! $isConcealed) ? $getMaxValue() : null,
                        'maxlength' => (! $isConcealed) ? $getMaxLength() : null,
                        'min' => (! $isConcealed) ? $getMinValue() : null,
                        'minlength' => (! $isConcealed) ? $getMinLength() : null,
                        'placeholder' => $getPlaceholder(),
                        'readonly' => $isReadOnly(),
                        'required' => $isRequired() && (! $isConcealed),
                        'step' => $getStep(),
                        'type' => $type,
                        $applyStateBindingModifiers('wire:model') => $statePath,
                        'x-bind:type' => $isPasswordRevealable ? 'isPasswordRevealed ? \'text\' : \'password\'' : null,
                        'x-mask' . ($mask instanceof \Filament\Support\RawJs ? ':dynamic' : '') => filled($mask) ? $mask : null,
                        'class' => $isPasswordRevealable ? 'fi-input-password-revealable' : null,
                    ], escape: false)
            "
        />
```

This works perfectly. I have a question though.. Do you want me to create a new css file to apply this:
```css
.fi-input-password-revealable::-ms-reveal {
    display: none;
}
```

If so, in which package directory? Considering that the code is in the `text-input` component in the `forms` package directory, but the CSS is applied to the `fi-input` component in the `support` directory..

zepfietje:
Just read through the entire conversation above, @andrewdwallo @danharrin.

First of all, I'd suggest not messing too much with focus behavior.
The issue of losing input focus when clicking an affix can be solved by making a click event on the affix focus the input. This would also improve the current behavior where clicking a prefix icon for example doesn't focus the input. Let's discuss the desired behavior regarding this in a separate pull request.

Regarding clicking affix icons, I'd suggest sticking to the browser behavior. The issue here should be solved in a different way: by preventing the input wrapper from rendering its focus state when an action inside it is focused. This is something that's been on my radar for some time, and can be fixed using the `:has()` CSS pseudo-class.

All in all, I'm going to close this PR and link to a new PR that fixes the input wrapper focus state when an affix action is focused.

PR #10929, opened 2024-01-19
Allow markAsRequired() to be used independently of required() to denote a required field.

RalphMRivera:
## Description

Allow markAsRequired() to be used independent of required()

In scenarios where a red asterisk must be present to denote a required field, but `->rules(['required'])` is being used instead of `->required()`, using `->markAsRequired()` will generate the red asterisk.

Example:

```php
TextInput::make('sample')
    ->rules(['required'])
    ->markAsRequired()
```

Currently, only `->markAsRequired(false)` has the effect of removing the asterisk when paired with `->required(),` but `->markAsRequired()` or `->markAsRequired(true)` have no effect even when paired with `->required()`.

This change is helpful in cases where a project ignores browser-side error handling in favor of Laravel validation, and the author wants to display a required asterisk without having to fall back on solutions such as using a `label()` that generates an HtmlString that includes a div with styling classes and an asterisk.

- [x] Visual changes (if any) are shown using screenshots/recordings of before and after.

Original code without proposed changes:

```php
TextInput::make('sample')
    ->rules(['required'])
    ->markAsRequired()
```

image

With Proposed changes:

```php
TextInput::make('sample')
    ->rules(['required'])
    ->markAsRequired()
```

image

## Code style

- [x] `composer cs` command has been run.

## Testing

- [x] Changes have been tested.

## Documentation

- [x] Documentation is up-to-date.

PR #10957, opened 2024-01-21
fix: isMarkedAsRequired should be false by default

imlokesh:
## Description

Recent update (#10929) improves the logic for showing the asterisk label for fields marked as required. But `isMarkedAsRequired` should be `false` by default for this to work correctly. 

Currently, all fields will show an asterisk unless we explicitly call `markAsRequired(false)` on the form component. 

## Code style

## Testing

## Documentation


danharrin:
I think this breaks disabling the asterisk altogether, so going to push up a refactor

PR #10959, opened 2024-01-21
Feature/export action file type

Umar-Farooq-Shafi:
## Description

Allow the exporter to export only the `csv` or `xlsx` file.

## Code style

## Testing

## Documentation


danharrin:
Adding a new DB column like this would be a breaking change, so I've replaced it with #10963

Umar-Farooq-Shafi:
thanks

PR #10961, opened 2024-01-21
Bugfix panel assets

PhiloNL:
## Description
Fix a small bug when using `$panel->assets` (https://github.com/filamentphp/filament/pull/10954) that would cause a multi-dimension array to be created.


danharrin:
Thanks, think I screwed up my refactor 🤣

PR #10970, opened 2024-01-21
Bugfix: Check if resource is scoped to tenant when using spatie/laravel-translatable plugin

KarelBrijs:
## Issue

When using the spatie-laravel-translatable plugin within a resource that is scoped to a tenant the record creation method was missing a check for that thus resulting in an error when saving a resource. This PR adds that check and fixes that issue.


danharrin:
Thanks!

PR #10973, opened 2024-01-21
Add extraAttributes for js assets

binaryfire:
Allows doing things like `defer` etc: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attributes. Particularly useful for custom pages.


danharrin:
You can do `defer()` for this

PR #10981, opened 2024-01-22
Fixed error and replaced function with a better one

thethunderturner:
## Description

Basically the routes isnt always as stated. In reality it would be something like:
`filament.<panel_name>.resources.posts.edit`. Overall, I find this way to be very inconvenient and it could lead to further issues in the future if you decide for example to change the name of the slug. Therefore it is better to reference the resource directly!

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/tables/docs/10-advanced.md
+++ b/packages/tables/docs/10-advanced.md
@@ -107,7 +107,7 @@ public function table(Table $table): Table
 {
     return $table
         ->recordUrl(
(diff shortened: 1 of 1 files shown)


zepfietje:
This is just an example in the Table Builder package docs.
It doesn't depend on panels, so `Resource::getUrl()` doesn't make sense here.

Issue #5972, opened 2023-03-19
[3.x] Markdown editor loses content after modal reopens


danharrin:
### Package

filament/filament

### Package Version

v3.x

### Laravel Version

v10.x

### Livewire Version

v2.x

### PHP Version

8.2

### Problem description

Content of markdown editor disappears when it gets loaded. It reappears when you focus the field.

### Expected behavior

Markdown content rendered.

### Steps to reproduce

Add a markdown editor to a simple resource or relation manager. Ensure each record has content saved to the database for this field.

Edit one table row, then close it. Then open another one. The markdown content is gone. Now focus the field and it is rendered correctly.

### Reproduction repository

https://github.com/filamentphp/demo

PR #10982, opened 2024-01-22
fix: Markdown Editor initialization in modal

### LINKED ISSUE #10926: MarkdownEditor has no content when opened in a modal
### Package

filament/forms

### Package Version

v3.3.2

### Laravel Version

v10.41.0

### Livewire Version

v3.3.5

### PHP Version

PHP 8.3.0

### Problem description

This seems somehow similar to https://github.com/filamentphp/filament/issues/5972 but slightly different

Content of markdown editor is not shown when it gets loaded in a modal. It reappears when you focus the field.

this happens only if the modal is opened from a page that contains a markdown editor itself

### Expected behavior

Markdown content rendered.

### Steps to reproduce

Starting from the reproduction repository code:

1. `composer install`
2. `php artisan migrate --seed --fresh`
3. `php artisan serve`
4. visit http://127.0.0.1:8000/admin
5. login with default credentials
6. open Cars resource → edit "test car"
7. in "Parts" relation, edit "test part"
8. "description" markdown editor field is empty
9. click on it in order to get focus
10. the "this text *should* be visible" content appears

### Reproduction repository

https://github.com/fabio-ivona/filament-markdown-issue-reproduction

### Relevant log output

_No response_

andrewdwallo:
## Description

This PR aims to resolve the issue referenced in #10926 and #5972 where the Markdown editor was not properly initializing when opened within a resource-heavy modal. The proposed changes focus on adjusting the initialization process of the markdown editor component to account for the asynchronous and dynamic nature of the modal's content.

### Summary
Users experience a problem where the Markdown editor, when loaded within a modal, fails to display its content immediately. The problem is exacerbated by the resource-intensive nature of the modal, which can delay the rendering and initialization of the editor.

The root cause appears to be linked to the timing of the Markdown editor's initialization in relation to the modal's opening sequence. Components like EasyMDE/CodeMirror, used within the editor, require accurate measurements of their container's dimensions, which can be misreported during the modal's dynamic rendering process.

### Solution Implementation

Implemented `ax-load="visible"` for the Markdown editor within modals. This approach leverages the `IntersectionObserver` to initiate the loading of the editor only when it becomes visible within the modal. The use of `ax-load="visible"` optimizes the initialization of the Markdown editor, particularly in the modal context, by delaying its loading until it is actually in view. This effectively resolves the issue of the editor content not being immediately visible upon modal opening. The rendering of the Markdown Editor is still just as fast as with using the default `ax-load` behavior. This doesn't effect Markdown Editor components that are not located within a modal, or at least it is not noticeable at all.

### Before

[](https://github.com/filamentphp/filament/assets/104294090/3d7070d5-d777-46d7-bf88-451a46d0bf4d)

### After

[](https://github.com/filamentphp/filament/assets/104294090/691e6fdf-836e-451d-b66e-3d0012338520)

### Testing

- Validated the fix by reproducing the original issue and confirming that the Markdown editor now displays its content as soon as the modal opens.
- Tested in various scenarios, including pages with and without additional Markdown editors, to ensure that the lazy-loading implementation does not adversely affect other components.
- Ensured browser compatibility and responsiveness.

### DIFF
--- a/packages/forms/resources/views/components/markdown-editor.blade.php
+++ b/packages/forms/resources/views/components/markdown-editor.blade.php
@@ -20,11 +20,7 @@
             "
         >
             <div
-                @if (FilamentView::hasSpaMode())
-                    ax-load="visible"
-                @else
-                    ax-load
-                @endif
+                ax-load="visible"
                 ax-load-src="{{ \Filament\Support\Facades\FilamentAsset::getAlpineComponentSrc('markdown-editor', 'filament/forms') }}"
                 x-data="markdownEditorFormComponent({
                             isLiveDebounced: @js($isLiveDebounced()),


danharrin:
Wow, nice! Thank you so much for figuring this out!

andrewdwallo:
@danharrin yeah of course! Sorry for the long explanation 😂 probably could have made that much shorter..

PR #10995, opened 2024-01-23
Change the return type of the getNavigationBadge() method in Resource.php

misaf:
When we use declare(strict_types=1) at the beginning of model resource files, and we want to return an integer value in the getNavigationBadge method, we encounter an error related to strict types. To resolve this issue, I added a new return type to address the strict types error.

## Description

## Code style

## Testing

## Documentation


danharrin:
There are many places in Filament where we use the string return type for data displayed in the UI, and I do not think it's practical to replace these instances with a new type every time.

However, you can use `strval()` or `(string) 0000` to cast on your side :)

PR #10996, opened 2024-01-23
RenderHook constants

margarizaldi:
## Description

Memorization is just a pain.

## Code style

## Testing

## Documentation


danharrin (changes requested):
Please convert these to use class constants, then we don't need to change the code to be able to support enums.

`const PANELS_AUTH_LOGIN_FORM_AFTER`

Then, go through all instances of where we register the render hooks within the Filament codebase, and replace them with their respective render hook

If this is too much work, I totally understand, but please close the PR

margarizaldi:
No worries... will try...

margarizaldi:
@danharrin do you mean like this?
```blade
{{ \Filament\Support\Facades\FilamentView::renderHook(\Filament\Support\RenderHook::PANELS_TENANT_MENU_BEFORE) }}

```

danharrin:
Yes, but lets use `Filament\Support\View\RenderHook`

zepfietje:
One thing I dislike about this approach is that the current string approach makes it clear how render hooks are namespaced using `.`.

Also, I think using enums will be better than constants eventually, since it allows you to list all cases for example.

margarizaldi:
> One thing I dislike about this approach is that the current string approach makes it clear how render hooks are namespaced using `.`.

I think with constants, the class can be separated into different classes, each stays in respective repos. Instead of `Filament\Support\View\RenderHook::PANELS_XXX` we can use
- `Filament\...\PanelRenderHook::XXX`
- `Filament\Tables\...\TableRenderHook::XXX`
- `Filament\Widgets\...\WidgetRenderHook::XXX`

> 
> Also, I think using enums will be better than constants eventually, since it allows you to list all cases for example.

Yes, at the first time I proposed enum but for some reason Dan suggested to use constants, which is make sense at some points. We can still list them using ReflectionClass for example.
Using enum will require another checking or more for every method. CMIIW.

I'll be happy to follow your decision after all.

danharrin:
I think that having a render hook class for each package is quite nice, like `PanelsRenderHook`, `TablesRenderHook`, and keeps the definitions out of the support package when they relate to other packages.

I love enums, I don't see any benefit of using enums here.

With regards to the loss of `-` and `.` specificity, it's a shame, but I don't think it will result in any confusion for users. I highly doubt there will be clashes in the names of hooks.

Can you please separate out the individual RenderHook classes for each package, and then update the docs to match?

margarizaldi:
> Can you please separate out the individual RenderHook classes for each package, and then update the docs to match?

Sure, that would be as simple as find & replace, but let me know your preference:
1. `Filament\Support\View\PanelsRenderHook::AUTH_LOGIN_FORM_AFTER`
2. `Filament\Support\View\RenderHooks\Panels::AUTH_LOGIN_FORM_AFTER`
3. `Filament\Support\View\RenderHooks\PanelsRenderHook::AUTH_LOGIN_FORM_AFTER`

danharrin:
`Filament\View\PanelsRenderHook`, `Filament\Forms\View\FormsRenderHook`, `Filament\Tables\View\TablesRenderHook` etc

danharrin:
Thanks!

PR #11012, opened 2024-01-23
Forms: Introduce new `ignoreRecord` argument for searchable selects

ryangjchandler:
I think this is common enough in applications to add first-party support for it – save some lines of code and tidy things up a little.

When dealing with recursive relationships, e.g. `Page` in a CMS might be the child of another `Page` via a `parent_id` column, you quite often want to prevent selecting the current record as the parent.

The new `ignoreRecord` argument handles this for you by simply excluding the current record from the result set, saving yourself some lines of code that you'd normally write to do this via `modifyQueryUsing`.

### DIFF
--- a/packages/forms/docs/03-fields/03-select.md
+++ b/packages/forms/docs/03-fields/03-select.md
@@ -214,7 +214,7 @@ This can be easily be done using the `ignoreRecord` argument.
 use Filament\Forms\Components\Select;
 
 Select::make('parent_id')
-    ->relationship('parent', 'name', ignoreRecord: true)
+    ->relationship(name: 'parent', titleAttribute: 'name', ignoreRecord: true)
 ```
 
 ### Customizing the relationship query
--- a/packages/forms/docs/03-fields/03-select.md
+++ b/packages/forms/docs/03-fields/03-select.md
@@ -208,7 +208,7 @@ Select::make('author_id')
 
 When working with recursive relationships, you will likely want to remove the current record from the set of results.
 
-This can be easily be done using the `ignoreRecord` argument.
+This can be easily be done using the `ignoreRecord` argument:
 
 ```php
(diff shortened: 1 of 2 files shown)


danharrin:
I think its nice, but I don't like the fact that we're introducing a new argument into an already crowded method signature

ryangjchandler:
> I think its nice, but I don't like the fact that we're introducing a new argument into an already crowded method signature

Tell that to the `unique()` method!

danharrin:
Yeah... can you add some docs for this?

ryangjchandler:
Yep, will do this afternoon.

ryangjchandler:
@danharrin Docs updated.

packages/forms/docs/03-fields/03-select.md
```diff
@@ -204,6 +204,19 @@ Select::make('author_id')
...
+
+When working with recursive relationships, you will likely want to remove the current record from the set of results.
+
+This can be easily be done using the `ignoreRecord` argument.
+
+```php
+use Filament\Forms\Components\Select;
+
+Select::make('parent_id')
+    ->relationship('parent', 'name', ignoreRecord: true)
```
danharrin:
```suggestion
    ->relationship(name: 'parent', titleAttribute: 'name', ignoreRecord: true)
```

packages/forms/docs/03-fields/03-select.md
```diff
@@ -204,6 +204,19 @@ Select::make('author_id')
     ->preload()
 ```
 
+### Excluding the current record
+
+When working with recursive relationships, you will likely want to remove the current record from the set of results.
+
+This can be easily be done using the `ignoreRecord` argument.
```
danharrin:
```suggestion
This can be easily be done using the `ignoreRecord` argument:
```

PR #11014, opened 2024-01-23
feat: Select relationship title attribute based on query expression

lviobio:
## Description

**The issue:**
I want to sort my users by full name, without creating virtual column in database.

**Solution:**
Allow to specify titleAttribute of forms/Select component, using query expression:

```php
use Illuminate\Database\Query\Expression;

Select::make('author_id')
    ->relationship(
        name: 'author',
        titleAttribute: new Expression('concat(first_name, \' \', last_name) as full_name'),
    )
```

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/forms/docs/03-fields/03-select.md
+++ b/packages/forms/docs/03-fields/03-select.md
@@ -235,6 +235,19 @@ Select::make('author_id')
     ->relationship(name: 'author', titleAttribute: 'full_name')
 ```
 
+Or you can use `Illuminate\Database\Query\Expression` to do the same thing:
+
+```php
+use Filament\Forms\Components\Select;
+use Illuminate\Database\Query\Expression;
+
+Select::make('author_id')
+    ->relationship(
+        name: 'author',
+        titleAttribute: new Expression('concat(first_name, \' \', last_name) as full_name'),
+    )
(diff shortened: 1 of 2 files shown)


lviobio:
I think i can remove alias existence check, and add it manually if it not present in sql, because name of alias is not used anywhere, only to correctly execute pluck, preventing errors when expression string used as object key.

And this part of code repeated three times:
`if (str_contains($relationshipTitleAttribute, '->')) {`
will be good to extract it to method too

danharrin:
Any disadvantage of auto-generating the alias, as a random string? Feels like something we could automate so that the user doesn't have to think about it

lviobio:
> Any disadvantage of auto-generating the alias, as a random string? Feels like something we could automate so that the user doesn't have to think about it

No, i will add auto-generating of alias if it's missing

danharrin:
That sounds like a good idea, right?

danharrin:
Hey @azzumed, are you planning to get this reviewable again with the alias suggestion implemented?

lviobio:
Yes, i will add new changes soon.

danharrin:
Please re-open this when the changes are implemented

PR #11067, opened 2024-01-25
feat: Allow columns to be omitted from exports by default.

morphsteve:
## Description

Currently the modal for the new export action has all columns selected by default. In the case where an export includes a column that is only occasionally needed, it would be convenient for it to be _deselected_ by default.

This PR makes a simple modification to `ExportColumn` and `CanSelectRecords`, adding a `$defaultEnabled` property and appropriate methods (there may be a more suitable name for this) to allow the default state of a column to be configured.

(No visual changes.)

## Code style

## Testing

(Changes tested manually.)

## Documentation

### DIFF
--- a/packages/actions/src/Exports/ExportColumn.php
+++ b/packages/actions/src/Exports/ExportColumn.php
@@ -58,9 +58,9 @@ class ExportColumn extends Component
         return $this;
     }
 
-    public function enabledByDefault(bool | Closure $isEnabledByDefault): static
+    public function enabledByDefault(bool | Closure $condition): static
     {
-        $this->isEnabledByDefault = $isEnabledByDefault;
+        $this->isEnabledByDefault = $condition;
 
         return $this;
     }
--- a/packages/actions/src/Exports/ExportColumn.php
+++ b/packages/actions/src/Exports/ExportColumn.php
@@ -77,7 +77,7 @@ class ExportColumn extends Component
 
     public function isEnabledByDefault(): bool
     {
-        return $this->evaluate($this->isEnabledByDefault) ?? true;
+        return (bool) $this->evaluate($this->isEnabledByDefault)
     }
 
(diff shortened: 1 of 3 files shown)


danharrin (changes requested):
Please rename this API to just `default()`, `isDefault()` and `$isDefault`.

morphsteve:
Happy to @danharrin, however `default()` collides with `HasCellState::default()`.

danharrin:
Ah, gotcha. Can we have `enabledByDefault()`, `$isEnabledByDefault` and `isEnabledByDefault()` then please?

packages/actions/src/Exports/ExportColumn.php
```diff
@@ -66,6 +75,11 @@ public function getExporter(): ?Exporter
         return $this->exporter;
     }
 
+    public function isEnabledByDefault(): bool
+    {
+        return $this->evaluate($this->isEnabledByDefault) ?? true;
```
danharrin:
```suggestion
        return (bool) $this->evaluate($this->isEnabledByDefault)
```

packages/actions/src/Exports/ExportColumn.php
```diff
@@ -56,6 +58,13 @@ public function exporter(?Exporter $exporter): static
         return $this;
     }
 
+    public function enabledByDefault(bool | Closure $isEnabledByDefault): static
+    {
+        $this->isEnabledByDefault = $isEnabledByDefault;
```
danharrin:
```suggestion
    public function enabledByDefault(bool | Closure $condition): static
    {
        $this->isEnabledByDefault = $condition;
```

morphsteve:
Done @danharrin.

danharrin (changes requested):

danharrin:
Thanks!

PR #11092, opened 2024-01-26
Adds the globalSearchDebounce method.

yannisc:
Adds the globalSearchDebounce method that allows to set the debounce for the global search.


danharrin:
Thanks!

PR #11096, opened 2024-01-26
Allow disabling of ->defaultSort()

ctoma:
## Description

Currently there is no way to disable `->defaultSort()`. This PR allows a boolean to be passed to `->defaultSort()` that allows it to be disabled. It also updates the docs to mention the default behavior.

## Explanation

Currently all tables are sorted by the primary key if no `->defaultSort()` is set, or if `null` is passed. The problem is that this interferes with global scopes, forces unnecessary duplication of sorting logic in the code, and also duplication in the SQL (see the third example).

Here are some SQL examples where I have a resource named `Category` and a global scope that sorts by the `name` column.

no `->defaultSort()` declared and global scope set:
`SELECT * FROM `categories` ORDER BY `categories`.`id` ASC, `name` ASC LIMIT 50 OFFSET 0`

`->defaultSort(null)` and global scope set:
`SELECT * FROM `categories` ORDER BY `categories`.`id` ASC, `name` ASC LIMIT 50 OFFSET 0`

`->defaultSort('name')` and global scope set:
`SELECT * FROM `categories` ORDER BY `name` ASC, `name` ASC LIMIT 50 OFFSET 0`

And now this PR allows us to have:
`->defaultSort(false)` and global scope set:
`SELECT * FROM `categories` ORDER BY `name` ASC LIMIT 50 OFFSET 0`

Note: Personally I find it strange that passing `->defaultSort(null)` still results in a sort, but of course changing that behavior would be a breaking change so I figured this would be the most suitable. I'm happy to make any changes to the PR as you see fit. Thanks!

### DIFF
--- a/packages/tables/src/Concerns/CanSortRecords.php
+++ b/packages/tables/src/Concerns/CanSortRecords.php
@@ -97,10 +97,6 @@ trait CanSortRecords
 
     protected function applyDefaultSortingToTableQuery(Builder $query): Builder
     {
-        if (! $this->getTable()->getIsDefaultSortEnabled()) {
-            return $query;
-        }
-
         $sortColumnName = $this->getTable()->getDefaultSortColumn();
         $sortDirection = ($this->getTable()->getDefaultSortDirection() ?? $this->tableSortDirection) === 'desc' ? 'desc' : 'asc';
 
--- a/packages/tables/src/Concerns/CanSortRecords.php
+++ b/packages/tables/src/Concerns/CanSortRecords.php
@@ -97,6 +97,10 @@ trait CanSortRecords
 
     protected function applyDefaultSortingToTableQuery(Builder $query): Builder
     {
+        if (!$this->getTable()->getIsDefaultSortEnabled()) {
+            return $query;
+        }
+
         $sortColumnName = $this->getTable()->getDefaultSortColumn();
         $sortDirection = ($this->getTable()->getDefaultSortDirection() ?? $this->tableSortDirection) === 'desc' ? 'desc' : 'asc';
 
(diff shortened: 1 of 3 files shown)


danharrin (changes requested):
I have an idea - instead of introducing `defaultSort(false)`, we can detect if an order has already been applied in a global scope.

To do this, you can use `if (filled($query->getQuery()->orders))` and return early before applying the primary key ordering

ctoma:
`->getQuery()->orders` won't work because it does not apply global scopes. Instead I used `->toBase()->orders` which does apply global scopes.

I'm not sure if doing it this way results in global scopes being applied twice and whether or not that could result in any side effects under certain scenarios.

Other than that everything looks good, here is the behavior now with the changes:

no `->defaultSort()` declared and no global scope set:
`SELECT * FROM `categories` ORDER BY `categories`.`id` ASC LIMIT 50 OFFSET 0`

no `->defaultSort()` declared and global scope set:
`SELECT * FROM `categories` ORDER BY `name` ASC LIMIT 50 OFFSET 0`

`->defaultSort(null)` and global scope set:
`SELECT * FROM `categories` ORDER BY `name` ASC LIMIT 50 OFFSET 0`

`->defaultSort('created_at')` and global scope set:
`SELECT * FROM `categories` ORDER BY `created_at` ASC, `name` ASC LIMIT 50 OFFSET 0`

packages/tables/src/Concerns/CanSortRecords.php
```diff
@@ -113,6 +113,10 @@ protected function applyDefaultSortingToTableQuery(Builder $query): Builder
             return $query->orderBy($sortColumnName, $sortDirection);
         }
 
+        if (filled($query->toBase()->orders)) {
```
danharrin: I think this block needs to be moved to line 128, before we apply the ID sort? Otherwise, `defaultSort()` with a custom query function doesn't work with the global scope

danharrin: moved it

danharrin:
Thanks!

PR #11097, opened 2024-01-26
Fix vertical centering of logo

stefket:
## Description

Due to the `inline-flex` class, the logo isn't centered correctly. By using the class `flex` instead, the problem is fixed.

## Before and after
before-after

### DIFF
--- a/packages/panels/resources/views/components/logo.blade.php
+++ b/packages/panels/resources/views/components/logo.blade.php
@@ -7,9 +7,9 @@
 
     $getLogoClasses = fn (bool $isDarkMode): string => \Illuminate\Support\Arr::toCssClasses([
         'fi-logo',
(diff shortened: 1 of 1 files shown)


zepfietje:
Thanks for noticing and this fix, @stefket!
Could you share some more info on why `flex` works and `inline-flex` doesn't?

stefket:
Using `flex` creates a block-level container, making vertical alignment more predictable with Flexbox properties. `inline-flex` generates an inline-level container, introducing challenges in achieving consistent vertical centering due to baseline alignment behaviors. Opting for `flex` simplifies vertical alignment, offering a more reliable solution.

zepfietje:
Thanks, @stefket!

zepfietje (approved):

PR #11100, opened 2024-01-26
feat: allow to prevent next step on wizard component

AcTiv3MineD:
## Description

Add the option for users to stop going to the next step on a wizard if needed.

## Testing

## Documentation


danharrin:
Instead of returning, people should throw the `Filament\Support\Exceptions\Halt`, which we can then catch and prevent the behaviour. Should be easier to implement :)

AcTiv3MineD:
Thanks for the feedback, I just implemented the new behaviour.

danharrin:
Thanks

PR #11113, opened 2024-01-27
Fixing BelongsToThrough Error

misaf:
After integrating \Znck\Eloquent\Relations\BelongsToThrough to support this relation when saving a form, an error arises if this relationship is utilized in the form.

The error specifically pertains to a 'Call to undefined method Znck\Eloquent\Relations\BelongsToThrough::associate().' Some individuals in the Discord community have also reported encountering this problem.

In my opinion, the best suggestion would be to examine the saveRelationshipsUsing() method. It seems that there is no method to prevent this incorrect association, as this relation does not require an 'associate' or 'sync' method, similar to belongsToMany. To resolve this issue, I added some lines in select.php, and since then, the problem has not recurred.

## Description

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -965,6 +965,10 @@ class Select extends Field implements Contracts\CanDisableOptions, Contracts\Has
         $this->saveRelationshipsUsing(static function (Select $component, Model $record, $state) {
             $relationship = $component->getRelationship();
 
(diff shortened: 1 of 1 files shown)


danharrin:
Could you please open a bug report with a reproduction repository, so I can double check this is the correct approach?

danharrin:
Okay, I've had a think about this, I think you're right. Thanks

PR #11120, opened 2024-01-28
Graceful handling of deleted export dirs

binaryfire:
You often need to automate cleanups in a multi-tenant environment to keep disk space under control. I'm running a scheduled job to delete export folders older than x days.

Currently when the user clicks a download link for an export with a deleted folder:

- The csv export returns a file of 0 bytes
- The xlsx export fails with a server error 

This PR aborts with 419 "Page expired" when the export directory has been deleted. IMHO 419 is a better choice than 404 here because it makes it clear to the user what's going on. 404 implies that something's broken, 419 implies that the export has just "expired".


binaryfire:
image

danharrin:
LGTM, thanks!

PR #11124, opened 2024-01-28
Fix - Close results container in spa mode

lguichard:
## Description

Currently, in spa mode the results container not close after click. This PR fix that.

Thanks for your amazing project 💛


danharrin:
Thanks

zepfietje:
This is not the right fix, as clicking on a result group heading now also closes the panel, which is unexpected.

zepfietje:
Fixed it in https://github.com/filamentphp/filament/commit/7555ab48906ce4595ba2795376482a15c9ba10ad.

Issue #11128, opened 2024-01-28
`KeyValue` `->reorderable()` doesn't work


sandersjj:
### Package

filament/filament

### Package Version

v3.2.16

### Laravel Version

v10.41.0

### PHP Version

PHP8.2.15

### Problem description

```
KeyValue::make('meta')
    ->reorderable()

```    
This should reorder the key/values and save them in the db as such. However when you change the order and reload the page or go back to the edit page. The ordering has not changed 

### Expected behavior

The Ordering set by the user should be respected

### Steps to reproduce

1. go to http://filament-issue.test/admin/products/1/edit
2. Save an order. 
3. Go back to http://filament-issue.test/admin/products/1/edit
4. Change the order and save. 
5. Reload and the order has not changed 

### Reproduction repository

https://github.com/sandersjj/filament-issue

danharrin:
Duplicate of #9030? Please check your Livewire version

sandersjj:
> Duplicate of #9030? Please check your Livewire version

I have checked it and I am running livewire v3.4.2.

[This Pr](https://github.com/livewire/livewire/pull/7671) has been merged in v3.4.0 so it didn't solve the issue yet.

joowdx:
@sandersjj, have you found the fix yet?

sandersjj:
> @sandersjj, have you found the fix yet?

No I ended up not using the KeyValue field.

StyxUA:
I think the problem itself is in trying to add the ordering to one-dimensional array.

[json.org](https://json.org) states that:
> An _object_ is **an unordered set** of name/value pairs.

That not only means that order of name/value pairs is (could be) not preserved, but also that two objects which differ only by order of name/value pairs are (should be) considered the same.

On the other hand, [php.net](https://www.php.net/manual/en/language.types.array.php) states that:
> An _array_ in PHP is actually **an ordered map**.

The order is preserved for string keys, but numeric keys could not be ordered inherently.  
Also, as we can see, Livewire tracks changes in JS objects with no regards to name/value pairs order, in order to minimize payload (and imho it's a good choice).

Thus, the KV component itself should either not have "ordering" feature, or change the storage mechanism to "array of objects".

For example, instead of storing as:
```{ one: "One", two: "Two" }```
it could be:
```[ { key: "one", value: "One" }, { key: "two", value: "Two" } ]```

This way, the ordering will be preserved both in frontend and backend. Here is the small demonstration:  https://wirebox.app/b/41729

[...]

PR #11129, opened 2024-01-28
fix: Handle Empty State for KeyValueEntry

### LINKED ISSUE #11127: Infolist KeyValueEntry handles state with value empty array as NULL
### Package

filament/filament

### Package Version

v3.2.16

### Laravel Version

v10.42.0

### Livewire Version

_No response_

### PHP Version

PHP 8.3.1

### Problem description

I am trying to hook up a KeyValueEntry to an Infolist view to preview the set of properties on a given Product model, and experiencing an exception in cases where the properties list is set to **`[]`** within the DB table
It looks like the component attempts to set the `state` value to `null` in case the attribute value was returned as `[]`

> The model attribute `properties` exists within the casts array inside the model.
```php
protected $casts = [
     'properties' => ['array']
];
```

I have checked the [docs](https://filamentphp.com/docs/3.x/infolists/entries/key-value) and found nothing special about optional values and empty states

Here is the exception I am getting:
image

Here is a video of the issue:

https://github.com/filamentphp/filament/assets/21292175/01b23d05-8e54-4ef7-9c13-55a7f9cdb4be

### Expected behavior

The component should set the `state` to an empty array instead of `null`.
This way this line: `vendor/filament/infolists/resources/views/components/key-value-entry.blade.php:33` won't throw an exception

**Nice to Have**
The component should have an empty state. I've tried to override the `state` and update the `null` value to `[]` by overriding the `getState()` method on the `KeyValueEntry` component
```php
public function getState(): mixed
    {
        $state = parent::getState();

        if ($state === null) {
            return [];
        }

        if (is_array($state)) {
            return collect($state);
        }

        if ($state instanceof Collection) {
            return $state;
        }

        return $state;
    }
```
But the UI does not look like nice as a Table with only the header will be shown:
image

It would be nice if the template handles a beautiful empty state
```html
@if($getState)
      {{-- Existing logic stays as is--}}
@else
       <tr>
           <td class="px-3 py-2 text-center text-sm font-medium text-gray-700 dark:text-gray-200"
                        colspan="2">
                        {{$getEmptyLabel()}}
            </td>
        </tr>
@endif
```
image

### Steps to reproduce

The steps to re-produce are as follows:

- Create a Simple model ie, `Product` with a `not null` json attribute
- Cast the model attribute to `array`
- Create a Fillament resource for the Model
- Create a View Page (Infolist Panel) for the resource and link it to the resource pages
- Remove the edit page so that any record will redirect you to the `view` page route
```php
    {
        return [
            'index' => Pages\ListProducts::route('/'),
            'create' => Pages\CreateProduct::route('/create'),
            //'edit' => Pages\EditProduct::route('/{record}/edit'),
            'view' => Pages\ViewProduct::route('/{record}'),
        ];
    }
```
- Seed a Product with `properties` equal to `[]`
- Visit the products list and click on the product with no properties
- Observe the exception

### Reproduction repository

https://github.com/jafar-albadarneh/filament-keyvalue-entry-issue

### Relevant log output

_No response_

jafar-albadarneh:
## Description

Currently, `KeyValueEntry` handles empty arrays `[]` as  `null` state very well. 
The case happens when the model attribute is optional and its value is an empty array `[]`. 
More about the issue is described [here](https://github.com/filamentphp/filament/issues/11127) 

- [ * ] Visual changes (if any) are shown using screenshots/recordings of before and after.

***Before***
image

***After***
image

## Code style

## Testing

## Documentation


sandersjj:
[This bug report ](https://github.com/filamentphp/filament/issues/11128) might be releated.

packages/infolists/src/Components/KeyValueEntry.php
```diff
@@ -15,6 +16,23 @@ class KeyValueEntry extends Entry
...
+        if ($state === null) {
+            return collect();
+        }
+
+        if (is_array($state)) {
+            return collect($state);
+        }
+
+        return $state;
+    }
```
danharrin: Could you please explain what this override is being used for? I think we should avoid it if possible

jafar-albadarneh: @danharrin Of course, for some reason the state of the casted array attribute is being converted to `null`, thus causing the @foreach to fail within the template.
I tried my best to track where the value of the state being overwritten and set to `null` when it's `[]`
The proposed override attempts to change the state to `[]` when it's null.

You might have a good point about avoiding such an override since the template now checks for the state value before attempting the loop.
```php
@if($getState())
```

danharrin: `@if (filled($state = ($getState() ?? [])))`, `@foreach ($state ...`

jafar-albadarneh: @danharrin yeah sounds better. What's your take on the empty state template change?

danharrin:
What about it? My only comment would be renaming `label` to `message`, probably

packages/infolists/src/Components/KeyValueEntry.php
danharrin: What about it? My only comment would be renaming label to message, probably

packages/infolists/docs/03-entries/06-key-value.md
```diff
@@ -51,3 +51,14 @@ use Filament\Infolists\Components\KeyValueEntry;
 KeyValueEntry::make('meta')
     ->valueLabel('Property value')
 ```
+
+## Customizing the empty value pair's label
+
+You may customize the message for empty value pairs using the `emptyMessage()` method:
```
danharrin:
```suggestion
## Customizing the empty message

You may customize the message that is displayed when the entry is empty using the `emptyMessage()` method:
```

danharrin:
Thank you for picking up on our code style and structure, and trying to make this fit into the rest of the framework as well as possible. It is appreciated!

danharrin:
Realised this is actually just placeholder() like the other entries, so renamed to that

jafar-albadarneh:
Makes complete sense

PR #11136, opened 2024-01-29
All: Add missing closure dependency resolution bindings

Voltra:
## Description

This PR provides the missing dependency resolution names for components that didn't have any, as well as provide very few additional handy (contextual) aliases.

It is a follow up of sorts for #10613 and follows the convention of using `$evaluationIdentifier` as mentioned there, instead of a `$self` that'd be shared by all.

With this PR you can, for instance, reuse the query modifier function in the badge of a tab:

```php
namespace App\Filament\Resources\PlaceResource\Pages;

use App\Enums\PlaceStatus;
use App\Filament\Resources\PlaceResource;
use App\Models\Place;
use Filament\Resources\Components\Tab;

class ListPlaces extends ListRecords
{
    public function getTabs(): array
    {
        return [
            Tab::make('All'),
            Tab::make('Active')
                ->modifyQueryUsing(fn(Builder $query) => $query->whereIn('status', [
                    PlaceStatus::CONSIDERING,
                    PlaceStatus::ACTIVELY_LOOKING,
                    PlaceStatus::OK,
                ]))
                ->badge(fn(Tab $tab) => $tab->modifyQuery(Place::query())->count()),
        ];
    }
}
```

Before that it was impossible as `Tab` was one of the components that didn't have an `$evaluationIdentifier`.

## Code style

## Testing

## Documentation

> Will edit the docs website's pages related to parameters resolution. Quite a few components didn't have an evaluation identifier, so it'll be quite a bit of work.


danharrin (changes requested):
Hi, thanks for your work on this.

I can tell how much time has been put in, however, I would really appreciate if you asked us if it would be useful so we can make sure you aren't wasting your time.

In this case, I don't think it makes sense to have a different identifier for every different type of component. I think `$component` works fine for all form components, `$tab` works for all tabs, `$group` works for all group objects, etc.

Please remove all identifier overrides and resolution method overrides, we do not need those if we are not overriding the evaluation identifiers. If a class already has an identifier in a parent class, please do not override it in this PR. The only files changed in this PR should be new identifiers added for classes that do not already have them.

Once you've cleaned it up, I'll add in a feature where you can resolve the current object based on its param type instead of name.

Voltra:
Please kindly sort out the inconsistencies with the enounced design philosophy (philosophies?) in the current core implementation before I do any more work on this enhancement.

danharrin:
What's inconsistent currently? Isn't consistency what you're trying to achieve with your PR?

Voltra:
1. `"self"` as a default fallback binding (thus it'd be `$self`) can't be accepted, it is recommended `$evaluationIdentifier` be used on a per-component basis instead
2. Many component have `$evaluationIdentifier` which does allow for customization, at the cost of only ever having one possibility per component
3. Having to rely on the generically chosen names (think fallbacks) like `component` can be cumbersome when we want a `tab` for instance. Or even for third-party plugins where changing the root component with no internal change might still induce breakage (e.g. was `component` becomes `group`)
4. Can't overly add semantic names that fit with what the component is, even adding fallbacks and aliases in an override of `resolveDefaultClosureDependencyForEvaluationByName` to keep consistency across components

I'm trying to add proper binding resolution for components that can evaluate closures but can't grant the user access to the component instance itself inside said closure.

The consistence aspect of it only comes from how it already is and the design decisions and philosophies around them.

danharrin:
Now, all parameter names can be used: 8dbdbe134658ea7451a040b232a926ad017798f0

PR #11148, opened 2024-01-29
Set table column width

divdax:
## Description

Let's set the width of a table column using `﻿->columnWidth()`.

It adds `['style' => 'width: ' . $width]` to the `$extraHeaderAttributes`

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -811,6 +811,10 @@
                         @endif
 
                         @foreach ($columns as $column)
+                            @php
+                                $columnWidth = $column->getWidth();
+                            @endphp
+
                             <x-filament-tables::header-cell
                                 :actively-sorted="$getSortColumn() === $column->getName()"
                                 :alignment="$column->getAlignment()"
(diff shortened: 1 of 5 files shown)


danharrin:
What are you trying to achieve? Does `grow(false)` work instead?

divdax:
I try to shrink the id, category, datetime columns to its minimum.

`->grow(false)`:

image


`->columnWidth('1%')`:

image

### `->columnWidth()` accepts `int | string`:

`->columnWidth(1)` -> `['style' => 'width: 1px']`
`->columnWidth(200)` -> `['style' => 'width: 200px']`
`->columnWidth('20%')` -> `['style' => 'width: 20%']`

packages/tables/resources/views/index.blade.php
```diff
@@ -822,10 +826,13 @@ class="text-sm font-semibold text-gray-950 dark:text-white"
                                     \Filament\Support\prepare_inherited_attributes($column->getExtraHeaderAttributeBag())
                                         ->class([
                                             'fi-table-header-cell-' . str($column->getName())->camel()->kebab(),
-                                            'w-full' => $column->canGrow(default: false),
+                                            'w-full' => blank($columnWidth) && $column->canGrow(default: false),
                                             '[&:not(:first-of-type)]:border-s [&:not(:last-of-type)]:border-e border-gray-200 dark:border-white/5' => $column->getGroup(),
                                             $getHiddenClasses($column),
                                         ])
+                                        ->style([
+                                            'width' => $columnWidth,
```
malzariey, line 834: @danharrin Thank you for your great effort, correct me if am wrong. Shouldn't this line (for example) be like this
                        'width:' .  $columnWidth => $columnWidth != null,

Your Advise is highly appreciated.

danharrin: I've adjusted it

adevade:
This does not work well with `ImageColumn`, since the width gets applied to the image as well as the column. The `ImageColumn` has the same methods defined as in the `HasWidth` trait.

zepfietje:
Could you create an issue for that, @adevade?

adevade:
@zepfietje Sure, and done! ✌️ https://github.com/filamentphp/filament/issues/13354

mokhosh:
@divdax This doesn't work with `IconColumn`, does it? Any idea how we can achieve this with icon columns?

CicerBro:
Am I missing something? columnWidth() seems to be a non existent function on Filament 3.2.131

danharrin:
It was renamed to `width()`

PR #11164, opened 2024-01-30
Allow relationship filters using custom keys

Skullbock:
## Description

When using relationships that do not rely on primary keys, the select filter `->relationship` method still uses the primary key to filter instead of the defined foreign key. This PR reintroduces a method that was present in 2.x that gets the related key based on the relationship type, and when present uses that one instead of the primary key to filter.

## Code style

command run, but not commited changes to tests not touched by this PR

## Testing

We tested this in `BelongsTo` relationship types.

## Documentation

### DIFF
--- a/packages/tables/src/Filters/Concerns/HasRelationship.php
+++ b/packages/tables/src/Filters/Concerns/HasRelationship.php
@@ -5,6 +5,9 @@ namespace Filament\Tables\Filters\Concerns;
 use Closure;
 use Filament\Support\Services\RelationshipJoiner;
 use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+use Illuminate\Database\Eloquent\Relations\HasManyThrough;
 use Illuminate\Database\Eloquent\Relations\Relation;
 
 trait HasRelationship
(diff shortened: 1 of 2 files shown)


danharrin:
Do the keys in the options already pluck from the correct column? Or do those need to be changed too in the Select component?

Skullbock:
In my tests with the belongs to relations the select options where fine. That was what lead me to the issue, because the query was like "accounts.id = 'other text string'" which was correctly picked up by the select

emjay64:
It seems this caused my SelectFilter to stop functioning:
```
SelectFilter::make('parent_id')
                        ->relationship('parentSku', 'sku', fn ($query, $livewire) => $query->where('parentage_level', 'parent')->where('mp', $livewire->activeTab))
                        ->label('Parent SKU')
                        ->multiple()
                        ->searchable()
                        ->preload(),
```

Here is the relationship
```
    public function parentSku(): BelongsTo
    {
        return $this->belongsTo(self::class, 'parent_id', 'id')->withDefault(['sku' => null]);
    }
```
All I get in the filtered table now is the parentsku instead of all the children as well.

danharrin:
Please open an issue with a reproduction repository and I will investigate it

emjay64:
Thank you @danharrin 

https://github.com/EmJay646/filament-select-filter

In the Listing resource when filtering by Parent Sku I get no results.

danharrin:
Please open an issue and attach that, then I can look into it

PR #11165, opened 2024-01-30
feat: If labels are being used, include the labels in the headers of the example CSV.

bzy107:
## Description
The addition of functionality for headers during the download of example CSVs in the import modal.
スクリーンショット 2024-01-31 1 19 29

|use label|not use|
|---|---|
|スクリーンショット 2024-01-31 1 18 44result:スクリーンショット 2024-01-31 1 20 28|スクリーンショット 2024-01-31 1 21 07result:スクリーンショット 2024-01-31 1 21 37|

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/actions/src/Concerns/CanImportRecords.php
+++ b/packages/actions/src/Concerns/CanImportRecords.php
@@ -304,7 +304,7 @@ trait CanImportRecords
                     }
 
                     $csv->insertOne(array_map(
-                        fn (ImportColumn $column): string => $column->getName(),
+                        fn (ImportColumn $column): string => $column->getLabel() ?? $column->getName(),
                         $columns,
                     ));
 


zhanang19:
Since this PR is changing default, I think this will make an issue for everyone who already using `ImportColumn` with label which resulting in unexpected CSV header. How about using flag like `labelAsHeader()` instead of changing the default?

danharrin:
This isn't the desired behaviour IMO - as the computer-readable column name should take priority. That way, its guaranteed that the column will be matched when the example file is uploaded. The labels do not need to be unique, but the names do. The example should be downloadable and reuploadable.

bzy107:
Thank you for your comment @danharrin.

The reasons for my submission of this PR are twofold:

- Labels are being utilized in validation.
- The characters specified in the labels are being used as headers in the exporter.

Additionally, since I am not a native English speaker, I would like the phrases specified in the labels to be used as headers. Based on these points, I believe that labels should also be used in examples.

Alternatively, should we consider implementing a flag like `labelAsHeader()` to address this?

danharrin:
How about `exampleHeader()` where you can pass in whatever you want, which is more flexible?

bzy107:
got it. I'll issue a PR if necessary.

PR #11188, opened 2024-01-31
Improving Table performance by caching eloquent queries

tiagof:
### The "problem"
In a ListRecords page, every time an Action is clicked the database query is re-executed.
(In fact this happens every time Livewire "talks" to the server)
A single action with a form seems to trigger the exact same query to the DB at least 3 times (4 times if it requires confirmation).

With high number of records and/or complex queries (e.g. where the sorting is done on a related table column) this has a significant impact in the user experience.

### Proposed solution
This PR aims to improve the performance by leveraging Laravel's cache and avoid extra queries.

#### How it works:
- this feature must be enabled explicitly by setting `protected bool $shouldCacheQueries = true;` in any Page that extends `ListRecords`
- cache key is defined by the query's SQL
- cache is tagged using the Component's id.
- queries are cached by a default of 600 seconds
- cache is flushed anytime an Action executes successfully, to reflect changes in the table.

#### Notes:
- This solution relies on a cache driver that supports tagging, such as Redis.
- If not available, queries are simply not cached.
- If enabled without this support, will throw an exception.

I'd love to get some feedback, **even if only to kick-off a discussion on how to improve such scenarios.**

## Code style

## Testing

## Documentation

- [ Not yet ] Documentation is up-to-date.

### DIFF
--- a/packages/tables/src/Concerns/CachesQueries.php
+++ b/packages/tables/src/Concerns/CachesQueries.php
@@ -1,0 +1,62 @@
+<?php
+
+namespace Filament\Tables\Concerns;
+
+use Closure;
+use Filament\Resources\Pages\ListRecords;
+use Illuminate\Support\Facades\Cache;
+
+trait CachesQueries
+{
+    protected bool $shouldCacheQueries = false;
+
+    protected int $keepCachedQueriesForSeconds = 600;
+
+    public function cacheQueries(bool $cache = true): static
+    {
+        $this->shouldCacheQueries = $cache;
+
+        return $this;
+    }
+
+    public function isCachingQueries(): bool
+    {
+        return $this->shouldCacheQueries;
+    }
+
+    public function keepCacheFor(int $seconds): static
+    {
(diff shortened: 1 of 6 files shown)


PascaleBeier:
I personally suggest leaving this in userspace since there is currently no caching strategy for Eloquent queries in filament. You do have full control to implement query caching that way.

danharrin:
Hey, I agree that this can be a very subjective implementation, so worth a plugin or something. Thanks anyway!

tiagof:
> I personally suggest leaving this in userspace since there is currently no caching strategy for Eloquent queries in filament. You do have full control to implement query caching that way.

Could you specify (high-level) how to achieve this in userspace?
Can't find anything in the docs that allows me to override the query "execution"....

danharrin:
All the methods you implement in this PR exist on the List page class, from the InteractsWithTable trait

tiagof:
I'll (re)check that.
Many thanks!!

PR #11017, opened 2024-01-23
Make entire repeater and builder item header collapsible

saade:
Repeaters and Builders currently only collapses when clicking on the collapse action or the label, which can be annoying if the item label is too short (the click area is too small)

This PR uses the same aproach as #9978 using `x-on:click.stop` on the other divs that shouldn't toggle the collapsed state of the component.

## Before
https://github.com/filamentphp/filament/assets/14329460/5aee9f19-5bc8-41cb-a1e7-998c0cad89a6

## After
https://github.com/filamentphp/filament/assets/14329460/40d72d9d-015e-4174-a7ae-b882127b1bb6

PR #11206, opened 2024-02-01
Improve UX of forms builder/repeater with `collapsible()`

austincarpenter:
When using the repeater/builder with `collapsible()` clicking the label toggles the collapsed/expanded status. However, as an end user I'd assume the label *and* empty space is clickable to toggle this, and that clicking the buttons of course would do their own thing.

301461780-7d6750e9-e391-46ae-bc9d-d81deae7e8cf

Adding a simple `flex-grow` on the element containing the label does the trick.


danharrin:
Can be handed in #11017

austincarpenter:
Apologies, didn't see a PR was in progress.

Issue #10558, opened 2023-12-29
Add `plain`/`transparent` button variant


zepfietje:

danharrin:
I think I'm going to close this one as I'm not entirely sure the use cases that would be covered that aren't already covered by other button variants in our design system.

zepfietje:
It replaces the icon button as passing just an icon to a plain button looks identical.

The plain button mainly serves as a way to have an action without attracting too much visual attention, e.g. the cancel button in modal dialogs.

danharrin:
Why would you use a plain button over a gray button for Cancel in a modal for example? I just fail to see the value here when we already have an icon button.

danharrin:
This request has been sitting inactive for a while, and it isn't something that I am going to PR to implement myself, so I am going to remove it from the roadmap. However, I would still accept a PR for this if it were ever created. Many thanks!

PR #11217, opened 2024-02-01
Add `fi-color-name` classes for named colors

tnylea:
Hey guys 👋 this is a very small PR; however, I may have a few more down the road 😊

## Description

I am trying to add [custom styles to a few of the filament components](https://filamentphp.com/docs/3.x/support/style-customization#discovering-hook-classes) and not able to make some modifications. This PR is an attempt to try make those mods possible. There are only 2 fixes in this PR.

### 1. The Button/index component

Inside of the Button component you are displaying the text/slot content in the component like so:

```
<span
    @if ($hasFormProcessingLoadingIndicator)
        x-show="! isProcessing"
    @endif
    class="{{ $labelClasses }}"
>
    {{ $slot }}
</span>
```

However, if I only wanted to display a button with an icon it doesn't look right because the parent container has a `.space-x` utility class, and the button looks like this:

jenky-button

I thought to myself... No problem. I'll just use the CSS `:empty` property to hide the `<span>` element if it is empty; however, the problem is that when the button text is empty the white-space inside that element is being rendered in the DOM like so:

```
<span class="fi-btn-label">
        
    </span>
```

The CSS `:empty` property counts whitespace as being content so therefore, it is not empty. So, if we change that line to be:

```
<span
    @if ($hasFormProcessingLoadingIndicator)
        x-show="! isProcessing"
    @endif
    class="{{ $labelClasses }}"
>{{ $slot }}</span>
```

I can add the following CSS:

```
.fi-btn-label:empty{
    @apply hidden;
}
```

And now that icon button is looking pretty good:

good-button

This was the only alternative that I could find for stylizing this correctly, but let me know if there's another way that I could be doing this.

### 2. The Badge Component

I can't seem to find a way to override the styles in the badge component for a particular color without a class that specifies the color:

```
is_string($color) ? "fi-badge-color-{$color}" : null,
```

This will allow me to add custom styles to that badge for that specific color.

I did see that this is available in the button component and unfortunately seen that there was a deprecation notice for this class, on this line here (https://github.com/filamentphp/filament/blob/3.x/packages/support/resources/views/components/button/index.blade.php#L60). Are you guys going to remove that from the button? I actually think it's a great class to have for adding custom CSS to each button color and I would love to have it in the badge component as well.

## Code style

- [x] `composer cs` command ran successfully

## Testing

## Documentation

  No additional documentation is needed for this use-case.

Thanks so much guys for all the hardwork you do 🙏

Appreciate it!

Let me know if you have any follow-ups, or have an answer about the deprecation of those classes 😊


danharrin:
Hey! For the button label, try the `label-sr-only` attribute. IMO you should still define the label, and then use this attribute, so that it is accessible.

Or you could just go with the icon-button component instead?

tnylea:
@danharrin Thanks for the comment!

Ok, I can do that with the Icon Button. That should resolve the issue for that component. How about for the button. Are you guys planning on removing this `is_string($color) ? "fi-btn-color-{$color}" : null,`

Is it possible to have this added to the badge and the icon button components:

```
is_string($color) ? "fi-badge-color-{$color}" : null,
```

```
is_string($color) ? "fi-icon-btn-color-{$color}" : null,
```

I've just updated the PR to only those two custom color classes. I really feel that these classes on these components will be a big help for the developer to override the styles for a specific badge or icon button. Let me know your thoughts 😊

Thanks so much!

zepfietje:
For v4 I'm planning to support icon-only buttons for non-icon-buttons (so regular buttons and the plain variant to be added: https://github.com/filamentphp/filament/issues/10558).
Let's not touch that part for now to prevent breaking changes.

As for the color classes, could you elaborate on your specific use case?
I've thought about adding `fi-color-purple` classes for named colors, but in many cases the current `fi-color-custom` classes should suffice.
Let me know.

tnylea:
Hey @zepfietje 

Thanks for getting back to me 😊. So, I'm trying to create a custom design system built on top of the Filament components. In this specific design system, I have 3 styles of badges. `primary`, `secondary`, and `danger`. So, I want to add some custom styles to each of those badges; however, there isn't a way to differentiate those elements based on the specific color.

Here is a screenshot of our current badge component with the current code:

cannot


As you can see, there is not a specific class that I can use to stylize our `primary` vs our `secondary` badge component.

With the changes that I have suggested, users will be able to stylize the badge based on a specific class:

can-1

can-2

We could always just create our own `badge` component, but instead want to use Filament components and override the styles with some custom CSS, but the way that it currently stands it's not possible to do so without these classes.

Let me know if that makes sense and/or you have any other questions. Really appreciate it 🙏

Look forward to hearing from you soon 😊

zepfietje (changes requested):
Thanks for clarifying, @tnylea.

Honestly I don't think the color classes should be used to create different "styles" of components.

You may want to consider creating a new API for your project like "variant" instead.

However, I'm open to adding `fi-color-primary`, `fi-color-blue`, etc. classes for named colors to all places where we currently have `fi-color-custom` and `fi-color-gray`.

Please rework this PR to add this to the entire codebase. If you don't want to work on it, feel free to close the PR. 🙂

tnylea:
Thanks for the reply @zepfietje, I'll be updating the PR to include that class for all components in the next few days 👏

Appreciate it and I'll talk to you soon!

danharrin:
Not sure if this has been covered, but I think the color classes can be generalised instead of component-specific to reduce their length

zepfietje:
Yeah they should just be `fi-color-name` as per my comment.
Some old classes may still exist, but they should be marked as deprecated, which I've probably already done for all of them.

zepfietje:
Also, I think colors from the default Tailwind palette should get a named class when used like `Color::Purple` (`fi-color-purple`).

danharrin:
> Also, I think colors from the default Tailwind palette should get a named class when used like `Color::Purple` (`fi-color-purple`).

I think that would be tricky, they are just class constants backed by arrays. If they are registered as named colors then its easy, but lets not worry about that.

zepfietje:
> I think that would be tricky, they are just class constants backed by arrays. If they are registered as named colors then its easy, but lets not worry about that.

Yeah I know, maybe something for the future.

rcostadev (commented):
is good

tnylea:
Hey guys, just wanted to keep you updated. I've made sure to add all the suggested changes. I was going to put together a storybook of most of the Filament components and test these out. I'll also see about adding a few updates to the docs to let users know about the color classes they can use.

Thanks a ton. Hoping to wrap this up tomorrow 😉

packages/support/resources/views/components/button/index.blade.php
```diff
@@ -57,8 +57,8 @@
                 'gray' => 'fi-color-gray',
                 default => 'fi-color-custom',
             },
-            // @deprecated `fi-btn-color-*` has been replaced by `fi-color-gray` and `fi-color-custom`.
-            is_string($color) ? "fi-btn-color-{$color}" : null,
+            // @deprecated `fi-btn-color-*` has been replaced with `fi-color-*`
+            is_string($color) ? "fi-color-{$color}" : null,
```
zepfietje, line 62: Old class needs to stay for BC.

packages/support/resources/views/components/dropdown/header.blade.php
```diff
@@ -38,6 +38,7 @@
                     'gray' => 'text-gray-400 dark:text-gray-500',
                     default => 'text-custom-500 dark:text-custom-400',
                 },
+                is_string($color) ? "fi-color-{$color}" : null,
```
zepfietje: Please add to parent component instead.

packages/support/resources/views/components/link.blade.php
```diff
@@ -79,6 +79,7 @@
             default => 'text-custom-600 dark:text-custom-400',
         } => ! $labelSrOnly,
         'sr-only' => $labelSrOnly,
+        is_string($color) ? "fi-color-{$color}" : null,
```
zepfietje: Should be on parent.

zepfietje (changes requested):

packages/support/resources/views/components/button/index.blade.php
tnylea: Added that back 😉

Thanks!

tnylea:
@zepfietje Just went ahead and made those updates.

Let me know if there are anything else that needs to change. I've tested on my end and these custom color classes should be working fine now with this PR allowing me to customize them better.

Really appreciate it! Talk to you soon.

zepfietje (dismissed):

zepfietje:
I've added a lot of missing `fi-color-*` classes, and made sure `fi-color-gray` didn't have duplicates in the codebase. Should be ready to be merged now.

Thanks, @tnylea!

PR #11228, opened 2024-02-02
preserve current url when switching tenant if current user can access…

eelco2k:
This PR adds functionality to preserve the current page/url when switching to another tenant. It will change the tenant.

### DIFF
--- a/packages/panels/resources/views/components/tenant-menu.blade.php
+++ b/packages/panels/resources/views/components/tenant-menu.blade.php
@@ -133,7 +133,7 @@
         <x-filament::dropdown.list>
             @foreach ($tenants as $tenant)
                 <x-filament::dropdown.list.item
-                    :href="(Route::current()->getController()->canAccess()) ? route(Route::currentRouteName(), ['tenant' => $tenant->id]) : filament()->getUrl($tenant)"
+                    :href="(Route::current()->getController()->canAccess()) ? route(Route::currentRouteName(), \Arr::collapse([Route::current()->parameters(),['tenant' => $tenant->id]])) : filament()->getUrl($tenant)"
                     :image="filament()->getTenantAvatarUrl($tenant)"
                     tag="a"
                 >
--- a/packages/panels/resources/views/components/tenant-menu.blade.php
+++ b/packages/panels/resources/views/components/tenant-menu.blade.php
@@ -27,8 +27,8 @@
 
     function hasTenantRouteSwitchingAccess(\Illuminate\Database\Eloquent\Model $tenant): bool
     {
-        $routeParams = Route::current()->parameters();
-        $routeController = Route::current()->getController();
+        $routeParams = \Illuminate\Support\Facades\Route::current()->parameters();
+        $routeController = \Illuminate\Support\Facades\Route::current()->getController();
         $tenantUrlTenantModel = app(filament()->getTenantModel());
 
         if (method_exists($routeController, 'getResource')) {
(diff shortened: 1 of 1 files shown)


binaryfire:
I'm using a combination of domain-based multi-tenancy (via separate package) + Filament's multi-tenancy (each "domain" tenant has multiple Filament tenants). Filament's current tenant switching logic works great in this scenario. Will this PR break that in any way?

eelco2k:
Which package do you use for multi tenancy based on domain? This PR only is for the Filament  Tenancy menu switcher. 
before it generates the tenant menu URL’s it checks if the page you’re on also is viewable for the other tenant. If this is the case it will generate the url fully otherwise base url

for example:

http://company.domain.com/admin/1/roles/34/edit
Is the current page I’m on. And I have 3 tenants with ID’s 1, 2 and 3
It will check if url
http://company.domain.com/admin/2/roles/34/edit
and
http://company.domain.com/admin/3/roles/34/edit
Is accessible even before you have clicked/switched in the tenant menu. 
if it will be accessible that URL will be used otherwise it will use
http://company.domain.com/admin/2
If tenant 2 has no access to roles/34/edit

This is mainly for Resources which have:
$isScopedToTenant = false;
and Filament Pages

hope this makes sense.

danharrin:
Hey, thanks for the PR. However, this isn't a behaviour we are looking to change at the moment, sorry. This is due to the inconsistent nature of some pages being accessible & some not, creating inconsistent redirect behaviour between pages. In the future, if you would like to check if a PR would be merged or not, please ask a member of the team and we can chat about it.

eelco2k:
Understandable.. but is there any method to override this? Without touching the core? 
and if not .. are there ways to build this so that it becomes overidable/extendable?

danharrin:
The only way currently would be publishing this view, I think

tonypartridge:
> The only way currently would be publishing this view, I think

Hey Dan, I actually think this PR has some foundation. Imagine a payroll app where they are handling payroll for different companies, it would massively reduce clicks. Same with any other app with many users processing certain information.

What if we add a panel config? 

->preserveUrlInTenantSwitcher() 

So then people can decide is it's right for their panel or not? Since the redirect 'generally' might go to a page without permissions but in that scenario it would trigger a 403/404 in my understanding.

tonypartridge:
For now you can override the view and this is an improved version of the code for keeping to the resource if possible too.

```php
    @if ($canSwitchTenants)
        <x-filament::dropdown.list>
            @php
                $route = \Illuminate\Support\Facades\Route::current();
                $routeParams = $route->parameters();
                $routeController = $route->getController();
                $tenantUrlTenantModel = app(filament()->getTenantModel());
                $originalTenant = filament()->getTenant();
                $hasResource = method_exists($routeController, 'getResource');
            @endphp

            @foreach ($tenants as $tenant)
                @php
                    $hasAccess = false;
                    $resourceAccessible = false;

                    if ($hasResource) {
                        filament()->setTenant($tenantUrlTenantModel->find($tenant->id), true);
                        $tenantUrlResource = app($routeController->getResource());
                        $tenantUrlModel = app($tenantUrlResource::getModel());

                        if (isset($routeParams['record'])) {
                            $record = $routeController::getResource()::resolveRecordRouteBinding($routeParams['record']);
                            $tenantUrlResourceAction = array_values(array_intersect(
                                array_keys($tenantUrlResource->getPages()),
                                explode('.', $route->getAction()['as'])
                            ))[0];
                            $recordAccess = $tenantUrlResource::can($tenantUrlResourceAction, $tenantUrlModel->find($routeParams['record']));

                            $hasAccess = !empty($record) && $recordAccess;
                        } else {
                            $hasAccess = $routeController->canAccess();
                        }

                        if (!$hasAccess) {
                            $resourceAccessible = $tenantUrlResource::canViewAny();
                        }

                        filament()->setTenant($originalTenant, true);
                    } else {
                        $hasAccess = $routeController->canAccess();
                    }

                    $redirectUrl = $hasAccess
                        ? route($route->getName(), array_merge($routeParams, ['tenant' => $tenant->id]))
                        : ($resourceAccessible
                            ? $tenantUrlResource::getUrl('index', ['tenant' => $tenant->id])
                            : url(filament()->getCurrentPanel()->getPath() . '/' . filament()->getCurrentPanel()->getTenantRoutePrefix() . '/' . $tenant->getAttributeValue('id'))
                        );
                @endphp

                        :href="$redirectUrl"
                        :image="filament()->getTenantAvatarUrl($tenant)"
                        tag="a"
                >
                    {{ filament()->getTenantName($tenant) }}
            @endforeach
        </x-filament::dropdown.list>
    @endif
```

PR #11238, opened 2024-02-03
Add action modal form selector

binaryfire:
Adds a selector for forms in modals. I needed a clean way to select all forms in a page excluding those in modals. This let's me do it via `document.querySelectorAll('form:not(.fi-action-form)');`


zepfietje:
```css
form:not(.fi-modal form)
```

binaryfire:
@zepfietje That doesn't work because `.fi-modal` is on the div inside the form, not on the form element itself.

Tried that and a lot of other combinations - it's very difficult to target modal forms without having a selector on the form itself.

binaryfire:
Never mind, figured out a workaround use `:has`

binaryfire:
Scratch that, browser support is pretty bad for `:has`. Doesn't work in Firefox on mobile, for example. So still haven't been able to find a viable solution without a class on the form itself.

I could write some complex js to get all the forms, then filter through them searching any that contain `.fi-modal` and remove them from the array, but that would be pretty bad for performance.

PR #11257, opened 2024-02-05
Modal refactor: beta

danharrin:


IgordeOliveira:
Using "vanilla" modals ( only the from --simple resources ), no issue at all, and everything working as expected

atmonshi:
~all modals now are full height in LW component~
but I recompile the assets and all goods

Screenshot 2024-02-05 at 6 53 19 PM


Screenshot 2024-02-05 at 6 59 32 PM

ashleyshenton:
On v3.2.25-beta1, modal state is not cleared when closing one modal and clicking another. For example, a table with view or edit buttons that open a modal will always show the contents of the first one you open. This also affects the create button above the table.

Another example is if you click edit on a row first, then close the modal followed by clicking the create button which also opens a modal, it will have the same pre-filled contents as edit modal you first opened.

Here's a video of the issue using the filament demo, upgraded to v3.2.25-beta1.


https://github.com/filamentphp/filament/assets/26444425/9052f442-8b07-4535-a551-67ca7b32ff1f


On the other hand, the modal animation works smoothly and scrolling still works on iOS devices on this version.

nkeena:
After upgrading I'm noticing that the `unsavedChangesAlerts` is acting a little strange with modals. Any time you close out a modal without making changes you get alert when navigating to another page.

Nested modals aren't acting as expected for me. When using the `createOptionForm` method on a select field with a relationship, closing the child modal also closes the parent. Submitting the form does not close the child modal either.

https://github.com/filamentphp/filament/assets/4267904/327a979d-4f9c-41d0-9261-de9806b634d9

danharrin:
The `unsavedChangesAlerts` behaviour with modals is deliberate, and was there before this PR. For example, if a user clicks a Delete button, doesn't see the modal, and then thinks they deleted the record and navigates away.

danharrin:
@ashleyshenton @nkeena both those issues should be fixed now on 3.x

nkeena:
> @ashleyshenton @nkeena both those issues should be fixed now on 3.x

Still doesn't seem to be working for me. Is this the right version? `3.2.25-beta1`

danharrin:
The fix isnt released yet but you could test it by pulling in a fork of 3.x locally

sandersjj:
Has this been released yet? Because if so by, by me the modals are still too big 
Screenshot 2024-02-07 at 20 46 56

danharrin:
Recompile NPM

sandersjj:
> Recompile NPM
Of course. Works like a charm. Thanks soo much.

PR #11259, opened 2024-02-05
feat: Make the ID of the record unique

bzy107:
## Description
This modification pertains to the export of intermediate tables. 
Exporting the resource page and the intermediate table may not be recommended.

In the case of the intermediate table where there are multiple records with the same key (in this image, item_id: 1 corresponds to three records), we are ensuring uniqueness to avoid multiple executions of item_id: 1 within the chunk.

スクリーンショット 2024-02-06 2 18 37

The image shows the results when the chunk is set to 2.
|before|after|
|---|---|
|スクリーンショット 2024-02-06 2 32 37|スクリーンショット 2024-02-06 2 32 16|

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/actions/src/Exports/Jobs/PrepareCsvExport.php
+++ b/packages/actions/src/Exports/Jobs/PrepareCsvExport.php
@@ -74,7 +74,7 @@ class PrepareCsvExport implements ShouldQueue
             $recordsCount = count($records);
 
             if (($totalRows + $recordsCount) > $this->export->total_rows) {
-                $records = array_slice(array_unique($records), 0, $this->export->total_rows - $totalRows);
+                $records = array_slice($records, 0, $this->export->total_rows - $totalRows);
                 $recordsCount = count($records);
             }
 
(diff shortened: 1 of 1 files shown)


danharrin (changes requested):
There are two levels of chunking - the query first gets chunked by 1000, then that gets chunked into 100 and dispatched. Currently, you are only verifying uniqueness before dispatching chunks of 100, when in reality this needs to happen at the query level, so that chunks of 1000 are unique too.

I believe you need to remove your current solution, and add a `$baseQuery->distinct($qualifiedKeyName)` around line 114?

If you would like to make these changes and test them, please go ahead. If not, please close this PR.

bzy107:
Thank you for the review. The fixes were made and everything is working smoothly.

|before| after|
|--|--|
|スクリーンショット 2024-02-06 21 39 47|スクリーンショット 2024-02-06 21 41 10|

danharrin:
Thanks

PR #11275, opened 2024-02-06
Fix: Optional MorphToSelect does not set key column to null

coleshirley:
## Description

I have an optional MorphToSelect field:
```php
MorphToSelect::make('morphable')
    ->types([
        MorphToSelect\Type::make(MorphType1::class)
            ->titleAttribute('name'),
        MorphToSelect\Type::make(MorphType2::class)
            ->titleAttribute('name'),
        MorphToSelect\Type::make(MorphType3::class)
            ->titleAttribute('name'),
    ]),
```
When I set the type field back to "Select an option" and save the `morphable_id` column is still set even though the morphable_type is set to null.

This PR adds an override of the `dehydrateState()` method that manually sets the "key" column to null if the field is not required

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/forms/src/Components/Concerns/HasState.php
+++ b/packages/forms/src/Components/Concerns/HasState.php
@@ -36,6 +36,8 @@ trait HasState
 
     protected bool | Closure $isDehydrated = true;
 
+    protected bool | Closure $isDehydratedWhenHidden = false;
+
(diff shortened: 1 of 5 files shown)


danharrin:
Thanks! I've given this a refactor and a few tests to make it an official API, instead of overriding just for this component

PR #11294, opened 2024-02-07
Panel Domain Routing

Skullbock:
## Description

Expanding on the concept of [Multi Tenancy](https://filamentphp.com/docs/3.x/panels/tenancy), i've tried to implement support for domain or subdomain based routing using [Laravel's Subdomain Roting](https://laravel.com/docs/10.x/routing#route-group-subdomain-routing).

The idea is to provide a configurable sudomain (or even domain) setup to identify the tenant.
Will post here the updated docs to better explain the new options:

When using a tenant, you might want to use domain or subdomain routing like `admin.example.com/posts` instead of a route prefix like `/admin/posts` . You can do that with the `tenantDomain()` method, alongside the `tenant()` configuration method:

```php
use Filament\Panel;

{
    return $panel
        // ...
        ->tenant(Team::class)
        ->tenantDomain('{tenant}.example.com', , slugAttribute: 'slug');
}
```

You can even use full domains for the routing:

```php
use Filament\Panel;

{
    return $panel
        // ...
        ->tenant(Team::class)
        ->tenantDomain('{tenant}', , slugAttribute: 'domain');
}
```

Remember that the model itself should also be updated to use the correct `getRouteKeyName()` method or similar, and that the model field name should contain a valid domain host, like `example.com` or `subdomain.example.com`. if you are using a full domain routing instead of the previous subdomain routing example.

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/panels/docs/11-tenancy.md
+++ b/packages/panels/docs/11-tenancy.md
@@ -677,6 +677,57 @@ public function panel(Panel $panel): Panel
 
 Before, the URL structure was `/admin/1` for tenant 1. Now, it is `/admin/team/1`.
 
+## Using a domain to identify the tenant
+
+When using a tenant, you might want to use domain or subdomain routing like `admin.example.com/posts` instead of a route prefix like `/admin/posts` . You can do that with the `tenantDomain()` method, alongside the `tenant()` configuration method:
+
+```php
+use Filament\Panel;
+
+public function panel(Panel $panel): Panel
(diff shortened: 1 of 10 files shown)


danharrin:
This is a great implementation, I didn't know it could be so simple. Thank you

Skullbock:
Thanks @danharrin for everything you do 💪🏻

PR #11298, opened 2024-02-07
DB Notifications: include user_id

davidwhicker:
Quite often I find myself needing to send back a notication/email back to the originator of the first notification.

For example: an Invoice unlock request. Admin then unlocks but needs to alert the originator or the unlock. The one requesting is not always the owner of the Invoice.

This PR stores the user_id of the originator for easy lookup when dealing with the logic of the request.

### DIFF
--- a/packages/notifications/src/Notification.php
+++ b/packages/notifications/src/Notification.php
@@ -79,6 +79,7 @@ class Notification extends ViewComponent implements Arrayable
             'title' => $this->getTitle(),
             'view' => $this->getView(),
             'viewData' => $this->getViewData(),
+            'user_id'=> $this->getAuthUser(),
         ];
     }
 
@@ -309,5 +310,11 @@ class Notification extends ViewComponent implements Arrayable
                 'The notification with the given title was sent'
             );
         }
+    }
+
+    public function getAuthUser()
+    {
+        $user = Auth::user(); // Attempt to get the authenticated user
+        return $user ? $user->id : null; // Return user ID if authenticated, otherwise return null
     }
 }


danharrin:
Hmm, I'm not going to merge this for now, but you can override the Notification class and bind a new one to the container that contains this, if you want the feature

davidwhicker:
Ah that’s a shame. I assume Filament has no way of storing the originator ID or any other data ( like native laravel db notifications ? The overriding is little beyond my limited scope. Thanks for considering.  On 8 Feb 2024, at 08:29, Dan Harrin ***@***.***> wrote:﻿
Hmm, I'm not going to merge this for now, but you can override the Notification class and bind a new one to the container that contains this, if you want the feature

tonypartridge:
So you have the notifiable_type. Are you having global notifications or notifications against different models? If so, can;t you call that model and get the user_id from the initiation?

danharrin:
I just don't know how I feel about storing the user ID in notification data when it's not really information about the content of the notification itself, it feels like you can put the user's ID in the URL of an action on the notification (as per your example)

davidwhicker:
> So you have the notifiable_type. Are you having global notifications or notifications against different models? If so, can;t you call that model and get the user_id from the initiation? 
> 
> 

In theory yes. My issue was the owner/creator of the post/invoice may not be the one requesting the unlock to edit. I suppose I could store a request_user_id in the table and alert said user on unlock.

tonypartridge:
Yeah I'm with Dan on this one. If it's related to the unlock etc, you should handle this as part of the model/observing. You are essentially using notifications for processes by the sounds of it. So on the model changing and it needing to go back to the creator, or updater etc you should have this handled on the model. Either in a series of events, or approval actions table whereby you can see each action that's happened on the model. 

But as Dan said, you can just add the column yourself and override/extend the model

davidwhicker:
> Yeah I'm with Dan on this one. If it's related to the unlock etc, you should handle this as part of the model/observing. You are essentially using notifications for processes by the sounds of it. So on the model changing and it needing to go back to the creator, or updater etc you should have this handled on the model. Either in a series of events, or approval actions table whereby you can see each action that's happened on the model. 
> 
> 
> 
> But as Dan said, you can just add the column yourself and override/extend the model 

Makes sense. Always learning. Thanks guys.

PR #11317, opened 2024-02-08
Fix FileUpload `grid` layout

### LINKED ISSUE #11031: multiple image using `->panelLayout('grid')` issue
### Package

filament/filament

### Package Version

3.2.12

### Laravel Version

10.42

### Livewire Version

3.4

### PHP Version

8.3

### Problem description

While uploading multiple image, if the last (display at first) image is very small then it breaks the UI.

You can see clearly from this video
[Screencast from 01-24-2024 08:13:25 PM.webm](https://github.com/filamentphp/filament/assets/28250303/0340c8c3-7ef0-43da-9065-212e4cbed095)

### Expected behavior

No breaking UI

### Steps to reproduce

install
upload 2 or more images including very small and big image then try reordering, put smallest image on first

### Reproduction repository

https://github.com/valpuia/filament-fileupload

### Relevant log output

_No response_

saade:
fixes #11031 

### Before

https://github.com/filamentphp/filament/assets/14329460/7cb78801-2c63-4ea9-a976-61775c1ae50c

### After

https://github.com/filamentphp/filament/assets/14329460/33431763-0765-4ab9-9487-8c1a1ab95919


saade:
Just added a new fluent configuration `->grid($columns)` for controlling the grid cols dinamically.

This method is a shortcut to
```php
->panelLayout('grid')
->itemAspectRatio(1)
```

and will default to the same number of columns we already have (2 < lg > 3) to prevent visual breaking changes.

Feel free to revert [the last commit](https://github.com/filamentphp/filament/pull/11317/commits/04128782942ed8fcf4854f7de175f58c7a321f38) if you think `->grid()` is too much. Personally i have a use case to force only two columns since we use media queries instead of container queries. 3 columns is too much for me even on lg.

danharrin:
I did revert `grid()` because I can see it being abused, and also not having control over the number of columns at each breakpoint, so we should probably hold off

danharrin:
Thanks

saade:
> I did revert `grid()` because I can see it being abused, and also not having control over the number of columns at each breakpoint, so we should probably hold off

My first iteration was considering responsive breakpoints, but it sounded too much and i removed it 🤦🏼. If we implement this in the future, we should use container queries, because the content on the form has a max width, so it does not make sense to show lets say: 3 columns on lg and 4 on xl because the form container will have the same width in those two breakpoints.

bernhardh:
This change breaks the way my upload fields (`SpatieMediaLibraryFileUpload`) are shown. 

Before this (v3.2.28) it looked like this:

image


After upgrade to latest (including this PR) it looks like this:

image

This problem still exists, when removeing everything from the field, except the necessary methods:

```php
SpatieMediaLibraryFileUpload::make("summerteaserImg")
   ->collection("summerteaserImg")
```

Adding `->imagePreviewHeight("500px")` or a large conversion `->conversion("default")` doesn't help to fix it.

Any ideas how to fix this?

Issue #11204, opened 2024-02-01
Saving relationship with an Select field and BelongsTo relation does not work properly


sstraakenbroek:
### Package

filament/filament

### Package Version

v3.2.21

### Laravel Version

v10.43.0

### PHP Version

PHP 8.3.2

### Problem description

When having a PostResource with an relation to Author or Category. When creating a new Post and enter the relation with a Select field. The relations is only saved when model is unguarded, or the relation id is added to fillable property. The associate is never saved.

In the PostResource having the following:

```php
Forms\Components\Select::make('blog_author_id')->relationship('author', 'name'),
```

The will associate the relation with:

```php
$this->saveRelationshipsUsing(static function (Select $component, Model $record, $state) {
    ...

    if (! $relationship instanceof BelongsToMany) {
        $relationship->associate($state);

        return;
    }

    ...
});
```

But when you look at the CreateRecord:

```php
public function create(bool $another = false): void
{
    ...

    try {
        ...

        $this->record = $this->handleRecordCreation($data);

        $this->form->model($this->getRecord())->saveRelationships();

        $this->callHook('afterCreate');
    } catch (Halt $exception) {
        return;
    }

    ...
}
```

The record is created AND saved first by `handleRecordCreation`. When not having the relations id in the fillable, this will throw an MySQL error when working with relations that can't be nullable. When using `Model::unguard`, then the relation ids are saved by filling the model, but not by associate.

Inside the `saveRelationships` the method `saveRelationshipsUsing` is called where the relation is associate. But after the associate there is never called a save method for saving the model with the associated relationships. Therefore the relations are not safe properly.

### Expected behavior

When creating the Post model, you need to **make** the model first, then associate the relationships and after that save the model.

Also it would be nice if you not unguard the Model inside the `AppServiceProvider::boot` method according to the [documentation](https://filamentphp.com/docs/3.x/panels/getting-started#unguarding-all-models). This will unguard it for the whole application. Better solution would be unguarded inside the `Filament/AdminPanelProvider`. For example:

```php
class AdminPanelProvider extends PanelProvider
{
    public function boot(): void
    {
        Filament::serving(function () {
            Model::unguard();
        });
    }
   ...
}
```

Another solution is to use forceFill/forceCreate on the create/update to bypass fillables in Filament core.

### Steps to reproduce

See the Reproduction repository, there you don't have any Model unguarding and no relationships id are added to the fillabled. Therefore creating a new Post with an category or author doesn't save the relationship.

[...]

PR #11329, opened 2024-02-09
Save records with BelongsTo relations using the relation name without having to set fillable in the model

johanzandstra:
## Description

When using Select fields with BelongsTo relations, we had to set the relation field (mostly a foreign key) as $fillable in the Model or had to set the whole Model as Unguarded. To protect this foreign key fields in other parts of the application we don't want to set these fields as fillable normally. 

Example:

The model `Project.php` has the relation company and only the  attribute `name` is set as fillable
```php
class Project extends Model
{

    protected $fillable = [
        'name',
    ];

    public function company(): BelongsTo
    {
        return $this->belongsTo(Company::class);
    }
}
```

The `ProjectResource` has the following form with a Select Field `company`
```php
    {
        return $form
            ->schema([
            Select::make('company_id')
                ->relationship(name: 'company', titleAttribute: 'name')
                ->preload()
                ->searchable(),
            TextInput::make('name')
                ->required(),
        ]);
    }
```

The Select field is using the relation registered in the model. In the current situation, when creating a new record using this example form and Model, would throw an Exception: `Add fillable property [company_id] to allow mass assignment on [App\Models\Project].`.

With minor changes to the CreateRecord and EditRecord, we are now able to save the records using Filament without the need to set the relation fields as $fillable.

For now, this implementation is using forceFill() to keep the implementation more simple. This will also fix the issue https://github.com/filamentphp/filament/issues/11204 as intended.

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/actions/src/CreateAction.php
+++ b/packages/actions/src/CreateAction.php
@@ -55,7 +55,7 @@ class CreateAction extends Action
                     $record = $translatableContentDriver->makeRecord($model, $data);
                 } else {
                     $record = new $model();
-                    $record->fill($data);
+                    $record->forceFill($data);
                 }
 
                 if ($relationship = $this->getRelationship()) {
--- a/packages/actions/src/EditAction.php
+++ b/packages/actions/src/EditAction.php
@@ -52,7 +52,7 @@ class EditAction extends Action
                 if ($translatableContentDriver = $livewire->makeFilamentTranslatableContentDriver()) {
                     $translatableContentDriver->updateRecord($record, $data);
                 } else {
-                    $record->update($data);
+                    $record->forceFill($data)->save();
                 }
             });
 
--- a/packages/forms/src/Components/Concerns/BelongsToModel.php
+++ b/packages/forms/src/Components/Concerns/BelongsToModel.php
@@ -34,7 +34,7 @@ trait BelongsToModel
             return;
         }
 
-        if ($this->getRecord() === null || (!$this->getRecord()?->exists && !$this->getRecord()?->isDirty())) {
+        if ($this->getRecord() === null || (! $this->getRecord()?->exists && ! $this->getRecord()?->isDirty())) {
             return;
         }
 
--- a/packages/forms/src/Components/Concerns/BelongsToModel.php
+++ b/packages/forms/src/Components/Concerns/BelongsToModel.php
@@ -26,7 +26,7 @@ trait BelongsToModel
         return $this;
     }
 
-    public function saveRelationships(bool $recordShouldExist = true): void
+    public function saveRelationships(): void
     {
         $callback = $this->saveRelationshipsUsing;
 
@@ -34,7 +34,7 @@ trait BelongsToModel
             return;
         }
 
-        if ($this->getRecord() === null || ($recordShouldExist && ! $this->getRecord()->exists)) {
+        if (! ($this->getRecord()?->exists)) {
             return;
         }
 
--- a/packages/forms/src/Components/Concerns/BelongsToModel.php
+++ b/packages/forms/src/Components/Concerns/BelongsToModel.php
@@ -26,7 +26,7 @@ trait BelongsToModel
         return $this;
     }
 
-    public function saveRelationships(): void
+    public function saveRelationships(bool $recordShouldExist = true): void
     {
         $callback = $this->saveRelationshipsUsing;
 
@@ -34,7 +34,7 @@ trait BelongsToModel
             return;
         }
 
-        if ($this->getRecord() === null || (! $this->getRecord()?->exists && ! $this->getRecord()?->isDirty())) {
+        if ($this->getRecord() === null || ($recordShouldExist && ! $this->getRecord()->exists)) {
             return;
         }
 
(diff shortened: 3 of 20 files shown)


danharrin:
To be honest, this does feel like an overly complex solution to me. I understand the problem - you don't want to add the foreign key as fillable. That's fine. But in that case, Filament should probably just switch to using `forceFill()`, which would solve the problem in all cases. I don't see any issues with this since all the data we pass to the model is validated anyway.

johanzandstra:
@danharrin forceFill() is an option too, but then we can't name the Select field using the relation name.
```            
Select::make('company')
                ->relationship(name: 'company', titleAttribute: 'name')
                ->preload()
                ->searchable(),
```
But we are forced to use the field name:
```            
Select::make('company_id')
                ->relationship(name: 'company', titleAttribute: 'name')
                ->preload()
                ->searchable(),
```

Next to that we are not using the Eloquent associate() method to associate the related model but just force the mass assignment to the model directly (what is some how bad practice in my opinion) and  that can  potentially expose the application to vulnerabilities.

I'm happy to change this implementation to a forceFill() implementation for the time being If you think that is the way to go for now btw.

danharrin:
Naming the select field after the relation name (for BelongsTo at least) is not the guidance of the documentation, the docs should use the foreign key for this reason

What vulnerabilities are there in mass assignment if we are passing validated data only?

johanzandstra:
I have updated the implementation to use the forceFill() as suggested for now.

There are no vulnerabilities in mass assignment using Filament so i don't see problems using mass assignment in this part of the application with validated data only.

johanzandstra:
@danharrin thanks for taking care of the Actions and documentation too! Looks promising!

mddickens:
I updated to 3.2.30 and my $guarded arrays are now ignored. I suspect this change is the root problem. Reverted to prior release and all works now.

danharrin:
It's deliberate, please read our discussion. We only use validated data when passing to models.

danharrin:
If you want to prevent a field from being saved, this option has been available since September 2021 as a feature of Filament - https://filamentphp.com/docs/3.x/forms/advanced#preventing-a-field-from-being-dehydrated

danharrin:
To be honest, I think I'm going to revert this in the next release, it feels a little too risky for my liking if people were relying on guarded when they should have been using `dehydrated(false)`. It's user error, but I think we need to protect them in this case

danharrin:
@johanzandstra you can override any of the methods we added in this PR / abstract to traits if you want to preserve that behaviour. Or make your foreign key fillable

mddickens:
> It's deliberate, please read our discussion. We only use validated data when passing to models.
I saw that before, but in my case, I have a number of fields in my form that have no corresponding db column and need to be present in the data array, so I used the guarded array to block them from the generated SQL. Things like toggles whose values are used when creating or editing records.  If my problem is so unusual, I must not be using Filament correctly. Wouldn't be the first time, I think

danharrin:
The idea is that everything in the resulting data array is saveable. But I think we will have to wait until v4 for this change.

mddickens:
> The idea is that everything in the resulting data array is saveable. But I think we will have to wait until v4 for this change.

I get it now, but where I got confused and perhaps others have as well is that there are more than one "data" array.  There's the one in the parameter passed into mutateFormDataBeforeCreate/Edit, etc functions and there's the data object array at the component level. The data array you send to the function has apparently been copied from the component and stripped of the fields marked with dehydrated(false) so it can be saved as is, but if you use any of the field types for options that don't correspond with a database column, you have to refer to them with $this->data([]). A fine distinction which escaped my 67 year old brain back when I programmed the form in question.

wychoong:
> Or make your foreign key fillable

This should be documented instead of applying `forceFill` for every v2~3 forms out there, it will make upgrading to v4 a difficult process (yes I know we should have tests)

Since Filament internal process of record creation and linking requires foreign key fillable, and only **validated foreign key** are filled by Filament 😉, it should be safe to just add that to $fillable. Userland code should validate the data anyhow or use the attach/associate.

johanzandstra:
> The idea is that everything in the resulting data array is saveable. But I think we will have to wait until v4 for this change.

@danharrin I understand that the change has maybe more impact then initially thought. For now we will work around this in projects with the unguarded solution.
 
The forceFill() was not my initial idea but I wanted to go for the more complex attach/associate concept as @wychoong mentioned without breaking the existing fillable logics in models as I first proposed in https://github.com/filamentphp/filament/pull/11329/commits/89536107a332f99132b5e33b6996a118a71a5afe

danharrin:
One option you can use in your panel provider is `->bootUsing(fn () => Model::unguard())` which only unguards models when the panel is being used

skills-up:
@danharrin I was looking for the _right_ place to unguard the models instead of `app::boot()`, as my app has an admin panel interface using filament, as well as a regular mobile app where `Model::unguard()` is risky. Your comment finally answered it for me.

CosmicodeStudio:
> One option you can use in your panel provider is `->bootUsing(fn () => Model::unguard())` which only unguards models when the panel is being used

This should be in the documentation, and not making developers add Model::unguard() everywhere. Some might not know the risk of doing that.

PR #11330, opened 2024-02-09
Allow to inject current import & export column instance

zhanang19:
## Description
This will allow us to access current column instance when defining column.

Here is the example on my project to use same definition between label & example header without defining it twice.

```php
ImportColumn::make('birth_date')
    ->label('Date of birth')
    ->exampleHeader(fn (ImportColumn $column) => $column->getLabel())
```

## Code style

## Testing

## Documentation


danharrin:
I may be wrong here, but in the latest version, did you double check if `ImportColumn $column` works already without defining the evaluation param like you did here?

zhanang19:
On my side, it is not working using Filament v3.2.29, here is the [reproduction repository](https://github.com/zhanang19/try-filament/tree/issue/cannot-inject-import-column-on-importer).

When downloading example CSV from import modal, I got an error `Call to a member function getRecord() on null` [check this error page for more detailed trace](https://flareapp.io/share/V7j9g6LP). After [this line](https://github.com/zhanang19/try-filament/blob/3d4ffd87b5e2d9e3de9ea5ad0db3413d97cc535c/app/Filament/Imports/UserImporter.php#L20) commented out its working well.

danharrin:
The `getRecord()` error is actually unrelated and I've just fixed it, should be working fine now

PR #11343, opened 2024-02-09
Fix using view for page headings

zepfietje:
## Description

Currently `view()` returned from `getHeading()` or `getSubheading()` gets converted when it's passed as a property from the page Blade view to the header Blade component. The approach in this PR prevents this from happening.

@danharrin, please let me know what you think of this fix and if we need to apply this to more places.

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/panels/resources/views/components/header/index.blade.php
+++ b/packages/panels/resources/views/components/header/index.blade.php
@@ -1,9 +1,13 @@
 @props([
     'actions' => [],
     'breadcrumbs' => [],
-    'heading',
-    'subheading' => null,
 ])
+
+@php
+    $heading = $attributes->get('heading');
+    $subheading = $attributes->get('subheading');
+    $attributes = $attributes->except(['heading', 'subheading']);
+@endphp
 
 <header
     {{ $attributes->class(['fi-header flex flex-col gap-4 sm:flex-row sm:items-center sm:justify-between']) }}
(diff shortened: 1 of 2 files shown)


packages/panels/resources/views/components/page/index.blade.php
```diff
@@ -32,8 +32,12 @@
             <x-filament-panels::header
                 :actions="$this->getCachedHeaderActions()"
                 :breadcrumbs="filament()->hasBreadcrumbs() ? $this->getBreadcrumbs() : []"
-                :heading="$heading"
-                :subheading="$this->getSubheading()"
+                :attributes="
```
danharrin, line 36:
```suggestion
                {{-- These are passed through in the bag otherwise Laravel converts View objects to strings prematurely --}}
                :attributes="
```

packages/panels/resources/views/components/header/index.blade.php
```diff
@@ -1,10 +1,14 @@
 @props([
     'actions' => [],
     'breadcrumbs' => [],
-    'heading',
-    'subheading' => null,
 ])
 
+@php
+    $heading = $attributes->get('heading');
```
danharrin, line 8:
```suggestion
    // These are passed through in the bag otherwise Laravel converts View objects to strings prematurely.
    $heading = $attributes->get('heading');
```

PR #11349, opened 2024-02-10
Allow checking to see if a record can be deleted before deleting

binaryfire:
Solves: https://x.com/maurozadu/status/1751351926711545863?s=20

Currently there's no clean way to check for integrity constraint violations before deleting a record. Hiding the action can lead to a confusing UX because it's not clear to the user why the delete button is there for some records and not for others. Using `before` works but it means repeatedly creating the same if / else / notification boilerplate.

This PR adds a convenient way of running a check before deletions occur, and showing a customisable notification if it isn't allowed:

```PHP
DeleteAction::make()
    ->deletable(fn ($record) => $record->teams->isEmpty())
    ->notDeletableNotificationTitle('Cannot delete user')
    ->notDeletableNotificationBody(fn ($user) => "You cannot delete {$user->name} because they are the owner of one or more teams."),
```

image

### DIFF
--- a/packages/actions/docs/07-prebuilt-actions/04-delete.md
+++ b/packages/actions/docs/07-prebuilt-actions/04-delete.md
@@ -37,6 +37,21 @@ public function table(Table $table): Table
 }
 ```
 
+## Checking if a record is allowed to be deleted
+
+In some cases, you may need to determine if a record can be deleted based on certain conditions. For example, you might want to prevent a user from being deleted if they are the owner of a "Team" model, to avoid integrity constraint violations.
+
+You can use the `deletable()` method to run pre-deletion logic. This method accepts a callback and should return `true` if the record can be deleted, or `false` otherwise.
+
+If the `deletable()` method returns `false`, the deletion process will be skipped and a notification will be displayed to the user. You can customize the title and body of this notification using the `notDeletableNotificationTitle()` and `notDeletableNotificationBody()` methods:
+
+```php
+DeleteAction::make()
+    ->deletable(fn ($record) => $record->teams->isEmpty())
+    ->notDeletableNotificationTitle('Cannot delete user')
+    ->notDeletableNotificationBody(fn ($user) => "You cannot delete {$user->name} because they are the owner of one or more teams."),
+```
+
 ## Redirecting after deleting
 
 You may set up a custom redirect when the form is submitted using the `successRedirectUrl()` method:
--- a/packages/actions/docs/07-prebuilt-actions/06-force-delete.md
+++ b/packages/actions/docs/07-prebuilt-actions/06-force-delete.md
@@ -37,6 +37,21 @@ public function table(Table $table): Table
 }
 ```
 
+## Checking if a record is allowed to be force-deleted
+
+In some cases, you may need to determine if a record can be force-deleted based on certain conditions. For example, you might want to prevent a user from being force-deleted if they are the owner of a "Team" model, to avoid integrity constraint violations.
+
+You can use the `deletable()` method to run pre-deletion logic. This method accepts a callback and should return `true` if the record can be force-deleted, or `false` otherwise.
+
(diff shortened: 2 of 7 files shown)


binaryfire:
This is still WIP. I need to make the same changes to the table action and add some documentation.

binaryfire:
Table delete actions and docs updated, all good for review.

danharrin (changes requested):
Please alter the behaviour so the notification shows instead of the modal

binaryfire:
> Please alter the behaviour so the notification shows instead of the modal

Seems like my `isDeletable()` trait method is only evaluated inside `$this->action()`. Why would that be? I tried doing the following but it always returns true when used in this location:

```PHP
if ($this->isDeletable()) {
    ray($this->isDeletable());

    $this->modalHeading(fn (): string => __('filament-actions::delete.single.modal.heading', ['label' => $this->getRecordTitle()]));

    $this->modalSubmitActionLabel(__('filament-actions::delete.single.modal.actions.delete.label'));

    $this->modalIcon(FilamentIcon::resolve('actions::delete-action.modal') ?? 'heroicon-o-trash');

    $this->requiresConfirmation();
}
```

danharrin:
Too early in the lifecycle of the class to to it in `setUp()`, that runs when the class is constructed, before the user passes any config to it. You'd need to add a `modalHidden()` condition I think, and keep the code in `action()` or `before()`. `before()` might be better tbh as then it is overridable individually, then you just call `$action->halt()` too

binaryfire:
`modalHidden()` is called before `before()` so I just used that.

It sets a property which is checked in `action()` so that `isDeletable()` is only ever evaluated once.

This would be a great place to use L11's `once()`. Might be worth refactoring to that at some point in the future. Same end result but could be more readable.

wychoong:
Maybe 

```php
 ->notDeletableNotification(fn ($notification) => …)

```

binaryfire:
Do you mean define the whole notification manually eg. `Notification::make()->`…. ? The idea here is to provide the same clean DX as `->modalHeading()`, `->modalDescription()` etc. The title and body methods both accept closures so it’s still pretty easy to customize the notification.

wychoong:
Something like this

https://github.com/filamentphp/filament/blob/e07cc0fd787c3767ab1034c9a0096490626fb29d/packages/actions/src/Concerns/CanNotify.php#L18-L39

binaryfire:
So you want to be able to use custom notifications rather than being stuck with a `danger` notification with only a customizable title and body? Yeah that's probably better. Eg. if you were doing `->deletable(fn ($record) => $record->teams->isEmpty())` it might be nice to include a notification action that takes you to that user's teams.

What do you think @danharrin?

danharrin:
At that level of customization, I do not really understand the point of the new feature when it could be accomplished manually in roughly the same number of lines

binaryfire:
Actually your point about skipping the modal was a good one - the UX is much better. You can't do that by just using `->before()`. It would get kinda messy trying to do it manually.

I think the majority of people would be happy with the current implementation. And if you need more you can still go manual.

wychoong:
> rather than being stuck with a `danger` notification with only a customizable title and body?

Yes

> it could be accomplished manually in roughly the same number of lines

But not always need to customize everything down to the color. And the api will similar with Action’s notification without implementing all notification api

danharrin:
```php
->modalHidden(fn ($record) => $record->teams->isEmpty())
->before(fn ($record, $action) => $record->teams->isEmpty() && Notification::make()
    ->title('Delete?')
    ->danger()
    ->send() && $action->halt())
```

I really don't think it's that bad tbh

danharrin:
And then users can remove the `modalHidden()` if they want to show the modal anyway, instead of forcing that behaviour, you know? Just feels like it's not enough complexity to warrant an abstraction, but you could create a macro for it, with parameters for each option!

binaryfire:
Yeah that's fair. No worries.

PR #11376, opened 2024-02-12
feature: Collapsed sidebar group dropdowns

danharrin:


edjeavons:
@danharrin Just giving you a heads up that I think this has introduced a bug in menus that _don't_ have `sidebarCollapsibleOnDesktop()` enabled. They it seems to throw the "Navigation group [GroupName] has an icon but one or more of its items also have icons" Exception every time, even then no sub-items actually have icons. 

I don't have time to setup a test repo to log a proper bug report at the moment but will try to do so later. Just thought you might like to know ASAP.

danharrin:
I'll have a look @edjeavons but please proceed with a reproduction repo in case I cannot reproduce

danharrin:
Does the new release fix it?

edjeavons:
Will check now. In the meantime I've just make a repo at https://github.com/edjeavons/filament-issue-navigation-groups which illustrates it.

edjeavons:
Yes that seems to have resolved it. Thanks for the quick response, especially outside of the proper issue tracker.

PR #11386, opened 2024-02-13
Add mask to TextInputColumn

rappasoft:
This is a super simple addition to TextInputColumn that adds support for `mask()`.

The only thing I didn't do since this is duplicate code from `TextInput` would be to extract it out to a trait like `HasMask` or something. But I can if you'd like.


danharrin:
Thanks! I usually start extracting into traits if there are 3 usages across packages, or 2 usages inside the same package

Issue #11382, opened 2024-02-12
Target class [Filament\Support\Commands\Aliases\MakeIssueCommand] does not exist.


omarherri:
### Package

filament/filament

### Package Version

v3.2.32

### Laravel Version

v10

### Livewire Version

v3.0.0

### PHP Version

PHP 8.3.0

### Problem description

After the last update, I get an error when i try to update the package saying:
Target class [Filament\Support\Commands\Aliases\MakeIssueCommand] does not exist.

### Expected behavior

Update without any bug

### Steps to reproduce

Update to the latest version.

### Reproduction repository

https://github.com/filamentphp/support

### Relevant log output

```shell
I guess after the update of filamentphp/support package. 

This was added use Filament\Support\Commands\Aliases\MakeIssueCommand as MakeIssueCommandAlias;

It should be use Filament\Support\Commands\MakeIssueCommand as MakeIssueCommandAlias;
```

TWithers:
Just ran into this as well running composer update. Rolled back for now.

CyberPunkCodes:
So sad that this is "closed" because of a repo URL.

Just purchased Mailcoach (Yes, aware it isn't your product) and this error is breaking the installation.

Prime example of why requiring a repo URL is ridiculous.

<img width="1172" alt="Screen Shot 2024-02-12 at 8 41 25 PM" src="https://github.com/filamentphp/filament/assets/9557392/c8f994a8-bdf9-45f0-8b74-8a4f21688238">

akr4m:
https://github.com/filamentphp/filament/pull/11387

danharrin:
This only seems to be an issue when you install Filament without `filament/filament`, like Mailcoach does. This issue states `filament/filament` as the package but I don't think it's accurate.

I've released a patch to move the class into the correct directory, should be fine now

And the repo URL requirement isn't going anywhere. It has not hindered this issue getting fixed, and remember that issues like this could just quote filamentphp/demo if the issue is there (which it isn't, hence why some further investigation into the fact that it only happens when you dont have the panel builder is installed would have resulted in more accurate PRs)

PR #11387, opened 2024-02-13
MakeIssueCommandAlias not exists.

akr4m:
In latest release, `MakeIssueCommand` on `Aliases` doesn't exists.

image

So, from service provider, it has been removed also. But the main `MakeIssueCommand` command is present there for `make:filament-issue`


danharrin:
https://github.com/filamentphp/filament/issues/11382#issuecomment-1940503503

PR #11408, opened 2024-02-14
There are some minor performance and structural issues.

misaf:
Instead of using the where method to retrieve media collections, we are using the filter method. This can lead to performance issues, particularly with heavy media rows.

## Description

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/spatie-laravel-media-library-plugin/src/Tables/Columns/SpatieMediaLibraryImageColumn.php
+++ b/packages/spatie-laravel-media-library-plugin/src/Tables/Columns/SpatieMediaLibraryImageColumn.php
@@ -62,6 +62,10 @@ class SpatieMediaLibraryImageColumn extends ImageColumn
     public function getImageUrl(?string $state = null): ?string
     {
         $record = $this->getRecord();
+
+        if ($this->hasRelationship($record)) {
+            $record = collect($this->getRelationshipResults($record))->first();
+        }
         
         $media = $record->getFirstMedia(filters: fn(Media $media): bool => $media->uuid === $state);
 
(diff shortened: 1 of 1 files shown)


danharrin:
N+1 issue since this is being run once for each cell, I guess?

misaf:
Before this PR, when we attempted to use this column with a specific collection name in the log file, we observed something like this:

`select * from `media` where `media`.`model_id` in (1, 3) and `media`.`model_type` = ? ["App\\Models\\Product\\Product"]
`

However, after this PR, we now have:

`select * from `media` where `media`.`model_id` in (1, 3) and `media`.`model_type` = ? and `collection_name` = ? ["App\\Models\\Product\\Product", "test-collection"]
`

The problem arises when we encounter a large number of rows on every request; all rows are retrieved, and then we filter by collection name on the PHP side. I believe it would be more efficient to have this filtering on the SQL side.

danharrin:
Before the PR, all media was eager loaded, and those collections were used to filter

Now, all media is still eager loaded, but now each cell also runs a query to filter, right? Which is a new N+1?

misaf:
Hello again danharrin,

I've found another bug in this column. We cannot use a relationship on media with more than one join.

For example, we can't use `'productPrice.currency.image'` before this PR; we could only use, for instance `'currency.image'`

in this situation because we have `'$record->getRelationValue($this->getRelationshipName());'` However, I've changed it to `'collect($this->getRelationshipResults($record))->first();'` as I'm trying to refactor it.

I believe it was better before.

danharrin:
I have rewritten this in #11548 to remove the N+1 issues etc.

danharrin:
I have reverted part of the PR that loads only the media from the collection, as it breaks if you have multiple media columns with different collections

PR #11409, opened 2024-02-14
fix the search in the table and now you can search with  multipl lang…

YacoubAl-hardari:
**There was error in searching in the table** 

لقطة شاشة 2024-01-22 172902

 **i have solved.**

i edited the function applySearchConstraintToQuery in the following path src\SpatieLaravelTranslatableContentDriver.php


danharrin:
Please share your MySQL version, I think I've seen people attempt something similar before and it causes issues

YacoubAl-hardari:
> Please share your MySQL version, I think I've seen people attempt something similar before and it causes issues

image

YacoubAl-hardari:
fix the Bug For Searching   

https://github.com/filamentphp/filament/assets/94101869/ffeb7b39-c3c9-4dc7-882b-018983085c5a

danharrin:
The idea is that the language you have selected (in LocaleSwitcher) is the one you are searching in. Currently, we do not have plans to break this behaviour, but you can always pass in a custom function to `searchable()` to change the query that is used to search

PR #11410, opened 2024-02-14
Use new `hasActualContent` method on slots

comes:
## Description

Since Laravel will shortly become a new method for slots `hasActualContent()` (see https://github.com/laravel/framework/pull/49966), I think it might be a good idea to use the framework instead of filaments own regex.

This PR will update to the new function.


zepfietje:
Could you use the actual method in the Blade components instead of our helper?
Also mark the helper as deprecated please.

Probably we also need to bump the minimum required Laravel version because your method was only introduced recently?

zepfietje (changes requested):
Thanks, @comes!

Could you use the actual method in the Blade components instead of our helper?
Also mark the helper as deprecated please.

Probably we also need to bump the minimum required Laravel version because your method was only introduced recently?

danharrin:
@zepfietje our helper allows Htmlable (HtmlStrings and View objects etc) whereas Laravel's is only available on component slots, so I have refactored it to use the new method and keep our helper

PR #10972, opened 2024-01-21
Laravel 11

danharrin:
Issues

- [x] [Service provider auto-registration](https://github.com/laravel/framework/blob/master/src/Illuminate/Foundation/Console/ProviderMakeCommand.php#L49-L52)
- [x] Tests are failing due to database setup issue

Dependencies that need updating for core packages to function:

- [x] https://github.com/AnourValar/eloquent-serialize
- [x] https://github.com/kirschbaum-development/eloquent-power-joins

Dependencies that need updating for CI to function:

- [x] https://github.com/spatie/laravel-medialibrary
- [x] https://github.com/spatie/laravel-ray
- [x] https://github.com/spatie/eloquent-sortable
- [x] https://github.com/spatie/laravel-tags
- [x] https://github.com/staudenmeir/belongs-to-through
- [x] https://github.com/staudenmeir/eloquent-has-many-deep

PR #11429, opened 2024-02-15
Laravel 11: Remove Doctrine DBAL

hafezdivandari:
## Description
Doctrine DBAL has been removed from Laravel 11.x:
* laravel/framework#48864
* [Laravel 11.x Upgrade Guide](https://laravel.com/docs/master/upgrade#doctrine-dbal-removal)

This PR removes `doctrine/dbal` from filament and uses following Laravel's native schema methods, that are stable since Laravel 10.38, instead:

* `Schema::getColumns($table)`
* `Schema::getColumnListing($table)`
* `Schema::getIndexes($table)`

PS 1: It's my first time contributing to filament, please let me know if I've missed anything, and sorry about that!
PS 2: This PR targets `laravel-11` branch (#10972), please let me know if you prefer anything else.
PS 3: I tried to keep the changes as minimum as possible, but this PR changes the methods' signatures on `packages/support/src/Commands/Concerns/CanReadModelSchemas.php`, IDK if you consider this as a breaking change or not.

## Code style

## Testing

## Documentation


packages/forms/src/Commands/Concerns/CanGenerateForms.php
```diff
@@ -89,40 +87,40 @@ protected function getResourceFormSchema(string $model): string
...
+                'money',
             ])) {
                 if ($componentData['type'] === Forms\Components\TextInput::class) {
                     $componentData['numeric'] = [];
                 }
 
-                if (filled($column->getDefault())) {
-                    $componentData['default'] = [$column->getDefault()];
+                if (filled($column['default'])) {
+                    $componentData['default'] = [$column['default']];
```
hafezdivandari: Doctrine DBAL doesn't support column's **default with expression** but Laravel does. Therefore, Doctrine always parses the default expression to get a literal value, but we don't do that on Laravel - returning the default expression itself. The best we can do here is to parse default expression manually. for example:

```php
$table->integer('foo')->default(23);
$table->string('bar')->default('hello');
```

`$column['default']` for `foo` would be:
* MySQL: `23`
* PostgreSQL: `'23'::smallint`
* SQL Server: `((23))`
* SQLite: `'23'`

`$column['default']` for `bar` would be:
* MySQL: `hello`
* MariaDB: `'hello'`
* PostgreSQL: `'hello'::character varying`
* SQL Server: `('hello')`
* SQLite: `'hello'`

packages/forms/src/Commands/Concerns/CanGenerateForms.php
```diff
@@ -89,40 +87,40 @@ protected function getResourceFormSchema(string $model): string
...
                 Forms\Components\Textarea::class,
-            ]) && ($length = $column->getLength())) {
-                $componentData['maxLength'] = [$length];
+            ]) && isset($type['length'])) {
+                $componentData['maxLength'] = [$type['length']];
 
-                if (filled($column->getDefault())) {
-                    $componentData['default'] = [$column->getDefault()];
+                if (filled($column['default'])) {
+                    $componentData['default'] = [$column['default']];
```
hafezdivandari: same as above: https://github.com/filamentphp/filament/pull/11429/files#r1491473432

packages/support/src/Commands/Concerns/CanReadModelSchemas.php
```diff
@@ -107,6 +96,61 @@ protected function guessBelongsToRelationshipTitleColumnName(AbstractAsset $colu
...
+            // 'enum' => 'enum',
+            // 'set' => 'set',
+            // 'inet', 'cidr', 'macaddr', 'macaddr8' => 'string',
+            // 'bit', 'varbit' => 'bit',
+            // 'xml' => 'xml',
+            // 'year' => 'year',
+            // 'interval' => 'interval',
+            // 'geometry', 'geometrycollection', 'linestring', 'multilinestring', 'multipoint', 'multipolygon', 'point', 'polygon', 'box', 'circle', 'line', 'lseg', 'path' => 'geometry',
+            // 'geography' => 'geography',
+            // 'tsvector', 'tsquery' => 'text',
```
hafezdivandari: These column types are also possible, I put them here for you to do the mapping as you want. For example you may want to have a select input for `enum` column type.

packages/support/src/Commands/Concerns/CanReadModelSchemas.php
```diff
@@ -107,6 +96,61 @@ protected function guessBelongsToRelationshipTitleColumnName(AbstractAsset $colu
...
+        $values = str_contains($column['type'], '(')
+            ? str_getcsv(Str::between($column['type'], '(', ')'), ',', "'")
+            : null;
+
+        $values = is_null($values) ? [] : match ($type) {
+            'string', 'char', 'binary' => ['length' => (int) $values[0]],
+            // 'enum', 'set' => ['values' => $values],
+            // 'float', 'decimal', 'double' => ['precision' => (int) $values[0], 'scale' => isset($values[1]) ? (int) $values[1] : null],
+            // 'datetime', 'timestamp', 'time', 'interval' => ['precision' => (int) $values[0]],
+            // 'geometry', 'geography' => ['subtype' => $values[0] ?? $column['type_name'] ?? null, 'srid' => isset($values[1]) ? (int) $values[1] : null],
```
hafezdivandari: Related to https://github.com/filamentphp/filament/pull/11429/files#r1491482327

You don't have any usage for these, but I put them here in case you wanted to add something in the future.

packages/forms/src/Commands/Concerns/CanGenerateForms.php
hafezdivandari: Fixed at 4295f0d

danharrin:
Thank you

sandersjj:
Hello, I am having an issue with this:

When I run this migration:
```
      Schema::table('products', function (Blueprint $table) {
            $table->integer('stock')->nullable()->change();
        });
```
I get the error that it needs `Doctrine/DBAL`

I am on Laravel 10.46

danharrin:
You can install DBAL in your project as per the Laravel documentation, ideally you should not be relying on the fact that Filament had it installed by default since it is also in the Laravel instructions

sandersjj:
I understand and I will install the package in my dev. But I find it weird. Shouldn't this PR break many more projects? Since this migration is quite up to date and is still be used in Laravel.

danharrin:
If you want to submit a PR to add it back you can, but this does feel like mostly user error

danharrin:
I'll just do it in the next release

hafezdivandari:
@sandersjj on Laravel 10, you may call `Schema::useNativeSchemaOperationsIfPossible()` method within the boot method of your `App\Providers\AppServiceProvider` class. See laravel/framework#45487

sandersjj:
@hafezdivandari thanks this works!

PR #11432, opened 2024-02-15
Add pivot data to Select and CheckboxList components

saade:
Its common to have pivot attributes on pivot tables. Currently there's no support to filling those properties when saving a Select or CheckboxList field.

**The `->pivotData()` method**

If you're using a `BelongsToMany` relationship and your pivot table has additional columns, you can use the `pivotData()` method to specify the pivot data that should be saved:

```php

Select::make('technologies')
    ->relationship(titleAttribute: 'name')
    ->pivotData([
        'is_primary' => true,
    ])
```


saade:
Just stumbled across this again, and i think this should be `->mutateRelationshipDataBeforeXUsing()`?

PR #11440, opened 2024-02-16
docs: notify on import errors & setup database notifications

ali-issa:
Enhanced documentation to cover a new notification feature for CSV import errors, including a reminder on the necessity of setting up database notifications for feature functionality.


danharrin:
Thanks, but since this is mentioned at very start of the document in the setup instructions, I would rather not add it twice

PR #11442, opened 2024-02-16
Add more CSS hook classes

PhiloNL:
## Description

To make it easier to override the styling when developing custom themes I've added two additional CSS classes to the stat widget as this will make it easier to target the label and value using a CSS selector. And also a few to the table.


packages/tables/resources/views/components/actions/cell.blade.php
```diff
@@ -1,7 +1,7 @@
 <x-filament-tables::cell
     :attributes="\Filament\Support\prepare_inherited_attributes($attributes)"
 >
-    <div class="whitespace-nowrap px-3 py-4">
+    <div class="fi-ta-action-cell whitespace-nowrap px-3 py-4">
```
zepfietje:
```suggestion
    <div class="fi-ta-actions-cell whitespace-nowrap px-3 py-4">
```

packages/tables/resources/views/components/selection/group-cell.blade.php
```diff
@@ -4,7 +4,7 @@
             ->class(['fi-ta-group-selection-cell bg-gray-50 dark:bg-white/5 w-1'])
     "
 >
-    <div class="px-3">
+    <div class="fi-ta-selection-group-cell px-3">
```
zepfietje: Already has a hook class on the root element.

packages/tables/resources/views/components/actions/cell.blade.php
```diff
@@ -1,7 +1,7 @@
 <x-filament-tables::cell
     :attributes="\Filament\Support\prepare_inherited_attributes($attributes)"
 >
-    <div class="whitespace-nowrap px-3 py-4">
+    <div class="fi-ta-actions-cell whitespace-nowrap px-3 py-4">
```
zepfietje: Please move all of these to the root element.

zepfietje (changes requested):

packages/tables/resources/views/components/actions/cell.blade.php
PhiloNL: In that case, it would not solve the ease of selecting this element. If I move the class up, and I want to change the padding on `<div class="whitespace-nowrap px-3 py-4">` I will have to use selectors like `> div` or target `.px-3{ }` in order to override the padding. What would you recommend?

zepfietje: We'll probably have to add a class to the `<div>` too eventually.
`fi-ta-actions-cell` should still be on the root element in my opinion, to be consistent. Then a new class can be added to the `<div>`.

zepfietje (approved):

PhiloNL:
Sorry about that, missed the notification about your reply. Thanks @zepfietje 🙏

PR #11445, opened 2024-02-16
Improve JS code style

soymgomez:
## Description

In this PR I have changed 2 things, on the one hand the JS files that are created when `build:js` are minified now use the extension `.min.js.`

On the other hand I have implemented StandardJS in the Workflow and I have run it once to correct the style errors.

## Code style

No PHP changes have been made

## Testing

## Documentation

I have modified the double-speak documentation of the JS file generation to refer to the new output extension.


zepfietje:
Thanks for the PR, but we already use Prettier to consistently format JS inside JS and Blade files.

PR #11447, opened 2024-02-16
add default placeholder for select and input fields depending on the Field Label 

ibrahimalkhalafalnimer:
…label
   add  default placeholder for Select and Input fields depending on the field label

## Description
All text fields will be filled with this placeholder: Enter {field label} => Field label name ---> Placeholder: Enter Name.
And all select fields will be filled with a placeholder like this: Select {field label} => Field label country---> Placeholder: Select Country.

### DIFF
--- a/packages/forms/resources/lang/en/components.php
+++ b/packages/forms/resources/lang/en/components.php
@@ -413,6 +413,8 @@ return [
 
         'search_prompt' => 'Start typing to search...',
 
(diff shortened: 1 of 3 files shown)


zepfietje:
I don't think it's a good idea to add these defaults.
Feel free to apply them to your own projects using global configuration. 🙂

PR #11448, opened 2024-02-16
Feature: XLSX format header row styling

shepard153:
## Description

This PR adds `CellStyle` class that allows users to style header row of the exported XLSX document by overriding the `getHeadingStyle()` method on the exporter class.

I'm not 100% sure about naming and implementation method so any suggestions are welcome.

Example styling of headers:
image

## Code style

## Testing

## Documentation


danharrin:
Can you refactor this to remove the new CellStyle object, then people can just use the raw object from the original package, which should be much easier for us to maintain

shepard153:
> Can you refactor this to remove the new CellStyle object, then people can just use the raw object from the original package, which should be much easier for us to maintain

Yes, of course. That was my original idea but I wasn't sure if you're ok with it. I'll add it in my next commit

shepard153:
Should we add URL to the package docs? There is more about styling and `Style` object allows to add borders too.

[https://github.com/openspout/openspout/blob/4.x/docs/documentation.md#styling](https://github.com/openspout/openspout/blob/4.x/docs/documentation.md#styling)

danharrin:
Thanks!

PR #11449, opened 2024-02-16
Add label for Input wrapper Blade components

mischasigtermans:
## Description

This pull request introduces labels for the Input Wrapper Blade component, as this is currently not supported. By clearly associating text labels with input fields, we enhance user experience, making forms more intuitive and easier to navigate.

screenshot-001250 (2024-02-16)@2x

## Code style

Simply add the new `label` attribute to the `<x-filament::input-wrapper>` component.

## Testing

## Documentation

### DIFF
--- a/packages/support/docs/09-blade-components/02-input-wrapper.md
+++ b/packages/support/docs/09-blade-components/02-input-wrapper.md
@@ -65,6 +65,21 @@ To disable the input, you must also pass the `disabled` attribute to the wrapper
 </x-filament::input.wrapper>
 ```
 
+## Adding a label above the input
+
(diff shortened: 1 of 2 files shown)


zepfietje:
The label isn't the responsibility of the input wrapper component.
Instead, it's part of the field wrapper.

I'm not sure if it's currently possible to use the field wrapper as a Blade component instead of using Filament Forms, but you could give it a try.
If it doesn't work, I'm happy to review a PR that adds support for using it standalone. 🙂

mischasigtermans:
@zepfietje In that case, wouldn't it make more sense to have the Field wrapper available as core component?

PR #11454, opened 2024-02-17
UI: Repeater and Builder header contrast on light mode

saade:
Repeater and Builder items have a good contrast in dark mode, however in light mode, it has the same background as the container.

### Before

<table>
  <tr>
    <th>Light</th>
    <th>Dark</th>
  </tr>
  <tr>
    <td>
Screenshot 2024-02-16 at 21 42 35
    </td>
    <td>       
Screenshot 2024-02-16 at 21 42 52
    </td>
  </tr>
</table>

### After
Screenshot 2024-02-16 at 21 45 51


zepfietje:
Thanks for the PR, @saade.
While I do understand the reasoning behind the change, I don't think it's a design change we can or should introduce in a minor version.

PR #11469, opened 2024-02-18
Truncate section description when it's too long

danswiser:
## Description

I've had a few cases where the description text of a section has overflowed and caused UI issues.

In my past experience, I've never had an issue truncating text by default and I can't imagine a case where truncating the description text here isn't preferred as the default.

| Before | After |
|---|---|
| before | after |

## Testing

- [x] Tested with/without `->headerActions([])`.
- [x] Tested with/without `->compact()`.
- [x] Tested with/without `->collapsed()`.

## Documentation

- [x] No changes needed.


danharrin:
Personally, I think it should wrap instead of truncating

zepfietje:
Agree, replaced by https://github.com/filamentphp/filament/pull/11613.

PR #11470, opened 2024-02-18
Added `form` attribute to Button component

mischasigtermans:
## Description

Include the `form` attribute directly on the rendered button for use outside of the `<form>` element.

## Code style

## Testing

## Documentation


danharrin:
This attribute already has behaviour associated - it should be the name of a Livewire submit action. In this case, it needs to be the ID of a form, which would conflict and be a breaking change

PR #11471, opened 2024-02-18
Use the guards defined in panel configuration

AlexisSerneels:
In the current implementation, the "web" and "auth" guards are hard-coded in the Action Export routes.
This causes a bug where we can't download exports if we have specified a custom guard in the panel configuration.
This PR fixes that.

### DIFF
--- a/packages/actions/routes/web.php
+++ b/packages/actions/routes/web.php
@@ -2,12 +2,16 @@
 
 use Filament\Actions\Exports\Http\Controllers\DownloadExport;
 use Filament\Actions\Imports\Http\Controllers\DownloadImportFailureCsv;
+use Filament\Facades\Filament;
 use Illuminate\Support\Facades\Route;
 
-Route::get('/filament/exports/{export}/download', DownloadExport::class)
-    ->name('filament.exports.download')
-    ->middleware(['web', 'auth']);
+foreach (Filament::getPanels() as $panel) {
+    /** @var \Filament\Panel $panel */
+    Route::get('/filament/exports/{export}/download', DownloadExport::class)
+        ->name('filament.exports.download')
+        ->middleware(array_merge($panel->getMiddleware(), $panel->getAuthMiddleware()));
 
-Route::get('/filament/imports/{import}/failed-rows/download', DownloadImportFailureCsv::class)
-    ->name('filament.imports.failed-rows.download')
-    ->middleware(['web', 'auth']);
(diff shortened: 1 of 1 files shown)


danharrin:
This is not going to work as a solution as the export features can work outside of panels, even without filament/filament installed

AlexisSerneels:
Any tip to fix this? I have to open an issue anyway as there is a bug with the current implementation.

danharrin:
There will have to be some way of overriding the middleware, probably by binding a new controller to the container. Please make a PR to move the middleware to [controller middleware](https://laravel.com/docs/10.x/controllers#controller-middleware)

maartenpaauw:
What about replacing the array `['web', 'auth']` with a middleware group named something like `filament.actions`? This group would naturally include the same middlewares as currently configured.

Then, within the panel service provider, we could add and remove middlewares to the group using the `$router->removeMiddlewareFromGroup(...);` and `$router->prependMiddlewareToGroup(...);` methods. I can promptly open a PR to refactor those two routes to use middleware groups. Later on, I can potentially address the initial issue regarding downloading exports as a non-default auth guard user. Currently I'm a bit short on time to fix both.

Saifallak:
> What about replacing the array `['web', 'auth']` with a middleware group named something like `filament.actions`? This group would naturally include the same middlewares as currently configured.
> 
> Then, within the panel service provider, we could add and remove middlewares to the group using the `$router->removeMiddlewareFromGroup(...);` and `$router->prependMiddlewareToGroup(...);` methods. I can promptly open a PR to refactor those two routes to use middleware groups. Later on, I can potentially address the initial issue regarding downloading exports as a non-default auth guard user. Currently I'm a bit short on time to fix both.

how would you do that? can you show example?

AlexisSerneels:
@Saifallak 
```
app(Router::class)
            ->removeMiddlewareFromGroup('filament.actions', 'auth')
            ->prependMiddlewareToGroup('filament.actions', 'web')
            ->prependMiddlewareToGroup('filament.actions', 'auth:admin');
```

Saifallak:
> ```
> app(Router::class)
>             ->removeMiddlewareFromGroup('filament.actions', 'auth')
>             ->prependMiddlewareToGroup('filament.actions', 'web')
>             ->prependMiddlewareToGroup('filament.actions', 'auth:admin');
> ```

In boot of service provider right?
Many thanks ❤️🙏

maartenpaauw:
Sorry! I didn't have time to create a PR yet, but I would do something like that indeed.

PR #11487, opened 2024-02-19
make Js asset when a .js file is passed to $panel->viteTheme()

JoshTrebilco:
## Explanation of Issue
When using a custom theme with a js file that imports the css file, running `npm run dev` will import the js file as a `link` tag instead of a `script` tag. This causes the browser to display `" rel="stylesheet" data-navigate-track />` at the top of the body.

I believe importing css via js is a valid use case because the Laravel documentation states that you can import CSS via JavaScript using only the JavaScript entry point in the `vite.config.js` file. https://laravel.com/docs/10.x/vite#loading-your-scripts-and-styles

## Expected Behavior
The js file should be imported as a `script` tag.

## Actual Behavior
The js file is imported as a `link` tag.

## TL;DR Reproduction Steps
Install Filament in a fresh Laravel project, create a custom theme, add a theme.js file next to the css file, import the css file in the js file, set the js file in the vite + panel config, run `npm run dev`, login to panel to see error.

## Full Reproduction Steps (with commit links)
- [Install Fresh Laravel](https://github.com/JoshTrebilco/filament-js-theme/commit/0ad4af94200ceddbf020e10cd916bce4a345074a)
  - `laravel new "project-name"`
- Run migrations
  - `php artisan migrate:fresh`
- Require Filament
  - `composer require filament/filament:"^3.2" -W`
- [Install Filament](https://github.com/JoshTrebilco/filament-js-theme/commit/853f4a5d073798db4e2d71b0f974ebd83ed628bb)
  - `php artisan filament:install --panels`
  - accept all defaults
- Create a user
  - `php artisan make:filament-user`
- [Create a custom theme](https://github.com/JoshTrebilco/filament-js-theme/commit/bf551b597590c2b4ec53496a7f2e25b1c277272e)
  - `php artisan make:filament-theme`
- [Configure Theme](https://github.com/JoshTrebilco/filament-js-theme/commit/fc2c27adbf6daf061fb8cda2612666af414d3435)
  - add `theme.css` to `vite.config.js` and `AdminPanelProvider.php`
- [Create `theme.js` file that imports the css file
](https://github.com/JoshTrebilco/filament-js-theme/commit/8f97cbfd8ec92a538cbf9df8d82fd7855b7c8cdb)
  - `import './app.css';`
- [Replace `theme.css` with `theme.js` in `vite.config.js` and `AdminPanelProvider.php`](https://github.com/JoshTrebilco/filament-js-theme/commit/e84b73c1e4d5542f05bf0d02ecad7a6f20b9451d)
- Run `npm run dev`
- Login to panel to see error
  - `npm run dev` imports the js file as a `link` tag instead of a `script` tag. This causes the browser to display `" rel="stylesheet" data-navigate-track />` at the top of the body.

## Proposed Solution
This PR fixes the issue by adding a check to see if the theme file is a js file and if it is, it will import it as a `script` tag. If it is not a js file, it will default to importing it as a `link` tag.

The PR is a fairly naive approach to solving the issue. When calling `->viteTheme()` it simply gets the file extension from the first file in the theme array using the `pathinfo()` php function, and then makes and returns an instance of `Filament\Support\Assets\Js` instead of `Filament\Support\Assets\Theme` if the file extension is `js`.

The existing functionality of importing a theme with a css entry point still works as expected.

### DIFF
--- a/packages/panels/src/FilamentManager.php
+++ b/packages/panels/src/FilamentManager.php
@@ -17,6 +17,7 @@ use Filament\Models\Contracts\HasTenants;
 use Filament\Navigation\MenuItem;
 use Filament\Navigation\NavigationGroup;
 use Filament\Navigation\NavigationItem;
+use Filament\Support\Assets\Js;
 use Filament\Support\Assets\Theme;
 use Filament\Support\Enums\MaxWidth;
 use Filament\Support\Facades\FilamentAsset;
@@ -425,7 +426,7 @@ class FilamentManager
         return $this->getCurrentPanel()->getTenantRegistrationUrl($parameters);
     }
 
-    public function getTheme(): Theme
+    public function getTheme(): Theme | Js
     {
         return $this->getCurrentPanel()->getTheme();
     }
--- a/packages/panels/src/Panel/Concerns/HasTheme.php
+++ b/packages/panels/src/Panel/Concerns/HasTheme.php
@@ -3,6 +3,7 @@
 namespace Filament\Panel\Concerns;
 
 use Filament\Enums\ThemeMode;
+use Filament\Support\Assets\Js;
 use Filament\Support\Assets\Theme;
 use Filament\Support\Facades\FilamentAsset;
 use Illuminate\Contracts\Support\Htmlable;
@@ -21,6 +22,8 @@ trait HasTheme
 
     protected ThemeMode $defaultThemeMode = ThemeMode::System;
 
+    protected string | array $themeExtension = 'css';
+
     /**
      * @param  string | array<string>  $theme
      */
@@ -28,6 +31,10 @@ trait HasTheme
     {
         $this->viteTheme = $theme;
         $this->viteThemeBuildDirectory = $buildDirectory;
+        $this->themeExtension = collect((array) $theme)
+            ->map(fn ($item) => pathinfo($item, PATHINFO_EXTENSION))
+            ->filter()
+            ->first() ?: $this->themeExtension;
 
         return $this;
     }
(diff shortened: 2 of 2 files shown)


danharrin:
Hi there. To be honest, I've never really thought about this as a use case. Could you explain the advantages of importing a CSS file in a JS file like this? It sounds like it brings along extra baggage for not much benefit

IMO this introduces technical debt (`getTheme()` can now return Theme or Js, which is unexpected IMO) so I am very hesitant to merge unless there is an excellent reason to use a Js file here.

JoshTrebilco:
Hi Dan, thanks for taking the time to review this. You make a great point about introducing technical debt by returning Theme or Js from `getTheme()`, I'll see if I can find a more maintainable approach.

I don't know enough about vite to provide an excellent reason why to use a js file here instead of css, however, it seems that the Laravel team uses this pattern in both [breeze](https://github.com/laravel/breeze/blob/1.x/stubs/inertia-react/vite.config.js) and [jetstream](https://github.com/laravel/jetstream/blob/4.x/stubs/inertia/vite.config.js) so there's probably a reason why they chose to do so? I'll do some research to see what the benefits / trade offs are.

The two (not excellent) reasons why it would be useful for me are:

1. I'm adding Filament to an [existing project](https://discord.com/channels/883083792112300104/1196403745773334578/1196403745773334578) and it would be nice to be consistent with the vite config 
image

2. I have some pages in my panel that I'd like to use [prism](https://prismjs.com/#basic-usage-bundlers) for code highlighting, having a Js file that I can use to import this package into my panel seems like a clean way to add custom js/css to Filament. (I haven't tried this yet, so not sure if this is a valid option)

Looking forward to discussing this further.

danharrin:
Hi, since there is no obvious benefit, I'm going to close this, I don't think there's a nice way to support it since `Theme extends Css`, not `Js`. With regards to Prism, you can absolutely import that, by registering a `Js` asset in the panel with it compiled in.

PR #11502, opened 2024-02-20
Fix download exports and failed row imports as non-default auth guard

maartenpaauw:
Currently, I have multiple panels (3 in total) that use different auth guards because they are all associated with different models. They all make use of Filament's export and import functionality. Presently, only the default auth guard allows downloads of exports and handling of failed row imports.

This pull request's code gathers all Filament panels, retrieves their associated auth guards, and passes them as arguments to the auth middleware. This change will enable users authenticated with non-default auth guards to download exports and handle failed row imports.

## Description

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/actions/routes/web.php
+++ b/packages/actions/routes/web.php
@@ -2,12 +2,20 @@
 
 use Filament\Actions\Exports\Http\Controllers\DownloadExport;
 use Filament\Actions\Imports\Http\Controllers\DownloadImportFailureCsv;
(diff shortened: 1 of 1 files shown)


danharrin:
Duplicate of #11471

maartenpaauw:
Ah, I didn't realize there was already a pull request opened proposing a similar solution. You've made a valid point about the actions package being installable separately. I'll consider an alternative solution.

PR #11457, opened 2024-02-17
Support section footer actions

mischasigtermans:
## Description

This pull request adds a new footer section to the Section Blade component. This enhancement aims to improve the structure and flexibility of content presentation, allowing developers to include additional information, links, or actions at the bottom of each section.

screenshot-001252 (2024-02-17)@2x

## Code style

Add a slot named `footer` or `footerActions` to add the footer to a section. You may control the `footerActions` alignment with the `footerActionsAlignment` attribute on the Section component.

- [x] `composer cs` command has been run.

## Testing

Tested to work with `aside`, `collapsible` and `compact`.

- [x] Changes have been tested.

## Documentation

- [x] Documentation is up-to-date.

Issue #10594, opened 2024-01-01
Possibility to use Form Builder in (chart) widgets for filters.


danharrin:
### Discussed in https://github.com/filamentphp/filament/discussions/10593

<div type='discussions-op-text'>

<sup>Originally posted by **flolanger** January  1, 2024</sup>
Like discussed in the new year event, it would be really great to have the possibility to use the form builder for filters of chart widgets (not global filters), instead of just having one select that needs to be filled with an array.

I would really like to see this feature natively, as I don’t need or want to use apex charts and just stay with chart js.

Thank you for your work 🙏</div>

PR #11518, opened 2024-02-21
Improve section header

mischasigtermans:
## Description

This pull request improves the header of the Section Blade component.

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/support/docs/09-blade-components/02-section.md
+++ b/packages/support/docs/09-blade-components/02-section.md
@@ -111,49 +111,6 @@ You may render additional content at the end of the header, next to the heading 
 </x-filament::section>
 ```
 
-## Adding a footer to the section
-
-You can add a footer to the section by using the `footer` slot:
-
-```blade
-<x-filament::section>
-    {{-- Content --}}
-
-    <x-slot name="footer">
-        {{-- Footer content --}}
-    </x-slot>
-</x-filament::section>
-```
-
-Alternatively, you can add actions into the footer by using the `footerActions` slot:
-
-```blade
-<x-filament::section>
-    {{-- Content --}}
-
-    <x-slot name="footerActions">
-        {{-- Footer content actions --}}
-    </x-slot>
-</x-filament::section>
-```
-
-### Changing the footer actions alignment
-
-By default, footer action's will be aligned to the start. If you wish to change the alignment of the footer actions, you can use the `footerActionsAlignment` attribute and pass it `start`, `center` or `end`:
-
-```blade
-<x-filament::section
-    footerActionsAlignment="end"
->
-    {{-- Content --}}
-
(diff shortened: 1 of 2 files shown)


danharrin:
Please explain each of these changes with reasoning / what didnt work before

mischasigtermans:
Sure thing.

Better alignment for icons when not using a description:
screenshot-001334 (2024-02-28)@2x

Make sure that `headerActions` won't trigger `collapsible`:
screenshot-001335 (2024-02-28)

Beter use of space for mobile:
screenshot-001336 (2024-02-28)@2x

`headerActions` still align left when there is no heading:
screenshot-001337 (2024-02-28)@2x

Compact still works:
screenshot-001338 (2024-02-28)@2x

Aside still works:
screenshot-001339 (2024-02-28)@2x

It also works with the new footer from PR #11457.

danharrin:
Can you take a picture of all those cases again but with collapsible enabled?

mischasigtermans:
Just a heading:
screenshot-001340 (2024-02-29)

With description:
screenshot-001341 (2024-02-29)

Header action collapse trigger fix:
screenshot-001342 (2024-02-29)

Mobile:
screenshot-001345 (2024-02-29)

Works with header end:
screenshot-001346 (2024-02-29)

Keeps aligned to the right on mobile where previously it didn't:
screenshot-001347 (2024-02-29)@2x

Skipped aside. It behaves the same as before, but probably not a combo that will be used.

danharrin:
How about mobile with header actions and collapsible as per the original set of the screenshots?

mischasigtermans:
Works the same as with `headerEnd`. If you want to check the different options, please checkout the PR for yourself as that's probably faster.

danharrin:
I'm asking for that specific example since it is main problematic issue with this component that we need to keep fixing

mischasigtermans:
It's exactly like this, but then headerActions. You can test them on the dev environment I sent you.

image

danharrin:
Zep - for your review. I am not a fan of moving the actions to be inline with the heading and description on mobile, as it would result in squashing when the heading or description is over a certain length, or when there is more than one action or an action with a long label. Personally I think they should be on their own line when on mobile.

I also think there is scope to utilize a container query to decide whether they are inline or not, instead of a responsive breakpoint

zepfietje:
I'm going to hold off on merging this.
As @danharrin explained above, actions are rendered below the heading on mobile on purpose.
This PR brings back an issue that Dan fixed some time ago.

The header end slot should probably be removed in v4 and replaced by the header actions slot, since the header end is only used for select filters on widgets. Essentially that's a select action, and an option for filter button will be added with https://github.com/filamentphp/filament/issues/10594.

Feel free to address the `x-on:click.stop` issue in a separate PR.

PR #11532, opened 2024-02-22
fix: update pgsql generate_search_column_expression

mgkimsal:
## Description

PostgreSQL column names are case sensitive, and need to be quoted for certain behaviours.  The searchable() functionality currently handles case-sensitivity in the column's *value*, but not case-sensitivity in the column's name.

Most of the regular type query behaviour already quote column names, but the 'generate_search_column_expression' method does not.

If you've created mixed-case names via a migration file, the field will be created with surrounding double-quote marks

```php
$table->string('MixedCase')->nullable();
```

will create a postgres field name of "MixedCase" not MixedCase.

Regular Eloquent usage typically quotes the column name

```php
$c = Company::where('MixedCase', "dave")->get()
```

works fine.

The searchable method uses generate_search_column_expression which has

```php
$column = match ($driverName) {
    'pgsql' => "{$column}::text",
    default => $column,
};
```

For pgsql connections, we'd get MixedCase::text which then gets 'lower()'ed later on.  But this query breaks because MixedCase doesn't exist as a column.  "MixedCase" does.

Changing the column naming behaviour to
```php
$column = match ($driverName) {
    'pgsql' =>  "\"{$column}\"::text",
    default => $column,
};
```
addresses that.

## Code style

No style changes.

'composer cs' shows 'command "cs" not defined.

## Testing

Making a mixedCase-named column is now searchable, whereas before, it triggered a breaking error.

## Documentation

No doc changes required.

### DIFF
--- a/packages/support/src/helpers.php
+++ b/packages/support/src/helpers.php
@@ -167,8 +167,15 @@ if (! function_exists('Filament\Support\generate_search_column_expression')) {
     {
         $driverName = $databaseConnection->getDriverName();
 
+        if (Str::lower($column) !== $column) {
+            $column = match ($driverName) {
+                'pgsql' => (string) str($column)->wrap('"'),
+                default => $column,
+            };
+        }
+
         $column = match ($driverName) {
-            'pgsql' => "\"{$column}\"::text",
+            'pgsql' => "{$column}::text",
             default => $column,
         };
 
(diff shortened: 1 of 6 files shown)


danharrin:
As far as I've read, this is nothing to do with lower(), it's just when you're using uppercase characters in the column name

danharrin:
I've updated it to only happen when you are using an uppercase character

mgkimsal:
Thanks @danharrin.   I knew there would likely be a more comprehensive way to address this, but don't have the depth of knowledge on this that you do (obviously).  Thank you for taking the issue seriously and addressing so promptly (and in a cleaner way!)

PR #11534, opened 2024-02-22
Update EditProfile handleRecordUpdate() to verify new email after changing It

devhoussam1998:
## Description

Update EditProfile handleRecordUpdate() to verify new email after changing It
https://www.youtube.com/watch?v=ZUW5PftAuuk

## Code style

## Testing


danharrin:
Judging by the PHPStan issues, this has not been manually tested (Undefined variable: $user). Additionally, this is a breaking change. Also, you can override this behaviour by overriding this method in your own class that extends EditProfile.

PR #11540, opened 2024-02-23
English grammar corrections

danswiser:
## Description

Just correcting some English grammar. **Forgot your password** might be the proper grammar but **Forgot password** is the most common text used for a password reset link.

## Code style

- [x] No changes

## Testing

- [x] No testing needed

## Documentation

- [x] No changes


zepfietje:
Just checked what Google and Airbnb do (just some random big platforms) and they both do use the same wording as suggested in this PR.

PR #11543, opened 2024-02-23
Add troubleshooting to 'Adding a form to a Livewire component'

vjanssens:
## Description

This PR adds a troubleshooting section to 'Adding a form to a Livewire component' with a problem I faced recently. When installing Filament in Laravel, you usually don't have to add the `@filemanetStyle` and `@filamentScripts` to your layout, unless you add forms to your front-end application. Therefore I think it would be good to point users to the instructions once again.

## Code style

## Testing

## Documentation


danharrin:
I understand you don't need to do this for the admin panel, but there are many other instructions to follow other than adding these directives to your layout - such as setting up Tailwind with the preset. I hope most people will read the installation instructions of the package they use before they read pages in that package documentation

PR #11578, opened 2024-02-26
Update 01-getting-started.md in Infolists

dissto:
Add missing `static` keyword

## Description

The docs were missing the static keyword which would result in `Cannot make static method ... non static`

## Code style

## Testing

## Documentation


danharrin:
This guide also applies to infolists outside of resources where a static function isnt nessecary

PR #11595, opened 2024-02-27
Use Testbench configuration feature

crynobone:
This fixes Laravel 11 tests


tests/src/TestCase.php
```diff
@@ -49,12 +51,7 @@ protected function getPackageProviders($app): array
         ];
     }
 
-    protected function defineDatabaseMigrations(): void
-    {
-        $this->loadMigrationsFrom(__DIR__ . '/../database/migrations');
```
crynobone, line 54: Using this causes Laravel to migrate the database twice, using `RefreshDatabase` trait (which starts a database transaction) and explicitly via `loadMigrationsFrom()`. 

Using `testbench.yaml` configurations only registers the paths to `Migrator` and runs the migration once via `RefreshDatabase` eliminating the issue introduced by caching PDO instances between tests: https://github.com/laravel/framework/pull/47912

testbench.yaml
```diff
@@ -0,0 +1,5 @@
+migrations:
+  - tests/database/migrations
+
+workbench:
+  install: true
```
crynobone, line 5: Having `install: true` and `migrations` other than `false` will migrate the default laravel migrations. Which means we no longer requires `create_password_reset_tokens_table` and `create_users_table`.

danharrin:
Very helpful, thank you very much!

Issue #11521, opened 2024-02-22
Improve Choices.js select field appearance while script is loading


binaryfire:
### Package

filament/filament

### Package Version

3.2.35

### Laravel Version

10.3.3

### Livewire Version

3.4.6

### PHP Version

8.3

### Problem description

There's a pretty jarring visual glitch when using non-native select fields. This is what is shown while the field is loading:

<img width="909" alt="image" src="https://github.com/filamentphp/filament/assets/5361908/5224dbd9-f0d7-4b38-9443-6a0ec5cd1e88">

It's very noticable for people on slow connections. See this video: https://www.youtube.com/watch?v=Ruax9OpgDmU

I understand that the js is lazy loaded (and that's great) but I think there should be a better placeholder while the field is still loading. Currently UI looks broken for several seconds for people on slow connections. It's also an issue when the server is far away from the user (eg. Australia -> Germany) regardless of connection speed. 

The large white box with the dropdown arrow that shows while the field is loading isn't clickable either, which is even more confusing for users.

### Expected behavior

Non-native select fields should have some kind of placeholder while Choices.js is loading to prevent them from looking broken.

### Steps to reproduce

1. Launch the reproduction repository
2. Open the browser and throttle the connection to 2G
3. Clear the browser cache, then open the create product page

The select will look broken for a few seconds.

### Reproduction repository

https://github.com/binaryfire/filament-reproduction-repo

zepfietje:
Would indeed be nice to improve the appearance of the field when Choices.js hasn't loaded yet.
For us it's a low priority improvement to make though, so feel free to attempt a PR that addresses the issue already.

Happy to review. 🙂

valentin-morice:
Adding this block of code to the default HTML select input displayed before Alpine async-loads its component gets it to look more decent.

```
{{
       $attributes->class([
            "w-full rounded-lg border-none bg-white/0",
       ])
}}
```

![image](https://github.com/filamentphp/filament/assets/100000204/25229f9c-6b85-4d9b-b491-3304f0ea6b03)

![image](https://github.com/filamentphp/filament/assets/100000204/c45da099-3914-4a98-8574-cdaf9524e4ac)

As far as I can see, there's no hard-coded classes in the view, so I'm not sure it's good practice. Let me know what you think, if it looks acceptable I'll make a proper PR.

danharrin:
#11631

binaryfire:
Hi @valentin-morice 

[...]

PR #11631, opened 2024-03-01
add styling to default html select

valentin-morice:
## Description

Fix for issue #11521
The select shown before the Alpine component loads isn't styled properly.

306832743-5224dbd9-f0d7-4b38-9443-6a0ec5cd1e88
<i>Before</i>

image
<i>After</i>

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/forms/resources/views/components/select.blade.php
+++ b/packages/forms/resources/views/components/select.blade.php
@@ -159,6 +159,11 @@
                                 'multiple' => $isMultiple(),
                             ], escape: false)
                     }}
+                    {{
+                        $attributes->class([
+                            "w-full rounded-lg border-none bg-white/0 h-9",
+                        ])
+                    }}
                 ></select>
             </div>
         @endif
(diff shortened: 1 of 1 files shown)


packages/forms/resources/views/components/select.blade.php
```diff
@@ -159,6 +159,11 @@
                                 'multiple' => $isMultiple(),
                             ], escape: false)
                     }}
+                    {{
+                        $attributes->class([
+                            "w-full rounded-lg border-none bg-white/0 h-9",
```
danharrin: Can that `bg-white/0` be replaced with `bg-transparent`?

danharrin (changes requested):

packages/forms/resources/views/components/select.blade.php
danharrin: tested and it seems to be fine

PR #11648, opened 2024-03-02
Allow Laravel 11 dev version

mokhosh:
## Description

Currently when you run ` php artisan filament:install --panels` on a Laravel 11 project, you get this

```bash
 file_get_contents(/Users/mokhosh/Sites/drbanoo/config/app.php): Failed to open stream: No such file or directory

  at vendor/filament/support/src/Commands/Concerns/CanGeneratePanels.php:65
```

This is because the code checks for the version being higher than `11.0` which is actually considered higher than `11.x-dev` by php's `version_compare` function, because according to php documentation:

> The function first replaces _, - and + with a dot . in the version strings and also inserts dots . before and after any non number so that for example '4.3.2RC1' becomes '4.3.2.RC.1'. Then it compares the parts starting from left to right. If a part contains special version strings these are handled in the following order: any string not found in this list < dev < alpha = a < beta = b < RC = rc < # < pl = p. This way not only versions with different levels like '4.1' and '4.1.2' can be compared but also any PHP specific version containing development state.

This PR compares the current version against `11.x-dev` which should have been the case to begin with.

## Testing

- [x] Changes have been tested manually, and it works on a dev Laravel 11 app.

### DIFF
--- a/packages/support/src/Commands/Concerns/CanGeneratePanels.php
+++ b/packages/support/src/Commands/Concerns/CanGeneratePanels.php
@@ -50,7 +50,7 @@ trait CanGeneratePanels
             ]);
         }
 
-        $isLaravel11OrHigherWithBootstrapProvidersFile = version_compare(App::version(), '11.0', '>=') &&
+        $isLaravel11OrHigherWithBootstrapProvidersFile = version_compare(App::version(), '11.x-dev', '>=') &&
             /** @phpstan-ignore-next-line */
             file_exists($bootstrapProvidersPath = App::getBootstrapProvidersPath());
 


danharrin:
1) I don't really want code existing that is only relevant while the stable version doesn't exist, that feels like clutter?
2) If you are using a dev version you need to be aware that some things wont work exactly right, and this is one of them. Its very easy to fix, just register the service provider
3) This PR (I think) breaks Filament when the stable version is released since the stable version constraint doesn't include -dev

mokhosh:
@danharrin fair enough. 

although it won't break, since these are all standard version codes and php's `compare_version` accounts for all of them.

PR #11661, opened 2024-03-03
filament/upgrade: Make rector rules compatible with required rector version

dbohn:
## Description

When upgrading a filament v2 application to v3, I tried to use the automatic upgrade strategy provided with the `filament/upgrade` package and noticed, that the step, where rector was called, did not produce any change at all.

When running the rector rule set manually, rector complained about some outdated patterns, that were used. 

So instead of doing a manual upgrade, I updated the rules as follows:

* Beginning with rector 0.19, the namespace `Rector\Core\` was moved to `Rector\`: https://github.com/rectorphp/rector/releases/tag/0.19.0
* Beginning with rector 0.17, querying the parent node is not possible anymore, as rector does not populate parent lookups. I updated the `SimpleMethodChangeRector` and `SimplePropertyChangeRector` to target the `Class_` instead of the method or property targeted. This is in line with the suggested update strategy described here: https://getrector.com/blog/rector-017-brings-more-robust-and-lighter-node-tree
* The `MODIFIERS_` constant in the `Class_` are deprecated, so I updated them to use the `Modifiers` class instead.

## Code style

## Testing

I tested the changes on a local project, that by far does not use every feature of Filament, but I tried to test every rectored scenario while implementing the changes. There are no automated tests in place for this package as far as I can see.

## Documentation

There are no changes to any user interface, so no action required here, I think!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,7 @@
         "pestphp/pest-plugin-laravel": "^2.0",
         "phpstan/extension-installer": "^1.1",
         "phpstan/phpstan": "^1.10",
+        "rector/rector": "^1.0",
         "spatie/laravel-medialibrary": "^10.0|^11.0",
         "spatie/laravel-ray": "^1.29",
         "spatie/laravel-tags": "^4.2",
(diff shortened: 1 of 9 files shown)


danharrin:
Hi, thank you very much for your work on this. I didn't have time to upgrade previously and had to comment out the Rector tests which were failing. Those should now be passing.

I have bumped the Rector version to v1

PR #11664, opened 2024-03-03
Add getDefaultName for Columns

SanderMuller:
## Description

This makes it possible to create custom general columns like a CreatedAtColumn without having to override the `::make()` method. This is already possible for Filters and Actions. The implementation is pretty much a copy paste from `Filament\Tables\Filters::make()`, see https://github.com/filamentphp/filament/blob/3.x/packages/tables/src/Filters/BaseFilter.php#L31

## Code style

## Testing

## Documentation

Note, `getDefaultName` is not mentioned anywhere in the documentation. Extending the section on custom Filters, Actions and Columns could be useful.


danharrin:
Had to revert this, realised its a breaking change for anyone overriding the `make()` method with the old signature

PR #11666, opened 2024-03-04
Fix options() docblock in HasOptions.php

ziming:
## Description

I have the following code that PHPStan level 5 falsely say is error

```php
public function filtersForm(Form $form): Form
    {
        $currentYear = (int) date('Y');
        $years = range(2020, $currentYear);
        if ($this->filters['year'] === null) {
            $this->filters['year'] = $currentYear;
        }

        return $form
            ->schema([
                Select::make('year')
                    ->required()
                    ->default($currentYear)
                    ->options(
                        //  Parameter #1 $options of method Filament\Forms\Components\Select::options() expects array<array<string>|string>|Closure|Illuminate\Contracts\Support\Arrayable|string|null, array<int, int>  given.    
                        array_combine($years, $years)
                    ),
            ]);
    }
```

```php
 Parameter #1 $options of method Filament\Forms\Components\Select::options() expects 
array<array<string>|string>|Closure|Illuminate\Contracts\Support\Arrayable|string|null, array<int, int>  
         given.    
```

So this attempt to fix it by changing the docblock to allow both string & int array

### DIFF
--- a/packages/tables/src/Filters/Concerns/HasOptions.php
+++ b/packages/tables/src/Filters/Concerns/HasOptions.php
@@ -21,7 +21,7 @@ trait HasOptions
     protected ?Closure $getSearchResultsUsing = null;
 
     /**
-     * @param  array<string | array<string>> | Arrayable | class-string | Closure | null  $options
+     * @param  array<string|int | array<string|int>> | Arrayable | class-string | Closure | null  $options
      */
     public function options(array | Arrayable | string | Closure | null $options): static
     {


zepfietje:
Thanks for bringing this up, @ziming!

I'm not sure if this is by design (anything to do with int to string conversions?), so will assign this to Dan for review. 🙂

danharrin:
I understand where you're coming from here. But technically, we do not use the features of integer in this function, so adding it to the type hint is basically useless. We are expecting it to be string, in case we need to manipulate it with one of the string functions. There are countless areas in the codebase where we accept strings only, and I don't think it's maintainable to have every mention as `string | int` instead. Also, what about floats? It would just be a huge mess.

I think you should cast your int to a string, or use `strval()`.

ziming:
Okay will do although it is not very convenient

PR #11667, opened 2024-03-04
Refine applyGroupingToTableQuery to Control Eager Loading for isTableReordering

misaf:
This pull request refines the behavior of the applyGroupingToTableQuery method to provide better control over eager loading, specifically when isTableReordering is enabled.

## Description

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -37,11 +37,11 @@
         fn (\Filament\Tables\Actions\BulkAction | \Filament\Tables\Actions\ActionGroup $action): bool => $action->isVisible(),
     );
     $groups = $getGroups();
-    $areGroupingSettingsVisible = count($groups) && (! $areGroupingSettingsHidden());
     $description = $getDescription();
     $isGroupsOnly = $isGroupsOnly() && $group;
     $isReorderable = $isReorderable();
     $isReordering = $isReordering();
+    $areGroupingSettingsVisible = (! $isReordering) && count($groups) && (! $areGroupingSettingsHidden());
     $isColumnSearchVisible = $isSearchableByColumn();
     $isGlobalSearchVisible = $isSearchable();
     $isSearchOnBlur = $isSearchOnBlur();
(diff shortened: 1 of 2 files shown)


zepfietje (changes requested):
Please fill out the entire pull request template (as stated in the template) and make sure `composer cs` has been run to fix code style issues.

misaf:
This important bug involves an attempt to implement lazy loading, which could result in errors.

Additionally, I ran composer cs to fix coding style issues

danharrin:
I do not understand. While reordering, the group headings and descriptions are not visible. So why does the data need to be eager loaded?

misaf:
Thank you for your feedback, @danharrin 

When reordering, our intention is to sort the table by column name and then respond to the user with the grouped data. Currently, when using grouping, applyEagerLoading is applied without any issues. However, when we introduce reordering into the mix, this method first checks if reordering is in progress. If it is, it returns the query without applying eager loading. Consequently, in the response, we are using productCategory.name without eager loading, which leads to performance issues and potential N+1 problems.

I've attached two images showing the SQL query logs during the reordering process. These logs demonstrate that without eager loading, we encounter the N+1 issue, resulting in additional queries and potential performance bottlenecks.


Screenshot 2024-03-04 at 21 06 14
Screenshot 2024-03-04 at 20 51 45

danharrin:
I still do not understand, which part of Filament do the queries originate from if the group headings are not displayed while reordering? Surely if the group headings are hidden while reordering, they do not need to display any data

misaf:
if the group headings are indeed visible during reordering, then the data associated with those groups still needs to be accurately loaded and displayed to the user.

I've uploaded a new video to better illustrate the situation. @danharrin 

https://houshang-flowers.com/screen.mov

danharrin:
How did you make group headings visible while reordering? I thought that was disabled

misaf:
I didn't enable group headings during reordering intentionally. It seems to be the default behavior.

danharrin:
I've made it so that group headings aren't present while reordering.

PR #11668, opened 2024-03-04
Refactor getRelationshipName Method for Clarity and Consistency

misaf:
This pull request enhances the getRelationshipName method by refactoring it for improved clarity and consistency.

## Description

## Code style

## Testing

## Documentation


zepfietje:
Thank you for contributing, @misaf!

However, I don't think this change makes the code more readable.
In Filament we make use of the early return pattern in a lot of places (like here).
Also, the `Stringable` value is explicitly cast to string, which we'll lose with this change.

PR #11681, opened 2024-03-04
Improve wizard steps responsive design

### LINKED ISSUE #10903: Design of form wizard with many steps
### Package

filament/filament

### Package Version

v3.2.3

### Laravel Version

v10.41.0

### Livewire Version

v3.3.5

### PHP Version

PHP 8.3.2

### Problem description

Form Wizard Steps header UI overflows and becomes hidden if it has many steps. This confuses the end user as steps keep appearing on clicking Next button but no steps showing on Steps Header.

### Expected behavior

The header should move to second row and remain within container of the wizard. 

### Steps to reproduce

Just runt the given repo app. Goto Test resource and try to add a new resource. It will show wizards steps header broken.

### Reproduction repository

https://github.com/AhmadFida2/filament-wizard-issue

### Relevant log output

_No response_

valentin-morice:
## Description

Wizard steps overflow their container when too many steps are created. Title is now left-aligned to match a potential description. Fix for #10903.

297796597-be7f474a-8acf-4eb4-a88a-65954683a5e0
<i>Before</i>

image
<i>After</i>

## Code style

## Testing

## Documentation


packages/forms/resources/views/components/wizard.blade.php
```diff
@@ -190,7 +191,7 @@ class="fi-fo-wizard-header-step-indicator text-sm font-medium"
                     <div class="grid justify-items-start">
                         @if (! $step->isLabelHidden())
                             <span
-                                class="fi-fo-wizard-header-step-label text-sm font-medium"
+                                class="fi-fo-wizard-header-step-label text-sm font-medium text-left"
```
```suggestion
                                class="fi-fo-wizard-header-step-label text-sm font-medium text-start"
```

packages/forms/resources/views/components/wizard.blade.php
```diff
@@ -123,6 +123,7 @@
             'border-b border-gray-200 dark:border-white/10' => $isContained,
             'rounded-xl bg-white shadow-sm ring-1 ring-gray-950/5 dark:bg-gray-900 dark:ring-white/10' => ! $isContained,
         ])
+        @style(['overflow: scroll'])
```
danharrin: Can this be replaced with the `overflow-x-auto` class instead of an inline style?

danharrin (changes requested):

packages/forms/resources/views/components/wizard.blade.php
valentin-morice: Yes, it works. I've made a commit.

zepfietje:
Does this prevent long step titles from wrapping, @valentin-morice?
I'm worried this may break some existing uses of wizard steps.

Say you have a wizard with two steps, which have long titles that wrap across two lines (before this change). Are they still wrapped and contained in the container size, or does horizontal scrolling kick in, preventing the titles from wrapping to the next line?

Honestly, I think there should be a better fix for this, if it is a real bug to start with. Really hard, since we don't control how Filament users use wizard steps (e.g. 20 steps in a small wizard container of 200px wide isn't going to work anyway).

valentin-morice:
image

The titles do wrap in two steps. It starts to overflow when you reach the fifth step, and that's when the steps become scrollable.

image
image

danharrin:
I think there needs to be a min-width added to those so they don't squash as much

zepfietje:
Min-width would be tricky, because the min-width shouldn't be larger than the max label width.

valentin-morice:
@danharrin @zepfietje  min-width seems to work. I could also measure the size of the steps with JS and fold the design into the mobile view if the width gets too narrow, but it requires using Alpine and would get more complex than just a few Tailwind classes.
Quick question, since Filament does not ship with all Tailwind classes, how can I add those needed for adjusting this part of the codebase? Right now I'm resorting to blade style directives.

zepfietje:
I prefer to not use JS for these things if possible.

Filament compiles Tailwind CSS classes in its views, so there's nothing you need to do apart from running `npm build`.

valentin-morice:
image
Added a min-width of 11rem.

zepfietje:
@valentin-morice @danharrin, I don't like the min-width solution.
I feel like there must be another way.

Could you play around with adding `w-max` to the steps to see if they grow larger then when the container is scrollable?

zepfietje (approved):

zepfietje:
(Above approval was just for testing.)

elmergustavo:
Thank you for your contribution.

valentin-morice:
image

@zepfietje `w-max` has the advantage to let each step stretch according to it's content. However, it doesn't seem like there'd be any line break regardless of the size of the label.

image

`min-width` on the other hand, will produce line breaks, but will lead to strange layouts like this:
image

zepfietje:
What about using `w-max` in combination with a max width?
That should give us the benefits of both, right?

valentin-morice:
image
It does. However, this `max-w` causes the label to wrap even when the steps aren't scrollable and there's enough space in the container for them to stretch their full length (for exemple, when there's only two steps).
image

zepfietje:
Replaced by https://github.com/filamentphp/filament/pull/11803 where I implemented my suggestion.

PR #11687, opened 2024-03-05
fix: icon position for action group in header actions

andrewdwallo:
## Description

This PR attempts to fix an issue where when using `iconPosition()` for an `ActionGroup` in the `getHeaderActions()` method of a custom page (in my case), it was not doing anything.

I am not sure if the code I proposed here would be considered the "best" or "correct" way to do it, but I thought it would be nice for this to be able to be used.

It is worth noting that when using `iconPosition()` for an `ActionGroup` with `iconButton()`, placing the `:icon-position` prop in the view without the check I did results in the error "`trim(): Argument #1 ($string) must be of type string, Filament\Support\Enums\IconPosition given`" coming from the `<x-filament::icon-button>` component view, so maybe there is a better way to handle this.

### Example
```php
{
    return [
        ActionGroup::make([
            Action::make('exportCSV'),
            Action::make('exportPDF'),
        ])
            ->label('Export')
            ->button()
            ->dropdownPlacement('bottom-end')
            ->icon('heroicon-c-chevron-down')
            ->iconPosition(IconPosition::After),
    ];
}
```

### Before
Screenshot 2024-03-04 at 10 03 28 PM

### After
Screenshot 2024-03-04 at 10 03 02 PM

## Code style

## Testing

### DIFF
--- a/packages/actions/resources/views/components/group.blade.php
+++ b/packages/actions/resources/views/components/group.blade.php
@@ -11,7 +11,6 @@
     'dynamicComponent' => null,
     'group' => null,
     'icon' => null,
-    'iconPosition' => null,
     'iconSize' => null,
     'iconButton' => false,
     'label' => null,
@@ -31,7 +30,6 @@
             ->dropdownPlacement($dropdownPlacement)
             ->dropdownWidth($dropdownWidth)
             ->icon($icon)
-            ->iconPosition($iconPosition)
             ->iconSize($iconSize)
             ->label($label)
             ->size($size)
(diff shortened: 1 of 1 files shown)


andrewdwallo:
Maybe a conditional check in the ActionGroup class and passing it through with `extraAttributes(['icon-position' => $iconPosition])` would be better.. not sure.

zepfietje:
Shouldn't this be fixed by making a change to `icon-button-group.blade.php` instead?

danharrin:
No, I think it needs fixing in button-group.blade.php and link-group.blade.php, similar to how we pass the icon position in button-action.blade.php and link-action.blade.php

danharrin:
Found the problem and fixed with an alt solution

andrewdwallo:
Sounds good, thanks!

PR #11691, opened 2024-03-05
➕ Add font weights as strings to enum

bdsoha:
## Description

Add font weights as strings to `FontWeight` enum to be used as classes in templates.

## Code style

## Testing

### DIFF
--- a/packages/support/src/Enums/FontWeight.php
+++ b/packages/support/src/Enums/FontWeight.php
@@ -2,21 +2,21 @@
 
 namespace Filament\Support\Enums;
 
-enum FontWeight
+enum FontWeight: string
 {
-    case Thin;
+    case Thin = 'thin';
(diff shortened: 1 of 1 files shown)


zepfietje:
Assigning to you, @danharrin, but I'm not sure about this change.

danharrin:
Yeah, since this change has not really been explained properly I don't see why it would be useful?

zepfietje:
Agree. Also it isn't clear what type of casing (kebab, snake, whatever) we should use for the backed value. If @bdsoha means passing the backed value straight to `font-{$value}` to get a Tailwind CSS class (which looks like is the goal according to https://tailwindcss.com/docs/font-weight), I don't think that's good practice.

I'm going to close this PR, as a match expression in your template should be used instead.

PR #11740, opened 2024-03-07
changed check for active sidebar item from null ->undefined

mvenghaus:
if no active sidebar item is found you get undefined instead of null


danharrin:
Not in my browser, so you'll need to add a check for both

mvenghaus:
which browser are you using? i'm using latest chrome on linux

mvenghaus:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

danharrin:
`.offsetParent` can be null

PR #11741, opened 2024-03-07
sortablejs - group option support

mvenghaus:
This option would give the possibilty to drag items from/to mutliple or nested lists.

```html
List 1
<div x-data="{}"
     x-sortable
     x-sortable-group="same"
>
```

```html
List 2
<div x-data="{}"
     x-sortable
     x-sortable-group="same"
>
```

### DIFF
--- a/packages/support/resources/js/sortable.js
+++ b/packages/support/resources/js/sortable.js
@@ -11,6 +11,7 @@ export default (Alpine) => {
         }
 
         el.sortable = Sortable.create(el, {
+            group: el.getAttribute('x-sortable-group'),
             draggable: '[x-sortable-item]',
             handle: '[x-sortable-handle]',
             dataIdAttr: 'x-sortable-item',


packages/support/resources/js/sortable.js
```diff
@@ -11,6 +11,7 @@ export default (Alpine) => {
         }
 
         el.sortable = Sortable.create(el, {
+            group: el.getAttribute('x-sortable-group'),
```
danharrin, line 14: Can this not just be `[x-sortable-group]` like the others?

mvenghaus: the other don't need a value

danharrin: and what happens if the attribute is not defined

mvenghaus: If not defined it gets null -> behaves like now.

Example why you need a value.

you have 4 (2 groups) lists and you want only to move items inside groups.

```
-------       -------
|     |       |     |
|     |  <->  |     |
|     |       |     |
-------       -------

-------       -------
|     |       |     |
|     |  <->  |     |
|     |       |     |
-------       -------
```

PR #11742, opened 2024-03-07
Handle fallback url of media

akunbeben:
## Description

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/spatie-laravel-media-library-plugin/src/Tables/Columns/SpatieMediaLibraryImageColumn.php
+++ b/packages/spatie-laravel-media-library-plugin/src/Tables/Columns/SpatieMediaLibraryImageColumn.php
@@ -16,6 +16,31 @@ class SpatieMediaLibraryImageColumn extends ImageColumn
 
     protected string | Closure | null $conversion = null;
 
+    protected function setUp(): void
+    {
+        parent::setUp();
+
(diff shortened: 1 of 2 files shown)


akunbeben:
Hey guys! @danharrin @zepfietje 

In this PR, I handle the fallback URL if the SpatieMediaLibraryImageColumn didn't have media.
I tried it using `->default()` method. But, no luck. So, I handle it this way. 

Is it allowed to use `// @phpstan-ignore-line`? I've tried to modify the docblock but still can't pass the phpstan analyze. Because the `/** @var Model $record */` is a plain Model instance.

Before
before

After
after

packages/spatie-laravel-media-library-plugin/src/Tables/Columns/SpatieMediaLibraryImageColumn.php
```diff
@@ -57,7 +58,7 @@ public function getImageUrl(?string $state = null): ?string
             $media = $record->getRelationValue('media')->first(fn (Media $media): bool => $media->uuid === $state);
 
             if (! $media) {
-                continue;
+                return $record->getFirstMediaUrl($this->getCollection()); // @phpstan-ignore-line
```
danharrin: This will break using the column on a -many relationship since once one record doesn't have media, the whole thing will fail

akunbeben:
Wow, thanks Dan!😄

PR #11749, opened 2024-03-07
Suggest resources when creating a page

mokhosh:
## Description

Currently when you run `php artisan make:filament-page` you are asked to type the name of your resource. This PR makes it possible to suggest the name of resources as you type, or by selecting them directly from the dropdown.

I have moved the part of the code that identifies the current panel up and before asking for the resource, without changing it, so we can use it to get the resources of the currently selected panel.

Old:

Screenshot 2024-03-07 at 17 21 32

New:

Screenshot 2024-03-07 at 17 20 36

## Code style

## Testing

### DIFF
--- a/packages/panels/src/Commands/MakePageCommand.php
+++ b/packages/panels/src/Commands/MakePageCommand.php
@@ -71,7 +71,7 @@ class MakePageCommand extends Command
         $resourceInput = $this->option('resource') ?? suggest(
             label: 'Which resource would you like to create this in?',
             options: collect($panel->getResources())
-                ->filter(fn (string $path): bool => str($path)->contains('\\Resources\\'))
+                ->filter(fn (string $namespace): bool => str($namespace)->contains('\\Resources\\'))
                 ->map(
                     fn (string $namespace): string => str($namespace)
                         ->afterLast('\\Resources\\')
--- a/packages/panels/src/Commands/MakePageCommand.php
+++ b/packages/panels/src/Commands/MakePageCommand.php
@@ -76,7 +76,8 @@ class MakePageCommand extends Command
                     fn (string $namespace): string => str($namespace)
                         ->afterLast('\\Resources\\')
                         ->beforeLast('Resource')
-                )->all(),
+                )
+                ->all(),
             placeholder: '[Optional] UserResource',
         );
 
--- a/packages/panels/src/Commands/MakePageCommand.php
+++ b/packages/panels/src/Commands/MakePageCommand.php
@@ -73,7 +73,7 @@ class MakePageCommand extends Command
             options: collect($panel->getResources())
                 ->map(
                     fn ($resource): string => str($resource)
-                        ->afterLast('\\')
+                        ->afterLast('\\Resources\\')
                         ->beforeLast('Resource')
                 )->all(),
             placeholder: '[Optional] UserResource',
--- a/packages/panels/src/Commands/MakePageCommand.php
+++ b/packages/panels/src/Commands/MakePageCommand.php
@@ -73,7 +73,7 @@ class MakePageCommand extends Command
             options: collect($panel->getResources())
                 ->filter(fn (string $namespace): bool => str($namespace)->contains('\\Resources\\'))
                 ->map(
-                    fn (string $namespace): string => str($namespace)
+                    fn (string $namespace): string => (string) str($namespace)
                         ->afterLast('\\Resources\\')
                         ->beforeLast('Resource')
                 )
--- a/packages/panels/src/Commands/MakePageCommand.php
+++ b/packages/panels/src/Commands/MakePageCommand.php
@@ -76,6 +76,7 @@ class MakePageCommand extends Command
                         ->afterLast('\\')
                         ->beforeLast('Resource')
                 )->all(),
+            placeholder: '[Optional] UserResource',
         );
 
         if (filled($resourceInput)) {
--- a/packages/panels/src/Commands/MakePageCommand.php
+++ b/packages/panels/src/Commands/MakePageCommand.php
@@ -47,6 +47,26 @@ class MakePageCommand extends Command
         $resourceClass = null;
         $resourcePage = null;
 
+        $panel = $this->option('panel');
+
+        if ($panel) {
+            $panel = Filament::getPanel($panel);
+        }
+
+        if (! $panel) {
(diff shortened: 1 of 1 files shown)


danharrin:
How easy is it to not select a resource?

mokhosh:
Just pressing enter

danharrin:
The screenshot doesn't show the Optional placeholder, where is that?

mokhosh:
https://github.com/filamentphp/filament/assets/6499685/a795ad2b-260f-49fe-a0ea-c041e488e8d5

mokhosh:
> The screenshot doesn't show the Optional placeholder, where is that?

Sorry took the screenshot before adding the placeholder back. The video has the latest changes.

packages/panels/src/Commands/MakePageCommand.php
```diff
@@ -47,8 +48,34 @@ public function handle(): int
...
+                default: Filament::getDefaultPanel()->getId()
+            )] : Arr::first($panels);
+        }
+
+        $resourceInput = $this->option('resource') ?? suggest(
+            label: 'Which resource would you like to create this in?',
+            options: collect($panel->getResources())
+                ->map(
+                    fn ($resource): string => str($resource)
+                        ->afterLast('\\')
```
danharrin:
```suggestion
                        ->afterLast('\\Resources\\')
```

to handle nested directories?

mokhosh: Hmm. What would a nested directory path look like?

danharrin: App\Filament\Resources\Blog\PostResource

danharrin:
Can you make it so that if `\\Resources` is not in the path, it doesn't get included in the list. `filter()` the collection. Thanks!

mokhosh:
Is this what you had in mind?

Also, doesn't the panel handle these? Since we're getting all the resources from the panel itself?

danharrin:
If resources are manually registered it wouldn't be able to guess it properly

packages/panels/src/Commands/MakePageCommand.php
```diff
@@ -47,8 +48,35 @@ public function handle(): int
...
+                    $panels,
+                ),
+                default: Filament::getDefaultPanel()->getId()
+            )] : Arr::first($panels);
+        }
+
+        $resourceInput = $this->option('resource') ?? suggest(
+            label: 'Which resource would you like to create this in?',
+            options: collect($panel->getResources())
+                ->filter(fn ($path) => str_contains($path, '\\Resources\\'))
```
danharrin:
```suggestion
                ->filter(fn (string $path): bool => str($path)->contains('\\Resources\\'))
```

mokhosh:
But we would still want those manually registered resources, no?

packages/panels/src/Commands/MakePageCommand.php
```diff
@@ -47,8 +48,35 @@ public function handle(): int
...
+        }
+
+        $resourceInput = $this->option('resource') ?? suggest(
+            label: 'Which resource would you like to create this in?',
+            options: collect($panel->getResources())
+                ->filter(fn (string $path): bool => str($path)->contains('\\Resources\\'))
+                ->map(
+                    fn ($resource): string => str($resource)
+                        ->afterLast('\\Resources\\')
+                        ->beforeLast('Resource')
```
danharrin:
```suggestion
                    fn (string $namespace): string => str($namespace)
                        ->afterLast('\\Resources\\')
                        ->beforeLast('Resource')
```

packages/panels/src/Commands/MakePageCommand.php
```diff
@@ -47,8 +48,35 @@ public function handle(): int
...
+                    $panels,
+                ),
+                default: Filament::getDefaultPanel()->getId()
+            )] : Arr::first($panels);
+        }
+
+        $resourceInput = $this->option('resource') ?? suggest(
+            label: 'Which resource would you like to create this in?',
+            options: collect($panel->getResources())
+                ->filter(fn (string $path): bool => str($path)->contains('\\Resources\\'))
```
danharrin:
```suggestion
                ->filter(fn (string $namespace): bool => str($namespace)->contains('\\Resources\\'))
```

packages/panels/src/Commands/MakePageCommand.php
```diff
@@ -47,8 +48,35 @@ public function handle(): int
...
+
+        $resourceInput = $this->option('resource') ?? suggest(
+            label: 'Which resource would you like to create this in?',
+            options: collect($panel->getResources())
+                ->filter(fn (string $namespace): bool => str($namespace)->contains('\\Resources\\'))
+                ->map(
+                    fn (string $namespace): string => str($namespace)
+                        ->afterLast('\\Resources\\')
+                        ->beforeLast('Resource')
+                )->all(),
```
danharrin:
```suggestion
                )
                ->all(),
```

packages/panels/src/Commands/MakePageCommand.php
```diff
@@ -47,8 +48,36 @@ public function handle(): int
...
+                default: Filament::getDefaultPanel()->getId()
+            )] : Arr::first($panels);
+        }
+
+        $resourceInput = $this->option('resource') ?? suggest(
+            label: 'Which resource would you like to create this in?',
+            options: collect($panel->getResources())
+                ->filter(fn (string $namespace): bool => str($namespace)->contains('\\Resources\\'))
+                ->map(
+                    fn (string $namespace): string => str($namespace)
```
danharrin:
```suggestion
                    fn (string $namespace): string => (string) str($namespace)
```

packages/panels/src/Commands/MakePageCommand.php
mokhosh: Type hints definitely, but wouldn't the native `str_contains` suffice? Using `str` helper seems a bit overkill.

packages/panels/src/Commands/MakePageCommand.php
mokhosh: Isn't this unnecessary since the return type already casts to string?

packages/panels/src/Commands/MakePageCommand.php
danharrin: I don't like the readability, the order of arguments is ambiguous

PR #11769, opened 2024-03-08
Don't apply length constraints to numeric (decimal) input fields

buddhaCode:
## Description

When applying the length validation to a `numeric()` `TextInput` field, the result is not what I expected it to be.

Example:

```php
TextInput::make('numeric')
  ->numeric()
  ->maxLength(2)
```

Results in:

Bildschirmfoto 2024-03-08 um 17 33 07

The input has exactly two digits. But the validations fails.

I guess, it's not easy to tell the overall length of a decimal number. [According to the Laravel docs, the min_digits, digits and max_digits rules should only be applied to integer values](https://laravel.com/docs/10.x/validation#rule-digits). Not to float values. Also, I couldn't find any solution for floats in Laravel's preset validation rules. There is a rule `decimal` to check the length of the decimal part. But that's only a part of the whole story.

 I suggest to not apply the `minLength()`, `length()` and `maxLength()` validations to any decimal number at all.

This PR changes the behavior accordingly.

Background:

How did I stumble across this issue? I applied a standard check to all my TextInputs.

```php
TextInput::configureUsing(function (TextInput $textinput) {
            $textinput
                ->maxLength(255)
            ;
        });
```

This also applies to fields where I call the `numeric()` method on. Unfortunately, there is no way to remove the length constraint after it's set once.

@danharrin: If this change, shouldn't suit into the framework, maybe there should be antoher way to unset pre-configured validations.

## Code style

## Testing

Didn't find any software tests, which I could fix or extend.

## Documentation

### DIFF
--- a/packages/forms/docs/03-fields/02-text-input.md
+++ b/packages/forms/docs/03-fields/02-text-input.md
@@ -252,8 +252,6 @@ TextInput::make('code')
     ->length(8)
 ```
 
-The length validation doesn't apply to `numeric()` fields, as Laravel has no rules to validate the overall length of a decimal number. You can use the `decimal` rule to validate the number of decimal places.
-
 ### Size validation
 
 You may validate the minimum and maximum value of a numeric input by setting the `minValue()` and `maxValue()` methods:
--- a/packages/forms/docs/03-fields/02-text-input.md
+++ b/packages/forms/docs/03-fields/02-text-input.md
@@ -252,6 +252,8 @@ TextInput::make('code')
     ->length(8)
 ```
 
+The length validation doesn't apply to `numeric()` fields, as Laravel has no rules to validate the overall length of a decimal number. You can use the `decimal` rule to validate the number of decimal places.
+
 ### Size validation
 
 You may validate the minimum and maximum value of a numeric input by setting the `minValue()` and `maxValue()` methods:
--- a/packages/forms/src/Components/Concerns/CanBeLengthConstrained.php
+++ b/packages/forms/src/Components/Concerns/CanBeLengthConstrained.php
@@ -3,6 +3,7 @@
 namespace Filament\Forms\Components\Concerns;
 
 use Closure;
+use Filament\Forms\Components\Contracts\CanHaveNumericState;
 
 trait CanBeLengthConstrained
 {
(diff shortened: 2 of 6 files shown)


danharrin:
Yeah this would be a breaking change. I've made it so you can unset the validation rules now

buddhaCode:
Thank you, @danharrin. That helps a lot.

But I think, there is still something wrong with the digits rules on decimal numbers. Maybe that's a breaking change. But I also looks like bug to me.
Looking at what Laravel does here, this validation always evaluates to false, if the value contains non-digit characters like the decimal point. Please [have look have the regex here](https://github.com/laravel/framework/blob/f2bff1c04ee0eb978d2daef4fdc48d27e23f7bf5/src/Illuminate/Validation/Concerns/ValidatesAttributes.php#L1483).

danharrin:
So what are we supposed to do in Filament, if this is Laravel behaviour?

buddhaCode:
I think there a two points:

1. Filament uses the digits rules wrong. According to the [Laravel Docs, the digits rules can only be applied to **integer** values](https://laravel.com/docs/11.x/validation#rule-digits). But Filament also applies this rules to decimal numbers aka floats.
> digits:value
>
> The **integer** under validation must have an exact length of value.

I think, this is clearly a misusage.

2. Currently, there is no rule to validate the total length of decimal numbers/floats. One can validate the length of the decimal part with `decimal`. But there is no way to validate the whole number part or the total length. That's sad. Maybe the `digits` rule in Laravel should be able to validate the total length of floats, too. But currently, the rule works as documented and can only applied to integer values.

danharrin:
@buddhaCode maybe you should resubmit this PR for Laravel? https://github.com/laravel/framework/pull/47976

buddhaCode:
Oh thats a nice one :)

What do you say to my 1st bullet point? Don't you think this is a valid point? What do I miss?

danharrin:
Since there isn't an alternative what do we do?

buddhaCode:
What do you think about Tims solution in the linked PR? Guess it's not trivial to apply this without knowledge about the desired length of the decimal part.

danharrin:
Test it out by passing the rule to `->rule()` yourself, I don't know if I want to modify Filament to force the user to pass in a number of decimal places though?

buddhaCode:
>  I don't know if I want to modify Filament to force the user to pass in a number of decimal places though?

Nope. You shouldn't do that. I agree with you.

My conclusion is, that (min|max)`length()` and combination with `numeric()` simply doesn't work.

PR #11816, opened 2024-03-12
Merge extra Attributes of  ShowPasswordAction/HidePasswordAction

lukas-frey:
Merge the extra attributes so that they can be configured outside of the filament core.

Explanation:
I needed to override the extraAttributes via the global `ShowPasswordAction::configureUsing` method, but my changes got overridden by this.

### DIFF
--- a/packages/forms/src/Components/TextInput/Actions/HidePasswordAction.php
+++ b/packages/forms/src/Components/TextInput/Actions/HidePasswordAction.php
@@ -25,7 +25,7 @@ class HidePasswordAction extends Action
         $this->extraAttributes([
             'x-cloak' => true,
             'x-show' => 'isPasswordRevealed',
-        ]);
+        ], true);
 
         $this->alpineClickHandler('isPasswordRevealed = false');
     }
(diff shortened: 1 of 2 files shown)


packages/forms/src/Components/TextInput/Actions/HidePasswordAction.php
```diff
@@ -25,7 +25,7 @@ protected function setUp(): void
         $this->extraAttributes([
             'x-cloak' => true,
             'x-show' => 'isPasswordRevealed',
-        ]);
+        ], true);
```
danharrin:
```suggestion
        ], merge: true);
```

packages/forms/src/Components/TextInput/Actions/ShowPasswordAction.php
```diff
@@ -24,7 +24,7 @@ protected function setUp(): void
 
         $this->extraAttributes([
             'x-show' => '! isPasswordRevealed',
-        ]);
+        ], true);
```
danharrin:
```suggestion
        ], merge: true);
```

PR #11856, opened 2024-03-14
Ability to add extra attributes in the Menu Items.

jaksonleao:


danharrin (changes requested):
This doesn't feel properly tested, `x-filament::dropdown.list.item` is present in other places in user-menu.blade.php and tenant-menu.blade.php and those haven't been updated with the attributes?

Also you mention "filter menu items" in the docs sections when this should just be "menu items"

Also, you need to run `composer cs` as per the PR template that you deleted.

danharrin:
Closing due to issues mentioned above. Please reopen when that comment has been addressed. Please test your work.

PR #11869, opened 2024-03-15
Fix: getStartStep to accommodate for hidden components

yob-yob:
## Description

This pull request will fix the issue of `getStartStep()` in `wizard.php@line:212`  It does not accommodate for a scenario where a Step Component is hidden.

- [✅] Visual changes (if any) are shown using screenshots/recordings of before and after.

There isn't really any visual changes, it only fixes an issue.

## Code style

- [✅] `composer cs` command has been run.

I was able to run `composer cs`.

## Testing

- [✅] Changes have been tested.

There was no tests for the Wizard component (I might be wrong here, so if you can redirect me to the right files, I will gladly write a test for this issue.)

## Documentation

- [✅] Documentation is up-to-date.

I'm Guessing there is no need for Documentation for this change?

### DIFF
--- a/packages/forms/src/Components/Wizard.php
+++ b/packages/forms/src/Components/Wizard.php
@@ -214,7 +214,9 @@ class Wizard extends Component
         if ($this->isStepPersistedInQueryString()) {
             $queryStringStep = request()->query($this->getStepQueryStringKey());
 
-            foreach ($this->getChildComponents() as $index => $step) {
+            $components = collect($this->getChildComponents())->filter(fn ($component) => ! $component->isHidden())->values();
+
+            foreach ($components as $index => $step) {
                 if ($step->getId() !== $queryStringStep) {
                     continue;
                 }
--- a/packages/forms/src/Components/Wizard.php
+++ b/packages/forms/src/Components/Wizard.php
@@ -214,9 +214,7 @@ class Wizard extends Component
         if ($this->isStepPersistedInQueryString()) {
             $queryStringStep = request()->query($this->getStepQueryStringKey());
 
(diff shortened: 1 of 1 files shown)


yob-yob:
PhpStan Errors does not concern my changes, what should I do about this?

yob-yob:
@danharrin 

TL;DR: wrap `array_filter` with `array_values`

Should I open another Merge Request? or can this one be reopened?

....
....

Hi Dan, I already got the latest version of Filament which is using this commit, and I just noticed that my code still doesn't work as expected so I viewed the code in Vendor and found out that you have contributed a different approach to solving the issue.

however, `$this->getChildComponentContainer()->getComponents()` still returns `Steps` that is supposed to be hidden....

making this pull request not work at all... 

I did try to find out what happens with you try to use `->getComponents()`

vendor/filament/forms/src/Concerns/HasComponents.php 

@ Line: 112

```
 return array_filter(
            $components,
            fn (Component $component) => $component->isVisible(),
        );
```

This means that it should respect the visibility of the components... but in my case it doesn't work... 

So did some more digging...

I found out that array_filter doesn't resent the index count of the array...

meaning that if a wizard has 4 steps... and the 3 step is hidden... the array would look like this

```
[0 => step, 1 => step, 2 => step, 4 => step]
```

and because the `Wizard.php@getStartStep()` uses the `index` of a step to initialize the `startStep` it still wouldn't work...

so the fix here should be to wrap `array_filter` with `array_values`

danharrin:
Merged PRs cannot be reopened

yob-yob:
I have made a follow-up pull request #12190

PR #11875, opened 2024-03-15
Bread Crumb Component Updates and Additions

tnylea:
This PR will add the ability to include the `Current` page text into the breadcrumbs. By default if you do not include the `current` attribute in the component, it will work as it did before.

Additionally, in the component, it would not default to a direction, so if you were to render the following code:

```
<x-breadcrumbs
    :breadcrumbs="[
        '/' => 'Home',
        '/dashboard' => 'Dashboard',
        '/dashboard/users' => 'Users'
    ]"
/>
```

You would end up with something that looks like the following:

CleanShot 2024-03-15 at 09 53 57@2x

Not the desired output.

So, in this PR `ltr` will be default and if you want to use `rtl` you would have to specify that by including the `dir` attribute in the component and setting it to `rtl`

## 📼 Quick Video.

I've also created a quick video to show you what the breadcrumbs look like in this PR. I then swap over to the original (old) breadcrumbs component that is currently in the repo. You'll see the differences and updates in this video.

https://www.loom.com/share/73273290a6ee4daa938692bd382a0b0a

## Testing

I've made sure to thoroughly test the component and make sure it is rendering correctly both in light and dark mode.

## Documentation

I've also updated the documentation to include these updates.

Please let me know if you would like me to make any further changes or need clarification on anything.

Thanks!

### DIFF
--- a/packages/support/docs/09-blade-components/02-breadcrumbs.md
+++ b/packages/support/docs/09-blade-components/02-breadcrumbs.md
@@ -40,5 +40,5 @@ Home > Dashboard > Users
 If you specify `rtl` as the direction, the breadcrumbs will look similar to the following:
 
 ```
-Home < Dashboard < Users
+Users < Dashboard < Home
 ```
--- a/packages/support/docs/09-blade-components/02-breadcrumbs.md
+++ b/packages/support/docs/09-blade-components/02-breadcrumbs.md
@@ -16,3 +16,29 @@ The breadcrumbs component is used to render a simple, linear navigation that inf
 ```
 
 The keys of the array are URLs that the user is able to click on to navigate, and the values are the text that will be displayed for each link.
+
+### RTL Support
+
+If you need to rotate the breadcrumb separator for RTL support, simply add the `dir` attribute with the value of `rtl`:
+
+```
+<x-filament::breadcrumbs 
+    :breadcrumbs="[
+        '/' => 'Home',
+        '/dashboard' => 'Dashboard',
+    ]"
+    dir="rtl"
+/>
+```
+
+By default the breadcrumb separator direction is set to `ltr`, and will look similar to the following:
+
+```
+Home > Dashboard > Users
+```
+
+If you specify `rtl` as the direction, the breadcrumbs will look similar to the following:
+
+```
+Home < Dashboard < Users
+```
--- a/packages/support/resources/views/components/breadcrumbs.blade.php
+++ b/packages/support/resources/views/components/breadcrumbs.blade.php
@@ -61,7 +61,7 @@
                         'ltr:hidden',
                     ])
                 />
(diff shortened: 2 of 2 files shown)


danharrin:
Hi Tony, what's the reason you don't have `dir` on the `<html>` tag? That's what we do in the Panel Builder so it applies to all elements. I am not a fan of introducing a new attribute for this.

danharrin:
Also, the "current" behaviour can already be achieved, by appending the label of the current page onto the `:breadcrumbs`, but without a key (so the key is numeric). That's also what we do in the panel builder.

danharrin:
I believe my comments will help you out, so I'm going to close this. But if you have further clarification, just reply here and we can reopen.

tnylea:
@danharrin Great, thanks for the clarification. 

So, say that someone is not using the Panel builder, do you recommend adding the `dir` to the body tag of the page, or where do you recommend adding that? It would be nice if the `breadcrumbs` component defaulted to `dir='ltr'` as opposed to showing both carets (left and right), making the component look broken by default.

Let me know :)

Thanks!

danharrin:
If we added `dir='ltr'` to the component, it would override the body setting, right?

tnylea:
Yes it would override the default setting. 

By default you would *not* want to require that developers have to add this attribute, and that `ltr` would be default. But, currently that is not the case.

If you include this inside of a page that is not inside of the panel builder:

```
<x-breadcrumbs
	:breadcrumbs="[
		'/' => 'Home',
		'/dashboard' => 'Dashboard',
		'/dashboard/users' => 'Users',
		'' => 'John Doe'
	]"
/>
```

You end up with the component that looks like this:

CleanShot 2024-03-18 at 10 08 11@2x

To fix this, you would need to add `dir='ltr'` to the parent or the component itself:


```
<div dir="ltr">
	<x-breadcrumbs
		:breadcrumbs="[
			'/' => 'Home',
			'/dashboard' => 'Dashboard',
			'/dashboard/users' => 'Users',
			'' => 'John Doe'
		]"
	/>
</div>
```

CleanShot 2024-03-18 at 10 09 59@2x

So, including this as a default prop:

```
@props([
    'breadcrumbs' => [],
    'dir' => 'ltr'
])
```

It would set the default as `ltr` whether or not it was set anywhere else. Let me know if that makes sense.

Perhaps this is the expected behavior though 🤷‍♂️, the developer would just need to know that they need the `dir='ltr'` attribute set somewhere in a parent element of that component.

danharrin:
But we'd need to `:dir="null"` in the admin panel, or pass in the RTL setting again? Otherwise it would override our `<html>` setting?

danharrin:
I understand its not desired but is there another way to do this in CSS?

danharrin:
We want `ltr:hidden` to apply even when the dir attribute is not present

PR #11893, opened 2024-03-17
Fix duplicate headers error import

sandersjj:
Currently when you import a file with duplicate header names you will get a `SyntaxError`: `\League\Csv\SyntaxError`. This in production throws a "Server Error" and the users has no clue what is happening. This PR ties to solve this. 

## Code style

- [ X] `composer cs` command has been run.

## Testing

- [X ] Changes have been tested.

## Documentation

- [X ] Documentation is up-to-date.


tonypartridge:
I am going to guess that Syntax error will be thrown for more than just duplicate column names and as such would need conditioning?

sandersjj:
> I am going to guess that Syntax error will be thrown for more than just duplicate column names and as such would need conditioning?

You are right. I have covered them now in my PR. Please let me know what you thoughts are.

danharrin (changes requested):
I feel like this should be a validation rule on the FileUpload instead of after the import process has been initiated. It should be a custom rule.

danharrin:
Since that is basically a whole rewrite, please reopen this when you're done

sandersjj:
> I feel like this should be a validation rule on the FileUpload instead of after the import process has been initiated. It should be a custom rule.

Hey @danharrin thanks for your feedback. 
I don't mind adding custom  a validation rule to FileUpload. However I am a bit lost of how to do this. Do you mind to give me some guidelines on how to get started on this?

Thanks so much.

danharrin:
Is this enough? https://filamentphp.com/docs/3.x/forms/validation#custom-rules

sandersjj:
Hey @danharrin,

Thanks for your feedback. 
We have looked into this for a while. The conclusion is that we don't know how to move forward. A few considerations:

1. Only once the file is processed the error wil trigger. See https://github.com/filamentphp/filament/blob/7e92e571cd83afcf4e9bc1785abb7bcdedce56eb/packages/actions/src/Concerns/CanImportRecords.php#L177 This is when you actually press the import button. And not when the file is uploaded to the system.

2. We have tried to add a custom rule like so: 
```php
  FileUpload::make('file')
      ->rules([
          function () {
              return function (string $attribute, $value, Closure $fail) {
                  var_dump($attribute, $value);
              };
          },
      ])
```
`$value` is a `Livewire\Features\SupportFileUploads\TemporaryUploadedFile`. Should we manually try to process it and catch the exception? 
Or should we parse the headers and implement the same logic as LeagueCsv for validating the headers in this validation rule?
3. You said it should be a separate rule. We have looked in the code base and could not find any other custom Laravel validation rules. How should we handle this?                
                
Looking forward to your feedback.
Thanks.

danharrin:
2) Process it and parse, yes
3) Use a closure based rule, you dont need a separate class. We do it in other places too I think

PR #11905, opened 2024-03-18
Update 11-tenancy.md

seansch:
## Description

The currently suggested use of `auth()->check()` in a global scope causes an infinite loop and crashes the server when used. This is a relevant Laravel issue: https://github.com/laravel/framework/issues/18218. It seems that the auth check queries for the user at some point, which triggers the scope again, causing the loop.

I found that using `auth()->hasUser()` achieves the desired result without causing the loop condition.

Alternatively, https://github.com/filamentphp/filament/pull/9968/commits/92b09186262dd64572d6fe8f8acc132c2a43fd5a could be reverted, moving `auth()->check()` back outside of the `addGlobalScope`, which does work for the given example. The downside of suggesting that is if you use a class-based scope, like what's suggested on https://laravel.com/docs/11.x/eloquent#global-scopes, you cannot use `auth()->check()` inside of the `apply` method because it leads to the same looping condition.

Edit: I'm realizing now that this most likely only effects a global scope applied to the User model.

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/panels/docs/11-tenancy.md
+++ b/packages/panels/docs/11-tenancy.md
@@ -24,7 +24,7 @@ class Post extends Model
     protected static function booted(): void
     {
         static::addGlobalScope('team', function (Builder $query) {
-            if (auth()->check()) {
+            if (auth()->hasUser()) {
                 $query->where('team_id', auth()->user()->team_id);
                 // or with a `team` relationship defined:
                 $query->whereBelongsTo(auth()->user()->team);


danharrin:
Yeah, considering it only affects the User model, would `auth()->user()` even work? Which is also what we are depending on here. Feels like it's not something we should be concerned with for our documentation IMO? Let me know if you disagree and we can reopen

seansch:
It was something I fought with for a while last night, I'm using a basic one-to-many relationship for users belonging to different companies.

Using `auth()->hasUser()` works in the User model scope because it only checks if auth has an instance of User(`$this->user`) rather than loading the associated User model(`$this->user()`) in the check and triggering a  db query.

Illuminate\Auth\GuardHelpers:
```
    public function hasUser()
    {
        return ! is_null($this->user);
    }

    public function check()
    {
        return ! is_null($this->user());
    }
```

This change might help someone avoid the wall I ran into. Adding the scope to the User model was the first one I tried adding after reading the docs and confusion ensued.

seansch:
I did a little more testing to be sure that this works as intended.

I'm using this global scope on my User model:

```
    protected static function booted(): void
    {
        static::addGlobalScope('company', function (Builder $query) {
            if (auth()->hasUser()) {
                $query->whereBelongsTo(auth()->user()->company);
            }
        });
    }
```

I made a middleware to log all DB queries and then loaded the user view page.

These are the queries with the scope applied:

```
[2024-03-18 10:20:12] local.INFO: Query Log: [
{"query":"select * from `users` where `id` = ? limit 1","bindings":[100000],"time":0.99},
{"query":"select * from `companies` where `companies`.`id` = ? limit 1","bindings":[1],"time":0.25},
{"query":"select count(*) as aggregate from `users` where `users`.`company_id` in (?)","bindings":[1],"time":0.23},
{"query":"select * from `users` where `users`.`company_id` in (?) order by `users`.`id` asc limit 10 offset 0","bindings":[1],"time":0.26},
{"query":"select * from `companies` where `companies`.`id` in (1)","bindings":[],"time":0.16}]
```

And there are the queries with the scope disabled:

```
[2024-03-18 10:20:47] local.INFO: Query Log: [
{"query":"select * from `users` where `id` = ? limit 1","bindings":[100000],"time":0.8},
{"query":"select count(*) as aggregate from `users`","bindings":[],"time":0.18},
{"query":"select * from `users` order by `users`.`id` asc limit 10 offset 0","bindings":[],"time":0.23},
{"query":"select * from `companies` where `companies`.`id` in (1, 2, 3)","bindings":[],"time":0.19}]
```

Calling `auth()->user()->company` doesn't appear to cause any issues inside of the scope.

PR #11908, opened 2024-03-18
Add `wire:navigate` to Action url method

binaryfire:
Adds `wire:navigate` logic to the Action `->url()` method.

`FilamentView::hasSpaMode()` wasn't working - maybe because I'm testing in a standlone SimplePage which I've manually associated with a panel using `setCurrentPanel()`? Getting it from the panel works though.

### DIFF
--- a/packages/actions/resources/views/components/action.blade.php
+++ b/packages/actions/resources/views/components/action.blade.php
@@ -7,6 +7,11 @@
 @php
     $isDisabled = $action->isDisabled();
     $url = $action->getUrl();
+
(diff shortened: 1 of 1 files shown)


danharrin:
Please open an issue with a reproduction as the current implementation should already do this

binaryfire:
@danharrin Did some more digging and yeah, it’s an edge case. I have some standalone pages that I’m routing to directly. It’s happening in those.

Not worth investigating. I’ll just add wire:navigate via extraAttributes.

PR #11914, opened 2024-03-18
Feat: Add lifecyclehook events to actions

SebastiaanKloos:
## Description

## Code style

## Testing

## Documentation


danharrin (changes requested):
Surely this is going to break all non-Filament\Actions\Action actions? Since they don't extend that object? I think it needs to be MountableAction?

SebastiaanKloos:
> Surely this is going to break all non-Filament\Actions\Action actions? Since they don't extend that object? I think it needs to be MountableAction?

Ahh yeah. Thanks. Fixed now! 😎

SebastiaanKloos:
Thankyou💛

PR #11930, opened 2024-03-19
Add disk to local in upload csv file on import action

sbc640964:
## Description

When using an environment variable
```env
FILAMENT_FILESYSTEM_DISK=s3
```

The import operation cannot retrieve the file from the bucket (it creates a mixed URL between the bucket and the - temporary - file location on the server).

This fix will set the upload field on import to always use the local file system.


danharrin:
Please open an issue with more details, the system was designed for use on all disks including S3

PR #11947, opened 2024-03-19
Update 03-creating-records.md

thethunderturner:
It should be $this->data not $this->getRecord(). It doesn't make sense to check a record that hasn't been created yet.

## Description

Fix in the docs. Should use $this->data in the beforeCreate. It doesnt make sense to check a record that hasnt been created yet.

## Code style

## Testing

## Documentation

### DIFF
--- a/packages/panels/docs/03-resources/03-creating-records.md
+++ b/packages/panels/docs/03-resources/03-creating-records.md
@@ -162,7 +162,7 @@ use Filament\Notifications\Notification;
 
 protected function beforeCreate(): void
 {
-    if (! $this->getRecord()->team->subscribed()) {
+    if (! $this->data->team->subscribed()) {
         Notification::make()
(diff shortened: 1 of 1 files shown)


danharrin:
`data` is not an object its an array so that would never really work, how about we replace it with `auth()->user()->team`

packages/panels/docs/03-resources/03-creating-records.md
```diff
@@ -162,7 +162,7 @@ use Filament\Notifications\Notification;
 
 protected function beforeCreate(): void
 {
-    if (! $this->getRecord()->team->subscribed()) {
+    if (! $this->data->team->subscribed()) {
```
danharrin:
```suggestion
    if (! auth()->user()->team->subscribed()) {
```

PR #11976, opened 2024-03-21
Qualify sort column

dododedodonl:
## Description

When changing the default sort, eg. `->defaultSort('id', 'desc')` the column has to be prefixed with the table name in certain scenarios to prevent:
`SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'id' in order clause is ambiguous`

This can be automatically done by filament in these cases.

I am not sure if only `Filament\Tables\Concerns\CanSortRecords` should be changed, or also the `getSortColumnForQuery()` method in the `Group` class and `InteractsWithTableQuery` trait.

## Code style

## Testing

- [X] Changes to `Filament\Tables\Concerns\CanSortRecords` have been tested
- [ ] Changes to `getSortColumnForQuery()` method in the `Group` class and `InteractsWithTableQuery` trait have not been extensively tested, but works on system

## Documentation

-

### DIFF
--- a/packages/tables/src/Columns/Concerns/InteractsWithTableQuery.php
+++ b/packages/tables/src/Columns/Concerns/InteractsWithTableQuery.php
@@ -146,7 +146,7 @@ trait InteractsWithTableQuery
             [];
 
         if (! count($relationships)) {
-            return $sortColumn;
+            return $query->getModel()->qualifyColumn($sortColumn);
         }
 
         $currentRelationshipName = array_shift($relationships);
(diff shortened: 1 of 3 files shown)


dododedodonl:
I now notice, in https://github.com/filamentphp/filament/blob/b80fddbdbbbb7ac9d372cf3c7749c3e2680014a7/packages/tables/src/Concerns/CanSortRecords.php#L129 the `$sortDirection` is not used. Is that intended?

danharrin:
The error only happens if the column name is the same in more than one table involved in the query. This is a breaking change for those who might be sorting with a column on a joined table, which works as long as it has a unique name. Manual qualification is fine in this situation IMO.

PR #12005, opened 2024-03-23
Introduce Enhanced Table Row Reordering with Dragged Record Identification

kaspernowak:
## Description

This pull request introduces enhancements to the table row reordering feature within Filament by incorporating the identification of the dragged record alongside the new order of records. The implementation modifies the Sortable.js configuration to dispatch a custom event containing both the new order array and the dragged record's ID. A new method, reorderTableWithDragged, has been introduced to handle this additional data on the backend, providing developers with more flexibility when implementing custom reordering logic that depends on the identity of the dragged record.

The changes aim to enhance the developer experience by making it easier to implement more complex reordering behaviors, such as conditional reordering or reordering with additional side effects that depend on the specific record being moved.

## Visual changes

None. This change affects the functionality of table row reordering without altering the visual appearance or requiring changes to the UI.

## Functional changes
The code style has been adhered to by running the composer cs command.
The changes have been thoroughly tested to ensure they do not break existing functionality within Filament.
Documentation updates are not applicable as this change does not introduce new concepts that require documentation or alter existing documentation.

- [ ] Documentation updates are not applicable as this change does not introduce new concepts that require documentation or alter existing documentation.


danharrin:
I appreciate the effort here, but it doesn't feel like we can implement this in a way that doesn't introduce technical debt. I don't really like introducing the new method, it feels like this really belongs in a major version where we change the method signature of the existing `reorderTable()` method. Which I understand is disappointing for you, but I just don't feel comfortable merging this in its current state.

Feel free to make a new PR to the v4 branch where you introduce a simple `$draggedRecordKey` parameter to the existing `reorderTable()` method. You should not need to change the JS, just pass this in the view from `$event.target.sortable.item.getAttribute('x-sortable-item')` I think?

kaspernowak:
@danharrin Thank you for the review!

It didn't feel good to add the new method either, but it was the only way to avoid breaking changes.
I have created a new PR for the 4.x branch based on your recommendations here:
https://github.com/filamentphp/filament/pull/12096.

Does v4 have a planned release date?

danharrin:
There is no release date, we've only just started working on it

PR #12007, opened 2024-03-23
Native DatePicker: min and max HTML attribute shouldn't contain time

buddhaCode:
## Description

Setting the `minDate()`/`maxDate()` restrictions on a native Date input doesn't restrict the selection in the GUI. Currently, the min/max attribute on the HTML tag contains the time, even when one is just using the `DatePicker` (not `DateTimePicker`). [According to the HTML docs](https://html.spec.whatwg.org/multipage/input.html#date-state-(type=date)), a date input should only have the date and ommit the time.

```php
DatePicker::make('Native Date Picker')
  ->minDate(now()->subDays(3))
  ->maxDate(now()->addDays(3))
```

Produces:

```html
<input [...] max="2024-03-26 18:14:56" min="2024-03-20 18:14:56" type="date" [...]>
```

But should produce:

```html
<input [...] max="2024-03-26" min="2024-03-20" type="date" [...]>
```

The native date time input works fine. But [according to HTML standard](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#valid-local-date-and-time-string), the space between the date and time should be an `T`. But Chrome also accepts the space. And the GUI only restricts the date selection, not the time selection. But I guess thats regular chrome behavior.

The JS date and date time picker work fine as expected.

## Visual changes

Before (tested with Chrome 123 on Mac/Safari 17.2.1 on Mac/Firefox 124 on Mac):

Bildschirmfoto 2024-03-23 um 19 15 05

After:

Bildschirmfoto 2024-03-23 um 19 16 08

## Functional changes


packages/forms/resources/views/components/date-time-picker.blade.php
```diff
@@ -48,8 +48,8 @@
                             'inlinePrefix' => $isPrefixInline && (count($prefixActions) || $prefixIcon || filled($prefixLabel)),
                             'inlineSuffix' => $isSuffixInline && (count($suffixActions) || $suffixIcon || filled($suffixLabel)),
                             'list' => $datalistOptions ? $id . '-list' : null,
-                            'max' => $getMaxDate(),
-                            'min' => $getMinDate(),
+                            'max' => $hasTime() ? $getMaxDate() : substr($getMaxDate(), 0, 10),
```
danharrin:
```suggestion
                            'max' => $hasTime() ? $getMaxDate() : \Carbon\Carbon::parse($getMaxDate())->toDateString(),
```

packages/forms/resources/views/components/date-time-picker.blade.php
```diff
@@ -48,8 +48,8 @@
...
+                            'min' => $hasTime() ? $getMinDate() : substr($getMinDate(), 0, 10),
```
danharrin:
```suggestion
                            'min' => $hasTime() ? $getMinDate() : \Carbon\Carbon::parse($getMinDate())->toDateString(),
```

buddhaCode:
Hey Dan, thanks for reviewing my PRs. What can I do, that you don't have to rewrite my PRs completely? ;)

danharrin:
There's not some general tip I can write that will improve them. I do this to most PRs. There's just some insights that most people don't have that are a little specific to people who have built so much of the framework. In this case, I didn't like the original implementation since I thought it was fragile - if the type of the max date / min date variable changed in the future away from a string then substr wouldn't really work, and if the date format changed to include extra characters then it also would break as the substr indexes would be out. I just changed it to make it so that it hopefully wouldn't break in these cases, to ease maintenance in the future.

With the other PR, there were other places that also needed the implementation. I also copied the implementation from the text column/entry instead of introducing `Collection::wrap()` for consistency. I didn't like that a object collection was being created in all cases even though we didn't use any of the collection methods, it felt like an inefficiency.

PR #12010, opened 2024-03-24
Conditionally apply margin to notification contents

dododedodonl:
## Description

When not using for example a title in a notification, the body still has the margin-top applied.

### Before
before

### After
after

## Functional changes

  - when running this, very many changes were suggested, I only applied the ones on the notification blade file


zepfietje (approved):
Thanks!

PR #12011, opened 2024-03-24
Update 02-listing-records.md

andruu:
If using the static prefix you get the following error `Cannot make non static method Filament\Resources\Pages\ListRecords::table() static in class`

Making non static works


zepfietje:
Thanks!

danharrin:
This is wrong, the docs clearly state `table() method of the resource`

zepfietje:
Sorry, I did check but missed that final word apparently...

PR #12020, opened 2024-03-24
Add `handleRegistration()` method to register page

dododedodonl:
## Description

When customizing the user creation process, the whole register method has to be overwritten. With this, only the creation itself can be customized.

When making this edit, I was not sure what return type to choose. I ended up with choosing `Authenticatable`, the one required for `->login()`. Should the `$user` parameter of `sendEmailVerificationNotification()` not also be changed to `Authenticatable` instead of `Model`?

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/panels/src/Pages/Auth/Register.php
+++ b/packages/panels/src/Pages/Auth/Register.php
@@ -97,7 +97,7 @@ class Register extends SimplePage
     /**
      * @param  array<string, mixed>  $data
      */
-    protected function handleRegistration(array $data): Authenticatable
+    protected function handleRegistration(array $data): Model
     {
         return $this->getUserModel()::create($data);
     }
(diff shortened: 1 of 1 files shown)


zepfietje:
> When making this edit, I was not sure what return type to choose. I ended up with choosing `Authenticatable`, the one required for `->login()`. Should the `$user` parameter of `sendEmailVerificationNotification()` not also be changed to `Authenticatable` instead of `Model`?

Changing the param type of existing methods would be a breaking change, right?

Maybe @danharrin knows the reasoning for using `Model` instead of `Authenticatable`?

packages/panels/src/Pages/Auth/Register.php
```diff
@@ -93,6 +94,14 @@ public function register(): ?RegistrationResponse
         return app(RegistrationResponse::class);
     }
 
+    /**
+     * @param  array<string, mixed>  $data
+     */
+    protected function handleRegistration(array $data): Authenticatable
```
danharrin:
```suggestion
    protected function handleRegistration(array $data): Model
```

danharrin:
Changing to Model for consistency. I don't think there's much reason to have one over the other, but there might be a situation where someone is implementing a custom login/registration flow that does not use Laravel's built in auth logic, so Authenticatable might not be applied. Model just feels safer considering we are interacting with Eloquent database features, and more open to extension.

PR #12023, opened 2024-03-25
Fix logout method when using multi panel.

brkfun:
## Description
When you try to logout from panel-A which is guarded by "web:a".
it will also logs you out from panel-n which is guarded by web:n.
with this PR you can carry your session data remove old login details in session and refresh your session token.
You can remove any comment you want. I wrote them to explain what they are doing.

## Visual changes
Nothing effected.

## Functional changes
Logout will be only done for guard the panel provided. Which makes you logout from one panel only.

### DIFF
--- a/packages/panels/src/Http/Controllers/Auth/LogoutController.php
+++ b/packages/panels/src/Http/Controllers/Auth/LogoutController.php
@@ -9,11 +9,25 @@ class LogoutController
 {
     public function __invoke(): LogoutResponse
     {
-        Filament::auth()->logout();
+        $session = session()->all(); // We will get old session values.
+        Filament::getCurrentPanel()->auth()->logout(); // We will log out only from current panel.
 
-        session()->invalidate();
-        session()->regenerateToken();
+        session()->invalidate(); // it will invalidate and destory our session data.
+        session()->regenerateToken(); // it will regenerate session token.
+        session($this->removeGuardVariables($session)); // it will remove guard values such as password_hash and login from old session data.
+        
+        return app(LogoutResponse::class); // it will send user to logout response.
+    }
 
-        return app(LogoutResponse::class);
+    public function removeGuardVariables(array $session): array
+    {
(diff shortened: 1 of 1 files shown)


brkfun:
Someone tried to implement this at : https://github.com/filamentphp/filament/pull/10408 here.
So i want to finish it.

danharrin:
This feels really dangerous. What if there are other session values that are introduced in the future that are then not handled by this code when putting them back into the session?

Is there a way to only put the session values back that are imperative to keeping the user logged in? It's probably safer then and more maintainable

zepfietje:
Personally I think this code should be in your own projects instead of Filament core.
Some may expect to be logged out of all panels, where others would like to log out of the current panel only.

Combined with this being high risk, I don't think it's worth merging this now, @danharrin.

brkfun:
> Personally I think this code should be in your own projects instead of Filament core. Some may expect to be logged out of all panels, where others would like to log out of the current panel only.
> 
> Combined with this being high risk, I don't think it's worth merging this now, @danharrin.

This one is true as i don't mentioned i already do this in my project with bind. If filament can implement this as feature that would be better than i created. That's a good point if i had a time i would love to write this as a new feature. 
Like configuration to log out from all panels. or an internal bind for LogoutController etc. what do you all have in your mind for this feature is more important.

> This feels really dangerous. What if there are other session values that are introduced in the future that are then not handled by this code when putting them back into the session?
> 
> Is there a way to only put the session values back that are imperative to keeping the user logged in? It's probably safer then and more maintainable

So we can remove everything also instead of login_ and password_hash_ but, if that is inside of the session it means it is intended. Which your notification package is also holding notifications in session. Also "login_" and "password_hash_" are default values they can also change if user implemented something. It has to be configurable from configuration maybe...

danharrin:
Yeah, as per Zep's comment, I think it's safer if users implement this themselves if they know what they're doing. You could release a plugin if you wanted to redistribute

PR #12038, opened 2024-03-26
Adding application/vnd.ms-excel to accepted file types for importing

notscottsmith:
## Description

Firefox looks to be detecting the CSV files as "application/vnd.ms-excel" and not text/csv, et al. Adding this mime type to CanImportRecords.php solves this issue for Firefox users.

## Visual changes

image
image

## Functional changes


packages/actions/src/Concerns/CanImportRecords.php
```diff
@@ -75,7 +75,7 @@ protected function setUp(): void
             FileUpload::make('file')
                 ->label(__('filament-actions::import.modal.form.file.label'))
                 ->placeholder(__('filament-actions::import.modal.form.file.placeholder'))
-                ->acceptedFileTypes(['text/csv', 'text/x-csv', 'application/csv', 'application/x-csv', 'text/comma-separated-values', 'text/x-comma-separated-values', 'text/plain'])
+                ->acceptedFileTypes(['application/vnd.ms-excel', 'text/csv', 'text/x-csv', 'application/csv', 'application/x-csv', 'text/comma-separated-values', 'text/x-comma-separated-values', 'text/plain'])
```
```suggestion
                ->acceptedFileTypes(['text/csv', 'text/x-csv', 'application/csv', 'application/x-csv', 'text/comma-separated-values', 'text/x-comma-separated-values', 'text/plain', 'application/vnd.ms-excel'])
```

danharrin:
I agree that if Excel is outputting CSVs with this type, then we should support it. But I am worried that it will open the input up to users also uploading `.xls` with the same type?

notscottsmith:
> I agree that if Excel is outputting CSVs with this type, then we should support it. But I am worried that it will open the input up to users also uploading `.xls` with the same type?

Excel formatted files from the 97-2008 era (.xls) are recognised as the same mime-type so yes, there is a percentage chance of that occurring. Modern Excel files (.xlsx) return as application/vnd.openxmlformats-officedocument.spreadsheetml.sheet. Perhaps a secondary check to exclude file extensions of type .xls?

The [PhpSpreadsheet](https://phpspreadsheet.readthedocs.io/en/latest/) package allows for the reading/importing of xls/xlsx files, but switching over to that library that would also mean changing the underlying processes within the import/export of files.

danharrin:
Not interested in changing the underlying process as it would slow the import process down *a lot*, currently it is very lean

What's the best way to implement the extension check?

notscottsmith:
I'd say the fastest would be using the builtin [pathinfo](https://www.php.net/manual/en/function.pathinfo.php) function supplied with the PATHINFO_EXTENSION option - this would yield just the extension from the file.

$ext = pathinfo($filename, PATHINFO_EXTENSION);

danharrin:
Please add and test `->rule('extensions:csv')` and ensure that only .csv files can be uploaded and not .xls

notscottsmith:
> Please add and test `->rule('extensions:csv')` and ensure that only .csv files can be uploaded and not .xls

In the CanImportRecords.php file? I'm not sure where to add what you're asking...

danharrin:
Does it work on the FileUpload?

buddhaCode:
> The [PhpSpreadsheet](https://phpspreadsheet.readthedocs.io/en/latest/) package allows for the reading/importing of xls/xlsx files, but switching over to that library that would also mean changing the underlying processes within the import/export of files.

@notscottsmith Filament currently uses `openspout/openspout` which also supports Excel files anyway.

danharrin:
Only for exports, not imports. It would change the underlying process and make it slower

buddhaCode:
`openspout/openspout` does support reading and writing Excel files.
Yes, but Filament's import action is limited to CSVs only.

I already asked in Discord a few days ago, why there is no Excel import. You clearly have your reasons.

But, why would an Excel import be slower, compared to a CSV import? And does it really matter? The import would be processed in background as a queue job anyway. So there would be no direct effect on the user experience.

danharrin:
We do not use OpenSpout for reading CSVs, we use league/csv.

The time/memory consuming part is reading the Excel file row by row and dispatching jobs. The rest is the same. That part is very performant with CSVs as each record is on a new line. Excel exports will never be able to match the efficiency of CSV ones and it could result in a lot of expectations being placed on the system that the core team does not have the resources to deal with right now. I'd need to see some very strong numbers to consider supporting it.

aSeriousDeveloper:
> Please add and test `->rule('extensions:csv')` and ensure that only .csv files can be uploaded and not `.xls`

I've made this change on a local repo and it half-works.

For uploading a `.csv` via Firefox, it now works as expected.

`.xls` files can still be uploaded and it allowed you to select a column. It contains garbled data.

image

However, once you press submit you're given a validation error that your file must have the `.csv` extension.

image

### Context

Additionally, I was able to gather some extra context about the edge case and why it happens. Windows stores file Content-Types within the registry, and these are used to check what software should open a file. If one has MS Office / Excel installed, the Content-Type for `.csv` files is changed from `text/csv` to `application/vnd.ms-excel`.

From here, browsers differ from how they gather MIME Type. Seemingly, chromium based browsers have their own built-in checks and gather that `.csv` files are of type `text/csv`. However, Firefox (which uses Gecko) simply uses the Content-Type within the registry that in this case ends up being `application/vnd.ms-excel`.

# Update

I have a slightly better fix for this locally now. Not sure if / how to push to someone else's PR but I can include the snippet here.

If you perform the validation within `afterStateUpdated`, then catch, set the file to null, then rethrow the validation error, you stop `.xls` files being uploaded.

image

Here's a snippet for how I've built the `FileUpload` form field:

```php
FileUpload::make('file')
    ->label(__('filament-actions::import.modal.form.file.label'))
    ->placeholder(__('filament-actions::import.modal.form.file.placeholder'))
    ->rules(['extensions:csv'])
    ->acceptedFileTypes(['text/csv', 'text/x-csv', 'application/csv', 'application/x-csv', 'text/comma-separated-values', 'text/x-comma-separated-values', 'text/plain', 'application/vnd.ms-excel'])
    ->afterStateUpdated(function (Forms\Set $set, ?TemporaryUploadedFile $state, FileUpload $component, Livewire $livewire) use ($action) {
        if (! $state instanceof TemporaryUploadedFile) {
            return;
        }

        $name = $component->getName();
        try {
            $livewire->validateOnly($component->getStatePath());
        }
        catch(ValidationException $e) {
            $component->state(null);
            throw $e;
        }

        // Rest of CSV stream logic goes here...
        
        }, []));
    })
    ->storeFiles(false)
    ->visibility('private')
    ->required()
    ->hiddenLabel(),
```

danharrin:
Thanks for the investigation @aSeriousDeveloper, I've made those changes to this PR myself so let's see how it goes

giodi:
After updating, the aforementioned validation error "The file field must have one of the following extensions: csv" persists. 
grafik
This seems because the file extension of the uploaded CSV gets changed to .txt in the TemporaryUploadedFile.
I found some discussions in the livewire project about this: [Proposal for Enhanced File Extension Detection Using getClientOriginalExtension() ](https://github.com/livewire/livewire/discussions/8083)

Would it be an option to add txt to the rule?

danharrin:
@giodi please PR that addition

PR #12048, opened 2024-03-26
Make commands lazy

stephenjude:
## Description

This PR makes the Filament commands "lazy", i.e., they do not need to be instantiated when running other commands.

Currently, when you have Filament installed in a Laravel project, if you were to run php artisan `filament:install` all of the Filament commands are instantiated. This is just how Symfony console commands work.

The `AsCommand` attribute was introduced to allow commands to be "lazy", i.e., they do not need to be instantiated when running other commands.

@timacdonald has been doing some work across the Laravel ecosystem to standardise this and I thought to drop by here and do the same.

See: https://github.com/laravel/framework/pull/50617

## Functional changes

### DIFF
--- a/packages/actions/src/Commands/Aliases/MakeExporterCommand.php
+++ b/packages/actions/src/Commands/Aliases/MakeExporterCommand.php
@@ -3,7 +3,9 @@
 namespace Filament\Actions\Commands\Aliases;
 
 use Filament\Actions\Commands;
+use Symfony\Component\Console\Attribute\AsCommand;
 
+#[AsCommand(name: 'filament:exporter')]
 class MakeExporterCommand extends Commands\MakeExporterCommand
 {
     protected $hidden = true;
--- a/packages/tables/src/Commands/MakeTableCommand.php
+++ b/packages/tables/src/Commands/MakeTableCommand.php
@@ -8,9 +8,11 @@ use Filament\Support\Commands\Concerns\CanReadModelSchemas;
 use Filament\Tables\Commands\Concerns\CanGenerateTables;
 use Illuminate\Console\Command;
 use Illuminate\Support\Str;
+use Symfony\Component\Console\Attribute\AsCommand;
 
 use function Laravel\Prompts\text;
 
+#[AsCommand(name: 'make:livewire-table')]
 class MakeTableCommand extends Command
 {
     use CanGenerateTables;
--- a/packages/actions/composer.json
+++ b/packages/actions/composer.json
@@ -19,8 +19,7 @@
         "illuminate/support": "^10.45|^11.0",
         "league/csv": "^9.14",
         "openspout/openspout": "^4.23",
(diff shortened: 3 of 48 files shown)


packages/actions/src/Commands/Aliases/MakeExporterCommand.php
```diff
@@ -3,7 +3,9 @@
 namespace Filament\Actions\Commands\Aliases;
 
 use Filament\Actions\Commands;
+use Symfony\Component\Console\Attribute\AsCommand;
 
+#[AsCommand(name: 'make:filament-exporter', aliases: ['filament:exporter'])]
```
danharrin: For the alias command classes, do the arguments need to be swapped? So `make:filament-exporter` is an alias of `filament:exporter`? Maybe I am understanding it wrong, but Symfony's docs do not explain this

danharrin (changes requested):

packages/tables/src/Commands/MakeTableCommand.php
```diff
@@ -24,11 +26,13 @@ class MakeTableCommand extends Command
 
     public function handle(): int
     {
-        $component = (string) str($this->argument('name') ?? text(
-            label: 'What is the table name?',
-            placeholder: 'Products/ListProducts',
-            required: true,
-        ))
+        $component = (string) str(
```
danharrin: Please remove all formatting changes in this file and also MakeWidgetCommand

packages/actions/src/Commands/Aliases/MakeExporterCommand.php
danharrin: Maybe we dont need `aliases: ` at all yet since we aren't using the `$aliases` property on the commands? These could be treated as separate commands where the alias classes just reference their own name?

Then we refactor to `$aliases` and get rid of these classes in v4

stephenjude: Yes, I think we should treat these alias as separate command. I will go ahead and do that.

packages/tables/src/Commands/MakeTableCommand.php
stephenjude: Could this formatting be as a result of `composer cs` command or maybe from my IDE formatting.

packages/actions/src/Commands/Aliases/MakeExporterCommand.php
stephenjude: > For the alias command classes, do the arguments need to be swapped? So `make:filament-exporter` is an alias of `filament:exporter`? Maybe I am understanding it wrong, but Symfony's docs do not explain this

For swapping of the arguments, I don't think so but I will test it out and see where that leads me.

stephenjude:
@danharrin I have made all necessary changes and  rebased the commits into one. I think this PR is ready. Let me know if there anything else that needs my attention.

PR #4916, opened 2022-11-12
fix: Wizard submit validation

danharrin:
Fixes #4496.

PR #12054, opened 2024-03-27
Handle cases where Wizard Step is not found (hidden steps)

AlexisSerneels:
## Description

Found related issue : https://github.com/filamentphp/filament/pull/4916#issuecomment-1315657935

I encountered the same exception in my project:
```
Undefined array key -1 at /vendor/filament/forms/src/Components/Wizard.php:78
```

It happens because the method `findIndex(...)`returns -1 if no item is found.
Thus the `getStepIndex()`returns -1 which does not exists and throws a 500 Exception.

This PR fixes this issue by returning the first Wizard step instead.

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/forms/resources/views/components/wizard.blade.php
+++ b/packages/forms/resources/views/components/wizard.blade.php
@@ -49,6 +49,7 @@
 
         getStepIndex: function (step) {
             let index = this.getSteps().findIndex((indexedStep) => indexedStep === step)
+            
             if (index === -1) {
                 index = this.getSteps()[0]
             }
--- a/packages/forms/resources/views/components/wizard.blade.php
+++ b/packages/forms/resources/views/components/wizard.blade.php
@@ -51,7 +51,7 @@
             let index = this.getSteps().findIndex((indexedStep) => indexedStep === step)
             
             if (index === -1) {
-                index = this.getSteps()[0]
+                return 0
             }
             
             return index
--- a/packages/forms/resources/views/components/wizard.blade.php
+++ b/packages/forms/resources/views/components/wizard.blade.php
@@ -48,7 +48,7 @@
         },
 
         getStepIndex: function (step) {
-            let index = this.getSteps().findIndex((indexedStep) => indexedStep === step);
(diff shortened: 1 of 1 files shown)


packages/forms/resources/views/components/wizard.blade.php
```diff
@@ -48,7 +48,11 @@
         },
 
         getStepIndex: function (step) {
-            return this.getSteps().findIndex((indexedStep) => indexedStep === step)
+            let index = this.getSteps().findIndex((indexedStep) => indexedStep === step);
+            if (index == -1) {
+                index = this.getSteps()[0]
+            }
+            return index
```
```suggestion
            if (index === -1) {
                index = this.getSteps()[0]
            }
            
            return index
```

packages/forms/resources/views/components/wizard.blade.php
```diff
@@ -48,7 +48,12 @@
         },
 
         getStepIndex: function (step) {
-            return this.getSteps().findIndex((indexedStep) => indexedStep === step)
+            let index = this.getSteps().findIndex((indexedStep) => indexedStep === step);
```
```suggestion
            let index = this.getSteps().findIndex((indexedStep) => indexedStep === step)
```

packages/forms/resources/views/components/wizard.blade.php
```diff
@@ -48,7 +48,12 @@
         },
 
         getStepIndex: function (step) {
-            return this.getSteps().findIndex((indexedStep) => indexedStep === step)
+            let index = this.getSteps().findIndex((indexedStep) => indexedStep === step)
+            if (index === -1) {
```
danharrin, line 53:
```suggestion
            
            if (index === -1) {
```

danharrin:
To be honest, it doesn't feel like you've tested this. `index = this.getSteps()[0]` assigns the ID of the step instead of its index doesn't it? Doesn't it just need to be `index = 0`?

AlexisSerneels:
Sorry about the wrong copy/paste into Github Issue PR. You are correct.
Thank you for the strict comparison operator correction.

PR #12058, opened 2024-03-27
fix(docs): import and use of livewire testing class

discoverlance-com:
Pest does not have the livewire testing function, `use function Pest\Livewire\livewire;` but we can get it from Livewire,  `use Livewire\Livewire;` .

## Description

I updated the docs testing markdown files to use the livewire class for testing, `use Livewire\Livewire;`  and changed the places where the pest livewire was being used to  instead use `Livewire::test` .

## Visual changes

Example from `packages/actions/docs/09-testing.md`:
Screenshot 2024-03-27 15 38 59

## Functional changes

### DIFF
--- a/packages/actions/docs/09-testing.md
+++ b/packages/actions/docs/09-testing.md
@@ -13,12 +13,12 @@ Since all actions are mounted to a Livewire component, we're just using Livewire
 You can call an action by passing its name or class to `callAction()`:
 
 ```php
-use function Pest\Livewire\livewire;
+use Livewire\Livewire;
 
 it('can send invoices', function () {
     $invoice = Invoice::factory()->create();
 
-    livewire(EditInvoice::class, [
+    Livewire::test(EditInvoice::class, [
         'invoice' => $invoice,
     ])
         ->callAction('send');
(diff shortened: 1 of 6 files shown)


danharrin:
It does have it - https://pestphp.com/docs/plugins#content-livewire ?

Am I missing something? Please do let me know if I'm being stupid here

discoverlance-com:
> It does have it - https://pestphp.com/docs/plugins#content-livewire ?
> 
> Am I missing something? Please do let me know if I'm being stupid here

Oh yes, No you are not. I missed it I think. It's just that in the default pest documentation, it's not really described there so I also missed it and assumed there's no such function since I could not see it in the documentation or after installing pest. Can it be included instead in the docs? To state that one would need to install that plugin? Because it looks as if it's present directly from using pest. In the documentation of filament testing, there's a link to [livewire testing helpers](https://livewire.laravel.com/docs/testing) but it's also pointing to the livewire docs on testing which has all the testing functionalities that the pest plugin exposes so it's not really apparent from the docs that one needs the plugin from pest.

Thanks for the correction.

danharrin:
Yeah, you could add a link to the plugin if you want to all testing pages

discoverlance-com:
> Yeah, you could add a link to the plugin if you want to all testing pages

Can that be done directly or should I suggest a different pull request with such changes?

danharrin:
It's probably easier if you do a new PR because there are so many changes to revert here

PR #12070, opened 2024-03-28
Enhancement: Tooltip for disabled buttons, title attribute on icon-button without tooltip

andrewdwallo:
## Description

This PR resolves an issue where icon buttons displayed both tooltips and title attributes simultaneously. Now, the title attribute is only shown when a tooltip is not present. Additionally, it introduces the ability for disabled buttons to display tooltips, achieved by wrapping the elements in a `<span>` when a tooltip is applied. This wrapper is omitted for elements without tooltips.

To address formatting errors reported by `prettier-plugin-blade`, such as "encountered an unexpected closing tag" due to Prettier's interpretation of Blade syntax within conditionals, I've implemented `{{-- format-ignore-start --}}` and `{{-- format-ignore-end --}}` tags around the affected code blocks. This ensures compatibility without altering the intended HTML structure.

### Before
Screenshot 2024-03-27 at 11 15 18 PM

### After
Screenshot 2024-03-27 at 11 13 08 PM
Screenshot 2024-03-27 at 11 14 18 PM

## Functional changes

### DIFF
--- a/packages/forms/resources/views/components/wizard.blade.php
+++ b/packages/forms/resources/views/components/wizard.blade.php
@@ -48,12 +48,14 @@
         },
 
         getStepIndex: function (step) {
-            let index = this.getSteps().findIndex((indexedStep) => indexedStep === step)
-            
+            let index = this.getSteps().findIndex(
+                (indexedStep) => indexedStep === step,
+            )
+
             if (index === -1) {
                 return 0
             }
-            
+
             return index
         },
 
--- a/packages/support/resources/views/components/badge.blade.php
+++ b/packages/support/resources/views/components/badge.blade.php
@@ -72,21 +72,26 @@
     $hasTooltip = filled($tooltip);
 @endphp
 
-<{{ $tag }}
-    @if ($tag === 'a')
-        {{ \Filament\Support\generate_href_html($href, $target === '_blank', $spaMode) }}
-    @endif
-    @if ($keyBindings || $hasTooltip)
+@if ($hasTooltip)
+    {{-- format-ignore-start --}}<span
         x-data="{}"
-    @endif
-    @if ($keyBindings)
-        x-mousetrap.global.{{ collect($keyBindings)->map(fn (string $keyBinding): string => str_replace('+', '-', $keyBinding))->implode('.') }}
-    @endif
-    @if ($hasTooltip)
+        class="inline-flex"
         x-tooltip="{
             content: @js($tooltip),
             theme: $store.theme,
         }"
+    >{{-- format-ignore-end --}}
+@endif
+
+<{{ $tag }}
+    @if ($tag === 'a')
+        {{ \Filament\Support\generate_href_html($href, $target === '_blank', $spaMode) }}
+    @endif
+    @if ($keyBindings)
+        x-data="{}"
+    @endif
+    @if ($keyBindings)
+        x-mousetrap.global.{{ collect($keyBindings)->map(fn (string $keyBinding): string => str_replace('+', '-', $keyBinding))->implode('.') }}
     @endif
     {{
         $attributes
@@ -233,3 +238,7 @@
         @endif
     @endif
 </{{ $tag }}>
+
+@if ($hasTooltip)
+    {{-- format-ignore-start --}}</span>{{-- format-ignore-end --}}
+@endif
--- a/packages/support/resources/views/components/button/index.blade.php
+++ b/packages/support/resources/views/components/button/index.blade.php
@@ -173,21 +173,26 @@
     />
 @endif
 
-<{{ $tag }}
-    @if ($tag === 'a')
-        {{ \Filament\Support\generate_href_html($href, $target === '_blank', $spaMode) }}
-    @endif
-    @if (($keyBindings || $hasTooltip) && (! $hasFormProcessingLoadingIndicator))
+@if ($hasTooltip)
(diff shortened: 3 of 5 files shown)


zepfietje (commented):
Tooltips and the title attribute don't necessarily provide the same information, so we can't just get rid off the title attribute when a tooltip is present.

Also, wrapping the button in another element will likely cause styling issues and [isn't advised by the creator of Tippy.js](https://github.com/floating-ui/floating-ui/issues/1841#issuecomment-1543495036) (we will probably migrate to the successor: Floating UI).

I'm going to close this PR, but feel free to look into different approaches such as using `aria-disabled`. You do need to make sure then that all disabled states are also handled when "disabling" using `aria-disabled` instead of using the `disabled` HTML attribute.

andrewdwallo:
I tested everything and it didn't cause any styling issues. It's valid html markup. Also I asked @danharrin if I could make a PR to only have the title attribute on the icon button when the tooltip is not present and he said yes. As for migrating to Floating UI for the tooltip I'm totally down for it, just hope it happens. I guess I will just make a custom solution for this for now then.

andrewdwallo:
Also why would you want to keep the title attribute when a user is using a tooltip?? If accessibility is a concern then why not just use aria-label instead of title?

danharrin:
Maybe it's worth removing the title and replacing with aria-label **only when the title and the tooltip are identical**?

zepfietje:
> **only when the title and the tooltip are identical**?

That's the important part that I meant indeed.

andrewdwallo:
And what if they are different? What if the tooltip is conveying something entirely different from the label (which is used for the title attribute)? Wouldn't that cause confusion? The title attribute is only being used for the icon-button right now.

zepfietje:
If tooltip and title content are different, they should both be shown.
The title describes the action (because icon buttons don't have a visible label) and tooltips may contain whatever info you decide.

andrewdwallo:
Idk I guess I just think the reason why someone may use the tooltip in the first place is to describe the action.. which is why I think the title attribute should be hidden when a tooltip is shown. I actually can't think of any other reason why someone would have a tooltip for an icon button other than to describe the action.  Buts it's fine.

zepfietje:
A simple example:

Label/title: "Block user"
Tooltip: "Blocking a user will hide all their public posts and prevent them from logging in to their account."

It would be an accessibility issue if screenreader read the entire tooltip instead of the actual action as indicated by the label.

andrewdwallo:
And what about when you would most likely also need to do the reverse of this?

For example:
"""
Label/title: "Unblock user"
Tooltip: "Unblocking a user will show their public posts and allow them to login to their account.
"""

Or I guess in this case it would show this if the label was the same:

"""
Label/title: "Block user"
Tooltip: "Unblocking a user will show their public posts and allow them to login to their account.
"""

andrewdwallo:
I understand yalls point though. Thanks.

zepfietje:
Block and unblock are different actions, so it's up to the dev to handle that however they like in their projects. I don't see how that's something that should be handled in Filament.

andrewdwallo:
> Block and unblock are different actions, so it's up to the dev to handle that however they like in their projects. I don't see how that's something that should be handled in Filament.

I mean they could be different actions if you want to duplicate code and if you don't want to utilize dynamism in Filament.

For example, for the code in the picture above:
```php
Tables\Actions\Action::make('markAsReviewed')
          ->label('Mark as Reviewed')
          ->iconButton()
          ->icon(static fn (Transaction $record) => $record->reviewed ? 'heroicon-s-check-circle' : 'heroicon-o-check-circle')
          ->color(static fn (Transaction $record, Tables\Actions\Action $action) => match (static::determineTransactionState($record, $action)) {
                        'reviewed' => 'primary',
                        'unreviewed' => Color::rgb('rgb(' . Color::Gray[600] . ')'),
                        'uncategorized' => 'gray',
          })
          ->tooltip(static fn (Transaction $record, Tables\Actions\Action $action) => match (static::determineTransactionState($record, $action)) {
                        'reviewed' => 'Reviewed',
                        'unreviewed' => 'Mark as Reviewed',
                        'uncategorized' => 'Categorize first to mark as reviewed',
          })
          ->disabled(fn (Transaction $record): bool => $record->isUncategorized())
          ->action(fn (Transaction $record) => $record->update(['reviewed' => ! $record->reviewed])),
```

Block and Unblock could be utilized the same way as above. I guess it just depends on the developers coding style and practices.

danharrin:
I think there should be:

`titleTooltip(?bool $condition)` (null by default):
- `null` only show title tooltip when the tooltip is different, else aria label
- `true` always show the title tooltip regardless of the tooltip
- `false` always hide the title tooltip and replace with an aria-label regardless of the tooltip

andrewdwallo:
That would work.

zepfietje:
> Block and Unblock could be utilized the same way as above. I guess it just depends on the developers coding style and practices.

Still doesn't matter, because you'd use a dynamic label and tooltip depending on what the action is currently used for.

andrewdwallo:
Yep you're right. No worries I will just use a blank/hidden label and a dynamic aria-label. It's what I want for my use case anyway.

danharrin:
Another request: titleTooltip() should also accept a string, and if it is set then it should use that text all the time instead regardless of tooltip()

PR #12091, opened 2024-03-29
Make it possible to submit form from outside itself

dododedodonl:
## Description

Usecase: move the "save" button to the header actions of EditRecord or CreateRecord.  
Using the [form HTML feature](https://stackoverflow.com/a/12567605)

- [packages/support/resources/views/components/button/index.blade.php#L199](https://github.com/filamentphp/filament/blob/75f4cbd2353066a737026e21d6931d5217376a98/packages/support/resources/views/components/button/index.blade.php#L199) form selection will probably not work anymore

## Visual changes

None

## Functional changes

  - `npm run prettier` suggests quite some changes to untouched files, I did not commit those

### DIFF
--- a/packages/panels/resources/views/resources/pages/create-record.blade.php
+++ b/packages/panels/resources/views/resources/pages/create-record.blade.php
@@ -4,8 +4,12 @@
         'fi-resource-' . str_replace('/', '-', $this->getResource()::getSlug()),
     ])
 >
+    @php
+        $identifier = $this->getId() . '.forms.' . $this->getFormStatePath();
+    @endphp
     <x-filament-panels::form
-        :wire:key="$this->getId() . '.forms.' . $this->getFormStatePath()"
+        :id="$identifier"
+        :wire:key="$identifier"
         wire:submit="create"
     >
         {{ $this->form }}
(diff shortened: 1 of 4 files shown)


dododedodonl:
When speaking to @zepfietje, he told `formToSubmit` was intended for [this](https://stackoverflow.com/questions/7020659/submit-form-using-a-button-outside-the-form-tag/12567605#12567605) use case. But I am confused as to what `\Filament\Actions\Concerns\CanSubmitForm->formToSubmit` contains.
- in `EditRecord` it is set to `save` on the submit action
- in `CreateRecord` it is set to `create` on the submit action
It looks like the value only gets set as `wire:target` on the loading indicator.

@danharrin can you maybe shed some light on this?

danharrin:
It isn't intended for that use case, its just intended for setting the correct wire:target so the loading indicator displays at the correct time.

We cannot change the behaviour of the existing property (BC) and theres no real need to introduce an ID or wire:key on the form.

Please open a PR as follows:
- New `formId(?string $id)` method in that trait
- New `?string $formId = null` property in that trait
- New `getFormId(): ?string` method in that trait
- Utilize the `getFormId()` to set `form` on the button if it is set

Thanks!

PR #12095, opened 2024-03-29
BugFix: Import Action Look at File System Disk Driver for s3

mmccall10:
## Description

When a filesystems disk name is not s3 the import action fails with `fopen` regardless of `FILAMENT_FILESYSTEM_DISK` and `FILESYSTEM_DISK` being set to the s3 driver. This PR changes the lookup to check the driver of the configured disks vs the name.

## Visual changes
N/A

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/CanImportRecords.php
+++ b/packages/actions/src/Concerns/CanImportRecords.php
@@ -349,7 +349,7 @@ trait CanImportRecords
     {
         $filePath = $file->getRealPath();
 
-        if (config('filesystems.disks.'.config('filament.default_filesystem_disk').'.driver') !== 's3') {
+        if (config('filesystems.disks.' . config('filament.default_filesystem_disk') . '.driver') !== 's3') {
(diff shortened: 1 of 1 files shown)


packages/actions/src/Concerns/CanImportRecords.php
```diff
@@ -349,7 +349,7 @@ public function getUploadedFileStream(TemporaryUploadedFile $file)
     {
         $filePath = $file->getRealPath();
 
-        if (config('filament.default_filesystem_disk') !== 's3') {
+        if (config('filesystems.disks.'.config('filament.default_filesystem_disk').'.driver') !== 's3') {
```
```suggestion
        if (config('filesystems.disks.' . config('filament.default_filesystem_disk') . '.driver') !== 's3') {
```

telkins:
Happy to see this fix.  We ran into this a few weeks ago and just haven't had the bandwidth to try to address it via PR.

Thx, @mmccall10  and @danharrin !  🤓

PR #12106, opened 2024-03-30
Add formId to actions

dododedodonl:
## Description

Continuation of #12091 

## Visual changes

None

## Functional changes

  - `npm run prettier` suggests quite some changes to untouched files, I did not commit those

### DIFF
--- a/packages/actions/src/Concerns/CanSubmitForm.php
+++ b/packages/actions/src/Concerns/CanSubmitForm.php
@@ -8,6 +8,8 @@ trait CanSubmitForm
 
     protected ?string $formToSubmit = null;
 
+    protected ?string $formId = null;
+
     public function submit(?string $form): static
     {
         $this->canSubmitForm = filled($form);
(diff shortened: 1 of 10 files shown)


dododedodonl:
https://github.com/filamentphp/filament/pull/12091#issuecomment-2027815214
> (...)
> theres no real need to introduce an ID or wire:key on the form.
> (...)

@danharrin The `wire:key` was already there. The `form=""` attribute has to contain the ID of the form to submit, so for this to work in a panel, the form has to have an ID. Is adding a default one, equal to the `wire:key`, a valid option?

packages/actions/src/Concerns/CanSubmitForm.php
```diff
@@ -25,4 +27,16 @@ public function getFormToSubmit(): ?string
...
     }
+
+    public function formId(?string $id): static
+    {
+        $this->formId = $id;
+
+        return $this;
+    }
+
+    public function getFormid(): ?string
```
danharrin:
```suggestion
    public function getFormId(): ?string
```

danharrin:
> The wire:key was already there

My mistake

> for this to work in a panel, the form has to have an ID

For user friendliness, we can probably just assign the edit/create forms an ID of `form`

danharrin:
Thanks

PR #12111, opened 2024-03-30
Use CSS content-based sizing property for textarea elements

ericmp33:
## Description

Javascript is not needed anymore to calculate the textarea height sizing. Instead, we can use the new field sizing CSS property called `field-sizing`.

⚠️ Important note - this property is experimental, which means not all browsers suport it yet. I make the PR so y'all know it exists and when more browsers suport it can be merged. For now, I wouldn't merge it.

This PR is about the textarea, but this content based sizing style also works for more than a textarea.

Relevant links:
- https://caniuse.com/?search=field-sizing
- https://developer.chrome.com/docs/css-ui/css-field-sizing

## Visual changes

There is no visual changes.

## Functional changes

### DIFF
--- a/packages/forms/resources/js/components/textarea.js
+++ b/packages/forms/resources/js/components/textarea.js
@@ -1,14 +1,14 @@
 export default function textareaFormComponent({ initialHeight }) {
     return {
         init: function () {
-            this.render()
+            if (this.$el.scrollHeight > 0) {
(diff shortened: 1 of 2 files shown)


danharrin:
Thanks for letting us know, that's a great replacement. But we don't typically keep future PRs open, as they can cause congestion in the repository. It would be great if you could reopen or @ me when you think the browser support is good enough!

PR #12120, opened 2024-03-31
Update Register.php

sitenzo:
## Description

Updated the registration to handle relation data if user customize the form, also added the default hooks for is a user wants to manipulate the data it is possible to do so without the override of the register function

## Visual changes

None,

## Functional changes

### DIFF
--- a/packages/panels/src/Pages/Auth/Register.php
+++ b/packages/panels/src/Pages/Auth/Register.php
@@ -87,15 +87,15 @@ class Register extends SimplePage
 
             $this->callHook('afterValidate');
 
-            $data = $this->mutateFormDataBeforeCreate($data);
+            $data = $this->mutateFormDataBeforeRegister($data);
 
-            $this->callHook('beforeCreate');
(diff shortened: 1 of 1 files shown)


packages/panels/src/Pages/Auth/Register.php
```diff
@@ -77,9 +81,23 @@ public function register(): ?RegistrationResponse
...
-            return $this->handleRegistration($data);
+            $this->callHook('afterValidate');
+
+            $data = $this->mutateFormDataBeforeCreate($data);
+
+            $this->callHook('beforeCreate');
+
+            $this->record = $this->handleRegistration($data);
+
+            $this->form->model($this->record)->saveRelationships();
```
dododedodonl, line 96: Shouldn't this also be done in the handleRegistration method?

danharrin: This is consistent with the create page

danharrin:
Thanks

PR #12122, opened 2024-03-31
Fix long label in CheckboxList Component

agencetwogether:
## Description

Adds the `break-all` class for label in CheckboxList.
When using multi columns and searchable() method, if a label is very long, search layout overflow.

## Visual changes

Before `break-all`
before

After => `break-all`applied
after

## Functional changes

### DIFF
--- a/packages/forms/resources/views/components/checkbox-list.blade.php
+++ b/packages/forms/resources/views/components/checkbox-list.blade.php
@@ -196,7 +196,7 @@
 
                         <div class="grid text-sm leading-6">
                             <span
-                                class="fi-fo-checkbox-list-option-label font-medium text-gray-950 dark:text-white"
+                                class="fi-fo-checkbox-list-option-label break-all font-medium text-gray-950 dark:text-white"
                             >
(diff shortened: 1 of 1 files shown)


packages/forms/resources/views/components/checkbox-list.blade.php
```diff
@@ -196,7 +196,7 @@ class="fi-fo-checkbox-list-option-label flex gap-x-3"
 
                         <div class="grid text-sm leading-6">
                             <span
-                                class="fi-fo-checkbox-list-option-label font-medium text-gray-950 dark:text-white"
+                                class="fi-fo-checkbox-list-option-label break-all font-medium text-gray-950 dark:text-white"
```
zepfietje:
```suggestion
                                class="fi-fo-checkbox-list-option-label overflow-hidden break-words font-medium text-gray-950 dark:text-white"
```

zepfietje:
The initial approach wasn't right, as it would break all words without trying to preserve words where possible. I've updated this PR to use the right approach.

PR #12128, opened 2024-04-01
Relation manager content tab improvements

binaryfire:
## Description

Adds the ability to set the form tab's position and icon when using `hasCombinedRelationManagerTabsWithContent()`. This is useful when the primary focus of the page should be adding related records instead of editing the owner record itself. In these cases, the owner record form is more like a "settings" section which should take a back seat in the UI.

## Visual changes

Before:
image

After, with `getContentTabIcon()` and `getContentTabPosition()` set in the page class:
image


packages/panels/resources/views/components/resources/relation-managers.blade.php
```diff
@@ -26,7 +28,10 @@
                 $tabs = $managers;
 
                 if ($content) {
-                    $tabs = array_replace([null => null], $tabs);
+                    match ($contentTabPosition) {
+                        \Filament\Support\Enums\ContentPosition::After => $tabs = $tabs + ["" => null],
```
danharrin: Can this be done with `array_merge()`, is the result the same? That would be consistent. Also with a `null` array key instead of an empty string?

packages/support/src/Enums/ContentPosition.php
danharrin, line 1: Please move this to `Filament\Resources\Pages` and rename to `ContentTabPosition`

packages/panels/docs/03-resources/07-relation-managers.md
```diff
@@ -665,6 +665,28 @@ public function hasCombinedRelationManagerTabsWithContent(): bool
 }
 ```
 
+## Setting an icon for the form tab
```
danharrin:
```suggestion
### Setting an icon for the form tab
```

packages/panels/docs/03-resources/07-relation-managers.md
```diff
@@ -665,6 +665,28 @@ public function hasCombinedRelationManagerTabsWithContent(): bool
...
+On the Edit or View page class, override the `getContentTabIcon()` method:
+
+```php
+public function getContentTabIcon(): ?string
+{
+    return 'heroicon-m-cog';
+}
+```
+
+## Setting the position of the form tab
```
danharrin:
```suggestion
### Setting the position of the form tab
```

packages/panels/docs/03-resources/07-relation-managers.md
```diff
@@ -665,6 +665,28 @@ public function hasCombinedRelationManagerTabsWithContent(): bool
...
+
+By default, the form tab is rendered on the left. To render it on the right, you can override the `getContentTabPosition()` method on the Edit or View page class:
```
danharrin:
```suggestion
By default, the form tab is rendered before the relation tabs. To render it after, you can override the `getContentTabPosition()` method on the Edit or View page class:
```

danharrin (changes requested):

packages/support/src/Enums/ContentPosition.php
binaryfire: @danharrin Do you mean `Filament\Resources\Pages\Enums`? There's no `Enums` dir under `Filament\Resources\Pages` atm

danharrin: I mean what I wrote

binaryfire: Ok done. Although my OCD is being triggered by having an enum outside an Enums directory.

danharrin:
Thanks

a-tarek:
@binaryfire :wave: 
is there a way to set the default page ? 
I am setting this up but when visiting the Edit post page it goes straight to the settings sub tab 

`/admin/posts/1/edit`

is there a way to set the default page to 
`/admin/posts/1/edit/activeRelationManager=0`  ?


```php
{
  return [
      'edit' => Pages\EditPost::route('/{record}/edit')
    ];
 }
```

PR #12146, opened 2024-04-02
Custom background color on modal blade component

tanthammar:
Example when you want to style the background of a modal component:

```html
<x-filament::modal 
    id="custom-error" 
    background-color="!bg-danger-500"
    close-button-classes="!bg-white/5 !rounded-full !text-gray-200 hover:!text-white"
    >

        <x-slot name="heading">Missing something</x-slot>
        
        <div class="text-white">Custom error</div>

</x-filament::modal>


danharrin:
Closing in favor of #12155 which would let you do this with CSS hooks as usual

PR #12148, opened 2024-04-02
Add isFiltered property and getter

binaryfire:
EDIT: I reverted the changes and just added an `isFiltered` property instead. See https://github.com/filamentphp/filament/pull/12148#issuecomment-2036231828

Currently a table's `emptyState()` is also used when no search / filter results are found. So using the example from the docs, that means you'd see this when you've already got bunch of posts but your search doesn't match anything: 

image

Which doesn't really make sense... 

This PR adds a separate "no results" state which is displayed when no search / filter results are found. The `emptyState()` is now only shown when a resource is truly empty, meaning it can be fully customized as a welcome screen (inviting users to create their first record) without worrying about how it'll look when no search results are returned. It uses sensible defaults and can be customized the same way as `emptyState()`.

I'm not sure what you want to do re: docs so I didn't change the md file numbering. Do you want to combine the empty state and no result state docs into a one section? I also need someone else to create the screenshots. No idea how to use AutoScreenshot.

How things look after the PR:

image

image


danharrin (changes requested):
This feels like it's introducing a ton of extra duplicate code for something so similar to what we have already. How about introducing a new evaluation argument to the existing empty state configuration methods, where users can inject `bool $isFiltered` into their functions and adjust the text etc based on that? Currently this is a breaking change as previous empty state customisations users have made no longer apply to the filtered empty state?

You can probably reduce the number of added lines down to 50 or something like this

binaryfire:
> How about introducing a new evaluation argument to the existing empty state configuration methods, where users can inject `bool $isFiltered` into their functions and adjust the text etc based on that? 
> 

@danharrin Not sure what you mean. Can you give me a few lines of code to illustrate? Then I can take it from there.

danharrin:
```php
return $this->evaluate($this->noResultStateDescription, [
    'isFiltered' => $isFiltered,
]) ?? __('filament-tables::table.no_result.description', [
    'model' => $this->getPluralModelLabel(),
]);
```

```php
->noResultStateDescription(fn (bool $isFiltered): string => match ($isFiltered) {
    true => 'No posts',
    false => 'No posts yet',
})
```

binaryfire:
@danharrin Thanks. You used `noResultStateDescription()` in your example. Did you mean `emptyStateDescription()` ?

danharrin:
Yes

binaryfire:
@danharrin If we're not going to add separate methods for a no result state, I think adding an `isFiltered` property / getter is enough. Then people can do this:

```php
->emptyStateDescription(function () {
    if ($this->table->isFiltered() || $this->hasTableSearch()) {
        return 'No matching posts found .';
    }

    return 'Create a post to get started.';
})
```

I've updated the PR accordingly.

binaryfire:
`emptyStateActions()` doesn't accept a closure so this doesn't fully solve the problem. And considering the logic in `pushEmptyStateActions`, I'm not sure if trying to make it accept a closure is the best approach.

danharrin:
`->hidden(fn () => )` on the action?

danharrin:
What's the advantage of the new method over `getActiveFiltersCount()`? Do we really need the property? Feels weird at the moment, the new method doesn't work unless `getActiveFiltersCount()` is called BEFORE to set the property, which feels fragile.

Maybe refactor to get rid of the property and just call `getActiveFiltersCount()` internally and return a bool?

binaryfire:
> What's the advantage of the new method over getActiveFiltersCount()? Do we really need the property? Feels weird at the moment, the new method doesn't work unless getActiveFiltersCount() is called BEFORE to set the property, which feels fragile.

Sorry, that was a leftover from some other experiments. Def not needed.

And yeah, using `->hidden(fn () => )` on the action will do it. Not sure why I didn't think of that...

danharrin:
perfect

PR #12154, opened 2024-04-02
Add Support for S3 Compatible Storage when importing data

WheelieDev:
## Description

Following on  from PR [#12095](https://github.com/filamentphp/filament/pull/12095), I have made it easier to use S3 compatible storage that is set up with a different disk name. 

Although you could simply masquerade your chosen service as the S3 disk, this didn't feel particularly user-friendly or promote developer choice. This my first code-related open source PR so any feedback very welcome 

## Visual changes
None

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/CanImportRecords.php
+++ b/packages/actions/src/Concerns/CanImportRecords.php
@@ -495,9 +495,6 @@ trait CanImportRecords
         return $this->evaluate($this->options);
     }
 
-    /**
-     * @throws Exception if the disk is not an S3 disk
-     */
     private function getDisk(): string
     {
         $disk = $this->evaluate($this->disk);
--- a/packages/actions/src/Concerns/CanImportRecords.php
+++ b/packages/actions/src/Concerns/CanImportRecords.php
@@ -495,7 +495,7 @@ trait CanImportRecords
         return $this->evaluate($this->options);
     }
 
-    private function getDisk(): string
+    public function getDisk(): string
     {
         $disk = $this->evaluate($this->disk);
 
--- a/packages/actions/src/Concerns/CanImportRecords.php
+++ b/packages/actions/src/Concerns/CanImportRecords.php
@@ -503,7 +503,7 @@ trait CanImportRecords
         $disk = $this->evaluate($this->disk);
 
         if (config('filesystems.disks.' . $disk . '.driver') !== 's3') {
-            throw new Exception('The disk must be an S3 compatible disk.');
+            throw new Exception('The disk for the storage of import data can only be customised when using an S3 compatible disk.');
         }
         return $disk;
     }
--- a/packages/actions/src/Concerns/CanImportRecords.php
+++ b/packages/actions/src/Concerns/CanImportRecords.php
@@ -499,7 +499,7 @@ trait CanImportRecords
     {
         $disk = $this->evaluate($this->disk);
 
-        if (config('filesystems.disks.' . $disk . '.driver') !== 's3') {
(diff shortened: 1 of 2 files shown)


packages/actions/src/Concerns/CanImportRecords.php
```diff
@@ -347,17 +350,18 @@ protected function setUp(): void
...
             return fopen($filePath, mode: 'r');
         }
 
         /** @var AwsS3V3Adapter $s3Adapter */
-        $s3Adapter = Storage::disk('s3')->getAdapter();
+        $s3Adapter = Storage::disk($disk)->getAdapter();
 
         invade($s3Adapter)->client->registerStreamWrapper(); /** @phpstan-ignore-line */
-        $fileS3Path = 's3://' . config('filesystems.disks.s3.bucket') . '/' . $filePath;
+        $fileS3Path = 's3://' . config("filesystems.disks.$disk.bucket") . '/' . $filePath;
```
```suggestion
        $fileS3Path = 's3://' . config("filesystems.disks.{$disk}.bucket") . '/' . $filePath;
```

packages/actions/src/Concerns/CanImportRecords.php
```diff
@@ -483,4 +494,18 @@ public function getOptions(): array
     {
         return $this->evaluate($this->options);
     }
+
+    /**
+     * @throws Exception if the disk is not an S3 disk
+     */
```
```suggestion
```

packages/actions/src/Concerns/CanImportRecords.php
```diff
@@ -483,4 +494,18 @@ public function getOptions(): array
...
+    private function getDisk(): string
```
```suggestion
    public function getDisk(): string
```

packages/actions/src/Concerns/CanImportRecords.php
```diff
@@ -483,4 +494,18 @@ public function getOptions(): array
...
+    {
+        $disk = $this->evaluate($this->disk);
+
+        if (config('filesystems.disks.' . $disk . '.driver') !== 's3') {
```
```suggestion
        if (config("filesystems.disks.{$disk}.driver") !== 's3') {
```

packages/actions/src/Concerns/CanImportRecords.php
```diff
@@ -483,4 +494,18 @@ public function getOptions(): array
...
+            throw new Exception('The disk for the storage of import data can only be customised when using an S3 compatible disk.');
```
danharrin, line 503: Strictly this isn't true though, check line 356 which can support local disks.

IMO, this should be a nullable `$disk` property, null by default, and then should fallback to `config('filament.default_filesystem_disk')` when null. Then it can also be used on line 356 and we are still respecting the default disk

danharrin (changes requested):

WheelieDev:
Thank you so much @danharrin  for reviewing and making those improvements. I have been thinking, as we are using livewire temp files under the hood , we might want to change approach. and check the disk being used for such files instead?

danharrin:
So do the files never touch the real disk then? If that's the case, sure we should change it to not read filament.default_filesystem_disk

danharrin:
Hi @TomEasterbrook can you please clarify the above?

WheelieDev:
@danharrin sorry for the delay in coming back to you on this, Easter got in the way and I've only just been able to pick it back up. We are definitely so relying on the temporary uploaded file provided by Livewire. This means that the disk for such files must be relied upon when considering other logic. 

I'm going to have another crack at this now and will add some commits this evening.

danharrin:
Hi, gonna close this due to inactivity, but we'd still love to see this change, so please @ me when you've finished working on it or resubmit the PR and we can handle it. Many thanks!

PR #12155, opened 2024-04-03
Add `extraModalWindowAttributes`

binaryfire:
This PR adds a unique class to action modal windows, allowing individual modals to be targeted via CSS. This is helpful when trying to match the styling of a specific modal to iFramed content (eg. Stripe forms).

Solves: https://x.com/binaryfire/status/1773133588709413339?s=20

Results:

```php
Action::make('addPost')
```

```css
/* theme.css */

.fi-modal-window-add-post {
    @apply !bg-blue-500;
    @apply dark:!bg-blue-500;
}
```

image


danharrin:
Can we please not leak the concept of "action-name" outside of the actions package, please replace this arg with `modal-window-class` or something which can pass through the entire class string to the modal

binaryfire:
@danharrin Done. Also moved the class name logic to CanOpenModal

zepfietje:
What about making this more generic?
Something like `extraModalWindowAttributes`.
I don't like using something specifically for classes as we also don't do this in most other places.

By the way, have you tried using CSS with `:has(iframe)`, @binaryfire?

Not sure if we should add a class to all modals if it's only needed in very select cases, @danharrin.

tanthammar:
If I understand correctly the name of an action should be unique (within each component).

So, if I want three red modals, I have to create three separate css definitions for the same thing?

If this is the case, I would like to suggest the following;

```php

protected string | Closure | null $modalWindowClass = null;


public function modalWindowClass(string | Closure | null $classes = null): static
{
    $this->modalWindowClass = $classes;

    return $this;
}


public function getModalWindowClass(): string
{
    if($this->modalWindowClass) {
        return $this-evaluate($this->modalWindowClass);
    }
    return 'fi-modal-window-' . str($this->getName())->camel()->kebab();
}

```
On the note regarding naming, if we go with something generic like extraFooAttributes then we can't merge the value as per this PR, because the naming implies to only be merged once, into one html element in the component.
In this case it's best to clearly state that we deal with css classes.

Continuing from #12146 

The file `support/resources/views/components/modal/index.blade.php` needs to apply the `modalWindowClass` in two more places. Otherwise the bg-color won't be applied to the modals header and footer.
On the other hand it can be added to the documentation that a custom class must include nested css for those slots...

binaryfire:
Hi @zepfietje. 

The iframe was just one example - there have been several times I've wanted to change the styling of specific modals without iframes too (similar to @tanthammar). This is especially common when you're integrating actions into a public website instead of an admin panel
Resource pages automatically get a `fi-resource-{resource name}` class - this PR is aiming to do something similar for modals. That seems a lot cleaner to me than using CSS selectors 





I think it makes sense to do something similar for modals. Feels a lot cleaner to me than 

To apply the same styling to multiple modals you could do this:

```css
.fi-modal-window-add-post,
.fi-modal-window-add-tag {
    @apply !bg-blue-500;
    @apply dark:!bg-blue-500;
}
```


Also, you can create a single css block that targets multiple classes, so that's still a clean way

binaryfire:
Ignore the last comment... it got posted before I finished writing it.

@zepfietje The iframe was just one example - there have been several times I've wanted to change the styling of specific modals without iframes too, similar to @tanthammar. This is really common when integrating actions into a public website instead of an admin panel. Websites use "popups" with many different styles. A "Sign up to our mailing list" modal will look completely different to a "Get the free eBook" modal. And I need a way of allowing users to style these dynamically using values from the db. That's only possible with a predictable selector.

Resource pages automatically get a `fi-resource-{resource name}` class. This PR aims to do the same thing for modals. That seems a lot cleaner than using `:has` selectors, which would have to be different each time depending on the modal content.

Modals feel like a bit of a second class citizen with respect to styling atm. They're anonymous and don't "belong" to the action that triggered them. There are several ways to customize an action's trigger, but not its modal.

I agree `extraModalAttributes` would be cool but I think that should be in addition to this class? Similar to how pages have both `extraBodyAttrbutes` and a `fi-resource-{resource name}` class. For example I can target specific a list page using a combination of `fi-resource-list-records-page` and `fi-resource-{resource name}` without adding any extra attributes. This would allow the same kind of dynamic CSS targeting for modals.

What do you think?

@tanthammar To apply the same styling to multiple modals you could do this:

```css
.fi-modal-window-add-post,
.fi-modal-window-add-tag {
    @apply !bg-blue-500;
    @apply dark:!bg-blue-500;
}
```

> The file support/resources/views/components/modal/index.blade.php needs to apply the modalWindowClass in two more places. Otherwise the bg-color won't be applied to the modals header and footer.

The header and footer divs don't have bg classes of their own so they will inherit the main window's bg color (see my screenshot above). But if you still need to alter something in the header or footer, you can use a selector like `.fi-modal-window-add-post .fi-modal-header`. The new class is added to the parent window, which makes it easy to target specific child elements too.

danharrin:
I would be open to add `extraModalWindowAttributes()` to the class, seems like a worthwhile addition. We can remove any autogenerating class names and let the user decide which class names they want to add.

tanthammar:
> I would be open to add `extraModalWindowAttributes()` to the class, seems like a worthwhile addition. We can remove any autogenerating class names and let the user decide which class names they want to add.

That is the most fitting suggestion, as it follows other Filament component patterns.
But, I assume that the position where the attributes are inserted cannot be in blade `@class([])` as it is now? (in this PR)

@binaryfire "The header and footer divs don't have bg classes of their own"
Sticky header/footer both have `bg-white dark:bg-gray-900`
Plus, I really, really want to avoid having to edit/recompile custom css every time I create a custom styled modal.


Dare I suggest a ->color() option like Notifications? I mean, both our pr's is about trying to colorize the modal, rather than adding custom attributes?

binaryfire:
I refactored to `extraModalAttributes`, which adds attributes directly to the modal container. I think that's better because you can alter the overlay now too. And it simplifies the class and method names.

Here's an example of the kind of slick, sophisticated modal styles you can create:

```php
Action::make('addPost')
            ->label('New post')
            ->icon('heroicon-s-plus')
            ->modal()
            ->modalWidth(MaxWidth::Large)
            ->extraModalAttributes([
                'class' => 'my-modal',
            ])
            ....
```
```css
.my-modal {
    @apply !bg-red-500;

    .fi-modal-content {
        @apply !bg-blue-500;
    }

    .fi-modal-header, .fi-modal-footer {
        @apply !bg-green-500;
    }

    input {
        @apply !bg-yellow-500;
    }
}
```

image

zepfietje:
> I would be open to add `extraModalWindowAttributes()` to the class, seems like a worthwhile addition. We can remove any autogenerating class names and let the user decide which class names they want to add.

Yeah this is exactly what I would propose. Modals just feel a little different in this regard when comparing to resource page CSS classes.

binaryfire:
Ok I've refactored to `extraModalWindowAttributes`. The attributes are added to the modal window now instead of the modal container.

packages/support/resources/views/components/modal/index.blade.php
```diff
@@ -170,7 +171,7 @@
                         x-transition:leave-start="scale-100 opacity-100"
                         x-transition:leave-end="scale-95 opacity-0"
                     @endif
-                    @class([
+                    {{ $attributes->class([
```
danharrin: Instead of `$attributes`, you should use a new bag that only contains the window attributes. Otherwise, you now have leaking of all attributes from the main bag onto the window as well

Maybe pass through the entire `$extraModalWindowAttributeBag` instead of just `$extraModalWindowAttributes`

danharrin (changes requested):

packages/support/resources/views/components/modal/index.blade.php
binaryfire: Done. I thought `$attributes` was unique-per-element for some reason🤦‍♂️ 

Another alternative could be passing through the array like before and constructing the bag in the `@php` block of `index.blade.php`. Might be a bit cleaner?

danharrin (changes requested):
1) Please use `getExtraModalWindowAttributeBag()` instead of `getExtraModalWindowAttributesBag()` as per other instances of similar functionality
2) This needs to be extended to `extra-modal-window-attributes-bag` -> `extra-modal-window-attribute-bag`
3) Please check `HasExtraInputAttributes` and similar traits. The implementation is quite different, where the attributes are stored as `array<array<mixed> | Closure>` and merged in the getter. There is also a `$merge` option when setting the attributes. You can just copy that and tweak the naming

danharrin:
Thanks

PR #12167, opened 2024-04-03
Fix quirk with beforeSave hook on EditRecord page

tjodalv:
## Description

This PR should fix a quirk with the `beforeSave()` hook on the EditRecord page. The issue is that currently, when saving data, the EditRecord page stores record relationships before the `beforeSave()` hook is called. I have introduced a new argument in the `getState()` method that skips saving relationships when getting the form state. After the `beforeSave()` hook is called, then the relationships are saved.

## Functional changes

### DIFF
--- a/packages/forms/src/Concerns/HasState.php
+++ b/packages/forms/src/Concerns/HasState.php
@@ -2,6 +2,7 @@
 
 namespace Filament\Forms\Concerns;
 
+use Closure;
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Support\Arr;
 
@@ -220,9 +221,11 @@ trait HasState
     /**
      * @return array<string, mixed>
      */
-    public function getState(bool $shouldCallHooksBefore = true): array
+    public function getState(bool $shouldCallHooksBefore = true, ?Closure $afterValidate = null): array
     {
         $state = $this->validate();
+
+        value($afterValidate);
 
         if ($shouldCallHooksBefore) {
             $this->callBeforeStateDehydrated();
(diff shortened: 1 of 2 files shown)


packages/panels/src/Resources/Pages/EditRecord.php
```diff
@@ -140,14 +140,16 @@ public function save(bool $shouldRedirect = true): void
 
             $this->callHook('beforeValidate');
 
-            $data = $this->form->getState();
+            $data = $this->form->getState(skipSavingRelationships: true);
```
danharrin:
```suggestion
            $data = $this->form->getState(shouldSaveRelationships: false);
```

packages/forms/src/Concerns/HasState.php
```diff
@@ -220,14 +220,17 @@ public function statePath(?string $path): static
...
+    public function getState(bool $shouldCallHooksBefore = true, bool $shouldSaveRelationships = false): array
     {
         $state = $this->validate();
 
         if ($shouldCallHooksBefore) {
             $this->callBeforeStateDehydrated();
-            $this->saveRelationships();
-            $this->loadStateFromRelationships(andHydrate: true);
+
+            if (! $shouldSaveRelationships) {
```
danharrin:
```suggestion
            if ($shouldSaveRelationships) {
```

packages/forms/src/Concerns/HasState.php
```diff
@@ -220,14 +220,17 @@ public function statePath(?string $path): static
     /**
      * @return array<string, mixed>
      */
-    public function getState(bool $shouldCallHooksBefore = true): array
+    public function getState(bool $shouldCallHooksBefore = true, bool $shouldSaveRelationships = false): array
```
danharrin:
```suggestion
    public function getState(bool $shouldCallHooksBefore = true, bool $shouldSaveRelationships = true): array
```

tjodalv:
This is even better. Tnx!

PR #12186, opened 2024-04-04
Improve PostgreSQL Ordering Compatibility in PrepareCsvExport Job

kaspernowak:
## Description

This pull request enhances the PrepareCsvExport job to handle PostgreSQL's DISTINCT ON requirements more gracefully. It ensures the primary key is always the first in the ORDER BY clause, maintaining compatibility with PostgreSQL's expectations without disrupting existing functionality for users of other database engines. The necessity for this adjustment became evident when utilizing `->defaultSort()` on tables intended for export, which introduced an additional ordering layer and highlighted the DISTINCT ON discrepancy specifically for PostgreSQL users.

## Visual changes

N/A - This change affects backend functionality only, with no direct visual impact on the user interface.

## Functional changes

## Additional notes

During testing, I observed that when modifying queries using modifyQueryUsing on the ExportAction, the method $query->orderBy('sort') does not clear existing orderings, leading to unexpected sorting results when using PostgreSQL, not when using MySQL. In contrast, $query->reorder()->orderBy('sort') effectively resets and applies the desired ordering. This behaviour underscores the importance of using reorder() in modifyQueryUsing when intending to override previous orderings completely. I suggest further discussion or a separate issue to explore best practices and potentially enhance documentation or functionality around this observation.

### DIFF
--- a/packages/actions/src/Exports/Jobs/PrepareCsvExport.php
+++ b/packages/actions/src/Exports/Jobs/PrepareCsvExport.php
@@ -63,6 +63,19 @@ class PrepareCsvExport implements ShouldQueue
         $keyName = $query->getModel()->getKeyName();
         $qualifiedKeyName = $query->getModel()->getQualifiedKeyName();
 
+        if (config('database.default') === 'pgsql') {
+            $originalOrderings = collect($query->getQuery()->orders)
+                ->reject(function ($order) use ($keyName, $qualifiedKeyName) {
+                    return in_array($order['column'], [$keyName, $qualifiedKeyName]);
+                })
+                ->unique('column');
+
+            $query->reorder()->orderBy($qualifiedKeyName);
+            foreach ($originalOrderings as $ordering) {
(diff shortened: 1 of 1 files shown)


ifrancoe:
We had the same problem, defaultSort was set to "date" and we were not able to export to CSV. We tried the proposed fix and the export worked as expected.

packages/actions/src/Exports/Jobs/PrepareCsvExport.php
```diff
@@ -63,6 +63,19 @@ public function handle(): void
         $keyName = $query->getModel()->getKeyName();
         $qualifiedKeyName = $query->getModel()->getQualifiedKeyName();
 
+        if (config('database.default') === 'pgsql') {
```
```suggestion
        /** @var Connection $databaseConnection */
        $databaseConnection = $query->getConnection();

        if ($databaseConnection->getDriverName() === 'pgsql') {
```

current logic is flawed if the model uses a non-default db connection

PR #12187, opened 2024-04-04
Style notifications based on status

tanthammar:
When manually creating Notifications, the `->color()` option adds `fi-color-foo` to the component.

When Filament generates notifications it never sets the `$color` attribute, only `$status`.
Which makes it very hard to create custom css that targets the notification, when only `fi-color-gray` is added.

This change adds the `$status` value to the blade component. Which means we get a css class that we can target.

Example coloring Filament auto generated form validation error notifications as well as custom notifications with `color('danger')`:

```css
.fi-no-notification.fi-color-danger,
.fi-no-notification.fi-status-danger {
    > div:first-child {
        @apply bg-danger-500 !important;
    }

    .fi-no-notification-title,
    .fi-no-notification-body {
        @apply text-white !important;
    }

    .fi-no-notification-icon,
    .fi-icon-btn-icon {
        @apply text-danger-300 !important;
    }
}
```

### DIFF
--- a/packages/notifications/resources/views/notification.blade.php
+++ b/packages/notifications/resources/views/notification.blade.php
@@ -5,6 +5,7 @@
     use Illuminate\Support\Arr;
 
     $color = $getColor() ?? 'gray';
+    $status = $getStatus();
     $isInline = $isInline();
     $title = $getTitle();
     $hasTitle = filled($title);
@@ -52,6 +53,7 @@
                     default => 'fi-color-custom ring-custom-600/20 dark:ring-custom-400/30',
                 },
                 is_string($color) ? 'fi-color-' . $color : null,
+                is_string($status) ? 'fi-status-' . $status : null,
             ],
         },
     ])


zepfietje:
You should use the existing [color API](https://filamentphp.com/docs/3.x/notifications/sending-notifications#setting-a-background-color) then and target the `fi-color-*` CSS classes.

tanthammar:
@zepfietje Please reopen. 
When Filament sends notifications, It never uses color, just status.
I have no problem when sending notifications myself.

zepfietje:
Status doesn't use color indeed, that's by design.
It only applies a default icon and icon color.

In your example in the description, you're trying to apply a background color style.
This should be done using the dedicated color API as I described above.

Let me know if I'm missing something.

tanthammar:
@zepfietje 
I am trying to style notifications that Filament generates, out of my control.
For instance form validation errors. 
They get the class fi-color-gray. 
By adding the status as a class to the notification enables me to target auto generated notifications.
All statuses get color-gray... 
How can I target them with css classes.

It also gives the ability for different styling depending if one uses `$notification->color('success')` or `$notification->success()`

zepfietje:
> I am trying to style notifications that Filament generates, out of my control. 

Which notifications are you talking about specifically?

Status is basically just a shorthand for a preset icon and icon color currently.
So targeting the color is still the way to go I think.
Let's see if you can customize the notifications you're referring to.

tanthammar:
> Which notifications are you talking about specifically?
> 
All notifications that Filament generates, out of my control.
Like, form field validation error notifications.
If you search for Notification::make in the vendor/filament...
None of them sets the color, only the status.
And they all end up with the color-gray class.
I just want a flag so I can target them in my custom css.

zepfietje:
Please be more specific about the notifications that are out of your control.
Filament doesn't generate form field validation error notifications as far as I know.
Other notifications like action success notifications are customizable already.

tanthammar:
"Filament doesn't generate form field validation error notifications as far as I know." - It does ...

Other Filament generated notifications:

CanExportRecords, rows 130, 241
CanImportRecords, rows 183, 256, 287
CanNotify, rows 21, 58
ExportCompletion, row 55
EmailVerificationPrompt, rows 71, 88
RequestPasswordReset, rows 55, 89, 97
ResetPassword, rows 69, 102, 110
EditProfile, row 212
Login, row 56
Register, row 68
EditTenantProfile, row 167
CreateRecord, row 139
EditRecord, row 226


May I ask why you are reluctant to add the status class to the notification?

zepfietje:
> It does ...

Thanks for pointing me at those!
The action notifications that don't allow for direct customization should probably be refactored to use `successNotification` and `failureNotification` as is used elsewhere.

> May I ask why you are reluctant to add the status class to the notification?

Because the status API isn't a style property like color and icon. Like I said before, it's a shorthand for icon and icon color. If you're trying to use it to add a background color to a notification, that's not the right approach; the dedicated color API should be used for that (as I described in my previous comment).

tanthammar:
My point is that if you allow a status class in the notification blade view.
It is free for everyone to easily target styling for both notification status and/or color.
And the rest of Filament code base can stay as it is.

PR #12188, opened 2024-04-04
Change danger icon

tanthammar:
I hope I am not offending anyone but I think the default notification, danger, icon is a bit strange. 
It looks like a button to close the notification...

This suggested icon follows the design of the warning icon.

Current warning, and suggested danger icon
image

Current danger icon
image

### DIFF
--- a/packages/notifications/src/Concerns/HasIcon.php
+++ b/packages/notifications/src/Concerns/HasIcon.php
@@ -14,7 +14,7 @@ trait HasIcon
     public function getIcon(): ?string
     {
         return $this->baseGetIcon() ?? match ($this->getStatus()) {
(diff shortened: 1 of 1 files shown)


saade:
Sharing my two cents here...

> It looks like a button to close the notification...

I do agree with that, but i think danger relates to errors hence the `x` icon. I don't see any onther icon that fits better here... maybe the filled version of the icons? TailwindUI uses that too.
https://tailwindui.com/components/application-ui/feedback/alerts


For the other icons, to me, they resemble:
`exclamation-circle` - Information
`exclamation-triangle` - Warning

zepfietje:
I don't think this is a good idea, as currently warning and danger are much more distinguishable.

Since the icon is colored red too, and positioned inline in front of the title, I don't see how this would be confusing.

Feel free to override the default icon for your own projects with whatever works for you though. 🙂

PR #12191, opened 2024-04-04
Fix $persist breaking Alpine when Tabs are used in modals

jackwh:
## Description

We can persist the currently active `Tab` to localStorage:

```php
Tabs::make('Example Tabs')
                ->persistTab(true)
                ->id('example-tabs')
                ->tabs([...])
```

I can see the value being updated and written in devtools when the tab is changed. However, when tabs are used in a modal which is being dismissed, the value of `tab` is incorrectly set to `undefined` during a Livewire commit hook.

When opening the modal again, the UI breaks and nothing is rendered. This error appears in the console:

Screenshot 2024-04-04 at 17 08 10@2x

The fix seems to be to fall back to the current tab, instead of trying to persist a potentially undefined array offset.

## Visual changes

N/A

## Functional changes

### DIFF
--- a/packages/forms/resources/views/components/tabs.blade.php
+++ b/packages/forms/resources/views/components/tabs.blade.php
@@ -48,7 +48,7 @@
                     const tabs = getTabs()
 
                     if (! tabs.includes(tab)) {
-                        tab = tabs[@js($getActiveTab()) - 1]
+                        tab = tabs[@js($getActiveTab()) - 1] || tab
                     }
                 })
             })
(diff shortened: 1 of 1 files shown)


jackwh:
Related Alpine PR: https://github.com/alpinejs/alpine/pull/4128

thethunderturner:
Hello, I have a similar issue
image
Not sure if they are related but I just have a modal action in my header, and when I press confirm, the action is "loading" and I get this error in the console

packages/forms/resources/views/components/tabs.blade.php
```diff
@@ -48,7 +48,7 @@
                     const tabs = getTabs()
 
                     if (! tabs.includes(tab)) {
-                        tab = tabs[@js($getActiveTab()) - 1]
+                        tab = tabs[@js($getActiveTab()) - 1] || tab
```
danharrin:
```suggestion
                        tab = tabs[@js($getActiveTab()) - 1] ?? tab
```

PR #12198, opened 2024-04-05
Add Reset Password notification locale

donatiss:
## Description

Send reset password notification by user selected locale (app->getLocale()) 

## Visual changes

none

## Functional changes

### DIFF
--- a/packages/panels/src/Http/Responses/Auth/PasswordResetResponse.php
+++ b/packages/panels/src/Http/Responses/Auth/PasswordResetResponse.php
@@ -12,7 +12,8 @@ class PasswordResetResponse implements Responsable
     public function toResponse($request): RedirectResponse | Redirector
     {
         return redirect()->to(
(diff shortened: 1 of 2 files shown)


danharrin:
Correct me if I'm wrong, but Filament isn't using this URL parameter at all even though you're setting it. So it feels weird to introduce it. But you can absolutely override any of these classes in your own app if you want to make this change. Feel free to push back if I am missing something and we can think about reopening

donatiss:
app()->getLocale() is Laravel function. So when i'm using multi languages, password reset always send email in default language, not user selected language.

danharrin:
I know, but we dont run setLocale() anywhere do we?

donatiss:
Laravel doesn't by default?

`The default language for your application is stored in the config/app.php configuration file's locale configuration option, which is typically set using the APP_LOCALE environment variable. You are free to modify this value to suit the needs of your application.`

danharrin:
It doesnt read the query string though?

donatiss:
I'll check it next week

PR #12199, opened 2024-04-05
Add optional Custom Url Target

JustMetalCake:
## Description

**Use case:** Being able to set a custom target has helped us address a request raised by a client, who wanted to know if it was possible to return to the same tab every time they selected a particular action; however still have both the admin panel & tab in question open.

We noticed that WordPress uses a `target="wp-post-1234"` format to achieve this. This gives flexibility for other use cases too.

I have added a section in the Actions docs to help document this addition.

Huge thanks to @chrispage1 for the support & help on this!

## Visual changes

No visual changes.

## Functional changes

### DIFF
--- a/packages/actions/.stubs.php
+++ b/packages/actions/.stubs.php
@@ -43,6 +43,8 @@ namespace Livewire\Features\SupportTesting {
 
         public function assertActionShouldOpenUrlInNewTab(string | array $name): static {}
 
+        public function assertActionHasUrlTarget(string | array $name, string $target): static {}
+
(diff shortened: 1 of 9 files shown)


danharrin:
Thanks, but it feels weird having this next to the boolean `_blank` option in v3, feels cluttered. Please reopen this to the v4 branch, and refactor to fit better with the existing `_blank` option. The existing setter can be an alias for `urlTarget('_blank')`. The `generate_href_html()` helper can have the argument changed.

PR #12236, opened 2024-04-07
add render hook in bulk action indicator

fabianosfbr:
## Description

Add render hook in bulk action indicator

## Functional changes

### DIFF
--- a/packages/tables/resources/views/components/selection/indicator.blade.php
+++ b/packages/tables/resources/views/components/selection/indicator.blade.php
@@ -36,28 +36,29 @@
     </div>
 
     <div class="flex gap-x-3">
+        {{ \Filament\Support\Facades\FilamentView::renderHook(\Filament\Tables\View\TablesRenderHook::SELECTION_INDICATOR_ACTIONS_BEFORE, scopes: static::class) }}
 
-        {{ \Filament\Support\Facades\FilamentView::renderHook(\Filament\Tables\View\TablesRenderHook::BULKACTION_INDICATE_AFTER, scopes: static::class) }}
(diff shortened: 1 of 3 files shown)


packages/tables/resources/views/components/selection/indicator.blade.php
```diff
@@ -36,23 +36,29 @@ class="text-sm font-medium leading-6 text-gray-700 dark:text-gray-200"
...
+        {{ \Filament\Support\Facades\FilamentView::renderHook(\Filament\Tables\View\TablesRenderHook::SELECTION_INDICATOR_ACTIONS_BEFORE, scopes: static::class) }}
 
-        <x-filament::link
-            color="danger"
-            tag="button"
-            :x-on:click="$deselectAllRecordsAction"
-        >
-            {{ __('filament-tables::table.selection_indicator.actions.deselect_all.label') }}
-        </x-filament::link>
+        <div class="flex gap-x-3">
```
zepfietje, line 41: Do we really need another wrapping div here, @danharrin?

danharrin: The render hook name wouldnt make sense if it wasnt directly wrapping an element IMO, could be confusing, BEFORE and AFTER hooks always directly surround the same element

zepfietje: Is it clear from the PR description if `before` and `after` should be used or `start` and `end`?

danharrin: START and END must be the first and last elements within their parent, which they are not since the `{{ $end }}` slot exists

PR #12261, opened 2024-04-09
Fix createOptionUsing with multiple() and no initial options

archilex:
## Description

When using createOptionUsing() with multiple() BUT without previously selecting options, `$component->getState()` is `null` which, when unpacking the array throws a `Only arrays and Traversables can be unpacked` error. 

This PR fixes this by assigning an empty array when the state is null. 

## Functional changes

### DIFF
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -290,7 +290,7 @@ class Select extends Field implements Contracts\CanDisableOptions, Contracts\Has
 
                 $state = $component->isMultiple()
                     ? [
-                        ...$component->getState(),
+                        ...$component->getState() ?? [],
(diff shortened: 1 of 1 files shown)


danharrin:
Kinda feels like something is wrong in your code tbh, the state for a multiselect should never be null if it has been hydrated properly - maybe missing `form->fill()` to fill the default `[]`?

Nevertheless, this feels like a low impact change we can make

archilex:
@danharrin You are right in that this was an issue with my code. There is another select prior to this one that resets this select when it gets updated. I was setting it to `null` on `afterStateUpdated()` instead of an empty array which caused this error. Apologies for the oversight. Feel free to revert this (or keep it to protect others who make a similar mistake).

PR #12265, opened 2024-04-09
Replace 'overflow-y-hidden' to 'hidden' for tab content

### LINKED ISSUE #12151: Layout shift when tabs with a repeater are used on form
### Package

filament/filament

### Package Version

3.2.62

### Laravel Version

11

### Livewire Version

3

### PHP Version

8.2

### Problem description

Hi!

I need to use tabs with localization. Each tab has a repeater component, which in an example, creates a size table for the product. The problem is that if you use the repeater in tabs, then extra space appears at the page's bottom. 

I tested this on Google Chrome, Safari and Firefox. This issue reproduces in **Chrome** and **Safari**. In Firefox page is normal.

To minimize the effects of extensions, I opened the guest window. If opened the **first tab**:
image

(dev tools - html height) 
image

As you can see, the page has space and a broken header (most likely related to page height). 

But when I switch to the **second tab** - no more space.
image

After minimal debugging, I found that the reason is in the **repeater in tabs**
image

because, when I put many input fields at the same height, the page is normal
image

### Expected behavior

When switching between tabs, there is no free space at the page's bottom on first tab.

### Steps to reproduce

I have prepared a demo project that reproduces the problem.

To create a test product with a size table, run the seeder `php artisan db:seed`

After that, open Google Chrome, login, open a product resource and edit the product.

### Reproduction repository

https://github.com/Oleksandr-Moik/filament-demo-tabs-repeater

### Relevant log output

_No response_

Oleksandr-Moik:
I will replace the 'overflow-y-hidden' with the 'hidden' class for tabs. I don't know why 'overflow' doesn't work, but with 'display:none' it works and fixes my problem.

## Description

(Closes #12151)

## Visual changes

I updated the code in the demo repository. 

Before/after with tabs on the form (see on scrollbar height and header)
form-before
form-after

## Functional changes

- [ ] Code style has been fixed by running the composer cs command.

### DIFF
--- a/packages/forms/resources/views/components/tabs/tab.blade.php
+++ b/packages/forms/resources/views/components/tabs/tab.blade.php
@@ -8,7 +8,7 @@
         'mt-6' => ! $isContained,
     ]);
 
-    $inactiveTabClasses = 'invisible h-0 overflow-y-hidden p-0';
+    $inactiveTabClasses = 'invisible h-0 hidden p-0';
 @endphp
 
 <div
--- a/packages/forms/resources/views/components/wizard/step.blade.php
+++ b/packages/forms/resources/views/components/wizard/step.blade.php
@@ -8,7 +8,7 @@
         'mt-6' => ! $isContained,
     ]);
 
-    $inactiveStepClasses = 'invisible h-0 overflow-y-hidden p-0';
+    $inactiveStepClasses = 'invisible h-0 hidden p-0';
 @endphp
 
 <div
--- a/packages/infolists/resources/views/components/tabs/tab.blade.php
+++ b/packages/infolists/resources/views/components/tabs/tab.blade.php
@@ -8,7 +8,7 @@
         'mt-6' => ! $isContained,
     ]);
 
-    $inactiveTabClasses = 'invisible h-0 overflow-y-hidden p-0';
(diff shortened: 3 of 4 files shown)


Oleksandr-Moik:
This is very interesting, I played around with the layout, and I found one interesting thing - when I removed the layout of the `hiddenLabel` function call from the input fields for the repeater, it no longer breaks

For example, form tabs without `hiddenLabel`
image

and with `hiddenLabel`
image

with `hiddenLabel`, but with a fix from this pull request
image


----

Same behaviour with **Wizard** steps and **Sections** with the repeater.

Sections: 
without hidden labels
image

with hidden labels
image

and now with a fix
image


Wizard:
without hidden labels
image

with hidden labels
image

with a fix
image


---

All these examples have already been added to the demo repository - https://github.com/Oleksandr-Moik/filament-demo-tabs-repeater.

zepfietje:
Thanks for attempting to fix this issue, @Oleksandr-Moik.
Unfortunately, this has been tried a couple times before and isn't the right fix.
For reasons I can't directly remember, the content actually needs to be visible (not `display: hidden`).

Could you try a new PR to experiment with fixes to hidden labels?

Oleksandr-Moik:
Ok) I'll try this weekend. Maybe it's right to fix hidden labels, not content, you're right)

PR #12272, opened 2024-04-10
Only render table empty state actions when any are visible

zepfietje:
## Description

Currently table empty state actions output an empty div when no actions are actually visible. This PR makes sure to first check if any actions are visible.

## Functional changes

### DIFF
--- a/packages/tables/resources/views/components/empty-state/index.blade.php
+++ b/packages/tables/resources/views/components/empty-state/index.blade.php
@@ -34,7 +34,7 @@
             </x-filament-tables::empty-state.description>
         @endif
 
-        @if (array_filter($actions, fn ($action): bool => $action->isVisible()))
+        @if ($actions)
             <x-filament-tables::actions
                 :actions="$actions"
                 :alignment="Alignment::Center"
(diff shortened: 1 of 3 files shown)


packages/tables/resources/views/components/empty-state/index.blade.php
```diff
@@ -34,7 +34,7 @@ class="fi-ta-empty-state-icon h-6 w-6 text-gray-500 dark:text-gray-400"
             </x-filament-tables::empty-state.description>
         @endif
 
-        @if ($actions)
+        @if (array_filter($actions, fn ($action): bool => $action->isVisible()))
```
danharrin: Instead, how about removing this check altogether, and wrapping the `<div>` on line 40 of actions.blade.php in `@if ($actions)`?

Since in that view, the actions are already being filtered with some more nuanced logic regarding action groups etc

danharrin (changes requested):

packages/tables/resources/views/components/empty-state/index.blade.php
danharrin: changed that

zepfietje: Good call, thanks.

PR #12275, opened 2024-04-10
Add unsaved changes for SPA mode

binaryfire:
This uses Livewire's [new cancellable navigate feature](https://github.com/livewire/livewire/pull/8093). 

Edit: This is working properly with `wire:navigate` links now. Checking if the component exists using `if (typeof @this !== 'undefined')` solved the component not found errors.

I'm still seeing component not found errors with `beforeUnload` though (but it is working...). Couldn't figure that out - someone else will need to take a look.

### DIFF
--- a/packages/panels/docs/09-configuration.md
+++ b/packages/panels/docs/09-configuration.md
@@ -200,8 +200,6 @@ public function panel(Panel $panel): Panel
 }
 ```
 
-> Please note: this feature is not compatible with [SPA mode](#spa-mode).
-
 ## Enabling database transactions
 
 By default, Filament does not wrap operations in database transactions, and allows the user to enable this themselves when they have tested to ensure that their operations are safe to be wrapped in a transaction. However, you can enable database transactions at once for all operations by using the `databaseTransactions()` method:
--- a/packages/panels/resources/lang/en/unsaved-changes.php
+++ b/packages/panels/resources/lang/en/unsaved-changes.php
@@ -1,0 +1,7 @@
+<?php
+
+return [
+
+    'wire_navigate_alert' => 'You have unsaved changes. Are you sure you want to leave this page?',
+
+];
--- a/packages/panels/resources/views/components/page/unsaved-data-changes-alert.blade.php
+++ b/packages/panels/resources/views/components/page/unsaved-data-changes-alert.blade.php
@@ -19,7 +19,7 @@
 
                 document.addEventListener('livewire:navigate', (event) => {
                     if (typeof @this !== 'undefined') {
-                        if (! window.shouldPreventNavigation()) {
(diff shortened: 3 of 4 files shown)


binaryfire:
@danharrin I've taken this as far as I can. Hopefully someone else can solve the remaining issues.

binaryfire:
Edit: Checking if the component exists using if (typeof @this !== 'undefined') solved this for `wire:navigate` links.

Actually there shouldn't be any need use `{ once: true }` with `livewire:navigate`. If I remove the hash check, cancelling `wire:navigate` works as expected and the event only fires once.

I think the problem is where `livewire:navigate` sits in the navigation lifecycle. It's happening [before navigation has completed](https://livewire.laravel.com/docs/navigate#javascript-hooks), which means $wire isn't available yet.

So everything works fine the first time the unsaved changes alert is shown. But every subsequent navigation causes an `Uncaught Could not find Livewire component in DOM tree` error. And that's also when `livewire:navigate` starts behaving weirdly and firing multiple times per click.

binaryfire:
Adding `if (typeof @this !== 'undefined')` solved the issues with the `wire:navigate` alert. 

Still getting component not found errors with `beforeUnload` though.

binaryfire:
Hey @zepfietje. I’ve taken this as far as I can. 

The alert for internal links (using `wire:navigate`) works without issues.

But there are component not found console errors being thrown when navigating to external URLs (i.e. on the `beforeUnload` event).

Someone else will need to take a look at that. I was unable to solve it.

danharrin:
Hi, apologies but none of the Filament team know enough about the JS side of Livewire to contribute any more than you do. On top of that, we do not use SPA mode so investing a lot of time into this would not really be a good use of our personal time.

We would love it if you continued to seek a solution, and we'd be more than happy to merge this PR. If not, I totally understand that it's a time-sink, but we aren't planning to invest a large amount of time into completing others' PRs. Sorry about that.

binaryfire:
No worries, makes sense. The intention was to finish it of course, but I couldn’t solve those last few console errors. It’s 100% functional but I’m assuming you don’t want to merge something which causes console errors.

I’ll try and take another look at it down the line.

danharrin:
Yeah, it's just a shame, we would definitely get complaints about the console errors as many people have JS error reporting set up in production

binaryfire:
@danharrin Hey can you re-open this. I figured it out. I thought the issue was coming from `unsaved-data-changes-alert.blade.php`. I didn't realise there was also a `beforeunload` event in `unsaved-action-changes-alert.blade.php`. Applying the same `@this` check there fixes the problem.

binaryfire:
Never mind, just opened a new one

PR #12283, opened 2024-04-11
Support saving without notification

zepfietje:
## Description

Currently there's no way to manually call `save()` and only send a notification in some cases. My use case is a page that's part of a flow of steps, which has a form with its own save button. However, when users forget to click the form submit button, the form should still be saved when they continue to the next page/step. In that case, the saved notification should not be sent, while it should be sent when the form gets saved directly from the form's submit button.

I don't exactly know our breaking change policy regarding method signature changes, so I've implemented this in the ideal way disregarding breaking changes.

## Functional changes

### DIFF
--- a/packages/panels/src/Resources/Pages/EditRecord.php
+++ b/packages/panels/src/Resources/Pages/EditRecord.php
@@ -131,7 +131,7 @@ class EditRecord extends Page
         return $data;
     }
 
-    public function save(bool $shouldRedirect = true): void
+    public function save(bool $shouldSendSavedNotification = true, bool $shouldRedirect = true): void
     {
         $this->authorizeAccess();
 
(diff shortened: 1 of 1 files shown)


packages/panels/src/Resources/Pages/EditRecord.php
```diff
@@ -131,7 +131,7 @@ protected function mutateFormDataBeforeFill(array $data): array
         return $data;
     }
 
-    public function save(bool $shouldRedirect = true): void
+    public function save(bool $shouldSendSavedNotification = true, bool $shouldRedirect = true): void
```
danharrin: BC, please swap the property order

zepfietje:
```suggestion
    public function save(bool $shouldRedirect = true, bool $shouldSendSavedNotification = true): void
```

zepfietje:
Updated, @danharrin.
Is the order something to change in v4?

danharrin:
if you can also write the Rector script, go for it. but I don't think it's worth my time to work that out

zepfietje:
From a quick search I don't think that'll be easy unfortunately.

danharrin:
yeah

PR #12285, opened 2024-04-11
Use more specific return type for form component methods

zepfietje:
## Description

Currently `Component` is used as return type for methods like `getEmailFormComponent()` on the registration page. If you want to add an additional `rule()` to the component by overriding it and calling `parent::getEmailFormComponent()`, PHPStan throws errors. This is because the method uses a generic `Component` return type. This PR changes the return types of those methods to be more specific to fix this issue. Since return types can be changed anyway when the method is overridden, this doesn't cause any issues.

## Functional changes

### DIFF
--- a/packages/panels/docs/08-users.md
+++ b/packages/panels/docs/08-users.md
@@ -240,9 +240,9 @@ In the `form()` method of the example, we call methods like `getNameFormComponen
 If you'd like to customize a field in an authentication form without needing to define a new `form()` method, you could extend the specific field method and chain your customizations:
 
 ```php
-use Filament\Forms\Components\Component;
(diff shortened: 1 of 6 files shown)


danharrin:
BC if the user currently has the method returning Component and now we change it to TextInput, return types can only be MORE specific in child classes, not less specific. Also, this closes down the possibility of using other types of fields OR layout components in these methods. You could just change the return type in your own code to fix the PHPStan issues

zepfietje:
Sorry, I didn't think of that.
I'll have to add a PHPStan annotation in my code then for it to recognize the base Filament form component that gets returned is actually a `TextInput`.

PR #12292, opened 2024-04-11
Allow disabling broadcasting per-panel

binaryfire:
Allows broadcasting (Echo) to be disabled per-panel.


packages/notifications/docs/04-broadcast-notifications.md
```diff
@@ -54,3 +54,18 @@ public function toBroadcast(User $notifiable): BroadcastMessage
         ->getBroadcastMessage();
 }
 ```
+
+## Disabling broadcasting on specific panels
```
danharrin: Please move this to panels/configuration and explain that Echo will attempt connection if you dont disable it

PR #12308, opened 2024-04-12
Fixes bug when attaching a belongsToMany relation that has an orderByPivot

edwwaarrdd:
## Description

When trying to attach a BelongsToMany relation through the RelationManager where the belongsToMany relation has an **orderByPivot** method it will give an error in the SQL query.

This used to work but seems to give an error since filament v3. 
I think its because of the distinct query that has been added in RelationshipJoiner.php class.
As the orderBy query with a distinct needs that orderBy field to be added to the select.

I added a simple foreach that adds the existing orderBy columns to the select. 
This seems to fix the error and has no extra side effects afaik.

For extra info also see this discussion I found discussing this exact problem:
https://github.com/filamentphp/filament/discussions/8022

### DIFF
--- a/packages/support/src/Services/RelationshipJoiner.php
+++ b/packages/support/src/Services/RelationshipJoiner.php
@@ -97,9 +97,10 @@ class RelationshipJoiner
                     continue;
                 }
 
-                if (Str::startsWith($order['column'], "{$relationshipQuery->getModel()->getTable()}.")) {
+                if (str($order['column'])->startsWith("{$relationshipQuery->getModel()->getTable()}.")) {
                     continue;
                 }
+                
                 $relationshipQuery->addSelect($order['column']);
             }
         }
--- a/packages/support/src/Services/RelationshipJoiner.php
+++ b/packages/support/src/Services/RelationshipJoiner.php
@@ -91,6 +91,12 @@ class RelationshipJoiner
             $relationshipQuery
                 ->distinct()
                 ->select($relationshipQuery->getModel()->getTable() . '.*');
+
+            foreach ($relationshipQuery->getQuery()->orders as $order) {
+                if (array_key_exists('column', $order)) {
+                    $relationshipQuery->addSelect($order['column']);
+                }
+            }
         }
 
         return $relationshipQuery;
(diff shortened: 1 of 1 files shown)


packages/support/src/Services/RelationshipJoiner.php
```diff
@@ -91,6 +91,12 @@ function (array $where) use ($relationship): bool {
             $relationshipQuery
                 ->distinct()
                 ->select($relationshipQuery->getModel()->getTable() . '.*');
+
+            foreach ($relationshipQuery->getQuery()->orders as $order) {
+                if (array_key_exists('column', $order)) {
+                    $relationshipQuery->addSelect($order['column']);
```
danharrin: Are the `$order['column']` values always qualified with the table name?

If so, can we add a new condition to ONLY `addSelect()` IF the column name does NOT start with `"{$relationshipQuery->getModel()->getTable()}."`

Because we already have a select for `->select($relationshipQuery->getModel()->getTable() . '.*')` which should catch some of these?

danharrin (changes requested):

packages/support/src/Services/RelationshipJoiner.php
edwwaarrdd: Just checked the code for the eloquent builder and it seems that for an order query it always calls `$this->model->getQualifiedKeyName()` which adds the table name in front the column name.
I'll update the PR to check if the column matches the existing select.

edwwaarrdd: Added the extra check! 
I don't know if you like the early returns with the continue's. But it seems like that would be the most readable?
Let me know if you want it changed 😃

packages/support/src/Services/RelationshipJoiner.php
```diff
@@ -91,6 +91,17 @@ function (array $where) use ($relationship): bool {
...
                 ->select($relationshipQuery->getModel()->getTable() . '.*');
+
+            foreach ($relationshipQuery->getQuery()->orders as $order) {
+                if (! array_key_exists('column', $order)) {
+                    continue;
+                }
+
+                if (Str::startsWith($order['column'], "{$relationshipQuery->getModel()->getTable()}.")) {
+                    continue;
+                }
```
danharrin:
```suggestion
                if (str($order['column'])->startsWith("{$relationshipQuery->getModel()->getTable()}.")) {
                    continue;
                }
                
```

danharrin:
Thanks!

ryanmortier:
@danharrin this caused some type of regression for me.

> foreach() argument must be of type array|object, null given

danharrin:
Fixed in upcoming version

PR #12310, opened 2024-04-12
Feature/custom xlsx writer

shepard153:
## Description
Since OpenSpout XLSX writer is a little bit limited (eg. doesn't format currency cells) I've tried to use different writer. The problem was that whenever I've tried to launch my custom job with different writer it always failed because there was another XLSX job in the queue. So here is my proposition to implement `customWriter()` method so it will skip the original job and proceed with the rest of the queued tasks.

## Functional changes


ralphjsmit:
I recognize the issue for not (easily) being able to format cells. Wouldn't it be a better solution to include a write in Filament and add native support for formatting of cells and perhaps some simple styling (bold, italic).

shepard153:
The problem is that there is no perfect package. I've actually added styling to Filament exports and been working on conditional styling but dropped it since OpenSpout doesn't suit all my needs. The package I use now is [SimpleXLSXGen](https://github.com/shuchkin/simplexlsxgen) but it does not support chunking. [PHP_XLSXWriter](https://github.com/mk-j/PHP_XLSXWriter) looks promising but I think it's no longer maintained. That's why I came with solution above.

danharrin:
Personally, I don't think this is the right solution. The method doesn't really allow you override anything "custom", just disable the existing job. IMO, you just need to bind a new `CreateXlsxFile` class to the container to replace the existing one, no? Then you can implement any processing you want.

Please let me know if I am misunderstanding

PR #12316, opened 2024-04-13
Add completed icon to step

malzariey:
## Description

The update introduces an extra method in the Step class which allows setting a custom icon for completed steps. This new feature enhances the UI customization abilities, enabling users to further adjust the appearance of their progress steps based on their specific needs.

In the wizard blade view, new conditionals have been introduced. These conditionals check if a custom completed icon has been set by the user. If no custom icon has been set, the default icon is used. This ensures that the user interface remains consistent and intuitive, regardless of the customization level.

### Before

image

### After

image

## Functional changes

- Changes have been tested, and it is confirmed that they don't break existing functionality.
- Up-to-date documentation has been included.
- The code style has been fixed by running the composer cs command, ensuring it adheres to the project's style guide.


packages/forms/resources/views/components/wizard.blade.php
```diff
@@ -160,13 +160,23 @@ class="fi-fo-wizard-header-step-icon-ctn flex h-10 w-10 shrink-0 items-center ju
...
+                            />
+                        @else
+                            <x-filament::icon
+                                alias="forms::components.wizard.completed-step"
+                                icon="heroicon-o-check"
+                                x-cloak="x-cloak"
+                                x-show="getStepIndex(step) > {{ $loop->index }}"
+                                class="fi-fo-wizard-header-step-icon h-6 w-6 text-white"
+                            />
+                        @endif
```
```suggestion
                        <x-filament::icon
                            :alias="filled($completedIcon) ? null : 'forms::components.wizard.completed-step'"
                            :icon="$completedIcon ?? 'heroicon-o-check'"
                            x-cloak="x-cloak"
                            x-show="getStepIndex(step) > {{ $loop->index }}"
                            class="fi-fo-wizard-header-step-icon h-6 w-6 text-white"
                        />
```

danharrin (changes requested):

danharrin:
Thanks, especially for doing the docs screenshots

packages/forms/resources/views/components/wizard.blade.php
malzariey: Hello @danharrin,

Firstly, I really appreciate your update.

Just to make sure I understand correctly, should we not retrieve `$completedIcon` from the $step first?

malzariey: @danharrin  I will submit a new pull request, Wizard got an execption. `Undefined variable $completedIcon`

Want to make sure why was the alias changes to null on custom?

atmediauk:
In a scenario where the wizard is skippable and the buttons are disabled e.g

```
->skippable()
->nextAction(fn (Action $action): Action => $action->extraAttributes(['x-show' => 'false']))
->previousAction(fn (Action $action): Action => $action->extraAttributes(['x-show' => 'false']))
```
The redundant checked icon is still displayed. Hopefully this video helps explain what I mean - https://jmp.sh/2YXDsYQN

I tried passing null to ```->completedIcon(null)``` (https://filamentphp.com/docs/3.x/forms/layout/wizard#customizing-the-icon-for-completed-steps) but this has no impact due to the ```@else``` fallback.

Would it be possible to add the option to omit the completed icon by passing null? E.g ```->completedIcon(null)```

PR #12356, opened 2024-04-16
fix: Duplicate request to count records

webmaxx:
## Description

If you select to display all records, then the request to count the number of records is executed 2 times

### Before

fix-before

### After

fix-after

## Functional changes

### DIFF
--- a/packages/tables/src/Concerns/CanPaginateRecords.php
+++ b/packages/tables/src/Concerns/CanPaginateRecords.php
@@ -32,7 +32,7 @@ trait CanPaginateRecords
 
         /** @var LengthAwarePaginator $records */
         $records = $query->paginate(
-            perPage: $perPage === 'all' ? $total : $perPage,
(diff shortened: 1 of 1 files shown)


packages/tables/src/Concerns/CanPaginateRecords.php
```diff
@@ -28,12 +28,14 @@ public function updatedTableRecordsPerPage(): void
...
     {
         $perPage = $this->getTableRecordsPerPage();
+        $total = $query->toBase()->getCountForPagination();
 
         /** @var LengthAwarePaginator $records */
         $records = $query->paginate(
-            $perPage === 'all' ? $query->toBase()->getCountForPagination() : $perPage,
-            ['*'],
-            $this->getTablePaginationPageName(),
+            perPage: $perPage === 'all' ? $total : $perPage,
```
danharrin:
```suggestion
            perPage: ($perPage === 'all') ? $total : $perPage,
```

PR #12363, opened 2024-04-16
SPA mode unsaved changes

binaryfire:
Adds unsaved changes functionality to SPA mode using Livewire's new [cancellable navigate feature](https://github.com/livewire/livewire/pull/8093).

This PR is fully working with no errors. It was necessary to add `if (typeof @this !== 'undefined')` checks to prevent component not found errors in SPA mode due to `wire:navigate`'s lifecycle.

I've tested this in both SPA mode and non-SPA mode with internal links, external links and navigating away with open modals. Everything I checked works.


tonypartridge:
Q.  shouldn't we bump Livewire requirement to ^3.4.10 to be safe?

binaryfire:
@tonypartridge Good point. Yes definitely - this depends on min 3.4.10.

taliptako:
Right now, it changes the warning message in the SPA mode. But non-spa mode still has the old one. I think we should use the same message in both.

binaryfire:
@taliptako That isn't possible AFAIK. Every browser uses different language for their `beforeUnload` alert (and it can't be changed). Eg. this is what they look like in Chrome and Firefox:
image
image

Unsaved changes alerts in SPAs use custom events, which means they have to use custom alerts instead of the browser's `beforeUnload` alert. It's the same with any JS framework- see Vue Router's navigation guard docs for example: https://router.vuejs.org/guide/advanced/navigation-guards.html. 
image

It's the same with `wire:navigate`.

So there's no way to match the browser's `beforeUnload` alert exactly. If you try and match the language of Chrome's, it won't match Firefox's. And so on... Feel free to make a PR if you think the SPA alert language can be improved, but I think the current alert is simple and clear.

T3mmyKay:
Is there a way to perform action other actions like say the user is navigating away and didn't submit the form I want to capture, that and store it that he cancelled that action

PR #12370, opened 2024-04-17
Add database notifications config info

janiskelemen:
## Description

The documentation is missing this important step, which is needed to enable database notifications.

### DIFF
--- a/packages/notifications/docs/01-installation.md
+++ b/packages/notifications/docs/01-installation.md
@@ -4,6 +4,14 @@ title: Installation
 
 **The Notifications package is pre-installed with the [Panel Builder](/docs/panels).** This guide is for using the Notifications package in a custom TALL Stack application (Tailwind, Alpine, Livewire, Laravel).
 
+## Enable Database Notifications
+
+In order to enable database notifications you need to add the following to your AppPanelProvider configuration:
+
+```php
+->databaseNotifications()
+```
+
 ## Requirements
 
 Filament requires the following to run:


dreftas:
It is documented at [Panel builder -> notifications](https://filamentphp.com/docs/3.x/panels/notifications#overview)

janiskelemen:
@drefats I think it makes sense to mention it in the notifications section itself. Even Povilas mentioned it in one of his videos :)

https://youtu.be/2qrPdEW4zxE?t=172

packages/notifications/docs/01-installation.md
```diff
@@ -4,6 +4,14 @@ title: Installation
 
 **The Notifications package is pre-installed with the [Panel Builder](/docs/panels).** This guide is for using the Notifications package in a custom TALL Stack application (Tailwind, Alpine, Livewire, Laravel).
 
+## Enable Database Notifications
+
+In order to enable database notifications you need to add the following to your AppPanelProvider configuration:
+
+```php
+->databaseNotifications()
```
valpuia604, line 12: I think it's better like this
```php
use Filament\Panel;
 
{
    return $panel
        // ...
        ->databaseNotifications();
}
```

packages/notifications/docs/01-installation.md
```diff
@@ -4,6 +4,14 @@ title: Installation
 
 **The Notifications package is pre-installed with the [Panel Builder](/docs/panels).** This guide is for using the Notifications package in a custom TALL Stack application (Tailwind, Alpine, Livewire, Laravel).
 
+## Enable Database Notifications
+
+In order to enable database notifications you need to add the following to your AppPanelProvider configuration:
```
valpuia604, line 9: Instead of `AppPanelProvider configuration`, use like `panel configuration file` because we have to put this inside Filament panel provider, it can be Admin, App, Customer etc

danharrin:
I don't think the Installation guide is the right place for this. How about putting it on [this page](https://filamentphp.com/docs/3.x/notifications/database-notifications), and please copy the example from [here](https://filamentphp.com/docs/3.x/panels/notifications#overview)? I will merge a PR following that, but will close this one since it involves a completely separate file anyway.

PR #12348, opened 2024-04-16
Fixed bug issue on TextArea with autosize after validation failed

emamalias:
## Description

Reference issue #11428 

Fixes bug issue on TextArea form component with autosize() return the height to initial height after validation failed

## Visual changes

https://github.com/filamentphp/filament/assets/13638285/e22ea86a-2e85-4faf-bb01-f2456f4ae672

## Functional changes

- [x] Code style has been fixed by running the `composer cs` command.
- [x] Changes have been tested to not break existing functionality.
- [x] Documentation is up-to-date.

PR #12374, opened 2024-04-17
fix: removed form-validation-error attribute on TextArea component

emamalias:
## Description
Ref: #12348 
Removed `x-on:form-validation-error.window` as it is not needed for autosize bug issue


zepfietje:
Thanks, @emamalias!
Reference: https://github.com/filamentphp/filament/pull/12348#discussion_r1568402627.

zepfietje:
What about `x-intersect.once="render()"`?

emamalias:
> What about `x-intersect.once="render()"`?

Yep. I think it is safe to remove. I was hesitant to remove as it might help the performance especially for long forms?

danharrin:
I am returning the intersect. #8193

PR #12376, opened 2024-04-17
Allow multiple `routes()` registries on `Panel`

juliomotol:
## Description

This PR aims to allow Developers and Plugin Authors to register `routes()`, `authenticatedRoutes()`, `tenantRoutes()`, `authenticatedTenantRoutes()` without interfering with one another.

The previous implementation only allow a single instance of closure. So if a developer decides to add a route, it may not be registered if they use a plugin that adds a route.

## Visual changes

N/A

## Functional changes


danharrin:
Nice, thanks

PR #12379, opened 2024-04-17
Fix: Select HasManyThrough relationship

saade:
## Description

When using a Select with a `HasManyThrough` relationship, the component fails to load its state from relationship because `getResults()` returns a collection not a single record. 

## Functional changes

### DIFF
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -811,7 +811,7 @@ class Select extends Field implements Contracts\CanDisableOptions, Contracts\Has
             if (
                 ($relationship instanceof BelongsToMany) ||
                 ($relationship instanceof HasManyThrough)
-            ){
+            ) {
                 /** @var Collection $relatedModels */
                 $relatedModels = $relationship->getResults();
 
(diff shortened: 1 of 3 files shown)


packages/forms/src/Components/Select.php
```diff
@@ -818,18 +821,15 @@ public function relationship(string | Closure | null $name = null, string | Clos
                     //
                     // https://github.com/filamentphp/filament/issues/1111
                     $relatedModels
-                        ->pluck($relationship->getRelatedKeyName())
+                        ->pluck($relationship->getRelated()->getKeyName())
```
```suggestion
                        ->pluck(($relationship instanceof BelongsToMany) ? $relationship->getRelatedKeyName() : $relationship->getRelated()->getKeyName())
```

saade: @danharrin Out of curiosity, whats the difference here?

danharrin: check the implementation of getRelatedKeyName in BelongsToMany, it seems to differ, I think BelongsToMany has its own related key definition

saade: Cool, i thought `->getRelatedKeyName()` and `->getRelated()->getKeyName()` results would be the same

PR #12398, opened 2024-04-19
Add usual search / naming to disable placeholder selection

tonypartridge:
Struggled to find this in the docs, so included some keywords to improve find-ability

### DIFF
--- a/packages/tables/docs/04-filters/02-select.md
+++ b/packages/tables/docs/04-filters/02-select.md
@@ -103,7 +103,7 @@ SelectFilter::make('author')
 
 ## Disable placeholder selection
 
-You can prevent the placeholder (null option) from being selected using the `selectablePlaceholder()` method:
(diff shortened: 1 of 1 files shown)


packages/tables/docs/04-filters/02-select.md
```diff
@@ -103,7 +103,7 @@ SelectFilter::make('author')
 
 ## Disable placeholder selection
 
-You can prevent the placeholder (null option) from being selected using the `selectablePlaceholder()` method:
+You can prevent the placeholder (null option) also known as the (Select All) option from being selected using the `selectablePlaceholder()` method:
```
zepfietje, line 106:
```suggestion
You can prevent the placeholder (null option with default label "All") from being selected using the `selectablePlaceholder()` method:
```

danharrin:
```suggestion
You can remove the placeholder (null option), which disables the filter so all options are applied, using the `selectablePlaceholder()` method:
```

PR #12454, opened 2024-04-23
Fix infolist entry label

zepfietje:
## Description

Infolist entries of nested relationship attributes currently get a wrong label which is different from what a table column with the same name gets.

| | Column/entry name | Table column label | Infolist entry label
|---|---|---|---|
| **Before this fix** | generation.project.team.name | Team | Generation |
| **After this fix** | generation.project.team.name | Team | Team |

## Functional changes

### DIFF
--- a/packages/infolists/src/Components/Entry.php
+++ b/packages/infolists/src/Components/Entry.php
@@ -41,7 +41,8 @@ class Entry extends Component implements HasHintActions
     public function getLabel(): string | Htmlable | null
     {
         $label = parent::getLabel() ?? (string) str($this->getName())
-            ->before('.')
+            ->beforeLast('.')
+            ->afterLast('.')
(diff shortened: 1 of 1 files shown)


danharrin:
BC, also inconsistent with schema component labelling. In v4 these will be the same algo. I think table cols and infolist entries deserve different labels since entires go into much more depth and there could be more than one entry for the same relationship. Also JSON entries?

zepfietje:
I don't think this is a breaking change, rather a bug fix.

Which algorithm will be used in v4?

Personally I don't see why infolist entries and table columns are named inconsistently.
I do understand your remark about JSON entries, but still then the algorithm in this PR could work fine for most cases. Even if it didn't work correctly for JSON, usage of relationships is generally much higher than JSON fields I think.

PR #12456, opened 2024-04-23
Fix unintentionally scrolling sidebar

timrspratt:
When using certain field components that have `<template>` tags in for modals and alpine collapse functionality, the sidebar can end up scrolling incorrectly over the sticky header.

The header is `sticky` with a `z-20`, so this PR adds the same to the sidebar.

Top header wrapper:
Screenshot 2024-04-23 at 12 25 58

Sidebar wrapper before:
image

Sidebar wrapper after:
image

Video showing issue:

https://github.com/filamentphp/filament/assets/20431294/5da54f86-a5fc-4143-a755-f7dfee818192

I have checked scenarios for mobile menu, top navigation, collapse navigation and all appear to be correct.

### DIFF
--- a/packages/panels/resources/views/components/sidebar/index.blade.php
+++ b/packages/panels/resources/views/components/sidebar/index.blade.php
@@ -41,7 +41,7 @@
         ])
     }}
 >
-    <div class="overflow-x-clip">
+    <div class="sticky top-0 z-20 overflow-x-clip">
         <header
             class="fi-sidebar-header flex h-16 items-center bg-white px-6 ring-1 ring-gray-950/5 dark:bg-gray-900 dark:ring-white/10 lg:shadow-sm"
         >


zepfietje (changes requested):
Thanks for the PR, @timrspratt!

I think position `fixed` should be used instead, since the sticky position in this case would never act as relative since it's always at the top 0 already, right?

Also, could you please check if the collapsible sidebar animation on desktop still works as expected? Please post a screen recording of that here.

timrspratt:
I'll have a play with `fixed` properly when I can. A quick try just now and it looks like `w-full` would also be needed.

zepfietje:
Are you able to pick this up again, @timrspratt?

zepfietje:
I'm going to close this PR, since there hasn't been any activity and requires some more testing before it can be merged.

Feel free to open a new PR when you've had a chance to look at my change request, @timrspratt. 🙂

PR #12457, opened 2024-04-23
Unable to set collapsed() in Builder field due to hardcoded setting

DanielFeldbrugge:
## Description
The component to add new Builder blocks to the field contains hardcoded collapsed(false) which prevents from setting this setting at Resource level.

## Visual changes
No visual changes

## Functional changes

### DIFF
--- a/packages/forms/src/Components/Builder.php
+++ b/packages/forms/src/Components/Builder.php
@@ -144,8 +144,6 @@ class Builder extends Field implements Contracts\CanConcealComponents, Contracts
 
                 $component->getChildComponentContainer($newUuid)->fill();
 
-                $component->collapsed(false, shouldMakeComponentCollapsible: false);
-
                 $component->callAfterStateUpdated();
             })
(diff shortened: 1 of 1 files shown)


danharrin:
The resource setting should apply to existing items. This line of code is present to expand new items in the builder so they can be filled out quickly, its a UX thing. Please open an issue w/ a reproduction repository if you cannot get existing (saved) items to collapse

DanielFeldbrugge:
When you meantion that existing items are collapsed this works the way it shoot. So I agree if you add new items to the builder and you want to directly fill in the form. However there will also be usecases (like in my case) that you first want to add the blocks, rearrange them and then add information to te items. For example a site builder or a form builder (first setup the fields like email, phone), then openup the item. 

If I create a PR to make this an optional setting like (addActionCollapsed and betweenActionCollapsed) would this be on you consideration to at this to the core? If not set, the default is still present.

danharrin:
`expandNewItems(false)` probably

PR #12515, opened 2024-04-26
Add option to remove topbar from the profile page

saade:
### Description

Every auth page has an option to remove the topbar except for the profile page. This PR adds the same configuration option as the other pages.


packages/panels/src/Pages/Auth/EditProfile.php
danharrin, line 1: Please extract a trait to use here and SimplePage, probably HasTopbar

danharrin (changes requested):
As above

saade:
@danharrin Done. Took a step further and extracted `$hasMaxWidth` too. Feel free to revert it.

danharrin:
Thanks!

PR #12533, opened 2024-04-28
Closure support to forms method on CanBeValidated trait

uncrackable404:
## Description

This PR implements support for closure evaluation in the ->forms($closure) method on classes that use the CanBeValidated trait. While closure evaluation already exists for individual rules, there is currently no implementation that returns an array of rules. For our use case, this implementation allows us to dynamically modify the rules for a specific field. Centralizing the rules on an Enum.

## Visual changes
Before
Screenshot 2024-04-28 alle 18 30 38

After
Screenshot 2024-04-28 alle 18 34 16

Enum
Screenshot 2024-04-28 alle 18 31 26

## Functional changes
- [x] Changed Trait `vendor/filament/forms/src/Components/Concerns/CanBeValidated.php`
- [x] Test passed


danharrin:
Thanks

PR #12540, opened 2024-04-29
euc-kr is a superset of iso-8859-1

corean:
## Description

Fix encoding issues when importing CSV files in Excel 2007

When importing CSV files into Excel 2007, there were encoding issues. While converting from ISO-8859-1 to UTF-8 worked, encoding errors persisted.

After trying various approaches, it was discovered that EUC-KR is a subset of ISO-8859-1. By adding code to handle EUC-KR encoding, the issue was resolved.

This commit addresses the encoding problems by incorporating the necessary changes to handle EUC-KR encoding when importing CSV files into Excel 2007.

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/CanImportRecords.php
+++ b/packages/actions/src/Concerns/CanImportRecords.php
@@ -410,6 +410,11 @@ trait CanImportRecords
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks

PR #12381, opened 2024-04-18
add charset converter & detect encoding on csv upload

valentin-morice:
## Description

Fixes the issue #12063. Implements a dropdown that allows the user to select an encoding for their csv, which is then converted to UTF-8. [mb_detect_encoding()](https://www.php.net/manual/en/function.mb-detect-encoding.php) could be used to detect formatting, but is apparently (somewhat?) deprecated, as pointed out [here](https://www.php.net/manual/en/function.mb-detect-encoding.php#127650).
If implemented, this solution might require to modify documentation.
My first choice was to disable the upload field when no encoding is selected, but dynamic disabling doesn't seem to work on this filed. I've used `hidden()` instead.

## Visual changes

image
*Displays an additional dropdown*

## Functional changes

- [X] Code style has been fixed by running the `composer cs` command.
- [X] Changes have been tested to not break existing functionality.
- [x] Documentation is up-to-date.

PR #12571, opened 2024-05-01
bugfix: stream_get_contents retrieves content in bytes, which may result in unexpected content

bzy107:
## Description
The change in #12381 caused a phenomenon where the mapping string of headers during file upload gets garbled. 
スクリーンショット 2024-05-01 21 30 28

It was found that when there is a multibyte character at the 1000th byte, the garbled content obtained by stream_get_contents is retrieved.
it's advisable to avoid using `stream_get_contents($resource, 1000)`.
スクリーンショット 2024-05-01 21 29 22

Proposal:
Instead of retrieving up to the 1000th byte, we will modify it to retrieve the first line (header) of the CSV and determine the character code. In this case, there is no need to worry about multibyte characters.

The list of header mappings after the fix
スクリーンショット 2024-05-01 21 31 29

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/CanImportRecords.php
+++ b/packages/actions/src/Concerns/CanImportRecords.php
@@ -398,8 +398,8 @@ trait CanImportRecords
     {
         $fileHeader = fgets($resource);
 
-        // Encoding of a subset should be declared before the encoding of its superset
-        $encodingLists = [
+        // The encoding of a subset should be declared before the encoding of its superset.
+        $encodings = [
             'UTF-8',
             'SJIS-win',
             'EUC-KR',
@@ -410,7 +410,7 @@ trait CanImportRecords
             'EUC-JP',
         ];
 
-        foreach ($encodingLists as $encoding) {
+        foreach ($encodings as $encoding) {
             if (! mb_check_encoding($fileHeader, $encoding)) {
                 continue;
             }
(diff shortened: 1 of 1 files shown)


danharrin:
The idea is that we don't want to read the entire file into memory, only the first 1000 bytes. I think this change could result in performance issues in large datasets. What can we do about that?

bzy107:
>  I think this change could result in performance issues in large datasets. 

Thank you for your input.

With this change, I don't think there's a need to worry about performance concerns since we're only checking the first line (header). If the header string has enough data to affect performance, then there might be issues with the table design in the first place.

My concern lies with `stream_get_contents` potentially causing unintended character encoding issues with multibyte strings.
In such cases, we cannot determine the correct character code.

If there is no garbling:
```bash
php > $f = fopen('./a.txt', mode: 'r');
php > var_dump(stream_get_contents($f, 6));
string(6) "ああ"
```

If there is garbling:
```bash
php > $f = fopen('./a.txt', mode: 'r');
php > var_dump(stream_get_contents($f, 5));
string(5) "あ?"
```

If `stream_get_contents` could be specified by the number of characters instead of bytes, I don't think this issue would have occurred, but I didn't think of it at the time.

danharrin:
Thanks for the clarification

PR #12587, opened 2024-05-02
Simplify abortion using abort_unless

Omranic:


danharrin:
We want to abort IF the URL is blank, not UNLESS. Not trying to be rude here but please test your work, or it just creates more load for us.

Also, I'm not entirely sure the benefit. I don't like the fact that the inline variable is created inside the function call and then used underneath. I think I prefer the old way, sorry. If the variable init wasn't inside the func call, then I'd probably approve.

PR #12593, opened 2024-05-02
Make SpatieMediaLibraryFileUpload component more robust if media is null

sten:
There were some lines left that deal with $media possibly being null. 

## Description

When the $media variable is null, some lines would cause null pointer exceptions and thus crashes.

## Functional changes


danharrin:
Thanks

PR #12605, opened 2024-05-03
Fix for bug #12602

### LINKED ISSUE #12602: Checkboxes in a CheckboxList, when included in a Wizard step, toggles all checkboxes in the group regardless of the one toggled
### Package

filament/forms

### Package Version

v3.2.72

### Laravel Version

v3.2.72

### Livewire Version

v3.4.11

### PHP Version

PHP 8.2.18

### Problem description

When including a JS powered CheckboxList form field in a Wizard step, where the step can dynamically contain a schema, the CheckboxList toggles all on every checkbox click.

### Expected behavior

When toggling a checkbox in a CheckboxList only that checkbox should be toggled

### Steps to reproduce

Setup the project
```
composer install
npm install
npm run build
```

Start the webserver
```
php artisan serve
```

Go to the Page resource page http://127.0.0.1:8000/admin/pages/create

Click Add to content blocks

Select the radio button `Select + Checkboxes + Text` and click `Next`
Toggle any of the checkboxes and you will see all the checkboxes toggle

Mov to

### Reproduction repository

https://github.com/matthewnessworthy/filamentphp-wizard-bug

### Relevant log output

_No response_

dmitry-udod:
## Description

Fix for bug #12602

https://github.com/filamentphp/filament/assets/4639175/da13cfb2-2a4f-4092-95a6-7b2870dea798

## Visual changes

Nope

## Functional changes

### DIFF
--- a/packages/forms/src/Components/Concerns/HasState.php
+++ b/packages/forms/src/Components/Concerns/HasState.php
@@ -269,6 +269,17 @@ trait HasState
         Arr::set($hydratedDefaultState, $statePath, $defaultState);
     }
 
+    public function fillStateWithDefaults(): void
+    {
+        if (! Arr::has((array) $this->getLivewire(), $this->getStatePath())) {
+            $this->state($this->defaultState);
+        }
+
+        foreach ($this->getChildComponentContainers(withHidden: true) as $container) {
+            $container->fillStateWithDefaults();
+        }
+    }
+
     public function fillStateWithNull(): void
     {
         if (! Arr::has((array) $this->getLivewire(), $this->getStatePath())) {
--- a/packages/forms/src/Components/Wizard.php
+++ b/packages/forms/src/Components/Wizard.php
@@ -87,7 +87,7 @@ class Wizard extends Component
                             $currentStep->callBeforeValidation();
                             $currentStep->getChildComponentContainer()->validate();
                             $currentStep->callAfterValidation();
-                            $nextStep?->fillStateWithNull();
+                            $nextStep?->fillStateWithDefaults();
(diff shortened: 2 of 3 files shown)


danharrin:
I don't think this is a problem specific to wizards. It happens whenever a dynamic schema is involved, and my advice is usually to use `afterStateUpdated()` on the component that determines the schema to `$set()` the field values for the new fields. I do feel a bit weird filling the defaults after the form is first loaded, it could be unexpected behaviour.

Gonna need to leave this open while I consider it

dmitry-udod:
@danharrin Hi! Sure. Just FYI, the problem here is that without applying default values for next step fields, we don't have this data at all. Please check the video.


https://github.com/filamentphp/filament/assets/4639175/be897016-69cf-45ab-918b-a429c3349c1e

danharrin:
Yeah, so it's the same problem as something like this:

```php
TextInput::make('test'),
Group::make(function ($get) {
    if ($get('test') === 'foo') {
        return [CheckboxList::make('test2')->options(['one' => 'One', 'two' => 'Two'])];
    }

    return [];
}),
```

Which is solved by:

```php
TextInput::make('test')
    ->afterStateUpdated(fn ($set, $state) => ($state === 'foo') ?
        $set('test2', []) :
        null
    ),
Group::make(function ($get) {
    if ($get('test') === 'foo') {
        return [CheckboxList::make('test2')->options(['one' => 'One', 'two' => 'Two'])];
    }

    return [];
}),
```

danharrin:
This solution is actually documented, so closing this

PR #12627, opened 2024-05-04
feat: loading section customizable height

hpiaia:
## Description

Sometimes the fixed `h-32` height on the loading section can cause a lot of layout shift when the components are loaded.

This PR adds `$placeholderHeight` to components that can lazy loaded.

I've wrapped the common lazy properties and methods into a new `CanBeLazy` concern and created the `$placeholderHeight` property, so users can override it in each component.

I am just not sure where to put the documentation for this since its a shared concept between multiple components.

Also, I would love a word from you to see what do you think about this, @danharrin.

## Functional changes

### DIFF
--- a/packages/support/resources/views/components/loading-section.blade.php
+++ b/packages/support/resources/views/components/loading-section.blade.php
@@ -29,5 +29,5 @@
     :twoXlStart="$columnStart['2xl'] ?? null"
     class="fi-loading-section"
 >
-    <x-filament::section class="animate-pulse" style="height: {{ $placeholderHeight }};" />
+    <x-filament::section class="animate-pulse" style="height: {{ $placeholderHeight }}" />
 </x-filament::grid.column>
--- a/packages/support/resources/views/components/loading-section.blade.php
+++ b/packages/support/resources/views/components/loading-section.blade.php
@@ -11,7 +11,7 @@
         ];
     }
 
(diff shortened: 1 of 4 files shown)


packages/support/resources/views/components/loading-section.blade.php
```diff
@@ -10,6 +10,8 @@
             'default' => $columnStart ?? null,
         ];
     }
+
+    $placeholderHeight = $getPlaceholderHeight() ?? '8rem'; // 8rem = h-32
```
danharrin:
```suggestion
    $placeholderHeight = $getPlaceholderHeight() ?? '8rem';
```

packages/support/resources/views/components/loading-section.blade.php
```diff
@@ -27,5 +29,5 @@
     :twoXlStart="$columnStart['2xl'] ?? null"
     class="fi-loading-section"
 >
-    <x-filament::section class="h-32 animate-pulse" />
+    <x-filament::section class="animate-pulse" style="height: {{ $placeholderHeight }};" />
```
danharrin:
```suggestion
    <x-filament::section class="animate-pulse" style="height: {{ $placeholderHeight }}" />
```

danharrin:
Thank you

PR #12629, opened 2024-05-04
Support previews in builder

awcodes:
## Description

This feature supports rendering blade in place of forms in the Builder form field with functionality to show the form in an action modal.

## Visual changes

No visual changes.

## Functional changes

- [ X] Code style has been fixed by running the `composer cs` command.
- [ X] Changes have been tested to not break existing functionality.
- [ X] Documentation is up-to-date.


binaryfire:
@awcodes Thanks for this awesome PR. One performance question - will the views of all blocks in the form be re-rendered when a single block is edited and saved?

Ideally it would work like server-rendered blocks in WP Gutenberg, where changes to a block only cause that one block to re-render.

awcodes:
@binaryfire my understanding until possibly v4 is released is that it will still rerender the entire thing. But since it's only one view per block it will still be vastly more performant than it is with render the form for each block. It's a lot less to dom diff and since it's a blade view it will probably be cached anyway if the data hasn't changed.

binaryfire:
@awcodes Ok thanks. Hopefully v4 will allow for partial updates. I'm thinking of long pages with sliders, galleries etc. The Blade / html isn't really the issue, it's all those JS libraries and associated media files needing to be reloaded on each edit. 

You can see the same problem in the Customizer in WP themes that don't use postMessage / partial refresh. There's a very noticeable reload delay after making each change. The good themes' Customizer implementations, [ACF Extended's dynamic previews](https://www.acf-extended.com/features/fields/flexible-content/dynamic-render) etc. all only refresh the edited section, which makes a big difference.

199ocero:
This is feature is so good @awcodes. Hoping it will merge soon as there is project currently need this. Thank you for this <3

dhruva81:
Can't wait for this feature to be available. I hope it gets merged soon. Thanks, @awcodes!

binaryfire:
@danharrin Is there a way of only refreshing the one specific preview that was edited rather than all of them each time an edit is made? I've been testing some older WordPress stuff that doesn't use partial refresh (i.e. where the whole page is refreshed each time one section is edited) and it gets painful for pages with a lot of media and/or which have sections that use heavy js libraries like carousels etc. Adam mentioned this might be possible with v4?

PR #12656, opened 2024-05-06
Inline global search field suffix

zepfietje:
## Description

The global search field suffix is currently not inline, which means it takes up more space than needed. Also, the prefix icon is in fact inline. Most places I checked have the suffix inline, so I'd suggest streamlining this in Filament too.

Maybe we could introduce an alternative API later to offer more customization options, like this:
```php
->globalSearchField(fn (TextInput $component): TextInput $component->inlineSuffix(false))
```

### Before

Screenshot 2024-05-06 at 17 23 52

### After

Screenshot 2024-05-06 at 17 23 35

## Functional changes

### DIFF
--- a/packages/panels/resources/views/components/global-search/field.blade.php
+++ b/packages/panels/resources/views/components/global-search/field.blade.php
@@ -13,10 +13,11 @@
     </label>
 
     <x-filament::input.wrapper
-        inline-prefix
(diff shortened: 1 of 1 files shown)


zepfietje:
@lukas-frey @danharrin

lukas-frey:
Looks better 👍 I always forget about the inline versions :D

PR #12660, opened 2024-05-07
Actually hide hidden item actions

zepfietje:
## Description

Repeater and builder item actions that are conditionally hidden/visible still render as disabled. This PR fixes that bug by checking if the action is visible for the current item before rendering the action.

## Functional changes


packages/forms/resources/views/components/repeater/index.blade.php
```diff
@@ -79,7 +82,16 @@ class="items-start gap-4"
                                 $extraItemActions,
                                 fn (Action $action): bool => $action(['item' => $uuid])->isVisible(),
                             );
-                            $itemHasToolbar = $isReorderableWithDragAndDrop || $isReorderableWithButtons || filled($itemLabel) || $isCloneable || $isDeletable || $isCollapsible || $visibleExtraItemActions;
+                            $cloneAction = $cloneAction(['item' => $uuid]);
```
danharrin, line 85: I don't think you need to reassign the var, invoking it modifies the existing object

danharrin (changes requested):
Please do the same for the builder & simple repeater

packages/forms/resources/views/components/repeater/index.blade.php
zepfietje: I just tried and it doesn't, so will need to stay as is.

zepfietje:
Done, @danharrin! :)

PR #12672, opened 2024-05-07
Fix bug #8870

### LINKED ISSUE #8870: Keybindings only work first time after page load
### Package

filament/filament

### Package Version

v3.x

### Laravel Version

v10.x

### Livewire Version

v3.x

### PHP Version

PHP 8.x

### Problem description

A defined key binding only works the first time after a page load. Then no action is fired.

This is the only related topic I could find: https://github.com/filamentphp/filament/discussions/3970

### Expected behavior

I expect the action to fire every time I press the key combination.

### Steps to reproduce

1. Go to https://demo.filamentphp.com/shop/products/1/edit
2. Press `CTRL+S`/`CMD+S` and see the page save
3. Try again, and it does not work until the page is fully refreshed

### Reproduction repository

### Relevant log output

_No response_

dmitry-udod:
## Description

The fix for bug #8870 also resolves the issue where the Command+D (Delete button at the top) command only works once on the edit page.

https://github.com/filamentphp/filament/assets/4639175/12e9f6dc-1faf-4124-b9b0-ce5587daffee

## Visual changes

Nope

## Functional changes


packages/support/resources/views/components/button/index.blade.php
```diff
@@ -183,7 +187,8 @@
         x-data="{}"
     @endif
     @if ($keyBindings)
-        x-mousetrap.global.{{ collect($keyBindings)->map(fn (string $keyBinding): string => str_replace('+', '-', $keyBinding))->implode('.') }}
+        id="{{ $keyBindingsAsString }}"
+        x-mousetrap.global.{{ $keyBindingsAsString }}="document.getElementById(@js($keyBindingsAsString)).click()"
```
danharrin: Can we not use `$el.click()` instead of setting an ID?

dmitry-udod: @danharrin That was my first idea, but it’s not working as expected. :(


https://github.com/filamentphp/filament/assets/4639175/5fe46759-e03d-4b0a-85d6-622794e16753

danharrin: Can we use this? I just really want to avoid ID conflicts

https://alpinejs.dev/magics/id

dmitry-udod: Good idea! I will try to implement it.

dmitry-udod: @danharrin Seems magic works :)


https://github.com/filamentphp/filament/assets/4639175/e0fe7b10-a853-4a0e-be4e-b8d2c769b443

danharrin:
Thanks!

PR #12685, opened 2024-05-08
Check login() method exists, punt to setUser()

cheesegrits:
## Description

As Discussed in Discord, in CSV import and export jobs use login() if the method is available in the guard, else punt to setUser().

## Visual changes

No visual changes

## Functional changes


devfrey:
Why does Filament interact with the guard inside a job?

Also, can you link the Discord discussion? I can't find it.

danharrin:
It was in a private channel. This PR is to undo a breaking change that was made a few weeks ago, we are removing `login()` altogether in v4.

PR #12690, opened 2024-05-08
Add language assistance

assertchris:
## Description

This PR adds language assistance helpers (like Hemingway) to the MarkdownEditor component.

## Visual changes

demo

## Functional changes

### DIFF
--- a/packages/forms/resources/js/components/markdown-editor.js
+++ b/packages/forms/resources/js/components/markdown-editor.js
@@ -1,3 +1,11 @@
+const {
+    getSentenceDifficulty,
+    getPassivePhrases,
+    getAdverbPhrases,
+    getComplexPhrases,
+    getQualifiedPhrases,
+} = require('@assertchris/ellison')
+
 window.CodeMirror = require('codemirror/lib/codemirror')
 
 require('codemirror')
@@ -90,6 +98,7 @@ export default function markdownEditorFormComponent({
     maxHeight,
     minHeight,
     placeholder,
+    hasLanguageAssistance,
     state,
     translations,
     toolbarButtons,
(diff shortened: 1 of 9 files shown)


assertchris:
Just to note, I removed inline styles (replaced with tailwind classes) shortly before recording the demo but didn't build the stylesheet before recording. The dots are circular with a re-build of the stylesheet. In case you noticed and it bugs you as much as it does me.

packages/forms/resources/js/components/markdown-editor.js
```diff
@@ -199,6 +215,169 @@ export default function markdownEditorFormComponent({
...
+
+                    const lines = cm.doc.children[0].lines
+
+                    for (let line of lines) {
+                        let sentenceOffset = 0
+
+                        const sentences = getSentenceDifficulty(line.text)
+                        const lineNumber = cm.getLineNumber(line)
+                        const lineElement =
+                            this.$refs.editor.parentNode.parentNode.querySelectorAll(
```
assertchris, line 276: Shameful, I know. CM doesn't provide a way to get the line elements, and this event handler isn't given them either. Can't use renderLine (which does get a line element) because marking in that generates an infinite loop.

danharrin:
To be honest, I am a little overwhelmed with having to maintain something like this

danharrin:
I originally thought that your package would be doing a bit more of the heavy lifting and it was just a case of requiring it

assertchris:
Would it be better if that (roughly) 160 lines of component JS was moved to another lib? I suspect the template stuff is equally to blame for the apprehension; but I don't know how I could package that separately in a way that would keep it just as tightly coupled as it is now.

danharrin:
Yeah, is there a way you can bundle a CM plugin? Maybe you can inject the HTML through that plugin also? I'm not familiar there

assertchris:
I think there's still be significant stuff that needs to be elsewhere (templates, alpine state etc.)

assertchris:
Ok, I'll make another plan. Disappointing amount of work making it this far down a dead end road though. :/

danharrin:
Yeah, I really am sorry about that. Hopefully you can see it from our perspective too though. I'm happy to include it in core as long as we're not the ones responsible for maintaining that code.

assertchris:
Yep, no worries. Thanks for your time.

PR #12694, opened 2024-05-08
Add afterInit hook

assertchris:
## Description

This provides a way to run JS code after the markdown field has been initialised, so that it can be extended by anything in the view.

## Functional changes


packages/forms/resources/js/components/markdown-editor.js
```diff
@@ -210,6 +211,10 @@ export default function markdownEditorFormComponent({
 
                 Alpine.raw(this.editor).value(this.state ?? '')
             })
+
+            if (afterInit) {
+                afterInit.apply(this)
```
danharrin: Can afterInit be a simple function instead of an object with a functional property?

danharrin (changes requested):

packages/forms/resources/js/components/markdown-editor.js
assertchris: I don't understand what you mean. Could you give me an example of what you're thinking?

danharrin:
```suggestion
                afterInit(this)
```

assertchris: ok, so the function would receive an instance of the markdown editor component object, instead of having to resolve it via `this`? I'm ok with that.

danharrin: oh wait, .apply() is some feature of JS to set `this`? I didn't know

danharrin:
Please rename afterInit to setUpUsing

danharrin:
And pass props in alphabetical order

danharrin:
Thanks!

PR #12711, opened 2024-05-09
Feat: Add search value to Select field's relationship modifyQueryUsing

takashato:
## Description

Add Select field component's `$search` value to the evaluation of the callback `$modifyQueryUsing` of field `Select::relationship(modifyQueryUsing: $modifyQueryUsing)`

## Visual changes

No visual change

## Functional changes


mybouhssina:
$search is only injected in reationship() function, which causes the following exception when 'loadFromRelationshipsUsing' is executed:
`An attempt was made to evaluate a closure for [Filament\Forms\Components\Select], but [$search] was unresolvable.`
am I missing something?
Here is my code in a Resource Class:
```PHP
Select::make('categories')
                ->relationship('categories', modifyQueryUsing: fn(Builder $query, ?string $search) => $query)
                ->multiple()
```
My filament version is 3.2.88

danharrin:
What are you trying to do? It was not expected behaviour that we pass that into the `modifyQueryUsing()` as its also used when listing options without a search term. Usually you would just override `getSearchResultsUsing()`

mybouhssina:
> What are you trying to do? It was not expected behaviour that we pass that into the `modifyQueryUsing()` as its also used when listing options without a search term. Usually you would just override `getSearchResultsUsing()`

Maybe I ve misunderstood the intended usage, I ended up using getSearchResultsUsing. But, can you please give me the correct usage for the parameter added in this PR?

Thank you !

danharrin:
Reviewing it again, its probably because there are other places in the method that execute the function and don't pass in the search, and its one of those that is causing your problem

danharrin:
#14024 should fix it and will be in the next release

mybouhssina:
Great, thank you !

PR #12725, opened 2024-05-09
Update item.blade.php

snellingio:
Fixes `{ 'class' : ! tab === 'tab' }`

## Description

Currently there is a bug in the Alpine comparison logic, and because of this, inactive styles are never being applied to tab items. This is because the alpine-active is a `===` comparison.

## Visual changes

Before: 
image

After:
image

## Functional changes


danharrin:
I've just run the formatter and it got rid of this change again. I tried wrapping it in `{{-- format-ignore-start --}}` and it didn't work. So sorry, this isn't going to work like this

Issue #12583, opened 2024-05-02
Summaries error in MSSQL Environment


BukkitAT:
### Package

filament/filament

### Package Version

v3.2.62

### Laravel Version

v11.2.0

### PHP Version

8.3.6

### Problem description

When I am using the summaries feature from table builder in mssql with a complex table name it generate a select query that has a invalid as for the subquery it uses the table name but points are not allowed in the as

The Problem is for examble in Summirzer.php

```php
$query = DB::connection($query->getModel()->getConnectionName())
    ->table($query->toBase(), $query->getModel()->getTable());
```

### Expected behavior

The subquery should generate valid aliases for tables, even in MSSQL environments with complex schema names.

### Steps to reproduce

Create a model with a complex schema name
Use table builder with summarization in MSSQL environment.

```php
TextColumn::make('EFK_VERKAUFSWERT_NETTO')
                    ->summarize(Sum::make())
```

Observe the error due to invalid table aliases in the generated subquery's AS clause.

### Reproduction repository

https://github.com/CruxAllez/filament-issue-mssql

danharrin:
I am not a user of this driver, so I would appreciate you investigating the issue and opening a PR if you can, I don't know if any of the Filament team are going to take active action on this bug report so I don't want you to be left waiting.

PR #12734, opened 2024-05-10
Fix summaries error in MSSQL Environment 

BukkitAT:
This pull request addresses an issue encountered when using the table summarization feature with complex table names in Microsoft SQL Server. The problem arises because the generated SQL subquery uses the full table name as an alias, which is invalid syntax in MSSQL when the table name includes special characters like dots.

This PR fixes the issue reported in [#12583](https://github.com/filamentphp/filament/issues/12583) on the Filament repository.


danharrin:
Thanks

PR #12750, opened 2024-05-12
Fix active indicators when using group options

NathanaelGT:
## Description

When select filter options is grouped, the indicator will not show anything by default. This PR fixes that behaviour

## Visual changes

Before:
image

After:
image

## Functional changes

### DIFF
--- a/packages/tables/src/Filters/SelectFilter.php
+++ b/packages/tables/src/Filters/SelectFilter.php
@@ -65,7 +65,7 @@ class SelectFilter extends BaseFilter
                         ->all();
                 } else {
                     $labels = collect($filter->getOptions())
-                        ->mapWithKeys(fn($label, $value) => is_array($label) ? $label : [$value => $label])
+                        ->mapWithKeys(fn ($label, $value) => is_array($label) ? $label : [$value => $label])
                         ->only($state['values'])
                         ->all();
                 }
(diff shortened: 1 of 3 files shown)


packages/tables/src/Filters/SelectFilter.php
danharrin, line 1: Instead of the `mapWithKeys()`, does it not work with `flatten()`?

danharrin (changes requested):

packages/tables/src/Filters/SelectFilter.php
NathanaelGT: Yes it does, completely forgot about that. I have tested using `collapse()` and that also works

danharrin: I think flatten is probably clearer here

NathanaelGT: Flatten is a bit buggy
```php
Tables\Filters\SelectFilter::make('status')
                    ->default([StudentStatus::Active->value])
                    ->options(StudentStatus::class)
                    ->searchable()
                    ->multiple()
```

That filter will not show any indicator when using flatten

NathanaelGT: Apparently flatten will reset the keys

danharrin: Ah thats annoying, does collapse do that and still work with non-grouped options?

NathanaelGT: I just learned that collapse only works when the array is multi-dimensional. So, it does not

Looks like mapWithKeys is the only viable option

danharrin: Thanks for looking into this

PR #12782, opened 2024-05-14
Feat: Extra field wrapper attributes

atmonshi:
New Method: `extraFieldWrapperAttributes`, for more fields customization without the need for publishing the views.

## Description

the new method `extraFieldWrapperAttributes` will allow to add classes to the`field-wrapper`. my use case it to highlight some fields by custom CSS.

all other `extra*Attributes` will only applies to the input itself

## Visual changes:

Screenshots:
- with `extra*Attributes`:
Screenshot_2024-05-13_110915

- with extraFieldWrapperAttributes:
Screenshot_2024-05-13_105903

## Usage:
```php
TextInput::make('name')
    ->extraFieldWrapperAttributes(['class' => 'components-locked'])
```

## Functional changes


packages/forms/resources/views/components/radio.blade.php
danharrin, line 1: You don't need to pass extraFieldWrapperAttributes into every view like that, in the field wrapper you can call `$field->getExtraFieldWrapperAttributes()` like all the other properties like the label. That should make this PR much much smaller

danharrin (changes requested):

packages/forms/resources/views/components/radio.blade.php
atmonshi: ooh, how I didn't notice all that!!
sorry 😅 reverted and refactored

danharrin (changes requested):
Please do the same for the infolist entry wrapper, and remove the changes to composer.lock

atmonshi:
> Please do the same for the infolist entry wrapper, and remove the changes to composer.lock

all done 👍, and revert composer.lock

packages/infolists/docs/03-entries/01-getting-started.md
```diff
@@ -296,6 +296,15 @@ TextEntry::make('slug')
 
 These get merged onto the outer `<div>` element of each entry in that entry.
 
+Passing extra HTML attributes to the entry wrapper:
```
danharrin:
```suggestion
You can also pass extra HTML attributes to the entry wrapper which surrounds the label, entry, and any other text:
```

packages/forms/docs/03-fields/01-getting-started.md
```diff
@@ -209,6 +209,15 @@ TextInput::make('categories')
     ->extraInputAttributes(['width' => 200])
 ```
 
+Passing extra HTML attributes to the field wrapper:
```
danharrin:
```suggestion
You can also pass extra HTML attributes to the field wrapper which surrounds the label, entry, and any other text:
```

danharrin:
Thanks!

PR #12799, opened 2024-05-15
Check for Duplicate / Empty columns before attempting to submit import

aSeriousDeveloper:
## Description

When importing data through the pre-built import action, duplicate column headers will cause a 500 error to be thrown. This is due to the League/Csv package throwing an Uncaught SyntaxError when these are detected.

This fix catches the syntax error and does some logic to determine whether:

- The duplicate headers are all empty
- There are duplicate header names

A different message is displayed for each, for better user experience. The reason for uniquely detecting empty headers is that seemingly, some spreadsheet software may include completely empty columns even after saving. I'm not fully certain why these would be preserved as this doesn't seem to occur in the latest version of Excel, but it's an issue that some users have had so far when attempting to import data.

### Error when submitting spreadsheet with duplicate columns:
image

### New message when submitting with duplicate column headers
image

### New Message when submitting with duplicate empty column headers
image

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/CanImportRecords.php
+++ b/packages/actions/src/Concerns/CanImportRecords.php
@@ -84,7 +84,7 @@ trait CanImportRecords
                 ->rule('extensions:csv,txt')
                 ->rules([
                     'extensions:csv,txt',
-                    File::types(['csv','txt'])->rules([
+                    File::types(['csv', 'txt'])->rules([
                         function (string $attribute, mixed $value, Closure $fail) use ($action) {
                             $csvStream = $this->getUploadedFileStream($value);
                             $csvReader = CsvReader::createFromStream($csvStream);
--- a/packages/actions/src/Concerns/CanImportRecords.php
+++ b/packages/actions/src/Concerns/CanImportRecords.php
@@ -102,13 +102,9 @@ trait CanImportRecords
                             }
 
                             $columns = Arr::where($duplicates, fn (string $value): bool => $value !== '');
-                            if (empty($columns)) {
-                                $fail(__('filament-actions::import.modal.form.rules.empty_columns'));
-                            } else {
-                                $fail(__('filament-actions::import.modal.form.rules.duplicate_columns', [
-                                    'columns' => implode(', ', $columns),
-                                ]));
-                            }
+                            $fail(trans_choice('filament-actions::import.modal.form.rules.duplicate_columns', count($columns), [
+                                'columns' => implode(', ', $columns),
+                            ]));
                         },
                     ]),
                 ])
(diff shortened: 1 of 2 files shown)


danharrin (changes requested):
Thanks! Can you please investigate refactoring this into a custom validation rule that we pass to `->rule()` of the `FileUpload`? That should declutter it a little bit and control sending the validation message to the correct field etc.

aSeriousDeveloper:
> Thanks! Can you please investigate refactoring this into a custom validation rule that we pass to `->rule()` of the `FileUpload`? That should declutter it a little bit and control sending the validation message to the correct field etc.

For building a custom validation rule, what would you prefer as best practice? Separate namespace within the support / actions package for Validation rules and place in there? or something else?

I'm currently investigating as a validation closure function but that's hitting some roadblocks with file validation, so building it as a separate class makes sense; I just need to know where in the repository would be most ideal.

EDIT: Disregard this, I was just making a bad call to the validator for validating files. I've built this as an anonymous validation function now. I'll add it to the PR for another check. On closer inspection, might be slightly difficult to move to its own rule class, as it depends on some functions that are specific to the CanImportRecords trait.

danharrin:
Ideally it would be a closure rule as its used in only a single place, is there a feature of class based rules that isnt available in closure ones?

Otherwise, Filament\Actions\Imports\Rules I think

aSeriousDeveloper:
Was able to make the validation into a closure rule, not fully sure how to attach my commits to the requested changes in all honesty 😬 , but it should now check for duplicate columns and return the names of them as a validation error.

danharrin (changes requested):
Thanks. Can you please move the translations to `import.modal.form.file.rules`? The current place is for translations related to the failure CSV that is generated after the import finishes?

packages/actions/src/Concerns/CanImportRecords.php
```diff
@@ -80,7 +81,37 @@ protected function setUp(): void
...
+                                $columns = Arr::where($duplicates, fn (string $value): bool => $value !== '');
+
+                                if (empty($columns)) {
+                                    $fail(__('filament-actions::import.failure_csv.empty_columns'));
+                                } else {
+                                    $fail(__('filament-actions::import.failure_csv.duplicate_columns', [
+                                        'columns' => implode(', ', $columns),
+                                    ]));
+                                }
+                            }
```
danharrin: Can this be refactored to utilise early returns so that we don't have nested conditions?

danharrin (changes requested):

packages/actions/src/Concerns/CanImportRecords.php
aSeriousDeveloper: if(! empty($duplicates)) removed from nesting and returns

aSeriousDeveloper:
Nested ifs have been removed, refactored translation string to only use 1 key instead of requiring additional if statement splitting between 2 keys.

danharrin:
Thanks!

sandersjj:
@aSeriousDeveloper thanks for making this PR. I have tested this on my application. The validation rule works. 
However I don't get a nice validation message. I have checked that the translations are available in Dutch (nl)

Screenshot 2024-05-23 at 17 46 42

PR #12826, opened 2024-05-16
Feat: Adding grouping direction hidden method for the table grouping

bensondevs:
## Description

I have a client that asks me to disable the sorting direction of the table group because we want to use the specified sorting direction as preset configuration that cannot be changed. To enable that for the user, we need to add the hide feature for the grouping direction for the user.

## Visual changes

Before:
image
image

After:
image
image

## Functional changes


packages/tables/resources/views/index.blade.php
```diff
@@ -42,6 +42,7 @@
     $isReorderable = $isReorderable();
     $isReordering = $isReordering();
     $areGroupingSettingsVisible = (! $isReordering) && count($groups) && (! $areGroupingSettingsHidden());
+    $isGroupingDirectionHidden = (! $isGroupingDirectionHidden());
```
danharrin: This doesn't make logical sense with the !, the value is true if its actually false

danharrin: You'll need to swap around the `@if` statements in the groups.blade.php view to fix the change as well

danharrin (changes requested):

packages/tables/resources/views/index.blade.php
danharrin: Or, a better alternative could be renaming to `$isGroupingDirectionVisible` and passing the groups.blade.php attribute as just `:direction="true / false"` which I think is still clear?

bensondevs: Updated as suggested, please check

danharrin: Thanks

PR #12829, opened 2024-05-16
feat: add support to override media model on spatie media file upload

WayneHarris:
## Description

Adding support to override the media model on SpatieMediaLibraryFileUpload in line with the ability to do so in the Spatie Media Libday since version [11.4.0](https://github.com/spatie/laravel-medialibrary/releases/tag/11.4.0)

## Functional changes


WayneHarris:
@danharrin, What are your thoughts on this? I can test and review further if you think it is worthwhile.

danharrin (changes requested):
I think the part that reads `getMediaModel()` from the record is fine, but I don't understand the need for the new property and methods to set a component-specific media class? Especially when it's only used within the method that reorders files, not anywhere else? I think this could be greatly simplified to support the new method on the model?

WayneHarris:
> I think the part that reads `getMediaModel()` from the record is fine, but I don't understand the need for the new property and methods to set a component-specific media class? Especially when it's only used within the method that reorders files, not anywhere else? I think this could be greatly simplified to support the new method on the model?

Agree, I did wonder on this too, I'll confirm it's redundant and remove it.

danharrin:
Thank you

PR #12836, opened 2024-05-17
Moves nepali language translations from np directory to ne

pratikkuikel:
## This PR moves the existing nepali translations from np directory to ne directory

In Laravel community packages, the Nepali language locale is recognized as ne. This change aligns Filament translations with the standard directory used by other packages and translation managers, ensuring compatibility and consistency.


danharrin:
This is a breaking change for anyone currently using this language code.

danharrin:
This has been targeted for v4 and will be a step in the upgrade guide.

pratikkuikel:
I think a way to introduce this change as non-breaking would be to make both np and ne directories co-exist like they do right now in actions translations and remove np translations in 4.x . I will make another PR if this is possible.
Actions has ne translations from this PR : https://github.com/filamentphp/filament/pull/12131

danharrin:
I would rather not have duplicate files.

PR #12846, opened 2024-05-17
Update getNavigationBadge method to allow Closures

emmadesilva:
## Description

This allows badges to use dynamic or deferred data. Since the actual NavigationItem::badge method supports this I think it makes sense.

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/panels/src/Resources/Resource.php
+++ b/packages/panels/src/Resources/Resource.php
@@ -2,8 +2,8 @@
 
 namespace Filament\Resources;
 
-use Exception;
 use Closure;
+use Exception;
 use Filament\Clusters\Cluster;
 use Filament\Facades\Filament;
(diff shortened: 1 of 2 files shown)


danharrin:
I think a better solution would be to swap `static::getNavigationBadge()` with `static::getNavigationBadge(...)` so it gets run on demand as a Closure instead of called once? Then people don't need to change how the badge gets returned

Also, does this impact how many times the badge method is run per request, do we need to cache the result?

danharrin:
Closing as my proposed solution is a rewrite anyway

emmadesilva:
> Closing as my proposed solution is a rewrite anyway

Cool! Yeah this was more of a proof of concept to see if it's a good fit. I actually ended up going with your on demand closure route in my application when implementing a workaround yesterday.

PR #12853, opened 2024-05-18
Fix: Incorrectly hiding the 'Select all X' table link

### LINKED ISSUE #12781: The behavior when using `selectCurrentPageOnly`
### Package

filament/filament

### Package Version

v3.2.78

### Laravel Version

v11.0.7

### Livewire Version

v3.4.10

### PHP Version

PHP8.3.7

### Problem description

When `selectCurrentPageOnly` is used, if the same number of items as the pagination count are checked across pages, the "select all :count" display disappears.

### Expected behavior

When `selectCurrentPageOnly` is used, I think the checks should not be counted across pagination.

### Steps to reproduce

1. On the first page, check fewer items than the number of items displayed.
2. On the second page, check enough items to make the total equal to the number of items displayed.
As a result, the "select all :count" display disappears.

As shown in the video

https://github.com/filamentphp/filament/assets/28666304/d2be969b-c084-4f45-bb9d-123824154d4b

### Reproduction repository

https://github.com/bzy107/filament-select-bug

### Relevant log output

_No response_

dev-idkwhoami:
## Description

This PR is related to issue #12781.

The "Select all X" link was hidden by comparing the amount of selected records with the "selectable" records on the page.
But since records stay selected across pages this caused the undesired behavior.

The code change instead checks if the current pages records all are part of the selected records.

## Visual changes

**Before recording can be found in the issue.**

image

There are a total of 5 selected records matching the pages record count but the link is still shown. 
That was previously not possible.

## Functional changes


bzy107:
Thank you for fixing my bug report.
However, I think it's necessary to add a condition to handle cases where selectCurrentPageOnly is not used.

Here is a video showing the case where selectCurrentPageOnly is not used:

https://github.com/filamentphp/filament/assets/28666304/f782510f-0f8d-4b48-b130-7b46af500b24

dev-idkwhoami:
There is another issue.
When switching pages it doesn't refresh so when u go from a page where it should not be visible to a page where it should it stays hidden and the other way around as well.

danharrin:
I've changed it so instead of a random string that refreshes on every request, it will only refresh when the page is changed

danharrin:
Can you test that it still works and let me know please?

dev-idkwhoami:
I pulled the changes you made couldn't find any change in behavior 👍 

brave_O9WtiiY1eK

danharrin:
Thanks!

PR #12860, opened 2024-05-19
Fix: Visual bug when using selectCurrentPageOnly in combination with grouping

dev-idkwhoami:
## Description

While confirming this another PR worked properly with groups. I noticed when selectCurrentPageOnly is used in combination with grouping it causes a visual bug.

The checkbox would display as checked if none or not all records of that group on the current page are selected.

The code change simply adds a livewire key to the group checkbox component so livewire knows to update it.

## Visual changes

Before:
brave_EAz9SiDsWX

After:
brave_XXD7BXT1nW

## Functional changes


danharrin (changes requested):
Can you please refactor this to use a `$page` argument instead of a random string, like I did in your other PR?

And since we'll be calling `$this->getTablePage()` in two places, maybe save this value as `$page` at the top of the index.blade.php file alongside the other variables?

danharrin:
If you want to wait until the other PR is merged, then merge that in to here before you make the changes, that's also fine

PR #12879, opened 2024-05-21
Dispatch DatabaseNotificationsSent from import/export jobs

rossbearman:
## Description

Uses the `$isEventDispatched` parameter on `sendToDatabase()` to dispatch the `DatabaseNotificationsSent` event when import or export jobs notify the user.

Ensures that database notifications from import/export will be broadcast when using Echo to update database notifications. Additionally allows developers to take further action when these notifications are created by listening for the event.

## Functional changes

- [ ] `DatabaseNotificationsSent` will be dispatched from jobs created by the `CanImportRecords` trait and the `ExportCompletion` job.

### DIFF
--- a/packages/actions/src/Concerns/CanImportRecords.php
+++ b/packages/actions/src/Concerns/CanImportRecords.php
@@ -334,7 +334,7 @@ trait CanImportRecords
                                     ->markAsRead(),
                             ]),
                         )
-                        ->sendToDatabase($import->user, true);
+                        ->sendToDatabase($import->user, isEventDispatched: true);
                 })
                 ->dispatch();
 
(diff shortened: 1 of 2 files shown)


packages/actions/src/Concerns/CanImportRecords.php
```diff
@@ -334,7 +334,7 @@ function (string $attribute, mixed $value, Closure $fail) use ($action) {
                                     ->markAsRead(),
                             ]),
                         )
-                        ->sendToDatabase($import->user);
+                        ->sendToDatabase($import->user, true);
```
```suggestion
                        ->sendToDatabase($import->user, isEventDispatched: true);
```

packages/actions/src/Exports/Jobs/ExportCompletion.php
```diff
@@ -74,6 +74,6 @@ public function handle(): void
                     $this->formats,
                 )),
             )
-            ->sendToDatabase($this->export->user);
+            ->sendToDatabase($this->export->user, true);
```
danharrin:
```suggestion
            ->sendToDatabase($this->export->user, isEventDispatched: true);
```

PR #12921, opened 2024-05-24
fix: Apply filters when the encoding is something other than UTF-8

bzy107:
## Description
Modify to apply filters when the encoding is something other than UTF-8 during CSV import.

## Functional changes


danharrin:
Thanks

PR #12923, opened 2024-05-24
Allow to use html in checkboxlist labels

bernhardh:
## Description

Just a minor change of a single return type from `string` to `Htmlable`, so that you can use HTML inside a checkbox label.

## Visual changes

As an example:

```
CheckboxList::make('newsletterSubscriptions')
      ->getOptionLabelFromRecordUsing(function (NewsletterSubscription $record) {
          return new HtmlString( $record->name. " (<small> ". $record->userCount . " Abonnenten</small>)");
      })
```

Before: 

image

After:

image

## Functional changes

- [ x] Code style has been fixed by running the `composer cs` command.
- [ x] Changes have been tested to not break existing functionality.
- [ x] Documentation is up-to-date.

### DIFF
--- a/packages/forms/src/Components/CheckboxList.php
+++ b/packages/forms/src/Components/CheckboxList.php
@@ -251,7 +251,7 @@ class CheckboxList extends Field implements Contracts\CanDisableOptions, Contrac
         return $this->getOptionLabelFromRecordUsing !== null;
     }
 
-    public function getOptionLabelFromRecord(Model $record): string
+    public function getOptionLabelFromRecord(Model $record): Htmlable
     {
         return $this->evaluate(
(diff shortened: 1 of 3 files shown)


packages/forms/src/Components/CheckboxList.php
```diff
@@ -251,7 +251,7 @@ public function hasOptionLabelFromRecordUsingCallback(): bool
         return $this->getOptionLabelFromRecordUsing !== null;
     }
 
-    public function getOptionLabelFromRecord(Model $record): string
+    public function getOptionLabelFromRecord(Model $record): Htmlable
```
danharrin:
```suggestion
    public function getOptionLabelFromRecord(Model $record): string | Htmlable
```

PR #12928, opened 2024-05-24
fixes missing default validation messages on FileUpload

mariusticha:
## Description

We encountered an unexpected behaviour in our running app after a composer update, so we investigated a little and found the origin in this PR: https://github.com/filamentphp/filament/pull/12772

The addition of custom validation messages on a FileUpload has broken the default validation messages thrown in a failing rule.

### Code Example:

```php

// App\Rules\TestRule.php
public function validate(string $attribute, mixed $value, Closure $fail): void
{
    $fail('default validation message');
}


// TestComponent.php
FileUpload::make('test')
    ->rule(new TestRule)

```

### Current Behaviour

On submit the error message shown says 'App\Rules\TestRule'

image

### Expected Behaviour

On submit the error message shown says 'default validation message'

## Functional changes

### DIFF
--- a/packages/forms/src/Components/BaseFileUpload.php
+++ b/packages/forms/src/Components/BaseFileUpload.php
@@ -602,9 +602,7 @@ class BaseFileUpload extends Field implements Contracts\HasNestedRecursiveValida
             $validator = Validator::make(
                 [$name => $files],
                 ["{$name}.*" => ['file', ...parent::getValidationRules()]],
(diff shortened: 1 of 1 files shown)


aSeriousDeveloper:
This change also fixes the validation message that came with #12799.
Currently, the error message given is reverted to `Illuminate\Validation\Rules\File`.

danharrin:
Thanks

PR #12940, opened 2024-05-25
UploadedFile should be converted to TemporaryUploadedFile

recca0120:
## Description

UploadedFile::fake() should covert to TemporaryUploadedFile

## Functional changes

### DIFF
--- a/packages/forms/src/Testing/TestsForms.php
+++ b/packages/forms/src/Testing/TestsForms.php
@@ -7,6 +7,7 @@ use Filament\Forms\ComponentContainer;
 use Filament\Forms\Components\Field;
 use Filament\Forms\Components\TextInput;
 use Filament\Forms\Contracts\HasForms;
+use Illuminate\Http\UploadedFile;
 use Illuminate\Support\Arr;
 use Illuminate\Testing\Assert;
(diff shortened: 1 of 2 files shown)


danharrin:
Please explain what the issue is here because you have provided very little context about what you are fixing

recca0120:
when I use SpatieMediaLibraryFileUpload in test, it should create a record to media table.
before v3.2.76, it work correctly, but after v3.2.76, it is broken
I found this
v3.2.76
https://github.com/filamentphp/filament/blob/04f2bcff66799d14a883818cdda8997fecc9ca3b/packages/forms/src/Testing/TestsForms.php#L35
it will call Livewire\Features\SupportTesting\Testable::uplolad
https://github.com/livewire/livewire/blob/b818a5b85f5ee041b6568fea16a2926a7b91c193/src/Features/SupportTesting/Testable.php#L129-L138

but v3.2.77
only set value to form
https://github.com/filamentphp/filament/blob/v3.2.77/packages/forms/src/Concerns/InteractsWithForms.php#L65

this change is missing Livewire\Features\SupportTesting\Testable::uplolad

danharrin:
Thank you for providing more detail

PR #12958, opened 2024-05-27
Fix Groupable contract implementation order

juliangums:
## Description
The `grouped()` method is implemented on the `StaticAction` class and not the `Action` class:
https://github.com/juliangums/filament/blob/patch-2/packages/actions/src/StaticAction.php#L88
It will not produce any errors currently because `Action` extends `StaticAction`, but it isn't as clean and leads to issues in static analysers like Psalm when extending Action with a custom action.

## Functional changes

### DIFF
--- a/packages/actions/src/Action.php
+++ b/packages/actions/src/Action.php
@@ -5,7 +5,7 @@ namespace Filament\Actions;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Js;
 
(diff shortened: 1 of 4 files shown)


danharrin:
Hi, there are other inheritors of StaticAction that have this interface too, which would also need removing.

juliangums:
@danharrin thanks. Located them and removed it.

PR #12960, opened 2024-05-27
fix return type for navigation icon

juliangums:
Given that `$navigationIcon` is a nullable string as well as `$activeNavigationIcon` (https://github.com/juliangums/filament/blob/3.x/packages/panels/src/Resources/Resource.php#L75),
`getNavigationIcon()` and `getActiveNavigationIcon()` should not be able to return an Htmlable. 
It also causes issues with static analysers if I want to use it like
```php
                MenuItem::make()
                    ->icon(SomeResource::getNavigationIcon()),
```

### DIFF
--- a/packages/panels/src/Resources/Resource.php
+++ b/packages/panels/src/Resources/Resource.php
@@ -804,7 +804,7 @@ abstract class Resource
         static::$navigationParentItem = $item;
     }
 
-    public static function getNavigationIcon(): string | Htmlable | null
+    public static function getNavigationIcon(): ?string
     {
         return static::$navigationIcon;
     }
@@ -814,7 +814,7 @@ abstract class Resource
         static::$navigationIcon = $icon;
     }
 
-    public static function getActiveNavigationIcon(): string | Htmlable | null
+    public static function getActiveNavigationIcon(): ?string
     {
         return static::$activeNavigationIcon ?? static::getNavigationIcon();
     }


danharrin:
Breaking change, and returning an object is something we support - initialising an object is not possible when defining a property. We do not explicitly support creating menu items from resources in the way that you are describing, you could PR support for Htmlable menu item icons if you wish.

juliangums:
@danharrin forget about my use case. I think that made it more confusing instead of clarifying the problem. Actually, you are currently not supporting defining an icon as an Htmlable. It wouldn't make any sense imho but it can also not be set as the setter method only accepts a string:
https://github.com/filamentphp/filament/blob/3.x/packages/panels/src/Resources/Resource.php#L812
and overriding the property in a class that extends Resource will also not work, as the type must be 'null|string' (as in base class '\Filament\Resources\Resource').

So how can an Htmlable ever be returned? Therefore, my change shouldn't break anything, either. It was never possible to set or return an Htmlable in the past and it isn't in the present, and it probably shouldn't be as we are talking about a string that defines the icon like 'heroicon-o-building-library'. Does that make sense, or am I missing something?

danharrin:
As per the docs, it's valid to override the getter instead of using the setter, and keeps all the code in the same place. Anyone using the Htmlable feature will do that, its a common pattern to override the getter throughout Filament.

juliangums:
I see where you're coming from now. That will make sense.

PR #12972, opened 2024-05-28
feature: Custom import file validation

bzy107:
## Description
This is a feature addition for cases where you want to customize the validation of import files by adding a custom validation method.

スクリーンショット 2024-05-28 16 44 39

## Functional changes


danharrin:
Hi, thanks for your contribution. However, this API doesn't really work in a similar way to any of our other APIs, we don't often require users to extend classes to configure them like this. I would suggest you contribute a new method `->fileRules()` to the existing action class, which users can call and pass through new rules. Then you can store these in a property on the action, and then call a getter method for that property when defining the form to merge the rules in. We would accept a new PR using this approach :)

PR #12980, opened 2024-05-28
feature: Custom import file validation

bzy107:
## Description
Revised PR for https://github.com/filamentphp/filament/pull/12972

## Functional changes

### DIFF
--- a/packages/actions/docs/07-prebuilt-actions/08-import.md
+++ b/packages/actions/docs/07-prebuilt-actions/08-import.md
@@ -707,23 +707,20 @@ ImportColumn::make('name')
     ->validationAttribute('full name')
 ```
 
-## Customizing Import File Validation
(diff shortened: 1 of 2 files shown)


danharrin:
Thanks

buzkall:
I've tried this to add xlsx as a valid extension, but I still get the validation error of 'csv, txt' because the fileRules rule is appended to the validation and the first one fails.

I'm setting the rule like this
```
  ImportAction::make()
      ->importer(ExpenseImporter::class)
      ->fileRules([
                      File::types(['csv', 'txt', 'xlsx']),
                  ]),
```

And the debugger shows this:

2024-06-04 at 09 57 33

What's the correct way to allow xlsx extension?

danharrin:
The system cannot handle XLSX input files yet, regardless of the validation

PR #12984, opened 2024-05-28
Use `h-full` instead of `h-screen` for sidebar

### LINKED ISSUE #12983: Weird scrolling behavior when Laravel Debugbar is expanded
### Package

filament/filament

### Package Version

v3.2.83

### Laravel Version

v11.8.0

### Livewire Version

v3.5.0

### PHP Version

PHP 8.3.7

### Problem description

When expanding the [Laravel debugbar](https://github.com/barryvdh/laravel-debugbar) the scrolling does not behave as expected. The whole site should scroll together but instead, the navbar on top of the main content remains fixed (as expected) but the part on top of the sidebar scrolls up.

I've tried it with Chrome, Firefox, and Safari (also using incognito to disable all the extensions) and it happens the same with all of them. 

I've only found this problem using the debug bar for now, but this may happen with other out-of-filament elements. It does not occur when the screen is just smaller than the content

https://github.com/filamentphp/filament/assets/11754473/92dca9eb-fb64-4ded-960f-765da1363af9

Smaller screen

### Expected behavior

Scrolling the page should leave the whole navigation bar fixed on top.

### Steps to reproduce

Navigate to any resource, expand the Laravel Debugbar and scroll down.

### Reproduction repository

https://github.com/josepdecid/events-manager-laravel

### Relevant log output

_No response_

alexmanase:
## Description

Fixes https://github.com/filamentphp/filament/issues/12983

## Visual changes

Before:

SCR-20240529-blbj

After:

SCR-20240529-bldq

## Functional changes

### DIFF
--- a/packages/panels/resources/views/components/sidebar/index.blade.php
+++ b/packages/panels/resources/views/components/sidebar/index.blade.php
@@ -35,7 +35,7 @@
     @endif
     {{
         $attributes->class([
-            'fi-sidebar fixed inset-y-0 start-0 z-30 flex flex-col h-screen content-start bg-white transition-all dark:bg-gray-900 lg:z-0 lg:bg-transparent lg:shadow-none lg:ring-0 lg:transition-none dark:lg:bg-transparent',
+            'fi-sidebar fixed inset-y-0 start-0 z-30 flex flex-col h-full content-start bg-white transition-all dark:bg-gray-900 lg:z-0 lg:bg-transparent lg:shadow-none lg:ring-0 lg:transition-none dark:lg:bg-transparent',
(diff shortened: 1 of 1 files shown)


danharrin:
I've noticed this as well and it seems to fix, thank you

danharrin:
Actually, the container is now no longer the full height of the screen, which could break custom themes if they add a right border to the container. I cannot merge this in its current state, sorry. If you can find a solution while keeping the container full height we can review.

archilex:
If I recall correctly, this issue is actually due to debugbar. Zep and I discussed this a while ago when reworking some of the sidebar stuff. Turning off debug bar makes the problem go away so maybe a PR to debugbar would be a better fix.

PR #12985, opened 2024-05-28
Change css invisibility strategy in Wizard steps

### LINKED ISSUE #9426: CreateRecord wizard navbar and page UI bugs
### Package

filament/filament

### Package Version

v3.0.88

### Laravel Version

v10.30.1

### Livewire Version

v3.0.10

### PHP Version

PHP 8.1

### Problem description

I wanted to implement a wizard for creating records of a resource, so I followed this section of the documentation:

https://filamentphp.com/docs/3.x/panels/resources/creating-records#using-a-wizard

Everything is functioning correctly from a logical standpoint. However, I am encountering several UI issues.

The first problem is that the navigation bar breaks when I scroll, which is likely related to the second problem. The sidebar within the navigation bar becomes fixed after a certain point in scrolling.

image

The second problem is that the page height is excessively long for no apparent reason, resulting in a large white space below the content. It appears that subsequent steps following an initial long step have the same height, even if their content is shorter.

image

### Expected behavior

I expect the navigation bar to remain fixed in its entirety when I scroll through the wizard steps, and for the page height to adjust according to the size of the step's content.

### Steps to reproduce

To reproduce these bugs, simply open the create form for the resource in the provided reproduction repository below and proceed to the second step (after completing the fields in the first step, of course). While on the second step, you can observe all the bugs (this also applies to the 3rd step, as its content is smaller than the first step). You can also return to the first step, and when you scroll all the way to the bottom, you'll encounter the navigation bar bug. This issue is not present before you navigate to the second step for the first time.

### Reproduction repository

https://github.com/JUEChristopher/wizard_ui_bugs

### Relevant log output

_No response_

sebapastore:
## Description

Fixes #9426 

Digging into the issue, I found that both problems described are part of the same root cause, which is a CSS issue. The issue is related to the CSS properties `height: 0; visibility: hidden;` in hidden (or invisible) steps. The problem appears when the previous step is long enough to display the browser scrollbar, and the next step is shorter.

#### CSS examples
As it's a CSS-related issue, I made this simple HTML to make it easy to see what is happening:

[Example 0](https://html-preview.github.io/?url=https://github.com/sebapastore/css-visibility-hidden-issue/blob/main/test0.html): Here you can find a big Header and a Footer (nothing is hidden).
[Example 1](https://html-preview.github.io/?url=https://github.com/sebapastore/css-visibility-hidden-issue/blob/main/test1.html): Here the Header is hidden with `height: 0; visibility: hidden;`, and you can see there is a blank space below the Footer that shouldn't be there.
[Example 2](https://html-preview.github.io/?url=https://github.com/sebapastore/css-visibility-hidden-issue/blob/main/test2.html): Here the Header is hidden with `height: 0; overflow: hidden; position: absolute;`, and you can see it is working as expected.

#### Why not `display: none`?

I searched why `display: none` is not being used. I found that `height: 0; visibility: hidden;` was introduced in a "tab with file" related issue. Here is the [issue](https://github.com/filamentphp/filament/issues/1967), and here is the [solution](https://github.com/filamentphp/filament/pull/1997) that introduced `height: 0; visibility: hidden;` instead of `x-show`. I guess the `step` component "inherited" these properties from `tab`. I tried to reproduce the "tab with file" issue in the Wizard but couldn't, so I think that maybe in the Wizard we are able to use display: none if that was the only reason not to use it. However, in the meantime, I found this other solution.

#### Observations

Another observation is that in the HTML examples I provided, there is really no need to use `position: absolute;`, but in the Wizard, it is necessary for it to work.

The changes have been tested in Safari, Chrome and Firefox.

### DIFF
--- a/packages/forms/resources/views/components/tabs/tab.blade.php
+++ b/packages/forms/resources/views/components/tabs/tab.blade.php
@@ -8,7 +8,7 @@
         'mt-6' => ! $isContained,
     ]);
 
-    $inactiveTabClasses = 'invisible h-0 overflow-y-hidden p-0';
+    $inactiveTabClasses = 'absolute h-0 overflow-hidden p-0';
 @endphp
 
 <div
--- a/packages/forms/resources/views/components/wizard/step.blade.php
+++ b/packages/forms/resources/views/components/wizard/step.blade.php
@@ -8,7 +8,7 @@
         'mt-6' => ! $isContained,
     ]);
 
-    $inactiveStepClasses = 'invisible h-0 overflow-y-hidden p-0';
+    $inactiveStepClasses = 'absolute h-0 overflow-hidden p-0';
 @endphp
 
 <div
--- a/packages/infolists/resources/views/components/tabs/tab.blade.php
+++ b/packages/infolists/resources/views/components/tabs/tab.blade.php
@@ -8,7 +8,7 @@
         'mt-6' => ! $isContained,
     ]);
 
-    $inactiveTabClasses = 'invisible h-0 overflow-y-hidden p-0';
+    $inactiveTabClasses = 'absolute h-0 overflow-hidden p-0';
 @endphp
 
 <div
--- a/packages/support/resources/views/components/section/index.blade.php
+++ b/packages/support/resources/views/components/section/index.blade.php
@@ -172,7 +172,7 @@
             @if ($collapsed || $persistCollapsed)
                 x-cloak
             @endif
-            x-bind:class="{ 'invisible h-0 overflow-y-hidden border-none': isCollapsed }"
+            x-bind:class="{ 'absolute h-0 overflow-hidden border-none': isCollapsed }"
         @endif
         @class([
             'fi-section-content-ctn',


danharrin:
Thank you very much for looking into this and giving us a detailed explanation, it is appreciated

danharrin:
Can you please search the codebase for `invisible h-0` and have a look at if the other places need the same fix? I think there are 3 other files

sebapastore:
Of course, I will look into the other cases!

danharrin:
Could you also explain the thinking behind the change to overflow please?

sebapastore:
`invisible h-0` is being used in 3 more places:

* [Form Tabs](https://github.com/filamentphp/filament/blob/340c30345c2f3f1d84cfbaf97c22b86e3b1ad37d/packages/forms/resources/views/components/tabs/tab.blade.php#L11): In this case, I was able to reproduce the same bug that was found in Wizard, and the same solution works.

* [Infolist Tabs](https://github.com/filamentphp/filament/blob/340c30345c2f3f1d84cfbaf97c22b86e3b1ad37d/packages/infolists/resources/views/components/tabs/tab.blade.php#L11): In this case, I wasn't able to reproduce the bug. Anyway, I tried changing the CSS to use the same solution and it worked.

* [Collapsible Sections](https://github.com/filamentphp/filament/blob/340c30345c2f3f1d84cfbaf97c22b86e3b1ad37d/packages/support/resources/views/components/section/index.blade.php#L175): In this case, I wasn't able to reproduce the bug. Anyway, I tried changing the CSS to use the same solution and it worked.

sebapastore:
> Could you also explain the thinking behind the change to overflow please?

Apparently, the strategy of setting `height: 0; visibility: hidden;` should work, but it doesn't in some scenarios. Knowing we can't use `display: none` due to known related bugs, I had to search for an alternative. Since `height: 0` is already being set, forcing `overflow-x: hidden` should do the trick. In some scenarios, it does, but it's not enough for the 'Wizard case', so I had to add `position: absolute` as well.

danharrin:
> `invisible h-0` is being used in 3 more places:
> 
> * [Form Tabs](https://github.com/filamentphp/filament/blob/340c30345c2f3f1d84cfbaf97c22b86e3b1ad37d/packages/forms/resources/views/components/tabs/tab.blade.php#L11): In this case, I was able to reproduce the same bug that was found in Wizard, and the same solution works.
> * [Infolist Tabs](https://github.com/filamentphp/filament/blob/340c30345c2f3f1d84cfbaf97c22b86e3b1ad37d/packages/infolists/resources/views/components/tabs/tab.blade.php#L11): In this case, I wasn't able to reproduce the bug. Anyway, I tried changing the CSS to use the same solution and it worked.
> * [Collapsible Sections](https://github.com/filamentphp/filament/blob/340c30345c2f3f1d84cfbaf97c22b86e3b1ad37d/packages/support/resources/views/components/section/index.blade.php#L175): In this case, I wasn't able to reproduce the bug. Anyway, I tried changing the CSS to use the same solution and it worked.

Thanks for your investigation, can you push the commits for those components please?

sebapastore:
I have updated the other three components. I have tested that FileUpload is working correctly when editing a form with tabs, in order to make sure the issue #1967 is not present.

danharrin:
Thank you

PR #12998, opened 2024-05-29
The import or export notification header depends on success, partial or failure completion

Sergey-Karapetyan:
When importing or exporting data, the notification title always reads “Import/Export completed.” Receiving an “Import Completed” message when some or all rows were not imported is inconvenient. To resolve this issue, I made the import or export notification header depend on successful, partial, or failed completion.

### DIFF
--- a/packages/actions/resources/lang/de/export.php
+++ b/packages/actions/resources/lang/de/export.php
@@ -46,6 +46,12 @@ return [
 
             'title' => 'Export abgeschlossen',
 
+            'titles' => [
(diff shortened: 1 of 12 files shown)


danharrin:
Hi, unfortunately we cannot merge this change as it would break all the other languages that have not been updated, and any apps that override these translation strings.

I would accept a separate PR for `completionNotificationTitle()`, which accepts a function and `$import` and overrides the default string. You could customize based on that import record

PR #13016, opened 2024-05-30
Fix: ucwords -> title

rhynodesigns:
Fix: Method Illuminate\Support\Str::ucwords does not exist.

https://laravel.com/docs/11.x/strings#method-title-case


danharrin:
The method exists as a macro - https://github.com/filamentphp/filament/blob/3.x/packages/support/src/SupportServiceProvider.php#L129-L134

PR #13021, opened 2024-05-30
Fix #12858

### LINKED ISSUE #12858: MDE behaves differently when submitted via mouse click vs keyBinding
### Package

filament/forms

### Package Version

3.2.80

### Laravel Version

 11.7.0

### Livewire Version

3.4.12

### PHP Version

8.2.18

### Problem description

The behavior of clicking on submit vs pressing the registered `keyBinding()` is different.
Clicking will clear the form. (correct)
Pressing the shortcut key does not. (incorrect)

### Expected behavior

Pressing the shortcut key should also clear the form.

### Steps to reproduce

1. Type something into the markdown editor at `/admin/issue`
2. Press `ctrl + enter`

### Reproduction repository

https://github.com/bilogic/filament-issue/tree/click-different-from-keyBindings

### Relevant log output

https://github.com/filamentphp/filament/assets/946010/fb144bf9-802f-4003-8dcd-ec4c2148e742

_No response_

alexmanase:
## Description

Fixes https://github.com/filamentphp/filament/issues/12858. 

This PR focus on the button, before executing the action in order for EasyMDE (and other components) to be able to change the state.

## Visual changes

It focuses on the button of the action. 

## Functional changes

### DIFF
--- a/packages/support/resources/views/components/badge.blade.php
+++ b/packages/support/resources/views/components/badge.blade.php
@@ -82,7 +82,7 @@
     @endif
     @if ($keyBindings)
         x-bind:id="$id('key-bindings')"
-        x-mousetrap.global.{{ collect($keyBindings)->map(fn (string $keyBinding): string => str_replace('+', '-', $keyBinding))->implode('.') }}="document.getElementById($el.id).click()"
+        x-mousetrap.global.{{ collect($keyBindings)->map(fn (string $keyBinding): string => str_replace('+', '-', $keyBinding))->implode('.') }}="document.getElementById($el.id).focus(); document.getElementById($el.id).click()"
     @endif
     @if ($hasTooltip)
         x-tooltip="{
--- a/packages/support/resources/views/components/badge.blade.php
+++ b/packages/support/resources/views/components/badge.blade.php
@@ -82,7 +82,10 @@
     @endif
     @if ($keyBindings)
         x-bind:id="$id('key-bindings')"
-        x-mousetrap.global.{{ collect($keyBindings)->map(fn (string $keyBinding): string => str_replace('+', '-', $keyBinding))->implode('.') }}="document.getElementById($el.id).focus(); document.getElementById($el.id).click()"
+        x-mousetrap.global.{{ collect($keyBindings)->map(fn (string $keyBinding): string => str_replace('+', '-', $keyBinding))->implode('.') }}="
+            document.getElementById($el.id).focus()
+            document.getElementById($el.id).click()
+        "
     @endif
     @if ($hasTooltip)
         x-tooltip="{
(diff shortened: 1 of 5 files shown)


danharrin:
I'm concerned about the new UX. I don't think you should be defocused from the input when you save, imagine you are writing a long paragraph and want to save as you go with a keyboard shortcut, you don't want to lose your place necessarily

alexmanase:
One solution that I can think of would be to store the keybindings in the global `$store` and to register the key maps when the component is initiated. 

https://codemirror.net/5/doc/manual.html#keymaps

Example:

```js
$store.keymaps.foreach((keymap) => {
   this.editor.codemirror.setOption("extraKeys", {
      keymap: function(cm) {
      }
   });
})
```

bilogic:
> I'm concerned about the new UX. I don't think you should be defocused from the input when you save, imagine you are writing a long paragraph and want to save as you go with a keyboard shortcut, you don't want to lose your place necessarily

@danharrin 

I can think of 3 options

1. How about having another parameter on `->keyBindings(['mod+enter'], stealFocus: true);` to steal focus with default set to `false`?

2. Or is there a way to provide a closure that steals focus first to `->keyBindings()`? Then we won't have to change filament to steal the focus.

3. Or is there a way to clear the MDE from the action?

All 3 can solve the problem of MDE not clearing itself after `mod+enter`

bilogic:
> One solution that I can think of would be to store the keybindings in the global `$store` and to register the key maps when the component is initiated.

I can't quite connect how this will solve the problem. Could you explain? Thank you.

bilogic:
I like to share some thoughts here, + the 4th option
### Thoughts
I gave this some thought and here is my reasoning, are we dealing with a

- HTML form?
- or document save?

A simple HTML form clears itself after a POST, e.g. create page. A form could re-populate its fields if there are validation errors or on an edit page, but that is a result of the response to the POST.

I looked at https://demo.filamentphp.com/blog/posts/1/edit and fields such as `Title` and `slug` have their values returned in the response, the MDE however, is different because the response does not contain its existing value.

Thus, I'm inclined to think that what we have is a form (not document save), but there are some special fields like the MDE which makes it behave like a document.

A very readable solution would be to specify to the action, to clear the MDE after it has been triggered.

### 4th option
1. When keyBinding is triggered, store the `document.activeElement`, see https://stackoverflow.com/a/497108/3872647, do read the comments after the 2020 edit
2. Action button steals focus
3. Action button executes
4. Restore focus to the activeElement

Seems like it addressses all issues, how does that sound?

bilogic:
I tested out the 4th option, it does not solve the problem. (MDE still does not clear because step 4 executes before save completes)

I think option 1 is the best so far, perhaps we could use `->focusIfKeypressed()`?

danharrin:
The true solution will be to work out how to eliminate the focus check in the markdown editor and still avoiding the infinite state update loop.

danharrin:
Going to close this since I don't imagine this PR is going to evolve into something I would merge, the code required to reimplement the focus check is in a completely different place. Definitely open to further PRs for a solution

PR #13028, opened 2024-05-31
Fix: 404 for XLSX file on export

kondratovbr:
## Description

The job that creates XLSX file after export is scheduled after the notification job. Meaning user can get notification about export completion before that XLSX file is created. This causes the button "Download .xlsx" to return 404. The issue is only noticeable for larger exports when `CreateXlsxFile` job can take at least a few seconds.

## Visual changes

No visual changes.

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/CanExportRecords.php
+++ b/packages/actions/src/Concerns/CanExportRecords.php
@@ -219,14 +219,13 @@ trait CanExportRecords
                         fn (PendingBatch $batch) => $batch->name($jobBatchName),
                     )
                     ->allowFailures(),
-                ...(($hasXlsx && (! $hasCsv)) ? [$makeCreateXlsxFileJob()] : []),
+                ...($hasXlsx ? [$makeCreateXlsxFileJob()] : []),
                 app(ExportCompletion::class, [
                     'export' => $export,
                     'columnMap' => $columnMap,
                     'formats' => $formats,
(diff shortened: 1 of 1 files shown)


danharrin:
The correct behaviour is that if the XLSX file doesn't exist when you try and download it, it should generate the file immediately and stream it to you, check the XlsxDownloader action class

danharrin:
With your change, if the user only needs the CSV then the notification gets delayed until the XLSX file is generated which I think makes the UX worse

kondratovbr:
> The correct behaviour is that if the XLSX file doesn't exist when you try and download it, it should generate the file immediately and stream it to you, check the XlsxDownloader action class

You're right, it does. It's not particularly fast for a large export, but it does it.

Thank you so much for your input, you just helped me find a significant issue with our production setup.



> With your change, if the user only needs the CSV then the notification gets delayed until the XLSX file is generated which I think makes the UX worse

UX-wise I'd personally rather get that notification a few seconds later but have it work quickly right away (also not risk hitting execution time limit for really large exports). But my opinion may be skewed - our customers are using these exports for quite large numbers of entries (20k+ and a couple dozen columns with large chunks of text in them) and they're only interested in .xlsx files.

danharrin:
If you disable the CSV export, then the notification will wait for the XLSX one to finish before getting sent :)

PR #13038, opened 2024-05-31
Fix custom size CSS class

underdpt:
## Description

When using button component with custom size an exception is thrown:

```php
<x-filament::button
    size="gap-1.5 px-4 py-3 text-xl"
>
    New
</x-filament::button>
```

## Visual changes

Nothing

## Functional changes

### DIFF
--- a/packages/support/resources/views/components/button/index.blade.php
+++ b/packages/support/resources/views/components/button/index.blade.php
@@ -39,6 +39,8 @@
         $size = filled($size) ? (ActionSize::tryFrom($size) ?? $size) : null;
     }
 
+    $sizeClassSuffix = $size instanceof ActionSize ? $size->value : null;
+
     $iconSize ??= match ($size) {
         ActionSize::ExtraSmall, ActionSize::Small => IconSize::Small,
         default => IconSize::Medium,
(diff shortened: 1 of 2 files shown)


packages/support/resources/views/components/button/index.blade.php
```diff
@@ -62,9 +64,9 @@
             // @deprecated `fi-btn-color-*` has been replaced by `fi-color-*` and `fi-color-custom`.
             is_string($color) ? "fi-btn-color-{$color}" : null,
             is_string($color) ? "fi-color-{$color}" : null,
-            "fi-size-{$size->value}" => $size instanceof ActionSize,
+            "fi-size-{$sizeClassSuffix}" => $size instanceof ActionSize,
```
zepfietje: Could you update the PR to just use the value inline? No need to have the array value as condition if we already conditionally return the string or null.

zepfietje (changes requested):

zepfietje:
Could you apply this change to all other components that need it, @underdpt? Link, icon button, badge, etc.

underdpt:
Hi, 

It seems only link is affected, i've made the requested changes and applied them to the link component.

Thanks

packages/support/resources/views/components/button/index.blade.php
```diff
@@ -62,9 +64,9 @@
             // @deprecated `fi-btn-color-*` has been replaced by `fi-color-*` and `fi-color-custom`.
             is_string($color) ? "fi-btn-color-{$color}" : null,
             is_string($color) ? "fi-color-{$color}" : null,
-            "fi-size-{$size->value}" => $size instanceof ActionSize,
+            $sizeClassSuffix ? "fi-size-{$sizeClassSuffix}" : null,
```
zepfietje:
```suggestion
            ($size instanceof ActionSize) ? "fi-size-{$size->value}" : null,
```

zepfietje (changes requested):

packages/support/resources/views/components/button/index.blade.php
underdpt: Done. 

Long day, sorry

zepfietje (approved):
Thanks!

PR #13040, opened 2024-05-31
The import or export notification header depends on getCompletedNotificationTitle function

Sergey-Karapetyan:
When importing or exporting data, the notification title always reads “Import/Export completed.” Receiving an “Import Completed” message when some or all rows were not imported is inconvenient. To resolve this issue, I made the import or export notification header depend on getCompletedNotificationTitle function to change the title if you want.


danharrin (changes requested):
Please refactor this so the return type is `string` and returns the default translation string within the importer/exporter

danharrin (approved):

PR #13048, opened 2024-05-31
Fix collapsible Block behavior when containing Wizard

### LINKED ISSUE #12623: Wizard Step after step 2 gets `overflow-hidden` class on expanded collapsible container
### Package

filament/filament

### Package Version

v3.2.72

### Laravel Version

v3.2.72

### Livewire Version

v3.4.11

### PHP Version

PHP 8.2.18

### Problem description

When going past step 2 (i.e., step 3+) on a Wizard that is collapsible, the container gets the collapsed classes (`fi-collapsed overflow-hidden`) even though the container is open and no collapse action was triggered by the user

The issue with this is any container JS dropdowns get hidden/cut by the container.

Here's a quick look at the classes that get added from step 3+
https://github.com/filamentphp/filament/assets/5653887/fc5e28cb-4a82-455a-8fc9-74552d563e49

### Expected behavior

The container should not have `overflow-hidden` while it is open.

### Steps to reproduce

Setup the project
```
composer install
npm install
npm run build
```

Start the webserver
```
php artisan serve
```

Go to the Page resource page http://127.0.0.1:8000/admin/pages/create

Click Add to content blocks

Select the radio button `Widgets` and click `Next`
Select the radio button `Select + Checkboxes + Text` and click `Next`

Inspect the wizard step `<li>` classes

it will have something like this:
```
fi-fo-builder-item rounded-xl bg-white shadow-sm ring-1 ring-gray-950/5 dark:bg-white/5 dark:ring-white/10 fi-collapsed overflow-hidden
```
but it should be like this
```
fi-fo-builder-item rounded-xl bg-white shadow-sm ring-1 ring-gray-950/5 dark:bg-white/5 dark:ring-white/10
```

### Reproduction repository

https://github.com/matthewnessworthy/filamentphp-wizard-bug

### Relevant log output

_No response_

sebapastore:
## Description

Fixes #12623 

I found that the issue was related with [this line](https://github.com/filamentphp/filament/blob/6de3b9b55abb4c061ead93b1dc643d8100f3e0df/packages/forms/src/Components/Builder.php#L159) in the `Builder` component. The problem occurs when you have when you have a `$builder->collapsed(true)` and a Wizard with 3 steps. In Step 2 there is also a minor visual bug where the collapsible icon rotates as if the block is collapsed, but it isn't.

I couldn't find any reason why `$component->collapsed(false, shouldMakeComponentCollapsible: false);` should be present in `Builder`. I tested deleting `$component->collapsed(false, shouldMakeComponentCollapsible: false);` from `Builder` and it is working correctly, eliminating the bug.

I also tested the other two parts where `$component->collapsed(false, shouldMakeComponentCollapsible: false);` is used in `Builder`, and found that they can be deleted too.

### DIFF
--- a/packages/forms/src/Components/Builder.php
+++ b/packages/forms/src/Components/Builder.php
@@ -156,8 +156,6 @@ class Builder extends Field implements Contracts\CanConcealComponents, Contracts
 
                 $component->getChildComponentContainer($newUuid ?? array_key_last($items))->fill();
 
-                $component->collapsed(false, shouldMakeComponentCollapsible: false);
-
                 $component->callAfterStateUpdated();
             })
             ->livewireClickHandlerEnabled(false)
@@ -220,8 +218,6 @@ class Builder extends Field implements Contracts\CanConcealComponents, Contracts
 
                 $component->getChildComponentContainer($newKey)->fill();
 
-                $component->collapsed(false, shouldMakeComponentCollapsible: false);
-
                 $component->callAfterStateUpdated();
             })
             ->livewireClickHandlerEnabled(false)
@@ -268,8 +264,6 @@ class Builder extends Field implements Contracts\CanConcealComponents, Contracts
                 }
 
                 $component->state($items);
-
-                $component->collapsed(false, shouldMakeComponentCollapsible: false);
 
                 $component->callAfterStateUpdated();
             })
--- a/packages/panels/resources/views/components/layout/base.blade.php
+++ b/packages/panels/resources/views/components/layout/base.blade.php
@@ -28,7 +28,7 @@
         @endphp
 
         <title>
-            {{ (filled($title) ? "{$title} - " : null) }} {{ $brandName }}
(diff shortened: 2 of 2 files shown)


danharrin:
These lines are present so new items on `collapsed()` builders/repeaters get expanded when they are created, so the form is visible and is editable

matthewnessworthy:
@danharrin would you mind checking the associated bug report and perhaps offer an alternative solution if you have one please? 

It appears that new blocks are added as expanded (as expected), but the initial state is somehow not fully applied so it exists in an expanded state, but also with some of the classes of a collapsed state
the linked issue has a live repo with steps to reproduce the issue

Issue: https://github.com/filamentphp/filament/issues/12623

danharrin:
This issue is in a queue of others and they are dealt with in order.

https://github.com/orgs/filamentphp/projects/2/views/18

Funding the issue with the button in the issue description may make it take higher priority.

matthewnessworthy:
@danharrin thanks! i didn't realise you had a defined queue

PR #13056, opened 2024-06-01
Auth classes rate limit notification method

olehasoft:
## Description

Move the rate limit notification code into a separate method to make it easier to extend the authentication method of the Login class.

## Functional changes

### DIFF
--- a/packages/panels/src/Pages/Auth/EmailVerification/EmailVerificationPrompt.php
+++ b/packages/panels/src/Pages/Auth/EmailVerification/EmailVerificationPrompt.php
@@ -68,17 +68,7 @@ class EmailVerificationPrompt extends SimplePage
                 try {
                     $this->rateLimit(2);
                 } catch (TooManyRequestsException $exception) {
-                    Notification::make()
-                        ->title(__('filament-panels::pages/auth/email-verification/email-verification-prompt.notifications.notification_resend_throttled.title', [
-                            'seconds' => $exception->secondsUntilAvailable,
-                            'minutes' => ceil($exception->secondsUntilAvailable / 60),
-                        ]))
-                        ->body(array_key_exists('body', __('filament-panels::pages/auth/email-verification/email-verification-prompt.notifications.notification_resend_throttled') ?: []) ? __('filament-panels::pages/auth/email-verification/email-verification-prompt.notifications.notification_resend_throttled.body', [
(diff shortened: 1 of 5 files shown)


danharrin (changes requested):
Please copy the implementation of `getCreatedNotification()` on CreateRecord, where the notification is still sent by the parent class if null is not returned. This allows easier fluent customisation of the notification without needing to redefine the entire method.

It should be `getRateLimitedNotification()`

Also please check other instances of rate limited notifications to implement this on consistently.

olehasoft:
@danharrin thank you for your attention!

Implemented `getRateLimitedNotification()` method, also added `getRateLimitedNotificationTitle()` and `getRateLimitedNotificationBody()` methods.

I decided that it would be better to pass `TooManyRequestsException` from now on, because its other properties might be useful.

Also `TooManyRequestsException` already has public `$this->minutesUntilAvailable = ceil($this->secondsUntilAvailable / 60)` so I used that instead of repeating.

olehasoft:
If I implemented correctly, should I add similar changes to this PR for the other `Filament\Pages\Auth` classes: `Register`, `PasswordReset`, `RequestPasswordReset`, `EmailVerificationPrompt`?

In this case, I additionally suggest transferring everything to the `HasRateLimitedNotification` trait with property like this:
```php
protected string $rateLimitedNotificationTranslateBaseKey = 'filament-panels::pages/auth/login.notifications.throttled';
```

danharrin:
Thank you @olehasoft. `getRateLimitedNotificationTitle()` and `getRateLimitedNotificationBody()` can probably be merged into the main function, they are there for backwards compatibility on the other components. That should make it easier to apply to the other pages :)

olehasoft:
@danharrin merged into one `getRateLimitedNotification()` method with code as close as possible to the original.

Although it always returns `Notification`, because `__('filament-panels::pages/auth/login.notifications.throttled.title')` will never return `null`, I left the `?Notification` return declaration for use in extending.

If this meets your requirements, I suggest applying the solution to all the `Filament\Pages\Auth` classes I mentioned above.

danharrin:
Looks perfect, so its ready to be copied to the other classes. Thanks!

olehasoft:
@danharrin all done, thanks!

danharrin (approved):

danharrin:
Thank you for your patience with this review

PR #13074, opened 2024-06-03
Feature: Shift click to select table records (like Gmail and Linear)

neverything:
## Description

Allow multiple records in tables to be selected using shift click as in Gmail or Linear. Was working on this for an app I'm building and saw https://github.com/filamentphp/filament/discussions/13070.

- Alters the table Alpine component by adding two methods to handle the shift + click to select records
- ~Implements the [`updatedPaginators`](https://github.com/livewire/livewire/blob/main/src/Features/SupportPagination/HandlesPagination.php#L51-L68) method to dispatch a browser event in `Filament\Tables\Concerns\InteractsWithTable`~
- ~Implements generic `updated` in `Filament\Tables\Concerns\InteractsWithTable` to dispatch `tableUpdated` event to reinitalize checkbox handling in table.js~
- Using `Livewire.hook('element.init')` and comparing `component.id === this.$wire.id` to check if we need to run `initializeCheckboxes` again.

~**Potential issue:** Implementing `updated` in `Filament\Tables\Concerns\InteractsWithTable` can be an issue if Filament users already implemented this method. If this approach is the correct one, this needs documentation.~

## Functional changes

   - Tested in regular table
   - Tested in relation manager
   - Tested in table widget
   - Tested in table with grid layout
   - Tested with table grouping
   - Tested with search
   - Tested with `deferLoading()`
   - Tested with a table in a Livewire component
- [x] Documentation is up-to-date. (no changes made)

## Visual changes
Not visible in the video: `Shift` + click to select/unselect the records between.
filament-pr-13074

## Ideas for improvement

- Maybe there is a simpler way to get updates from the table, like pagination in the Alpine component?
- Eventually this should be conditionally turned on and off

## Changes needed

- [x] In order for grouping and search to work, another event is needed
- [x] Testing with Livewire component with a Filament table
- [x] Testing with inline editing of table
- [x] Checking if `Livewire.hook('element.init', ...` is impacting performance

**Disclaimer:** This is my first PR, please let me know how I can do better, happy to work on it.


neverything:
Looking for feedback: If anyone has the chance, please let me know if this is going into the right direction? I'm not sure if there isn't a better way to capture changes of the records shown in the table.

packages/tables/resources/js/components/table.js
```diff
@@ -157,5 +167,54 @@ export default function table() {
...
+                let range = [start, end].sort((a, b) => a - b)
+                for (let i = range[0]; i <= range[1]; i++) {
+                    checkboxes[i].checked = checkbox.checked
+                    let value = checkboxes[i].value
+                    if (checkbox.checked) {
+                        this.selectRecords([value])
+                    } else {
+                        this.deselectRecords([value])
+                    }
+                }
```
danharrin: Instead of calling `selectRecords()` and `deselectRecords()` multiple times, is it possible to construct an array of checkboxes to be selected and an array to be deselected, and then call each method just once after the loop has finished?

danharrin (changes requested):

packages/tables/resources/js/components/table.js
neverything: Good catch, fixed in https://github.com/filamentphp/filament/pull/13074/commits/3728b38cf95bee57f8411b5a7ae1ac1ade7a3065

danharrin:
@neverything let me know when you have finished testing performance and this should be ready to merge

neverything:
> @neverything let me know when you have finished testing performance and this should be ready to merge

@danharrin thanks for your checks and replies. I did not find a measurable impact when using `Livewire.hook('element.init', ...`, but I could be wrong. Tested in Chrome with a few different tables, but might be different if there are tons of Livewire components on a single page.

The other thing that users might want, but I haven't figured out an easy way, is to turn this feature on and off.

Besides that it works well and I couldn't find any issues and it should be good to go if you agree.

danharrin:
Personally I don't think anyone would want to disable this feature as its basically a shortcut instead of a change in behaviour, I'd be open to another PR in the future if anyone needs to do that

neverything:
I agree, makes sense to have it on by default. Thank you @danharrin

elbaylot:
Thanks for this amazing feature @neverything 😍

danharrin:
Thanks

neverything:
As reported by @ekrist1 in the [release](https://github.com/filamentphp/filament/discussions/13177#discussioncomment-9728588), my change broke `spa()` mode.

If anyone has an idea to check against `this.$wire.id` after a `livewire:navigated` event, feel free to let me know.

As mentioned in the [comment](https://github.com/filamentphp/filament/discussions/13177#discussioncomment-9729623) on the release:

Something like this:
```javascript
Livewire.hook('element.init', () => {
    this.watchForCheckboxClicks()
})
```
instead of:
```javascript
Livewire.hook('element.init', ({ component }) => {
    if (component.id === this.$wire.id) {
        this.watchForCheckboxClicks()
    }
})
```

But doing this on every `element.init` might be a bit crazy and I haven't found an easier way yet.

neverything:
Potential fix in https://github.com/filamentphp/filament/pull/13186

PR #13082, opened 2024-06-04
fix: handle panel ID correctly in make:filament-resource command

### LINKED ISSUE #13081: Incorrect Panel ID Handling in make:filament-resource Command with --panel Flag
### Package

filament/filament

### Package Version

v3.2.85

### Laravel Version

v11.9.2

### Livewire Version

_No response_

### PHP Version

PHP 8.3.7

### Problem description

When creating a Filament panel using the artisan command `php artisan make:filament-panel MyPanel`, the command calls `generatePanel` from the trait `CanGeneratePanels`, and sets the panel ID by modifying the first letter to lowercase (using `Str::lcfirst` - `CanGeneratePanels.php` line 18). So the panel name for `php artisan make:filament-panel MyPanel` by design becomes `myPanel`.

When executing `php artisan make:filament-resource User --generate`, the resource is created in `app/Filament/System/Resources/UserResource.php`, but it always asks for manual input of "Which panel would you like to create this in?". Selecting the panel from the options available works, as the panel IDs are correctly retrieved using the `getPanels` method, which gets it from the app container instance. The problem appears when the flag `--panel` is used without the correct panel ID.

Since `Str::lcfirst` is not an explicitly intuitive behavior (for single-word panel names, it could look like the panel name could also be using only all lowercase), someone creating new Filament resources could, for example, mistakenly run `php artisan make:filament-resource User --generate --panel=MyPanel`, or `php artisan make:filament-resource User --generate --panel=mypanel`, or `php artisan make:filament-resource User --generate --panel=my-panel`.

Running that does not generate an error but instead returns a default Panel, because in the class `PanelRegistry`, the `get` method returns `$this->getDefault` if the ID is not found. This can create confusion for the developer trying to understand why the resource being created for the panel specified in the command line is not being set in `Filament\Resources\MyPanel`, but instead in `Filament\Resources`.

**Proposed Solution:**
Modify the `handle` method in the `MakeResourceCommand` class to check the panel ID correctly. If the provided panel ID does not match, attempt to match using `Str::lcfirst`. If still not found, return an error message to the user.

### Expected behavior

The command should correctly identify the panel by its name, even if the case of the provided ID does not exactly match, due to the first letter being capitalized. It should also provide an error message if the panel is not found, instead of defaulting to the default panel. The folder to be created will have the panel name with the first letter capitalized.

### Steps to reproduce

1. Create a Filament panel using `php artisan make:filament-panel MyPanel`.
2. Attempt to create a resource in this panel using `php artisan make:filament-resource User --generate --panel=MyPanel`.
3. Notice that the resource is created in the default panel instead of MyPanel.

### Reproduction repository

https://github.com/mwguerra/filament-fix-panel-id-handling-reproduction.git

### Relevant log output

```shell
$ php artisan make:filament-resource User --generate --panel=MyPanel

   INFO  Filament resource [app/Filament/Resources/UserResource.php] created successfully.  

$ php artisan make:filament-resource User --generate --panel=InexistentPanel

 ┌ UserResource.php already exists, do you want to overwrite it? ┐
 │ Yes                                                           │
 └───────────────────────────────────────────────────────────────┘

 ┌ ListUsers.php already exists, do you want to overwrite it? ──┐
 │ Yes                                                          │
 └──────────────────────────────────────────────────────────────┘

 ┌ CreateUser.php already exists, do you want to overwrite it? ─┐
 │ Yes                                                          │
 └──────────────────────────────────────────────────────────────┘

 ┌ EditUser.php already exists, do you want to overwrite it? ───┐
 │ Yes                                                          │
 └──────────────────────────────────────────────────────────────┘

   INFO  Filament resource [app/Filament/Resources/UserResource.php] created successfully.  

$ php artisan make:filament-resource User --generate --panel=myPanel        

   INFO  Filament resource [app/Filament/MyPanel/Resources/UserResource.php] created successfully.  

$
```

mwguerra:
**Description**:
This pull request addresses issue #13081. The `make:filament-resource` command was not handling panel IDs correctly when the `--panel` flag was used. The command now correctly checks the panel ID, attempts to match using `Str::lcfirst`, and provides an error message if the panel is not found.

**Visual changes**:
There are no visual changes associated with this PR.

**Functional changes**:

Closes #13081

**Code Changes**:
Here are the relevant changes in the MakeResourceCommand.php:
```php
<?php

namespace Filament\Commands;

use Filament\Clusters\Cluster;
use Filament\Forms\Commands\Concerns\CanGenerateForms;
use Filament\Panel;
use Filament\Support\Commands\Concerns\CanIndentStrings;
use Filament\Support\Commands\Concerns\CanManipulateFiles;
use Filament\Support\Commands\Concerns\CanReadModelSchemas;
use Filament\Tables\Commands\Concerns\CanGenerateTables;
use Illuminate\Console\Command;
use Illuminate\Support\Arr;
use Illuminate\Support\Str;
use Symfony\Component\Console\Attribute\AsCommand;

use function Laravel\Prompts\select;
use function Laravel\Prompts\text;

#[AsCommand(name: 'make:filament-resource')]
class MakeResourceCommand extends Command
{
    use CanGenerateForms;
    use CanGenerateTables;
    use CanIndentStrings;
    use CanManipulateFiles;
    use CanReadModelSchemas;

    protected $description = 'Create a new Filament resource class and default page classes';

    protected $signature = 'make:filament-resource {name?} {--model-namespace=} {--soft-deletes} {--view} {--G|generate} {--S|simple} {--panel=} {--model} {--migration} {--factory} {--F|force}';

    public function handle(): int
    {
        // Existing code...

        $panelOption = Str::lcfirst($this->option('panel'));

        if ($panelOption) {
            $panel = Filament::getPanel($panelOption);

            if ($panel->getId() !== $panelOption) {
                $this->components->error("Panel [{$panelOption}] not found.");
                return static::INVALID;
            }
        }

        if (! $panelOption) {
            $panels = Filament::getPanels();

            /** @var Panel $panel */
            $panel = (count($panels) > 1) ? $panels[select(
                label: 'Which panel would you like to create this in?',
                options: array_map(
                    fn (Panel $panel): string => $panel->getId(),
                    $panels,
                ),
                default: Filament::getDefaultPanel()->getId()
            )] : Arr::first($panels);
        }

        // Rest of the code...
    }
}
```

### DIFF
--- a/packages/panels/src/Commands/MakeClusterCommand.php
+++ b/packages/panels/src/Commands/MakeClusterCommand.php
@@ -45,7 +45,7 @@ class MakeClusterCommand extends Command
         $panel = $this->option('panel');
 
         if ($panel) {
-            $panel = Filament::getPanel($panel);
+            $panel = Filament::getPanel($panel, isStrict: false);
         }
 
         if (! $panel) {
--- a/packages/panels/src/Commands/MakePageCommand.php
+++ b/packages/panels/src/Commands/MakePageCommand.php
@@ -53,7 +53,7 @@ class MakePageCommand extends Command
         $panel = $this->option('panel');
 
         if ($panel) {
-            $panel = Filament::getPanel($panel);
+            $panel = Filament::getPanel($panel, isStrict: false);
         }
 
         if (! $panel) {
--- a/packages/panels/src/Commands/MakeRelationManagerCommand.php
+++ b/packages/panels/src/Commands/MakeRelationManagerCommand.php
@@ -62,7 +62,7 @@ class MakeRelationManagerCommand extends Command
         $panel = $this->option('panel');
 
         if ($panel) {
-            $panel = Filament::getPanel($panel);
+            $panel = Filament::getPanel($panel, isStrict: false);
         }
 
         if (! $panel) {
--- a/packages/panels/src/Commands/MakeResourceCommand.php
+++ b/packages/panels/src/Commands/MakeResourceCommand.php
@@ -85,7 +85,7 @@ class MakeResourceCommand extends Command
         $panel = $this->option('panel');
 
         if ($panel) {
-            $panel = Filament::getPanel($panel);
+            $panel = Filament::getPanel($panel, isStrict: false);
         }
 
         if (! $panel) {
--- a/packages/panels/src/Commands/MakeResourceCommand.php
+++ b/packages/panels/src/Commands/MakeResourceCommand.php
@@ -12,6 +12,7 @@ use Filament\Support\Commands\Concerns\CanReadModelSchemas;
 use Filament\Tables\Commands\Concerns\CanGenerateTables;
 use Illuminate\Console\Command;
 use Illuminate\Support\Arr;
+use Illuminate\Support\Str;
 use Symfony\Component\Console\Attribute\AsCommand;
 
 use function Laravel\Prompts\select;
@@ -82,13 +83,18 @@ class MakeResourceCommand extends Command
         $pluralModelClass = (string) str($modelClass)->pluralStudly();
         $needsAlias = $modelClass === 'Record';
 
-        $panel = $this->option('panel');
+        $panelOption = Str::lcfirst($this->option('panel'));
 
-        if ($panel) {
-            $panel = Filament::getPanel($panel);
+        if ($panelOption) {
+            $panel = Filament::getPanel($panelOption);
+
+            if ($panel->getId() !== $panelOption) {
+                $this->components->error("Panel [{$panelOption}] not found.");
+                return static::INVALID;
+            }
         }
 
-        if (! $panel) {
+        if (! $panelOption) {
             $panels = Filament::getPanels();
 
             /** @var Panel $panel */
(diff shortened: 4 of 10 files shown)


danharrin (changes requested):
I think the fix belongs in `PanelRegistry::get()`. If an ID is not null but not found, it should throw an exception. If the ID is not passed, it should return the default.

Instead of `$this->panels[$id]`, you should create a temporary array where all the panel IDs are fully lowercase and have `-` and `_` removed. Then do the same to the ID. Then the method can match `FooBar` to `foobar` `fooBar` `FoObAr`, `foo-bar`, `foo_bar` as well

danharrin:
Actually, I think there should be a new `$isStrict = true` method on the `get()` of the Filament facade, manager, and panel registry, which optionally uses this new index of panels. As we don't need that behaviour in every place.

Then we pass through `isStrict: false` from all the commands. It's not just the resource one, theres others like `make:filament-page`, make sure to check all usages

mwguerra:
I've made some updates to the `getPanel` method and introduced a default `isStrict` parameter set to `true`. This means that, by default, the method operates in strict mode unless specified otherwise. Only when commands directly involving user input are executed is `isStrict` set to `false`.

For better separation of concerns and improved readability, I extracted the logic into a protected `findPanel` method. This method adheres to the `isStrict` parameter and returns `null` if a panel isn't found, ensuring strict behavior as intended.

However, the `getPanel` method still maintains the original fallback mechanism. If no panel is found, it returns the default panel. This approach ensures that we avoid user experience issues that could arise from strictly returning `null`. For instance, methods like `getRouteName` could be negatively impacted if they attempt to access a non-existent panel.

If you have any questions or need further adjustments, I'm happy to help. Just let me know!

danharrin:
Perfect, thanks for understanding!

PR #13086, opened 2024-06-04
Able to add input attributes to disabled rich editor

nielsvh2103:
## Description
This will make it so that you can add attributes to your disabled rich editor input.
We actually have a scenario where we want to add some classes to a disabled rich editor input.

Im sure we are not the only one that might need this :)

Snippet:
```
          RichEditor::make('ourField')
                            ->extraInputAttributes(['class' => 'myclasses']) // this will be possible as its not possible now
                            ->label('Our field')
                            ->disabled()
```

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/forms/resources/views/components/rich-editor.blade.php
+++ b/packages/forms/resources/views/components/rich-editor.blade.php
@@ -12,7 +12,12 @@
                 state: $wire.{{ $applyStateBindingModifiers("\$entangle('{$statePath}')") }},
             }"
             x-html="state"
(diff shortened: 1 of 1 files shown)


danharrin:
Hi, we strictly allow the input attribute bag on HTML input elements only. `<div>` is not an input element.

To add styles to a disabled rich editor, you can target the `.fi-fo-rich-editor.fi-disabled` class in your CSS file

nielsvh2103:
I see @danharrin 

Though it'd be nice to add atleast styling (or classes) or some kind of way to specific editors, even if it is a div element.
Now it might be rare for 2 disabled editors to have different css, but it can still happen depending on what someone wants.

Targeting the class in my css file will be global, is there a way to do it per editor/page?

danharrin:
You could use `extraFieldWrapperAttributes(['id' => 'myEditor1'])` on the field to add attributes to the outer element of the field wrapper, then target `#myEditor1 .fi-fo-rich-editor.fi-disabled`

PR #13099, opened 2024-06-05
feat: set locale on DateTimePicker field

darkclow4:
## Description

This PR adds the ability to customize locale on date time picker, especially when we use display format 'F' and we want to use locale different from locale on app config.

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/forms/docs/03-fields/08-date-time-picker.md
+++ b/packages/forms/docs/03-fields/08-date-time-picker.md
@@ -85,7 +85,7 @@ DatePicker::make('date_of_birth')
 
 <AutoScreenshot name="forms/fields/date-time-picker/display-format" alt="Date time picker with custom display format" version="3.x" />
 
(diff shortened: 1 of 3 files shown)


packages/forms/docs/03-fields/08-date-time-picker.md
```diff
@@ -85,6 +85,17 @@ DatePicker::make('date_of_birth')
 
 <AutoScreenshot name="forms/fields/date-time-picker/display-format" alt="Date time picker with custom display format" version="3.x" />
 
+You may also configure locale when you use display format on the field and you want to use different locale from your app config. For this, you can use the `locale()` method.
```
danharrin:
```suggestion
You may also configure the locale that is used when rendering the display, if you want to use different locale from your app config. For this, you can use the `locale()` method:
```

danharrin:
Perfect implementation, thanks!

darkclow4:
It's my pleasure

PR #13102, opened 2024-06-05
Ignore textarea style height by default, not just on auto size

### LINKED ISSUE #13090: Using textarea with live() resets the height when typing
### Package

filament/forms

### Package Version

3.2.83

### Laravel Version

11.10

### Livewire Version

Latest (3.5.0)

### PHP Version

8.3.7

### Problem description

Edit: i just realized it always happens unless you add autosize (we do not want autosize in this case)

Using live() in a textarea does not work properly.

We tried using it in a repeater as well on live(true) (with on blur), that didnt work either

Happens when typing or pasting (ctrl + v)

https://github.com/filamentphp/filament/assets/60341437/df7f8783-de03-4e5c-a162-46b8b514a6d7

### Expected behavior

Simple, the resized height should preserve and not go back to initialheight (that is set in the view)

### Steps to reproduce

```
                Textarea::make('description')
                                    ->label(false)
                                    ->maxLength('180')
                                    ->nullable()
                                    ->live(),
```

Just add live, or live(true) to a textarea input
Should be set for you in my reproduction repository.

### Reproduction repository

https://github.com/nielsvh2103/filament-issue

### Relevant log output

_No response_

nielsvh2103:
## Description

See issue #13090 
This just changes the fact that it always ignores style height instead of just on autosize enabled.

Because when resizing and then typing, it resets (by using live() or on blur)

I've tested it manually by checking with live, live(onblur), without live() and by using autosize aswell.
Should all work now, instead of without autosize it being resetting.

Unless there are scenario's where you don't want to ignore the height that is, but can't think of any reason someone would want that :)

## Visual changes

Fixes textarea resizing without using autosize, it wont reset every keystroke / on blur.

## Functional changes

- [x] Code style has been fixed by running the `composer cs` command. (no changes)

### DIFF
--- a/packages/forms/resources/views/components/textarea.blade.php
+++ b/packages/forms/resources/views/components/textarea.blade.php
@@ -42,13 +42,13 @@
                 @endif
                 ax-load-src="{{ \Filament\Support\Facades\FilamentAsset::getAlpineComponentSrc('textarea', 'filament/forms') }}"
                 x-data="textareaFormComponent({ initialHeight: @js($initialHeight) })"
-                x-ignore
                 x-intersect.once="render()"
                 x-on:input="render()"
                 x-on:resize.window="render()"
-                wire:ignore.style.height
                 {{ $getExtraAlpineAttributeBag() }}
             @endif
+            x-ignore
+            wire:ignore.style.height
             {{
                 $getExtraInputAttributeBag()
                     ->merge([


danharrin:
Thanks

mozex:
@danharrin This PR raises a problem with a disabled textarea. `wire:ignore.style.height` is the same as `wire:ignore`, and when the textarea is disabled dynamically, `wire:ignore` prevents it from changing the disabled state.

danharrin:
hmm, strange how that managed to slip in somewhere, I thought that was valid syntax when I reviewed it

do you have any alternative suggested solution for being able to persist the height?

nielsvh2103:
Hi, to come back to it as well

it seems this was already an issue, i merely placed this to work on all textarea's
Which means it would've been broken before it was merged in this way also;

```
            @if ($shouldAutosize)
                @if (FilamentView::hasSpaMode())
                    ax-load="visible"
                @else
                    ax-load
                @endif
                x-intersect.once="render()"
                x-on:input="render()"
                x-on:resize.window="render()"
            x-ignore
            wire:ignore.style.height
                {{ $getExtraAlpineAttributeBag() }}
            @endif
```

though it would just be broken for the autosize. ( i just confirmed this )

im checking it out as well now

maybe its as simple as checking if its disabled or not..

danharrin:
I think the solution will be to store the height in an Alpine variable and then `x-bind:style` to that variable. The variable will be persisted between requests then. Thoughts?

mozex:
I tried many things to achieve this, but the only solution was to call render() on every request, which is not ideal. Maybe storing height in a variable will work.

nielsvh2103:
> I tried many things to achieve this, but the only solution was to call render() on every request, which is not ideal. Maybe storing height in a variable will work.

Yeah because render() is meant for the autosize to persist on input, you don't want to call render on every request indeed.

I just tried simply by 

```
            @if(!$isDisabled)
            x-ignore
            wire:ignore.style.height
            @endif
```

But this re introduces the issue of the height not persisting also when disabled is being removed dynamically, it will jump back to the rows height (initialHeight that is loaded in the view)

Therefore i think @danharrin solution idea might fit here.

danharrin:
If possible can one of you try and make a PR based on my comment?

nielsvh2103:
Ill check it out

mozex:
I just tested the idea, and it works; if you need, I can make a PR.

nielsvh2103:
> I just tested the idea, and it works; if you need, I can make a PR.

I'd say go ahead, i haven't tried yet as im busy :)
If you make a PR i can check it in my scenario's also!

mozex:
I just opened #13210; @danharrin, can you look at it?

nielsvh2103:
Could both of you check out https://github.com/filamentphp/filament/pull/13211 @mozex @danharrin

PR #13125, opened 2024-06-06
Prevent infolist entry grid blowout

ralphjsmit:
## Description

I have an a custom infolist component. The contents of this custom component are sometimes wider than the actual space allowed in the infolist grid layout. This is handled in the custom component by setting an `overflow-x-scroll`. 

However, this overflow class isn't applied om the infolist, because the infolist entry wrapper grid gives the children automatically a width of `auto`, meaning that because the infolist entry wrapper has this grid, all the children are [automatically given as much width as they can occupy](https://css-tricks.com/preventing-a-grid-blowout/), without giving the developer an option to use a scroll overflow for example.

I tested by and this doesn't change any behaviour for existing custom components that are overflow, only for components that have an `overflow-...` property set.

Thanks!

## Functional changes

### DIFF
--- a/packages/infolists/resources/views/components/entry-wrapper/index.blade.php
+++ b/packages/infolists/resources/views/components/entry-wrapper/index.blade.php
@@ -123,6 +123,7 @@
                         Alignment::Right => 'text-right',
                         default => $alignment,
                     },
+                    'min-w-0',
                 ])
             >
                 @if ($url)
--- a/packages/infolists/resources/views/components/entry-wrapper/index.blade.php
+++ b/packages/infolists/resources/views/components/entry-wrapper/index.blade.php
@@ -101,7 +101,7 @@
 
         <div
             @class([
-                'grid gap-y-2',
+                'grid auto-cols-fr gap-y-2',
                 'sm:col-span-2' => $hasInlineLabel,
             ])
         >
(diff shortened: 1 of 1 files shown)


packages/infolists/resources/views/components/entry-wrapper/index.blade.php
```diff
@@ -123,6 +123,7 @@
                         Alignment::Right => 'text-right',
                         default => $alignment,
                     },
+                    'min-w-0',
```
zepfietje: Wouldn't `grid-cols-1` be a slightly cleaner fix in this case?
It's the equivalent of `grid-template-columns: repeat(1, minmax(0, 1fr));`, which fixes the overflow issue at the grid level (as explained in the article you linked to).

zepfietje (changes requested):

packages/infolists/resources/views/components/entry-wrapper/index.blade.php
zepfietje: Tested and achieves the same result.

zepfietje (approved):

zepfietje:
Thanks for bringing this up and referencing that article, @ralphjsmit!

ralphjsmit:
Thank you for the thorough review! Yeah, I agree the grid-template-columns solution is better to do it on the grid level. I can work well with Tailwind, but my main focus is now on PHP/backend, so I don't know all these CSS intricacies 😄 Good weekend!

zepfietje:
No worries, Ralph! It's all a team effort. Enjoy your Sunday. 😊

PR #13130, opened 2024-06-07
Add missing class string type annotations

juliangums:
This is more in line with what's been done here for example: https://github.com/filamentphp/filament/blob/3.x/packages/actions/src/Exports/Exporter.php#L66

There are a few more places where this could also be added. If needed/wanted I can add them to the PR too.


danharrin:
Can you please investigate the PHPstan failures as a result of this PR?

danharrin:
I have had to remove the resource annotation, PHP CS Fixer incorrectly changes it from `class-string<Resource>` to `class-string<resource>`

PR #13163, opened 2024-06-09
Update documentation to include persistentMiddleware method usage

alifirzb:
## Functional changes


danharrin:
This is the tenancy documentation and persistent middleware applies to non tenant panels as well. This method is for internal purposes and users can use Livewire's API to do this: https://livewire.laravel.com/docs/security#configuring-persistent-middleware

PR #13165, opened 2024-06-09
Fix wrong type in docs

thiagolcks:
In the example it was using an array when it should be a string or mixed, but not an array.


zepfietje:
Good catch, thanks!

PR #13171, opened 2024-06-10
Allow multiple lines to be added to the import example file.

Sergey-Karapetyan:
There was a need to add more than 1 line to the import example file, but this is not possible now. Added an "examples" function to columns that takes an array and generates a file based on it.

### DIFF
--- a/packages/actions/src/Concerns/CanImportRecords.php
+++ b/packages/actions/src/Concerns/CanImportRecords.php
@@ -334,22 +334,13 @@ trait CanImportRecords
                         $columns,
                     ));
 
-                    /*
-                    Сount the number of lines in the file.
-                    It should be equal to the largest number of examples in the columns.
-                    */
-                    $rowsCount = 0;
-                    array_map(
-                        function (ImportColumn $column) use (&$rowsCount) {
-                            $rowsCount = count($column->getExamples()) > $rowsCount ? count($column->getExamples()) : $rowsCount;
-                        },
+                    $rowsCount = array_reduce(
                         $columns,
+                        function ($maxCount, ImportColumn $column) {
+                            return max($maxCount, count($column->getExamples()));
+                        }
                     );
 
-                    /*
-                    Get an array of “examples” in which the number of elements
-                    is equal to the number of lines in the file.
-                    */
                     $examples = [];
                     foreach ($columns as $column) {
                         $columnExamples = $column->getExamples();
@@ -359,9 +350,6 @@ trait CanImportRecords
                         }
                     }
 
-                    /*
-                    Filling out the file
-                    */
                     $csv->insertAll($examples);
 
                     return response()->streamDownload(function () use ($csv) {
--- a/packages/actions/src/Concerns/CanImportRecords.php
+++ b/packages/actions/src/Concerns/CanImportRecords.php
@@ -338,7 +338,8 @@ trait CanImportRecords
                         $columns,
                         function ($maxCount, ImportColumn $column) {
                             return max($maxCount, count($column->getExamples()));
-                        }
+                        },
+                        0
                     );
 
                     $examples = [];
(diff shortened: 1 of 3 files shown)


packages/actions/src/Concerns/CanImportRecords.php
danharrin, line 1: Please remove the comments from this file, we should restructure the code so it is more readable without comments

packages/actions/src/Concerns/CanImportRecords.php
```diff
@@ -334,18 +334,36 @@ protected function setUp(): void
...
+                    Сount the number of lines in the file.
+                    It should be equal to the largest number of examples in the columns.
+                    */
+                    $rowsCount = 0;
+                    array_map(
+                        function (ImportColumn $column) use (&$rowsCount) {
+                            $rowsCount = count($column->getExamples()) > $rowsCount ? count($column->getExamples()) : $rowsCount;
+                        },
                         $columns,
                     );
```
danharrin: Please swap this for an `array_reduce()`, where the initial count/carry is 0 and if the number of examples is greater than the carry then it gets returned instead

packages/actions/src/Concerns/CanImportRecords.php
```diff
@@ -334,18 +334,36 @@ protected function setUp(): void
...
+                    Get an array of “examples” in which the number of elements
+                    is equal to the number of lines in the file.
+                    */
+                    $examples = [];
+                    foreach ($columns as $column) {
+                        $columnExamples = $column->getExamples();
+
+                        for($i = 0; $i < $rowsCount; $i++) {
+                            $examples[$i][] = $columnExamples[$i] ?? '';
+                        }
```
danharrin: Is it possible to refactor this to a simple `foreach()` through `array_values($column->getExamples())` and using the indexes from that array?

packages/actions/src/Imports/ImportColumn.php
danharrin, line 1: Please remove the `$example` property, mark the `getExample()` method as `@deprecated Use `getExamples()` instead.` and make it return `Arr::first($this->getExamples())`. Then alter the `example()` method to just call `$this->examples($example)`.

packages/actions/src/Imports/ImportColumn.php
```diff
@@ -116,6 +121,16 @@ public function example(mixed $example): static
...
 
+    /**
+     * @param  array<mixed>  $examples
+     */
+    public function examples(array $examples): static
+    {
+        $this->examples = $examples;
+
+        return $this;
+    }
```
danharrin: This should be able to handle `mixed` instead of an array.

1) If the examples are an array or a Closure, set them directly as `$this->examples`
2) If the examples are not an array or a Closure, use `Arr::wrap()` and then set `$this->examples`

packages/actions/src/Imports/ImportColumn.php
```diff
@@ -53,6 +53,11 @@ class ImportColumn extends Component
 
     protected mixed $example = null;
 
+    /**
+     * @var array<mixed>
+     */
+    protected array $examples = [];
```
danharrin:
```suggestion
    /**
     * @var array<mixed> | Closure
     */
    protected array | Closure $examples = [];
```

packages/actions/src/Imports/ImportColumn.php
```diff
@@ -437,6 +452,14 @@ public function getExample(): mixed
         return $this->evaluate($this->example);
     }
 
+    /**
+     * @return array<mixed>
+     */
+    public function getExamples(): array
+    {
+        return array_merge([$this->evaluate($this->example)], $this->examples);
+    }
```
danharrin: First, call `$this->evaluate($this->examples)`, then `Arr::wrap()` that, then return it

danharrin (changes requested):
Finally, please add a small paragraph to the documentation demonstrating adding multiple examples to an import column

packages/actions/src/Concerns/CanImportRecords.php
Sergey-Karapetyan: I think not. Because one column may have fewer examples than another. We need to go through a `for`, check if the column contains fewer examples than the number of lines in the file, if so, add an empty value.

danharrin:
Thanks for making those changes

PR #13180, opened 2024-06-10
feat: introduce `defaultToNA()`

dasundev:
## Description
This pull request introduces a new static method to set a default cell state in the filament table builder. In some cases, certain table columns may contain null values, which can negatively impact user experience if left empty. Although you can already use `default('N/A')` to address this, I believe adding `defaultToNA()` would provide a more concise option.

```php
Tables\Columns\TextColumn::make('mobile')
    ->defaultToNA();
```

With this enhancement, the output for columns with null values will now elegantly display as "N/A".

## Functional changes

### DIFF
--- a/packages/support/src/Concerns/HasCellState.php
+++ b/packages/support/src/Concerns/HasCellState.php
@@ -60,6 +60,13 @@ trait HasCellState
         return $this;
     }
 
+    public function defaultToNA(): static
+    {
+        $this->defaultState = 'N/A';
+
+        return $this;
+    }
+
     public function isDistinctList(): bool
     {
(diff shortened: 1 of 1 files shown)


dasundev:
Once approved, I'll proceed to update the documentation accordingly before merging the pull request.

danharrin:
Hi, sorry but I am not interested in this change, it feels too niche and with not enough value. It is also very English-centric which does not work for an international package. In the future, I would hate for you to waste any more time on new features that don't get merged, so please create a discussion and @ me beforehand so we can chat about it

dasundev:
Got it. I'll make sure to create a discussion next time.

PR #13196, opened 2024-06-11
Add tooltip option to `->since()`

dododedodonl:
## Description
Often, I find myself combining `->since()` on a column with a tooltip having the exact date. With this change, it is easy to do this and specify the format.

## Visual changes

Adds a tooltip with the specified date format.

Changes
```php
                    ->since()
                    ->tooltip(static fn (TextColumn $column, ?string $state) => filled($state)
                        ? Carbon::parse($state)
                            ->setTimezone($column->getTimezone())
                            ->translatedFormat('d-m-Y H:i:s')
                        : null
                    ),
```
into
```php
                    ->since()
                    ->dateTimeTooltip()
```

## Functional changes

### DIFF
--- a/packages/tables/src/Columns/Concerns/CanFormatState.php
+++ b/packages/tables/src/Columns/Concerns/CanFormatState.php
@@ -82,7 +82,7 @@ trait CanFormatState
         return $this;
     }
 
-    public function since(?string $timezone = null): static
+    public function since(?string $timezone = null, ?string $tooltipFormat = null): static
     {
         $this->isDateTime = true;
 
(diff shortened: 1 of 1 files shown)


danharrin (changes requested):
How about `->since()->dateTooltip()` and `->dateTimeTooltip()` and `->sinceTooltip()`? Then we can also add a default format that is applied in the same way to the existing `->date()` and `->dateTime()` methods

danharrin:
Cheers

zepfietje:
@dododedodonl, care to add some docs for this new feature too? 😊

(I like this new API by the way.)

PR #13210, opened 2024-06-12
fix textarea auto resize problem

mozex:
## Description

as discussed in #13102 I stored the scroll height of textarea in an Alpine variable which fixed the issue without needing to add wire:ignore but there are 2 problems.

1. when `autosize()` and `live()` are used at the same time, there is a flicker while resizing.
2. when `autosize()` is not used but textarea is live, the problem that #13102 was fixing remains because wire:ignore is removed.

@danharrin can you take a look at these.

## Functional changes

### DIFF
--- a/packages/forms/resources/js/components/textarea.js
+++ b/packages/forms/resources/js/components/textarea.js
@@ -1,5 +1,7 @@
 export default function textareaFormComponent({ initialHeight }) {
     return {
+        height: initialHeight + 'rem',
+
         init: function () {
             this.render()
         },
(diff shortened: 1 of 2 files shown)


nielsvh2103:
Looks fine to me, but this is in autosize only, it will recreate the issue with autosize off.

mozex:
> Looks fine to me, but this is in autosize only, it will recreate the issue with autosize off.

Yes, as I mentioned in the description, this is a problem that still exists. do you have any suggestions?

nielsvh2103:
Ill make somethign real quick and make a PR also, ill let you know

nielsvh2103:
Check out https://github.com/filamentphp/filament/pull/13211

danharrin:
Thanks but replaced by 13211

PR #13211, opened 2024-06-12
Fix textarea resizing preserving heights with/without autosize

nielsvh2103:
## Description

As discussed in previous pr
https://github.com/filamentphp/filament/pull/13102

this should fix both issues including dynamic disabling/enabling, autosize on and off
if you have any suggestions for the mouseup after resizing to preserve height on autosize off, please do because couldn't get both to work without it.

extra;

i also changed SPA mode, im sure that wasnt supposed to be in autosize only either :")
same for alpine attributes, was only applied to when autosize is on

## Visual changes

Height is preserved now on autosize or normal resize

## Functional changes


packages/forms/resources/views/components/textarea.blade.php
```diff
@@ -34,21 +34,21 @@
...
-                x-on:input="render()"
-                x-on:resize.window="render()"
-                {{ $getExtraAlpineAttributeBag() }}
+                x-intersect.once="onResize()"
+                x-on:input="onResize()"
+                x-on:resize.window="onResize()"
             @endif
-            x-ignore
-            wire:ignore.style.height
+            x-on:mouseup="onResize()"
```
danharrin: Does this need to be part of the `@if` block above?

packages/forms/resources/js/components/textarea.js
```diff
@@ -1,14 +1,34 @@
...
-                this.$el.style.height = this.$el.scrollHeight + 'px'
+                this.$el.style.height = this.height;
             }
         },
+
+        onResize: function () {
+            if (this.$el.scrollHeight > 0) {
+                // Only set the height if it differs from the current height
+                const newHeight = this.$el.scrollHeight + 'px';
+                if (this.height !== newHeight) {
```
danharrin:
```suggestion
                
                if (this.height !== newHeight) {
```

packages/forms/resources/js/components/textarea.js
```diff
@@ -1,14 +1,34 @@
...
+                    this.$el.style.height = this.height;
+                }
+            }
+        },
+
+        addResizeListener: function () {
+            const observer = new ResizeObserver(() => {
+                this.height = this.$el.style.height;
+            });
+            observer.observe(this.$el);
```
danharrin:
```suggestion
            
            observer.observe(this.$el);
```

danharrin (changes requested):

packages/forms/resources/views/components/textarea.blade.php
nielsvh2103: Oh my bad, i think its actually redundant, can be removed probably, let me check real quick

mozex:
Thanks, but there is a problem: resizing only increases the height, but when you delete some text, it doesn't get resized.

nielsvh2103:
> Thanks, but there is a problem: resizing only increases the height, but when you delete some text, it doesn't get resized.

Hi, can you confirm this wasnt already the case? im sure it was , i made sure before i made the PR

mozex:
> > Thanks, but there is a problem: resizing only increases the height, but when you delete some text, it doesn't get resized.
> 
> Hi, can you confirm this wasnt already the case? im sure it was , i made sure before i made the PR

In the PR I made, it worked, but before that, I don't know.

packages/forms/resources/views/components/textarea.blade.php
nielsvh2103: Removed in new commit

nielsvh2103 (commented):
Should be done, let me check the resize on autosize removing text

mozex:
> Should be done, let me check the resize on autosize removing text

You have to set the height to the initial height on every resize and get the scrollHeight to get the reduced height, and that was causing the flicker I was talking about in my PR.
```javascript
render: function () {
    if (this.$el.scrollHeight > 0) {
        this.$el.style.height = initialHeight + 'rem'
        this.height = this.$el.scrollHeight + 'px'
        this.$el.style.height = this.height
    }
},
```

nielsvh2103:
> > Should be done, let me check the resize on autosize removing text
> 
> You have to set the height to the initial height on every resize and get the scrollHeight to get the reduced height, and that was causing the flicker I was talking about in my PR.
> 
> ```js
> render: function () {
>     if (this.$el.scrollHeight > 0) {
>         this.$el.style.height = initialHeight + 'rem'
>         this.height = this.$el.scrollHeight + 'px'
>         this.$el.style.height = this.height
>     }
> },
> ```

Yeah this will cause the flickering, setting it to 'auto' also, need to think of another approach

danharrin:
Maybe still keep the inline style attribute which gets used before Alpine kicks in

nielsvh2103:
> Maybe still keep the inline style attribute which gets used before Alpine kicks in

Doesn't seem to work directly, any ideas?

mozex:
This wouldn't solve it, the only solution is to set it to initial height and figure out a way to stop flickering on live mode

nielsvh2103:
Yeah im trying to figure out how to stop the flickering in live mode, but can't really get something in the top of my head at the moment

danharrin:
You could set the style property in JS in the same way as before as well as updating the Alpine property maybe?

nielsvh2103:
> You could set the style property in JS in the same way as before as well as updating the Alpine property maybe?

Tried, doesnt seem to work, its the thing when using live and alpine together, though im sure it didnt go back in the current stable version either

wire ignore would fix this but yeah breaks removing disabled in livewire 

by setting the height back to auto for resizing the other way around, a debounce could work possibly

I don't have any time left right now, got few appointments, if you got any suggestions or possibly a fix let me know

danharrin:
I've pulled it down, made Livewire requests, and can't see any issues or flicker, so going to go ahead with this.

Please let me know if theres anything I missed and we can open a separate PR

https://github.com/filamentphp/filament/assets/41773797/7060af51-beaa-4b3e-86dd-b23e2a99aa3c

nielsvh2103:
Ah thanks!
I see why this works and why my version wasn't working properly, re adding ->style() didn't do anything, but setting initialheight on resize and checking if heights are the same is what i forgot to do, that caused the flickering.

I tested it and works fine for both occasions (autosize, without, and dynamic disabling), thanks! @danharrin

PR #13218, opened 2024-06-12
Feature/add icons to builder header

stursby:
## Description

This PR adds a `blockIcons()` method to the Builder class.

This is very helpful for quickly visually scanning builder blocks to see what kind of block type they are.

It is disabled by default, but when enabled, will pull in the icon set in the individual blocks.

### Before
image

### After
image

## Functional changes


danharrin:
Thanks

PR #13278, opened 2024-06-16
Next wizard step test helper

eduPHP:
## Description

This pull request creates a new test helper that calls for a wizard's next step

Closes #11622

## Functional changes

Add the `nextFormWizardStep()` action to the test component, optionally you can provide the `$currentStep` (default 0) and `$formName` (default 'form')

Sample:
```php
use function Pest\Livewire\livewire;

it('moves to next wizard step', function () {
    livewire(CreatePostUsingWizard::class)
        ->nextFormWizardStep()
        ->assertHasFormErrors(['title']);
});
```


packages/forms/src/Testing/TestsForms.php
```diff
@@ -401,4 +401,20 @@ public function assertFormFieldIsVisible(): Closure
...
+    {
+        return function (int $currentStep = 0, string $formName = 'form'): static {
+            /** @phpstan-ignore-next-line  */
+            $this->assertFormExists($formName);
+
+            /** @var ComponentContainer $form */
+            $form = $this->instance()->{$formName};
+
+            $this->call('dispatchFormEvent', 'wizard::nextStep', $form->getStatePath(), $currentStep);
+            $this->refresh();
```
danharrin: Can you please explain what problem this line fixes?

packages/forms/docs/09-testing.md
```diff
@@ -286,6 +286,45 @@ test('comments section is enabled' function () {
...
+```
+
+For the subsequent steps you can pass a specific step's index to `currentStep`, which starts from 0 as the first step:
+
+```php
+use function Pest\Livewire\livewire;
+
+it('moves to the wizards second step', function () {
+    livewire(CreatePostUsingWizard::class)
+        ->nextFormWizardStep(currentStep: 1)
```
danharrin: I might be wrong, but doesn't this example put you on the 3rd step, since the currentStep is 1 which is the 2nd step?

danharrin (changes requested):

packages/forms/docs/09-testing.md
eduPHP: Right, I will update the example to have a clearer test description.
That aside, shouldn't it be better to input the desired wizard step instead of the current's index?

packages/forms/src/Testing/TestsForms.php
eduPHP: It was a test, forgot to remove it, thanks

packages/forms/docs/09-testing.md
danharrin: maybe rename it to goToStep or setStep?

eduPHP: That would be nice, I am also thinking on adding assertions for the current step `->assertWizardCurrentStep(2)` and `->assertWizardStepsCount(3)` or so, but it would be necessary to add a property to track it since it is set only on frontend, and that would allow us to also use `nextFormWizardStep` without the need to input the current step
In a project I a currently working, we track it like this:

```php

public int $currentStep = 1;

// override this to get the current step
// it would need to track the previous step too to be more accurate
public function dispatchFormEvent(...$args): void
{
    parent::dispatchFormEvent(...$args);
    // saves the current step, but only if validation is successfull
    // since in this event it starts from 0 we add +1
    // it should also be the next step, so +2
    if ('wizard::nextStep' === $args[0]) {
        $this->currentStep = $args[2] +2;
    }
}

```

eduPHP: I got some neat ideas for improving this, I am going to work on it tonight and update it up to tomorrow, please confirm if it is a good idea to add a current step property and its getter method to the form/wizard object (maybe on another pull request), if so I would increment/decrement it on the existing listener for wizard::nextStep and wizard::previousStep

mokhosh: I was actually surprised that you can't access the current step.

eduPHP:
Updated with some more helpers and a fix/improvement to the wizard component.

I hope I didn't go too far, please let me know if it requires more documentation or if it would be better to split this in multiple PRs

danharrin:
To be honest I don't understand the new property and step assertion, it doesn't get persisted between requests so I am unsure of the use cases of `goToStep(2)->assertStep(2)`, it just feels like it would never catch a bug in a user's codebase, if you wanted to assert that a validation error happened when switching steps then we do have an assertion for that

eduPHP:
> To be honest I don't understand the new property and step assertion, it doesn't get persisted between requests so I am unsure of the use cases of `goToStep(2)->assertStep(2)`, it just feels like it would never catch a bug in a user's codebase, if you wanted to assert that a validation error happened when switching steps then we do have an assertion for that

I would use the it more like fillForm([...])->goToNextStep()->assertNoErrors()->assertStep(2)

The currentIndex property I think is just a missing piece, my use case was to store the current form progress to continue later, and I needed to hack around to get an accurate current wizard step

danharrin:
PHPStan is failing, please fix that

mokhosh:
> To be honest I don't understand the new property and step assertion, it doesn't get persisted between requests so I am unsure of the use cases of `goToStep(2)->assertStep(2)`, it just feels like it would never catch a bug in a user's codebase, if you wanted to assert that a validation error happened when switching steps then we do have an assertion for that

I have a use case where I update some information outside of the form after each step of a wizard, and I need to assert that the info gets updated when I go to the next step, so this would make a lot of sense.

eduPHP:
> PHPStan is failing, please fix that

Fixed ✅

packages/forms/resources/views/components/wizard.blade.php
```diff
@@ -29,6 +29,12 @@
                 return
             }
 
+            $wire.dispatchFormEvent(
+                'wizard::previousStep',
+                '{{ $statePath }}',
+                this.getStepIndex(this.step),
+            )
```
danharrin: This code is not actually run during tests, and doesn't do anything when you're using the wizard in real life, it just creates a network request with no purpose. I think it can be removed?

packages/forms/src/Components/Wizard.php
```diff
@@ -58,12 +60,27 @@ protected function setUp(): void
     {
         parent::setUp();
 
+        $this->currentStepIndex = $this->getStartStep() - 1;
```
danharrin: Please utilize the setter method here instead of setting the property directly

packages/forms/src/Components/Wizard.php
```diff
@@ -246,4 +265,16 @@ public function isStepPersistedInQueryString(): bool
     {
         return filled($this->getStepQueryStringKey());
     }
+
+    private function setCurrentStepIndex(int $currentStepIndex): static
```
danharrin: Please make this function protected so it is available to extended wizard classes. Please drop the "set" from the name so it is consistent with other setter methods. Please rename the parameter to just `$index`.

packages/forms/src/Testing/TestsForms.php
```diff
@@ -401,4 +402,111 @@ public function assertFormFieldIsVisible(): Closure
...
+            /** @phpstan-ignore-next-line  */
+            $this->assertFormExists($formName);
+
+            /** @var ComponentContainer $form */
+            $form = $this->instance()->{$formName};
+
+            /**
+             * @var Wizard $wizard
+             */
+            $wizard = $form->getComponent(fn ($component) => $component instanceof Wizard);
```
danharrin: Please add a parameter type and return type to the function

packages/forms/src/Testing/TestsForms.php
```diff
@@ -401,4 +402,111 @@ public function assertFormFieldIsVisible(): Closure
...
+            Assert::assertTrue(count($wizard->getChildComponents()) >= $step, "Wizard does not have a step $step");
```
danharrin: Could this be an assertion to check if an array key exists in the child components instead?

packages/forms/src/Testing/TestsForms.php
```diff
@@ -401,4 +402,111 @@ public function assertFormFieldIsVisible(): Closure
...
+            /** @phpstan-ignore-next-line  */
+            $this->assertFormExists($formName);
+
+            /** @var ComponentContainer $form */
+            $form = $this->instance()->{$formName};
+
+            /**
+             * @var Wizard $wizard
+             */
+            $wizard = $form->getComponent(fn ($component) => $component instanceof Wizard);
```
danharrin: Please add a parameter type and return type to the function

packages/forms/src/Testing/TestsForms.php
danharrin: Please wrap the variable interpolation in `{}`, so `{$step}`

packages/forms/src/Testing/TestsForms.php
```diff
@@ -401,4 +402,111 @@ public function assertFormFieldIsVisible(): Closure
...
+            Assert::assertEquals(
+                $step,
+                $current = $wizard->getCurrentStepIndex() + 1,
+                "Failed asserting that wizard is on step $step, current step is {$current}"
```
danharrin: Please wrap the variable interpolation in `{}`, so `{$step}`

packages/forms/src/Testing/TestsForms.php
```diff
@@ -401,4 +402,111 @@ public function assertFormFieldIsVisible(): Closure
...
+            /** @phpstan-ignore-next-line  */
+            $this->assertWizardStepExists($desiredStep, $formName);
+
+            /** @var ComponentContainer $form */
+            $form = $this->instance()->{$formName};
+
+            /**
+             * @var Wizard $wizardComponent
+             */
+            $wizardComponent = $form->getComponent(fn ($component) => $component instanceof Wizard);
```
danharrin: Please add a parameter type and return type to the function

packages/forms/src/Testing/TestsForms.php
```diff
@@ -401,4 +402,111 @@ public function assertFormFieldIsVisible(): Closure
...
+             * @var Wizard $wizard
+             */
+            $wizard = $form->getComponent(fn ($component) => $component instanceof Wizard);
+            Assert::assertTrue(count($wizard->getChildComponents()) >= $step, "Wizard does not have a step $step");
+
+            return $this;
+        };
+    }
+
+    public function assertWizardCurrentStepIs(): Closure
```
danharrin: I think we should drop the "is" from the end of this method name

packages/forms/src/Testing/TestsForms.php
```diff
@@ -401,4 +402,111 @@ public function assertFormFieldIsVisible(): Closure
...
+                "Failed asserting that wizard is on step $step, current step is {$current}"
+            );
+
+            return $this;
+        };
+    }
+
+    public function goToWizardStep(): Closure
+    {
+        return function (int $desiredStep = 1, string $formName = 'form'): static {
```
danharrin: I think the parameter should just be `$step` and not `$desiredStep`

packages/forms/src/Testing/TestsForms.php
```diff
@@ -401,4 +402,111 @@ public function assertFormFieldIsVisible(): Closure
...
+
+            /** @var ComponentContainer $form */
+            $form = $this->instance()->{$formName};
+
+            /**
+             * @var Wizard $wizardComponent
+             */
+            $wizardComponent = $form->getComponent(fn ($component) => $component instanceof Wizard);
+            if ($desiredStep === -1) {
+                $desiredStep = count($wizardComponent->getChildComponents()) + 1;
```
danharrin: Can you explain what `-1` signifies please? Is it the last step or something?

packages/forms/src/Testing/TestsForms.php
```diff
@@ -401,4 +402,111 @@ public function assertFormFieldIsVisible(): Closure
...
+            }
+
+            $stepIndex = $desiredStep <= 1 ? 0 : $desiredStep - 2;
```
danharrin: Please explain what is happening here

packages/forms/src/Testing/TestsForms.php
```diff
@@ -401,4 +402,111 @@ public function assertFormFieldIsVisible(): Closure
...
+            /** @phpstan-ignore-next-line  */
+            $this->assertFormExists($formName);
+
+            /** @var ComponentContainer $form */
+            $form = $this->instance()->{$formName};
+
+            /**
+             * @var Wizard $wizardComponent
+             */
+            $wizardComponent = $form->getComponent(fn ($component) => $component instanceof Wizard);
```
danharrin: Please add a parameter type and return type to the function

packages/forms/src/Testing/TestsForms.php
```diff
@@ -401,4 +402,111 @@ public function assertFormFieldIsVisible(): Closure
...
```
danharrin: Please add a parameter type and return type to the function

packages/forms/src/Testing/TestsForms.php
```diff
@@ -401,4 +402,111 @@ public function assertFormFieldIsVisible(): Closure
...
+        return function (string $formName = 'form'): static {
+            /** @phpstan-ignore-next-line  */
+            $this->assertFormExists($formName);
+
+            /** @var ComponentContainer $form */
+            $form = $this->instance()->{$formName};
+
+            /**
+             * @var Wizard $wizardComponent
+             */
```
danharrin: Please condense this down to one line, like the other `@var` comments for consistency

packages/forms/src/Testing/TestsForms.php
```diff
@@ -401,4 +402,111 @@ public function assertFormFieldIsVisible(): Closure
...
```
danharrin: Please condense this down to one line, like the other `@var` comments for consistency

packages/forms/src/Testing/TestsForms.php
```diff
@@ -401,4 +402,111 @@ public function assertFormFieldIsVisible(): Closure
...
+        return function (int $desiredStep = 1, string $formName = 'form'): static {
+            /** @phpstan-ignore-next-line  */
+            $this->assertWizardStepExists($desiredStep, $formName);
+
+            /** @var ComponentContainer $form */
+            $form = $this->instance()->{$formName};
+
+            /**
+             * @var Wizard $wizardComponent
+             */
```
danharrin: Please condense this down to one line, like the other `@var` comments for consistency

packages/forms/src/Testing/TestsForms.php
```diff
@@ -401,4 +402,111 @@ public function assertFormFieldIsVisible(): Closure
...
+        return function (int $step, string $formName = 'form'): static {
+            /** @phpstan-ignore-next-line  */
+            $this->assertFormExists($formName);
+
+            /** @var ComponentContainer $form */
+            $form = $this->instance()->{$formName};
+
+            /**
+             * @var Wizard $wizard
+             */
```
danharrin: Please condense this down to one line, like the other `@var` comments for consistency

packages/forms/src/Testing/TestsForms.php
```diff
@@ -401,4 +402,111 @@ public function assertFormFieldIsVisible(): Closure
...
```
danharrin: Please condense this down to one line, like the other `@var` comments for consistency

danharrin (changes requested):

packages/forms/resources/views/components/wizard.blade.php
eduPHP: The purpose of this event is to update the current step index on the backend. I will add a test this week when I have the time

eduPHP:
I will work on the changes probably on the weekend, thanks for taking the time to go through the code :green_heart:

packages/forms/resources/views/components/wizard.blade.php
danharrin: The current step on the backend doesn't get persisted as form components are not Livewire components, so the event is not required outside of testing IMO

danharrin:
Hi, not to put on too much pressure here, but we've decided that from Monday we will no longer be merging in PRs for new features for v3. This will free up some time we need to work on v4. Bug PRs will continue to be merged. If you can get this fixed before then, that would be great. If not, thats absolutely fine, but I will close this and if you make a new PR it will need to target v4. If you have any questions let me know. Thanks!

packages/forms/resources/views/components/wizard.blade.php
eduPHP: Removed

packages/forms/src/Components/Wizard.php
eduPHP: Updated

packages/forms/src/Testing/TestsForms.php
eduPHP: It takes you to the last form step, just wanted to cover the option since I was on it already, doesn't seem so useful now, I will remove it and maybe submit again in the future if I feel it is a missing functionality

packages/forms/src/Testing/TestsForms.php
eduPHP: The helper expect the step you want to go to, so if you are on step 1, you input step 2
The event expects the current step you are on, not the destination, and it starts from 0
So, -1 for the correct index, - 2 for the current step

eduPHP:
> Hi, not to put on too much pressure here, but we've decided that from Monday we will no longer be merging in PRs for new features for v3. This will free up some time we need to work on v4. Bug PRs will continue to be merged. If you can get this fixed before then, that would be great. If not, thats absolutely fine, but I will close this and if you make a new PR it will need to target v4. If you have any questions let me know. Thanks!

Updated, hopefully I made it in time.

Sorry about the delay, weekends are also being so busy lately since we are traveling visiting some relatives

mokhosh:
Looks good. 

I have a question and an idea. 

Question: do the test helpers support multiple wizards in a form? Doesn't look like it. 

Idea: ability to go to steps based on their names. 

Neither of these should block merging imo tho.

eduPHP:
> Looks good.
> 
> I have a question and an idea.
> 
> Question: do the test helpers support multiple wizards in a form? Doesn't look like it.
> 
> Idea: ability to go to steps based on their names.
> 
> Neither of these should block merging imo tho.

I didn't think it was possible to have multiple wizards in a single form 🤯 I will think in a way to handle it, maybe accept a third option with an index or identifier

I also thought of suporting a named step, but decided to leave it for a future merge request to keep it simple on this one

danharrin:
Thank you for your continued work and revisions on this PR

mokhosh:
@danharrin @eduPHP What do you think about making `currentStepIndex` `public` so it's persisted between requests and can be used as a useful piece of information?

danharrin:
Form components are Livewire components, its not persisted.

mokhosh:
This doesn't work with wizards inside an action does it? 🧐

mokhosh:
It works 🥳

```php
->goToNextWizardStep('mountedActionForm')
```

PR #13319, opened 2024-06-19
Fix for possible Path Traversal vulnerability.

FeatureCreepKing:
## Description
Unsanitized input from the HTTP request body flows into unlink, where it is used as a path. This may result in a Path Traversal vulnerability and allow an attacker to delete arbitrary files.

## Functional changes

- [Y] Code style has been fixed by running the `composer cs` command.
- [Y] Changes have been tested to not break existing functionality.
- [Y] Documentation is up-to-date.

### DIFF
--- a/packages/support/src/Commands/Concerns/CanManipulateFiles.php
+++ b/packages/support/src/Commands/Concerns/CanManipulateFiles.php
@@ -15,11 +15,12 @@ trait CanManipulateFiles
     protected function checkForCollision(array $paths): bool
     {
         foreach ($paths as $path) {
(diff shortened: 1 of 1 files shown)


danharrin:
Excuse my bluntness please. How on earth would data from the HTTP request body ever enter a trait that is only used by Artisan commands? If this is possible out of the box with Filament without you having to wire up a controller to a command, please email me as soon as possible as this is a security vulnerability that should be addressed in private, as per our [security policy](https://github.com/filamentphp/filament/security/policy).

PR #13320, opened 2024-06-19
Conditionally hide summaries

D-Jordy:
## Description
Adds a `CanBeHidden` trait to the `Summarizer` so they can be conditionally hidden in the same way a `Column` can

## Visual changes
none

## Functional changes


zepfietje (changes requested):
Are you sure you have run `composer cs`?
Seems like the Blade views aren't properly formatted.

Also, please document this feature as per the pull request template.

D-Jordy:
done. my bad, prettier did not run.

danharrin:
Visibility checks should be moved out of the view and into the getter method that returns the summaries for a column, you can return the filtered array

D-Jordy:
done ^^

PR #13321, opened 2024-06-19
Don't lazy-load the GlobalSearch component

jackwh:
## Description

- The `Filament\Livewire\GlobalSearch` component is lazy-loaded, despite it being a very simple component with effectively no overhead on initial load.
- Because it's lazy-loaded it triggers a pointless network request after the initial page load
- It also causes the component to "jump" in after a delay (like a FOUC)

I've simply removed the `lazy` flag for better user experience. I don't see this being a breaking change.

## Visual changes

No visual changes, except the search component is now present on the page immediately without unnecessary network requests:

Screenshot 2024-06-19 at 15 10 49@2x

## Functional changes


zepfietje:
I'm not aware of any particular reason why it's currently lazy-loaded.
So I'm fine with this change, but requesting a final review from @danharrin.

PR #13323, opened 2024-06-19
Add page end/start render hooks to simple page

ralphjsmit:
This PR adds the `page.start` and `page.end` render hooks to the simple page component. Alternatively, if one would like to keep these render hooks for "actual" panel pages, we could introduce a `simple-page.start` and `simple-page.end` hook or sth.

Thanks!


danharrin:
Please introduce the new hook since you might need different hook content to deal with the different layouts and its easy to register the same view to multiple hooks

ralphjsmit:
Done 👍

PR #13328, opened 2024-06-19
Broadcasting using the Panel builder - configuration & documentation

brentkelly:
## Description

I've been playing around with getting broadcasting going under the Panel Builder, however have found a few gaps in the documentation which made things a bit hard to follow. The existing documentation in the Notification section is targeted at standalone installations, and there is no mention of it in the Panel Builder documentation.

I managed to piece it together with a bunch of searching around along with some trial & error, but I think this is worth documenting.

I was also initially tripped up by the default commented configuration in `config/filament.php` containing an invalid route for the auth endpoint. The [official docs](https://laravel.com/docs/11.x/broadcasting#authorizing-channels) list this as `/broadcasting/auth` - not `/api/v1/broadcasting/auth` which returns a 404 if you attempt to use it with a default broadcasting installation.

This PR addresses these 2 areas.

## Visual changes

No visual changes aside from the additional documentation to `/docs/panels/notifications`.

## Functional changes

### DIFF
--- a/packages/panels/docs/07-notifications.md
+++ b/packages/panels/docs/07-notifications.md
@@ -32,3 +32,29 @@ public function panel(Panel $panel): Panel
         ->databaseNotificationsPolling('30s');
 }
 ```
+
+## Broadcast notifications
+
+The Panel Builder comes with a level of inbuilt support for real-time broadcast notifications. However there are a number of areas you will need to install and configure to wire everything up and get it working.
+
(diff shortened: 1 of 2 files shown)


danharrin:
Thanks. I have removed the assets step, since the Echo script is always published on install whether or not it is configured, and there is no Vite JavaScript file within your app that gets included in a panel out of the box.

PR #13349, opened 2024-06-21
Cache getFlatComponents

matsesh:
### Summary

This PR introduces caching for the `getFlatComponents` method to enhance performance by avoiding redundant computations. This change is particularly beneficial for scenarios involving large sets of components, where the flattening process can be computationally expensive.

### Changes

- **Caching**: Added a caching mechanism to store the results of the `getFlatComponents` method based on the visibility of components (`withHidden` flag).

### Detailed Changes

1. **Caching Implementation**:
   - Introduced a new property `$cachedFlatComponents` to store cached results.
   - Modified `getFlatComponents` to check the cache before computing the flat list of components.
   - Ensured that the cache key reflects the state of the `withHidden` flag to maintain correct behavior for both visible and hidden components.

### Testing

- Verified that the caching mechanism works as expected and does not alter the existing functionality.
- Ensured that the output of `getFlatComponents` is consistent with the previous implementation.
- Tested with both `withHidden` set to true and false to confirm correct behavior in both scenarios.

### Comparison

The main advantage in speed is visible when we work with repeaters that use livewire. Without this cache, it is difficult to work with repeaters with a large number of components.

#### Before
Without Cache

#### After
With Cache


danharrin:
This is a very risky change, any request that calls the method and then adds new components to the state such as in a repeater, builder or another dynamic schema could be affected. I will need time to test this to ensure there aren't any knock-on effects.

matsesh:
[danharrin](https://github.com/danharrin), sure, thanks, will do double-check from my end too.

UPD: From my end, no issues were found with the repeaters.

danharrin:
I think this is a useful change, but I am going to merge it into v4 as I am not 100% confident this will not cause issues in some cases. Thank you.

PR #13358, opened 2024-06-22
Fix using HtmlString for HasIcon

pxlrbt:
## Description

Tried to fix this issue on Discord: 
https://discord.com/channels/883083792112300104/1254017792902959124

According to the docs you should be able to pass a view to register an icon alias. This gets casted to a string though and is then outputted as an image.

Also `view()` doesn't work. I think the issue is that `View` gets converted to string when passed as a Blade component param. Not sure how to fix this. I think the issue is how we pass `$icon` in `actions/resources/views/components/action.blade.php`.

CC: @zepfietje not sure how this was intended to work. Maybe you can have a look at this.

## Functional changes

### DIFF
--- a/packages/support/docs/03-icons.md
+++ b/packages/support/docs/03-icons.md
@@ -33,7 +33,7 @@ use Filament\Support\Facades\FilamentIcon;
 
 FilamentIcon::register([
     'panels::topbar.global-search.field' => 'fas-magnifying-glass',
-    'panels::sidebar.group.collapse-button' => view('icons.chevron-up'),
+    'panels::sidebar.group.collapse-button' => new HtmlString(view('icons.chevron-up')->render()),
 ]);
 ```
 
--- a/packages/support/src/Concerns/HasIcon.php
+++ b/packages/support/src/Concerns/HasIcon.php
@@ -5,16 +5,17 @@ namespace Filament\Support\Concerns;
 use Closure;
 use Filament\Support\Enums\IconPosition;
 use Filament\Support\Enums\IconSize;
+use Illuminate\Support\HtmlString;
 
 trait HasIcon
 {
-    protected string | Closure | null $icon = null;
+    protected string | HtmlString | Closure | null $icon = null;
 
     protected IconPosition | string | Closure | null $iconPosition = null;
 
     protected IconSize | string | Closure | null $iconSize = null;
 
-    public function icon(string | Closure | null $icon): static
+    public function icon(string | HtmlString | Closure | null $icon): static
     {
         $this->icon = $icon;
 
@@ -35,7 +36,7 @@ trait HasIcon
         return $this;
     }
 
-    public function getIcon(): ?string
+    public function getIcon(): string | HtmlString | null
     {
         return $this->evaluate($this->icon);
     }


ju5t:
`view()` gets converted to html because Laravel calls `sanitizeComponentAttribute` in [CompilesComponents](https://github.com/laravel/framework/blob/11.x/src/Illuminate/View/Compilers/Concerns/CompilesComponents.php#L211). A view contains `__toString()`. It then runs `e($view)` which finds out its `Htmlable` and renders the view.

I don't see why we couldn't pass an Htmlable but can pass an HtmlString so this feels like a bug in Laravel.

ju5t:
@pxlrbt

> I think the issue is that View gets converted to string when passed as a Blade component param

It only happens when passing `Renderables`'s. `Htmlable`'s are fine. I've raised an issue with the Laravel team. It feels like a bug, but perhaps there's a good reason for it.

As discussed -- this may be a breaking change and not the solution short term.

See: https://github.com/laravel/framework/issues/51889

danharrin:
Please swap it to use Htmlable and get rid of the HtmlString wrapper. If we need to do extra processing into an HtmlString internally once its passed in then we will do that, but this PR can still be updated as such for now until we have a response from Laravel.

packages/support/docs/03-icons.md
```diff
@@ -33,7 +33,7 @@ use Filament\Support\Facades\FilamentIcon;
 
 FilamentIcon::register([
     'panels::topbar.global-search.field' => 'fas-magnifying-glass',
-    'panels::sidebar.group.collapse-button' => view('icons.chevron-up'),
+    'panels::sidebar.group.collapse-button' => new HtmlString(view('icons.chevron-up')->render()),
```
zepfietje, line 36: Shouldn't passing a view just work? Maybe it needs to be `fn () => view()` instead?

zepfietje (changes requested):

packages/support/docs/03-icons.md
pxlrbt: Not with the current solution. A view gets rendered as a string when passed as an attribute and isn't of type `Htmlable` anymore. I might try the slot solution.

ju5t: Just to clarify and to have my comment on Discord here too.

A view gets rendered to a string because it implements the magic method `__toString()` and `HasIcon` did not accept `Htmlable`. PHP converts it to a string to match the `string` type on properties and methods.

The reason it doesn't work when you also accept `Htmlable` is because views also implement `Renderable`. All data passed to views an its derivatives (e.g. Blade components) will get processed by [`gatherData`](https://github.com/laravel/framework/blob/11.x/src/Illuminate/View/View.php#L221) and rendered to HTML.

As a solution/workaround I have added the `Htmlable` type to the `HasIcon` trait and included the following in `getIcon` to make sure we're not passing a Renderable to views.

```php
$icon = $this->icon instanceof Renderable
    ? new HtmlString($this->icon->render())
    : $this->icon;
```

`HtmlString` is the simplest solution in this case, but type hints should still be `Htmlable` as it's the only implementation you need for HTML content in Laravel.

ju5t:
Is there anything I can do to help with this PR?

danharrin:
I would welcome a PR to fix this, but this isn't the right way to go about it IMO. You should be able to pass in a View object as per the docs without having to wrap within an HtmlString.

PR #13377, opened 2024-06-24
feature: Custom mono and serif fonts

### LINKED ISSUE #13294: Load multiple fonts
Currently you can only configure a single font to be loaded in the panel configuration.
However, some of Filament's components use the `mono` font family.
Adding support for loading a font into a specific Tailwind CSS family would allow you to configure additional fonts like `mono` from the panel configuration instead of having to resort to render hooks.

danharrin:
Closes #13294


packages/panels/tailwind.config.preset.js
```diff
@@ -2,7 +2,9 @@ import defaultPreset from '../support/tailwind.config.preset.js'
 import defaultTheme from 'tailwindcss/defaultTheme'
 
 defaultPreset.theme.extend.fontFamily = {
+    mono: ['var(--mono-font-family)', ...defaultTheme.fontFamily.mono],
     sans: ['var(--font-family)', ...defaultTheme.fontFamily.sans],
```
zepfietje, line 6: @danharrin, shall we rename the regular font to `sans-font-family` here just like the others?

danharrin: no because I feel like it should feel clearly like the default

zepfietje: Sure, fine by me

Issue #13252, opened 2024-06-14
Extract empty state component


zepfietje:

zepfietje:
Can be done in v3 by adding the original `fi-*` CSS classes using [Blade slot attributes](https://laravel.com/docs/11.x/blade#slots).

danrichards:
I would love to see an empty state pattern in lists and relation managers. It's a great way to tell the user to create their first record or just explain how to start using something.

danharrin:
This request has been sitting inactive for a while, and it isn't something that I am going to PR to implement myself, so I am going to remove it from the roadmap. However, I would still accept a PR for this if it were ever created. Many thanks!

danharrin:
I implemented it in #17932 as I needed it for something, although the table empty state is not using this new component yet

PR #13408, opened 2024-06-26
Add empty state to chart widget

juliangums:
@danharrin as discussed in https://github.com/filamentphp/filament/discussions/13311 I have added a feature to display an empty state on chart widgets. Before this is merged, I'd like to ask a couple of things as I am not entirely familiar with everything in Filament:
1) I added a string and translations for `en` and `de`. Is that enough? How are all the other translations handled for so many languages?
2) I didn’t use screenshots or links in the docs as I didn’t know how to do them.
3) The implementation is very much based on the work done on tables. But as we aren't defining these things by calling methods on an object but rather by defining a class that extends ChartWidget, I removed all the setters from it. But I guess that also means closures don't really make sense for the attributes or do they? I want to make sure this is in the spirit of Filament so any guidance would be much appreciated.
4) Is everything else as you'd expect? Anything I missed?

Let me know and I'll get the rest sorted.
Otherwise, it's ready to be tested.

## Visual changes
Screenshot 2024-06-26 at 23 22 49

## Functional changes

### DIFF
--- a/packages/panels/resources/lang/de/widgets/empty-state.php
+++ b/packages/panels/resources/lang/de/widgets/empty-state.php
@@ -1,0 +1,7 @@
+<?php
+
+return [
+
+    'heading' => 'Keine Daten vorhanden',
+
+];
--- a/packages/panels/resources/lang/en/widgets/empty-state.php
+++ b/packages/panels/resources/lang/en/widgets/empty-state.php
@@ -1,0 +1,7 @@
+<?php
+
+return [
+
+    'heading' => 'Nothing to display',
+
+];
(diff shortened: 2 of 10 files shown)


zepfietje (changes requested):
Thanks for this PR, @juliangums!

I don't like the duplication of empty state Blade views.
Since I was already planning to extract the empty state component from tables, shall we get https://github.com/filamentphp/filament/issues/13252 done before merging your new feature?

juliangums:
@zepfietje that would make a ton of sense. Let me know when that’s done or if you need any help just instruct me.

zepfietje:
Great! It's at the top of my list, so will get to it as soon as possible.

I'll mark this PR as draft until we can move forward. 👍

zepfietje:
Started working on https://github.com/filamentphp/filament/pull/13642.

danharrin:
Going to close this since there was no movement on #13642. Next step would be putting in a v4 PR for an empty state Blade component, and then opening another PR to add that to the chart widget.

lambda-bajara:
@juliangums  @zepfietje So this was not deployed? Is it only going to be launched in v4?
Just to make sure... I'm needing this 😅

PR #13415, opened 2024-06-27
Allow up to 5 stats widgets per row

milenmk:
## Description

You cannot have more than 3 stats widgets per row as this is hardcoded in the view file `stats-overview-widget.blade.php`.

With this PR, you can have up to 5 by default. If a user wants more than 5, it is also possible. However, a custom CSS might have to be added.

### Before
screen_before

### After
screen_after

## Functional changes


zepfietje (changes requested):
Thanks for the PR, @milenmk!

I agree this should support any number of columns.
Could you make this dynamic using the existing grid component?

milenmk:
> Could you make this dynamic using the existing grid component?

Could you elaborate a bit more please?

zepfietje:
Check out the widgets blade component to see how the grid Blade component is used to allow for any number of columns instead of hard-coding the possible values.

zepfietje:
Let me know when this is ready to be reviewed, @milenmk :)

milenmk:
Hi. I think this should do it.

1. We have 3 widgets per row by default
2. No screen breakpoints are hardcoded i.e. the use of `getColumns()` method is optional
3. The number of columns for each screen size is null by default i.e. if `getColumns()` is not used there will be 3 widgets per row for each screen size 
4. As I've stated in the addition to the docs that I've made, the end user can alter the number of widgets per row for any breakpoint, be it only for one or 2-3 of them.

danharrin:
Hi, this is not the correct approach unfortunately. You are concatenating values from PHP into Tailwind classes, which will break Tailwind since it is not able to scan for the classes to compile. Please follow what @zepfietje and open a PR using the `grid` component. In addition, users with 5 cards currently should not see a change in the display of the widget until they override the columns configuration, otherwise that would be a breaking change.

PR #13420, opened 2024-06-27
Text Column Badge Tooltip

Skullbock:
## Description

This PR introduces the ability to add tooltips to text column badges, and supports also list of badges. 
Since this PR introduces new methods related to badges, it also creates a dedicated `HasBadge` trait, to match the rest of the codebase.

## Visual changes
Before

Screenshot 2024-06-27 alle 17 32 17

After
Screenshot 2024-06-27 alle 17 28 56

## Functional changes

### DIFF
--- a/packages/tables/resources/views/columns/text-column.blade.php
+++ b/packages/tables/resources/views/columns/text-column.blade.php
@@ -178,10 +178,15 @@
                         ])
                     >
                         @if ($isBadge)
+                            @php
+                                $badgeTooltip = $getBadgeTooltip($state);
+                            @endphp
+
                             <x-filament::badge
                                 :color="$color"
                                 :icon="$icon"
                                 :icon-position="$iconPosition"
+                                :tooltip="$badgeTooltip"
                             >
                                 {{ $formattedState }}
                             </x-filament::badge>
--- a/packages/tables/resources/views/columns/text-column.blade.php
+++ b/packages/tables/resources/views/columns/text-column.blade.php
@@ -178,15 +178,11 @@
                         ])
                     >
                         @if ($isBadge)
-                            @php
-                                $badgeTooltip = $getBadgeTooltip($state);
-                            @endphp
-
                             <x-filament::badge
                                 :color="$color"
                                 :icon="$icon"
                                 :icon-position="$iconPosition"
-                                :tooltip="$badgeTooltip"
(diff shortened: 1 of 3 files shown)


packages/tables/resources/views/columns/text-column.blade.php
```diff
@@ -178,10 +178,15 @@
                         ])
                     >
                         @if ($isBadge)
+                            @php
+                                $badgeTooltip = $getBadgeTooltip($state);
```
zepfietje: No need to store in a variable if it's only used once.

zepfietje (changes requested):
Can't we just use the existing tooltip method instead?

Skullbock:
That was my initial thought as well, but isn't it a breaking change since I think it should require a method signature change as week, since getTooltip would need to receive the state?

zepfietje:
Fair enough.
Let's see what @danharrin thinks?

packages/tables/resources/views/columns/text-column.blade.php
Skullbock: Fixed!

danharrin:
Personally, I don't think a dedicated "badge tooltips" feature is the right approach here. Text columns with multiple values do not always use badges, they can use a concatenated list of values, or a list with new lines for each item. These should also have consistent tooltip support?

Here are some plans of action:
- Support individual tooltips for each state item for all configurations of the text column: badges, list items, or concatenated list items
- Or, fix the support for badge tooltips so that all badges show the same tooltip when hovered over, so its consistent. This is less work
- Please also consider the "text entry" component of infolists, which needs consistent tooltip support with text columns

Since both of these plans require a complete rewrite, I am going to close this PR, but I am open to others as above.

Skullbock:
Thanks @danharrin, makes sense to me

Can you clarify: "fix the support for badge tooltips so that all badges show the same tooltip when hovered over, so its consistent"?

What would that look like? As i said before, to make the badge tooltip content customizable for each repeatable value requires a method signature change, that is a breaking change in itself, or am i missing something?

Will gladly PR something like it if you can provide the "direction"
Thanks

danharrin:
At the moment, you've added a feature so that all badges can show different tooltips. The way to make this consistent with non-badge columns is to make it so that all badges show the same tooltip. That is a bug fix at its most simplest.

PR #13422, opened 2024-06-27
Add Dynamic Body Classes Based on URL Path

tnylea:
## Description

I need a way to add styles to different pages that I create in Filament. I didn't see a way to target a CSS selector that was unique for the Page or the URL that I was on.

I someone visits a page at `/admin/posts`. The body will have a class that is:

```
.fi-url-admin-posts
```

This will allow for better refined styles for different pages of the admin panel builder.

Is there another way that I should implement this, if there is let me know, but I thought this was a simple way to add classes to a page that can be uniquely identified by a CSS selector.

## Visual changes

No visual changes, but here is an example of a page with the custom class applied to it.

CleanShot 2024-06-27 at 18 30 08@2x

## Functional changes

If you let me know that this is alright, I will also add a section to the documentation to explain how to utilize this dynamic class 😁

Thanks guys! Keep up the awesome work.


zepfietje (changes requested):
Maybe we should add the CSS class to the page component instead of the layout? Also I think it should be explicit instead of adding classes to all pages while only actually needed in some cases.

nwconsulting:
Hi, can I suggest that we also add the page id from the database as a class e.g. "fi-page-id-??" where the "??" is the numeric id. This will ensure that when doing style changes it wont matter if the URL changes the CSS references will stay the same.

danharrin:
I think this belongs on the body since it can then be used to target classes outside of the page in the layout, without needing to rely on `:has`

danharrin:
I am concerned this doesn't work in SPA mode since the layout remains the same, have you tested that @tnylea ?

zepfietje:
I'm wondering if we should use [HTML data attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) instead of CSS classes here.

nwconsulting:
> I'm wondering if we should use [HTML data attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) instead of CSS classes here.

That would get my vote (for what it is worth!) because it would also then allow developers to add other data themselves and reference it via JS and CSS on the pages.

danharrin:
Hi, not to put on too much pressure here, but we've decided that from Monday we will no longer be merging in PRs for new features for v3. This will free up some time we need to work on v4. Bug PRs will continue to be merged. If you can get this fixed before then, that would be great. If not, thats absolutely fine, but I will close this and if you make a new PR it will need to target v4. If you have any questions let me know. Thanks!

danharrin:
Hi, unfortunately closing as above, but open to it being reopened and targeted at v4

PR #13427, opened 2024-06-28
Add current state to the error messages on column test assertions

gehrisandro:
## Description

This PR improves the assertion error message for column state tests by adding the current value.

Here is an example of the changed message:

_Failed asserting that a table column with name [title] has value of [My title] for record [1] on the [App\Models\Post] component. **Current value: [Other Title]**_

## Functional changes


danharrin:
I haven't tried this, but would switching from `assertTrue` to `assertSame` give the user more information here?

gehrisandro:
I will give it a try.

gehrisandro:
@danharrin `assertSame` works fine, but I went with `assertEquals` because before it used lazy comparison `==`.

danharrin:
Thanks!

PR #13445, opened 2024-07-01
Fix footer actions on content grid not aligning to the base of the container

rossbearman:
When a table is displayed as a grid using `contentGrid()`, if the heights of each row box differs, the footer actions will stick to the bottom of the content, rather than sitting flush with the bottom of the box.

A couple of people have encountered and reported this, and this fix has worked in those cases, but **this PR has not been thoroughly tested**, though it is a small change.

## Visual Difference

Before: 
image

After:
image

## Functional changes

- [ ] Conditionally switch to using `flex h-full items-stretch` on content grid rows in place of `flex items-center`

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -542,8 +542,9 @@
 
                                 <div
                                     @class([
-                                        'flex items-center' => ! $contentGrid,
-                                        'flex h-full items-stretch' => $contentGrid,
+                                        'flex',
+                                        'items-center' => ! $contentGrid,
+                                        'h-full items-stretch' => $contentGrid,
                                         'ps-1 sm:ps-3' => (! $contentGrid) && $hasItemBeforeRecordContent,
                                         'pe-1 sm:pe-3' => (! $contentGrid) && $hasItemAfterRecordContent,
                                         'ps-1' => $contentGrid && $hasItemBeforeRecordContent,
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -542,7 +542,8 @@
 
                                 <div
                                     @class([
-                                        'flex items-center',
+                                        'flex items-center' => ! $contentGrid,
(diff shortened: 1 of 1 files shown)


zepfietje:
I'm going to hold off on merging this PR, because I'm not sure if having the actions aligned at the bottom is desired for every use case.

rossbearman:
Would it make sense to have this available as an option on the content grid instead?

zepfietje:
Ideally there'd just be an actions component you could use to create the exact layout you want. Not sure if we're open to an API to configure the current actions position. What do you think, @danharrin?

danharrin:
You'd need to be able to configure the parent container to align the actions to the bottom, so I don't think thats right

zepfietje:
From your comment it's unclear to me what direction we want to take, @danharrin.
Care to clarify? 🙂

danharrin:
Personally I can't think of a reason why this PR cannot be merged as-is

zepfietje:
Alright, let's reconsider this then.

packages/tables/resources/views/index.blade.php
```diff
@@ -542,7 +542,8 @@ class="col-span-full"
 
                                 <div
                                     @class([
-                                        'flex items-center',
+                                        'flex items-center' => ! $contentGrid,
```
zepfietje: Please extract the common classes.

zepfietje (commented):
Could you please share screenshots of before and after for this change with bulk actions enabled, @rossbearman? Need to make sure this doesn't break any of the combinations (reorder indicator or bulk select checkbox).

packages/tables/resources/views/index.blade.php
zepfietje: Are you able to clean this up, @rossbearman? 🙂

rossbearman: It's on my list to get to, but I'm snowed under with work currently.

danharrin:
Going to close this as it has been fixed in v4 and there is a lack of activity here.

PR #13447, opened 2024-07-01
Fix view type string vs null|string compatibility

NeftaliYagua:
Fix compatibility on line 99 getDefaultView()
Expected type 'string'. Found 'Filament\Support\Components\view-string|null' public function getDefaultView(): ?string

## Functional changes


danharrin:
How to you get this error in the first place? The condition is `filled()`, it will not return null

PR #13471, opened 2024-07-04
fix: better handling of non-latin column names on import

beholdr:
## Description

Using `Str::lower` instead of `strtolower` here because `ImportColumn->getGuesses()` uses `Str::lower` for column guesses, and it return different result for non-latin strings.

## Visual changes

No visual changes

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/CanImportRecords.php
+++ b/packages/actions/src/Concerns/CanImportRecords.php
@@ -117,7 +117,7 @@ trait CanImportRecords
 
                     $csvColumns = $csvReader->getHeader();
 
-                    $lowercaseCsvColumnValues = array_map('strtolower', $csvColumns);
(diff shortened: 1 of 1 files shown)


packages/actions/src/Concerns/CanImportRecords.php
```diff
@@ -117,7 +117,7 @@ protected function setUp(): void
 
                     $csvColumns = $csvReader->getHeader();
 
-                    $lowercaseCsvColumnValues = array_map('strtolower', $csvColumns);
+                    $lowercaseCsvColumnValues = array_map(fn (string $column): string => Str::lower($column), $csvColumns);
```
danharrin: Does this work?

```suggestion
                    $lowercaseCsvColumnValues = array_map(Str::lower(...), $csvColumns);
```

danharrin (changes requested):

beholdr:
Yep, this works too.

PR #13489, opened 2024-07-05
Added $get('./') syntax to get current repeater item

cheesegrits:
## Description

This change allows the use of $get('./') to return the complete data array for the current instance of a repeater.  It basically extends the '../' syntax to include the concept of "current directory" in the file system analogy.

Also added $get() section to the Builder docs, as I wanted to document that this also works for builders, and there wasn't a section for it.

## Visual changes

No visual changes

## Functional changes


danharrin:
I kinda feel like `$get('./')` should compile down to `$get('')` / `$get()` (without a param) and they should behave in the same way, in the same way you could do `$get('../another-repeater-key')` which doesnt require anything to signify that you are getting all data at that path?

danharrin:
Put simply, `$get()` without a param could return all data from the current path 🤔

cheesegrits:
I did think about that, was kinda torn between $get(), $get(''), $get('.') and $get('./').

Decided against $get() because it feels like the method should require an arg, as it currently does, just for simple mistake proofing.

Of the three string variants, felt like './' was the most explicit.  Kind of feels like it should need a conscious decision to ask for the entire data array.

But I'll go with whatever you feel is best.  Like, $get(), $get(''), $get('.') and $get('./') could all be synonymous.

And I should probably change the doc to say "will return the array of all data from the current path" rather than "current repeater item".

cheesegrits:
@danharrin did you decide anything on this?

danharrin:
I think this is fair

> $get(), $get(''), ~~$get('.')~~ and $get('./') could all be synonymous

`ltrim('./')` and then check if the string is `blank()`?

cheesegrits:
I'm closing this one and doing another PR as somehow this one got totally fubared with extraneous commits.

PR #13497, opened 2024-07-07
Add defaultAction to docs

archilex:
This is an awesome built in feature that definitely needs to be documented. I stumbled upon it when I had to implement this exact feature, but it was a welcome suprise to see you've already implemented it! So many uses cases! I chose an onboarding type example, but change it to anything you like.  

### DIFF
--- a/packages/panels/docs/04-pages.md
+++ b/packages/panels/docs/04-pages.md
@@ -54,21 +54,33 @@ protected function getHeaderActions(): array
 }
 ```
 
-### Default action
+### Opening an action modal when a page loads
 
-You can also easily trigger an action to be launched when a page loads. Just add a method that returns your action and then set the `$defaultAction` property to the name of that action:
+You can also open an action when a page loads by setting the `$defaultAction` property to the name of the action you want to open:
 
 ```php
 use Filament\Actions\Action;
 
-public $defaultAction = 'onboardingAction';
+public $defaultAction = 'onboarding';
 
 public function onboardingAction(): Action
 {
-    return Action::make('onboardingAction')
+    return Action::make('onboarding')
         ->modalHeading('Welcome')
-        ->visible(fn () => ! auth()->user()->isOnBoarded());
+        ->visible(fn (): bool => ! auth()->user()->isOnBoarded());
 }
+```
+
+You can also pass an array of arguments to the default action using the `$defaultActionArguments` property:
+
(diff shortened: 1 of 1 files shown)


ralphjsmit:
Didn't know this existed, super-cool! I really needed this in the past when opening a modal on a page and forcing the user to perform the action otherwise the page wasn't allowed to be accessed. Perfect stuff.

archilex:
@ralphjsmit Our need was kind of similar to yours. We needed to show a warning on a custom page that continuing would do XYZ and we needed the user to confirm this was ok before proceeding. If they canceled they'd be redirected back. So many use cases for this feature.

ralphjsmit:
Exactly! Super nice & very useful. I messed around with it earlier with e.g. calling `mountAction()` from the `mount()`, but that didn't work. Calling `mountAction()` from a booted method did work iirc, but that wasn't also ideal. So this is super useful indeed and surprising that it's not yet documented. But glad it gets documented at last 🫡

danharrin:
Thanks

packages/panels/docs/04-pages.md
```diff
@@ -54,6 +54,35 @@ protected function getHeaderActions(): array
...
+You can also pass an array of arguments to the default action using the `$defaultActionArguments` property:
+
+```php
+public $defaultActionArguments = ['step' => 2];
+```
+
+Alternatively, you can open an action modal when a page loads by specifying the `defaultAction` as a query string parameter to the page:
+
+```
+/admin/products/edit/932510?defaultAction=onboarding
```
saade, line 83:
```suggestion
/admin/products/edit/932510?action=onboarding
```

Hey @archilex, @danharrin i think this is wrong? The correct query param is `action`. `defaultAction` doesn't work as Povilas just pointed in his last video.

saade: I know for sure that `action` works because i'm using it in production. Not sure about `defaultAction`.

wychoong: https://github.com/filamentphp/filament/blob/e90998fe6aa4681569f09b6ebb01e5346d3a1644/packages/actions/src/Concerns/InteractsWithActions.php#L42-L49

danharrin: it was already fixed in the docs, I just haven't published them since

PR #13509, opened 2024-07-08
Hint icon tabindex

dreftas:
## Description

Parameter to change `tabindex=""` for a form field hint icon.

Usage:

```php

TextInput::make('name')
    ->hintIcon('heroicon-m-question-mark-circle', tooltip: 'Need some more information?', tabIndex: -1)
```
or 

```php

TextInput::make('name')
    ->hintIcon('heroicon-m-question-mark-circle')
    ->hintIconTabIndex(-1)
```
Accepts `int`, `Closure` or `null` values, default is `null`.

It is useful when hint icon needs to have specific `tabindex`, or `tabindex="-1"` when you need the possibility to disable focusing on that hint icon when navigating through form using `Tab` key.

## Visual changes

No visual changes.

## Functional changes

### DIFF
--- a/packages/forms/docs/03-fields/01-getting-started.md
+++ b/packages/forms/docs/03-fields/01-getting-started.md
@@ -189,6 +189,17 @@ TextInput::make('name')
     ->hintIcon('heroicon-m-question-mark-circle', tooltip: 'Need some more information?')
 ```
 
+#### Changing tab index for a hint icon
+
+Additionally, you can change `tabindex` for a hint icon, using the `tabIndex` parameter of `hintIcon()`:
+
+```php
+use Filament\Forms\Components\TextInput;
+
+TextInput::make('name')
+    ->hintIcon('heroicon-m-question-mark-circle', tooltip: 'Need some more information?', tabIndex: -1)
+```
+
 ## Adding extra HTML attributes
 
 You can pass extra HTML attributes to the field, which will be merged onto the outer DOM element. Pass an array of attributes to the `extraAttributes()` method, where the key is the attribute name and the value is the attribute value:
--- a/packages/forms/resources/views/components/field-wrapper/hint.blade.php
+++ b/packages/forms/resources/views/components/field-wrapper/hint.blade.php
@@ -3,6 +3,7 @@
     'color' => 'gray',
     'icon' => null,
     'tooltip' => null,
+    'tabindex' => null
 ])
 
 <div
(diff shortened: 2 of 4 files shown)


zepfietje:
Hint icons aren't focusable by default (at least from my testing), so I'm not sure what this change is for.

Please clarify if you think I'm wrong. 🙂

dreftas:
I'll update with before & after video later :)

dreftas:
@zepfietje 
Here is video how it works now when there is hint, tooltip and icon on a field:

https://github.com/user-attachments/assets/76e964ce-3625-40b9-9b68-be854813a8ad

And here is video how it would work having this feature and adding `->hintIconTabIndex(-1)` which adds `tabindex="-1"` to an icon (svg when using heroicons):

https://github.com/user-attachments/assets/7289dac7-2161-406f-9a99-394378631b82

As you can see in second video, icon is not focused when navigating through form fields using TAB key. This is very useful with forms where you put quite a lot data and do not want to click on every field with a mouse and using only TAB key, and if form has 10 fields and every field has this kind of hint icon it is good for the first time to check what field is used for (some description in a tooltip), but when doing this kind of data insert every day multiple times, the thing that hint icon is always focused before actual input is very uncomfortable.

Of course there are other form fields parts which are also focusable with tab key and it would be cool to have control over that too, but this PR only covers hint icons. For example searchable select with create options has remove icon (X) and create button (+) focusable too and again, if it is a lot of fields and there are these kind of selects, to "jump" to another field when next one is this kind of select, TAB key must be pressed 4 times (1 - focus select itself, 2 - focus X removal icon, 3 - focus create/edit option, 4 - focus another field)

image

zepfietje:
Like I said, the hint icon isn't focusable for me.
Could you please provide the exact OS and browser (including versions) you're using?
Please also test if you can reproduce it on other browsers.

As for the suffix action, you should already be able to customize that using `extraAttributes`.

dreftas:
> Like I said, the hint icon isn't focusable for me. Could you please provide the exact OS and browser (including versions) you're using? Please also test if you can reproduce it on other browsers.
> 
> As for the suffix action, you should already be able to customize that using `extraAttributes`.

Tested with various browsers, all of these focuses on hint icon.

- Opera: 112.0.5197.24
- Google Chrome: 126.0.6478.127 (64-bit)
- Microsoft Edge: 126.0.2592.87 (64-bit)

Only **Firefox: 128.0 (64-bit)** does not focus on those hint icons.

I'm using Windows 11 Home version 23H2 (Build: 22631.3880)

dreftas:
@zepfietje is this will be considered to be reopened, or I should make my own ways for this thing, because this wouldn't be implemented in any way ? :)

zepfietje:
I can't reproduce this on Safari and Chrome on macOS.
Could you please post a video recording, @dreftas?

PR #13512, opened 2024-07-09
fix: allow view() to be used as icon

ju5t:
## Description

This is an updated PR of #13358.

In the [docs](https://filamentphp.com/docs/3.x/support/icons#replacing-the-default-icons) it mentions you can use a view as an icon within `FilamentIcon::register`. This didn't work.

A view gets rendered to a string because it implements the magic method `__toString()` and `HasIcon` did not accept `Htmlable`, but does accept strings. PHP converts it to a string to match the string type on properties and methods.

The reason it doesn't work when you also accept `Htmlable` is because views also implement `Renderable`. All data passed to views an its derivatives (e.g. Blade components) will get processed by [gatherData](https://github.com/laravel/framework/blob/11.x/src/Illuminate/View/View.php#L221) whichs renders Renderables to HTML.

This is why we're wrapping the rendered view into an `HtmlString` as that's not `Renderable`.

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/support/src/Concerns/HasIcon.php
+++ b/packages/support/src/Concerns/HasIcon.php
@@ -40,9 +40,9 @@ trait HasIcon
 
     public function getIcon(): string | Htmlable | null
     {
-        $icon = $this->icon instanceof Renderable
-            ? new HtmlString($this->icon->render())
-            : $this->icon;
+        $icon = $this->icon instanceof Renderable ?
+            new HtmlString($this->icon->render()) :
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks

ju5t:
@danharrin your last commit doesn't use the result of `$this->evaluate`. This is also why the tests are failing now.

danharrin:
Apologies, will fix

PR #13514, opened 2024-07-09
Support for adding an icon to a page heading

tjodalv:
## Description

This is something I use often in my Filament projects. It allows you to add an icon next to a page heading. I’ve updated the documentation to illustrate how to implement it. You can customize the icon’s position (before or after the heading), it's color and size.

## Functional changes


zepfietje:
Thank you for contributing to Filament, @tjodalv!

However, I don't think an icon next to the heading is such a common pattern that it warrants an API in Filament's core.
Since you can easily achieve the same result in your own projects by returning an `HtmlString` from the `getHeading()` method, I'm going to close this PR.

PR #13520, opened 2024-07-09
make select component accept null value

obeidareda37:
Package
Form builder

Package Version
v3.2

When I used the Select component on a form and tried to fetch values ​​from a relationship, I ran into some challenges. Specifically, when there is any empty record in the relationship table, it throws an error. However, if all records contain values, this works fine.

The first selection works fine because all columns of records in the form have a value
The second select return error because there are a nullable column in model

This is what the error looks like

Screenshot 2024-07-09 at 9 07 24 PM

This form in filament resource

Screenshot 2024-07-09 at 9 12 34 PM

and this relation 

Screenshot 2024-07-09 at 9 13 06 PM

In This PR The Select has been made to accept a value of null.

### DIFF
--- a/packages/forms/src/Components/Concerns/CanDisableOptions.php
+++ b/packages/forms/src/Components/Concerns/CanDisableOptions.php
@@ -36,14 +36,14 @@ trait CanDisableOptions
     /**
      * @param  array-key  $value
      */
-    public function isOptionDisabled($value, string $label): bool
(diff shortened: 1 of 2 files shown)


danharrin:
Thanks, but this is the correct solution: https://github.com/filamentphp/filament/discussions/13517#discussioncomment-10009372

PR #13526, opened 2024-07-10
Feat: ImportAction events

smiliyas:
## Description

Three new events have been added to the ImportAction:

- `ImportInitiated` - Triggered just before the import job starts
- `ImportCompleted` - Triggered after all jobs are processed

Although we typically use the `Import` model's observers, the `options` and `columnMap` data in `ImportAction` are not accessible within these observers. In such scenarios, these new event triggers will be useful, as all of the `Import` model, `options` and `columnMap` data are available in these events.

One more event for `ImportCsv` job

- `ImportCsvProcessed` - Triggered at the end of each `ImportCsv` job executions

## Visual changes

NA

## Functional changes


packages/actions/src/Imports/Events/ImportCsvProcessed.php
```diff
@@ -0,0 +1,45 @@
...
+{
+    use Dispatchable;
+
+    /**
+     * @param array<Throwable>  $exceptions
+     */
+    public function __construct(
+        protected Import $import,
+        protected int $processedRows,
+        protected int $successfulRows,
```
danharrin: Can you please confirm why you are passing these here since they are already available in the `$import` object?

packages/actions/src/Imports/Jobs/ImportCsv.php
```diff
@@ -110,6 +111,8 @@ public function handle(): void
 
         $this->import->save();
 
+        event(new ImportCsvProcessed($this->import, $processedRows, $successfulRows, $exceptions));
```
danharrin: Please pass the column map and options in here too, before the exceptions

packages/actions/src/Concerns/CanImportRecords.php
danharrin, line 1: Please swap the order of arguments so the column map is passed in before the options

danharrin (changes requested):

packages/actions/src/Imports/Events/ImportCsvProcessed.php
smiliyas: The `$import` object includes only the cumulative numbers for all processed jobs; however, the specified parameters are used to determine the statistics for the current job numbers.

smiliyas:
@danharrin 
All requested changes have been completed. Please let me know if any further adjustments are needed.

danharrin:
Thanks

PR #13530, opened 2024-07-10
Show bulk actions below table

howdu:
## Description
Add option to display bulk actions below table

```php
->bulkActions([
])
->bulkActionsPosition(Tables\Enums\BulkActionsPosition::BelowTable)
```

## Visual changes
Screenshot 2024-07-10 at 13 38 58

Screenshot 2024-07-10 at 13 35 58

## Functional changes


danharrin:
@zepfietje would appreciate the green light on the design before I proceed into the code please

zepfietje:
Personally I don't think having the actions at the bottom results in a good UI/UX.
Could you clarify your reasoning here, @howdu?

howdu:
@zepfietje it was a client who requested it - their reasoning was they work their way down the table selecting all the relevant records and always finish at the bottom where they wanted the action button instead of having to scroll to the top.

The table is implemented as a checklist where they need to work through all the rows selecting all the relevant records at once.

zepfietje:
Hmm, I'm wondering if the bulk select checkbox and select/deselect actions should be in the footer too then?

howdu:
@zepfietje I've added this above the action let me know what you think.
Screenshot 2024-07-21 at 07 45 37

Screenshot 2024-07-21 at 07 45 54

Also added the option to display both above and below table. You can get clever with not displaying this on a small tables.
```php
->bulkActionsPosition(
    fn (Table $table) => $table->getRecords()->count() > 5
        ? Tables\Enums\BulkActionsPosition::AboveAndBelowTable
        : Tables\Enums\BulkActionsPosition::AboveTable
)
```

packages/tables/resources/views/index.blade.php
```diff
@@ -1250,6 +1254,36 @@ class="fi-ta-record-checkbox"
             @endif
         </div>
 
+        @if ($hasBulkActionsBelowTable && (! $isReordering) && count($bulkActions))
+            <div
```
zepfietje, line 1258: Is there a way to minimize duplication here and in the header?
Maintenance of the table index view is already very hard, so I don't want to introduce additional complexity.

zepfietje (changes requested):

zepfietje:
@howdu, shouldn't it also show the bulk selection checkbox?
It should be identical to what's in the header, right?

howdu:
@zepfietje I think adding the bulk checkbox would overcomplicate this. I was trying to keep it simple and work the same as filter position above/below.

zepfietje:
Alright, I understand now since the bulk selection checkbox is in the column headers row.
What does this PR look like when positioning bulk actions *and* filters in the footer of the table, @howdu?

howdu:
@zepfietje it displays above the pagination so doesn't affect that. 

Screenshot 2024-08-09 at 14 03 39

Screenshot 2024-08-09 at 14 04 43

packages/tables/resources/views/index.blade.php
howdu: @zepfietje please let me know what changes you think best to improve this. 

Perhaps using `capture` blade directive? But due to it not including the reorder action in the footer I think the `selection-indicator` and bulk action`filament-tables::actions` would need to be captured separately?

danharrin:
Going to close this as it feels stale. I'm not entirely sure I like how the bulk actions look under the table and I can't say I've had many requests for it either. If there is a particular explanation behind it, let me know and we can consider accepting a PR in v4.

PR #13537, opened 2024-07-11
Text column and entry tooltip per item

Skullbock:
## Description

This PR is a followup of #13420 as per agreement with @danharrin
It introduces the ability to add tooltips to text columns that contain repeatable text: badges, list items, or concatenated list items.

Method  name is up for change if it's not good enough!

## Visual changes
Before:
Screenshot 2024-07-11 alle 10 41 17
Screenshot 2024-07-11 alle 10 41 29
Screenshot 2024-07-11 alle 10 41 41

After:
Screenshot 2024-07-11 alle 10 41 23
Screenshot 2024-07-11 alle 10 41 33
Screenshot 2024-07-11 alle 10 41 45

## Functional changes

### DIFF
--- a/packages/tables/resources/views/columns/text-column.blade.php
+++ b/packages/tables/resources/views/columns/text-column.blade.php
@@ -125,6 +125,8 @@
                         $lineClamp = $getLineClamp($state);
                         $size = $getSize($state);
                         $weight = $getWeight($state);
+                        $itemTooltip = $getItemTooltip($state);
+                        $hasItemTooltip = filled($itemTooltip);
 
                         $iconClasses = \Illuminate\Support\Arr::toCssClasses([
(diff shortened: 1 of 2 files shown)


zepfietje:
Can't we just use the existing tooltip API, @danharrin?

danharrin:
Yeah I think we probably could but it could be a breaking change if someone is calling that method. I might push this back to v4

Skullbock:
I agree that the best way would be to be able to reuse the tooltip api with a state parameter similar to other apis in filament but that's a breaking change I fear. 

If I can help on the v4 port let me know. Feel free to close this if it's not feasible for v3

Thanks!

danharrin:
I'll leave this open as a reminder to work on it myself, thanks

danharrin:
Replaced by #14967 for v4

PR #13542, opened 2024-07-11
Fix: Make Select loadStateFromRelationshipsUsing() consider HasOne and HasMany relationships

smpita:
## Description

When using a `Filament\Forms\Components\Select` on a `HasOne` or `HasMany` relationship, a `BelongsTo` is assumed (see the line `/** @var BelongsTo $relationship */`)

When a related model is found `$relationship->getOwnerKeyName()` gets called resulting in a runtime failure due to `getOwnerKeyName` not being a valid function on `HasOne` or `HasMany`

## Visual changes

No visual differences

## Functional changes

### DIFF
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -854,8 +854,9 @@ class Select extends Field implements Contracts\CanDisableOptions, Contracts\Has
                 $relatedRecords = $relationship->getResults();
 
                 $component->state(
-                    $relatedRecords
-                        ->pluck($relationship->getForeignKeyName())
+                    $relatedRecords->pluck(
+                        $relationship->getForeignKeyName()
+                    ),
                 );
 
                 return;
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -851,6 +851,7 @@ class Select extends Field implements Contracts\CanDisableOptions, Contracts\Has
             }
 
             if ($relationship instanceof HasMany) {
+                /** @var Collection $relatedRecords */
                 $relatedRecords = $relationship->getResults();
 
                 $component->state(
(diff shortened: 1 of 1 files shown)


packages/forms/src/Components/Select.php
```diff
@@ -848,6 +848,16 @@ public function relationship(string | Closure | null $name = null, string | Clos
                 return;
             }
 
+            if ($relationship instanceof HasOneOrMany) {
+                $relatedModel = $relationship->getResults();
+
+                $component->state(
+                    $relatedModel->getAttribute(
+                        $relatedModel->getKeyName(),
```
danharrin: I think you need the get the key name from the relationship in case it was customised?

danharrin (changes requested):

packages/forms/src/Components/Select.php
danharrin: Also what happens if this is a HasMany instead of a HasOne? Then you need to pluck from a collection I think

smpita: Good catches.
I also missed the early return after setting the state.

packages/forms/src/Components/Select.php
```diff
@@ -848,6 +850,31 @@ public function relationship(string | Closure | null $name = null, string | Clos
                 return;
             }
 
+            if ($relationship instanceof HasMany) {
+                /** @var Collection $relatedRecords */
+                $relatedRecords = $relationship->getResults();
+
+                $component->state(
+                    $relatedRecords->pluck(
+                        $relationship->getForeignKeyName()
```
```suggestion
                        $relationship->getForeignKeyName(),
```

danharrin (changes requested):

danharrin:
Thank you

PR #13551, opened 2024-07-12
Fixed visibility of action buttons on form with deferred filters.

den1n:
## Visual changes

Before.

image

After.

image

### DIFF
--- a/packages/tables/src/Table/Concerns/HasFilters.php
+++ b/packages/tables/src/Table/Concerns/HasFilters.php
@@ -205,7 +205,8 @@ trait HasFilters
                 Action::make('resetFilters')
                     ->label(__('filament-tables::table.filters.actions.reset.label'))
                     ->color('danger')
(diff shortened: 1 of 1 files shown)


danharrin:
Does this affect the appearance of the buttons when they are not in a filters modal and are above the table / in the popover instead? Did they used to be links or have they always been normal buttons?

den1n:
> Does this affect the appearance of the buttons when they are not in a filters modal and are above the table / in the popover instead

The appearance of the buttons is the same on all filter layout (modal, dropdown, above/below content).

> Did they used to be links or have they always been normal buttons?

I think, they had no type before. The apply action is added recently by commit https://github.com/filamentphp/filament/commit/863cd3de74ba42cb41d0a417dd786a3878888519#diff-43d81083da71caf720a1e909491afd09b93e76146f75376c4994192435e069f4R196-R212.

There is no link, button or icon method call anywhere.

danharrin:
Thanks

PR #13166, opened 2024-06-09
Add fallback default locale for format curreny

datlechin:
## Description

Add fallback default locale for format curreny.

This update simplifies the code for passing the locale parameter when chaining to the `money()` method. If no locale is provided, the application’s default locale will be used automatically.

## Functional changes

- [x] Code style has been fixed by running the `composer cs` command.
- [x] Changes have been tested to not break existing functionality.
- [x] Documentation is up-to-date.

PR #13556, opened 2024-07-12
Override default number locale in Tables and Infolists

### LINKED ISSUE #13340: Number::useLocale not working
### Package

filament/filament

### Package Version

v3.2.92

### Laravel Version

v11.11.1

### Livewire Version

v3.5.1

### PHP Version

PHP 8.3.8

### Problem description

When adding the number helper function it doesn't work as expected.
I used ar locale because it changes the numbers completely.

### Expected behavior

To change the Locale according to the provided locale in the AppServiceProvider

### Steps to reproduce

1. Make a filament project
2. Make a test resource
3. Add Number::useLocale('ar'); to AppServiceProvider boot() function
4. check

https://github.com/Zacktamondo/FilamentNum

### Relevant log output

_No response_

dmason30:
## Description

Caused by: https://github.com/filamentphp/filament/pull/13166 (3.2.87)

In my opinion #13166 should be reverted instead of this change. If they want to set a default locale for the number helper they should have just used `Number::useLocale(config('app.locale'))`. 

If not, this workaround is the new way to set a global default number locale in filament for tables and infolists.

The filament docs suggest using `Number::useLocale($locale)` to set the default number formatting globally.

However, as reported in #13340  this does not override the `config('app.locale')` locale that filament is now passing to `Number::currency` and `Number::format` functions by default since 3.2.87.

Table docs:
image

Here in Infolists docs:
image

## Visual changes

None

## Functional changes
I have added a `Table::$defaultNumberLocale` static variable that can be used to override the table default.

I have added the same for `Infolist::$defaultNumberLocale`.

### DIFF
--- a/packages/infolists/docs/03-entries/02-text.md
+++ b/packages/infolists/docs/03-entries/02-text.md
@@ -85,12 +85,12 @@ TextEntry::make('stock')
     ->numeric(locale: 'nl')
 ```
 
-Alternatively, you can set the default locale used across your app using the `Number::useLocale()` method in the `boot()` method of a service provider:
+Alternatively, you can set the default locale used across your app using the `Infolist::$defaultNumberLocale` method in the `boot()` method of a service provider:
 
 ```php
-use Illuminate\Support\Number;
+use Filament\Infolists\Infolist;
 
-Number::useLocale('nl');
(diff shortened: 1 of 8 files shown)


danharrin:
Thanks!

PR #13559, opened 2024-07-13
Allow table groups to accept a closure

archilex:
## Description

Allow table groups to accept a closure:

```php
return $table
    ->groups(function (Component $livewire) {
        return match ($livewire->activeTab) {
            'pending' => ['order_date'],
            'shipped' => ['shipping_date'],
            'delivered' => [
                  Group::make('delivered_date')->label('Delivered'),
             ],
        };
    })
```

## Functional changes

### DIFF
--- a/packages/tables/src/Table/Concerns/CanGroupRecords.php
+++ b/packages/tables/src/Table/Concerns/CanGroupRecords.php
@@ -15,6 +15,11 @@ trait CanGroupRecords
     /**
      * @var array<string, Group>
      */
+    protected array $cachedGroups;
+
+    /**
+     * @var array<string | Group> | Closure
+     */
     protected array | Closure $groups = [];
 
     protected bool | Closure $isGroupsOnly = false;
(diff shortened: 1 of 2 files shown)


packages/tables/src/Table/Concerns/CanGroupRecords.php
danharrin, line 1: Before, multiple calls to `getGroups()` would not need to reconstruct the array of ID-keyed groups. Now, that happens.

Please introduce a new `cachedGroups` property where you save the keyed groups into, so that doesn't run multiple times.

Please also fix the `@var` docblock on the `$groups` property and argument to introduce the closure and string value and remove the string key.

danharrin (changes requested):

packages/tables/src/Table/Concerns/CanGroupRecords.php
archilex: @danharrin Good call. Added the `cachedGroups` property and also added a `getCachedGroups()` method to be consistent with the rest of the code base.

danharrin: I think in this case, we can avoid the new method and implement this within `getGroups()` please, I usually add the extra method if the original method is overridable by the user which would break the caching

archilex: Sounds good! Done.

danharrin:
Thanks!

PR #13564, opened 2024-07-13
Docs update: Add info to hide column label

kachelle:
## Description
Add info to hide column label.

## Visual changes
new: 
image

## Functional changes

### DIFF
--- a/packages/tables/docs/03-columns/01-getting-started.md
+++ b/packages/tables/docs/03-columns/01-getting-started.md
@@ -77,6 +77,15 @@ TextColumn::make('title')
     ->translateLabel() // Equivalent to `label(__('Title'))`
 ```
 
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks, but this is not official syntax or recommended since the column will not be accessible without a label

kachelle:
> Thanks, but this is not official syntax or recommended since the column will not be accessible without a label

How would you recommend hiding a column header?

danharrin:
If you want to make a PR, add a new method `hiddenHeaderLabel()` (getter is `isHeaderLabelHidden()`, property is `$isHeaderLabelHidden`), which applies a `sr-only` class around the label

kachelle:
Ok! Maybe I'll try to make my first technical PR.

PR #13567, opened 2024-07-13
Fix #13518

### LINKED ISSUE #13518: Stats overview widget charts randomly change color when resizing browser
### Package

filament/filament

### Package Version

v3.2.92

### Laravel Version

v11.14.0

### Livewire Version

v3.5.2

### PHP Version

PHP 8.2.21

### Problem description

When using stat widgets, the chart color will load fine on a hard reload but when you resize the browser window, will randomly flip between colors.

WqE9mpX

f1IyqJJ

I tested in both Chrome and Safari so I don't think it's a browser issue.

More discussion here: https://github.com/filamentphp/filament/discussions/10637

**Edit:** They also appear to change color with sidebar collapse and also polling updates.

### Expected behavior

Stay the original color.

### Steps to reproduce

Open demo app dashboard and resize browser. Observe stat widget chart colors change.

oele-dev:
Fix #13518 

use per-dataset settings instead of default colors to prevent the responsive behaviour change the chart color

## Description

To the chart initialization for stats widget the responsive option of chartjs override the defaults color settings randomly according to the amount of chart instances, to prevent it, this require set the colors using per-dataset settings.

## Visual changes

after
Screen-Recording-2024-07-13

## Functional changes

### DIFF
--- a/packages/widgets/resources/js/components/stats-overview/stat/chart.js
+++ b/packages/widgets/resources/js/components/stats-overview/stat/chart.js
@@ -41,13 +41,6 @@ export default function statsOverviewStatChart({
         },
 
         initChart: function () {
-            Chart.defaults.backgroundColor = getComputedStyle(
-                this.$refs.backgroundColorElement,
-            ).color
-
-            Chart.defaults.borderColor = getComputedStyle(
-                this.$refs.borderColorElement,
-            ).color
 
             return new Chart(this.$refs.canvas, {
                 type: 'line',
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks! Is it also worth doing this for the normal chart widget too in `components/chart.js`?

oele-dev:
Not really, the normal chart widget only has one instance of chartjs, there is no conflict there.

danharrin:
Can you make it consistent though please? :) You can also use more than one chart widget on a page, not sure if that affects it

oele-dev:
In the normal chart widget the instance can receive an array with multiple datasets so we can't use the per dataset configuration, the current definition is the best option or do you want to read all the data and set the required properties? but I have found many options set by default attribute.

In the stats chart we can define multiple instances of chartjs from a single filament widget and all of them are defined using the same value of refs “canvas”. so the same refs name creates a conflict with the chartjs resize function but in the normal widget there is only one instance of chartjs so there is no conflict here, even when you have many widgets per page.

danharrin:
Thanks for the explanation!

PR #13585, opened 2024-07-15
Display current state for placeholder field

datlechin:
By default, when no `->content()` is specified, the content is empty. This PR automatically adds the current state to the content when none is specified.

## Functional changes


danharrin:
This is a breaking change, and achievable with `content(fn ($component) => $component->getState())`

PR #13593, opened 2024-07-16
Fix: avoid table polling while reordering

### LINKED ISSUE #12108: Bug: Table Polling Interferes with Row Reordering in Filament Tables
### Package

filament/filament

### Package Version

v3.2.49

### Laravel Version

v10.48.2

### Livewire Version

v3.4.9

### PHP Version

PHP 8.2

### Problem description

When using Filament tables with both polling (`->poll()`) and row reordering (`->reorderable()`) enabled, I encountered a visual and functional glitch during the drag-and-drop reordering process. Specifically, if the table's polling interval triggers a refresh while a row is being dragged, the dragged row's placeholder reverts to its original position, and the reordering operation fails to complete as expected. This issue disrupts the user experience, making row reordering unreliable in dynamically updated tables.

### Expected behavior

I expected that during a drag-and-drop operation, the table's polling would either be paused or intelligently managed to not interfere with the reordering process. The dragged row should maintain its position relative to the cursor, and upon dropping, the row should reorder correctly, regardless of any ongoing or intermediate polling updates.

### Steps to reproduce

1. Setup a Filament Table: Create a Filament table component with row reordering enabled (->reorderable()) and polling set to a short interval (e.g., ->poll('2s')).
3. Add Rows to the Table: Ensure the table has multiple rows that can be reordered.
4. Attempt to Reorder Rows: Click and hold to drag a table row to a new position.
5. Observe During Poll Interval: Keep the row held and wait for the polling interval to trigger a refresh before releasing the mouse button.
https://github.com/filamentphp/filament/assets/54978466/56d39d95-88ea-4ab9-a786-ceda4b828dc7

6. Release and Check Position: Release the row and observe that the reordering does not reflect the expected new position.

Note: The Reproduction repository is a fork of the official filament demo site (https://github.com/filamentphp/demo) with modifications to the ProductResource which can be used to test the issue from step 4 to 6. The headstart template can't run `composer install` and it has duplicate migrations, making it difficult to work with, that's why I used the Filament demo repo instead.

### Reproduction repository

https://github.com/kaspernowak/reorder-and-poll-demo

### Relevant log output

_No response_

smiliyas:
## Description

closes #12108 

## Visual changes

NA

## Functional changes

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -284,7 +284,7 @@
         @endif
 
         <div
-            @if ($pollingInterval = $getPollingInterval())
+            @if (! $isReordering && ($pollingInterval = $getPollingInterval()))
                 wire:poll.{{ $pollingInterval }}
             @endif
             @class([
@@ -358,7 +358,9 @@
                                             <x-filament::input.select
                                                 x-model="column"
                                             >
-                                                <option value="">{{ $defaultSortOptionLabel }}</option>
+                                                <option value="">
+                                                    {{ $defaultSortOptionLabel }}
+                                                </option>
 
                                                 @foreach ($sortableColumns as $column)
                                                     <option
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -284,7 +284,7 @@
         @endif
 
         <div
-            @if (! $isReordering && ($pollingInterval = $getPollingInterval()))
+            @if ((! $isReordering) && ($pollingInterval = $getPollingInterval()))
                 wire:poll.{{ $pollingInterval }}
             @endif
             @class([


packages/tables/resources/views/index.blade.php
```diff
@@ -284,7 +284,7 @@ class="fi-ta-header-toolbar flex items-center justify-between gap-x-4 px-4 py-3
         @endif
 
         <div
-            @if ($pollingInterval = $getPollingInterval())
+            @if (! $isReordering && ($pollingInterval = $getPollingInterval()))
```
```suggestion
            @if ((! $isReordering) && ($pollingInterval = $getPollingInterval()))
```

danharrin (changes requested):

kaspernowak:
Thanks @smiliyas for this fix!
I do see an issue with this "blanket" solution though, as it will stop all polling when reordering is enabled, instead of just when a record is actively being dragged. A better approach IMO would be to set a parameter to true when the sortablejs onStart event is triggered and set to false when the onEnd event is triggered. 
This fix effectively forces me to remove the reordering functionality from my table, as polling is essential for me when I work with my orders table and doing delivery routing in rush hour, as I would not see new orders when reordering records.

@danharrin Is it feasible to tap into the sortablejs onStart and onEnd events here?

danharrin:
You will have to test it yourself, I don't know if removing the attribute from JS would be enough in case an additional Livewire render happens while dragging

kaspernowak:
@danharrin Maybe there could be a quick fix to this. Is there a way to manually trigger a table polling/refresh/update event in a `ListRecords` component method? Like as we can do `$this->getTableRecords()`, can we do something like `$this->refreshTable()` that forces a table update?

danharrin:
From JavaScript or PHP?

kaspernowak:
@danharrin from PHP in an overwritten `reorderTable` method.

danharrin:
It happens every time Livewire renders anyway

kaspernowak:
@danharrin Whenever I reorder my records, I trigger actions that update related records. The updated values of the related records do not show before I refresh the page (or disable reordering). So I want to force this update at the end of my reordering actions, is there a way to do this?

danharrin:
That is a Livewire method, so after dropping the page should rerender and you should see updates.

kaspernowak:
It turned out to be an issue with how I pulled the table records with `$this->getTableRecords()` in my overwritten method. defining a public parameter for `$tableRecords` on my List page for the table records, pulling the table records in my getHeaderWidgets method with `$this->getTableRecords()` and storing it in the public parameter instead, and using `$tableRecords` in my overwritten method did not result in failed table record updates on reordering. I don't fully understand livewire and how table records are handled on filament pages, but it seems like it works for now. Sorry for the off topic, just wanted to conclude this.

PR #13597, opened 2024-07-16
Create 03-group.md

carstenjaksch:
Documentation about the `Group` layout component is missing.

This PR is a proposal. I hope the few words and comments are helpful as a starting point.


danharrin:
Hi, we've intentionally not documented it previously because it's difficult to define what it should be used for, as its not really for controlling layout. The `relationship()` method, for example, is available for all layout components and not just this one. Thanks for the contribution but going to close because I think a lot of work is required to make this mergeable

PR #13603, opened 2024-07-17
Fix nested Eager Loading for EntanglesStateWithSingularRelationship Components Concern

yebor974:
## Description

If you use many nested grids / field sets and others with relationship, all nested relations are reloaded in `getCachedExistingRecord` function even if they have been eager loaded.
And so, too many requests are executed.

## Visual changes

No visual changes.

## Functional changes

### DIFF
--- a/packages/forms/src/Components/Concerns/EntanglesStateWithSingularRelationship.php
+++ b/packages/forms/src/Components/Concerns/EntanglesStateWithSingularRelationship.php
@@ -203,14 +203,18 @@ trait EntanglesStateWithSingularRelationship
             return $this->cachedExistingRecord;
         }
 
-        $relationshipName = $this->getRelationshipName();
-
-        if (! $relationshipName) {
+        if (! $parentRecord = $this->getRecord()) {
             return null;
         }
 
-        if ($this->getModelInstance()->relationLoaded($relationshipName)) {
-            $record = $this->getModelInstance()->getRelationValue($relationshipName);
+        $relationshipName = $this->getRelationshipName();
+
+        if (blank($relationshipName)) {
+            return null;
+        }
+
+        if ($parentRecord->relationLoaded($relationshipName)) {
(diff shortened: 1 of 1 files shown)


packages/forms/src/Components/Concerns/EntanglesStateWithSingularRelationship.php
```diff
@@ -203,7 +203,17 @@ public function getCachedExistingRecord(): ?Model
...
 
-        $record = $this->getRelationship()?->getResults();
+        $relationshipName = $this->getRelationshipName();
+
+        if (! $relationshipName) {
+            return null;
+        }
+
+        if ($this->getModelInstance()->relationLoaded($relationshipName)) {
+            $record = $this->getModelInstance()->getRelationValue($relationshipName);
```
danharrin: I think we should first save `$parentRecord` as `$this->getRecord()`, early return null if that is null, then use that instead of calling `$this->getModelInstance()` twice?

danharrin (changes requested):

packages/forms/src/Components/Concerns/EntanglesStateWithSingularRelationship.php
```diff
@@ -203,7 +203,21 @@ public function getCachedExistingRecord(): ?Model
             return $this->cachedExistingRecord;
         }
 
-        $record = $this->getRelationship()?->getResults();
+        if (! $parentRecord = $this->getRecord()) {
```
danharrin:
```suggestion
        $parentRecord = $this->getRecord();

        if (! $parentRecord) {
```

packages/forms/src/Components/Concerns/EntanglesStateWithSingularRelationship.php
yebor974: it's done.

danharrin:
Thanks

PR #13606, opened 2024-07-17
Make actions array optional

carstenjaksch:
Just a small change so that the `make()` method doesn't require an array as argument.

**Use case:**

We can create custom action classes, for example `SomeAction`, and use that like so:`SomeAction.make()`.

I would like to create a custom action **group** class so that I can reuse it in several resources. However, I cannot use it like the above example. I need to give it an empty array: `SomeActionGroup::make([])`

This doesn't look good.

Making the array optional and having an empty array by default shouldn't be an issue because the `actions()` method sets an empty array too (on the `actions` property).

### DIFF
--- a/packages/actions/src/ActionGroup.php
+++ b/packages/actions/src/ActionGroup.php
@@ -69,7 +69,7 @@ class ActionGroup extends ViewComponent implements HasLivewire
     /**
      * @param  array<StaticAction | ActionGroup>  $actions
      */
-    public static function make(array $actions): static
+    public static function make(?array $actions = []): static
     {
         $static = app(static::class, ['actions' => $actions]);
         $static->configure();


tonypartridge:
I am confused, at what point would you use an ActionGroup without an array?

carstenjaksch:
Sorry for the confusion @tonypartridge.

There will be multiple resources (Page, Post, Product, ...) that will use this action group to display a list of links to the respective url. Each resource will have multiple associated websites, thus I need the action group.

The reason I don't need to pass an array to `make()` is because I build my own in the class – since this will be used multiple times and should be global.

Here is my class as an example:

```php
<?php

namespace App\Filament\Actions;

use App\Models\Site;
use Closure;
use Filament\Actions\Action;
use Filament\Actions\ActionGroup;
use Illuminate\Contracts\Database\Eloquent\Builder;
use Illuminate\Support\Facades\App;

class VisitActionGroup extends ActionGroup
{
    protected array $actions = [];

    protected Closure|array|string|null $color = 'info';

    {
        // TODO: Somehow get the slug of the resource instance that contains this action group.

        parent::setUp();

        $this->icon(Site::icon());

        $this->label(__('app/filament/actions.visit-group.label'));

        $this->button();

        $this->color($this->color);

        // TODO: Enhance query with sites that are related to the resource instance containing this action group.
        $sites = Site::withWhereHas('domains', function (Builder $query) {
            $query->where('default', true);
        })->get();

        $sites->each(function (Site $item) {
            $this->actions[] = Action::make('visit')
                ->label($item->name)
                ->color($this->color)
                ->icon('heroicon-o-arrow-top-right-on-square')
                ->url(function () use ($item): string {
                    $domain = $item->domains[0]->domain;

                    if (App::isLocal()) {
                        $domain = $item->localDomain();
                    }

                    return $domain;
                })
                ->openUrlInNewTab();
        });

        $this->actions($this->actions);
    }
}
```

carstenjaksch:
Well, in order to access the resource record, I needed to override `make()` and `__construct()` from the parent class anyways.

So I can do this:

```php
    {
        return [
            VisitActionGroup::make(record: $this->getRecord()),
            DeleteAction::make(),
        ];
    }
```

New class implementation:

```php
<?php

namespace App\Filament\Actions;

use App\Models\Site;
use Closure;
use Filament\Actions\Action;
use Filament\Actions\ActionGroup;
use Illuminate\Contracts\Database\Eloquent\Builder;
use Illuminate\Support\Facades\App;

class VisitActionGroup extends ActionGroup
{
    protected array $actions = [];

    protected Closure|array|string|null $color = 'info';

    protected mixed $record = null;

    /**
     * @param  array<StaticAction | ActionGroup>  $actions
     */
    public function __construct(array $actions, mixed $record)
    {
        $this->actions($actions);
        $this->record = $record;
    }

    /**
     * @param  array<StaticAction | ActionGroup>  $actions
     */
    public static function make(?array $actions = [], mixed $record = null): static
    {
        $static = app(static::class, ['actions' => $actions, 'record' => $record]);
        $static->configure();

        return $static;
    }

    {
        if ($this->record->slug?->slug || $this->record->type == 'home') {
            parent::setUp();

            $this->icon(Site::icon());

            $this->label(__('app/filament/actions.visit-group.label'));

            $this->button();

            $this->color($this->color);

            // TODO: Enhance query with sites that are related to the resource instance containing this action group.
            $sites = Site::withWhereHas('domains', function (Builder $query) {
                $query->where('default', true);
            })->get();

            $sites->each(function (Site $item) {
                $this->actions[] = Action::make('visit')
                    ->label($item->name)
                    ->color($this->color)
                    ->icon('heroicon-o-arrow-top-right-on-square')
                    ->url(function () use ($item): string {
                        $protocol = 'https';
                        $domain = $item->domains[0]->domain;
                        $slug = $this->record->slug->slug;

                        if (App::isLocal()) {
                            $domain = $item->localDomain();
                            $protocol = 'http';
                        }

                        return $protocol.'://'.$domain.'/'.$slug;
                    })
                    ->openUrlInNewTab();
            });

            $this->actions($this->actions);
        }
    }
}
```

Since this makes sense only for custom class implementations, I don't know if this makes sense in Filament code.

Daniel-H123:
> Well, in order to access the resource record, I needed to override `make()` and `__construct()` from the parent class anyways.

This makes this PR not necessary anymore right? I don't see why anyone would want to pass an empty array, without access to the record, and adding the record in Filament code doesn't really make sence.

carstenjaksch:
It might be a goodie for devs, who don't need (and want) to override `make()` and want to call their custom class without an empty array: `CustomActionGroup::make()`

But for me, it is not necessary anymore.

danharrin:
Yeah I think we can safely say that if you have a custom group class you can override `make()` to have a param default

PR #13615, opened 2024-07-17
ToggleFilter to use in tables

andreazanini:
## Description

Since using filament I feeled the need to have a fast usable filter on top of the tables, so instead of having selects or search for specific queries i implemented this filter.

Here an example usage

```
ToggleFilter::make('stato_assistenza_id')
->options(StatoAssistenza::pluck('stato'))
->multiple()
->grouped()
->default([0, 1, 2, 3, 4, 8, 9])
->extraAttributes(['class' => 'h-16'])
```

or not grouped in normal filter
```
ToggleFilter::make('stato_assistenza_id')
->options(StatoAssistenza::pluck('stato'))
->multiple()
->default([0, 1, 2, 3, 4, 8, 9])
```

## Visual changes

image

image

## Functional changes

I had to adapt toggle buttons class to allow extra attributes and have better control over it, since in my case some labels were two lines and the height was not adapting

### DIFF
--- a/packages/forms/resources/views/components/toggle-buttons/grouped.blade.php
+++ b/packages/forms/resources/views/components/toggle-buttons/grouped.blade.php
@@ -53,6 +53,7 @@
                 grouped
                 :icon="$getIcon($value)"
                 tag="label"
+                :attributes="$getExtraAttributeBag()"
             >
                 {{ $label }}
             </x-filament::button>
--- a/packages/forms/resources/views/components/toggle-buttons/index.blade.php
+++ b/packages/forms/resources/views/components/toggle-buttons/index.blade.php
@@ -71,6 +71,7 @@
                     :for="$inputId"
                     :icon="$getIcon($value)"
                     tag="label"
+                    :attributes="$getExtraAttributeBag()"
                 >
                     {{ $label }}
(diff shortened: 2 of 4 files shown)


danharrin:
Hi, thanks for the PR! However, I am going to decline it right now, for a few reasons
- It's a lot of extra code to maintain, which is sometimes ok but I haven't had this feature request before
- It can be achieved already using a Filter class with a `form()` quite easily and with full customisation
- It could be shipped in a plugin if you wanted to release it!

The reason why we have a `SelectFilter` is because it is so popular that basically everyone uses it, whereas I don't think I can say the same about this as I haven't had requests. I also think that a select makes use of the limited space in the filters area much better than toggle buttons do.

In the future, I would really appreciate it if you made feature requests in Discussions and `@danharrin` in them so I can give you the green light or not, otherwise you might spend lots of time like this on something that is not used. Sorry about that.

andreazanini:
Understood, will probably make a plugin. What instead of adding the extra attributes to toggle buttons? Should that be accepted if I make a different PR?

danharrin:
The extraAttributes should always be applied to the single root element in the field, not in multiple places. I would accept a PR for extraButtonAttributes()

PR #13619, opened 2024-07-17
Add option to tie resource to relation manager

Daniel-H123:
## Description

This PR will add the functionality as partially described in https://github.com/filamentphp/filament/discussions/213

It will add the option to link a resource to a relation manager, so view-, edit- and create actions can link to their respective resource pages.

## Functional changes

If the user wants to tie a resource to a relation manager, it can simply be done by defining the $resource variable inside the ResourceManager.

```php
<?php

namespace Filament\Resources\RelationManagers;

class RandomRelationManager extends RelationManager
{
    protected static string $relationship = 'randomRelation';

    protected static ?string $resource = RandomResource::class;
...
```
### Customization per action
If the user for some reason wants to use the modal instead of the page in only one of the actions, it can be specified in the Action::make() command. I first didn't want to add that variable there, but adding it as a function would become confusing, since it cannot be called after the make() function, because it needs to be executed before `$static->configure();` is called.

```php
->actions([
    Tables\Actions\ViewAction::make(), // uses view page if exists on resource
    Tables\Actions\EditAction::make(useModal: true), // uses modal
])
```

Updated Action::make() function:

```php
public static function make(?string $name = null, bool $useModal = false): static
    {
        $static = app(static::class, [
            'name' => $name ?? static::getDefaultName(),
        ]);

        if ($static instanceof MountableAction) {
            $static->preferModal($useModal);
        }

        $static->configure(); // $static->useModel needs to be set before called

        return $static;
    }
```

### Example repository
I've created an example repository of how the PR is supposed to work.
https://github.com/Daniel-H123/filament-example-relationmanager-with-resource

___

### DIFF
--- a/packages/actions/src/Concerns/CanOpenModal.php
+++ b/packages/actions/src/Concerns/CanOpenModal.php
@@ -91,6 +91,19 @@ trait CanOpenModal
      */
     protected string | array | Closure | null $modalIconColor = null;
 
+    protected bool $useModal = false;
+
+    public function preferModal($bool = false): static
+    {
+        $this->useModal = $bool;
+        return $this;
+    }
+
+    public function shouldUseModal(): bool
+    {
+        return $this->useModal;
+    }
+
     public function closeModalByClickingAway(bool | Closure | null $condition = true): static
     {
         $this->isModalClosedByClickingAway = $condition;
--- a/packages/actions/src/StaticAction.php
+++ b/packages/actions/src/StaticAction.php
@@ -60,11 +60,16 @@ class StaticAction extends ViewComponent implements Contracts\Groupable
         $this->name($name);
     }
 
-    public static function make(?string $name = null): static
+    public static function make(?string $name = null, bool $useModal = false): static
     {
         $static = app(static::class, [
             'name' => $name ?? static::getDefaultName(),
         ]);
+
+        if ($static instanceof MountableAction) {
+            $static->preferModal($useModal);
+        }
+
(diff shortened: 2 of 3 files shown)


Daniel-H123:
Documentation needs to be updated, but would like some feedback on my implementation before I start doing that.

Also `composer` cs doesn't seem to work for me, as it tries to update 3k+ files, while I only changed 3 files. 😞

danharrin:
Hi, apologies but I am not interested in this change at the moment. It's standard practice right now to be able to pass a resource URL to the URL of the action and it's about 1 line, and I don't want to have to introduce the new features such as `useModal` which I don't necessarily think are required to achieve something like this.

In v4, I have already implemented a similar feature as part of the nested resources feature, where you can link a relation manager to a `$relatedResource` and have it automatically route to that. Hopefully that solves it for you :)

Daniel-H123:
> Hi, apologies but I am not interested in this change at the moment. It's standard practice right now to be able to pass a resource URL to the URL of the action and it's about 1 line, and I don't want to have to introduce the new features such as `useModal` which I don't necessarily think are required to achieve something like this.
> 
> In v4, I have already implemented a similar feature as part of the nested resources feature, where you can link a relation manager to a `$relatedResource` and have it automatically route to that. Hopefully that solves it for you :)

No worries. Didn't really came to the solution of adding the url function, but that would work perfectly fine in my usecase.

Nice to see it added in v4 👍🏼

PR #13620, opened 2024-07-17
Fix array_reduce in new groups closure

archilex:
Fixes array_reduce to use `$carry` in #13559. It's only returning one group now. 


danharrin:
Good catch, thanks!

PR #13624, opened 2024-07-18
Prevent resetting action record

tanthammar:
## Description

creds: @joshhanley

### Mountable actions:
This PR ensures that, if an action's record is dynamically assigned, it does not revert to its initial value.

## Example
You define a header action to issue refunds on an `InvoiceResource` page.
```php
class ManageInvoice extends ViewRecord

{
    return [
        RefundInvoiceAction::make(),
    ];
}
```

You declare additional modal actions where you will need to dynamically assign the record.
```php
Action::make('refundInvoice')
    ->registerModalActions([
        RefundStripePaymentAction::make()->record(see below),
    ])
```

This `$action->record` and its "sub-actions", are by default the `Invoice` model

On the `ManageInvoice` page you loop through the `$invoice->payments` and you pass the `Payment` to the refund action.

```blade
@foreach ($this->record->payments as $payment)
     {{ $action->getModalAction('refundStripePayment')(['payment_id' => $payment->id]) }}
@endforeach
```

You dynamically set the action's record using the passed payment id
```php
RefundStripePaymentAction::make()->record(fn ($arguments) => Payment::findOrFail($arguments['payment_id'])),
```

### Problem
The **prepareModalAction()** reverts the actions record back to the `$invoice`.
```php
public function prepareModalAction(StaticAction $action): StaticAction
{
    //...
    if (
        ($this instanceof HasRecord) &&
        ($action instanceof HasRecord)
    ) {
        $action->record($this->getRecord()); //<-------------------- back to the $invoice again
    }
    //...
}
```

### Fix
This modification ensures that, if an action's record is dynamically assigned, it does not revert to its initial value.
```php
public function prepareModalAction(StaticAction $action): StaticAction
{
    //...
    if (
        ($this instanceof HasRecord) &&
        ($action instanceof HasRecord) &&
        invade($action)->record === null //<------------------------prevent reset
    ) {
        $action->record($this->getRecord());
    }
    //...
}
```

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/CanOpenModal.php
+++ b/packages/actions/src/Concerns/CanOpenModal.php
@@ -434,7 +434,8 @@ trait CanOpenModal
 
         if (
             ($this instanceof HasRecord) &&
-            ($action instanceof HasRecord)
+            ($action instanceof HasRecord) &&
+            invade($action)->record === null
         ) {
             $action->record($this->getRecord());
         }
--- a/packages/actions/src/Concerns/CanOpenModal.php
+++ b/packages/actions/src/Concerns/CanOpenModal.php
@@ -435,7 +435,7 @@ trait CanOpenModal
         if (
             ($this instanceof HasRecord) &&
             ($action instanceof HasRecord) &&
-            invade($action)->record === null
+            ! $action->hasRecord()
         ) {
             $action->record($this->getRecord());
         }
--- a/packages/actions/src/Concerns/CanOpenModal.php
+++ b/packages/actions/src/Concerns/CanOpenModal.php
@@ -435,7 +435,7 @@ trait CanOpenModal
         if (
             ($this instanceof HasRecord) &&
             ($action instanceof HasRecord) &&
-            ! $action->hasRecord()
+            (! $action->hasRecord())
         ) {
             $action->record($this->getRecord());
         }
--- a/packages/tables/src/Columns/Concerns/HasRecord.php
+++ b/packages/tables/src/Columns/Concerns/HasRecord.php
@@ -19,4 +19,9 @@ trait HasRecord
     {
         return $this->record ?? $this->getLayout()?->getRecord();
     }
+
+    public function hasRecord(): bool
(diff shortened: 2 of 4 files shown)


packages/actions/src/Concerns/CanOpenModal.php
```diff
@@ -434,7 +434,8 @@ public function prepareModalAction(StaticAction $action): StaticAction
 
         if (
             ($this instanceof HasRecord) &&
-            ($action instanceof HasRecord)
+            ($action instanceof HasRecord) &&
+            invade($action)->record === null
```
danharrin: Please introduce a new method called `hasRecord()` in InteractsWithRecord and HasRecord, which returns true if the record is not null. So that we do not need to invade the object

danharrin (changes requested):

packages/actions/src/Concerns/CanOpenModal.php
```diff
@@ -434,7 +434,8 @@ public function prepareModalAction(StaticAction $action): StaticAction
 
         if (
             ($this instanceof HasRecord) &&
-            ($action instanceof HasRecord)
+            ($action instanceof HasRecord) &&
+            ! $action->hasRecord()
```
danharrin:
```suggestion
            (! $action->hasRecord())
```

packages/tables/src/Columns/Concerns/HasRecord.php
```diff
@@ -19,4 +19,9 @@ public function getRecord(): ?Model
     {
         return $this->record ?? $this->getLayout()?->getRecord();
     }
+
+    public function hasRecord(): bool
+    {
+        return $this->record !== null;
+    }
```
danharrin: Please move this to the `Filament\Actions\Contracts\HasRecord` interface, it just needs the method signature not the implementation

danharrin (changes requested):

danharrin:
Thanks

PR #13634, opened 2024-07-18
Proposal: Add `filament:optimize` and `filament:optimize-clear` commands

dissto:
## Description

This PR adds two new commands under the filament namespace: 
- `php artisan filament:optimize`
- `php artisan filament:optimize-clear`

Its just a wrapper for the 2 recommended commands `icons:cache` and `filament:cache-components`

## Visual changes

https://github.com/user-attachments/assets/624a1015-629c-4d7b-a084-12bd83ea8470

## Functional changes


packages/support/src/Commands/OptimizeCommand.php
```diff
@@ -0,0 +1,28 @@
...
+
+use Illuminate\Console\Command;
+use Symfony\Component\Console\Attribute\AsCommand;
+
+#[AsCommand(name: 'filament:optimize')]
+class OptimizeCommand extends Command
+{
+    protected $signature = 'filament:optimize';
+
+    protected $description = 'Cache components and blade icons to increase performance.';
```
danharrin:
```suggestion
    protected $description = 'Cache components and Blade icons to improve performance.';
```

packages/support/src/Commands/OptimizeCommand.php
```diff
@@ -0,0 +1,28 @@
...
+
+    public function handle(): int
+    {
+        $this->components->info('Caching components and blade icons.');
```
danharrin:
```suggestion
        $this->components->info('Caching components and Blade icons...');
```

packages/support/src/Commands/OptimizeCommand.php
```diff
@@ -0,0 +1,28 @@
...
+
+        collect([
+            'Caching components' => fn () => $this->callSilent('filament:cache-components') == 0,
+            'Caching blade icons' => fn () => $this->callSilent('icons:cache') == 0,
```
danharrin:
```suggestion
            'Caching Blade icons' => fn (): bool => $this->callSilent('icons:cache') === static::SUCCESS,
```

packages/support/src/Commands/OptimizeCommand.php
```diff
@@ -0,0 +1,28 @@
...
+    protected $signature = 'filament:optimize';
+
+    protected $description = 'Cache components and blade icons to increase performance.';
+
+    public function handle(): int
+    {
+        $this->components->info('Caching components and blade icons.');
+
+        collect([
+            'Caching components' => fn () => $this->callSilent('filament:cache-components') == 0,
```
danharrin:
```suggestion
            'Caching components' => fn (): bool => $this->callSilent('filament:cache-components') === static::SUCCESS,
```

packages/support/src/Commands/OptimizeClearCommand.php
```diff
@@ -0,0 +1,28 @@
...
+
+use Illuminate\Console\Command;
+use Symfony\Component\Console\Attribute\AsCommand;
+
+#[AsCommand(name: 'filament:optimize-clear')]
+class OptimizeClearCommand extends Command
+{
+    protected $signature = 'filament:optimize-clear';
+
+    protected $description = 'Remove cached components and blade icons.';
```
danharrin:
```suggestion
    protected $description = 'Remove cached components and Blade icons.';
```

packages/support/src/Commands/OptimizeClearCommand.php
```diff
@@ -0,0 +1,28 @@
...
+
+    public function handle(): int
+    {
+        $this->components->info('Clearing cached components and blade icons.');
```
danharrin:
```suggestion
        $this->components->info('Clearing cached components and Blade icons...');
```

packages/support/src/Commands/OptimizeClearCommand.php
```diff
@@ -0,0 +1,28 @@
...
+
+        collect([
+            'Clearing cached Components' => fn () => $this->callSilent('filament:clear-cached-components') == 0,
```
danharrin:
```suggestion
            'Clearing cached components' => fn (): bool => $this->callSilent('filament:clear-cached-components') === static::SUCCESS,
```

packages/support/src/Commands/OptimizeClearCommand.php
```diff
@@ -0,0 +1,28 @@
...
+            'Clearing cached Blade Icons' => fn () => $this->callSilent('icons:clear') == 0,
```
danharrin:
```suggestion
            'Clearing cached Blade icons' => fn (): bool => $this->callSilent('icons:clear') === static::SUCCESS,
```

danharrin (changes requested):
Please add a check around the commands that cache components / clear cached components, so if the panel builder is not installed it only handles the icons. Otherwise the command will fail if the user runs this while not having the panel builder installed, since it is just part of the support package.

dissto:
> Please add a check around the commands that cache components / clear cached components, so if the panel builder is not installed it only handles the icons. Otherwise the command will fail if the user runs this while not having the panel builder installed, since it is just part of the support package.

I have borrowed the [check from the install command](https://github.com/filamentphp/filament/blob/3.x/packages/support/src/Commands/InstallCommand.php#L70). Or did you have something different in mind? 

I additionally made that check into a Trait and applied all your suggestions.🤓

packages/support/src/Concerns/CanCacheComponents.php
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Filament\Support\Concerns;
+
+use Filament\PanelProvider;
+
+trait CanCacheComponents
+{
+    protected function canCacheComponents(): bool
+    {
+        return class_exists(PanelProvider::class);
```
danharrin: Please instead check to see if the cache-components command class exists

packages/support/src/Concerns/CanCacheComponents.php
danharrin, line 1: Please move this into `Filament\Support\Commands\Concerns`

danharrin (changes requested):

packages/support/src/Concerns/CanCacheComponents.php
danharrin: Please rename the trait and method to `CanCachePanelComponents`

danharrin (changes requested):

dissto:
Applied the latest suggestions 🤓

danharrin (changes requested):
Please swap it back so that the components are cached/cleared first before the icons so its consistent with the messages and docs

dissto:
> Please swap it back so that the components are cached/cleared first before the icons so its consistent with the messages and docs

Got it.

danharrin:
Thanks

PR #13653, opened 2024-07-21
Allowing $get() to get full array of current path

cheesegrits:
## Description

Allow $get() with no args, or $get('') or $get('./'), to return the entire data array for the current path.  This is particularly useful when dealing with repeaters or builders.

Note that this was discussed with Dan in another version of this PR that I closed because I somehow managed to completely screw it up with a bunch of extraneous commits.

Also note that this usage it not restricted to repeaters, it'll always return the data array for the current path regardless, I just couldn't find anywhere else to document the changes, so the documentation is in the repeater / builder sections.

## Visual changes

No visual changes.

## Functional changes

### DIFF
--- a/packages/forms/docs/03-fields/12-repeater.md
+++ b/packages/forms/docs/03-fields/12-repeater.md
@@ -475,7 +475,7 @@ You are trying to retrieve the value of `client_id` from inside the repeater ite
 
 You can use `../` to go up a level in the data structure, so `$get('../client_id')` is `$get('repeater.client_id')` and `$get('../../client_id')` is `$get('client_id')`.
 
-The special case of `$get()` with no arguments, or `$get('')` or `$get('./')`, will always return the full data array for the current path.
+The special case of `$get()` with no arguments, or `$get('')` or `$get('./')`, will always return the full data array for the current repeater item.
 
 ## Repeater validation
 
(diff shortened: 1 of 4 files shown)


packages/forms/docs/03-fields/12-repeater.md
```diff
@@ -475,6 +475,8 @@ You are trying to retrieve the value of `client_id` from inside the repeater ite
 
 You can use `../` to go up a level in the data structure, so `$get('../client_id')` is `$get('repeater.client_id')` and `$get('../../client_id')` is `$get('client_id')`.
 
+The special case of `$get()` with no arguments, or `$get('')` or `$get('./')`, will always return the full data array for the current path.
```
danharrin:
```suggestion
The special case of `$get()` with no arguments, or `$get('')` or `$get('./')`, will always return the full data array for the current repeater item.
```

packages/forms/docs/03-fields/13-builder.md
```diff
@@ -368,6 +368,36 @@ Builder::make('content')
...
+]
+```
+
+You are trying to retrieve the value of `client_id` from inside the builder item.
+
+`$get()` is relative to the current builder item, so `$get('client_id')` is looking for `$get('builder.item1.client_id')`.
+
+You can use `../` to go up a level in the data structure, so `$get('../client_id')` is `$get('builder.client_id')` and `$get('../../client_id')` is `$get('client_id')`.
+
+The special case of `$get()` with no arguments, or `$get('')` or `$get('./')`, will always return the full data array for the current path.
```
danharrin:
```suggestion
The special case of `$get()` with no arguments, or `$get('')` or `$get('./')`, will always return the full data array for the current builder item.
```

danharrin (changes requested):

danharrin:
Thanks!

PR #13671, opened 2024-07-23
Show tooltip for disabled action button

ericmp33:
## Description

We would like to show a tooltip on a disabled action button to explain why that button is disabled. https://github.com/filamentphp/filament/discussions/4271

One thing to note is that this change might be an unexpected behavior for developers who previously had tooltips configured to display only when actions were enabled. We should consider mentioning this update in the `What's Changed` section of the release notes to avoid confusion.

## Visual changes

If the action has a tooltip, despite being disabled, the tooltip will be shown.

## Functional changes

### DIFF
--- a/packages/actions/resources/views/components/action.blade.php
+++ b/packages/actions/resources/views/components/action.blade.php
@@ -7,31 +7,35 @@
 @php
     $isDisabled = $action->isDisabled();
     $url = $action->getUrl();
+    $tooltip = $action->getTooltip();
+    $tooltipIsNotEmpty = str($tooltip)->isNotEmpty();
 @endphp
 
-<x-dynamic-component
-    :color="$action->getColor()"
-    :component="$dynamicComponent"
-    :disabled="$isDisabled"
-    :form="$action->getFormToSubmit()"
-    :form-id="$action->getFormId()"
-    :href="$isDisabled ? null : $url"
-    :icon="$icon ?? $action->getIcon()"
-    :icon-size="$action->getIconSize()"
-    :key-bindings="$action->getKeyBindings()"
-    :label-sr-only="$action->isLabelHidden()"
-    :tag="$url ? 'a' : 'button'"
-    :target="($url && $action->shouldOpenUrlInNewTab()) ? '_blank' : null"
-    :tooltip="$action->getTooltip()"
-    :type="$action->canSubmitForm() ? 'submit' : 'button'"
-    :wire:click="$action->getLivewireClickHandler()"
-    :wire:target="$action->getLivewireTarget()"
-    :x-on:click="$action->getAlpineClickHandler()"
-    :attributes="
-        \Filament\Support\prepare_inherited_attributes($attributes)
-            ->merge($action->getExtraAttributes(), escape: false)
-            ->class(['fi-ac-action'])
-    "
->
-    {{ $slot }}
-</x-dynamic-component>
+@if ($tooltipIsNotEmpty) <div x-data x-tooltip.raw="{{ $tooltip }}"> @endif
+    <x-dynamic-component
+        :color="$action->getColor()"
+        :component="$dynamicComponent"
+        :disabled="$isDisabled"
+        :form="$action->getFormToSubmit()"
+        :form-id="$action->getFormId()"
+        :href="$isDisabled ? null : $url"
+        :icon="$icon ?? $action->getIcon()"
+        :icon-size="$action->getIconSize()"
+        :key-bindings="$action->getKeyBindings()"
+        :label-sr-only="$action->isLabelHidden()"
+        :tag="$url ? 'a' : 'button'"
+        :target="($url && $action->shouldOpenUrlInNewTab()) ? '_blank' : null"
+        :tooltip="$tooltip"
+        :type="$action->canSubmitForm() ? 'submit' : 'button'"
+        :wire:click="$action->getLivewireClickHandler()"
+        :wire:target="$action->getLivewireTarget()"
+        :x-on:click="$action->getAlpineClickHandler()"
+        :attributes="
+            \Filament\Support\prepare_inherited_attributes($attributes)
+                ->merge($action->getExtraAttributes(), escape: false)
+                ->class(['fi-ac-action'])
+        "
+    >
+        {{ $slot }}
+    </x-dynamic-component>
+@if ($tooltipIsNotEmpty) </div> @endif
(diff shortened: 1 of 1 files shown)


danharrin:
Please test this by adding two buttons inside `<div class="grid grid-cols-2">` and `<div class="flex justify-end">` and `<div class="flex justify-between">`. Please screenshot each case and post them here, so we can review if there are any impacts of having a wrapper.

packages/actions/resources/views/components/action.blade.php
```diff
@@ -7,31 +7,35 @@
...
+        :form="$action->getFormToSubmit()"
+        :form-id="$action->getFormId()"
+        :href="$isDisabled ? null : $url"
+        :icon="$icon ?? $action->getIcon()"
+        :icon-size="$action->getIconSize()"
+        :key-bindings="$action->getKeyBindings()"
+        :label-sr-only="$action->isLabelHidden()"
+        :tag="$url ? 'a' : 'button'"
+        :target="($url && $action->shouldOpenUrlInNewTab()) ? '_blank' : null"
+        :tooltip="$tooltip"
```
wychoong: If it’s not disabled it will be 2 tooltips?

ericmp33: oops! you are right. let me fix it!

ericmp33:
> Please test this by adding two buttons inside `<div class="grid grid-cols-2">` and `<div class="flex justify-end">` and `<div class="flex justify-between">`. Please screenshot each case and post them here, so we can review if there are any impacts of having a wrapper.

I created 3 different full page lw components with 2 action buttons per each. Both actions have tooltip, but one action is enabled and the other one is disabled (I also tried disabling it both actions, results look the same).

The class part looks the same for all 3. The view part changes on the classes used.

@danharrin is this what you expected me to try out?

https://github.com/user-attachments/assets/d768d216-4117-464f-8a12-db8530963b7d

Component classes:
```php
<?php

namespace App\Livewire;

use App\View\Components\AppLayout;
use Filament\Forms\Components\Actions;
use Filament\Forms\Form;
use Livewire\Component;

class Test1 extends Component implements HasForms
{
    use InteractsWithForms;

    public function render(): View
    {
        return view('livewire.test1')->layout(AppLayout::class);
    }

    public function form(Form $form): Form
    {
        return $form->schema([
            Actions::make([
                Action::make('star')
                    ->icon('heroicon-m-star')
                    ->tooltip('First tooltip'),
                Action::make('resetStars')
                    ->icon('heroicon-m-x-mark')
                    ->color('danger')
                    ->tooltip('Second tooltip')
                    ->disabled(),
            ]),
        ]);
    }
}
```
Component views:
```html
// test 1
<div class="grid grid-cols-2">
    {{ $this->form }}
</div>
```
```html
// test 2
<div class="flex justify-end">
    {{ $this->form }}
</div>
```
```html
// test 3
<div class="flex justify-between">
    {{ $this->form }}
</div>
```

wychoong:
> is this what you expected

Probably is to directly using actions https://filamentphp.com/docs/3.x/actions/adding-an-action-to-a-livewire-component#adding-the-action

ericmp33:
> > is this what you expected
> 
> Probably is to directly using actions https://filamentphp.com/docs/3.x/actions/adding-an-action-to-a-livewire-component#adding-the-action

Oh yeah I see. Okay, I test it again.

ericmp33:
New tests:

https://github.com/user-attachments/assets/d5230dc9-1d40-420e-b183-2e566728548c


Classes:
```php
<?php

namespace App\Livewire;

use App\View\Components\AppLayout;
use Filament\Actions\Concerns\InteractsWithActions;
use Filament\Actions\Contracts\HasActions;
use Filament\Forms\Components\Actions;
use Filament\Actions\Action;
use Filament\Forms\Form;
use Livewire\Component;

class Test1 extends Component implements HasForms, HasActions
{
    use InteractsWithForms;
    use InteractsWithActions;

    public function render(): View
    {
        return view('livewire.test1')->layout(AppLayout::class);
    }

    public function oneAction(): Action
    {
        return Action::make('one')
            ->disabled()
            ->tooltip('One');
    }

    public function twoAction(): Action
    {
        return Action::make('two')
            // ->disabled()
            ->tooltip('Two');
    }
}
```

Views:
```html
// test 1
<div class="grid grid-cols-2">
    {{ $this->oneAction }}
    {{ $this->twoAction }}

    <x-filament-actions::modals />
</div>
```

```html
// test 2
<div class="flex justify-end">
    {{ $this->oneAction }}
    {{ $this->twoAction }}

    <x-filament-actions::modals />
</div>
```

```html
// test 3
<div class="flex justify-between">
    {{ $this->oneAction }}
    {{ $this->twoAction }}

    <x-filament-actions::modals />
</div>
```

packages/actions/resources/views/components/action.blade.php
```diff
@@ -7,31 +7,34 @@
...
-</x-dynamic-component>
+@if ($tooltipIsNotEmpty) <div x-data x-tooltip.raw="{{ $tooltip }}" class="flex"> @endif
+    <x-dynamic-component
+        :color="$action->getColor()"
+        :component="$dynamicComponent"
+        :disabled="$isDisabled"
+        :form="$action->getFormToSubmit()"
+        :form-id="$action->getFormId()"
+        :href="$isDisabled ? null : $url"
+        :icon="$icon ?? $action->getIcon()"
```
saumanarsad: What will be the scenario where $icon is null?

ericmp33: checkout the video [here](https://github.com/filamentphp/filament/pull/13671#issuecomment-2248645952), where icon is null in all btns. or u mean something else?

zepfietje:
Could you please do the same video for icon button actions, @ericmp33?
Currently those have a negative margin to account for spacing, which this PR might break.
Not sure though, hence I'm asking for the video :)

ericmp33:
I updated the tests & fixed the code. Here you have a table to know what functions I used for both actions in each test.
|  | `->icon('heroicon-s-pencil')` | `->iconButton()` |
| ------------- | ------------- | ------------- |
| TestX  | `false`  | `false`  |
| TestXb | `true`  | `false`  |
| TestXc | `true`  | `true`  |

*For each test, the 1st action is always disabled and the 2nd it's not.

Before:

https://github.com/user-attachments/assets/aeaca375-2afc-4ff0-8934-42e0f58694c6


After:

https://github.com/user-attachments/assets/77a8106f-fd40-4645-8e27-aafcd3911500

danharrin:
Think I'm going to push this to v4 since it does have some capability to be a breaking change with certain style customizations

PR #13677, opened 2024-07-24
fix: get correct primary key name from media class

LuizCristino:
## Description

When user changes the media model, he has the possibility of changing the model primary key, this line will cause an error if the primary key is any different of `id`

## Visual changes

There are none.

## Functional changes


danharrin:
Thanks

PR #13680, opened 2024-07-24
Update 13-builder.md rename view path

bansal:
The path used in preview method is not same as the resource path.

## Description

There is incorrect path in documentation.

## Visual changes

Screenshot 2024-07-24 at 11 06 15 AM

## Functional changes


danharrin:
Good catch, thanks

PR #13707, opened 2024-07-25
Add Filament:generateResourceUrl() method

geisi:
Hey Filament team! 👋

First, I want to say how much I love working with Filament. You've created an amazing tool, and I'm excited to contribute to the project for the first time.

I've been working on a little DX improvement that could make life easier for Filament developers. It's all about generating URLs for Resources more smoothly.

I constantly typed out `SomeResource::getUrl()` with similar parameters repeatedly. It got me thinking - what if we could simplify this?

So, I came up with a new method: `Filament::getResourceUrl()`. The cool thing is that you can pass in a model instance or class, and it figures out the rest. Here's a quick before and after:

Before:
```PHP
$post = Post::first();
PostResource::getUrl('view', ['record' => $post]);
PostResource::getUrl('edit', ['record' => $post]);
PostResource::getUrl('index');
```
After:
```PHP
$post = Post::first();
Filament::getResourceUrl($post, 'view');
Filament::getResourceUrl($post, 'edit');
Filament::getResourceUrl(Post::class, 'index');
```
I think this could be a nice win for a couple of reasons:

1. It's way easier to remember and type, especially when you're in the flow of working with your models.
2. You don't have to juggle those `['record' => $post]` parameters for editing and viewing routes anymore.

This change is not a breaking change. It's just a new option that works well with all the existing stuff.

I've added some tests to ensure everything works as expected, and I'd be happy to update the docs if you think this is worth including.

What do you think? I'm open to feedback or suggestions if you see ways to improve this.

Thanks for checking this out!
## Functional changes

### DIFF
--- a/packages/panels/src/Facades/Filament.php
+++ b/packages/panels/src/Facades/Filament.php
@@ -71,7 +71,7 @@ use Illuminate\Support\Facades\Facade;
  * @method static string | null getRequestPasswordResetUrl(array $parameters = [])
  * @method static string getResetPasswordUrl(string $token, CanResetPassword | Model | Authenticatable $user, array $parameters = [])
  * @method static array getResources()
- * @method static array getResourceUrl(string | Model $model, string $name = 'index', array $parameters = [], bool $isAbsolute = false, ?string $panel = null, ?Model $tenant = null)
+ * @method static array getResourceUrl(string | Model $model, string $name = 'index', array $parameters = [], bool $isAbsolute = false, ?Model $tenant = null)
  * @method static string getSidebarWidth()
  * @method static Model | null getTenant()
  * @method static string | null getTenantAvatarUrl(Model $tenant)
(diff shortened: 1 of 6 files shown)


Jonas-johansen:
That's pretty neat

Wondering what happens when you have multiple resources for the same model though? 
Personally i tend to have a admin resource and a tenant/customer resource for the same model, on different panels.

geisi:
Hey Jonas good point this is already handled by this PR. If you have multiple panels you have to pass the panel in this case.

Otherwise it would throw an exception "Multiple Resources found for the model".

danharrin (changes requested):
Hey! Thanks for your contribution.

Please be aware that due to the nature of looping through each resource to check its model, this is going to be less performant than the original method.

Please refactor this code to belong in the Panel class. You can make use of the `getModelResource()` method which already finds a resource for a model.

The FilamentManager can `getCurrentPanel()` and then forward the call onto the panel as per the other methods.

Please make sure to update the docblock in the Filament facade with the new method.

geisi:
@danharrin Thanks for your review. I've refactored the `getResourceUrl` method and added the Facade doc block as requested.

Regarding moving this to the Panel class, I'm not sure I see the benefit. It would result in a more verbose syntax:

```PHP
Filament::getCurrentPanel()->getResourceUrl($post,'edit');
```

Could you explain the reasoning behind this suggestion? I'm open to making this change.

danharrin:
The method on the facade can still exist, but it proxies to the method on the panel. This is so you can still generate a resource URL for another panel even if that panel is not active.

geisi:
@danharrin Thanks for clarifying. Now I got it. I have applied the changes.

danharrin:
Thanks!

PR #13715, opened 2024-07-26
fix: missed header column with grouped only list

nazarpashkevich:
## Description

I faced a problem when displaying columns in the table when using grouping. If only grouped records are displayed in the table (->groupsOnly()) then one column is lost in the header of the table (in the record it`s a grouping column), and all column names move 1 cell to the left.

## Visual changes
before:
image

after:
image

## Functional changes

I tested this all on my project on our cases, whether it can break the functionality - I think you guys know better :)

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -833,6 +833,10 @@
                             @endif
                         @endif
 
+                        @if ($isGroupsOnly)
+                            <x-filament-tables::header-cell>{{ $group->getLabel() }}</x-filament-tables::header-cell>
+                        @endif
+
                         @foreach ($columns as $column)
                             @php
                                 $columnWidth = $column->getWidth();
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -834,7 +834,9 @@
                         @endif
 
                         @if ($isGroupsOnly)
-                            <x-filament-tables::header-cell>{{ $group->getLabel() }}</x-filament-tables::header-cell>
+                            <x-filament-tables::header-cell>
+                                {{ $group->getLabel() }}
+                            </x-filament-tables::header-cell>
                         @endif
 
                         @foreach ($columns as $column)


cheesegrits:
@nazarpashkevich hey, Dan asked me to take a look at this (I sponsored the addition of the Grouping and Summaries features, and use them heavily).  I can't replicate what you are seeing, and if I apply the change in this PR, I just get the grouped column header showing twice.

Can you show me the full table schema you are using for this?  Or maybe provide a repro repo I can try (the Filament demo site is a great place to start for creating repros).

danharrin:
Gonna close this for now but we can reopen if its reproducable

nazarpashkevich:
@cheesegrits, hi, I`ve created a fresh project with the same table structure, please take a look https://github.com/nazarpashkevich/filament-test, and there is the same problem, I don't know, maybe it has to do with the specialty of my fields structure. Thank you
image

cheesegrits:
@nazarpashkevich OK, I'll take a look as soon as I have time.

I'm sure you are seeing this problem, the issue is that the fix you implemented breaks other usages of grouping, so I need to nail down exactly what combination of settings are causing your issue, and a way to fix it that doesn't break other combinations.

cheesegrits:
@nazarpashkevich The sail config doesn't work.

Can you either redo it without sail (using sqlite), or build the Sail config / .env assuming that the person installing it already has stuff running on the standard ports for web, mysql and vite.

I tried adding APP_PORT, FORWARD_DB_PORT and VITE_PORT in the .env using unused ports on my machine, but the 'sail artisan migrate' fails with "php_network_getaddresses: getaddrinfo for mysql failed", and I don't have time to debug Sail issues.

cheesegrits:
I got it going.  Working on it.

cheesegrits:
OK, I've confirmed the issue, but the fix in this PR doesn't work for all usages.  I've also found a different but related problem, when using multiple groupings, which is actually affected my deployed sites - I don't know if it's always been like that, or if its a regression.

I'll work on this as I have time this week.

nazarpashkevich:
thanks, glad we got it figured out, if there's anything I can do to help, let me know

PR #13718, opened 2024-07-26
Proposal: add `brandLogoInTopBar()` method to `Panel` component

wickham-c:
## Description

Adding a `brandLogoInTopBar()` method to the `Panel` component is proposed, which has the effect of hiding the brand logo in the sidebar and instead showing it in the topbar (after the "Expand sidebar" & "Collapse sidebar" buttons). 

This has the benefit of **always displaying the brand logo in the topbar**, even when the sidebar is collapsed.

The proposed approach is preferable to using the...
- `PanelsRenderHook::TOPBAR_START` render hook, as the brand logo then appears to the left of the "Expand sidebar" button when the sidebar is collapsed, which looks strange.
- `PanelsRenderHook::GLOBAL_SEARCH_BEFORE` render hook, as the brand logo then appears right-aligned on the topbar, to the left of the global search container / user menu, which also looks strange.

### _Before (without calling proposed method)_
```php
$panel
    ->brandLogo(fn () => view('logos.logo'))
    ->sidebarFullyCollapsibleOnDesktop()
```
_...with sidebar expanded..._

image

_...with sidebar collapsed..._

image

### _After (calling proposed method)_
```php
$panel
    ->brandLogo(fn () => view('logos.logo'))
    ->brandLogoInTopBar()
    ->sidebarFullyCollapsibleOnDesktop()
```
_...with sidebar expanded..._

image

_...with sidebar collapsed..._

image

## Functional changes

When proposed `brandLogoInTopBar()` method is employed, the brand logo is always displayed, regardless of whether the sidebar is collapsed or not.

- [x] Code style has been _checked_ by running the _`./vendor/bin/pint --test` command on each of the individual files changed. The two modified Blade templates passed. For the other two files, Pint suggested a trivial "fix" of removing spaces between alternative function argument types and function return types, which seems less readable so suggested action not taken._
- [ ] Documentation is up-to-date. _Will submit documentation changes if the proposal is approved._

### DIFF
--- a/packages/panels/resources/views/components/sidebar/index.blade.php
+++ b/packages/panels/resources/views/components/sidebar/index.blade.php
@@ -53,12 +53,14 @@
                     x-transition:enter-end="opacity-100"
                 @endif
             >
-                @if ($homeUrl = filament()->getHomeUrl())
-                    <a {{ \Filament\Support\generate_href_html($homeUrl) }}>
+                @if (! filament()->hasBrandLogoInTopBar())
+                    @if ($homeUrl = filament()->getHomeUrl())
+                        <a {{ \Filament\Support\generate_href_html($homeUrl) }}>
+                            <x-filament-panels::logo />
+                        </a>
+                    @else
                         <x-filament-panels::logo />
-                    </a>
-                @else
-                    <x-filament-panels::logo />
+                    @endif
                 @endif
             </div>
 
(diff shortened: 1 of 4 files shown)


danharrin:
Hi, apologies but I am not a huge fan of how this looks when the sidebar is expanded. In the future, it would be great if you could open a discussion and @danharrin before implementing a feature request, just to make sure we agree that it is a good idea and to avoid any wasted time on your part. Many thanks for the contribution

PR #13630, opened 2024-07-18
Fix: alpine errors when textarea is in modals

awcodes:
## Description

Resolves #13621 by adding x-ignore to the blade component to prevent alpine from initializing before the alpine component has been loaded.

## Visual changes

none.

## Functional changes

- [ X] Code style has been fixed by running the `composer cs` command.
- [ X] Changes have been tested to not break existing functionality.
- [X ] Documentation is up-to-date.

PR #13722, opened 2024-07-27
Revert "Fix: alpine errors when textarea is in modals"

zepfietje:
Reverts filamentphp/filament#13630

The original PR broke autosizing textareas across Livewire requests.
On every Livewire request, the textarea goes back to its initial size.

Linking to the PR that initially fixed autosizing: https://github.com/filamentphp/filament/pull/13211.
`x-ignore` seems to have been removed there.

I'd like to merge this PR that reverts the breaking fix by @awcodes.
Do you have any ideas for a definite fix we can introduce next?

### DIFF
--- a/packages/forms/resources/views/components/textarea.blade.php
+++ b/packages/forms/resources/views/components/textarea.blade.php
@@ -34,7 +34,6 @@
         "
     >
         <textarea
-            x-ignore
             @if (FilamentView::hasSpaMode())
                 ax-load="visible"
             @else


awcodes:
As long as it's an async alpine component we are going to have to have the x-ignore. Without it alpine tries to initialize before the script can be loaded, resulting in alpine breaking down completely and preventing it from executing any further methods.

awcodes:
Now that I'm aware of the other issue I'll see what I can think of.

zepfietje:
Perfect, let me know once you've found a potential fix, @awcodes.

awcodes:
@zepfietje I think the only way we can do this, would be to entangle the state and switch to `x-model` instead of `wire:model` then call the resize method on state change with a watcher.

Appears to be working for me with this approach.

Should I push changes to this PR for you to review?

zepfietje:
Hmm, how would that solve this issue?
The textarea currently reverts to its initial size after any Livewire request, not input.

awcodes:
Let me run some more tests. I just recreated what Dan did in his video on the other PR.

danharrin:
This PR definitely should be merged, as Adam said the `x-ignore` is required for all Async Alpine components

zepfietje:
@awcodes let me know once you've got a PR up with your fix.
Then I'll test if it fixes the issue in my case.

danharrin:
As per #13748

PR #13735, opened 2024-07-29
fix: choicesjs toggle disable/enable

wychoong:
## Description

this fixes an issue where a Select is disabled based on another field
```php
Forms\Components\Toggle::make('toggle')
    ->live(),
Forms\Components\Select::make('quantities')
    ->disabled(fn ($get) => ! $get('toggle'))
    ->native(false)
    ->options(function ($get) {
        if (! $get('toggle')) {
            return [1 => 1];
        }

        return range(1, 10);
    }),
```

the issue is when a disabled select is changed to enabled later, it doesn't remove the `disabled` attribute from the select element.

this PR fixes it by wrapping it in a div container to get the `disabled` property
```php
x-data="{
    disabled: @js($isDisabled),
    init(){
        const container = $el.nextElementSibling
        container.dispatchEvent(new CustomEvent('toggle-state', { detail : { disabled: this.disabled } }))
    }
}"
```

```php
  x-on:toggle-state="$event.detail.disabled ? select.disable() : select.enable()"
```
this listener uses the ChoicesJs instance to toggle enable/disable, so that when click on the dropdown it triggers Choices event
https://github.com/filamentphp/filament/blob/e90998fe6aa4681569f09b6ebb01e5346d3a1644/packages/forms/resources/js/components/select.js#L91-L104

## Visual changes

N/A

## Functional changes

### DIFF
--- a/packages/forms/resources/views/components/select.blade.php
+++ b/packages/forms/resources/views/components/select.blade.php
@@ -98,10 +98,14 @@
             <div
                 x-data="{
                     disabled: @js($isDisabled),
-                    init(){
+                    init() {
                         const container = $el.querySelector('div[data-select-container]')
-                        container.dispatchEvent(new CustomEvent('toggle-state', { detail : { disabled: this.disabled } }))
-                    }
+                        container.dispatchEvent(
+                            new CustomEvent('toggle-state', {
+                                detail: { disabled: this.disabled },
+                            }),
+                        )
+                    },
                 }"
             >
                 <div
--- a/packages/forms/resources/views/components/select.blade.php
+++ b/packages/forms/resources/views/components/select.blade.php
@@ -108,8 +108,7 @@
                         )
                     },
                 }"
-            >
-            </div>
+            ></div>
             <div
                 x-ignore
                 @if (FilamentView::hasSpaMode())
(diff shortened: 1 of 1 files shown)


packages/forms/resources/views/components/select.blade.php
```diff
@@ -95,6 +95,20 @@
...
             </x-filament::input.select>
         @else
+            <div
+                class="hidden"
+                x-data="{
+                    disabled: @js($isDisabled),
+                    init() {
+                        const container = $el.nextElementSibling
+                        container.dispatchEvent(
+                            new CustomEvent('toggle-state', {
```
danharrin: Please rename the event to `set-select-property`

packages/forms/resources/views/components/select.blade.php
```diff
@@ -95,6 +95,20 @@
                 @endforeach
             </x-filament::input.select>
         @else
+            <div
+                class="hidden"
+                x-data="{
+                    disabled: @js($isDisabled),
```
danharrin: Please rename this AND the event prop to `isDisabled`

danharrin (changes requested):
Thanks for this!

packages/forms/resources/views/components/select.blade.php
```diff
@@ -95,6 +95,20 @@
...
+                    init() {
```
danharrin:
```suggestion
                    init: function () {
```

PR #13748, opened 2024-07-30
Persist Textarea autosize

awcodes:
## Description

Resolves #13722. Prevents textarea autosize from being reset during livewire requests.

## Visual changes

https://github.com/user-attachments/assets/984fa123-3f57-4528-a79c-53b4e1b3e191

## Functional changes


packages/forms/resources/views/components/textarea.blade.php
```diff
@@ -34,19 +34,25 @@
         "
     >
         <textarea
+            wire:ignore
```
danharrin: We cannot `wire:ignore` here unfortunately since it would break dynamically making the textarea `disabled` for example

danharrin (changes requested):

packages/forms/resources/views/components/textarea.blade.php
awcodes: hmmm. I not sure of any other way to do it then, since the livewire request is resetting the `$initialHeight` in the php evertime it is rerendered.

I'll keep thinking about it.

danharrin: You could store the initial height in the value of a hidden input within the component, and then reference that in the JS instead of a property

awcodes:
What is the acceptable percentage for css features? Come on Firefox and safari. https://caniuse.com/?search=field-sizing

zepfietje:
> What is the acceptable percentage for css features? Come on Firefox and safari. https://caniuse.com/?search=field-sizing

66.52% really is too low unfortunately.
I always aim for about +90% support, though it also depends on which browsers and versions aren't supported.

zepfietje:
Thanks a lot for working on this, @awcodes!
Please re-request my review once all discussions have been resolved.
Also, hit me up if there's anything you want to brainstorm about.

awcodes:
@zepfietje Latest commit works. Feels a little gross to me, though. The main problem is that for some reason the Livewire requests are dropping the `x-bind:style` attributes. If that can be figured out, this could possibly cleaner. But it works.

zepfietje:
Thanks, @awcodes.

Are we ok with this approach, @danharrin?
If so, I'll properly test this PR before merging, so let me know :)

zepfietje:
Can confirm this PR does indeed fix the issue for me.

packages/forms/resources/views/components/textarea.blade.php
```diff
@@ -33,43 +33,48 @@
...
+                @endif
+                ax-load-src="{{ \Filament\Support\Facades\FilamentAsset::getAlpineComponentSrc('textarea', 'filament/forms') }}"
+                x-data="textareaFormComponent({
+                    initialHeight: @js($initialHeight),
+                    shouldAutosize: @js($shouldAutosize),
+                    state: $wire.{{ $applyStateBindingModifiers("\$entangle('{$statePath}')") }},
+                })"
+                @if ($shouldAutosize)
+                    x-intersect.once="resize()"
+                    x-on:input="resize()"
```
danharrin: Now we have the state watcher, should this line be removed maybe?

packages/forms/resources/views/components/textarea.blade.php
```diff
@@ -33,43 +33,48 @@
...
-                ax-load
-            @endif
-            ax-load-src="{{ \Filament\Support\Facades\FilamentAsset::getAlpineComponentSrc('textarea', 'filament/forms') }}"
-            x-data="textareaFormComponent({ initialHeight: @js($initialHeight) })"
-            @if ($shouldAutosize)
-                x-intersect.once="resize()"
-                x-on:input="resize()"
-                x-on:resize.window="resize()"
-            @endif
-            x-bind:style="{ height }"
```
danharrin, line 50: Now this has been removed, are we actually using the height property anymore? Can it be removed?

danharrin (changes requested):

packages/forms/resources/views/components/textarea.blade.php
```diff
@@ -33,43 +33,47 @@
...
-                        'cols' => $getCols(),
-                        'disabled' => $isDisabled,
-                        'id' => $getId(),
-                        'maxlength' => (! $isConcealed) ? $getMaxLength() : null,
-                        'minlength' => (! $isConcealed) ? $getMinLength() : null,
-                        'placeholder' => $getPlaceholder(),
-                        'readonly' => $isReadOnly(),
-                        'required' => $isRequired() && (! $isConcealed),
-                        'rows' => $rows,
-                        $applyStateBindingModifiers('wire:model') => $statePath,
```
danharrin, line 66: Does it still work if we keep this `wire:model` line as well, and remove the `x-model`? Otherwise, I think `live(onBlur: true)` and debounce control would break for this component, as these aren't supported by entangle?

You can still pass the entangled prop into the JS as it should keep itself in sync with the `wire:model`, but it can probably be simplified to `state: $wire.$entangle('{{ $statePath }}')` as the state binding modifiers have no effect if you never set the state

danharrin (changes requested):

packages/forms/resources/views/components/textarea.blade.php
awcodes: Appears to be ok. I'm not sure of any side effects of entangle and wire:model, though. Will default to your judgement on this.

danharrin:
Great, thanks

zepfietje:
Thanks guys! 💛

davidwhicker:
It looks like there is a slight issue on the release:

```
Textarea::make('collectFrom')->autosize()->hintActions(                   
        [
            \Filament\Forms\Components\Actions\Action::make('collect_client_address_modal')
                ->label('Client')
                ->icon('heroicon-o-magnifying-glass')
                ->modalSubmitActionLabel('Add Address')
                ->form(fn(Get $get)=>self::searchClientActionSchema($get))
                ->action(fn (array $data, Set $set) => $set('collectFrom', Client::find($data['client_id'])->address ?? null)),
        ]
    ),
```
                        
                        
When using a hintAction WITH a form it will not resize the textarea field when passing a value back. IF a modal is not required on the hintAction and custom logic is solely used in action() then it will work.

In short, it seems passing back a value into the textarea from a hintaction MODAL it will not update. 

The above code will fail to resize however the below will work:
```
Textarea::make('collectFrom')->autosize()->hintActions(                   
        [
            \Filament\Forms\Components\Actions\Action::make('collect_client_address_modal')
                ->label('Client')
                ->icon('heroicon-o-magnifying-glass')
                ->modalSubmitActionLabel('Add Address')
                ->action(fn (array $data, Set $set) => $set('collectFrom', Client::find(1072)->address ?? null)),
        ]
    ),
```


To confirm, the textarea does not collapose like before, its just not updating to its new size.

danharrin:
@awcodes do you have the time to look into this? no rush, lmk if not

awcodes:
Yea. Was going to see about it tomorrow. Can't imagine why it wouldn't work since the state is entangled and $set directly works.

awcodes:
autosize is working in the demo for me with:

```php
->hintAction(
    FormAction::make('fill')
        ->form([
            Forms\Components\Textarea::make('modal_content'),
        ])
        ->action(function ($component, $data, Forms\Set $set) {
            $set('some_content', $data['modal_content']);

        })
)
```

https://github.com/user-attachments/assets/d6749cea-7c5e-4453-af5b-a9cbdc782696

danharrin:
The issue doesnt have a reproduction either...

davidwhicker:
> autosize is working in the demo for me with:
> 
> ```
> ->hintAction(
>     FormAction::make('fill')
>         ->form([
>             Forms\Components\Textarea::make('modal_content'),
>         ])
>         ->action(function ($component, $data, Forms\Set $set) {
>             $set('some_content', $data['modal_content']);
> 
>         })
> )
> ```
> 
>  Screen.Recording.2024-08-12.at.9.48.22.AM.mov

I am using this outside of Panels in a livewire component if this helps? I have tried your exact code and the issue remains. I appreicatae I have not supplied a repo so I understand if you would like to close this one.

awcodes:
Nothing in that code is panel specific.

davidwhicker:
> Nothing in that code is panel specific.

hmmm I'm at a loss then. Only other thing I can think of is my Textarea with the hintAction on is already on a Modal. So the Textarea that the hintAction returns a value to is on a Modal to begin with. Two levels of Modals. It may have no bearing but worth mentioning.

awcodes:
This is why we need a reproduction repo. 😅

danharrin:
every. single. time. 🤣

PR #13814, opened 2024-08-05
Add better mimetype support for FileUpload component

bbredewold:
## Description

This PR adds better front-end filetype validation to the FileUpload field.

The FileUpload field uses Filepond for uploading files.
When setting `->acceptedFileTypes()` with the accepted mimeTypes, Filepond will pre-validate the input files before sending them to the server where Laravel's validator also validates the mimeTypes. So far so good.

### The problem

The problem with front-end validation, if that the mimeType guessing is done by the client. So the OS and the browser is responsible to return a mimeType of a selected file. When the type cannot be resolved, Filepond receives a `null` and the validation fails.

In our case, when uploading WOFF / WOFF2 (webfonts) files, the validation fails because the browser tells Filepond that the mimeType is `null`. The mimetypes should be `font/woff` and `font/woff2`.

### The fix

Since we cannot rely on the OS and browser to resolve the mimeTypes for some more exotic file extensions, I added a package which contains a more complete database with mimeType-extension combinations.

Utilising the `fileValidateTypeDetectType` property of the `filepond-plugin-file-validate-type` filepond-extension, we can lookup the correct mimeType when one cannot be resolved.

## Visual changes

Using:
```php
FileUpload::make('custom_font')
    ->acceptedFileTypes(['font/woff2', 'application/font-woff2']),
```

Before:
BRB2024-08-05 at 10 25 34@2x

After:
BRB2024-08-05 at 10 26 28@2x

## Functional changes


danharrin (changes requested):
Can you please read what it says [here](https://pqina.nl/filepond/docs/api/plugins/file-validate-type/#custom-type-detection) about returning a promise from the function, where the promise is rejected if the type is not detectable? So if the package returns `null`, I guess

bbredewold:
> Can you please read what it says [here](https://pqina.nl/filepond/docs/api/plugins/file-validate-type/#custom-type-detection) about returning a promise from the function, where the promise is rejected if the type is not detectable? So if the package returns `null`, I guess

Hi Dan, my idea was to always resolve, even when no mimeType could be found in the package. This way, it returns to the default behaviour, by returning a `null`.

danharrin:
I would prefer to follow the FilePond docs in case it is more strict in the future, please

bbredewold:
> I would prefer to follow the FilePond docs in case it is more strict in the future, please

That seems wise :)
Should we keep the async syntax or do you want to work with actual Promises, like the documentation.

```js
fileValidateTypeDetectType: async (source, detectedType) => {
    const mimeType = detectedType || mime.getType(
        source.name.split('.').pop(),
    )
    
    if (! mimeType) {
        throw new Error('MIME type could not be detected')
    }

    return mimeType
}
```

vs

```js
fileValidateTypeDetectType: (source, detectedType) => {
    return new Promise((resolve, reject) => {
        const mimeType = detectedType || mime.getType(source.name.split('.').pop());
        
        if (! mimeType) {
            reject(new Error('MIME type could not be detected'));
        } else {
            resolve(mimeType);
        }
    });
}
```

danharrin:
Thanks!

PR #13836, opened 2024-08-06
Added 'error' as alias to 'danger' to HasStatus.php

syntaxlexx:
## Description
Added 'error' as an alias to 'danger' 

Having to debate whether it is 'danger' or 'error', this update shall help eliviate that as they should 'ideally' be the same.
On one hand, we have 'success', and it would be ideal having 'error' as its pollar opposite.

Also, after getting used to Laravel session flashing 'success' and 'error',  this would greatly reduce potential errors and the mental fatigue associated in switching between frameworks. 

## Functional changes

### DIFF
--- a/packages/notifications/src/Concerns/HasStatus.php
+++ b/packages/notifications/src/Concerns/HasStatus.php
@@ -25,6 +25,11 @@ trait HasStatus
         return $this->status('danger');
     }
 
+    public function error(): static
+    {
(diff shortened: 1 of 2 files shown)


zepfietje:
These statuses are named after the semantic color names, which are primary, info, success, warning, and danger. We decided to go with "danger" instead of "error" because the red color doesn't always signal a mistake (error).

PR #13843, opened 2024-08-06
Docs update: Customizing the page title -> better example

kachelle:
## Description

Bettter example, to make explicit you can access the record property in this function. Since this is most likely the thing you will want to do.

## Visual changes
Before:
image

After:
image
## Functional changes


danharrin:
This is docs for any page in the entire panel, and not all pages have access to the record, only edit/view pages in a resource.

PR #13888, opened 2024-08-11
Fix cache invalidation in getCachedExistingRecords() method in nesting repeater component 

liyuze:
The Repeater component's `getChildComponentContainers()` method returns a list of components by calling the `getClone()` method, which produces entirely new clone objects each time. As a result, the `getCachedExistingRecords()` method of all non-top-level Repeater components will always select DB because the `cachedExistingRecords` property remains unset.

## Description

Many internal methods, such as getValidationRules(), getValidationMessages(), and getValidationAttributes(), also call getChildComponentContainers(). Each time these methods are invoked, they query the database and cache the results in a new object's cachedExistingRecords property. The getClone() method causes the getCachedExistingRecords() method to lose its original design purpose within nested components.

## Visual changes

form ui
ui

show edit page
edited_show

before save
no_edit_save

now save
edited_save

## Functional changes

### DIFF
--- a/packages/forms/src/Components/Repeater.php
+++ b/packages/forms/src/Components/Repeater.php
@@ -45,7 +45,7 @@ class Repeater extends Field implements Contracts\CanConcealComponents, Contract
 
     protected bool | Closure $isInset = false;
 
-    protected ?Collection $cachedExistingRecords = null;
+    protected static array $cachedExistingRecords = [];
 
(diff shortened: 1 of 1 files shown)


danharrin:
Sorry, I don't think we can do this, it would break on Octane for sure which might even be a security issue if there is data leakage between requests

PR #13890, opened 2024-08-11
Add authorization to register method

4m1n0s:
## Description

Issue scenario:
if the user keep the page open (skipping the mount method that check for authorization), and keep submitting the form, it will work because there is no authorization check on the register method.

## Functional changes


danharrin:
Isn't the form button disabled after it is clicked the first time, that is Livewire behaviour I think?

danharrin:
Either way I think this is an OK change since someone could trigger it from JS

danharrin:
whoops

Issue #13899, opened 2024-08-12
DeleteAction within ActionGroup resulting in fatal error in unit test


BradnDodd:
### Package

filament/filament

### Package Version

3.2.40

### Laravel Version

11.20.0

### Livewire Version

3.5.4

### PHP Version

8.3

### Problem description

When testing a DeleteAction within an ActionGroup it produces the following errors; 

```
Failed asserting that a table action with name [delete] exists on the [App\Livewire\ListPosts] component.
Failed asserting that null is an instance of class Filament\Tables\Actions\Action.
```

When debugging through it appears that the `flatActions` variable against the Action group stores some of the actions with a capitalised name but when testing its searching for an entirely lowercase name thus producing the above errors  

### Expected behavior

A successful result on a unit test for a delete action

### Steps to reproduce

Checkout this reproduction repo , composer install then run php artisan test. There is a single unit test there that should reproduce the bug 

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/BradnDodd/filament-bug/

### Relevant log output

```shell
Failed asserting that a table action with name [delete] exists on the [App\Livewire\ListPosts] component.
Failed asserting that null is an instance of class Filament\Tables\Actions\Action.
```

BradnDodd:
https://github.com/filamentphp/filament/pull/13900 Fix for this bug here. From clicking around the UI this seems to still work on the interface and also fixes my unit test issue.

danharrin:
https://github.com/filamentphp/filament/pull/13900#issuecomment-2283800832

PR #13900, opened 2024-08-12
Fix issue 13899 DeleteAction within ActionGroup resulting in fatal error in unit test 

BradnDodd:
## Description

Fixes the https://github.com/filamentphp/filament/issues/13899 bug. Due the the code storing actions with one name and retrieving them with a different name

## Visual changes

No visual changes 

## Functional changes


danharrin:
I don't know why you are using `Delete` as the name instead of sticking to the default `delete`. The name is used to generate the label by capitalising it, you don't need to do that for the framework. The correct way is to use a `camelCase` action name, or stick to the default: `DeleteAction::make()` instead of `DeleteAction::make('Delete')`

Your test would pass if you used `Delete` as the name instead of `delete`

The reason why I am not merging this is because it would break tests for users that legitimately want to change the name of the action, or use the same action multiple times in the same table with different names

Issue #13907, opened 2024-08-12
Tenancy does not work when Guest access is set in Filament


akshit-arora:
### Package

filament/filament

### Package Version

v3.2

### Laravel Version

11.20.0

### Livewire Version

v3.5

### PHP Version

PHP 8.2

### Problem description

If the filament has been set to access without user authentication, i.e. the guest access, enabling the tenancy gives an error.

```
Filament\FilamentManager::getUserDefaultTenant(): Argument #1 ($user) must be of type Filament\Models\Contracts\HasTenants|Illuminate\Database\Eloquent\Model|Illuminate\Contracts\Auth\Authenticatable, null given, called in D:\laravel\filament-guest-tenant-issue\vendor\laravel\framework\src\Illuminate\Support\Facades\Facade.php on line 357
```

Tenancy is expecting a logged in user.

### Expected behavior

The tenancy should not be dependent on the user authentication. User authentication can be disabled and guest access can be set from the AdminPanelProvider. The tenancy should work in both the cases.

### Steps to reproduce

Set up the guest access in Filament as described in the docs [here](https://filamentphp.com/docs/3.x/panels/users#setting-up-guest-access-to-a-panel)

Apply tenancy in the Panel configuration

```
use App\Models\Team;
use Filament\Panel;
 
public function panel(Panel $panel): Panel
{
    return $panel
        // ...
        ->tenant(Team::class);
}
```

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/akshit-arora/filament-guest-tenancy-issue

danharrin:
Closing as not supported as per https://github.com/filamentphp/filament/pull/13908#issuecomment-2284467526

PR #13908, opened 2024-08-12
fix: Tenancy with guest access

akshit-arora:
## Description

Fixes issue #13907 
Tenancy can work when guess access is enabled

## Functional changes

### DIFF
--- a/packages/panels/src/Events/TenantSet.php
+++ b/packages/panels/src/Events/TenantSet.php
@@ -13,7 +13,7 @@ class TenantSet
 
     public function __construct(
         protected Model $tenant,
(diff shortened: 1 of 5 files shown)


danharrin:
I am concerned by the default that when there isn't a user authenticated, one can access every tenant in the system. I think it could result in some accidental data leaks in edge case situations if the auth middleware is not run for example. Feels very very dangerous. 

I haven't had any other feature requests for this functionality to work with guest access, and I don't know how to proceed without introducing new configuration which would probably be used very rarely.

Sorry, because of the lack of other request and my security concerns I am going to choose not to merge this at the moment.

PR #13740, opened 2024-07-30
fix: nested modals scrolling issue

wychoong:
## Description

fixes #10925 

using $nextTick for `open` action to make sure a modal is open only after another is closed

## Visual changes

before

https://github.com/user-attachments/assets/76cd84b3-b0e2-4135-9ecf-3af727430f5a

after

https://github.com/user-attachments/assets/c7711669-822c-421d-9a48-4f1fb793fdd8

## Functional changes

- [x] Code style has been fixed by running the `composer cs` command.
- [x] Changes have been tested to not break existing functionality.
- [x] Documentation is up-to-date.

PR #13921, opened 2024-08-13
fix async alpine not loading reliably in spa mode

wychoong:
## Description
usage of $nextTick to open modal in #13740 seems affecting components loading in spa mode
this PR added a new event to trigger ax-load https://async-alpine.dev/docs/strategies/#event

the issue happens randomly and do not have consistent steps to replicate, but I observed is after browser window size change.
1. load page
2. resize window
3. open modal (close and open again if it doesn't)

with the best effort test it seems to fix the `select` component. 
applied the change to all **form** components using `ax-load="visible"`
not applied to table & widget is not usable in modal

discord ref: https://discord.com/channels/883083792112300104/970354547723730955/1272603144148488255

## Visual changes
N/A

## Functional changes

### DIFF
--- a/packages/forms/resources/views/components/color-picker.blade.php
+++ b/packages/forms/resources/views/components/color-picker.blade.php
@@ -43,7 +43,7 @@
         <div
             x-ignore
             @if (FilamentView::hasSpaMode())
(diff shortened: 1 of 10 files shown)


danharrin:
Thanks

PR #13923, opened 2024-08-13
Fix alpine class binding for inactive tab items

gehrisandro:
## Description

There is a bug, which prevents inactive tab items from getting the correct CSS classes.

## Visual changes

This is primarily visible in dark mode. In light mode only the hover state is missing.

Before:
CleanShot 2024-08-13 at 07 57 17@2x

After:
CleanShot 2024-08-13 at 07 56 53@2x

## Functional changes

### DIFF
--- a/packages/support/resources/views/components/tabs/item.blade.php
+++ b/packages/support/resources/views/components/tabs/item.blade.php
@@ -51,7 +51,7 @@
     @endif
     @if ($hasAlpineActiveClasses)
         x-bind:class="{
-            @js($inactiveItemClasses): ! {{ $alpineActive }},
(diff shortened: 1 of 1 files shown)


danharrin:
Unfortunately Prettier undoes this change when you run `cs`. Can you try using `{{-- format-ignore-start --}}` and `{{-- format-ignore-end --}}` around these lines, and then running `cs` to check it still works please?

gehrisandro:
I am very sorry. Didn't realize, `composer cs` affects blade files too.

Your suggestion works fine 👍

danharrin:
Thank you!

PR #13936, opened 2024-08-13
Form getState: adding shouldCallHooksBeforeExceptComponents array parameter

### LINKED ISSUE #8: Compatibility with relationship to translatable models
Problem description:

In v3.2.81, translatable models with a relationship to another translatable model don't get that relationship's data updated on locale change via the language switch.

Expected Behavior:

It should switch to the translated version of the content for that relationship

Potential solution:

https://github.com/lara-zeus/translatable/blob/c00fbd28c6e37f2476498befb1925a071f414d18/src/Resources/Pages/EditRecord/Concerns/Translatable.php#L84

In my [PR](https://github.com/filamentphp/filament/pull/11450) I had encountered that issue, and using `$this->form->getState(false)` fixed it.

Versions:
- filament/filament 3.2.81
- filament/spatie-laravel-translatable-plugin 3.2.81
- ralphjsmit/laravel-filament-seo 1.3.2
- z3d0x/filament-fabricator 2.2.2

Setup:
- `\App\Models\Overrides\Page` extends `\Z3d0X\FilamentFabricator\Models\Page` and uses `\RalphJSmit\Laravel\SEO\Support\HasSEO` as well as `Spatie\Translatable\HasTranslations` (NB: the `seo` property is not part of the `$translatable` array since it's a relationship and not a db column)
- `\App\Models\Overrides\SEO` extends `\RalphJSmit\Laravel\SEO\Models\SEO` and uses `Spatie\Translatable\HasTranslations`
- `\App\Filament\Resources\PageResource` extends `\Z3d0X\FilamentFabricator\Resources\PageResource`, uses `Filament\Resources\Concerns\Translatable` and all related pages are properly overridden and configured
- A (filament) section for the `seo` relationship is displayed on the (edit or create) page, its translatable attributes are editable thanks to the section's schema

cawecoy:
## Description

Making it convenient to fix https://github.com/lara-zeus/translatable/issues/8 (we can't translate certain components like Repeaters).

To fix it, we need to be able to use `getState()` method with `$shouldCallHooksBefore = true` but not for every component. In other words, we need to make the "call hooks before" ignore certain components like:

```php
// using shouldCallHooksBefore but not for components named 'answers' and 'modules'
$this->form->getState(
    shouldCallHooksBefore: true,
    shouldCallHooksBeforeExceptComponents: ['answers', 'modules']
)
```

Why?

Currently, if we use `$this->form->getState(shouldCallHooksBefore: true)` which applies "call hooks before" to all components, then certain components like Repeaters fail to be translatable.

And if we use `$this->form->getState(shouldCallHooksBefore: false)` which doesn't apply "call hooks before" to any component, then certain components like FileUpload fail to be translatable.

## Functional changes

### DIFF
--- a/packages/forms/src/Concerns/BelongsToModel.php
+++ b/packages/forms/src/Concerns/BelongsToModel.php
@@ -15,9 +15,13 @@ trait BelongsToModel
         return $this;
     }
 
-    public function saveRelationships(): void
+    public function saveRelationships(?array $exceptComponents = []): void
     {
         foreach ($this->getComponents(withHidden: true) as $component) {
+            if (method_exists($component, 'getName') && in_array($component->getName(), $exceptComponents)) {
+                continue;
+            }
+            
             $component->saveRelationshipsBeforeChildren();
 
             $shouldSaveRelationshipsWhenDisabled = $component->shouldSaveRelationshipsWhenDisabled();
@@ -27,16 +31,20 @@ trait BelongsToModel
                     continue;
                 }
 
-                $container->saveRelationships();
+                $container->saveRelationships($exceptComponents);
             }
 
             $component->saveRelationships();
         }
     }
 
-    public function loadStateFromRelationships(bool $andHydrate = false): void
+    public function loadStateFromRelationships(bool $andHydrate = false, ?array $exceptComponents = []): void
     {
         foreach ($this->getComponents(withHidden: true) as $component) {
+            if (method_exists($component, 'getName') && in_array($component->getName(), $exceptComponents)) {
+                continue;
+            }
+            
             $component->loadStateFromRelationships($andHydrate);
 
             foreach ($component->getChildComponentContainers(withHidden: true) as $container) {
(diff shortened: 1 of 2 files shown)


danharrin:
What specific hooks cause repeaters to break?

danharrin:
Because it would be ideal to fix the hooks themselves instead of introducing this

Voltra:
It's usually an issue with hydration/dehydration. With `getState()` it'd be `callBeforeStateDehydrated`. 

There is the following repo for repro of bugs with translating relationships (which is fixed with `shouldCallHooksBefore: false`, but breaks other components' translatability like `FileUpload`'s): https://github.com/lara-zeus/translatable-demo

danharrin:
The reason why I handed off development of the package is because I didn't have time to properly deal with issues like this. I don't think this solution is one that I'd merge as it does create complexity and also has issues if you have two fields with the same name in different state paths.

I would be open to a PR that fixes the repeater hooks that are problematic in a way that does not affect the entire hydration process. Many thanks.

PR #13942, opened 2024-08-14
fix: Header actions render hook alignment

danharrin:


packages/panels/resources/views/components/header/index.blade.php
```diff
@@ -31,17 +31,18 @@ class="fi-header-subheading mt-2 max-w-2xl text-lg text-gray-600 dark:text-gray-
...
     </div>
 
-    {{ \Filament\Support\Facades\FilamentView::renderHook(\Filament\View\PanelsRenderHook::PAGE_HEADER_ACTIONS_BEFORE, scopes: $this->getRenderHookScopes()) }}
+    <div
+        @class([
+            'flex shrink-0 items-center gap-3',
+            'sm:mt-7' => $breadcrumbs,
+        ])
+    >
+        {{ \Filament\Support\Facades\FilamentView::renderHook(\Filament\View\PanelsRenderHook::PAGE_HEADER_ACTIONS_BEFORE, scopes: $this->getRenderHookScopes()) }}
```
zepfietje, line 40: Isn't this now more like "start" and "end" instead of "before" and "after"?

danharrin: It would be `start` and `end` if the hooks were inside the actions component, but they are outside it so it is `before` and `after`

zepfietje: Why was the previous position wrong then?

danharrin: The previous hook position was not useful in order to add elements either side of the actions, since the parent element has `justify-between` and would position one of the elements in the middle of the screen

PR #13952, opened 2024-08-15
fix: File upload compact circle error messages

danharrin:


packages/forms/resources/views/components/file-upload.blade.php
```diff
@@ -142,6 +139,13 @@
             />
         </div>
 
+        <div
```
zepfietje, line 142: Any reason why you're using `<div>` here instead of `<p>`, @danharrin?
Also, why not just reuse the existing `error-message` Blade component?

danharrin: `p` would work, not using the Blade component as this is not the field wrapper

zepfietje: Perhaps extract the common code from the field wrapper error message and this file then, because it needs to be themeable using CSS hook classes as well.

PR #13954, opened 2024-08-15
fix(constraints): human-readable summary display

maartenpaauw:
## Description

Implement a human-readable summary based on the configured `getOptionLabelsUsingCallback()` or `getOptionLabelUsingCallback()` methods. If both callbacks are `null`, the summary will fall back to using the selected options retrieved through `getOptions()`.

### Before

before

### After

after

## Functional changes

### DIFF
--- a/packages/tables/src/Filters/QueryBuilder/Constraints/SelectConstraint/Operators/IsOperator.php
+++ b/packages/tables/src/Filters/QueryBuilder/Constraints/SelectConstraint/Operators/IsOperator.php
@@ -29,8 +29,19 @@ class IsOperator extends Operator
     public function getSummary(): string
     {
         $constraint = $this->getConstraint();
+        $settings = $this->getSettings();
 
-        $values = Arr::wrap($this->getSettings()[$constraint->isMultiple() ? 'values' : 'value']);
+        if ($constraint->isMultiple()) {
+            $callback = $constraint->getOptionLabelsUsingCallback();
+            $valuesKey = 'values';
+        } else {
+            $callback = $constraint->getOptionLabelUsingCallback();
+            $valuesKey = 'value';
+        }
+
+        $values = $callback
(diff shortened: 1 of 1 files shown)


maartenpaauw:
I understand that the focus is currently on v4, so if it's more convenient, feel free to close this for now. I'll be happy to resubmit a rebased pull request once v4 is released.

danharrin:
Please send the query builder code for this constraint that causes the issue

maartenpaauw:
Of course! I should have included it in the description from the start.

```php
QueryBuilder::make()
    ->constraints([
        SelectConstraint::make('state')
            ->multiple()
            ->getOptionLabelUsing(static fn (string $value): string => Str::title($value)) // Optional
            ->getOptionLabelsUsing(static fn (array $values): array => Arr::map($values, Str::title(...))) // Optional
            ->options([
                'received' => 'Received',
                'under_consideration' => 'Under Consideration',
                'rejected' => 'Rejected',
            ]),
    ]);
```

If this query builder is used without both optional lines, selecting `Under Consideration` will cause the summary to display its key value `under_consideration` instead of the label, without the fix in this PR.

danharrin:
So, can we not use the options array to translate the keys into labels, instead of the functions?

maartenpaauw:
Using the options array directly to translate the keys into labels would indeed work for most cases and would meet my needs as well. However, to avoid any confusion, I added support for both the `getOptionLabelsUsing()` and `getOptionLabelUsing()` methods. If neither method is configured by the developer, the code falls back to using the `getOptions()` method to retrieve the necessary values.

If you prefer, I can remove the callback check and always rely on the `getOptions` array, which would simplify the code. Let me know your preference, and I’ll update the implementation accordingly!

danharrin:
Thanks

maartenpaauw:
You're welcome! Thanks for cleaning up the code and merging 🚀

PR #13957, opened 2024-08-16
Check model view policy on related records pages

ryanmitchell:
## Description

If you are using a ManageRelatedRecords page it checks the `viewAny` policy for the relation, but it doesn't check the `view` policy for the record itself. This makes it possible to 'game' the URLs by changing the IDs in the address bar, potentially being able to see something you don't have permission to see.

This PR ensures the `view` policy on the record is checked alongside the existing `viewAny` check on the relation.

## Visual changes

There are no visual changes

## Functional changes

### DIFF
--- a/packages/panels/src/Resources/Pages/ManageRelatedRecords.php
+++ b/packages/panels/src/Resources/Pages/ManageRelatedRecords.php
@@ -105,7 +105,7 @@ class ManageRelatedRecords extends Page implements Tables\Contracts\HasTable
         $model = $record->{static::getRelationshipName()}()->getQuery()->getModel()::class;
 
         try {
-            return authorize('viewAny', $model, static::shouldCheckPolicyExistence())->allowed();
+            return authorize('view', $record, static::shouldCheckPolicyExistence())->allowed() && authorize('viewAny', $model, static::shouldCheckPolicyExistence())->allowed();
         } catch (AuthorizationException $exception) {
             return $exception->toResponse()->allowed();
         }


danharrin:
Hi, unfortunately we cannot merge this. The objective reason is that it is a breaking change for anyone relying on the current behaviour. And my personal reasoning is that this authorization was a conscious decision that I took, since the concept of "view"ing an entire records details may be authorized differently to a records management of relationships. It isn't necessarily true that a user needs to be able to view all details about a record just to manage its relationships to others, some parts of that might be sensitive. In the same way, we do not check if a user is "view"able before editing either. Keeping the authorization minimal (only check the current permission and if they can view the entire resource or not) and then allowing developers to override this on their pages is our philosophy currently.

So as mentioned above, you can override this method on your pages if you wanted to change this behaviour, you could even extract it into a trait if you wanted to reuse it across multiple pages.

ryanmitchell:
Thanks @danharrin. 

I opened it as a PR when I saw the issue required a GitHub repo to reproduce anyway (somewhat off-putting). 

I understand your reasoning, I guess you're trying to make it work for a lot of different needs. We'll opt for overriding the method on our resources by default.

danharrin:
Hi, please note that the docs mention opening a discussion (https://github.com/filamentphp/filament/discussions) and not an issue (https://github.com/filamentphp/filament/issues)

ryanmitchell:
My mistake - thanks

PR #13958, opened 2024-08-16
Prevent page from jumping to the top if the section is not collapsed

KarlBoosterBaer:
## Description

If a page has several sections that are collapsed by default, and the page size exceeds the window size when the sections are expanded, then under certain circumstances the page jumps back to the top of the page after a Livewire update. Apparently, when the page content is re-rendered, the original page height (when the sections are collapsed) is used and the browser jumps back to the top.

## Functional changes

### DIFF
--- a/packages/support/resources/views/components/section/index.blade.php
+++ b/packages/support/resources/views/components/section/index.blade.php
@@ -72,6 +72,7 @@
             },
         ])
     }}
+    wire:ignore.self
 >
     @if ($hasHeader)
         <header


danharrin:
I'm wondering if this means we can also revert #8511, can you test that maybe? Also @cheesegrits if you have time and could take a look and see if this singular attribute means we can move the `x-data` back into the conditional?

KarlBoosterBaer:
@danharrin I moved the x-data directive back into the condition and tried it in one of my projects. The bug still exists. The addded attribute doesnt fix issue #8511.

danharrin:
Thanks for checking!

danharrin:
Hi, unfortunately I've had to revert this PR. Initially, it caused problems with widgets, as they conditionally pass an `animate-pulse` class to the section for their loading state. After this PR, that was no longer getting removed because of the `wire:ignore`, so I wrapped the loading state in a `wire:replace` element to fix

Unfortunately, this introduced a separate problem with relation managers, which also uses the same loading state, due to a morphing issue that allowed the `wire:replace` to transfer from the loading state onto the relation manager, which in turn caused interactivity problems within that table.

I would accept a new PR for this issue, but unfortunately we cannot use `wire:ignore` on this outer element. Is it feasible to add it to one of the inner elements in the section, if we restructured the HTML so some of the collapsible attributes were inside?

PR #13973, opened 2024-08-17
Fix oldFormState containing new values

bumbummen99:
## Description

When Livewire tries to [update multiple properties](https://github.com/livewire/livewire/blob/e7f49c360a8218047e68d2182d65bd3f99d5508e/src/Mechanisms/HandleComponents/HandleComponents.php#L294) in one request, it does trigger an *update* event for [every property](https://github.com/livewire/livewire/blob/e7f49c360a8218047e68d2182d65bd3f99d5508e/src/Mechanisms/HandleComponents/HandleComponents.php#L314). 

This will run the [`updatingInteractsWithForms`](https://github.com/filamentphp/filament/blob/8de1a8d0f861d3a824edd0ee89ef455492d18932/packages/forms/src/Concerns/InteractsWithForms.php#L291) trait hook and in turn set `oldFormState` to equal data of the component for every property.

When reacting on a properties *updated* event, this will cause the expected `$old` parameter to be wrong, as `oldFormState` has already been overwritten by updating multiple properties at once so it does include the new and not the old value.

I would propose to only set entries in the `oldFormState` in case they are not already set, as to prevent the entries being overwritten by new values.

## Visual changes

There are none.

## Functional changes

- [x] Only set `oldFormState` entries in case they are not set

### DIFF
--- a/packages/forms/src/Concerns/InteractsWithForms.php
+++ b/packages/forms/src/Concerns/InteractsWithForms.php
@@ -288,9 +288,8 @@ trait InteractsWithForms
     {
         $statePath = (string) str($statePath)->before('.');
 
-        if (! Arr::has($this->oldFormState, $statePath)) {
-            $this->oldFormState[$statePath] = data_get($this, $statePath);
-        }
+        // https://github.com/filamentphp/filament/pull/13973
+        $this->oldFormState[$statePath] ??= data_get($this, $statePath);
     }
 
     public function getOldFormState(string $statePath): mixed
--- a/packages/forms/src/Concerns/InteractsWithForms.php
+++ b/packages/forms/src/Concerns/InteractsWithForms.php
@@ -288,7 +288,9 @@ trait InteractsWithForms
     {
         $statePath = (string) str($statePath)->before('.');
 
-        $this->oldFormState[$statePath] = data_get($this, $statePath);
+        if (! Arr::has($this->oldFormState, 'data')) {
+            $this->oldFormState[$statePath] = data_get($this, $statePath);
+        }
     }
 
     public function getOldFormState(string $statePath): mixed
--- a/packages/forms/src/Concerns/InteractsWithForms.php
+++ b/packages/forms/src/Concerns/InteractsWithForms.php
@@ -288,7 +288,7 @@ trait InteractsWithForms
     {
         $statePath = (string) str($statePath)->before('.');
 
-        if (! Arr::has($this->oldFormState, 'data')) {
+        if (! Arr::has($this->oldFormState, $statePath)) {
             $this->oldFormState[$statePath] = data_get($this, $statePath);
         }
     }


danharrin (changes requested):
Not every form has a `data` property, that all depends on the state path of the form.

I am confused about why `data` would ever be a key in `oldFormState`, since we are setting the key of the full form component. Are you suggesting that when setting `data.one.two`, this function is called for `data`, `data.one`, AND `data.one.two`?

bumbummen99:
> Not every form has a `data` property, that all depends on the state path of the form.
> 
> I am confused about why `data` would ever be a key in `oldFormState`, since we are setting the key of the full form component. Are you suggesting that when setting `data.one.two`, this function is called for `data`, `data.one`, AND `data.one.two`?

Yes you are right, so for example if i change `data.heading`, `data.blocks.XYZ.content`, and so on it will use the string before `.` as the key to be set in `oldFormState`, so only `data` is set. I'll update the code accordingly.

Edit: Should be correct now. Locally i have used the Arr helper below my changes too but i do not want to change more lines than required.

danharrin:
Can you please create a reproduction repository for the original bug? I'm honestly quite lost now, if the entire state path is in the array key then I don't understand how multiple property updates (with different state paths) would overwrite each other, so I'd like to check it out

bumbummen99:
> Can you please create a reproduction repository for the original bug? I'm honestly quite lost now, if the entire state path is in the array key then I don't understand how multiple property updates (with different state paths) would overwrite each other, so I'd like to check it out

The overwrite happens because data is written to oldFormState, but data could already have been modified by a previous properties "update cycle". So the first time Filament executes the updating hook, oldFormState['data'] is set from NULL to its initial, old value. But when said hook is executed again, the already modified data is written to oldFormState again. This means, if i run the updating hook over and over, oldFormState will be the "old" data of the last iteration done by Livewire to change the property and therefore include all previous property changes that are actually part of this update request / stack.


I will put up an example repository to illustrate the issue using examples exactly as shown from the V3 documentation.

bumbummen99:
@danharrin I have set up the mentioned repository as well as a test to verify the issue:

https://github.com/bumbummen99/filament-bug-oldFormState

Please take note of the mentioned links in the [NewsResource](https://github.com/bumbummen99/filament-bug-oldFormState/blob/master/app/Filament/Resources/NewsResource.php) as tose are copy & paste examples from the documentation.

Also i have added an explaination of the issue to the test case:
https://github.com/bumbummen99/filament-bug-oldFormState/blob/master/tests/Feature/Livewire/CreateNewsTest.php

bumbummen99:
Is there anything else I can provide or do to help get this request merged?

danharrin:
I haven't had a chance to look at the resources you have provided yet, but when I do I will let you know if I need anything else. Thank you for your patience

danharrin:
Thank you very much for the reproduction repository! I now understand the issue and the fix.

PR #13987, opened 2024-08-19
Ensure User is not null when getting default tenenat.

tonypartridge:
Some edge cases I have seen is we try and redirect after the user has logged out and as such it fails. In those instances it shouldn't try to redirect based on getUserDefaultTenant as they are logged out and instead we should check we have a user before passing one which maybe null. 

This ensure the next step is hit instead.

### DIFF
--- a/packages/panels/src/Panel/Concerns/HasRoutes.php
+++ b/packages/panels/src/Panel/Concerns/HasRoutes.php
@@ -173,7 +173,7 @@ trait HasRoutes
 
         $hasTenancy = $this->hasTenancy();
 
-        if ((! $tenant) && $hasTenancy && $this->auth()->user()) {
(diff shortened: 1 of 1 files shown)


packages/panels/src/Panel/Concerns/HasRoutes.php
```diff
@@ -173,7 +173,7 @@ public function getUrl(?Model $tenant = null): ?string
 
         $hasTenancy = $this->hasTenancy();
 
-        if ((! $tenant) && $hasTenancy) {
+        if ((! $tenant) && $hasTenancy && $this->auth()->user()) {
```
danharrin:
```suggestion
        if ((! $tenant) && $hasTenancy && $this->auth()->hasUser()) {
```

danharrin (changes requested):

PR #14018, opened 2024-08-21
Fix aria-label for sortable headers, add aria-sort on active sort

### LINKED ISSUE #13986: Screenreader can't announce table headers in sortable table
### Package

filament/filament

### Package Version

v3.2.100

### Laravel Version

v11.20.0

### Livewire Version

v3.5.4

### PHP Version

8.3.9

### Problem description

When a column is marked as being sortable the header cell is rendered as a button, but in turn also gets an aria-label set based on the sorting direction. See `header-cell.blade.php`. This makes sortable tables unusable for people using a screen reader since the context for the column is lost.

### Expected behavior

At the very least the screenreader should be able to identify the column based on the label that is set. Even better would be to dynamically set the `aria-sort="sortDirection"`. However this field should only be present on the current active sortable field. See [MDN web docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-sort) for more details.

### Steps to reproduce

The issue can be reproduced by marking a column as sortable, and using the screenreader to identify the column header. It will be read as the sorting direction.

### Relevant log output

_No response_

mikeslinkman:
## Description

Resolves #13986 changed the aria-label when using sortable header cells. Before this a screenreader would read out the sorting direction for every column instead of the label. Also added the aria-sort when the column is actively sorted, as per the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-sort) added `ascending` / `descending` as the possible values

cc @danharrin as discussed in the issue

## Visual changes

No visual changes

## Functional changes

### DIFF
--- a/packages/tables/resources/views/components/header-cell.blade.php
+++ b/packages/tables/resources/views/components/header-cell.blade.php
@@ -23,11 +23,11 @@
     <{{ $sortable ? 'button' : 'span' }}
         @if ($sortable)
             type="button"
-            aria-label="{{$slot}}"
+            aria-label="{{ $slot }}"
             wire:click="sortTable('{{ $name }}')"
         @endif
-        @if($activelySorted)
-            aria-sort="{{ $sortDirection === 'asc' ? 'ascending' : 'descending'}}"
+        @if ($activelySorted)
+            aria-sort="{{ __($sortDirection === 'asc' ? 'filament-tables::table.sorting.fields.direction.options.asc' : 'filament-tables::table.sorting.fields.direction.options.desc') }}"
         @endif
         @class([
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks! I've moved the `aria-sort` attribute onto the header cell element since that's how its shown in the MDN docs, does that still work ok for you?

mikeslinkman:
> Thanks! I've moved the `aria-sort` attribute onto the header cell element since that's how its shown in the MDN docs, does that still work ok for you?

You're right that was my mistake, works like it should!

danharrin:
Thanks for checking

PR #14050, opened 2024-08-25
Fix: Table grouping key

wychoong:
## Description
I have a table that I want to display in groups, where it should be group by `id` and display by the `name` column
the issue happens when the `name` is the same but different record.id
from the doc it seems I should be able to use [getTitleFromRecordUsing](https://filamentphp.com/docs/3.x/tables/grouping#setting-a-group-key) to achieve this but it makes no difference

```php
->defaultGroup('event.id')
  ->groups([
      Group::make('event.id')
          ->getTitleFromRecordUsing(fn ($record) => $record->event->name)
          // ->getKeyFromRecordUsing(fn ($record): string => $record->event->id) # if using `event.name` as the group
          ->titlePrefixedWithLabel(false),
  ])
```

## Visual changes

**before**
`event.id`
image

 apply `->getTitleFromRecordUsing(fn ($record) => $record->event->name)`
image

**after**
image

## Functional changes
not tested with grid table, so not applying to it

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -975,7 +975,7 @@
                                 $recordGroupTitle = $group?->getTitle($record);
                             @endphp
 
-                            @if ($recordGroupTitle !== $previousRecordGroupTitle)
+                            @if ($recordGroupTitle !== $previousRecordGroupTitle || $recordGroupKey !== $previousRecordGroupKey)
                                 @if ($hasSummary && (! $isReordering) && filled($previousRecordGroupTitle))
                                     <x-filament-tables::summary.row
                                         :actions="count($actions)"


danharrin:
To be honest it was deliberate behaviour that multiple groups with the same title did not start a new group. I think it looks a bit odd and broken, do you know what I mean?

Also, the way the sorting works, we rely on the title being the primary sort column of the table, so all the records in the group are in the same place. Wouldn't this change break using a secondary column to sort the records INSIDE a group on, since records that have the same name but not the same ID may not necessarily be next to each other once they are sorted by name?

wychoong:
> To be honest it was deliberate behaviour that multiple groups with the same title did not start a new group. I think it looks a bit odd and broken, do you know what I mean?

Yeah I understand that. But also when the ID is different it should be different. Eg: group by student’s name, and there is 2 John, it wouldn’t be correct 

> Wouldn't this change break using a secondary column to sort the records INSIDE a group on, since records that have the same name but not the same ID may not necessarily be next to each other once they are sorted by name?

TBH I didn’t test with table sort. But being same title/name but different entity should be different group
this is align with a normal sql `order by key_column, secondary_column` result

Ideally the group label should based on the table ordering, using the getKeyFromRecordUsing to determine if we moved to the next group, and using getTitleFromRecordUsing for the title label

Alternatively, wouldn’t mind to have a `formatTitleFromRecordUsing` that acts like table formatState 
```php
->defaultGroup('event.id')
  ->groups([
      Group::make('event.id')
          ->formatTitleFromRecordUsing(fn ($record) => $record->event->name)
          ->titlePrefixedWithLabel(false),
  ])
```

danharrin:
I'm still not on board with displaying multiple groups with the same title, but how about adding a check for differences in the description instead? Then you can add unique data to the description to create clear group distinctions.

The grouping system was specifically built to work with unique groups such as categories, so unless there are obvious distinctions I don't think this solution fits in with that philosophy. Also, without knowing how this affects secondary sorting within groups, I am hesitant to merge it. You could also make the title more unique by adding an identifier to it.

wychoong:
> how about adding a check for differences in the description instead?

Although not ideal, but I can work with this

> You could also make the title more unique by adding an identifier to it.

This taking too much attention

> without knowing how this affects secondary sorting within groups

I will probably open an issue/discussion with a repo to investigate with this

wychoong:
For now I will rework this PR with the description

PR #14053, opened 2024-08-26
Feature: Prune sensitive data from failed_import_rows data

MACscr:
## Description

This pull request adds a `sensitive()` method to the `ImportColumn` class, allowing developers to mark specific columns as sensitive. When a column is marked as sensitive, its data will be excluded from the JSON data stored for failed import rows. This is particularly important for handling encrypted or sensitive data that should not be stored in plain text within the application.

The `filterSensitiveData()` method has been introduced in the `ImportCsv` job to handle the filtering of sensitive data before it is logged. The method checks each column for the `isSensitive()` flag and ensures that any data from these columns is removed before saving the failed row data.

## Visual changes

N/A

## Functional changes

- [x] Changes have been tested to ensure they do not break existing functionality.
- [x] Documentation has been updated to include the new `sensitive()` method for `ImportColumn`.


danharrin:
Thanks

PR #14060, opened 2024-08-26
Update InteractsWithForms.php

RohanSakhale:
## Description

Fix towards access level method override issue for classes having `ViewRecord` as extended class

## Visual changes

Screenshot 2024-08-26 at 7 44 07 PM

## Functional changes


danharrin:
We are unable to do that as it would break it for anyone using the trait with the protected version of the method.

Also, the `ViewRecord` page includes the `InteractsWithForms` trait already so you don't need to include it again.

PR #14074, opened 2024-08-27
Improve type support for model-related methods

maartenpaauw:
## Description

Enhanced type support by adding docblock annotations with `class-string<Model>` and, in some cases, adding the native type `Model & Authenticatable` to ensure the object extends the `Model` class and implements the `Authenticatable` interface.

Addressed an edge case for `SpatieTagInput` by checking if the `getTagClassName` method exists before calling it. This was necessary because there isn't a docblock annotation to describe an instance of a class using a specific trait.

## Visual changes

None.

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/InteractsWithActions.php
+++ b/packages/actions/src/Concerns/InteractsWithActions.php
@@ -291,6 +291,9 @@ trait InteractsWithActions
         );
     }
 
+    /**
+     * @return Model|class-string<Model>|null
+     */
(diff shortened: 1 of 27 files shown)


danharrin:
Hi, I'm willing to do this but please target it at Filament v4. Thanks

maartenpaauw:
Alright! I'll do that!

maartenpaauw:
Pull request targeted to v4: #14083

PR #14083, opened 2024-08-29
Improve type support for model related methods

maartenpaauw:
## Description

Enhanced type support by adding docblock annotations with `class-string<Model>` and, in some cases, adding the native type `Model & Authenticatable` to ensure the object extends the `Model` class and implements the `Authenticatable` interface.

Addressed an edge case for `SpatieTagInput` by checking if the `getTagClassName` method exists before calling it. This was necessary because there isn't a docblock annotation to describe an instance of a class using a specific trait.

## Functional changes

### DIFF
--- a/packages/spatie-laravel-tags-plugin/src/Forms/Components/SpatieTagsInput.php
+++ b/packages/spatie-laravel-tags-plugin/src/Forms/Components/SpatieTagsInput.php
@@ -100,7 +100,7 @@ class SpatieTagsInput extends TagsInput
         }
 
         $model = $this->getModel();
-        $tagClass = $model && method_exists($model, 'getTagClassName') ? $model::getTagClassName() : config('tags.tag_model', Tag::class);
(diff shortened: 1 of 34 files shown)


packages/spatie-laravel-tags-plugin/src/Forms/Components/SpatieTagsInput.php
```diff
@@ -100,7 +100,7 @@ public function getSuggestions(): array
         }
 
         $model = $this->getModel();
-        $tagClass = $model ? $model::getTagClassName() : config('tags.tag_model', Tag::class);
+        $tagClass = $model && method_exists($model, 'getTagClassName') ? $model::getTagClassName() : config('tags.tag_model', Tag::class);
```
danharrin:
```suggestion
        $tagClass = ($model && method_exists($model, 'getTagClassName')) ? $model::getTagClassName() : config('tags.tag_model', Tag::class);
```

danharrin (changes requested):

danharrin:
Thanks

PR #14094, opened 2024-08-29
Update 05-viewing-records.md

sertxudev:
## Description

The infolist method inside `ViewRecord` isn't static.

### DIFF
--- a/packages/panels/docs/03-resources/05-viewing-records.md
+++ b/packages/panels/docs/03-resources/05-viewing-records.md
@@ -26,7 +26,7 @@ By default, the View page will display a disabled form with the record's data. I
 use Filament\Infolists;
 use Filament\Infolists\Infolist;
 
-public static function infolist(Infolist $infolist): Infolist
+public function infolist(Infolist $infolist): Infolist
 {
     return $infolist
         ->schema([


danharrin:
As per the line above

> you can define an `infolist()` method on the resource class:

sertxudev:
@danharrin The code block is incorrect because the `infolist` method inside the `ViewRecord` class is not static. 

https://github.com/filamentphp/filament/blob/143e9b86d9633de53ecb086face0ba8bc6ba373b/packages/panels/src/Resources/Pages/ViewRecord.php#L219-L222

image

Later in this docs page the `infolist` method appears as non static.

https://github.com/filamentphp/filament/blob/143e9b86d9633de53ecb086face0ba8bc6ba373b/packages/panels/docs/03-resources/05-viewing-records.md?plain=1#L138-L148

valpuia604:
@sertxudev read this https://filamentphp.com/docs/3.x/infolists/entries/getting-started#overview

danharrin:
> you can define an infolist() method on the resource class:

The method is defined on the resource class, not on the resource page class

> Later in this docs page the infolist method appears as non static.

The section of the docs is about adding an additional page with an infolist, in which case the infolist is defined on the page and the method doesn't need to be static

sertxudev:
Gotcha! Filament is so customizable that it is sometimes difficult to understand how to do it correctly. Thanks, awesome work!

PR #14101, opened 2024-08-30
Adding Page Sub Navigation Hooks

dbpolito:
## Description

This adds hooks for Page Sub Navigation

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/panels/resources/views/components/page/index.blade.php
+++ b/packages/panels/resources/views/components/page/index.blade.php
@@ -64,25 +64,33 @@
             ])
         >
             @if ($subNavigation)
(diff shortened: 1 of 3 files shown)


danharrin:

Please open a discussion and `@danharrin` so we can talk about the implementation, I'm unsure if using the same hooks for both sidebar and topbar nav is useful? What if the HTML between them differs? I'd need to see a use case.

dbpolito:
Well they are both sub-navigation... you can't have both on the same page, right? and you can scope for the pages you want it differently...

I'm totally fine having different hooks too...

altho i don't see this as a new feature... i understand...

dbpolito:
I created this discussion: https://github.com/filamentphp/filament/discussions/14134

Thanks

PR #14121, opened 2024-09-02
✨ Allow additional arguments on `authorize()`

juliomotol:
## Description

This PR aims to allow additional `$arguments` on `authorize()`.

Given the ff policy:

```php
class PostPolicy
{
    public function view(User $user, Post $post, PostTypeEnum $type): bool
    {
        // ...
    }
}
```

In our case, we have different `Resource`s for each type of `Post` (don't ask me why, I suggested to just use filters but the client insisted on separating each type), and certain users may only have access on certain types of `Post`s 

At the moment, Filament's `authorize` method is unable to pass a `$type` argument to the policy method unlike in Laravel's `authorize` method.

## Visual changes

n/a

## Functional changes

### DIFF
--- a/packages/panels/src/Resources/Concerns/InteractsWithRelationshipTable.php
+++ b/packages/panels/src/Resources/Concerns/InteractsWithRelationshipTable.php
@@ -22,6 +22,9 @@ trait InteractsWithRelationshipTable
 
     protected static bool $shouldSkipAuthorization = false;
 
+    /** @var array<mixed> */
+    protected static array $authorizationArguments = [];
+
     public static function checkPolicyExistence(bool $condition = true): void
     {
         static::$shouldCheckPolicyExistence = $condition;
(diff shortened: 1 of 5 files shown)


danharrin:
Hi, apologies but we are no longer accepting new features for v3 as per #13744. Also, please check out our [contributing guide](https://filamentphp.com/docs/3.x/support/contributing#development-of-new-features) which talks about proposing the feature before submitting the PR, to ensure you aren't spending time building something that we won't merge. I am not convinced of the feature yet so discussing it would be best first. For example, why not just check `$post->type` or `$post::class`?

juliomotol:
> we are no longer accepting new features for v3 as per #13744

I'm sorry, I totally forgot about this. 

> For example, why not just check `$post->type` or `$post::class`?

The issue arises on the `viewAny` policy since there are no actual records passed. 

I'll go ahead and make a discussion thread for this. thanks!

PR #14123, opened 2024-09-02
Export directory made configurable.

iyashpal:
## Description
I have been working with digital ocean spaces. After the successful setup of the dev, stage, and prod environment. I can configure the upload directory for the form fields but cannot set it for the exports. I end up with a default export directory for all environments. The issue was fixed by configuring the `default_export_directory` in the config file and overriding the `getFileDirectory` method in the `Filament\Actions\Exports\Models\Export` model to utilize this new setting.

 

## Visual changes

Before:
image

After:
image

## Functional changes

### DIFF
--- a/packages/actions/src/Exports/Models/Export.php
+++ b/packages/actions/src/Exports/Models/Export.php
@@ -96,6 +96,6 @@ class Export extends Model
 
     public function getFileDirectory(): string
     {
-        return 'filament_exports' . DIRECTORY_SEPARATOR . $this->getKey();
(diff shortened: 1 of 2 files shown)


danharrin:

Also, I'm unsure if I would accept this PR for v4, since I think you can achieve what you're looking for using the `root` config option of `config/filesystems.php`, which applies a prefix to all files in the disk. You could add an env var set to `development` `stage` or `production`

iyashpal:
@danharrin I think I missed this `root` configuration. Thank you so much.

PR #14124, opened 2024-09-02
Use assertEqualsCanonicalized() for summary testing

cheesegrits:
## Description

The assertTableColumnSummarySet() method currently fails for Values summaries if the expected state array is not sorted the same as the actual value.  As the ordering/sorting of the Values summary is not specified, this makes testing difficult.  Using assertEqualsCanonicalized() instead of assertSame() fixes this.

Note that other potential differences between using an "Equals" test vs a "Same" test are not applicable, as we internally normalize strings vs numeric before calling the assertion anyway.

Also added a line to the doc noting that expected and actual values are normalized.

## Visual changes

No visual changes

## Functional changes

### DIFF
--- a/packages/tables/docs/12-testing.md
+++ b/packages/tables/docs/12-testing.md
@@ -638,7 +638,7 @@ it('can average values in a column', function () {
 
 The first argument is the column name, the second is the summarizer ID, and the third is the expected value.
 
-Note that the expected and actual values are normalized, such that 123.12 is considered the same as "123.12", and ['Fred', 'Jim'] is the same as ['Jim', 'Fred'].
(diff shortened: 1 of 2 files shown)


packages/tables/docs/12-testing.md
```diff
@@ -638,6 +638,8 @@ it('can average values in a column', function () {
 
 The first argument is the column name, the second is the summarizer ID, and the third is the expected value.
 
+Note that the expected and actual values are normalized, such that 123.12 is considered the same as "123.12", and ['Fred', 'Jim'] is the same as ['Jim', 'Fred'].
```
danharrin:
```suggestion
Note that the expected and actual values are normalized, such that `123.12` is considered the same as `"123.12"`, and `['Fred', 'Jim']` is the same as `['Jim', 'Fred']`.
```

danharrin (changes requested):

danharrin:
Thanks

PR #14127, opened 2024-09-03
Add slideOver support to RelationManager actions

FdClemente:
## Description

Introduced a new boolean property $useSlideOver in the RelationManager to allow opening forms in a slide-over instead of a modal. This enhancement provides a smoother and less intrusive user experience when interacting with forms, particularly useful for workflows that benefit from preserving the context of the underlying page. The configureCreateAction and configureEditAction methods have been updated to use this new property, enabling developers to easily toggle between modal and slide-over views.

## Visual changes

Since the introduction of the $useSlideOver property affects the UI, the changes primarily involve the replacement of modal dialogs with slide-over panels when the property is set to true.

## Functional changes

### DIFF
--- a/packages/panels/src/Resources/RelationManagers/RelationManager.php
+++ b/packages/panels/src/Resources/RelationManagers/RelationManager.php
@@ -88,6 +88,8 @@ class RelationManager extends Component implements Actions\Contracts\HasActions,
 
     protected static ?string $badgeTooltip = null;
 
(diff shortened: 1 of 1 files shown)


danharrin:

In this case, the user can also just pass `slideOver()` to each action object instead of needing to use this, so I'm unsure if we'd accept this feature in v4 either?

PR #14145, opened 2024-09-04
Change plural to singular in getInverseRelationship in HasQuery

### LINKED ISSUE #14090: AssociateAction derives wrong inverse relationship name
Inverse relationship is derived as plural of the model name while it should be singular.

pelmered:
## Description

This fixes #14090 where the inverse of a one to many-relationship in the `AssociateAction` uses plural for the relation name and subsequently also the foreign key. This is according to the Laravel convention. 

## Visual changes

N/a

## Functional changes

- [ ] Changes from the inverse relationship name from plural to singular in for example `AssociateAction` 

Looking at the Laravel documentation as an example: https://laravel.com/docs/11.x/eloquent-relationships#one-to-many-inverse. 
On the post resource we add a relationship manager for comments, and in the relationship manager we add an `AssociateAction` to associate comments with the current post. 
The current code would look for the inverse relationship with the name `posts`  (plural) instead of `post` (singular). 

This is only tested for one to many-relationships and many to many-relationships. The `$relationship instanceof HasMany` if-statement probably needs to be updated to cover more cases.


packages/tables/src/Table/Concerns/HasQuery.php
```diff
@@ -144,9 +145,19 @@ public function getInverseRelationship(): ?string
...
+        $inverseRelationship = $this->evaluate($this->inverseRelationship);
+
+        if ($inverseRelationship) {
+            return $inverseRelationship;
+        }
+
+        $inverseRelationship = str(class_basename($relationship->getParent()::class));
+
+        if ($relationship instanceof HasMany) {
+            return (string) $inverseRelationship->singular()->camel();
```
danharrin: Can you give an example of how `singular()` on this line affects the output please? I assume the parent model class name to already be singular.

Also, I believe the check above should be for the `HasOneOrMany` class instead of just `HasMany`

danharrin (changes requested):

danharrin:
Hi @pelmered are you still planning to work on this?

pelmered:
> Hi @pelmered are you still planning to work on this?

Sorry, I have been really busy the last few weeks. And I will probably be busy for the next few weeks as well. 
I would be happy if someone took over. Otherwise I can work on it when I have a bit less things on my table, but that will probably not be until mid October.

packages/tables/src/Table/Concerns/HasQuery.php
zepfietje: You're totally right about both!

zepfietje:
No worries, @pelmered!
I just took over this PR, implementing the comments by @danharrin.
Tested on a real project and it does fix the bug.

Ready for merge! 😊🚀

pelmered:
> No worries, @pelmered! I just took over this PR, implementing the comments by @danharrin. Tested on a real project and it does fix the bug.
> 
> Ready for merge! 😊🚀

Great! Thank you! 😁🚀

PR #14157, opened 2024-09-05
Database Notifications - Customizable "Mark All as Read" Action

leandrocfe:
## Description

This feature introduces the `databaseNotificationsMarkAllAsReadAction()` method, which allows passing a closure that returns a custom action. It enables full customization of the **Mark All as Read** button using the same methods available for **customizing trigger buttons**.

```php
{
    return $panel
        ...
        ->databaseNotifications()
        ->databaseNotificationsMarkAllAsReadAction(fn (Action $action) => $action
            ->button()
            ->outlined()
            ->size('sm')
            ->color('success')
        )
...
```

## Visual changes

image

## Functional changes


leandrocfe:
If this is approved, I can also submit the Clear Notifications actions.

packages/notifications/src/Livewire/DatabaseNotifications.php
```diff
@@ -28,6 +29,8 @@ class DatabaseNotifications extends Component
 
     public static ?string $authGuard = null;
 
+    public static ?Action $markAllNotificationsAsReadAction = null;
```
danharrin: Instead of storing the action, please add `HasActions` and `InteractsWithActions` to this component, and then rename `getMarkAllNotificationsAsReadAction` to `markAllNotificationsAsReadAction`. Then you can use `$this->markAllNotificationsAsReadAction` to render

danharrin (changes requested):

packages/notifications/resources/views/database-notifications.blade.php
```diff
@@ -1,6 +1,7 @@
 @php
     $notifications = $this->getNotifications();
     $unreadNotificationsCount = $this->getUnreadNotificationsCount();
+    $markAllNotificationsAsReadAction = $this->markAllNotificationsAsReadAction();
```
danharrin: Since you're now using `InteractsWithActions`, you can access the action as a property directly, and it will be cached

```php
:mark-all-notifications-as-read-action="$this->markAllNotificationsAsReadAction"
```

packages/notifications/src/Livewire/DatabaseNotifications.php
danharrin: You won't need this line anymore

packages/notifications/src/Livewire/DatabaseNotifications.php
```diff
@@ -113,6 +119,19 @@ public function getTrigger(): ?View
         return view($viewPath);
     }
 
+    public function markAllNotificationsAsReadAction(): Action
+    {
+        if ($action = static::$markAllNotificationsAsReadAction) {
+            return $action;
+        }
+
```
danharrin: We don't need this

danharrin:
@leandrocfe please check my remaining comments

leandrocfe:
> @leandrocfe please check my remaining comments

@danharrin did you see my answers?

danharrin:
I don't see any responses, sorry

leandrocfe:
> I don't see any responses, sorry

I fixed it implementing as you told me
https://github.com/filamentphp/filament/pull/14157#issuecomment-2528142984

But I think we need

```php
if ($action = static::$markAllNotificationsAsReadAction) {
    return $action;
}
```

Because when we are using Notifications in standalone mode, we can do this in the provider:

```php
public function boot(): void
{
    DatabaseNotifications::trigger('notifications-trigger');
    DatabaseNotifications::$markAllNotificationsAsReadAction = Action::make('markAllNotificationsAsRead')
        ->label('Mark all as read')
        ->button()
        ->action('markAllNotificationsAsRead');
}
```

leandrocfe:
> I don't see any responses, sorry

Hey @danharrin,

I fixed it implementing `HasActions` and `HasForms` as you told me:
https://github.com/filamentphp/filament/pull/14157/commits/d71eb2d8e38aad7b213581a10768bc9842a0f20a

But I think we need this

```php
if ($action = static::$markAllNotificationsAsReadAction) {
    return $action;
}
```

Because when we are using Notifications in standalone mode, we can do this in the provider:

```php
public function boot(): void
{
    DatabaseNotifications::trigger('notifications-trigger');
    DatabaseNotifications::$markAllNotificationsAsReadAction = Action::make('markAllNotificationsAsRead')
        ->label('Mark all as read')
        ->button()
        ->action('markAllNotificationsAsRead');
}
```

Does it make sense?

danharrin:
Hey @leandrocfe, I refactored this instead to allow the user to define their own database notifications LW component, where they can override these methods. Otherwise, I am worried the panel configuration is starting to get very large.

PR #14165, opened 2024-09-06
Fix autofocus not working in wizard steps

### LINKED ISSUE #14164: Autofocus inside a wizard step still gets ignored
### Package

filament/filament

### Package Version

v3.2.110

### Laravel Version

v11.20

### Livewire Version

_No response_

### PHP Version

8.3.9

### Problem description

It was reported that "If you set autofocus->() to an input inside a step inside a wizard - it will be ignored." here https://github.com/filamentphp/filament/issues/4301

And there are a few PRs to fix this
https://github.com/filamentphp/filament/pull/4408
https://github.com/filamentphp/filament/pull/10113

But the autofocus is still not working, even though everything seems to be wired.

### Expected behavior

Autofocus should work inside a wizard.

### Steps to reproduce

Create a wizard.
Put an input inside a step.
Set autofocus to true on the input.

https://github.com/dissto/filament

### Relevant log output

_No response_

mokhosh:
## Description

Fixes https://github.com/filamentphp/filament/issues/14164

## Functional changes

    - [x] I tried to check `blame` and the history of `step.blade.php` but couldn't figure out why `tabindex` was put there in the first place.

### DIFF
--- a/packages/forms/resources/views/components/wizard/step.blade.php
+++ b/packages/forms/resources/views/components/wizard/step.blade.php
@@ -12,6 +12,17 @@
 @endphp
 
 <div
+    x-init="
+        const allChildren = $el.getElementsByTagName('*')
+
+        for (const el of allChildren) {
+            if (el.hasAttribute('autofocus')) {
+                return;
+            }
+        }
+
+        $el.setAttribute('tabindex', '0')
+    "
     x-bind:class="{
         @js($activeStepClasses): step === @js($id),
         @js($inactiveStepClasses): step !== @js($id),
--- a/packages/forms/resources/views/components/wizard/step.blade.php
+++ b/packages/forms/resources/views/components/wizard/step.blade.php
@@ -12,7 +12,7 @@
 @endphp
 
 <div
-    x-bind:tabindex="$el.querySelector('[autofocus]') || '0'"
+    x-bind:tabindex="$el.querySelector('[autofocus]') ? '-1' : '0'"
     x-bind:class="{
         @js($activeStepClasses): step === @js($id),
         @js($inactiveStepClasses): step !== @js($id),
--- a/packages/forms/resources/views/components/wizard/step.blade.php
+++ b/packages/forms/resources/views/components/wizard/step.blade.php
@@ -13,13 +13,7 @@
 
 <div
     x-init="
-        const allChildren = $el.getElementsByTagName('*')
-
-        for (const el of allChildren) {
-            if (el.hasAttribute('autofocus')) {
-                return
-            }
-        }
+        if ($el.querySelector('[autofocus]')) return
 
         $el.setAttribute('tabindex', '0')
     "
--- a/packages/forms/resources/views/components/wizard/step.blade.php
+++ b/packages/forms/resources/views/components/wizard/step.blade.php
@@ -30,7 +30,6 @@
                 'aria-labelledby' => $id,
                 'id' => $id,
                 'role' => 'tabpanel',
-                'tabindex' => '0',
             ], escape: false)
             ->merge($getExtraAttributes(), escape: false)
             ->class(['fi-fo-wizard-step outline-none'])
--- a/packages/forms/resources/views/components/wizard/step.blade.php
+++ b/packages/forms/resources/views/components/wizard/step.blade.php
@@ -17,7 +17,7 @@
 
         for (const el of allChildren) {
             if (el.hasAttribute('autofocus')) {
-                return;
+                return
             }
         }
 
--- a/packages/forms/resources/views/components/wizard/step.blade.php
+++ b/packages/forms/resources/views/components/wizard/step.blade.php
@@ -12,11 +12,7 @@
 @endphp
 
 <div
-    x-init="
-        if ($el.querySelector('[autofocus]')) return
-
-        $el.setAttribute('tabindex', '0')
-    "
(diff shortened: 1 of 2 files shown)


zepfietje:
It seems that tabindex has been on the element since the beginning.
Since it's similar for tabs and relation managers, I don't feel comfortable merging this change without understanding why tabindex was used in the first place.

That doesn't mean this PR can never be merged, but it'd be better if we first understood the use of tabindex so we can make this change across all components when needed.

danharrin:
> In HTML, the value tabindex="0" indicates that an element should be included in the default navigation order, based on its position in the source code. This allows elements that aren't normally focusable, like <div>, <span>, and <p>, to be accessible by keyboard.

So since this is on a `<div>`, it suggests that it should be focusable. However, I don't think it should be, this is just the container for the form components and not the form components themselves. This is also not the button that actually opens the step. So I think it can safely be removed.

danharrin:
I don't see it on relation managers, if it was on a relation manager it would be on the table itself, not the tab button. This view is not for a tab button.

danharrin:
Actually, I think I have found the reason here: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/#accessibilityfeatures

danharrin:
@mokhosh Maybe a better solution involves JavaScript, which checks if any elements with the `autofocus` attribute exist inside the container, and if there are then we don't include this attribute.

mokhosh:
@danharrin @zepfietje 

I'm checking if any of the attributes have autofocus, if they don't I add the `tabindex` to the `step`.

I also found another issue in my tests. That the input `wrapper` also gets the `tabindex` as part of the `attributes` and interferes with focusing. Fixed that too.

Tested this in a project and works perfectly with backwards compatibility.

packages/forms/resources/views/components/wizard/step.blade.php
```diff
@@ -12,6 +12,17 @@
...
 
 <div
+    x-init="
+        const allChildren = $el.getElementsByTagName('*')
+
+        for (const el of allChildren) {
+            if (el.hasAttribute('autofocus')) {
+                return;
+            }
+        }
```
danharrin: How about using `$el.querySelector('[autofocus]')` to check if there are any elements with that attribute instead of fetching everything and looping through

mokhosh: Lol, you know what, that sounds smarter.

Done.

packages/forms/resources/views/components/wizard/step.blade.php
```diff
@@ -12,6 +12,11 @@
 @endphp
 
 <div
+    x-init="
+        if ($el.querySelector('[autofocus]')) return
+
+        $el.setAttribute('tabindex', '0')
+    "
```
zepfietje:
```suggestion
    x-init="$el.querySelector('[autofocus]') && $el.setAttribute('tabindex', '0')"
```

zepfietje: This should be terser and still work right?
Actually, can't we just use `x-bind:tabindex` instead?

packages/support/resources/views/components/input/wrapper.blade.php
```diff
@@ -85,7 +85,7 @@
     @endif
     {{
         $attributes
-            ->except(['wire:target'])
+            ->except(['wire:target', 'tabindex'])
```
zepfietje, line 88: What is this needed for?

zepfietje (changes requested):

packages/support/resources/views/components/input/wrapper.blade.php
mokhosh: This is because the blade compiler duplicates all attributes of the input on the wrapper div including the tabindex which is not the intention and creates another unwanted tab target and steals the focus.

packages/forms/resources/views/components/wizard/step.blade.php
mokhosh: I'll test both and let you know.

mokhosh:
Thanks for all the suggestions @danharrin and @zepfietje. It's looking lovely.

I think we're there guys.

packages/forms/resources/views/components/wizard/step.blade.php
```diff
@@ -12,6 +12,7 @@
 @endphp
 
 <div
+    x-bind:tabindex="$el.querySelector('[autofocus]') || '0'"
```
zepfietje: This reads a little nicer.
Could you check if this still works fine (I would expect so), @mokhosh?

```suggestion
    x-bind:tabindex="$el.hasAttribute('autofocus') || '0'"
```

zepfietje (changes requested):

packages/forms/resources/views/components/wizard/step.blade.php
mokhosh: No, it doesn't because it's not about this element having the `autofocus` attribute, it's about having any children that have `autofocus`.

zepfietje: Obviously, I forgot 🤦‍♂️🤣

mokhosh:
@danharrin Sorry for mentioning. Is there anything blocking this from being merged? I'm only asking because there's been a couple of releases recently.

danharrin:
This looks good to me, thanks

PR #14170, opened 2024-09-06
Adding table actions header cell new hook class

Crowly34:
## Description

Adding an additional hook class for the action table cell header to make it easy to add my own custom theme, I followed the structure of the others, we could also just use `fi-ta-header-cell`, I'm open to suggestions


danharrin (changes requested):
There are a few other places where this will also need applying, like on the `header-cell` above when the cell has a label, and for the other action column positions (before columns, after columns)

Also I think it should be `fi-ta-actions-header-cell` as per `fi-ta-summary-header-cell`

PR #14178, opened 2024-09-07
The toggleAllCheckboxes function should not check a disable checkbox

### LINKED ISSUE #14137: CheckboxList disabled are present in data form
### Package

filament/forms

### Package Version

v3.2.110

### Laravel Version

v10

### Livewire Version

3.4.12

### PHP Version

php 8.3

### Problem description

when I make a checkboxlist with one is disabled I can change the status with bulk action
image
after click on select all
image

and on the data array I can see I have 4 rows of data

image

### Expected behavior

Normally the "select all" must select only enabled row

and even if some box are checked but disabled they must not appear in the data array 

### Steps to reproduce

in a form code

1. make a checkbox list
2. use disableOptionWhen to disable one 
3. set bulkToggleable at true

in view

click on "select all" => all the row will be selected
validate =>the disabled row will be sent 

example of code

```php    
    {
        return $form
            ->schema([
                CheckboxList::make('technologies')
                    ->bulkToggleable()
                    ->options([
                        'tailwind' => 'Tailwind CSS',
                        'alpine' => 'Alpine.js',
                        'laravel' => 'Laravel',
                        'livewire' => 'Laravel Livewire',])
                    ->disableOptionWhen(fn (string $value): bool => $value === 'livewire')
            ]);
    }
```

https://github.com/ricksieper/filament_checkboxlist

### Relevant log output

_No response_

hamidroohani:
Close #14137

## Description

There is a function in the js part of the checkbox named `toggleAllCheckboxes`, that will effect on all checkboxes without considering the disabled attribute. in this merge request I made a condition to prevent changing disable checkboxxes.

## Functional changes

- [ ] The toggleAllCheckboxes in JS part of checkbox will effect only on non disable checkboxes

### DIFF
--- a/packages/forms/resources/views/components/checkbox-list.blade.php
+++ b/packages/forms/resources/views/components/checkbox-list.blade.php
@@ -33,8 +33,10 @@
                 this.visibleCheckboxListOptions.forEach((checkboxLabel) => {
                     checkbox = checkboxLabel.querySelector('input[type=checkbox]')
 
-                    checkbox.checked = state
-                    checkbox.dispatchEvent(new Event('change'))
+                    if (!checkbox.disabled) {
+                        checkbox.checked = state
+                        checkbox.dispatchEvent(new Event('change'))
+                    }
                 })
 
                 this.areAllCheckboxesChecked = state
--- a/packages/forms/resources/views/components/checkbox-list.blade.php
+++ b/packages/forms/resources/views/components/checkbox-list.blade.php
@@ -33,10 +33,12 @@
                 this.visibleCheckboxListOptions.forEach((checkboxLabel) => {
                     checkbox = checkboxLabel.querySelector('input[type=checkbox]')
 
-                    if (!checkbox.disabled) {
-                        checkbox.checked = state
-                        checkbox.dispatchEvent(new Event('change'))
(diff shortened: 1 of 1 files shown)


danharrin (commented):
Instead, maybe modify the query selector on line 15 to not select checkboxes with the `disabled` attribute? Then those checkboxes wont end up in the array in the first place

hamidroohani:
> Instead, maybe modify the query selector on line 15 to not select checkboxes with the `disabled` attribute? Then those checkboxes wont end up in the array in the first place

the thing is, those are labels and not checkboxes, how ever we can filter the labels like:

```
checkboxListOptions: Array.from(
    $root.querySelectorAll('.fi-fo-checkbox-list-option-label')
).filter(label => {
    const checkbox = label.querySelector('input[type=checkbox]');
    return checkbox && !checkbox.disabled;
}),
```

hamidroohani:
@danharrin 
also please remember that If the `disabled` state of a checkbox changes dynamically after the initial load, the list of labels won’t automatically update, unless you re-run the filtering process.
Please tell me your final comment, then if needed I will change it again

danharrin:
Gotcha, good point

danharrin:
Thanks

PR #14187, opened 2024-09-09
fix creating resources in vendor when plugin uses cluster

wychoong:
## Description

when a plugin register cluster discovery in plugin's panel, it register vendor's namespace and path and when runs make:* commands it detected as the only registered namespace and path, it creates resources/page/RM in the vendor directory

example: https://github.com/lunarphp/lunar/blob/2b31dcabc4cc7f169051be63648836b75bf91d18/packages/admin/src/LunarPanelManager.php#L231

added filter to remove directories in vendor `mb_strpos($resourceDirectory, base_path('vendor')) === 0`

## Functional changes

### DIFF
--- a/packages/panels/src/Commands/MakePageCommand.php
+++ b/packages/panels/src/Commands/MakePageCommand.php
@@ -205,24 +205,25 @@ class MakePageCommand extends Command
                 $pageDirectories[array_search($namespace, $pageNamespaces)] :
                 (Arr::first($pageDirectories) ?? app_path('Filament/Pages/'));
         } else {
-            $resourceDirectories = [
-                app_path('Filament/Resources/'),
-                ...$panel->getResourceDirectories(),
-            ];
-            $resourceNamespaces = [
-                'App\\Filament\\Resources',
-                ...$panel->getResourceNamespaces(),
-            ];
+            $resourceDirectories = $panel->getResourceDirectories();
+            $resourceNamespaces = $panel->getResourceNamespaces();
+
+            foreach ($resourceDirectories as $resourceIndex => $resourceDirectory) {
+                if (mb_strpos($resourceDirectory, base_path('vendor')) === 0) {
+                    unset($resourceDirectories[$resourceIndex]);
+                    unset($resourceNamespaces[$resourceIndex]);
+                }
+            }
 
             $resourceNamespace = (count($resourceNamespaces) > 1) ?
                 select(
                     label: 'Which namespace would you like to create this in?',
                     options: $resourceNamespaces
                 ) :
-                Arr::first($resourceNamespaces);
+                (Arr::first($resourceNamespaces) ?? 'App\\Filament\\Resources');
             $resourcePath = (count($resourceDirectories) > 1) ?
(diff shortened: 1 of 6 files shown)


wychoong:
cc: @glennjacobs

danharrin:
Please create a test project with https://github.com/InterNACHI/modular, with a module that contains a plugin that discovers resources. Register the plugin in the app's admin provider. Then check that you can still create files within the module and they aren't stripped out (because technically the module is symlinked into `/vendor`, so I don't know how this is going to behave or break)

wychoong:
> Please create a test project with https://github.com/InterNACHI/modular, with a module that contains a plugin that discovers resources. Register the plugin in the app's admin provider. Then check that you can still create files within the module and they aren't stripped out (because technically the module is symlinked into `/vendor`, so I don't know how this is going to behave or break)

@danharrin it seems to be working fine
the test repo: https://github.com/wychoong/filament-modular  (change your local repo)

the options when running `filament:resource`
image

dd of `$resourceDirectories`
image

danharrin (changes requested):
Can you apply this to discovered page, widget and cluster directories too, not just for resources?

Also, please use `str($resourceDirectory)->startsWith(base_path('vendor'))` for readability in these checks

wychoong:
> Can you apply this to discovered page, widget and cluster directories too, not just for resources?

do you think its safe to filter in  `Filament\Panel\Concerns\HasComponents` trait?

danharrin:
I don't think so, no? Because I think we use those directories for actual file discovery, not just these commands

wychoong:
> Can you apply this to discovered page, widget and cluster directories too,

done

packages/panels/src/Commands/MakePageCommand.php
```diff
@@ -70,10 +70,20 @@ public function handle(): int
...
+                unset($resourceDirectories[$resourceIndex]);
+                unset($resourceNamespaces[$resourceIndex]);
+            }
+        }
+
         $resourceInput = $this->option('resource') ?? suggest(
             label: 'Which resource would you like to create this in?',
             options: collect($panel->getResources())
-                ->filter(fn (string $namespace): bool => str($namespace)->contains('\\Resources\\'))
+                ->filter(fn (string $namespace): bool => str($namespace)->contains('\\Resources\\') && str($namespace)->startsWith($resourceNamespaces))
```
danharrin, line 86: Can you explain this please?

wychoong: In a plugin where it register resources `->resources(...)`, when using the `filament:page` command it will list all resources including in vendor. this added check only allow non-vendor resource namespaces

before 
image

after
image

danharrin: Does this also need to happen for resource widgets?

wychoong: not really, the whole codebase only this command has this `suggest`

danharrin: Alright thanks

PR #14193, opened 2024-09-10
Fix: Prevent Livewire Modal Error on HTTP 419

mischasigtermans:
This fix resolves the Livewire modal error caused by HTTP 419 for users with disabled cookies, allowing Filament admin panels to function without interruptions or unnecessary popups.


danharrin:
Thanks, but as discussed too risky and not enough explanation about the problem yet

PR #14200, opened 2024-09-11
fix: Unmounting actions dispatches close-modal when modal is already closed

### LINKED ISSUE #10581: Modal Actions
### Package

filament/filament

### Package Version

v3.1

### Laravel Version

v10

### Livewire Version

v3

### PHP Version

PHP 8.2

### Problem description

When the modal (any modal) opened and closed the all modals in the same component will become unresponsive until the first modal completed.
This problem will appear obviously if you have slow connection or your server is slow, and this will be very annoying.

I have reported this issue before and was told that there is no solution and that appropriate indicators have been used.

But when the modal close the button to open the modal will be disabled but if i am clicked directly on this button it will give me loading indicator after a while and nothing will happened because the network requests not finished at this moment (this will happen if connection is slow).

### Expected behavior

Why we need two network requests after modal closing?
Is that because the first request using to dispatch event from livewire and livewire will dispatch the event from the Alpinejs which will send the second request?
If this case we can send the event from Alpinejs directly.

And we need more robust and expressive indicator in this case.

Finally why you need to stop opening modals when previous modals closing? is this for refreshing table because some records may deleted or updated? can we disable actions only on these records?

### Steps to reproduce

demo with set throttle on network

### Reproduction repository

### Relevant log output

_No response_

wout-o:
## Description

This addresses the issue described in https://github.com/filamentphp/filament/issues/10581.

[The modal's close() function](https://github.com/filamentphp/filament/blob/5a8b9994e9356a0d0702b7c307d811496440ef12/packages/support/resources/views/components/modal/index.blade.php#L73)  already closes the current modal and it also dispatches the `modal-closed` event, which when [caught](https://github.com/filamentphp/filament/blob/5a8b9994e9356a0d0702b7c307d811496440ef12/packages/actions/resources/views/components/modals.blade.php#L40) will cause the action's unmount function to be called. This function, however, does not take this into account and will always dispatch the `close-modal` event again.

This PR introduces a new variable `$shouldDispatchCloseModal` for the actions' unmount function, which when set to `false` will prevent the `close-modal` event from being dispatched again. Since this variable has a default value of `true` it should not break any other uses of the functions.

## Visual changes

Improved UX when rapidly closing and reopening modals on complex pages or when the user has a slower connection.

### Actions
Delete button
Before: [https://demo.filamentphp.com/shop/orders/1/edit](https://demo.filamentphp.com/shop/orders/1/edit)
After: [https://demo.filament.thenines.be/shop/orders/1/edit](https://demo.filament.thenines.be/shop/orders/1/edit)

### Table actions
Edit button
Before: [https://demo.filamentphp.com/blog/categories](https://demo.filamentphp.com/blog/categories)
After: [https://demo.filament.thenines.be/blog/categories](https://demo.filament.thenines.be/blog/categories)

### Form actions
Reset button
Before: [https://demo.filamentphp.com/shop/orders/1/edit](https://demo.filamentphp.com/shop/orders/1/edit)
After: [https://demo.filament.thenines.be/shop/orders/1/edit](https://demo.filament.thenines.be/shop/orders/1/edit)

### Infolist Actions
Author's suffix action
Before: _not present in the demo_
After: [https://demo.filament.thenines.be/blog/posts/1](https://demo.filament.thenines.be/blog/posts/1)

## Functional changes

### DIFF
--- a/packages/actions/resources/views/components/modals.blade.php
+++ b/packages/actions/resources/views/components/modals.blade.php
@@ -37,7 +37,7 @@
                     return
                 }
 
-                $wire.unmountAction(false)
+                $wire.unmountAction(false, false)
             "
             x-on:opened-form-component-action-modal.window="if ($event.detail.id === '{{ $this->getId() }}') close()"
         >
@@ -99,7 +99,7 @@
                     return
                 }
 
-                $wire.unmountTableAction(false)
+                $wire.unmountTableAction(false, false)
             "
             x-on:opened-form-component-action-modal.window="if ($event.detail.id === '{{ $this->getId() }}') close()"
         >
@@ -217,7 +217,7 @@
                     return
                 }
 
-                $wire.unmountInfolistAction(false)
+                $wire.unmountInfolistAction(false, false)
             "
             x-on:opened-form-component-action-modal.window="if ($event.detail.id === '{{ $this->getId() }}') close()"
         >
@@ -271,7 +271,7 @@
                 const mountedFormComponentActionShouldOpenModal = {{ \Illuminate\Support\Js::from($action && $this->mountedFormComponentActionShouldOpenModal()) }}
 
                 if (mountedFormComponentActionShouldOpenModal) {
-                    $wire.unmountFormComponentAction(false)
+                    $wire.unmountFormComponentAction(false, false)
                 }
             "
         >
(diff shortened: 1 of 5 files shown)


danharrin:
This looks great, thank you so much for taking the time!

krognome:
So this fix is live and I can get this fix by running: composer update filament/* is this correct??

PR #14205, opened 2024-09-11
options limit within ManageRelatedRecords

JamieCee20:
## Description

Found the inability to set an options limit on the attach action that was a part of a ManageRelatedRecords class, since the Select is defined in the function, unable to add options limit to the Select::make() directly

## Functional changes


danharrin:

In this case, we won't merge in v4 either as its already possible - `->recordSelect(fn (Select $select) => $component->optionsLimit(100))`

PR #14220, opened 2024-09-12
fix: Parent modal not reopening after child modal is closed

wout-o:
## Description

This PR fixes a bug I introduced in https://github.com/filamentphp/filament/pull/14200 which prevents parent modals from reopening when a child modal gets closed.

## Functional changes


danharrin:
How about nested table actions within the footer of another table action? Or the same for page/header actions?

wout-o:
> How about nested table actions within the footer of another table action? Or the same for page/header actions?

Those only dispatch the `close-modal` event so these don't require any additional checks.

danharrin:
Thank you

PR #14224, opened 2024-09-12
Update Login.php -> function authenticate()

FilipeNSV:
## Description

This pull request adds a session flush right after the authentication credentials are successfully verified. This ensures that any previous session data is cleared, enhancing session security and preventing conflicts. The session is regenerated to establish a new session ID for the authenticated user.

## Visual changes

No visual changes.

## Functional changes

- Added `session()->flush()` after successful authentication to clear old session data.

## Checklist

### DIFF
--- a/packages/panels/src/Pages/Auth/Login.php
+++ b/packages/panels/src/Pages/Auth/Login.php
@@ -64,6 +64,8 @@ class Login extends SimplePage
             $this->throwFailureValidationException();
         }
 
+        session()->flush();
+
         $user = Filament::auth()->user();
 
         if (


malzariey:
In most cases, you need to store data in the session before login and process it after login.

FilipeNSV:
It's true, but I believe it's important to clean the session after checking
credentials, before login. In my app, I had some problems with this. It's
just a simple suggestion, for me, I believe works better.

Em sex., 13 de set. de 2024 às 05:17, Majid Al Zariey <
***@***.***> escreveu:

danharrin:
This is a breaking change for anyone relying on data set before login.

Also, neither Breeze nor Jetstream/Fortify do this, so I'm prepared to say it's not a best practice according to Laravel.

You can always change the behaviour by overriding this method on the login page in your panel.

FilipeNSV:
Thank you so much for clarifying and taking the time to respond. I now
have a better understanding and truly appreciate your insights.

Em sex., 13 de set. de 2024 às 18:35, Dan Harrin ***@***.***>
escreveu:

PR #14249, opened 2024-09-15
Add feature to show date with iso format in Table & Infolist

agencetwogether:
## Description

Here I added new methods to the TextColumn and TextEntry, which functions to format date directly in real format of app locale. Specially useful for plugins developers.

PS : I know we can have same result with `formatStateUsing` but I think these methods added are like helpers.

## Visual changes
**TextColumn**
Capture d’écran 2024-09-15 à 14 19 28

**TextEntry**
Capture d’écran 2024-09-15 à 14 22 35

## Functional changes

### DIFF
--- a/packages/infolists/docs/03-entries/02-text.md
+++ b/packages/infolists/docs/03-entries/02-text.md
@@ -47,6 +47,15 @@ TextEntry::make('created_at')
     ->dateTime()
 ```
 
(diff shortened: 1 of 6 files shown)


danharrin:

I would accept a PR for this feature when v4 is released, if you are still interested. However, I think that the word `iso` in the feature should be before `date`. So, `isoDate()` etc instead of `dateIso()`. Thank you!

PR #14276, opened 2024-09-17
Add feature to show date with iso format in Table & Infolist

agencetwogether:
Like suggested in [#14249](https://github.com/filamentphp/filament/pull/14249#issuecomment-2354783995) I added this for v4.x

## Description

Here I added new methods to the TextColumn and TextEntry, which functions to format date directly in real format of app locale. Specially useful for plugins developers.

PS : I know we can have same result with `formatStateUsing` but I think these methods added are like helpers.

## Functional changes


danharrin (changes requested):
Can you please add the missing `isoTime()` method to both traits?

agencetwogether:
Hello, 
Sorry for delay, but isoTime method has been added 😉

danharrin:
Thanks

PR #14303, opened 2024-09-20
Add aria label to action th

### LINKED ISSUE #14299: Table actions column header WCAG compliance
### Package

filament/tables

### Package Version

v3.2.113

### Laravel Version

v11.23.5

### Livewire Version

v3.5.6

### PHP Version

8.3.4

### Problem description

I encountered an issue during accessibility checks: when creating a table with an actions column, the table header for that column lacks an accessible name, which is required for elements with the `columnheader` role.

W3C docs: https://www.w3.org/TR/wai-aria-1.2/#columnheader

I found this using the WAVE accessibility evaluation tool. https://wave.webaim.org/

### Expected behavior

Filament currently renders the actions column header as this
```
<th class="w-1"></th>
```
I see 2 solutions to remedy, either;

1. Swap the `<th>` element to a `<td>` so it no longer has the `columnheader` role
```
<td class="w-1"></td>
```
2. Add an `aria-label` to the `<th>` element
```
<th class="w-1" aria-label="actions"></th>
```

### Steps to reproduce

- Clone repository
- Install dependencies
- Make user with `php artisan make:filament-user`
- Login with user at `/admin/login`
- Navigate to users resource
- Use browser element inspector or WAVE extension to see the error with the column header

https://github.com/honeyguppy/filament-issue-table-wcag

### Relevant log output

_No response_

yogesh16:
## Description
This pull requests, add `aria-label="actions"` to action th
Issue #14299 

## Visual changes
Before 
Screenshot 2024-09-20 at 7 26 38 AM
After
Screenshot 2024-09-20 at 7 26 29 AM

## Functional changes


ju5t:
> the table header for that column lacks an accessible name, which is required for elements with the columnheader role.

This is not true. I agree that it's something that should be added, but it's not required. I would argue that the label should be more descriptive though. From an accessibility standpoint a user wouldn't know what to expect if there are just 'actions'.

Perhaps you can loop over all the defined actions in the table and make it look like:

```html
aria-label="Edit, move or delete records"
```

Or if it's just one action:

```html
aria-label="Edit records"
```

danharrin:
There could be many actions, especially if they are grouped, so I don't think listing them is what we should do here.

Also, this current PR is only suitable for English apps since the string is not translated

ju5t:
@danharrin that's true. Perhaps it's possible to count all specified actions to decide if the translation should be plural?

danharrin:
Sure, but it doesn't just need to adapt to the plural it also needs to be translated

yogesh16:
@danharrin Fixed Plural and added localization.
I will update other languages as well.

danharrin:
Thanks

PR #14314, opened 2024-09-21
Correct view path in notifications documentation

Dinesh-loma:
corrected the view path 

Previous : filament-notifications.notification

correct one : notifications.notifications

##  path mentioned in view() was wrong. It was giving error. So I changed the path , and use correct path which was mention in example section
issue was in view path
``` 
Notification::configureUsing(function (Notification $notification): void {
    $notification->view('filament-notifications.notification');
});
```

## Visual changes
Before :
image

After : 
image

## Functional changes

No functionality change. just Typo fixed


zepfietje:
Good catch, thanks!

PR #14328, opened 2024-09-23
Make table function static in relation manager stub to match base function definition

lupinitylabs:
## Description

For as long as I have been using the relation manager generator, it generated the table prototype as `public function table(Table $table): Table`, which rightfully leads to the IDE complaining that its base function in `Resource.php` is `static`, and thus, the overriding function should be, too.

grafik

This PR fixes this by making the function definition in the stub static.

## Visual changes

None.

## Functional changes

### DIFF
--- a/packages/panels/stubs/RelationManager.stub
+++ b/packages/panels/stubs/RelationManager.stub
@@ -24,7 +24,7 @@ class {{ managerClass }} extends RelationManager
             ]);
     }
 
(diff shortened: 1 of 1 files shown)


lupinitylabs:
Ignore this; obviously the table function is inherited from `InteractsWithTable`, not from `Resource`. 
No idea why PHPStorm gets this wrong. 🙏

PR #14338, opened 2024-09-25
Update 05-viewing-records.md

muesker:
## Description

Currently the documentation for the "infolist" function is shown as non static method but thats not correct.
https://github.com/filamentphp/filament/blob/3.x/packages/panels/src/Resources/Resource.php#L131

## Current Documentation

image

## Functional changes


dissto:
Possibly related to:  https://github.com/filamentphp/filament/pull/11578#issuecomment-1964731064

zepfietje:
Yeah this part of the docs is about adding an infolist to an additional view page of a resource, so the method should not be static.

PR #14340, opened 2024-09-25
Fix parent component dehydrate state when hidden

coleshirley:
## Description

`dehydratedWhenHidden` does not work when a form layout is hidden. I am not completely confident in the fix but the test case should cover the use case I am experiencing 

```
Section::make('Quantity')
   ->schema([
      TextInput::make('quantity_maximum'),
      ...
   ])
   ->hidden(fn(Get $get) => filled($get('other_field')))
   ->dehydratedWhenHidden()
```

## Visual changes

none

## Functional changes


packages/forms/src/Components/Concerns/HasState.php
```diff
@@ -192,8 +192,8 @@ public function dehydrateState(array &$state, bool $isDehydrated = true): void
             }
         }
 
-        foreach ($this->getChildComponentContainers() as $container) {
-            if ($container->isHidden()) {
+        foreach ($this->getChildComponentContainers(withHidden: true) as $container) {
+            if ($container->getParentComponent()?->isHiddenAndNotDehydrated()) {
```
danharrin: I might be wrong, but isn't `$this` the same as `$container->getParentComponent()`?

coleshirley: You are correct! Updated

PR #14343, opened 2024-09-25
Allow for color interface to be used in color

howdu:
## Description
Allow for Enum with ColorInterface to be returned from closure.

```php
TextColumn::make('test')
   ->color(fn (?string $state): ColorEnum => $state ? ColorEnum::Success : ColorEnum::Danger)
```

E.g
```php
use Filament\Support\Colors\Color;
use Filament\Support\Contracts\HasColor as ColorInterface;

enum ColorEnum: string implements ColorInterface
{
    case Danger = 'danger';
    case Success = 'success';

    public function getColor(): string|array|null
    {
        return match ($this) {
            self::Danger => Color::Red,
            self::Success => Color::Green,
        };
    }
}
```

## Visual changes

Without this change it always entered the `filled` if statement and throws error. 
Screenshot 2024-09-24 at 20 02 41

## Functional changes

### DIFF
--- a/packages/forms/resources/js/components/file-upload.js
+++ b/packages/forms/resources/js/components/file-upload.js
@@ -247,7 +247,7 @@ export default function fileUploadFormComponent({
                     .map((file) =>
                         file.source instanceof File
                             ? file.serverId
-                            : this.uploadedFileIndex[file.source] ?? null,
+                            : (this.uploadedFileIndex[file.source] ?? null),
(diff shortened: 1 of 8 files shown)


zepfietje:
What exactly would be the use case of this, @howdu?

howdu:
Actually sorry yeah reconsidering this it's not needed.

PR #14377, opened 2024-09-28
fix: update resource namespace option in commands

wotta:
## Description
Added the missing option to set the resource namespace when calling `make:filament-resource` using the `Artisan::call` method.

Maybe a bit unusual PR but I ran into this while creating some functionality and have "clusters" enabled. While this is enabled I couldn't programmatically run this command anymore because there was no way of adding that option thus getting errors.
This change resolves it for me.

## Visual changes
No visual changes unless you want a screenshot of the new command options. Just let me know😁

## Functional changes

### DIFF
--- a/packages/panels/src/Commands/Aliases/MakeResourceCommand.php
+++ b/packages/panels/src/Commands/Aliases/MakeResourceCommand.php
@@ -10,5 +10,5 @@ class MakeResourceCommand extends Commands\MakeResourceCommand
 {
     protected $hidden = true;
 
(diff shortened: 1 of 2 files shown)


danharrin:
Instead of this flag that just selects the first, shouldnt there be an argument where you pass in the name of the namespace, like the model namespace, so you can select a cluster if you need that? Because I find this PR quite confusing as a true/false option

danharrin:
Hi, the CLI has now been refactored for v4, so if you want to implement this please put it on that branch, with the suggested change above too

PR #14403, opened 2024-09-30
Update 07-advanced.md

Kamav21:
handle nullable , unique field when using 'dehydrateStaaeUsing() ' if field left blank it returns empty string  when using strtolower()

## Functional changes


danharrin:
Hi, I have not experienced such validation errors with null values, and empty string conversions like this are a native Laravel behaviour when parsing requests. I am going to need some more detail as to when this happens in order to evaluate if this is something that needs to be fixed inside Filament instead of documented.

PR #14407, opened 2024-10-01
Extract base stylesheet

zepfietje:
## Description

This pull request extracts the styles added to Tailwind's base layer by Filament.
The goal is to support importing the base stylesheet separately so developers can import Tailwind CSS into a CSS layer if desired.

This change does not affect existing projects or the compiled Filament stylesheet at all.

## Functional changes

### DIFF
--- a/packages/panels/resources/css/base.css
+++ b/packages/panels/resources/css/base.css
@@ -1,0 +1,10 @@
+@layer base {
+    :root.dark {
+        color-scheme: dark;
+    }
+
+    /* When scrolling to validation error, do not hide element behind the top bar */
+    [data-field-wrapper] {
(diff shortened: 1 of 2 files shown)


ThijmenKort:
I was using `@tailwind` in a custom css file. Due to this change my css was broken.
Is there somewhere this is documented? I can't seem to find it in the Filament or Tailwind docs to use `@import`.

zepfietje:
Make sure to only import Tailwind once (which is done in Filament's main stylesheet).
`@import` is needed when having other imports in the CSS file because `@tailwind` won't work then: https://tailwindcss.com/docs/using-with-preprocessors#build-time-imports.

PR #14419, opened 2024-10-02
Added markdown extensions support

Plytas:
## Description

As proposed in https://github.com/filamentphp/filament/discussions/1294#discussioncomment-10812259, this PR adds support for markdown extensions.

One important change - HTML is sanitized before being passed to `markdown()` function. This still sanitizes user input, but allows markdown to produce "unsafe" HTML like `<iframe>`.

I'm starting this as draft PR to receive feedback. I'll be able to update docs once desired API is confirmed. Also, let me know if this needs tests and point me to an example I can follow.

## Visual changes

No visual changes.

## Functional changes


Plytas:
Ping @danharrin @ryangjchandler.

danharrin (changes requested):
I like the idea! Few things:
- I think all the properties and methods should be renamed to `commonMarkOptions()` and `commonMarkExtensions()` for clarity
- I think it should apply to the TextColumn and TextEntry, "summarizer" classes too, which can render Markdown using the `markdown()` method
- Since we're using it in multiple places, extract the properties and methods to a trait `CanConfigureCommonMark` in `Filament\Support\Concerns`

Plytas:
@danharrin I have updated the PR with requested changes.

The failing actions are for Laravel 10. `Str::markdown()` does not accept extensions as it was introduced in [v11.13.0](https://github.com/laravel/framework/releases/tag/v11.13.0). Is there a good way to check if extension support is available and only then pass them in?

danharrin:
Laravel 10 will soon be out of security release support, so I will probably drop it in v4 anyway. I will update this PR when I do so.

Thanks for the rest of it, it looks perfect!

Plytas:
You might have to lock it to at least v11.13.0 then.

danharrin:
Thanks!

PR #14456, opened 2024-10-07
Fix: Add the bind value that is initialized in reorder()

bzy107:
## Description
In the case of PostgreSQL, we are performing reorder during individual processing at the time of export. However, when the query contains bindings, the bindings are initialized, leading to an SQL error.

In this fix, we search for the initialized bind attributes before and after the reorder, and re-bind the initialized values to prevent SQL errors.

## Functional changes


bzy107:
There is an error in PHPStan, but the type of the target is an array.

danharrin (changes requested):
Please move the logic back into the main function as it is only called once

Please fix the PHPStan issue, you may need to add a Docblock `@var`

PR #14480, opened 2024-10-09
Add a new render hook at the end of view record page

krzysztofengineer:
## Description

I need to inject some code at the end of the view record template.It would be much easier if there was a render hook.

### DIFF
--- a/packages/panels/resources/views/resources/pages/view-record.blade.php
+++ b/packages/panels/resources/views/resources/pages/view-record.blade.php
@@ -1,47 +1,45 @@
 <x-filament-panels::page
-    @class([
-        'fi-resource-view-record-page',
-        'fi-resource-' . str_replace('/', '-', $this->getResource()::getSlug()),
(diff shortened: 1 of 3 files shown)


dissto:
Isn't `PanelsRenderHook::SIMPLE_PAGE_END` or `PanelsRenderHook::PAGE_END` just what you need? 🤔

danharrin:
Yeah or `PAGE_FOOTER_WIDGETS_BEFORE`

krzysztofengineer:
Unfortunately I need to target only view pages. I need to call this from a package so I can’t scope only particular resources

danharrin:
In the view you could check `@if ($this instanceof ViewRecord)`

krzysztofengineer:
> In the view you could check `@if ($this instanceof ViewRecord)`

so should I publish a view-record view from a package?

danharrin:
no? in the hook view

PR #14488, opened 2024-10-10
add optional label to Form Fields and Table Column to their __construct.

Abather:
## Description

Most of the time any field or column created needed to give it label instead of doing it in to lines make it one line.

## Visual changes

`Filament\Forms\Components\Field` and `Filament\Tables\Columns\Column` Before:
```
    final public function __construct(string $name)
    {
        $this->name($name);
        $this->statePath($name);
    }
```

`Filament\Forms\Components\Field` and `Filament\Tables\Columns\Column` After:
```
    final public function __construct(string $name, ?string $label = null)
    {
        $this->name($name);

        if (! is_null($label)) {
            $this->label($label);
        }
    }
```

also the change made to `make` function:
```
    public static function make(string $name, ?string $label = null): static
    {
        $static = app(static::class, ['name' => $name, 'label' => $label]);
        $static->configure();

        return $static;
    }
```

## Functional changes

- [ ✅ ] Code style has been fixed by running the `composer cs` command.
- [ ✅ ] Changes have been tested to not break existing functionality.

### DIFF
--- a/packages/forms/src/Components/Field.php
+++ b/packages/forms/src/Components/Field.php
@@ -15,15 +15,20 @@ class Field extends Component implements Contracts\HasHintActions, Contracts\Has
 
     protected string $viewIdentifier = 'field';
 
-    final public function __construct(string $name)
+    final public function __construct(string $name, ?string $label = null)
     {
         $this->name($name);
+
+        if(!is_null($label)) {
+            $this->label($label);
+        }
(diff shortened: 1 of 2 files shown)


danharrin:
Hi, sorry I am not interested in this API change. I prefer the methods to be clearly separate.

PR #14501, opened 2024-10-11
Fix CheckboxList does not display descriptions when using an enum for dynamic options

### LINKED ISSUE #14500: Bug: CheckboxList does not display descriptions when using an enum for dynamic options
### Package

filament/filament

### Package Version

v3.2.108

### Laravel Version

v11.21.0

### Livewire Version

v3.5.6

### PHP Version

8.3.12

### Problem description

When rendering a CheckboxList field with dynamic options where an enum is returned by the options closure, the descriptions are not displayed. This issue occurs even though the enum implements both HasDescription and HasLabel.

Code example:

```php
enum Options: string implements HasDescription, HasLabel
{
    case FOO = 'foo';
    case BAR = 'bar';

    public function getDescription(): ?string
    {
        return match ($this) {
            self::FOO => 'Foo Description',
            self::BAR => 'Bar Description',
        };
    }

    public function getLabel(): ?string
    {
        return match ($this) {
            self::FOO => 'Foo Label',
            self::BAR => 'Bar Label',
        };
    }
}

///

class TestComponent extends Livewire
{
    public function form(Form $form): Form
    {
        return $form
            ->schema([
                CheckboxList::make('test')->options(fn () => Options::class),
            ]);
    }
```

### Expected behavior

The descriptions (Foo Description and Bar Description) should be displayed below their respective labels (Foo Label and Bar Label) in the rendered form.

### Steps to reproduce

1. Render a form with a CheckboxList field
2. Write a closure returning the options
3. Return an enum class which implements HasLabel and HasDescription

https://github.com/standaniels/filament/tree/fix-descriptions-for-dynamic-options

### Relevant log output

_No response_

standaniels:
Closes #14500 

As this is my first pull request I need help with checking if this PR breaks existing functionality.

## Description

When rendering a CheckboxList field with dynamic options where an enum is returned by the options closure, the descriptions are not displayed. This issue occurs even though the enum implements both HasDescription and HasLabel.

## Visual changes

None.

## Functional changes

### DIFF
--- a/packages/forms/src/Components/Concerns/HasDescriptions.php
+++ b/packages/forms/src/Components/Concerns/HasDescriptions.php
@@ -52,13 +52,15 @@ trait HasDescriptions
             $descriptions = $descriptions->toArray();
         }
 
+        $options = $this->evaluate($this->options);
+
         if (
             empty($descriptions) &&
-            is_string($this->options) &&
-            enum_exists($this->options) &&
-            is_a($this->options, HasDescription::class, allow_string: true)
+            is_string($options) &&
+            enum_exists($options) &&
+            is_a($options, HasDescription::class, allow_string: true)
         ) {
-            $descriptions = array_reduce($this->options::cases(), function (array $carry, HasDescription & UnitEnum $case): array {
+            $descriptions = array_reduce($options::cases(), function (array $carry, HasDescription & UnitEnum $case): array {
                 if (filled($description = $case->getDescription())) {
                     $carry[$case?->value ?? $case->name] = $description;
                 }
--- a/tests/src/Forms/Components/CheckboxListTest.php
+++ b/tests/src/Forms/Components/CheckboxListTest.php
@@ -1,0 +1,93 @@
+<?php
+
+use Filament\Forms\Components\CheckboxList;
+use Filament\Forms\Form;
+use Filament\Support\Contracts\HasDescription;
+use Filament\Support\Contracts\HasLabel;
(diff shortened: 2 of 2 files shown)


ceejayoz:
I think this behavior might be intentional - not every `CheckboxList` needs descriptions, even if the enum has them.

You can already do `->options(Enum::class)->description(Enum::class)` and it'll pick it up.

standaniels:
I agree not every `CheckboxList` needs descriptions, but when you pass an enum class to the options method descriptions are automatically added if this enum implements `HasDescription`. See the test in the first commit of this PR:

```php
it('can render descriptions when enum implements HasDescription', function () {
    expect(Options::class)->toImplement(HasDescription::class);
    livewire(TestComponentWithFixedOptions::class)
        ->assertSeeText('Foo Description')
        ->assertSeeText('Bar Description');
});
```

This tests passes in the current version of Filament if you do `->options(Enum::class)`.

EDIT
Also, `descriptions` only accepts `array | Arrayable | Closure`. So `->descriptions(Enum::class)` doesn't work.

danharrin:
The reason why it's like this is because evaluating the options closure twice (once to fetch options, once to fetch descriptions) could cause queries to be duplicated etc. `->options(EnumName::class)` should work, but `->options(fn () => EnumName::class)` won't. If you would like to use a dynamic enum like this then you will need to define your own function in `->descriptions()` to return the array of descriptions.

standaniels:
Alright, I understand. Thanks for the clarification!

PR #14517, opened 2024-10-13
Allow poll action on tables

ahmedash95:
## Description
Since Livewire supports polling with the ability [to trigger actions](https://livewire.laravel.com/docs/wire-poll#:~:text=You%20can%20also%20specify%20an%20action%20to%20fire%20on%20the%20polling%20interval), it would be a great feature to have in Filament tables. With each table refresh, an event could be triggered, or any custom action the developer wishes to initiate. This would add more flexibility to the table’s functionality.

## Visual changes

N/A

## Functional changes

    - I couldn’t find any tests specifically for table polling, so I tested it on a separate test project, and everything seems to be working fine.
    - Will do once this pull request gets accepted!

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -284,8 +284,12 @@
         @endif
 
         <div
-            @if ((! $isReordering) && ($pollingInterval = $getPollingInterval()))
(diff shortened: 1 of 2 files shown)


danharrin:

I do not think I understand the use case for this feature either at the moment.

PR #14523, opened 2024-10-15
Add `assertActionHasGroupedIcon` test helper, document Action's `groupedIcon` behavior

tobias-bischoff:
## Description

I noticed that there's no equivalent `assertActionHasGroupedIcon` to the `assertActionHasIcon` [test helper](https://filamentphp.com/docs/3.x/actions/testing#button-appearance) when we want to assert that an action inside an action group has a specific icon which may differ from its regular icon. In fact, the docs do not mention `groupedIcon` at all, so I added documentation for this as well.

While custom actions [fall back](https://github.com/filamentphp/filament/blob/459c21543377267b8fbbd24bd02e4d8cf6313d42/packages/actions/src/Concerns/HasGroupedIcon.php#L20) to their `icon()`-defined icon if set, prebuilt actions [have a default](https://github.com/filamentphp/filament/blob/459c21543377267b8fbbd24bd02e4d8cf6313d42/packages/actions/src/CreateAction.php#L46). This means that prebuilt actions or their subclasses will show a different icon inside action groups even if the user has defined a different icon with `icon()`. I guess the behavior is intentional to allow setting a [default grouped icon](https://filamentphp.com/docs/3.x/support/icons#replacing-the-default-icons), so I opted to explain this behavior in the new docs rather than proposing to remove it.

## Visual changes

This PR has no visual changes.

## Functional changes

### DIFF
--- a/packages/actions/.stubs.php
+++ b/packages/actions/.stubs.php
@@ -29,6 +29,8 @@ namespace Livewire\Features\SupportTesting {
 
         public function assertActionHasIcon(string | array $name, string $icon): static {}
 
+        public function assertActionHasGroupedIcon(string | array $name, string $icon): static {}
+
         public function assertActionDoesNotHaveIcon(string | array $name, string $icon): static {}
 
         public function assertActionHasLabel(string | array $name, string $label): static {}
--- a/packages/actions/docs/03-trigger-button.md
+++ b/packages/actions/docs/03-trigger-button.md
@@ -128,6 +128,32 @@ Action::make('edit')
 
 <AutoScreenshot name="actions/trigger-button/icon-after" alt="Trigger with icon after the label" version="3.x" />
 
+## Using a different icon inside an action group
+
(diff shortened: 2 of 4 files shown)


danharrin:
Hey! I'm trying to move away from adding new testing helpers, as the number of properties we have on objects would make this impossible to complete, we'd just constantly be adding new helpers for new properties.

In v4, you'll be able to do this `->assertActionExists('name', fn (Action $action): bool => $action->getGroupedIcon === 'expected-icon')`, or use any other getter inside that function

In v3, you could use `->instance()->getAction($name)` like you do inside the testing helper and then check the grouped icon on that

danharrin:
With regards to the docs update, `groupedIcon()` is more of an internal API that I don't want to document yet, but we should fix the bug where `->icon()` doesn't reset the `->groupedIcon()` to null when it is called which should avoid the need for the extra docs

tobias-bischoff:
Hi @danharrin ! Thanks for taking the time to review this. I didn't think of a testing helper as a feature, which indeed it is. Just a tiny amount of time was spent and I learned new things about Filament internals, so no worries! Excited for v4.

I had a hunch that fixing the `->icon()` / `groupedIcon` behavior would be cleaner, but was not sure if this is intended or a bug 😅 I'll draft a separate PR with a fix attempt later this week.

Thank you for your work!

PR #14553, opened 2024-10-17
Improve Repeater's `grid()` method to accept closure that can be evaluated later

tjodalv:
## Description

The `grid()` method of the Repeater component already accepts arrays, integers, or strings. This PR introduces support for using closures that can be evaluated later. If the value `'auto'` is provided to the `grid()` method, a closure will be registered to automatically return the number of items in the Repeater. Additionally, users can provide their own custom closure to calculate the grid size based on specific logic.

## Functional changes

### DIFF
--- a/packages/forms/docs/03-fields/12-repeater.md
+++ b/packages/forms/docs/03-fields/12-repeater.md
@@ -403,6 +403,8 @@ Repeater::make('qualifications')
 
 This method accepts the same options as the `columns()` method of the [grid](../layout/grid). This allows you to responsively customize the number of grid columns at various breakpoints.
 
+You can configure the grid to automatically adjust its size based on the number of items in the repeater by passing `'auto'` to the `grid()` method. Alternatively, you can provide a closure to calculate the grid size dynamically, allowing for custom logic based on the repeater's state or other conditions.
+
 ## Adding a label to repeater items based on their content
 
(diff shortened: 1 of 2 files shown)


niladam:
I think this is a great addition.

I was already looking to this, in order to dynamically set the grid based on some other key.

This allows for doing stuff like this:

```php
TextInput::make('products_number')
    ->label(__('Number of products'))
    ->reactive()
    ->numeric(),

Repeater::make('visible_products')
    ->reactive()
    ->schema([
        Select::make('product_id')
            ->label(__('Choose a product'))
            ->searchable()
            ->options(fn() => Product::active()->pluck('title', 'id')),
    ])
    ->usesPosition()
    ->hiddenLabel()
    ->grid(fn(Get $get) => $get('products_number') ?? 4)
```


Good job @tjodalv and thank you!

danharrin:

I would accept a PR for this feature when v4 is released, if you are still interested, with the following changes:

- Not sure how I feel about `auto` at the moment, feels a too bit magic and probably could be accomplished with `->grid(fn (array $state): int => count($state))`. Let's not include that
- The logic will need to be fixed so that when calling `->grid()` multiple times, it will merge onto the array instead of replacing the array, as per the original behaviour that you have removed.

Thank you!

PR #14558, opened 2024-10-17
Have next and previous functions disabled when their actions are disabled in the Wizard component

whoisnegrello:
## Description

It's a small fix to have the next and previous functions disabled when their actions are disabled in the Wizard component

## Visual changes

image

## Functional changes


danharrin (changes requested):
Please save the `$previousAction` and `$nextAction` into variables for reuse. Then wrap the `x-on:click` with an `@if` checking `$previousAction->isDisabled()` etc. Then you can remove the changes to the JavaScript.

danharrin:
Thanks

PR #14566, opened 2024-10-18
Add TOP and BOTTOM hooks on Sub Navigation

davidwhicker:
## Description

Wrapped current Sub Navigation UL with a DIV which then enabled the use of two new hooks to add views above or below the Sub Navigation.

## Visual changes

No visual change unless a hook is used. Please see TEST HOOK text to show the use of the BOTTOM hook. 

Screenshot 2024-10-18 at 15 59 02

This is my first PR so if anything is not in the correct manner please let me know.

### DIFF
--- a/packages/panels/resources/views/components/page/index.blade.php
+++ b/packages/panels/resources/views/components/page/index.blade.php
@@ -75,7 +75,7 @@
                 <div class="md:hidden">
                     {{ \Filament\Support\Facades\FilamentView::renderHook(\Filament\View\PanelsRenderHook::PAGE_SUB_NAVIGATION_SELECT_AFTER, scopes: $this->getRenderHookScopes()) }}
                 </div>
-            
+
                 @if ($subNavigationPosition === SubNavigationPosition::Start)
                     {{ \Filament\Support\Facades\FilamentView::renderHook(\Filament\View\PanelsRenderHook::PAGE_SUB_NAVIGATION_START_BEFORE, scopes: $this->getRenderHookScopes()) }}
 
--- a/packages/panels/resources/views/components/page/index.blade.php
+++ b/packages/panels/resources/views/components/page/index.blade.php
@@ -64,14 +64,18 @@
             ])
         >
             @if ($subNavigation)
-                {{ \Filament\Support\Facades\FilamentView::renderHook(\Filament\View\PanelsRenderHook::PAGE_SUB_NAVIGATION_SELECT_BEFORE, scopes: $this->getRenderHookScopes()) }}
+                <div class="md:hidden">
+                    {{ \Filament\Support\Facades\FilamentView::renderHook(\Filament\View\PanelsRenderHook::PAGE_SUB_NAVIGATION_SELECT_BEFORE, scopes: $this->getRenderHookScopes()) }}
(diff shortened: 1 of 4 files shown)


danharrin (changes requested):
Please rename these to `PAGE_SUB_NAVIGATION_SIDEBAR_START` and `PAGE_SUB_NAVIGATION_SIDEBAR_END`

davidwhicker:
> Please rename these to `PAGE_SUB_NAVIGATION_SIDEBAR_START` and `PAGE_SUB_NAVIGATION_SIDEBAR_END`

That should be done on the latest commit for you.

danharrin (changes requested):
I've made some changes, to make the hooks more consistent with the rest of the codebase & fix a breaking change related to the hook class. Can you check they still work please?

Please add the hooks to the documentation

davidwhicker (commented):
Works for me. I have updated the docs. 

I did notice that when shrinking down to mobile size, the hook is not visible. I then tried the PAGE_SUB_NAVIGATION_SELECT_AFTER which works on mobile, however the hook is visible on Desktop ( now have both hooks visible with the same content ). I may have to dig a little deeper to get this to work nicely on mobile only as well, unless you know?

davidwhicker:
I have just pushed up a commit for the mobile fix I mentioned in my previous comment. I hope this is acceptable. To me it makes sense to have the PAGE_SUB_NAVIGATION_SELECT_BEFORE and PAGE_SUB_NAVIGATION_SELECT_AFTER hooks only be visible when the Sub Nav SELECT itself is.

danharrin:
Thanks

packages/panels/resources/views/components/page/index.blade.php
```diff
@@ -64,13 +64,17 @@
             ])
         >
             @if ($subNavigation)
-                {{ \Filament\Support\Facades\FilamentView::renderHook(\Filament\View\PanelsRenderHook::PAGE_SUB_NAVIGATION_SELECT_BEFORE, scopes: $this->getRenderHookScopes()) }}
+                <div class="md:hidden">
```
zepfietje, line 67: @danharrin, this breaks the UI on mobile because it adds an additional item in the flex container which has a gap:
Screenshot 2024-11-05 at 11 17 23

danharrin: Can the element be wrapped in a check to see if the render hook has output first then?

zepfietje: That might be a good solution for cases where the render hook isn't used.
However, the issue would still be there when using the render hook and viewing the page on mobile I think.

danharrin: If the render hook is used the space is justified IMO

zepfietje: Why would it be ok then?

danharrin: So there is a normal space between the content you put in the render hook and the select?

zepfietje: Alright, let's apply that fix then.

PR #14591, opened 2024-10-21
Support non-streamed downloads

Robertbaelde:
## Description

Some hosting environments do not support streamed downloads (mainly environments running on lambda like Laravel Vapor or Bref). This makes the export functionality not work because the downloaders assume we can do a streamed download. 

This PR adds a config to enable/disable streamed downloads. When disabled, the downloader will write the contents to one file and uploads it to the disk (so we don't have to re-do the generation on multiple downloads).  

## Visual changes

N/A

## Functional changes

### DIFF
--- a/packages/actions/docs/07-prebuilt-actions/09-export.md
+++ b/packages/actions/docs/07-prebuilt-actions/09-export.md
@@ -691,3 +691,8 @@ public function view(User $user, Export $export): bool
     return $export->user()->is($user);
 }
 ```
+
+## Streamed downloads
+
+Some hosting environments do not support streamed downloads. If you encounter issues with downloading files, you can disable streamed downloads by setting the `filament.supports_stream_downloads` config variable to `false`.
+It is possible to define the location of the temporary file used for generating downloads by setting the `filament.temp_directory` config variable to the desired path.
--- a/packages/actions/src/Exports/Downloaders/CsvDownloader.php
+++ b/packages/actions/src/Exports/Downloaders/CsvDownloader.php
@@ -24,7 +24,7 @@ class CsvDownloader implements Downloader
             return $disk->download($filePath);
         }
 
-        if (!config('filament.supports_stream_downloads', true)) {
+        if (! config('filament.supports_stream_downloads', true)) {
             return $this->handleNonStreamedDownload($disk, $directory, $fileName, $filePath);
         }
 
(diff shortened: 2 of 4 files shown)


danharrin (changes requested):
Please can you remove the part that saves new files to the disk? I am conscious we are already writing a lot of this data to the disk and XLSX files are already generated ahead of time, so we can probably avoid that for now while still allowing non-streamed downloads.

Please be aware that you may hit the memory limit more often if you don't stream

I would also double check that Vapor definitely doesnt support streamed downloads, because that doesn't really sound right to me 🤔

austincarpenter:
To get things working on Vapor, you need to add the `X-Vapor-Base64-Encode` header in the response. (See https://docs.vapor.build/projects/development.html#binary-responses)

I only use the XLSX downloader so I've simply subclassed `XlsxDownloader`, replacing the `__invoke()` function to add the header:

```php
public function __invoke(Export $export): StreamedResponse
{
    return tap(parent::__invoke($export), function (StreamedResponse $response) {
        $response->headers->set('X-Vapor-Base64-Encode', true);
    });
}
```

```php
$this->app->bind(BaseXlsxDownloader::class, XlsxDownloader::class);
```

danharrin:
Is there a way we can detect Vapor within Filament and do this on the users behalf? Instead of disabling streaming?

Robertbaelde:
> Is there a way we can detect Vapor within Filament and do this on the users behalf? Instead of disabling streaming?

For Vapor: `isset($_ENV['VAPOR_SSM_PATH']) || isset($_SERVER['VAPOR_SSM_PATH'])`
For Bref: `env('LAMBDA_TASK_ROOT')` (detects Lambda runtime) or `env('BREF_LOOP_MAX')`

Will try to reshape this PR a bit with the given feedback. I'm afraid for Bref downloads larger than 6MB should come from a disk (s3), see: https://bref.sh/docs/use-cases/http/binary-requests-responses 

~Not sure if/how Vapor mitigates that though~
Edit: Vapor also has the 6MB file limit

danharrin:
Hi, please reopen this if you decide to implement the requested changes. Thank you!

bretto36:
@danharrin I still get 502 Bad Gateway errors when using laravel vapor and filament on large downloads. Even some that are < 1MB 

https://docs.vapor.build/projects/development#binary-responses

Lambda limits responses to 6MB. If you need to serve a larger file, consider returning a signed, temporary S3 URL that your user may use to download the file directly from S3.

----

Even if there was a way to apply custom Downloaders i could create a 3rd party package that people could simply use the formats() method on the Exporter class to push to their own

PR #14601, opened 2024-10-23
Add console command optimization in FilamentServiceProvider

damms005:
Sponsored by a relatively new addition to the Laravel framework: https://laravel.com/docs/11.x/packages#optimize-commands.

This will be helpful in local dev, especially whne clearing the cache and forgetting to clear Filament's cache.

### DIFF
--- a/packages/panels/src/FilamentServiceProvider.php
+++ b/packages/panels/src/FilamentServiceProvider.php
@@ -97,7 +97,7 @@ class FilamentServiceProvider extends PackageServiceProvider
                 ], 'filament-stubs');
             }
 
-            if (method_exists($this, 'optimize')) {
+            if (method_exists($this, 'optimizes')) {
                 $this->optimizes(
                     optimize: 'filament:optimize',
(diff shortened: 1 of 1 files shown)


packages/panels/src/FilamentServiceProvider.php
```diff
@@ -96,6 +96,13 @@ public function packageBooted(): void
                     $file->getRealPath() => base_path("stubs/filament/{$file->getFilename()}"),
                 ], 'filament-stubs');
             }
+
+            if (method_exists($this, 'optimize')) {
+                $this->optimizes(
```
danharrin: Did you test this? I think the method name in the check is wrong.

danharrin (changes requested):

zepfietje:
@danharrin shouldn't this be in the support package's `SupportServiceProvider.php` file?

danharrin:
Yeah good catch, fixed

zepfietje:
Great, thanks for fixing! 🫶

PR #14602, opened 2024-10-23
Add console command optimization in FilamentServiceProvider

damms005:
Sponsored by a relatively new addition to the Laravel framework: https://laravel.com/docs/11.x/packages#optimize-commands.

This will be helpful in local dev, especially whne clearing the cache and forgetting to clear Filament's cache.


danharrin:
Closing as v3 will merge into v4 anyway and we have #14601

damms005:
Was thinking about adding support for 3.x too, since 3.x also supports L11 where this command was added.

Thoughts? @danharrin

danharrin:
the 3.x PR is still open, it will be merged into 4.x too

PR #14604, opened 2024-10-23
Save pagination preference per entity rather than per entire model

juliangums:
## Description

I had multiple users confused about the pagination. Intuitively, one would assume that pagination preference is only for the table they are on. This is true for page tables, but not resource tables and relationship managers. They share the same class name, and therefore every entity shares the same preferences currently. My code change makes this more in line with the initial idea of storing different settings per table.

## Functional changes

### DIFF
--- a/packages/tables/src/Concerns/CanPaginateRecords.php
+++ b/packages/tables/src/Concerns/CanPaginateRecords.php
@@ -88,8 +88,9 @@ trait CanPaginateRecords
     public function getTablePerPageSessionKey(): string
     {
         $table = class_basename($this::class);
+        $id = method_exists($this, 'getOwnerRecord') ? $this->getOwnerRecord()->id : '';
 
-        return "tables.{$table}_per_page";
+        return "tables.{$table}_per_page.{$id}";
     }
 
     /**


danharrin:
Hey, personally I don't like this as a default as changing it on one table should change it on all tables of that type IMO, but you may override it in your project if you wish by adding this method to the table class.

juliangums:
@danharrin how would you go about overriding this? Feels like I would have to do some hacky stuff, copy the trait, modify it and override the namespace etc or is there a simpler solution?

I'm indifferent but I had three users raising this as a bug independently so I think I will have to change this somehow

danharrin:
`getTablePerPageSessionKey` is a method that is present on every Livewire component using the table builder (relation manager class, resource's list/manage class, table widget class), so you can define the method on those classes. You could extract the override into a trait or parent class if you wish.

juliangums:
@danharrin I can find all classes using it and then how do I define the method? As a macro like this in a service provider?
```php
foreach ($classes as $class){
        $class::macro('getTablePerPageSessionKey', function () {
            $table = class_basename($this::class);
            $id = method_exists($this, 'getOwnerRecord') ? $this->getOwnerRecord()->id : '';

            return "tables.{$table}_per_page.{$id}";
        });
}
```
(not working for me :()

danharrin:
hmm no you just have to define it as a method in the class itself

```php
public function getTablePerPageSessionKey(): string
{
    $table = class_basename($this::class);
    $id = method_exists($this, 'getOwnerRecord') ? $this->getOwnerRecord()->id : '';

    return "tables.{$table}_per_page.{$id}";
}
```

danharrin:
In this case you'll only need to add it to relation manager classes

juliangums:
Thanks. That helps. But I have about 30 of them. That'd be a lot of duplicated code. I made it a trait and I am using it everywhere, but it's not a great solution as I might forget to add it to a new class in future. No idea how to enforce such an override consistently across all classes.

danharrin:
Maybe something like a PEST arch test to ensure all classes that extends `RelationManager` use your trait?

juliangums:
Probably the best solution. Thanks @danharrin !

PR #14614, opened 2024-10-24
Fix the missing repeater field alignement feature when using the simple layout

alanpilloud:
## Description

According to [this part of the documentation](https://filamentphp.com/docs/3.x/forms/fields/repeater#aligning-the-add-action-button), we are able to align the Add Button of a repeater. But it doesn't work when we use the simple layout.

This PR adds this feature, inspired by how it is made for the standard layout.

## Visual changes

image

## Functional changes


danharrin:
Thanks

PR #14630, opened 2024-10-25
Updated: Repeater's `grid()` method can except closure as parameter

tjodalv:
## Description

This is a follow up of #14553 

Hi! Thanks for the feedback, and apologies for any oversight on our end regarding the contribution guidelines. I've reviewed the notes, and I've created a new pull request where:

- The 'auto' functionality has been removed, as suggested.
- The logic for the `->grid()` function has been adjusted so that calling `->grid()` multiple times merges onto the array, maintaining the original behavior. Additionally, `->grid()` now accepts a closure. Closure will be executed last allowing it to  override any breakpoint value that is being set previously.

I'm looking forward to any further feedback, and thank you for considering this improvement for v4!

## Functional changes


danharrin:
Thanks

PR #14646, opened 2024-10-28
fix: icon button should be disabled while loading

julieni:
## Description

Icon button should be disabled once clicked, just like the standard button.

Noticed the issue when a customer double-clicked on an icon-button to create an option from a form select component: the modal form is opened twice and the validation process fails.

## Visual changes

None

## Functional changes


danharrin:
Good catch, thanks

PR #14668, opened 2024-10-30
Make StatsOverviewWidget use Filament::Grid

mariomeyer:
## Description

Change `StatsOverviewWidget` to use `Filament::Grid`

Before this change if a user tried to set `getColumns()` on `StatusOverviewWidget` to something greater than 4 it would not properly format the div and end up rendering a single line for each stat.

The code implements `Filament::Grid` and changes `getColumns()` to return `int | array`. For backwards compatibility `int` has been kept and handled in the blade.

### Before

image

### After

image

## Functional changes

### DIFF
--- a/packages/widgets/docs/02-stats-overview.md
+++ b/packages/widgets/docs/02-stats-overview.md
@@ -165,3 +165,22 @@ To disable this behavior, you may override the `$isLazy` property on the widget 
 ```php
 protected static bool $isLazy = false;
 ```
+
+## Changing columns layout
+
+By default, a maximum of 4 stats will apper per row. If you have more than 4 stats, the rows will contain 3 stats each.
+
+You can override this using the `getColumns()` method. Different screen sizes can also be handled using this method.
+
+```php
+protected function getColumns(): int | array
+{
+    return [
+        'sm' => 1,
+        'md' => 2,
+        'lg' => 2,
+        'xl' => 3,
+        '2xl' => 3,
+    ];
+}
+```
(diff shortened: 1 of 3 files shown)


danharrin:

I am going to handle this myself for v4, but thank you for your contribution 👍

mariomeyer:
@danharrin Thanks for your feedback. I opened the PR directly because I saw there was a previous PR (#13415) about this issue that wasn't approved due to lack of proper implementation. 

Don't see this exactly as a new feature as well, this is a bug. As you can set `getColumns()` on `StatusOverviewWidget` but it doesn't properly handle what is being passed.

 I could have simply handled the cases for more columns on the blade which would fix it, but that was the reason #13415 was rejected. And that's the only reason why I did the `Filament::Grid` implementation.

danharrin:
I consider this a new feature since I don't think we tell users they can set more than 4 columns at the moment, and you're introducing array syntax.

mariomeyer:
Ok. But I specifically made backwards compatibility with int syntax, so it wouldn't break anything for prior implementations.

danharrin:
Yeah that's all good, I'm not worried about BC here. The reason I have closed feature requests is due to management of my own time and prioritisation of the new version over spending time reviewing new features for v3.

If an implementation is perfect or the feature is easy to pass then I sometimes break that rule and merge a PR.

If a feature is relatively unimportant and there are things in the code I will need to address before merging like this, then it's less likely to get my time. And that shouldn't really be a problem for you since it's possible to use a custom view on your widget to achieve this in your app without altering core.

Thanks.

zepfietje:
I think supporting more than 4 columns might be regarded as a bug fix in some sense.
If support for that can be added in a simple PR, I think we'd be happy to review and merge that, though I'll let @danharrin decide on this one.

duykb2015:
@mariomeyer You saved me, thank you!

PR #14675, opened 2024-10-31
Fix table empty view with no records

mishkx:
## Description

Table additional headers was not displayed with no records.

## Visual changes

Before
image

After
image

## Functional changes

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -704,7 +704,7 @@
                         />
                     </x-filament-tables::table>
                 @endif
-            @elseif (($records !== null) && count($records))
+            @else
                 <x-filament-tables::table
                     :reorderable="$isReorderable"
                     :reorder-animation-duration="$getReorderAnimationDuration()"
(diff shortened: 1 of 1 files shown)


danharrin:
Hmm I don't know about this, I'm pretty sure there are reasons why we don't render the table when there are no records. Probably column widths jumping after content is added?

Also, how does this affect the table layout feature, when there is no `<table>` element?

danharrin:
I think I remember why this is: the empty state scrolls when you have enough columns that make the table overflow horizontally?

zepfietje:
> I think I remember why this is: the empty state scrolls when you have enough columns that make the table overflow horizontally?

Yup that's the reason, @danharrin.

I do think the table column headers need to be displayed conditionally in order to undo applied filters or searches.

danharrin:
The applied filters and searches can be undone from the filter indicators I think

zepfietje:
Yeah but that UX is bad when entering a search query in the single field search input.
When it returns no results, the input disappears, so you can just hit backspace and update your search query.

danharrin:
Yeah so this isn't really mergeable in its current state due to the overflow scroll issue...  think we should tackle this in v4 to minimise the potential of breaking changes 👍 Thanks @mishkx

PR #14681, opened 2024-11-01
Scroll on pagination change in list records

mvenghaus:
By default the scroll position is not affected when changing the page. I've added a method to scroll the wire component into the view.


danharrin:
Does this scroll the table all the way to the top? Or just make sure that the pagination UI is still visible

Please can you refactor this to not use `$this->js()`, it's not really a feature I'm really a fan of and we don't use it elsewhere in Filament.

mvenghaus:
> Does this scroll the table all the way to the top? Or just make sure that the pagination UI is still visible
> 
> Please can you refactor this to not use `$this->js()`, it's not really a feature I'm really a fan of and we don't use it elsewhere in Filament.

it scrolls up to the table. i changed it to an event instead of using js

packages/tables/resources/views/index.blade.php
```diff
@@ -121,6 +121,7 @@
     @endif
     ax-load-src="{{ \Filament\Support\Facades\FilamentAsset::getAlpineComponentSrc('table', 'filament/tables') }}"
     x-data="table"
+    x-on:table-pagination-changed.window="if ($event.detail.id === '{{ $this->getId() }}') $el.scrollIntoView();"
```
```suggestion
    x-on:filament-table-pagination-changed.window="if ($event.detail.id === '{{ $this->getId() }}') $el.scrollIntoView()"
```

packages/panels/src/Resources/Pages/ListRecords.php
```diff
@@ -345,4 +345,9 @@ public function getSubNavigation(): array
 
         return [];
     }
+
+    public function updatedPaginators(): void
+    {
+        $this->dispatch('table-pagination-changed', id: $this->getId());
```
danharrin:
```suggestion
        $this->dispatch('filament-table-pagination-changed', id: $this->getId());
```

packages/panels/src/Resources/Pages/ListRecords.php
danharrin, line 1: Can you please move this to the `Filament\Tables\Concerns\CanPaginateRecords` trait, so it applies to all tables (using `InteractsWithTables`) and not just the list page?

danharrin (changes requested):

danharrin:
I just tested this, here is what it did:

https://github.com/user-attachments/assets/e3a0ef7c-e665-4f52-bac0-60af3f91058b

I do not want it to scroll the table up to the top like this, as when you click on a new item I still want to be able to click the next item without scrolling again.

I think the only use case I want it to scroll for is when you are on a page with many records and then you switch to a page with very few records, to scroll the user up to the pagination controls, but not the top.

If you want to attempt that use case, I would accept a new PR, but closing for now as that is not currently the behaviour of this PR and I'm not sure if it's achievable.

PR #14688, opened 2024-11-02
Disallow EXIF data from uploads to be deleted when resizing

boris-glumpler:
## Description

Atm, when you upload an image without transforming it, then all EXIF data is retained. As soon as a transformation happens, tho, the EXIF data is stripped due to the default setting for `imageTransformOutputStripImageHead` being `true`.  This PR just makes it consistent to always retain the EXIF data.

See https://pqina.nl/filepond/docs/api/plugins/image-transform/#properties

## Visual changes

None

## Functional changes


boris-glumpler:
Thanks!

PR #14698, opened 2024-11-04
Two-factor authentication

### LINKED ISSUE #12662: 2FA

danharrin:
Config is defined per-panel, and the implementation is driver based with two default drivers: Google 2FA-compatible apps, and "email code" 2FA


binaryfire:
Hi @danharrin. I just realised that requiring 2FA prior to authentication won't work with several third party auth packages (eg. [laravel-passwordless-login](https://github.com/grosv/laravel-passwordless-login ) and [laravel-magiclink](https://github.com/cesargb/laravel-magiclink)). Magic links need to log the user in first, so 2FA will need to be handled via middleware + a custom code entry page.

I had a quick look at the MFA classes and they seem flexible enough to be used that way, but I just wanted to confirm in case I missed something.

danharrin:
I wouldn't want to change the system to support 2FA after authentication because there is a ton of additional complexity and potential for security issues, so as long as I don't need to do anything that's fine.

Personally if you wanted the experience of these packages, I would suggest you implement a simple custom version that utilises signed URLs to log the user in from their email.

binaryfire:
@danharrin Sorry let me clarify. I'm not suggesting the current implementation should change. From what I could tell, all the logic related to verifying 2FA _before_ authentication is in the `Login` page itself. And the classes in `src/MultiFactorAuthentication` are completely decoupled from that. Just wanted to confirm I've got that right? Because if that's the case, devs can build whatever custom implementation they want using those classes and methods.

Yeah I used to use signed URLs but `laravel-magiclink` has some nice features like [one-time downloads](https://github.com/cesargb/laravel-magiclink?tab=readme-ov-file#download-file-action) and [one-time views](https://github.com/cesargb/laravel-magiclink?tab=readme-ov-file#view-action). Wouldn't be too hard to build I guess but the package works well.

binaryfire:
Regardless of whether you're using signed URLs or a package for magic links, you still need be able to use 2FA. Otherwise you're only relying on one factor (email access). AFAIK the only way to do that is by doing the 2FA verification post-auth.

PR #14727, opened 2024-11-06
feat : ignore cache control

MathieuMenegain:
## Description

Add a parameter to bypass the cache system when running the `widgets`, `pages`, or `resources` methods. This can be used for a dynamic system that wants to keep the static content cached but load independent dynamic modules alongside.

## Visual changes

Before : 
`$panel->resources($resources);`

After : 
`$panel->resources($resources, ignoreCached: true);`

## Functional changes

### DIFF
--- a/packages/panels/src/Panel/Concerns/HasComponents.php
+++ b/packages/panels/src/Panel/Concerns/HasComponents.php
@@ -101,10 +101,11 @@ trait HasComponents
 
     /**
      * @param  array<class-string>  $pages
+     * @param  bool  $ignoreCached
      */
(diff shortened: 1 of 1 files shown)


danharrin:
Hi, thanks for your contribution!

My concern here is that on applications that use Octane, the pages will remain registered between requests even if they do not apply to that request. That is a potential security issue where users could access pages registered for another user.

As such, I am not going to merge this. Generally, I only merge features that work seamlessly across environments, and many people use Octane.

However, you can use a macro on the `Panel` object to add your own method that can register pages dynamically in this way.

PR #14729, opened 2024-11-06
Update 01-getting-started.md

nicoorfi:


danharrin:
This is deliberate, the method only needs to be static for resource classes, other usages of infolists do not use static methods.

PR #14742, opened 2024-11-08
fix(Page): allow the use of MaxWidth enum

saade:
## Description

Currently, both the view and property getter allows the use of the MaxWidth enum, however, the class property only allows string or null.

This PR fixes that.

## Functional changes


wychoong:
breaking change.
submitted in #10755 😆

PR #14744, opened 2024-11-08
fix: global search in SPA mode when switching tenants

### LINKED ISSUE #14196: SPA multi-tenancy global search does not respect tenant switch
### Package

filament/filament

### Package Version

v3.2.110

### Laravel Version

v11.22.0

### Livewire Version

v3.5.6

### PHP Version

PHP 8.3.7

### Problem description

When using global search in SPA mode with multi-tenancy **after** switching to a different tenant, the results are displayed from a previous tenant. I believe that global search component is not reloaded and still uses initial tenant scope.

https://github.com/user-attachments/assets/6ab988f1-c4ca-4467-a313-f97a0b45023b

In this video I am on `Team 1` tenant. I type `post` in global search and observe `This post is on Team 1` record. After switching to tenant `Team 2` I type `post` in global search and still observe `This post is on Team 1` record.

### Expected behavior

I expected search to work the same as it does in non SPA mode. Switching tenant and using global search should correctly scope the query to new tenant.

This is how it works in non SPA mode.

https://github.com/user-attachments/assets/59eb1104-409c-4886-8e2f-6e383b1cd0c4

### Steps to reproduce

1. Clone the repo
2. `composer install`
3. `php artisan migrate --seed`
4. Log in to the app on `/` route with the credentials: email - `test@example.com`, password - `password`
5. Type `post` in the global search bar
6. Observe that `This post is on Team 1` is visible
7. Switch the tenant to `Team 2`
8. Type `post` in the global search bar
9. Observe that `This post is on Team 1` is still visible

https://github.com/Plytas/filament-spa-search

### Relevant log output

_No response_

ralphjsmit:
## Description

This PR fixes the global search issue, where the GlobalSearch is not reset after switching tenants in SPA mode. The reason that it was not taking the new `Filament::getTenant()` is that the component was inside an `x-persist`, so that means that it was not re-mounted and that it would keep the previous URL path also with the previous tenant inside it. As Livewire mocks the original request, this resulted in the previous tenant being used for search.

This solution changes it by adding the current tenant ID to the `x-persist` identifier, so that the topbar end is persisted when navigating within the same tenant, but is refreshed and re-mounted when switching tenants.

Another benefit is that this will re-load the notifications and the user menu, so in case the user avatar or something changed, that would also benefit from the re-load and not getting persisted.

PS: Also, I'm not sure if `x-persist` is officially documented, so perhaps better to switch to `@persist` (though you likely have had your reasons/preference).

Thanks!

## Functional changes

### DIFF
--- a/packages/panels/resources/views/components/topbar/index.blade.php
+++ b/packages/panels/resources/views/components/topbar/index.blade.php
@@ -152,13 +152,13 @@
         @endif
 
         <div
-            x-persist="topbar.end"
+            x-persist="topbar.end{{ filament()->hasTenancy() ? '-' . filament()->getTenant()->getKey() : '' }}"
             class="ms-auto flex items-center gap-x-4"
         >
             {{ \Filament\Support\Facades\FilamentView::renderHook(\Filament\View\PanelsRenderHook::GLOBAL_SEARCH_BEFORE) }}
 
             @if (filament()->isGlobalSearchEnabled())
-                @livewire(Filament\Livewire\GlobalSearch::class, key('global-search' . (filament()->hasTenancy() ? '-' . filament()->getTenant()->getKey() : '')))
+                @livewire(Filament\Livewire\GlobalSearch::class)
             @endif
 
             {{ \Filament\Support\Facades\FilamentView::renderHook(\Filament\View\PanelsRenderHook::GLOBAL_SEARCH_AFTER) }}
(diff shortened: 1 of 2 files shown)


packages/panels/resources/views/components/topbar/index.blade.php
```diff
@@ -152,7 +152,11 @@ class="fi-topbar-close-sidebar-btn lg:hidden"
         @endif
 
         <div
-            x-persist="topbar.end"
+            @if (filament()->hasTenancy())
+                x-persist="topbar.end-{{ filament()->getTenant()->getKey() }}"
```
danharrin:
```suggestion
                x-persist="topbar.end.tenant-{{ filament()->getTenant()?->getKey() }}"
```

PR #14767, opened 2024-11-11
Fixed bug on windows where the system puts \ instead of / while forming the s3 url (Import action).

bmspereira-07:
Fixed bug on windows where the system puts \ instead of / while forming the s3 url.

## Description

When using the import action on windows, it throws an error when trying to get the file from the s3 storage. This happens because windows tries to build the url with \ instead of /. This commit fixes it.

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/CanImportRecords.php
+++ b/packages/actions/src/Concerns/CanImportRecords.php
@@ -398,7 +398,7 @@ trait CanImportRecords
 
             invade($s3Adapter)->client->registerStreamWrapper(); /** @phpstan-ignore-line */
             
-            $fileS3Path = str_replace('\\','/','s3://' . config("filesystems.disks.{$fileDisk}.bucket") . '/' . $filePath);
+            $fileS3Path = (string) str('s3://' . config("filesystems.disks.{$fileDisk}.bucket") . '/' . $filePath)->replace('\\', '/');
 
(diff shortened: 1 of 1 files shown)


packages/actions/src/Concerns/CanImportRecords.php
```diff
@@ -397,7 +397,8 @@ public function getUploadedFileStream(TemporaryUploadedFile $file)
             $s3Adapter = Storage::disk($fileDisk)->getAdapter();
 
             invade($s3Adapter)->client->registerStreamWrapper(); /** @phpstan-ignore-line */
-            $fileS3Path = 's3://' . config("filesystems.disks.{$fileDisk}.bucket") . '/' . $filePath;
+            
+            $fileS3Path = str_replace('\\','/','s3://' . config("filesystems.disks.{$fileDisk}.bucket") . '/' . $filePath);
```
danharrin: I think this is slightly more readable, what do you think?

```suggestion
            $fileS3Path = (string) str('s3://' . config("filesystems.disks.{$fileDisk}.bucket") . '/' . $filePath)->replace('\\', '/');
```

bmspereira-07: Yes sir! I just went with str_replace for compatibility.

PR #14785, opened 2024-11-13
Add missing Indonesian translation

Kepsondiaz:
## Description

In This PR i've added missing translations for Indonesian language

### DIFF
--- a/packages/actions/resources/lang/id/import.php
+++ b/packages/actions/resources/lang/id/import.php
@@ -13,6 +13,9 @@ return [
             'file' => [
                 'label' => 'Berkas',
                 'placeholder' => 'Unggah berkas CSV',
+                'rules' => [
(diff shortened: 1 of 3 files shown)


danharrin:
Can you please confirm where you are sourcing the translations from? Because you have now submitted quite a few PRs for different languages.

I would rather a speaker of these languages wrote the translations, so if these were generated by Google Translate or similar software please let me know.

Kepsondiaz:
For certain languages, I use Google Translate, DeepL, and ChatGPT to ensure that the translation is accurate in both meaning and word choice.

danharrin:
Hi, thank you for your contributions. I have discussed this with the team, and moving forward we would prefer all translations to be written by native speakers to ensure quality. Your PRs will be useful as starting points for those if anyone wants to pick them up. Many thanks.

PR #14815, opened 2024-11-15
Add nestedRecursiveRules to ToggleButtons

tanthammar:
## Description

This PR adds the missing trait to be able to define array rules on `ToggleButtons::make('foo')->multiple()`


danharrin:
Can you please specify which rules would work well on this field? Since these are pre-defined options and not variable like `TagsInput`

tanthammar:
Here is an example:

```php
ToggleButtons::make('booked_seat_id')
      ->multiple()
      ->inline()
      ->grouped()
      ->options($booking->bookedSeats->pluck('bookable.name', 'id'))
      ->nestedRecursiveRules(Rule:in($booking->bookedSeats->pluck('id')->toArray()))
```

PR #14819, opened 2024-11-16
Support globally setting max simple page content width

zepfietje:
## Description

Currently it's not possible to globally set the max width for simple pages like registration and login. This pull request adds support for globally configuring the width just like is already possible for regular pages.

## Functional changes


danharrin (changes requested):
Thoughts on `simpleMaxContentWidth()` instead of `simplePageMaxWidth()`? Existing setting is `maxContentWidth()`

zepfietje:
Hmm, ideally the APIs would be `maxPageContentWidth` and `maxSimplePageContentWidth`, don't you think? Just "simple" without "page" might not be clear enough.

Regarding "max width" vs "max content width", I'm aware of that inconsistency but just stuck to the existing naming. Happy to change anything where possible :)

zepfietje:
@danharrin actually I'm not sure if I like this configuration on the panel provider as global config methods. What do you think about introducing a static property on the page classes, `$defaultMaxContentWidth`?

danharrin:
1) Static properties can only really be set once and not per-request otherwise it breaks Octane
2) Static properties can't differentiate widths based on which panel is used

Personally I usually only use static properties for BC or when something can't be a panel configuration option

Why don't you like the panel config method pattern?

zepfietje:
1. Didn't know about that, good point.
2. Static properties would have to be set in `bootUsing()` of the panel config.

I do like the panel config method pattern in general, but the max page content width APIs feel a bit weird since there's two different types of pages. Maybe it's just a naming issue we can easily resolve though.

danharrin:
Can you rename to `maxSimplePageContentWidth()` so I can merge this please?

zepfietje:
Done! 😊

danharrin (changes requested):
Sorry, could this be `simplePageMaxContentWidth` and `getSimplePageMaxContentWidth` instead please, so then when you search `maxContentWidth` both are returned, if you know what I mean

zepfietje:
Smart, done that!

zepfietje:
Could you merge and tag a new release, @danharrin?
This just missed the latest release and there's a couple people waiting on this change :)

PR #14826, opened 2024-11-18
Add automatic scrolling of wizard steps

### LINKED ISSUE #14665: Wizard header auto horizontal scroll
### Package

filament/filament

### Package Version

v3.2.120

### Laravel Version

v10.48.22

### Livewire Version

v3.4.10

### PHP Version

PHP 8.2.18

### Problem description

I'm not sure if this is a bug or intended behaviour but I was expecting the steps header of the wizard to automatically scroll to the next step. Without this the user may never see the step labels or descriptions. Screenshot:

image

I've checked all existing issues and nobody seems to have brought it up before. There is some talk about it on Discord where I have added a [partial solution](https://discord.com/channels/883083792112300104/1279965710180683821/1301156631555473480).

### Expected behavior

I expected that the wizard header automatically performed a horizontal scroll to step 2, step 3 and so on after each step validates.

I can kind of work around this, with great difficultly, by dispatching the step number through custom events via registerListeners and then listening in alpine with $wire.on.

### Steps to reproduce

Just create a form wizard that has enough steps to overflow the x-axis of the wizard header.

https://github.com/stuartcusackie/filament-wizard-header-scroll-issue

### Relevant log output

_No response_

austincarpenter:
## Description

This PR ensures the wizard scrolls the current step into view when moving between steps.

https://github.com/filamentphp/filament/issues/14665

## Visual changes

https://github.com/user-attachments/assets/9239f0f4-0c9e-4e0c-9b28-9f99118fac0b

## Functional changes

### DIFF
--- a/packages/forms/resources/views/components/wizard.blade.php
+++ b/packages/forms/resources/views/components/wizard.blade.php
@@ -39,9 +39,9 @@
 
         scroll: function () {
             this.$nextTick(() => {
-                this.$root.scrollIntoView({ behavior: 'smooth', block: 'start' })
-
-                this.$refs.header.children[this.getStepIndex(this.step)].scrollIntoView({ behavior: 'smooth', block: 'start' })
+                this.$refs.header.children[
+                    this.getStepIndex(this.step)
+                ].scrollIntoView({ behavior: 'smooth', block: 'start' })
             })
         },
 
--- a/packages/forms/resources/views/components/wizard.blade.php
+++ b/packages/forms/resources/views/components/wizard.blade.php
@@ -41,7 +41,7 @@
             this.$nextTick(() => {
                 this.$root.scrollIntoView({ behavior: 'smooth', block: 'start' })
 
-                this.$refs.headerSteps.children[this.getStepIndex(this.step)].scrollIntoView({ behavior: 'smooth', block: 'start' })
+                this.$refs.header.children[this.getStepIndex(this.step)].scrollIntoView({ behavior: 'smooth', block: 'start' })
             })
         },
 
@@ -137,7 +137,7 @@
             'border-b border-gray-200 dark:border-white/10' => $isContained,
             'rounded-xl bg-white shadow-sm ring-1 ring-gray-950/5 dark:bg-gray-900 dark:ring-white/10' => ! $isContained,
         ])
-        x-ref="headerSteps"
+        x-ref="header"
     >
         @foreach ($getChildComponentContainer()->getComponents() as $step)
             <li
--- a/packages/forms/resources/views/components/wizard.blade.php
+++ b/packages/forms/resources/views/components/wizard.blade.php
@@ -21,7 +21,7 @@
             this.step = this.getSteps()[nextStepIndex]
 
             this.autofocusFields()
-            this.scrollToTop()
+            this.scroll()
         },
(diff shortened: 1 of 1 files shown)


packages/forms/resources/views/components/wizard.blade.php
```diff
@@ -34,13 +34,15 @@
...
 
-        scrollToTop: function () {
-            this.$nextTick(() =>
-                this.$root.scrollIntoView({ behavior: 'smooth', block: 'start' }),
-            )
+        scroll: function () {
+            this.$nextTick(() => {
+                this.$root.scrollIntoView({ behavior: 'smooth', block: 'start' })
+
+                this.$refs.headerSteps.children[this.getStepIndex(this.step)].scrollIntoView({ behavior: 'smooth', block: 'start' })
```
zepfietje: Following the header CSS class, this can probably just be header instead.

```suggestion
                this.$refs.header.children[this.getStepIndex(this.step)].scrollIntoView({ behavior: 'smooth', block: 'start' })
```

zepfietje (changes requested):
I like this UX improvement, @austincarpenter! 😄

packages/forms/resources/views/components/wizard.blade.php
austincarpenter: Too easy, have pushed the change!

zepfietje (approved):
Tested this PR locally and simplified the scroll method since scrolling the step into view no longer requires the wizard itself to be scrolled into view.

Looks good to me! 👌

zepfietje:
Ready for merge, @danharrin :)

ziming:
@austincarpenter 

Is there an easy way to get scrollToTop() back?

My client not happy after a composer update that the behaviour is no longer scrollToTop() as you changed from scrollToTop() to scroll() in this PR

ziming:
Is there an easy way to get scrollToTop() back without locking filament to an older version?

My employer is unhappy after a composer update that the behaviour is no longer scrollToTop() as you changed from scrollToTop() to scroll() in this PR as he felt the UX is now much worse (in his opinion)

For the maintainers, would you all be open to a PR for us to specify what type of scrolling we want? @danharrin

For those of us who want to go back to the previous scroll behaviour

Thank you

danharrin:
I dont know what you mean. As far as I know before, even though the method is `scrollToTop` it just scrolled until the wizard was in view, now it ensures you are actually at the top. Where does your employer want it to scroll to? Because now it **actually** scrolls you to the top.

ziming:
It's after office hours now, let me ask him again next week. He said something along the lines of feeling that the previous scrolling behaviour of the form wizard is much better

danharrin:
Without specifics I'm not sure I understand.

austincarpenter:
`scrollToTop()` to `scroll()` was just a cosmetic name change of the function. 

Before the entire wizard would scroll into view, now just the header scrolls into view when changing steps. However, since the header _is_ at the top of the wizard, if on step 1 for example it was scrolled to the bottom, when going to step 2, I would assume the browser would first need to scroll whatever element it needs to actually see the header element in the viewport, then scroll to the step element inside the header. Perhaps this isn't the case and it's just scrolling the header horizontally outside the viewport.

PR #14827, opened 2024-11-18
Disable extra empty query in DefaultGlobalSearchProvider on page load/navigate  for every resource that can be searched globally

Cybrarist:
## Description

when global search active for resources, an empty search query runs when the page loads, the following query is running for each resource available globally.

`select * from 'models' where ('name' like '%%') limit 50`

as it's not needed until the user searches.

## Visual changes

Before :
Screenshot 2024-11-18 at 11 09 54 AM

After:
Screenshot 2024-11-18 at 11 10 32 AM

## Functional changes

if you don't want to use collection you can use an array instead
```php
$resourceResults = ($query) ? $resource::getGlobalSearchResults($query) : [];

if ($resourceResults && ! $resourceResults->count()) {
    continue;
}
```

### DIFF
--- a/packages/panels/src/GlobalSearch/DefaultGlobalSearchProvider.php
+++ b/packages/panels/src/GlobalSearch/DefaultGlobalSearchProvider.php
@@ -3,6 +3,7 @@
 namespace Filament\GlobalSearch;
 
 use Filament\Facades\Filament;
+use Illuminate\Database\Eloquent\Collection;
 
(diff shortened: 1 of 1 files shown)


danharrin:
Sorry but I am very confused about this bug. I can't reproduce it locally, and the code makes it sound impossible to happen - the `GlobalSearch` component does not call `getResults()` if the `$search` is blank, which then gets passed as `$query`.

Can you please create a reproduction repository and post it as an issue? So I can then look at the issue myself.

PR #14835, opened 2024-11-18
bug fix Infolist columns not respected on lg screen

### LINKED ISSUE #14714: Infolist columns not respected on lg screen
The lg screen breakpoint uses 2 columns even though 3 should be used starting from the sm breakpoint:

```php
$infolist
    ->columns([
        'default' => 1,
        'sm' => 3,
    ])
```

uthadehikaru:
## Description

this pull request is to fix issue Infolist columns not respected on lg screen #14714

## Functional changes

### DIFF
--- a/packages/infolists/src/Concerns/HasColumns.php
+++ b/packages/infolists/src/Concerns/HasColumns.php
@@ -3,6 +3,7 @@
 namespace Filament\Infolists\Concerns;
 
 use Filament\Infolists\ComponentContainer;
+use Illuminate\Support\Facades\Log;
 
 trait HasColumns
 {
@@ -18,7 +19,7 @@ trait HasColumns
     {
         if (! is_array($columns)) {
             $columns = [
-                'lg' => $columns,
+                'sm' => $columns,
             ];
         }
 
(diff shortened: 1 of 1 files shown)


zepfietje:
Thanks for contributing, @uthadehikaru!
This isn't the right fix unfortunately.

Currently when passing an integer for columns, it only applies from screen size lg (which is by design). The issue (as described in the GitHub issue) occurs when passing an array.
Please test your fix using the provided example and let me know if you need any feedback, happy to help :)

uthadehikaru:
> Thanks for contributing, @uthadehikaru!
> This isn't the right fix unfortunately.
> 
> Currently when passing an integer for columns, it only applies from screen size lg (which is by design). The issue (as described in the GitHub issue) occurs when passing an array.
> Please test your fix using the provided example and let me know if you need any feedback, happy to help :)

Hi @zepfietje thx for feedback. I tested using the scenario from the issue. Is there any case that I can do to solve this issue? Thx

zepfietje:
If you could also test passing a single integer to the columns method and ensure that that still only applies from the lg breakpoint instead of sm (that this PR would change), that would help you find the final solution I think.

PR #14853, opened 2024-11-20
feat: allow limiting column map in exports

ralphjsmit:
## Description

This PR adds the option to limit the `ExportAction` to certain columns. Previously, the only choice we had to limit columns was to have a column mapping form where the user can choose columns. However, sometimes we either want the column mapping form not to show all columns, or to not show a column mapping form and directly export a subset of columns.

This adds a `->columns()` method that will allow the user to either:
- Specify which columns get shown in the column mapping form, if `->columnMapping(true)` (default).
- Specify which subset of columns get outputted in the export without showing a form, if `->columnMapping(false)`.

Basically the four choices are now:

```php
// Show column mapping form, include all columns
```

```php
// Show column mapping form, include only specified columns
->columns(['id', 'full_name', 'email'])
```

```php
// Hide column mapping form, export all columns
->columnMapping(false)
```

```php
// Hide column mapping form, export specified columns
->columns(['id', 'full_name', 'email'])
->columnMapping(false)
```

## Functional changes

### DIFF
--- a/packages/actions/docs/07-prebuilt-actions/09-export.md
+++ b/packages/actions/docs/07-prebuilt-actions/09-export.md
@@ -142,6 +142,31 @@ ExportAction::make()
     ->columnMapping(false)
 ```
 
+### Limiting selectable columns
+
+By default, the user will be asked which columns to export. You can limit the list of columns shown for export by using the `columns()` method:
+
+```php
+use App\Filament\Exports\ProductExporter;
+
+ExportAction::make()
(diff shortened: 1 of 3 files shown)


zepfietje:
Been thinking about this before but didn't find the chance to work on it.
I think this is the missing piece for exports! 🚀

danharrin:

I would accept a PR for this feature when v4 is released, if you are still interested. Thank you!

PR #14854, opened 2024-11-20
Improve sync connection support for `ExportAction`

ralphjsmit:
It can be very helpful to run an export on the sync queue if it does not take too long and then immediately display a notification to download it, instead of needing to wait on a job being picked up by a queue and then monitoring the notifications bell icon. In my tests running an export in sync was even quicker than waiting for an (empty) queue to pick it, process all the individual batch parts and then send back to the user. Particularly for less tech-savvy users this is not intuitive.

The ExportAction can already run on the `sync` connection, but this PR improves support for that by removing the first export started notification and changing the export completed notification to a persistent in-page notification.

Scherm­afbeelding 2024-11-20 om 21 11 51

Thanks!

### DIFF
--- a/packages/actions/src/Exports/Jobs/ExportCompletion.php
+++ b/packages/actions/src/Exports/Jobs/ExportCompletion.php
@@ -74,6 +74,14 @@ class ExportCompletion implements ShouldQueue
                     $this->formats,
                 )),
             )
-            ->sendToDatabase($this->export->user, isEventDispatched: true);
+            ->when(
+                $this->connection === 'sync',
+                fn (Notification $notification) => $notification->persistent()
+            )
+            ->when(
+                $this->connection === 'sync',
+                fn (Notification $notification) => $notification->send(),
+                fn (Notification $notification) => $notification->sendToDatabase($this->export->user, isEventDispatched: true),
+            );
     }
 }
--- a/packages/actions/src/Exports/Jobs/ExportCompletion.php
+++ b/packages/actions/src/Exports/Jobs/ExportCompletion.php
@@ -76,11 +76,9 @@ class ExportCompletion implements ShouldQueue
             )
             ->when(
                 $this->connection === 'sync',
-                fn (Notification $notification) => $notification->persistent()
-            )
-            ->when(
-                $this->connection === 'sync',
-                fn (Notification $notification) => $notification->send(),
+                fn (Notification $notification) => $notification
+                    ->persistent()
+                    ->send(),
                 fn (Notification $notification) => $notification->sendToDatabase($this->export->user, isEventDispatched: true),
             );
     }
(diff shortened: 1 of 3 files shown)


packages/actions/src/Exports/Jobs/ExportCompletion.php
```diff
@@ -74,6 +74,14 @@ public function handle(): void
...
-            ->sendToDatabase($this->export->user, isEventDispatched: true);
+            ->when(
+                $this->connection === 'sync',
+                fn (Notification $notification) => $notification->persistent()
+            )
+            ->when(
+                $this->connection === 'sync',
+                fn (Notification $notification) => $notification->send(),
+                fn (Notification $notification) => $notification->sendToDatabase($this->export->user, isEventDispatched: true),
+            );
```
zepfietje: Maybe merge these?

```suggestion
            ->when(
                $this->connection === 'sync',
                fn (Notification $notification) => $notification
                    ->persistent()
                    ->send(),
                fn (Notification $notification) => $notification->sendToDatabase($this->export->user, isEventDispatched: true),
            );
```

zepfietje (changes requested):
Nice one, @ralphjsmit!

I agree there's a strong use case for this (also from personal experience). Thanks for improving the UX around those exports. 😊

packages/actions/src/Exports/Jobs/ExportCompletion.php
ralphjsmit: Thanks, for sure committed! Reason I kept it separate was that the other identifiers above also were in it's own `->when()`'s each, but this is also perfect.

ralphjsmit:
Thanks! Yeah I also like the exports feature, so glad to add some improvements based on real-life cases.

The biggest thing yet I would really wish for on exports is the ability to format (Excel) cells to their native type (like integer, money), text formatting and set column widths. But that unfortunately seems very difficult/unsupported on the OpenSpout package... 

Update: it seems to be possible to style on row-level already, but not yet on the individual/cell column level. Though I dived a bit more and OpenSpout does seem to support passing individual Style objects to individual Cells, so perhaps if we can re-work our logic a little that _could_ work out... 🤔

Update: leaving this here for later reference: https://github.com/openspout/openspout/blob/4.x/src/Writer/XLSX/Manager/Style/StyleRegistry.php

packages/actions/src/Exports/Jobs/ExportCompletion.php
zepfietje: Yeah no big deal, just like grouping the conditions a little better.

zepfietje:
Interesting! Keep us posted when you've experimented with that. 😄

zepfietje (approved):

Issue #10816, opened 2024-01-15
When there is a SQL error in the importer, it causes multiple retries.


bzy107:
### Package

filament/filament

### Package Version

v3.1.47

### Laravel Version

v10.40.0

### Livewire Version

v3.3.5

### PHP Version

8.2.11

### Problem description

This issue is related to #10814 and occurred when using PostgreSQL. Additionally, the queue is running in the database.

```
QUEUE_CONNECTION=database
```

When a job fails due to a SQL error during CSV import, the `getJobRetryUntil()` function retries the job for the specified time. However, it currently also retries the error for the specified number of times. Therefore, I want to modify it to only retry once, but currently, it is only possible to specify a time.

As a solution, I would like to make `getJobRetryUntil()` nullable so that the job is only executed once.

This image is from a case where getJobRetryUntil was set to 3 seconds.
```php
    public function getJobRetryUntil(): CarbonInterface
    {
        return now()->addSeconds(3);
    }
```
![スクリーンショット 2024-01-16 2 19 41](https://github.com/filamentphp/filament/assets/28666304/bddf3a37-ee46-4284-a9fc-1d3f4b3ef8d3)

**I will create a pull request for this issue later.**

### Expected behavior

If getJobRetryUntil() is set to null, it will be executed only once.
```php
    public function getJobRetryUntil(): ?CarbonInterface
    {
        return null;
    }
```
![スクリーンショット 2024-01-16 2 25 27](https://github.com/filamentphp/filament/assets/28666304/f11acf4d-39ac-4a6d-8426-25ddf17b3491)

### Steps to reproduce

Please refer to #10814 for more information.

### Reproduction repository

https://github.com/bzy107/filament-select-bug

danharrin:
It appears that this is a feature request rather than a bug report, so closing

Issue #14823, opened 2024-11-17
Import file with long name (or arabic name)


Saifallak:
### Package

filament/filament

### Package Version

v3.2.124

### Laravel Version

v11.x

### Livewire Version

v3.x

### PHP Version

8.3.x

### Problem description

getting an error when importing large file names or Arabic file name ex:

```
SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'file_path' at row 1 (Connection: mysql, SQL: insert into `imports` (`user_id`, `file_name`, `file_path`, `importer`, `total_rows`, `updated_at`, `created_at`) values (209, عملاء شهر يناير 2024 البدر للمبيعات اون لاين واوف لاين .csv, /code/storage/app/livewire-tmp/97dKPqjXMEhVTL3x1swh3I6hfqyYKK-meta2LnZhdmE2KfYoSDYtNmH2LEg2YrZhtin2YrYsSAyMDI0INin2YTYqNiv2LEg2YTZhNmF2KjZiti52KfYqiDYp9mI2YYg2YTYp9mK2YYg2YjYp9mI2YEg2YTYp9mK2YYgLmNzdg==-.csv, App\Filament\Imports\ContactImporter, 53, 2024-11-17 08:09:15, 2024-11-17 08:09:15))
```

### Expected behavior

imported succssfully. 

### Steps to reproduce

just import file named ` عملاء شهر يناير 2024 البدر للمبيعات اون لاين واوف لاين .csv`

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/AQuadic/NO-REPO-NEEDED

### Relevant log output

```shell
Sorry for not including repo, but it can be reproduced on any repo and on demo. just file name like I said.
```

btxtiger:
`1406 Data too long for column 'file_path'`

Looks like your `file_path` column is too small, probably `varchar(255)`. You can try it with `text` type

jonaaix:
@Saifallak 
1406 Data too long for column `'file_path'`

Looks like your `file_path` column is too small, probably `varchar(255)`. You can try it with `text` type

Saifallak:
> @Saifallak 1406 Data too long for column `'file_path'`
> 
> Looks like your `file_path` column is too small, probably `varchar(255)`. You can try it with `text` type

Yes you're right, but this is the migration created by filament, that's why i opemed issue to be changed in the library itself

[...]

PR #14856, opened 2024-11-20
Fix: Rely on Laravel's default job retry settings for Imports/Exports

ahinkle:
## Description
This PR addresses an issue with the default retry behavior for failed imports, which can lead to unintended costs and resource usage. Specifically, the default behavior allows a bad import to continuously retry for 24 hours, potentially resulting in increased expenses due to services like AWS SQS and exception logging.

While the 24-hour retry window was designed to account for temporary issues like a database outage, it is more common for failures to result from code bugs or invalid data. Reducing the retry duration prevents these issues from escalating into excessive costs while still allowing jobs to retry a reasonable number of times.

We retain the existing method for those who prefer it while updating the default to align with your application (or Laravel's) defaults.

Fixes  https://github.com/filamentphp/filament/discussions/13113 , https://github.com/filamentphp/filament/issues/10816

## Functional changes

### DIFF
--- a/packages/actions/docs/07-prebuilt-actions/08-import.md
+++ b/packages/actions/docs/07-prebuilt-actions/08-import.md
@@ -681,7 +681,7 @@ If you'd like to customize the middleware that is applied to jobs of a certain i
 
 ### Customizing the import job retries
 
-By default, the import system will retry a job for 24 hours. This is to allow for temporary issues, such as the database being unavailable, to be resolved. That functionality is defined in the `getJobRetryUntil()` method on the importer class:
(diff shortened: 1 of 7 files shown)


Saifallak:
I faced this issue, due code bug (#14823), and it was re trying for a day and costed high usage in server

danharrin:
Hi, thanks for the PR. At the moment, this isn't a change we can make as it would be breaking. But you can implement this by overriding the methods in your own app. In v4 I have changed the retry strategy for imports/exports already so hopefully that will be better.

PR #14865, opened 2024-11-22
Add support for dashboard tabs

ralphjsmit:
This PR adds support for adding tabs on your dashboard. In case you have many widgets, it makes sense to split them by separate tabs. The implementation is based on the Resource tabs implementation, with the only difference that the Dashboard tabs are implemented on the Alpine-side instead of Livewire due to performance (if Livewire is still (lazy) loading widget on a tab, then switching tabs would be unresponsive until all (slow) widgets are completely loaded).

Users can mark their widgets as belonging to a tab using a `protected static ?string $tab = 'Label';` property. This is the same principle as eg the `$navigationGroup` property, that also matches navigation group by it's label.

https://github.com/user-attachments/assets/727b5260-7859-4d00-866b-d6b6e5a4d3c0

Thanks!

### DIFF
--- a/packages/panels/docs/05-dashboard.md
+++ b/packages/panels/docs/05-dashboard.md
@@ -215,6 +215,32 @@ class Dashboard extends BaseDashboard
 
 Handling data from the filter action is the same as handling data from the filters header form, except that the data is validated before being passed to the widget. The `InteractsWithPageFilters` trait still applies.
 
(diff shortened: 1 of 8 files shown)


danharrin:

In v4, I think we should allow this to be achieved by overriding the "schema" of the page instead. You can then render widgets using inside a `Tabs` component with any combination.

ralphjsmit:
Ok, thanks, that about the "schema" of the page sounds great!

PR #14898, opened 2024-11-27
Make getLocaleLabelUsing usable in service provider

johanmolen:
## Description

Right now we can not do the following within the panel service providers:

```php
    {
        return $panel
            ->plugins([
                SpatieLaravelTranslatablePlugin::make()
                    ->defaultLocales(
                        array_map(
                            static fn ($locale) => type($locale)->asString(),
                            config()->array('app.locales')
                        )
                    )
                    ->getLocaleLabelUsing(fn (string $locale, ?string $displayLocale = null): ?string => $locale),
            ])
    }
```

Because `getLocaleLabelUsing` is not returning the plugin itself, which is expected by the panel `plugins()` method.
This PR fixes that.


danharrin:
Thanks

PR #14962, opened 2024-12-02
Added sample generated image

ModestasV:
## Description

With this PR, I am checking for the need to improve documentation. In this specific example, there will be an image added to:

Panels -> Navigation -> Customizing the user menu

This example adds one image but should become a starting point for generating more images. In other words - I would love some feedback if we figured out the workflow correctly and if we should continue adding them :)

ps. Inspired by https://x.com/PovilasKorop/status/1862104660154241338 - to bring more visuals to documentation.

## Visual changes

No changes to the Filament Application.

Change only affects documentation (I did not find a way to build it locally to add screenshots, sorry!)

## Functional changes


docs-assets/screenshots/schema.js
```diff
@@ -2954,4 +2954,18 @@ export default {
...
+    'panels/navigation/customizing-user-menu': {
+        url: 'top-bar',
+        selector: 'body > div.fi-layout.flex.min-h-screen.w-full.flex-row-reverse.overflow-x-clip > div.fi-main-ctn.w-screen.flex-1.flex-col.opacity-0 > div > nav > div > div > div.fi-dropdown-panel.absolute.z-10.w-screen.divide-y.divide-gray-100.rounded-lg.bg-white.shadow-lg.ring-1.ring-gray-950\\/5.transition.dark\\:divide-white\\/5.dark\\:bg-gray-900.dark\\:ring-white\\/10.\\!max-w-\\[14rem\\]',
+        viewport: {
+            width: 1080,
+            height: 640,
+            deviceScaleFactor: 3,
+        },
+        before: async (page) => {
+            await page.click('body > div.fi-layout.flex.min-h-screen.w-full.flex-row-reverse.overflow-x-clip > div.fi-main-ctn.w-screen.flex-1.flex-col.opacity-0 > div > nav > div > div > div.fi-dropdown-trigger.flex.cursor-pointer')
```
danharrin: I really REALLY appreciate the work you put into this, adding screenshots to the docs is a goal we have and it's great to see people investigate the screenshot system we have in order to contribute, as it scares a lot of people off

My only feedback here is to suggest that we add CSS classes to Filament so these selectors can be simplified, I don't want them to break if we change Tailwind classes etc.

danharrin (changes requested):

docs-assets/screenshots/schema.js
ModestasV: Not going to lie - it was a bit scary to use it! But now, it is pretty obvious.

My question unrelated to this - is there a way to generate a preview with documentation? Would love to be sure it was actually added :)

As for a custom class - will see what I can do and update it accordingly, thanks for feedback!

danharrin: Once the PR is merged, the image will be hosted on GitHub & then you can run `npm run generate-docs` from the `filamentphp.com` repo to check it out, but I will do that for you when I deploy next

ModestasV: I changed the selector path to simplify it (avoiding copying from the browser).

I think this is a reasonable approach since just a few components are visible anyway.

danharrin (changes requested):
Can you try and work out a way to add some space around the outside of the screenshots please? Essentially zoom out so you can see some of the elements around the outside of the dropdown, so it looks more natural. Have a look at the other screenshots for inspiration maybe

ModestasV:
Will see what's possible :)

danharrin (approved):
This looks great, thanks!

docs-assets/screenshots/schema.js
```diff
@@ -2954,4 +2954,21 @@ export default {
             deviceScaleFactor: 3,
         },
     },
+    'panels/navigation/customizing-user-menu': {
+        url: 'top-bar',
```
danharrin:
```suggestion
        url: 'topbar',
```

docs-assets/app/routes/web.php
```diff
@@ -29,3 +30,4 @@
 Route::get('/infolists/layout', InfolistsLayoutDemo::class);
 Route::get('/notifications', NotificationsDemo::class);
 Route::get('/tables', TablesDemo::class);
+Route::get('/top-bar', Topbar::class);
```
danharrin:
```suggestion
Route::get('/topbar', Topbar::class);
```

docs-assets/app/app/Livewire/Topbar.php
```diff
@@ -0,0 +1,23 @@
...
+    protected static string $view = 'livewire.topbar';
+
+    public function mount()
+    {
+        filament()
+            ->getCurrentPanel()
+            ->userMenuItems([
+                MenuItem::make()
+                    ->label('Settings')
+                    ->url(fn(): string => '#topbar-link-example')
```
danharrin:
```suggestion
                    ->url(fn (): string => '#')
```

danharrin (changes requested):

danharrin:
Thanks!

ModestasV:
Oh, you accepted this 😅 Will be adding more images (practically all of them to the Navigation section) with next PR then :)

danharrin:
Thanks! I have deployed the first screenshot to the website

ModestasV:
Hmm, it doesn't seem to have loaded... Did I make a mistake anywhere? Or is there some delay?

danharrin:
For some reason Forge didn't pick up the deployment, but I've done it again

PR #14970, opened 2024-12-02
fix: Disabled button tooltip

### LINKED ISSUE #14842: Show tooltip for disabled actions
Opening this issue to track since the PR has been closed by the author: https://github.com/filamentphp/filament/pull/13671.

danharrin:
What do you think about this solution @zepfietje? If we remove the `pointer-events-none` when there is a tooltip and remove the href/click handlers instead, we can still listen for hover events.

If you approve, I can apply it to the link/dropdown item/icon button and others.


danharrin:
Closes #14842

zepfietje:
How exactly does this work?
Wasn't the issue that disabled buttons don't fire hover events?

danharrin:
Yes, so now the buttons are not disabled and they still emit pointer events.

PR #14975, opened 2024-12-02
Feature: Uncontained Fieldset

### LINKED ISSUE #11355: Uncontained fieldset
- [ ] Remove border

thethunderturner:
## Description

Closes #11355. This has to be discussed with @zepfietje before merging

## Visual changes

image
image

## Functional changes


danharrin:
Please check #14968 for how I implemented it in the Section, with the `-not-` class instead, and using the `CanBeContained` trait.

thethunderturner:
> Please check #14968 for how I implemented it in the Section, with the `-not-` class instead, and using the `CanBeContained` trait.

Will do. Thanks

danharrin:
The reason I want the `-not` class is because my philosophy is that an element should have a minimal number of CSS classes to achieve the "default" look, so if someone uses the raw classes instead of the Blade component they can achieve it easier.

packages/support/resources/css/components/fieldset.css
danharrin, line 1: I don't think this is right, the code needs to apply if there is `:not()` this class present. Can you make sure you test it locally please?

danharrin (changes requested):
Please add more space between the legend and the content, I think it needs to at least look the same as `gap-6`

But I would like to hear what Zep's idea was for this, as I don't think I'd ever use an uncontained fieldset

packages/support/resources/css/components/fieldset.css
thethunderturner: Yeah you're right. Im still working on it though :sweat_smile:

thethunderturner:
> Please add more space between the legend and the content, I think it needs to at least look the same as `gap-6`
> 
> But I would like to hear what Zep's idea was for this, as I don't think I'd ever use an uncontained fieldset

Unfortunately `gap` wont work because we're using the `display: block` property which prevents gap from having an effect. Instead I used `padding-top` which in this case I believe has the same result.

thethunderturner:
Hey @zepfietje, is there any update on this?

zepfietje:
@thethunderturner could you share screenshots of the current state of this PR?

thethunderturner:
> @thethunderturner could you share screenshots of the current state of this PR?

Will do!

danharrin (approved):

PR #14993, opened 2024-12-04
fix: Modal heading getting too close to close button

### LINKED ISSUE #14938: Modal heading getting too close to close button
Possible to add right margin just to top line?

Screenshot 2024-11-30 at 09 48 51

RoyHridoy:
## Description

Fixed: #14938 
This PR fixes the issue of the modal heading being too close to the close button.

### Before
Screenshot 2024-12-04 125107

### After
image

## Functional changes


danharrin (changes requested):
Please use `me-5` instead of `mr-5` to support RTL layouts.

Please only add this class when the close button is enabled. It is probably easier if you pass this class through from the `modal.index` view since you can access the close button setting there already.

RoyHridoy:
The task is complete. I've updated the layout to use `me-5` instead of `mr-5` for RTL support and ensured the class is only added when the close button is enabled. The logic is passed through from the `modal.index` view, leveraging the close button setting already available there.

Let me know if there's anything else you'd like adjusted!

danharrin:
Thanks

PR #15004, opened 2024-12-05
Select table records by pivot if needed

johanmolen:
## Description

If you define `allowDuplicates` on a BelongsToMany relation manager (the pivot table contains the `id`), and add a bulk action like `DetachBulkAction`. 

The action runs `getSelectedTableRecords` with `$shouldFetchSelectedRecords` set to `true`. 
In that case it always runs the wrong query, because it should use the pivot query.

This pull request should fix that, but im not sure if it could break other things.

I opted for a variable `$usePivotQuery` to make the `if` statement more readable, but I can also inline it if you want.

### DIFF
--- a/packages/tables/src/Concerns/HasBulkActions.php
+++ b/packages/tables/src/Concerns/HasBulkActions.php
@@ -349,10 +349,9 @@ trait HasBulkActions
 
         $table = $this->getTable();
 
-        if (
-            $shouldFetchSelectedRecords ||
-            (! ($table->getRelationship() instanceof BelongsToMany && $table->allowsDuplicates()))
-        ) {
+        $usePivotQuery = $table->getRelationship() instanceof BelongsToMany && $table->allowsDuplicates();
+
+        if (! $usePivotQuery) {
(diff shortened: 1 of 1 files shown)


danharrin:
Does

```
if (
    (! $shouldFetchSelectedRecords) ||
    (! ($table->getRelationship() instanceof BelongsToMany && $table->allowsDuplicates()))
) {
```

work?

Because otherwise I think this breaks using a `BelongsToMany` with `fetchSelectedRecords(false)`

johanmolen:
@danharrin should `fetchSelectedRecords(false)` touch the pivot query part? 

Right now the code only executes the pivot query if both the relationship is `BelongsToMany` and `allowsDuplicates` is `true`.

If `allowsDuplicates` is false but the relationship is `BelongsToMany` it will still enter the `if` statement right?

danharrin:
It should make it so the `$records` is just an array of IDs instead of an array of models

johanmolen:
Ah I see! Let me try.

johanmolen:
@danharrin works indeed, changed the PR.

danharrin:
Thanks

PR #15016, opened 2024-12-06
Fix style table record actions on Grid layout

A909M:
## Description

This  PR fixes an issue when arranging records into a grid. A problem occurs where "Record Actions" do not wrap automatically if there are too many actions because `wrap='-sm'` on `filament-tables::actions` enable  `sm:flex-nowrap` . This causes the actions to overflow the container (div) and results in a broken layout, especially on larger screens.

## Visual changes

_**before**_

_**After**_

after

## Functional changes


danharrin:
This was a deliberate decision within our design to not allow wrapping unless you are on mobile, to ensure that cards do not grow. I suggest that in this case, you should use an action group to ensure that there are not actions that are overflowing.

Issue #14713, opened 2024-11-05
Sub-navigation hooks wrapper breaks mobile UI


zepfietje:
https://github.com/filamentphp/filament/pull/14566#discussion_r1829091512

PR #15030, opened 2024-12-09
Fix padded appearance near sub-navigation select element

Persaeus:
## Description

With 1e65d98, on mobile the sub-navigation select element appears to have padding above and below when the render hooks for `PAGE_SUB_NAVIGATION_SELECT` are empty.

This is because the wrappers for the hooks, even when empty, are affected by the flexbox gap.

This PR adds `display: contents` to each to undo this effect.

## Visual changes

| Before | After |
|:------:|:-----:|
| before | after |

## Functional changes

### DIFF
--- a/packages/panels/resources/views/components/page/index.blade.php
+++ b/packages/panels/resources/views/components/page/index.blade.php
@@ -64,7 +64,7 @@
             ])
         >
             @if ($subNavigation)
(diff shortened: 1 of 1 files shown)


danharrin:
Very nice fix, thanks

danharrin:
Fixes #14713

PR #15039, opened 2024-12-09
add a new helper method for actions for add body text

sohailahmad07:
## Description
Hello,

I’ve added two new helper methods, successNotificationBody and failureNotificationBody, to simplify the process of adding body text to notifications. These methods make it easier to define a clean and concise body for notifications when needed.

Both methods are completely optional—if no body text is provided, the body will default to null and will not be displayed. This ensures flexibility while keeping the notification system lightweight and user-friendly.
## Functional changes

- [ * ] Code style has been fixed by running the `composer cs` command.
- [ * ] Changes have been tested to not break existing functionality.
- [ * ] Documentation is up-to-date.

### DIFF
--- a/packages/actions/docs/07-prebuilt-actions/01-create.md
+++ b/packages/actions/docs/07-prebuilt-actions/01-create.md
@@ -101,6 +101,13 @@ CreateAction::make()
     ->successNotificationTitle('User registered')
 ```
 
+To customize the body of this notification, use the `successNotificationBody()` method:
+
+```php
+CreateAction::make()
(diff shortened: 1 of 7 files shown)


danharrin:
Hi, please check out our [contributing guide](https://filamentphp.com/docs/3.x/support/contributing#development-of-new-features) which talks about proposing the feature before submitting the PR, to ensure you aren't spending time building something that we won't merge.

In this case, the feature is the same as this:

```php
->successNotification(fn (Notification $notification) => $notification->body('...'))
```

And this method supports passing any Notification method so it is more flexible.

PR #15041, opened 2024-12-09
Fix `Route [login] not defined` error

ralphjsmit:
This PR fixes the common `Route [login] not defined` error. I've had this issue popping up occasionally for years. I dug into it this time and the reason for the error comes from when Laravel does not recognize the remember token in the `AuthenticateSession` middleware, in turn logs out and then throws an `AuthenticationException`. This `AuthenticationException` has no redirect url, so the Exception handler defaults to `route('login')`. 

This PR fixes the issue by introducing an extended `AuthenticateSession` middleware that contains a Filament-specific `redirectTo()` method, in the same way as the `Authenticate` middleware. We could recommend people to replace this middleware and/or make it part of the automated Rector V4 upgrade script.

### DIFF
--- a/packages/panels/src/Http/Middleware/AuthenticateSession.php
+++ b/packages/panels/src/Http/Middleware/AuthenticateSession.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace Filament\Http\Middleware;
+
+use Filament\Facades\Filament;
+use Illuminate\Session\Middleware\AuthenticateSession as Middleware;
+
+class AuthenticateSession extends Middleware
+{
+    protected function redirectTo($request): ?string
+    {
+        return Filament::getLoginUrl();
+    }
+}
--- a/docs-assets/app/app/Http/Kernel.php
+++ b/docs-assets/app/app/Http/Kernel.php
@@ -10,6 +10,7 @@ use App\Http\Middleware\TrimStrings;
 use App\Http\Middleware\TrustProxies;
 use App\Http\Middleware\ValidateSignature;
 use App\Http\Middleware\VerifyCsrfToken;
+use Filament\Http\Middleware\AuthenticateSession;
 use Illuminate\Auth\Middleware\AuthenticateWithBasicAuth;
 use Illuminate\Auth\Middleware\Authorize;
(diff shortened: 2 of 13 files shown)


packages/panels/src/Http/Middleware/AuthenticateSession.php
```diff
@@ -0,0 +1,14 @@
...
+namespace Filament\Http\Middleware;
+
+use Filament\Facades\Filament;
+use Illuminate\Session\Middleware\AuthenticateSession as Middleware;
+
+class AuthenticateSession extends Middleware
+{
+    protected function redirectTo($request): ?string
+    {
+        return Filament::getLoginUrl();
```
danharrin, line 12: What about this?

```suggestion
        return parent::redirectTo() ?? Filament::getLoginUrl();
```

Also have you checked if the correct panel is always used to generate URLs and not just the default panel? I can't remember if the middleware to identify the panel runs before this or not.

danharrin (changes requested):

packages/panels/src/Http/Middleware/AuthenticateSession.php
ralphjsmit: The reason I did it like this is just to follow how `Authenticate::class` middleware `redirectTo()` method does it. 

I don't think we should call the `parent::redirectTo()`, because if we would do that, then it will give priority to the `Illuminate\Session\Middleware\AuthenticateSession::redirectUsing(...)` method. It is not unthinkable that people have this callback configured for their entire (client-facing) application without thinking about Filament and handling the specific login URL for Filament. So then people trying to access Filament could suddenly get redirected to a whole other login page based on the `redirectUsing()` callback. So I would say as this is a Filament middleware then you don't need the `redirectUsing()` functionality, just configure it via a panel if you want.

The panel is identified before this middleware, because the Panel `getMiddleware()` method *first* returns the `panel:{id}` middleware alias and then all the other middleware provided in `middleware([...])`. I also tested this myself on a non-default panel. So it should be good.

danharrin:
Thanks

PR #15046, opened 2024-12-10
Fix TagsInput closing filters when removing tag from list

TheSkybreaker:
## Description

Hello,

When using `TagsInput` to make a custom table filter, clicking the "X" to remove a tag causes the filters to close. 
If the user needs to remove more than one tag they have to open the filters again and remove the tags one by one (or reset the filters and insert again what to keep)

Stopping the event propagation on click prevents this side effect

## Visual changes

None

## Functional changes


danharrin:
Thanks

PR #15051, opened 2024-12-10
Split logout item from user menu

dododedodonl:
## Description

Split logout item into its own component in oder to easier override and customize its behavior by overriding a more specific view.

Usecase: need to remove certain items from local storage when a user logs out

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/panels/resources/views/components/logout-item.blade.php
+++ b/packages/panels/resources/views/components/logout-item.blade.php
@@ -1,0 +1,13 @@
+@props([
+    'logoutItem' => null,
+])
+
+<x-filament::dropdown.list.item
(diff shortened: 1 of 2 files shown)


danharrin:
Hey, I would rather implement something like `extraAttributes()` on the item, instead of encouraging people to override the views.

I would accept a PR to do that in v4, thank you.

dododedodonl:
> Hey, I would rather implement something like `extraAttributes()` on the item, instead of encouraging people to override the views.
> 
> I would accept a PR to do that in v4, thank you.

Will menu item / logout button be a full actions in v4?

danharrin:
Actually yes, so `extraAttributes()` should already be there

PR #15062, opened 2024-12-11
fix: Remove x-ignore around nested Alpine components

danharrin:
The Async Alpine (v1) version was locked a while ago due to issues released in a minor version. We cannot receive updates to this package.

2 weeks ago we locked Livewire in [v3.2.125](https://github.com/filamentphp/filament/releases/tag/v3.2.125) as it caused issues when using tables.

A [discussion](https://github.com/filamentphp/filament/discussions/14901#discussioncomment-11504282) revealed the reason the tables broke was because [Alpine.js started not initialising components within `x-ignore`d elements](https://github.com/alpinejs/alpine/pull/4428).

The `x-ignore` needs to be present when using Async Alpine v1, and our tables use Async Alpine and contain nested Alpine.js components.

Since it does not look like this change is going to be reverted in Alpine.js, and it is not great having the Livewire version locked in Filament as it prevents users from being able to use new Livewire features elsewhere in their apps, I am removing Async Alpine from the tables so I can remove the `x-ignore` and unlock Livewire.


bilogic:
can we unlock livewire already? Happy to file a PR

danharrin:
I do not think the change has been reverted, so no. This PR was reverted as more issues were discovered.

bbredewold:
Is there an issue / discussion were we can be kept informed about this? Really waiting to unlock Livewire again. Can we help?

danharrin:
I am not expecting Alpine.js to revert the change, so I currently think Livewire is going to be locked like this for the remainder of v3. It is unlocked in v4 as we were able to update a separate dependency which fixes the effect.

bilogic:
when will v4 become official? I can't find any info EOL dates for v3

danharrin:
No Filament versions have EOL dates at the moment, we would still release security fixes for v1 or v2

danharrin:
Please see #15559

PR #15064, opened 2024-12-11
DatePicker is set to readOnly(), but it still can clicked and open calendar popup

untillnesss:
## Description
I'm using DatePicker with readOnly() and native(false) options, but I can still do the Click action and it displays the calendar popup.

## Visual changes
No visual changes.

## Functional changes
Just add IF if it isReadOnly.


danharrin:
Thanks

Issue #14080, opened 2024-08-28
When the tab title is in Chinese, the Form Tab is not separated correctly.


cityisempty:
### Package

filament/filament

### Package Version

V3.2.0

### Laravel Version

v11

### PHP Version

PHP 8.3

### Problem description

When I have two tabs in the form, if both tabs have Chinese titles, then their contents are merged into one, and both tabs are in a selected state.
<img width="897" alt="截屏2024-08-28 22 12 25" src="https://github.com/user-attachments/assets/ce389fd8-41e6-4885-8efd-8c571583ffd0">
<img width="577" alt="截屏2024-08-28 22 12 58" src="https://github.com/user-attachments/assets/6e7fd665-7e34-4f3e-8808-0e98368d561c">

### Expected behavior

When using tabs, Chinese characters should be handled correctly.

### Steps to reproduce

``` 
Tabs\Tab::make(__('基本信息'))->schema([
TextInput::make('title')
                            ->label(__('Post Title'))
                            ->required()
                            ->maxLength(255)
                            ->live(onBlur: true)]),
Tabs\Tab::make(__('(E)扩展信息'))->schema([
                        Placeholder::make(__('扩展信息')),
    
                        Hidden::make('user_id')
                            ->default(auth()->user()->id)
                            ->required(),
    
                        Hidden::make('post_type')
                            ->default('post')
                            ->required()])
```  

When I use the above code, the error occurs. If I change it to the following code, the result is correct.

```
Tabs\Tab::make(__('Title '))->schema([
TextInput::make('title')
                            ->label(__('Post Title'))
                            ->required()
                            ->maxLength(255)
                            ->live(onBlur: true)]),
Tabs\Tab::make(__('Other Info '))->schema([
                        Placeholder::make(__('扩展信息')),
    
                        Hidden::make('user_id')
                            ->default(auth()->user()->id)
                            ->required(),
    
                        Hidden::make('post_type')
                            ->default('post')
                            ->required()])
```

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/cityisempty/filament-issue

zxc88645:
The issue comes from
[forms/src/Components/Tabs/Tab.php](https://github.com/filamentphp/filament/blob/v3.2.110/packages/forms/src/Components/Tabs/Tab.php)
[infolists/src/Components/Tabs/Tab.php](https://github.com/filamentphp/filament/blob/v3.2.110/packages/infolists/src/Components/Tabs/Tab.php)

[...]

PR #15065, opened 2024-12-11
fix: handle tag switching for non-ASCII characters

zxc88645:
## Description

Fixed an issue where tabs or steps could not be switched when non-ASCII characters (e.g., Chinese) were used as their titles.

#14080 

## Visual changes

none

## Functional changes


danharrin:
Thanks for the fast response!

PR #15078, opened 2024-12-13
Refactor: To allow passing in a callback function to `assertTableActionDataSet` and `assertTableBulkActionDataSet`

chester-sykes:
## Description

This PR refactors the `assertTableActionDataSet` and `assertTableBulkActionDataSet` so that you can pass in a callback to get access to the data that is currently set in the table action, like you can with the `assertFormSet`.
 
I think this feature will be useful for testing the `FileUpload` as currently I don't know how you would test that a image is set with `assertTableActonDataSet` because it uses UUIDs so you can't assert that the right file has been set.

Also when I ran `composer cs`, it removed this `use Filament\Support\Assets\AlpineComponent;` from `packages/tables/src/TablesServiceProvider.php`, I checked and I couldn't see it being used anywhere. I commited it even though its not related to my PR, not sure if I should have. If not then please let me know and I'll revert.

## Visual changes

No visual changes.

## Functional changes

### DIFF
--- a/packages/tables/.stubs.php
+++ b/packages/tables/.stubs.php
@@ -13,7 +13,7 @@ namespace Livewire\Features\SupportTesting {
 
         public function setTableActionData(array $data): static {}
 
-        public function assertTableActionDataSet(array $data): static {}
+        public function assertTableActionDataSet(array | Closure $state): static {}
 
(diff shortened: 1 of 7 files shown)


danharrin:
Thanks

PR #15080, opened 2024-12-13
Add class to target logout action

dododedodonl:
## Description

A way to be able to target the logout button/form without having to override the user menu view (instead of #15051).

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/panels/resources/views/components/user-menu.blade.php
+++ b/packages/panels/resources/views/components/user-menu.blade.php
@@ -90,6 +90,7 @@
             :icon="$logoutItem?->getIcon() ?? \Filament\Support\Facades\FilamentIcon::resolve('panels::user-menu.logout-button') ?? 'heroicon-m-arrow-left-on-rectangle'"
             method="post"
             tag="form"
+            class="fi-user-menu-logout-form"
         >
             {{ $logoutItem?->getLabel() ?? __('filament-panels::layout.actions.logout.label') }}
         </x-filament::dropdown.list.item>


danharrin:
Since this element isn't going to exist in v4 in the same way (after moving to actions), I am hesitant to introduce it as a hook class, since it will then create a breaking change when upgrading. Sorry.

dododedodonl:
> Since this element isn't going to exist in v4 in the same way (after moving to actions), I am hesitant to introduce it as a hook class, since it will then create a breaking change when upgrading. Sorry.

Any suggestions of how to trigger javascript on logout then?

danharrin:
What is the Javascript going to do?

dododedodonl:
> What is the Javascript going to do?

Clear 3 keys from local storage

danharrin:
How about doing that in a render hook `<script>` on the login page, when the login page is loaded?

dododedodonl:
> How about doing that in a render hook `<script>` on the login page, when the login page is loaded?

It effectively signs the user out from a different but linked system running on a subpath of the domain, so that would not be viable because of security.

danharrin:
How about adding a value to the session on logout (in the LogoutResponse object or similar), and then only outputting the script on login if that is present

dododedodonl:
> How about adding a value to the session on logout (in the LogoutResponse object or similar), and then only outputting the script on login if that is present

Well, the problem is, users are not logged out from the linked system when they think they are. That system can be independently reached, but logs in using the same laravel.

PR #15098, opened 2024-12-16
Update 07-advanced.md, document isAbsolute for $get

Lockyc:
Added missing documentation of $get function. Found the solution in https://github.com/filamentphp/filament/discussions/4843 by @danharrin

## Description

I was having difficulty accessing property values which I was unable to resolve until I stumbled across @danharrin's solution. I had come across the `isAbsolute` property, but i did not realise it was the absolute path of the Livewire component.

I have added the information @danharrin provided with a code example in the appropriate section of the Advanced Forms page.

## Functional changes

### DIFF
--- a/packages/forms/docs/07-advanced.md
+++ b/packages/forms/docs/07-advanced.md
@@ -143,6 +143,16 @@ function (Get $get) {
 }
 ```
 
+By default `$get` uses relative paths, you can use the second parameter `isAbsolute` to provide the absolute path of the field. This path is not "absolute" from the current form, it is absolute for the entire Livewire component. You will need to prefix the whichever state path you're trying to access (usually `data.` or `mountedActionData.` 
+```php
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks! Going to keep this out of the documentation still though, as it's tricky to explain how to find the correct absolute path (do you choose `data` or `mountedActionData` or something else)? I think it might cause more confusion than it's worth, and I want to encourage users to use `../../` to traverse the state paths instead. Maybe it will get simpler in the future.

Lockyc:
👍 ok no worries, I figured that might be the case.\
I hoped it would just help the next person find the solution and figure it out with some troubleshooting.

I only found the `../../` traversal pattern after this pr.\
Its documented in the Repeater documentation, not with the rest of the `$get` documentation.\
I'm not sure it would have solved my problem anyway.

PR #15118, opened 2024-12-18
fix: Table column headers hidden on empty state

### LINKED ISSUE #13734: isIndividual column search removes table when no results
### Package

filament/filament

### Package Version

v3.2.96

### Laravel Version

v11.7.0

### Livewire Version

v3.5.1

### PHP Version

PHP 8.2

### Problem description

I have a TextColumn that I want to make searchable and set isIndividual to true:
```php
TextColumn::make('user.full_name')
  ->label('Name')
  ->sortable(['first_name', 'last_name', 'company'])
  ->searchable(['first_name', 'last_name', 'company'], isIndividual: true),
```

Which works great. But when I search for a name that does not exist, the table and search field is removed. 
This adds an annoying element with using isIndividual: true, as you have to clear the added filter and retype the search query every time you search for something that does not exist, instead of just changing your search input.

### Expected behavior

I expected that the table and search field would stay visible, when no search results are found, and allow me to change the search input directly.

### Steps to reproduce

Go to the filament demo site to the customer resource:
https://demo.filamentphp.com/shop/customers

Type in a name that does not exist in the first name column.

NA

### Relevant log output

_No response_

danharrin:
Closes #13734.


avinmaster:
Thanks!

PR #15124, opened 2024-12-18
Fix inconsistent escaping behaviour

ralphjsmit:
There is some inconsistency with regards to how attributes are escaped. In most places they are already not escaped, this PR prevents the escaping on the rest of the places in the framework where we are using `->merge()`. 

This solves issues where eg `->extraEntryWrapperAttributes([
    'class' => '[&_dt>span]:text-gray-500',
])` is escaped and ends up as `[&amp...]` in the html.


danharrin:
Can you please do this in v4? Technically it is a breaking change and while I think we should make it consistent, it could cause security issues for people who are expecting these to be escaped. As a workaround in v3 you might be able to pass an HtmlString() where you don't want these escaped.

PR #15139, opened 2024-12-19
Resolve exception for search on json columns on a relationship

dododedodonl:
## Used code
Using this snippet, a table can display the json columns content (only present in 1 of the 4 rows):
```php
Tables\Columns\TextColumn::make('contactDetailSubmission.data.name')
    ->searchable()
    ->label(__('Name')),
```
- `contactDetailSubmission` is a relationship
- `data.name` is the `name` field in the `data` JSON column

afbeelding

## Problem
However, when searching (eg. `g`), this produces an exception:
afbeelding

## Result after change
This fix resolves the exception while preserving the original behavior:
afbeelding

## Visual changes

--

## Functional changes


dododedodonl:
Are any changes needed to this PR?

danharrin:
I'm gonna need to refactor it, might even push it into v4 so I can make changes to allow it to be a bit cleaner.

This shouldn't be blocking your work since you can pass a custom function to `searchable()` to handle this instead

TRYCOOF (commented):
nose works

PR #15151, opened 2024-12-20
allowing to add a closure to handle visibility for some actions

ht3aa:
## Description

Problem: can't control the visibility of some actions in repeater form field without using the (deleteAction, addAction...) methods. also the ($state) of each item isn't evaluated like other fields as expected.

Solution: adding ability to pass a Closure to visibility method for actions

- add action
- add between action
- delete action
- clone action
- move down action
- move up action
- reorder action

and evaluate the state to directly get the state of the item using 
`
$state[$arguments['item']]
`

## Visual changes
here are some photos for delete action

before:
Screenshot from 2024-12-20 13-50-14

after:
Screenshot from 2024-12-20 13-51-23

## Functional changes

### DIFF
--- a/packages/forms/docs/03-fields/12-repeater.md
+++ b/packages/forms/docs/03-fields/12-repeater.md
@@ -613,7 +613,7 @@ Repeater::make('members')
 
 ### Action Visibility
 
-You can modifiy the visibiltiy for (add, add between, delete, clone, move up, move down, reorder) actions by using the related methods:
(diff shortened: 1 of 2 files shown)


danharrin:

In this case, I would not like to merge this since it is already possible by overriding the action method and passing `->hidden()`, as you said in the original PR. If we added customisation methods for every since action on the repeater too, it would get out of hand, which is why we have the generalised functions for modifying the Action object.

PR #15159, opened 2024-12-21
Disable escaping for attribute merging

ralphjsmit:
Re-target of #15124 to 4.x.


ralphjsmit:
The tests are failing but does not seem to be related. Thanks!

PR #15160, opened 2024-12-21
Prevent creating non-existent models from container

ralphjsmit:
Currently, the closure evaluation system works great by resolving any dependencies not provided by name or type from the container. This is very useful. However, there is one situation where it isn't useful and that is when it is evaluating empty/non-existent Eloquent models from the container.

For example, in Filament V3 we had the nice feature of being able to type an Eloquent model as both `User $record` and `User $user`, which would both work nicely. However, sometimes there are situations where you _think_ you have a `$record`, but actually for some reason you don't (or you have a different type of record when working with eg relationships or repeaters). If you would then still type it as `User $user`, you would get an empty/non-existent Eloquent model from the container. This has caught me in very complex forms out quite a few times and therefore I generally prefer `User $record` now just for this, whilst for the cleanliness I would prefer `User $user`. 

This PR fixes the issue by preventing the container from instantiating an empty Eloquent model if it was not able to resolve dependency using type or name. This prevents users from typing things like `User $user`, thinking they have the record but they actually don't. It should not really impact users, as no one would instantiate their empty models using the container and would only help to prevent bugs. I also added a test to prove the behavior.

Thanks!

### DIFF
--- a/packages/support/src/Concerns/EvaluatesClosures.php
+++ b/packages/support/src/Concerns/EvaluatesClosures.php
@@ -84,7 +84,7 @@ trait EvaluatesClosures
 
         if (
             filled($typedParameterClassName)
-            && ! is_subclass_of($typedParameterClassName, Model::class)
+            && (! is_subclass_of($typedParameterClassName, Model::class) || app()->bound($typedParameterClassName))
         ) {
             return app()->make($typedParameterClassName);
         }
--- a/packages/support/src/Concerns/EvaluatesClosures.php
+++ b/packages/support/src/Concerns/EvaluatesClosures.php
@@ -4,6 +4,7 @@ namespace Filament\Support\Concerns;
 
 use Closure;
 use Illuminate\Contracts\Container\BindingResolutionException;
+use Illuminate\Database\Eloquent\Model;
 use ReflectionFunction;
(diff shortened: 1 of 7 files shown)


ralphjsmit:
There are some changes in un-touched files because of Pint.

packages/support/src/Concerns/EvaluatesClosures.php
```diff
@@ -81,7 +82,10 @@ protected function resolveClosureDependencyForEvaluation(ReflectionParameter $pa
             return $this;
         }
 
-        if (filled($typedParameterClassName)) {
+        if (
+            filled($typedParameterClassName)
+            && ! is_subclass_of($typedParameterClassName, Model::class)
```
danharrin: If the model has been bound to the container manually, we should still resolve it. Someone might want to bind the current tenant to the container for easy access, for example.

danharrin (changes requested):

packages/support/src/Concerns/EvaluatesClosures.php
ralphjsmit: Good point, changed!

danharrin:
Thanks

danharrin:
Thanks

Issue #14814, opened 2024-11-15
CSV Export Loses Global Sort Order Due to Using Chunks (Observed with PostgreSQL)


kaspernowak:
### Package

filament/filament

### Package Version

v3.2.116

### Laravel Version

v11.29.0

### Livewire Version

v3.5.12

### PHP Version

PHP 8.2.22

### Problem description

When exporting data to CSV using Filament's Export action, the global sorting order is not maintained when the data is processed in chunks. This issue has been observed with PostgreSQL under these conditions:

1. The chunk size is smaller than the total number of records
2. Any sort order is applied, whether through:
   - Custom sort via `modifyQueryUsing()`
   - Default table sort configuration
   - User-triggered column sorting in the Filament table

Importantly, the sorting works correctly when the chunk size is larger than the total number of records, confirming this is specifically a chunking issue rather than a general sorting problem.

Current workaround:
- Setting the chunk size larger than the total number of records resolves the issue
- However, this is not a scalable solution for large datasets

Additional context:
The issue appears to be related to how ORDER BY clauses are handled during chunked queries. The fact that the sorting works correctly when processing all records in a single chunk suggests that the chunking mechanism isn't properly maintaining the global sort order when splitting the data.

This affects all sorting methods available in Filament:
- Default table sorting
- Interactive column sorting
- Custom query modifications

This might require either:
1. A modification to how chunks are processed to maintain global order
2. A database-specific implementation of the chunking strategy
3. A warning in the documentation about this limitation

Environment:
- Database: PostgreSQL 14.x
- Records: Works correctly when chunk size > total records
- Sorting: Affects all available sorting methods in Filament

### Expected behavior

The CSV export should maintain a consistent global sort order regardless of:
1. The chunk size used
2. The total number of records
3. How the sort order was specified (default table sort, column sorting, or custom sort)
4. The specific sort order applied

The chunking process should preserve the global ordering while still allowing for efficient processing of large datasets.

### Steps to reproduce

See ReadMe in Reproduction repository.

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/kaspernowak/filament-issue

bzy107:
I will write about what I have discovered through my investigation.

I analyzed the SQL queries executed by MySQL and PostgreSQL. Please note that the investigation is still ongoing.
I will update this document as progress is made.

First, regarding the differences in the SQL queries executed by MySQL and PostgreSQL:
[...]

PR #15181, opened 2024-12-25
Fix: Add conditions to DISTINCT and modify reorder conditions in PostgreSQL

bzy107:
## Description
This is a fix for issue #14814.

The problem was that in PostgreSQL, the results were always ordered by `id`. I added a condition to sort by the value specified on the screen if sorting is applied; otherwise, the results will fall back to sorting by `id`.

Additionally, the `DISTINCT` clause was always applied to `id`. A condition was added to handle cases where sorting is based on id.

## Functional changes

### DIFF
--- a/packages/actions/src/Exports/Jobs/PrepareCsvExport.php
+++ b/packages/actions/src/Exports/Jobs/PrepareCsvExport.php
@@ -69,13 +69,18 @@ class PrepareCsvExport implements ShouldQueue
 
         if ($databaseConnection->getDriverName() === 'pgsql') {
             $originalOrders = collect($query->getQuery()->orders)
-                ->reject(fn (array $order): bool => in_array($order['column'] ?? null, [$keyName, $qualifiedKeyName]))
+                ->reject(fn (array $order): bool => in_array($order['column'] ?? null, [$qualifiedKeyName]))
                 ->unique('column');
 
             /** @var array<string, mixed> $originalBindings */
             $originalBindings = $query->getRawBindings();
 
-            $query->reorder($qualifiedKeyName);
+            if (! empty($originalOrders->all())) {
+                $query->reorder($originalOrders[0]['column'], $originalOrders[0]['direction']);
+                $originalOrders->forget(0);
+            } else {
+                $query->reorder($qualifiedKeyName);
+            }
 
             foreach ($originalOrders as $order) {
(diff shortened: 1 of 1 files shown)


packages/actions/src/Exports/Jobs/PrepareCsvExport.php
```diff
@@ -69,13 +69,18 @@ public function handle(): void
 
         if ($databaseConnection->getDriverName() === 'pgsql') {
             $originalOrders = collect($query->getQuery()->orders)
-                ->reject(fn (array $order): bool => in_array($order['column'] ?? null, [$keyName, $qualifiedKeyName]))
```
danharrin, line 72: Can you please explain the removal of the non-qualified key name from this array? That is the part I am most unsure about in this PR

danharrin (changes requested):

packages/actions/src/Exports/Jobs/PrepareCsvExport.php
bzy107: This is because sorting in DESC order causes an issue when the table has an id column.
line 82`$query->reorder($qualifiedKeyName);` does not specify a sort direction, so it always defaults to ASC.

スクリーンショット 2024-12-31 141555

result: 
```csv
ID,DATA_OF_BIRTH,NAME,TYPE,HAS_RECOVERED
100,2016-11-21,Dixie,cat,false
99,2018-11-29,Arturo,cat,true
...
2,2019-04-27,Pattie,rabbit,false
1,2018-07-02,Krystal,dog,true
150,2019-04-27,Sydnee,dog,false
149,2019-12-13,Leonel,rabbit,false
```

packages/actions/src/Exports/Jobs/PrepareCsvExport.php
```diff
@@ -69,13 +69,18 @@ public function handle(): void
...
+                ->reject(fn (array $order): bool => in_array($order['column'] ?? null, [$qualifiedKeyName]))
```
```suggestion
                ->reject(fn (array $order): bool => ($order['column'] ?? null) === $qualifiedKeyName)
```

danharrin (changes requested):

packages/actions/src/Exports/Jobs/PrepareCsvExport.php
bzy107: Thank you for the review. fixed it.

danharrin:
Thank you for the investigation work and fix

PR #15196, opened 2024-12-28
Allow closure based table `Group` labels

ralphjsmit:
This PR allows the `getLabel()` method of Table Groups to be based on a closure. It also ensures that the group evaluation parameters for `$table`/`$livewire` are correctly received. In line with other code the table is set on the group as soon as possible (when pushing the groups), or on instantiation of the groups when only providing the string attribute.

Thanks!

### DIFF
--- a/packages/tables/src/Table/Concerns/CanGroupRecords.php
+++ b/packages/tables/src/Table/Concerns/CanGroupRecords.php
@@ -72,6 +72,10 @@ trait CanGroupRecords
 
     public function defaultGroup(string | Group | null $group): static
     {
+        if ($group instanceof Group) {
+            $group->table($this);
+        }
+
         $this->defaultGroup = $group;
 
         return $this;
@@ -82,6 +86,12 @@ trait CanGroupRecords
      */
     public function groups(array | Closure $groups): static
     {
+        foreach ($groups as $group) {
+            if ($group instanceof Group) {
+                $group->table($this);
+            }
+        }
+
         $this->groups = $groups;
 
         return $this;
@@ -159,7 +169,8 @@ trait CanGroupRecords
             return $group;
         }
 
-        return Group::make($this->defaultGroup);
+        return Group::make($this->defaultGroup)
+            ->table($this);
     }
 
     /**
@@ -171,7 +182,8 @@ trait CanGroupRecords
             $this->evaluate($this->groups),
             function (array $carry, $group): array {
                 if (! $group instanceof Group) {
-                    $group = Group::make($group);
+                    $group = Group::make($group)
+                        ->table($this);
                 }
 
                 $carry[$group->getId()] = $group;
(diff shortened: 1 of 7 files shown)


packages/tables/src/Table/Concerns/CanGroupRecords.php
```diff
@@ -82,6 +86,12 @@ public function defaultGroup(string | Group | null $group): static
      */
     public function groups(array | Closure $groups): static
     {
+        foreach ($groups as $group) {
+            if ($group instanceof Group) {
+                $group->table($this);
+            }
+        }
```
danharrin: If groups are passed in a closure I don't think the table gets passed, I think the table needs to be injected in when the groups property is consumed

packages/tables/src/Table/Concerns/CanGroupRecords.php
```diff
@@ -72,6 +72,10 @@ public function groupingDirectionSettingHidden(bool | Closure $condition = true)
 
     public function defaultGroup(string | Group | null $group): static
     {
+        if ($group instanceof Group) {
+            $group->table($this);
+        }
```
danharrin: Can we move this to the place where the defaultGroup gets read so if we introduce closure support this won't break

danharrin (changes requested):

packages/tables/src/Table/Concerns/CanGroupRecords.php
ralphjsmit: Ah you're right, not sure how I missed that. Fixed now!

packages/tables/src/Table/Concerns/CanGroupRecords.php
ralphjsmit: Done

ralphjsmit:
Thanks for the review, I just updated it and also pushed a little test.

tests/src/Tables/TestCase.php
```diff
@@ -11,6 +12,7 @@ protected function getPackageProviders($app): array
         return [
             ...parent::getPackageProviders($app),
             TablesServiceProvider::class,
+            PowerJoinsServiceProvider::class,
```
danharrin: Can you move this to the base test case please? There is reasonable possibility we will need this elsewhere in the future.

tests/src/Tables/GroupingTest.php
danharrin, line 1: Can `invade($livewire)->table` be replaced with `$livewire->getTable()` since it is public?

danharrin (changes requested):

tests/src/Tables/TestCase.php
ralphjsmit: Done

tests/src/Tables/GroupingTest.php
ralphjsmit: Ah nice, didn't even know that method. Done

danharrin:
Thanks

PR #15215, opened 2024-12-31
Added TimePicker and time method with generate argument

achyutkneupane:
## Description

When using `--generate` with `php artisan make:filament-resource`, it detects `date`, `datetime`, and `timestamps` but not `time`.
So, I added the time field and column generation.

## Visual changes

The description seems clear but here is what it generates:

image

## Functional changes


danharrin:
Hi, apologies but we are no longer accepting new features for v3 as per #13744. I would accept a PR for this feature when v4 is released, if you are still interested. Thank you!

achyutkneupane:
@danharrin This thing is missing in v4 too. How about a PR for this on v4?

danharrin:
Yes I would accept a v4 PR

PR #15217, opened 2024-12-31
Added TimePicker and time method with generate argument

achyutkneupane:
## Description

When using `--generate` with `php artisan make:filament-resource`, it detects `date`, `datetime`, and `timestamps` but not `time`.
So, I added the time field and column generation.

## Visual changes

The description seems clear but here is what it generates:

image

image

## Functional changes

### DIFF
--- a/packages/tables/src/Commands/FileGenerators/Concerns/CanGenerateModelTables.php
+++ b/packages/tables/src/Commands/FileGenerators/Concerns/CanGenerateModelTables.php
@@ -143,6 +143,13 @@ trait CanGenerateModelTables
                 }
 
                 if (in_array($type['name'], [
+                    'time',
+                ])) {
+                    $columnData['time'] = [];
+                    $columnData['sortable'] = [];
+                }
+
+                if (in_array($type['name'], [
                     'datetime',
                     'timestamp',
                 ])) {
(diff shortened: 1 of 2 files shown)


packages/tables/src/Commands/FileGenerators/Concerns/CanGenerateModelTables.php
```diff
@@ -142,6 +142,13 @@ public function getTableColumns(?string $model = null, array $exceptColumns = []
                     $columnData['sortable'] = [];
                 }
 
+                if (in_array($type['name'], [
+                    'time',
+                ])) {
```
danharrin:
```suggestion
                if ($type['name'] === 'time') {
```

danharrin (changes requested):

packages/tables/src/Commands/FileGenerators/Concerns/CanGenerateModelTables.php
achyutkneupane: I followed what was done with the `date` type.

I will make the change accordingly.

danharrin: Good spot, I didn't see that. Will fix

PR #15218, opened 2024-12-31
Select field on enum type when using generate

achyutkneupane:
## Description

When using `--generate` with `php artisan make:filament-resource`, `enum` table type will generate `Forms\Components\Select`.

Along with:
- `options` from allowed array
- Default value

It follows the basic label transformation script (kebab -> replace -> ucfirst) for options.

## Visual changes

image

image

image

## Functional changes


danharrin (changes requested):
Usually people also pair this with an enum Eloquent model cast, or even use the enum class instead of an enum column. How about detecting on the model if an enum cast is used, and if it is pass the enum class name to the `->options()`? And if there is no enum cast and the column is an enum type, then do this instead.

achyutkneupane:
@danharrin I went through a lot of research on how to set class as an argument.

Thanks to the `Literal` import you used below, I got the idea.

achyutkneupane (commented):
@danharrin Can you verify the changes?

danharrin:
Thanks, I will take on the changes from here

danharrin:
Thanks

PR #15220, opened 2025-01-01
Update date-time-picker.js add Iraq and saudia an Kuwait local

NazarAli:
add local for iraq and saudia and kuwait and tunis

## Functional changes


danharrin:
Since these aren't available locales in Filament how will these locales be used?

NazarAli:
it found in package dayjs

https://github.com/iamkun/dayjs/tree/dev/src/locale

danharrin:
I know, but how do you use it in Filament? If you set the locale to one of these, you will not have any of the other translations in the UI

danharrin:
Closing as I don't think this works since Filament does not have these locales as options in the main translation files

Issue #13643, opened 2024-07-19
`ExportAction` / `ImportAction` doesn't respect $modelLabel defined in Resource class


ayrtonandino:
### Package

filament/actions

### Package Version

v3.2.95

### Laravel Version

v11.16.0

### Livewire Version

v3.5.4

### PHP Version

PHP 8.3.6

### Problem description

`Filament\Actions\ExportAction` and `Filament\Actions\ImportAction` doesn't respect $modelLabel defined in Resource class.

### Expected behavior

To produce the same result as `Filament\Tables\Actions\ExportBulkAction`, `Filament\Tables\Actions\ExportAction` and `Filament\Tables\Actions\ImportAction`.

### Steps to reproduce

Install filament-demo as usual, app is set to Spanish to show malformed labels.

## For `Filament\Actions\ExportAction`:

Go to [http://127.0.0.1:8000/shop/products/brands](http://127.0.0.1:8000/shop/products/brands), select some records for bulkActions, see produced exporter labels, "Exportar brands" should be "Exportar marcas", same as labels generated by  `Filament\Tables\Actions\ExportBulkAction` and `Filament\Tables\Actions\ExportAction`.

![ExportAction img](https://github.com/user-attachments/assets/21188e95-ed3e-4906-b41e-73ede63eb38d)

## For `Filament\Actions\ImportAction`

Go to [http://127.0.0.1:8000/shop/products/categories](http://127.0.0.1:8000/shop/products/categories), see produced importer labels, "Importar categories" should be "Importar categorías", same as label generated by  `Filament\Tables\Actions\ImportAction`.

![ImportAction img](https://github.com/user-attachments/assets/923524d8-ada8-41ed-bba0-439d3fbe737a)

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/ayrtonandino/filament-exporter-importer-bug

porozhnyy:
They also ignore the override of the `getTableRecordKey` method.

danharrin:
This should be addressed in v4 as part of an architecture rethink

PR #15237, opened 2025-01-04
fix: respect the model label defined in the resource class when using ExportAction 3.x

borjajimnz:
## Description

As described in https://github.com/filamentphp/filament/issues/13643, I'm not sure if this approach is the best, but it does fix the bug.

The thing is, the only way I found to get the model label is through the Livewire / table component. Is there another way to do this?

## Functional changes


danharrin:
I think this has been addressed in v4 as part of an architecture rethink, but this fix isn't correct as I don't think you can pass a custom model label just for this action anymore

Issue #13552, opened 2024-07-12
Namespace Collision with Model Name "Form"


rvalenzuelaferrada:
### Package

filament/filament

### Package Version

v3.2.93

### Laravel Version

v11.9

### PHP Version

PHP 8.2

### Problem description

When creating a resource using FilamentPHP, I encountered a namespace collision. Specifically, I created a resource named **Form** using the command:

```
php artisan make:filament-resource Form --soft-deletes --view --model --migration --factory
```

The command completes successfully with the following output:

```
INFO  Migration [database/migrations/2024_07_12_143938_create_forms_table.php] created successfully.
INFO  Filament resource [app/Filament/Resources/FormResource.php] created successfully.
```

However, the generated files contain a namespace conflict. My model's name Form collides with the Filament class Form, causing a fatal error. The error arises in the FormResource.php file due to the following lines:

```php
➜ use App\Models\Form;
use Filament\Forms;
➜ use Filament\Forms\Form;
```
This results in a collision between App\Models\Form and Filament\Forms\Form.

I received the following error message when attempting to use the model:
```php
Symfony\Component\ErrorHandler\Error\FatalError

Cannot use Filament\Forms\Form as Form because the name is already in use

at app/Filament/Resources/FormResource.php:9
    5▕ use App\Filament\Resources\FormResource\Pages;
    6▕ use App\Filament\Resources\FormResource\RelationManagers;
    7▕ use App\Models\Form;
    8▕ use Filament\Forms;
➜   9▕ use Filament\Forms\Form;
   10▕ use Filament\Resources\Resource;
   11▕ use Filament\Tables;
   12▕ use Filament\Tables\Table;
   13▕ use Illuminate\Database\Eloquent\Builder;

Whoops\Exception\ErrorException

Cannot use Filament\Forms\Form as Form because the name is already in use

at app/Filament/Resources/FormResource.php:9
    5▕ use App\Filament\Resources\FormResource\Pages;
    6▕ use App\Filament\Resources\FormResource\RelationManagers;
    7▕ use App\Models\Form;
    8▕ use Filament\Forms;
➜   9▕ use Filament\Forms\Form;
   10▕ use Filament\Resources\Resource;
   11▕ use Filament\Tables;
   12▕ use Filament\Tables\Table;
   13▕ use Illuminate\Database\Eloquent\Builder;
```

### Expected behavior

[...]

PR #15238, opened 2025-01-04
fix: add Form and Table to prevent collision while creating a resource

borjajimnz:
## Description

This prevents class collisions when the model is called "Form", "Table", or "Resource", as described in https://github.com/filamentphp/filament/issues/13552 

## Functional changes


danharrin:
Thanks

Issue #14852, opened 2024-11-20
Closure inside getStateUsing() called 3 times per row


jonaaix:
### Package

filament/tables

### Package Version

v3.2.124

### Laravel Version

v11.33.2

### Livewire Version

default

### PHP Version

8.3.13

### Problem description

When using `->getStateUsing()` to fill a column with a custom value, the Closure seems to be called 3 times per row!
<img width="825" alt="image" src="https://github.com/user-attachments/assets/3007e23d-ba5d-4c07-a331-563f2c1c8d2f">

**If this closure contains a custom query, this will lead to triple query execution per row!**
<img width="1263" alt="Bildschirmfoto 2024-11-20 um 16 16 22" src="https://github.com/user-attachments/assets/bf33e0f6-b75c-4fe8-bc1a-7d73d03b0832">

### Expected behavior

The Closure should be executed once on page load

### Steps to reproduce

Example Column:

```php
Tables\Columns\TextColumn::make('permission_number')
  ->label('Permissions')
  ->badge()
  ->getStateUsing(function (User $user) {
      return $user->permissions()->count();
  }),
```

### Reproduction repository (issue will be closed if this is not valid)

[https://github.com/btxtiger/filament-getstateusing-bug-14852](https://github.com/btxtiger/filament-getstateusing-bug-14852)

Cybrarist:
i think you should use formatStateUsing instead of getStateUsing()

jonaaix:
@Cybrarist this does not work, since the state of the column should be _**any**_ query or value (probably also a result of an HTTP request, Cache, other service, ...), and not dependant on a model column.

Cybrarist:
@btxtiger  oh ok, i get what you mean, then why not doing it like the following 

```php 
$table->modifyQueryUsing(function (Builder $query) {
      $query->withCount('permissions');
})
```

and then you use 'permissions_count' in your TextColumn.

jonaaix:
@Cybrarist yes this would be a workaround for this specific example, but would not solve the actual problem.
Let me show you a different example:

```php
Tables\Columns\TextColumn::make('permission_granted')
  ->label('Permissions')
  ->badge()
  ->getStateUsing(function (User $user) {
      return Http::get('https://dummy-auth-provider.com/api/check-permissions', ['uuid' => $user->uuid])
          ->json('permissions_granted') ? 'Granted' : 'Not Granted';
  });
```

[...]

PR #15246, opened 2025-01-04
fix: prevent multiple evaluations when getStateUsing is called #14852

borjajimnz:
## Description

As mentioned in https://github.com/filamentphp/filament/issues/14852, this issue was causing multiple evaluations. This fix ensures that only the final evaluation is performed, which I believe is the correct approach.

## Functional changes

### DIFF
--- a/packages/support/src/Concerns/HasCellState.php
+++ b/packages/support/src/Concerns/HasCellState.php
@@ -76,7 +76,7 @@ trait HasCellState
         }
 
         $state = ($this->getStateUsing !== null) ?
-            $this->evaluate($this->getStateUsing) :
+            $this->getStateUsing :
             $this->getStateFromRecord();
 
         if (is_string($state) && ($separator = $this->getSeparator())) {


mc-fu:
this breaks the use of fn and function within getStateUsing!

borjajimnz:
> this breaks the use of fn and function within getStateUsing!

Have you tested it @mc-fu ? is not breaking nothing for me.

https://www.youtube.com/watch?v=VzFf0Vas0j8

packages/support/src/Concerns/HasCellState.php
```diff
@@ -76,7 +76,7 @@ public function getState(): mixed
         }
 
         $state = ($this->getStateUsing !== null) ?
-            $this->evaluate($this->getStateUsing) :
+            $this->getStateUsing :
             $this->getStateFromRecord();
 
         if (is_string($state) && ($separator = $this->getSeparator())) {
```
danharrin, line 82: The breaking comment is correct, if the function returns a string that needs separating here or a blank value it would break

danharrin (changes requested):

mc-fu:
Yeah, sorry, I needed to be more precise:

When using "`->money('EUR')`" option on this as well, I get the following error with your fix:

`Too few arguments to function App\Filament\Resources\StlResource::App\Filament\Resources\{closure}(), 0 passed in /var/www/testdev/vendor/laravel/framework/src/Illuminate/Collections/helpers.php on line 236 and exactly 1 expected `

borjajimnz:
This is more complex than I initially thought; I'll need some time to process it. I'll reopen the PR if I find a way to handle it.

robinmoisson:
@borjajimnz curious if you considered caching the state result [like this](https://github.com/filamentphp/filament/issues/14852#issuecomment-2650199984)? I'm not super familiar with the Filament codebase so I don't know if it'd break something else.

Issue #14876, opened 2024-11-25
Table footer renders above content instead of bottom


GigaGiorgadze:
### Package

filament/filament

### Package Version

v3.2

### Laravel Version

v11.31

### Livewire Version

3.4

### PHP Version

8.3

### Problem description

table footer does not appear in its proper places and instead renders at the top outside of table content. this problem is only fixed when i use Split layout component 
![image](https://github.com/user-attachments/assets/a782c635-8259-4fcb-afed-df13cdf91991)
without split layout
![image](https://github.com/user-attachments/assets/f4c87bc8-eee2-41b0-86aa-f0a8e7c77d13)
with split layout

### Expected behavior

footer should render at the bottom of the content above navigation, just like Summaries appear and i should not need to use split layout to fix it

### Steps to reproduce

download repository URL, it is normal laravel project with nothing fancy so basic setup of database and key generation will be required nothing more, after that create a user and log in. users table renders footer which appears at the top but posts table renders footer which appears properly at the bottom because posts uses split layout while users does not

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/GigaGiorgadze/table-footer-filament-issue

borjajimnz:
Columns and records are passed as props to your view. You can simply access them like this.

```php
@props([
    'columns',
    'records'
])

<td colspan="{{ count($columns) }}">
    Your footer :)
</td>
```

I don’t think this is actually a bug; it seems more like a small gap in the documentation. Maybe a PR with this information could be a good idea? @danharrin

danharrin:
Sounds good

GigaGiorgadze:
@borjajimnz ah i see, thank you very much ❤️

danharrin:
Thanks @borjajimnz

PR #15247, opened 2025-01-04
Feature/issues 14876 add footer content tables docs

borjajimnz:
## Description

Add contentFooter to tables/advanced docs with a brief description regarding the issue discussed in https://github.com/filamentphp/filament/issues/14876

## Functional changes


danharrin:
Thanks for this. I think the developer experience is bad for this feature so I am going to continue to omit it from the documentation. I will think about improving the developer experience in v4 by allowing the user to use a `<div>` instead of a table element, and add to the documentation if I think it is better.

PR #15261, opened 2025-01-06
Fix: `RelationshipJoiner` raw & sub-query orders

ralphjsmit:
This PR fixes two situations in the `RelationshipJoiner` that did not work:

- `->orderByRaw("CASE WHEN role = 'PIC' THEN 1 ELSE 2 END")`
- `->orderBy(fn (Builder $query) => $query->selectRaw("CASE WHEN role = 'PIC' THEN 1 ELSE 2 END"))`

This PR fixes both cases to make the `RelationshipJoiner` handle them correctly. There might always be edge cases regarding this part of the `RelationshipJoiner`, like when the raw order by contains a binding, but at least this makes the joiner handle a few more common cases.

### DIFF
--- a/packages/support/src/Services/RelationshipJoiner.php
+++ b/packages/support/src/Services/RelationshipJoiner.php
@@ -5,6 +5,7 @@ namespace Filament\Support\Services;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Relations\BelongsToMany;
 use Illuminate\Database\Eloquent\Relations\Relation;
+use Illuminate\Database\Query\Expression;
 use Illuminate\Database\Query\JoinClause;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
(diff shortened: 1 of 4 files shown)


danharrin:
The code looks great, but would you mind writing some tests for this? It's getting to a point where the conditions are quite complex and I would hate for them to break in the future.

ralphjsmit:
Yes, sure! Will do later this week.

ralphjsmit:
@danharrin I just added tests for the `RelationshipJoiner`, and also confirmed that it failed on the old code just to be sure.

danharrin:
Thanks Ralph!

howdu:
This doesn't work if the raw sql order contains the direction.

E.g
```php
->orderByRaw($this->getQuery()->getGrammar()->wrap('name->' . $locale) . ' ' . $direction);
```

Produces this sql:
```sql
select
  distinct `tags`.*,
  json_unquote(json_extract(`name`, '$."en"')) asc
from
  `tags`
  left join `taggables` on `tags`.`id` = `taggables`.`tag_id`
```

danharrin:
Please open an issue or a PR with a failing test

Issue #15264, opened 2025-01-06
HasOneThrough parent class changed in laravel 11.x


musaberatbahadir:
### Package

filament/filament

### Package Version

v3.2

### Laravel Version

v11.31

### PHP Version

8.2

### Problem description

Because of parent class of HasOneThrough is changed in laravel 11.x [here](https://github.com/laravel/framework/commit/4529498671704a2ea4bae4af1e3f8ee48ca8dfd0#diff-9308216b3c4246a75d7f24d33bb5cd2e9979191ade49811820cb58884a4fcf5cR17) CreateAction is throwing `Call to undefined method Illuminate\Database\Eloquent\Relations\HasOneThrough::save()` 

### Expected behavior

Relationship control needs to be changed to from `HasManyThrough` to `HasOneOrManyThrough` on `CreateAction` `setUp`

### Steps to reproduce

`composer install` and run the tests on given repository, I've created a small scenario to get the same error I mentioned in description

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/musaberatbahadir/create-action-bug

PR #15265, opened 2025-01-06
Fix: used new parent relation on instanceof control for has one through

musaberatbahadir:
## Description
fix for the [issue](https://github.com/filamentphp/filament/issues/15264) I created. CreateAction is checking the correct class on instance of control now.

### DIFF
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -1145,7 +1145,7 @@ class Select extends Field implements Contracts\CanDisableOptions, Contracts\Has
         return parent::getLabel();
     }
 
-    public function getRelationship(): BelongsTo | BelongsToMany | HasOneOrMany | HasManyThrough | BelongsToThrough | null
+    public function getRelationship(): BelongsTo | BelongsToMany | HasOneOrMany | HasManyThrough | HasOneOrManyThrough | BelongsToThrough | null
     {
         if (blank($this->getRelationshipName())) {
             return null;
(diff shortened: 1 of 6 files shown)


danharrin (changes requested):
This would break Laravel 10.x compatibility, right? Can you do a `class_exists()` check and check the correct parent class based on that please?

musaberatbahadir:
@danharrin I've added the check you have mentioned above

musaberatbahadir:
@danharrin I also realized that `instanceof HasManyThrough` control causes other errors on my project like in `Filament\Tables\Table\Concerns\HasQuery`, do you want me to replace all of them as I did on the last commit, or do you wanna introduce some config level check by controlling laravel version or class_exists function?
there are 6 more `instanceof HasManyThrough` control

danharrin:
Please can you do the `class_exists()` in all places?

musaberatbahadir:
Done! @danharrin

packages/forms/src/Components/Select.php
danharrin, line 1: Does the return type of `getRelationship()` need updating? Any other references to this class?

danharrin: If we did, I think we would need to retain both `HasManyThrough` and `HasOneOrManyThrough` on the return type for backwards compatibility, right?

danharrin (changes requested):

packages/forms/src/Components/Select.php
musaberatbahadir: Good catch! I fixed that too.

danharrin:
Thanks

PR #15267, opened 2025-01-06
Fix forced lower case search for json field

dododedodonl:
## Description

When using `->forceSearchCaseInsensitive()` on a column that is accessed via json extraction (eg. `data.field`), it results in an exception. This is because `lower(data->field)` does not work for mysql (and mariadb), it should be ```lower(json_extract(`data`, '$."field"'))```.

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/support/src/helpers.php
+++ b/packages/support/src/helpers.php
@@ -181,6 +181,12 @@ if (! function_exists('Filament\Support\generate_search_column_expression')) {
         };
 
         if ($isSearchForcedCaseInsensitive) {
+            if (in_array($driverName, ['mysql', 'mariadb'], true) && str_contains($column, '->')) {
+                [$field, $path] = invade($databaseConnection->getQueryGrammar())->wrapJsonFieldAndPath($column);
+
+                $column = "json_extract({$field}{$path})";
+            }
+
             $column = "lower({$column})";
         }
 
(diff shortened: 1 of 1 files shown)


packages/support/src/helpers.php
```diff
@@ -181,6 +181,12 @@ function generate_search_column_expression(string $column, ?bool $isSearchForced
         };
 
         if ($isSearchForcedCaseInsensitive) {
+            if (in_array($driverName, ['mysql', 'mariadb'], true) && str_contains($column, '->')) {
```
danharrin: If the column already contains `json_extract()` (as per the check above), do we still need to do this?

danharrin (changes requested):

packages/support/src/helpers.php
dododedodonl: I am not quite sure, in what situation would there be `json_extract(` and `->` in the column? I presumed either one of them would be present, as `->` is a short hand/alias for `json_extract(`.  When the column starts with `json_extract(` it is probably not needed to do this (and would probably not work correctly)

dododedodonl: I changed it to exclude the situation when the column already starts with `json_extract(`, which is probably the most careful solution

danharrin: `json_extract(` is added by the translatable plugin for translatable fields before being passed to this function

danharrin:
Thanks

PR #15274, opened 2025-01-07
bug: Fix type error in Asset

davidyell:
## Description

The `$path` class property is `null` by default but the accessor method can't return `null` which causes a `TypeError`

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/support/src/Assets/Asset.php
+++ b/packages/support/src/Assets/Asset.php
@@ -13,15 +13,15 @@ abstract class Asset
 
     protected string $package;
 
-    protected ?string $path = null;
+    protected string $path = null;
 
-    final public function __construct(string $id, ?string $path = null)
+    final public function __construct(string $id, ?string $path = '/')
     {
         $this->id = $id;
         $this->path = $path;
     }
 
-    public static function make(string $id, ?string $path = null): static
+    public static function make(string $id, ?string $path = '/'): static
     {
         return app(static::class, ['id' => $id, 'path' => $path]);
     }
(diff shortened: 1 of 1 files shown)


davidyell:
Seems there is more upstream code which seems to always assume this is not null, due to the accessor method. So perhaps the real fix is to not allow a `null` at all, or default to string `/`

danharrin:
Can you hint at how you got here, what were you doing when you encountered this problem? I just need a bit more context really

davidyell:
So I was trying to ensure that my Laravel 11 apps `app.js` was included into the Filament layout.

Then I found myself over here https://filamentphp.com/docs/3.x/support/assets#registering-javascript-files

Which gives the code example of 

```php
use Filament\Support\Assets\Js;
 
FilamentAsset::register([
    Js::make('custom-script', __DIR__ . '/../../resources/js/custom.js'),
]);
```

So I tried that, but then found that if I omitted the `$path` param, I got a TypeError.

danharrin:
Yeah honestly not really sure how this code happened, seems suspicious. Thanks

danharrin:
I found why this needs to be optional (only just noticed PHPStan didn't run because it wasn't approved here, would have caught it)

https://github.com/filamentphp/filament/blob/cc95f8a020d98d585e04ba5017ca175071043fe2/packages/panels/src/Panel/Concerns/HasTheme.php#L57

I am going to revert this and allow null to be returned from `getPath()` instead

PR #15281, opened 2025-01-08
ensures `filament.xxx.home` is always available

bilogic:
## Description

Addresses #15257

1. Ensures that the route name `filament.<panel-id>.home` won't be overwritten when the dashboard page is used
2. The route `/` still works with or without the dashboard, so backwards compatible

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/panels/routes/web.php
+++ b/packages/panels/routes/web.php
@@ -92,7 +92,8 @@ Route::name('filament.')
                                             $routes($panel);
                                         }
 
-                                        Route::get('/', RedirectToHomeController::class)->name('home');
+                                        Route::get('/', RedirectToHomeController::class);
+                                        Route::get('/redirect-to-home', RedirectToHomeController::class)->name('home');
 
                                         Route::name('tenant.')->group(function () use ($panel): void {
                                             if ($panel->hasTenantBilling()) {


danharrin:
Can you redirect between panels by generating a URL like `Filament::getPanel('admin')->getUrl()`? I don't like the idea of having a route that doesn't do anything to be honest

danharrin:
Gonna close this because of the above, but please let me know what I am missing, I think it should be relatively easy to generate a URL for each panel without needing the ghost route. I can reopen if you let me know how I am wrong.

bilogic:
> Can you redirect between panels by generating a URL like `Filament::getPanel('admin')->getUrl()`? I don't like the idea of having a route that doesn't do anything to be honest

I wouldn't say it does nothing. It had to be this way so that no existing project will break, I'm sure there are some folks who hard coded the URL of `/` for dashboards.

The alternative is to change dashboard's `$routePath` to be `/dashboard` as shared here #15257

While `Filament::getPanel('admin')->getUrl()` is a way to get the URL, the Laravel convention of `route('a.b.c')` is much easier remember and to look up using artisan, I mean afterall, Filament did name every panel with a `home` route, except that it is now overwritten by Dashboard.

And that is main issue I'm trying to address: the inconsistency.

If dashboard is not used, `home` route exists, but if dashboard was used, `home` does not exists. It took quite a bit of debugging to figure this out, if this PR is merged, it should save others from this trouble.

danharrin:
In Filament, you should never really rely on route names for anything, even pages in a resource. Route names are internally generated and used, and are subject to change. The reason why we have the `getUrl()` helpers on the panel, resources and pages is to allow you not to have to consider the route name. Avoiding hard-coded route names will make major version upgrades easier.

bilogic:
I work alot with packages and tried this out for some time, here are my thoughts

1. Route names are an abstraction that lets app and packages to link to one another without getting into the nuts and bolts
2. Some notable examples of route names would be `login`, `logout`, we don't need to know what is actually handling it
3. Middleware also follow the same concept, i.e. `auth`, `guest` etc, replaceble without massive code changes
4. By using `->getUrl()`, my packages often have to point to a very specific class and incurs alot of namespace changes when there are better replacements
5. About subject to change, I think filament's main aim is to provide an admin panel, i.e. the panel concept is very unlikely to "change"
6. Thus, the `filament.<panel-name>` prefix is probably very stable
7. I think it is a better trade off to make route names consistent for each filament version rather than label them as "do not use"

PR #15288, opened 2025-01-08
fix: call base isHidden method in addition to checking action

### LINKED ISSUE #15287: Can't set hidden state on Action within form
### Package

filament/forms

### Package Version

v3.2.132

### Laravel Version

v11.37.0

### Livewire Version

v3.5.12

### PHP Version

PHP 8.2.25

### Problem description

When building a form that contains an action, I should be able to use the `visibleOn()` method on the action to only show the action on specified operations.

For example with the following code, I would expect the "Generate password" action to only be visible on the 'create' action.

```php
    {
        return $form
            ->schema([
                TextInput::make('password')
                    ->label('Password')
                    ->password()
                    ->required()
                    ->revealable()
                    ->columnSpanFull(),
                
                Action::make('generate_password')
                    ->label('Generate password')
                    ->action(fn (Set $set) => $set('password', Str::random(16)))
                    ->toFormComponent()
                    ->visibleOn(['create']),
            ]);
    }
```

Currently, the "Generate password" action is incorrectly shown on both the "Create" _and_ "Edit" operation.

Calling `visibleOn()`, `visible()` or `hidden()` has no effect on the Action.

### Expected behavior

I would expect the above example to only show the "Generate password" action on the "create" operation, because of the call to `->visibleOn(['create'])`.

### Steps to reproduce

- Create a Form
- Add an `Action` to the form, and call `->visibleOn(['create'])` on it
- See that the Action is still visible on the Edit page.

https://github.com/danjohnson95/filament-issue

### Relevant log output

_No response_

danjohnson95:
## Description

Fixes an issue where an `Action` within a form could not be hidden.

Resolves https://github.com/filamentphp/filament/issues/15287

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/forms/src/Components/Actions/ActionContainer.php
+++ b/packages/forms/src/Components/Actions/ActionContainer.php
@@ -2,10 +2,15 @@
 
 namespace Filament\Forms\Components\Actions;
 
+use Filament\Forms\Components\Concerns;
 use Filament\Forms\Components\Component;
 
 class ActionContainer extends Component
 {
+    use Concerns\CanBeHidden {
+        isHidden as baseIsHidden;
+    }
+
     protected string $view = 'filament-forms::components.actions.action-container';
 
     protected Action $action;
@@ -31,6 +36,6 @@ class ActionContainer extends Component
 
     public function isHidden(): bool
     {
-        return $this->action->isHidden();
+        return $this->baseIsHidden() || $this->action->isHidden();
     }
 }


danharrin:
https://github.com/filamentphp/filament/issues/15287#issuecomment-2579416806

Further: this does seem like a small change to make it work the way you think it should, but it technically does have a performance impact since extra checks are being made each time the actions are rendered or called. Since this is not supported usage, I am inclined not to merge for that reason, sorry.

PR #15290, opened 2025-01-08
Allow multiple disable option conditions

Abdulmajeed-Jamaan:
## Description

I have the following use case: repeater that has `Select` column and `RichEditor` which used to select a locale and sets its content.

I want to use the `disableOptionsWhenSelectedInSiblingRepeaterItems` to make it unique, and add extra condition to disable a specific locale from being selected by user but displayable if exists (ai generated content, which has specific locale identifier).

In other words, i need extra condition on top of `disableOptionsWhenSelectedInSiblingRepeaterItems`.

## Functional changes

Modified the trait `CanDisableOptions` to have similar structure as the `HasExtraAttributes` which allows multiple conditions to be added optionally.

```PHP
->disableOptionsWhenSelectedInSiblingRepeaterItems()
->disableOptionWhen(fn(string $value): bool => str($value)->contains('ai'), true)
```

### DIFF
--- a/packages/forms/src/Components/Concerns/CanDisableOptions.php
+++ b/packages/forms/src/Components/Concerns/CanDisableOptions.php
@@ -3,21 +3,22 @@
 namespace Filament\Forms\Components\Concerns;
 
 use Closure;
+use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
 
 trait CanDisableOptions
 {
     /**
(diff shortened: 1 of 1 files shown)


Abdulmajeed-Jamaan:
Would like to learn if there are other proper ways to solve my use case. Or should i proceed with updating the docs ?

danharrin:
Thanks

PR #15296, opened 2025-01-09
✨ docs: added authorization section in 10-clusters.md

chrisreedio:
## Description

Introduced `canAccess()` method to control cluster visibility. Just a copy and paste with a `s/page/cluster/i` of the authorization section of the Page docs. 

!! I ran `composer cs` and it oddly changed (only) the line endings in 900+ files when I only added around 10 lines to the cluster docs page. Using PHPStorm and not seen this before so not sure why it'd do that. (I ran this after the commit in this PR so those line ending changes discarded)

## Visual changes

image

## Functional changes

### DIFF
--- a/packages/panels/docs/10-clusters.md
+++ b/packages/panels/docs/10-clusters.md
@@ -62,6 +62,17 @@ use App\Filament\Clusters\Settings;
 protected static ?string $cluster = Settings::class;
 ```
 
+## Authorization
+
+You can prevent clusters from appearing in the menu by overriding the `canAccess()` method in your Cluster class. This is useful if you want to control which users can see the cluster in the navigation, and also which users can visit the cluster directly:
+
+```php
+public static function canAccess(): bool
+{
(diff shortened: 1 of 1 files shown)


danharrin:
There is a bit of a risk here that users might take this as actually authorising access to all pages in the cluster. This is just controlling navigation and authorisation on the cluster redirect at the moment, rather than preventing the user from viewing pages in the cluster. For that reason, I would prefer this was not documented to avoid that potentially serious confusion from a security POV.

Like other pages, I would suggest that clusters use the `shouldRegisterNavigation()` method to control navigation registration. This method name has much less responsible connotations. You could use it like so:

```php
public static function shouldRegisterNavigation(): bool
{
    if (! parent::shouldRegisterNavigation()) {
        return false; // The parent check returns false if all pages in the cluster are unauthorized.
    }

    return auth()->user()->canManageSettings();
}
```

PR #15302, opened 2025-01-09
Refactor: `assertTableFilterExists` to allow passing in a callback

chester-sykes:
## Description

I've refactored the `assertTableFilterExists` test function to allow you to pass a callback function, so that you can access the filter itself like you can do with the `assertTableColumnExists`.

I couldn't see `assertTableFilterExists` in the docs so I've added it in as well as docs on the new functionality.

I also couldn't see the stub anywhere so I've added that as well.

I think this feature will be useful if you want to test a specific part of your filter, for example making sure that the label is right.

## Visual changes

No visual changes.

## Functional changes

### DIFF
--- a/packages/tables/docs/12-testing.md
+++ b/packages/tables/docs/12-testing.md
@@ -349,6 +349,33 @@ it('can remove all table filters', function () {
 });
 ```
 
+### Existence
+
+To ensure that a filter exists, you can use the `assertTableFilterExists()` method:
+
+```php
+use function Pest\Livewire\livewire;
+
+it('has an author filter', function () {
+    livewire(PostResource\Pages\ListPosts::class)
+        ->assertTableFilterExists('author');
+});
+```
+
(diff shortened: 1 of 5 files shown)


packages/tables/docs/12-testing.md
```diff
@@ -349,6 +349,33 @@ it('can remove all table filters', function () {
...
+You may pass a function as an additional argument in order to assert that a filter passes a given "truth test". This is useful for asserting that a filter has a specific configuration:
+
+```php
+use function Pest\Livewire\livewire;
+use Filament\Tables\Filters\SelectFilter;
+
+it('has an author filter', function () {    
+    livewire(PostResource\Pages\ListPosts::class)
+        ->assertTableFilterExists('author', function (SelectFilter $column): bool {
+            return $column->getLabel() === 'Select Author';
```
danharrin:
```suggestion
            return $column->getLabel() === 'Select author';
```

danharrin:
Good spot, thanks

Issue #15308, opened 2025-01-09
Multiple file upload race condition sorting


angus-mcritchie:
### Package

filament/filament

### Package Version

v3.2.132

### Laravel Version

v11.37.0

### Livewire Version

v3.5.12

### PHP Version

8.3.15

### Problem description

Upload 2 or more files, the sorting is based on which one uploads first rather than the visible sort order of the uploader.

This happens because [FilePond core options](https://pqina.nl/filepond/docs/api/instance/properties/#core-properties) default value of `maxParallelUploads` is `2`.

https://github.com/user-attachments/assets/fcf67f9c-3752-4511-9c1f-c9e5ceff9dbd

A simple work around would be allowing to change the `maxParallelUploads` to `1` and everything would work.

But ideally, we should allow `maxParallelUploads` to change to more than `1` and still resoect the upload order regardless which file uploads first.

### Expected behavior

Select files to upload, order will show `1.svg` first, then `2.svg` and after upload for them to remain in the same order regardless which one is uploaded first.

### Steps to reproduce

clone the [reproduction repository](https://github.com/angus-mcritchie/filament-image-race-sorting)
run `php artisan migrate`
run `php artisan db:seed`
Login and go to the http://localhost/admin/users/1/edit
Upload 2 or more files (with different file sizes helps reproduce more easily) and see the order change based on which on is uploaded first

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/angus-mcritchie/filament-image-race-sorting

danharrin:
Thanks for the PR, another setting you can try that affects ordering is `appendFiles()`

angus-mcritchie:
Thanks, @danharrin. I have tried the `appendFiles()` method, but this bug still happens regardless of whether you append or prepend the images. I am confident it's a race condition.

It's especially noticeable when you upload many (like 10 or more) images, and in one case, I had 5 images out of order.

Thanks for merging https://github.com/filamentphp/filament/pull/15309. Should we keep this open as this is still a bug? The PR adds a workaround.

If I were to make a PR to fix this properly, would you prefer it to be done on v3 or v4?

danharrin:
Sure, I would accept a PR in v3 since this is a bug

gcorp-llc:
hi I'm run composer update to project and show error Undefined variable $getMaxParallelUploads for admin panel 
[...]

PR #15309, opened 2025-01-09
feat/bug-fix: add maxParallelUploads option to FileUpload component

angus-mcritchie:
## Description
This pull request adds a `maxParallelUploads()` method to control how many parallel uploads happen at once.
Although this is a useful method by itself, it also allows a simple workaround for https://github.com/filamentphp/filament/issues/15308

## Visual changes
There are no visible changes except for the [documentation](https://filamentphp.com/docs/3.x/forms/fields/file-upload#uploading-multiple-files), which is highlighted in red.

image

## Functional changes

- [ x] Code style has been fixed by running the `composer cs` command.
- [ x] Changes have been tested to not break existing functionality.
- [ x] Documentation is up-to-date.


packages/forms/docs/03-fields/09-file-upload.md
```diff
@@ -70,6 +70,19 @@ class Message extends Model
 }
 ```
 
+### Control the maximum parallel uploads
```
danharrin:
```suggestion
### Controlling the maximum parallel uploads

```

danharrin (changes requested):
The code is perfect, thank you very much for adhering to the style of the rest of the codebase

Tigla:
After these changes I get an error:
```
Undefined variable $getMaxParallelUploads (View: /vendor/filament/forms/resources/views/components/file-upload.blade.php)
```
A field without multiple uploads.
What could this be related to?

danharrin:
Could you create a reproduction repository please? I can't imagine why that would be the case.

Tigla:
Sorry, the reason is not in these changes, but in another package that substitutes the source classes, as a result of which this error occurs: https://github.com/joshembling/image-optimizer/issues/38

DevMushref:
All my forms that has file uploads got this error: `Undefined variable $getMaxParallelUploads`

xkakjd88178:
I’m having this error too.

danharrin:
Could you create a reproduction repository please? I can't imagine why that would be the case.

bytemewebdev:
I ran composer update, then I got this error

devSviat:
After the update, I encountered the error "Undefined variable $getMaxParallelUploads" in all forms that involve file uploads, including the Import action.

**Update:**  The issue was resolved by removing the "joshembling/image-optimizer" package.

danharrin:
It does seem like this plugin is causing the issues. It is using the Filament file upload view, but not extending the `BaseFileUpload`, so any time we add new methods to the view it will break. The plugin should probably extend the class, or copy the view into the plugin so that it does not matter.

bytemewebdev:
Anyone got a workaround?

DevMushref:
> Anyone got a workaround?

Remove `joshembling/image-optimizer` from your `composer.json` and run `composer update`

PR #15319, opened 2025-01-11
Fix: Add expanded close tag option in datalist (bug in Chrome)

borjajimnz:
## Description

This fixes a bug that occurs when using a datalist in Chrome, which throws the following error:

"filamentphp uncaught typeerror: cannot read properties of null (reading 'before')"

demo: https://youtu.be/CNnH1AI8tbk

## Functional changes


danharrin:
Thanks

PR #15337, opened 2025-01-13
Fix a few inconsistencies with v4 docs

Omranic:
Just fixing a few inconsistencies in the new v4 documentation.


danharrin:
Hi, thanks but these docs are going to be completely rewritten at some point before release, and some of these suggestions are incorrect so I am going to take it from here. Thank you very much though

PR #15340, opened 2025-01-14
Prevent stacked, editable table columns to trigger wrapping a.href 

tanthammar:
## Description
Fixes: If you have your table columns stacked, and click on an editable column the outer `a.href` will be triggered and the record will open instead of editing the value of the column.

## Visual changes
None

## Functional changes
-  The editable column becomes editable instead of only triggering the redirect.

No, tried multiple times, still spaces are added that were not there before, sorry...

yes, 
- Checkbox:  `click.stop.prevent` was needed on the wrapping div. Else `a.href` is triggered if `Alpine.isLoading` is `true`
- Text input, Toogle and Select columns was enough with adding `.prevent `on the existing `x-on:click.stop`

-[X] Documentation is up-to-date.

### DIFF
--- a/packages/tables/resources/views/columns/checkbox-column.blade.php
+++ b/packages/tables/resources/views/columns/checkbox-column.blade.php
@@ -4,7 +4,7 @@
 @endphp
 
 <div
-        x-data="{
+    x-data="{
         error: undefined,
 
         isLoading: false,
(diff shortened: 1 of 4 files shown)


danharrin:
Thanks

jvitasek:
Hey @tanthammar, I have a feeling this caused a problem documented here: https://github.com/filamentphp/filament/issues/15895

Mind taking a look?

tanthammar:
> Hey @tanthammar, I have a feeling this caused a problem documented here: https://github.com/filamentphp/filament/issues/15895
> 
> 
> 
> Mind taking a look?

Yes of course!

danharrin:
#16098

tanthammar:
@jvitasek Fixed it today, thank you for reporting and sorry to keep you waiting.

PR #15345, opened 2025-01-14
Fi: Sub Navigation Icon Fallback

davidwhicker:
When using sub-navigation, you can usually omit a navigation icon, and a dot with a line will be displayed. However, when combining Pages with a ManageRelatedRecords page, omitting the icon from ManageRelatedRecords defaults to a rectangular icon, disrupting the consistent appearance. For clarity it seems that ManageRelatedRecords pages in the Sub Navigation falls back to the rectangle icon. 

Before:
Screenshot 2025-01-14 at 21 43 54

After:
Screenshot 2025-01-14 at 21 43 40


danharrin:
This is consistent with the edit/view pages which also have icons by default, you can override the method to return null if you want to remove the icon.

Issue #13416, opened 2024-06-27
KeyValue UI not updating while using $set()


rickzim:
### Package

filament/filament

### Package Version

v3.2.115

### Laravel Version

v11.25.0

### Livewire Version

v3.5.6

### PHP Version

v8.3.10

### Problem description

When using `$set()` on a `Forms\Components\KeyValue` field, the UI will not update properly to the newly set values when changed for the first time. However, internally the value is changed.

![2024-06-27 14_13_40-Edit Post - Laravel](https://github.com/filamentphp/filament/assets/72699017/3f0b4643-1b8e-4a75-8e93-bc643a1d775d)

Simular Issues:
https://github.com/filamentphp/filament/issues/9120
https://github.com/filamentphp/filament/issues/2939

### Expected behavior

When i change the `KeyValue` values, i expect the internal value to be the same as the UI.

### Steps to reproduce

1. Navigate to `Posts`
2. Select an item
3. Click on of the buttons to change the value of the `Meta` field.
4. Refresh if not changed.

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/rickzim/demo-filament-key-value

borjajimnz:
@rickzim thanks for the way you put the things to test it in your repository, it seems to be fixed already.

https://www.youtube.com/watch?v=w7AL-wxNhSg

cc @danharrin

danharrin:
Thanks

rickzim:
@borjajimnz @danharrin Interesting, i've just download the demo and updated all the packages, but for me its still not working somehow...

Laravel: 11.37.0
PHP: 8.4
Filament: 3.2.133
Livewire: 3.5.12

danharrin:
@rickzim can you submit a video with the reproduction steps and how they differ from the one sent by @borjajimnz ?

rickzim:
@danharrin I hope this is good enough
![recording](https://github.com/user-attachments/assets/d2fca01a-5dac-49dc-94d5-f23f79876c63)

The first time when i use set(), it will update the values but wont update the UI. After that, it works perfectly fine.

rickzim:
@borjajimnz 
It seems that on an empty array and/or null it works for me aswell. 
The problem is that when the field already contains data, it won't update the UI.

borjajimnz:
Finally I could reproduce this yesterday. It seems to be something related with a tweak to prevent loop updates.

Will give a try when I have some free time.

borjajimnz:
[...]

PR #15346, opened 2025-01-14
Fix: key-value form component initial rows/state synchronization

borjajimnz:
## Description

As described in https://github.com/filamentphp/filament/issues/13416, there was an issue with the frontend state not reflecting the actual state. This occurred because, during the update process, shouldUpdateRows was initially set to false, preventing the update.

After modifying this behavior, the issue has been resolved, as demonstrated in the following example.

## Visual changes

https://youtu.be/lKEOCOkdDos

## Functional changes


danharrin:
Thanks

Issue #14990, opened 2024-12-03
->emptyStateDescription() does not work as expected when using configureUsing() in a provider.


RalphMRivera:
### Package

filament/filament

### Package Version

3.2.127

### Laravel Version

11.34.2

### Livewire Version

v3

### PHP Version

8.2.15

### Problem description

When configuring the default behavior for a table in a service provider, the empty state heading can be defined, but the empty state description can not.

### Expected behavior

The defaults for both the empty state heading and description should appear on a list page that has no records.

### Steps to reproduce

Step 1 - Add the following to the provider file:

```php
Table::configureUsing(function (Table $table): void {
    $table
        ->emptyStateDescription('This is my custom description!')
        ->emptyStateHeading('This is my custom heading!');
});
```

Step 2 - Go to an empty list page.

Note: The description does work if manually added to the list page.

```php
class CustomerListPage extends ListRecords
{
    public function table(Table $table): Table
    {
        return $table->emptyStateDescription('This is my custom description!');
    }
}
```

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/RalphMRivera/filament-issue

RalphMRivera:
The relevant lines are here:

https://github.com/RalphMRivera/filament-issue/blob/a52e512106900d8873a98481ae7db8b354fac4b2/app/Providers/AppServiceProvider.php#L23-L27

danharrin:
In v4 I have changed the feature so that any parent defined `configure()` statements are called before `setUp()`, and any child ones are called after `setUp()` - c91a51d9bd0e62f053e05b9e90ee8600e5820345

This will hopefully prevent similar issues in the future

RalphMRivera:
That's great to hear. Thanks. What does this mean for v3? Do you think this will get corrected there as well?

danharrin:
No, this will not be unfortunately.

PR #15348, opened 2025-01-15
Fix: configure using emptyStateDescription

borjajimnz:
## Description

As described in https://github.com/filamentphp/filament/issues/14990, I'm not sure if this is the expected behavior. At least for me, it seems to be. @danharrin 

This is the current priority order:

1. Resource
2. Configure using
3. If nothing is set, but the create action is there, it will display the "Create a %% to get started." text.

## Visual changes

https://youtu.be/1gzuXzNHWDs

## Functional changes

### DIFF
--- a/packages/tables/src/Table.php
+++ b/packages/tables/src/Table.php
@@ -85,9 +85,15 @@ class Table extends ViewComponent
     {
         parent::setUp();
 
-        $this->emptyStateDescription(function (Table $table): ?string {
+        $configuredDescription = $this->getEmptyStateDescription() ?? null;
+
+        $this->emptyStateDescription(function (Table $table) use ($configuredDescription): ?string {
             if (! $table->hasAction('create')) {
-                return null;
+                return $configuredDescription ?? null;
+            }
+
+            if ($configuredDescription) {
+                return $configuredDescription;
             }
 
             return __('filament-tables::table.empty.description', [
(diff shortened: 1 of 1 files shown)


CodeWithDennis:
Came here to look for exactly this issue.

```PHP
 Table::configureUsing(function (Table $table): void {
     $table->emptyStateDescription(__('It looks like we couldn\'t find any records'));
 });
 ```
 
 Is this gonna be merged in the current version or waiting for V4?

**Comment by danharrin** [comment]
I'm not sure yet, I don't think the implementation is right here but I've not had time to work out a proper alternative

**Comment by CodeWithDennis** [comment]
This idea is "somewhat" related but doesn't fix the issue itself.

Wouldn't it make sense to replace the following with a default description as well, since that's what `getEmptyStateHeading` also does?

```php
public function getEmptyStateDescription(): string | Htmlable | null
{
    return $this->evaluate($this->emptyStateDescription);
}
```

Replace with:

```php
public function getEmptyStateDescription(): string | Htmlable | null
{
    return $this->evaluate($this->emptyStateDescription) ?? __('filament-tables::table.empty.description', [
        'model' => $this->getModelLabel(),
    ]);
}
```

danharrin:
Would break being able to remove the description with `emptyStateDescription(null)`

danharrin:
In v4 I have changed the feature so that any parent defined `configure()` statements are called before `setUp()`, and any child ones are called after `setUp()` - c91a51d9bd0e62f053e05b9e90ee8600e5820345

This will hopefully prevent similar issues in the future

PR #15350, opened 2025-01-15
Fix: Respect nullified record action and url in tables

zsnakeee:
This PR fixes an issue where setting `recordAction` or `recordUrl` to `null` in a table configuration would be overridden by the default values.

## Changes

1. Added tracking of whether `recordAction` and `recordUrl` have been explicitly set:
   - Added `hasRecordActionBeenSet` and `hasRecordUrlBeenSet` flags
   - Added getter methods to check these flags

2. Modified table builders to respect user configuration:
   - Only apply default values when the user hasn't set these values explicitly
   - This allows nullification to work properly

3. Added comprehensive tests:
   - Tests for basic table configuration
   - Tests for relation manager table configuration
   - Tests for both nullification and custom values
   - Tests for closure-based values

## Testing

Added a new test file `tests/Tables/TableConfigurationTest.php` that verifies:
- Record actions and URLs can be nullified
- Both settings can be set together
- Both settings can be nullified together
- Closure values are properly evaluated
- RelationManager respects all these settings

## Breaking Changes

None. This change fixes a bug where user configuration was being ignored, restoring the expected behavior.

## Security Impact

None. This is a purely functional change that does not affect security.

### DIFF
--- a/packages/panels/src/Resources/Concerns/InteractsWithRelationshipTable.php
+++ b/packages/panels/src/Resources/Concerns/InteractsWithRelationshipTable.php
@@ -54,11 +54,14 @@ trait InteractsWithRelationshipTable
 
     protected function makeTable(): Table
     {
-        return $this->makeBaseTable()
+        $table = $this->makeBaseTable()
             ->relationship(fn (): Relation | Builder => $this->getRelationship())
             ->modifyQueryUsing($this->modifyQueryWithActiveTab(...))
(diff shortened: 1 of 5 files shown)


danharrin:
Can you please explain why `configureUsing()` would be used for these methods? I do not think I agree that they should overwrite the relation manager and resource configurations for the table, and it would be a breaking change to do this. If a user was using the table builder outside the panel and wanted to configure this without affecting the panel, it would break their panel.

PR #15354, opened 2025-01-15
Fix: Prevent column groups table row from appearing if all columns in every group are (toggled) hidden

ralphjsmit:
This PR fixes an issue where this empty bar would appear if you have a `ColumnGroup`, but all columns inside the group are either `->hidden()` (e.g. via a closure) or toggled hidden by the toggleable setting.

CleanShot 2025-01-15 at 18 10 27@2x

I am not sure if it's for performance reasons, but I would say that it should be possible e.g. in V4 to refactor away from this manual setting of a boolean when pushing a new `ColumnGroup`.

Thanks!

### DIFF
--- a/packages/tables/src/Table/Concerns/HasColumns.php
+++ b/packages/tables/src/Table/Concerns/HasColumns.php
@@ -154,7 +154,7 @@ trait HasColumns
             }
         }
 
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks

PR #15355, opened 2025-01-16
Feat: Add `unmountTableBulkAction()` testing method

ralphjsmit:
I noticed this `unmountTableBulkAction()` method was missing still.

Thanks!


danharrin:
This would be gone in v4 so I am not going to merge it, you can still call `->call('unmountTableBulkAction')` in your app if you want to though.

PR #15362, opened 2025-01-17
Fix: Added Google Cloud Storage support for Import Records.

Jessedev1:
## Description
When using an external file system like Google Cloud Storage, a stream must be opened to access the file; otherwise, it cannot be read. While the implementation for S3 already existed, support for GCS was missing. Therefore, I implemented it.

## Visual changes
This is not applicable.

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/CanImportRecords.php
+++ b/packages/actions/src/Concerns/CanImportRecords.php
@@ -3,45 +3,46 @@
 namespace Filament\Actions\Concerns;
 
 use Closure;
-use Filament\Actions\Action;
-use Filament\Actions\ImportAction;
-use Filament\Actions\Imports\Events\ImportCompleted;
-use Filament\Actions\Imports\Events\ImportStarted;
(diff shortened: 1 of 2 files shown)


danharrin:
As a rule, we do not support GCS anywhere else in Filament. By merging this PR, it feels like I am then committing to providing full support for it elsewhere, which makes me very hesitant. I have no experience with it and it seems to differ from S3 quite a lot. Since I have not had this request before, I am willing to bet there are a relatively low number of users who are using GCS with Filament.

As such, I am not going to merge this. Maybe if GCS gains more popularity in the future, I will choose to introduce official support everywhere. For now, you could extend the `ImportAction` in your own application to add GCS support, or even release a package which does so. Sorry. Please check out our [contributing guide](https://filamentphp.com/docs/3.x/support/contributing#development-of-new-features) which talks about proposing the feature before submitting the PR, to ensure you aren't spending time building something that we won't merge.

PR #15376, opened 2025-01-19
Fix str_getcsv(): the $escape parameter must be provided as its default value will change after updating to PHP 8.4

ziming:
## Description

After updating to PHP 8.4, I keep getting the following message in Sentry (hundreds of them)

"Fix str_getcsv(): the $escape parameter must be provided as its default value will change"

"str_getcsv(): the $escape parameter must be provided as its default value will change"

CleanShot 2025-01-19 at 10 32 05@2x

So after some googling around, I found this drupal discussion about it and provided a default value

https://www.drupal.org/project/drupal/issues/3477324

The default value I provide is the 1 in PHP.net (which is the current default, but I guess the default could change in the future hence that warning)

https://www.php.net/manual/en/function.str-getcsv.php

CleanShot 2025-01-19 at 10 34 41@2x

CleanShot 2025-01-19 at 10 54 40@2x

The official php documentation recommends setting $escape to empty string instead of the default `"\\"` but I did not set to empty string to preserve existing behaviour (`"\\"`) out of caution. It is possible that empty string is the better choice & won't break things but I do not know enough to make that judgement

CleanShot 2025-01-19 at 10 55 19@2x

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/support/src/Commands/Concerns/CanReadModelSchemas.php
+++ b/packages/support/src/Commands/Concerns/CanReadModelSchemas.php
@@ -132,7 +132,7 @@ trait CanReadModelSchemas
         };
 
         $values = str_contains($column['type'], '(')
-            ? str_getcsv(Str::between($column['type'], '(', ')'), ',', "'")
+            ? str_getcsv(Str::between($column['type'], '(', ')'), enclosure: "'", escape: '\\')
             : null;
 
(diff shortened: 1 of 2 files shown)


danharrin:
I think keeping the escape is fine, thank you

PR #15387, opened 2025-01-21
Get user by the panel guard.

Abdelhammied:
Add Getting user by the panel guard for different guards and assert the filament function exists for panel usage

## Description
Add Getting user by the panel guard for different guards and assert the filament function exists for panel usage.
Because the code was written to get the default guard only, the added code can check if filament exists so if the import and export actions have been used alone or with filament both cases will be handled

## Visual changes

image
image

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/CanExportRecords.php
+++ b/packages/actions/src/Concerns/CanExportRecords.php
@@ -146,6 +146,10 @@ trait CanExportRecords
 
             $user = auth()->user();
 
+            if (function_exists('filament')) {
+                $user = auth(filament()->getAuthGuard())->user();
+            }
+
             if ($action->hasColumnMapping()) {
                 $columnMap = collect($data['columnMap'])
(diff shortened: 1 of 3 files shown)


danharrin:
In version 4 there is customisation available for the auth guard, but I think this PR would introduce some unintended behaviour. Outside of a panel route, `filament()` will return the default panel, so if the user is using imports/exports outside of panels they would then have to be logged into the same guard as the default panel, otherwise it would not work. Thank you for the suggestion though.

Abdelhammied:
So how can we use a different guard for the export and import actions, in panels, 
Can be done through a different approach without using the filament function, should it be tried?

danharrin:
How can it be done without using the Filament function, and how would that improve what I described?

Abdelhammied:
Can you please check this new pr ? 

https://github.com/filamentphp/filament/pull/15404

PR #15392, opened 2025-01-21
Add option to force horizonal steps header in form wizard

ziming:
## Description

Currently on mobile devices, form wizard steps header is almost always vertically stacked even if I hide the all step header labels with ->hiddenLabel(). This makes the form wizard look unpleasant as can be seen below

CleanShot 2025-01-21 at 21 14 34

What my PR does is you can now do this:

```php
Wizard::make($steps)
    ->forceHorizontalStepsHeader();
```

And the horizontal steps header will be forced. Looking like this

CleanShot 2025-01-21 at 21 20 12

The horizontal scroll bar if not all can fit into the screen

CleanShot 2025-01-21 at 21 20 45

Hence making it better for those of us who didn't like the vertically stacked steps header look and want to force horizontal at all times even on mobile

### DIFF
--- a/packages/forms/resources/views/components/wizard.blade.php
+++ b/packages/forms/resources/views/components/wizard.blade.php
@@ -134,6 +134,7 @@
         role="list"
         @class([
             'fi-fo-wizard-header grid divide-y divide-gray-200 dark:divide-white/5 md:grid-flow-col md:divide-y-0 md:overflow-x-auto',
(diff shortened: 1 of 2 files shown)


danharrin:
Hi, I would suggest using CSS to add these classes to `.fi-fo-wizard-header`. Then we do not need to introduce a new PHP method for a styling preference. Many thanks.

PR #15399, opened 2025-01-22
Update all Armenian language files

arshaviras:
## Description

Added Armenian (hy) language translations to improve localization support for Filament. This update provides better accessibility for Armenian-speaking users.


danharrin:
Thanks

PR #15404, opened 2025-01-22
Set Custom Auth Guard For The Export And The Import Classes.

Abdelhammied:
Add Getting user by the panel guard for different guards and assert the filament function exists for panel usage

## Description
Add Getting user by the panel guard for different guards and assert the filament function exists for panel usage.
Because the code was written to get the default guard only, the added code can check if filament exists so if the import and export actions have been used alone or with filament both cases will be handled

## Visual changes

image
image

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/CanExportRecords.php
+++ b/packages/actions/src/Concerns/CanExportRecords.php
@@ -144,7 +144,7 @@ trait CanExportRecords
                 return;
             }
 
-            $user = auth()->user();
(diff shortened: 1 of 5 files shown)


danharrin:
As I described in the other PR, a similar feature has already been built for v4 but it's not the same API is this. I am not going to merge an API like this into v3 that will then just be deprecated as it creates maintenance work for people.

In this situation, I would suggest you extend `ImportAction` and `ExportAction` in your own app and override the methods you need to customize the guard.

PR #15293, opened 2025-01-09
Sushi Compatible Summarizers

eele94:
## Description

I couldn't use the summarizers when I used Sushi models.

## Functional changes

- [x] Code style has been fixed by running the `composer cs` command.
- [x] Changes have been tested to not break existing functionality.
- [x] Documentation is up-to-date.

PR #15407, opened 2025-01-22
Fix summarizer connection resolver on non-default connections

saulens22:
## Description

Laravel model method `resolveConnection`, when called without arguments, returns default database connection, even if model has another connection set. 

Usually it doesn't matter if syntax is the same between connections but in our case it was returning `MariaDbConnection` instead of `SqlServerConnection`, which has different char escaping.

Laravel itself manually passes connection name to `resolveConnection` like here: https://github.com/laravel/framework/blob/6e0e2cbd584a2988348409fe15bb2ead87aeafec/src/Illuminate/Database/Eloquent/Model.php#L1817

I believe it is better to use `getConnection` as it sets connection properly as in example above.

## Functional changes

### DIFF
--- a/packages/tables/src/Columns/Summarizers/Summarizer.php
+++ b/packages/tables/src/Columns/Summarizers/Summarizer.php
@@ -126,7 +126,7 @@ class Summarizer extends ViewComponent
 
         $asName = (string) str($query->getModel()->getTable())->afterLast('.');
 
-        $query = $query->getModel()->getConnection()
+        $query = $query->getModel()->resolveConnection($query->getModel()->getConnectionName())
             ->table($query->toBase(), $asName);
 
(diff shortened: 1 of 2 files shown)


danharrin:
Hey @eele94, in #15293 you ensured that summarisers were compatible with Sushi models. Does this PR break the compatibility again?

saulens22:
I checked Sushi model file, and it doesn't have `getConnection` function, so it breaks.

I have changed the lines to `$query->getModel()->resolveConnection($query->getModel()->getConnectionName())` as these methods exist on both Laravel and Sushi

eele94:
@danharrin thanks for mentioning. I believe @saulens22 has found the solution😃

danharrin:
Thanks

PR #15418, opened 2025-01-23
docs: update installation docs to mention testing

rrmesquita:
## Description

When testing a Laravel app, the application environment automatically switches to `testing`, so if `canAccessPanel` is not implemented, tests will fail (as they should) but the developer might not know why instantly.

Feel free to keep it shorter if you think this change don't add much value.

## Visual changes

None

## Functional changes


danharrin:
Good spot, thanks

Issue #15416, opened 2025-01-23
`Add row` on KeyValue form field not working when contains data


rickzim:
### Package

filament/filament

### Package Version

v3.2.134

### Laravel Version

v11.39.1

### Livewire Version

v3.5.12

### PHP Version

PHP 8.4.3

### Problem description

When using Forms\Components\KeyValue field to update a record, you're no longer able to add new rows.

Introduced in:
Filament v3.2.134

Might be related to:
https://github.com/filamentphp/filament/pull/15346

Continuation of: 
https://github.com/filamentphp/filament/issues/13416

### Expected behavior

Rows being added to the KeyValue field.

### Steps to reproduce

1.  Navigate to `Posts`
2. Select an item
3. Click on `Add row`

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/rickzim/demo-filament-key-value

### Relevant log output

```shell

```

borjajimnz:
Thanks @rickzim this one is the good one. regards.

rickzim:
@borjajimnz I should be the one thanking you! awesome work dude :) and thanks for everything!

danharrin:
Reverted the PR, reopening the other

PR #15420, opened 2025-01-23
Fix that resolves the issue when trying to add a new key-value pair.

borjajimnz:
## Description

Fix that resolves the issue when trying to add a new key-value pair described in https://github.com/filamentphp/filament/issues/15416.
This was my fault in the last fix. now is solved. soz!

## Visual changes

https://www.youtube.com/watch?v=WP4hgMzWZ4A

## Functional changes


rickzim:
Hey! i used the updated code in the demo project but now theres another weird issue (maybe its just me but it doesn't hurt to test it)

if you add new rows to a KeyValue field, and populate it, it will start doing weird things like removing other rows.

It is the easiest to trigger if you add multiple empty lines at ones, and then try to populate them.

danharrin:
Adding rows definitely needs to update the state otherwise it will be out of sync

danharrin:
I am going to revert the old PR and make a new release

danharrin:
Going to close this, but please reopen when the above issue is addressed

PR #15427, opened 2025-01-24
Update dark mode background color.

lomotech:
## Description

Update default dark mode background color for input and datepicker select button

## Visual changes

Correcting background color for dark mode to be darker (from white to bg-gray-800)

before 

Screenshot 2025-01-24 at 7 53 05 PM

after 

Screenshot 2025-01-24 at 7 53 16 PM

## Functional changes

### DIFF
--- a/packages/forms/resources/views/components/date-time-picker.blade.php
+++ b/packages/forms/resources/views/components/date-time-picker.blade.php
@@ -168,7 +168,7 @@
                                     type="number"
                                     inputmode="numeric"
                                     x-model.debounce="focusedYear"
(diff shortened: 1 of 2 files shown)


danharrin:
Hmm, what is overriding the `bg-transparent`?

danharrin:
I have not seen this behaviour before

lomotech:
so there is some suspect that overridding the bg-transparent?  maybe that is the clue. i will check it out. im using firefox dev btw + filament tree plugin. tq.

lomotech:
update: ok correct. there is something in that plugin overwrite the bg-transparent..

Issue #14999, opened 2024-12-04
Toggle Buttons not working if there are more than one Relation Manager


arshaviras:
### Package

filament/filament

### Package Version

3.2.123

### Laravel Version

11

### Livewire Version

3

### PHP Version

8.3

### Problem description

Toggle Buttons not working if there are more than one Relation Manager, need to double click, issue is available in official Filament Demo page.

### Expected behavior

It need to to click once not double click

### Steps to reproduce

https://demo.filamentphp.com/shop/customers/1/edit?activeRelationManager=1
go to this link and click Edit, on modal form click any of Toggle Button and you can expect an issue.

Here is the video to show where to reproduce an issue
https://stream.new/v/f2plrRWQJINhI7CJCkvsTxZOd01KUNa9YuX011ymRsr00E

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/arshaviras/demo

bzy107:
same #13661

arshaviras:
And still not resolved :((

arshaviras:
Very interesting why this bug not resolving such a long time?

danharrin:
Closed by #15430

Issue #14913, opened 2024-11-28
Content in a modal's first field is highlighted when modals open


binaryfire:
### Package

filament/filament

### Package Version

3.2.125

### Laravel Version

11.34.2

### Livewire Version

3.5.12

### PHP Version

8.3

### Problem description

When a modal is opened, the first field's  contents are highighted/selected (as if the user has clicked into the field and and pressed Ctrl + A to select all).

Also, the first field is autofocused by default (even if `->autofocus()` isn't used).

![image](https://github.com/user-attachments/assets/af076cf0-4e71-485c-bacc-a0a2430da53d)

### Expected behavior

- Field contents shouldn't be highlighted. Otherwise as soon as a user types something the field's contents are overwritten, which is a major UX issue.

- Autofocusing shouldn't happen by default. It makes it more likely a user will unintentionally change the value of a field with an accidental keystroke. It should only happen if `->autofocus()` is used. 

### Steps to reproduce

1. Head to https://demo.filamentphp.com/shop/orders/1
2. Scroll down to the first Payment and click the "Edit" table action
3. The Reference field will be autofocused and its contents will be highlighted

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/filamentphp/demo

danharrin:
This is Alpine.js `x-trap` behaviour for accessibility: https://alpinejs.dev/plugins/focus#x-trap

binaryfire:
I checked out a few articles and ok re: autofocus, but are you sure auto-selecting everything is correct for accessibility? I didn't see that behaviour in modal examples on any of the accessibility sites.

Check out the example here: https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/

If you fill out some fields and tab through them, only the focus changes. The field contents aren't ever selected. The only place I've seen "select-all-on-tab" is the Alpine link you shared and it seems a lot less accessible - users have to press an arrow key to deselect the content if they want to edit it. Otherwise the field is cleared.

danharrin:
I agree that it isn't convenient for users that don't require the feature, but again the select-all thing is also Alpine and not Filament

https://github.com/user-attachments/assets/1569f1bb-ba4d-4e04-884f-7cee075974c7

So please raise the issue over there and see the reasoning

binaryfire:
Ok will do. But IMHO it'd be better to not use `x-trap` until it's fixed because I think it's breaking accessibility much more than it's helping it.

[...]

Issue #13661, opened 2024-07-22
Modal Behavior Issue with Multiple Relations


bzy107:
### Package

filament/filament

### Package Version

v3.2.95

### Laravel Version

v11.16.0

### Livewire Version

v3.5.4

### PHP Version

v8.2.11

### Problem description

When creating or editing relations in the edit screen with multiple relations, the modal does not close when clicking outside of it while focusing on a TextInput or Textarea. 
The focus remains on the TextInput or Textarea. 
This issue does not occur when there is only one relation.

click outside of modal:
![スクリーンショット 2024-07-22 19 03 32](https://github.com/user-attachments/assets/409f0188-3298-4f2d-954c-39343178e5f2)

You can verify this on the following demo pages:
Single relation: https://demo.filamentphp.com/shop/orders/8/edit
Multiple relations: https://demo.filamentphp.com/shop/customers/5/edit?activeRelationManager=1
I will investigate this issue if time permits, but for now, I am submitting this bug report.

### Expected behavior

The modal should close when clicking outside of it.

### Steps to reproduce

Login with credentials: test@test.com/testtest

1. Access /admin/patient-seconds/1/edit
2. Click the "new treatment" button
3. Focus on the textarea or input field
4. Click outside the modal

This will reproduce the issue where the modal does not close.

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/bzy107/filament-select-bug

hussamitani:
What browser are you using? On what OS are you working? I tried it on Chrome and Firefox (Windows).  I had no issues. Modal disappeared after clicking outside

bzy107:
I tested it on both Chrome and Firefox, and the result was the same. (In Safari, the functionality of the modal itself was broken.)


https://github.com/user-attachments/assets/840957b6-92a8-40f6-a2e1-b0ea6ec70396

danharrin:
Closed by #15430

PR #15430, opened 2025-01-24
Fix form field focusing on modal actions in relation managers

### LINKED ISSUE #15289: clicking in a modal with focused element out of the screen makes the modal jump to said element
### Package

filament/filament

### Package Version

v3.2.132

### Laravel Version

v11.5.0

### Livewire Version

v3.5.12

### PHP Version

PHP 8.3.13

### Problem description

On a Modal, if the focused element is not currently on the screen (scrolled out of view) clicking in the modal outside a element triggers a jump to the focused element. 
The same behaviour happens when clicking in a `Toggle Button` , wich makes selecting an option of a toggle impossible if the focused element is out of view.

image
here i have selected the Text Input 

image
Having scrolled the Focused TextInput out of view, I will not be able to click in this ToggleButton without jumping to the TextInput

chrome-capture-2025-1-8

### Expected behavior

The expected behaviour would be either :
- Selecting a option would focus the toggle button, not allowing the modal to jump anywhere.
- making it a option to disable this behaviour of scrollingIntoView when the focused element is not currently seen on screen

### Steps to reproduce

- Go to https://demo.filamentphp.com/shop/products/brands/1/edit 
- Click in the CreateAction "New Product" of the Products Relation Manager
- Click on the "Name" TextInput
- Scroll all the way down (zoom window if needed)
- Click in the modal or outside of the modal (anywhere but a focusable element)

The behaviour seen is the same when i click a ToggleButton

### Relevant log output

_No response_

archilex:
## Description
Opening a modal from a relation manager causes focus issues with form fields. Some of the symptoms of this bug are:

1. When any RM modal is opened (header action, table action, filters in a slideover, etc.), text fields remains selected even after clicking out of the field. 
2. Typing after clicking out of the field continues to update the field.
3. Since the text field remains selected, trying to trigger buttons, toggles, doesn’t work.
4. If your form is long and scrolls, clicking a button at the bottom will scroll you back to the still selected field.
5. To trigger a button/toggle/radio deck etc. you have to double click quickly, which apparentetly breaks through the focus issue (however only for that click).

I’ve tracked this down to the `tabindex="0"` on the relation manager when multiple relation managers are used. The tabindex seems to be conflicting with the modal's focusing. 

This is probably some type of conflict/bug with livewire or alpine.js's focus trap, as `tabindex="0"` has been in core for more than 2 years. But ultimately, I'm unsure as to why `tabindex="0"` is used at all. You can already tab to the RM's buttons/links without it. And Filament doesn't add tabindex to subnavigation groups for example. (That said I'm not an accesibility expert either).

This is only an issue when multiple relation managers (or relation manager groups) are on the page (ie when the tab navigation component is displayed). Displaying a single relation manager or single relation manager group doesn't cause this issue as `tabindex="0"` is only added conditionaly when more than one RM are displayed via tabs.

I did try rolling back filament and livewire as far as I could go and still ran into this issue so not sure what version of filament/livewire/alpine this was introduced in

## Visual changes
Here's a video of the issue from Filament's demo page, but this issue has also been confirmed in our production app, as well as my plugin demo app.

https://github.com/user-attachments/assets/9992d084-0dc0-4a13-9f4d-548bd53f261e

## Functional changes
This PR just removes `tabindex="0"` from the conditional. 

I've also tested that:
1. The relation manager's actions, links, etc. can still be tabbed to,
2. When opening a modal, the autofocus is still applied to the first element, 
3. Focus trapping still works. 
4. `->modalAutoFocus(false)` also continues to work

If there is a better way to handle this then let me know. 

After doing some more investigating this should also address and close https://github.com/filamentphp/filament/issues/15289, https://github.com/filamentphp/filament/issues/14999, https://github.com/filamentphp/filament/issues/14913

There could be other issues out there, but that's as far as I looked.

### DIFF
--- a/packages/panels/resources/views/components/resources/relation-managers.blade.php
+++ b/packages/panels/resources/views/components/resources/relation-managers.blade.php
@@ -72,7 +72,6 @@
             @if (count($managers) > 1)
                 id="relationManager{{ ucfirst($activeManager) }}"
                 role="tabpanel"
-                tabindex="0"
             @endif
             wire:key="{{ $this->getId() }}.relation-managers.active"
             class="flex flex-col gap-y-4"


bzy107:
The issue with #13661 has also been resolved. Congratulations!

arshaviras:
The issue #14999 also resolved, thanks!

gustavocaiano:
Can confirm the issue #15289 is resolved with this. Thank you!

danharrin:
Thanks!

PR #15439, opened 2025-01-25
Fix: Allow to override ExportFormat with custom enum instance

jeremiealvin:
## Description
This code allow to override the ExportFormat enum by implementing a common interface on a new enum.
This allows the developer to create a custom exporter with other formats than the default specified in ExportFormat enum.

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/CanExportRecords.php
+++ b/packages/actions/src/Concerns/CanExportRecords.php
@@ -5,6 +5,7 @@ namespace Filament\Actions\Concerns;
 use AnourValar\EloquentSerialize\Facades\EloquentSerializeFacade;
 use Closure;
 use Filament\Actions\ExportAction;
+use Filament\Actions\Exports\Enums\Contracts\ExportFormat as ExportFormatInterface;
 use Filament\Actions\Exports\Enums\ExportFormat;
 use Filament\Actions\Exports\ExportColumn;
 use Filament\Actions\Exports\Exporter;
@@ -54,7 +55,7 @@ trait CanExportRecords
     protected string | Closure | null $fileName = null;
 
     /**
-     * @var array<ExportFormat> | Closure | null
+     * @var array<ExportFormatInterface> | Closure | null
      */
     protected array | Closure | null $formats = null;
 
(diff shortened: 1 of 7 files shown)


danharrin (changes requested):
Please rename it to `ExportFormat` and import it in the original enum as `ExportFormatInterface`

Please move it to `Enums/Contracts`

jeremiealvin:
@danharrin done ✅

danharrin:
@jeremiealvin can you please let me know how the `DownloadExport` knows which enum to use, since the value is resolved from the query string? I don't think you have tested this code in your project, right?

jeremiealvin:
I was obligated to create a new route and a new controller that replicate the default one.
2025-02-04 at 22 20 33@2x
2025-02-04 at 22 22 32@2x

I know it is not the best but maybe to improve overloading we can add a setter for the $format variable in the `DownloadExport`.  Then we can extend the default class to overload the format.

danharrin (changes requested):
Can you add a new `getFormatFromRequest(Request $request): ExportFormatInterface` to the controller, so you only have to override that instead of the entire `__invoke()` which is more risky?

Also please test `$this->app->bind(DownloadExport::class, YourCustomDownloadExport::class)` in a service provider, does it work? That is better than redefining the route

jeremiealvin:
Yes, I'll give it a try, but since an enumeration can't be extended, do you think replacing DownloadExport with a new bind is a good thing?
If we do that, we're forced to redeclare the Csv and Xlsx formats in our new enum to retain the enum base options.

danharrin:
I can't think of any other options for this feature

jeremiealvin:
I've thought about it and it's not a problem after all. To avoid replication, developers can just tryFrom their custom enum first, and if null tryFrom the Filament enum.

danharrin:
How about the download links in the notification?

jeremiealvin:
My only problem about notification was about the translation of the download link.
calling the `translateLabel()` method gives me the possibility to have a custom translation in my json file.

here is an example:

2025-02-12 at 22 01 52@2x

jeremiealvin:
@danharrin I have added (and pushed) the getFormatFromRequest method to allow overriding

danharrin:
Thanks, I did rename to `resolveFormatFromRequest()` in the end

PR #15446, opened 2025-01-27
Fix: QueryBuilder's TextConstraint for Postgres

Zanzofily:
## Description

When a table name or column name contains uppercase letters, both table name and column name gets wrapped between double quotes inside any `TextContraint` operator.

for example: `Flight` table name with `code` field name becomes `"Flight.code"` which results in column not found error. Instead, it should be `"Flight".code`, and if column name contains uppercase letters, it should be `"Flight"."Code"`.

Example error: ```Undefined column: 7 ERROR:  column "Flight.code" does not exist
LINE 1: ...count(*) as aggregate from "Flight" where (lower("Flight.c...
                                                             ^ (Connection: pgsql, SQL: select count(*) as aggregate from "Flight" where (lower("Flight.code"::text)::text like search_string%))```

It might not be very common when it comes to Laravel naming conventions, but I encountered this issue while building a dashboard for an app that was initially developed with Next.js and prisma, where this naming convention is common.

## Functional changes


danharrin:
Thanks!

PR #15452, opened 2025-01-28
Update getNavigationBadge() to allow int|string|null return type

iRaziul:
## Description
Updated `getNavigationBadge` method to support multiple return types.  
The method can now return `int`, `string`, or `null` instead of only `string|null`.  
This allows more flexibility for navigation badges that might need to display numeric values or whenever using strict types.

## Visual changes
No visual changes - this is a type declaration update that doesn't affect runtime behavior.

## Functional changes
- [x] Code style has been fixed by running the `composer cs` command
- [x] Documentation is up-to-date (assuming method docs were updated if needed)

### DIFF
--- a/packages/panels/src/Resources/Resource.php
+++ b/packages/panels/src/Resources/Resource.php
@@ -822,7 +822,7 @@ abstract class Resource
         return static::$navigationLabel ?? static::getTitleCasePluralModelLabel();
     }
 
-    public static function getNavigationBadge(): ?string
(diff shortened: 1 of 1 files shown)


danharrin:
Hi, I think there are many places like this in Filament that *could* accept an int instead of a string, but if you choose to enable strict types you should probably just cast the string on your end. It would be too much work to update the entire Filament codebase to accept ints where we also accept strings, and I don't want it to be inconsistent. Sorry.

PR #15462, opened 2025-01-29
Add override possibility to page aware widgets

dododedodonl:
## Description

For some resource list pages, I have extra logic and parameters in the `mount` method. In order for widgets to be able to render correctly, some data has to be passed into the mount method params. This makes it possible to implement this logic easily.

## Visual changes

None.

## Functional changes


danharrin:
Thanks

Issue #11084, opened 2024-01-26
File upload doesn't work property if it's after a hidden step (form wizard)


sprtk-ches:
### Package

filament/filament

### Package Version

v3.2.14

### Laravel Version

10.42.0

### Livewire Version

v3.4.1

### PHP Version

8.2.15

### Problem description

If you have a wizard with a hidden step, and you conditionally show that step, the file upload after that hidden step will not work property and with throw the following JS error

```
Uncaught (in promise) TypeError: this.pond is null
```

This issue was introduced in [version 3.1.24](https://github.com/filamentphp/filament/releases/tag/v3.1.24) (I believe this was the PR https://github.com/filamentphp/filament/pull/10371)

### Expected behavior

The file upload works properly even if it's after a hidden-step-turned-visible

### Steps to reproduce

1. Create a form with a wizard and 3 steps
2. Add a file upload component in step 3
3. Make step 2 conditionally hidden
4. Toggle the 2nd step's visibility status
5. Open the browser console, go to step 3 and try to upload a file
6. You will see an error in the console

Here is a simple form that can reproduce this

```php
Wizard::make([
    Wizard\Step::make('Step 1')
        ->schema([
            Toggle::make('something')
                ->label('Show step 2')
                ->live()
        ]),
    Wizard\Step::make('Hidden step')
        ->visible(fn (Get $get) => $get('something'))
        ->schema([]), // nothing to do here
    Wizard\Step::make('Step 3')
        ->schema([
            FileUpload::make('media')
                ->hint('Open the browser console and try to upload a file')
        ])
])
```

### Reproduction repository

https://github.com/sprtk-ches/laravel-filament

### Relevant log output

```shell
Uncaught (in promise) TypeError: this.pond is null
    B http://filament.local/js/filament/forms/components/file-upload.js?v=3.2.14.0:40
    i http://filament.local/js/filament/forms/components/file-upload.js?v=3.2.14.0:1
    setTimeout handler*Ml http://filament.local/js/filament/forms/components/file-upload.js?v=3.2.14.0:1
    i http://filament.local/js/filament/forms/components/file-upload.js?v=3.2.14.0:1
    i http://filament.local/js/filament/forms/components/file-upload.js?v=3.2.14.0:1
    fire http://filament.local/js/filament/forms/components/file-upload.js?v=3.2.14.0:1
    b http://filament.local/js/filament/forms/components/file-upload.js?v=3.2.14.0:2
[...]

PR #15467, opened 2025-01-29
Always render DOM for hidden wizard steps

freshleafmedia:
## Description

When applying conditional `hidden()` reactivity to wizard steps you get weird rendering of subsequent steps on previous ones.

There is a repro repo here: https://github.com/freshleafmedia/filament-wizard-step-repro

### Before
https://github.com/user-attachments/assets/fe00c4ee-9ce5-486a-aa97-a7f6a0514add

### After
https://github.com/user-attachments/assets/7cce06b9-75ff-4a33-a18f-ed8c22830c65

## Functional changes

### DIFF
--- a/packages/forms/resources/views/components/wizard.blade.php
+++ b/packages/forms/resources/views/components/wizard.blade.php
@@ -256,7 +256,7 @@
         @endforeach
     </ol>
 
-    @foreach ($getChildComponentContainer()->getComponents() as $step)
+    @foreach ($getChildComponentContainer()->getComponents(true) as $step)
         {{ $step }}
     @endforeach
 
--- a/packages/forms/resources/views/components/wizard.blade.php
+++ b/packages/forms/resources/views/components/wizard.blade.php
@@ -119,7 +119,6 @@
         type="hidden"
         value="{{
             collect($getChildComponentContainer()->getComponents())
-                ->filter(static fn (\Filament\Forms\Components\Wizard\Step $step): bool => $step->isVisible())
                 ->map(static fn (\Filament\Forms\Components\Wizard\Step $step) => $step->getId())
                 ->values()
                 ->toJson()


freshleafmedia:
I also noted that the hidden `stepsData` input filters out hidden steps but the `getComponents()` call doesn't return them by default.

See https://github.com/filamentphp/filament/blob/ae86b8c77c6349f5350ace412b22b5d1a55f7c4f/packages/forms/resources/views/components/wizard.blade.php#L121-L125

danharrin:
@freshleafmedia regarding your latest comment does something need to be changed, I don't follow what is wrong.

freshleafmedia:
@danharrin It's not causing a problem and doesn't need changing.

To explain: `getComponents()` only returns visible components, which makes the `filter()` call redundant.

I only mention it because it suggests the wizard was built on the assumption that `getComponents()` returns hidden components, which AFAIK it doesn't.

Sorry for the confusion.

danharrin:
Ah okay, can you remove the redundant filter then please? And are you sure we don't now need to include step data for hidden steps after this change?

Also, how does this problem affect Tabs which operates on a similar principle, should hidden tab content be rendered too?

sprtk-ches:
Yeap tried it. 

Closes https://github.com/filamentphp/filament/issues/11084

That being said, it does render the body of hidden steps, so I'm not sure if it's something to consider. For example, if you have something that you do not want to show, it will be visible if you inspect the DOM

Here is an example 
image


As you can see, the solutions step is not visible, but you can see the body of that tab in the DOM

danharrin:
Yeah I think that is an necessary evil

archilex:
@freshleafmedia I ran into the same issue and solved it by adding `wire:key="{{ $this->getId() }}"` to the wrapping div of the wizard step blade file. Not sure if that's a better solution / or even fully fixes the problem for everyone, but thought I'd mention it. 

Edit: I just pulled down your repo to test and indeed `wire:key="{{ $this->getId() }}"` fixes it in yours as well with the added benefit that we're not rendering the hidden steps.

Haven't tested if it closes #11084

sprtk-ches:
Ok I tried @archilex's solution and it does appear to work (also solves https://github.com/filamentphp/filament/issues/11084 )

Not sure if you want to create another PR with your solution?

Anyway, up to you @danharrin but I disagree that it's a necessary evil. You could have a hidden step with sensitive information visible to certain users only that would still be rendered.

danharrin:
@sprtk-ches sure, if there is an alternative that is preferable.

archilex:
@freshleafmedia feel free to update your PR with the alternate fix if you like. Or I can open a separate PR if you prefer.

freshleafmedia:
@archilex Adding the key seems to work but for the wrong reason, adding `wire:key="{{ $this->getId() }}"` gives the following:



The keys are all the same, which is not what is [intended](https://livewire.laravel.com/docs/components#adding-wirekey-to-foreach-loops). Adding `wire:key="{{ $id }}"` gives unique keys but then the original problem returns.

I believe the root of the problem is a conflict of control between Livewire and AlpineJS. By getting Livewire to render all steps up-front AlpineJS is solely responsible for controlling the step visibility.

freshleafmedia:
@danharrin Are you happy to merge this PR as-is?

danharrin:
How about adding the step ID into the `wire:key` somewhere? No, judging by the responses in the thread it doesn't sound like it's safe to render hidden steps anyway

freshleafmedia:
Setting the key doesn't resolve the problem.

I have tried `wire:key="{{ $id }}"`, `$id` being the step id, but the problem persists.

The only comment which alluded to safety you dismissed as necessary evil, has that changed?

archilex:
@freshleafmedia Dan means to append the step key to the component key: `wire:key="{{ $this->getId() }}.{{ $id }}”. That way livewire is forced to rerender the components but each key is unique. 

Might want to add “step” in there. You can search the code base and see how this strategy is implemented in other places.

freshleafmedia:
@archilex Just tried `wire:key="{{ $this->getId() }}.step.{{ $id }}"`, this makes the keys globally unique but the problem persists.

archilex:
@freshleafmedia You're right. Looks like the component key isn't changing between requests. I assumed it was since just adding `{{ $this->getId() }}` was working. Looks like that leaves the brute force way of either appending a random key or just using a random key `wire:key="{{ \Illuminate\Support\Str::random() }}"` which does appear to fix the issue.

danharrin:
I'm not going to merge in a solution with a random key, it will cause all sorts of problems on re-render

danharrin:
Thanks for the discussion, but closing until a new alternative is found. Maybe try rendering the containers for the wizard steps but not their content or something? I would welcome a new PR when this is investigated further.

freshleafmedia:
@danharrin I'm confused.

Please can you tell me what is wrong with this solution that requires a _new_ alternative?

The only problem that was raised in the previous comments was that if you inspect the DOM you can see the hidden content but you dismissed that as a necessary evil. Was that decision reversed?

I'm not trying to be difficult, I just don't understand what I need to change to get this bug fixed.

danharrin:
Yeah I think I was wrong, I don't think hidden components should be rendered

PR #15476, opened 2025-01-30
Add the ability to sort the search results in global search resource

derpoho:
## Description

Adds the ability to set `$globalSearchSort` to any integer to sort the output of the global search by resource.
As fallback it should respect the filament native sorting based on resource loading. Event if some resources have it set, others have not, it will respect that and sort the set resources first and the others afterwards. So there shouldn't be any breaking change to existing installations.

## Visual changes

Change in the sorting of resources.

## Functional changes


derpoho:
Additionally: As far as i can see and not knowing what @danharrin is planning with search in v4 it should also be compatible with v4.

danharrin:

You can customize this in your own app using a "global search provider", which is a class you can pass to `$panel->globalSearch(YourCustomProvider::class)`. Check `Filament\GlobalSearch\Providers\DefaultGlobalSearchProvider` for the default provider implementation to copy from.

I would accept a PR once v4 is released, but I think it should be implemented differently. I think the changes need to be made inside the `DefaultGlobalSearchProvider` instead of the results object like it is currently.

derpoho:
roger!

PR #15508, opened 2025-02-04
Update 05-viewing-records.md

DBawazir2002:
The infolist is not a static method.


danharrin:
> you can define an `infolist()` method on the resource class

On the resource class it is static

DBawazir2002:
You can review the ViewRecord class to see that the method is non-static, while in the resource, it is a static method.

danharrin:
I know, but read the sentence above the code example you edited

Issue #10436, opened 2023-12-22
Use container queries in more places


zepfietje:

zepfietje:
@danharrin, as part of this we should probably add an API to control components at different container query breakpoints, much like the current screen breakpoints.

danharrin:
Quick note: anything that wraps an action group cannot be a container since https://github.com/floating-ui/floating-ui/issues/3067

`contain: layout` on the dropdown causes unwanted overflow clipping for bulk actions above a table

PR #15522, opened 2025-02-05
Container queries

danharrin:


danharrin:
Closes #10436

packages/schemas/resources/views/schema.blade.php
```diff
@@ -36,6 +36,10 @@ function (Component | Action | ActionGroup $component) use (&$hasVisibleComponen
 @endphp
 
 @if ((! $isDirectlyHidden()) && $hasVisibleComponents)
+    @if ($isRoot)
+        {!! '<div class="fi-grid-ctn">' !!}
```
saade, line 40: Out of curiosity, what's the diff between these two?

```blade
@if($isRoot)
    <div class="fi-grid-ctn">
@endif
```

danharrin: The Prettier formatter doesn't like that 😁

saade: LOL 🤦🏼‍♂️

PR #15523, opened 2025-02-05
add estonian translations

GasparSukk:
## Description

Add Estonian translations to filament

P.S. I can be the maintainer for ET language if neccessary.


danharrin:
Hi, thanks for this! However, we are not accepting any new languages for Filament v3. You could use these files in your app in the meantime. In v4, please PR the language in and we will merge it :) Many thanks.

PR #15542, opened 2025-02-06
filament() helper return types

erikgaal:
## Functional changes


packages/panels/src/global_helpers.php
```diff
@@ -4,6 +4,10 @@
 use Filament\FilamentManager;
 
 if (! function_exists('filament')) {
+    /**
+     * @param string|null $plugin
```
danharrin: Do we need this line for the other to work? If not please remove it

erikgaal: nah we don't i'll remove it

PR #15582, opened 2025-02-11
Allow to disable a modal trigger

CharlieEtienne:
## Description

Currently there is no way to prevent a modal from being triggered. Even if the trigger button is disabled, the modal would still open, as in this example:

```blade
<x-filament::modal id="modalTest">
    <x-slot name="trigger">
        <x-filament::button :disabled="true">I'm disabled</x-filament::button>
    </x-slot>
    Hello, I'm a modal
</x-filament::modal>
```

This PR lets users disable the trigger element like this:

```blade
<x-filament::modal id="modalTest">
    <x-slot name="trigger" :disabled="true">
        <x-filament::button :disabled="true">I'm disabled</x-filament::button>
    </x-slot>
    Hello, I'm a modal
</x-filament::modal>
```

It also adds documentation for this change.

## Functional changes

### DIFF
--- a/packages/support/docs/09-blade-components/02-modal.md
+++ b/packages/support/docs/09-blade-components/02-modal.md
@@ -231,7 +231,7 @@ By default, modals will autofocus on the first focusable element when opened. If
 
 ## Disabling the modal trigger button
 
-By default, the trigger button will open the modal even if it disabled, since the click event listener is registered on a wrapping element of the button itself. If you want to prevent the modal from opening, you should also use the `disabled` attribute on the trigger slot:
+By default, the trigger button will open the modal even if it is disabled, since the click event listener is registered on a wrapping element of the button itself. If you want to prevent the modal from opening, you should also use the `disabled` attribute on the trigger slot:
 
 ```blade
 <x-filament::modal>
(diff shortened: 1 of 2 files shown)


packages/support/resources/views/components/modal/index.blade.php
```diff
@@ -106,7 +106,9 @@
 >
     @if ($trigger)
         <div
-            x-on:click="open"
+            @if(!$trigger->attributes['disabled'] ?? false)
```
danharrin:
```suggestion
            @if (! $trigger->attributes->get('disabled'))
```

Does this work?

packages/support/docs/09-blade-components/02-modal.md
danharrin, line 1: Does `<x-slot name="trigger" disabled>` work, as I updated in the docs?

packages/support/resources/views/components/modal/index.blade.php
CharlieEtienne: Yes it works for me at least

packages/support/docs/09-blade-components/02-modal.md
CharlieEtienne: Both work:

`<x-slot name="trigger" disabled>` and `<x-slot name="trigger" :disabled="true">`

danharrin (approved):

danharrin:
Thanks

PR #15585, opened 2025-02-11
Update choices.js to ^11.0.2

informatJonas:
## Description

Update choices.js to version ^11.0.2 (reason for this, in this version the grouping & single option in a select was fixed)

Adjustment in select.js => `setChoices` (An additional array element was mistakenly added during grouping, this is removed using `choices.filter(...)`)

### Before
image

### After
image

## Functional changes

- [ ] Code style has been fixed by running the `composer cs` command. (I have not customized any Composer relevant files)


danharrin:
Hi, this is a breaking change and has been upgraded for v4

PR #15594, opened 2025-02-12
Fix conditionally hiding summaries

damiandejong:
## Description

The ability to conditionally hide summaries as described in the [documentation](https://filamentphp.com/docs/3.x/tables/summaries#conditionally-hiding-the-summary) by injecting e.g. `$query` is currently not working. When injecting `$query` it always returns `null` and also the other dependencies of a summarizer cannot be reached. The reason is that they are not yet injected when the visibility condition is evaluated.

With the changes in this PR, the visibility condition is evaluated after the injection instead of beforehand.

## Functional changes

### DIFF
--- a/packages/tables/resources/views/components/summary/index.blade.php
+++ b/packages/tables/resources/views/components/summary/index.blade.php
@@ -28,6 +28,11 @@
 @endphp
 
 @if ($hasPageSummary)
+    @php
+        $query = $this->getPageTableSummaryQuery();
+        $selectedState = $this->getTableSummarySelectedState($query)[0] ?? [];
+    @endphp
+
     <x-filament-tables::row
         class="fi-ta-summary-header-row bg-gray-50 dark:bg-white/5"
     >
@@ -46,7 +51,7 @@
         @endif
 
         @foreach ($columns as $column)
-            @if ($placeholderColumns || $column->hasSummary())
+            @if ($placeholderColumns || $column->hasSummary($query))
                 @php
                     $alignment = $column->getAlignment() ?? Alignment::Start;
 
@@ -54,7 +59,7 @@
                         $alignment = filled($alignment) ? (Alignment::tryFrom($alignment) ?? $alignment) : null;
                     }
 
-                    $hasColumnHeaderLabel = (! $placeholderColumns) || $column->hasSummary();
+                    $hasColumnHeaderLabel = (! $placeholderColumns) || $column->hasSummary($query);
                 @endphp
 
                 <x-filament-tables::summary.header-cell
@@ -92,11 +97,6 @@
             <td></td>
         @endif
     </x-filament-tables::row>
-
-    @php
-        $query = $this->getPageTableSummaryQuery();
-        $selectedState = $this->getTableSummarySelectedState($query)[0] ?? [];
-    @endphp
 
     <x-filament-tables::summary.row
         :actions="$actions"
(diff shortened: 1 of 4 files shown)


danharrin (changes requested):
Please look at the other usages of `getSummarizers()`.

damiandejong:
@danharrin I have checked the other usages. Most of the other usages check whether the column actually has a summary by calling `hasSummary()`. I am not sure whether it's the best solution, but I got it working by passing the `$query` in `getSummarizers()` instead of injecting it at the moment of rendering the actual summarizer. See the latest commit.

There is only one usage left for which there is no evaluation of the injected parameters possible, if I understand everything correctly. When the columns are pushed to the table instance, `hasSummary()` is evaluated to determine whether the table has a summary overall. For each column it counts the summarizers in `getSummarizers`, but at that point in the lifecycle, it cannot access the the table instance yet. I am correct or do I miss something that might solve this?

danharrin:
When you submitted the PR, I assume your original changes were tested and worked, right? So am I wrong in suggesting we look at the other usages? Would you be satisfied if we just merged your original commit? What implications are there here?

damiandejong:
@danharrin The changes in the original PR were tested and confirmed to work without breaking anything.

That said, I appreciate the suggestion to review other usages of getSummarizers. In this PR, filtering summarizers for those usages will be skipped. Those usages only check whether the table has any summarizers to determine whether to show or hide the summary overview. Ideally, it would be best to incorporate this evaluation, but I currently don’t see how—likely because the solution goes beyond my current understanding of Filament’s architecture.

With this PR, the table will still consider itself as having a summary even if all summarizers are hidden due to condition evaluation. This happens because it only counts the number of summarizers, which results in displaying an empty summary row.

On the other hand, summarizers are not currently filtered correctly, or at least, conditionally hiding them based on injected parameters throws an exception—despite this being documented as a valid use case. My PR resolves this exception and addresses my specific use case, so I would be satisfied if you merged the original commit, but I understand if you prefer to merge only a solution that covers all usages.

damiandejong:
@danharrin I will try to be more specific on what the implications are when not evaluating the dependency injection at the other usages of `getSummarizers()`.

Let's say, I want to condition a summarizer like this:
``
Tables\Columns\Summarizers\Sum::make()->hidden(fn (Builder $query) => $query->exists())
``

According to the docs, this should be possible. See (here)[https://filamentphp.com/docs/3.x/tables/summaries#conditionally-hiding-the-summary]. However, this results in the following exception: 
`Call to a member function exists() on null`

This problem is solved with my initial PR. See the screenshot below.
image

The only implication is that when all summarizers would be evaluated as hidden, the summary row will still be displayed as below. In my opinion, that's not really an issue.
image

Therefore, I will revert my second commit and stick with the first commit to at least solve the exceptions thrown and I can hide the custom summarizers in my use case that result in 0.

bscarola:
This change introduced a negative performance side effect: hidden summarizers now always have their queries executed. This breaks a somewhat common workflow of using an Action to hide/show summaries on a large table for performance reasons. I also find the blank summary rows to be a confusing UX.

This can potentially be worked around by passing an empty array to Column->summarize(), however that method does not currently accept a closure.

danharrin:
@bscarola please open an issue with a reproduction so we can better track a bug fix for that

bernhardh:
Same issue here. I will try to create a reproducable repo for that and create a separate issue.

PR #15612, opened 2025-02-15
Add ability to hook into a clone

ralphjsmit:
Currently, it is not possible to add custom macros on e.g. form components that persist a custom property. For example, take the following set of macro:

```php
Filament\Support\Components\ViewComponent::macro('someIdentifier', function (string $identifier) {
       // Where to store? Dynamic properties are deprecated from 8.2 on and cause many file system log writes.
});

Filament\Support\Components\ViewComponent::macro('getSomeIdentifier', function () {
       // Where to retrieve?
});
```

A solution I tried was to make a singleton object to store the identifiers based on e.g. in a `WeakMap`, however that does not work because (it seems) that components are cloned and therefore then there is no way to hook them also into the WeakMap.

This PR adds some tests for the cloning and the ability to run a callback any time a component is cloned, thus providing the ability to have a form of "persistent" properties throughout a macro.

Thanks!

### DIFF
--- a/packages/forms/src/Components/Concerns/Cloneable.php
+++ b/packages/forms/src/Components/Concerns/Cloneable.php
@@ -23,14 +23,13 @@ trait Cloneable
 
         return $this;
     }
-	
-	
-	public function afterClone(Closure $callback): static
-	{
-		$this->cloneCallbacks[] = $callback;
-		
-		return $this;
-	}
+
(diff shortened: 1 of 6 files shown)


danharrin:
Please can you submit this to v4?

ralphjsmit:
@danharrin Just targeted to v4, see #15614.

Issue #15611, opened 2025-02-15
acceptedFileTypes on SpatieMediaLibraryFileUpload doesn't work for 3DM and SKP files


fmeccanici:
### Package

filament/spatie-laravel-media-library-plugin

### Package Version

v3.2.140

### Laravel Version

v11.42.1

### Livewire Version

v3.5.20

### PHP Version

PHP 8.3.16

### Problem description

I want to reject IFC files but allow 3DM and SKP files. To accomplish this I used the `SpatieMediaLibraryFileUpload` component in my `Resource` form and wanted to use the `->acceptedFileTypes()` function. I checked the source code and it sets rules for the mime types, so I researched which mime types where detected by PHP for these files:

```php
$threeDmFile->getMimeType();
"application/octet-stream"
```

```php
$ifcFile->getMimeType();
"text/plain"
```

```php
$skpFile->getMimeType();`
"application/vnd.sketchup.skp"
```

So I added the ones I want to allow to the component:

```php
Forms\Components\SpatieMediaLibraryFileUpload::make('designs')
    ->collection('designs')
    ->multiple()
    ->required()
    ->acceptedFileTypes([
        'application/vnd.sketchup.skp',
        'application/octet-stream'
    ]),
```
What happens after uploading all three files is that they all fail in their validation:

![Image](https://github.com/user-attachments/assets/0c84f18b-7941-40e4-be26-fa84d7f18102)

So I decided to check which mime types are detected:

```php
Forms\Components\SpatieMediaLibraryFileUpload::make('designs')
    ->collection('designs')
    ->multiple()
    ->required()
    ->afterStateUpdated(function ($state) {
        ray(collect($state)->map->getMimeType()->toArray());
    })
```

This leads me to the following discovery:

![Image](https://github.com/user-attachments/assets/65a94e47-d1ee-4294-b42f-6c1983a0cae5)

The SKP file is correctly detected but the IFC and 3DM files are empty strings. From this observation you would expect the SKP file to be correctly validated but this is not the case (see first image). However when I add `'application/vnd.koan'` :

```php
->acceptedFileTypes([
    'application/vnd.sketchup.skp',
    'application/octet-stream',
    'application/vnd.koan'
]),
```
The SKP file is correctly uploaded:

![Image](https://github.com/user-attachments/assets/f907713f-c189-4155-ba01-bc0e956b6fff)

[...]

PR #15616, opened 2025-02-16
Add MIME type mapping support to file upload component

fmeccanici:
## Description
I solved the [issue](https://github.com/filamentphp/filament/issues/15611) I posted recently, because I needed more customisability on how filepond resolves the mime type. They provide an example in their [docs](https://pqina.nl/filepond/docs/getting-started/examples/validate-file-type/):

```js
import { create, registerPlugin } from 'filepond';
import 'filepond/dist/filepond.css';

// Import the File Type Validation plugin
import FilePondPluginFileValidateType from 'filepond-plugin-file-validate-type';

// Register the plugin with FilePond
registerPlugin(FilePondPluginFileValidateType);

// Get a file input reference
const input = document.querySelector('input[type="file"]');

// Create a FilePond instance
create(input, {
    // Only accept images
    acceptedFileTypes: ['image/*'],

    fileValidateTypeDetectType: (source, type) =>
        new Promise((resolve, reject) => {
            // test if is xls file
            if (/\.xls$/.test(source.name)) return resolve('application/vnd.ms-excel');

            // accept detected type
            resolve(type);
        }),
});
```
This is exactly what I implemented but configurable from the `FileUpload.php` component.

### Before
```php
FileUpload::make('designs')
    ->acceptedFileTypes([
        'application/octet-stream',
        'x-world/x-3dmf',
    ])
```
https://github.com/user-attachments/assets/c14fa18d-82f6-4639-913f-16b72014f298

### After
```php
FileUpload::make('designs')
    ->mimeTypeMap([
          '3dm' => 'x-world/x-3dmf'
      ])
    ->acceptedFileTypes([
        'application/octet-stream',
        'x-world/x-3dmf',
    ])
```
https://github.com/user-attachments/assets/516cb8ab-b9de-48ff-bbbd-8fd3428e3541

## Functional changes
You can now specify the mime type mapping (extension => mime type) on the `FileUpload.php` component like this:

```php
FileUpload::make('designs')
    ->mimeTypeMap([
        '3dm' => 'x-world/x-3dmf'
    ])
    ->acceptedFileTypes([
        'application/vnd.sketchup.skp',
        'application/octet-stream',
        'x-world/x-3dmf',
    ])
```

### DIFF
--- a/packages/forms/docs/05-validation.md
+++ b/packages/forms/docs/05-validation.md
@@ -602,3 +602,18 @@ Page::$reportValidationErrorUsing = function (ValidationException $exception) {
         ->send();
 };
 ```
+## Custom MIME Type Mapping
+
+Some file formats may not be recognized correctly by the browser when uploading files. Filament allows you to manually define MIME types for specific file extensions using the `mimeTypeMap()` method:
+
+```php
+FileUpload::make('designs')
+    ->mimeTypeMap([
+        '3dm' => 'x-world/x-3dmf',
+        'skp' => 'application/vnd.sketchup.skp',
+    ])
+    ->acceptedFileTypes([
+        'x-world/x-3dmf',
+        'application/vnd.sketchup.skp',
+    ]);
+```
--- a/packages/forms/docs/05-validation.md
+++ b/packages/forms/docs/05-validation.md
@@ -602,18 +602,3 @@ Page::$reportValidationErrorUsing = function (ValidationException $exception) {
         ->send();
 };
 ```
-## Custom MIME Type Mapping
-
-Some file formats may not be recognized correctly by the browser when uploading files. Filament allows you to manually define MIME types for specific file extensions using the `mimeTypeMap()` method:
-
-```php
-FileUpload::make('designs')
-    ->mimeTypeMap([
-        '3dm' => 'x-world/x-3dmf',
-        'skp' => 'application/vnd.sketchup.skp',
-    ])
-    ->acceptedFileTypes([
-        'x-world/x-3dmf',
-        'application/vnd.sketchup.skp',
-    ]);
-```
--- a/packages/forms/resources/js/components/file-upload.js
+++ b/packages/forms/resources/js/components/file-upload.js
@@ -74,7 +74,6 @@ export default function fileUploadFormComponent({
     uploadUsing,
     mimeTypeMap,
 }) {
-
     return {
         fileKeyIndex: {},
 
@@ -204,8 +203,14 @@ export default function fileUploadFormComponent({
                 },
                 fileValidateTypeDetectType: (source, detectedType) => {
                     return new Promise((resolve, reject) => {
-                        const extension = source.name.split('.').pop().toLowerCase()
-                        const mimeType = mimeTypeMap[extension] || detectedType || mime.getType(extension)
+                        const extension = source.name
+                            .split('.')
+                            .pop()
+                            .toLowerCase()
+                        const mimeType =
+                            mimeTypeMap[extension] ||
+                            detectedType ||
+                            mime.getType(extension)
 
                         mimeType ? resolve(mimeType) : reject()
                     })
--- a/packages/forms/resources/js/components/file-upload.js
+++ b/packages/forms/resources/js/components/file-upload.js
@@ -58,6 +58,7 @@ export default function fileUploadFormComponent({
     maxSize,
     minSize,
     maxParallelUploads,
+    mimeTypeMap,
     panelAspectRatio,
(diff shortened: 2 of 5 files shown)


packages/forms/resources/js/components/file-upload.js
```diff
@@ -72,6 +72,7 @@ export default function fileUploadFormComponent({
     uploadingMessage,
     uploadProgressIndicatorPosition,
     uploadUsing,
+    mimeTypeMap,
```
danharrin: Please ensure you keep the code style consistent and define this alphabetically

packages/forms/resources/views/components/file-upload.blade.php
```diff
@@ -39,6 +39,7 @@
                     deleteUploadedFileUsing: async (fileKey) => {
                         return await $wire.deleteUploadedFile(@js($statePath), fileKey)
                     },
+                    mimeTypeMap: @js($getMimeTypeMap()),
```
danharrin: Please define this alphabetically

packages/forms/src/Components/FileUpload.php
danharrin, line 1: Please look at the other areas of the codebase where we allow `array | Closure`, and ensure closure evaluation is available for this method also.

packages/forms/docs/05-validation.md
danharrin, line 1: Please do not define this in the validation section. It's not true backend validation, and it also only applies to the file upload field. Please add it here: https://filamentphp.com/docs/3.x/forms/fields/file-upload#file-type-validation

danharrin (changes requested):

fmeccanici:
@danharrin I applied the suggested improvements

batnieluyo:
This MR break the file extension validation on frontend

packages/forms/resources/views/components/file-upload.blade.php
```diff
@@ -71,6 +71,7 @@
                     maxFiles: @js($getMaxFiles()),
                     maxSize: @js(($size = $getMaxSize()) ? "{$size}KB" : null),
                     minSize: @js(($size = $getMinSize()) ? "{$size}KB" : null),
+                    mimeTypeMap: @js($getMimeTypeMap()),
```
batnieluyo, line 74: this not should be 

```
mimeTypeMap: @js(($mimeMap = $getMimeTypeMap()) ? $mimeMap  : []),
``` 

??

batnieluyo (commented):
Issue on frontend validation mimeTypeMap

danharrin:
Can you please expand on what the error is?

habibi-dev:
It looks like this PR introduced a breaking change. The acceptedFileTypes option no longer works as expected. No matter what file type I specify, the uploader rejects all files with the message "Invalid file type."

This behavior effectively makes acceptedFileTypes useless, which seems like a bug rather than an intended feature. If there's a new way to handle file type validation, could you please clarify it in the documentation?

Has anyone else experienced this issue? 🚀

danharrin:
@habibi-dev no, I don't have the problem after this change. What plugins do you have installed?

packages/forms/resources/views/components/file-upload.blade.php
danharrin: Why? The return type of `getMimeTypeMap()` is `array` and not `?array`

habibi-dev:
@danharrin
image

danharrin:
Please create a new minimal reproduction repository for the bug, and open an issue

habibi-dev:
Issue Resolved! 🎉

The problem was that my views were outdated. After running:
```php artisan vendor:publish ```
to publish the latest changes, everything started working correctly. Thanks for the guidance! 🚀

**Comment by danharrin** [comment]
https://filamentphp.com/content/danharrin-publishing-views-in-laravel

**Comment by grzegorzkamil** [comment]
> It looks like this PR introduced a breaking change. The acceptedFileTypes option no longer works as expected. No matter what file type I specify, the uploader rejects all files with the message "Invalid file type."
> 
> This behavior effectively makes acceptedFileTypes useless, which seems like a bug rather than an intended feature. If there's a new way to handle file type validation, could you please clarify it in the documentation?
> 
> Has anyone else experienced this issue? 🚀

Yes, I have the same issue.

```
            Forms\Components\FileUpload::make('thumbnail')
                ->label(__('Thumbnail'))
                ->image()
                ->directory('images/course/thumbnails')
                ->required(),
```
                
causes an issue.

Here are the screenshots:

image

image

danharrin:
Have you published the Blade views?

PR #15634, opened 2025-02-19
Fix forced case search with uppercase string

TheSkybreaker:
## Description
When using `->forceSearchCaseInsensitive()` on a column that is accessed via json extraction it doesn't return results if terms searched are in uppercase.

## Visual changes
None

## Functional changes


danharrin:
Hello, we have handed over maintenance of this plugin to trusted developers, here: https://github.com/lara-zeus/translatable. You should open a new PR there for a resolution. It may already be fixed if you use their fork also.

The current plan is to officially deprecate the plugin and suggest the new plugin when v4 is released.

Many thanks

PR #15650, opened 2025-02-20
fix: incorrect loaded state with HasOneOrMany select

SychO9:
## Description

When using the `Select` component on a HasMany (or HasOne) relationship, the loaded records are incorrect. The `loadStateFromRelationshipsUsing` callback implementation uses the foreign key value when it should use the local key of the related model.

For example, if you have a Group (id,name) model and a User (id,name,group_id) model. right now it plucks the `group_id` values of the related users.

```php
class Group extends Model
{
    protected $guarded = ['id'];

    public function users(): HasMany
    {
        return $this->hasMany(User::class);
    }
}

class GroupResource extends Resource
{
    protected static ?string $model = Group::class;

    {
        return $form
            ->schema([
                    ->required(),
                Forms\Components\Select::make('users')
                    ->searchable()
                    ->multiple(),
            ]);
    }
}
```

## Visual changes

**Before**
image

**After**
image

## Functional changes

### DIFF
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -862,7 +862,7 @@ class Select extends Field implements Contracts\CanDisableOptions, Contracts\Has
 
                 $component->state(
                     $relatedRecords
-                        ->pluck($relationship->getForeignKeyName())
+                        ->pluck($relationship->getLocalKeyName())
                         ->all(),
                 );
 
@@ -874,7 +874,7 @@ class Select extends Field implements Contracts\CanDisableOptions, Contracts\Has
 
                 $component->state(
                     $relatedModel?->getAttribute(
-                        $relationship->getForeignKeyName(),
+                        $relationship->getLocalKeyName(),
                     ),
                 );
 
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -988,7 +988,7 @@ class Select extends Field implements Contracts\CanDisableOptions, Contracts\Has
         $this->saveRelationshipsUsing(static function (Select $component, Model $record, $state) use ($modifyQueryUsing) {
             $relationship = $component->getRelationship();
 
-            if ($relationship instanceof HasOne || $relationship instanceof HasMany) {
+            if (($relationship instanceof HasOne) || ($relationship instanceof HasMany)) {
                 $query = $relationship->getQuery();
 
                 if ($modifyQueryUsing) {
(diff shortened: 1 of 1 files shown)


SychO9:
I was also confused by the fact it doesn't automatically save the relationship changes like other relation types. I see that it's skipped on purpose in the `saveRelationshipsUsing` callback, is there a particular reason for that or would an addition to make it happen be welcome?

danharrin:
If you want to implement saving, please go for it

SychO9:
Awesome, I've pushed a commit here, let me know if you would rather see that separated into another pull request, I wasn't sure what the preference on that would be.

packages/forms/src/Components/Select.php
```diff
@@ -988,6 +988,42 @@ static function (Select $component): bool {
         $this->saveRelationshipsUsing(static function (Select $component, Model $record, $state) use ($modifyQueryUsing) {
             $relationship = $component->getRelationship();
 
+            if ($relationship instanceof HasOne || $relationship instanceof HasMany) {
```
danharrin: Maybe check `HasOneOrMany` instead of both?

SychO9: That would then also include the morph relation types (MorphOne, MorphMany) which would fail with the added logic

danharrin: Ah okay

packages/forms/src/Components/Select.php
```diff
@@ -988,6 +988,42 @@ static function (Select $component): bool {
...
+
+                        if ($modifyQueryUsing) {
+                            $component->evaluate($modifyQueryUsing, [
+                                'query' => $query,
+                                'search' => null,
+                            ]);
+                        }
+
+                        $query->update([
+                            $relationship->getForeignKeyName() => $record->getKey(),
```
danharrin: Do we need to use the key name from the relationship here instead of the primary key?

packages/forms/src/Components/Select.php
```suggestion
            if (($relationship instanceof HasOne) || ($relationship instanceof HasMany)) {
```

danharrin (changes requested):

packages/forms/src/Components/Select.php
SychO9: good catch! yes absolutely. Pushed a fix

danharrin:
Thanks

danharrin:
I'm going to revert the saving part of this, sorry. It does seem to break some apps (see #15980). I will bring it back in v4

PR #15662, opened 2025-02-21
fix: Exporter not caching export columns

AurelDemiri:
## Description

This PR fixes ExportColumns not being correctly cached by the Exporter.


danharrin:
Good catch, thanks

PR #15695, opened 2025-02-24
Laravel 12 support

danharrin:
- [x] https://github.com/AnourValar/eloquent-serialize/pull/21
- [x] https://github.com/ryangjchandler/blade-capture-directive/pull/23


danielhe4rt:
Hey Dan, when we should be able to use L12 with Filament? :eyes:

hamrak:
> Hey Dan, when we should be able to use L12 with Filament? 👀

filament/actions depends on "anourvalar/eloquent-serialize" ... https://github.com/AnourValar/eloquent-serialize/pull/21 ... we must wait :)

danharrin:
~~Still awaiting https://github.com/ryangjchandler/blade-capture-directive/pull/23, but I merged and released this so that as soon as that package is ready, you can install Filament.~~

Update - released!

PR #15698, opened 2025-02-24
SelectFilter: allowHtml added

DannZolik:
## Description

Added the `CanAllowHtml` trait to the `SelectFilter` class and updated the `getFormField` function to support HTML in filters. Previously, while the `Select` form component allowed HTML in options, it was not possible to use `SelectFilter` with HTML, making the feature inconsistent.

### Before
image

### After
image

## Functional changes

- [ ] Added the `CanAllowHtml` trait to `SelectFilter`.
- [ ] Updated the `getFormField` method: Now checks if HTML is allowed using `isHtmlAllowed()`. Calls `$field->allowHtml()` when HTML is enabled.

### DIFF
--- a/packages/tables/src/Filters/SelectFilter.php
+++ b/packages/tables/src/Filters/SelectFilter.php
@@ -3,6 +3,7 @@
 namespace Filament\Tables\Filters;
 
 use Closure;
+use Filament\Forms\Components\Concerns\CanAllowHtml;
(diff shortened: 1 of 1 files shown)


danharrin:
Trying to prevent duplicating every single select method, `->modifyFormFieldUsing(fn (Select $field) => $field->allowHtml())`

PR #15699, opened 2025-02-24
fix: Replace with the fake helper

bzy107:
## Description
Use faker laravel helper

## Functional changes


danharrin:
No thank you, this is slower than using the object that is already loaded

bzy107:
Understood. I was just trying to align it with the other code.
https://github.com/filamentphp/filament/blob/b8dece3c3102e6844b7f6cc8e2bde8569991051e/docs-assets/app/database/factories/UserFactory.php#L22

danharrin:
I would appreciate a PR to make it consistent everywhere

Issue #12554, opened 2024-04-30
When using QueryBuilder with Actions memory exhausts almost instantly


Sprunkas:
### Package

filament/filament

### Package Version

v3.2.72.0

### Laravel Version

v11.5.0

### Livewire Version

v3.4.11

### PHP Version

PHP 8.3.4

### Problem description

When clicking second time "View" in "Products" menu table after QueryBuilder filter is applied I get "Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes)". Even tried to increase memory limit to 2 GB, execution time to 10 min., but still didn't helped. 

### Expected behavior

It should open modal just like it opens first time, without a lot of memory consumption and etc.

### Steps to reproduce

1. Add `Tables\Actions\ViewAction::make()`, in ProductResource table method inside actions([])
2. Open "Products" menu
3. Apply any **QueryBuilder** filter
4. Click "View" button to open modal
5. Close modal
6. Click again "View" button to open modal

### Reproduction repository

https://github.com/Sprunkas/filament-demo

donnyxray:
I experience something very similar:

* I navigate to a resource and view its table
* Apply a filter to it - one that uses querybuilder (this works)
* Navigate elsewhere
* Navigate to the first resource again: memory exhausted error

I then have to "fix" it by manually appending GET parameters to disable the filter.

alexmanase:
I started to investigate the problem and it seems that after pressing "View Action" the query with the selected filter is run over and over. In this way it creates an infinite loop.

https://github.com/filamentphp/filament/assets/10696975/fd313a35-c4eb-410e-9060-25c87e96570e

I will be still looking into the issue.

DonatasJabl:
Any news on this issue? Having save problem.

dmitry-udod:
Really odd bug :(

If we comment out this line, everything works fine.

https://github.com/filamentphp/filament/blob/ca68fd43381e4d9ddcb16c529511dc59bb09b424/packages/forms/src/Concerns/InteractsWithForms.php#L378

But I'm not sure if these changes won't break something.

The code above calls the `getMountedTableActionForm` method.


https://github.com/filamentphp/filament/blob/ca68fd43381e4d9ddcb16c529511dc59bb09b424/packages/tables/src/Concerns/InteractsWithTable.php#L264

After drilling down to the problem, we will found the following code:

[...]

PR #15720, opened 2025-02-25
test: Add querybuilder test

bzy107:
## Description
I added a test to investigate the cause of #12554.

Additionally, I reproduced the test that results in the error from issue #12554, but I haven't included it in this commit. If you plan to investigate, please add the following test to your investigation.

```php
it('can modal actions during text constraint', function () {
    $posts = Post::factory()->count(10)->create();
    $content = $posts->first()->content;
    $post = Post::where('content', $content);

    livewire(PostsQueryBuilderTable::class)
        ->assertCanSeeTableRecords($posts)
        ->queryBuilderTable('content', 'contains', $content)
        ->assertCanSeeTableRecords($post->get())
        ->callTableAction(DeleteAction::class, $post->first());

    assertSoftDeleted($post->first());
});
```
Here are the results of this test:
```bash
   FAILED  Tests\src\Tables\Filters\QueryBuilderTest > it can modal actions during text constraint                                                                                                                                                  Error
  Maximum call stack size of 8339456 bytes (zend.max_allowed_stack_size - zend.reserved_stack_size) reached. Infinite recursion?
```

## Functional changes

### DIFF
--- a/packages/tables/.stubs.php
+++ b/packages/tables/.stubs.php
@@ -177,7 +177,7 @@ namespace Livewire\Features\SupportTesting {
 
         public function assertCanNotSeeTableRecords(array | Collection $records): static {}
 
-        public function queryBuilderTable(string $column, string $operator, $data = null): static {}
+        public function queryBuilderTable(string $filter, string $operator, $data = null): static {}
 
         public function assertCountTableRecords(int $count): static {}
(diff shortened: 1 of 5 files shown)


danharrin:
Replaced by #15884

PR #15728, opened 2025-02-26
Feature: Move affixes

### LINKED ISSUE #14873: Move affixes "inside" input
Make sure clicking prefix/suffix icon focuses input.

thethunderturner:
## Description

Closes #14873 

This seems like the simplest way to "move" the affixes. When clicking on a prefix/suffix, find the nearest container container with ".fi-input-wrp", and focus on the input. Let me know if you think there is a simpler way. 

## Visual changes

none

## Functional changes

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -129,16 +129,16 @@
         },
         {
             "name": "aws/aws-sdk-php",
-            "version": "3.340.0",
+            "version": "3.340.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/aws/aws-sdk-php.git",
-                "reference": "435aa3c9867d11df4ba4d5c3d11c7b1696dc3898"
+                "reference": "22cee29c0ca8d93a7a9c03d13739217373a21fcc"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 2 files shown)


thethunderturner:
For clarity, this only works with `TextInput`. Do we need this to work with other components as well? For example, `Select` and `ColorInput`? If so, when it focuses, do we want the dropdown to open (for Select) and the popup to select color (for ColorPicker)? Then I have to change my approach.

danharrin:
I think Zep might mean that the input size actually increases and the affixes become click-through, so this will likely depend on his review

zepfietje:
> I think Zep might mean that the input size actually increases and the affixes become click-through, so this will likely depend on his review

Yes that's exactly what I meant, though then we might have to calculate the input size (padding to account for affixes) dynamically based on the affix size.

I think the current approach in this PR also has issues when used with affix actions? When action click events bubble up and aren't stopped, they will trigger focus on the input element.

danharrin:
Hi, I am closing this as I think it is quite far away from what Zep was looking for in the original issue

PR #15752, opened 2025-03-01
Refactor: Use outline instead of ring

### LINKED ISSUE #13733: Use outline instead of ring
Switch to native CSS outline for focus states.

thethunderturner:
## Description

Closes #13733

## Visual changes

none

## Functional changes


thethunderturner:
Current issue: There are blue borders when focusing on an input. 
image
In `packages/support/resources/css/components/input/wrapper.css` I found a fix by adding
```css
:focus {
    box-shadow: none !important;
    outline: none !important;
}
```
which in TW would be written as
```css
:focus {
    @apply shadow-none! outline-none!;
}
```
However, the TW version doesnt work, and still shows the blue borders. Here is an overview of the generated css
Using TW:
```css
.fi-input-wrp :focus {
    --tw-shadow: 0 0 #0000;
    --tw-outline-style: none;
    box-shadow: var(--tw-inset-shadow),var(--tw-inset-ring-shadow),var(--tw-ring-offset-shadow),var(--tw-ring-shadow),var(--tw-shadow)!important;
    outline-style: none!important
}
```
Using native css
```css
.fi-input-wrp :focus {
    box-shadow: none!important;
    outline: none!important
}
```

packages/forms/resources/css/components/markdown-editor.css
```diff
@@ -73,11 +73,11 @@
         color: var(--color-cm-gray);
     }
 
-    & .EasyMDEContainer .cm-s-easymde .cm-string {
+    & .EasyMDEContainer .cm-s-easymde .cm-stoutline {
```
saade: This is probably a bad result of a find and replace of the word `ring` (st`ring`).

packages/tables/resources/css/row.css
```diff
@@ -56,7 +56,7 @@
     }
 }
 
-.fi-ta-reordering {
+.fi-ta-reordeoutline {
```
saade: This one too.

saade (changes requested):

packages/forms/resources/css/components/markdown-editor.css
thethunderturner: yeah, I know. Im working on some fixes. Thats why I put the PR as draft

zepfietje:
Part of the migration from ring to outline should be using a consistent outline color.
Currently the primary color is used, which is an option, or we default to the browser's accent color as per the default behavior.

This improves current cases of the danger color being used for inputs with validation errors, which in hindsight is suboptimal.

thethunderturner:
> Part of the migration from ring to outline should be using a consistent outline color. Currently the primary color is used, which is an option, or we default to the browser's accent color as per the default behavior.
> 
> This improves current cases of the danger color being used for inputs with validation errors, which in hindsight is suboptimal.

Which one do we want to use? primary or browsers default? I think primary should work better because browsers could potentially have different default colors. Let me know if I misunderstood something!

danharrin:
I am going to close this as I think there are considerable further steps that need to be taken in order for it to fulfil the intent of the original issue

PR #15753, opened 2025-03-01
feat: add getCurrentPage method to Filament facade

danielgnh:
## Description

This Pull Request is adding new `getCurrentPage()` method to the Filament facade to allow easy **access to the current page** class from anywhere in the application. Really useful for determining the current page context, when working with multiple pages and implementing **conditional logic based on the active page**.

## Implementation Details

- Added `getCurrentPage()` method to `FilamentManager` that delegates to the current panel
- Method returns `?string` representing the current page class or null if none found
- Maintains type safety with nullable return type

## Example Usage

```php
  
public function getPage(): ?string
{
    return Filament::getCurrentPage();
}

### DIFF
--- a/packages/panels/src/FilamentManager.php
+++ b/packages/panels/src/FilamentManager.php
@@ -284,6 +284,14 @@ class FilamentManager
         return $this->getCurrentPanel()->getPages();
     }
 
+    /**
+     * @return string | null
+     */
+    public function getCurrentPage(): ?string
+    {
+        return $this->getCurrentPanel()?->getCurrentPage();
+    }
+
     public function getPanel(?string $id = null, bool $isStrict = true): Panel
     {
         return app(PanelRegistry::class)->get($id, $isStrict);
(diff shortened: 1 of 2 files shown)


danharrin:

Can you please explain where you need to be able to access the current page within your codebase?

danielgnh:
> Hi, apologies but we are no longer accepting new features for v3 as per #13744. Also, please check out our [contributing guide](https://filamentphp.com/docs/3.x/support/contributing#development-of-new-features) which talks about proposing the feature before submitting the PR, to ensure you aren't spending time building something that we won't merge.
> 
> 
> 
> Can you please explain where you need to be able to access the current page within your codebase?

Hi @danharrin, it would be nice to have a method for accessing the active page, especially from middleware, since this is a real use case for me. 

With multi-level permissions for pages across different panels, I need to check user access for each page in the application. 

Previously, I could have used a trait to override the canAccess method for pages. However, with the additional steps I now require, maintaining a single point of control for all access checks is essential. This method offers a nice, clean, and simple way to implement it.

danharrin:
Thanks for the extra detail!

I'm afraid that I haven't had many other requests for this yet, so I don't think it's something I would accept as a feature of the framework. That could change in the future based on the velocity of requests :) Fortunately, this is a helper you can add into your project or a package.

PR #15759, opened 2025-03-03
Enhancement: Add Closure support to HasTopbar trait

fadlee:
## Description

I've modified the `HasTopbar` trait to accept a Closure as a parameter, similar to other conditional features in Filament. 

The changes include:
1. Adding the `use Closure` statement
2. Updating the property and method type hints to support both `bool` and `Closure`
3. Using the `evaluate()` method to properly handle the Closure when checking if the topbar should be displayed

### Motivation

While implementing Filament in our application, we needed to dynamically control the visibility of the topbar based on the authenticated user's role. The current implementation only supports boolean values, which limits the ability to make runtime decisions about topbar visibility.

## Visual changes

No changes.

## Functional changes

- [*] Changes have been tested to not break existing functionality.

### DIFF
--- a/packages/panels/src/Panel/Concerns/HasTopbar.php
+++ b/packages/panels/src/Panel/Concerns/HasTopbar.php
@@ -2,11 +2,13 @@
 
 namespace Filament\Panel\Concerns;
 
+use Closure;
+
 trait HasTopbar
 {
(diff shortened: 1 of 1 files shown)


mgkimsal:
@fadlee  - thanks!

PR #15785, opened 2025-03-05
Filament session key improvement

NankoPrinzhorn:
## Description

In some scenarios, the default behavior of the table session, which keeps track of enabled and disabled columns, does not work as expected.

### Issue  
In our use case, we have multiple Livewire components in different directories with the same class name, for example:  

- `App\Livewire\Admin\ManualDeployment\Index.php`  
- `App\Livewire\Admin\Users\Index.php`  
- `App\Livewire\Admin\ExposureRequests\Index.php`  

Each of these classes has its own Filament table and uses the `InteractsWithForms` trait. However, they all share the same session key from the `getTableColumnToggleFormStateSessionKey` function.  

#### Cause  
This happens because `class_basename($this::class);` returns `"Index"` for all three classes. As a result, when an ID column is marked as `->toggable()`, its visibility state is shared across all tables.  

For example:  
- Enabling the ID column in one table will also enable it in the other two tables.  
- Disabling the ID column in one table will hide it in all tables.  

This unintended behavior causes inconsistencies when toggling columns across different tables.  

---

## Solution  

To resolve this, the session key should be based on the fully qualified class name rather than just the basename. However, to prevent overly long session keys—especially in projects with deeply nested directories—we apply hashing.  

### Previous Implementation  
```php
$table = class_basename($this::class); // -> "Index"
```

### Updated Implementation
```php
$table = hash('sha256', $this::class); // -> "619b97a8ad39037aa755dfc64a1fc00782020e23a7cd4013c8209cd29afe534b"
```
By using a SHA-256 hash of the full class name, we ensure that each table gets a unique session key while keeping it manageable in length.

## Functional Changes  

- [x] Code style has been fixed using `composer cs`.  
- [x] Changes have been tested to ensure no existing functionality breaks.  
- [x] Documentation is up to date.  

Let me know if any additional details are needed! 😊

### DIFF
--- a/packages/panels/src/Pages/Dashboard/Concerns/HasFilters.php
+++ b/packages/panels/src/Pages/Dashboard/Concerns/HasFilters.php
@@ -78,7 +78,7 @@ trait HasFilters
 
     public function getFiltersSessionKey(): string
     {
-        $livewire = class_basename($this::class);
+        $livewire = hash('sha256', $this::class);
 
         return "{$livewire}_filters";
     }
--- a/packages/panels/src/Pages/Dashboard/Concerns/HasFilters.php
+++ b/packages/panels/src/Pages/Dashboard/Concerns/HasFilters.php
@@ -78,7 +78,7 @@ trait HasFilters
 
     public function getFiltersSessionKey(): string
     {
-        $livewire = hash('sha256', $this::class);
+        $livewire = hash('md5', $this::class);
 
         return "{$livewire}_filters";
     }
--- a/packages/tables/src/Concerns/CanPaginateRecords.php
+++ b/packages/tables/src/Concerns/CanPaginateRecords.php
@@ -87,7 +87,7 @@ trait CanPaginateRecords
 
     public function getTablePerPageSessionKey(): string
     {
-        $table = hash('sha256', $this::class);
+        $table = hash('md5', $this::class);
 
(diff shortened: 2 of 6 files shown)


danharrin (changes requested):
There are many similar `SessionKey(): string` methods, can you update them please to be consistent?

danharrin (changes requested):
Also, since this is not sensitive data, lets use `MD5` instead of SHA256 so the hashes are shorter

danharrin:
Thanks

NankoPrinzhorn:
Thanks for the merge @danharrin, could you create a new tag with the latest changes maybe? Much appreciate

danharrin:
In the meantime before the next release, you can define these methods on the Livewire classes in your app to set this up.

jmwamser:
Just wanted to follow up here. I like the change, BUT I see this as a BC break.

I had a few widgets that would look up the session value so they could maintain the same filter. Without a direct way to have the widget get the key I had to copy this logic to the widget. 

Is there a better way I could do this? or could the method be static?

danharrin:
Making the method static would be a breaking change unfortunately. Maybe you could create a new static `::getTableFiltersDataFromSession()` method on the class in your project, so the logic doesn't leak out of there? Or you could override the method we changed to have a more predictable key, or even return the old logic?

PR #15792, opened 2025-03-06
fix: handle a possible issue from output buffer

gillobis:
## Description

When exporting an xlsx file, sometimes the output buffer is not cleaned up properly, generating an inaccessible  excel file when downloaded. This is caused by an extra space that is added to the file. 
Calling the ob_end_clean() function before sending the download response fix this problem.

## Functional changes

### DIFF
--- a/packages/actions/src/Exports/Downloaders/XlsxDownloader.php
+++ b/packages/actions/src/Exports/Downloaders/XlsxDownloader.php
@@ -26,6 +26,7 @@ class XlsxDownloader implements Downloader
         if ($disk->exists($filePath = $directory . DIRECTORY_SEPARATOR . $fileName)) {
             $response = $disk->download($filePath);
             ob_end_clean();
+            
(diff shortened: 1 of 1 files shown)


packages/actions/src/Exports/Downloaders/XlsxDownloader.php
```diff
@@ -24,7 +24,9 @@ public function __invoke(Export $export): StreamedResponse
         $fileName = $export->file_name . '.xlsx';
 
         if ($disk->exists($filePath = $directory . DIRECTORY_SEPARATOR . $fileName)) {
-            return $disk->download($filePath);
+            $response = $disk->download($filePath);
+            ob_end_clean();
+            return $response;
```
danharrin, line 30:
```suggestion
            
            return $response;
```

danharrin (changes requested):

PR #15793, opened 2025-03-06
fix: Update tailwind.config.preset.js darkMode selector

mgkimsal:
## Description

Tailwind v 3.4.1 changed its darkMode selector behavior.  The Filament tailwind.config.preset doesn't work with the base.blade.php themeing behaviour as expected.

https://github.com/tailwindlabs/tailwindcss/pull/12717 notes the new 'selector' mode.

The Filament JS in base.blade.php sets a 'dark' class on the top document element, and the tailwind.config 'selector' behaviour will look for that 'dark' class to determine if dark mode variants will be used, as opposed to the system-specified preferences.  

The core issue was having a system-specified 'dark model' would not allow the 'light' theme to work.  This *may* only have affected custom themes - unsure about that.

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/support/tailwind.config.preset.js
+++ b/packages/support/tailwind.config.preset.js
@@ -2,7 +2,7 @@ import forms from '@tailwindcss/forms'
 import typography from '@tailwindcss/typography'
 
 export default {
-    darkMode: 'class',
+    darkMode: 'selector',
     content: ['./vendor/filament/**/*.blade.php'],
     theme: {
         extend: {


danharrin:
The PR you linked reverts the breaking change, so I am unsure what the problem is? I have not seen any other reports of problems myself

danharrin:
@mgkimsal please clarify the above if I am wrong, but closing for now.

PR #15822, opened 2025-03-10
Add Global Controller Registration Support via `filamentChartJsRegister`

BukkitAT:
## Description

This pull request introduces a mechanism for globally registering Chart.js controllers and elements using a new global array: `window.filamentChartJsRegister`. Currently, Filament's chart implementation supports adding additional plugins via `window.filamentChartJsPlugins` passed in the chart configuration. However, this method does not support registering new controllers—such as those required for treemap charts—which must be registered globally using `Chart.register()`. 

With this change, before initializing a chart, the code checks if `window.filamentChartJsRegister` is defined and contains any entries. If so, it registers these controllers/elements globally via `Chart.register(...window.filamentChartJsRegister)`. This ensures that custom controllers, which cannot be registered via the plugins array, are available for use without affecting existing functionality.

## Visual changes

_No visual changes are expected. The modification only impacts the chart initialization process to support additional controller registration. Charts that rely on custom controllers (e.g., treemap charts) will now render correctly._

## Functional changes


danharrin:

In v4, I am considering building a new system for asynchronously loading plugins for Chart.js (and the rich editor). In this implementation, it will load the JS on pages even without charts.

PR #15851, opened 2025-03-13
Allows the use of multiple afterStateHydrated hooks.

xGrz:
Allows the use of multiple afterStateHydrated hooks.

When using `formatStateUsing` in a form field, you cannot later use `afterStateHydrated` for other purposes with advanced form fields (it overwrites formatting).

This version allows the use of multiple afterStateHydrated hooks in the order they were added to the field.


danharrin:

Unfortunately, this is also a breaking change. It does not allow users to customize the existing `afterStateHydrated()` hook by overriding it, since the new hook will be run after the old one instead of replacing it.

Can you please describe the use case you have for this feature?

PR #15859, opened 2025-03-14
Update HasBody trait's body() to accept HtmlString

timsinakiran:
## Description

The ->body() method already parses HtmlString but the function's param type is limited to string only.

## Visual changes

This is a screenshot of the notification body parsing HtmlString.

Screenshot 2025-03-14 at 13 25 03

## Functional changes


danharrin:
There is no difference, you can pass a string containing HTML (in this case). It cannot truly be treated as an HtmlString since it needs to be serialised to JSON for sending

PR #15860, opened 2025-03-14
Configurable utility route prefix

binaryfire:
I'm running into a similar issue to this: https://github.com/filamentphp/filament/discussions/11494

If the Laravel app is in a subdirectory, or in my case if only a specific path like `/app` is reverse proxied to Laravel app, then all routes need to be under that path. This PR allows customising the prefix for "utility routes" added by filament to something other than `/filament`.

This is a problem with other packages too eg. Livewire:
- https://github.com/livewire/livewire/discussions/7480
- https://github.com/livewire/livewire/pull/8752#issuecomment-2542826198

It's also useful for branding in case you want to keep urls generic eg. `/system/export`


danharrin (changes requested):
Can we call these "system" or "internal" routes instead? What do you think?

binaryfire:
Yeah sure. System sounds better to me but either works. Which do you prefer?

danharrin:
Yeah let's do system then

binaryfire:
> Yeah let's do system then

Done

danharrin (approved):

PR #15866, opened 2025-03-14
Add optional navigate param to action redirect

binaryfire:
Setting `spaUrlExceptions` for each action redirect you want to exclude gets pretty clunky. This PR allows us to pass a second optional parameter to `->redirect()` to manually set navigate on individual actions.

This also makes `$action->redirect()` consistent with the Livewire `redirect()` method signature.


danharrin (changes requested):
I've just noticed that in most places in the codebase, we aren't passing the URL to `hasSpaMode()`, so the list of exceptions doesn't apply. Can we add that here? Also could you go through the rest of the codebase and find places where we use `FilamentView::hasSpaMode()` without passing the URL?

When passing the URL, you don't need to do `is_app_url()` after since it is already included in that check

binaryfire:
No problems. But do you mean in lieu of this, or in addition to it? It'd still be good to be able to set the redirect behaviour per-action like with Livewire's `redirect()`.

Might be cleaner to make a separate PR to fix passing the URL to `hasSpaMode()` across the whole codebase.

danharrin:
Yeah it would be an addition, but would make the default behaviour more consistent. I just made a release so we have time to just add it in this PR

binaryfire:
@danharrin I've updated the codebase to pass the url to every occurrence of `hasSpaMode()`.  `is_app_url` is only used in one place now. Do you want me to remove the helper and move the logic directly into hasSpaMode()? Or should we leave it in case it's useful for something in the future?

danharrin (approved):

PR #15888, opened 2025-03-18
Add missing focus state highlighting in Toggle component

remiebelingmerifond:
## Description
Moving Focus to a Toggle Input by tabbing is possible, but does not come with any visual Feedback at all.

This Behavior can be observerd in the Demo Application (Status Section => Visible Toggle):
https://demo.filamentphp.com/shop/products/products/1/edit

## Visual changes

**Before (light):**
image

**After (light)**
image

**Before (Dark)**
image

**After (Dark).**
image

## Functional changes
No functional Changes.
Only one added CSS-Class

### DIFF
--- a/packages/forms/resources/views/components/toggle.blade.php
+++ b/packages/forms/resources/views/components/toggle.blade.php
@@ -68,7 +68,7 @@
                     ], escape: false)
                     ->merge($getExtraAttributes(), escape: false)
                     ->merge($getExtraAlpineAttributes(), escape: false)
-                    ->class(['fi-fo-toggle relative inline-flex h-6 w-11 shrink-0 cursor-pointer rounded-full border-2 border-transparent outline-none transition-colors duration-200 ease-in-out disabled:pointer-events-none disabled:opacity-70 focus-visible:ring-primary-500 focus-visible:ring-2'])
+                    ->class(['fi-fo-toggle relative inline-flex h-6 w-11 shrink-0 cursor-pointer rounded-full border-2 border-transparent outline-none transition-colors duration-200 ease-in-out disabled:pointer-events-none disabled:opacity-70 focus-visible:ring-primary-500 focus-visible:ring-offset-1 focus-visible:ring-2'])
             }}
         >
             <span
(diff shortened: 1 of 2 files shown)


danharrin (changes requested):
Can you please check all the other inputs, I think we use a ring there and not an outline. Please also look for other toggle components, like the toggle table column

remiebelingmerifond:
You were right—I aligned the focus style with the input wrapper component.  
The toggle now uses a ring as well.  

I also checked if there are other toggle components.  
As far as I can tell, the only one is the toggle column.  

So, I tried to apply the focus indication there as well.  
However, that escalated a bit.  

It turned out that the toggle column wasn’t keyboard-accessible at all yet.  

I’m completely new to Alpine, and Livewire, but I gave it a shot to fix that anyway.  
I hope my approach isn’t completely off.  

It seems to work, but I am really not sure if it follows best practices.  

There’s still one issue I haven’t fixed yet, because every solution I came up with felt really hacky.  

When you toggle a switch in a toggle column, using space, the UI updates.  
Since the toggle has the key:  

```blade
wire:key="{{ $this->getId() }}.table.record.{{ $recordKey }}.column.{{ $getName() }}.toggle-column.{{ $state ? 'true' : 'false' }}"
```  
the element is replaced, causing it to lose focus.  

I messed around with preserving the focus state as a Livewire property and refocusing in `x-init` of the switch if needed.  
But as i said before this did not feel to be the right way and kinda hacky.

This Problem appears on regular toggles that are "live" as well.

I'm not really sure how to proceed from here.

I could either check how other components handle this issue and try to apply a similar approach to toggles in a separate PR to keep the scope small.

Or I could try to find a solution and add it to this PR.

If, based on the code you've seen in this issue, you’d prefer that I leave this to more experienced Alpine developers, I can simply create an issue and step back—which would be absolutely fine with me, since my only intention was to add a simple outline on focus originally. 😄

danharrin:
Yeah can we please remove the changes from that column file, it is quite fragile to change

remiebelingmerifond:
done

danharrin (approved):

PR #15889, opened 2025-03-18
Prevent catching side-effects of database transactions being committed

ralphjsmit:
This PR prevents Filament from catching any exceptions resulting in side-effects _after an already committed transaction_. Previously, these side-effects were still part of the `try/catch` blocks and therefore their exceptions could be catched, and triggering an extra rollback. That would cause the following flow to happen: `begin transaction` + `commit` + `rollback`.

I have also added a test in which you can observe the behavior.

### DIFF
--- a/tests/src/Panels/Resources/Pages/ListRecordsTest.php
+++ b/tests/src/Panels/Resources/Pages/ListRecordsTest.php
@@ -117,7 +117,7 @@ it('can bulk delete posts', function () {
     }
 });
 
-it('actions will not interfere with database transactions on an error', function () {
+test('actions will not interfere with database transactions on an error', function () {
     $post = Post::factory()->create();
 
     $transactionLevel = DB::transactionLevel();
@@ -135,7 +135,7 @@ it('actions will not interfere with database transactions on an error', function
         ->toBe($transactionLevel);
 });
 
-it('table actions will not interfere with database transactions on an error', function () {
+test('table actions will not interfere with database transactions on an error', function () {
     $post = Post::factory()->create();
 
     $transactionLevel = DB::transactionLevel();
--- a/tests/src/Panels/Resources/Pages/ListRecordsTest.php
+++ b/tests/src/Panels/Resources/Pages/ListRecordsTest.php
@@ -143,7 +143,7 @@ test('table actions will not interfere with database transactions on an error', 
     $transactionLevel = DB::transactionLevel();
 
     try {
-        livewire(PostResource\Pages\ListPosts::class)
+        livewire(ListPosts::class)
             ->callTableAction('randomize_title', $post);
     } catch (Exception $e) {
         // This can be catched and handled somewhere else, code continues...
--- a/tests/src/Panels/Resources/Pages/ListRecordsTest.php
+++ b/tests/src/Panels/Resources/Pages/ListRecordsTest.php
@@ -131,6 +131,8 @@ test('actions will not interfere with database transactions on an error', functi
         // This can be catched and handled somewhere else, code continues...
     }
 
+    // Original transaction level should be unaffected...
+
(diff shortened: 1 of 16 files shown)


ralphjsmit:
The PHPStan issue in Laravel 12 suddenly appears here:

```
 129    Comparison operation ">" between int<1, max> and 0 is always true.  
         🪪  greater.alwaysTrue                                              
         💡 Because the type is coming from a PHPDoc, you can turn off this  
            check by setting treatPhpDocTypesAsCertain: false in your        
            phpstan.neon.dist.          
```

It seems to be introduced by https://github.com/laravel/framework/pull/55044, causing the return time of the `->sum()` method to be `int<1, max>` instead of `int<0, max>`. I don't think the PR looks strange, so it appears that PHPStan for some reason understands that the closure in this count returns `<1, max>` instead of `<0, max>`. I already tried making this a ternary like `$file['removed'] ? count($file['removed']) : 0` to indicate that this closure can also return `0` but that did not have any effect.

```php
$removedKeysCount = $files->sum(fn ($file): int => count($file['removed']));
```

liamduckett:
Hey, I might be able to help here - I'm unfamiliar with Filament (although I've heard great things!) but the author of that Laravel PR.

The problemed section of code:
```php
$missingKeysCount = $files->sum(fn ($file): int => count($file['missing']));
$removedKeysCount = $files->sum(fn ($file): int => count($file['removed']));

if ($missingKeysCount == 0 && $removedKeysCount == 0) {
    info("[✓] Foo!\n");
}
elseif ($missingKeysCount > 0 && $removedKeysCount > 0) {
    warning("[!] Bar\n");
}
elseif ($missingKeysCount > 0) {
    warning("[!] Baz.\n");
}
// For us to get this far:
// 1. They aren't both 0
// 2. They aren't both +
// 3. $missingKeysCount is -
// By combining 2 and 3, we therefore know that $removedKeysCount must be +
// "must be +" is equivalent to int<1, max>
elseif ($removedKeysCount > 0) {
    warning("[!] Qux.\n");
}
```

[PHPStan playground equivalent](https://phpstan.org/r/3e98aa37-735f-456b-bd71-47eb2c5ae7df)

Consider we have 4 possible scenarios:
1. `$a` is positive, `$b` is positive
2. `$a` is positive, `$b` is zero
3. `$a` is zero, `$b` is positive
4. `$a` is zero, `$b` is zero

1 is ruled out by the second condition (the first `elseif`)
2 is ruled out by the third condition (the second `elseif`)
4 is ruled out by the first condition (the `if`)

Therefore by the time we get to the fourth condition (the third `elseif`) we must have scenario 3, in which `$b` is definitely positive - and therefore it will always be `true` - and can be safely replaced with an `else`.

I'd have never caught that manually - another reason I love static analysis!

This isn't related to your PR - should probably be resolved (if desired) in another branch.

ralphjsmit:
@liamduckett Thanks for your message! Wow I had to read this two times, but you are right indeed. In this case I would argue that the `} elseif ($removedKeysCount > 0) {` still provides value from a DX/simplicity perspective, but for static analysis it is then the same as `} else {`. I didn't look at the previous if-conditions but I jumped straight to the `->sum()` thing 😄

PS: definitely give it a try, I'm sure you will love it!

tests/src/Panels/Resources/Pages/ListRecordsTest.php
danharrin, line 1: Can you move the EditRecord test into the EditRecordTest please

danharrin (changes requested):

tests/src/Panels/Resources/Pages/ListRecordsTest.php
ralphjsmit: @danharrin Good point, just moved.

PR #15890, opened 2025-03-18
Render table action arguments into Livewire click handler

ralphjsmit:
I noticed that table actions do not include their arguments rendered into the Livewire click handler. This can be an issue if you provide a custom `->modalContent()` inside a table action, and then manually render a table action in that content that needs arguments. Arguments are normally supported to table actions.

Thanks!

### DIFF
--- a/packages/tables/src/Actions/Action.php
+++ b/packages/tables/src/Actions/Action.php
@@ -36,7 +36,11 @@ class Action extends MountableAction implements Groupable, HasRecord, HasTable
             return null;
         }
 
-        return $this->generateJavaScriptClickHandler('mountAction');
+        $arguments = $this->getArguments();
+
+        return $this->generateJavaScriptClickHandler('mountTableAction', [
+            ...$arguments ? [$arguments] : [],
+        ]);
     }
 
     public function getLivewireTarget(): ?string
@@ -49,7 +53,11 @@ class Action extends MountableAction implements Groupable, HasRecord, HasTable
             return null;
         }
 
-        return $this->generateJavaScriptClickHandler('mountTableAction');
+        $arguments = $this->getArguments();
+
+        return $this->generateJavaScriptClickHandler('mountTableAction', [
+            ...$arguments ? [$arguments] : [],
+        ]);
     }
 
     public function getLivewireClickHandler(): ?string
--- a/packages/tables/src/Actions/Action.php
+++ b/packages/tables/src/Actions/Action.php
@@ -32,7 +32,7 @@ class Action extends MountableAction implements Groupable, HasRecord, HasTable
             return $handler;
         }
 
-        if (!$this->canAccessSelectedRecords()) {
+        if (! $this->canAccessSelectedRecords()) {
             return null;
         }
 
@@ -45,7 +45,7 @@ class Action extends MountableAction implements Groupable, HasRecord, HasTable
             return $target;
         }
 
-        if (!$this->canAccessSelectedRecords()) {
+        if (! $this->canAccessSelectedRecords()) {
             return null;
         }
 
@@ -54,7 +54,7 @@ class Action extends MountableAction implements Groupable, HasRecord, HasTable
 
     public function getLivewireClickHandler(): ?string
     {
-        if (!$this->isLivewireClickHandlerEnabled()) {
+        if (! $this->isLivewireClickHandlerEnabled()) {
             return null;
         }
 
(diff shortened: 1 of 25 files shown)


danharrin (changes requested):
I'm not sure I understand and this doesn't look right to me, the record ID is now no longer passed in

ralphjsmit:
@danharrin Hmm you are right, I didn't look into the `generateJavaScriptClickHandler()` code, and if there is no record the test passed. I will refactor this to include record.

ralphjsmit:
Regarding use case, consider the following:

```php
$table
   ->actions([
       Tables\Actions\ViewAction::make()
          ->modalContent(fn (Model $record, Tables\Actions\ViewAction $action) => view('some-view', ['record' => $record, 'action' => $action]))
           ->registerModalActions([
                Tables\Actions\Action::make('view_some_relationship')
                    ->slideOver()
                    // Uses `$arguments` to determine the exact `SomeRelationship` model...
            ])
    ])
```
And then the `some-view`:
```blade
<div>
   @foreach ($record->someRelationship as $someRelationship)
      {{ $action->getModalAction('view_some_relationship')(['some_relationship_id' => $someRelationship->getKey()])
   @endforeach
</div>
```

danharrin:
I think `generateJavaScriptClickHandler` needs the changes

Also please check the HasTable interface, I think the `mountTableAction` method is out of date with InteractsWithTable

packages/tables/src/Actions/Action.php
```diff
@@ -35,7 +36,11 @@ public function getAlpineClickHandler(): ?string
             return null;
         }
 
-        return $this->generateJavaScriptClickHandler('mountAction');
```
ralphjsmit, line 38: My eye fell upon this in the `getAlpineClickHandler()`, this is inside a table action. Is it correct that this should be `mountAction()` indeed, or should this be `mountTableAction()` too?

packages/tables/src/Actions/Action.php
```diff
@@ -69,18 +78,38 @@ public function getLivewireClickHandler(): ?string
...
+        return $this->generateJavaScriptClickHandler('mountTableAction', [
+            ...$arguments ? [$arguments] : [],
+        ]);
     }
 
-    protected function generateJavaScriptClickHandler(string $method): string
+    /**
+     * @param  array<array-key, array<mixed, mixed>>  $parameters
+     */
+    protected function generateJavaScriptClickHandler(string $method, array $parameters = []): string
```
danharrin, line 91: Why do we need to pass the params in, can we not fetch them inside the method? Like we do for the main Action class?

ralphjsmit: Yes we can, but in theory you can also pass in a different method then `mountTableAction`, and we wouldn't be sure if such other method would also accept `$arguments`. 

Though to be fair that also applies to the `$record` parameter, if you pass in a different method into `generateJavaScriptClickHandler()` then Filament also cannot be sure that the 2nd parameter is `$record`.

packages/tables/src/Actions/Action.php
danharrin: I think this should be `mountTableAction`

danharrin: But potentially it needs to be `$wire.`? What is is like in the main Action class?

packages/tables/src/Actions/Action.php
ralphjsmit: Another option would be to rename `generateJavaScriptClickHandler()` into something like `getJavascriptMountTableActionClickHandler()` to make it more explicit that this method is *only* intended to be used for `mountTableAction()`.

danharrin: The method for the Livewire and JavaScript click handlers will always have the same signature, I thought the difference was that one was prefixed with `$wire.`...

packages/tables/src/Actions/Action.php
ralphjsmit: In the `Actions\StaticAction` it is `close()`, and for notification actions it is `markAsRead()` or `markAsUnread()`.

danharrin:
Sorry, I'm going to close this, I'm worried about breaking changes and this is for sure addressed in v4 where everything is unified.

ralphjsmit:
Ok, you can easily run the tests and compare with before/after, but then I will work around this in a different way for now, and change it in V4 👍

danharrin:
Yeah you can always call the `mountTableAction` in your own code if you want

PR #15909, opened 2025-03-20
In a wizard, prevent users from pressing the 'Next' button rapidly and skipping steps while the live form is still loading

arshaviras:
## Description

In a Wizard page preventing to press Next button rapidly and skipping steps when in form there is live() form.
Without this user can click next button rapidly and skip mandatory steps without validation.

## Visual changes

Disabling Next button when live() is working

## Functional changes

No change

### DIFF
--- a/packages/forms/resources/views/components/wizard.blade.php
+++ b/packages/forms/resources/views/components/wizard.blade.php
@@ -293,6 +293,7 @@
                 "
             @endif
             x-bind:class="{ 'hidden': isLastStep(), 'block': ! isLastStep() }"
+            wire:loading.class="opacity-50 pointer-events-none"
         >
             {{ $nextAction }}
         </span>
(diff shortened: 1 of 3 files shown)


danharrin:
Is there a `disabled` attribute on the button already when it is loading? I think that is the default behaviour of Livewire, anyway

arshaviras:
Yes in a button there is `wire:loading.attr="disabled"`, but when ->live() is working button not disabling.

For example without ->live() is in progress pressing Next button rapidly, button again not disabling, but steps are not skipped, it will go to second step as expected.
Problem only persist when in the form some of element ->live is in progress (for example 3 sec), and user pressing Next rapidly (without waiting to finish ->live finishing), than how many time user clicked, that many times will be skipped steps.

In this PR I just add class to span `"opacity-50 pointer-events-none"` to prevent user clicking button when ->live is in process. Maybe you will have more right approach to preventing this problem.

Thanks

danharrin:
I have removed the opacity change, as when changing step the opacity then becomes half of what it should be

arshaviras:
yes when changing step, but in this case not clicking next, it just loader working, opacity will indicate user that button is disabled.
Overall do what you think is right.
Thanks

danharrin:
Ah I see that there isn't a loading opacity at the moment

danharrin (approved):

PR #15871, opened 2025-03-16
Widget Command Livewire Path Solution

ksaif534:
Hello.

This is a solution to this Github issue: [https://github.com/filamentphp/filament/issues/14637](https://github.com/filamentphp/filament/pull/url)

### **Description**

When the make:filament-widget command is run to create a new widget and the resource is provided and when the user selects [App\Livewire] alongside other Livewire components after selecting Table from the first prompt, a file permission error was thrown and the directory was not being created. The solution now creates a new file TestWidget.php within the app/Livewire directory when it's prompted by the user.

### **Visual Changes**

**After the Solution:**

widget-command-livewire-solution

**Before the Solution:**

widget-command-livewire-solution-before

### **Functional Changes**

I've fixed the style with the composer cs command, the file changes have been tested.

Issue #14637, opened 2024-10-27
WidgetCreateCommand - throws exception when resource provided to command


samapraku:
### Package

filament/widgets

### Package Version

v3.2.119

### Laravel Version

v10.48.22

### PHP Version

PHP 8.3.0

### Problem description

When I create a widget with make:filament-widget and I provide a resource to the command and select where to create the widget -  [App\Livewire] alongside other Livewire components, the command exits with the exceptions below

```
artisan make:filament-widget TestWidget --resource=InvoiceResource

 ┌ What type of widget do you want to create? ──────────────────┐
 │ Table                                                        │
 

 ┌ Where would you like to create this? ────────────────────────┐
 │ [App\Livewire] alongside other Livewire components           │

```

Exception  
```
InvoiceResource
   ErrorException 

  mkdir(): Permission denied

  at vendor/laravel/framework/src/Illuminate/Filesystem/Filesystem.php:650
    646▕     {
    647▕         if ($force) {
    648▕             return @mkdir($path, $mode, $recursive);
    649▕         }
  ➜ 650▕         return mkdir($path, $mode, $recursive);
    651▕     }
    652▕ 
    653▕     /**
    654▕      * Move a directory.

      +17 vendor frames 

  18  artisan:35
      Illuminate\Foundation\Console\Kernel::handle()

```

On a newly created project, the exception below is thrown
```
   ErrorException 

  mkdir(): Read-only file system

```

### Expected behavior

The command should create a widget successfully when the resource is provided and create alongside other livewire components option is selected.

### Steps to reproduce

1. Execute the command and pass a resource

`artisan make:filament-widget TestWidget --resource=InvoiceResource`

2. Select `Table` as widget type
3. Where would you like to create this?  -  `[App\Livewire] alongside other Livewire components `

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/samapraku/filament-widget-command-bug

### Relevant log output

```shell
ErrorException 

  mkdir(): Read-only file system

  at vendor/laravel/framework/src/Illuminate/Filesystem/Filesystem.php:651
    647▕         if ($force) {
    648▕             return @mkdir($path, $mode, $recursive);
    649▕         }
    650▕ 
  ➜ 651▕         return mkdir($path, $mode, $recursive);
    652▕     }
    653▕ 
    654▕     /**
    655▕      * Move a directory.

      +18 vendor frames 

  19  artisan:13
[...]

PR #15910, opened 2025-03-20
Widget command livewire path solution(adjustment)

ksaif534:
Hello.

This is a modification/adjustment to my PR Solution [https://github.com/filamentphp/filament/pull/15871](url) to the issue [https://github.com/filamentphp/filament/issues/14637](url)

### **Description**

When the `make:filament-widget` command is run to create a new widget and the resource is provided and when the user selects `[App\Livewire] alongside other Livewire components` after selecting `Table` from the first prompt, a file permission error was thrown and the directory was not being created. The solution now creates a new file `TestWidget.php` within the `app/Livewire` directory when it's prompted by the user. I've fixed the problems in the initial solution by modifying the `MakeWidgetCommand.php` and also the test `WidgetCommandLivewireTest.php` according to the `v4` coding style.

### **Visual Changes:**

**Before the Solution:**

filament-widget-livewire-path-issue

### **After the Solution:**

filament-widget-livewire-path-solution

### **Functional Changes:**

I've fixed the style with the composer cs command, the file changes have been tested.

### DIFF
--- a/packages/widgets/src/Commands/MakeWidgetCommand.php
+++ b/packages/widgets/src/Commands/MakeWidgetCommand.php
@@ -163,7 +163,11 @@ class MakeWidgetCommand extends Command
 
         $path = (string) str($widget)
             ->prepend('/')
-            ->prepend($resource === null ? $path : "{$resourcePath}\\{$resource}\\Widgets\\")
+            ->prepend(
+                ! $panel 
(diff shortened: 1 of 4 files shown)


danharrin:
I have decided this fix is too complex to maintain, and it should be fixed already in v4. These options in v3 are incompatible, it can't be in a resource and also have a class outside the resource's directory. You could also just move the class after it is created.

PR #15915, opened 2025-03-21
[2.x]: Upgrade DOMPurify

danharrin:
There is a security issue with earlier versions of this dependency.


karan218:
this PR create lots off issue on filament we are not able to open any popup or anything it's update alpine version to 3.14.9 which breaks everything...please revert it asap

danharrin:
Please try v2.17.58

PR #15920, opened 2025-03-21
Add support for page-specific relation managers with policy enforcement

andrewdwallo:
## Description
This PR introduces a new `getRelations()` method for ViewRecord or EditRecord pages that allows developers to specify page-specific relation managers while maintaining proper policy enforcement.

## Current Situation
Currently, developers who want to customize relation managers for specific ViewRecord pages often override the `getRelationManagers()` method directly. While this works functionally, it inadvertently bypasses the policy checks from `canViewForRecord()` that would normally be applied.

## Enhancement
This PR adds:
- A new protected `getRelations()` method
- Updated logic in `getRelationManagers()` to use relation managers from `getRelations()` when available, while still applying proper policy filters

## Benefits
- Provides a documented, official way to customize relation managers per view page
- Ensures all relation managers respect visibility policies
- Maintains backward compatibility
- Improves security by preventing accidental policy bypasses

## Usage Example
```php
class ViewProject extends ViewRecord
{
    protected function getRelations(): array
    {
        return [
            ProjectUsersRelationManager::class,
            ProjectTasksRelationManager::class,
        ];
    }
}
```

## Migration for existing code
Developers who were previously overriding `getRelationManagers()` directly should update their code to use `getRelations()` instead to benefit from proper policy enforcement.

## Functional changes

### DIFF
--- a/packages/panels/docs/03-resources/05-viewing-records.md
+++ b/packages/panels/docs/03-resources/05-viewing-records.md
@@ -1,7 +1,6 @@
 ---
 title: Viewing records
 ---
-
 import LaracastsBanner from "@components/LaracastsBanner.astro"
 
 ## Creating a resource with a View page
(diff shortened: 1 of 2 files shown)


danharrin (changes requested):
Please name it `getAllRelationManagers()`, please return `$this->getResource()::getRelations()` from it so the conditional can be removed

packages/panels/docs/03-resources/05-viewing-records.md
```diff
@@ -170,6 +170,45 @@ public function infolist(Infolist $infolist): Infolist
...
+protected function getAllRelationManagers(): array
+{
+    return [
+        RelationManagers\ContactsRelationManager::class,
+        RelationManagers\AddressesRelationManager::class,
+    ];
+}
+```
+
+If `getAllRelationManagers` isn't defined or returns an empty array, any relation managers defined in the resource will be used.
```
danharrin:
```suggestion
If `getAllRelationManagers()` isn't defined, any relation managers defined in the resource will be used.
```

danharrin (changes requested):

PR #15924, opened 2025-03-23
Cache cell state to prevent multiple execution of the closure in `state()/getStateUsing()`

### LINKED ISSUE #14852: Closure inside getStateUsing() called 3 times per row

robinmoisson:
## Description

In the table view of a panel, closures in `->state(...)` or `->getStateUsing(...)` get called three times per row, instead of one. This can severely impact performance if the closure is a query that takes time to execute - especially if multiple columns are using this.

The issue was reported here: #14852

Here is a reproduction repository of the issue: https://github.com/robinmoisson/filament-state-bug (relevant code [is here](https://github.com/robinmoisson/filament-state-bug/blob/main/app/Filament/Resources/UserResource.php#L22-L31))

See [this comment](https://github.com/filamentphp/filament/issues/14852#issuecomment-2661140603): the problem is that the state is computed in the `text-column.blade.php` blade template to display the value (which is expected), but it is also computed in the `isClickDisabled` method, which is called twice (it would be called once if not for a blade quirk). This leads to the state closure being called 3 times.

My proposal to fix this is to cache the state upon first computation, as I don't see any easy way of passing the computed state to `isClickDisabled`.

Also I saw after making the change the contributing guide saying I should tag @danharrin to the conversation before implementing the change - apologies! I've been using this fix locally but I'm not familiar with all the filament codebase, let me know if this wouldn't be a good solution.

## Visual changes

From the repro rep, before:
Screenshot from 2025-03-23 09-40-52

After:
Screenshot from 2025-03-23 09-40-49

## Functional changes

The state of each column state is now cached by record key upon first computation.

(For "Changes have been tested to not break existing functionality." below, I ran the tests with `composer test:pest` and there wasn't any issue, let me know if I should test it in a different way)

### DIFF
--- a/packages/support/src/Concerns/HasCellState.php
+++ b/packages/support/src/Concerns/HasCellState.php
@@ -24,12 +24,10 @@ trait HasCellState
 
     protected ?string $inverseRelationshipName = null;
 
-    protected array $cachedStatePerRecordId = [];
-
-    protected function getRecordId(): mixed
-    {
-        return $this->getRecord()?->getKey();
-    }
+    /**
+     * @var array<string, mixed>
+     */
+    protected array $cachedState = [];
 
     public function inverseRelationship(?string $name): static
     {
@@ -78,12 +76,14 @@ trait HasCellState
 
     public function getState(): mixed
     {
-        if (! $this->getRecord()) {
+        $record = $this->getRecord();
+
+        if (! $record) {
             return null;
         }
 
-        if (array_key_exists($this->getRecordId(), $this->cachedStatePerRecordId)) {
-            return $this->cachedStatePerRecordId[$this->getRecordId()];
+        if (array_key_exists($record->getKey(), $this->cachedState)) {
+            return $this->cachedState[$record->getKey()];
         }
 
         $state = ($this->getStateUsing !== null) ?
@@ -101,9 +101,7 @@ trait HasCellState
             $state = $this->getDefaultState();
         }
 
-        $this->cachedStatePerRecordId[$this->getRecordId()] = $state;
-
-        return $state;
+        return $this->cachedState[$record->getKey()] = $state;
     }
 
     public function getStateFromRecord(): mixed
--- a/packages/support/src/Concerns/HasCellState.php
+++ b/packages/support/src/Concerns/HasCellState.php
@@ -24,6 +24,13 @@ trait HasCellState
 
     protected ?string $inverseRelationshipName = null;
 
+    protected array $cachedStatePerRecordId = [];
+
+    protected function getRecordId(): mixed
+    {
+        return $this->getRecord()?->getKey();
+    }
+
     public function inverseRelationship(?string $name): static
     {
         $this->inverseRelationshipName = $name;
@@ -75,6 +82,10 @@ trait HasCellState
             return null;
         }
 
+        if (array_key_exists($this->getRecordId(), $this->cachedStatePerRecordId)) {
+            return $this->cachedStatePerRecordId[$this->getRecordId()];
+        }
+
         $state = ($this->getStateUsing !== null) ?
             $this->evaluate($this->getStateUsing) :
             $this->getStateFromRecord();
@@ -89,6 +100,8 @@ trait HasCellState
         if (blank($state)) {
             $state = $this->getDefaultState();
         }
+
+        $this->cachedStatePerRecordId[$this->getRecordId()] = $state;
 
         return $state;
     }


danharrin (approved):

jonaaix:
@robinmoisson many thanks!!

robinmoisson:
Thank you for your very fast review and merge @danharrin, the more I interact with filament the more I'm impressed :)

My pleasure @btxtiger, thanks for the brainstorming!

howdu:
@robinmoisson this is great thanks but should it be fetching the key from `getTableRecordKey`?

E.g we've got a custom table widget with group by and the primary key is not loaded causing a `MissingAttributeException` when it's called `cacheState()`. 

```php
public function table(Tables\Table $table): Tables\Table
{
    return $table->query(fn (): Builder => Product::query()->select(['location_id'])->groupBy('location_id'))
        ->columns([
            Tables\Columns\TextColumn::make('location_id')
        ]);
}

public function getTableRecordKey(Model $record): string
{
    return $record->location_id;
}
```

robinmoisson:
@howdu ah, interesting! It looks like the code has changed a bit since this PR but from what I see it's still using `$record->getKey()` and not `$this->getTableRecordKey($record)`.

I don't know the internals of filament very well, so I'm not positive what the implications are. I'll just mentioned that I saw the problem you mention (with my version of the code it didn't throw an error, but all rows were the same since they were cached with an empty string as key) and opened [another PR](https://github.com/filamentphp/filament/pull/16013) to bypass the cache if there was no record found. If your idea allows to restore the use of caching in those cases that'd be great.

(I haven't checked the latest state of the code in detail, so feel free to double check if I'm still understanding things correctly)

rafaucau:
It looks like this PR broke my app that uses composite primary keys. I'm using the [Thiagoprz\CompositeKey\HasCompositeKey](https://github.com/thiagoprz/eloquent-composite-key) trait in my model, and since version 3.3.8, I'm getting an `Array to string conversion` error. Everything works fine in 3.3.7 and earlier. Previously I could solve this with the `getTableRecordKey` method, but it no longer works with the new caching.

danharrin:
@robinmoisson would you mind opening a PR to use that method please?

robinmoisson:
That's an interesting case, I'll try to do it this week. If I understand correctly this would bring the caching to tables without primary keys too, which would be a performance improvement as well.

danharrin:
@howdu @rafaucau the change has been made in #16072

yannisc:
After filament 3.3.8 that introduced this caching, I get wrong data in my table that uses intermediate tables and raw sql to get the right data. Commenting out the following lines  in HasCellState.php fixes the problem:

        if (array_key_exists($record->getKey(), $this->cachedState)) {
             return $this->cachedState[$record->getKey()];
        }

Are there a any plans to revert to previous code or any way to disable this caching functionality?

danharrin:
If you check the record key, is it empty or duplicative across the table?

yannisc:
If by record key you mean the id, it's the same (as it should be) in both the old and the new version of filament.

danharrin:
@yannisc I don't understand the issue so please open an issue with a reproduction repository, it will be treated as priority since it is a breaking change

yannisc:
I think it will be easier to provide here more info. I have this tab modify query on the ListPayments page:

                    ->modifyQueryUsing(fn (Builder $query) =>
                        Payment::leftJoin(
                            DB::raw('(SELECT payment_id, SUM(amount) AS total_amount
                                FROM order_payment GROUP BY payment_id) pp'), function($join) {
                                    $join->on('payments.id', '=', 'pp.payment_id');
                                })
                        ->whereColumn('payments.amount', '!=', 'pp.total_amount')
                        ->orWhereNull('pp.payment_id')
                            )
                ];
            }

        if (auth()->user()->hasRole('Sales Agents') ||
            auth()->user()->hasRole('Billing') ||
            auth()->user()->hasRole('Super Admin')
        ) {
            $tabs += ['sellerscommissions' => Tab::make()
                ->label('Seller Commissions')
                ->modifyQueryUsing(fn (Builder $query) =>
                    $query
                        ->join('order_payment', 'payments.id', '=', 'order_payment.payment_id')
                        ->join('orders', function($join) {
                            $join->on('order_payment.order_id', '=', 'orders.id')
                                ->when(isset($this->tableFilters['renewal']['value']), function($join) {
                                    $join->where('orders.renewal', $this->tableFilters['renewal']['value']);
                             });
                        })
                        ->join('order_user', 'orders.id', '=', 'order_user.order_id')
                        ->join('users', 'order_user.user_id', '=', 'users.id')
                        ->selectRaw('
                            payments.*, orders.name, first_name, last_name,
                            order_payment.amount as order_payment_amount,
                            order_payment.amount * net_amount / total_amount as amount_without_vat,
                            CASE WHEN orders.renewal = 1 THEN order_commission_rate / 2 ELSE order_commission_rate END as order_commission_rate,
                            (CASE WHEN orders.renewal = 1 THEN order_commission_rate / 2 ELSE order_commission_rate END) * order_payment.amount / 100 * net_amount /
                            total_amount as commission_amount
                        ')
                    )
            ];

It worked ok before the update and produced a table like this:

image

After the update, I get the same Commission (and all other order fields) for all rows. Hope that helps.

danharrin:
Sorry, no I don't understand. I will need some simple example in a reproduction repository to be able to fix this.

yannisc:
ok, here it is: https://github.com/filamentphp/filament/issues/16165

danharrin:
This feature is being reverted and re-introduced in v4. The issues are all user error, but breaking anyway.

yannisc:
Thank you! It works now.

I also tried your proposal of adding order_payment.id to the sql query, but it didn't fix the problem.

danharrin:
Each record key needs to be unique, and when I pulled down your repository and added that line it did work. It needs to be last in the list to override the `payment.*` ID selection

yannisc:
You are right. It works on the reproduction repo, as I had a simpler version there, it needs more work on my actual app. Thanks a lot!!

PR #15937, opened 2025-03-25
Docs V4 - Forms section

leandrocfe:
### Docs
- Fix links, format code blocks, add `Aside` components.

### Code
- Fix operator precedence in x-show by wrapping negation in parentheses.

## Functional changes

### DIFF
--- a/packages/forms/docs/01-overview.md
+++ b/packages/forms/docs/01-overview.md
@@ -90,15 +90,6 @@ TextInput::make('name')
     ->label(__('fields.name'))
 ```
 
-Optionally, you can have the label automatically translated [using Laravel's localization features](https://laravel.com/docs/localization#introduction) with the `translateLabel()` method:
-
-```php
-use Filament\Forms\Components\TextInput;
-
-TextInput::make('name')
-    ->translateLabel() // Equivalent to `label(__('Name'))`
-```
-
 ### Hiding a field's label
 
 It may be tempting to set the label to an empty string to hide it, but this is not recommended. Setting the label to an empty string will not communicate the purpose of the field to screen readers, even if the purpose is clear visually. Instead, you should use the `hiddenLabel()` method, so it is hidden visually but still accessible to screen readers:
--- a/packages/forms/docs/01-overview.md
+++ b/packages/forms/docs/01-overview.md
@@ -286,7 +286,7 @@ Select::make('role')
 Toggle::make('is_admin')
     ->hiddenJs(<<<'JS'
         $get('role') !== 'staff'
-    JS)
+        JS)
 ```
 
 Although the code passed to `hiddenJs()` looks very similar to PHP, it is actually JavaScript. Filament provides the `$get()` utility function to JavaScript that behaves very similar to its PHP equivalent, but without requiring the depended-on field to be `live()`.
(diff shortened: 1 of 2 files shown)


packages/forms/docs/01-overview.md
```diff
@@ -90,6 +90,15 @@ TextInput::make('name')
...
 
+Optionally, you can have the label automatically translated [using Laravel's localization features](https://laravel.com/docs/localization#introduction) with the `translateLabel()` method:
+
+```php
+use Filament\Forms\Components\TextInput;
+
+TextInput::make('name')
+    ->translateLabel() // Equivalent to `label(__('Name'))`
+```
+
```
danharrin: Please remove this, I didn't include it because I don't think it should exist and its only there for backwards compatibility

packages/forms/docs/01-overview.md
```diff
@@ -261,7 +270,7 @@ Select::make('role')
         'user' => 'User',
         'staff' => 'Staff',
     ])
-    ->live()
+    ->live(),
```
danharrin: In situations where there are multiple unrelated example like this I deliberately do not  include commas

packages/forms/docs/01-overview.md
```diff
@@ -281,11 +290,11 @@ Select::make('role')
...
         'staff' => 'Staff',
-    ])
+    ]),
 
 Toggle::make('is_admin')
-    ->hiddenJs(>>>JS
-        \$get('role') !== 'staff'
+    ->hiddenJs(<<<'JS'
+    $get('role') !== 'staff'
     JS)
```
danharrin: Please indent the two lines after the `<<<'JS'`

packages/forms/docs/01-overview.md
```diff
@@ -1209,11 +1229,10 @@ To generate a slug from a title while the user is typing, you can use the [`afte
 ```php
 use Filament\Schemas\Components\Utilities\Set;
-use Illuminate\Support\Str;
 
 TextInput::make('title')
     ->live(onBlur: true)
-    ->afterStateUpdated(fn (Set $set, ?string $state) => $set('slug', Str::slug($state)))
+    ->afterStateUpdated(fn (Set $set, ?string $state) => $set('slug', str($state)->slug())),
```
danharrin: Please revert the use of the fluent string helper, I only really suggest that when there are multiple operations

packages/schemas/src/Schema.php
```diff
@@ -261,7 +261,7 @@ function (Component | Action | ActionGroup $component) use (&$hasVisibleComponen
                                 <?php } ?>
                                 <?php if (filled($xShow = match ([filled($hiddenJs), filled($visibleJs)]) {
                                     [true, true] => "(! {$hiddenJs}) && ({$visibleJs})",
-                                    [true, false] => "! {$hiddenJs}",
+                                    [true, false] => "! ({$hiddenJs})",
```
danharrin, line 264: Does the line above also need parenthesis around `{$hiddenJs}`

danharrin (changes requested):

packages/forms/docs/01-overview.md
```diff
@@ -284,8 +284,8 @@ Select::make('role')
     ])
 
 Toggle::make('is_admin')
-    ->hiddenJs(>>>JS
-        \$get('role') !== 'staff'
+    ->hiddenJs(<<<'JS'
+        $get('role') !== 'staff'
     JS)
```
danharrin: The last line needs to be in line with the content, so please indent

danharrin (changes requested):

PR #15947, opened 2025-03-26
Fix: Better detection of encoding for .csv files

manuzoli:
## Description

When importing a .csv file into Filament, incorrect file encoding can sometimes be detected. This issue occurs because only the header of the file is used to determine the encoding. If the header is short or lacks distinctive characters, the detection can produce false positives — multiple encodings might appear valid.

To improve accuracy, I modified the code to scan the first 8 KB of the file instead of just the header. This change still ensures fast performance but significantly improves detection accuracy and reduces false positives. Especially for older formats like 

## Visual changes

There are no visual changes by this update. 

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/CanImportRecords.php
+++ b/packages/actions/src/Concerns/CanImportRecords.php
@@ -440,7 +440,18 @@ trait CanImportRecords
     protected function detectCsvEncoding(mixed $resource): ?string
     {
         rewind($resource);
-        $contentSample = fread($resource, 8192);
+
+        $lineCount = 0;
+        $contentSample = '';
+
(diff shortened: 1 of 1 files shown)


bzy107:
I think issue #12571 will recur.

manuzoli:
Thank you very much—you’re absolutely right. I’ve updated my commit so that the first 20 lines of the CSV are now used to detect the charset. This approach is still performant, should provide ample data for accurate charset detection, and shouldn’t impact performance.

danharrin:
@bzy107 what do you think of the changes made?

bzy107:
I don't think #12571 will recur due to the fixes.

danharrin (approved):

PR #15951, opened 2025-03-27
Fix: Exporting excel files not working in Laravel Vapor environments

adriaardila:
## Description

CSV exports work correctly but we had an issue where the exports for excel didn't work in our production server, but they did locally. After some debugging I found this error:

`Fatal error: Uncaught Exception: Error encoding runtime JSON response: Malformed UTF-8 characters, possibly incorrectly encoded in /var/task/vendor/laravel/vapor-core/src/Runtime/NotifiesLambda.php:49`

The error came from this part in XlsxDownloader:

```
        if ($disk->exists($filePath = $directory . DIRECTORY_SEPARATOR . $fileName)) {
            $response = $disk->download($filePath);

            if (ob_get_length() > 0) {
                ob_end_clean();
            }

            return $response;
        }
```

I modified it to stream the download the same way csv exports are streamed and now it works fine in my production and staging server, so it should work the same on other environments but it also works on vapor environments now:

```
        if ($disk->exists($filePath = $directory . DIRECTORY_SEPARATOR . $fileName)) {
            return response()->streamDownload(function () use ($disk, $filePath) {
                if (ob_get_level()) {
                    ob_end_clean();
                }

                echo $disk->get($filePath);

                flush();
            }, $fileName, [
                'Content-Type' => 'application/vnd.ms-excel',
            ]);
        }
```

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/actions/src/Exports/Downloaders/XlsxDownloader.php
+++ b/packages/actions/src/Exports/Downloaders/XlsxDownloader.php
@@ -24,13 +24,13 @@ class XlsxDownloader implements Downloader
         $fileName = $export->file_name . '.xlsx';
 
         if ($disk->exists($filePath = $directory . DIRECTORY_SEPARATOR . $fileName)) {
-            $response = $disk->download($filePath);
-
-            if (ob_get_length() > 0) {
-                ob_end_clean();
-            }
-
-            return $response;
+            return response()->streamDownload(function () use ($disk, $filePath) {
+                if (ob_get_level()) ob_end_clean();
+                echo $disk->get($filePath);
+                flush();
+            }, $fileName, [
+                'Content-Type' => 'application/vnd.ms-excel',
+            ]);
         }
 
         $writer = app(Writer::class);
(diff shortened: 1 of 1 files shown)


edgarMiralles:
This is what I was looking for. Thank you!

danharrin:
@gillobis can you please test this? Since you made the change in #15792

gillobis:
> @gillobis can you please test this? Since you made the change in #15792

I've just tested the modification and it works fine.

danharrin (changes requested):
For a large export, I am worried about the performance impact. I think the `download()` method you removed doesn't load the whole file into memory, whereas `get()` does. Am I wrong about that?

adriaardila:
> For a large export, I am worried about the performance impact. I think the `download()` method you removed doesn't load the whole file into memory, whereas `get()` does. Am I wrong about that?

I've tried it in my production server with a 5000 rows export and it worked fine. This is the same way files are downloaded with the CsvDownloader class, do you think it's gonna be much different when it's an excel file?

danharrin:
The CSV files are chunked so there is limited data in memory, thats not possible with excel

adriaardila:
> The CSV files are chunked so there is limited data in memory, thats not possible with excel

Got it, let me see if I can find a solution using download() that also doesn't break vapor environments

adriaardila:
@danharrin found a much better solution, apparently vapor has this header that when set in the response will base64 encode the binary, avoiding the utf-8 malformatting error, it should still work the same for other environments and is only relevant for vapor environments:

`$response->headers->set('X-Vapor-Base64-Encode', 'True');`

danharrin (approved):
Much better. Thanks for improving this.

PR #15962, opened 2025-03-28
fix: Add missing Filepond translations to support more locales in File Upload

exabyssus:
## Description

Currently used Filepond version 4.31.1 supports more locales than are imported into file-upload component. Therefore those locales can't be used. 

Added missing locales `am`, `az`, `el`, `et`, `he`, `hr`, `ja`, `ko`, `lt`, `lv`, `sk`.
 
## Visual changes

None

## Functional changes


packages/forms/resources/js/components/file-upload.js
```diff
@@ -784,28 +795,39 @@ import zh_CN from 'filepond/locale/zh-cn'
...
+    az,
     ca,
     ckb,
     cs,
     da,
     de,
+    el,
     en,
     es,
+    et,
```
danharrin: This language doesn't have any translations in Filament at the moment, so please remove it

danharrin (changes requested):

packages/forms/resources/js/components/file-upload.js
exabyssus: Ok, it's been removed.

danharrin (approved):

PR #15994, opened 2025-04-03
Ability to check if macro exists

ralphjsmit:
This PR allows to check if a macro is registered using a `Component:hasMacro()` method, similar to Laravel's `hasMacro()` method.

Thanks!

### DIFF
--- a/packages/support/src/Concerns/Macroable.php
+++ b/packages/support/src/Concerns/Macroable.php
@@ -41,7 +41,7 @@ trait Macroable
 
     public static function hasMacro(string $name): bool
     {
-        return isset(static::$macros[$name]);
+        return (bool) static::getMacro($name);
     }
 
     /**
--- a/packages/support/src/Concerns/Macroable.php
+++ b/packages/support/src/Concerns/Macroable.php
@@ -39,6 +39,11 @@ trait Macroable
         static::$macros = [];
     }
 
+    public static function hasMacro(string $name): bool
+    {
+        return isset(static::$macros[$name]);
+    }
+
     /**
      * @param  array<array-key>  $parameters
      */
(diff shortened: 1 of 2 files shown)


packages/support/src/Concerns/Macroable.php
```diff
@@ -39,6 +39,11 @@ public static function flushMacros(): void
         static::$macros = [];
     }
 
+    public static function hasMacro(string $name): bool
+    {
+        return isset(static::$macros[$name]);
```
danharrin: This will return true even if the macro is registered for a different class

danharrin (changes requested):

packages/support/src/Concerns/Macroable.php
ralphjsmit: Apologies, you are right indeed. I just copied the implementation from Laravel assuming Filament had the same. I updated the test to now account for this and also check macros on child components.

packages/support/src/Concerns/Macroable.php
```diff
@@ -39,6 +39,11 @@ public static function flushMacros(): void
         static::$macros = [];
     }
 
+    public static function hasMacro(string $name): bool
+    {
+        return (bool) static::getMacro($name);
```
ralphjsmit, line 44: Doing it this way (instead of with some `isset()` calls) ensure that we don't need to repeat the logic for checking the macros on parent classes, which I don't think is great to copy and maintain twice.

packages/support/src/Concerns/Macroable.php
danharrin: The reason we have our own trait instead of using Laravel's is because it handles inherited classes much better

danharrin (approved):

PR #16008, opened 2025-04-04
Clear cache before cell-state assertions to support changing state during tests

### LINKED ISSUE #16007: Caching cell state breaks testing assertions
### Package

filament/filament

### Package Version

v3.3.8

### Laravel Version

v11.44.2

### Livewire Version

v3.6.2

### PHP Version

8.3.19

### Problem description

The updates here seem to cache the record, which makes tests that run something like the block below fail, whereas they passed prior to the change:
```
it('update table state test', function () {
    // Get all the patients (only one is seeded)
    $patients = Patient::all();

    livewire(ListPatients::class)
        ->assertCanSeeTableRecords($patients)
        // Make sure the patient's status is null
        ->assertTableColumnStateSet('status', null, $patients->first())
        // Update the status to 'active'
        ->assertTableColumnExists('status', function (SelectColumn $column) {
            $column->updateState('active');
            return true;
        }, $patients->first())
        // Verify the the status was updated
        ->assertTableColumnStateSet('status', 'active', $patients->first());
});
```

The record itself has been properly updated. However, the cached version remains the same. This appears to be related to the changes introduced here:
https://github.com/filamentphp/filament/commit/a36b1822f7f7b5eceaff80d9ac181a43c03db0a2

I've added a proposal to clear the cache immediately before an assertion during testing: 
https://github.com/filamentphp/filament/pull/16008

I've seen other discussions regarding octane, so this adjustment is limited to testing scenarios only.

### Expected behavior

Passing test, allowing state to be modified and verified while running a single test.

### Steps to reproduce

1. Create a table that contains a SelectColumn
2. Assert the initial state in a livewire test
3. Chang the state
3. Assert the new state exits

https://github.com/Mrkbingham/app-filament-issue-3.x

### Relevant log output

```shell
Failed asserting that a table column with name [status] has value of [active] for record [1] on the [App\Filament\Resources\PatientResource\Pages\ListPatients] component.
Failed asserting that null matches expected 'active'.
```

Mrkbingham:
## Description

Invalidates cache immediately before state assertions on tables, so knock-on effects from state-changes can be verified during testing.

## Visual changes

n/a

## Functional changes


danharrin (changes requested):
Should the state not be cleared within `updateState()` instead of in the testing helpers?

danharrin (approved):

danharrin:
I've refactored this a bit, I agree that it's best to take this approach to capture all potential places where the state could be updated and not just `updateState()`

danharrin (approved):

danharrin (commented):
Apologies for the formatting changes, another PR did not format their work before committing

danharrin (approved):

PR #14411, opened 2024-10-01
Feature: XLSX Writer Options

mmonari:
## Description

This PR adds a `getXlsxOptions()` method to the Exporter class to allow users to further customize their exported XLSX documents, such as column widths, or any other `OpenSpout\Writer\XLSX` options  as documented at the [OpenSpout Documentation](https://github.com/openspout/openspout/blob/4.x/docs/documentation.md)

## Code style
- [x] Code style has been fixed by running the `composer cs` command.

## Testing
- [x] Changes have been tested to not break existing functionality.

## Documentation
- [ ] Documentation is up-to-date.

### Note:
As this is my first PR, I appreciate your understanding regarding my commit history. The first one represents the actual changes, the second was after executing a `composer cs`  and it should have been described differently.

PR #16025, opened 2025-04-08
Feature: XLSX Configure Writer

gpibarra:
## Description

In PR https://github.com/filamentphp/filament/pull/14411, options were added for the new Writer, but some features cannot be enabled through Options, for example https://github.com/openspout/openspout/blob/4.x/docs/documentation.md#autofilter-xlsx-writer and https://github.com/openspout/openspout/blob/4.x/docs/documentation.md#sheet-view-xlsx-writer.

This method is called before closing the Writer, but after all records have been written, because it is useful to have `$writer->getCurrentSheet()->getWrittenRowCount()` to know how many rows have been written.

## Functional changes

### DIFF
--- a/packages/actions/src/Exports/Exporter.php
+++ b/packages/actions/src/Exports/Exporter.php
@@ -14,6 +14,7 @@ use Illuminate\Database\Eloquent\Model;
 use Illuminate\Queue\Middleware\WithoutOverlapping;
 use OpenSpout\Common\Entity\Style\Style;
 use OpenSpout\Writer\XLSX\Options;
+use OpenSpout\Writer\XLSX\Writer;
 
 abstract class Exporter
 {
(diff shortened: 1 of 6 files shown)


packages/actions/src/Exports/Exporter.php
```diff
@@ -209,6 +210,11 @@ public function getXlsxWriterOptions(): ?Options
         return null;
     }
 
+    public function configureXlsxWriterBeforeClose(Writer &$writer): Writer
```
danharrin: I don't think we need the `&` symbol here as all objects in PHP are passed by reference, please let me know if this breaks the feature though

danharrin (changes requested):

packages/actions/src/Exports/Exporter.php
gpibarra: In fact, it is not necessary in this case

danharrin (approved):

PR #16043, opened 2025-04-10
Add Operation Enum Support to Forms::hiddenOn()

onairmarc:
## Description

This PR creates an `Operation` enum for forms that can be used with the `hiddenOn` method. 

## Visual changes

There are no visible changes to the rendered HTML. The only thing that changes is the use of an enum instead of magic string.

## Functional changes


packages/support/src/Enums/Operation.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+/*
+ * Copyright (c) 2025. Encore Digital Group.
```
danharrin: Sorry, I just can't merge this in, this is an open source project under MIT. Please remove this yourself or close the PR, but I can't remove it for you.

packages/support/src/Enums/Operation.php
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+namespace Filament\Support\Enums;
+
+enum Operation: string
+{
+    case Create = 'create';
+    case View = 'view';
+    case Edit = 'edit';
+    case Delete = 'delete';
```
danharrin: Please remove this as it isn't an operation that would be used for a schema

danharrin (changes requested):

packages/support/src/Enums/Operation.php
onairmarc: Ope my bad! I thought I had turned that off! Happy to remove it!

packages/support/src/Enums/Operation.php
onairmarc: I'm assuming you meant `Operation::Delete`. If that's not the case and you need me to change something else, please let me know.

danharrin:
Thanks, I am merging this into v4 as its a bit of a mindset shift for the current method usage

danharrin (approved):

onairmarc:
Sounds good to me! Thanks for all your hard work! Super excited for v4!

PR #16044, opened 2025-04-11
Wrapped fieldname to solve SQL Error when sorting

jeroendeleur:
Wrapped fieldname with backticks, because when the primary key of a model is a reserved word, MySql will throw an error.

## Description

When the model has a reserved word as the PK e.g. `key` sorting records will throw the following SQL error.

SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'key = 'cZKBUpPyUR' then 1 when key = 'EQ5zh0Eqhf' then 2 when key = 'lwmhTQ8C...' at line 1 (Connection: mysql, SQL: update `age_groups` set `sort` = case when key = 'cZKBUpPyUR' then 1 when key = 'EQ5zh0Eqhf' then 2 when key = 'lwmhTQ8Cjd' then 3 when key = '0XaE8LzBZN' then 4 when key = 'tUkKkfH3P2' then 5 when key

I've added backticks around $modelKeyName on line 47 of the CanReorderRecords concern, this will solve the issue.

## Visual changes

No visual changes

## Functional changes

- [ v] Code style has been fixed by running the `composer cs` command.
- [ v] Changes have been tested to not break existing functionality.
- [ v] Documentation is up-to-date.


danharrin:
I think this will break every other driver apart from MySQL? right?

jeroendeleur:
> I think this will break every other driver apart from MySQL? right?

This is a fair point. Is it an idea to replace it with upsert? 

Replacing

```php
$model
    ->newModelQuery()
    ->whereIn($modelKeyName, array_values($order))
    ->update([
        $orderColumn => DB::raw(
            'case ' . collect($order)
                ->map(fn ($recordKey, int $recordIndex): string => 'when `' . $modelKeyName . '` = ' . DB::getPdo()->quote($recordKey) . ' then ' . ($recordIndex + 1))
                ->implode(' ') . ' end'
        ),
    ]);
```

with

```php
$collection = collect($order)
                ->map(fn ($recordKey, int $recordIndex) => [$modelKeyName => $recordKey, $orderColumn => $recordIndex]);
$model->upsert($collection->toArray(), [$modelKeyName], [$orderColumn]);
```

danharrin:
Hmm I'm not sure I've seen upsert used in this way. Maybe get the query grammar object, and then use `invade($grammar)->wrapValue($string)` to wrap the string in the correct quotes for that database engine?

jeroendeleur:
That was my first aproach, but google sent me through the doctrine/dbal rabbit hole. I have changed the code to:

```PHP
$wrappedModelKeyName = DB::connection()->getSchemaGrammar()->wrap($modelKeyName);
```

And then used the new variable in the raw SQL.

I have tested it on, MySql, MariaDB, SQLite and Postgres drivers, not on SQL server as I don't have a test bench for that.

packages/tables/src/Concerns/CanReorderRecords.php
```diff
@@ -36,14 +36,15 @@ public function reorderTable(array $order): void
 
             $model = app($this->getTable()->getModel());
             $modelKeyName = $model->getKeyName();
+            $wrappedModelKeyName = DB::connection()->getSchemaGrammar()->wrap($modelKeyName);
```
danharrin:
```suggestion
            $wrappedModelKeyName = $model->getConnection()->getSchemaGrammar()->wrap($modelKeyName);
```

in case the model has a customized `$connection` set, otherwise the default is used

danharrin (changes requested):

danharrin (approved):

danharrin:
Thanks for fixing it!

DirtyRacer1337:
This cause #16076 error

maybe something like that would work for all?
```php
$modelKeyName = $model->getKeyName();
$wrappedModelKeyName = $model->getConnection()->getSchemaGrammar()?->wrap($modelKeyName) ?? $modelKeyName;
```

danharrin:
@DirtyRacer1337 can you please explain the situation why a "connection" object could return a null grammar object?

DirtyRacer1337:
> @DirtyRacer1337 can you please explain the situation why a "connection" object could return a null grammar object?

In my case and mentioned issue ` $model->getConnection()->getSchemaGrammar()` return null

I'm using postgresql, maybe that's important

Just checked, everything works on previous version

danharrin:
What does `getConnection()` return, is the information about that object correct?

danharrin:
Does it work if you swap `$model->getConnection()->getSchemaGrammar()` for `$model->getConnection()->getQueryGrammar()`?

DirtyRacer1337:
Query Grammar not null, Schema Grammar null

danharrin:
Does it work though with the query grammar?

DirtyRacer1337:
I'm not sure what it should return, but in my case with `$model->getConnection()->getQueryGrammar()->wrap($model->getKeyName())`
it return `"id"`

DirtyRacer1337:
> What does `getConnection()` return, is the information about that object correct?

изображение

DirtyRacer1337:
so then solution would be like
```php
$modelKeyName = $model->getKeyName();
$modelGrammar = $model->getConnection()->getSchemaGrammar() ?? $model->getConnection()->getQueryGrammar();
$wrappedModelKeyName = $modelGrammar->wrap($modelKeyName);
```

or all together

```php
$modelKeyName = $model->getKeyName();
$modelGrammar = $model->getConnection()->getSchemaGrammar() ?? $model->getConnection()->getQueryGrammar();
$wrappedModelKeyName = $modelGrammar?->wrap($modelKeyName) ?? $modelKeyName;
```

arshaplus:
Throws error when sorting:

`Call to a member function wrap() on null at /var/www/html/project/vendor/filament/tables/src/Concerns/CanReorderRecords.php:39`

DB driver is PostgreSQL.

danharrin:
I think we can actually remove the `getSchemaGrammar()` as the query grammar sounds like what we actually need here. I'll make a PR

danharrin:
#16084

PR #16065, opened 2025-04-16
fix(spa): unsaved data changes alert

Domianidze:
## Description

This PR addresses an issue in SPA mode where the "unsaved changes" alert logic was based on whether any form had been submitted. This logic introduced two key issues:

1. **Editing and re-editing a form**: After submitting a form (e.g., updating a record), if a user made another change and attempted to navigate away, the unsaved changes alert would **not** appear — because `formSubmitted` was already `true` from the first submission.

2. **Additional forms on the same page**: If a user interacted with another form (e.g., a create action in a select field), submitting that would also set `formSubmitted = true`, which incorrectly prevented the alert from showing even if the main form had unsaved changes.

To fix this, the `formSubmitted` check was removed. However, this introduced another issue — the alert began appearing during programmatic redirects (e.g., after a create or delete action), which is likely why the `formSubmitted` check was originally added.

This PR resolves that by replacing the form submission check with a more targeted condition: it now checks if a redirect is happening programmatically using `$wire?.__instance?.effects?.redirect`. If so, the alert is not triggered. We've used this approach extensively in our own app and have not encountered any flaws with it.

Please let me know if there are edge cases we may have missed, or if there’s a better approach you’d recommend.

## Visual changes

Before:

chrome-capture-2025-4-16 (1)

After:

chrome-capture-2025-4-16

## Functional changes

### DIFF
--- a/packages/panels/resources/views/components/page/unsaved-data-changes-alert.blade.php
+++ b/packages/panels/resources/views/components/page/unsaved-data-changes-alert.blade.php
@@ -6,23 +6,15 @@
     @if (FilamentView::hasSpaMode())
         @script
             <script>
-                let formSubmitted = false
-
-                document.addEventListener(
-                    'submit',
-                    () => (formSubmitted = true),
-                )
-
                 shouldPreventNavigation = () => {
-                    if (formSubmitted) {
+                    if ($wire?.__instance?.effects?.redirect) {
                         return
                     }
 
                     return (
                         window.jsMd5(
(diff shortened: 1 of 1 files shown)


danharrin:
#12363 seems to be where this logic was introduced, what do you think of this PR @binaryfire?

binaryfire:
@danharrin I've tested this and it works. This is a good improvement and should be merged. Nice work @Domianidze

danharrin (approved):

PR #16066, opened 2025-04-16
Xlsx style cell column

gpibarra:
## Description

Currently, you can style the header and the rest of the cells (separately). This PR allows you to format each of the exported columns (value cells only; header row cells are not included).
When you style a column, you can call the function multiple times, and each style (or callback) is saved in an array. When obtaining the column style, all styles are calculated and merged. This is designed to allow styling, for example, in the make/setup function for potential new column classes that extend the `ExportColumn` class.
Styles (or callbacks) can be fixed, meaning they do not depend on the cell value. They will be cached to improve performance and avoid merging each row of data in the export file. Additionally, you can assign a style based on the cell value, which will be executed on each row of data (OpenSpout does not support Excel conditional formatting).

## Functional changes

### DIFF
--- a/packages/actions/docs/12-export.md
+++ b/packages/actions/docs/12-export.md
@@ -602,6 +602,8 @@ You can only specify a single character, otherwise an exception will be thrown.
 
 ## Customizing XLSX files
 
+### Styling all cells
+
 If you want to style the cells of the XLSX file, you may override the `getXlsxCellStyle()` method on the exporter class, returning an [OpenSpout `Style` object](https://github.com/openspout/openspout/blob/4.x/docs/documentation.md#styling):
 
 ```php
@@ -614,6 +616,8 @@ public function getXlsxCellStyle(): ?Style
         ->setFontName('Consolas');
 }
 ```
+
+### Styling all header cells
 
 If you want to use a different style for the header cells of the XLSX file only, you may override the `getXlsxHeaderCellStyle()` method on the exporter class, returning an [OpenSpout `Style` object](https://github.com/openspout/openspout/blob/4.x/docs/documentation.md#styling):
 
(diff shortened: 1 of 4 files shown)


gpibarra:
In the documentation, I corrected a change to this PR: https://github.com/filamentphp/filament/pull/16025
It wasn't clear whether the function needed to be renamed or not, but only the documentation had been changed. Feel free to rename the function.

danharrin (changes requested):
To be honest I am concerned about the performance implications of calling another method for each cell. Maybe the implementation should just be `makeXlsxRow()` static method on the Exporter where you can pass a function that accepts `$data, $style, $columns` and returns a `Row` object. The `$columns` would be an array of column names that match the `$data` in order, so you can use `Row::fromValuesWithStyles()` yourself. If that function is not passed, the `makeXlsxRow()` would just return the default `Row::fromValues()` object. This way, we aren't really introducing new overhead by looping over columns etc. Thoughts?

danharrin:
At this point, maybe the entire writer creation process should just be moved onto the Exporter class, so it can be overriden in any way instead of needing all these new config methods

gpibarra:
I agree. Initially, I only called getStyle once before the record loop. But the fact that the library doesn't support conditional formatting motivated me to create an additional method. I think it's perhaps too much customization for the default case. I would leave the basic style and not the one that depends on the value (called for each cell in each row). 

Regarding the writer depending on the exporter, I don't see it as a problem, but if not, it's not that difficult to override the logic. These changes I proposed were born from a project with Filament v3, and I modified the writer logic with a bind from a service provider. It's not complicated.

danharrin:
Can yon move all the writer logic into the exporter class then please? We could even remove that `configureXlsxWriterBeforeClose()` method tbh

danharrin:

gpibarra:
Sorry for the delay, I'll get back to working on this PR tomorrow.

danharrin:
Replaced by 476b92c7f1a04dc29066eb271bd498f8d25b6a6c

PR #16068, opened 2025-04-17
Fix "isLimited is not defined" error in TextColumn expandableLimitedList functionlity.

### LINKED ISSUE #16047: Alpine Expression Error: isLimited is not defined
### Package

filament/tables

### Package Version

v3.3.10

### Laravel Version

v12.8.1

### Livewire Version

v3.6.2

### PHP Version

PHP 8.3.20

### Problem description

I have a table which has this column ([here](https://github.com/ericmp33/example/blob/main/app/Filament/Resources/UserResource.php#L34)):
```php
Tables\Columns\TextColumn::make('items.name')
    ->badge()
    ->listWithLineBreaks(false)
    ->limitList(1)
    ->expandableLimitedList()
    ->limit(15)
    ->searchable()
```

I open the browser devtools and I see:

Image

Image

I tried to debug it down and I would guess the error is [on this line](https://github.com/filamentphp/filament/blob/3.x/packages/tables/resources/views/columns/text-column.blade.php#L296).

### Expected behavior

I expected to be able to toggle the "Show 1 more" / "Show 1 less" buttons as in earlier versions.

### Steps to reproduce

1. Clone the repo
2. Execute migrations
3. Go to /admin/users
4. Open web devtools
5. Checkout [this file](https://github.com/filamentphp/filament/blob/3.x/packages/tables/resources/views/columns/text-column.blade.php#L296)

https://github.com/ericmp33/example

misterneo:
Fix expandableLimitedList "isLimited is not defined" error by updating x-filament::link attributes 

## Description

This pull request fixes the "isLimited is not defined" error that occurs when the `expandableLimitedList()` functionality is activated. The issue was caused by the use of the `tag="button"` attribute in the `<x-filament::link>` component.

This fix removes the `tag="button"` attribute and replaces it with `tag="div"`, which has been added to the global Link component’s tag options. It also ensures that `role="button"` is retained to maintain accessibility.

Additionally, the `.stop` modifier was added to the `x-on:click.prevent` directive to prevent event propagation. (Modal or page navigation when the cell is clicked).

## Visual changes
No visual changes were introduced. The component retains its default styling and behavior as an `<a>` tag.

Before:

https://github.com/user-attachments/assets/47402d13-6343-4f08-9e99-9216ac026190

After:

https://github.com/user-attachments/assets/912a00d6-34c2-49fd-bdca-b4a5bcb6f05b

## Functional changes

### DIFF
--- a/packages/support/resources/views/components/link.blade.php
+++ b/packages/support/resources/views/components/link.blade.php
@@ -301,4 +301,53 @@
         @endif
     </button>
     @trim
+@elseif ($tag === 'div')
+    <div
+        role="button"
+        @if ($keyBindings || $hasTooltip)
+            x-data="{}"
+        @endif
+        @if ($keyBindings)
+            x-bind:id="$id('key-bindings')"
+            x-mousetrap.global.{{ collect($keyBindings)->map(fn (string $keyBinding): string => str_replace('+', '-', $keyBinding))->implode('.') }}="document.getElementById($el.id).click()"
+        @endif
+        @if ($hasTooltip)
+            x-tooltip="{
+                content: @js($tooltip),
+                theme: $store.theme,
+            }"
+        @endif
+        {{ $attributes->class([$linkClasses]) }}
+    >
+        @if ($icon && $iconPosition === IconPosition::Before)
+            <x-filament::icon
+                :alias="$iconAlias"
+                :icon="$icon"
+                :class="$iconClasses"
+                :style="$iconStyles"
+            />
+        @endif
+
+        <span class="{{ $labelClasses }}" style="{{ $labelStyles }}">
+            {{ $slot }}
+        </span>
+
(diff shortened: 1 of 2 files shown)


danharrin:
Please can the link tag be a `<div>` instead of `<a>`, that is a little more semantically correct in this case as there is no `href`

misterneo:
@danharrin  I understand your concern, but I'm not sure what you mean. Are you suggesting replacing the `<x-filament::link>` with a `<div>`, or adding a new tag option inside the Link component to allow it to render as a `<div>`?"

danharrin:
The latter please

misterneo:
Done! Please give it a quick review and let me know if anything needs tweaking.

danharrin (approved):

danharrin:
Thanks

PR #16097, opened 2025-04-22
fix: Set default maxSize on SpatieMediaLibraryFileUpload based on media-library configuration

dev:
## Description

This PR adds a default maxSize setting to the SpatieMediaLibraryFileUpload component that automatically uses the value from the media-library configuration. The maxSize is now set to `round(config('media-library.max_file_size') / 1000)` by default.
Currently, if developers forget to set maxSize manually, the media-library throws an exception when users try to upload files that exceed the size limit. By setting a default maxSize value that matches the media-library configuration, users will receive a validation error instead of an exception, which improves the user experience.

## Visual changes

N/A

## Functional changes

### DIFF
--- a/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
+++ b/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
@@ -52,6 +52,8 @@ class SpatieMediaLibraryFileUpload extends FileUpload
     {
         parent::setUp();
 
+        $this->maxSize(round(config('media-library.max_file_size') / 1000));
+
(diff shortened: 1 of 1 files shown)


danharrin (changes requested):
I think collections can also define a max size, so you should pass a function to `maxSize()` to check that before falling back on the default. We might do this already for accepted file types too but I can't remember.

danharrin:

danharrin:
Please re-open the PR once those changes are made. Thanks!

dev:
Hi @danharrin, I'm sorry, I totally forgot about this.

I looked into this but couldn't find anything about max file size on collection level.
It does have `acceptsMimeTypes` and `collectionSizeLimit` (for number of files), but nothing for file size limits per collection. 

Are you sure collections can define a max size? Or can you point me in the right direction?

PR #16100, opened 2025-04-22
Fix: change color() in all actions to defaultColor() where required, so that it can be overidden

getupkid:
## Description

Changed color() in ViewActions to defaultColor() so that it can be overidden if need be using configureUsing().

I wanted to change the color of my ViewActions globally but even when using ViewAction::configureUsing() I noticed that the color was not changing from the default "gray". I noticed that in all ViewActions, the color is strictly set to "gray". So the only way to override this was to specify ->color() on every single ViewAction in my project.

So, to be able to configure to color of the ViewActions globally, I've updated $this->color('gray') to $this->defaultColor('gray'). This way, it should not harm anything as people who have used ViewAction::color(xxx) should not be affected.

### Before:
```
ViewAction::configureUsing( function(TableViewAction $action) {
     $action->color(Color::Lime);
});
```
image

### After:
(with ConfigureUsing override):
```
ViewAction::configureUsing( function(TableViewAction $action) {
     $action->color(Color::Lime);
});
```
image

(without ConfigureUsing override):
image

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/CanExportRecords.php
+++ b/packages/actions/src/Concerns/CanExportRecords.php
@@ -255,7 +255,7 @@ trait CanExportRecords
             }
         });
 
-        $this->color('gray');
+        $this->defaultColor('gray');
 
         $this->modalWidth('xl');
 
--- a/packages/actions/src/Concerns/CanImportRecords.php
+++ b/packages/actions/src/Concerns/CanImportRecords.php
@@ -383,7 +383,7 @@ trait CanImportRecords
                 }),
         ]);
 
-        $this->color('gray');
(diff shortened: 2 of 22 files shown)


danharrin (changes requested):
Can you update all of the actions, not just these please?

getupkid:
@danharrin The PR has been updated to include other actions whereby color() was already defined.

danharrin:
In v4 I have changed the feature so that any parent defined `configure()` statements are called before `setUp()`, and any child ones are called after `setUp()` - c91a51d9bd0e62f053e05b9e90ee8600e5820345

This will hopefully prevent similar issues in the future

PR #16106, opened 2025-04-22
Feature: Personal pivot data

trin4ik:
## Description

In the case of `relationship` + `multiple` Selects and CheckboxLists, it is possible to bind `pivotData` 
```php
Select::make('primaryTechnologies')
    ->relationship(name: 'technologies', titleAttribute: 'name')
    ->multiple()
    ->pivotData([
        'is_primary' => true,
    ])
```
https://filamentphp.com/docs/3.x/forms/fields/select#saving-pivot-data-to-the-relationship
But there is no way to pass specific `pivotData`. 
For example, I want to make `is_primary=true` for the first relationship only, such as this:
```php
Select::make('technologies')
    ->relationship(name: 'technologies', titleAttribute: 'name')
    ->multiple()
    ->pivotData(
        fn (array $state): array => collect($state)
            ->mapWithKeys(fn ($technology, $index) => [
                $technology => [
                    'is_primary' => $index === 0, // set `is_primary` only for first technology
                ],
            ])
            ->toArray()
    )
```
Also a classic example: when creating a team and adding admins to it, I want to create the first one as `owner` and not just `admin`.
Taking into account that `pivotData` accepts Closure into which `Forms\Get`, state etc. are binded, it turns out to be quite convenient and flexible. 

## Visual changes

N/A

## Functional changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,5 @@
         "trix": "^2.1",
         "uuid-browser": "^3.1.0",
         "vanilla-colorful": "^0.7.2"
-    },
-    "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
+    }
 }
--- a/package.json
+++ b/package.json
@@ -63,5 +63,6 @@
         "trix": "^2.1",
         "uuid-browser": "^3.1.0",
         "vanilla-colorful": "^0.7.2"
-    }
+    },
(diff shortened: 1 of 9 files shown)


danharrin:

I think once v4 is released you could PR this feature again. I would suggest that instead of passing the entire `$state`, just the current repeater item state is passed to the function, and then you return whatever the pivot data should be for that item, instead of needing to map through the items.

trin4ik:
> Hi, apologies but we are no longer accepting new features for v3 as per #13744. Also, please check out our [contributing guide](https://filamentphp.com/docs/3.x/support/contributing#development-of-new-features) which talks about proposing the feature before submitting the PR, to ensure you aren't spending time building something that we won't merge.

ok, no problem, i will create a discussion about it. 

> I think once v4 is released you could PR this feature again. I would suggest that instead of passing the entire `$state`, just the current repeater item state is passed to the function, and then you return whatever the pivot data should be for that item, instead of needing to map through the items.

it is not about repeater component, it is for select/checkboxlist with `multiple` flag. anyway, i will expand on that in the discussion.

PR #16114, opened 2025-04-24
Query builder german translations

andelkovic-damjan:
## Description

Updated German translations for the query builder

## Visual changes

No visual changes

## Functional changes

### DIFF
--- a/packages/tables/resources/lang/de/filters/query-builder.php
+++ b/packages/tables/resources/lang/de/filters/query-builder.php
@@ -33,7 +33,7 @@ return [
 
     ],
 
-    'no_rules' => '(Keine Bedinungen)',
+    'no_rules' => '(Keine Bedingungen)',
 
     'item_separators' => [
         'and' => 'UND',
@@ -107,7 +107,7 @@ return [
             'is_date' => [
 
                 'label' => [
-                    'direct' => 'ist Datum',
+                    'direct' => 'ist ein Datum',
                     'inverse' => 'ist kein Datum',
                 ],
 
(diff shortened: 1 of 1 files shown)


packages/tables/resources/lang/de/filters/query-builder.php
```diff
@@ -84,8 +84,8 @@
                 ],
 
                 'summary' => [
-                    'direct' => ':attribute is after :date',
-                    'inverse' => ':attribute is not after :date',
+                    'direct' => ':attribute ist nach :date',
+                    'inverse' => ':attribute ist vor :date',
```
pxlrbt: @danharrin Is there a reason the english version for `inverse` is "is not after :date" and not "is before :date". Is it inclusive of the date? Then "ist vor" wouldn't be the accurate translation and we'd need to use ":attribute ist nicht nach :date"

danharrin: It is inclusive, thats why.

pxlrbt: Then it should be `:attribute ist nicht nach :date` or `:attribute is vor oder am :date` which would translate to ":attribute is before or at :date"

andelkovic-damjan: Is it okay now? I am not actually German speaking.
Also should is_date label instead be ist ein Datum?

pxlrbt: > Is it okay now? I am not actually German speaking.
Also should is_date label instead be ist ein Datum?

Yes "ist ein Datum" would be better.
And while you are at it, could you fix `no_rules`. The current version has an `g` missing and should be:

`'no_rules' => '(Keine Bedingungen)',`

pxlrbt (approved):

PR #16124, opened 2025-04-25
Add globbing to theme stub

binaryfire:
Add glob patterns to theme stub so subdirectories are parsed

### DIFF
--- a/packages/panels/stubs/ThemeCss.stub
+++ b/packages/panels/stubs/ThemeCss.stub
@@ -1,4 +1,4 @@
 @import '../../../../vendor/filament/filament/resources/css/theme.css';
 
-@source '../../../../app/Filament{{ classDirectory }}';
-@source '../../../../resources/views/filament{{ viewDirectory }}';
+@source '../../../../app/Filament{{ classDirectory }}**/*';
+@source '../../../../resources/views/filament{{ viewDirectory }}**/*';
--- a/packages/panels/stubs/ThemeCss.stub
+++ b/packages/panels/stubs/ThemeCss.stub
@@ -1,4 +1,4 @@
 @import '../../../../vendor/filament/filament/resources/css/theme.css';
 
-@source '../../../../app/Filament{{ classDirectory }}**/*';
-@source '../../../../resources/views/filament{{ viewDirectory }}**/*';
+@source '../../../../app/Filament/{{ classDirectory }}**/*';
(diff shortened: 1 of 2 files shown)


packages/panels/stubs/ThemeCss.stub
```diff
@@ -1,4 +1,4 @@
 @import '../../../../vendor/filament/filament/resources/css/theme.css';
 
-@source '../../../../app/Filament{{ classDirectory }}';
-@source '../../../../resources/views/filament{{ viewDirectory }}';
+@source '../../../../app/Filament{{ classDirectory }}**/*';
+@source '../../../../resources/views/filament{{ viewDirectory }}**/*';
```
danharrin: Does the `classDirectory` and `viewDirectory` already end with `/`? Do we not need that before `**/*`?

danharrin (changes requested):

packages/panels/stubs/ThemeCss.stub
binaryfire: Yeah they already do.

binaryfire: Ah actually it works for panels in subdirs (which is what I'm using).

```
@import '../../../../vendor/filament/filament/resources/css/theme.css';

@source '../../../../app/Filament/AdminProjects/**/*';
@source '../../../../resources/views/filament/admin-projects/**/*';
```

But not in normal projects:
```
@import '../../../../vendor/filament/filament/resources/css/theme.css';

@source '../../../../app/Filament**/*';
@source '../../../../resources/views/filament**/*';
```

My bad, I'll take another look

danharrin: The variables should not end with `/`, the `/` should be added to the stub file

binaryfire: Works for both cases now:

No subdir:
```
@import '../../../../vendor/filament/filament/resources/css/theme.css';

@source '../../../../app/Filament/**/*';
@source '../../../../resources/views/filament/**/*';
```

Subdir:
```
@import '../../../../vendor/filament/filament/resources/css/theme.css';

@source '../../../../app/Filament/AdminProjects/**/*';
@source '../../../../resources/views/filament/admin-projects/**/*';
```

danharrin (approved):

PR #16130, opened 2025-04-25
feat: support Redirector type in Logout response

saade:
This _shouldn't_ break any existing applications since the docs does not tell users to `extend` the base class, but instead `implement` the contract.


danharrin:
Can you please explain the situation that causes this to be returned in a normal Filament app without this class overidden?

saade:
Gonna be real with you — I can’t remember exactly what the issue was 😅 but I do remember that [this exception](https://flareapp.io/share/87nkQN65) was being thrown, and this fix took care of it.

I haven’t seen it happen again, so I can’t really reproduce it to give you more context.

Either way, the PR is safe (the implementation is the same as the other responses), but totally fine to close it if you feel it’s unnecessary.

Thanks a ton! 🙌

danharrin (approved):

danharrin:
Thanks

PR #16154, opened 2025-04-28
Prevent 2FA setup modal from being closed accidentally

binaryfire:
Prevents accidentally breaking the 2FA setup flow or closing before recovery codes are copied.

### DIFF
--- a/packages/panels/src/Auth/MultiFactor/GoogleTwoFactor/Actions/SetUpGoogleTwoFactorAuthenticationAction.php
+++ b/packages/panels/src/Auth/MultiFactor/GoogleTwoFactor/Actions/SetUpGoogleTwoFactorAuthenticationAction.php
@@ -54,7 +54,6 @@ class SetUpGoogleTwoFactorAuthenticationAction
             ->modalWidth(Width::Large)
             ->closeModalByClickingAway(false)
             ->closeModalByEscaping(false)
-            ->modalCloseButton(false)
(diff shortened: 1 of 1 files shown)


packages/panels/src/Auth/MultiFactor/GoogleTwoFactor/Actions/SetUpGoogleTwoFactorAuthenticationAction.php
```diff
@@ -52,6 +52,9 @@ public static function make(GoogleTwoFactorAuthentication $googleTwoFactorAuthen
                 ]);
             })
             ->modalWidth(Width::Large)
+            ->closeModalByClickingAway(false)
+            ->closeModalByEscaping(false)
+            ->modalCloseButton(false)
```
danharrin: Personally I think we should keep the close button even if the other two are disabled, so that users can quit from the recovery code step without needing to go Back

danharrin (changes requested):

packages/panels/src/Auth/MultiFactor/GoogleTwoFactor/Actions/SetUpGoogleTwoFactorAuthenticationAction.php
binaryfire: Done

danharrin (approved):

PR #16157, opened 2025-04-29
fix: read only label date time picker

untillnesss:
## Description

Date Time Picker has been set to readOnly, if you click on the Label, it still opens the Date Time Picker dialog.

## Visual changes

What is currently happening.

https://github.com/user-attachments/assets/79bce719-29b2-4ae2-848f-df3ee584f24a

After repair

https://github.com/user-attachments/assets/c145a1df-1394-441d-a486-40c4a3750141

## Functional changes

### DIFF
--- a/packages/forms/resources/views/components/date-time-picker.blade.php
+++ b/packages/forms/resources/views/components/date-time-picker.blade.php
@@ -23,6 +23,7 @@
     :component="$getFieldWrapperView()"
     :field="$field"
     :inline-label-vertical-alignment="\Filament\Support\Enums\VerticalAlignment::Center"
+    :isReadOnly="$isReadOnly()"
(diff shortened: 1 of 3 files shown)


danharrin:
I think you should use `disabled()` instead of `readOnly()`

untillnesss:
But, if I use the `disabled()` option, it doesn't get entered into the `$data` array when the form is submitted.

danharrin:
Check the docs - https://filamentphp.com/docs/3.x/forms/fields/getting-started#disabling-a-field

Use `disabled()->dehydrated()`

untillnesss:
But, if DatePicker is readonly, it should not be able to open the Calendar dialog when its label is Clicked.

PR #16174, opened 2025-04-30
add missing dependency

wietsewarendorff:
As mentioned [here](https://github.com/filamentphp/filament/discussions/16163), I think the staudenmeir/belongs-to-through dependency is missing. I am not sure if this is the correct way to update the forms-subtree, let me know if I have to create this PR somewhere else...


danharrin:
It's not a dependency, it's optional. You can only use the feature if you have a relationship set up with that package anyway. It does not cause errors if it is missing AFAIK.

PR #16177, opened 2025-04-30
Add EmbeddedWidget

binaryfire:
Provides a consistent API for adding widgets to page content:

```PHP
    public function content(Schema $schema): Schema
    {
        return $schema
            ->components([
                $this->getFormContentComponent(),
                EmbeddedWidget::make(SummaryWidget::class, [
                    'total' => $total,
                ]),
                EmbeddedTable::make()
            ]);
    }
```

### DIFF
--- a/packages/schemas/src/Components/EmbeddedWidget.php
+++ b/packages/schemas/src/Components/EmbeddedWidget.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace Filament\Schemas\Components;
+
+use Closure;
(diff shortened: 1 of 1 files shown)


danharrin:
Hey, sorry but I don't really understand the point except the different name. The `Livewire::make()` method can be used for widgets without modifications. The `render()` method in your class is never called because it is never instantiated, and the `getLivewire()` method returns the parent component not the child.

PR #16181, opened 2025-05-01
Feat: Add Affix/Prefix support to MorphToSelect component

a21ns1g4ts:
## Description
This PR adds support for prefix and suffix elements in the MorphToSelect component, allowing for better visual integration with other form elements. The change addresses the need for consistent form field decoration across the application.

## Visual changes
Before:
[MorphToSelect without any affixes]

After:
[MorphToSelect with prefix icon and suffix prefix]

## Functional changes
- Added prefix and suffix props to MorphToSelect component
- Implemented affix positioning and styling
- Maintained all existing component functionality
- Updated component documentation with affix examples

- [] Code style has been fixed by running the `composer cs` command
- [] Documentation is up-to-date

### DIFF
--- a/packages/forms/src/Components/MorphToSelect.php
+++ b/packages/forms/src/Components/MorphToSelect.php
@@ -16,6 +16,7 @@ class MorphToSelect extends Component
     use Concerns\CanBeSearchable;
     use Concerns\HasLoadingMessage;
     use Concerns\HasName;
+    use Concerns\HasAffixes;
 
     /**
      * @var view-string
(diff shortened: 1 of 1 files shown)


danharrin:
Can you explain the use for this? And what made you choose applying it to the "key" select instead of the "type" select?

a21ns1g4ts:
I chose to apply it to the "key" because it was simpler, to be honest. But I think it could also be done in the "type" select. What do you think?

Alternatively, we could keep the functionality in both "key" and "type"—what's your opinion? I can implement it in "type" as well.

Using "key" was useful because the action I'm performing isn't specific to each type. But I don’t see a problem in adding it to "type" either, especially since I don’t have many types yet. However, if I had many types later, it might still be good to keep it in "key" as well.

What do you think?

image

danharrin:
Instead of adding affixes directly to the `MorphToSelect`, you should instead implement `modifyTypeSelectUsing()` and `modifyKeySelectUsing()` methods, where a user can pass something like `fn (Select $select) => $select->prefix('xxxxxx')` or any other customization method available. Then we dont need to keep a copy of every one and you can also specify which select it should apply to. Have a look at `modifyRecordSelectUsing()` in the `AttachAction` for inspiration on how to make the PR. Gonna close this one as it's a separate suggestion, but will welcome the new PR as described.

PR #16183, opened 2025-05-01
Add guide on using Tailwind CSS v3 and v4 with Vite

zepfietje:
## Description

As there still is a lot of confusion around using Filament with Tailwind CSS v3 in Laravel projects that have Tailwind CSS v4 installed, and the recommended approach of using npx doesn’t work well for all cases with plugins and themes, I wrote a guide on how to use Filament in Tailwind CSS v4 projects while simultaneously using Vite to compile Filament’s Tailwind CSS v3 styles.

To prevent more questions about this on Discord and GitHub Discussions, I added a link to the guide to the existing note in the documentation.

## Functional changes


danharrin (changes requested):
Instead of linking to an external article I think it's probably better if the guide/process is improved. Any changes to the `make:theme` command or the guide should not be breaking.

zepfietje:
Sure, I’d be happy to review a PR that implements the steps outlined in the guide! 😊

Perhaps it’s better to link to the guide in the meantime until the `make:theme` command has been updated?

alexandersix:
Hey @zepfietje - I appreciate the initiative of getting this article written!

I'm 100% in agreement with @danharrin here, this is the kind of content that doesn't actually fit into the "content" part of our content strategy–it would be better served as a direct update to the Filament theming guide or to the process of making a theme itself (which I think would be the best option, in my opinion).

With regards to your proposal of linking to the guide, due to the temporary nature of the link, I think that it would be best if we didn't add it to the site. My view on the content strategy for Filament is that our content should be permanent and useful. Even if it references previous versions of Filament, there is still use there for it to remain on the site for archival purposes or for the folks who are still _using_ previous versions of Filament. I also don't want to throw people for a loop by adding a piece of content to the site, just for them to return to the Filament site and not be able to find it later on.

I definitely do want to reiterate, though, that I think this would be incredibly useful as a direct update to the guide/process, especially in the run-up to Filament v4! I know I've had a few people mention how excited they are to upgrade specifically to use the new versions of Tailwind across their entire app (including Filament), so something like this would be a great little gift for folks in that specific situation 🎉

zepfietje:
Totally agree the best solution would be to update the theme creation process/command, @alexandersix! 😄

PR #16185, opened 2025-05-01
added id check to modal closed event

alperenersoy:
## Description

When we open another modal in an action modal and click the close button in the top right corner to close the child modal, it also closes the parent action modal and keeps the black overlay. 

Related plugin issue: https://github.com/alperenersoy/filament-export/issues/146

## Visual changes

Before:

[Screencast From 2025-05-01 21-29-34.webm](https://github.com/user-attachments/assets/b695bc92-5a39-499b-a3b3-d8b6cede2fe3)

After:

[Screencast From 2025-05-01 21-51-07.webm](https://github.com/user-attachments/assets/992d6190-7a64-4d26-9032-898def2509fa)

## Functional changes

### DIFF
--- a/packages/actions/resources/views/components/modals.blade.php
+++ b/packages/actions/resources/views/components/modals.blade.php
@@ -27,7 +27,7 @@
             :wire:key="$action ? $this->getId() . '.actions.' . $action->getName() . '.modal' : null"
             x-on:closed-form-component-action-modal.window="if (($event.detail.id === '{{ $this->getId() }}') && $wire.mountedActions.length) open()"
             x-on:modal-closed.stop="
-                if ($event.detail.id !== '{{ $this->getId() }}-action') {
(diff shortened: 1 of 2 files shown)


danharrin:
We don't really support nested modals where the modals are from two separate Livewire components, we only support nested modals within a Livewire component and I am afraid that this might break that, so I will need to test it thoroughly when I have time.

danharrin (changes requested):
Does it still work if you change the check to be "if `$event.detail.id` STARTS WITH `livewireId-` instead of matching the entire string? That would be safer change to make IMO

alperenersoy:
> Does it still work if you change the check to be "if `$event.detail.id` STARTS WITH `livewireId-` instead of matching the entire string? That would be safer change to make IMO

@danharrin It works the same way with `startsWith` as expected. Since we are using a string prototype function, I also added null conditionals to avoid errors.

danharrin (approved):

PR #16194, opened 2025-05-02
Add `->userMenu()` for disabling user menu

binaryfire:
There are times when you may want a topbar without a user menu. For example, I use a dedicated panel for application settings with a simple "Back" button to return to the main dashboard:

image

This would also be useful for a dedicated onboarding panel etc.

The only way to achieve this at the moment is by hiding the user menu dropdown trigger with CSS. This PR adds a `->userMenu()` method for disabling the user menu entirely, similar to the `->topBar()` method.

### DIFF
--- a/packages/panels/resources/views/livewire/simple-user-menu.blade.php
+++ b/packages/panels/resources/views/livewire/simple-user-menu.blade.php
@@ -1,7 +1,5 @@
 <div>
-    @if (filament()->hasUserMenu())
-        <x-filament-panels::user-menu />
-    @endif
+    <x-filament-panels::user-menu />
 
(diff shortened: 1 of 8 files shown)


packages/panels/resources/views/livewire/simple-user-menu.blade.php
danharrin, line 1: Please move this check so the entire `SimpleUserMenu` LW component is not rendered if its false

binaryfire: Done

danharrin (approved):

PR #16230, opened 2025-05-09
Gate password reset process with canAccessPanel logic

robinmalburn:
## Description

Currently, the password reset and password reset request process don't check the the `canAccessPanel()` method on the FilamentUser contract, which leads to a few possible issues:

- User experience inconsistency: A user can successfully reset their password, only to be told their credentials are invalid when attempting to log in.
- Unintended interactions with notifications / events: Where custom notifications or events are in use in a multi-panel configuration, actions may be taken against users who should not be eligible to receive those notifications / events.

## Visual changes

 Existing password reset request process - user requesting a reset on a panel they shouldn't have access to.

https://github.com/user-attachments/assets/d8706521-5f0d-4ff5-9af4-c3abe5f81795

Existing password reset process - user completing reset on a pabel they shouldn't have access to, along with subsequent login attempt following reset.

https://github.com/user-attachments/assets/f43015c4-2226-4efd-8cb6-d6cf1dcbc530

New password reset request process - user requesting a reset, now receives an ambiguous notification so as to not bleed information about whether a notification has been sent or not.

https://github.com/user-attachments/assets/b3247ce8-8a44-4891-ba19-1d81b616bf61

New password reset process - when a user attempts to reset a password on a panel they do not have access to, for example where access has been removed _after_ the reset request was raised, the user is notified that the user account is invalid for the request.

https://github.com/user-attachments/assets/1a911082-b9f8-4e4e-888e-e18f7b7f65a2

## Functional changes

The process has been updated with the following changes:

- Password reset request:
  -  This process checks that a user is a valid instance of a FilamentUser contract, and checks that the user has access to the current panel via the `canAccessPanel` method.  A password reset request is only dispatched where a user as valid access to the panel.
  - Unless requests are throttled, a standard notification is used to advise that an email will have been sent if the email is valid in the system, to avoid leaking user access information.
  - Tests updated to reflect these changes.
- Password reset:
  - This process checks that a user is a valid instance of a FilamentUser contract, and checks that the user has access to the current panel via the `canAccessPanel` method. 
  - Where a user has valid access to the panel, the password is updated as normal.
  - Where a user does not have valid access to the panel, e.g. where access was revoked after a reset request was generated, the user is notified that the user is invalid.
  - Tests updated to reflect these changes.

### DIFF
--- a/packages/panels/resources/lang/en/pages/auth/password-reset/request-password-reset.php
+++ b/packages/panels/resources/lang/en/pages/auth/password-reset/request-password-reset.php
@@ -37,6 +37,8 @@ return [
             'body' => 'Please try again in :seconds seconds.',
         ],
 
+        'sent' => 'If that email address is valid in our records, we will send you an email to reset your password.',
+
     ],
 
 ];
--- a/packages/panels/resources/lang/en/pages/auth/password-reset/request-password-reset.php
+++ b/packages/panels/resources/lang/en/pages/auth/password-reset/request-password-reset.php
@@ -32,12 +32,14 @@ return [
 
     'notifications' => [
 
+        'sent' => [
+            'body' => 'If your account doesn\'t exist, you will not receive the email.',
+        ],
+
         'throttled' => [
             'title' => 'Too many requests',
             'body' => 'Please try again in :seconds seconds.',
         ],
-
-        'sent' => 'If that email address is valid in our records, we will send you an email to reset your password.',
 
     ],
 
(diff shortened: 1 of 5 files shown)


eyesfivemserver:
The performance optimizations here are quite clever. Nice engineering work!

danharrin (approved):

danharrin:
Thanks! I've tweaked the notification so it still uses the old translation, but adds a body below as extra information. It also ensures that the title is not too long

PR #16240, opened 2025-05-10
Update Custom Filter Docs

bmckay959:
I think making mention of this in the upgrade guide might be helpful.

With my project, I ended up manually updating all the resource forms with this(The upgrade script was putting the wrong return type on all of them)

## Description

New custom filters should move away from `->schema()` in favor of `->form()`


ayestoro:
Great implementation of this feature. I particularly like the approach to error handling.

fivemeyestoreserver:
Interesting solution to this problem! I've been working on similar challenges recently.

PR #16241, opened 2025-05-11
Fix: ensure getType() always returns valid string

rtraselbd:
## Description

The `getType()` method can currently return invalid values (e.g., `null`, array, or other non-string values) when `$this->type` is evaluated as a closure or dynamic value. This violates the expected return type of `string` and may lead to unexpected behavior in form components.

This PR ensures that `getType()` always returns a valid string by:

* Checking if the evaluated `$type` is a non-empty string before returning it.

---

## Visual changes

N/A – No visual changes introduced. This is a backend logic improvement.

---

## Functional changes

* [x] Code style has been fixed by running the `composer cs` command.
* [x] Changes have been tested to not break existing functionality.
* [x] Documentation is up-to-date (not applicable to this logic-only change).

### DIFF
--- a/packages/forms/src/Components/TextInput.php
+++ b/packages/forms/src/Components/TextInput.php
@@ -213,7 +213,9 @@ class TextInput extends Field implements CanHaveNumericState, Contracts\CanBeLen
 
     public function getType(): string
     {
-        if ($type = $this->evaluate($this->type)) {
+        $type = $this->evaluate($this->type);
+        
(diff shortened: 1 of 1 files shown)


danharrin:
It should throw an exception (as it does now) instead of silently failing like this - the evaluated function is in the users code so they can fix it to return a string

danharrin:
If I am missing something please let me know

PR #16247, opened 2025-05-11
Optional HTML validation messages

binaryfire:
Adds support for HTML validation messages. This is opt-in and therefore secure by default. HTML rendering needs to be enabled via `->htmlValidationMessages()`.

There are times where you want custom rules to return HTML, for example to format the output as a list:
image

Or to add a link:
https://youtu.be/_k4zwJIUPH0?si=WOjRrjI0jm1ZuQ8p&t=58
image

The message wrapper uses a `<div>` instead of a `<p>` when `->htmlValidationMessages()` is used. This doesn't break anything for non-HTML messages:

image

It might be simpler to just use `<div>` for both?


danharrin (changes requested):
Is there a way we can allow `HtmlString` objects to be used as validation messages, instead of needing to opt-in to HTML rendering across all messages?

binaryfire:
The first thing I tried was using `HtmlString` with an `instanceof HtmlString` check in the view. The problem is it's converted to a string:

https://github.com/laravel/framework/blob/bcfa7c32acc17a7e49892efc5040fcaae3f207da/src/Illuminate/Contracts/Validation/ValidationRule.php#L14

https://github.com/laravel/framework/blob/bcfa7c32acc17a7e49892efc5040fcaae3f207da/src/Illuminate/Support/HtmlString.php#L62

So AFAIK that's not possible?

binaryfire:
I've renamed things to be consistent with `allowHtml()` and added the same XSS warning to the docs. 

I've been trying to think of when not escaping validation messages would be a security issue. Only thing I can think of is if messages are using `:input`, but that would only allow a malicious user to inject JS for themselves (which is both harmless and pointless). I've grilled AI as well and it hasn't been able to come up with anything.

No harm in escaping by default, but I don't think HTML rendering for a field is an issue. Either way, it's opt-in and the same warning as `allowHtml()` is there now.

danharrin (approved):

PR #16254, opened 2025-05-12
feat: ability to evaluate Closures in `viewData()`

ralphjsmit:
I wanted to insert a Blade view into a form and I noticed that surprisingly it was not possible to calculate the `viewData()` using a closure. As the `$viewData` property is an array that is merged on every call to the `viewData()` method, a convention should be found to support these closures.

I implemented this in the same way as e.g. the `->hintActions()` method, which accepts either an array of closures or an array of actions. In this case, the `viewData()` method will accept a regular array or an array with closures. If a closure is found on a numeric key, it is assumed the key means nothing and the end result is flat-merged to the top-level. If an closure is found with a string key, the closure is evaluated and set as the result of the original string key. See also the tests for examples of the three scenarios.

Thanks!

### DIFF
--- a/packages/notifications/src/Notification.php
+++ b/packages/notifications/src/Notification.php
@@ -57,14 +57,6 @@ class Notification extends ViewComponent implements Arrayable
         return $static;
     }
 
-    /**
-     * @return array<string, mixed>
-     */
-    public function getViewData(): array
-    {
-        return $this->viewData;
-    }
-
     public function toArray(): array
     {
         return [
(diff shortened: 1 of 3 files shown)


danharrin (changes requested):
Does it not make more sense in this case to allow array or closure instead of an array of closures, and then you store the array of arrays and closures. In the getter, you then evaluate the closures and flatten the array by one level to merge it?

ralphjsmit:
Yes, that would be a great option as well. I just changed it, but will check the failing test tomorrow .

ralphjsmit:
All good now, I removed the override of the `getViewData()` method in `Notification` class.

danharrin (approved):

PR #16262, opened 2025-05-14
fix(infolists): avoid unnecessary spacing when no label content is used

mvenghaus:
## Description

When the label is hidden there is a small spacing because of the empty label container.

## Visual changes

image

## Functional changes


danharrin:
The element still needs to be there when `inlineLabel()` is used

Also does this change also need to be applied to form fields?

mvenghaus:
> The element still needs to be there when `inlineLabel()` is used
> 
> Also does this change also need to be applied to form fields?

when inlineLabel is used then the label should have content and therefore be displayed, or?

yes .. same for Form Fields

danharrin:
No, the container needs to be there to preserve spacing. Multiple `inlineLabel()` can be stacked on top of each other which creates a two column effect.

mvenghaus:
@danharrin okay .. i've pushed an update and added the same for forms

danharrin (approved):

danharrin:
Thanks

PR #16271, opened 2025-05-14
Disable step button if previous action is disabled

leandrogehlen:
## Description

This pull request fixes a bug in the wizard component (`wizard.blade.php`) where the step selector allowed selection even when the `previousAction` was disabled. This behavior was unintended and has now been corrected to ensure proper functionality.


leandrogehlen:
There are multiple issues related to the navigation tabs.

When using tabs, any customizations made to actions — such as label, color, or other properties — are discarded or not applied correctly.

Suggestion: A possible solution would be to disable direct tab selection and enforce navigation only through button actions. This could help preserve action customizations and ensure consistent behavior across tabs.

PR #16275, opened 2025-05-15
Add Optional Truncate-Table Support to ImportAction

manuzoli:
## Description
This update introduces an optional feature to truncate the target database table before executing an import. This is particularly useful when Filament is not the source of truth, and external systems are providing the authoritative data set. In such cases, ensuring the database only contains the latest imported records is critical to avoid data inconsistencies.

You can now use the new ->truncateBeforeImport(true) method on the ImportAction to enable this behavior:
 
```
ImportAction::make()
    ->importer(ProductImporter::class)
    ->truncateBeforeImport(true)
```

This gives developers fine-grained control over the import behavior when syncing external data.

## Visual changes
there are no visual changes. 

## Functional changes
- [x] Added truncateBeforeImport(bool $shouldTruncate) method to ImportAction.
- [x] Automatically truncates the relevant table when the flag is set to true.
- [x] Documentation has been updated accordingly.

### DIFF
--- a/packages/actions/docs/07-prebuilt-actions/08-import.md
+++ b/packages/actions/docs/07-prebuilt-actions/08-import.md
@@ -601,6 +601,16 @@ ImportAction::make()
 
 If you are encountering memory or timeout issues when importing large CSV files, you may wish to reduce the chunk size.
 
+## Deleting all existing data before an import
+
+To remove all existing records before running an import, enable the `truncateBeforeImport()` option on the import action:
+
+```php
+ImportAction::make()
+    ->importer(ProductImporter::class)
+    ->truncateBeforeImport(true)
+```
+
 ## Changing the CSV delimiter
 
 The default delimiter for CSVs is the comma (`,`). If your import uses a different delimiter, you may call the `csvDelimiter()` method on the action, passing a new one:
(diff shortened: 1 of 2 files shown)


danharrin:

In this case, I don't think it warrants a new feature. Can you do this in the `before()` function of the action?

manuzoli:
Thank you very much for your reply. I only found the contributing guide after I had already submitted the pull request—my apologies.

The before() function appears to be called before every entry, based on my testing. What I needed was something that gets called just once, right before the first import. I couldn't get that to work. I'll use my changes as a patch for now and hope to contribute this feature in the 4.x version.

danharrin:
Sorry, what do you mean by "before every entry"?

danharrin:
The hook will run once, before the `action` function is run

manuzoli:
Thanks for your response! I’ll take another look at it—mine was triggering for every entry, but it’s possible I made a mistake somewhere. I’ll double-check and follow up here.

If it turns out to be working as expected, my pull request might not be necessary after all. That said, I still believe the change could improve simplicity for some developers, maybe in v4.x.

danharrin:
Since this is just a one line change I don't think I'd add a dedicated method for it, considering I can't actually think of many reasons why you would want to truncate before an import in a production system.

PR #16276, opened 2025-05-16
fix(section): ensure action motion with non-English modal headings

GabrielShell:
Previously, non-English characters in Section Action's ModalHeading caused motion effects failure due to invalid ID generation. This Commit Solve the proplem.

## Description
Resolves invalid ID generation in Section Action components when using non-English ModalHeadings, which caused motion effects failure.

## Functional changes


danharrin (commented):
Please provide an example of a heading that breaks this.

danharrin:
Please open an issue or provide more information so I can understand the problem, otherwise I am unsure if this is the correct fix

Issue #12824, opened 2024-05-16
KeyValue field updates state from the second time


finoghentov:
### Package

filament/filament

### Package Version

3.2

### Laravel Version

10.10

### Livewire Version

3.4

### PHP Version

8.1

### Problem description

I have a simple resource editing form.

```php
KeyValue::make('options'),
Select::make('type')
  ->options(array_column(BoostType::cases(), 'name'))
  ->afterStateUpdated(function (Component $component, Set $set) {
        $set('options', ['test' => rand(0, 1000)]);
  })
  ->nullable(false)
  ->live()
```
Key value update state only from the second time

The problem in this [script](https://github.com/filamentphp/filament/blob/3.x/packages/forms/resources/js/components/key-value.js)

Initializing the component with NON empty value triggers `updateState` method which switches `shouldUpdateRows` flag to false

https://github.com/filamentphp/filament/blob/3.x/packages/forms/resources/js/components/key-value.js#L15
```js
if (this.rows.length <= 0) {
    this.rows.push({ key: '', value: '' })
} else {
    this.updateState()
}
```
https://github.com/filamentphp/filament/blob/3.x/packages/forms/resources/js/components/key-value.js#L105
```js
this.shouldUpdateRows = false
```

While first updating field state we trigger next code which switch flag to true
https://github.com/filamentphp/filament/blob/3.x/packages/forms/resources/js/components/key-value.js#L72
```js
if (!this.shouldUpdateRows) {
    this.shouldUpdateRows = true
    return
}
```

I've found that this was required to fix this [bug](https://github.com/filamentphp/filament/issues/1107) but we need to find a better decision. Maybe something like this:
```js
updateState: function () {
    let state = {}

    let keys = [];

    for (let i in this.rows) {
        if (
            keys.includes(this.rows[i].key) ||
            this.rows[i].key === '' ||
            this.rows[i].key === null
        ) {
            return;
        }

        keys.push(this.rows[i].key);
        state[this.rows[i].key] = this.rows[i].value
    }

    this.state = state
},
```
Remove `shouldUpdate` flag and do not update the state if we've found empty or duplicated keys. But maybe we should make a warning to make users understand that with duplicated keys state won't be updated

### Expected behavior

I expect that changing the Select value will update the KeyValue state, but it works only If I change Select 2 times. 
It works only if you are updating the resource with some filled value in KeyValue.

### Steps to reproduce

- Create resource
- Paste example code in resource editing form
[...]

Issue #1107, opened 2022-01-02
KeyValue form component erase values when entering new matching key


telkins:
### Package

filament/filament

### Package Version

v2.7.14

### Laravel Version

v8.77.1

### Livewire Version

v2.8.2

### PHP Version

8.1.0

### Bug description

The KeyValue form component seems to erase values when entering a new key if it matches a pre-existing key...even if you're not yet done typing.

### Steps to reproduce

Let's say I have an array that already has `'key' => 'value'` and I add a new row to add another set: `'key_code' => 'abc123'`.

Once I start to type the new key in, as soon as it matches the existing key, it clears the existing value and clears out the new row.

danharrin:
Fixed by #1113.

PR #16290, opened 2025-05-19
Fix/improve key-value component functionality

alex-elivate:
## Description
**updateRows**: I’m including any changes that the user is currently making even if invalid.
**updateState**: Sync the rows but only the ones that don’t have an issue. (@finoghentov solution) 
**Validation**: However if a user enters a duplicate key/value and submits, the responsibility of checking this falls on the backend validation. This will not generate an error if a user enters a duplicate key and saves the form, it will silently skip the attempt. If we want front end validation I propose a new method to handle this.

https://github.com/filamentphp/filament/issues/12824
Fixed the issue of not updating the first state change.

https://github.com/filamentphp/filament/issues/1107
Maintained the fix for temporary duplicate keys.

## Visual changes

https://github.com/user-attachments/assets/b9ab1ae1-aabb-42b5-bacc-bf1dd378fe55

## Functional changes
Removed temporary hack and added logic for tracking changes and updating state and UI.

I ran the Pests as far as it would go before getting a memory error. No tests failed.


danharrin:
Thanks for your work here!

In the past, I have tried to make similar fixes to the KeyValue and I have always ended up having to abandon them, as they end up breaking other cases. I feel like it is quite a fragile component to change.

As such, I would feel more comfortable making these changes in v4 in case they break something. Would you please test v4 to see if the issue still persists, and if it does copy the changes over to there? Many thanks

PR #16293, opened 2025-05-19
Correct discover_app_classes for Windows paths

ConnorHowell:
## Description

When attempting to create a new resource it is currently failing to detect class names when running the command on a Windows machine due to the use of forward slashes when filtering out classes from the composer autoload.

**Output of `base_path('vendor/composer/../../')`:**
```
D:\Sites\filamentv4\vendor/composer/../..
```
**Path returned by composer autoload class map:**
```
D:\Sites\filamentv4\vendor\composer/../../app/Models/User.php
```

Using the built-in PHP constant for "DIRECTORY_SEPARATOR" in the path fixes it, not sure why composer only uses backslashes on windows until just before moving up a directory...

**Output of `base_path('vendor' . DIRECTORY_SEPARATOR . 'composer/../../')`:**
```
D:\Sites\filamentv4\vendor\composer/../..
```

## Functional changes

### DIFF
--- a/packages/support/src/helpers.php
+++ b/packages/support/src/helpers.php
@@ -301,7 +301,7 @@ if (! function_exists('Filament\Support\discover_app_classes')) {
 
         return collect($classLoader->getClassMap())
             ->filter(function (string $file, string $class) use ($parentClass): bool {
-                if (! str($file)->startsWith(base_path('vendor/composer/../../'))) {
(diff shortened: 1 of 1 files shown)


ConnorHowell:
@danharrin I should have mentioned in the title, this fix is for v4 by the way😃

danharrin (approved):

PR #16295, opened 2025-05-19
Hidden fields incorrectly dehydrated in Section with state path

alanost:
## Description

This PR addresses a bug where conditionally hidden fields within a Section component (with state path) are incorrectly dehydrated. Currently, when a field is both:
- Hidden based on a condition
- Inside a Section with a state path
- Marked as non-dehydrated

The field's value is still being dehydrated when it should not be.

Example scenario:
```php
Section::make("User Information")
    ->statePath('data')
    ->schema([
        Radio::make('show_input')
            ->label('Control Input Visibility')
            ->options([
                 'show' => 'Show',
                  'hide' => 'Hide',
             ])
              ->live(),,
        TextInput::make('name')
            ->hidden(fn (Get $get): bool => $get('show_input') === 'hide')
            ->dehydrated(fn (Get $get): bool => $get('show_input') === 'show')
    ])
```

When `show_input` is 'hide', the 'name' field is still being dehydrated despite being hidden and explicitly configured not to dehydrate.

I couldn't fix the bug further, but I created a test to demonstrate this issue

## Functional changes
- [x] Added test case demonstrating the bug

### DIFF
--- a/packages/forms/src/Components/Concerns/HasState.php
+++ b/packages/forms/src/Components/Concerns/HasState.php
@@ -186,14 +186,10 @@ trait HasState
             return;
         }
 
-        if ($this->getStatePath(isAbsolute: false)) {
+        if ($this->hasStatePath()) {
             foreach ($this->getStateToDehydrate() as $key => $value) {
                 Arr::set($state, $key, $value);
             }
-        }
-
-        if ($this->isHiddenAndNotDehydrated()) {
-            return;
         }
 
         foreach ($this->getChildComponentContainers(withHidden: true) as $container) {
@@ -438,7 +434,11 @@ trait HasState
 
     public function isDehydrated(): bool
     {
-        return (bool) $this->evaluate($this->isDehydrated);
+        if (! $this->evaluate($this->isDehydrated)) {
+            return false;
+        }
+
+        return ! $this->isHiddenAndNotDehydrated();
     }
 
     public function isDehydratedWhenHidden(): bool
--- a/packages/forms/src/Concerns/HasState.php
+++ b/packages/forms/src/Concerns/HasState.php
@@ -72,10 +72,6 @@ trait HasState
     public function dehydrateState(array &$state = [], bool $isDehydrated = true): array
     {
         foreach ($this->getComponents(withHidden: true) as $component) {
-            if ($component->isHiddenAndNotDehydrated()) {
-                continue;
-            }
-
             $component->dehydrateState($state, $isDehydrated);
         }
 
@@ -89,10 +85,6 @@ trait HasState
     public function mutateDehydratedState(array &$state = []): array
     {
         foreach ($this->getComponents(withHidden: true) as $component) {
-            if ($component->isHiddenAndNotDehydrated()) {
-                continue;
-            }
-
             if (! $component->isDehydrated()) {
                 continue;
             }
--- a/tests/src/Forms/StateTest.php
+++ b/tests/src/Forms/StateTest.php
@@ -7,6 +7,9 @@ use Filament\Forms\Get;
 use Filament\Tests\Forms\Fixtures\Livewire;
 use Filament\Tests\TestCase;
 use Illuminate\Support\Str;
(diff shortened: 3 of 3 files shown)


DirtyRacer1337:
this kind one broke for me now
```php
Section::make('Old')
    ->statePath('old')
    ->hidden($conditions)
    ->schema([
            TextInput::make('name')
                ->disabled(true)
                ->hidden(fn (mixed $state): bool => !$state)
                ->dehydrated(),
]),
Section::make('New')
    ->statePath('new')
    ->hidden($conditions)
    ->schema([
            TextInput::make('name')
                ->disabled(false)
                ->hidden(fn (mixed $state): bool => !$state)
                ->dehydrated(),
]),
```

how to make it now save both sections? both of them empty now

yannisc:
This change broke my app.

I had an action SendPaymentNotification::dispatch($data);

where $data included amount_due, order_id and order_name from the form:

                Action::make('SendPaymentNotification')
                ->label('Send Payment Notification')
                ->icon('heroicon-o-bell')
                ->visible(fn(Order $order) => $order->account->orders()->where('amount_due', '>', 0)->exists())
                ->form([
                    Wizard::make([
                        Step::make('Orders')
                            ->label('Orders')
                            ->schema([
                                Repeater::make('Order Amounts')
                                ->label('Amount Dues')
                                ->columns(5)
                                ->reorderable(false)
                                ->addable(false)
                                ->deletable(false)
                                ->default(function(Order $order) {
                                    return array_map(function($order) {
                                        return [
                                            'order_id' => $order['id'],
                                            'order_name' => $order['name'],
                                            'amount_due' => $order['amount_due']
                                        ];
                                    }, $order->account
                                        ->orders()
                                        ->where('amount_due', '>', 0)
                                        ->where('cancelled', false)
                                        ->get(['id', 'name', 'amount_due'])->toArray()
                                    );
                                })
                                ->schema([
                                    TextInput::make('order_id')
                                        ->label('id')
                                        ->hidden()
                                        ->disabled(),
                                    TextInput::make('order_name')
                                        ->disabled()
                                        ->columnSpan(4)
                                        ->label('Order'),
                                    TextInput::make('amount_due')
                                        ->label('Amount')
                                        ->numeric()
                                        ->rules('numeric')
                                        ->suffix('€')
                                ]),
                            ]),

but after the update, $data does not include amount_due and order_id.

danharrin:
It's a bug that `order_id ` was dehydrated before the change previously, since that is a hidden field. But I don't know why `amount_due` is not included. Is that hidden in some other way?

yannisc:
No, it's not hidden, but it's in the first step of the form wizard.

danharrin:
Are there other steps in the wizard?

yannisc:
yes, there are 3 next steps

danharrin:
Do they have other form components with the same name as the repeater?

danharrin:
@yannisc does this fix your problem? https://github.com/filamentphp/filament/pull/16340

It hasn't been released so you will need to make the changes manually in /vendor or use a local fork of Filament to test

yannisc:
let me check...

yannisc:
no, same error.

danharrin:
Please create a reproduction repository so I can work out what's going on in your case

jbardnz:
This change also seemed to have caused an issue with my setup:

```
Toggle::make("is_conditional")
    ->reactive(),

Radio::make("type")
    ->options([
        "conditional_mapping_or" => "ANY condition below is met (OR)",
        "conditional_mapping" => "ALL conditions below are met (AND)",
    ])
    ->default("mapping")
    ->required()
    ->visible(fn(Get $get) => $get("is_conditional"))
    ->live(),
```


Here I have a hidden radio field inside a Repeater. Before this update if `is_conditional` was set to `false` the default value of `mapping` would be saved in my JSON. 

After this update the `type` property is not saved in my JSON at all if the field is hidden.

danharrin:
Yeah that was the bug we fixed here, unless explicitly marked hidden fields should not be saved

PR #16311, opened 2025-05-21
fix path when creating custom field/entry

andcarpi:
## Description

When using a command to create a custom field or entry the namespace is different from the file path, resulting in the component not being autoloaded/found.

Namespace: App\Filament\Infolist\Components
Path: App\Infolist\Components

Imagem do WhatsApp de 2025-05-20 à(s) 12 07 35_8ab6e775

This commit makes the path equal to the NS as App\Filament\Infolist\Components

## Visual changes

No changes

## Functional changes


danharrin:
Thanks, but I fixed in a different place

PR #16312, opened 2025-05-21
add missing generics for eloquent-builder

wietsewarendorff:
## Description

add missing generics for eloquent-builder

## Functional changes


danharrin:
Can you help me understand, in what way are these providing more information, since `Model` is the most generic you can get? You can't have a query builder without a model.

wietsewarendorff:
you are so right, what was I thining... let me close this one and I will create a new one with the templates

PR #16319, opened 2025-05-22
Updated `getColumns` method to accept an `array $options` parameter

paulohps:
Updated `getColumns` method to accept an `array $options` parameter, enabling more flexible column configuration. Adjusted relevant calls and logic to pass options where necessary. This improves customizability of the export functionality.

## Description

This pull request solves the problem of passing the action options to the getColumns method of the export class, this allows generating export with dynamic columns based on the passed options.

## Visual changes

There were no visual changes.

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/CanExportRecords.php
+++ b/packages/actions/src/Concerns/CanExportRecords.php
@@ -98,7 +98,7 @@ trait CanExportRecords
                         ])
                             ->verticallyAlignCenter()
                             ->statePath($column->getName()),
-                        $action->getExporter()::getColumns(),
+                        $action->getExporter()::getColumns($this->getOptions()),
                     );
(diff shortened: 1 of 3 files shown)


danharrin:
I am quite sure that we cannot do this in v3 since it is a breaking change. Methods that extend their parents must implement all their parent's arguments, and anyone currently implementing this method would then not have the argument.

Please submit a PR for v4, and include a change to the upgrade script that adds the new parameter to the correct classes.

Please also consider if this needs to be added for imports as well as exports

Please also let me know if I am incorrect in my first statement

rguttersohn:
Doesn't look like this was added in v4 either.  Is there anyway to create dynamic export columns based on the current record or other dynamic data?

PR #16324, opened 2025-05-24
Fix Select relationship field generating empty Rule::in() when using JSON key in title (e.g., 'name->en')

thaqebon:
## Description
When a relationship title attribute uses JSON path notation (e.g., `'name->en'`), Laravel's `getAttributeValue('name->en')` cannot resolve the value.
As a result, `getOptionLabel()` returns `null`, which leads to returning an empty array for `Rule::in()`. This breaks validation by allowing no valid values.

https://github.com/filamentphp/filament/blob/b302607a77d871330f56934168c202fe6dc64acc/packages/forms/src/Components/Select.php#L1385-L1398

## Functional changes

### DIFF
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -948,7 +948,13 @@ class Select extends Field implements Contracts\CanDisableOptions, Contracts\Has
                 return $component->getOptionLabelFromRecord($record);
             }
 
-            return $record->getAttributeValue($component->getRelationshipTitleAttribute());
+            $relationshipTitleAttribute = $component->getRelationshipTitleAttribute();
+
+            if (str_contains($relationshipTitleAttribute, '->')) {
+                return Arr::get($record->toArray(), str_replace('->', '.', $relationshipTitleAttribute));
+            }
+
+            return $record->getAttributeValue($relationshipTitleAttribute);
         });
 
         $this->getSelectedRecordUsing(static function (Select $component, $state) use ($modifyQueryUsing): ?Model {
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -951,7 +951,7 @@ class Select extends Field implements Contracts\CanDisableOptions, Contracts\Has
             $relationshipTitleAttribute = $component->getRelationshipTitleAttribute();
 
             if (str_contains($relationshipTitleAttribute, '->')) {
-                return Arr::get($record->toArray(), str_replace('->', '.', $relationshipTitleAttribute));
+                return data_get($record->toArray(), str_replace('->', '.', $relationshipTitleAttribute));
             }
 
             return $record->getAttributeValue($relationshipTitleAttribute);
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -951,7 +951,7 @@ class Select extends Field implements Contracts\CanDisableOptions, Contracts\Has
             $relationshipTitleAttribute = $component->getRelationshipTitleAttribute();
 
             if (str_contains($relationshipTitleAttribute, '->')) {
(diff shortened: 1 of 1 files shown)


packages/forms/src/Components/Select.php
```diff
@@ -948,7 +948,13 @@ public function relationship(string | Closure | null $name = null, string | Clos
                 return $component->getOptionLabelFromRecord($record);
             }
 
-            return $record->getAttributeValue($component->getRelationshipTitleAttribute());
+            $relationshipTitleAttribute = $component->getRelationshipTitleAttribute();
+
+            if (str_contains($relationshipTitleAttribute, '->')) {
+                return Arr::get($record->toArray(), str_replace('->', '.', $relationshipTitleAttribute));
```
danharrin: Does this work?

```suggestion
                return data_get($record, str_replace('->', '.', $relationshipTitleAttribute));
```

danharrin (changes requested):

packages/forms/src/Components/Select.php
danharrin: Also I think this needs changing in `getOptionLabelsUsing()`?

danharrin: @thaqebon I can see you pushed a commit, but please check my suggestion again to see if the `toArray()` is actually required, and the other suggestion about `getOptionLabelsUsing()`

thaqebon:
> @thaqebon I can see you pushed a commit, but please check my suggestion again to see if the `toArray()` is actually required, and the other suggestion about `getOptionLabelsUsing()`

I already replied to you (as review), but no problem.

> Does this work?

It works with `data_get()`, but we still need to convert the record using `toArray()`. This is because `toArray()` returns all record data after casting and without any manipulation, allowing us to fetch the key properly.
I'm using Spatie Translatable, which under the hood returns the default language as a plain string, That means it's no longer an array or Json, so we can't extract keys from it directly.

>Also I think this needs changing in `getOptionLabelsUsing()`?

When multiple select is enabled, `getOptionLabelsUsing()` works correctly because it fetches labels and IDs from the database, which supports extracting JSON keys using the `->` notation.

danharrin:
> I already replied to you (as review), but no problem.

Sorry I don't think you submitted it, it's probably pending.

> without any manipulation

Which manipulation in this case is causing problems? IMO we probably do want properties to be manipulated by casts and getters

thaqebon:
In most cases, we don't need to use `toArray()`, However, when using packages that manipulate data (like the Spatie Translatable package) accessing properties directly may not work as expected.

For example, if we have a `title` property with two language keys (`'en'` and `'ar'`):
```php
// WITHOUT spatie translatable
$record->title; // Returns: ['en' => 'Welcome', 'ar' => 'مرحبًا']
data_get($record, str_replace('->', '.', $relationshipTitleAttribute)); // This will work.

// WITH spatie translatable
$record->title; // Returns: 'Welcome' (default locale value)
data_get($record, str_replace('->', '.', $relationshipTitleAttribute)); // This will NOT work

// with `toArray()` it will work in all cases
$record->toArray(); // Return ['title' => ['en' => 'Welcome', 'ar' => 'مرحبًا']]
data_get($record->toArray(), str_replace('->', '.', $relationshipTitleAttribute)); // This will work
```
So, in this case, using `toArray()` helps because it gives you the raw data structure after casting (from json to array), allowing you to access the full array of translations.

danharrin (changes requested):
I feel like this case with Spatie Translatable is very different to the standard use case, and I am concerned about using `toArray()` since it ends up transforming all attributes and relationships which will be slow on large datasets.

Please proceed with removing the `toArray()` for this reason. In your case, I would suggest using `getOptionLabelFromRecordUsing()` to provide a function which retrieves the translated label correctly

thaqebon:
Done.

danharrin (approved):

danharrin:
Thanks

PR #16236, opened 2025-05-09
Enable paste in FileUpload component

moizmasood10:
## Description

This pull request enables the ability for users to paste files directly into the FileUpload component by using the pasteable() method. Previously, pasting files was not supported. This functionality improves the user experience by allowing files to be pasted from the clipboard into the upload area, making the component more intuitive. This was done by simply making a method that enables the allowPaste property of Filepond.

## Visual changes

https://github.com/user-attachments/assets/79b1d8bf-8829-4e40-9bba-042bafbc2212

## Functional changes

- [x] Code style has been fixed by running the `composer cs` command.
- [x] Changes have been tested to not break existing functionality.
- [x] Documentation is up-to-date.

PR #16345, opened 2025-05-28
Added isPasteable option to File Upload 

bretto36:
## Description

Added an isPasteable parameter to match isDeletable

## Functional changes

This should allow people to specify ->isPasteable on the Filepond plugin

### DIFF
--- a/packages/forms/docs/03-fields/09-file-upload.md
+++ b/packages/forms/docs/03-fields/09-file-upload.md
@@ -412,6 +412,17 @@ FileUpload::make('attachment')
     ->deletable(false)
 ```
 
+## Allow submission of file from Clipboard paste
+
+You can accept submission via the clipboard paste by using `pasteable(true)`:
+
+```php
+use Filament\Forms\Components\FileUpload;
+
+FileUpload::make('attachment')
+    ->pasteable(true)
+```
+
 ## Prevent file information fetching
 
 While the form is loaded, it will automatically detect whether the files exist, what size they are, and what type of files they are. This is all done on the backend. When using remote storage with many files, this can be time-consuming. You can use the `fetchFileInformation(false)` method to disable this feature:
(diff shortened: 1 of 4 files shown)


bretto36:
@danharrin First time contributing so not sure if i'm supposed to run the npm run build or not. 

Thank you for the amazing package, i am thoroughly enjoying working with it. Such an upgrade over Nova...

wychoong:
https://github.com/filamentphp/filament/pull/16236#issuecomment-2868570237

danharrin:
Sorry, as mentioned above it would introduce other issues. Please let me know if you can fix those

bretto36:
@danharrin  Thanks Dan for the quick response. I can see there was a package update on Filepond - [Fix Paste on ContentEditabe Outside Filepond](https://github.com/pqina/filepond/blob/master/CHANGELOG.md#4314)

I've tried to update to that package and it still doesn't seem to work. I'll keep playing wtih Filepond. Hopefully v4 isn't in beta for too long, there are lots of cool features in there i want to use. Thanks for a great package!

bretto36:
@danharrin  apologies for pinging you so much - https://github.com/pqina/filepond/pull/1045

I have found the issue with the RichEditor - contenteditable attribute can be empty string or true

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/contenteditable

Value
The attribute must take one of the following values:

true or an empty string, which indicates that the element is editable.

I'll also see if the trix editor would be willing to change too to put true instead of blank string

danharrin:
In v4 we replace Trix with TipTap, can you test these changes in v4 please?

If they work in v4 we can merge it into that, but I wouldn't put in the effort to try and change Trix when we are abandoning it anyway

bretto36:
> In v4 we replace Trix with TipTap, can you test these changes in v4 please?
> 
> 
> 
> If they work in v4 we can merge it into that, but I wouldn't put in the effort to try and change Trix when we are abandoning it anyway

I'll do that. Thank you!

PR #16346, opened 2025-05-28
Update CanFormatState.php

Cegem-360:
## Description
In laravel Number::currency helper have optimal $precision parameter, in this change only add, that option default value = null waht is mean xx,00 in this change add option for xx EUR USD HUF etc..

## Update functionality
 Update functionality add exist but can't usable yet number formater helper
https://laravel.com/docs/12.x/helpers#method-number-currency
## Visual changes

https://laravel.com/docs/12.x/helpers#method-number-currency

That image decribe everything

## Functional changes

### DIFF
--- a/packages/infolists/docs/03-entries/02-text.md
+++ b/packages/infolists/docs/03-entries/02-text.md
@@ -113,6 +113,14 @@ use Filament\Infolists\Components\TextEntry;
 TextEntry::make('price')
     ->money('EUR')
 ```
+There is also a `precision` argument for `money()` that allows you to dispaly the original value in Integer format. This could be useful if you stores the price with integer, for example:
+
+```php
+use Filament\Infolists\Components\TextEntry;
+
+TextEntry::make('price')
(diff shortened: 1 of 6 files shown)


danharrin (changes requested):
Please update the other references to `Number::currency()`

Cegem-360:
> Please update the other references to `Number::currency()`

I added all other money() function as you requested and also i added to the docs

danharrin:
Sorry, just realised we do actually have this already in v4 with a different name: https://github.com/filamentphp/filament/blob/4.x/packages/tables/src/Columns/Concerns/CanFormatState.php#L212

So instead of needing to do the work backporting and fixing merge conflicts, I would suggest just using `->formatStateUsing()` with the Laravel helper in the function, and then in v4 you can go back to using the `money()` helper. Thanks!

Cegem-360:
Thank you for help, i don't know that implemented in 4.x., i very exited about new version. ( tailwind v4 support)

PR #16354, opened 2025-05-29
Introduce `uniqueForTenant()` to scope unique rule per tenant

thaqebon:
## Description

This PR introduces a new method `uniqueForTenant()` to form fields, allowing to easily apply tenant-scoped unique validation when using multitenancy features.

---

### DIFF
--- a/packages/forms/docs/23-validation.md
+++ b/packages/forms/docs/23-validation.md
@@ -517,6 +517,7 @@ In multi-tenant applications, you may want a field to be unique only within the 
 ```php
 Field::make('email')->uniqueForTenant()
 ```
+This method accepts the same parameters as `Field::unique()`, such as `table`, `column`, `ignoreRecord`, `ignorable`, and `modifyRuleUsing`.
 
 ### ULID
 
--- a/packages/forms/docs/23-validation.md
+++ b/packages/forms/docs/23-validation.md
@@ -510,6 +510,13 @@ Field::make('email')
     })
 ```
 
+### Unique for Tenant
+
+In multi-tenant applications, you may want a field to be unique only within the current tenant's scope.
+
+```php
+Field::make('email')->uniqueForTenant()
+```
 
 ### ULID
 
--- a/packages/forms/src/Components/Concerns/CanBeValidated.php
+++ b/packages/forms/src/Components/Concerns/CanBeValidated.php
@@ -3,6 +3,7 @@
 namespace Filament\Forms\Components\Concerns;
 
 use Closure;
+use Filament\Facades\Filament;
 use Filament\Forms\Components\Contracts\CanBeLengthConstrained;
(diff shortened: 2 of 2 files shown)


andrefelipe18:
@thaqebon 
I have internal packages that work for both multi-tenant and non-multi-tenant systems. I use a helper called `isMultTenant` to check whether the system where the package is running is multi-tenant. How could I use this `uniqueForTenant` method only in multi-tenant systems? Shouldn't it accept some parameter or condition to determine whether this validation should be applied or not?

thaqebon:
@andrefelipe18 
This PR still needs some additional testing and improvements, so it's not yet ready for use.
Under the hood, it wraps `Field::unique()` with a custom `$modifyRuleUsing` to apply tenant scoping.

**Usage Examples:**
```php
Field::make('email')->uniqueForTenant();
Field::make('email')->uniqueForTenant(table: User::class);
Field::make('email')->uniqueForTenant(column: 'email_address');
Field::make('email')->uniqueForTenant(ignoreRecord: false);
Field::make('email')->uniqueForTenant(ignorable: $userModel);
```
**Update**
You can also customize the rule further
```php
Field::make('email')->uniqueForTenant(
   modifyRuleUsing: fn (Unique $rule) => $rule->whereNull('deleted_at')
);
```

andrefelipe18:
I hope this is approved, it looks incredible

danharrin:
Hi, thanks for this. It won't be required in v4 since it adds automatic tenancy global scoping to all models, and I'm not going to introduce methods that will get immediately removed in the next version.

thaqebon:
@danharrin No problem! How can I enable tenancy global scoping so that it also applies to validation rules?

thaqebon:
@danharrin This was for v4. Please note that the tenancy global scope does not affect the unique validation rule.

danharrin:
@thaqebon please explain why the global scope doesn't affect this query?

thaqebon:
Because `Rule::unique()` does not use Eloquent, it interacts with the database directly, so global scopes are not applied. If you need global scopes to take effect, you may have to create a custom validation rule that uses the Eloquent query builder instead.

**Update**
Also `Rule::exists()` works the same way.

danharrin:
I think both should probably be custom rules then to query using Eloquent, considering we pass an Eloquent model as the table name

thaqebon:
Do you mind if I create the custom rules and place them in `packages/support/src/Rules`, and update the `Field::unique()` accordingly?

danharrin:
I'm reconsidering this, we probably shouldn't apply all global scopes since then it will also ignore soft deletes.

I'll put this on my todos to handle, thanks for bringing it to my attention

danharrin:
I have added `scopedUnique()` and `scopedExists()` methods to v4

Issue #16037, opened 2025-04-10
Plugin [filament-edit-profile] is not registered for panel [app]


cweagans:
### Package

filament/filament

### Package Version

v3.3.10

### Laravel Version

v12.8.1

### Livewire Version

v3.6.2

### PHP Version

8.3.17

### Problem description

If I create two panels and enable a plugin in one of them, I get `Plugin [filament-edit-profile] is not registered for panel [app]`. I specifically do not want to register the plugin in the `app` panel -- only in the `user` panel.

### Expected behavior

I should be able to enable a plugin in one panel and not the other.

### Steps to reproduce

1. Create two panels - `app` and `user`
2. Set the path on the `app` panel to be `/` and the path on the `user` panel to be `/user`
3. Enable a plugin in the user panel

This is all reliably reproduced with the repo below. The only other thing I changed is the panel color so I can differentiate between the two panels prior to enabling the plugin. That repo also includes a ddev setup in case you want to reproduce the _exact_ environment that I'm using locally.

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/cweagans/filament-plugin-bug

### Relevant log output

```shell

```

cweagans:
Prior art: https://www.answeroverflow.com/m/1192802009548595320

I could not find a related bug for @awcodes ' repo linked on that page.

ralphjsmit:
This issue happens when a plugin is registering a page slug dynamically based on something like a `$plugin->slug('my-custom-slug')` method. It's very convenient for users but it can be unhandy when having a plugin only registered in some panels. In my case, adopting the plugin to the following code fixed the issue:

![Image](https://github.com/user-attachments/assets/2c8f1d94-a026-42d6-a944-0837e64c41ec)

Though it would be cool if the `getSlug` method could get access to the current `$panel` that the route is being registered to, like eg by accepting a `$panel` parameter.

icaliman:
I have the same issue.

The problem comes from `filament/routes/web.php`.

When routes are registered, **current panel** is not set and all plugins works with **default plugin**.

A fix will be to temporarily set the current panel in the `filament/routes/web.php`:

```php
Route::name('filament.')
    ->group(function () {
        foreach (Filament::getPanels() as $panel) {
            /** @var Panel $panel */

            filament()->setCurrentPanel($panel);  // <---------- THIS WILL FIX THE ISSUE
            
            $panelId = $panel->getId();
            $hasTenancy = $panel->hasTenancy();
```

[...]

PR #16355, opened 2025-05-29
feat: provide `Panel $panel` to static route-related methods

ralphjsmit:
This PR adds the `Panel $panel` variable to all static route-related methods that are used in the route registration process. This will allow plugin authors to resolve issues like https://github.com/filamentphp/filament/issues/16037 by allowing customizing e.g. the `->slug()` of a page using the plugin configuration object in situations where the plugin is not registered in the default panel.

I tested the Rector by adding method overrides to some resources in the demo project and upgrading those in the demo project to V4.


danharrin (approved):

danharrin (approved):

danharrin (changes requested):
One thing I just noticed is that while you fixed method definitions, you didn't fix calling these methods in the upgrade script. Any calls to these methods, especially common ones like `getSlug()`, should probably have `Filament::getCurrentOrDefaultPanel()` as the first arg

danharrin:
The alternative, which I would actully prefer, is to make these new params nullable to be honest

If the user **requires** the `$panel`, they can call `Filament::getCurrentOrDefaultPanel()` themselves

But making it nullable means its so much easier for the upgrade, especially for people overriding the resource Blade views in v3 which all call `getSlug()`

ralphjsmit:
@danharrin For making the new params optional, would it be an idea to only make the `$panel` param nullable in the `getSlug()` method, but keep it non-nullable in the other methods? Otherwise we would introduce in inconsistency between for example `getRouteMiddleware()`, where it is required and `getRoutePath()`, where it is then not required. I also don't think people are ever gonna call things like the `getRoutePath()` or `getRoutePrefix()` or only in very rare cases (I can add to the upgrade guide).

CleanShot 2025-05-30 at 22 06 30@2x

danharrin (changes requested):
Sure, let's just make `getSlug()` nullable then and ensure the upgrade script reflects that.

For consistency, can you also refactor `getRouteName()` and `getRouteBaseName()` to accept a nullable `Panel` object instead of a nullable string please?

ralphjsmit:
@danharrin I made the following changes:
- Made panel parameter optional in `getSlug()`
- Added Rector to either prepend `$panel` (if variable defined in scope) or `Filament::getCurrentOrDefaultPanel()` to any method calls to the modified methods
- Replaced methods that accepted `?string $panel = null` with `?Panel $panel = null`. Added a Rector rule to handle wrap the first argument of any method calls into a `Filament::getPanel(<value of original first argument>)`.
- Tested rules on a demo project:

CleanShot 2025-06-06 at 09 47 26@2x

CleanShot 2025-06-06 at 09 47 22@2x

ralphjsmit:
- The only methods where there is still `?string $panel = null` is in 5 `getUrl()`/`getIndexUrl()`-related methods, but I'm not sure if it's great DX to replace that with an actual Panel-object. What do you think?

packages/upgrade/src/Rector/SimpleMethodChangesRector.php
```diff
@@ -47,6 +64,21 @@ public function getChanges(): array
                     'getSubNavigationPosition' => function (ClassMethod $node): void {
                         $node->flags &= Modifiers::STATIC;
                     },
+                    'getRoutePath' => $prependPanelParamModifier,
+                    'getRelativeRouteName' => $prependPanelParamModifier,
+                    'getSlug' => $prependPanelParamModifier,
```
danharrin: I think this one needs to prepend a nullable panel object param

packages/upgrade/src/Rector/SimpleMethodChangesRector.php
```diff
@@ -47,6 +64,21 @@ public function getChanges(): array
...
+                ],
+            ],
+            [
+                'class' => [
+                    Resource::class,
+                ],
+                'changes' => [
+                    'getRelativeRouteName' => $prependPanelParamModifier,
+                    'getRoutePrefix' => $prependPanelParamModifier,
+                    'getSlug' => $prependPanelParamModifier,
```
danharrin: and here

danharrin (changes requested):

packages/upgrade/src/Rector/SimpleMethodChangesRector.php
ralphjsmit: Sharp, just done!

danharrin (approved):

danharrin:
Thanks Ralph!

PR #16360, opened 2025-05-29
feat(i18n): add merge tags translations for pt and es

andrefelipe18:
## Description

Add translations for merge tags in Portuguese and Spanish

## Functional changes

### DIFF
--- a/packages/forms/resources/lang/pt/components.php
+++ b/packages/forms/resources/lang/pt/components.php
@@ -382,6 +382,8 @@ return [
 
         ],
 
+        'no_merge_tag_search_results_message' => 'Nenhuma tag encontrada.',
+
         'tools' => [
             'attach_files' => 'Anexar ficheiros',
             'blockquote' => 'Bloco de citação',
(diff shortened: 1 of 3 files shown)


packages/forms/resources/lang/pt/components.php
```diff
@@ -382,6 +382,8 @@
 
         ],
 
+        'no_merge_tag_search_results_message' => 'Nenhuma tag encontrada.',
```
saade:
```suggestion
        'no_merge_tag_search_results_message' => 'Nenhuma tag dinâmica encontrada.',
```

packages/forms/resources/lang/pt_BR/components.php
```diff
@@ -382,6 +382,8 @@
 
         ],
 
+        'no_merge_tag_search_results_message' => 'Nenhuma tag encontrada.',
```
saade:
```suggestion
        'no_merge_tag_search_results_message' => 'Nenhuma tag dinâmica encontrada.',
```

saade (changes requested):

saade (approved):

andrefelipe18:
@saade I think `Nenhuma tag encontrada` is more natural

saade:
> @saade I think `Nenhuma tag encontrada` is more natural

So why are they called `Tags dinâmicas`? Shouldn't they be called `Tags` only?
— Since `dinâmica` is a characteristic of a `Tag`, I think it should be included, too.

@leandrocfe wdyt?

leandrocfe (approved):
I'm fine with just calling it "tag", but I think "dinâmica" makes it clearer that these aren't normal tags. They are special, dynamic ones.

PR #16367, opened 2025-05-30
Adding limit when searchable is true for Select Form Field

### LINKED ISSUE #16363: Select Field Relationship have no limit
### Package

filament/filament

### Package Version

v3.x

### Laravel Version

v12.x

### Livewire Version

v3.x

### PHP Version

8.2

### Problem description

I have over 1M record for projects table. when I use the code below it will get all the projects without any limit for it which making the website not responding.

```
Forms\Components\Select::make('project_id')
    ->searchable()
    ->relationship('project', 'title')
```

In order to make the select works I need to explicitly use `limit` in `options` and `getSearchResultsUsing`
```
Forms\Components\Select::make('project_id')
    ->searchable()
    ->options(Project::limit(10)->get()->pluck('title', 'id'))
    ->getSearchResultsUsing(function (?string $search) {
        return Project::where('title', 'like', "%{$search}%")->limit(10)->get()->pluck('title', 'id');
    })
```

is there any better way to do it? and why there is no limit in `relationship` method?

### Expected behavior

It should be some functionality to limit the amount of data that is retrieved from the model like using `limit` or something which make it easier to use the field.

### Steps to reproduce

- create a table with over 1M record (lets called it projects)
- use that table as a `belongsTo` relationship in order to use Select Field
- copy this code
```
Forms\Components\Select::make('project_id')
    ->searchable()
    ->relationship('project', 'title')
```

https://github.com/ht3aa/ticket-system-filament/blob/bug/app/Filament/Member/Resources/TicketResource.php

### Relevant log output

```shell

```

ht3aa:
## Description

This PR add the applity to apply limit on the `$this->options` when `->relationship with searchable` is applied on Select form field

## Visual changes

No visual changes

## Functional changes

### DIFF
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -778,6 +778,14 @@ class Select extends Field implements Contracts\CanDisableOptions, Contracts\Has
                 ]) ?? $relationshipQuery;
             }
 
+            $baseRelationshipQuery = $relationshipQuery->getQuery();
+
+            if (isset($baseRelationshipQuery->limit)) {
+                $component->optionsLimit($baseRelationshipQuery->limit);
+            } elseif ($component->isSearchable()) {
+                $relationshipQuery->limit($component->getOptionsLimit());
+            }
+
             $qualifiedRelatedKeyName = $component->getQualifiedRelatedKeyNameForRelationship($relationship);
 
             if ($component->hasOptionLabelFromRecordUsingCallback()) {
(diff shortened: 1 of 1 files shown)


ht3aa:
This PR solve the issue in [16363](https://github.com/filamentphp/filament/issues/16363)

danharrin (approved):

archilex:
@danharrin Looks like this changed introduces a bug and searching relationships no longer returns the correct results. It appears to be only searching within the original limited list.

danharrin:
@archilex does v3.3.19 fix the issue for you?

archilex:
Update: 3.3.19 actually broke the select for us, but Dan has already released 3.3.20 that fixes this and the initial issue.

PR #16380, opened 2025-06-01
Feature reorderable columns

### LINKED ISSUE #15486: Reorderable table columns

### LINKED ISSUE #15887: Grouped columns in toggleable dropdown
Currently when columns are grouped and toggleable, the grouped headings aren't present in the "toggleable" dropdown, so it can be confusing if many columns have similar labels but different groups.

Can we add the group headings in the toggleable dropdown to "group" the checkboxes together?

archilex:
## Description

This PR introduces reorderable columns to v4. Columns that are grouped in a column group are also supported. All toggling and reordering is now handled via alpine and is deferred by default (similiar to filters).

Also introduces a new indeterminate state for checkboxes allowing column groups headers to correctly display their indeterminate state.

Pending documentation

Closes #15486 

## Visual changes

Screenshot 2025-05-31 at 11 52 10 PM (2)

## Functional changes


packages/tables/src/Concerns/CanToggleColumns.php
```diff
@@ -110,4 +141,49 @@ protected function getTableColumnToggleFormMaxHeight(): ?string
...
+     *     name: string,
+     *     label: string,
+     *     toggled: bool,
+     *     toggleable: bool,
+     *     columns: array<int, mixed>
+     * }
+     */
+    protected function mapColumnGroup(ColumnGroup $group): array
+    {
+        return [
```
archilex, line 157: This should probably be a DTO, but I didn't see any in the code base so didn't want to introduce that just yet...but can if preferred

danharrin (changes requested):
Instead of "toggle-columns" "col-toggle", can this be the "column manager" please?

And yeah, I think a DTO would probably be great for this if you want to introduce one :)

archilex:
I've updated the naming of the css and component references I introduced, but can keep going throughtout the code base (methods, properties, etc.) if you would like. 

Improved the type hinting for added safety.

I also looked into implementing a wireable object and it does seem possible...just not sure how beneficial it will be. Either way, wanted to wait to get your thoughts on renaming the existing code base first.

danharrin:
Yes please, it would be nice if everything was consistent. Even in the translations, `column_toggle` could be `column_manager`

Please let me know if any breaking changes are introduced if you rename something. Its probably worth adding to the Rector upgrade script if that happens

archilex:
I've updated all the naming and deprecated older methods. With the deprecated methods I don't believe there are breaking changes. The demo app I've been testing against works without issue. The `CanToggleColumn` classes did get renamed, but those are internal so I don't expect many people would have overridden those.

As I was renaming I realized I forgot to add support for multiple columns in the dropdown. Since a group's columns have to stay together, using grid won't work as the height differences between groups and single columns causes large gaps. I already have a working fix for this using css columns, but I need some more time to implement, and I didn't want to hold this up for that.

archilex:
Added support for multiple columns in the form manager:

Screenshot 2025-06-07 at 10 07 02 PM

bin/build.js
```diff
@@ -155,6 +155,14 @@ tableColumns.forEach((column) => {
     })
 })
 
+compile({
+    ...defaultOptions,
+    entryPoints: [
+        `./packages/tables/resources/js/components/column-manager.js`,
+    ],
+    outfile: `./packages/tables/dist/components/column-manager.js`,
+})
```
danharrin: I think this probably doesn't need to be asynchronously loaded as a separate file, and can be included in the main `tables/index.js` bundle since the JS size is pretty small and it saves a network request.

docs-assets/screenshots/schema.js
```diff
@@ -2800,15 +2800,15 @@ export default {
         },
     },
     'tables/columns/toggleable': {
```
danharrin: Maybe we should change the name of this screenshot too, so its `tables/column-manager`?

packages/tables/src/View/TablesRenderHook.php
```diff
@@ -32,7 +32,17 @@ class TablesRenderHook
...
 
-    const TOOLBAR_TOGGLE_COLUMN_TRIGGER_BEFORE = 'tables::toolbar.toggle-column-trigger.before';
+    const TOOLBAR_COLUMN_MANAGER_TRIGGER_AFTER = 'tables::toolbar.toggle-column-trigger.after';
+
+    /**
+     * @deprecated Use `TOOLBAR_COLUMN_MANAGER_TRIGGER_BEFORE` instead.
+     */
+    const TOOLBAR_TOGGLE_COLUMN_TRIGGER_BEFORE = self::TOOLBAR_COLUMN_MANAGER_TRIGGER_BEFORE;
+
+    const TOOLBAR_COLUMN_MANAGER_TRIGGER_BEFORE = 'tables::toolbar.toggle-column-trigger.before';
```
danharrin: Small nit: can you please put these in alphabetical order?

packages/tables/src/Concerns/InteractsWithTable.php
```diff
@@ -45,23 +45,16 @@ public function bootedInteractsWithTable(): void
     {
         $this->table = $this->table($this->makeTable());
 
-        $this->cacheSchema('toggleTableColumnForm', $this->getTableColumnToggleForm());
-
         $this->cacheSchema('tableFiltersForm', $this->getTableFiltersForm());
 
         $this->cacheMountedActions($this->mountedActions);
 
+        $this->initializeColumnManager();
```
danharrin: Does this need to happen on every request, or just the first request when the component is mounted?

packages/tables/resources/views/components/column-manager.blade.php
danharrin, line 1: When there are only column groups and no columns outside groups, the column manager shows but is empty.

If you check the `4.x` branch of the Filament demo, you can observe this in the `PaymentsRelationManager` of the order resource, if you make a column toggleable.

packages/tables/src/Table/Concerns/HasColumnManager.php
```diff
@@ -0,0 +1,223 @@
...
+        foreach ($this->getColumns() as $column) {
+            if (! $column->isToggleable()) {
+                continue;
+            }
+
+            return true;
+        }
+
+        return false;
+    }
```
danharrin, line 256: Currently, the manager is enabled if any columns are toggleable. IMO, you should be able to enable the manager even if you just want reorderable columns. You should also be able to disable the reordering if you just want columns to be toggleable.

**To do this:**

I think there should be a `$table->columnManager(bool | Closure | null $condition = true)` method to enable the column manager without needing to make any of the columns toggleable. There should be a `bool | Closure | null $hasColumnManager = null` prop too (note `null` default, not `false`).

There should also be a `reorderableColumns(bool | Closure $condition = true)` method and `bool | Closure $hasReorderableColumns = true` property. This is enabled by default, but columns can only be reordered if the column manager is enabled.

The `hasColumnManager()` method should return the value of the prop if its a `bool`. If its null, it should check if the `$hasReorderableColumns` prop is true. If its false, it should then check if any columns are toggleable. Note: should we check if a column is hidden before checking if its toggleable?

If no columns are toggleable but the manager is enabled, the checkboxes should not be present in the dropdown. This allows them to still be reordered. If the column manager is enabled but reordering is disabled, we shouldn't show the reordering handles.

I know this is a lot of words for probably not a lot of code, so please let me know if I didn't explain anything clear enough.

packages/tables/resources/views/components/column-manager.blade.php
```diff
@@ -0,0 +1,152 @@
...
+<div class="fi-ta-col-manager">
+    <div
+        @if (FilamentView::hasSpaMode())
+            {{-- format-ignore-start --}}x-load="visible || event (x-modal-opened)"{{-- format-ignore-end --}}
+        @else
+            x-load
+        @endif
+        x-load-src="{{ \Filament\Support\Facades\FilamentAsset::getAlpineComponentSrc('column-manager', 'filament/tables') }}"
+        x-data="columnManagerComponent({
+                    columns: $wire.entangle('columnManager'),
```
danharrin: I understand the need to have data like `isToggleable` and `label` in the JS state of the component so its accessible here. However, if the user manages to change the label of a column using JS, or makes a column toggleable that shouldn't be, is this somewhat of a bug? Maybe not a security vulnerability, but should we be concerned? No is a perfectly valid answer to this, I'm just checking :)

danharrin (changes requested):

packages/tables/src/Concerns/InteractsWithTable.php
archilex: Every request unfortunately because if not sorting, searching, etc. will reset any reordering that has been done.

packages/tables/resources/views/components/column-manager.blade.php
archilex: I believe this is a session issue and you first loaded the RM, then added `->toggleable()` to a column, then refreshed. The issue is that when you first load the table it sets the session with the state of the columns which in this case the column wasn't toggleable. But when you make a column toggleable in the code, that doesn't affect the session. `hasToggleableColumns` evaluates the filament columns on each request which is why the dropdown shows, but `x-show="hasToggleable"` evaluates the data array which comes from the session.

You can also see this happen if you have multiple toggleable columns and then add/remove toggleable(). The manager won't get updated until you logout/clear the session.

This is also something that comes up from time to time with AdvancedTables as devs are iterating. The solution is to just clear the session/logout...Not sure the best solution here beyond devs being aware of this. Thoughts?

bin/build.js
archilex: done

packages/tables/src/View/TablesRenderHook.php
archilex: done

packages/tables/src/Table/Concerns/HasColumnManager.php
archilex: done

docs-assets/screenshots/schema.js
archilex: done

packages/tables/resources/views/components/column-manager.blade.php
archilex: Since the data is stored in the session, if they were able to tamper with that data, then yes, they could change the label/toggleable state. One solution to prevent this could be on `updatedColumnManager` to loop through the array again and then just set the label/toggleable to what was defined in `columns()`. That way any tampering would never persist. Thoughts?

packages/tables/resources/views/components/column-manager.blade.php
danharrin: Maybe calculate the "default" state of the JSON when it is first loaded, hash it, and then store that hash in the Livewire version of the JSON? If the hash is different, it's time to invalidate the Livewire version?

packages/tables/resources/views/components/column-manager.blade.php
danharrin: Yeah I think that sounds good! Might also tie to the conversation above about JSON invalidation? If its different than what we expect, reset the JSON entirely?

archilex:
Dan, implemented hashing for both initializing the columnManager (checking if the table column configuration has changed at all), and when applying any toggling/reordering changes (to see if the payload has been tampered with). Tried to be as efficient as possible, but ultimately each hash has to check the name/label/toggleable state of each column and group. 

I also discovered that livewire's `updatedColumnManager()` was being called for every updated item in the array. So I switched to a separate method that we have more control of and is only called once.

archilex:
@danharrin This latest version swaps out hashing in favor of syncing the column manager from the default state to pick up any changes that might have happened to the table. This simplifies the code and removes the need to store hashes in the session.

And as I shared (and in case others wonder), I also explored wireable objects (don't work) as well as implementing a DTO. A DTO works and cleans up the code base, however the DTOs add considerable overhead. Plain old arrays are just fast.

PR #16390, opened 2025-06-03
fix/key-value component

alex-elivate:
Component row syncing fix

## Description

https://github.com/filamentphp/filament/issues/12824 
state would not update the first time.
https://github.com/filamentphp/filament/issues/1107
duplicate keys were lost on refresh

## Visual changes

Before
https://github.com/user-attachments/assets/53670c9a-3299-4f24-bad2-71386c08fa5b

After
https://github.com/user-attachments/assets/f8380f8c-3d92-4d45-98a9-fed6ffbb959d

## Functional changes

Updated updateState to sync only on actual change
Updated updateRows to track and preserve empty keys

Thanks for making me work on this, I learned a lot! This was simplified from my last approach: https://github.com/filamentphp/filament/pull/16290


packages/forms/resources/js/components/key-value.js
```diff
@@ -71,14 +69,22 @@ export default function keyValueFormComponent({ state }) {
             })
         },
 
+        // https://github.com/filamentphp/filament/issues/1107
+        // https://github.com/filamentphp/filament/issues/12824
         updateRows: function () {
-            if (!this.shouldUpdateRows) {
-                this.shouldUpdateRows = true
+            const state = Alpine.raw(this.state)
+            const mergedRows = state.map(({ key, value }) => ({ key, value }))
```
danharrin, line 76: Can you please explain the `.map(({ key, value }) => ({ key, value }))` part? From what I can tell, you are mapping the array into the same format it was in originally?

danharrin (changes requested):

packages/forms/resources/js/components/key-value.js
alex-elivate: The deep copy ensures that changes in the UI don't affect the original state. I'm still learning about UI state management, but what keeps coming up is to ensure you have immutable objects or you'll have side effects. Functional methods are not my goto because it's easier to wrap my head around a good old fashioned foreach loop, but this one makes the code prettier.

danharrin: Ah so you are trying to loose the Alpine proxy reference? Are you sure this is required after you do the `Alpine.raw()`? Maybe `JSON.parse(JSON.stringify(state))` would work if not?

alex-elivate: I used Alpine.raw() because I saw it was added in v4 in other parts of this component and wanted to be consistent, but I didn't fully understand what it did under the hood. I'm reading up on it, but from what you're saying, it sounds like the object copying with map() is redundant. What's your take? I'm down for testing, I always have time to learn.

danharrin: I think it looks redundant, but it would be great if you could test removing the `.map()` to ensure that is the case please?

alex-elivate: I tested without the map. Alpine is not detecting a state change.  When I save the form, new fields are displaying, but not saving.

What I’ve learned:
Alpine.raw() removes the proxy, but copy is still needed or $this->state and $this->rows are the same object, breaking the reactivity needed to save the form. The form updates visually perfectly, but does not trigger a save/update, therefore new rows are randomly not persisted. (I say randomly because I was able to get it to save on some edge cases)

A lot of embarrassment demonstrating how little I know, but it’s worth the lesson. Thanks yet again.

danharrin (approved):

packages/forms/resources/js/components/key-value.js
danharrin: Thank you very much for the investigation and fix, it's very valuable!

PR #16402, opened 2025-06-04
Add option to get file url

aldesrahim:
## Description

I am facing an issue when using public S3 object, the handler to get the public url is not right, in my case it should be using `publicUrl` instead of `url`. So, I implemented a custom closure to override the url handler.

But, I think the method I am using to solve the problem is not suitable for most of the people that using FileUpload, and within this PR I am attaching an optional approach to handle the FileUpload's url generation. See https://github.com/aldesrahim/filament-file-upload/commit/b6c662bde2c937f53f2e6a9efe20e5cd9d6f8b76

## Functional changes


danharrin:
You can overwrite `getUploadedFileUsing()` to do this instead of introducing a new function

PR #16403, opened 2025-06-04
Add S3 public url handler

aldesrahim:
## Description

Handle public S3 when generating `ImageColumn` url

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/tables/src/Columns/ImageColumn.php
+++ b/packages/tables/src/Columns/ImageColumn.php
@@ -4,6 +4,7 @@ namespace Filament\Tables\Columns;
 
 use Closure;
 use Illuminate\Contracts\Filesystem\Filesystem;
+use Illuminate\Filesystem\AwsS3V3Adapter;
 use Illuminate\Filesystem\FilesystemAdapter;
 use Illuminate\Support\Facades\Storage;
 use Illuminate\View\ComponentAttributeBag;
@@ -163,9 +164,11 @@ class ImageColumn extends Column
             }
         }
 
+        $url = null;
+
         if ($this->getVisibility() === 'private') {
             try {
-                return $storage->temporaryUrl(
+                $url = $storage->temporaryUrl(
                     $state,
                     now()->addMinutes(5),
                 );
@@ -174,7 +177,15 @@ class ImageColumn extends Column
             }
         }
 
-        return $storage->url($state);
+        if (! $url && $storage instanceof AwsS3V3Adapter) {
+            try {
+                $url = $storage->publicUrl($state);
+            } catch (Throwable $exception) {
+                // This driver does not support creating public URLs.
+            }
+        }
+
+        return $url ?? $storage->url($state);
     }
 
     public function getDefaultImageUrl(): ?string


danharrin:
In #16402 you suggested that this method is not necessarily suitable for all users, why is that? And how does it differ considering you are suggesting that we use it for the `ImageColumn`?

This is probably my lack of knowledge, but I want to understand before I merge changes

aldesrahim:
> In #16402 you suggested that this method is not necessarily suitable for all users, why is that? And how does it differ considering you are suggesting that we use it for the `ImageColumn`?
> 
> This is probably my lack of knowledge, but I want to understand before I merge changes

Thanks for the response.

The "method is not necessarily suitable for all users" is referring to the PR I submitted (#16402), while this PR is the method I said will likely be used. Since in the other PR, I made a custom handler/closure just to override URL generation (add unnecessary complexity).

But in this PR, my focus is to handle the `publicUrl` method call rather than create a whole custom handler/method to override it.

wychoong:
can you share more on how are you using S3?
- is the s3 bucket publicly accessible
- are you using it with cloudfront
- is s3 only accessible thru cloudfront

danharrin:
Where is the `publicUrl()` method defined in `AwsS3V3Adapter`? Is it within your own code?

aldesrahim:
Sorry for late response. 

> Where is the `publicUrl()` method defined in `AwsS3V3Adapter`? Is it within your own code?

The `publicUrl` is defined within [The PHP League's AwsS3V3Adapter](https://github.com/thephpleague/flysystem/blob/3b8d678324d7651955483d7f416af8b97f121196/src/AwsS3V3/AwsS3V3Adapter.php#L476), which is the class that laravel's [AwsS3V3Adapter](https://github.com/laravel/framework/blob/c7a87ae20172593f7dd498d9b7779b56e051fa82/src/Illuminate/Filesystem/AwsS3V3Adapter.php#L10C1-L11C1) extends as `FileSystemAdapter`. I am not implemented the method by myself. (I am using Laravel  10, but in Laravel 12, it still extends the same class, so I think there are no differences between the Laravel versions)

> can you share more on how are you using S3?
> 
> * is the s3 bucket publicly accessible
> * are you using it with cloudfront
> * is s3 only accessible thru cloudfront

I am using minio (self-hosted S3 object), that one of my bucket's visibility is set to public, it doesn't require signed URL (generated by `temporaryUrl`)

This is the screenshot how my S3 bucket configured. The `Acess Policy` is set to `public`.
image

I am hoping if this PR is approved, it can be implemented too in `FileUpload` form's components.

danharrin:
I do not really understand. Your bucket is public, I guess your `ImageColumn` has `visibility('public')` or is using the default. Why does `url()` generate a temporary URL instead of a public one? I thought that the default behaviour of `url()` was to generate a public URL?

aldesrahim:
Ah, I see, this comes down to me misconfigured `AWS_URL` inside `.env` file, currently the `AWS_URL` was filled (w/o the bucket name) and the `Laravel's AwsS3V3Adapter` will concat with `prefixPath`, but if the `AWS_URL` is not set, it will call `getObjectUrl` same as `League's AwsS3V3Adapter` did.
image

Thank you for your patience and helping me through this. I apologize if this wastes your time.

PR #16407, opened 2025-06-05
CodeEditor: Add `->language()`, improve dark mode, add `destroy()` for cleanup

binaryfire:
CodeMirror only handles a single language at a time, so the current implementation wasn't working properly. This PR adds a `->language()` method which allows the user to select the language they want to use. By default it's set to plain text (none). 

**_Examples_**

Default (plain text):
```PHP
CodeEditor::make('code')
```
image

PHP:
```PHP
CodeEditor::make('code')
    ->language(Language::PHP),
```
image

JSON:
```PHP
CodeEditor::make('code')
    ->language(Language::JSON),
```
image


packages/forms/resources/js/components/code-editor.js
```diff
@@ -8,12 +8,32 @@ import { php } from '@codemirror/lang-php'
...
+    isDisabled,
+    language,
+    state,
+}) {
     return {
         editor: null,
-
         state,
 
+        getLanguageExtension() {
```
andrefelipe18: I think this makes it easier to add new languages

```javascript
getLanguageExtension() {
    const extensions = {
        css: css,
        html: html,
        javascript: javascript,
        json: json,
        php: php
    };

    return (extensions[language]?.()) || [];
},
```

binaryfire: @andrefelipe18 Good call. Updated.

andrefelipe18: I'm happy to contribute, let's hope PR passes

ralphjsmit: Is there an extension possible for Blade?

packages/forms/src/Components/CodeEditor/Enums/Language.php
```diff
@@ -0,0 +1,18 @@
...
+
+    case HTML = 'html';
+
+    case Javascript = 'javascript';
+
+    case JSON = 'json';
+
+    case PHP = 'php';
+
+    case PlainText = 'plaintext';
```
ralphjsmit: If it is "PlainText", should the key be `plain_text`?

```suggestion
    case PlainText = 'plain_text';
```

ralphjsmit (commented):
Nice addition! I missed this already when testing out the `CodeEditor` :)

packages/forms/src/Components/CodeEditor/Enums/Language.php
binaryfire: @ralphjsmit Yeah it should have been. Updated.

packages/forms/resources/js/components/code-editor.js
binaryfire: @ralphjsmit I looked into it and it didn't seem straightforward since languages need to be added at build time. CodeMirror doesn't have a plugin system or registry.

I'm maintaining my own fork which adds shell (via `codemirror/legacy-modes` + `StreamLanguage`) and Python. It increases bundle size though so not sure if Dan wants to open that door. The risk is people start requesting additional languages and the bundle size blows out.

There are 25 independent languages: https://github.com/orgs/codemirror/repositories?q=lang.
Plus all the legacy modes: https://github.com/codemirror/legacy-modes?tab=readme-ov-file#api-reference

I could add a few more core languages though. Shell, Python, SQL and YAML seem like good ones to me. That wouldn't add too much weight and would cover most use cases.

Someone with more JS experience might be able to figure out a way to dynamically add languages but that should be a separate PR. I'd like to just get the fundamentals working with this one.

ralphjsmit: Okay, yes I suppose that's a good consideration. Blade I don't see indeed as a plugin, probably too niche. SQL you could consider, then you capture pretty much all technologies used around Laravel.

binaryfire: @ralphjsmit Ah sorry, I misunderstood your question! I thought you were asking about dynamically adding more CodeMirror languages via PHP.

For Blade highlighting you’d need to write a custom language: https://codemirror.net/examples/lang-package/. Definitely doable but it’d take some work and would need to be maintained. So not something that should go in core.

ralphjsmit: Ah yeah okay, yeah just Blade-language support :) But if it requires a new plugin then that's not really an option for Filament.

binaryfire:
@danharrin I just added the OneDark theme (https://github.com/codemirror/theme-one-dark) to improve dark mode. OneDark is a first-party CodeMirror theme (it's their only theme) so there won't be any maintenance headaches. The default dark mode is _really_ bad.

The bundle size increase is negligible. 629kb without the theme, 632kb with it.

Before (no dark theme):
image

After (with OneDark):
image

I've also added a destroy method to ensure proper component cleanup.

And updated the CSS to use `overflow-hidden` to fix rounded corners because that fixes the scrollbar's rounded corners too:

image

packages/forms/resources/css/components/code-editor.css
```diff
@@ -1,11 +1,13 @@
...
+    @apply overflow-hidden;
+
     & .cm-editor {
         &.cm-focused {
             @apply outline-none!;
         }
 
         & .cm-gutters {
-            @apply min-h-48! rounded-s-lg border-e-gray-300! bg-gray-100! dark:border-e-gray-800! dark:bg-gray-950!;
+            @apply min-h-48! border-e-gray-300! bg-gray-100! dark:border-e-gray-800! dark:bg-gray-950!;
```
danharrin, line 10: Can you please explain why the `rounded-s-lg` was removed?

packages/forms/docs/20-code-editor.md
```diff
@@ -14,3 +14,15 @@ CodeEditor::make('code')
...
+## Changing the code editor's language
+
+You may change the language of the code editor using the `language()` method. The editor supports CSS, HTML, JavaScript, JSON, PHP and plain text. You can open the `Filament\Forms\Components\CodeEditor\Enums\Language` enum class to see the full list. To switch to use JavaScript as the language, you can use the `Language::Javascript` enum value:
+
+```php
+use Filament\Forms\Components\CodeEditor;
+use Filament\Forms\Components\CodeEditor\Enums\Language;
+
+CodeEditor::make('code')
+    ->language(Language::Javascript)
```
danharrin:
```suggestion
You may change the language of the code editor using the `language()` method. The editor supports CSS, HTML, JavaScript, JSON, PHP and plain text. You can open the `Filament\Forms\Components\CodeEditor\Enums\Language` enum class to see the full list. To switch to using JavaScript as the language, you can use the `Language::JavaScript` enum value:

```php
use Filament\Forms\Components\CodeEditor;
use Filament\Forms\Components\CodeEditor\Enums\Language;

CodeEditor::make('code')
    ->language(Language::JavaScript)
```

packages/forms/resources/js/components/code-editor.js
```diff
@@ -5,15 +5,40 @@ import { html } from '@codemirror/lang-html'
...
+    language,
+    state,
+}) {
     return {
         editor: null,
-
+        themeCompartment: new Compartment(),
         state,
 
+        isDarkMode() {
```
danharrin:
```suggestion
        isDarkMode: function () {
```

packages/forms/resources/js/components/code-editor.js
```diff
@@ -5,15 +5,40 @@ import { html } from '@codemirror/lang-html'
...
+            return document.documentElement.classList.contains('dark')
+        },
+
+        getThemeExtensions() {
```
danharrin:
```suggestion
        getThemeExtensions: function () {
```

packages/forms/resources/js/components/code-editor.js
```diff
@@ -5,15 +5,40 @@ import { html } from '@codemirror/lang-html'
...
+            return this.isDarkMode() ? [oneDark] : []
+        },
+
+        getLanguageExtension() {
```
danharrin:
```suggestion
        getLanguageExtension: function () {
```

packages/forms/resources/js/components/code-editor.js
```diff
@@ -5,15 +5,40 @@ import { html } from '@codemirror/lang-html'
...
+            const extensions = {
+                css: css,
+                html: html,
+                javascript: javascript,
+                json: json,
+                php: php,
```
danharrin:
```suggestion
                css,
                html,
                javascript,
                json,
                php,
```

packages/forms/src/Components/CodeEditor.php
```diff
@@ -12,4 +13,18 @@ class CodeEditor extends Field
...
+    {
+        $this->language = $language;
+
+        return $this;
+    }
+
+    public function getLanguage(): Language
+    {
+        return $this->language;
+    }
```
danharrin: Please allow closure functions to be passed and evaluated

Ensure that the `<UtilityInjection>` component is used in the docs to reflect this

packages/forms/src/Components/CodeEditor/Enums/Language.php
```diff
@@ -0,0 +1,18 @@
...
+{
+    case CSS = 'css';
+
+    case HTML = 'html';
+
+    case Javascript = 'javascript';
+
+    case JSON = 'json';
+
+    case PHP = 'php';
```
danharrin:
```suggestion
    case Css = 'css';

    case Html = 'html';

    case JavaScript = 'javascript';

    case Json = 'json';

    case Php = 'php';
```

danharrin (changes requested):

packages/forms/resources/css/components/code-editor.css
binaryfire: @danharrin Because adding `overflow-hidden` rounds the corners:

image

image

And I thought the bottom staying square looked better when the scrollbar appears:

image

(vs this):

image

Up to you though - I can re-add it if you prefer.

binaryfire:
@danharrin I've made all the changes except the CSS: https://github.com/filamentphp/filament/pull/16407#discussion_r2133638695

packages/forms/docs/20-code-editor.md
```diff
@@ -14,3 +14,17 @@ CodeEditor::make('code')
...
+
+```php
+use Filament\Forms\Components\CodeEditor;
+use Filament\Forms\Components\CodeEditor\Enums\Language;
+
+CodeEditor::make('code')
+    ->language(Language::JavaScript)
+```
+
+<UtilityInjection set="formFields" version="4.x">As well as allowing a static value, the language() method also accepts a function to dynamically calculate it. You can inject various utilities into the function as parameters.</UtilityInjection>
```
danharrin: This component needs to be imported at the top

packages/forms/src/Components/CodeEditor.php
```diff
@@ -12,4 +14,18 @@ class CodeEditor extends Field
      * @var view-string
      */
     protected string $view = 'filament-forms::components.code-editor';
+
+    protected Language | Closure $language = Language::PlainText;
```
danharrin: Instead of needing a `PlainText` case, I think that should just be represented by `null` language.

packages/forms/resources/js/components/code-editor.js
```diff
@@ -5,15 +5,40 @@ import { html } from '@codemirror/lang-html'
...
+        getLanguageExtension: function () {
+            const extensions = {
+                css,
+                html,
+                javascript,
+                json,
+                php,
+            }
+
+            return extensions[language]?.() || []
```
danharrin: Instead of returning an extension or an array, can we return an extension or null, and then when using the extension only register it if its not null?

packages/forms/docs/20-code-editor.md
```diff
@@ -14,3 +14,17 @@ CodeEditor::make('code')
 ```
 
 <AutoScreenshot name="forms/fields/code-editor/simple" alt="Code editor" version="4.x" />
```
danharrin, line 17: Now that there isn't syntax highlighting out of the box, the code used to generate this screenshot in `FieldsDemo.php` needs to be updated to ensure it has the correct language

packages/forms/docs/20-code-editor.md
danharrin, line 1: The "Introduction" section shouldn't mention the available languages anymore, as this implies they can be used without additional config, especially since the first code example doesn't use them.

danharrin (changes requested):

packages/forms/resources/js/components/code-editor.js
binaryfire: I think the current implementation is cleaner and more performant:

Current approach (returning []):
```JS
extensions: [
    basicSetup,
    keymap.of([indentWithTab]),
    // ... other extensions
    this.getLanguageExtension(), // Clean - returns [] when no lang set
    this.themeCompartment.of(this.getThemeExtensions()),
]
```

And we're only calling `getLanguageExtension()` once

Proposed approach (returning null) would require:
```JS
// Option 1: Calls getLanguageExtension() twice
...(this.getLanguageExtension() ? [this.getLanguageExtension()] : []),

// Option 2: More verbose
const langExt = this.getLanguageExtension();
if (langExt) extensions.push(langExt);
```

The language extension is already only being registered if set. If not set, `getLanguageExtension()` returns an empty array (i.e. no extension).

binaryfire:
@danharrin Done. 

- Removed `PlainText` case, use `null` by default instead
- Updated docs, added second screenshot to demonstrate syntax highlighting.
- The current code already only registers extensions when they exist (since `getLanguageExtension()` returns an empty array when no lang is set): https://github.com/filamentphp/filament/pull/16407#discussion_r2134384798

packages/forms/resources/js/components/code-editor.js
danharrin: I just don't like the inconsistency where an empty array represents no extension instead of `null`. It feels like we are just relying on the fact that CodeMirror appears to ignore empty arrays in the extension list by chance, which feels fragile to assume.

A constant can be used to store the resolved extension object so the method only needs to be run once

danharrin (approved):

PR #16419, opened 2025-06-06
Hidden fields in Section incorrectly dehydrated in Section with state path

alanost:
This PR addresses a bug where fields inside a hidden Section component (with a state path) are incorrectly included in the dehydrated state. Currently, when a field is:

- Placed inside a Section inside with a state path inside a section
- The Section is hidden based on a condition
- The field is marked as dehydrated only when visible

The field's value is still being dehydrated even though the Section (and thus the field) is hidden, and the dehydration condition should prevent this.

**Example scenario:**
```php
Section::make('User Information')
    ->statePath('data')
    ->schema([
        Radio::make('show_input')
            ->label('Control Input Visibility')
            ->options([
                'show' => 'Show',
                'hide' => 'Hide',
            ])
            ->live(),
        Section::make('Name')
            ->schema([
                TextInput::make('name')
                    ->label('Conditional Name'),
            ])
            ->hidden(fn (Get $get): bool => $get('show_input') === 'hide' || $get('show_input') === null)
            ->dehydrated(fn (Get $get): bool => $get('show_input') === 'show'),
    ])
```
When `show_input` is `'hide'`, the `'name'` field is still present in the dehydrated state, even though the Section is hidden and the dehydration condition should prevent this.

**What this PR does:**
- Adds a test case demonstrating the bug where fields inside a hidden Section are incorrectly dehydrated.
- Documents the expected behavior: hidden fields inside hidden Sections should not be present in the dehydrated state.

P.S. Similar to another pr I have created before https://github.com/filamentphp/filament/pull/16295

### DIFF
--- a/packages/forms/src/Components/Concerns/HasState.php
+++ b/packages/forms/src/Components/Concerns/HasState.php
@@ -4,6 +4,7 @@ namespace Filament\Forms\Components\Concerns;
 
 use Closure;
 use Filament\Forms\Components\Component;
+use Filament\Forms\Components\Section;
 use Filament\Forms\Get;
 use Filament\Forms\Set;
 use Illuminate\Contracts\Support\Arrayable;
@@ -189,7 +190,7 @@ trait HasState
             // we need to dehydrate the child component containers while
             // informing them that they are not dehydrated, so that their
             // child components get removed from the state.
-            foreach ($this->getChildComponentContainers() as $container) {
+            foreach ($this->getChildComponentContainers(withHidden: true) as $container) {
                 $container->dehydrateState($state, isDehydrated: false);
             }
 
--- a/packages/forms/src/Components/Concerns/HasState.php
+++ b/packages/forms/src/Components/Concerns/HasState.php
@@ -4,7 +4,6 @@ namespace Filament\Forms\Components\Concerns;
 
 use Closure;
 use Filament\Forms\Components\Component;
-use Filament\Forms\Components\Section;
 use Filament\Forms\Get;
 use Filament\Forms\Set;
 use Illuminate\Contracts\Support\Arrayable;
--- a/tests/src/Forms/StateTest.php
+++ b/tests/src/Forms/StateTest.php
@@ -618,7 +618,7 @@ test('components in a hidden parent component with a grandparent component with 
         ->fill();
 
     expect($container)
-        ->dehydrateState()->toBe(['data' => []]);
+        ->dehydrateState()->toBe(['data' => ['data' => []]]);
 });
 
 test('disabled components are excluded from state dehydration', function () {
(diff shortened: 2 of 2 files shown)


alanost:
Hey, just wanted to quickly add that this also affects fieldsets and grids, not just sections. Hopefully, there'll be a working solution soon. Big thanks to the Filament team!

danharrin (approved):

schroeder-hesel:
This PR introduces a breaking change that causes data loss in forms with conditional fields sharing the same state path.

## Issue Description

**Error**: `SQLSTATE[HY000]: General error: 1364 Field 'status' doesn't have a default value`

**Root Cause**: The new logic in `dehydrateState()` incorrectly removes state data when multiple components share the same state path, even when one of them is visible and should be dehydrated.

## Use Case

I have a form with multiple conditional `Fieldset` components, each containing a `Select::make('status')` field. Only one fieldset is visible at a time based on the selected item type (type_a, type_b, type_c, etc.). This is a common pattern for dynamic forms.

```php
// Fieldset 1 - visible when type is 'type_a'
Fieldset::make('Options')
    ->hidden(fn($get) => $get('item_type') !== 'type_a')
    ->schema([
        Select::make('status')->options($options_a),
    ]),

// Fieldset 2 - visible when type is 'type_b' 
Fieldset::make('Options')
    ->hidden(fn($get) => $get('item_type') !== 'type_b')
    ->schema([
        Select::make('status')->options($options_b),
    ]),

// Fieldset 3 - visible when type is 'type_c'
Fieldset::make('Options')
    ->hidden(fn($get) => $get('item_type') !== 'type_c')
    ->schema([
        Select::make('status')->options($options_c),
    ]),
```

## Problem with New Logic

The problematic code:
```php
if (! $rootContainer->hasDehydratedComponent($statePath)) {
    Arr::forget($state, $statePath);
}
```

This logic assumes a one-to-one mapping between state paths and components, but doesn't account for:

1. Multiple components sharing the same state path (like multiple `status` fields)
2. Only one component being visible/active at a time
3. Hidden components legitimately not being dehydrated

## Expected vs Actual Behavior

**Expected**: The visible `status` field should be dehydrated and its value saved to the database.

**Actual**: All `status` state is removed because the method finds hidden components with the same state path, causing the database insert to fail due to missing required field.

## Impact

This breaks any form using conditional fields with shared state paths - a common pattern in complex, dynamic forms. The change appears to be intended for optimization but introduces a regression for valid use cases.

## Suggested Fix

The logic should check if ANY component with the given state path should be dehydrated, not just whether the root container has a dehydrated component with that path. The current implementation doesn't properly handle the case where multiple components share state paths in conditional layouts.

phit:
this also caused a bunch of issues with dynamic forms for us, please revert this breaking change

danharrin:
Please open an issue with a reproduction repository containing the broken code, so we can investigate further. Reporting issues in PR comments is not ideal

PR #15234, opened 2025-01-03
Feat: list translations command option

VasasA:
## Description

This PR introduces a new option for 'check-translations' command for Filament 4.x. The 'check-translations' command helps to find missing and deleted translations. You need to see all translations to find all errors, like typos and changed English strings. You can list all translations with new `--list` option.
(Related unmerged [PR](https://github.com/filamentphp/filament/pull/15201).)

## Functional changes

Example usage: list Filament hungarian translations:
```bash
php artisan filament:check-translations hu --list
```
Command output:
kép

- [x] Code style has been fixed by running the `composer cs` command.
- [x] Changes have been tested to not break existing functionality.
- [x] Documentation is up-to-date.

PR #15201, opened 2024-12-29
Feat: list translations command

VasasA:
## Description

This PR introduces a new command to list translations for Filament 4.x.
"You need to fix the translations where the string has changed. If you want to check the strings, you can show all translations."

## Command Signature
kép

## Functional changes

Example usage: list filament hungarian translations
```bash
php artisan filament:list-translations hu
```
Command output:
kép

- [x] Code style has been fixed by running the `composer cs` command.
- [x] Changes have been tested to not break existing functionality.
- [x] Documentation is up-to-date.

Note: I used the code of '[check-translations](https://github.com/filamentphp/filament/pull/3881)' command.  @saade Thank You!

PR #16426, opened 2025-06-07
Translation Tool

pxlrbt:
I wrote this tool to have a better overview over the status of translations. It has following improvements over the `CheckTranslationsCommand`

- Lists translation managers
- Shows all translations keys as outdated if a file is missing completely
- **Summary with coverage for all locales**
- **No need to setup a Laravel app. You can check the translations right in the repo.**

**It's missing the following features:**

- Check the app folder of a project. Since it's written as a separate tool, it can't scan the translations of a project anymore. From a maintainer perspective this is not needed, but it's probably useful for developers, that have published the lang files. We could keep the current `CheckTranslationsCommand` for this purpose or go all-in and distribute this as a FrankenPHP binary.
- It will only check one locale at a time. I found the output for multiple locales confusing. I guess people would go from one locale to the next if they maintain multiple.
- It can't be run directly from the terminal as you need to select some options via Laravel Prompt.

What are your thoughts?


pxlrbt:
@danharrin I refactored the tool a litte. I use the root composer file again. Also moved to separate tables with the filename above as in the `CheckTranslationCommand` again, because lines where getting to long.

I included a command to list all translations side by side with their original names as asked in #15201 #15234
image

danharrin (approved):

VasasA:
Hey @pxlrbt ,
This tool works fine 👍 
Thank You for the 'List translations side-by-side' option too. Very useful with the highlights.
I tested it with these commands (WSL Ubuntu 24.04.1, PHP 8.3.6):
```bash
cd ~
git clone https://github.com/filamentphp/filament.git
cd filament
git checkout 4.x
composer install
./bin/translation-tool.php
```

adevade:
I just tested on Windows without WSL, and it doesn't run because of Laravel Prompts. Is it possible to configure fallbacks?

```log
PHP Fatal error:  Uncaught RuntimeException: Prompts is not currently supported on Windows. Please use WSL or configure a fallback. in C:\Users\adevade\packages\filamentphp-filament\vendor\laravel\prompts\src\Prompt.php:435
Stack trace:
#0 C:\Users\adevade\packages\filamentphp-filament\vendor\laravel\prompts\src\Prompt.php(117): Laravel\Prompts\Prompt->checkEnvironment()
#1 C:\Users\adevade\packages\filamentphp-filament\vendor\laravel\prompts\src\helpers.php(47): Laravel\Prompts\Prompt->prompt()
#2 C:\Users\adevade\packages\filamentphp-filament\bin\translation-tool.php(15): Laravel\Prompts\select('Choose the comm...', Array, 'status')
#3 {main}
  thrown in C:\Users\adevade\packages\filamentphp-filament\vendor\laravel\prompts\src\Prompt.php on line 435
```

pxlrbt:
Thanks for the feedback, @adevade. I hoped the fallbacks would just work. Maybe I need to set it up as a Laravel Zero app 😅 More boilerplate but probably faster to get it right on Windows 🙈

adevade:
@pxlrbt Here's where the Laravel Framework sets up the fallbacks: https://github.com/laravel/framework/blob/f717cce43c02e809c1ea77811ee46bb5e06b616d/src/Illuminate/Console/Concerns/ConfiguresPrompts.php#L28

Don't know what's more work. Implementing these again or just going Laravel Zero 😅

danharrin:
From what I've heard, how does anyone work on Windows without WSL? 😅 it sounds impossible in so many ways

adevade:
@danharrin Actually thanks to Herd it's not that bad, at least if you're just doing basic stuff 😅 Maybe I have to give WSL a try again, it's been a couple of years...

PR #16429, opened 2025-06-08
Replace button with span for sortable column headers

anasgets111:
## Description
Improve table header semantics: Replace button with span for sortable columns
What & Why:
- Replaces `<button>` wrapper with `<span>` + `cursor-pointer` for sortable table headers
- Maintains identical visual behavior and functionality
- Opens possibilities for richer header experiences (allows adding custom livewire components, and actions to the label without breaking)
## Visual changes
image
image
image

## Functional changes

### DIFF
--- a/packages/tables/resources/views/components/header-cell.blade.php
+++ b/packages/tables/resources/views/components/header-cell.blade.php
@@ -26,7 +26,11 @@
     <span
         @if ($sortable)
             aria-label="{{ trim(strip_tags($slot)) }}"
+            role="button"
+            tabindex="0"
             wire:click="sortTable('{{ $name }}')"
+            @keydown.enter="$wire.sortTable('{{ $name }}')"
+            @keydown.space.prevent="$wire.sortTable('{{ $name }}')"
         @endif
         @class([
             'group flex w-full items-center gap-x-1',
--- a/packages/tables/resources/views/components/header-cell.blade.php
+++ b/packages/tables/resources/views/components/header-cell.blade.php
@@ -29,8 +29,8 @@
             role="button"
             tabindex="0"
             wire:click="sortTable('{{ $name }}')"
-            @keydown.enter="$wire.sortTable('{{ $name }}')"
-            @keydown.space.prevent="$wire.sortTable('{{ $name }}')"
+            x-on:keydown.enter="$wire.sortTable('{{ $name }}')"
+            x-on:keydown.space.prevent="$wire.sortTable('{{ $name }}')"
         @endif
         @class([
             'group flex w-full items-center gap-x-1',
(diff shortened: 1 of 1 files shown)


muhammadsaeedparacha:
Bump

pxlrbt:
@muhammadsaeedparacha Did you really just bump a PR that was opened 5(!) hours ago? 😅

pxlrbt:
What's the goal of the change? To nest something inside the header cell? I think the `<button>` element is the most accurate element and we should keep it for accessibility.

muhammadsaeedparacha:
@pxlrbt 
Sorry I didn't mean any offense. I just meant this PR would really help with an issue. Currently if we inject a livewire component inside label, it doesn't work because it's nested inside a button. But anything else and it works like a charm.

I'll see if I can come up with a solution that can retain the button and still achieve the desired result

muhammadsaeedparacha:
@pxlrbt 
One solution is that we wrap the icon in a button, and the overall label in a span.
This way livewire components work inside header-cell + accessibility is retained in the on the sort icon.

What do you think?

muhammadsaeedparacha:
Or if we give it a role="button" and a tabindex, it should solve accessibility & keyboard navigation. Does this work?

```
<span
        @if ($sortable)
            aria-label="{{ trim(strip_tags($slot)) }}"
            role="button"
            tabindex="0"
            wire:click="sortTable('{{ $name }}')"
            @keydown.enter="$wire.sortTable('{{ $name }}')"
            @keydown.space.prevent="$wire.sortTable('{{ $name }}')"
        @endif
        @class([
            'group flex w-full items-center gap-x-1',
            'cursor-pointer' => $sortable,
            'whitespace-nowrap' => ! $wrap,
            'whitespace-normal' => $wrap,
            match ($alignment) {
                Alignment::Start => 'justify-start',
                Alignment::Center => 'justify-center',
                Alignment::End => 'justify-end',
                Alignment::Left => 'justify-start rtl:flex-row-reverse',
                Alignment::Right => 'justify-end rtl:flex-row-reverse',
                Alignment::Justify, Alignment::Between => 'justify-between',
                default => $alignment,
            },
        ])
    >
```

anasgets111:
> Or if we give it a role="button" and a tabindex, it should solve accessibility & keyboard navigation. Does this work?
> 
> ```
> <span
>         @if ($sortable)
>             aria-label="{{ trim(strip_tags($slot)) }}"
>             role="button"
>             tabindex="0"
>             wire:click="sortTable('{{ $name }}')"
>             @keydown.enter="$wire.sortTable('{{ $name }}')"
>             @keydown.space.prevent="$wire.sortTable('{{ $name }}')"
>         @endif
> ```

I tested this locally and it seems to work, does it handle all accessibility needs ?

danharrin:
Please add those attributes to the PR

anasgets111:
I have done that. Let me know if anything further is needed.

danharrin (approved):

PR #16462, opened 2025-06-11
Docs: v4 Installation on Windows

pxlrbt:
The caret ^ is used as an escape character in Powershell. No quoting helps. The tilde works though and is almost similar in semantics. `~4.0` will allow updated until `5.0`.

### DIFF
--- a/packages/upgrade/bin/filament-v4
+++ b/packages/upgrade/bin/filament-v4
@@ -91,7 +91,13 @@ foreach (json_decode(file_get_contents('composer.json'), true)['require'] as $pa
         continue;
     }
 
-    $requireCommands[] = "composer require {$package}:\"^4.0\" -W --no-update";
+    $isWindowsPowershell = str_starts_with(php_uname(), 'Windows');
+
+    if (! $isWindowsPowershell) {
+        $requireCommands[] = "composer require {$package}:\"^4.0\" -W --no-update";
+    } else {
+        $requireCommands[] = "composer require '\"{$package}:^4.0\"' -W --no-update";
+    }
 }
 
 $requireCommands = implode("</strong><br /><strong>", $requireCommands);
--- a/docs/01-introduction/02-installation.md
+++ b/docs/01-introduction/02-installation.md
@@ -38,7 +38,13 @@ Installation comes in two flavors, depending on whether you want to build an app
 
 ## Installing the panel builder
 
-Since Filament v4 is in beta, you will need to set the `minimum-stability` in your `composer.json` file to be `beta` before installing any packages:
+Since Filament v4 is in beta, you will need to set the `minimum-stability` in your `composer.json` file to be `beta` before installing any packages. Either adjust it manually or via CLI:
+
+```bash
+composer config minimum-stability beta
+```
+
+Your `composer.json` should look like this:
 
 ```json
 {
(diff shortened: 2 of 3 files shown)


danharrin (approved):

packages/upgrade/bin/filament-v4
```diff
@@ -91,7 +91,13 @@ foreach (json_decode(file_get_contents('composer.json'), true)['require'] as $pa
...
     }
 
-    $requireCommands[] = "composer require {$package}:\"^4.0\" -W --no-update";
+    $isWindowsPowershell = str_starts_with(php_uname(), 'Windows');
+
+    if (! $isWindowsPowershell) {
+        $requireCommands[] = "composer require {$package}:\"^4.0\" -W --no-update";
+    } else {
+        $requireCommands[] = "composer require '\"{$package}:^4.0\"' -W --no-update";
+    }
```
adevade, line 100: @pxlrbt I think this will fail if using CMD 😅 `php_uname()` returns _true_ for CMD as well, and the double quote thing throws a Composer error: `Could not parse version constraint ^4.0': Invalid version string "^4.0'"`

danharrin: `php_uname()` is a string, what does it return on CMD compared to Powershell?

adevade: Yeah sry, I meant that `$isWindowsPowershell` returns _true_ for both. `php_uname()` returns this string on my machine: `Windows NT G14 10.0 build 26100 (Windows 11) AMD64`. Same for both PWSH and CMD.

danharrin: Does ^^^^4.0 as the version constraint work on both?

adevade: | Command | PowerShell | CMD |
| --- | :---: | :---: |
| `composer require filament/filament:^4.0` | ❌ | ❌ |
| `composer require filament/filament:"^4.0"` | ❌ | 🟢 |
| `composer require "filament/filament:^4.0"` | ❌ | 🟢 |
| `composer require "filament/filament:^^4.0"` | ❌ | ❌ |
| `composer require "filament/filament:^^^^4.0"` | ❌ | ❌ |
| `composer require 'filament/filament:^4.0'` | ❌ | ❌ |
| `composer require '"filament/filament:^4.0"'` | 🟢 | ❌ |
| `composer require 'filament/filament:"^4.0"'` | 🟢 | ❌ |

danharrin: How's `composer require filament/filament:"^^^^4.0"`

adevade: Doesn't work on either. PowerShell tries to install exactly `4.0`, and on CMD Composer throws an error: `Could not parse version constraint ^^^^4.0: Invalid version string "^^^^4.0"`

pxlrbt: So `>=4.0,<5.0` on Windows? 😅

danharrin: Honestly `~` might be the only option for consistency on Windows, as much as I hate it.

danharrin: Before I release, does 4986df2aeb19b169ed1e7cc34d1a2513aa85aa04 look good?

adevade: The commands run on both terminals 👍

PR #16468, opened 2025-06-11
[4x]Allow array as record

maartenpaauw:
## Description

When using custom data records, an `array` is passed to both callback instead of a Eloquent `Model`. Therefore, we have to add `array` as typehint.

## Visual changes

None.

## Functional changes


tomschlick:
I'm running into this same issue. These fixes help.

Though there are more to add though to the HasCellState::hasRelationship() and many on the HasAuthorization trait. If this is accepted I can make a PR for those.

danharrin:
I appreciate the work, and feel free to reply to this telling me I'm wrong. 

I don't really understand this pull request. The idea behind the custom table data feature is that you *aren't* using Eloquent for your data.

Since resources in Filament are ONLY for managing Eloquent data sources, I don't intend that people use custom table data with resources. In my head, it doesn't make sense. You can't create a resource without an Eloquent model in the first place.

I intend that users create their own "custom pages" in a panel, where you can then add a table to the page and add whatever data you want inside. But inside a resource? I am not convinced that is the correct approach yet.

tomschlick:
@danharrin Ah I completely missed the ability to do custom pages. Thanks I'll go that route.

maartenpaauw:
Thanks for the explanation. I now realize I misunderstood when and how to use custom table data. It makes sense not to use it within resources. I’ll switch to using custom pages instead.

PR #16353, opened 2025-05-28
4.x - allowPaste opton for FileUpload restored

bretto36:
## Description

Adding this for v4, as requested by Dan. I'm setting the default value to true. I can attest that it works without the Filepond changes as TipTap puts `contenteditable="true"` instead of just `contenteditable` with no attribute value

## Visual changes

[Filament file paste.webm](https://github.com/user-attachments/assets/7850e1a1-7a8b-474e-8fc9-c35423690cf2)

## Functional changes

- [x] Code style has been fixed by running the `composer cs` command.
- [x] Changes have been tested to not break existing functionality.
- [x] Documentation is up-to-date.

PR #16472, opened 2025-06-12
Allow pasteable() for FileUpload after fixing underlying JS Package problem in filepond

bretto36:
## Description

Introducing pasteable() call as i have fixed the underlying Filepond issue causing the Editor to catch the paste. Defaulting to false in v3 to maintain backwards compatibility but v4 is true by default 

## Visual changes

As my previous PR #16353 for v4 - which works due to the switch to tiptap, this will work for the v3 rich editor as filepond is updated  pqina/filepond#1045 https://github.com/pqina/filepond/pull/1045

## Functional changes

- [*] Code style has been fixed by running the `composer cs` command.
- [*] Changes have been tested to not break existing functionality.
- [*] Documentation is up-to-date.


danharrin (changes requested):
Please check the v4 implementation of this feature, and ensure it is 1:1 consistent with how I added it. There's different default behaviour and different docs, as a starting point.

bretto36:
@danharrin Yes i didn't want it to default to on all of a sudden for people using v3, as they might be surprised if it went on all of a sudden. Only v4 has a change to make it default.

If oyu want me to make it default to true that that is fine i will do that, I just wanted to let you know the reasoning of why i left it default false.

danharrin:
Please copy v4 as close to 1:1 as possible

bretto36:
@danharrin  ready! thanks for your patience

danharrin (approved):

PR #16516, opened 2025-06-13
[4.x] Fix default sort

bzy107:
## Description
I noticed unexpected sorting behavior during the Filament v4 operational check.
The defaultSort() method is being used, but the model's primary key is still being sorted as well.
スクリーンショット 2025-06-14 065738

## Functional changes


danharrin:
The additional primary key sorting is intentional in v4 to ensure sorting consistency when the default sort column contains duplicate values. Does your Eloquent model not have a primary key? Please expand on this by opening a bug report with a reproduction repository

PR #16518, opened 2025-06-13
[v4] Removes redundant section and command in documentation

bryanjamesmiller:
## Description

Hi Dan, from what I can tell there is just a bit of redundancy in the 4.x docs.  This PR removes apparently redundant command and section as same is accomplished by "Creating a resource" section above and "php artisan make:filament-resource Customer" command seems to do the same thing if you add the generate flag.  I could definitely be wrong though as this is my first time using Filament :) 

## Visual changes

This is just a documentation update

## Functional changes


danharrin:
Hey Bryan! Good to see you. The `--generate` command reads the database schema and adds columns/fields to the resource automatically. If you don't pass this option, Filament will ask you again in one of the command prompts, so this is useful if you want to bypass that :) Thanks for the contribution though

Issue #16257, opened 2025-05-13
v4 Some styles don't render properly when Vite dev server is running


binaryfire:
### Package

filament/filament

### Package Version

4-alpha6

### Laravel Version

12

### Livewire Version

3

### PHP Version

8.4

### Problem description

Some button styles aren't rendering properly when using the Vite dev server. I'm not sure if this is a CORS issue that only affects some people. There was an attempted fix a few months ago but people still seem to be having problems: https://github.com/vitejs/vite/issues/19239.

When using `npm run dev`:
![Image](https://github.com/user-attachments/assets/5b658a29-70a4-4005-9a04-df8efced86ff)

When _**not**_ using `npm run dev`:
![Image](https://github.com/user-attachments/assets/8801c26d-a2d2-4db0-a53d-6d35bf8fdafc)

### Expected behavior

All styles should render properly when using `npm run dev`

### Steps to reproduce

1. Clone reproduction repo
2. Run `npm run dev` to start the Vite dev server
3. Go to `Vite Dev Styles Bug` page. The styling of the action button and toggle buttons are broken.

4. Stop the Vite dev server.
5. Go to `Test page`. The styling is ok.

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/binaryfire/filament4-vite-dev-server-styles

### Relevant log output

```shell

```

danharrin:
Is this not similar to #16121?

binaryfire:
No. Fixing the build errors got hot reloading working again, but turns out the styling issues were unrelated. They're still there and are happening when running the normal Vite dev server. I also just confirmed it's not CORS-related.

Are you dynamically generating any class names? Vite wouldn't know how to render those. Other than that, I'm not sure what could be causing it. It's probably something in the CSS as opposed to a Vite issue.

Are you able to reproduce the problem?

binaryfire:
Additional info:

- It appears CSS selectors using `+ &` (eg. `input:checked + &label`) aren't processed correctly by Vite's dev server, causing toggle buttons and ring outlines to not display properly. Similar issues:
https://github.com/tailwindlabs/tailwindcss/discussions/15524
https://github.com/parcel-bundler/lightningcss/issues/489

- The following changes to `packages/support/resources/css/components/button.css` work but aren't clean enough to merge: https://gist.github.com/binaryfire/15c2a9b9c08a4358d9ab4cc6318f6d82

binaryfire:
[...]

PR #16527, opened 2025-06-14
Vite dev server button styles: identify issue + POC fix

binaryfire:
_**NOTE: This PR is to demonstrate the problem and a possible fix. Hopefully you or Zep can figure out a cleaner solution.**_

CSS selectors using `+ &` (eg. `input:checked + &label`) are not being processed correctly by Vite's dev server, causing toggle buttons and ring outlines to not display properly. It's probably a JIT or Lightning CSS issue. Lightning CSS seems to have quirks with nested selectors:

https://github.com/tailwindlabs/tailwindcss/discussions/15524
https://github.com/parcel-bundler/lightningcss/issues/489

If you do the following:
- Clone the reproduction repo from https://github.com/filamentphp/filament/issues/16257
- Run `composer dev` to start the Vite dev server, then inspect the CSS on the action / toggle buttons on 'Test Page'
- Do the same thing using the compiled theme.css
- Then compare the two

You'll see the CSS that's missing

I refactored the problematic selectors to use explicit class names instead of parent selectors after combinators:
- Changed `input:checked + &label` → `input:checked + .fi-btn:is(label)`
- Split comma-separated selectors like `&:not(.fi-color), &label` into separate rules
- Moved checked state styles outside of the nested blocks

Not pretty but it works. I've done as much as I can but at least the problem has been identified. This one took a long time to debug 🙃

Might be worth opening an issue in the Tailwind repo to see if they can fix it.


danharrin:
Yeah I agree that this is not a pretty fix. I am aware of the issue but I don't think it should be our responsibility to have to write selectors like this, they are ugly and not maintainable. This problem doesn't affect production apps, so I think someone should submit a bug report to Tailwind to get it fixed. Thanks for your investigation though.

binaryfire:
> Yeah I agree that this is not a pretty fix. I am aware of the issue but I don't think it should be our responsibility to have to write selectors like this, they are ugly and not maintainable. This problem doesn't affect production apps, so I think someone should submit a bug report to Tailwind to get it fixed. Thanks for your investigation though.

@danharrin Fair enough but I really disagree on this one. `composer dev` is basically broken with Filament at the moment, and it's something lots of Laravel devs use. You can't even see if a toggle button is selected. It's very difficult to develop without it when you're using a custom theme or adding extra TW classes via -`>extraAttributes()`. Every small change requires a npm rebuild to view. It's pretty big pain point.

danharrin:
Have you tried rewriting the selector so part of it is nested and part isnt? Then at least the code will be roughly in the right place for making changes and I will feel better about merging.

binaryfire:
> Have you tried rewriting the selector so part of it is nested and part isnt? Then at least the code will be roughly in the right place for making changes and I will feel better about merging.

Yeah, multiple variations. Went around in circles for quite a while and couldn't find a clean solution. In the end I figured I'd PR this to help identify the problem and someone with better frontend skills could find a cleaner fix. Maybe mark it as help wanted?

danharrin:
Done. I think you should open that bug report for Tailwind though, that is the correct next step

Issue #16532, opened 2025-06-15
CSS Cascade Bug: Nested Sections with contained(false) Ignore :not() Selectors


andrewdwallo:
### Package

filament/filament

### Package Version

v4.0.0-beta4

### Laravel Version

v12.18.0

### Livewire Version

v3.6.3

### PHP Version

PHP 8.2

### Problem description

When sections are nested, the inner section with `contained(false)` still receives contained styles due to CSS cascade issues with `:not()` selectors. Or at least I think that's the issue.

### Expected behavior

Inner sections with `contained(false)` should not receive padding regardless of their parent section's contained state.

### Steps to reproduce

Example is in UserResource in reproduction repo.

1. Create a Filament form with nested sections:

```php
Section::make('User Information')
    ->columnSpanFull()
    ->schema([
        Section::make('Personal Details')
            ->contained(false)  // Should have no padding, but does
            ->schema([
                TextInput::make('name')->required(),
                TextInput::make('email')->required(),
            ]),
    ])
```

2. Inspect the inner section header in browser DevTools
3. Observe: Inner section has class `fi-section-not-contained` (correct)
4. Problem: Header still receives padding from outer section's CSS rules

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/andrewdwallo/filament-test

### Relevant log output

```shell

```

PR #16533, opened 2025-06-15
Fix CSS Cascade Bug for Nested Sections

andrewdwallo:
## Description

This PR fixes a CSS cascade bug #16532 where nested sections were incorrectly inheriting styling from their parent sections, affecting all section properties including `contained()`, `compact()`, `aside()`, `secondary()`, and `divided()` methods.

**Problem:** The CSS selectors for section styling were using descendant selectors instead of direct child selectors (`>`), causing styles to cascade down to nested sections and override their individual styling settings.

**Solution:** Updated CSS selectors to use direct child selectors (`>`) for layout-affecting properties while preserving descendant selectors for internal styling elements. This ensures proper section isolation regardless of nesting depth.

### Before
Screenshot 2025-06-14 at 10 33 15 PM

### After
Screenshot 2025-06-14 at 10 33 35 PM

## Functional changes


andrewdwallo:
I haven't tested other components, but I imagine this could potentially be an issue for other layout components, particularly components which can be nested within each other. 

**Edit:** I think it might have been just the section

danharrin (approved):

PR #16535, opened 2025-06-15
Fix TOTP code expiration time documentation

chengkangzai:
## Description
This PR fixes an inconsistency in the multi-factor authentication documentation regarding TOTP code expiration timing.
Changes Made

Updated the documentation to correctly state that with a window of 8, TOTP codes are valid for 8 minutes after generation (not 4 minutes as previously stated)
This aligns with the standard TOTP algorithm where each window typically represents 30-second intervals, making a window of 8 equal to approximately 4 minutes in each direction (8 × 30 seconds = 240 seconds = 4 minutes), totaling 8 minutes of validity
## Visual changes

Screenshot 2025-06-15 at 12 38 53 PM

## Functional changes

P.S. Loving the new beta features! Good Job Filament Team! 🚀

### DIFF
--- a/docs/07-users/02-multi-factor-authentication.md
+++ b/docs/07-users/02-multi-factor-authentication.md
@@ -275,7 +275,7 @@ public function panel(Panel $panel): Panel
 
 ### Changing the app code expiration time
 
-App codes are issued using a time-based one-time password (TOTP) algorithm, which means that they are only valid for a short period of time before and after the time they are generated. The time is defined in a "window" of time. By default, Filament uses an expiration window of `8`, which allows the code to be valid for 8 minutes after it is generated.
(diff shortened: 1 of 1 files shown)


danharrin (changes requested):
The way I read it, it's 4 minutes **after** not 4 minutes total. In my opinion, its not wrong. How can we reword it to make it clearer?

chengkangzai:
@danharrin I see, i was quite new to this topic.
i suggest 

```
By default, Filament uses an expiration window of `8`, which creates a 4-minute validity period on either side of the generation time (8 minutes total window).
```

for the last sentence, what do you think?

danharrin (approved):

PR #16545, opened 2025-06-15
Minor bugfix: Fixes NavigationItem's url parameter to not reset the shouldOpenUrlInNewTab flag

molbal:
## Description

Updates the url() method in NavigationItem, so that if in the builder pattern the url() parameter is called (omitting the optional $shouldOpenInNewTab parameter) after calling openUrlInNewTab, the shouldOpenUrlInNewTab property's state will not reset and the link will open in a new tab as expected.

### Fixed use case:

In the provided example below, before the fix, the link opened in the same page, and not a new tab.

```php

...
{
    {
        return $panel
        /// ...
            ->navigationItems([
                NavigationItem::make('Documentation')
                    ->icon(Heroicon::QuestionMarkCircle)
                    ->openUrlInNewTab()
                    ->url('https://...')
                    ->group('Help'),
            ])
       /// ...
 

```

## Visual changes

No visual changes

## Functional changes

- [x] Documentation is up-to-date. // No changes

### DIFF
--- a/packages/panels/src/Navigation/NavigationItem.php
+++ b/packages/panels/src/Navigation/NavigationItem.php
@@ -147,7 +147,7 @@ class NavigationItem extends Component
 
     public function url(string | Closure | null $url, bool | Closure $shouldOpenInNewTab = false): static
     {
-        $this->openUrlInNewTab($shouldOpenInNewTab);
+        $this->openUrlInNewTab($this->shouldOpenUrlInNewTab() || $shouldOpenInNewTab);
         $this->url = $url;
 
         return $this;
--- a/packages/panels/src/Navigation/NavigationItem.php
+++ b/packages/panels/src/Navigation/NavigationItem.php
@@ -147,10 +147,11 @@ class NavigationItem extends Component
 
     public function url(string | Closure | null $url, bool | Closure | null $shouldOpenInNewTab = null): static
     {
+        $this->url = $url;
+
         if ($shouldOpenInNewTab !== null) {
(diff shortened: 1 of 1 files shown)


packages/panels/src/Navigation/NavigationItem.php
```diff
@@ -145,9 +145,11 @@ public function sort(int | Closure | null $sort): static
         return $this;
     }
 
-    public function url(string | Closure | null $url, bool | Closure $shouldOpenInNewTab = false): static
+    public function url(string | Closure | null $url, bool | Closure | null $shouldOpenInNewTab = null): static
     {
-        $this->openUrlInNewTab($shouldOpenInNewTab);
+        if ($shouldOpenInNewTab !== null) {
+            $this->openUrlInNewTab($this->shouldOpenUrlInNewTab() || $shouldOpenInNewTab);
```
danharrin: This isn't required, right? We shouldn't call a getter from a setter

```suggestion
            $this->openUrlInNewTab($shouldOpenInNewTab);
```

danharrin (changes requested):

packages/panels/src/Navigation/NavigationItem.php
molbal: You are right, I only added it there, because $shouldOpenInNewTab can be a closure, and I did not want to copy the logic to evaluate it, but you are right, there is no need to evaluate it right away. I arrived at this after running some circles and did not notice this

danharrin (approved):

PR #16546, opened 2025-06-15
Prevent duplicate entries when using same Repeater ID across Wizard steps

alanost:
## Description

I discovered the following bug in my production application:

I have a Wizard with a Repeater that I use across two steps with the same ID. When I save the record and later return to edit it, I sometimes see that a nulled second entry is created (data: []), which is incorrect behavior.

I have created a dedicated test that tries to more or less reproduce this behavior. I fill the form with information for the repeater, but two arrays are automatically created instead of just one.

Expected Behavior:
When using a Repeater with the same ID across multiple Wizard steps, filling the form should create only one repeater item containing all the merged field data

```php
test('when repeaters share the same id across wizard steps data is filled only once', function () {
    $container = ComponentContainer::make(Livewire::make())
        ->statePath('data')
        ->components([
            Wizard::make()
                ->steps([
                    Step::make('step1')
                        ->schema([
                            Repeater::make('repeater')
                                ->schema([
                                    TextInput::make('name'),
                                ]),
                        ]),
                    Step::make('step2')
                        ->schema([
                            Repeater::make('repeater')
                                ->schema([
                                    TextInput::make('name2'),
                                ]),
                        ]),
                ]),
        ])
        ->fill([
            'repeater' => [
                'name' => 'lorem ipsum',
                'name2' => 'lorem ipsum',
            ],
        ]);

    $dehydratedState = $container->dehydrateState();

    expect($dehydratedState['data']['repeater'])
        ->toHaveCount(1) // <- fails, fill creates 2 items
        ->each(fn($item) => $item->toBe([
            'name' => 'lorem ipsum',
            'name2' => 'lorem ipsum',
        ]));
});
```

### DIFF
--- a/tests/src/Forms/StateTest.php
+++ b/tests/src/Forms/StateTest.php
@@ -1,12 +1,16 @@
 <?php
 
-use Filament\Forms\ComponentContainer;
-use Filament\Forms\Components\Component;
-use Filament\Forms\Components\Placeholder;
 use Filament\Forms\Get;
-use Filament\Tests\Forms\Fixtures\Livewire;
-use Filament\Tests\TestCase;
 use Illuminate\Support\Str;
+use Filament\Tests\TestCase;
+use Filament\Forms\Components\Wizard;
+use Filament\Forms\ComponentContainer;
+use Filament\Forms\Components\Repeater;
+use Filament\Forms\Components\Component;
+use Filament\Forms\Components\TextInput;
+use Filament\Forms\Components\Placeholder;
+use Filament\Forms\Components\Wizard\Step;
+use Filament\Tests\Forms\Fixtures\Livewire;
 
 uses(TestCase::class);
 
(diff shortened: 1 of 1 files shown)


danharrin:
Sorry, field names have to be unique, and you have two repeaters with the same name. This is not supported in Filament since it causes issues with Livewire dom diffing.

alanost:
Alright, got it. But do you still know what the best way is for two different repeaters to access the same data structure? So that in the first step, only the names of the household members are specified, and in the second step, the incomes of these household members are provided (also needs to be in sync)?


```php
Wizard::make()
    ->steps([
        Step::make('Step 1')
            ->schema([
                Repeater::make('member')
                    ->schema([
                        TextInput::make('name'),
                    ]),
            ]),
        Step::make('Step 2')
            ->schema([
                Repeater::make('member_income')
                    ->schema([
                        ToggleButtons::make('has_income')->boolean(),
                    ])
                    ->addable(false)
                    ->deletable(false)
                    ->itemLabel(fn (array $state): ?string => $state['name'] ?? null),
            ]),
    ])
```

danharrin:
You would need to set up an `afterStateUpdated()` on the first repeater to update the state of the second or something

alanost:
Okay, I tried but it's way too complicated for such a simple feature. Couldn't we just add an attribute to the repeater, i.e. ->key('id') to define the ID for the key under which the values should be stored? I imagine that could be helpful for many people, because sometimes you want to split one large repeater into two smaller ones.

alanost:
Ok, I think I figured out how to do this easily, namely by using repeaters with different IDs, but connecting the second repeater to the first one via `statePath`, like in this example:

```php
Wizard::make()
    ->steps([
        Step::make('Step 1')
            ->schema([
                Repeater::make('member')
                    ->schema([
                        TextInput::make('name'),
                    ]),
            ]),
        Step::make('Step 2')
            ->schema([
                Repeater::make('member_income')
                    ->statePath('member') 
                    ->schema([
                        ToggleButtons::make('has_income')->boolean(),
                    ])
                    ->addable(false)
                    ->deletable(false)
                    ->itemLabel(fn (array $state): ?string => $state['name'] ?? null),
            ]),
    ])
```

Is this theoretically possible, or would it also lead to bugs like with duplicate IDs? @danharrin

danharrin:
I can imagine issues with Filament being able to get the state out of both repeaters at once, but does it work for you?

PR #16550, opened 2025-06-16
[v4] return null when record is array

maartenpaauw:
## Description

When using custom data records, an `array` is retrieved from `getRecord` method. When the state is `null`, the record is passed to all relationship methods that only accept `Model` instances. Therefore, we need to early return `null` when record is an array.

## Visual changes

None.

## Functional changes

### DIFF
--- a/packages/support/src/Concerns/HasCellState.php
+++ b/packages/support/src/Concerns/HasCellState.php
@@ -118,6 +118,10 @@ trait HasCellState
             return $state;
         }
 
+        if (is_array($record)) {
+            return null;
+        }
+
         if (! $this->hasRelationship($record)) {
             return null;
         }
--- a/packages/support/src/Concerns/HasCellState.php
+++ b/packages/support/src/Concerns/HasCellState.php
@@ -118,7 +118,7 @@ trait HasCellState
             return $state;
         }
 
-        if (is_array($record)) {
+        if ($this instanceof Column && is_array($record)) {
             return null;
         }
 
--- a/packages/support/src/Concerns/HasCellState.php
+++ b/packages/support/src/Concerns/HasCellState.php
@@ -118,7 +118,7 @@ trait HasCellState
             return $state;
         }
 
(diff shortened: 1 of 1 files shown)


maartenpaauw:
I'd try to solve the PHPStan error in the latest commit, but I'm not entirely sure how to solve it. Feel free to drop last commit.

packages/support/src/Concerns/HasCellState.php
```diff
@@ -118,6 +118,10 @@ public function getStateFromRecord(): mixed
             return $state;
         }
 
+        if ($this instanceof Column && is_array($record)) {
```
danharrin:
```suggestion
        if (($this instanceof Column) && is_array($record)) {
```

danharrin (changes requested):

packages/support/src/Concerns/HasCellState.php
```diff
@@ -118,6 +118,10 @@ public function getStateFromRecord(): mixed
             return $state;
         }
 
+        if (($this instanceof Column) && is_array($record)) {
```
danharrin:
```suggestion
        if (($this instanceof Column) && is_array($record)) { /** @phpstan-ignore function.impossibleType */
```

danharrin (changes requested):

packages/support/src/Concerns/HasCellState.php
```diff
@@ -118,6 +118,10 @@ public function getStateFromRecord(): mixed
             return $state;
         }
 
+        if (($this instanceof Column) && is_array($record)) { /** @phpstan-ignore function.impossibleType */
```
danharrin:
```suggestion
        if (($this instanceof Column) && is_array($record)) { /** @phpstan-ignore function.impossibleType, booleanAnd.alwaysFalse */
```

danharrin (changes requested):

danharrin (approved):

Issue #16537, opened 2025-06-15
[4.x] Bug in defaultSort


bzy107:
### Package

filament/filament

### Package Version

v4.0.0-beta4

### Laravel Version

v11.45.1

### Livewire Version

v3.6.3

### PHP Version

PHP 8.3.15

### Problem description

Bug report #16516

I've set up the local environment and accessed http://localhost/admin/custom-treatments. 
This resulted in the following error:
Even though defaultSort is specified with a custom query, the model's primary key is still being used for an additional sort.
![Image](https://github.com/user-attachments/assets/bbf97df7-e6b8-4e63-a64c-70b05aa63750)

SQLSTATE[42S22]: Column not found: 1054 Unknown column 'treatments.id' in 'order clause'
```sql
select
    `sub`.`id`,
    `sub`.`description`,
    `sub`.`name`,
    `sub`.`type`
from
    (
        select
            `t`.`id`,
            `t`.`description`,
            `p`.`name`,
            `p`.`type`
        from
            `treatments` as `t`
            inner join `patients` as `p` on `t`.`patient_id` = `p`.`id`
        where
            `t`.`is_valid` = 1
    ) as `sub`
order by
    `sub`.`name` asc,
    `treatments`.`id` asc -- this is unexpect sort
limit
    10 offset 0
```

Even though defaultSort is configured, the table is still sorting by id. This is an unintended bug that didn't occur in v3.

### Expected behavior

Ensure http://localhost/admin/custom-treatments loads without errors.

### Steps to reproduce

local set up and php artisan migrate:fresh --seed
next, access: http://localhost/admin/custom-treatments

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/bzy107/filament-select-bug

### Relevant log output

```shell

```

danharrin:
In #16602, you can now disable this with `$table->defaultKeySort(false)`

PR #16552, opened 2025-06-16
Add orderBy qualified keyname for stable pagination sort

### LINKED ISSUE #15865: Missing table items on globally scoped models when sorting by created_at
### Package

filament/filament

### Package Version

v3.3.4

### Laravel Version

v12.2.0 also v11.44.2

### Livewire Version

v3.6.2

### PHP Version

8.2.26

### Problem description

1. Globally Scope a model.
2. Make sure there are at least 5 entries on the same date.
3. Make sure to page by 5 or ten ( larger pages seem to work )
4. Sort table by created_at.
5. Some items are missing from filament tables.

### Expected behavior

All items should appear eventually.

running regular laravel paginators. I.e.: 
```php
$x = \App\Models\Customer::paginate(perPage:10, page: 3)->items()
```
In tinker works as expected

### Steps to reproduce

Just migrate and seed the example repo.

If you set ```if ($cnt > 5) {``` in database/factories/CustomerFactory.php:26
Then id 41 is always missing when sorting by created_at.
If you set ```if ($cnt > 15) {``` then then 63 is missing.

https://github.com/helios-live/filament-paging-bug

### Relevant log output

```shell

```

barryvdh:
Makes sure to add a unique key to the orderBy to prevent missing/duplicate items when pagination because of unstable sorting, when keys are the same.

## Description

This adds a default sort to pagination on the keyname, so that a unique key is always available when sorting, in the case of keys with the same value. Fixes https://github.com/filamentphp/filament/issues/15865

## Visual changes
Sorting on provider type (which is AWS for multiple). See that 25 and 26 are on page 1 and 3. There are some values missing entirely when paginating.

Before
Screenshot 2025-06-16 at 11 27 31
After
Screenshot 2025-06-16 at 11 28 55

## Functional changes

The query changes to add a secondary orderBy. This could result to duplicate orderBy but I don't think this has any effect?

```
select * from `servers` where `servers`.`deleted_at` is null order by `provider` asc, `servers`.`id` asc limit 10 offset 20
select * from `servers` where `servers`.`deleted_at` is null order by `servers`.`id` asc, `servers`.`id` asc limit 10 offset 20
```

### DIFF
--- a/packages/tables/src/Concerns/CanPaginateRecords.php
+++ b/packages/tables/src/Concerns/CanPaginateRecords.php
@@ -30,6 +30,9 @@ trait CanPaginateRecords
     {
         $perPage = $this->getTableRecordsPerPage();
 
+        // Add orderBy key to ensure stable order results
+        $query->orderBy($query->getModel()->getQualifiedKeyName());
+        
         if (version_compare(App::version(), '11.0', '>=')) {
             $total = $query->toBase()->getCountForPagination();
 


danharrin:
Unfortunately, I think this is a breaking change. Some people are modifying the table queries to change the "source" table of the query, and some of those don't have primary keys that you can sort by.

In v4, I made a similar change several months ago, but it appears to have already caused issues for people using the new beta, so I will need to work out what to do there (#16537)

Could you use some of our existing APIs to do this in v3?

barryvdh:
Maybe we could add a flag to enable/disable this?

`$table->stableSort(false)` (when enabled by default in v4, or maybe opt-in in v3)

```
    protected function applySortingToTableQuery(Builder $query): Builder
    {
        ...
        
        $column->applySort($query, $sortDirection);

        if ($this->stableSort) {
             $query->orderBy($query->getModel()->getQualifiedKeyName());
        }

        return $query;
    }
```

barryvdh:
> Maybe we could add a flag to enable/disable this?
> 
> `$table->stableSort(false)` (when enabled by default in v4, or maybe opt-in in v3)
> 
> ```
>     protected function applySortingToTableQuery(Builder $query): Builder
>     {
>         ...
>         
>         $column->applySort($query, $sortDirection);
> 
>         if ($this->stableSort) {
>              $query->orderBy($query->getModel()->getQualifiedKeyName());
>         }
> 
>         return $query;
>     }
> ```
See https://github.com/filamentphp/filament/pull/16556

PR #16556, opened 2025-06-16
Make defaultKeySort configurable

### LINKED ISSUE #15865: Missing table items on globally scoped models when sorting by created_at

barryvdh:
Makes sure to add a unique key to the orderBy to prevent missing/duplicate items when pagination because of unstable sorting, when keys are the same.

## Description

Re-does #16552 

This adds a stableSort option to tables, to add an additional pagination key so that a unique key is always available when sorting, in the case of keys with the same value. Fixes https://github.com/filamentphp/filament/issues/15865

## Visual changes
Sorting on provider type (which is AWS for multiple). See that 25 and 26 are on page 1 and 3. There are some values missing entirely when paginating.

Before
Screenshot 2025-06-16 at 11 27 31
After
Screenshot 2025-06-16 at 11 28 55

## Functional changes

This adds a `$table->stableSort()` option which adds the sorting. Default behavior in v3 is off. Maybe in v4 we can turn it on by default. (and use `$table->stableSort(false)` instead.

The query changes to add a secondary orderBy. For custom queries, in the future we might make the stable sort key configurable?

### DIFF
--- a/packages/tables/src/Concerns/CanPaginateRecords.php
+++ b/packages/tables/src/Concerns/CanPaginateRecords.php
@@ -30,11 +30,6 @@ trait CanPaginateRecords
     {
         $perPage = $this->getTableRecordsPerPage();
 
-        if ($this->getTable()->getStableSort()) {
-            // Add orderBy key to ensure stable order results
-            $query->orderBy($query->getModel()->getQualifiedKeyName());
-        }
-
         if (version_compare(App::version(), '11.0', '>=')) {
             $total = $query->toBase()->getCountForPagination();
 
--- a/packages/tables/src/Concerns/CanPaginateRecords.php
+++ b/packages/tables/src/Concerns/CanPaginateRecords.php
@@ -30,6 +30,11 @@ trait CanPaginateRecords
     {
         $perPage = $this->getTableRecordsPerPage();
 
+        if ($this->getTable()->getStableSort()) {
+            // Add orderBy key to ensure stable order results
+            $query->orderBy($query->getModel()->getQualifiedKeyName());
+        }
+
         if (version_compare(App::version(), '11.0', '>=')) {
             $total = $query->toBase()->getCountForPagination();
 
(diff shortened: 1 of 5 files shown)


danharrin (changes requested):
Please implement this in `applySortingToTableQuery()` (check the 4.x branch), not pagination.

Please add a test to `ListRecordsTest`

Please rename the methods/property to `defaultKeySort(bool | Closure $condition = true)`, `$hasDefaultKeySort` and `hasDefaultKeySort()`

Please submit an additional PR to merge the branch into 4.x, allowing the default key sort to be turned off. The merge should enable the feature

Please document the behaviour change in the 4.x upgrade guide

I realise that there are alot of requests here, but I am super busy with other 4.x issues right now and I would appreciate the help to ensure that I don't have to go back and clean up this feature.

barryvdh:
Yes I understand. I tried to do the changes as suggested, and backport the checking of the existing sort. Had to extract it to a function because the defaultSort was still separate here and didn't want to change a lot.

I tried adding a test, but have two issues
 - Not sure how to enable the flag in the test
 -  Cannot reproduce a failing test, perhaps because of sqlite.

danharrin (approved):

Issue #16565, opened 2025-06-17
Get $get returns entire state rather than defined key


geofftech:
### Package

filament/filament

### Package Version

4.0.0-beta5

### Laravel Version

^12

### PHP Version

8.3

### Problem description

Using `$get('notes')` returns the entire state on an edit form rather than just the value of the `notes` property.

At this location 

https://github.com/filamentphp/filament/blob/91186c4d67dc9c40dffad2041311b21fe7186751/packages/schemas/src/Components/Utilities/Get.php#L30

Possibly the code

```
return $component->getState();
```

should be 

```
return data_get($component->getState(),  $key);
```

### Expected behavior

`$get('notes')` should return just the `notes` key

### Steps to reproduce

```
Section::make('Notes')
    ->schema([

        Textarea::make('notes')
            ->hiddenLabel()
            ->columnSpanFull(),

    ])
    ->collapsible()
    ->collapsed(
        fn(Get $get) => blank($get('notes'))
    ),
```

In sample project (below), 

1. login into `/admin` with `test@example.com` and `password`. 
2. goto users resource
3. edit user
4. notes section should be collapsed as notes field is null

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/geofftech/filament-get-bug

### Relevant log output

```shell

```

danharrin:
It was because your "notes" section had the same name as the field so it was trying to fetch the data for the section as a whole. I have fixed that oversight: 85434be82722c60bd605dd4f1b7adc4561ec0eed

PR #16566, opened 2025-06-17
$get honors $key property

geofftech:
## Description

`Get $get` does not honour the `$key` property and returns the entire state instead.

Solves #16565

## Visual changes

none

## Functional changes


danharrin:
The tests are failing here, and from what I can see the fix doesn't look correct for this. I will take a look at your issue, or you can take another look at the root cause. Alternatively, you could submit a new PR with a failing test that I can fix with the correct solution

PR #16581, opened 2025-06-18
fix(table-columns): sort table columns when a new column is added

mvenghaus:
## Description

When a new column is added to a table .. especially during development .. the new column is always added to the end of the table as long as the session persists.

To solve this I added a sort by the names of the default state.

## Visual changes

-

## Functional changes


danharrin:
I am a little concerned that this won't work with column groups, did you test those?

And does this effectively reset the reorderable columns configuration each time a column is added?

Would appreciate @archilex taking a look at this one (when you have time) please

archilex:
@danharrin Even without adding an additional column, this resorts the array which resets any column reordering that has been done by the user.

danharrin:
Thanks I kinda thought so, this cannot be merged as it is then since it breaks that feature

PR #16582, opened 2025-06-18
[4.x] Update to enhance phpstan level

BinaryKitten:
## Description

On stream with nuno - there was a request to update the phpstan level. I posted "Consider it done.." in a sort of "Challenge accepted" manner. 

This mostly adds docblocks to denote types - but a few minor code changes have been made to remove existing level 6 issues.

Currently this is a Work in progress, thus the draft status - more commits as i find time. Once i've squashed all the level 6 issues - then I'll move forward to Level 7+ each as a new PR.


packages/actions/src/Concerns/InteractsWithSelectedRecords.php
```diff
@@ -44,6 +45,11 @@ public function canAccessSelectedRecords(): bool
         return (bool) $this->evaluate($this->canAccessSelectedRecords);
     }
 
+    /**
+     * @return EloquentCollection<int, Model>|Collection<int, Model>|LazyCollection<int, Model>
```
danharrin, line 49: Please ensure proper spacing around ` | `

packages/notifications/src/Collection.php
```diff
@@ -5,10 +5,13 @@
...
 use Livewire\Wireable;
 
+/**
+ * @extends BaseCollection<array-key, Notification>
+ */
 class Collection extends BaseCollection implements Wireable
 {
     /**
-     * @param  array<array<string, mixed>>  $items
+     * @param  array<array-key, Notification>  $items
```
danharrin, line 14: I try and avoid `array<array-key` as it usually feels purposeless, should I think about this differently for some reason? Happy to learn

packages/forms/src/Components/BaseFileUpload.php
```diff
@@ -181,7 +181,7 @@ protected function setUp(): void
     }
 
     /**
-     * @param  array<string> | Arrayable | Closure  $types
+     * @param  array<string> | Arrayable<array-key, mixed> | Closure  $types
```
danharrin, line 184: Please ensure the array type matches the `Arrayable` type, should this be `string` too? Or is there a reason?

danharrin (changes requested):

packages/actions/src/Concerns/InteractsWithSelectedRecords.php
BinaryKitten: Yep this caught me by surprise as it's not really something I've seen a lot of. Will adjust.

packages/forms/src/Components/BaseFileUpload.php
BinaryKitten: It definitely seems like a mixup on my part here.
I think i read the int, string on above property and converted in mentally as the array-key ( which is short-hand for int | string )

Will correct.

packages/notifications/src/Collection.php
BinaryKitten: It's shorthand for `int | string`. Though this looks like might be error. Will confirm and check

BinaryKitten:
Still more work to do, pressed ready too soon. Work life in the way, will make more time shortly

packages/notifications/src/Collection.php
danharrin: I feel like since those are the only options for array keys anyway, the docblock isn't narrowing anything we didn't already know. Thoughts?

danharrin:
I understand you're probably just busy, so please feel free to reopen this when the task is complete. Until then, I would rather the PR was not sitting in the list with actionable ones. Thank you for the work so far.

PR #16587, opened 2025-06-18
Use $getLivewireKey in selects

archilex:
## Functional changes


danharrin:
`$livewireComponentId !== $livewireKey` here, I think the fix is going to need to be different. Can you explain more about what the issue is?

archilex:
Sorry about that. Should have dug deeper. Basically in Advanced Filters when a dependent searchable select is unhidden as a result of another select, it throws that `using $this not in object context` error. It started to appear after you fixed the partial rendering issue that was affecting query builder and advanced filters (but that is probably just because that first error occurred before getting here). Only happens when deferred filters is on. Might be an issue as well with just normal dependent selects in filters. Haven't tested that.

PR #16612, opened 2025-06-20
New Select input

danharrin:
Returning to a custom built select input to fix issues we have with Choices.js. This time, it doesn't use Alpine to render so hopefully there will be fewer issues.


mistermantas:
hi @danharrin - either for this new select or the old one in v3, is there a way to use it as a blade component (e.g. like <x-filament::button>) and pass along the `searchable=true` property?

PR #16621, opened 2025-06-21
fix(tables): wrap summarizer columns

mvenghaus:
## Description

In one of my projects I have auto generated report columns. Some of these columns are starting with a number .. "101_qty" for example.

In V3 this was no problem, but in V4 the column names of the summarizer are not wrapped in quotes and so it's not working.

## Visual changes

-

## Functional changes


danharrin (changes requested):
I think this should probably happen within each of the `getSelectStatements()` methods instead of at a higher level?

mvenghaus:
hmm .. or using a value object for a column !?

- table prefix
- table
- column
- __toString to get the full wrapped value

danharrin:
I think we can avoid a value object for now and just call `wrap()` in `getSelectStatements()`, only if the statements need it

danharrin:
@mvenghaus do you have some time to implement the proposed changes please?

mvenghaus:
@danharrin i'll get to it today

mvenghaus:
@danharrin sry for delay .. have a lot of work .. i pushed the changes

danharrin (approved):

PR #16628, opened 2025-06-21
refactor: extract Action context generation to method

ralphjsmit:
I need to construct an action mount click handler via JavaScript (as one of the arguments is coming from JS). In V4, the click handler became more complex with the bunch of differing context keys that are available. This extracts the context generation into a specific action method, so that I can just still leave the context generation part on the action.

Thanks!

### DIFF
--- a/packages/actions/src/Action.php
+++ b/packages/actions/src/Action.php
@@ -434,6 +434,22 @@ class Action extends ViewComponent implements Arrayable
             $argumentsParameter .= Js::from($arguments);
         }
 
+        $contextParameter = '';
+
+        if (count($context = $this->getContext())) {
+            $contextParameter .= ', ';
+            $contextParameter .= Js::from($context);
+
+            if ($argumentsParameter === '') {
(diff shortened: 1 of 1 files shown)


packages/actions/src/Action.php
```diff
@@ -434,6 +434,25 @@ protected function getJsClickHandler(): ?string
...
+            if ($argumentsParameter === '') {
+                $argumentsParameter = ', {}';
+            }
+        }
+
+        return "mountAction('{$this->getName()}'{$argumentsParameter}{$contextParameter})";
+    }
+
+    /**
+     * @return array<string, string|true>
```
ralphjsmit: Strictly speaking we can do with `array<string, string|true>`, but for future flexibility `array<string, mixed>` is also a fine option.

danharrin:
```suggestion
     * @return array<string, mixed>
```

danharrin (changes requested):

danharrin (approved):

PR #16629, opened 2025-06-21
[4.x] Feat: additional rich editor extensions

awcodes:
## Description

Adds the following extensions to the Rich Editor:
Gapcursor
Hard Break
Horizontal Rule
Lead
Small
TextAlign

## Visual changes

Screenshot 2025-06-21 at 3 08 16 PM

## Functional changes


packages/forms/src/Components/RichEditor/RichContentRenderer.php
```diff
@@ -235,19 +241,29 @@ public function getTipTapPhpExtensions(): array
...
+            app(SmallExtension::class),
             app(Strike::class),
             app(Subscript::class),
             app(Superscript::class),
             app(Text::class),
+            new TextAlign([
+                'types' => ['heading', 'paragraph'],
+                'alignments' => ['start', 'center', 'end', 'justify'],
+                'defaultAlignment' => 'start',
+            ]),
```
danharrin: Please use the `app()` helper so that users can swap this class out for their own in the container

packages/forms/docs/10-rich-editor.md
```diff
@@ -56,7 +56,7 @@ RichEditor::make('content')
     ->toolbarButtons([
         ['bold', 'italic', 'underline', 'strike', 'subscript', 'superscript', 'link'],
         ['h2', 'h3'],
```
danharrin: Please add the alignment tools to the docs here

danharrin: Also, opinion on whether the justify option should be documented and enabled by default?

package.json
```diff
@@ -30,29 +30,33 @@
...
-        "@tiptap/extensions": "3.0.0-beta.5",
-        "@tiptap/pm": "3.0.0-beta.5",
-        "@tiptap/suggestion": "3.0.0-beta.5",
+        "@tiptap/core": "3.0.0-beta.14",
+        "@tiptap/extension-blockquote": "3.0.0-beta.14",
+        "@tiptap/extension-bold": "3.0.0-beta.14",
+        "@tiptap/extension-code": "3.0.0-beta.14",
+        "@tiptap/extension-code-block": "3.0.0-beta.14",
+        "@tiptap/extension-document": "3.0.0-beta.14",
+        "@tiptap/extension-dropcursor": "3.0.0-beta.14",
```
danharrin: Does this still need to be installed if you pull it from the `extensions` package anyway?

danharrin (changes requested):
Generally, all the other icons in Filament are quite round by default. Can we find replacements for the new icons so they look similar to the other rounded icons? Ideally all MIT so we don't need license text in the SVGs. Have a look at the 24px Tabler icons, I think I used some of these before: https://tablericons.com

packages/forms/src/Components/RichEditor/RichContentRenderer.php
awcodes: Moved this out to a dedicated class so it can be used with `app()`. If you prefer to used the default extension we can, but the default uses 'left' and 'right' which don't work with rtl and would mean more complicated integration. By using 'start' and 'end' support for rtl works without any complicated logic.

danharrin: You don't need a dedicated class for it to be used with `app()`, you can pass an array of constructor params in the second arg of `app()`

packages/forms/docs/10-rich-editor.md
awcodes: Added to the readme with a note about `alignJustify`. I would be surprised if it is actually used by more than 5% of content creators. IMO, it should not be on by default.

package.json
awcodes: Could have sworn I removed it. But it is removed now as well as removing the '^' for the new extensions so that everything is locked to the same version.

packages/forms/docs/10-rich-editor.md
danharrin: In that respect, I think we should get rid of lead, small, hr and highlight from the default toolbar as well, I think very few people use those too

awcodes: Ok. I think hr is pretty common though.

schmeits:
Nice work! 👏
Is submitting a PR to Filament itself the recommended approach for extending the default editor's functionality, or would it be better to release this as a separate plugin package?

So far, I’ve created the following plugins:

- clearFormatting
- table
- tableDelete
- tableColumnAddBefore
- tableColumnDelete
- tableColumnAddAfter
- tableRowAddBefore
- tableRowDelete
- tableRowAddAfter
- tableCellMerge
- tableCellSplit
- tableHeaderToggle

Let me know what the preferred direction is. Happy to adapt! 👍

awcodes:
Tables are planned for core, if you want to submit it as it's own PR, I'm happy to let you do so. 😅

schmeits:
> Tables are planned for core, if you want to submit it as it's own PR, I'm happy to let you do so. 😅

The most difficult part is finding the correct icons for all the buttons 🤣

schmeits:
Will give it a try in the morning 🚀

CharlieEtienne:
@awcodes Maybe that's too much opinionated, but I'd personally like it if highlight could be styled, like TipTap does in the demo:

```css
mark {
    background-color: #faf594;
    border-radius: .4rem;
    -webkit-box-decoration-break: clone;
    box-decoration-break: clone;
    padding: .1rem .3rem;
}
```

awcodes:
@CharlieEtienne you can style it any way you want with a custom theme or in any stylesheet. We are not applying any styling. It's using the default browser styling for mark tags.

lukas-frey:
Is there somewhere a roadmap of planned plugins for the core package?

danharrin:
@lukas-frey what do you want to propose?

danharrin (approved):

lukas-frey:
> @lukas-frey what do you want to propose?

I've been experimenting with the collaboration plugin(s), so I wanted to see if that is planned.

danharrin:
I think the infrastructure required to use such a feature would probably cause me not to want that in core

Issue #11967, opened 2024-03-20
Many errors occurs with DatePicker in spa() mode


agencetwogether:
### Package

filament/filament

### Package Version

v3.2.57

### Laravel Version

v11.0.7

### Livewire Version

v3.4.9

### PHP Version

PHP 8.2.16

### Problem description

I split my form in multi sections, in first one, Select and Datepicker are `->native(false)`and `->live()`to permit to see a third component (another Select) when values are set.
When I choose an option on first Select, many errors occur in browser console 
```
Alpine Expression Error: displayText is not defined...
Alpine Expression Error: focusedMonth is not defined...
Alpine Expression Error: months is not defined...
Alpine Expression Error: focusedYear is not defined...
Alpine Expression Error: dayLabels is not defined...
Alpine Expression Error: emptyDaysInFocusedMonth is not defined...
Alpine Expression Error: daysInFocusedMonth is not defined...
```
Same problem when I edit a record, when I change option, same errors appear

- If I remove `->native(false)` of all Datepicker, no errors but I need minDate, maxDate etc.
- If I remove `->spa()` in `AdminPanelProvider`, no errors
- If I add, below `->spa()`
```
->spaUrlExceptions(fn (): array => [
            url('/admin'),
            PostResource::getUrl(),
        ])
```
erros occur too

### Expected behavior

No error should appears

### Steps to reproduce

You can see this problem by cloning reproduction repository

### Reproduction repository

https://github.com/agencetwogether/spa-datepicker

ArtMin96:
This is the same problem that has been around for a long time. If you don't have dynamic interactions, such as `min`/`max` dates and `disabled` dates, you can extend and add `wire:ignore` to make the problem disappear. But this is only temporary and cannot be considered as a solution. You can look at this PR https://github.com/filamentphp/filament/pull/9204

dmitry-udod:
@agencetwogether  Hello! The main reason this problem happens is because, in SPA mode, date pickers are initialized only when they appear on the viewport. Then, when you use `live()` for any form element and set a value for the 'live' field, Livewire tries to refresh uninitialized variables such as focusedMonth, focusedYear, etc.. https://github.com/filamentphp/filament/blob/3.x/packages/forms/resources/views/components/date-time-picker.blade.php#L154
which initialized in `dateTimePickerFormComponent` function 
https://github.com/filamentphp/filament/blob/3.x/packages/forms/resources/views/components/date-time-picker.blade.php#L75

[...]

PR #16632, opened 2025-06-22
Fix DatePicker in SPA mode

binaryfire:
_EDIT: I added logic to generate a unique key each time the `minDate`, `maxDate` and `disabledDates` values change. This forces the field to be re-rendered and makes it work properly with dynamic data + `wire:ignore`_

Fixes https://github.com/filamentphp/filament/issues/11967

The problem was that `wire:ignore` was missing. I'm not sure if there was a reason for this, but I tested the `DatePicker` config below and all functionality (`closeOnDateSelection`, `disabledDates`, `minDate`, `maxDate`) worked fine in both SPA mode and normal mode:

```PHP
DatePicker::make('date')
    ->native(false)
    ->required()
    ->closeOnDateSelection()
    ->disabledDates(fn () => [now()->addDays(1)->toDateString()])
    ->minDate(now()->subDays(20))
    ->maxDate(now()->addDays(15))
    ->live(),
```

I've added the example from https://github.com/filamentphp/filament/issues/11967 to my reproduction repo if you want to test it with the fix:

1. Clone https://github.com/binaryfire/filament-bug-reproduction
2. Replace `date-time-picker.blade.php` with the one from this PR
3. `migrate:fresh --seed`
4. Go to Bookings and edit the first entry


binaryfire:
_EDIT: Got it working by regenerating the key when the date picker option values change_

Oh I just saw this: https://github.com/filamentphp/filament/pull/9204#issuecomment-1831601689

And yeah, this doesn't work when `wire:ignore` is added:
```
Toggle::make('disabled_dates')
    ->label('Disable Dates')
    ->live(),

DatePicker::make('date')
    ->native(false)
    ->required()
    ->closeOnDateSelection()
    ->disabledDates(function (Get $get) {
        if ($get('disabled_dates')) {
            return [now()->addDays(1)->toDateString()];
        }
        return [];
    })
    ->minDate(now()->subDays(20))
    ->maxDate(now()->addDays(15))
    ->live(),
```

But isn't it normal to run into these kinds of JS issues without `wire:ignore`? All the other views that wrap JS components use it and work properly. Perhaps there's a workaround to make dynamic min/max/disabled dates work with `wire:ignore`? Maybe using `$watch` to watch the state or something.

PR #16633, opened 2025-06-22
[v4] Feature/navigation group enum support

lukas-frey:
## Description

This PR implements all three points as discussed in #16575 

## Visual changes

None

## Functional changes

- [x] Add support for the `HasLabel` interface when building navigation groups from UnitEnums.
- [x] Add support for the `HasIcon` interface when building navigation groups from UnitEnums.
- [x] Add support for ordering when building navigation groups from UnitEnums based on their order of definition in the enum.


lukas-frey:
I have two things to discuss for the PR:


**PHP8.2 support:**
I just noticed a few checks failed, since I used a PHP8.3 only feature - accessing the enum cases dynamically - since UnitEnum's don't have a `tryFrom` method. However since Filament 4 supports PHP8.2+, this will not work.

To fix this issue, I suggest searching through the `[Enum]::cases()` method and returning the correct enum instead. This should work even on PHP8.2+.

EDIT: Fixed in commit: https://github.com/filamentphp/filament/pull/16633/commits/1bf6123c2de58547d0168ab1da184981e76853df

**Navigation group icon priority:**
With the current implementation, if a navigation group enum is added to a resource which also has a navigation icon set, filament will throw the default exception saying that only one of them can have an icon, not both.

My question is, if this behaviour is wanted or if we want to add some kind of additional check, for example in the Resource's `getNavigationIcon()` method and simply do a check here if navigation group is an instance of `HasIcon` and the icon returns a non-null value, we will simply return `null` for the `getNavigationIcon()`. So basically the Icon from the enum will have higher priority.

danharrin:
If someone has a `$navigationIcon` set, and then they set a nav group with an icon, they can just remove the icon from the resource, right? I don't want it to automatically silence the error, because then people will report that the icon is not in the sidebar and its a "bug"

lukas-frey:
> If someone has a `$navigationIcon` set, and then they set a nav group with an icon, they can just remove the icon from the resource, right? I don't want it to automatically silence the error, because then people will report that the icon is not in the sidebar and its a "bug"

Yes exactly, they can just remove it.

danharrin (changes requested):
Please fix the PHPStan issues

lukas-frey:
> Please fix the PHPStan issues

Done

Issue #13467, opened 2024-07-04
Google Chrome devtools issue with FileUpload component: "Incorrect use of <label for=FORM_ELEMENT>"


branivreyes:
### Package

filament/filament

### Package Version

v3.2.92

### Laravel Version

v11.14

### Livewire Version

v3.5.2

### PHP Version

PHP 8.3.6

### Problem description

Google Chrome devtools show an issue with an incorrect use of label's _for_ attribute when using the FileUpload form component. I also tested it with Spatie's media component and it shows the exact same issue. Here a screenshot:

![image](https://github.com/filamentphp/filament/assets/86134720/87e8dae4-dc9d-4bd1-ba7c-fb7c24a0165e)

### Expected behavior

The issue message shouldn't appear.

### Steps to reproduce

1. Clone the reproduction repository and do the normal Laravel setup process.
2. Go to /admin/login route and enter this credentials:
    **email address:** test@example.com
    **password:** password
3. Go to _Users_ resource and then click on _New user_ button.
4. Open Google Chrome devtools and go to _Issues_ tab:
![image](https://github.com/filamentphp/filament/assets/86134720/ef9dc501-ec11-400a-9389-57d861facdb8)

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/branivreyes/filament-issue

danharrin:
Can you explain why the current usage is incorrect? Is it because Filepond is not a true file upload element?

alex-elivate:
The issue is the field wrapper generates the label for FileUpload's div container, which of course is not a form input. Would you be open to reviewing a PR implementing ARIA labeling and  role="group" only for the FileUpload while keeping existing 'for' attributes for other components? 
https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA16
(Example 2 - Labelling a slider)

alex-elivate:
The thing that concerns me with this technique is: 

> Clicking on a label focuses the associated form field. This does not occur with aria-labelledby. If this behaviour is required then use label or implement this functionality using scripting.

I don't know if this is needed with assisted technology, but doesn't FilePond handle these interactions?

danharrin:
Yes please!

PR #16639, opened 2025-06-22
Fix accessibility issue

alex-elivate:
## Description
https://github.com/filamentphp/filament/issues/13467
File upload components were causing accessibility warnings due to label associations. The issue was the field wrapper was adding “label for” attributes for a div, and of course, div is not an input.

## Visual changes
Screenshot 2025-06-22 at 3 56 18 AM

## Functional changes
This PR implements ARIA labeling and role="group" only for the FileUpload while keeping existing 'for' attributes for other components using the technique from https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA16 example 2.
It also substitutes the label with a div. This can be simplified by reusing the label component If you don’t mind a blue (upgrade from red?) accessibility notice, displaying:  

> A label isn't associated with a form field.
> To fix this issue, nest the input in the label or provide a for attribute on the label that matches a form field id.


danharrin (changes requested):
Instead of this being a file-upload specific override, can we just add an easy config option to `field-wrapper` that can switch the attributes/element we use for the label? I don't think we need a new Blade view or anything

alex-elivate:
The File upload blade component now passes renderAsDiv=true to the field wrapper. I should have thought of this, too focused on the fix.

danharrin (changes requested):
In other places we do a `tag` attribute (for the `label` component). In this case it would default to `label`, but you could then change it to `div`.

On the `field-wrapper`, I would name the new attribute `label-tag`

alex-elivate:
Thanks yet again for the feedback

alex-elivate:
Repeaters produce the same accessibility warnings.  Do you want `:label-tag="'div'"` added to the repeater components as well? I did test the repeaters and the change will clear the errors.

packages/forms/resources/views/components/field-wrapper/index.blade.php
```diff
@@ -58,7 +59,14 @@
     }}
 >
     @if ($label && $labelSrOnly)
-        <label for="{{ $id }}" class="sr-only">
+        <label
+            @if ($labelTag === 'label')
+                :for="$id"
+            @else
+                :id="$id . '-label'"
+            @endif
```
danharrin: I don't think these bindings are correct, `<label>` is not a Blade component so `:` should not be used before the attributes, and any PHP needs to be echoed

danharrin (changes requested):

packages/forms/resources/views/components/field-wrapper/index.blade.php
alex-elivate: You're absolutely correct.  I created a test and it fails.  How embarrassing.  Working on this now.
```
FileUpload::make('avatar')
    ->label('Avatar (Hidden Label)')
    ->hiddenLabel()
```

alex-elivate: This is the working output for sr-only class labels.
```
<div id="data.avatar-label" class="sr-only">
            Avatar (Hidden Label)
        </div>

<label for="data.nickname" class="sr-only">
            Nickname (Hidden Label)
        </label>
```

alex-elivate: image

danharrin (approved):

danharrin:
Thanks. I think we will need the same for these components, if you are interested in PRing:

- Builder
- Checkbox list
- Code editor (v4 only)
- Color picker
- Date time picker (non-native only)
- Key-value
- Livewire field (v4 only)
- Modal table select (v4 only)
- One time code input needs an ID to make the label map properly, but the label doesn't need to be a `<div>` like this (v4 only)
- Radio
- Repeater, simple repeater, table repeater (v4 only)
- Rich editor
- Select (non-native only)
- Slider (v4 only)
- Table select (v4 only)
- Tags input (v4 only)
- Toggle probably needs an ID

PR #16651, opened 2025-06-23
fix: Hidden summaries

### LINKED ISSUE #15919: Performance issue + view bug on invisible summaries
### Package

filament/filament

### Package Version

v3.3.4

### Laravel Version

v12.3.0

### Livewire Version

v3.6.2

### PHP Version

8.3.19

### Problem description

When hiding all summaries of a table, there is still a performance impact on the page which is weird. Plus the summary is still shown (just without the values). 

Just for the records: The usecase of hiding summaries (at all) is to use an `Action` to toggle them. This is usefull on large tables to improve performance or when having multiple summarizers to make the table cleaner for most users. 

I think this bug was introduced in PR https://github.com/filamentphp/filament/pull/15594

### Expected behavior

1. When using `->visible(false)` on a summary, the query for the summary shouldn't be used.
2. When using `->visible(false)`on ALL summaries, there should be no summary at all shown.

### Steps to reproduce

See repo url and use the `issue/table-summary` branch. This is just a fork of the filament demo repo. All changes are in the `OrderResource` class. To keep it simple, I only set the summarizer to `hidden`, without the Action an toggling logic.

When adding `->visible(false)` to both summaries, this is how it looked like in when using Filament (for example) `v3.2.123`:

Image

After upgrade to `v3.3.5` this is how it looks like:

Image

https://github.com/bernhardh/filamentdemo

### Relevant log output

```shell

```

danharrin:
Fixes #15919

### DIFF
--- a/packages/tables/resources/views/components/summary/index.blade.php
+++ b/packages/tables/resources/views/components/summary/index.blade.php
@@ -25,6 +25,9 @@
     }
 
     $hasPageSummary = (! $groupsOnly) && $records instanceof \Illuminate\Contracts\Pagination\Paginator && $records->hasPages();
+
+    $pageTableSummaryQuery = $hasPageSummary ? $this->getPageTableSummaryQuery() : null;
+    $allTableSummaryQuery = $this->getAllTableSummaryQuery();
 @endphp
 
 @if ($hasPageSummary)
@@ -54,7 +57,7 @@
                         $alignment = filled($alignment) ? (Alignment::tryFrom($alignment) ?? $alignment) : null;
                     }
 
-                    $hasColumnHeaderLabel = (! $placeholderColumns) || $column->hasSummary();
+                    $hasColumnHeaderLabel = (! $placeholderColumns) || ($pageTableSummaryQuery && $column->hasSummary($pageTableSummaryQuery)) || $column->hasSummary($allTableSummaryQuery);
                 @endphp
 
                 <x-filament-tables::summary.header-cell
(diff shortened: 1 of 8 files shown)


packages/tables/resources/views/components/summary/index.blade.php
```diff
@@ -25,6 +25,9 @@
     }
 
     $hasPageSummary = (! $groupsOnly) && $records instanceof \Illuminate\Contracts\Pagination\Paginator && $records->hasPages();
+
+    $pageTableSummaryQuery = $hasPageSummary ? $this->getPageTableSummaryQuery() : null;
```
maartenpaauw, line 29: @danharrin I think this line of code breaks tables with custom data. Because when using custom data, the underlying `getFilteredTableQuery` method return `null` and result in a type error.

packages/tables/resources/views/index.blade.php
```diff
@@ -24,7 +24,7 @@
     $filterIndicators = $getFilterIndicators();
     $hasColumnGroups = $hasColumnGroups();
     $hasColumnsLayout = $hasColumnsLayout();
-    $hasSummary = $hasSummary();
+    $hasSummary = $hasSummary($this->getAllTableSummaryQuery());
```
maartenpaauw, line 27: @danharrin Same for this line, the `getAllTableSummaryQuery ` method is trying to return null when using custom data.

danharrin:
#16671

PR #16667, opened 2025-06-24
Adding ability to specify the columns number in the field set of the export action.

ht3aa:
## Description

This feature will add an ability to specify the columns of the field set in the export action as well specify wither it should be inline labeled or not.

## Visual changes

Before:
Screenshot 2025-06-24 at 8 07 46 AM

After:
Screenshot 2025-06-24 at 8 08 29 AM

## Functional changes

### DIFF
--- a/packages/actions/docs/12-export.md
+++ b/packages/actions/docs/12-export.md
@@ -78,25 +78,6 @@ public function table(Table $table): Table
 
 The ["exporter" class needs to be created](#creating-an-exporter) to tell Filament how to export each row.
 
-## Customizing the column selection form layout
-
-When the column mapping feature is enabled (which is the default), users will see a form that allows them to select which columns to export and customize their labels. You can control the layout of this form using the `columns()` method.
-
-### Setting the number of columns
-
-By default, the column selection form displays in a single column layout. You can change this using the `columns()` method:
-
-```php
-use App\Filament\Exports\ProductExporter;
-use Filament\Actions\ExportAction;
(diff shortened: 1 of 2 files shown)


danharrin (changes requested):
Like with the "filters columns" feature, I think the modal width should be automatically adjusted based on the number of columns

Maybe `inlineLabel()` should be turned off as soon as there is more than one column, too. What change does the `inlineLabel()` make in this case, can you send a screenshot before and after for 1 column and 3 columns, with and without an inline label?

ht3aa:
@danharrin Thanks for viewing. I will work on your request changes.  



# inlineLabel()
## for 3 columns

This is when it enabled
Screenshot 2025-06-26 at 9 16 44 PM


This when it is disabled
Screenshot 2025-06-26 at 9 17 50 PM

## for 1 column
This is when it enabled
Screenshot 2025-06-26 at 10 15 13 PM

This when it is disabled
Screenshot 2025-06-26 at 10 16 29 PM


some ui changes is happening.

ht3aa:
I commit the changes. hope it meets your requirements.

danharrin:
Please completely remove `inlineLabel()` from the action and any configuration for it, I don't think we need it anymore in v4. It seems to look different in v3.

ht3aa:
I removed the `inlineLabel()` and all its configuration that I have made. please check it again

danharrin (approved):

PR #16676, opened 2025-06-24
[4.x] Feat: Rich editor table extensions

schmeits:
## Description

Adds the following extensions to the Rich Editor:

- Table
- Remove Table 
- Add Column Before
- Add Column After
- Remove Column
- Add Row Before
- Add Row After
- Delete Row
- Merge Cells
- Split Cells
- Toggle Header Row

## Visual changes

image

## Functional changes


danharrin:
I don't think these should be toolbar buttons, I think there should be some sort of "bubble menu" that opens above the cursor when you are inside a table, that contains these options. Otherwise the toolbar is very overwhelming.

If that is too large of a request, please close this PR and it will be implemented in the future when we have time to work it out

schmeits:
> I don't think these should be toolbar buttons, I think there should be some sort of "bubble menu" that opens above the cursor when you are inside a table, that contains these options. Otherwise the toolbar is very overwhelming.
> 
> If that is too large of a request, please close this PR and it will be implemented in the future when we have time to work it out

Hi Dan, I will check if I can manage to move them into a bubble menu. I'll let you know if I can get it to work ;)

schmeits:
You can now use Bubblemenu's using the following syntax

```PHP
Forms\Components\RichEditor::make('bio')
                    ->toolbarButtons([
                        ['bold', 'italic', 'underline', 'strike', 'subscript', 'superscript', 'link'],
                        ['h2', 'h3'],
                        ['blockquote', 'codeBlock', 'bulletList', 'orderedList'],
                        ['table' => [
                            'tableRemove',
                            'tableColumnAddBefore',
                            'tableColumnDelete',
                            'tableColumnAddAfter',
                            'tableRowAddBefore',
                            'tableRowDelete',
                            'tableRowAddAfter',
                            'tableHeaderToggle',
                        ]]]
                    )
```

All the array values are converted to Bubblemenu items

image

you could even do this

```PHP
Forms\Components\RichEditor::make('bio')
                    ->toolbarButtons([
                        ['bold' => ['italic', 'underline'], 'italic', 'underline', 'strike', 'subscript', 'superscript', 'link'],
```

image

danharrin (commented):
Can you please work out an API for floating toolbars that does not use the main toolbars API? Like `->floatingToolbars(['table' => [.......]])`

A user could then register a floating toolbar with something like `->floatingToolbars(['text' => [.......]])` which would be active whenever the user selects text, instead of needing to tie it to "bold" for example

I consider these separate toolbars from the main toolbar so I think it makes sense that their definition is separate

schmeits:
> Can you please work out an API for floating toolbars that does not use the main toolbars API? Like `->floatingToolbars(['table' => [.......]])`
> 
> A user could then register a floating toolbar with something like `->floatingToolbars(['text' => [.......]])` which would be active whenever the user selects text, instead of needing to tie it to "bold" for example
> 
> I consider these separate toolbars from the main toolbar so I think it makes sense that their definition is separate

Yes think that's the way to go... will give it another go ;)

schmeits:
New suggestion:

```PHP
              Forms\Components\RichEditor::make('bio')
                    ->toolbarButtons([
                        ['bold', 'italic', 'underline', 'strike', 'subscript', 'superscript', 'link'],
                        ['h2', 'h3'],
                        ['table'],
                        ['blockquote', 'codeBlock', 'bulletList', 'orderedList'],
                    ])
                    ->floatingToolbars([
                        'heading' => [
                            'h1', 'h2', 'h3',
                        ],
                        'paragraph' => [
                            'bold', 'italic', 'underline', 'strike', 'subscript', 'superscript',
                        ],
                        'table' => [
                            'tableRemove',
                            'tableColumnAddBefore', 'tableColumnDelete', 'tableColumnAddAfter',
                            'tableRowAddBefore', 'tableRowDelete', 'tableRowAddAfter',
                            'tableCellMerge', 'tableCellSplit',
                            'tableHeaderToggle',
                        ]
                    ])
```

image

image

image

schmeits:
The key is the node to trigger the floating menu

schmeits:
Maybe we can fix the dropdown menu's like the ones in https://fluxui.dev/components/editor if we can come up with a nice API for submenu items... would be nice for textAlign and Table functions

danharrin:
First thing I notice is that when there is a paragraph floating menu, it is open when the page loads even if the editor is out of focus

Screenshot 2025-06-26 at 13 38 53

There also seems to be overlap/z-index issues with the menu in various places

Screenshot 2025-06-26 at 13 39 53

I think we need to add some tooltips to the main toolbar buttons AND floating toolbar buttons, thoughts? Since they are becoming more complex

schmeits:
what kind of control is that in front of the floating menu?
image

schmeits:
I find the bubble menu for paragraphs a bit too intrusive.

danharrin:
> what kind of control is that in front of the floating menu?

That is a navigation item in the sidebar

schmeits:
> > what kind of control is that in front of the floating menu?
> 
> That is a navigation item in the sidebar

will reproduce and fix

schmeits:
> First thing I notice is that when there is a paragraph floating menu, it is open when the page loads even if the editor is out of focus
> 
> 
> 
> Screenshot 2025-06-26 at 13 38 53
> 
> 
> 
> There also seems to be overlap/z-index issues with the menu in various places
> 
> 
> 
> Screenshot 2025-06-26 at 13 39 53
> 
> 
> 
> I think we need to add some tooltips to the main toolbar buttons AND floating toolbar buttons, thoughts? Since they are becoming more complex

Both issues fixed

danharrin:
Can you add tooltips for all toolbar buttons please?

Also, I think it is probably worth moving the floating toolbars to BELOW the cursor instead of above, so they will never overlap the main toolbar

schmeits:
> Can you add tooltips for all toolbar buttons please?
> 
> Also, I think it is probably worth moving the floating toolbars to BELOW the cursor instead of above, so they will never overlap the main toolbar

ok sure... are tooltips used anywhere else so I can check out the code?

danharrin:
Yeah you can search `x-tooltip` in the codebase

schmeits:
Fixed:
- Align floating toolbar bottom
- Added x-tooltip to all the RichEditorTools (removed the title)

image

danharrin (approved):

danharrin:
Thanks!

schmeits:
Thank you! Learned a lot from your changes just now... next time I'll try to come closer to your coding style ;)

PR #16686, opened 2025-06-24
Fix table group checkboxes being treated as record checkboxes 

dvarilek:
## Description
This PR removes the `fi-ta-record-checkbox` class from all group checkboxes in table. Having this class applied caused two issues.

When grouping was applied, the group checkboxes were treated as record checkboxes in `table.js` functions ( `getRecordsOnPage()` and `handleCheckboxClick()`) which caused them to be picked up by `this.$root?.getElementsByClassName` and their values to be injected into the `selectedRecords` property. This caused the selection indicator to display an incorrect amount of records (+1) as seen on the videos attached below. 

### Selecting records on page when grouping is applied
https://github.com/user-attachments/assets/d6d81889-8dcb-450b-ab34-b6f43c5cd936

### Selecting records using shift click when grouping is applied
https://github.com/user-attachments/assets/5043b073-3534-47c7-89e9-08afc0f73f00

On both videos the values of checkboxes are console logged. It shows multiple 'ons' for each group because it just logs the keys, not the `selectedRecords` set.

## Visual changes
The `fi-ta-record-checkbox` class also made the groups taller than in 3.x, basically having the same height as the record row. I'm actually not sure if this was a deliberate design choice. Regardless, removing the class made them look like in v3. If the change was actually intentional I would make changes.

### v3 (original and current look after PR)
v3group
### v4 (before PR)
v4group

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -876,7 +876,7 @@
                                                 wire:key="{{ $this->getId() }}.table.bulk_select_group.checkbox.{{ $page }}"
                                                 wire:loading.attr="disabled"
                                                 wire:target="{{ implode(',', \Filament\Tables\Table::LOADING_TARGETS) }}"
-                                                class="fi-ta-record-checkbox fi-ta-group-checkbox fi-checkbox-input"
+                                                class="fi-ta-group-checkbox fi-checkbox-input"
                                             />
                                         @endif
 
@@ -1541,7 +1541,7 @@
                                                                     wire:key="{{ $this->getId() }}.table.bulk_select_group.checkbox.{{ $page }}"
                                                                     wire:loading.attr="disabled"
                                                                     wire:target="{{ implode(',', \Filament\Tables\Table::LOADING_TARGETS) }}"
-                                                                    class="fi-ta-record-checkbox fi-ta-group-checkbox fi-checkbox-input"
+                                                                    class="fi-ta-group-checkbox fi-checkbox-input"
                                                                 />
                                                             @endif
                                                         </td>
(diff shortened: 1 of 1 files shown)


danharrin (changes requested):
Please review all references to `fi-ta-record-checkbox` in the codebase and ensure that any CSS rules / JS functions still apply to `fi-ta-group-checkbox` if they need to be

danharrin:
If you have already done that, please confirm so

dvarilek:
Yeah, I checked and there are no JS functions that need to apply to `fi-ta-group-checkbox` elements. The ones that did before really shouldn't because of the problem with `selectedRecords`.

As for the CSS rules, removing the `fi-ta-record-checkbox` means that `.fi-ta-cell:has(.fi-ta-record-checkbox)` from `tables\resources\css\cell.css` no longer applies to the group checkbox containers. This isn't really a problem because there is `fi-ta-cell.fi-ta-group-selection-cell` that applies basically the same styles. The only difference is that the `:has()` selector one includes `py-4`, and `.fi-ta-group-selection-cell` does not. This causes the group to be shorter by that padding and look the same like in v3, which I'm not sure if that's the intended look for v4. Adding `py-4` to `fi-ta-cell.fi-ta-group-selection-cell` would make the group look the same like before this commit. The visual difference is shown in my screenshots above.

Unrelated to this change, but while testing, I noticed that when layout components are rendered in the table, grouping becomes broken even when the checkbox has `fi-ta-record-checkbox`. I tried fixing it and making another PR but I wasn't sure how to, so I'll make an issue for that
Screenshot 2025-06-26 140210

danharrin:
Thanks for the detail!

danharrin (approved):

PR #16693, opened 2025-06-25
[4.x] Feat: Configurable default view and size for record actions

damiandejong:
## Description

Currently, the default action view and size are fixed for record actions. I would like them to be configurable so that by using `Table::configureUsing()`, all record actions can be set to another view and/or size by default (in my case I want them to be icon buttons). 

## Visual changes

N.A.

## Functional changes


danharrin:
Hmm, instead of this, can you open up a PR that allows you to pass `$table->configureUngroupedRecordActionsUsing()`, where you can pass a `fn (Action $action) => ...` to customize whatever you want, including being able to use `$action->iconButton()` instead of needing to fetch the view string. If this is defined, the function should run AFTER the current defaults are applied, to ensure that they can be overwritten.

PR #16697, opened 2025-06-25
[v4] Docs: update installation instructions and component usage

leandrocfe:
### Update: Filament usage outside the panel

Improved installation instructions and usage guidelines for using Filament components outside the panel.

The following components have updated documentation:

- `Action`
- `Form`
- `Infolist`
- `Notification`
- `Schema`
- `Table`
- `Widget`

## Functional changes

### DIFF
--- a/docs/12-components/02-table.md
+++ b/docs/12-components/02-table.md
@@ -49,7 +49,7 @@ There are 3 tasks when adding a table to a Livewire component class:
 namespace App\Livewire;
 
 use App\Models\Shop\Product;
-use Filament\Schemas\Concerns\InteractsWithSchemas;  
+use Filament\Schemas\Concerns\InteractsWithSchemas;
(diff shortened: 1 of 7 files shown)


danharrin (changes requested):
Please review the actual dependencies of the packages, for example most of these don't require the `notifications` package

docs/12-components/02-table.md
danharrin, line 1: I think tables actually now require the HasActions interface and trait to work since we unified actions

danharrin (changes requested):

docs/12-components/02-table.md
leandrocfe: Ahh yes. I believe I tested it without the `actions` package at some point and it worked. 
Anyway, I'll include it. Thanks!

danharrin (approved):

PR #16705, opened 2025-06-26
[4.x] Use Email and Password Default Rule

lloricode:
## Functional changes


danharrin:
Hi, sorry but I think this is too opinionated. I can think of many places in an app where you would want an email/password text input to not follow these validation rules, which are meant for user accounts usually.

PR #16706, opened 2025-06-26
feat: add `->trim()` method to `TextInput`, `Textarea` and `TagsInput`

ralphjsmit:
This PR adds a `->trim()` method to several text-based components. It also extends the `->stripCharacters()` implementation to work on items inside a `TagsInput`.


danharrin:
Can you please retarget this at 4.x?

packages/schemas/src/Components/Concerns/HasState.php
```diff
@@ -68,6 +68,8 @@ trait HasState
      * @var array<string>
      */
     protected array $cachedStripCharacters;
+	
+	protected bool | Closure $isTrimmed = false;
```
ralphjsmit: Would you also want to cache the result of `isTrimmed`, similar to `cachedStripCharacters`?

ralphjsmit:
Done!

packages/schemas/src/Components/Concerns/HasState.php
```diff
@@ -496,7 +504,13 @@ public function mutateDehydratedState(mixed $state): mixed
 
     public function mutateStateForValidation(mixed $state): mixed
     {
-        $state = $this->stripCharactersFromState($state);
+        if (is_array($state)) {
+            $state = array_map($this->stripCharactersFromState(...), $state);
+            $state = array_map($this->trimState(...), $state);
```
danharrin: Please combine these maps so the state doesnt need to be looped twice

packages/schemas/src/Components/Concerns/HasState.php
danharrin: Yeah I think that would be good

danharrin (changes requested):
Is it also worth documenting the method for the other two supported components?

To be honest, I am a little hesitant to add extra methods to `HasState`, and the existing `stripCharacters()` should probably be moved out of there into its own trait also. Otherwise people will expect this to work for every component, and some of them it doesn't make sense with (Section?). Thoughts?

ralphjsmit:
> To be honest, I am a little hesitant to add extra methods to HasState, and the existing stripCharacters() should probably be moved out of there into its own trait also. Otherwise people will expect this to work for every component, and some of them it doesn't make sense with (Section?). Thoughts?

Are you primarily concerned about the performance of `HasState` or more about the theoretical implication that people would expect this to work for every component? In essence it's not really changing the status quo, as `stripCharacters()` is also already "available" for every component. I'm fine to extract this into a trait, and then manually copy these few methods that call `stripCharactersFromState()` and `trimState()` into the 3 components that support them. OK?

danharrin:
I am worried about both performance and also the fact that methods are present that don't work. Extracting into traits and then applying those traits sounds good to me.

packages/forms/src/Components/TagsInput.php
```diff
@@ -170,4 +174,48 @@ public function isReorderable(): bool
...
+    public function mutateDehydratedState(mixed $state): mixed
+    {
+        if (is_array($state)) {
+            $state = array_map(function (mixed $value) {
+                $value = $this->stripCharactersFromState($value);
+                $value = $this->trimState($value);
+
+                return $value;
+            }, $state);
+        } else {
```
ralphjsmit, line 187: Tried several versions to format this but couldn't find one that I really liked

CoringaWc:
It would be interesting to implement a trim on the columns of the importer classes so that the data imported from the columns of .csv files, which may for some reason come with unwanted spaces, are properly cleaned.

ralphjsmit:
@CoringaWc Not in the scope of this PR as this one is focused on form components, but you can always PR it yourself :)

danharrin:
AFAIK, importers already trim strings by default, so if that doesn't work please open an issue.

danharrin (changes requested):
TextInput and Textarea can be simplified as they never have array state. TagsInput can be simplified because it always has array state (or null I guess)

ralphjsmit:
@danharrin Lol, yes of course. Changed!

danharrin (approved):

danharrin:
Thanks

Issue #16243, opened 2025-05-11
Password validation only showing 1 password strength error at a time


binaryfire:
### Package

filament/filament

### Package Version

4-alpha6

### Laravel Version

12

### Livewire Version

3

### PHP Version

8.4

### Problem description

Laravel's password validation is configured to return all validation errors at once:

https://github.com/laravel/framework/blob/bcfa7c32acc17a7e49892efc5040fcaae3f207da/src/Illuminate/Validation/Rules/Password.php#L322

Showing all the failures each time is important for UX. Let's say you have this password strength config in a service provider:

```PHP
Password::defaults(function () {
    return Password::min(8)
                ->letters()
                ->numbers()
                ->symbols()
                ->mixedCase();
});
```

When a user tries to register with a password that doesn't meet these criteria, the expectation is that a message for every failed criteria is displayed. Currently Filament displays these errors one by one:

First try - length failure:
![Image](https://github.com/user-attachments/assets/a49e99a7-b144-4fbb-9196-15904250455d)

Second try - length corrected, mixed case failure:
![Image](https://github.com/user-attachments/assets/dd44657d-842f-4651-b96a-a006038e7b47)

Third try - mixed case corrected, symbol failure:
![Image](https://github.com/user-attachments/assets/d9ad4693-16cf-4303-a77f-58be0f74749e)

Fourth try - symbol corrected, number failure:
![Image](https://github.com/user-attachments/assets/800d34ee-8fd4-46c5-bca7-a6a1a127c7f2)

Also, each validation failure hits the rate limiter which adds to the UX issues:
![Image](https://github.com/user-attachments/assets/e16be3a8-c952-44a9-a3db-10893e27d526)

Showing password requirements as a helper text or a hint help a little, but displaying the full list of errors for the currently entered password is the most important visual feedback. I considered using a notification but I think that would feel strange. All other platforms display the list of errors under the password field. That's the UX users have been trained to expect.

### Expected behavior

All password field validation errors should be displayed at once.

### Steps to reproduce

1. Clone reproduction repo
2. Visit /register page and try using a password with just letters to start with.

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/binaryfire/filament4-password-validation

### Relevant log output

```shell

```

binaryfire:
This is how I'm working around it atm, in case it helps. Ideally the messages would be in an unordered list but validation error messages are escaped so I can't use HTML.

[...]

PR #16741, opened 2025-06-30
Allow showing multiple validation errors on fields + show all password errors

binaryfire:
Closes https://github.com/filamentphp/filament/issues/16243

Adds a `->showAllValidationMessages()` to fields. When true, multiple validation messages are shown in a list. This has been enabled for the password input on the `Register` page:

image


danharrin (changes requested):
Can we change the implementation so that it doesn't require us to enable the "html messages" feature? In case the messages aren't safe. Maybe just change how the messages are output in the Blade instead of constructing a new error message string.

I think `fused-group.blade.php`, which handles its own error messages, also needs to be updated.

binaryfire:
@danharrin Done. Fused group with a field using the `Password` rule:

- Without `->showAllValidationMessages()`:
image

- With `->showAllValidationMessages()`:
image

PR #16745, opened 2025-06-30
[v4 feature]: add option to limit the number of selectable records in a table

dvarilek:
## Description

This PR introduces the ability to limit the number of selectable records in a table without relying on `checkIfRecordIsSelectableUsing()` and selection property entaglement as proposed in #16558.

## Functional changes

By setting `maxSelectableRecords()`, the developer can limit how many records users can select in total. Checkboxes become disabled when toggling them on would exceed this limit. This applies to and works with all checkboxes within a table (record checkboxes, group checkboxes and page checkboxes). The 'select all' action gets conditionally hidden. Checkboxes can always be toggled off if they are on.

Since one of the primary reasons behind this feature was to imporve the experience with the newly added TableSelect 
components, I also added the `CanLimitItemsLength` trait to them. This means that form validation rules `minItems()` and `maxItems()` can now be used in these just like in a regular Select field. `maxItems` gets mapped to `masSelectableRecords`.

## Visual changes
Video showcasing the feature:

https://github.com/user-attachments/assets/fe9b1b91-65ed-4740-a357-20ff251c5aed

## Details

Apart from a simple developer API, two new functions were added to the table Alpine component.
The `areRecordsToggleable()`, as the name implies, is responsible for determining whether the provided records would have their checkboxes disabled. Since the condition for displaying 'select all' action became more complex, I was forced to move the logic into a standalone function - `canSelectAllRecords()` (not particularly happy about making this part of table.js but couldn't think of a better way).

When `maxSelectableRecords()` is null (default), it's expected that there is no limit.

Since it's possible to select multiple records with a shift click and the checkboxes can't just simply be disabled for that, I had to slightly modify the function that handles multi-selection.

One thing that might be a bit confusing for people (and they might even assume it's some bug) is what happens on lines 27-29 in `index.blade.php`, I personally think it's expected behaviour and it simplifies things quite a bit. (when `maxSelectableRecords` is 1 `canSelectMultipleRecords` is automatically false and vice versa)

## Changes


packages/tables/resources/js/components/table.js
```diff
@@ -294,12 +343,17 @@ export default ({
             let values = []
 
             for (let i = range[0]; i <= range[1]; i++) {
-                checkboxes[i].checked = checkbox.checked
```
danharrin, line 297: Doesn't removing this line break Shift bulk selection? What is replacing it?

danharrin (changes requested):
I think there needs to be some sort of backend work that limits the number of selected records in final collection, in case a malicious user bypasses the limit to crash a process or similar

packages/tables/resources/js/components/table.js
dvarilek: Pretty sure that Alpine handles the state update automatically with`x-bind:checked` that's present on record checkboxes. I tested it and it works fine without it, plus it makes things a bit easier there

dvarilek:
> I think there needs to be some sort of backend work that limits the number of selected records in final collection, in case a malicious user bypasses the limit to crash a process or similar

Yeah, I'm thinking that both `selectedTableRecords` and `deselectedTableRecords` could be made #[Locked] but that could be a breaking change for some users, though it would probably simplify things. I will look into it

danharrin:
I don't think we need to do that, I think the limit needs to be in the DB queries that fetch the selected records

danharrin (changes requested):
For the non-query parts this is good, but the reason I suggested adding a limit to the query is that a malicious user could potentially crash the request by loading too many records into memory, so I think that still needs to happen, right?

dvarilek:
I'm pretty sure that the records are not loaded into memory at all before the validation happens here. The builder / relationship instance just gets scoped, or am I missing something?

danharrin:
Instead of running an additional `count()` query and throwing an exception, I think a query limit would be better. No need to throw an exception if a user is acting maliciously.

dvarilek:
Yeah, that makes sense. I also made sure that this limit / exception is enforced when the table is set to single record selection, and I additionaly fixed an issue when toggling in that mode.

packages/tables/resources/views/index.blade.php
danharrin, line 1: Can you ensure that the new `@else` conditions added to the checkboxes are instead `@elseif`, where it only outputs the `x-bind:disabled` if there is a limit on the number of selectable records, or `$canSelectMultipleRecords` is false?

danharrin (changes requested):
Maybe `canSelectMultipleRecords()` etc can be refactored to just use `maxSelectableRecords(1)`, would that make things cleaner? Don't worry about breaking changes to those methods since they are new in v4.

packages/tables/resources/views/index.blade.php
dvarilek: Done, I didn't check for `$canSelectMultipleRecords ` everywhere where that was done in conditions before

dvarilek:
> Maybe `canSelectMultipleRecords()` etc can be refactored to just use `maxSelectableRecords(1)`, would that make things cleaner? Don't worry about breaking changes to those methods since they are new in v4.

To be honest, I think the `canSelectMultipleRecords` property was pretty redundant in the trait with `maxSelectableRecords` being present, so I removed it with its getter. Made the `multipleRecordsSelectable` into just a 'helper', pretty sure its cleaner and more obvious what's happening this way but I might be biased

danharrin:
I have tested this. It mostly works great, but I have found an edge case that breaks it:

This is a resource with `maxSelectableRecords(5)`. Groups can span multiple pages, but currently when a group checkbox is deciding whether or not it should be disabled, it only checks the number of records on the current page.

https://github.com/user-attachments/assets/de292eae-076c-403a-82b4-ac499576449c

In this example, the group checkbox should be disabled, since there is more than one record in the group, even though the records are on a different page.

I think we will need a solution similar to the `allSelectableRecordsCount` element, but for each group's total count. Maybe this could be stored in a `data-` attr on the group checkbox or something, then read from JS. If extra queries are required, these queries should only run if this feature is enabled, to not degrade the performance of apps that dont use the feature.

dvarilek:
Since the only way to get the total number of records in a group is by `$wire.getGroupedSelectableTableRecordKeys()` and counting the result, maybe storing the keys in the data attribute instead of the count would be better? At least this way it would be possible to have them cached and available for later when a group toggle is performed, this way there would be no need for an additional request over the wire when group toggling (`toggleSelectRecordsInGroup` table.js), making it slightly smoother.

danharrin (changes requested):
This looks good!

One thing that I did notice though, and it's definitely not a result of this PR, but probably should be fixed at the same time, is that when you select all records in a group on one page, and there are group records on another page too, they also get selected, which is correct. But when you deselect some on one page, the other page still thinks all in the group are selected.

Maybe we should *always* load the group keys into the HTML and use these to select/deselect/check if all group records are selected or not- instead of only when the `maxSelectableRecords()` feature is enabled.

What do you think about this?

dvarilek:
It make sense actually yes, but I noticed that currently, when fetching the group keys, models are being loaded into memory for some reason and I'm not quite sure why, so that would have to be resolved first I think

danharrin:
Good catch! Yeah I think that could be refactored, if records don't need to be checked if they are selectable or not.

packages/tables/src/Concerns/HasBulkActions.php
dvarilek, line 1: Turns out `pluck` hydrates the models under certain circumstances, one of them being when there is a cast on the attribute and the primary key has `int` by default

danharrin: Oh dear, that is a huge finding actually, I must have done this many many times in core.

dvarilek:
I’ve made it so that group checkboxes now affect all the record checkboxes in the group, even if it spans multiple pages. Also, whether the group checkbox is checked or not is now determined by all the record checkboxes, which I think is correct here.

I was also thinking about maybe adding some indicator that displays how many records are in a group in total (if a group spans multiple pages, users might assume that only records on the current page belong to that given group). Although I’m not sure where this count would be placed, maybe at the end of the group? Or maybe just a tooltip that’s displayed on hover? What do you think?

danharrin (approved):

danharrin:
I really *really* appreciate how patient you have been with my change requests @dvarilek. I am now happy with how this is behaving. Thank you so much!

PR #16752, opened 2025-07-01
Fix Postgres orderByRaw exports

### LINKED ISSUE #16184: Postgres-backed app with table columns using raw sorts cause prebuilt Export Action to crash
### Package

filament/filament

### Package Version

v3.3.14

### Laravel Version

v12.11.1

### Livewire Version

v3.6.3

### PHP Version

PHP 8.3.19

### Problem description

When:
1. An app is using postgres
2. A column, in a table, in that app, is `->sortable()`, with a closure
3. That closure uses `->orderByRaw()` on the query
4. A user attempts an export

The `PrepareCsvExport` queued job crashes, with the exception `Undefined array key "column"` [on this line](https://github.com/filamentphp/filament/blob/811ee353c806808d3a1b66c33eb5f955a72c14ee/packages/actions/src/Exports/Jobs/PrepareCsvExport.php#L79).

It would appear that `PrepareCsvExport` is expecting the `$originalOrders` collection to be shaped like this:
```
array:1 [
  0 => array:2 [
    "column" => "name"
    "direction" => "asc"
  ]
]
```

while  the `->orders` from an `->orderByRaw()` are shaped like this:
```
array:1 [
  0 => array:2 [
    "type" => "Raw"
    "sql" => "name ASC NULLS FIRST"
  ]
]
```

Thus, when the job attempts to access a `column` key on the order entry, it cannot find one.

### Expected behavior

The queued job should complete, with the query ordered by raw SQL.

### Steps to reproduce

1. Point a Filament app at a postgres database
2. Create a Filament column, in any Filament table, with some variation of `->sortable(fn (Builder $query) => $query->orderByRaw())` (Example repo uses `name DESC NULLS LAST`)
3. Create an exporter
4. Add a prebuilt Exporter Action on that table, configured with that exporter
5. Attempt an export

The queued job will crash, with the aforementioned exception.

https://github.com/V13Axel/filament-table-raw-order-exporter-bug

### Relevant log output

```shell

```

binaryfire:
Fixes https://github.com/filamentphp/filament/issues/16184

If you're happy with this, I'll port the changes to v4 as well.

### DIFF
--- a/packages/actions/src/Exports/Jobs/PrepareCsvExport.php
+++ b/packages/actions/src/Exports/Jobs/PrepareCsvExport.php
@@ -69,25 +69,45 @@ class PrepareCsvExport implements ShouldQueue
 
         if ($databaseConnection->getDriverName() === 'pgsql') {
             $originalOrders = collect($query->getQuery()->orders)
-                ->reject(fn (array $order): bool => ($order['column'] ?? null) === $qualifiedKeyName)
-                ->unique('column');
+                ->reject(function (array $order) use ($qualifiedKeyName): bool {
+                    if (($order['type'] ?? null) === 'Raw') {
+                        return false;
+                    }
+
+                    return ($order['column'] ?? null) === $qualifiedKeyName;
+                })
+                ->unique(function (array $order) {
+                    if (($order['type'] ?? null) === 'Raw') {
+                        return 'raw:' . ($order['sql'] ?? '');
+                    }
+
+                    return 'column:' . ($order['column'] ?? '');
+                });
 
             /** @var array<string, mixed> $originalBindings */
             $originalBindings = $query->getRawBindings();
 
             if (! empty($originalOrders->all())) {
-                $query->reorder($originalOrders[0]['column'], $originalOrders[0]['direction']);
+                $firstOrder = $originalOrders->first();
+
+                if (($firstOrder['type'] ?? null) === 'Raw') {
+                    $query->reorder();
+                    $query->orderByRaw($firstOrder['sql']);
+                } else {
+                    $query->reorder($firstOrder['column'], $firstOrder['direction']);
+                }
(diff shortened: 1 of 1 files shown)


packages/actions/src/Exports/Jobs/PrepareCsvExport.php
```diff
@@ -69,25 +69,45 @@ public function handle(): void
...
             } else {
                 $query->reorder($qualifiedKeyName);
             }
 
             foreach ($originalOrders as $order) {
-                if (blank($order['column'] ?? null) || blank($order['direction'] ?? null)) {
-                    continue;
+                if (($order['type'] ?? null) === 'Raw') {
+                    $query->orderByRaw($order['sql']);
+                } elseif (! blank($order['column'] ?? null) && ! blank($order['direction'] ?? null)) {
```
```suggestion
                } elseif (filled($order['column'] ?? null) && filled($order['direction'] ?? null)) {
```

packages/actions/src/Exports/Jobs/PrepareCsvExport.php
```diff
@@ -69,25 +69,45 @@ public function handle(): void
...
-                ->reject(fn (array $order): bool => ($order['column'] ?? null) === $qualifiedKeyName)
-                ->unique('column');
+                ->reject(function (array $order) use ($qualifiedKeyName): bool {
+                    if (($order['type'] ?? null) === 'Raw') {
+                        return false;
+                    }
+
+                    return ($order['column'] ?? null) === $qualifiedKeyName;
+                })
+                ->unique(function (array $order) {
```
```suggestion
                ->unique(function (array $order): string {
```

danharrin (changes requested):

danharrin:
No need to port to v4, the v3 branch gets merged into v4 regularly.

danharrin (approved):

danharrin:
Thanks!

PR #16761, opened 2025-07-01
Add PHP8.4 to test workflows

dmlogic:
## Description

With reference to [discussion 15388](https://github.com/filamentphp/filament/discussions/15388), this PR adds v8.4 to the php matrix for both the `tests` and `phpstan` GitHub workflows.

No code changes were needed for tests to pass. Two ignores were needed for phpstan.

I don't think the docs require updating unless support for PHP8.4 should be explicit rather than implied.

## Visual changes

None

## Functional changes

### DIFF
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


danharrin:
Cheers

PR #16763, opened 2025-07-01
Feature/call hook macros

Perturbatio:
## Description

Adds the ability to define hooks using macros as per https://github.com/filamentphp/filament/discussions/16743.

e.g. a user wants to define an `afterFill` method on a package's page and doesn't want to extend or override the entire class in order to do so.

This allows a user to write:

```PHP
InternalPage::macro('afterFill', function(){
     // ...code here
});
```

## Visual changes

None

## Functional changes

- [ x ] `callHook` method on `BasePage` now checks if a macro was defined that matches the hook.


danharrin (commented):
Have you tested registering a macro on one type of page to ensure that it doesnt leak to other types? Since the Macroable trait is on the base `Component` class

For example, register a `beforeFill` macro, does it run on both view and edit pages?

Perturbatio:
> Have you tested registering a macro on one type of page to ensure that it doesnt leak to other types? Since the Macroable trait is on the base `Component` class
> 
> For example, register a `beforeFill` macro, does it run on both view and edit pages?

You're right, I hadn't considered this.  I'm not sure it's an easy thing to do without either making a new macroable trait for this purpose (which I don't like), or modifying the core Laravel Macroable trait to allow macros to be scoped to the declaring class (which I've prototyped the changes for and does seem to work).

That would require PRs to L10+ in order to support Filament 3+, I'll have a think

danharrin:
Filament has its own Macroable trait which does the same, and I don't think Laravel would merge such a breaking change in core. I don't think we can override the core Livewire macroable trait either, right? I think we're going to have to stop this unfortunately, and plugin users will continue to use traits to add hooks to pages.

Perturbatio:
I'm not sure the way I've prototyped it would be a breaking change at all since anything defined via a call to ::macro would function as normal, except where a scoped macro was defined.  However I need to ensure I've thought it through more first.

danharrin:
If a user was previously registering a macro for a subclass and using it on a sibling class by accident, it would be breaking

Perturbatio:
> If a user was previously registering a macro for a subclass and using it on a sibling class by accident, it would be breaking

I was more thinking of adding a `scopedMacro` method that is functionally similar but was explicitly declared as scoped.
My solution is to split scoped macros out from macros in the Laravel Macroable:

```PHP

    /**
     * Register a custom scoped macro, scoped macros are only
     * available to the class in which they are registered.
     *
     * @param  string  $name
     * @param  object|callable  $macro
     * @return void
     */
    public static function scopedMacro($name, $macro)
    {
        static::$scopedMacros[static::class][$name] = $macro;
    }
```

and in the call and callStatic methods there would be a check for a scoped macro which will take precedence over a macro so it would only be invoked *if* a user explicitly declares a scopedMacro on the class which the method was invoked from.

Looking at the `\Filament\Support\Concerns\Macroable` implementation, you've overridden the core functionality, which would likely break things, I agree.

It is relatively straightforward however to avoid this by doing:

```PHP
        if (static::hasScopedMacro($method)) {
            $macro = static::$scopedMacros[static::class][$method];
        } else {
            $macro = static::$macros[$method];
        }
```

danharrin:
I would rather add a `EditRecord::registerHook()` or something to Filament instead of hacking macros

Then multiple hooks that are registered can be run AS WELL AS user defined hooks which could be useful, especially if multiple plugins register the same hooks

danharrin:
If you are going to do that, please do so in v4 instead of v3 though

Issue #16730, opened 2025-06-29
v4 - afterStateUpdatedJs cause error in Table/Actions installed as Individual Component


ucosatti:
### Package

filament/tables

### Package Version

4.0.Beta[10]

### Laravel Version

12

### PHP Version

8.4

### Problem description

This magnificent new feature called afterStateUpdatedJs is causing me an error when I add it to a form, both inside a table and in an external action using a "slideover" modal. When I open the modal, the function executes the code perfectly, but then when I close the modal or try to run the function again, I get the error: "An action tried to resolve without a name."

<img width="1611" alt="Image" src="https://github.com/user-attachments/assets/f4e02f0f-c878-44d7-8925-2ca575e6ce54" />

### Expected behavior

keep working

### Steps to reproduce

1- clone, migrate, seed users.
2- At root, clic edit a record (change the name if you want. Its also happends without touch anything). clic Edit again, now error should be shown.

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/ucosatti/fila.git

### Relevant log output

```shell

```

ucosatti:
works perfect, now! thanks!

PR #16764, opened 2025-07-01
fix: hydrate action components with data

### LINKED ISSUE #16732: v4 - ActionNotResolvableException when modals contain CodeEditors
### Package

filament/filament

### Package Version

v4.0.0-beta10

### Laravel Version

v12.19.3

### Livewire Version

v3.6.3

### PHP Version

PHP 8.4.5

### Problem description

Firing Actions as modals containing one or more CodeEditors throws `Filament\Actions\Exceptions\ActionNotResolvableException` on **second** or more consecutive executions of the action.

Image

### Expected behavior

No exception should be thrown and the action should fire as normal.

### Steps to reproduce

I've replicated this on a custom page for minimal demonstration, but you can replicate this on resource listing with the 'new resource' and 'edit' buttons as well as long as they're set as modals.

Page Example:

```php
<?php

namespace App\Filament\Pages;

use Filament\Actions\Action;
use Filament\Forms\Components\CodeEditor;
use Filament\Pages\Page;
use Filament\Schemas\Schema;

class ExamplePage extends Page
{
    protected string $view = 'filament.pages.settings';

    public function form(Schema $schema): Schema
    {
        return $schema->components([
            TextInput::make('name')
                ->hintActions([
                    Action::make('sample_action')
                        ->schema([
                            CodeEditor::make('some_code')
                        ])
                ]),
        ]);
    }
}
```

Using the above example code:

- Open the page
- Click the 'Test' action (Defined as a hint for easier demonstration)
- Enter some sample code in the CodeEditor
- Hit the modal submit button, 
- After modal closed then click the 'Test' action again, you'll be greeted by the exception

The same can be reproduced with a resource with all the correct state paths defined. The reproduction repository contains a reproduceable example on the UserResource:

- Open the User listing page
- Click edit to open the edit modal
- Edit the user ensuring there's something in the CodeEditor field, submit
- Click the 'Create User' button, be greeted by the exception

https://github.com/stephenlake/filament-v4-codeeditor-modals-bug

### Relevant log output

```shell

```

antonzapevalov:
fix #16732, #16730

### DIFF
--- a/packages/actions/src/Concerns/HasData.php
+++ b/packages/actions/src/Concerns/HasData.php
@@ -98,7 +98,7 @@ trait HasData
      */
     public function getRawData(): array
     {
-        return $this->getLivewire()->mountedActions[$this->getNestingIndex()]['data'] ?? [];
+        return $this->getLivewire()->mountedActionsData[$this->getNestingIndex()]['data'] ?? [];
     }
 
     /**
--- a/packages/actions/src/Concerns/InteractsWithActions.php
+++ b/packages/actions/src/Concerns/InteractsWithActions.php
@@ -36,6 +36,11 @@ trait InteractsWithActions
      */
     public ?array $mountedActions = [];
 
+    /**
+     * @var array<array<string, mixed>> | null
+     */
+    public ?array $mountedActionsData = [];
+
     protected ?int $originallyMountedActionIndex = null;
 
     /**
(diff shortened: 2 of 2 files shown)


danharrin:
Hi, I don't think this is the right solution, any references to `mountedActionsData` should be updated to `mountedActions`, not the other way round. Unless there is something I am missing? Please provide a description for your PRs so I can understand them a bit more.

antonzapevalov:
If “statePath” matches a component property, in this case “mountedActions”, Livewire starts adding ‘data’ to “mountedActions” when the modal window is reopened. You can see this in the “updates” parameter below.

image

PR #16773, opened 2025-07-02
Improve rich editor attributes

fukuhara-trust:
## Description

This pull request enables greater customization of the input area in the Filament rich editor by allowing developers to opt out of the default `fi-prose` styling. This is particularly useful for cases where users want to selectively apply Tailwind CSS utilities to elements within the editor input area.

The key changes include:

- Made the `fi-prose` class on the editor wrapper optional, while retaining it by default to preserve existing styling behavior.
- Introducing support for `editorProps` in the Tiptap instance configuration.

Reference:
https://tiptap.dev/docs/editor/api/editor#editorprops

This enhancement empowers developers to better integrate the editor into design systems or bespoke UI environments without relying on CSS overrides.

### Before

The input content was always wrapped in `fi-prose`, applying Filament’s prose-like typography styling regardless of user preference.

```
<div class="fi-fo-rich-editor-content fi-prose" x-ref="editor">
  <div contenteditable="true" role="textbox" translate="no" class="tiptap ProseMirror" tabindex="0">
    <!-- editor content ... -->
  </div>
</div>
```

### After

It is now possible to opt out of the `fi-prose` class. Styling can be configured by the developer using Tailwind or other CSS frameworks, optionally via `editorProps`. Additionally, arbitrary attributes (such as spellcheck) can also be configured.

```
<div class="fi-fo-rich-editor-content" x-ref="editor">
  <div contenteditable="true" role="textbox" translate="no" class="tiptap ProseMirror prose prose-sm sm:prose lg:prose-lg xl:prose-2xl" spellcheck="false" tabindex="0">
    <!-- editor content ... -->
  </div>
</div>
```

If `editorProps` is not set, the default code would be.

```
<div class="fi-fo-rich-editor-content" x-ref="editor">
  <div contenteditable="true" role="textbox" translate="no" class="tiptap ProseMirror fi-prose" tabindex="0">
    <!-- editor content ... -->
  </div>
</div>
```

Developers can now implement something like:

```php
RichEditor::make('content')
    ->extraInputAttributes([
        'class' => 'prose prose-sm sm:prose lg:prose-lg xl:prose-2xl',
        'spellcheck' => 'false',
    ]);
```

## Breaking changes

This is expected to be a breaking change due to modifications and additions to the HTML elements to which extraAttributes(), extraInputAttributes(), and extraFieldWrapperAttributes() are applied.

## Functional changes

---

Let me know if any changes are needed to align better with Filament’s direction.


danharrin:
Sorry, this is now making the component inconsistent with others. `extraAttributes()` should always be merged onto the outer element of the component.

I would accept a new PR for a simple `extraEditorAttributes()` that get passed straight to the `editorProps.attributes` config.

For customising the `fi-prose` class, I suggest you use CSS overrides instead of completely replacing the class, to avoid breaking the editor between updates.

Issue #16775, opened 2025-07-02
SpatieMediaLibraryFileUpload breaks in filament v4.0.0-beta11 if multiple() is false or omitted


JoshuaSigaban:
### Package

filament/spatie-laravel-media-library-plugin

### Package Version

v3.3.30

### Laravel Version

v12.19.3

### Livewire Version

v3.5

### PHP Version

PHP v8.3.22

### Problem description

When using SpatieMediaLibraryFileUpload in Filament v4.0.0-beta11, an error occurs when ->multiple(false) is used or when multiple() is omitted (i.e., single file upload mode):

`array_keys(): Argument #1 ($array) must be of type array, Livewire\Features\SupportFileUploads\TemporaryUploadedFile given
`

This breaks form loading or file processing entirely in single upload mode.

### Expected behavior

The SpatieMediaLibraryFileUpload component should correctly handle single file uploads when ->multiple(false) is used or when multiple() is omitted.

- It should not throw a type error.
- The uploaded file should be saved and displayed correctly on the edit page.
- The component’s state should be properly normalized and compatible with single-file mode.

### Steps to reproduce

1. Install the filament/spatie-laravel-media-library-plugin into a Filament v4.0.0-beta11 project.
2. Create a simple Eloquent model implementing Spatie\MediaLibrary\HasMedia.
3. Create a Filament resource for that model.
4. Add the following form field:

`
SpatieMediaLibraryFileUpload::make('document')->collection('attachments')
`
Save the form.

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/JoshuaSigaban/filament-media-bug

### Relevant log output

```shell
array_keys(): Argument #1 ($array) must be of type array, Livewire\Features\SupportFileUploads\TemporaryUploadedFile given
```

JoshuaSigaban:
**Proposed solution:**  
Fix is available in this pull request: [filamentphp/filament#16776](https://github.com/filamentphp/filament/pull/16776)

It handles single vs multiple file state properly and resolves the `array_keys()` type error in `SpatieMediaLibraryFileUpload`.

danharrin:
I think the issue is that you have the wrong plugin version installed - 3.x not 4.x

JoshuaSigaban:
I apologize, @danharrin.

It appears that the issue may have been version-related, although I'm still unsure why the repository didn’t work for you with v3, especially since it was working fine on my end with v3 until I explicitly forced it to v4.

[...]

PR #16776, opened 2025-07-02
Fix handling of single vs multiple file uploads in SpatieMediaLibrary…

JoshuaSigaban:
## ✅ Description

This PR fixes two critical issues in `SpatieMediaLibraryFileUpload` when used in **single file mode** (`->multiple(false)` or omitted) on **Filament v4.0.0-beta11**.

---

### 🐛 Issues Fixed

**1. Fatal type error during submission in single mode**  
Previously, the component attempted to apply `array_keys()` on a `TemporaryUploadedFile` or a string UUID, causing a fatal type error:

`array_keys(): Argument #1 ($array) must be of type array, Livewire\Features\SupportFileUploads\TemporaryUploadedFile given`

**2. Media deletion logic deleting files incorrectly or failing to delete**  
- When no changes were made to the upload field, existing media were still being deleted.
- When all files were removed in **single mode**, the `state` was a string or `null`, causing the deletion logic to early return.
- As a result, uploaded media either persisted incorrectly or were prematurely removed.

---

### 🔍 Issue Discovered During Usage

While using the patched component in a production application, it became clear that media files were being deleted even when no interaction occurred with the upload field.

✅ **The fix introduced in the second commit:**
- Added a `hasStateChanges()` method to detect actual state changes.
- Ensured `deleteAbandonedFiles()` only runs when a change is detected.
- Normalized the `state` value to handle both string and array formats reliably in both single and multiple modes.

---

### ✅ What this PR does

- ✅ Normalizes the component’s state to an array, even in single mode  
- ✅ Prevents `array_keys()` type errors  
- ✅ Prevents media deletion unless the state and DB media UUIDs differ  
- ✅ Ensures files are deleted properly when removed from the UI  
- ✅ Fully supports both single and multiple file uploads  
- ✅ Maintains backwards compatibility with current plugin usage  

Closes: #16775

---

## 🖼️ Visual Changes

There are no visual design changes, but:

- ✅ Files now render correctly in single file mode on edit pages  
- ✅ No error is thrown during form submit  
- ✅ Uploading and deletion works for both single and multiple configurations  

---

## ⚙️ Functional Changes

- [x] Fixes `array_keys()` type error in single file mode  
- [x] Normalizes state for consistent behavior in both modes  
- [x] Prevents incorrect media deletion when form is unchanged  
- [x] Ensures correct deletion when all files are removed  
- [x] Tested with standard `HasMedia` model and resource setup  

---

## 📦 Code Quality

### DIFF
--- a/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
+++ b/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
@@ -32,7 +32,7 @@ class SpatieMediaLibraryFileUpload extends FileUpload
      * @var array<string, mixed> | Closure | null
      */
     protected array | Closure | null $customHeaders = null;
-
+    
     /**
      * @var array<string, mixed> | Closure | null
      */
@@ -42,7 +42,7 @@ class SpatieMediaLibraryFileUpload extends FileUpload
      * @var array<string, array<string, string>> | Closure | null
      */
     protected array | Closure | null $manipulations = null;
-
+    
     /**
      * @var array<string, mixed> | Closure | null
      */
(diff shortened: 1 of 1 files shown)


danharrin:
https://github.com/filamentphp/filament/issues/16775#issuecomment-3031418017

Issue #16725, opened 2025-06-28
Select field shifts to top-left when using ->searchable() inside Table Repeater column in Filament 4


Zibi198585:
### Package

filament/filament

### Package Version

v4beta10

### Laravel Version

v12

### PHP Version

php 8.3

### Problem description

When using a `Select::make()` field inside a Table Repeater column (using the `table()` method) in Filament 4, adding the `->searchable()` method causes the select field to shift to the top-left corner of the cell or container. 

<img width="647" alt="Image" src="https://github.com/user-attachments/assets/44d2fda6-9489-4b8b-a155-112554742787" />

**Example code**

```php
Repeater::make('users')
    ->schema([
        // ...
    ])
    ->table([
        TextColumn::make('name')->label('Name'),
        SelectColumn::make('role')
            ->options([
                'member' => 'Member',
                'administrator' => 'Administrator',
                'owner' => 'Owner',
            ])
            ->searchable(), // This causes the issue
    ])

### Expected behavior

The Select field should remain properly aligned and styled within the table cell, as it does without `->searchable()`.

<img width="343" alt="Image" src="https://github.com/user-attachments/assets/46fa234e-a814-4d6c-ae2c-8de122b88558" />

### Steps to reproduce

1. Create a Table Repeater using the `table()` method as per [Filament documentation](https://filamentphp.com/docs/4.x/forms/fields/repeater#table-repeaters).
2. Add a `Select::make()` or `SelectColumn::make()` inside the table columns.
3. Add the `->searchable()` method to the select field.
4. Observe that the select field is shifted to the top-left corner.

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/Zibi198585/Rental2/blob/master/app/Filament/Resources/RentalDocuments/Schemas/RentalDocumentForm.php

### Relevant log output

```shell

```

ht3aa:
<img width="1163" alt="Image" src="https://github.com/user-attachments/assets/6f9a8446-a9bb-4cba-a46c-fbc803320d77" />

yeah same problem

bishwajitcadhikary:
![Image](https://github.com/user-attachments/assets/baba7709-bbac-41ee-99af-dd7051d3d6a7)

Same here

PR #16779, opened 2025-07-02
[4.x] Fix position of select dropdown inside table repeater

damiandejong:
## Description

Recently, I also came across the issue as described in #16725 and the changes in this PR solved the problem. I am not sure about the variable naming, so feel free to adapt it.

## Visual changes

Before:

image

After:

image

## Functional changes


danharrin:
Thanks but taking a slightly different approach #16788

PR #16781, opened 2025-07-02
fix: only evaluate `hint()` and `hintIcon()` callbacks after hydration

ralphjsmit:
In Filament V4, a fallback mechanism has been made to replicate the previous `hint()` and `hintIcon()` methods. However, the issue with this compared to V3 is that in the fallback mechanism, any closures passed to `hint()` or `hintIcon()` already get evaluated _before_ the component is hydrated. If you would for example modify the component state in the `->afterStateHydrated()` method, then this code has not yet run and the `$state` parameter passed to the `hint()` and `hintIcon()` callbacks is not yet the modified one. This PR fixes the fallback-mechanism to only evaluate the hint and hint icon callbacks when they the `afterLabel` child component schema is actually rendered, thus making sure that the `afterStateHydrated()` runs first.

Thanks!


packages/forms/src/Components/Concerns/HasHint.php
```diff
@@ -34,18 +34,34 @@ protected function setUpHint(): void
...
 
-            $hint = $component->getHint();
-
-            if (filled($hint)) {
-                $components[] = Text::make($hint)
-                    ->color($component->getHintColor());
+            if ($component->hasHint()) {
+                $components[] = Text::make(static function (Text $component) {
+                    /** @phpstan-ignore-next-line */
+                    return $component->getContainer()->getParentComponent()->getHint();
```
ralphjsmit: The PHPStan was complaining after that it didn't understand the return type of `getParentComponent()` as having the `getHint()` method (etc).

danharrin (changes requested):
Thanks, just missing return types on those new closures if you could add them please

danharrin:
Also, does this need applying to "helper text"?

ralphjsmit:
Added @danharrin.

danharrin (approved):

PR #16783, opened 2025-07-02
fix: `FusedGroup` should stretch entire height when one single-line component stretches

ralphjsmit:
The situation is not optimal in both cases, but this is before:

CleanShot 2025-07-02 at 20 24 25@2x

And after:

CleanShot 2025-07-02 at 20 23 36@2x


packages/schemas/resources/css/components/fused-group.css
```diff
@@ -6,6 +6,18 @@
     & .fi-sc {
         @apply gap-px rounded-lg bg-gray-950/10 dark:bg-white/20;
 
+        .fi-sc-component {
+            @apply h-full;
+
+            & > .fi-fo-field {
+                @apply h-full;
+            }
```
danharrin: I think this works and is more consistent, right?

```suggestion
        & .fi-sc-component {
            @apply h-full;

            & .fi-fo-field {
                @apply h-full;
            }
```

danharrin (changes requested):

danharrin:
Please also recompile the CSS

packages/schemas/resources/css/components/fused-group.css
ralphjsmit: Am not so at home in this type of CSS, but appears to work good as well!

ralphjsmit:
Done @danharrin

danharrin (approved):

PR #16817, opened 2025-07-04
Fix SPA mode component not found errors

### LINKED ISSUE #16401: TextareaFormComponent is not defined when in tabs
### Package

filament/filament

### Package Version

v3.3.20

### Laravel Version

v11.45.1

### Livewire Version

v3.6.3

### PHP Version

PHP 8.3.21

### Problem description

While using SPA mode, rendering a Filament form schema that includes tabs (schema-based tabs), placing a Textarea form component inside the second tab tab content triggers this error during runtime but if in first tab there is no error. The component fails to initialize correctly via Alpine.js, resulting in a JS exception.

But if without SPA no error show.

The relevant error trace:

```
Uncaught ReferenceError: TextareaFormComponent is not defined
    at [Alpine] textareaFormComponent
    at eval (at safeAsyncFunction)
    ...
Uncaught ReferenceError: state is not defined
    at [Alpine] state (at safeAsyncFunction)
    ...
```

Attached is the screenshot of the full error log as shown in the developer console:

Image

https://github.com/user-attachments/assets/50c7bda9-ea8e-4032-bfbd-7396ec743ae5

### Expected behavior

- No Alpine runtime error should be triggered.

### Steps to reproduce

1. Create a Filament form schema using tabs.
2. Inside the second the tab item, add a Forms\Components\Textarea::make() field.
3. Render the page.
4. Observe the browser console for the ReferenceError.

https://github.com/str4wh4t/app-filament-issue-3.x

### Relevant log output

```shell

```

### LINKED ISSUE #16681: v4 `state is not defined` and `xxxFormComponent is not defined` console errors with `live()` repeaters in tabs
### Package

filament/filament

### Package Version

4-beta9

### Laravel Version

12

### Livewire Version

3

### PHP Version

8.4

### Problem description

Using `->live()` on a `Repeater` in a tab causes `xxxFormComponent is not defined` and `state is not defined` console errors under certain conditions:

https://github.com/user-attachments/assets/fc6a0ce3-4ab7-4f08-ae7f-049b395d8675

### Expected behavior

Repeaters should work properly with `->live()` regardless of where they are in the schema.

### Steps to reproduce

1. Clone repo
2. Go to "Repeater Alpine Errors" page
3. Enter some info on the first tab, switch to the second tab, enter some info on the second tab
4. Reload the browser
5. Tabs are persisted so second tab should still be active. Enter some info on the second tab -> console errors

https://github.com/binaryfire/filament-bug-reproduction

### Relevant log output

```shell

```

binaryfire:
Fixes https://github.com/filamentphp/filament/issues/16681 and https://github.com/filamentphp/filament/issues/16401

I'm running into the same SPA mode issue as https://github.com/filamentphp/filament/issues/16401 with Wizards:

https://github.com/user-attachments/assets/599a18e0-d618-4c9b-8e34-0dc2c30abaf0

Reproduction repo: https://github.com/binaryfire/filament-bug-reproduction (`Wizard Text Area Not Found` page). 

Changing the Async Alpine loading strategy to [`eager`](https://async-alpine.dev/docs/strategies/#eager) fixes it, as reported [here](https://github.com/filamentphp/filament/issues/16401#issuecomment-2958579107). This means assets are loaded when the page loads instead of when the components are visible, but that's still much better than loading loading them all on the user's first visit. It'll probably reduce the visual glitch that happens when switching tabs / steps while waiting for the js to download too.

If you have an alternative approach in mind let me know and I'll check that out.


danharrin:
If we switch strategies like this, do we still need `|| event (x-modal-opened)` to ensure that components work in modals consistently? Or was that just an issue with `visible`?

binaryfire:
@danharrin Just tested the non-native select and code editor in a modal with only `x-load="eager"` (no `|| event (x-modal-opened)`). No errors:

[2025-07-07 at 10.23.48.880 PM.webm](https://github.com/user-attachments/assets/4c8cda83-be0e-491c-a44d-7af3e25e6f99)

It seems to solve a lot of problems. I'm guessing it'll help avoid some other potential `navigate` issues in the future too.

danharrin:
In which case, please completely remove those attributes everywhere, since `eager` is the default behaviour of the library

binaryfire:
Will do. Is there a reason why it was set to `visible` just for SPA mode? Just curious.

wychoong:
> Is there a reason why it was set to `visible` just for SPA mode?

from what I can find, it was introduced in https://github.com/filamentphp/filament/pull/9204 when still using AsyncAlpine v1
later on `event` based strategy was added in https://github.com/filamentphp/filament/pull/13921 after a bug introduced in a different fix https://github.com/filamentphp/filament/pull/13740 for page scroll lock

after https://github.com/filamentphp/filament/pull/15559 upgrade AsyncAlpine to v2, these could be just artefact that can be cleaned, if it:
- works fine with/without spa & inside/outside of modal
- page scroll lock works fine with modal & nested modal

side note: those issue might be fixed as per this issue https://github.com/Accudio/async-alpine/issues/52 that no workaround is required in filament anymore

danharrin:
It looks like the Markdown editor still needs `visible || event` otherwise it initialises inconsistently in modals. Very weird behaviour

danharrin (approved):

PR #16822, opened 2025-07-04
remove hasRelationship check

samarija:
## Description

When hasRelationship() returned false in line 67 the component was not added to $componentsWithThisRelationship. So the condition for the deletion of a record never matched.

## Functional changes

### DIFF
--- a/packages/schemas/src/Components/Concerns/EntanglesStateWithSingularRelationship.php
+++ b/packages/schemas/src/Components/Concerns/EntanglesStateWithSingularRelationship.php
@@ -63,8 +63,7 @@ trait EntanglesStateWithSingularRelationship
 
                     if (
                         ($childComponent instanceof CanEntangleWithSingularRelationships) &&
-                        ($childComponent->getRelationshipName() === $component->getRelationshipName()) &&
-                        ($childComponent->hasRelationship())
+                        ($childComponent->getRelationshipName() === $component->getRelationshipName())
                     ) {
                         $componentsWithThisRelationship[] = $childComponent;
 


danharrin:
I need more info about why this is a problem, because the description doesnt really tell me why `hasRelationship` is false in this case and there is no relevant bug report.

danharrin:
@samarija looking to understand more please, can you explain the above?

danharrin:
If you have any info to add, just reply here and I will reopen the PR.

samarija:
@danharrin , sorry for my late replay, i was offline for two weeks.

I tried to use the following functionality as described in the documentation:
https://filamentphp.com/docs/4.x/forms/overview#conditionally-saving-data-to-a-relationship

The part with the deletion of the related record, if the condition does not match, didn't work with filament 4, while it worked with filament 3. As far as i could see, the deleted `hasRelationship` -check returns false if the defined condition (see documentation) is not met. So the relevant component is not added to `$componentsWithThisRelationship`.

This seems to lead to a failing check in the following part:

```
// The first layout component using this relationship is the one that will save the relationship for all of them.
if (Arr::first($componentsWithThisRelationship) !== $component) {
   return;
}
```
So the later following part for the deletion of the related record is never reached in code:

```
if (! $component->hasRelationship()) {
      $record?->delete();

     return;
} 
```

After a second look it may make more sense to move the `$componentsWithThisRelationship` check below the deletion-code.

Sorry for the incomplete description. Next time i will try to make it better.

danharrin:
Can you please open an issue with a reproduction repository, so I can check the issue out locally?

samarija:
done #17066

PR #16826, opened 2025-07-05
Add attribute for in label

jeffersongoncalves:
Add attribute for in label in field-wrapper.blade.php.


danharrin:
Hmm I am going to need some better reasoning for this, you shouldn't need the attribute if the input exists within the `<label>` tag.

danharrin:
I completely misread this and you're absolutely right, my apologies.

danharrin (approved):

Issue #16720, opened 2025-06-27
Displayed column order does not match the order specified in the Table columns array


nabileisa:
### Package

filament/filament

### Package Version

v4.0.0-beta10

### Laravel Version

v12.19.3

### PHP Version

8.3.22

### Problem description

When creating a table in the filament panel , changing the order of the columns in the array  does not  change the order displayed in the panel. any new columns are added at the end on the table even if it is at the beginning of the columns array.
Tor example : 
this table  array 
```php
<?php

namespace App\Filament\Resources\Categories\Tables;

use Filament\Actions\BulkActionGroup;
use Filament\Actions\DeleteBulkAction;
use Filament\Actions\EditAction;
use Filament\Tables\Columns\TextColumn;
use Filament\Tables\Table;

class CategoriesTable
{
    public static function configure(Table $table): Table
    {
        return $table
            ->columns([
                TextColumn::make('slug')
                    ->label('Slug')
                    ->sortable()
                    ->searchable(),
                TextColumn::make('description')
                    ->label('Description')
                    ->limit(50) // Limit the description to 50 characters
                    ->searchable(),
                TextColumn::make('name')
                    ->label('Category Name')
                    ->sortable()
                    ->searchable(),

                TextColumn::make('icon')
                    ->label('Icon')
                    ->sortable()
                    ->searchable(),
            ])
            ->filters([
                //
            ])
            ->recordActions([
                EditAction::make(),
            ])
            ->toolbarActions([
                BulkActionGroup::make([
                    DeleteBulkAction::make(),
                ]),
            ]);
    }
}

```
renders this table which is obviously wrong
![Image](https://github.com/user-attachments/assets/25000e15-fa97-4db0-bd59-8cce663065ca)

### Expected behavior

columns are rendered in the same order in the columns array

### Steps to reproduce

- create a new laravel project 
- install filament beta
- create a table with multiple columns

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/nabileisa/filament4-beta10-bug

### Relevant log output

```shell

```

danharrin:
If you sign out and sign back in, does it work correctly?

nabileisa:
Yes, just tested and you are correct, if you sign out and sign back in it does display the correct order.

tabatii:
same problem

CoringaWc:
same problem

archilex:
[...]

PR #16835, opened 2025-07-05
Check enable Reorderable Columns

jeffersongoncalves:
Even though column reordering was not activated, there was a problem with not maintaining the order defined in the columns of the resource table.

Columns definition:

image

Bug reported in issue #16720 that changed the order of columns, even though column reordering was not enabled in the table.

image

Locally test the default behavior of columns in the correct order:

image

Debugging where to change the position of the columns I realized that the applyTableColumnManager from Concerns/HasColumnManager.php is called and the position of the columns was changed.

### DIFF
--- a/packages/tables/src/Concerns/HasColumnManager.php
+++ b/packages/tables/src/Concerns/HasColumnManager.php
@@ -63,6 +63,10 @@ trait HasColumnManager
      */
     public function applyTableColumnManager(?array $state = null): void
     {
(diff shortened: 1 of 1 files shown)


archilex:
@jeffersongoncalves Thanks for the PR, however early returning here would prevent new columns added by the dev to be synced to the table/column manager. I'm going to close this PR since this will be all fixed in the PR I'm working on for #16720. Thanks!

archilex (commented):
Closing

PR #16838, opened 2025-07-06
Hide inactive wizard steps on mobile

binaryfire:
Fixes: https://github.com/filamentphp/filament/discussions/16820

After the change:

**Large screens (no change)**
image

**Small screens**
image
image


packages/schemas/resources/css/components/wizard.css
```diff
@@ -2,11 +2,15 @@
     @apply flex flex-col;
 
     & .fi-sc-wizard-header {
-        @apply grid divide-y divide-gray-200 md:grid-flow-col md:divide-y-0 md:overflow-x-auto dark:divide-white/5;
+        @apply grid divide-gray-200 md:grid-flow-col md:divide-y-0 md:overflow-x-auto dark:divide-white/5;
```
danharrin: Adding this class back doesn't hurt anything, right? I am just thinking if a user wanted to disable this behaviour by overriding `&:not(.fi-active) {` in their own stylesheet, they wouldn't also need to redefine this.

```suggestion
        @apply grid divide-y divide-gray-200 md:grid-flow-col md:divide-y-0 md:overflow-x-auto dark:divide-white/5;
```

danharrin (changes requested):

packages/schemas/resources/css/components/wizard.css
binaryfire: When it's there, there's an extra line under the header on mobile.

image

danharrin: Okay, please remove all other divide classes from that line then.

binaryfire: Done

danharrin (approved):

PR #16845, opened 2025-07-06
V4 fix reorderable columns

### LINKED ISSUE #16720: Displayed column order does not match the order specified in the Table columns array

archilex:
Fixes #16720. Also, adds a reset button to reset the table to it's default state (ie, the state determined by the dev).

These changes also brought speed improvements when initializing the column manager when reoderable columns are not enabled, or when they are but the user hasn't reordered any columns. From 0.13 ms to 0.08 ms.

@danharrin to fix this behavior, new methods were introduced and to keep naming clear and consistent, I renamed other methods/variables to be consistent. Feel free to work your naming magic.

Also, I hope to build out a test suite for all this functionality in the near future to make sure future changes don't break current functionality.


packages/tables/src/Concerns/HasColumnManager.php
```diff
@@ -21,6 +21,11 @@ trait HasColumnManager
      */
     public array $tableColumns = [];
 
+    /**
+     * @var ?array<int, array{type: string, name: string, label: string, isToggled: bool, isToggleable: bool, columns?: array<int, array{type: string, name: string, label: string, isToggled: bool, isToggleable: bool}>}>
+     */
+    public ?array $cachedDefaultTableColumnState = null;
```
danharrin: This looks good and seems to work great, my only question is about this property. Does it need to be persisted between Livewire requests or can it be rebuilt? I think it can probably be protected, or it should be `#[Locked]` so it cant be tampered with on the frontend?

danharrin (changes requested):
The only real name changes I needed to make were to do with adding `table` to the names of methods. When we are adding new methods to traits like `InteractsWithTables`, which get added to classes in the users own codebase, we need to try and minimise clashes with any methods they already have. Names like `findMatchingItem()` are going to more likely clash with other methods than `findMatchingTableColumnStateItem()`. For methods that exist on Filament internal classes I worry less about this.

packages/tables/src/Concerns/HasColumnManager.php
archilex: Fixed. It should be protected. I had made it public when I was debugging and forgot to switch it back. And I'm only caching it for the request since multiple methods need to access this and this was cleaner than passing a variable between a bunch of methods.

danharrin (approved):

danharrin:
Thanks!

PR #16847, opened 2025-07-07
Add Hide Option method to option-based form components

ahinkle:
## Description

When working with enums in form inputs with options, developers may face a frustrating workflow when they need to conditionally show/hide options based on business logic. Here's a common scenario:

The enum setup (works great initially):

```php
enum OrderStatus: string implements HasLabel, HasIcon, HasColor
```

The Beautiful Component (Initially)
```php
ToggleButtons::make('status')
    ->options(OrderStatus::class)
```

The Reality Check (Business Logic Kicks In)

- Draft orders can't be set to 'Refunded' directly
- Only admins can set orders to 'Cancelled'
- Delivered orders can't go back to 'Processing'

While disabled can be useful, it's not enough when you're dealing with a large set of options, especially when clarity matters. It removes the noise entirely.

The Current Painful Solution

We lose all niceties of enums. 
```php
ToggleButtons::make('status')
    ->options(OrderStatus::availableOptions($order))
    ->icons(OrderStatus::icons())
    ->colors(OrderStatus::colors())
```

When we have to do some crufty thing in the enum to get the label from `availableOptions`.

### Solution

A `hideOptionWhen()` method solves this by allowing you to keep your beautiful enum setup while conditionally hiding options:

```php
ToggleButtons::make('status')
    ->options(OrderStatus::class)
    ->hideOptionWhen(fn (string $value): bool => $value === 'published')
```

## Visual changes

NA

## Functional changes


danharrin:
Hi, please can you target this at v4. Thanks

PR #16851, opened 2025-07-07
Add `extraAttributes` to tables

binaryfire:
I need to apply a bunch of styling to specific tables. Sometimes these are in pages with more than one table. Having `extraAttributes` on tables lets me add a class I can use as selector in my theme.

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -115,10 +115,10 @@
                 currentSelectionLivewireProperty: @js($getCurrentSelectionLivewireProperty()),
                 $wire,
             })"
-    @class([
+    {{ $attributes->merge($getExtraAttributes())->class([
         'fi-ta',
         'fi-loading' => $records === null,
-    ])
+    ]) }}
 >
     <input
         type="hidden"
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -115,10 +115,12 @@
                 currentSelectionLivewireProperty: @js($getCurrentSelectionLivewireProperty()),
                 $wire,
             })"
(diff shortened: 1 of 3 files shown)


packages/tables/resources/views/index.blade.php
```diff
@@ -115,10 +115,12 @@
                 currentSelectionLivewireProperty: @js($getCurrentSelectionLivewireProperty()),
                 $wire,
             })"
-    @class([
-        'fi-ta',
-        'fi-loading' => $records === null,
-    ])
+    {{
+        $attributes->merge($getExtraAttributes())->class([
```
danharrin: Does this work too?

```suggestion
        $getExtraAttributeBag()->class([
```

danharrin (changes requested):

packages/tables/resources/views/index.blade.php
binaryfire: Yeah it does. Changed.

binaryfire:
Pint really doesn't like what you're doing in `CanGenerateButtonHtml` lol. Do you want me to revert it?

binaryfire:
Reverted. Is there a way of telling Pint to ignore those lines? Otherwise we have to restore it every time we run `composer cs `.

danharrin:
I cant reproduce the Pint behaviour, do you have up to date dependencies?

binaryfire:
Weird, it must be on my end then. I just cloned a fresh copy of the Filament repo, checked out 4.x, ran composer + npm install, then ran composer cs:
image

danharrin (approved):

binaryfire:
@danharrin Btw I figured out why that file kept changing. My SonarScanner cli config was corrupted and it was wigging out.

PR #16855, opened 2025-07-07
Added toolbarButtons() change

Casmo:
Added toolbarButtons() update for v4.


danharrin:
Please can you open an issue instead, as this is an unintended breaking change. They should all be assigned to a group if you don't have nested arrays. I would like to fix it instead of documenting the bug.

Casmo:
Gotcha. https://github.com/filamentphp/filament/pull/16871.

PR #16862, opened 2025-07-08
Fix PostgreSQL JSON searching

glennjacobs:
## Description

When searching JSON fields in pgsql it would error with 

```
SQLSTATE[42703]: Undefined column: 7 ERROR: column "name" does not exist LINE 1: ...."deleted_at" is null) and (lower(attribute_data->name::text... ^
```

This was because the column search expression was 

```
lower(attribute_data -> name::text)::text LIKE '%chuck%'
```

rather than the correct

```
lower(attribute_data ->> 'name'::text)::text LIKE '%chuck%'
```

You will see there's two changes: 

1. `->` changes to `->>`
2. we are wrapping `name` in single inverted commas

## Tests

I have added two tests to show that the MySQL version remains as it was, and the PostgreSQL returns as expected.

## Functional changes

### DIFF
--- a/packages/support/src/helpers.php
+++ b/packages/support/src/helpers.php
@@ -171,19 +171,14 @@ if (! function_exists('Filament\Support\generate_search_column_expression')) {
         }
 
         $column = match ($driverName) {
-            'pgsql' => str($column)->contains('->')
-                ? str($column)
-                    ->beforeLast('->')
-                    ->append('->>')
-                    ->append("'")
-                    ->append(str($column)->afterLast('->'))
-                    ->append("'")
-                : $column,
-            default => $column,
-        };
-
-        $column = match ($driverName) {
-            'pgsql' => "{$column}::text",
+            'pgsql' => (str($column)->contains('->')
+                    ? str($column)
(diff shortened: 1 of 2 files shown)


packages/support/src/helpers.php
```diff
@@ -170,6 +170,18 @@ function generate_search_column_expression(string $column, ?bool $isSearchForced
             };
         }
 
+        $column = match ($driverName) {
```
danharrin, line 173: Should this be combined with the `match` below it maybe?

Also, does this change only need to take place when `lower()` is used, or is it also without `lower()`?

danharrin (changes requested):

packages/support/src/helpers.php
glennjacobs: Sure, I'll try combine the two.

It's required for both with and without `lower()`.

danharrin:
Thanks

danharrin (approved):

danharrin:
@glennjacobs what do you think of the fix just proposed here #16865

Issue #16864, opened 2025-07-08
SQL syntax error during global search when using : in table names


Capevace:
### Package

filament/filament

### Package Version

v3.3.30

### Laravel Version

v12.19

### Livewire Version

3.6.3

### PHP Version

8.3

### Problem description

I have some tables that have `:` characters in the table names inside a Postgres database. Laravel and Postgres both support this and it works without problem (with only some 3rd party libraries not handling it well).

It works in Filament, however sometime recently, the global search started throwing SQL errors in relation to the table names.

```
SQLSTATE[42601]: Syntax error: 7 ERROR: syntax error at or near ":" LINE 1: ...ect *
 from "test:dbname:periods" where (lower(test:dbname:pe... ^ (Connection: pgsql, SQL: 
select * from "test:dbname:periods" where 
(lower(test:dbname:periods.name::text)::text like %asd%) limit 50) POST 
127.0.0.1:8000 PHP 8.3.21 — Laravel 12.19.3
```

I've narrowed down the issue to the `generate_search_column_expression` function.

```php
function generate_search_column_expression(string $column, ?bool $isSearchForcedCaseInsensitive, Connection $databaseConnection): string | Expression
    {
        $driverName = $databaseConnection->getDriverName();

        if (Str::lower($column) !== $column) {
            $column = match ($driverName) {
                'pgsql' => (string) str($column)->wrap('"'),
                default => $column,
            };
        }
```

The problem disappears when modified like this:

```php
function generate_search_column_expression(string $column, ?bool $isSearchForcedCaseInsensitive, Connection $databaseConnection): string | Expression
    {
        $driverName = $databaseConnection->getDriverName();

            $column = match ($driverName) {
                'pgsql' => (string) str($column)
			->explode('.')
			->map(fn (string $part) => str($part)->wrap('"'))
			->implode('.'),
                default => $column,
            };
```

This wraps each segment of the column name (`x.y.z`) into their quoted alternatives (`"x"."y"."z"`). After this, the search behaves as expected.

### Expected behavior

You'd expect the model to be searched normally in the global search without error.

### Steps to reproduce

I've attached the example repository that has done these steps.

1. Use a Postgres DB
2. Create a model/migration/resource with a table name containing a period e.g. `contacts:friends` or `products:categories`
3. Enable global search on the resource
4. Use the global search with anything.

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/Capevace/filament-pgsql-quote-error

### Relevant log output

[...]

PR #16865, opened 2025-07-08
Fix SQL syntax error during global search when using : in table names #16864

Capevace:
## Description

I've experienced a bug with PGSQL and global search when using `:` in table names.

I've documented the issue in detail in #16864.

This PR applies the suggested fix in the issue, adding the quoting logic.

## Visual changes

There are no visual changes.

## Functional changes

### DIFF
--- a/packages/support/src/helpers.php
+++ b/packages/support/src/helpers.php
@@ -163,15 +163,11 @@ if (! function_exists('Filament\Support\generate_search_column_expression')) {
     {
         $driverName = $databaseConnection->getDriverName();
 
-        if (Str::lower($column) !== $column) {
-            $column = match ($driverName) {
-                'pgsql' => (string) str($column)->wrap('"'),
-                default => $column,
-            };
-        }
-
         $column = match ($driverName) {
-            'pgsql' => "{$column}::text",
+            'pgsql' => (string) str($column)
+                ->explode('.')
+                ->map(fn (string $part) => str($part)->wrap('"'))
+                ->implode('.') . '::text',
             default => $column,
         };
 
--- a/packages/support/src/helpers.php
+++ b/packages/support/src/helpers.php
@@ -166,8 +166,13 @@ if (! function_exists('Filament\Support\generate_search_column_expression')) {
         $column = match ($driverName) {
             'pgsql' => (
                 str($column)->contains('->')
-                    ? str($column)
-                        ->beforeLast('->')
+                    ? str(
+                        str($column)
+                            ->beforeLast('->')
+                            ->explode('.')
+                            ->map(fn (string $part) => str($part)->wrap('"'))
+                            ->implode('.')
+                    )
                         ->append('->>')
                         ->append("'")
                         ->append(str($column)->afterLast('->'))
--- a/tests/src/Support/helpersTest.php
+++ b/tests/src/Support/helpersTest.php
@@ -75,5 +75,24 @@ it('will generate json search column expression for pgsql', function () {
     $expression = generate_search_column_expression($column, $isSearchForcedCaseInsensitive, $databaseConnection);
 
     expect($expression->getValue($grammar))
-        ->toBe("lower(data->>'name'::text)");
+        ->toBe("lower(\"data\"->>'name'::text)");
 });
+
+it('will generate column expression for pgsql with colons in the name', function (string $column, string $text) {
+    $isSearchForcedCaseInsensitive = true;
+
+    $databaseConnection = Mockery::mock(Connection::class);
+    $databaseConnection->shouldReceive('getDriverName')->andReturn('pgsql');
+    $databaseConnection->shouldReceive('getConfig')->with('search_collation')->andReturn(null);
+
+    $grammar = new PostgresGrammar($databaseConnection);
+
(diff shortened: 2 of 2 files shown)


danharrin:
@Capevace can you fix the merge conflict please

Capevace:
Merged the changes in.

I'm unsure about the JSON logic, but I saw it was from another PR. I've left it as is if a `->` is in the column.

packages/support/src/helpers.php
```diff
@@ -163,15 +163,11 @@ function generate_search_column_expression(string $column, ?bool $isSearchForced
...
-            $column = match ($driverName) {
-                'pgsql' => (string) str($column)->wrap('"'),
-                default => $column,
-            };
-        }
-
         $column = match ($driverName) {
-            'pgsql' => "{$column}::text",
+            'pgsql' => (string) str($column)
+                ->explode('.')
```
glennjacobs: I think this needs to consider JSON fields, e.g. `->`.

For example, we wouldn't want to end up with ...

```
"data ->> 'name'"::text
```

If you update your branch you'll see my recent addition to fix JSON field searching, so hopefully this can be incorporated into that.

My update also provided some tests that you can borrow from to cover this update.

Please see https://github.com/filamentphp/filament/blob/3.x/packages/support/src/helpers.php#L173-L183

And https://github.com/filamentphp/filament/blob/3.x/tests/src/Support/helpersTest.php#L47-L79

danharrin: @glennjacobs that was merged in now, and JSON fields aren't being changed. Maybe they should be though?

Capevace: please check the re-merged version, i've added the logic from the other PR too

packages/support/src/helpers.php
```diff
@@ -163,22 +163,20 @@ function generate_search_column_expression(string $column, ?bool $isSearchForced
...
+            'pgsql' => (
+                str($column)->contains('->')
                     ? str($column)
                         ->beforeLast('->')
                         ->append('->>')
                         ->append("'")
                         ->append(str($column)->afterLast('->'))
                         ->append("'")
-                    : $column) . '::text',
+                    : str($column)
```
glennjacobs, line 180: This will only wrap the table/field names when it's not a JSON search, which I think is wrong.

danharrin:
It's funny how two very related PRs got opened very close to each other, after a long time with no changes to this part of the codebase 🤣

glennjacobs:
> It's funny how two very related PRs got opened very close to each other, after a long time with no changes to this part of the codebase 🤣

I promise we didn't plan it! lol

Capevace:
> It's funny how two very related PRs got opened very close to each other, after a long time with no changes to this part of the codebase 🤣

Yes very confusing at the moment 😂 

Dan, what do you think about the question by Glenn? 
I wasn't quite sure about the JSON syntax myself, so I left it in place. This fixes this issue for every other case tho and leaves the JSON logic in place.

What would the SQL look like with `:` and JSON? 

`"table:xyz"->abc`?

packages/support/src/helpers.php
Capevace: I agree, but I'm not too familiar with the pgsql JSON syntax.

danharrin:
Should it be like

```
lower("table"."column" ->> 'name'::text)::text LIKE '%chuck%'
```

I am no expert on this so it will need testing

Capevace:
> `lower("table"."column" ->> 'name'::text)::text LIKE '%chuck%'`

That's probably correct. In my head, implementing a fix for this would add some complexity, so I agree on the testing.

Would it be possible to merge as-is for now?

The current solution both fixes the issue for 1) normal table names + JSON columns and also 2) `:`-table names without JSON and doesn't break anything else. 

But this issue is currently causing a bug in our app, so it'd be nice to get this fixed first, and a better implementation treated as a separate issue perhaps?

glennjacobs:
How about this?

```php
        $column = match ($driverName) {
            'pgsql' => (
                str($column)->contains('->')
                    ? str(
                        str($column)
                            ->beforeLast('->')
                            ->explode('.')
                            ->map(fn (string $part) => str($part)->wrap('"'))
                            ->implode('.')
                    )
                        ->append('->>')
                        ->append("'")
                        ->append(str($column)->afterLast('->'))
                        ->append("'")
                    : str($column)
                        ->explode('.')
                        ->map(fn (string $part) => str($part)->wrap('"'))
                        ->implode('.')
            ) . '::text',
            default => $column,
        };
```

glennjacobs:
You'll need to update the test to update the expected output to

```php
    expect($expression->getValue($grammar))
        ->toBe("lower(\"data\"->>'name'::text)");
```

glennjacobs:
It would be good to add an additional test or two to cover your scenarios.

Capevace:
Just added your solution and some tests

glennjacobs:
> Just added your solution and some tests

Looks good to me!

Capevace:
> How about this?

...

> Looks good to me!


image


Thanks for your help, Glenn! 🙏🏻🫶

danharrin (approved):

danharrin:
Thanks guys

Issue #16841, opened 2025-07-06
FileUpload with multiple option creates invisible div over form


Lirux:
### Package

filament/filament

### Package Version

4.0.0-beta13

### Laravel Version

12.19.3

### Livewire Version

3.6.3

### PHP Version

8.3.19

### Problem description

FileUpload and SpatieMediaLibraryFileUpload with multiple() option creates invisible div. Possible to block other form elements.

<img width="1252" height="472" alt="Image" src="https://github.com/user-attachments/assets/1dfb67fc-1dec-4389-b877-55847ed39d9e" />

### Expected behavior

Not blocking other form elements.

### Steps to reproduce

Create form with FileUpload or SpatieMediaLibraryFileUpload field and add directly below the FileUpload fields an input field.

```php
TextInput::make('title')
    ->columnSpanFull(),

SpatieMediaLibraryFileUpload::make('images')
    ->label('Spatie File Upload')
    ->multiple(),

FileUpload::make('images')
    ->label('File Upload')
    ->multiple(),

TextInput::make('foo'),

TextInput::make('bar'),
```

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/Lirux/filament-file-upload-bug

### Relevant log output

```shell

```

demianottema:
This happens also for select fields where multiple() is used.

danharrin:
This is the related Filepond issue to track: https://github.com/pqina/filepond/issues/957

Lirux:
Created a pull request to fix this issue. https://github.com/filamentphp/filament/pull/16900

PR #16900, opened 2025-07-09
Fix filepond invisible blocking div

Lirux:
## Description

Fixes issue: https://github.com/filamentphp/filament/issues/16841

Overwrite the css classes because the filepond project seems to be not updated anymore.

Tested it for SpatieMediaLibraryFileUpload and FileUpload. With and without multiple().

## Visual changes
Before
Screenshot 2025-07-06 at 12 43 16

After
Screenshot 2025-07-09 at 22 55 05

Working like expected
https://github.com/user-attachments/assets/d1fe9644-2925-4cb1-91f0-0397cbb04cc8

## Functional changes

### DIFF
--- a/packages/forms/resources/css/components/file-upload.css
+++ b/packages/forms/resources/css/components/file-upload.css
@@ -41,6 +41,16 @@
         @apply rounded-full;
     }
 
+    & .filepond--root[data-style-panel-layout~='compact'],
+    & .filepond--root[data-style-panel-layout~='integrated'] {
+        & .filepond--list-scroller {
+            overflow: unset;
+            height: unset;
+            margin-top: 0;
+            margin-bottom: 0;
+        }
+    }
+
     & .filepond--panel-root {
         @apply bg-transparent;
     }
--- a/packages/forms/resources/css/components/file-upload.css
+++ b/packages/forms/resources/css/components/file-upload.css
@@ -30,7 +30,7 @@
     }
 
     & .filepond--root {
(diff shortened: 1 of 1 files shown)


danharrin:
What is the scroller used for? Are we not able to just apply an `overflow-hidden` to the `--root`, or does it have unintended consequences? These overrides feel like they could break something

Lirux:
So, I did some tests. 

Your suggestion to just put overflow-hidden would also work. 
- I tested it on mobile browser -> no issues
- I tested it the reorderable() option -> no issues
- I tested it create in modal
- I tested it create in slideover
- 
My fix is just a edited version from the filepond issue you linked.

There is no evidence that the list-scroller is used for other stuff.
- I tried with max-width on the filepond--root to simulate a fix height. [But the scrollbars didn't show up automatically](https://pqina.nl/filepond/docs/api/style/#limiting-the-height-of-filepond).
- The list scroller could have a "overflow" state. But it's not documented how to assign that to the [list scroller](https://github.com/pqina/filepond/blob/fef56ae870dc05caeb0edda7879393dca381732a/src/css/list-scroller.css#L12).
- It's possible that this list-scroller has no special function anymore?? [Last commit 6 years ago??](https://github.com/pqina/filepond/blob/fef56ae870dc05caeb0edda7879393dca381732a/src/js/app/view/listScroller.js)

danharrin (changes requested):
Thanks for the investigation. I think let's go with the `overflow-hidden` fix then please @Lirux since it changes less of the internal CSS.

Lirux:
Just added the overflow-hidden and removed my other attempt.

danharrin (approved):

danharrin:
Thanks

Issue #16904, opened 2025-07-10
Slideover nav menu always opens on page load when using `->sidebarCollapsibleOnDesktop()`


binaryfire:
### Package

filament/filament

### Package Version

4-beta15

### Laravel Version

12

### Livewire Version

3

### PHP Version

8.4

### Problem description

The problem described in [this issue](https://github.com/filamentphp/filament/issues/15056) wasn't fixed by this [PR](https://github.com/filamentphp/filament/pull/16647).

The issue isn't with the sidebar, it's with the behavior of the slideover nav menu. If you're using `->sidebarCollapsibleOnDesktop()` and the sidebar is expanded, the slideover opens automatically on small breakpoints (without clicking the hamburger icon). Which means users on small devices or who resize their browser window need to click away to close it:

https://github.com/user-attachments/assets/4801c88a-601a-4597-adc6-5cd2b06c16de

The logic for collapsing the sidebar and opening the slideover seem to be linked, instead of being handled separately. When the sidebar is collapsed, the problem doesn't occur.

### Expected behavior

Collapsing / expanding the sidebar shouldn't cause the slideover nav menu to open.

### Steps to reproduce

1. Clone reproduction repo
2. Follow the steps in the video above

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/binaryfire/filament-bug-reproduction

### Relevant log output

```shell

```

danharrin:
I don't think they need to be separated, we just need to add some sort of `ResizeObserver` so that when the browser gets smaller and the sidebar becomes an overlay, it also closes.

binaryfire:
Cool yeah that’s what I was thinking. I’ll put something by together.

Issue #15056, opened 2024-12-11
On mobile, filament panel loads with sidebar expanded


Persaeus:
### Package

filament/filament

### Package Version

v3.2.128

### Laravel Version

v11.34.2

### Livewire Version

v3.5.12

### PHP Version

PHP 8.4.1

### Problem description

The page loads with the sidebar expanded by default on mobile...

https://github.com/user-attachments/assets/6ecf2e3b-3d79-4170-ac19-13b390665ef4

The reason, is that the sidebar's `isOpen` parameter is initialized as `true`:

https://github.com/filamentphp/filament/blob/0a867664655559540fbb18506002652caa2efd03/packages/panels/resources/js/index.js#L7

### Expected behavior

The page should load normally and not be obscured by the sidebar slideover. The sidebar navigation can be accessed by clicking on the hamburger button

### Steps to reproduce

1. Open new browser window with private/incognito mode, so `localStorage` is empty
2. Use a mobile browser, or use something like "Responsive Design Mode" to simulate narrow screen width
3. Navigate to any panel page
4. Observe that the page loads with sidebar expanded

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/filamentphp/demo

binaryfire:
Confirming this is still an issue with Filament 4 beta 5. I'm building a desktop app with NativePHP + Filament and the app always opens with the sidebar open.

danharrin:
Changed the v4 default, I am too afraid to do so in v3 though #16647

PR #16905, opened 2025-07-10
Revert default collapsed sidebars

binaryfire:
Reverts https://github.com/filamentphp/filament/pull/16647. See: https://github.com/filamentphp/filament/issues/16904.

This PR didn't solve the issue (which was related to the slideover, not the sidebar). The default for the sidebar should still be `open`.


danharrin:
The slideover and the sidebar are the same component, which is why the "open" state is linked. The original issue was that when you visited the site on mobile, the sidebar was open, and it is now closed. Resizing the browser window was not mentioned by the original reporter.

binaryfire:
Could you please take another look at the video he shared in https://github.com/filamentphp/filament/issues/15056? I'm pretty sure he was talking about the slideover always being open. He also said:

> The page should load normally and not be obscured by the sidebar slideover. The sidebar navigation can be accessed by clicking on the hamburger button

danharrin:
Yes, I watched it- before the PR, when you signed in on mobile the sidebar was open instead of closed.

PR #16909, opened 2025-07-10
fix: `->searchable()` Select without `->getOptionLabelUsing()` validation

ralphjsmit:
Previously, if you had a searchable `Select`, it would have worked to only provide a `->getSearchResultsUsing()`. This works for a situation where the `Select` is never filled in on mount, as the `Select` will always display the label that came from the `->getSearchResultsUsing()`. It might be good to restore this behavior as otherwise people will be running into validation errors without them having any clue what it's caused by.

### DIFF
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -604,7 +604,13 @@ class Select extends Field implements Contracts\CanDisableOptions, Contracts\Has
     {
         $state = null;
 
-        $label = $this->evaluate($this->getOptionLabelUsing, [
+        $callback = $this->getOptionLabelUsing;
+
+        if (! $withDefault && ! $callback) {
+            throw new Exception("Select field [{$this->getStatePath()}] must have a [getOptionLabelsUsing()] callback set.");
+        }
+
+        $label = $this->evaluate($callback, [
             'value' => function () use (&$state): mixed {
                 return $state = $this->getState();
             },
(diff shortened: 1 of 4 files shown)


packages/forms/src/Components/Select.php
```diff
@@ -1415,7 +1415,14 @@ public function getInValidationRuleValues(): ?array
             return array_keys($this->getOptionLabels(withDefaults: false));
         }
 
-        return blank($this->getOptionLabel(withDefault: false)) ? [] : null;
+        $optionLabel = $this->getOptionLabel(withDefault: false);
+
+        if ($this->isSearchable() && ! $optionLabel) {
```
danharrin: I don't think we can trust that an absence of the option label indicates an absence of `getOptionLabelUsing()`, it could be that the option is invalid and no label is returned. In which case, if its also searchable, this would mean the validation is skipped, I think.

To be honest, I think that we should probably alter `getOptionLabel()` and `getOptionLabels()` to throw an exception if `$withDefault` is `false` and `$this->getOptionLabelUsing` is null, instead of silently failing. Users should be forced to implement the correct callbacks otherwise validation is not possible.

danharrin (changes requested):

packages/forms/src/Components/Select.php
ralphjsmit: Hey @danharrin, I tested that, but the `setUp()` method already provides a default `getOptionLabelUsing()` and `getOptionLabelsUsing()` callbacks, so the `$this->getOptionLabelUsing` is never `null`, so the exception is never thrown. Could work around it by detecting a custom option label set using e.g. a boolean property, but that would become a bit dirty quickly I think. What do you think?

danharrin:
I'll take this one over and try and find the right solution. Thanks for the report though, just want to be very careful not to introduce any validation security issues.

danharrin:
@ralphjsmit can you review #16920 please when you have a chance?

PR #16911, opened 2025-07-10
Hide sidebar slideover on mobile without collapsing desktop by default

binaryfire:
_Edit: Please see my second comment. Just pushed a new version that keeps track of and restores desktop state on resizing._

Setting `isOpen` to `false` by default fixed [#15056](https://github.com/filamentphp/filament/issues/15056) (sidebar auto-opening on mobile) but introduced a issue where the sidebar always starts collapsed on desktop - which usually isn't desirable.

This PR keeps the sidebar open by default on desktop while preventing it from auto-opening on mobile by:
1. Adding a ResizeObserver that closes the sidebar when resizing to mobile (< 1024px)
2. Checking on init to close the sidebar if loading on a mobile device

Since desktop and mobile share the same `isOpen` state, resizing a desktop window with an expanded sidebar to mobile and back results in a collapsed sidebar. This affects users who frequently snap/resize windows (like me). Maybe a future improvement could implement separate `isOpen` states for the desktop sidebar and mobile slideover to solve that.

Behaviour with this PR:
[Dubb-2025-07-10 at 08.30.20.677 PM.webm](https://github.com/user-attachments/assets/de70f01e-556e-4fc3-a1ff-5851b1923fc1)


binaryfire:
Just pushed a new version. This is a full fix using a separate `isOpenDesktop` key to keep track of the desktop state when resizing:

[Dubb-2025-07-10 at 09.28.12.029 PM.webm](https://github.com/user-attachments/assets/adda0338-c149-4779-bbdc-a7976fe87342)

danharrin (approved):

danharrin:
Thanks

PR #16920, opened 2025-07-11
Refactor select option validation

danharrin:
Replaces #16909


ralphjsmit (commented):
Looks good to me @danharrin. The exception is surely helpful.

PR #16928, opened 2025-07-12
Fix datepicker cannot be enabled

### LINKED ISSUE #16927: Datepicker cannot be enabled after initial page load
### Package

filament/forms

### Package Version

v3.3.31

### Laravel Version

v12

### Livewire Version

v3.6

### PHP Version

PHP 8.4

### Problem description

It is not possible to enable a Datepicker via subsequent requests when the Datepicker is initially in disabled state. Example: A Datepicker should only be enabled when a specific Select option is selected.

I believe the problem comes from #16762. Especially the added wire:ignore in date-time-picker.blade.php. To solve the wire:ignore should be removed. To compensate the removed wire:ignore, the v4 PR #16817 should be rolled out for v3 aswell since it solves the source of the problem #16762 tried to solve.

I would be happy to create a PR for it.

### Expected behavior

It should be possible to enable the Datepicker

### Steps to reproduce

When using the reproduction repo, just visit the Sample page (only Page available). There is a Select and a datepicker. The Select controls the disabled state of the Datepicker. Not try to enable the datepicker.

https://github.com/skaesdorf/filament-reproduction/tree/datepicker-cannot-be-enabled

### Relevant log output

```shell

```

skaesdorf:
Fixes #16927 and applies #16817 to v3

### DIFF
--- a/packages/forms/resources/views/components/color-picker.blade.php
+++ b/packages/forms/resources/views/components/color-picker.blade.php
@@ -1,6 +1,4 @@
 @php
-    use Filament\Support\Facades\FilamentView;
-
     $isDisabled = $isDisabled();
     $isLive = $isLive();
     $isLiveOnBlur = $isLiveOnBlur();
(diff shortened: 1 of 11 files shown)


danharrin:
In my opinion this is too risky to merge into v3 since there are large changes in behaviour. I would like to try and find an alternative solution to the issue. I may reopen this if I can't find an alternative though

PR #16932, opened 2025-07-12
Add `refresh-page` event / listener

binaryfire:
Adds a `refresh-page` event/listener, similar to `refresh-sidebar` and `refresh-topbar`. 

Now that we have a `content()` method that accepts any combination of components, it's easy to end up with complex pages where `record->refresh()` or re-filling the form doesn't refresh everything that needs to be. This provides a simple way to nuke everything if you have to.

Not sure if this should be added to the docs or not. If it should be, let me know which section to put it in.


danharrin:
Please be aware that this will not nuke everything on the page, but I don't think it should do that either. This just acts like a standard Livewire render that does nothing

PR #16938, opened 2025-07-12
fix: ensure $format defaults to null in isoDate(Time) methods

stefket:
## Description

This pull request addresses a missing default value for the $format parameter in the following methods:

- isoDate
- isoDateTime

These methods are designed to use a fallback format when no $format is provided (i.e., when $format is null). However, until now, the $format parameter did not have a default value, which meant that calling these methods without explicitly passing a format would result in an ArgumentCountError.

This PR sets the default value of $format to null to align the implementation with the intended behavior.

The same issue recently occurred in the iso*Tooltip methods. See: #16771 

## Visual changes

There are no visual changes

## Functional changes

### DIFF
--- a/packages/infolists/src/Components/Concerns/CanFormatState.php
+++ b/packages/infolists/src/Components/Concerns/CanFormatState.php
@@ -83,9 +83,11 @@ trait CanFormatState
         return $this;
     }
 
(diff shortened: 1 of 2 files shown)


danharrin (approved):

danharrin:
Thanks

PR #16941, opened 2025-07-12
fix/accessibility fix for complex elements 3.x

alex-elivate:
# Filament Accessibility Fixes Progress

## Components for Both 3.x and 4.x

-   [x] Builder
-   [x] Checkbox list
-   [x] Key-value
-   [x] Radio
-   [x] Rich editor
-   [x] Date time picker (non-native only) (see notes)
-   [ ] Select (non-native only) (see notes)
-   [ ] Toggle (needs ID) (see notes)

## Components for 4.x Only

-   [ ] Code editor
-   [ ] Color picker
-   [ ] Livewire field
-   [ ] Modal table select
-   [ ] Repeater, simple repeater, table repeater
-   [ ] Slider
-   [ ] Table select
-   [ ] Tags input
-   [ ] One time code input (needs ID for label mapping)

## Notes
- All listed components are on the test results for the screenshot.
- **Date time picker** (non-native) label is already pointing to a readonly input that's inside a button. Chrome and Light House do not show an error. Added id's for sub components.
- **Select** (non-native) label is pointing to a hidden select but the element is Choices.js. No Chrome errors found to fix.
- **Toggle** already has an id. 

## Visual changes
Before:
image

After:
image

## Functional changes

### DIFF
--- a/packages/forms/resources/views/components/builder.blade.php
+++ b/packages/forms/resources/views/components/builder.blade.php
@@ -34,7 +34,11 @@
     $statePath = $getStatePath();
 @endphp
 
-<x-dynamic-component :component="$getFieldWrapperView()" :field="$field" labelTag="div">
+<x-dynamic-component
+    :component="$getFieldWrapperView()"
+    :field="$field"
+    labelTag="div"
+>
     <div
         x-data="{}"
         {{
--- a/packages/forms/resources/views/components/builder.blade.php
+++ b/packages/forms/resources/views/components/builder.blade.php
@@ -34,7 +34,11 @@
     $statePath = $getStatePath();
 @endphp
 
-<x-dynamic-component :component="$getFieldWrapperView()" :field="$field" label-tag="div">
+<x-dynamic-component
+    :component="$getFieldWrapperView()"
+    :field="$field"
+    label-tag="div"
+>
     <div
         x-data="{}"
         {{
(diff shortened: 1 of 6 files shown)


alex-elivate:
Should I try to address some of these Lighthouse issues for the select component as part of this PR? or should this be a separate PR?

image

I think I need to update select.js or use an x-init hook.
I also found an option `labelId` in Choices.js that will set the aria-labledby for the choices div. I can grab the value from the select element. I don't think this is a lighthouse issue, but it's there for the taking.
The choices example page does not have these errors. https://choices-js.github.io/Choices/  I can maybe use this to troubleshoot.

alex-elivate:
It turns out there is a labelId option for Choices.js just for this reason:

```labelId
Type: String Default: ``

Input types affected: select-one, select-multiple

Usage: The labelId improves accessibility. If set, it will add aria-labelledby to the choices element.```

One possibility is to get the label Id from the aria-labelledby attribute already set by the component. But I don't know if select.js is used anywhere else.  I guess I could add a safety check and only set labelId if aria-labelledby has a value?

danharrin:
Since there are lots of changes here, lets refocus on only handling v4 support for now. Once we get v4 support, we can backport this to v3 if there is time.

I will change the target branch to v4, and the first task that will need to be done is resolving any merge conflicts.

alex-elivate:
Please may I submit a fresh PR for v4 and close this one? There are other components in v4 I wish to test.

danharrin:
Yeah thats absolutely fine

PR #16943, opened 2025-07-13
V4 fix hiding columns in column manager when hidden

### LINKED ISSUE #16880: [v4] `->visible(false)` should hide column everywhere
### Package

filament/filament

### Package Version

v4.0.0-beta14

### Laravel Version

v12.20.0

### Livewire Version

v3.6.3

### PHP Version

8.3.19

### Problem description

When `->visible(false)` is set on the column, it is stil present in the column manager menu. It is activated in the column manager, but not toggleable or visible in the table.

### Expected behavior

When `->visible(false)` is set on a column, it should not be visible anywhere

### Steps to reproduce

Add `->visible(false)` and `->toggleable(isToggledHiddenByDefault: true)`  to a column (updated_at in the example)

https://github.com/dododedodonl/filament-issues-v4/tree/issue-visible-toggleable

### Relevant log output

```shell

```

archilex:
Fixes #16880 

Dan, as we talked about separately, this PR _for the most part_ fixes the issue where table columns that use the `->hidden()`method were showing up in the column manager.

However, the one area where this is not working is when a dev is hiding a column based on the `activeTab` AND the table columns have been reordered. When those two conditions are met, then when switching to another tab that has different (or no) logic for hiding columns based on the active tab, neither the columns, nor the manager update on the first click. 

I actually now know what the issue is and how to fix it, but I will do that in a separate PR later on as that scenario is more of an edge case compared to the wider issue this PR is fixing.

### DIFF
--- a/packages/panels/src/Resources/Concerns/HasTabs.php
+++ b/packages/panels/src/Resources/Concerns/HasTabs.php
@@ -51,6 +51,10 @@ trait HasTabs
     public function updatedActiveTab(): void
     {
         $this->resetPage();
+
+        $this->cachedDefaultTableColumnState = null;
+
+        $this->applyTableColumnManager();
     }
 
     public function generateTabLabel(string $key): string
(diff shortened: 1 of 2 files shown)


danharrin (approved):

danharrin:
Thanks!

PR #16949, opened 2025-07-13
Fix the column bug that uses formatStateUsing which returns an HtmlString

jeffersongoncalves:
When formatStateUsing returns an instance of Htmlable.

This was limiting the HTML text and causing it to break.

A check was added to Str::limit and Str::words to verify whether the state is an instance of Htmlable.

Bug demonstration with the image below.

Captura de tela 2025-07-13 184301

Using ysfkaya/phone-input PhoneColumn has a formatStateUsing which returns an HtmlString.

https://github.com/ysfkaya/filament-phone-input/blob/main/src/Tables/PhoneColumn.php#L52

Then I identified the problem with the check before doing Str::limit and Str::words on a state that is an HTML.

image

### DIFF
--- a/packages/infolists/src/Components/Concerns/CanFormatState.php
+++ b/packages/infolists/src/Components/Concerns/CanFormatState.php
@@ -373,12 +373,14 @@ trait CanFormatState
             $state = $state->getLabel();
         }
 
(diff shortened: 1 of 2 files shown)


danharrin (approved):

danharrin:
Thanks

Issue #16906, opened 2025-07-10
Different collapsed sidebar padding on Chrome vs Firefox


binaryfire:
### Package

filament/filament

### Package Version

4-beta15

### Laravel Version

12

### Livewire Version

3

### PHP Version

8.4

### Problem description

The spacing on the right of the collapsed sidebar is different on Chrome vs Firefox.

Chrome is uneven (right > left):
<img width="657" height="621" alt="Image" src="https://github.com/user-attachments/assets/442a2186-517c-457d-911a-0f7e8f323fd6" />

Firefox looks correct:

<img width="707" height="675" alt="Image" src="https://github.com/user-attachments/assets/495c5b2f-c831-44d1-8b21-a80ca3c9eb63" />

### Expected behavior

The spacing should be consistent on all browsers

### Steps to reproduce

1. Download repo
2. Check collapsed sidebar in the inspector on both browsers

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/binaryfire/filament-bug-reproduction

### Relevant log output

```shell

```

angus-mcritchie:
This is caused by the difference in how the browsers' scroll bars work, as the `.fi-sidebar-nav` has the `scrollbar-gutter: stable;` style applied to it.

I assume this was applied so that opening/closing a navigation group wouldn't cause it to jump/shift when the scroll bar is shown.

**Chrome**
<img width="253" height="211" alt="Image" src="https://github.com/user-attachments/assets/868df6d6-034e-4aa0-aec5-7ba97d23ea98" />

**Firefox**
<img width="515" height="233" alt="Image" src="https://github.com/user-attachments/assets/233ea756-fa9d-482d-aa08-a54a754a70cc" />

I'm unsure how the maintainers would like to handle this or even if it will be considered a bug, @binaryfire.

But here are some potential solutions:

- Try to unify the browser's scrollbar width
- Leave it how it is
    - I think it's fine how it is. But I'm not sure why it's an issue in your case. Can you elaborate?
- Remove `scrollbar-gutter: stable` and put up with jumping on Chrome when opening/closing some navigation groups

What do you think @danharrin ?

binaryfire:
> This is caused by the difference in how the browsers' scroll bars work, as the .fi-sidebar-nav has the scrollbar-gutter: stable; style applied to it.

[...]

PR #16957, opened 2025-07-14
Add `OverlayScrollbars` for prettier scrollbars

binaryfire:
A lot of modern apps implement less obtrusive, better looking scrollbars. This PR implements that in Filament using [OverlayScrollbars](https://kingsora.github.io/OverlayScrollbars/). It looks much better, adds some nice behaviour like fade outs, and solves [issues like this](https://github.com/filamentphp/filament/issues/16906) which are caused by inconsistent scrollbar styling between browsers.

I did some research into the options. The main ones are [perfect-scrollbar](https://github.com/mdbootstrap/perfect-scrollbar?ref=hackernoon.com) (more or less abandoned) and [simplebar](https://github.com/Grsmto/simplebar?ref=hackernoon.com) (pretty chunky).

OverlayScrollbars is used by Spotify, Jetbrains and Storybook and it's pretty lightweight. I think it's the best choice. This is a good deep-dive into the various options: https://hackernoon.com/the-scrollbar-customization-in-css-and-js-2024-update

His conclusion was the same:
> Since 2022, I’ve moved away from perfect-scrollbar (unmaintained for 2 years now) and simplebar (too heavy) to overlayscrollbars. It’s lightweight (only 29.2kB), performant (though subjective without actual measurements), supports mobile devices, supports tree-shaking, and offers numerous library integrations (i.e., React, Vue, Angular, Svelte, and Solid).

I've included 2 themes. You might want to tweak them a bit.

All it takes to apply these scrollbars is adding `data-overlayscrollbars-initialize` and either `fi-scrollbar` or `fi-scrollbar-thin`, so devs can easily use them in custom components and plugins too.

**_Scrollbars before:_**
image

image

**_Scrollbars after:_**
image

image


danharrin (changes requested):
I must be honest, this PR feels incredibly heavy. Is #16906 the only issue it solves?

binaryfire:
> I must be honest, this PR feels incredibly heavy. Is #16906 the only issue it solves?

I was planning on making this PR prior to that issue. The default scrollbars look pretty bad IMHO - that's the main reason people use this in their apps.

danharrin:
I'm honestly not a huge fan of replacing browser/OS defaults for something as integral as scrolling

binaryfire:
No worries. Up to you.

binaryfire:
@danharrin I've simplified things, but if you still have concerns would you be open to just adding `.fi-scrollbar` and `data-scrollbar` to scrollable elements? No JS or CSS - just hooks for devs to implement their own scrollbar solution if they want. That might be better anyway. This might be too opinionated for core.

danharrin:
What do you think of my further simplification @binaryfire? Also added the ability to add custom options from userland

danharrin:
Also, is there any reason we need the extra `fi-scrollbar` class if we use the `data-` attr to initialise anyway?

binaryfire:
@danharrin LGTM. And no, there's no longer any point in adding `fi-scrollbar`

flolanger:
@danharrin This change has introduced scrolling issues on MacOS (and possibly other operating systems too). Previously, the scroll bar was always visible to the user and separate from the table content. Now, the scroll bar disappears after being clicked and dragged for the first time. It also stays within the content, which means that you might click on the last record instead of the scroll bar (the scroll bar is pretty tiny now as well).

IMO, the scroll bars should stay as they were for accessibility reasons, and no additional package should be required for them.

https://github.com/user-attachments/assets/ab9bc3f7-22c8-4a47-a530-5de85bacfb12
https://github.com/user-attachments/assets/f134cc04-6948-442b-9ab6-d17a938df55f

binaryfire:
@flolanger That’s not a bug - it’s the way `OverlayScrollbars` is currently configured. All those things can be tweaked, but I also get your point re: preferring the default scrollbars.

@danharrin What do you think about allowing this to be enabled via a panel method? Eg. `->overlayScrollbars()`, default false. 

I also reckon fade out should be off in the default config. Needing to mouse over things to make them reappear isn’t intuitive.

danharrin:
I think if its configurable, it may as well just be a plugin tbh.

binaryfire:
No worries. There’d still need to be selectors which can be used to target scrollable elements though. All the scrollbar packages rely on that.

PR #16980, opened 2025-07-16
Suggestion: Truncate long labels in select

skaesdorf:
When a select contains long labels, the overflow is just cutted in the options list. When selected, the label wraps into multiple lines, increasing the size of the select.

I suggest to truncate the text whenever it is longer than the select itself.

Select before:
Screenshot 2025-07-16 124112

After:
Screenshot 2025-07-16 123915


danharrin (changes requested):
I agree that the options list scrolling is bad, but by default I think it should wrap onto new lines.

Please introduce a new method - `wrapOptionLabels(false)`, with a `canOptionLabelsWrap()` getter and `$canOptionLabelsWrap` property to enable the truncation behaviour opt-in.

Please ensure it is in the docs.

skaesdorf:
Done

danharrin (changes requested):
Please revert any code styling changes, I see lots in the JS file for example. `composer cs` might fix this

skaesdorf:
Done. Sorry that was an unintended auto format from my ide

danharrin (approved):

danharrin:
Thanks

PR #16982, opened 2025-07-16
fix: scroll to active sidebar item logic

ralphjsmit:
This PR adds the scroll to sidebar item logic to scroll to the current active sidebar item.


danharrin:
We have `scroll-sidebar.js` to do this already, so not really sure why we should add it to the view. If that is broken, please fix the JS file instead

PR #16984, opened 2025-07-16
Suggestion: Allow empty string as placeholder for select

skaesdorf:
Currently, when setting an empty string as placeholder for a Select,  it is overruled by select.js `this.placeholder || 'Select an option'`. 

I think Selects should respect any string (especially empty strings) as a valid placeholder.


danharrin (changes requested):
I think you should be able to use `placeholder(null)` instead of an empty string, its much cleaner.

As such, please completely remove the default placeholders from the JS file in all cases. If no placeholder is given, none is displayed.

skaesdorf:
Done.

This actually was my plan initially, but I thought the placeholder in js were ment to be some kind of fallback.

danharrin (approved):

danharrin:
Thanks

PR #16989, opened 2025-07-17
Add Illuminate\Support\HtmlString support as state to RichEditor component

### LINKED ISSUE #16988: [v4.x] RichEditor allow HtmlString cast
### Package

filament/filament

### Package Version

v4.0.0-beta16

### Laravel Version

v12.20.0

### Livewire Version

v3.6.3

### PHP Version

PHP 8.4.5

### Problem description

My project uses a HtmlString cast which cast many fields to `\Illuminate\Support\HtmlString` which helps in many places.

However, the RichEditor doesn't support this. And when I try to use it on these field I get the following error.

Image

### Expected behavior

Work without error by casting HtmlString object to a string.

### Steps to reproduce

1. Navigate to edit user
2. Enter text into the `bio` field
3. Try and save the form
4. You'll see the error

https://github.com/angus-mcritchie/filament-issue/tree/rich-editor-issue

### Relevant log output

```shell

```

angus-mcritchie:
## Description
Fixes #16988

## Visual changes
none.

## Functional changes

### DIFF
--- a/packages/forms/src/Components/RichEditor/StateCasts/RichEditorStateCast.php
+++ b/packages/forms/src/Components/RichEditor/StateCasts/RichEditorStateCast.php
@@ -4,6 +4,7 @@ namespace Filament\Forms\Components\RichEditor\StateCasts;
 
 use Filament\Forms\Components\RichEditor;
 use Filament\Schemas\Components\StateCasts\Contracts\StateCast;
+use Illuminate\Support\HtmlString;
 
 class RichEditorStateCast implements StateCast
 {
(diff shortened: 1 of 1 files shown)


packages/forms/src/Components/RichEditor/StateCasts/RichEditorStateCast.php
```diff
@@ -59,6 +60,10 @@ public function get(mixed $state): string | array
      */
     public function set(mixed $state): array
     {
+        if ($state instanceof HtmlString) {
```
danharrin, line 63: Should be `Htmlable` instead of `HtmlString` for consistency

danharrin (changes requested):

packages/forms/src/Components/RichEditor/StateCasts/RichEditorStateCast.php
angus-mcritchie: Good call, see #17016

PR #17003, opened 2025-07-17
feat: ability to set `->defaultFocusedDate()` when opening empty `DateTimePicker`

ralphjsmit:
When the state of a non-native `DateTimePicker` is empty, clicking the field to open the panel always causes the dropdown to open at the current date. This is not so useful for fields in the case that e.g. a "default" exists, but the field state was still null. For example, take a field on a model called `custom_starts_at`, which is only filled in when there is actually a custom start date _set_. However, right now the panel always open at the current date, which is not convenient for that situation.

```php
DateTimePicker::make('custom_starts_at')
   ->defaultFocusedDate(fn (SomeModel $record) => $record->someRelationship->starts_at)
```

This way, if the field is `null`, toggling the panel will open the panel at the `$record->someRelationship->starts_at`. 

I checked and merging this into the V4-branch should be easy. Also willing to test it on V4 if needed.

Thanks!


packages/forms/resources/views/components/date-time-picker.blade.php
```diff
@@ -222,7 +224,7 @@ class="grid grid-cols-[repeat(7,minmax(theme(spacing.7),1fr))] gap-1"
                                                 ! dayIsSelected(day) &&
                                                 focusedDate.date() !== day &&
                                                 ! dayIsDisabled(day),
-                                            'bg-gray-50 dark:bg-white/5':
+                                            'bg-gray-100 dark:bg-white/10':
```
danharrin, line 227: What is this change related to?

danharrin (changes requested):

packages/forms/resources/views/components/date-time-picker.blade.php
ralphjsmit: @danharrin To make it slightly more obvious which date is currently selected.

danharrin (changes requested):
Ideally this would be a v4 PR. Is that ok or detrimental to your plans?

ralphjsmit:
@danharrin That would be unhandy tbh, because the app it would be for we can only upgrade to V4 around November, and it was a quite important request..

danharrin:
@ralphjsmit before I merge this, can you please also make the v4 PR so that I don't have to worry about doing that myself

ralphjsmit:
@danharrin Created!

danharrin (approved):

Issue #16898, opened 2025-07-09
[v4] difference in behavior of TextEntry in schema on view or edit page


dododedodonl:
### Package

filament/filament

### Package Version

v4.0.0-beta15

### Laravel Version

v12.20.0

### Livewire Version

v3.6.3

### PHP Version

PHP 8.3.21

### Problem description

A TextEntry that shows a relationship behaves different on a view page then on a edit page.

View page:
<img width="735" height="184" alt="Image" src="https://github.com/user-attachments/assets/260b5705-0f14-4b89-80da-cebca7384f08" />

Edit page (same model):
<img width="744" height="160" alt="Image" src="https://github.com/user-attachments/assets/9a18a404-1602-4edb-be9f-200b17b82afc" />

### Expected behavior

List should be identical as it is the same model

### Steps to reproduce

Use the same TextEntry on the eidt or view page
```php
TextEntry::make('otherModels.name')
    ->placeholder('-')
    ->bulleted(),
```

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/dododedodonl/filament-issues-v4/tree/issue-text-entry-field

### Relevant log output

```shell

```

PR #17006, opened 2025-07-17
[4.x] Fixes difference in behavior of TextEntry in schema on view or edit page

Jehizkia:
## Description

This PR fixes issue #16898 where TextEntry components on edit resource pages fail to display relationship data.

**Root Cause**
When TextEntry retrieves relationships on edit pages, the relationship path includes an extra 'data' item 
(e.g., `['data', 'otherModels']`) compared to view pages which only contain the relationship (e.g., `['otherModels']`).

**The Fix**
The solution allows the loops to continue when encountering non-relationship segments like 'data':

```php
// Line 28 in getStateRelationship() method:

foreach (explode('.', $statePath ?? $this->getStateRelationshipName()) as $nestedRelationshipName) {
    if (! $record->isRelation($nestedRelationshipName)) {
        $relationship = null;

        continue; // ← This allows the loop to skip 'data' and continue to actual relationships
    }

    $relationship = $record->{$nestedRelationshipName}();
    $record = $relationship->getRelated();
}
```

```php
// Line 77 in getStateRelationshipResults() method:

while (count($relationships)) {
    $currentRelationshipName = array_shift($relationships);

    $currentRelationshipValue = $record->getRelationValue($currentRelationshipName);

    // ... Collection handling code ...

    if (! $currentRelationshipValue instanceof Model) {
        continue; // ← This allows the loop to skip 'data' and continue processing
    }

    // ... rest of the method
}
```

## Visual changes

Before fix:
Screenshot 2025-07-17 at 15 37 58

After fix:
Screenshot 2025-07-17 at 15 38 15

## Functional changes


danharrin:
I don't think this is the correct fix, the correct fix would be to somehow determine what the actual relationship path is instead of using the full state path. Maybe it should be the relative state path, I'm not sure. If you are not sure of the correct fix I'll just handle the issue instead.

Jehizkia:
Hi @danharrin 

Thanks for checking it out. I wasn't really sure if this would be the best approach. However, it would be no problem if you decide to take this issue further, as my knowledge about the state paths you mentioned is limited.

Thanks.

PR #17009, opened 2025-07-17
fix: `ToggleColumn` disabled state

ralphjsmit:
This PR fixes two issues with the `ToggleColumn` in disabled state:
- The toggle would not have the disabled styling (fixed by changing to element type `<button>` instead of `<div>`, same as the re-usable toggle component).
- The tooltip has been moved from the ~$buttonAttributes~ to the container attributes, and the `disabled` attribute has been removed on the container as it is already present on the `$buttonAttributes`. This structure mimics the V3 structure.


danharrin:
The element needs to be `<div>` as it could be nested within a `<button>`, if there is a `recordAction()` on the table. Please submit a PR with an alternative solution that does not involve changing that element type

PR #17011, opened 2025-07-17
Fix support for schema-qualified column names in TextConstraint ContainsOperator (PostgreSQL)

### LINKED ISSUE #16996: Incorrect handling of schema-qualified table names in Query Builder in PostgreSQL
### Package

filament/tables

### Package Version

v4.0.0

### Laravel Version

v12.20.0

### Livewire Version

_No response_

### PHP Version

PHP 8.4.8

### Problem description

When using a PostgreSQL table with a schema-qualified column (e.g., custom_schema.example_table.column_name), the current implementation in the 

Filament\Tables\Filters\QueryBuilder\Constraints\TextConstraint\Operators\ContainsOperator class 

incorrectly parses the table and column using:

`[$table, $column] = explode('.', $qualifiedColumn);`

This logic assumes only two-part identifiers (table.column) and fails when handling schema-qualified columns (schema.table.column). As a result, the query builder misinterprets the components, leading to incorrect SQL generation or runtime errors.

### Expected behavior

The operator should correctly support both:

- table.column
- schema.table.column

by preserving the full schema-qualified table name and isolating the column. Suggested fix:

```php
$parts = explode('.', $qualifiedColumn);

if (count($parts) === 3) {
    [$schema, $table, $column] = $parts;
    $table = "{$schema}.{$table}";
} elseif (count($parts) === 2) {
    [$table, $column] = $parts;
}
```

This ensures compatibility with Laravel models and queries that use PostgreSQL schemas other than public.

### Steps to reproduce

- Ensure your PostgreSQL table is located in a non-default schema, e.g. demo_schema.items.
- Define a Laravel model pointing to the schema-qualified table:

```php
namespace App\Models;

use Illuminate\Database\Eloquent\Model;

class Item extends Model
{
    protected $table = 'demo_schema.items'; // schema-qualified table name
}
```

- Add a Query Builder filter using a TextConstraint on a column like name:

```php
use Filament\Tables\Filters\QueryBuilder;
use Filament\Tables\Filters\QueryBuilder\Constraints\TextConstraint;
use Filament\Tables\Filters\Layout\FiltersLayout;

->filters([
    QueryBuilder::make()
        ->constraints([
            TextConstraint::make('name'),
        ]),
], layout: FiltersLayout::AboveContent)
```

- Load the page and use the "contains" filter with a value. It will trigger an error similar to:

```explode(): Passed parameter is not a valid string...
SQLSTATE[42P01]: Undefined table: 7 ERROR: missing FROM-clause entry for table "demo_schema"```

https://github.com/boufaka/filament-text-constraint-issue

### Relevant log output

```
SQLSTATE[42P01]: Undefined table: 7 ERROR: missing FROM-clause entry for table "demo_schema" LINE 1: ...aggregate from "demo_schema"."items" where (lower(demo_schem... ^ (Connection: pgsql, SQL: select count(*) as aggregate from "demo_schema"."items" where (lower(demo_schema.items::text)::text like %mat%))
```

### PR DESCRIPTION
**Pull request by boufaka**
## Description

This PR fixes an issue in Filament\Tables\Filters\QueryBuilder\Constraints\TextConstraint\Operators\ContainsOperator where schema-qualified columns (e.g., demo_schema.items.title) are incorrectly parsed when building PostgreSQL filters.

The code currently uses:

```php
[$table, $column] = explode('.', $qualifiedColumn);
```

This breaks when the column name is schema-qualified (schema.table.column) because explode() only supports two parts. It results in incorrect SQL and errors like:

```
SQLSTATE[42P01]: Undefined table: 7 ERROR: missing FROM-clause entry for table "demo_schema"
```

### DIFF
--- a/packages/tables/src/Filters/QueryBuilder/Constraints/TextConstraint/Operators/ContainsOperator.php
+++ b/packages/tables/src/Filters/QueryBuilder/Constraints/TextConstraint/Operators/ContainsOperator.php
@@ -69,7 +69,7 @@ class ContainsOperator extends Operator
             if (count($parts) === 3) {
                 [$schema, $table, $column] = $parts;
                 $table = "{$schema}.{$table}";
-            } elseif (count($parts) === 2) {
+            } else {
                 [$table, $column] = $parts;
             }
 
--- a/packages/tables/src/Filters/QueryBuilder/Constraints/TextConstraint/Operators/ContainsOperator.php
+++ b/packages/tables/src/Filters/QueryBuilder/Constraints/TextConstraint/Operators/ContainsOperator.php
@@ -75,7 +75,7 @@ class ContainsOperator extends Operator
 
             if (Str::lower($table) !== $table) {
                 $table = collect(explode('.', $table))
-                    ->map(fn ($segment) => "\"{$segment}\"")
+                    ->map(fn (string $segment): string => "\"{$segment}\"")
                     ->implode('.');
             }
 
--- a/packages/tables/src/Filters/QueryBuilder/Constraints/TextConstraint/Operators/ContainsOperator.php
+++ b/packages/tables/src/Filters/QueryBuilder/Constraints/TextConstraint/Operators/ContainsOperator.php
@@ -64,14 +64,23 @@ class ContainsOperator extends Operator
         $isPostgres = $databaseConnection->getDriverName() === 'pgsql';
 
         if ($isPostgres) {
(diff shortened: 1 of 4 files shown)


danharrin (changes requested):
Please also fix this for StartsWith, EndsWith and Equals, if the bug is also present there

boufaka:
> Please also fix this for StartsWith, EndsWith and Equals, if the bug is also present there

Done

danharrin (approved):

danharrin:
Thanks

PR #17021, opened 2025-07-18
feat(4.x): spa prefetching

datlechin:
## Description

Implements SPA prefetch functionality for Filament panels. When enabled, pages are automatically prefetched on hover using `wire:navigate.hover`, making navigation feel more responsive.

https://github.com/user-attachments/assets/6ac296c7-5b83-4550-b85c-c7962ef5c8ae

## Functional changes

### DIFF
--- a/docs/05-panel-configuration.md
+++ b/docs/05-panel-configuration.md
@@ -222,7 +222,7 @@ public function panel(Panel $panel): Panel
 
 SPA prefetching enhances the user experience by automatically prefetching pages when users hover over links, making navigation feel even more responsive. This feature utilizes [Livewire's `wire:navigate.hover` functionality](https://livewire.laravel.com/docs/navigate#prefetching-links).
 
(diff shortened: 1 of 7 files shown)


binaryfire:
I'd be careful of this because it'll really hammer your server and database (with even a small number of concurrent users). Personally I'd recommened against this - prefetching is mainly used for websites and not in admin panels. In particular, using it on datatables seems like a bad idea.

Having an option to enable it on a component-by-component basis might be useful. But IMHO providing a way to enable this for an entire panel seems like a bad practice in terms of performance / scalability.

Just my 2c

binaryfire:
Although one thing it would be _really_ good for is NativePHP apps, since they're running locally... So this would actually be really useful for that! Maybe just include a warning in the docs about prefetching increasing the server load.

danharrin (approved):

PR #17025, opened 2025-07-18
Allow direct editing per type in morphtoselect

zachweix:
## Description

Currently there is no way to modify the select for morphToSelect Types directly. Instead, you can hack `modifyKeySelectUsing` and then have an if-else statement to check the label: `if ($select->getLabel() == 'label X)`. This allows you to edit the select inline directly

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/forms/docs/03-fields/03-select.md
+++ b/packages/forms/docs/03-fields/03-select.md
@@ -401,6 +401,36 @@ MorphToSelect::make('commentable')
 
 > Many of the same options in the select field are available for `MorphToSelect`, including `searchable()`, `preload()`, `native()`, `allowHtml()`, and `optionsLimit()`.
 
+#### Customizing the select component for each morphed type
+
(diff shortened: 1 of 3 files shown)


danharrin:
We are only accepting new features in v4, so I have corrected the target branch.

Can you fix the merge conflicts please?

danharrin:
I have looked at the v4 implementation and I think it will be quite tricky due to how I have refactored it already.

I believe this is already possible though, if you use the existing `modifyKeySelectUsing()` method, and then pass `$state` and check `$state['commentable_type']` etc before making modifications.

Hope this helps!

zachweix:
I can create a new merge request on the version 4 branch. While it is indeed true that you can get this functionality from what you have now, you need to put a lot of if-elses in potentially. If you have many types, and each one wants its own custom changes, then that function can get bloated. Doing it this way allows you to set each type on its own

danharrin:
I don't think you will be able to get it to cleanly work in v4

PR #17033, opened 2025-07-19
[v4] Fix: Section collapse icon does not rotate when collapsed

malzariey:
## Description

The problem:

The retated button is not an immediate descendant of `.fi-collapsed` class.

The updated code removes the > and uses the descendant combinator:

```css
&.fi-collapsed {
    & .fi-section-collapse-btn {
        @apply rotate-180;
    }
}
```

----

Edit:

The updated code adds the missing child `<header/>`:

```css
&.fi-collapsed {
    & > header > .fi-section-collapse-btn {
        @apply rotate-180;
    }
}
```
## Visual changes

None

## Functional changes


danharrin (changes requested):
That would affect nested sections I think, can you please instead fix the selector to add whichever intermediate elements we are missing, `> .fi-section-header >` or something?

malzariey:
> That would affect nested sections I think, can you please instead fix the selector to add whichever intermediate elements we are missing, `> .fi-section-header >` or something?

@danharrin  done

danharrin (approved):

PR #17039, opened 2025-07-20
[4.x] Fix French (fr) translations

hugodebard:
Fix non-compliant french translations.

### DIFF
--- a/packages/actions/resources/lang/fr/edit.php
+++ b/packages/actions/resources/lang/fr/edit.php
@@ -13,7 +13,7 @@ return [
             'actions' => [
 
                 'save' => [
-                    'label' => 'Sauvegarder',
+                    'label' => 'Sauvegarder les modifications',
                 ],
 
             ],
@@ -23,7 +23,7 @@ return [
         'notifications' => [
 
             'saved' => [
-                'title' => 'Sauvegardé(e)',
+                'title' => 'Sauvegardé',
             ],
 
         ],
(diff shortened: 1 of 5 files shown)


danharrin (approved):

loanbesson:
Hello, I agree with the “Cancel” action, but I disagree with the others.
From an ergonomic point of view, it's incorrect.
When you save, you know that you are saving the changes, so there's no need to add more detail to this action.
And it reduces usability on mobile devices.

hugodebard:
> Hello, I agree with the “Cancel” action, but I disagree with the others. From an ergonomic point of view, it's incorrect. When you save, you know that you are saving the changes, so there's no need to add more detail to this action. And it reduces usability on mobile devices.

Yes, that makes sense. What I did was simply update the translations to match the English translations. @danharrin , what do you think? Would it be interesting to replace 'Save changes' with 'Save'?

danharrin:
I don't know French, its up to you

loanbesson:
> > Hello, I agree with the “Cancel” action, but I disagree with the others. From an ergonomic point of view, it's incorrect. When you save, you know that you are saving the changes, so there's no need to add more detail to this action. And it reduces usability on mobile devices.
> 
> Yes, that makes sense. What I did was simply update the translations to match the English translations. @danharrin , what do you think? Would it be interesting to replace 'Save changes' with 'Save'?

On mobile devices, the button is now large. I said that to maintain a certain level of usability, but it's not a big deal. I'll change the label for my save actions on all my pages.
Thanks for your contribution.

hugodebard:
> I don't know French, its up to you

I was asking you about English, if you find it relevant to replace 'save changes' with 'save'. And therefore do the same thing in French, replace 'Sauvegarder les modifications' with 'Sauvegarder'. This would shorten the button time and have better ergonomics.

danharrin:
No, I would like English to remain the same

PR #17040, opened 2025-07-21
Improve column manager handling with hidden closures

archilex:
This PR fixes the outstanding issue from #16943 where a column is hidden based on the activeTab AND the table columns have been reordered. This is done by reverting some of the changes in the previous PR and instead of removing the column if it's hidden, it is instead filtered out in blade/js. 

It also improves the default syncing of the `isToggledHiddenByDefault()` in that if a column is initially created and is NOT `toggleable()`, but then `->toggleable(isToggledHiddenByDefault: true)` is subsequently added by the dev, the column will be correctly toggled off on the next refresh. Note, once a column is `toggleable()`, changing  the boolean value of `isToggledHiddenByDefault` will NOT update the column's state in the UI as this would potentially override any toggling the user has done. Of course, resetting the column manager will set the column to whatever this default is.


danharrin (approved):

danharrin:
Thanks!

PR #17041, opened 2025-07-21
fix: stops radio buttons being squashed

jsandfordhughescoop:
## Description

If the description of a radio button is wider than the container, it squishes the radio button due to the radio button not having a min width. I have fixed this by adding an `aspect-square` which makes sure the width and height are always the same.

## Visual changes

Before
CleanShot 2025-07-21 at 08 02 12

After
CleanShot 2025-07-21 at 08 02 33

## Functional changes


danharrin:
Thanks! I had a look at `CheckboxList` which doesn't have this issue, and we used `shrink-0` there, so I copied that instead for consistency.

danharrin (approved):

PR #17044, opened 2025-07-21
Update casts syntax in docs

MadBox-99:
## Description

Laravel 11 and later will be remove the attribute overiding, instead use function like in
 https://laravel.com/docs/eloquent-mutators#attribute-casting

## Functional changes

### DIFF
--- a/docs-assets/app/app/Models/Post.php
+++ b/docs-assets/app/app/Models/Post.php
@@ -10,9 +10,12 @@ class Post extends Model
 {
     use HasFactory;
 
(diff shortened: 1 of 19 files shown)


danharrin (changes requested):
Please completely remove the Laravel 10 version as we don't support that in Filament 4. No need for comments either

Please find every usage of `$casts` in the docs, as there are probably quite a few that need updating for consistency.

MadBox-99:
Done!

danharrin (approved):

Issue #16974, opened 2025-07-15
[v4.x] Severe performance degradation in  V4 when using $get() in deeply nested components


alancolant:
### Package

filament/filament

### Package Version

v4.x

### Laravel Version

v12.0.0

### PHP Version

8.4

### Problem description

In Filament v4, I've encountered a critical performance issue when using the `$get()` utility within deeply nested form structures — particularly when using `Builder` components that contain several layers of `Section`, `Grid`, or similar containers.

When `$get('some.path')` is used in these scenarios (e.g., inside a `visible()` closure or dynamic label), the `getSchemaComponent()` method appears to recurse excessively. This leads to drastically slower form submissions — sometimes taking **several minutes** for a relatively simple form with just 5 Builder blocks containing 2 nested sections each.

**Root Cause Analysis**
The bottleneck appears to stem from the following snippet inside the `$get()` utility logic

**Impact**
This issue makes Filament v4 impractical for production forms with dynamic content. 
Even relatively small forms become unusable and need several minutes to save.

### Expected behavior

**restores normal performance**.

---

**Proposed Solution**

It seems that passing `skipComponentChildContainersWhileSearching` causes `HasComponents::getComponent()` to recurse excessively, especially with large or deeply nested schemas. Benchmarks clearly indicate that this is the primary source of the degradation.

```php
$component = $livewire->getSchemaComponent(
    $this->component->resolveRelativeKey($key, $isAbsolute),
    withHidden: true,
    skipComponentChildContainersWhileSearching: $this->component,
);
```

Replacing it with:

```php
$component = $livewire->getSchemaComponent(
    $this->component->resolveRelativeKey($key, $isAbsolute),
    withHidden: true,
);
```

Alternatively, add a safeguard that checks whether the resolved key starts with the current component’s absolute key:

   ```php
   if (str_starts_with($resolvedKey, $this->component->getKey(absolute: true))) {
      ...
   }
   ```

In practice, this case seems unlikely and may be redundant for the `$get()` context.

### Steps to reproduce

1. Create a `Builder` with multiple blocks.
2. Within each block, nest multiple `Sections` or `Grids` (2–3 levels deep).
3. Add a visibility condition or similar logic using:

   ```php
   ->visible(fn (Get $get) => $get('field'))
   ```
4. Attempt to submit the form — observe the delay.
5. Remove `skipComponentChildContainersWhileSearching` → performance becomes acceptable.

### Reproduction repository (issue will be closed if this is not valid)

[...]

PR #17062, opened 2025-07-22
[4.x] Fix/recursive component performances

alancolant:
## Description

This PR aims to rewrite getComponent to handle the recursive traversal of components much more efficiently and significantly faster.

A detailed explanation can be found in the corresponding issue: #16974 

With a large number of nested components, such as when using the Builder, there no longer appears to be any noticeable performance impact — it now performs similarly to a smaller schema.

## Functional changes

- [x] Changes have been tested to not break existing functionality.  **(Based on my experience, I haven't observed any problems in any of my cases; however, other situations may indeed exist.)**

### DIFF
--- a/packages/schemas/src/Concerns/HasComponents.php
+++ b/packages/schemas/src/Concerns/HasComponents.php
@@ -12,6 +12,7 @@ use Filament\Schemas\Components\Text;
 use Illuminate\Contracts\Support\Htmlable;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
+use Illuminate\Support\Str;
 
 trait HasComponents
 {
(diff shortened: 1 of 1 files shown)


danharrin (changes requested):
Hey, I have merged #16981 which addresses most of the issue.

I would still like to explore whether or not we should go further with merging this in, but I would really like to better understand the code and evaluate whether or not it will cause issues in finding components in other places.

Would you mind adding a lot of comments to the code you've changed, explaining what is happening and also providing examples of each variable's content at each point?

At the same time, can you fix the new merge conflicts caused by #16981?

Many thanks!

alancolant:
Closed because my changes did not significantly improve performance. The code’s readability remains poor, and no tangible benefits were identified.

PR #17070, opened 2025-07-22
Change hasRoutes to hasRoute

jeffersongoncalves:
Follow the same pattern configured in the [filament/actions](https://github.com/filamentphp/filament/blob/4.x/packages/actions/src/ActionsServiceProvider.php#L28) package, which uses ->hasRoute('web').

Since it's just a single file, it would be a good idea to use ->hasRoute('web') instead of the other ->hasRoutes('web') method.

Or, if you prefer, change the filament/actions package to follow a standard configuration for the project's packages.

### DIFF
--- a/packages/panels/src/FilamentServiceProvider.php
+++ b/packages/panels/src/FilamentServiceProvider.php
@@ -52,7 +52,7 @@ class FilamentServiceProvider extends PackageServiceProvider
                 Commands\MakeThemeCommand::class,
                 Commands\MakeUserCommand::class,
             ])
-            ->hasRoutes('web')
+            ->hasRoute('web')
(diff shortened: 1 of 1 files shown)


danharrin:
Can you add a PR description please so I can understand why you are suggesting this?

jeffersongoncalves:
Updated the PR description as per the details mentioned in the service provider file change.

danharrin:
Please use `hasRoutes()` everywhere; even though there is only one file it contains multiple routes

jeffersongoncalves:
Opened PR #17075 with the same pattern using ->hasRoutes('web') to be standardized between filament packages.

danharrin:
Thanks

PR #17080, opened 2025-07-23
fix: Support HasEmbeddedView components for partialRendering

pxlrbt:
## Description

Partial rendering broke because it tries to call `->render()` on the component which does not exist on `HasEmbeddedView` components.

## Functional changes

### DIFF
--- a/packages/schemas/src/Components/Concerns/CanPartiallyRender.php
+++ b/packages/schemas/src/Components/Concerns/CanPartiallyRender.php
@@ -69,12 +69,8 @@ trait CanPartiallyRender
                 throw new Exception('A [key()] or [statePath()] is required to partially render a component.');
             }
 
-            $render = $this instanceof HasEmbeddedView
-                ? $this->toEmbeddedHtml(...)
-                : $this->render(...);
-
             return [
-                "schema-component::{$key}" => $render(),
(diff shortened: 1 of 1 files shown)


danharrin (changes requested):
If the component has a view, it should still call `render()`. I guess there is a method that does this check already, is it `toHtml()`?

pxlrbt:
Ah, great. I updated the PR.

danharrin (approved):

packages/schemas/src/Components/Concerns/CanPartiallyRender.php
```diff
@@ -70,7 +69,7 @@ public function partiallyRender(): void
             }
 
             return [
-                "schema-component::{$key}" => $this->toHtml(),
+                "schema-component::{$key}" => $this->hasView() ? $this->render() : $this->toHtml(),
```
pxlrbt, line 72: Wondering why you changed that? `->toHtml()` calls `$this->render()->render()` if there is a view

danharrin: It still needs to be a view object if its present, because we need to modify the view compiler to inject `$this` properly

PR #17093, opened 2025-07-24
[v4] Add option to pass additional arguments to Table Select Livewire component

dvarilek:
## Description

In the ModalTableSelect component, there is currently no way to modify the table based on the current state of other fields in the form.  This PR addresses this and adds the ability to pass additional arguments to the Livewire component hosting this table.

## Example usage

Suppose there is an Invoice that can contain multiple Services. Both Invoice and Services have some Category. However, this Invoice can only contain Services from the same Category. So the Table's query should get automatically scoped to contain only the correct Services with that Category.

```php
ModalTableSelect::make('services')
    ->relationship('services', 'name')
    ->multiple()
    ->tableConfiguration(ServicesTable::class)
    ->tableSelectArguments(function (Get $get) {
        return [
            'category_id' => $get('category_id'), // Suppose there is a category_id field
        ];
    })
```
```php
class ServicesTable
{
    public static function configure(Table $table): Table
    {
        return $table
            ->modifyQueryUsing(function (Builder $query, TableSelectLivewireComponent $livewire) {
                if ($categoryId = $livewire->tableSelectArguments['category_id'] ?? null) {
                    $query->where('category_id', $categoryId);
                }
                     
                return $query;
            })
           // ...
    }
}
```
## Functional changes


danharrin (approved):

danharrin:
great idea, many thanks

PR #17110, opened 2025-07-28
fix: convert `defaultFocusedDate()` to correct timezone

ralphjsmit:
I noticed I made one small mistake in #17003, because I assumed the timezone handling was in done in JS because of the `.tz(timezone)` part, but that appears actually to be the current "guessed" timezone from the browser. This PR fixes that by applying the `->setTimezone()` method on the default focused date.

Thanks!

### DIFF
--- a/packages/forms/src/Components/DateTimePicker.php
+++ b/packages/forms/src/Components/DateTimePicker.php
@@ -433,7 +433,7 @@ class DateTimePicker extends Field implements Contracts\HasAffixActions
     {
         $defaultFocusedDate = $this->evaluate($this->defaultFocusedDate);
 
-        if ($timezone = $this->getTimezone()) {
+        if (filled($defaultFocusedDate) && ($timezone = $this->getTimezone())) {
             $defaultFocusedDate->timezone($timezone);
         }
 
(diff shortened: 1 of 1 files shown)


packages/forms/src/Components/DateTimePicker.php
```diff
@@ -431,7 +431,13 @@ public function getMinDate(): ?string
 
     public function getDefaultFocusedDate(): ?string
     {
-        return $this->evaluate($this->defaultFocusedDate);
+        $defaultFocusedDate = $this->evaluate($this->defaultFocusedDate);
+
+        if (filled($defaultFocusedDate)) {
+            $defaultFocusedDate = $defaultFocusedDate->setTimezone($this->getTimezone());
```
danharrin, line 449: This assumes its a Carbon instance, when it doesn't need to be, right? Please check that

danharrin (changes requested):

packages/forms/src/Components/DateTimePicker.php
ralphjsmit: Good point, done.

danharrin (approved):

Issue #15717, opened 2025-02-25
Form FileUpload ->disabled not reacting


juanborras:
### Package

filament/forms

### Package Version

v3.3.0

### Laravel Version

v12.0.0 (Also with v11)

### PHP Version

8.2

### Problem description

FileUpload does not react to disabled() although it does to hidden()

I've setup a small app that adds 'photo' and 'enable_photo' attributes to the User model.

Inside the Form from the UserResource I've added two FileUpload components. Both use the 'enable_photo' to check whether or not enable/hide the Component.

### Expected behavior

When the 'enable photo' Toggle is toggled, only the FileUpload with the hidden() reacts to the value change of the Toggle (which has live(onBlur: true)). 

If the default value in the User model is set to 'false' the first load of the form is correct. When 'enable_photo' is set to 0, the first FileUpload appears disabled and the second one is hidden. If 'enable_photo' is set to 1, both FileUpload appears enabled. When the Toggle is toggled, only the FileUpload with the hidden() reacts.

### Steps to reproduce

Select Users, click on the toggle 'Enable photo'

https://github.com/user-attachments/assets/fd058eb7-acd6-4bb6-813d-cec5dd52cdbe

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/juanborras/fileupload-test.git

### Relevant log output

```shell
commited with sqlite db. User user@localhost.local, pass user
```

danharrin:
Hi, we are using FilePond for the file upload field, and like the other JS-powered fields in Filament, you cannot modify their properties after they are first loaded, since the configuration is already loaded into the JS component. This is a known issue and is not currently solvable, apologies.

gustavocaiano:
> This is a known issue and is not currently solvable, apologies.

Hi @danharrin, thank you for the clarification.
Would there be any way to manually trigger a reload of the FileUpload component when no longer disabled? I'm curious to know how others facing this issue have managed to work around it, as I'm currently unsure how to proceed. Any guidance would be greatly appreciated. Thank you!

gustavocaiano:
@juanborras have you got any workaround to that problem? In my specific case i can't use the hidden because the field gotta be there already.

danharrin:
I think I have an idea that I can try

juanborras:
> @juanborras have you got any workaround to that problem? In my specific case i can't use the hidden because the field gotta be there already.
> 
> 
> 

[...]

PR #17142, opened 2025-07-31
Dynamically disable JS inputs for v.3

### LINKED ISSUE #17363: FileUpload field with reactive disabled() closure doesn't re-render when dependent field changes
### Package

filament/forms

### Package Version

v3.3.36 

### Laravel Version

v12.24.0

### Livewire Version

v3.0

### PHP Version

PHP  8.3.6 

### Problem description

When using a `FileUpload` field with a `disabled()` closure that depends on another field (like `is_active`), the field doesn't re-render when the dependent field changes.

## Actual Behavior
The FileUpload field doesn't re-render, maintaining its previous state.

## Environment
- Filament Version: 3.3.36
- Laravel Version: [Your Laravel version]
- PHP Version: [Your PHP version]
- Browser: [Your browser]

## Code Example
```php
Forms\Components\Checkbox::make('is_active')
    ->label('Active'),

Forms\Components\FileUpload::make('image')
    ->disabled(fn($get) => !$get('is_active'))
    ->label('Image'),
```

## Additional Notes
This affects the user experience as users expect the form to be reactive to their input changes.

### Expected behavior

The FileUpload field should become enabled/disabled and re-render when the checkbox changes.

### Steps to reproduce

1. Create a form with a checkbox field (e.g., `is_active`)
2. Add a `FileUpload` field with `disabled(fn($get) => !$get('is_active'))`
3. Toggle the checkbox
4. The FileUpload field remains in its previous state instead of updating

https://github.com/Mohammedalhamami/filament-fileupload-disabled-bug

### Relevant log output

```shell

```

TheSkybreaker:
## Description

Implemented the fix of v.4 in #16891 for v.3
This means adding 

```blade
wire:key="{{ $this->getId() }}.{{ $statePath }}.{{ $field::class }}.{{
    substr(md5(serialize([
        $isDisabled,
    ])), 0, 64)
}}"
```
to `FileUpload`, `KeyValue`, `RichEditor` and `Select` as suggested in #15717 

## Functional changes

Reactive disabled should be working now.

- [] Changes have been tested to not break existing functionality - I tested only for the file upload 


packages/forms/resources/views/components/rich-editor.blade.php
```diff
@@ -377,6 +377,11 @@ class="trix-button trix-button--dialog"
                     @endif
                     x-ref="trix"
                     wire:ignore
+                    wire:key="{{ $this->getId() }}.{{ $statePath }}.{{ $field::class }}.{{
+                        substr(md5(serialize([
+                            $isDisabled,
```
danharrin, line 383: Can you test the PR please? I don't think this works because this variable is never assigned like it is in the other files, it is still a getter closure.

danharrin (changes requested):

packages/forms/resources/views/components/rich-editor.blade.php
TheSkybreaker: You're right, thanks
I added the `$isDisabled` variable and used it to be consistent with the other files in the latest commit

danharrin (approved):

PR #17145, opened 2025-07-31
fix: spacing within `RepeatableEntry`

ralphjsmit:
Currently there is no spacing between `RepeatableEntry`'s as the class `fi-in-repeatable` is applied to the `ul`, and the `ul` is therefore not a child of it.

### DIFF
--- a/packages/infolists/resources/css/components/repeatable.css
+++ b/packages/infolists/resources/css/components/repeatable.css
@@ -1,7 +1,5 @@
 .fi-in-repeatable {
-    &ul {
-        @apply gap-4;
-    }
+    @apply gap-4;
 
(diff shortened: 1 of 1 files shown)


danharrin (changes requested):
`&ul {` is not a child selector, so I don't really understand what the problem would be? `&ul {` applies to the parent element if it is a `ul`

ralphjsmit:
@danharrin Okay, hmm, let me double check that why it's then not working for me. For the `RepeatableEntry`, on your side does it have the correct gap class applied?

danharrin:
Yeah seems fine for me. Please open an issue if you can't work it out.

Screenshot 2025-08-01 at 07 47 29

PR #17158, opened 2025-08-01
fix: default state for simple `Repeater` fields

ralphjsmit:
The default state for simple repeater fields is already wrapped in the simple field name in the `->default()` method:

```php
parent::default(function (Repeater $component) use ($state) {
    $state = $component->evaluate($state);

    $simpleField = $component->getSimpleField();

    $items = [];

    foreach ($state ?? [] as $itemData) {
        if ($simpleField) {
            $itemData = [$simpleField->getName() => $itemData];
        }

        if ($uuid = $component->generateUuid()) {
            $items[$uuid] = $itemData;
        } else {
            $items[] = $itemData;
        }
    }

    $component->hydratedDefaultState = $items;

    return $items;
});
```

Therefore, when merging the hydrated default state with the items state we don't need to apply this wrapping again, otherwise it gets wrapped double.

### DIFF
--- a/packages/forms/src/Components/Repeater.php
+++ b/packages/forms/src/Components/Repeater.php
@@ -1019,8 +1019,6 @@ class Repeater extends Field implements CanConcealComponents, HasExtraItemAction
         $state = $this->getRawState();
         $items = $this->hydratedDefaultState;
 
-        $simpleFieldName = $this->getSimpleField()?->getName();
-
         foreach ($items as $itemKey => $itemData) {
             $items[$itemKey] = [
                 ...$state[$itemKey] ?? [],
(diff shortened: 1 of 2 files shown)


danharrin:
Can you please open an issue? I think this is there for a reason.

ralphjsmit:
OK, the alternative solution is to not apply the wrapping in the `->default()` method instead. Will try to create a quick repo.

danharrin:
This part of the codebase is incredibly fragile, as it needs to handle `defaultItems()` and `default()`, as well as deeply nested combinations of each

ralphjsmit:
Makes sense, would be good to add some tests handling common scenarios here.

danharrin:
I think we do have some in RepeaterTest, I might be wrong

danharrin (approved):

PR #17165, opened 2025-08-01
Merge extraAttributes in Action when postToUrl() is used

stefket:
## Description

This PR ensures that the extraAttributes() defined on the Action are applied to the <form> element when postToUrl() is used — just as they would be applied to the element rendered otherwise.

## Visual changes

No visual changes.

## Functional changes


danharrin:
This isn't correct since the trait should be usable without any other class methods existing on it already - hence the many parameters in the signature.

Can you please let me know exactly which extra attribute is not working when you use `postToUrl()`?

stefket:
@danharrin I need an action that uses postToUrl and visually looks like an outlined button.

However, using `$action->postToUrl(...)->button()->outlined()` doesn’t work, because it no longer renders as a form.

That’s why I thought about using something like `$action->postToUrl(...)->extraAttributes(['class' => 'fi-btn fi-outlined'])` to manually apply the outlined button styling.

danharrin:
That's fixed in #17182

PR #17178, opened 2025-08-03
Improve: fa translation in register.php

alisalehi1380:
image

@ariaieboy


ariaieboy:
It's a better translation, according to the screenshot you uploaded.

PR #17184, opened 2025-08-04
[BREAKING CHANGE IN BETA]: More friendly URL parameters

danharrin:


larsbo:
after this change the form tab is not activated on page load when using https://filamentphp.com/docs/4.x/resources/managing-relationships#combining-the-relation-manager-tabs-with-the-form.

opheus2:
I think this has introduced a bug. Tabs URL no longer work as expected.

Most likely introduced by #17185 

https://drive.google.com/file/d/1Zp7D-1VaJw5AAUApGVAln-QLE6yR-fL7/view?usp=sharing

danharrin:
https://github.com/filamentphp/filament/pull/17199

PR #17195, opened 2025-08-04
[v4] Fix table "select all" action with `selectCurrentPageOnly` enabled

dvarilek:
## Description
This PR fixes select all action in table incorrectly selecting all records when configured to only select records on the current page with `selectCurrentPageOnly()`. This happened because `selectAllRecords` always inverted tracking of selected records.

### Before

https://github.com/user-attachments/assets/1f9e008b-a78e-4efd-92dc-123ccb238e13

### After
https://github.com/user-attachments/assets/3cfcde14-0717-412a-80aa-35ac064bf011

## Functional changes


packages/tables/resources/views/index.blade.php
```diff
@@ -587,7 +588,7 @@ class="fi-ta-selection-indicator"
                                 color="primary"
                                 tag="button"
                                 x-on:click="selectAllRecords"
-                                :x-show="'canSelectAllRecords(' . \Illuminate\Support\Js::from($selectsCurrentPageOnly) . ')'"
+                                x-show="canSelectAllRecords"
```
danharrin: Just for consistency, does this work too?

```suggestion
                                x-show="canSelectAllRecords()"
```

danharrin (changes requested):

danharrin (approved):

PR #17207, opened 2025-08-06
[4.x] Use laravel provided namespace

steelvirus:
## Description
Currently, filament does not allow for custom namespaces in (some) commands and defaults, like the make:filament-panel command, it will always use `App\\` as prefix. Some commands allow for a custom namespace to be provided as an option. When using a namespace different from `App\\`, for example when developing a plugin/package (which is my use case) it would be a real improvement DX wise. Laravel itself already allows for custom namespaces (and paths). 

## Visual changes
No visual changes

## Functional changes
This change makes filament use the namespace (prefix) from the laravel app-instance (`::getNamespace`) and uses it as default, replacing hardcoded `App\\` occurrences. ~~The namespace can also be set using the `::setRootNamespace` method on the FilamentManager (Filament facade).~~ ~~The class also provides a `namespaceFor` method: `Filament::namespaceFor('Models\\')`.~~

- [x] No failing tests

This PR introduces a not-small-amount of changes, but simple ones, review should be straight forward ❤️

### DIFF
--- a/docs-assets/app/app/Providers/Filament/AdminPanelProvider.php
+++ b/docs-assets/app/app/Providers/Filament/AdminPanelProvider.php
@@ -32,12 +32,12 @@ class AdminPanelProvider extends PanelProvider
             ->colors([
                 'primary' => Color::Amber,
             ])
-            ->discoverResources(in: app_path('Filament/Resources'), for: app()->getNamespace() . 'Filament\\Resources')
-            ->discoverPages(in: app_path('Filament/Pages'), for: app()->getNamespace() . 'Filament\\Pages')
+            ->discoverResources(in: app_path('Filament/Resources'), for: "'" . app()->getNamespace() . "Filament\\Resources'")
+            ->discoverPages(in: app_path('Filament/Pages'), for: "'" . app()->getNamespace() . "Filament\\Pages'")
             ->pages([
                 Pages\Dashboard::class,
             ])
-            ->discoverWidgets(in: app_path('Filament/Widgets'), for: app()->getNamespace() . 'Filament\\Widgets')
+            ->discoverWidgets(in: app_path('Filament/Widgets'), for: "'" . app()->getNamespace() . "Filament\\Widgets'")
             ->widgets([
                 Widgets\AccountWidget::class,
                 Widgets\FilamentInfoWidget::class,
--- a/docs-assets/app/app/Providers/Filament/AdminPanelProvider.php
+++ b/docs-assets/app/app/Providers/Filament/AdminPanelProvider.php
@@ -2,7 +2,6 @@
 
 namespace App\Providers\Filament;
 
-use Filament\Facades\Filament;
 use Filament\Http\Middleware\Authenticate;
 use Filament\Http\Middleware\AuthenticateSession;
 use Filament\Http\Middleware\DisableBladeIconComponents;
@@ -33,12 +32,12 @@ class AdminPanelProvider extends PanelProvider
             ->colors([
                 'primary' => Color::Amber,
             ])
-            ->discoverResources(in: app_path('Filament/Resources'), for: Filament::namespaceFor('Filament\\Resources'))
-            ->discoverPages(in: app_path('Filament/Pages'), for: Filament::namespaceFor('Filament\\Pages'))
+            ->discoverResources(in: app_path('Filament/Resources'), for: app()->getNamespace() . 'Filament\\Resources')
+            ->discoverPages(in: app_path('Filament/Pages'), for: app()->getNamespace() . 'Filament\\Pages')
             ->pages([
                 Pages\Dashboard::class,
             ])
-            ->discoverWidgets(in: app_path('Filament/Widgets'), for: Filament::namespaceFor('Filament\\Widgets'))
+            ->discoverWidgets(in: app_path('Filament/Widgets'), for: app()->getNamespace() . 'Filament\\Widgets')
             ->widgets([
                 Widgets\AccountWidget::class,
                 Widgets\FilamentInfoWidget::class,
--- a/docs-assets/app/app/Providers/Filament/AdminPanelProvider.php
+++ b/docs-assets/app/app/Providers/Filament/AdminPanelProvider.php
@@ -2,6 +2,7 @@
 
 namespace App\Providers\Filament;
 
+use Filament\Facades\Filament;
 use Filament\Http\Middleware\Authenticate;
 use Filament\Http\Middleware\AuthenticateSession;
 use Filament\Http\Middleware\DisableBladeIconComponents;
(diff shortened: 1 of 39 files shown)


danharrin:
Hi, you mentioned this would be part of the Laravel configuration, I did not intend to approve additional configuration on the Filament side. Does Laravel not support configuration for picking a new `App\` namespace already?

steelvirus:
> Hi, you mentioned this would be part of the Laravel configuration, I did not intend to approve additional configuration on the Filament side. Does Laravel not support configuration for picking a new `App\` namespace already?

Hey :)

Yes, that is exactly what the Filament::getRootNamespace method uses by default, the App::getNamespace method. I understand not wanting additional configuration options (in this case, overwriting the laravel namespace when wanted) on the filament side. 

I'll remove the getRootNamespace and setRootNamespace methods, so only ::namespaceFor remains, removing added configuration from the filament side.

danharrin (changes requested):
If `::namespaceFor()` just concatenates `app()->getNamespace()`, I think we can remove that as well and just concatenate it in each place.

docs-assets/app/app/Providers/Filament/AdminPanelProvider.php
```diff
@@ -32,12 +33,12 @@ public function panel(Panel $panel): Panel
             ->colors([
                 'primary' => Color::Amber,
             ])
-            ->discoverResources(in: app_path('Filament/Resources'), for: 'App\\Filament\\Resources')
-            ->discoverPages(in: app_path('Filament/Pages'), for: 'App\\Filament\\Pages')
+            ->discoverResources(in: app_path('Filament/Resources'), for: Filament::namespaceFor('Filament\\Resources'))
+            ->discoverPages(in: app_path('Filament/Pages'), for: Filament::namespaceFor('Filament\\Pages'))
```
danharrin: Also, no generated files should contain calls to `Filament::namespaceFor()` or `app()->getNamespace()` - we should insert the namespace at the time of file generation. Otherwise it looks messy in the user's codebase, especially for 99.9% of users who do not use this feature.

steelvirus: Yes, makes sense, thanks for the input, will update.

steelvirus:
> If `::namespaceFor()` just concatenates `app()->getNamespace()`, I think we can remove that as well and just concatenate it in each place.

Allright, removing it :)

packages/panels/src/Commands/FileGenerators/PanelProviderClassGenerator.php
```diff
@@ -120,12 +120,12 @@ public function generatePanelMethodBody(): string
                     ->colors([
                         'primary' => {$this->simplifyFqn(Color::class)}::Amber,
                     ])
-                    ->discoverResources(in: app_path('Filament/{$componentsDirectory}Resources'), for: 'App\\Filament\\{$componentsNamespace}Resources')
-                    ->discoverPages(in: app_path('Filament/{$componentsDirectory}Pages'), for: 'App\\Filament\\{$componentsNamespace}Pages')
+                    ->discoverResources(in: app_path('Filament/{$componentsDirectory}Resources'), for: '{$this->simplifyFqn(app()->getNamespace() . "Filament\\{$componentsNamespace}Resources")}')
```
danharrin: Why did you wrap this in `simplifyFqn()`?

danharrin (changes requested):

packages/panels/src/Commands/FileGenerators/PanelProviderClassGenerator.php
steelvirus: Initially thought this makes sense in regards to the other places it was used, but makes no sense here. Gone

packages/panels/src/Commands/FileGenerators/PanelProviderClassGenerator.php
```diff
@@ -120,12 +122,12 @@ public function generatePanelMethodBody(): string
...
                     ])
-                    ->discoverResources(in: app_path('Filament/{$componentsDirectory}Resources'), for: 'App\\Filament\\{$componentsNamespace}Resources')
-                    ->discoverPages(in: app_path('Filament/{$componentsDirectory}Pages'), for: 'App\\Filament\\{$componentsNamespace}Pages')
+                    ->discoverResources(in: app_path('Filament/{$componentsDirectory}Resources'), for: '{$rootNamespace}Filament\\{$componentsNamespace}Resources')
+                    ->discoverPages(in: app_path('Filament/{$componentsDirectory}Pages'), for: '{$rootNamespace}Filament\\{$componentsNamespace}Pages')
                     ->pages([
                         {$this->simplifyFqn(Dashboard::class)}::class,
                     ])
-                    ->discoverWidgets(in: app_path('Filament/{$componentsDirectory}Widgets'), for: 'App\\Filament\\{$componentsNamespace}Widgets')
+                    ->discoverWidgets(in: app_path('Filament/{$componentsDirectory}Widgets'), for: '{$this->simplifyFqn(app()->getNamespace() . "Filament\\{$componentsNamespace}Widgets")}')
```
danharrin: As above

danharrin (changes requested):

packages/panels/src/Commands/FileGenerators/PanelProviderClassGenerator.php
steelvirus: Sorry lol, not generally blind :D

packages/panels/src/Commands/FileGenerators/PanelProviderClassGenerator.php
```diff
@@ -120,12 +122,12 @@ public function generatePanelMethodBody(): string
...
                     ])
-                    ->discoverResources(in: app_path('Filament/{$componentsDirectory}Resources'), for: 'App\\Filament\\{$componentsNamespace}Resources')
-                    ->discoverPages(in: app_path('Filament/{$componentsDirectory}Pages'), for: 'App\\Filament\\{$componentsNamespace}Pages')
+                    ->discoverResources(in: app_path('Filament/{$componentsDirectory}Resources'), for: '{$rootNamespace}Filament\\{$componentsNamespace}Resources')
+                    ->discoverPages(in: app_path('Filament/{$componentsDirectory}Pages'), for: '{$rootNamespace}Filament\\{$componentsNamespace}Pages')
                     ->pages([
                         {$this->simplifyFqn(Dashboard::class)}::class,
                     ])
-                    ->discoverWidgets(in: app_path('Filament/{$componentsDirectory}Widgets'), for: 'App\\Filament\\{$componentsNamespace}Widgets')
+                    ->discoverWidgets(in: app_path('Filament/{$componentsDirectory}Widgets'), for: '{$rootNamespace}Filament\\{$componentsNamespace}Widgets")}')
```
danharrin:
```suggestion
                    ->discoverWidgets(in: app_path('Filament/{$componentsDirectory}Widgets'), for: '{$rootNamespace}Filament\\{$componentsNamespace}Widgets')
```

danharrin (changes requested):

danharrin (approved):

PR #17220, opened 2025-08-07
Add HTML rendering support to ModalTableSelect labels

sawirricardo:
## Description
Introduces the ability to render option labels as HTML in the ModalTableSelect component by using the CanAllowHtml concern and updating the Blade view to conditionally render HTML when allowed.

This is to have feature parity with Select Component which also has allow html for its option label

## Visual changes

Before
image

After
image

## Functional changes

Documentation should be still the same with this section https://filamentphp.com/docs/4.x/forms/select#allowing-html-in-the-option-labels

### DIFF
--- a/packages/forms/resources/views/components/modal-table-select.blade.php
+++ b/packages/forms/resources/views/components/modal-table-select.blade.php
@@ -6,6 +6,7 @@
     $id = $getId();
     $isDisabled = $isDisabled();
     $isMultiple = $isMultiple();
+    $isHtmlAllowed = $isHtmlAllowed();
 @endphp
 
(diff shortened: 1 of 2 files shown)


danharrin:
You can use an `HtmlString` or `View` object to return HTML instead of needing to toggle another configuration option.

sawirricardo:
@danharrin I had used HtmlString, but didn't work...

danharrin:
Please open an issue or submit a fix for that instead then, but we don't need to introduce a new method for this one

sawirricardo:
noted... thanks

PR #17223, opened 2025-08-07
Add false to boolean icon allowing default icon to be disabled

howdu:
Allow the ability to pass false to a boolean icon column/entry and not display the default icon. 

```php
IconColumn::make('featured')
    ->trueIcon('heroicon-o-star')
    ->falseIcon(false)
```

Before - no way of disabling the default icon
Screenshot 2025-08-07 at 07 48 53

After
Screenshot 2025-08-07 at 07 49 28

Added `false` so that it does not affect previous behaviour, where both will still show the default icon.
```php
IconColumn::make('featured')
    ->trueIcon('heroicon-o-star'),
IconColumn::make('featured')
    ->trueIcon('heroicon-o-star')
    ->falseIcon(null)
```

## Functional changes

### DIFF
--- a/packages/infolists/src/Components/IconEntry.php
+++ b/packages/infolists/src/Components/IconEntry.php
@@ -36,14 +36,14 @@ class IconEntry extends Entry implements HasEmbeddedView
      */
     protected string | array | Closure | null $falseColor = null;
 
-    protected string | BackedEnum | Closure | null $falseIcon = null;
+    protected string | BackedEnum | Closure | false | null $falseIcon = null;
 
     /**
      * @var string | array<string> | Closure | null
      */
     protected string | array | Closure | null $trueColor = null;
 
-    protected string | BackedEnum | Closure | null $trueIcon = null;
+    protected string | BackedEnum | Closure | false | null $trueIcon = null;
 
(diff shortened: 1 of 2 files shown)


packages/infolists/src/Components/IconEntry.php
```diff
@@ -190,9 +196,15 @@ public function getTrueColor(): string | array
         return $this->evaluate($this->trueColor) ?? 'success';
     }
 
-    public function getTrueIcon(): string | BackedEnum
+    public function getTrueIcon(): string | BackedEnum | false
```
danharrin: Instead of returning `false` from this and `getFalseIcon()`, I think instead return `null`

packages/infolists/src/Components/IconEntry.php
```diff
@@ -273,11 +285,17 @@ public function toEmbeddedHtml(): string
         <div <?= $attributes->toHtml() ?>>
             <?php foreach ($state as $stateItem) { ?>
                 <?php
+                    $icon = $this->getIcon($stateItem);
```
danharrin: Please check indentation

danharrin (changes requested):

danharrin (approved):

PR #17230, opened 2025-08-07
Fix render HTML on option label in modal table select

sawirricardo:
## Description

Currently using HtmlString in getOptionLabelFromRecordUsing in Modal Table Select only render regular string instead of HTML

## Visual changes
Before
image
After
image

## Functional changes

### DIFF
--- a/packages/forms/src/Components/ModalTableSelect.php
+++ b/packages/forms/src/Components/ModalTableSelect.php
@@ -523,7 +523,7 @@ class ModalTableSelect extends Field
         return $this->getOptionLabelFromRecordUsing !== null;
     }
 
-    public function getOptionLabelFromRecord(Model $record): string | Htmlable | null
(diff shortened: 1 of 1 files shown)


packages/forms/src/Components/ModalTableSelect.php
```diff
@@ -523,7 +523,7 @@ public function hasOptionLabelFromRecordUsingCallback(): bool
         return $this->getOptionLabelFromRecordUsing !== null;
     }
 
-    public function getOptionLabelFromRecord(Model $record): string
+    public function getOptionLabelFromRecord(Model $record): string | Htmlable | null
```
danharrin:
```suggestion
    public function getOptionLabelFromRecord(Model $record): string | Htmlable
```

danharrin (changes requested):

danharrin (approved):

PR #17238, opened 2025-08-08
Fix PostgreSQL JSON field search syntax error (fixes #17237)

### LINKED ISSUE #17237: Bug: Filament v3.3.31 Breaks Search Queries on JSON Fields in PostgreSQL
### Package

filament/filament

### Package Version

v3.3.35

### Laravel Version

v12.21.0

### Livewire Version

v3.6.4

### PHP Version

8.2

### Problem description

**Regression in v3.3.31:** PostgreSQL search query generation fails for translatable fields, generating malformed JSON path operators. This is a general issue affecting PostgreSQL JSON field searches, not specifically related to Spatie\Translatable.

**Regression details:**
- ✅ Works correctly in v3.3.30
- ❌ Broken from v3.3.31 onwards
- Issue affects PostgreSQL JSON field search queries

**SQL Error:**
```
SQLSTATE[42601]: Syntax error: 7 ERROR: syntax error at or near "ar" 
LINE 1: ...d "publishing_status" = $2 and (lower("name"->>'>'ar''::text... ^
```

### Expected behavior

PostgreSQL search should generate valid SQL for JSON fields as it did in v3.3.30:

### Steps to reproduce

1. Set up Laravel project with PostgreSQL database
2. Upgrade Filament from v3.3.30 to v3.3.31 or later
3. Create any model with JSON/translatable fields
4. Create a Filament Resource with searchable field that uses JSON:
   ```php
   TextInput::make('name')
       ->label(__('resource.product.fields.name'))
       ->required()
       ->maxLength(255)
       ->searchable()  // or use table search
   ```

5. Perform a search on the field
6. SQL syntax error occurs in PostgreSQL

**Workaround:** Downgrade to v3.3.30 resolves the issue completely.

https://github.com/maytham553/ilament-postgresql-json-search-bug

### Relevant log output

```shell
SQLSTATE[42601]: Syntax error: 7 ERROR: syntax error at or near "ar" 
LINE 1: ...d "publishing_status" = $2 and (lower("name"->>'>'ar''::text... ^ 
(Connection: pgsql, SQL: select count(*) as aggregate from "products" where "products"."vendor_id" in (01jwr1ttfae9dbyh10nabhr6xd) and "publishing_status" = 2 and (lower("name"->>'>'ar''::text)::text like %asdf% or lower("name"->>'>'ar''::text)::text like %asdf% or lower("name"->>'>'ar''::text)::text like %asdf% or lower("group_id"::text)::text like %asdf%))
```

maytham553:
## Description
This PR fixes a bug with searching JSON fields in PostgreSQL. Before, search did not work and gave an error. Now, it works again.

**Fixes #17237**

- The problem started in version 3.3.31.
- The fix checks if the column already has `->>` and does not change it again.
- No changes for MySQL, SQLite, or SQL Server.

## Visual changes
No visual changes

## Functional changes

### DIFF
--- a/packages/support/src/helpers.php
+++ b/packages/support/src/helpers.php
@@ -165,10 +165,7 @@ if (! function_exists('Filament\Support\generate_search_column_expression')) {
 
         $column = match ($driverName) {
             'pgsql' => (
-                str($column)->contains('->>')
-                    ? $column
-                    : (
-                        str($column)->contains('->')
+                str($column)->contains('->')
                             ? (
                                 // Handle `table.field` part with double quotes
                                 str($column)
@@ -178,6 +175,18 @@ if (! function_exists('Filament\Support\generate_search_column_expression')) {
                                     ->implode('.')
                             ) . collect(str($column)->after('->')->explode('->')) // Handle JSON path parts
                                 ->map(function ($segment, $index) use ($column): string {
+                                    // if segment already contains '>something' (from ->> operator), preserve it
+                                    $isExplicitOperatorPrefixed = str($segment)->startsWith('>');
+                                    $segment = $isExplicitOperatorPrefixed ? (string) str($segment)->after('>') : $segment;
+
+                                    // Remove single quotes from segment if present to avoid redundant quoting
+                                    $isWrappedWithSingleQuotes = str($segment)->startsWith("'") && str($segment)->endsWith("'");
+                                    $segment = $isWrappedWithSingleQuotes ? (string) str($segment)->trim("'") : $segment;
+
+                                    if ($isExplicitOperatorPrefixed) {
+                                        return "->>'{$segment}'";
+                                    }
+
                                     $totalParts = substr_count($column, '->');
 
                                     return ($index === ($totalParts - 1))
@@ -189,7 +198,6 @@ if (! function_exists('Filament\Support\generate_search_column_expression')) {
                                 ->explode('.')
                                 ->map(fn (string $part): string => (string) str($part)->wrap('"'))
                                 ->implode('.')
-                    )
             ) . '::text',
             default => $column,
         };
--- a/packages/support/src/helpers.php
+++ b/packages/support/src/helpers.php
@@ -176,7 +176,7 @@ if (! function_exists('Filament\Support\generate_search_column_expression')) {
                                     ->explode('.')
                                     ->map(fn (string $part): string => (string) str($part)->wrap('"'))
                                     ->implode('.')
-                            ).collect(str($column)->after('->')->explode('->')) // Handle JSON path parts
+                            ) . collect(str($column)->after('->')->explode('->')) // Handle JSON path parts
                                 ->map(function ($segment, $index) use ($column): string {
                                     $totalParts = substr_count($column, '->');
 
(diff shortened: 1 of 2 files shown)


danharrin (changes requested):
Thanks.

This just skips most of the other transformation rules if `->>` already exists.

Would it not be better to preserve the behaviour of wrapping `table.field` in double quotes etc, even when `->>` is used?

Maybe your check for `->>` could be moved inside the `->map()` that is now on line 180?

danharrin:
@glennjacobs would you mind providing your thoughts too?

glennjacobs:
> @glennjacobs would you mind providing your thoughts too?

Will do. Just travelling back from holiday, so can look Monday.

maytham553:
Hi @danharrin,
Just to clarify — are you suggesting doing something like this within the `map`?

```php
->map(function ($segment, $index) use ($column): string {
    // If the segment already starts with '>', preserve it (used for ->> operator)
    if (str($segment)->startsWith('>')) {
        return '->' . $segment;
    }

    $totalParts = substr_count($column, '->');

    return ($index === ($totalParts - 1))
        ? "->>'{$segment}'"
        : "->'{$segment}'";
})
```

Let me know if I understood your idea correctly or if you had something else in mind.

danharrin:
@maytham553 thats the right place I think, but do you not need to check if it contains `->>` already or something?

maytham553:
@danharrin,

No — when `->>` is present, splitting by `'->'` removes the first `->`, so the segment we check only starts with a single `>`.
Example:

```php
$column = 'data->>name';

explode('->', $column);
// [
//   "data",
//   ">name" // starts with ">", so we detect it here
// ]
```

That’s why this already works:

```php
// If the segment starts with '>', it came from a '->>' operator.
// The first '->' was removed by explode(), so we re-add it.
if (str($segment)->startsWith('>')) {
    return '->' . $segment; // restores ->>
}
```

I’m not sure if there are any other edge cases I might be missing — what idea did you have in mind?

danharrin:
Okay that sounds good then, can you make those changes to the PR please?

maytham553:
@danharrin 

I’ve applied the changes and optimized the mapping logic as discussed. The segment now properly handles explicit `->>` operators and removes any redundant single quotes from segments before adding the JSON operators.


Everything tested and pushed.

packages/support/src/helpers.php
```diff
@@ -166,26 +166,38 @@ function generate_search_column_expression(string $column, ?bool $isSearchForced
...
+                            ? (
+                                // Handle `table.field` part with double quotes
+                                str($column)
+                                    ->before('->')
+                                    ->explode('.')
+                                    ->map(fn (string $part): string => (string) str($part)->wrap('"'))
+                                    ->implode('.')
+                            ) . collect(str($column)->after('->')->explode('->')) // Handle JSON path parts
+                                ->map(function ($segment, $index) use ($column): string {
+                                    // if segment already contains '>something' (from ->> operator), preserve it
```
danharrin:
```suggestion
                                    // If segment already starts with `>` (from `->>` operator), preserve it
```

danharrin (changes requested):

danharrin (approved):

danharrin:
Thanks @maytham553! Considering this is well tested I will merge it in, but Glenn can still provide feedback if he wants once he has a chance.

maytham553:
Thanks! — happy to adjust if @glennjacobs  spots anything.

glennjacobs:
> Thanks! — happy to adjust if @glennjacobs spots anything.

It all appears to be working great for me, cheers.

Issue #17024, opened 2025-07-18
[v4.x] dropdown isn’t anchored to the button when inside a registered modal action


archilex:
### Package

filament/filament

### Package Version

v4.0.0-beta17

### Laravel Version

v12.20.0

### PHP Version

PHP 8.3.0

### Problem description

When rendering actions with forms in an action's `modalContent()`, the nested action's form's fields appear in the parent's modal. Additionally, when using an `actionGroup` the dropdown is not anchored to the button.

<img width="858" height="324" alt="Image" src="https://github.com/user-attachments/assets/7729d14a-2529-4631-a681-054b8bac4401" />

<img width="1468" height="622" alt="Image" src="https://github.com/user-attachments/assets/a42a7396-1513-4898-b0c3-fef542de3d72" />

### Expected behavior

The nested form fields should not be displayed in the parent modal and the dropdown should be anchored to the button

### Steps to reproduce

Go to the User Resource and click on the header actions. The first one shows the textInput being displayed in the modal. The second shows the same, but since it's rendered in the a dropdown, the dropdown is not achored.

```php
protected function getHeaderActions(): array
  {
      return [
          Action::make('modalContentTest')
              ->registerModalActions([
                  $this->testAction()
              ])
              ->modalContent(view('modal-content-test')),
      ];
  }

  public function testAction(): Action
  {
      return Action::make('test')
          ->schema([
              TextInput::make('test') // This input appears in the parent modal when opening the modal
                  ->label('Test Input')
                  ->required()
          ])
          ->action(fn (array $data) => dump($data));
  }
```
```php
{{-- modal-content-test.blade.php --}}
<div>
    {{ $this->testAction }}
</div>
```

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/archilex/v4-modal-content-bug

### Relevant log output

```shell

```

danharrin:
As discussed yesterday, I think `->getModalAction()` needs to be called from the view to retrieve the correct Action instance properly, so it gets mounted as a nested action instead of a top-level action

archilex:
I just tested and you are correct! Thanks! However, the dropdown is still incorrectly positiion and not achored to the button. Do prefer I open a new issue, or just change the name of this one?

danharrin:
If you wouldn't mind changing the name that would be great

archilex:
[...]

PR #17251, opened 2025-08-09
Fix dropdowns inside of modals

stefket:
## Description

This pull request fixes an issue where dropdowns inside modals were not displayed in the correct position.
See the related issue for more details: [filamentphp/filament#17024](https://github.com/filamentphp/filament/issues/17024)

### Solution
The CSS property `contain: layout;` was added to the `.fi-modal-window` selector.
Since no corresponding Tailwind utility was found for this property, it has been implemented using plain CSS.

### Note
I wasn’t previously familiar with the `contain: layout;` property and only discovered it with the help of AI. However, all my tests have passed, and I couldn’t find any problems caused by this change.

## Functional changes


danharrin (approved):

stefket:
Thanks for the follow-up work 😊 @danharrin

PR #17261, opened 2025-08-11
fix: add helper methods to get Livewire ID

mansoorkhan96:
This pull request adds helper methods for schema fields to get the livewire id to avoid using `$this` in non object context.

## Problem

Fields like Select and CheckboxList, use `$this` object to get the Livewire ID. When passing these two fields to a custom Filter, the `$this` object is not available in the view files where it gets the Livewire ID using `$this->getId()`. Therefore, it throws an exception.

**Example:**

```php
Filter::make('name')
    ->schema([
        Select::make('operator')
            ->label('Operator')
            ->options([
                'contains' => 'Contains',
                'starts_with' => 'Starts with',
                'ends_with' => 'Ends with',
                'equals' => 'Equals',
                'not_contains' => 'Does not contain',
            ])
            ->default('contains')
            ->live()
            ->required(),
        TextInput::make('name'),
    ])
```

## Solution

This pull request adds helper methods for schema fields to get the Livewire ID to avoid using `$this` in non object context. So, now instead of using `$this->getId()`, now we can use `$getLivewireId()`.

### DIFF
--- a/packages/forms/resources/views/components/checkbox-list.blade.php
+++ b/packages/forms/resources/views/components/checkbox-list.blade.php
@@ -12,6 +12,7 @@
     $options = $getOptions();
     $livewireKey = $getLivewireKey();
     $wireModelAttribute = $applyStateBindingModifiers('wire:model');
+    $livewireId = $getLivewireId();
 @endphp
 
 <x-dynamic-component :component="$fieldWrapperView" :field="$field">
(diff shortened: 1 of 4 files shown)


danharrin:
Please open an issue for why you can't render the fields inside a custom filter, there must be something going wrong there for `$this` to not be injected properly.

mansoorkhan96:
My bad.
I tried to reproduce this in separate test app and these are the reasons it would happen.

1) If you force `4.0.0-beta14` (because i have issues with some plugins if i upgrade)
2) and if you use above mentioned Filter along with TernaryFilter

PR #17280, opened 2025-08-12
Check plugins compatibility when upgrading to v4

CharlieEtienne:
This PR introduces a new artisan command that will check installed plugins compatibility with Filament v4.

```bash
php artisan filament:check-plugins-compatibility-with-v4
```

Should print a table like this in the console:

image


packages/upgrade/src/Commands/CheckPluginsCompatibilityWithV4.php
```diff
@@ -0,0 +1,208 @@
...
+        return null;
+    }
+
+    /**
+     * Checks if a package requires filament/filament
+     *
+     * @param  string  $package  The full package name
+     * @return bool True if the package requires filament/filament
+     */
+    protected function packageRequiresFilament(string $package): bool
```
saade: Perhaps this can be refactored to look into `vendor/{$package}/composer.json` to avoid too many requests on the Packagist server? Since it's a simple check whether the package requires Filament or not.

CharlieEtienne: great suggestion, I will do that!

CharlieEtienne: Fixed in https://github.com/filamentphp/filament/pull/17280/commits/7fbac9d8f6988bb407681c8000a1e133c91d6bbf

CharlieEtienne:
As suggested by @danharrin on Discord, I updated the upgrade script to include this check.

It checks for incompatible plugins and exit upgrade script if any incompatible package is found:

Capture d’écran 2025-08-13 à 02 13 48

<hr>

If all packages are compatible, it continues the upgrade normally:

Capture d’écran 2025-08-13 à 02 40 17

CharlieEtienne:
Let me know if you want to keep the separate command, or just the script.

danharrin (changes requested):
Please remove the separate command as you mentioned

CharlieEtienne:
> Please remove the separate command as you mentioned

Done in [f82ab72](https://github.com/filamentphp/filament/pull/17280/commits/f82ab72075f04b22d467dff473e9aee0fb21cffc)

danharrin (approved):

danharrin:
Thanks, I added a few more experimental things to see what people think. I don't think the Tailwind command I added is ready to be added to the docs yet, but it would be good if people could test it. It freezes for me when I run it on the Filament demo for some reason.

CharlieEtienne:
That's good addition! I'll try to test it soon

urufudev:
Hi, will this command also be available in Filament v3?  
I think many users who are still on version 3 would like to check plugin compatibility before deciding to upgrade to v4.

danharrin:
The v4 upgrade scripts runs in Filament 3.

archilex:
@CharlieEtienne Hi! This script seems to fail with private/paid plugins as they aren't accessible through packgist. Any way this could be updated to either skip paid plugins or allow the dev to whitelist them in some sort of config?

danharrin:
@archilex is an exception thrown? If there is an error, it should just be skipped in the entire checking process.

archilex:
No...no exception is thrown. `$json = @file_get_contents($url);` returns false so it just gets thrown into the $incompatiblePlugins array

danharrin:
Maybe the `@` should be removed? I think that silences the exception

PR #17291, opened 2025-08-13
customizing the color of the Repeater Action button

Odilbukh:
This PR introduces support for customizing the color of the Repeater component in Filament.

### Changes
- Added a new `color()` method to allow setting custom colors for the Repeater header and UI elements.
- Defaults to the existing theme color if no custom color is specified.
- Maintains backward compatibility with existing Repeater usage.

### Motivation
Currently, the Repeater component's color is fixed and cannot be adjusted without overriding styles. This update enables developers to visually distinguish Repeaters in forms, improving UX and flexibility in UI customization.

### Example
Repeater::make('items')
    ->color('primary')

### BC Breaks
None.

### Related Issues
N/A

## Functional changes

### DIFF
--- a/packages/actions/resources/lang/it/delete.php
+++ b/packages/actions/resources/lang/it/delete.php
@@ -68,7 +68,6 @@ return [
             'missing_processing_failure_message' => ':count non possono essere eliminati.',
         ],
 
-
     ],
 
 ];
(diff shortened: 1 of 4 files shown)


danharrin:
`->addAction(fn (Action $action) => $action->color('success')`

We don't want to introduce duplicate configuration methods for everything that is possible, that is why these methods exist

PR #17293, opened 2025-08-13
Allow for file generation when no model is selected

### LINKED ISSUE #17287: [4.x] Error when create table widget without select models
### Package

filament/filament

### Package Version

v4.0.0

### Laravel Version

v12.23.1

### Livewire Version

v3.0.0

### PHP Version

PHP 8.4.11

### Problem description

When creating a new Table Widget using the command `php artisan filament:widget` and in the `What is the model?` column, I did not select any model, so an error occurred

```bash
 Nette\InvalidArgumentException 

  Value ‘’ is not a valid class/function/constant name.

  at vendor/nette/php-generator/src/PhpGenerator/PhpNamespace.php:102
     98▕                if (
     99▕                        !Helpers::isNamespaceIdentifier($name, allowLeadingSlash: true)
    100▕                        || (Helpers::isIdentifier($name) && isset(Helpers::Keywords[strtolower($name)]))
    101▕                ) {
  ➜ 102▕                        throw new Nette\InvalidArgumentException(“Value ‘$name’ is not a valid class/function/constant name.”);
    103▕ 
    104▕                } elseif ($alias && (!Helpers::isIdentifier($alias) || isset(Helpers::Keywords[strtolower($alias)]))) {
    105▕                        throw new Nette\InvalidArgumentException(“Value ‘$alias’ is not a valid alias.”);
    106▕                }

      +19 vendor frames 

  20  artisan:16
      Illuminate\Foundation\Application::handleCommand()
```

Image

### Expected behavior

The Table widget is created normally without a model.

### Steps to reproduce

Create new laravel with filament project

```bash
laravel new filament-widgets
cd filament-widgets
composer require filament/filament:"^4.0"
php artisan filament:install --panels
php artisan filament:widget
```

Image

https://github.com/vermaysha/filament-summarize-customdata

### Relevant log output

```shell

```

spcbfr:
## Description
Closes #17287

This PR allows for stub generation commands such as `filament:widget` to generate even when no model is provided. initially I was going to simply make the model field required, but decided against that to match the behavior of Version 3.x

## Functional changes

### DIFF
--- a/packages/tables/src/Commands/FileGenerators/Concerns/CanGenerateModelTables.php
+++ b/packages/tables/src/Commands/FileGenerators/Concerns/CanGenerateModelTables.php
@@ -22,7 +22,7 @@ trait CanGenerateModelTables
 
         return <<<PHP
             return \$table
-                ->query(fn (): {$this->simplifyFqn(Builder::class)} => {$this->simplifyFqn($model)}::query())
+                ->query({$this->outputQueryBody($model)})
                 ->columns([
                     {$this->outputTableColumns($model)}
                 ])
@@ -216,5 +216,14 @@ trait CanGenerateModelTables
         }
 
         return implode(PHP_EOL . '        ', $columns);
+    }
+    public function outputQueryBody(?string $model = null): string {
+        if(blank($model)) {
+            return "";
+        }
+
+        return (string) new  Literal(<<<PHP
+        fn (): {$this->simplifyFqn(Builder::class)} => {$this->simplifyFqn($model)}::query()
+        PHP);
     }
 }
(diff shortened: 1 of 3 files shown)


danharrin (changes requested):
Instead of needing to make all these changes, just default the model to be `Illuminate\Database\Eloquent\Model` if it is blank. Then the user can replace it with their own model when they want. But `query()` is not valid without a parameter.

The only change that needs to be made is in `MakeWidgetCommand.php` I think

spcbfr:
Updated, although I don't understand why this is better since you'd just get a "cannot instantiate Model class" error instead of "->query() cannot be empty"

danharrin:
Less typing to correct it

danharrin (approved):

PR #17297, opened 2025-08-13
fix: allow getting tenant menu items even if the tenant menu has been…

olliescase:
… disabled

## Description

In Filament v4, if you load the tenant menu and you have done `->tenantMenu(false)` in your panel configuration, the register tenant button will not show.

I assume this was intended as a performance fix, but it regressed functionality, making filament that little bit less customizable.

This allows the tenant menu items to be fetched even if the tenant menu is disabled, while still not affecting the performance of those who do not with to show the menu anywhere.

## Why

For my application, I use the render hooks in order to inject the tenant menu into the user menu items. I don't want it to register in the default location, so I did `tenantMenu(false)` to disable it.

In Filament v3, this worked fine, I could call upon the livewire component in a blade template and register this blade template in a Render Hook, in order to then show.

However, in Filament v4, this stopped working. The dropdown would show with the tenant selector, however the register tenant item would not.

I traced the issue to this problem, and have instead allowed the possibility of lazy-loading the menu items _if they should be called upon_

## Visual changes

Before:
image

After:

image

## Functional changes

### DIFF
--- a/packages/panels/src/Livewire/Concerns/HasTenantMenu.php
+++ b/packages/panels/src/Livewire/Concerns/HasTenantMenu.php
@@ -10,7 +10,7 @@ trait HasTenantMenu
     /**
      * @var array<Action>
      */
-    protected array $tenantMenuItems = [];
+    protected ?array $tenantMenuItems = null;
 
     public function bootHasTenantMenu(): void
     {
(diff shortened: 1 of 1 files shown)


danharrin:
Maybe just remove line 21-23? Then you don't need other changes?

olliescase:
@danharrin Done 👍  the only thing I will caveat this is now we are calling `getTenantMenuItems()` _always_ when tenancy is active, regardless of whether the menu is used or not.

One assumes that this should result in minimal/no performance impact, but it could have an effect if you have multiple components with this trait but only want to display it once.

danharrin:
@olliescase can you test the code I just pushed please?

olliescase:
Seems to work well!

danharrin (approved):

Issue #17298, opened 2025-08-13
Unable to remove table filter when filter is nested in a form


Oreo-Monster:
### Package

filament/tables

### Package Version

v4.0.0

### Laravel Version

v12.22.1

### Livewire Version

v3.6.4

### PHP Version

8.4

### Problem description

When a table filter is placed inside a form section in the TableFilterFormSchema, that filter will throw an exception when you try to remove.

From my poking around, it seem the issue starts at [line 79](https://github.com/filamentphp/filament/blob/f00683622449f11c19fccdd02b13a23dfc67c924/packages/tables/src/Concerns/HasFilters.php#L79) of HasFilters. The [getComponent](https://github.com/filamentphp/filament/blob/f00683622449f11c19fccdd02b13a23dfc67c924/packages/schemas/src/Concerns/HasComponents.php#L135) function of HasComponents when only given a string will do a shallow search of the array keys of the flattened schema. The problem is the filter components we are looking for have a bunch of prefixes from the sections, and we are only searching with the filters name. 

### Expected behavior

The filter to remove with no errors

### Steps to reproduce

# Reproduse in my project
To use my [test repo](https://github.com/Oreo-Monster/filament-v4-broken-table-filters.git). Clone it, `composer install`, `php artisan serve`, and go to the /admin/users endpoint. There will be two filters, one inside a section and is broken, the other is outside the section and works. Removing the broken filter will cause the error.

# Reproduse in another project
To recreate, make a table with filters defined in filters() and a defined filtersFormSchema(). In the form schema nest the filters in a Section. 

Now on the table, apply a filter, then remove it by clicking the X in the indicator. Upon removing it, you should get an exception:
```
foreach() argument must be of type array|object, null given
vendor/filament/tables/src/Concerns/HasFilters.php :90
```

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/Oreo-Monster/filament-v4-broken-table-filters.git

### Relevant log output

```shell

```

PR #17300, opened 2025-08-13
fix removeTableFilters to work with nested form filters

Oreo-Monster:
## Description
See #17298 for my full bug report. 

Heart of the issue is table filters that are nested inside a section are throwing an error when trying to remove them.

The problem is [removeTableFilters](https://github.com/filamentphp/filament/blob/f00683622449f11c19fccdd02b13a23dfc67c924/packages/tables/src/Concerns/HasFilters.php#L74) uses the name of the filter to find the filter component within the TableFiltersForm. When a filter is placed inside a section, the name of the filter is prefixed with the sections name as well. This causes the filter name to not match its form component, causing the exception.

To fix, I used a regex expression to find components instead of just matching the filter name. 
```
"/(?:\.|^)$filterName\$/"
```
This will match to any filter key that is just the filter name, or has a prefix and the key ends with ".filter_name". Hopefully this pattern is specific enough to not cause issues with a filters who's name is a substring of another filter.

## Functional changes

- [x] HasFilters removeTableFilter now matches filter name and key with a regex
- [x] Unit tests all passing
- [x] Linted with pint

### DIFF
--- a/packages/tables/src/Concerns/HasFilters.php
+++ b/packages/tables/src/Concerns/HasFilters.php
@@ -37,10 +37,10 @@ trait HasFilters
             ->schema($table->getFiltersFormSchema())
             ->when(
                 $table->hasDeferredFilters(),
-                fn (Schema $schema) => $schema
+                fn(Schema $schema) => $schema
                     ->statePath('tableDeferredFilters')
                     ->partiallyRender(),
-                fn (Schema $schema) => $schema
+                fn(Schema $schema) => $schema
                     ->statePath('tableFilters')
                     ->live(),
             );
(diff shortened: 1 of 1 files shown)


danharrin:
Hi, I don't think a regex should be used for this. Ideally, we should not have to modify the table package code when the bug is in the forms package. The component indexing by key should be fixed instead.

Oreo-Monster:
I agree the regex is a hacky solution, but I think this one is better:
```
        $filterFormGroup = $this->getTableFiltersForm()
            ->getComponent(fn($component) => $filterName === $component->getKey(isAbsolute: false));
```

Uses the component class to get the relative key instead of regex. Still a fix in the tables package, but I would argue the bug was caused by tables misusing the forms package. Not going to bug you with another pull request but thought this might be helpful

danharrin:
Shouldn't the array inside `getComponent()` already be keyed by key though? That is the correct solution here I think, but something is going on inside that function

danharrin:
I would love another PR to fix whatever is wrong within there, but making the change inside the tables package is just a hack I think.

Oreo-Monster:
getComponent relies on the getFlatComponents to get all the keys. getFlatComponents uses the absolute key of the components (with prefixes) to build a flat array. getFlatComponets does keep track of the top level parent component and can remove that from all the prefixes, but if there are more levels of nesting those prefixes won't be removed. 

You can see this in the recursive step, the containerKey is passed all the way down the stack. to get the true relative key in the recursion, you would have to pass the compoentKey instead of contianerKey.

The issue for flatCompoents is it uses the keys as array index, so if any components share keys but are nested differently, they will overwrite themselves. Using the callback gets around this, we still use the flat array with full keys, but we have access to each component's relative name.

Hopefully that explanation is not too confusing, will make a new PR now

PR #17304, opened 2025-08-13
wip: Add hintIconColor() method

zvizvi:


danharrin:
`->afterLabel(Icon::make('icon')->color('success'))`

We do not want to introduce more duplicate configuration methods, which is why hints are no longer documented in v4 but this is

PR #17313, opened 2025-08-14
fix removeTableFilters to work with nested form filters

Oreo-Monster:
## Description

See original PR for full discussion #17300


Oreo-Monster:
If you still don't like the taste of a callback in the tables package, you can fold it into getComponent with a parameter flag. like 
```
if($useRelative) {
   //array map the flat array to use the relative keys as values then do a search
  //Or just use the same callback and do the search directly
}
```

Or even make a wrapper function like `getCompoentRelative()` that puts the callback definition back in the HasComponent trait

danharrin:
I think this is the correct solution: #17509

PR #17314, opened 2025-08-14
feat: allow binding value object constant state

erikgaal:
## Description

We're using Value Objects heavily in our application, and would like to keep using them in Infolists. Changes in Filament 4's state mechanisms prevent us from using non-Model Value Objects effectively.

One issue lies in the fact that `constantState` only accepts `array`. We'd like to be able to bind objects to the state of schemas without having to convert it back to arrays.

Another blocker is that `RepeatableEntry` only sets the container state when it's a `Model` or an `array`, and silently ignores if it's anything else.

```php
public function content(Schema $schema): Schema
{
    return $schema
        ->state(new Profile(
            id: 'some-id',
            results: [
                new Result(
                    date: new \DateTime(),
                    score: 100,
                ),
            ],
        ))
        ->schema([
            TextEntry::make('id'),
            RepeatableEntry::make('results')
                ->schema([
                    TextEntry::make('date')->date(),

                    TextEntry::make('score')->numeric(),
                ])
        ]);
}
```

## Functional changes

### DIFF
--- a/packages/infolists/src/Components/RepeatableEntry.php
+++ b/packages/infolists/src/Components/RepeatableEntry.php
@@ -30,7 +30,7 @@ class RepeatableEntry extends Entry implements HasEmbeddedView
 
             if ($itemData instanceof Model) {
                 $container->record($itemData);
-            } elseif (is_array($itemData)) {
+            } elseif (is_array($itemData) || is_object($itemData)) {
                 $container->constantState($itemData);
             }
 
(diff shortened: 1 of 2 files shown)


danharrin (changes requested):
Please ensure `array` is declared before `object` in types since that is much more common

danharrin (approved):

PR #17324, opened 2025-08-15
ignoreRecord true on default on unique validation

dipesh79:
## Description
ignoreRecord true on default on unique validation

## Functional changes


danharrin:
It is true by default already, since `shouldUniqueValidationIgnoreRecordByDefault()` is true by default

PR #17334, opened 2025-08-15
[4.x] feat: make `afterStateUpdatedJs` work on table repeater

### LINKED ISSUE #17333: [4.x] The `afterStateUpdatedJs` does not work in table repeater.
### Package

filament/filament

### Package Version

v4.*.*

### Laravel Version

v12.*

### Livewire Version

v3.*

### PHP Version

v^8.2

### Problem description

If we have code like this into form or schema it does not work currently in table repeater but it works in normal repeater though.

```php
Repeater::make('dates')
        ->table([
            TableColumn::make('Start date'),
            TableColumn::make('End date'),
        ])
        ->schema([
            DateTimePicker::make('start_date')
                ->label('Start Date')
                ->afterStateUpdatedJs(<<<'JS'
                    if ($state && ($get('end_date') === null || $get('end_date') === '')) {
                        $set('end_date', $state);
                    }
                JS)
                ->required(),
            DateTimePicker::make('end_date')
                ->label('End Date')
                ->required(),
        ])->columns(2)
        ->columnSpan(2),
``` 

### Expected behavior

It should works similar to normal repeater.

### Steps to reproduce

clone the repo and setup project. 
open http:/{base-path}/admin/users/create
and try to update the start date and check if end date is updating as same as start date.

https://github.com/MrPunyapal/filament-table-repeater-issue

### Relevant log output

```shell

```

MrPunyapal:
## Description
fixes https://github.com/filamentphp/filament/issues/17333

### For this code.
```php
Repeater::make('dates')
      ->table([
          TableColumn::make('Start date'),
          TableColumn::make('End date'),
      ])
      ->schema([
          DateTimePicker::make('start_date')
              ->label('Start Date')
              ->afterStateUpdatedJs(<<<'JS'
                  if ($state && ($get('end_date') === null || $get('end_date') === '')) {
                      $set('end_date', $state);
                  }
              JS)
              ->required(),
          DateTimePicker::make('end_date')
              ->label('End Date')
              ->required(),
      ])->columns(2)
      ->columnSpanFull(),
```
### Before

https://github.com/user-attachments/assets/ead7c26c-6e3b-444a-842a-36d4f462b397

### After

https://github.com/user-attachments/assets/86f3fe44-11bb-444d-8491-897246583795

## Functional changes


MrPunyapal:
@danharrin I'm up for making `hiddenJs` and `visibleJs` work with table repeater if that make sense as we can see that we are making that intentionally visible even if that is hidden.
image

mszabeh:
I think it would be better 
if these two issues were separated 
**afterStateUpdatedJs** in this PR,
**hiddenJs** and **visibleJs** in another PR

mszabeh:
I think it would be better 
if these two issues were separated 
**afterStateUpdatedJs** in one PR,
**hiddenJs** and **visibleJs** in another PR

mszabeh:
Hi @MrPunyapal 
I need afterStateUpdatedJs fix
in my project.
Can you please, separate and create new PR. Or edit this PR?

MrPunyapal:
> Hi @MrPunyapal 
> I need afterStateUpdatedJs fix
> in my project.
> Can you please, separate and create new PR. Or edit this PR?

@mszabeh right now it's not about `afterStateUpdatedJs` I removed `hiddenJs` and `visibleJs` code before pushing.

mszabeh:
Hi @danharrin ,
Please merge this PR. Thanks

packages/forms/resources/views/components/repeater/table.blade.php
```diff
@@ -150,11 +150,38 @@ class="fi-fo-table-repeater-actions"
...
-                                                <td>
+                                                @php
+                                                    $componentStatePath = null;
+                                                    $canUseAlpineIntegration = false;
+
+                                                    try {
+                                                        $name = $component->getName();
+                                                        $componentStatePath = $component->getStatePath();
+                                                        $canUseAlpineIntegration = filled($componentStatePath) && $component->getLivewire();
+                                                    } catch (\Exception $e) {
```
danharrin: This doesn't look right, and could silence real code issues. What exception is being thrown that needs to be caught?

danharrin (changes requested):

packages/forms/resources/views/components/repeater/table.blade.php
MrPunyapal: I took inspiration from https://github.com/filamentphp/filament/blob/4.x/packages/schemas/src/Schema.php for this implementation, and I'm currently unable to identify the specific exception it was intended to catch. Therefore, I believe it can be removed until a real issue arises.

The code has been updated to remove unrelated code, and it still functions correctly.

packages/forms/resources/views/components/repeater/table.blade.php
```diff
@@ -150,11 +150,30 @@ class="fi-fo-table-repeater-actions"
...
                                             @php
-                                                $counter++
+                                                $counter++;
                                             @endphp
 
                                             @if ($component->isVisible())
-                                                <td>
+                                                @php
+                                                    $name = $component->getName();
+                                                    $componentStatePath = $component->getStatePath();
```
danharrin: All of the new code you introduced needs to be wrapped in a check to ensure that `! (($component instanceof Action) || ($component instanceof ActionGroup))`. Then you can also remove the `method_exists()` checks, I think? It should be very similar code to `Schema.php` but currently is it not.

danharrin (changes requested):

packages/forms/resources/views/components/repeater/table.blade.php
```diff
@@ -150,11 +152,34 @@ class="fi-fo-table-repeater-actions"
...
+                                                @if (! (($component instanceof Action) || ($component instanceof ActionGroup)))
+                                                    @php
+                                                        $name = $component->getName();
+                                                        $componentStatePath = $component->getStatePath();
+                                                    @endphp
+                                                    <td
+                                                        @if (filled($componentStatePath))
+                                                            x-data="filamentSchemaComponent({
+                                                                path: @js($componentStatePath),
+                                                                containerPath: @js(str_replace('.'. $name, '', $componentStatePath)),
```
danharrin: For each row in the table you should save `$itemStatePath = $item->getStatePath()` and use that instead of guessing based on the name. You need to remove the `$name` var as it doesn't work on layout components.

danharrin (changes requested):

packages/forms/resources/views/components/repeater/table.blade.php
MrPunyapal: thanks for the hint. updated the code.

danharrin: For each row, not each component. The `$itemStatePath` doesn't change for each component.

MrPunyapal: gotcha, updated the code

danharrin (approved):

PR #17350, opened 2025-08-16
Fix Toggle component error (#17347)

### LINKED ISSUE #17347: V4 Bug on Toggle
### Package

filament/filament

### Package Version

4.0

### Laravel Version

12.22.1

### Livewire Version

_No response_

### PHP Version

8.4.11

### Problem description

Toggle form component not set default value, throws exception

Image

In all cases, setting the default value to false always displays this error.

I went to the module file where this method is located and placed a log before it. The Toggle is always empty.

### Expected behavior

Not show error or setting default value is not causing problem.

### Steps to reproduce

I went to the module file where this method is located and placed a log before it. The Toggle is always empty.

https://github.com/IgorDePaula/toggle_bug.git

### Relevant log output

```shell

```

shadighorbani7171:
Fixes Toggle component error in Filament V4.

## What was wrong
Toggle component was throwing an error when using default(false).

## What I fixed
- Changed default(false) to default(null)
- Updated BooleanStateCast settings

## Files changed
packages/forms/src/Components/Toggle.php

Fixes #17347

### DIFF
--- a/packages/forms/src/Components/Toggle.php
+++ b/packages/forms/src/Components/Toggle.php
@@ -24,7 +24,7 @@ class Toggle extends Field
     {
         parent::setUp();
 
-        $this->default(false);
+        $this->default(null);
 
         $this->rule('boolean');
     }
@@ -36,7 +36,7 @@ class Toggle extends Field
     {
         return [
             ...parent::getDefaultStateCasts(),
-            app(BooleanStateCast::class, ['isNullable' => false]),
+            app(BooleanStateCast::class, ['isNullable' => true]),
         ];
     }
 }


danharrin:
I don't think the Toggle or Checkbox should be nullable as there is no way for the user to control that state. I will look at the issue myself and see if there is an alternative fix, but I don't think this is it. Sorry.

shadighorbani7171:
Hi, thanks for the feedback! No worries ,
I understand now that Toggle/Checkbox shouldn’t be nullable since users can’t actually control a third state.

I’m happy to leave this to you, but if you’d like, I can also explore another approach and open a new PR. If not, that’s completely fine too. Thanks again for the clarification! 🙏

danharrin:
If you can have a look at alternative ways of fixing the error, while still keeping the default as false and not allowing it to output null, that would be great :)

shadighorbani7171:
Ok thank you! I’ll check and test different ways to fix it while keeping the default as false. If I find a working solution, I’ll open a new PR. Thanks.

PR #17359, opened 2025-08-17
[4.x] Feat: Add TableEntry component to Infolists

oddvalue:
## Description

This PR adds a very simple TableEntry component to infolists for 2D arrays, similar to the KeyValueEntry.

## Visual changes

image

## Functional changes


danharrin (changes requested):
Please copy the CSS classes into their own file so we aren't piggy-backing off of the `key-value` styling

oddvalue:
> Please copy the CSS classes into their own file so we aren't piggy-backing off of the `key-value` styling

@danharrin no problem, have done that and added new translations too.

packages/infolists/resources/css/components/table.css
```diff
@@ -0,0 +1,23 @@
+.fi-in-table {
+    @apply w-full table-auto divide-y divide-gray-200 rounded-lg bg-white shadow-xs ring-1 ring-gray-950/5 dark:divide-white/5 dark:bg-white/5 dark:ring-white/10;
+
+    & th {
+        @apply px-3 py-2 text-start text-sm font-medium text-gray-700 dark:text-gray-200;
+    }
+
+    & tbody {
+        @apply divide-y divide-gray-200 font-mono text-base sm:text-sm sm:leading-6 dark:divide-white/5;
```
oddvalue:
```suggestion
        @apply divide-y divide-gray-200 text-base sm:text-sm sm:leading-6 dark:divide-white/5;
```

danharrin (changes requested):
Similar to how a `Repeater` has `->table()` mode, how about we introduced a new `->table()` mode for `RepeatableEntry` instead?

This way, we can mix different data types within the table, so columns can contain icons, images, different text styles, etc, since these are all handled by the individual entry classes.

Otherwise I feel like these other data types are inevitable future feature requests of the `TableEntry` anyway, and it will not remain "simple".

oddvalue:
@danharrin fair point. Have created a new PR for it: https://github.com/filamentphp/filament/pull/17865

PR #17369, opened 2025-08-18
fix: Page error notifications loop

### LINKED ISSUE #17332: v4: Failed Network Request Causing Excessive /livewire/update Calls
### Package

filament/filament

### Package Version

v4.0.0

### Laravel Version

v12.3.0

### Livewire Version

v3.6.4

### PHP Version

PHP 8.4

### Problem description

I noticed over the last few days that my server was receiving unusual CPU and network traffic. 

Image

After some investigation, I found that my NGINX log files were about 1GB, caused by requests being sent to my Filament app - I then noticed it was coming from my Mac which I had left idle on my Livewire v4 app. It was sending thousands of requests to my server every minute, and in some cases if the PHP server was off, it would result in 502 bad gateway errors, which meant Filament was sending the request to /livewire/update hundreds of times per second.

I was actually able to replicate this on the Filament demo site.

I'm happy to jump on a video call or record a video to demonstrate the behaviour.

### Expected behavior

For it to attempt and redirect to the login page, or after a certain amount of time, give up (in the case of 502 bad gateway)

### Steps to reproduce

1. Visit the demo site https://demo.filamentphp.com/login
2. Login to the site
3. Open your network tools
4. Clear all cookies for the current page
5. The next /livewire/update call will result in 419 as the session has expired
6. Hundreds of requests will be made

To reproduce the 502 bad gateway issue

1. Run Filament behind an NGINX proxy
2. Login to the site
3. Open your network tools
4. Turn off the Filament server, leaving the proxy on
5. Filament will send thousands of requests and get 502 bad gateway errors

### Relevant log output

```shell

```

danharrin:
Fixes #17332.

### DIFF
--- a/packages/panels/resources/views/components/page/index.blade.php
+++ b/packages/panels/resources/views/components/page/index.blade.php
@@ -167,8 +167,16 @@
             <script>
                 const errorNotifications = @js($this->getErrorNotifications())
 
+                let hasPageExpiredError = false
+
                 Livewire.hook('request', ({ fail }) => {
                     fail(({ status, preventDefault }) => {
+                        if ((status === 419) && hasPageExpiredError) {
+                            return
+                        }
+
+                        hasPageExpiredError = status === 419
+
                         preventDefault()
 
                         const errorNotification =
--- a/packages/panels/resources/views/components/page/index.blade.php
+++ b/packages/panels/resources/views/components/page/index.blade.php
@@ -167,15 +167,18 @@
             <script>
                 const errorNotifications = @js($this->getErrorNotifications())
 
-                let hasPageExpiredError = false
+                let hasFatalRequestError = false
 
                 Livewire.hook('request', ({ fail }) => {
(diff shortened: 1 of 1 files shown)


Sammyjo20:
Thanks @danharrin! Just a quick question more than anything - will there be a scenario when other HTTP codes other than 419 and 502 will get fired - which will then need to be caught in a further PR?

danharrin:
I'm honestly not sure, can you think of any that would cause the notification to fail being sent via HTTP request? When a request inside a Livewire hook fails it seems to retry infinitely which is weird

Sammyjo20:
@danharrin I can only assume that it'll do the loop for any HTTP error, I can see if I can hack my local instance to see if I can replicate it.

danharrin:
I think https://github.com/filamentphp/filament/pull/17519 is a better solution, it doesn't depend on the expected status code

PR #17386, opened 2025-08-18
[4.x] Fix(docs): Add missing form class to singular resources example

ace-of-aces:
## Description

This adds the `fi-sc-form` class to the singular resources blade view example, which provides proper padding to the form so the submit button is spaced from the form like in regular resources.

I also added an `id="form"` attribute for good measure, because it's also present in regular resource forms :)

## Visual changes

"Run" is the submit button in this example.

Before:
Screenshot 2025-08-18 at 17 51 48
After:
Screenshot 2025-08-18 at 17 51 58

## Functional changes

- [ ] ~~Code style has been fixed by running the `composer cs` command.~~

### DIFF
--- a/docs/03-resources/09-singular.md
+++ b/docs/03-resources/09-singular.md
@@ -107,7 +107,7 @@ The page Blade view should render the form and a Save button. The `wire:submit` 
 
 ```blade
 <x-filament::page>
-    <form wire:submit="save">
+    <form class="fi-sc-form" id="form" wire:submit="save">
         {{ $this->form }}
     
         <div>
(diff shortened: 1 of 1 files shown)


danharrin:
Hey, thanks, but this is just a basic example on how to create a form, and not an example on how to style it. You might be using it outside of a panel and not import the Form component class.

If you want to style a form like a normal resource form, I would use the Form schema component, which then applies `fi-sc-form` for you.

danharrin:
I just noticed this was for singular resources and not custom forms, my bad, but I still don't think users should need to apply these classes manually. I will update this page with better recommendations

danharrin (approved):

PR #17389, opened 2025-08-18
add team model example & pivot table to tenancy docs

kjjd84:
## Description

add missing team model example & pivot table to tenancy docs

## Visual changes

n/a

## Functional changes

### DIFF
--- a/docs/07-users/03-tenancy.md
+++ b/docs/07-users/03-tenancy.md
@@ -57,7 +57,30 @@ class PostObserver
 
 ## Setting up tenancy
 
-To set up tenancy, you'll need to specify the "tenant" (like team or organization) model in the [configuration](../panel-configuration):
+To set up tenancy, first you must create a model which represents tenants, e.g. `Team`:
+
+```php
+namespace App\Models;
+
+use Illuminate\Database\Eloquent\Factories\HasFactory;
+use Illuminate\Database\Eloquent\Model;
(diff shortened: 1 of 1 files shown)


danharrin:
Sorry, this isn't something I want to add. It makes the tenancy system feel more rigid than it actually is. The tenancy system should work with whatever way you handle tenancy in your app, with whichever model. Also, we don't hardcode the `members` relationship name anywhere as far as I can see, so that part is confusing me.

kjjd84:
> Sorry, this isn't something I want to add. It makes the tenancy system feel more rigid than it actually is. The tenancy system should work with whatever way you handle tenancy in your app, with whichever model. Also, we don't hardcode the `members` relationship name anywhere as far as I can see, so that part is confusing me.

@danharrin if you don't specify a members relation, you get an error in the app that it doesn't exist:

> Call to undefined method App\Models\Team::members()

this is because you have this in the `RegisterTeam` example in the docs:

```php
    protected function handleRegistration(array $data): Team
    {
        $team = Team::create($data);

        $team->members()->attach(auth()->user());

        return $team;
    }
```

the docs for multi-tenancy are just overall confusing and missing things required to actually get it working properly (e.g. the `Team` model, pivot table, relations, etc). i took this into consideration when i added my PR, based on the current docs and what is required to get a minimal working example by following the docs.

PR #17404, opened 2025-08-19
[4.x] Move `observeTenancyModelCreation()` to `ServiceProvider::boot()`

pxlrbt:
Registering the tenancy observer in `register()` instead of `boot()` will prevent packages configuring models via a config option, because it runs before the config is booted.

I could only test this with one of the packages I maintain and the existing test set of Filament itself. 

## Functional changes

### DIFF
--- a/packages/panels/src/Panel.php
+++ b/packages/panels/src/Panel.php
@@ -68,10 +68,6 @@ class Panel extends Component
 
     public function register(): void
     {
-        foreach ($this->getResources() as $resource) {
-            $resource::observeTenancyModelCreation($this);
-        }
-
         $this->registerLivewireComponents();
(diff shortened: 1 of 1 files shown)


danharrin (approved):

WheelieDev:
Hello,

Not sure if this is the correct place to put this so feel free to ask me to move it if necessary, but I think we may have introduced a bug with this change.

When in register, if I created a tenant aware model as part of say an action, it was automatically be scoped to the tenant, whereas now this doesn't happen and the tenant needs to be manually provided in someway.

This may of course be what you intended, but would welcome your thoughts

All the best

Tom

danharrin:
Hi Tom, where is the action being triggered in the lifecycle of the request?

PR #17416, opened 2025-08-19
This PR improves accessibility and usability of the password reveal toggle in the textinput component.

rishi111-code:
…oggle in the TextInput component.

Changes

Added aria-pressed, dynamic aria-label, and title attributes for screen readers and keyboard users.

Wrapped input and toggle in data-input-root container for reliable Alpine sync.

Ensured input type switches between password and text seamlessly.

Added new test (PasswordToggleAccessibilityTest) under packages/forms/tests/ to verify ARIA attributes render correctly.

Motivation
Currently, the password reveal toggle lacks accessibility attributes, which makes it unclear for assistive technology users and less friendly for keyboard navigation. This change ensures:

Screen reader users hear “Show password” / “Hide password”.

Toggle state is correctly announced via aria-pressed.

Input type stays in sync with Alpine state.

Backward Compatibility

No breaking changes.

UI/UX remains the same for visual users, with accessibility enhancements under the hood.

Testing

New PHPUnit test confirms presence of ARIA attributes and accessible markup.

All existing tests pass.

## Functional changes

### DIFF
--- a/packages/forms/resources/views/components/text-input.blade.php
+++ b/packages/forms/resources/views/components/text-input.blade.php
@@ -95,6 +95,7 @@
                 ->class(['fi-fo-text-input'])
         "
     >
+    <div data-input-root class="fi-fo-text-input flex items-center">
         <input
             {{
                 $inputAttributes->class([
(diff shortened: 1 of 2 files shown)


packages/forms/resources/views/components/text-input.blade.php
```diff
@@ -104,6 +105,29 @@ class="fi-fo-text-input-wrp"
                 ])
             }}
         />
+        <button
```
wychoong, line 108: this is rendering the toggle in **every single** text input field??

danharrin:
I'm sorry, this just looks like AI slop, it doesn't make any sense. I would love to improve the accessibility, but this is not the way.

PR #17419, opened 2025-08-19
feat(tables): add scrollToTop() to auto-scroll after pagination

saschaende:
feat(tables): add scrollToTop() to auto-scroll after pagination

Add Table->scrollToTop() option to scroll to top when table pagination changes. Introduce a new Table concern (CanScrollToTop) and dispatch a 'filament-tables-scroll-to-top' browser event from InteractsWithTable::setPage for the table’s page name. The table root view conditionally listens via Alpine and calls window.scrollTo(0, 0).

- Opt-in; disabled by default; no breaking changes
- No AppServiceProvider render hook required
- Usage: return $table->scrollToTop()


saschaende:
See https://filamentexamples.com/tutorial/pagination-auto-scroll-to-top

danharrin:
Hi, apologies but I am not going to implement this in the core right now. I don't think it is a great user experience to scroll up from the pagination since you then have to scroll again to switch pages, and this is already possible as explained in the tutorial. I might reconsider this again in the future if I have similar user feedback.

saschaende:
This would be an "optional" feature - by using $table->scrollToTop()

In my use case, with filament, people would have to scroll up manually for each page: https://ende.app/en/genres/electronica

Not having autoscroll is the worst user experience in this use case. Every website with a pagination has this, just browse products on amazon....

To be honest I dont understand why to decline an optional feature that is so important to make filament usable for frontend pages.

Yes I can use https://filamentexamples.com/tutorial/pagination-auto-scroll-to-top - but I really was searching for such a basic feature in the filamentphp documentation and wondered why this does not exist.

pxlrbt:
> I don't think it is a great user experience to scroll up from the pagination

I think this is a quite common pattern. 

> since you then have to scroll again to switch pages

What's the reason for switching pages? Probably checking the content of the next page. People usually don't navigate just for fun. So currently everybody has to first scroll to the top again to see the entries and then scroll down again to go to the next page.

I think this is good as as optional feature, especially since it's not really complex/hard to maintain.

I guess it depends on how much of the table you can see in the viewport. 

PS: This should scroll to the top of the table and not top of the page, but that's a detail that can be fixed, if @danharrin reconsiders this PR.

pxlrbt:
Btw. if this wasn't a Livewire app the default behaviour would be "scroll up" caused by the page change.

saschaende:
> PS: This should scroll to the top of the table and not top of the page, but that's a detail that can be fixed, if @danharrin reconsiders this PR.

I could fix this. Also implement smooth scrolling like I use on my website.

If more poeple start using Filament for Blogs, Shops.... this is a very important feature.

Examples from prominent websites:

- google: https://www.google.com/search?q=test
- amazon: https://www.amazon.com/s?k=keyboard&__mk_de_DE=%C3%85M%C3%85%C5%BD%C3%95%C3%91&crid=38T8871L5TJ3T&sprefix=keyboar%2Caps%2C189&ref=nb_sb_noss_2
- MisterSpex: https://www.misterspex.de/l/pg/100266

FilamentPHP is perfect also for building websites - not only backends.

saschaende:
Btw. - this is a Livewire 3 feature: https://livewire.laravel.com/docs/pagination#customizing-scroll-behavior

danharrin:
Please make it scroll to the top of the table instead of the page

saschaende:
> Please make it scroll to the top of the table instead of the page

Check, will do

saschaende:
@danharrin  I made a commit to my branch. Seems to work on my project.

Thank you for revisiting this topic.

packages/tables/src/Table/Concerns/CanScrollToTop.php
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Filament\Tables\Table\Concerns;
+
+use Closure;
+
+trait CanScrollToTop
+{
+    protected bool | Closure $shouldScrollToTopOnPageChange = false;
+
+    public function scrollToTop(bool | Closure $condition = true): static
```
pxlrbt: I think we should keep the same name for the method:

```suggestion
    public function scrollToTopOnPageChange(bool | Closure $condition = true): static
```

packages/tables/resources/views/index.blade.php
```diff
@@ -1249,7 +1257,7 @@ class="fi-ta-record-collapse-btn fi-icon-btn"
                                                             'fi-wrapped' => $columnGroup->canHeaderWrap(),
                                                             ((($columnGroupAlignment = $columnGroup->getAlignment()) instanceof \Filament\Support\Enums\Alignment) ? "fi-align-{$columnGroupAlignment->value}" : (is_string($columnGroupAlignment) ? $columnGroupAlignment : '')),
                                                             (filled($columnGroupHiddenFrom = $columnGroup->getHiddenFrom()) ? "{$columnGroupHiddenFrom}:fi-hidden" : ''),
-                                                            (filled($columnGroupVisibleFrom = $columnGroup->getVisibleFrom()) ? "{$columnGroupVisibleFrom}:fi-visible" : ''),
+                                                            (filled($columnGroupVisibleFrom = $columnGroup->getVisibleFrom()) ? "{$columnVisibleFrom}:fi-visible" : ''),
```
pxlrbt: Why was this changed?

packages/tables/src/Concerns/InteractsWithTable.php
```diff
@@ -244,6 +245,22 @@ public function resetPage($pageName = null): void
         $this->resetLivewirePage($pageName ?? $this->getTablePaginationPageName());
     }
 
+    /**
+     * Leitet setPage von Livewire weiter und dispatcht ggf. ein Browser-Event,
+     * um nach einem Tabellen-Seitenwechsel nach oben zu scrollen.
+     *
```
pxlrbt: Please remove the German comment

packages/tables/src/Concerns/InteractsWithTable.php
```diff
@@ -244,6 +245,22 @@ public function resetPage($pageName = null): void
...
+     * @param  ?string  $pageName
+     */
+    public function setPage($page, $pageName = 'page'): void
+    {
+        $this->setLivewirePage($page, $pageName);
+
+        if ($pageName === $this->getTablePaginationPageName() && $this->getTable()->shouldScrollToTopOnPageChange()) {
+            // Livewire v3 Browser-Event: nur die betroffene Tabelle soll reagieren
```
pxlrbt: Please remove the German comment

pxlrbt (changes requested):

saschaende:
@pxlrbt I commited the changes:

- keep the same name for the method: scrollToTopOnPageChange()
- remove german comment
- Undo changes that were accidentally
  committed

pxlrbt:
Looks good to me know. Maybe we should rename the Trait to `CanScrollToTopOnPageChange`, but I'll leave that to Dan.

pxlrbt (approved):

danharrin:
Please fix the PHPStan issue

saschaende:
> Please fix the PHPStan issue

updated setPage method parameter types

packages/tables/src/Concerns/InteractsWithTable.php
```diff
@@ -246,9 +246,10 @@ public function resetPage($pageName = null): void
     }
 
     /**
-     * @param  ?string  $pageName
-     */
-    public function setPage($page, $pageName = 'page'): void
+     * @param  ?int|string  $page
+     * @param  string  $pageName
```
pxlrbt: If we can use PHP types here, we don't have to repeat them in a docblock.

saschaende: Okay, I removed it.

saschaende:
Funny is that Livewire itself has no types:
public function setPage($page, $pageName = 'page')

:D

saschaende:
Anything to do here? When will this be merged? Sry I am new.

pxlrbt:
I think Dan was away last week and will check the open PRs when he's got time. Also there are 33 more PRs, so it just takes some time to look at all of them.

saschaende:
Ahhhhh okay 😂😘

danharrin:
Features are going to be reviewed at lower priority to bug fixes

saschaende:
Makes sense!

danharrin (approved):

PR #17424, opened 2025-08-20
feat: Add HTML support to table group titles

Baspa:
## Description
   
   This PR adds HTML support to Filament table group titles, allowing developers to use HTML in group headers for better customization.
   
   ## Changes
   
   - Update `Group::getTitle()` to support `HtmlString` instances
   - Change Blade template to use `{!! !!}` instead of `{{ }}` for group titles
   - Allow HTML rendering in group titles for better customization
   
   ## Example Usage
   
   ```php
   ->getTitleFromRecordUsing(function (Order $record): HtmlString {
       $date = (new DateTime($record->delivery_date))->format('d F Y');
       $product = $record->product->name ?? '';
       $color = $record->product->color ?? '';
   
       if ($color) {
           return new HtmlString("$date – <span style='color: $color; font-weight: bold;'>$product</span>");
       }
   
       return new HtmlString("$date – $product");
   })
   ```

## Visual result
   
Screenshot 2025-08-20 at 10 36 51

   ## Testing
   
   - [x] Tested with HTML strings in group titles
   - [x] Verified backward compatibility with plain strings
   - [x] Ensured proper HTML escaping for security

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -981,7 +981,7 @@
                                                         {{ $recordGroupLabel }}:
                                                 @endif
 
-                                                {{ $recordGroupTitle }}
+                                                {!! $recordGroupTitle !!}
                                             </{{ $secondLevelHeadingTag }}>
 
                                             @if (filled($recordGroupDescription = $group->getDescription($record, $recordGroupTitle)))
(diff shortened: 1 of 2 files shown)


danharrin:
I would accept a PR for this, but this is not the correct solution - it introduces XSS vulnerabilities if HTML is included in the title without using an `HtmlString` object. Instead, the method should just allow an `HtmlString` object, and if that is output in `{{}}` then you do not need `{!! !!}`.

PR #17427, opened 2025-08-20
feat: customizable summary header

Baspa:
## **What does this PR do?**

This PR adds the ability to customize or completely hide table summary headers in Filament tables. Currently, summary headers show labels like "Orders summary" which cannot be customized by developers.

## **New Features**

- **`summaryHeaderLabel()` method** - Set custom text for summary headers
- **Empty string support** - Pass `''` to completely hide summary headers
- **Closure support** - Use dynamic labels with closures
- **Backward compatibility** - Existing tables work unchanged

## **Usage Examples**

```php
// Hide summary header completely
->summaryHeaderLabel('')

// Custom static text
->summaryHeaderLabel('Summary')

// Dynamic text with closure
->summaryHeaderLabel(fn () => 'Custom ' . now()->format('Y-m-d'))
```

## **Problem Solved**

Currently, developers cannot customize table summary headers that show labels like:
- "Orders summary" 

This forces developers to either accept the default labels or disable summaries entirely.

##  **Implementation Details**

- **Trait Enhancement**: Added `summaryHeaderLabel()` method to `CanSummarizeRecords` trait
- **View Updates**: Modified summary Blade templates to support custom labels
- **Context Fix**: Ensured proper method calls between Livewire components and Table instances
- **Parameter Passing**: Added `custom-summary-header-label` prop to summary components

## **Files Changed**

- `packages/tables/src/Table/Concerns/CanSummarizeRecords.php` - Added new methods
- `packages/tables/resources/views/components/summary/index.blade.php` - Updated templates
- `packages/tables/resources/views/index.blade.php` - Updated main view

## **Breaking Changes**

None. This is a purely additive feature that maintains full backward compatibility.


danharrin:
To be honest, I don't really think that I see the point of customising the text all at once. In what situation is it useful to have the text repeating like this?

Screenshot 2025-09-18 at 22 52 09

Is the real goal of this just to allow the text to be removed? If so, I think you should probably just do that with CSS. I just really doubt how useful it is to "change" the label using this method... please feel free to reply and tell me how I'm wrong

Baspa:
> Is the real goal of this just to allow the text to be removed? If so, I think you should probably just do that with CSS. I just really doubt how useful it is to "change" the label using this method... please feel free to reply and tell me how I'm wrong

Removal would be sufficient for me, but I wasn't able to remove or hide it with only CSS unfortunately (didn't even get managed it with Claude). 

The issue is probably that I'm using HTML and datetime stamps in my group headers, but that's how I need to make it for my client:

Screenshot 2025-09-22 at 08 27 39

Maybe it would be better do support HTML in the summarizers? So it matches with the group titles? @danharrin

Baspa:
Ah I got it working using only CSS. But if you want I still would like to add support for HTML in the summarizer @danharrin

danharrin:
Sure, I would accept a PR to allow HTML there as long as it isn't introducing new XSS ability into existing code

PR #17430, opened 2025-08-20
[4.x] Add a new text-direction extension to RichEditor

mohamedsabil83:
This PR is for adding two new toolbar buttons that control the direction of headings and paragraphs (RTL/LTR)


mohamedsabil83:
Hi @danharrin and @awcodes. Can you review this PR and let me know what's missing to save the change? For now, when clicking either the RTL or LTR button, the direction is applied as expected on the editor, but not saved to the database.

danharrin:
Hi, I don't think this is a standard rich editor extension, so I think it would be best as a Filament plugin instead of in core. You can do it without needing to rebuild the entire component

mohamedsabil83:
Any suggestion to save the applied directions? The direction changed entire editor, but cleared when saved to the DB.

danharrin:
You could save it as a class maybe?

mohamedsabil83:
Yeah, maybe. I'll try that and make a separate package. Thank you, @danharrin.

PR #17443, opened 2025-08-20
Hide blank column labels in the column manager

### LINKED ISSUE #17378: Toggleable columns list shows columns with empty labels
### Package

filament/filament

### Package Version

v4.0.1

### Laravel Version

v12.3.0

### Livewire Version

v3.6.4

### PHP Version

PHP 8.3.24

### Problem description

The list of toggleable columns includes columns without labels.

Image

### Expected behavior

It should not show columns with no label in the list.

### Steps to reproduce

Make a column in a table with `->label(false)` also `->toggleable()`

### Relevant log output

```shell

```

archilex:
Fixes #17378 

Not showing a column with a blank label in the column manager was easy enough to implement in `column-manager.blade.php`

However, for column groups to work properly, a column with a blank label _must_ be toggleable so that the entire group can be toggled off. The easiest way to enforce that was in the `canBeToggled()` method. I don't love comments in methods, but what I added seems so random that I felt a comment was needed. Whenever I backfill tests, I should be able to remove that comment and let the test catch it. 

Since we need to check if reorderable columns is enabled when the label is blank, I cached that in a new property so we don't have to evaluate it each time.

Finally, I went ahead and chose option A from my last comment in the issue. If I need to implement option B just let me know.

### DIFF
--- a/packages/tables/resources/views/components/column-manager.blade.php
+++ b/packages/tables/resources/views/components/column-manager.blade.php
@@ -57,7 +57,7 @@
             }}
         >
             <template
-                x-for="(column, index) in columns.filter((column) => ! column.isHidden)"
+                x-for="(column, index) in columns.filter((column) => ! column.isHidden && column.label)"
                 x-bind:key="(column.type === 'group' ? 'group::' : 'column::') + column.name + '_' + index"
             >
                 <div
@@ -104,7 +104,10 @@
                                 class="fi-ta-col-manager-group-items"
                             >
                                 <template
-                                    x-for="(groupColumn, index) in column.columns.filter((column) => ! column.isHidden)"
+                                    x-for="
+                                        (groupColumn, index) in
+                                            column.columns.filter((column) => ! column.isHidden && column.label)
+                                    "
                                     x-bind:key="'column::' + groupColumn.name + '_' + index"
                                 >
                                     <div
--- a/packages/tables/src/Columns/Concerns/CanBeToggled.php
+++ b/packages/tables/src/Columns/Concerns/CanBeToggled.php
@@ -36,6 +36,12 @@ trait CanBeToggled
             return false;
         }
 
+        // When a column label is blank, it must be toggleable so that
+        // column groups can be collectively toggled on/off.
+        if (blank($this->getLabel())) {
+            return true;
+        }
+
         return (bool) $this->evaluate($this->isToggleable);
     }
 
--- a/packages/tables/src/Concerns/HasColumnManager.php
+++ b/packages/tables/src/Concerns/HasColumnManager.php
@@ -2,6 +2,7 @@
 
 namespace Filament\Tables\Concerns;
 
+use Exception;
 use Filament\Schemas\Schema;
 use Filament\Support\Components\Component;
 use Filament\Tables\Columns\Column;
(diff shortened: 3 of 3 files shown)


packages/tables/src/Columns/Concerns/CanBeToggled.php
```diff
@@ -36,6 +36,12 @@ public function isToggleable(): bool
             return false;
         }
 
+        // When a column label is blank, it must be toggleable so that
+        // column groups can be collectively toggled on/off.
+        if (blank($this->getLabel())) {
```
danharrin, line 41: Does this need casting to a string like in the other place?

packages/tables/resources/views/components/column-manager.blade.php
```diff
@@ -57,7 +57,7 @@ class="fi-ta-col-manager-ctn"
             }}
         >
             <template
-                x-for="(column, index) in columns.filter((column) => ! column.isHidden)"
+                x-for="(column, index) in columns.filter((column) => ! column.isHidden && column.label)"
```
danharrin, line 60: I think we should check if the label is `'' or null or undefined` instead of truthy, otherwise it might break falsy string labels like `0`?

packages/tables/resources/views/components/column-manager.blade.php
```diff
@@ -104,7 +104,10 @@ class="fi-ta-col-manager-reorder-handle fi-icon-btn"
                                 class="fi-ta-col-manager-group-items"
                             >
                                 <template
-                                    x-for="(groupColumn, index) in column.columns.filter((column) => ! column.isHidden)"
+                                    x-for="
+                                        (groupColumn, index) in
+                                            column.columns.filter((column) => ! column.isHidden && column.label)
```
danharrin, line 109: Same here

packages/tables/src/Concerns/HasColumnManager.php
```diff
@@ -218,10 +219,16 @@ protected function mapTableColumnGroupToArray(ColumnGroup $group): array
      */
     protected function mapTableColumnToArray(Column $column): array
     {
+        $label = (string) $column->getLabel();
+
+        if (blank($label) && $this->hasReorderableTableColumns()) {
+            throw new Exception("The column [{$column->getName()}] has a blank label. Columns must have labels when reorderable columns is enabled.");
```
danharrin:
```suggestion
            throw new Exception("The table column [{$column->getName()}] has a blank label. All columns must have labels when they are reorderable.");
```

danharrin (changes requested):

packages/tables/src/Columns/Concerns/CanBeToggled.php
archilex: The other place was casting to a string because an html string wasn't supported in livewire. That doesn't apply here so this is fine...unless you prefer to cast for consistency.

packages/tables/resources/views/components/column-manager.blade.php
archilex: In my testing, and I tested again to be sure, `->label(0)` and `->label('0')` displays as expected. I can change if you prefer, but this works as is.

packages/tables/resources/views/components/column-manager.blade.php
archilex: See above

danharrin (approved):

PR #17448, opened 2025-08-21
feat: html support in group titles

Baspa:
## Description
This PR adds HTML support to Filament table group titles, allowing developers to use HTML in group headers for better customization.

I created a pull request earlier #17424 but it was rejected due to XSS vulnerability. This should be fixed in this PR. 

## Changes
   
   - Update `Group::getTitle()` to support `HtmlString` instances
   - Change Blade template to use `{!! !!}` instead of `{{ }}` for group titles
   - Allow HTML rendering in group titles for better customization
   
   ## Example Usage
   
   ```php
   ->getTitleFromRecordUsing(function (Order $record): HtmlString {
       $date = (new DateTime($record->delivery_date))->format('d F Y');
       $product = $record->product->name ?? '';
       $color = $record->product->color ?? '';
   
       if ($color) {
           return new HtmlString("$date – <span style='color: $color; font-weight: bold;'>$product</span>");
       }
   
       return new HtmlString("$date – $product");
   })
   ```

## Visual result
   
Screenshot 2025-08-20 at 10 36 51

   ## Testing
   
   - [x] Tested with HTML strings in group titles
   - [x] Verified backward compatibility with plain strings
   - [x] Ensured proper HTML escaping for security


danharrin (changes requested):
The "example usage" you specified looks correct, but your PR contains an additional `html()` method? I don't think we need an extra method. If an `HtmlString` is provided, we shouldn't need to sanitise it - if someone wants to inject a script, that is their choice. If a string is provided, we shouldn't treat it as HTML.

Baspa:
> The "example usage" you specified looks correct, but your PR contains an additional `html()` method? I don't think we need an extra method. If an `HtmlString` is provided, we shouldn't need to sanitise it - if someone wants to inject a script, that is their choice. If a string is provided, we shouldn't treat it as HTML.

Alright I will change the PR today, thanks @danharrin  :)

danharrin (changes requested):
Please replace references of `HtmlString` with just `Htmlable`

danharrin (approved):

maartenpaauw:
@Baspa @danharrin `null` was removed as return type in this PR. Was that on purpose?

Baspa:
> @Baspa @danharrin `null` was removed as return type in this PR. Was that on purpose?

Oops, that was not on purpose. @maartenpaauw

danharrin:
It is back in #17547

maartenpaauw:
Thanks for the quick responses! I didn't spotted the open PR (I should have looked better 😅).

PR #17457, opened 2025-08-21
[4.x] Update rich text editor to support merge tags with labels

torosegon:
## Description

Added support for merge tags with labels in the Rich Text Editor, making template fields more readable and easier to insert.
Labeled merge tags provide better clarity for end users and reduce the chance of inserting the wrong field.

## Functional changes


packages/forms/docs/10-rich-editor.md
```diff
@@ -422,6 +422,24 @@ Merge tags are surrounded by double curly braces, like `{{ name }}`. When the co
...
+RichEditor::make('content')
+    ->mergeTags([
+        [
+            'id' => 'name',
+            'label' => 'Full Name'
+        ],
+        [
+            'id' => 'today',
+            'label' => 'Date of Today'
+        ],
```
danharrin: Please use `id => Label` instead

packages/forms/resources/js/components/rich-editor.js
```diff
@@ -246,14 +246,14 @@ export default function richEditorFormComponent({
...
-        insertMergeTag(id) {
+        insertMergeTag(id, label = null) {
             editor
                 .chain()
                 .focus()
                 .insertContent([
                     {
                         type: 'mergeTag',
-                        attrs: { id },
+                        attrs: { id, label: label || id },
```
danharrin: Please ensure that the label is not saved anywhere in the DB, and if the merge tag is inserted the current label gets loaded only when you open the editor

danharrin (changes requested):

packages/forms/docs/10-rich-editor.md
torosegon: Solved.

packages/forms/resources/js/components/rich-editor.js
torosegon: I removed the `data-label` attribute, it will now always read the label from the existing tags based on the ID, if available.

packages/forms/src/Components/RichEditor/TipTapExtensions/MergeTagExtension.php
```diff
@@ -44,6 +44,10 @@ public function addAttributes(): array
                 'parseHTML' => fn ($DOMNode) => $DOMNode->getAttribute('data-id') ?: null,
                 'renderHTML' => fn ($attributes) => ['data-id' => $attributes->id ?? null],
             ],
+            'label' => [
+                'parseHTML' => fn ($DOMNode) => $DOMNode->getAttribute('data-label') ?: null,
+                'renderHTML' => fn ($attributes) => ['data-label' => $attributes->label ?? null],
+            ],
```
danharrin: Is this required if we aren't saving labels?

danharrin (changes requested):

packages/forms/src/Components/RichEditor/TipTapExtensions/MergeTagExtension.php
torosegon: I removed it in the meantime.

danharrin (approved):

PR #17478, opened 2025-08-22
[4.x] Include content method when creating a CustomPage

Patabugen:
## Description

It took me a lot of digging to understand how to render components in the blade filed of a custom page, in hindsight it's very straightforward because it's the same as any other page!

The setup for the content method is mirrored on https://github.com/filamentphp/filament/blob/7ea7b8e6f2e44eba6f2e85b8a850f4fa6947e3bd/packages/schemas/src/Commands/FileGenerators/LivewireSchemaComponentClassGenerator.php#L94

It felt intuitive to include a default content() method the same way other pages do, so this PR adds that.

https://filamentphp.com/docs/4.x/navigation/custom-pages

## Visual changes

None

## Functional changes

- [ ] `php artisan make:filament-page` command includes a `content()` method in the generated class and `{{ $this->content }}` in the view file

The documentation doesn't include any code examples of custom pages, so there's nothing to edit.

All the tests are passing and phpstan is happy.

### DIFF
--- a/packages/panels/src/Commands/FileGenerators/CustomPageClassGenerator.php
+++ b/packages/panels/src/Commands/FileGenerators/CustomPageClassGenerator.php
@@ -4,6 +4,7 @@ namespace Filament\Commands\FileGenerators;
 
 use Filament\Clusters\Cluster;
 use Filament\Pages\Page;
+use Filament\Schemas\Schema;
 use Filament\Support\Commands\FileGenerators\ClassGenerator;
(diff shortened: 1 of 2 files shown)


danharrin:
Hi, I think having a Blade view is more familiar for the majority of users instead of a schema, so is a better default for generation. I would definitely accept a PR that adds "using a schema to define page content" to the custom pages docs though. In fact, you don't need a Blade view at all when you are rendering a schema, as the default view renders it for you.

PR #17479, opened 2025-08-22
[4.x] Less generic exceptions. Exception -> LogicException

adelf:
`\Exception` is too generic. In most cases, you're throwing it because of invalid configuration and invalid parameters. There is a special class for these cases: `LogicException`. phpDoc comment for this class:

`Exception that represents error in the program logic. This kind of exception should lead directly to a fix in your code.`

There is also a checked/unchecked exceptions problem. It exists only in Java, but PHP also has `@throws` phpDoc, and we have to use it for Exception, but not for LogicException, since it's unchecked.

I made changes for "actions" and "forms" packages. Somewhere, I used a more specific `InvalidArgumentException`(which extends `LogicException`). If you're ok with this, I can do it for the rest of the packages.


danharrin:
Yes I would accept a PR to bring the other packages in line too.

adelf:
I did an incorrect sync. Closed in favor of #17579

PR #17481, opened 2025-08-22
fix: Cast ModalTableSelect HasMany state to strings to fix selection

pxlrbt:
When using `ModalTableSelect` with a `HasMany` relationship and preselected values, currently it shows "3 selected records" but doesn't select the actual records. 

That's because the we compare string values in the frontend, to our IDs which are int values if not casted. The error was already fixed for `BelongsToMany` but apparently overlooked for `HasMany`

## Functional changes

### DIFF
--- a/packages/forms/src/Components/ModalTableSelect.php
+++ b/packages/forms/src/Components/ModalTableSelect.php
@@ -269,8 +269,9 @@ class ModalTableSelect extends Field
                 $relatedRecords = $relationship->getResults();
 
                 $component->state(
-                    // Cast the related keys to a string, otherwise JavaScript does not
-                    // know how to handle deselection.
+                    // @js() in the frontend casts numbers to strings.
+                    // Therefore, we need to cast array values to strings, too.
+                    // Otherwise, selection and deselection will break.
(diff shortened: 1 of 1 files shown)


danharrin (approved):

packages/forms/src/Components/ModalTableSelect.php
```diff
@@ -301,9 +300,8 @@ public function relationship(string | Closure | null $name = null, string | Clos
                 $relatedRecords = $relationship->getResults();
 
                 $component->state(
-                    // @js() in the frontend casts numbers to strings.
-                    // Therefore, we need to cast array values to strings, too.
-                    // Otherwise, selection and deselection will break.
+                    // Cast the related keys to a string, otherwise JavaScript does not
+                    // know how to handle deselection.
```
pxlrbt, line 304: @danharrin I hoped to improve this comment a bit, adding the root cause of it (`@js`). Also this affects selection AND deselection. 

I was a bit confused by "otherwise JavaScript does not know how to handle deselection" because it doesn't describe the why.

PR #17484, opened 2025-08-22
[4.x] Fix SelectColumn crash when using objects as state values 

A909M:
### Problem:

  SelectColumn crashes with "Cannot access offset of type [ObjectType] on array"  
      error when using object instances as state values, particularly common with:    
      - Spatie Model States
      - Any object used as column state 
      
      
Screenshot from 2025-08-23 02-17-42

## Functional changes

### DIFF
--- a/packages/tables/src/Columns/SelectColumn.php
+++ b/packages/tables/src/Columns/SelectColumn.php
@@ -31,6 +31,7 @@ use Illuminate\Support\Js;
 use Illuminate\Support\Str;
 use Illuminate\Validation\Rule;
 use Livewire\Attributes\Renderless;
+use Stringable;
 use Znck\Eloquent\Relations\BelongsToThrough;
 
 use function Filament\Support\generate_search_column_expression;
@@ -373,7 +374,7 @@ class SelectColumn extends Column implements Editable, HasEmbeddedView
 
             if ($state instanceof BackedEnum) {
                 $state = $state->value;
-            } elseif (is_object($state)) {
+            } elseif ($state instanceof Stringable) {
                 $state = (string) $state;
             }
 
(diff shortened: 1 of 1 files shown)


packages/tables/src/Columns/SelectColumn.php
```diff
@@ -373,6 +373,8 @@ public function getOptionLabel(bool $withDefault = true): ?string
 
             if ($state instanceof BackedEnum) {
                 $state = $state->value;
+            } elseif (is_object($state)) {
```
danharrin: Is `$state instanceof Stringable` more appropriate? Does it work for you?

danharrin (changes requested):

packages/tables/src/Columns/SelectColumn.php
danharrin: Also, is there anything else in the Filament codebase that needs to be updated to support value objects like this? If yes, I am leaning towards not supporting them in general. If no, is there a reason why this is the only place where it is broken?

A909M: > Is `$state instanceof Stringable` more appropriate? Does it work for you?

Yes, `$state instanceof Stringable` is more appropriate and works perfectly. It's more specific. I've updated the code to use this.

A909M: > Also, is there anything else in the Filament codebase that needs to be updated to support value objects like this? If yes, I am leaning towards not supporting them in general. If no, is there a reason why this is the only place where it is broken?

This is the only place that needs this fix. Tables SelectColumn works directly with raw 
  model attributes, so it receives object instances from casts. When `getOptionLabel()` uses the 
  object as an array key (`$options[$state]`), it fails because objects can't be array keys.

danharrin (approved):

A909M:
Hi @danharrin 
 What about adding exception `throw new InvalidArgumentException("The cast object for
  the {$this->getName()} attribute must implement Stringable or be declared as string.");` if `$state` is
   object?

PR #17486, opened 2025-08-23
[4.x] Fixes file visibility to configured filesystem disk

achyutkneupane:
Continues #17453 

## Description

This PR updates the `BaseFileUpload` component’s `getDiskName()` method to align with configuration.

## Visual changes

- no UI changes.

## Functional changes


danharrin:
I've reflected the changes across the image column and image entry to ensure they require the same configuration.

I also added an exception to the rule, so that if you use the `public` disk but have `visibility('private')` explicitly, you are switched to the `local` disk.

danharrin (approved):

achyutkneupane:
> I've reflected the changes across the image column and image entry to ensure they require the same configuration.
> 
> I also added an exception to the rule, so that if you use the `public` disk but have `visibility('private')` explicitly, you are switched to the `local` disk.

I was also planning to do so. But, was not sure if that was necessary or not.
Thankyou @danharrin

PR #17491, opened 2025-08-23
fix: css styling in non-contained aside section

lukas-frey:
## Description

It seems that when using `Section::make()->columnSpanFull()->contained(false)->aside()`, the width of the Section is different in comparison to the contained version.

This PR aims to fix it by adding the missing CSS.

See screenshot below.

## Visual changes

Before PR:
image

After PR:
image

## Functional changes


lukas-frey:
I added another commit to fix collapsible child sections:

Before PR:
image

After PR:
image

The toggle button was positioned wrong when the section was not collapsed.

danharrin:
The code in the second commit doesn't make sense to me so I have removed that, but if you could send the code in a new PR alongside the schema PHP code you are using so I can check it out, that would be great.

danharrin (approved):

PR #17492, opened 2025-08-24
Fix `initialOptions` label rendering for integer based initial states

gehrisandro:
## Description

When the current state of a multi select contains integer based values which are not within the loaded options, instead of the label, the ID is rendered.

The reason for the issue is, that the value for all options is cast to a string by default: https://github.com/filamentphp/filament/blob/4.x/packages/forms/src/Components/Select.php#L152

If the state is now an array of integers, the value can not be found in the list of initial options:
https://github.com/filamentphp/filament/blob/4.x/packages/support/resources/js/utilities/select.js#L621

## Visual changes

Before:
CleanShot 2025-08-24 at 07 38 51@2x

After:
CleanShot 2025-08-24 at 07 40 08@2x

## Functional changes

- [ ] Documentation is up-to-date. _(no change required)_

### DIFF
--- a/packages/support/resources/js/utilities/select.js
+++ b/packages/support/resources/js/utilities/select.js
@@ -600,7 +600,7 @@ export class Select {
                 }
 
                 // If not found, add to missing values
-                missingValues.push(value)
+                missingValues.push(value.toString())
             }
         }
 
--- a/packages/support/resources/js/utilities/select.js
+++ b/packages/support/resources/js/utilities/select.js
@@ -51,6 +51,12 @@ export class Select {
         statePath = null,
         onStateChange = () => {},
     }) {
+        if (Array.isArray(state)) {
+            state = state.map((item) => item.toString())
+        } else if (state !== null) {
+            state = state.toString()
+        }
+
         this.element = element
         this.options = options
         this.originalOptions = JSON.parse(JSON.stringify(options)) // Keep a copy of original options
@@ -600,7 +606,7 @@ export class Select {
                 }
 
                 // If not found, add to missing values
-                missingValues.push(value.toString())
+                missingValues.push(value)
             }
         }
 
--- a/packages/support/resources/js/utilities/select.js
+++ b/packages/support/resources/js/utilities/select.js
@@ -51,12 +51,6 @@ export class Select {
         statePath = null,
         onStateChange = () => {},
     }) {
-        if (Array.isArray(state)) {
-            state = state.map((item) => item.toString())
-        } else if (state !== null) {
-            state = state.toString()
-        }
-
         this.element = element
         this.options = options
         this.originalOptions = JSON.parse(JSON.stringify(options)) // Keep a copy of original options
@@ -606,7 +600,7 @@ export class Select {
                 }
 
                 // If not found, add to missing values
-                missingValues.push(value)
+                missingValues.push(value.toString())
             }
         }
 


danharrin (changes requested):
I think we need to fix the value we pass to `initialOptions` instead. If the `$state` var is an array, we should map it to strings, if it is a singular value and not `blank()` we should also cast to a string. Then we don't need to accommodate in multiple places in the JS. Any thoughts? and does this fix the issue?

gehrisandro:
> I think we need to fix the value we pass to `initialOptions` instead. If the `$state` var is an array, we should map it to strings, if it is a singular value and not `blank()` we should also cast to a string. Then we don't need to accommodate in multiple places in the JS. Any thoughts? and does this fix the issue?

The `initialOptionLabels` and `options` are both already cast to strings as they are both passed through the same `transformOptionsForJsUsing()` hook: https://github.com/filamentphp/filament/blob/4.x/packages/forms/src/Components/Select.php#L148-L155

What we actually needed to do, if we don't want to modify the JS, is to cast the `$state` into a string or an array of strings. But this could potentially break the users' application.

In my opinion, we have to fix this on the JS side. Instead of my initial PR, we could update the condition to cast everything to a string before comparing them. Something like that:

```diff
  // https://github.com/filamentphp/filament/blob/4.x/packages/support/resources/js/utilities/select.js#L621

  if (
      filled(initialOption) &&
      initialOption.value !== undefined &&
      initialOption.label !== undefined &&
-     missingValues.includes(initialOption.value)
+     missingValues.filter((value) => value.toString() === initialOption.value.toString()).length > 0
  ) {
      // ...
  }
```

The drawback here is that the cast has to happen over and over again.
But in return, it would even work if a user overrides the `transformOptionsForJsUsing()` and would not cast the options value to a string.

danharrin:
Why would casting the state to strings break the user's apps? We actually do the same when you are using `relationship()` when the state is loaded from the database keys

gehrisandro:
@danharrin Thank you for pointing out the handling in the `relationship()`. This actually helped me, to find the root cause of my problem.

What I did in Filament v3 is to fill the form with a list of IDs:
```php
$this->getSchema('form')->fill([
    'tags' => $this->taggable->userTags->pluck('id')->toArray(),
]);
```

In Filament v4 we now have to cast the IDs to strings:
```php
$this->getSchema('form')->fill([
    'tags' => $this->taggable->userTags->pluck('id')->map(fn (int $id): string => strval($id))->toArray(),
]);
```

I still think we should improve the JS to work with strings and integer based values, instead of casting the state to strings behind the scenes, which feels magically and unexpected to me.

If you think this isn't a relevant issue at all, feel free to close this PR.

danharrin:
I think we can make it easier by casting the state values to strings on the PHP side before they get passed to JS, probably somewhere in the view file. What do you think?

gehrisandro:
I gave it a try to cast the state to strings, but this isn't very easy. We would have to check for `Enums`, `Htmlable`, etc.
Basically everything that Laravel does here: https://github.com/laravel/framework/blob/12.x/src/Illuminate/Support/Js.php#L67

danharrin:
Maybe do it in the constructor of the JS Select object then?

gehrisandro:
Updated the PR to convert the state in the select constructor.

danharrin:
I'm really sorry, I think I have changed my mind about the fix. If we change the `state` variable, it will queue a Livewire update when the field is loaded to convert the ints to strings, which isn't ideal if the field is live as it might make an immediate network request.

Based on your initial fix, are there any other places that need updating to work correctly with int/float keys instead of strings?

gehrisandro:
> I'm really sorry, I think I have changed my mind about the fix. If we change the `state` variable, it will queue a Livewire update when the field is loaded to convert the ints to strings, which isn't ideal if the field is live as it might make an immediate network request.
> 
> Based on your initial fix, are there any other places that need updating to work correctly with int/float keys instead of strings?

No problem. I've just reverted my commit from before to return to my initial solution.

As far as I am aware of, there are no other changes necessary.

danharrin (approved):

danharrin:
Thanks, sorry for the extra work here

PR #17496, opened 2025-08-24
Allow bypassing failed incompatible plugin check during upgrade

chrisreedio:
## Description

The current upgrade script will now allow the user to bypass a failing compatible plugin check.

## Visual changes

image

---
### Aborted Upgrade output

image

## Functional changes

Since this is not a flag and is a prompt, it seems no additional documentation would be required.

### DIFF
--- a/packages/upgrade/bin/filament-v4
+++ b/packages/upgrade/bin/filament-v4
@@ -274,6 +274,11 @@ try {
             $version = ltrim($compatibility['version'], 'v');
             $isPrerelease = $compatibility['isPrerelease'];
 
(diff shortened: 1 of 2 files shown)


danharrin:
Thanks! This seemed to work and I can't see why it was marked as draft specifically

danharrin (approved):

chrisreedio:
Was mainly this one: `Changes have been tested to not break existing functionality.`

It works for me but didn't know if I was missing anything as there doesn't really seem to be a good way to write tests for this?

Glad it's working for you and glad I won't have to use my local version of it anymore :)

PR #17505, opened 2025-08-25
[4.x] Remove HasExtraInputAttributes trait from RichEditor

mathieutu:
Hi,
Thanks for your amazing work!

## Description

The component doesn't use the `extraInputAttributes` anymore. I had a regression while migrating.
No need to add the methods that are not used, and so the people using them in the past will now have an error while migrating.

## Visual changes
N/A

Screenshot 2025-08-25 at 17 07 44

## Functional changes


danharrin:
I don't think we can remove this trait, people might already be using the method in production and not realise it was silently broken before.

Can we instead add the method back somewhere in the rich editor Blade view, so you can control the min height / max height like you have in the example?

mathieutu:
Done in https://github.com/filamentphp/filament/pull/17524

PR #17507, opened 2025-08-25
Add a method to lazy load RichEditor images

CharlieEtienne:
## Description

This PR adds a new `lazyLoadImages` method to RichContentRenderer.

```php
use Filament\Forms\Components\RichEditor\RichContentRenderer;

RichContentRenderer::make($record->content)
    ->lazyLoadImages()
    ->toHtml()
```

This will simply add the `loading="lazy"`attribute to all images in RichEditor content:

```html
<img src="https://example.com/storage/img.jpg" loading="lazy">
```

## Functional changes

### DIFF
--- a/packages/forms/docs/10-rich-editor.md
+++ b/packages/forms/docs/10-rich-editor.md
@@ -151,6 +151,16 @@ RichContentRenderer::make($record->content)
     ->toHtml()
 ```
 
+If you want to lazy-load images, you can pass the `lazyLoadImages` option to the renderer:
+
+```php
+use Filament\Forms\Components\RichEditor\RichContentRenderer;
+
+RichContentRenderer::make($record->content)
+    ->lazyLoadImages()
+    ->toHtml()
+```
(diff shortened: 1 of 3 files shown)


danharrin:
We could add endless methods to customize the output like this, so maybe we just need a method where you can tap into the editor:

```php
use Filament\Forms\Components\RichEditor\RichContentRenderer;

RichContentRenderer::make($record->content)
    ->processNodesUsing(function (object &$node): void {
        if ($node->type !== 'image') {
            return;
        }
    
        $node->attrs->loading = 'lazy';
    })
    ->toHtml()
```

Would accept a PR for that

CharlieEtienne:
It would be nice indeed, but in my case I also had to modify `ImageExtension`. Would we need another method to tap into `addAttributes` methods?

danharrin:
Feel free to modify the extension as part of your PR, it seems like a sensible addition anyway

PR #17516, opened 2025-08-26
fix: remove sidebar collapse button rotation logic

ahsanmahmood09:
fix: prevent double rotation of sidebar collapse button

Removed the JavaScript code that was manually adding 'rotate-180' class to
sidebar group collapse buttons. This was causing the chevron icons to rotate
180 degrees twice when using the `collapsed()` method on NavigationGroup,
resulting in incorrect orientation.

The rotation is now handled properly by Filament's built-in functionality
when using the `collapsed()` method, making the manual rotation redundant
and problematic.

### DIFF
--- a/packages/panels/resources/views/components/sidebar/index.blade.php
+++ b/packages/panels/resources/views/components/sidebar/index.blade.php
@@ -165,9 +165,6 @@
                     group.querySelector(
                         '.fi-sidebar-group-items',
                     ).style.display = 'none'
-                    group
-                        .querySelector('.fi-sidebar-group-collapse-button')
-                        .classList.add('rotate-180')
                 })
         </script>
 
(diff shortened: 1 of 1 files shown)


danharrin:
I don't think I understand why this would be a problem to be honest. Can you please open an issue with a reproduction repository so I can check out what is happening?

ahsanmahmood09:
Thanks for getting back to me @danharrin I have created the repo with the issue description. Here is the link
https://github.com/filamentphp/filament/issues/17552

Please have a look and let me know your thoughts. thanks

danharrin (approved):
Corrected the class name which caused the root of the problem

ahsanmahmood09:
Thanks a lot @danharrin appreciate it.

PR #17524, opened 2025-08-26
fix: merge extraInputAttributes in RichEditor

mathieutu:
## Description

With v4, the `extraInputAttributes` were allowed, but not used anymore.

Following https://github.com/filamentphp/filament/pull/17505#issuecomment-3221723627, this PR apply the given attributes to the component, so migrating from v3 to v4 will be without any regression.

Thanks.

## Visual changes

Extra input attributes are reapplied.

## Functional changes

### DIFF
--- a/packages/forms/resources/views/components/rich-editor.blade.php
+++ b/packages/forms/resources/views/components/rich-editor.blade.php
@@ -19,6 +19,7 @@
         x-cloak
         :attributes="
             \Filament\Support\prepare_inherited_attributes($extraAttributeBag)
+                ->merge($getExtraInputAttributes())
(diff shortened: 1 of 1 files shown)


danharrin:
Is there a better element inside this component than the outer element, for example the element that represents the content area below the toolbar? When setting the input height in v3, did it also take the toolbar into account or not?

mathieutu:
@danharrin You're right. After comparing with v3, the inputAttributes applied directly on the content.

Sorry for going back and forth.

I've made the change:

V3:
Screenshot 2025-08-27 at 12 33 02


V4:
Screenshot 2025-08-27 at 12 32 38

PR #17526, opened 2025-08-26
introduce processNodesUsing method

CharlieEtienne:
## Description

This PR aims to introduce a way to customize TipTap nodes. It could be used to add or change HTML attributes.

```php
use Filament\Forms\Components\RichEditor\RichContentRenderer;

RichContentRenderer::make($record->content)
  ->processNodesUsing(function (object &$node): void {
      if ($node->type !== 'image') {
          return;
      }
  
      $node->attrs->loading = 'lazy';
  })
  ->toHtml()
```

## Functional changes


packages/forms/src/Components/RichEditor/RichContentRenderer.php
```diff
@@ -236,6 +238,22 @@ protected function processMergeTags(Editor $editor): void
         });
     }
 
+    public function processNodesUsing(Closure $callback): static
+    {
+        $this->nodeProcessor = $callback;
```
danharrin: I think you should be able to call the method multiple times to create a chain of processors

danharrin (changes requested):

CharlieEtienne:
My initial goal was to let users add new attributes dynamically. 

Unfortunately this won't be possible for now. We would have to modify `addAttributes` method in each extension to add these attributes. Or at least I didn't find a way to automate that... Maybe something could be done in tiptap-php directly...
Full discussion in Discord: https://discord.com/channels/883083792112300104/1409544539353518242

Anyway, this method opens up some possibilities

danharrin (changes requested):
If you want to add the lazy attribute to the image extension in this PR, go ahead

packages/forms/src/Components/RichEditor/TipTapExtensions/ImageExtension.php
```diff
@@ -32,6 +32,13 @@ public function addAttributes(): array
                 'parseHTML' => fn ($DOMNode) => $DOMNode->getAttribute('data-id') ?: null,
                 'renderHTML' => fn ($attributes) => ['data-id' => $attributes->id ?? null],
             ],
+            'loading' => [
+                'default' => null,
+                'parseHTML' => fn ($DOMNode) => $DOMNode->getAttribute('loading') ?: null,
+                'renderHTML' => fn ($attributes) => [
+                    'loading' => $attributes->loading ?? null,
+                ],
+            ],
```
danharrin:
```suggestion
            'loading' => [],
```

Does this work?

packages/forms/src/Components/RichEditor/RichContentRenderer.php
```diff
@@ -236,6 +241,24 @@ protected function processMergeTags(Editor $editor): void
...
+        $this->nodeProcessors[] = $callback;
+
+        return $this;
+    }
+
+    protected function processNodes(Editor $editor): void
+    {
+        if (empty($this->nodeProcessors)) {
+            return;
+        }
```
danharrin: I don't think we really need this, if the array is empty this doesn't do anything anyway

danharrin (changes requested):

packages/forms/src/Components/RichEditor/TipTapExtensions/ImageExtension.php
CharlieEtienne: yes. That was my initial code, then I copied adam's one, thinking he was seeing something I missed. But it seems to work fine with an empty array or even an empty string.

packages/forms/src/Components/RichEditor/RichContentRenderer.php
CharlieEtienne: sure, I removed it, thanks

danharrin (approved):

PR #17528, opened 2025-08-26
Added getResetPasswordNotificationClass function to facilitate future extension

tonegabes:
refact(password-reset): refact class RequestPasswordReset to make it easier to customize the reset password notification without override the whole method.

## Description

- Replaced the direct use of ResetPasswordNotification with a method that returns the notification class.
- Added getResetPasswordNotificationClass function to facilitate future extension.

## Functional changes


danharrin:
I think you should instead use a service provider to `app()->bind(OldNotification::class, NewNotification::class)` instead :)

PR #17535, opened 2025-08-27
fix: live debounce logic for rich editor

mathieutu:
## Description

Hi,
The debounce feature for the rich editor is broken in v4.
Indeed, `Alpine.debounce` returns a function that should be stored outside the event callback (because it has an internal state) and called when the update is done.

This PR:
- Stores the `debouncedCommit` function outside the editor.on('update') callback, ensuring that the debounce logic is not re-created on every update event (if it were inside the callback, a new debounced function would be instantiated each time, defeating the purpose of debouncing).
- Calls the function when an update is done, instead of just re-creating the function.

Thanks for your work,
Mathieu.

## Visual changes
N/A

## Functional changes


danharrin:
I think this is related to #17083. Can you check please @angus-mcritchie ?

mathieutu:
Indeed https://github.com/filamentphp/filament/pull/17083 introduced the bug.

@angus-mcritchie any idea when you could check it? We're waiting for this PR to be merged to deploy our migration to v4 (and the features developed afterward). 
Thanks.

angus-mcritchie:
@mathieutu I've just checked the bug #17083 fixed and it still works with these changes. Nice work mate!

All good by me @danharrin 👍

danharrin (approved):

PR #17538, opened 2025-08-27
Adding an explanation for the database notification queue target.

obaume:
## Description

Hi, I wanted to change the queue used for the database notifications. I tried finding relevent information in the documentation but didn't find anything. 
So i explored the code base and found my answer. I wanted to put it in the documentation to save other people some time in the future.
## Visual changes
N/A

## Functional changes
- [x] Code style has been fixed by running the composer cs command.


danharrin:
Thanks, but I feel like this is more of a Laravel thing and not something we should add to the Filament docs. We use the same database notifications driver as Laravel.

Issue #17375, opened 2025-08-18
V4 Table SelectColumn Does not update if the existing value is not in the provided options


bvboas:
### Package

filament/filament

### Package Version

V4.0.1 

### Laravel Version

12

### Livewire Version

3

### PHP Version

83.

### Problem description

When using a SelectColumn in a table, we can't update the value if the current model value is not in the options. That means, if the DB has a NULL value, we won't be able to set the any new value.

```php
SelectColumn::make('name')->options(['User Name' => 'One', 'Another Name' => 'Second']),
```
If the current record value is different from "User Name" or "Another Name", an error will be displayed:

```
The selected :attribute is invalid.
```

### Expected behavior

We should be able to update to any new value that is present in the 
```
->options()
```

array

### Steps to reproduce

Add a  SelectColumn with options, in which none is the current record value
```php
SelectColumn::make('name')->options(['User Name' => 'One', '2' => 'Second']),
```

<img width="2194" height="702" alt="Image" src="https://github.com/user-attachments/assets/39f48472-a9cd-4deb-8f35-0f781660f119" />

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/bvboas/filamentv4beta

### Relevant log output

```shell

```

envoyr:
I have the same issue.

PR #17540, opened 2025-08-27
fix column update if current state is null and not present in options…

envoyr:
## Description

This will address the issue #17375 and fix the column update if current state is null and not present in options.

## Visual changes

none

## Functional changes


danharrin:
Replaced by #17612

Issue #17546, opened 2025-08-27
[4.x] Grouping with null values


moory-se:
### Package

filament/filament

### Package Version

v4.0.4

### Laravel Version

v12.26.3

### PHP Version

PHP 8.3.0

### Problem description

When grouping tables where the group value might be null, the following error is thrown:

Filament\Tables\Grouping\Group::getTitle(): Return value must be of type Illuminate\Contracts\Support\Htmlable|string, null returned

In V3, this was perfectly fine. It did group all null values in one "group". I do believe this is the most intuitive way to solve it.

### Expected behavior

To keep the V3 behaviour where null-values are still grouped as one and that the table doesnt crash.

### Steps to reproduce

To reproduce this, simply add 
`->groups([
      'a_column_that_is_sometimes_null'
  ])` 
and make sure that the table contains atleast 1 row with a null value. Please note that empty strings also crasches (with another error).

### Reproduction repository (issue will be closed if this is not valid)

not-applicable

### Relevant log output

```shell

```

PR #17547, opened 2025-08-27
Allow null as title

moory-se:
Re-adds the possibility to group by null values

## Description

In V3, one could group on a field that was nullable. In V4, this breaks. This is because in V3 (https://github.com/filamentphp/filament/blob/3.x/packages/tables/src/Grouping/Group.php#L253), the getTitle() had null as a return type. This pull request re-adds null as a accepted return type.

This solves #17546.

## Functional changes

### DIFF
--- a/packages/tables/src/Grouping/Group.php
+++ b/packages/tables/src/Grouping/Group.php
@@ -250,7 +250,7 @@ class Group extends Component
         return Arr::get($record, $this->getColumn());
     }
 
-    public function getTitle(Model $record): string | Htmlable
+    public function getTitle(Model $record): ?(string | Htmlable)
     {
         $column = $this->getColumn();
 
--- a/packages/tables/src/Grouping/Group.php
+++ b/packages/tables/src/Grouping/Group.php
@@ -250,7 +250,7 @@ class Group extends Component
         return Arr::get($record, $this->getColumn());
     }
 
-    public function getTitle(Model $record): ?(string | Htmlable)
+    public function getTitle(Model $record): string | Htmlable | null
     {
         $column = $this->getColumn();
 


packages/tables/src/Grouping/Group.php
```diff
@@ -250,7 +250,7 @@ public function getKey(Model $record): mixed
         return Arr::get($record, $this->getColumn());
     }
 
-    public function getTitle(Model $record): string | Htmlable
+    public function getTitle(Model $record): ?(string | Htmlable)
```
danharrin: This is not valid PHP. It is very important to test your changes locally before you PR them. Have you tested that this is the only place that needs to be updated in order to support null groups?

```suggestion
    public function getTitle(Model $record): string | Htmlable | null
```

danharrin (changes requested):

packages/tables/src/Grouping/Group.php
moory-se: Sorry. I did this through AI.

I have now tested it locally (with your change) and it works. It do crash on empty strings, but that's also the case in V3. I think that this behaviour (allowing null) is not really intended so I'm quite sure the tests aren't perfect for this case. Still do believe that "passing tests and not breaking stuff between V3 and V4" is better than doing nothing, even if it's worse than adding new tests and documenting this behaviour.

philmareu: We've implemented this fix for a client's ERP today and they have reported everything is back to normal. @moory-se are you able to make the suggested update? I would be glad to help if needed.

moory-se: @philmareu I changed the PR now and hope to get it accepted :)

maartenpaauw:
Thank you for this PR! The removal of `null` return type was accidentally removed while adding `Htmlable` return type. This will revert that change. Code looks good!

moory-se:
Great. Fingers crossed this is merged soon so that we can go live with our V4 branch :)

danharrin (approved):

PR #17551, opened 2025-08-28
Fix FileUpload component initialization in tabs

danielreales7:
- Change x-load attribute from bare 'x-load' to 'x-load="visible"' in non-SPA mode
- Ensures FilePond components initialize properly when placed inside tabs
- Resolves issue where file uploads in secondary tabs would show background color instead of image previews
- Maintains consistency with SPA mode behavior which already uses x-load="visible || event (ax-modal-opened)"

## Functional changes

### DIFF
--- a/packages/forms/resources/views/components/file-upload.blade.php
+++ b/packages/forms/resources/views/components/file-upload.blade.php
@@ -28,7 +28,7 @@
         @if (FilamentView::hasSpaMode())
             {{-- format-ignore-start --}}x-load="visible || event (ax-modal-opened)"{{-- format-ignore-end --}}
         @else
-            x-load
+            x-load="visible"
         @endif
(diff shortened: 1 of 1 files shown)


danharrin:
I don't think this is really the right solution to be honest. The Async Alpine JS should be fetched immediately otherwise there is a waiting period for it download, but we may need add some (intersection?) observer checks in the file upload JavaScript which initialise the re-component when it becomes visible, if we detect that it is broken.

threesil:
I have the following:

```
Tab::make(__('attributes.gallery'))
    ->schema([
        SpatieMediaLibraryFileUpload::make('shop_product_gallery')
            ->label(__('attributes.gallery'))
            ->multiple()
            ->panelLayout('grid'),
    ])
```
Images aren't loaded when i open the tab.

When i remove the `->panelLayout('grid')` the images are loaded.

PR #17568, opened 2025-08-29
Update Form docs, to use better alternative

ropi-bc:
## Description

The function (`dehydrateStateUsing()`) mentioned for customizing dehydrated fields hides a landmine in Filament 4. 

What we ran into, is we need a few fields never to have null values, but to be empty strings instead.

Because of this https://github.com/filamentphp/filament/commit/571a0f16d9006147ea9a88b94ec9d30a99075999 commit empty strings are reverted back to null. If you use `mutateDehydratedStateUsing()` you can circumvent this issue

## Functional changes


ropi-bc:
I don't know if this proposal is applicable to the password hashing part too. The hash function should never return an empty string, that's why I didn't change that section.

danharrin:
Original issue resolved in #17789, please let me know if it doesn't help

PR #17572, opened 2025-08-29
feat: support Model instances in getRecords

jaulz:
## Description

Now, that Filament has custom data sources it would be nice to not only return arrays from the custom data source but also handle a list of different models (e.g. I want to display two models that are very similar).

Once you think, this is a good idea, I am happy to write a test for it 😊 

## Visual changes

n/a

## Functional changes


danharrin (approved):

jaulz:
@danharrin thanks!

PR #17574, opened 2025-08-30
Apply Explicit Bool Comparison for $failedRowsCount check

alisalehi1380:
## Description

Replaced the implicit truthy check of `$failedRowsCount` with an explicit comparison to 0.

This change was made according to [ExplicitBoolCompareRector rules](https://getrector.com/rule-detail/explicit-bool-compare-rector), to improve readability and avoid potential unintended truthy evaluations.

```diff
- if ($failedRowsCount = $import->getFailedRowsCount()) {
+ if (($failedRowsCount = $import->getFailedRowsCount()) !== 0) {
```

Reasoning:

-Makes the intention of checking “non-zero failed rows” explicit.
-Prevents ambiguity when the value might be 0, null, or other falsy values.
-Improves code clarity and complies with Rector’s explicit boolean comparison principle.

## Functional changes


danharrin:
`getFailedRowsCount()` has an `int` return type, so there is no way it can return a falsy value such as `null` or `''`. I think this suggestion negatively affects code readability.

PR #17588, opened 2025-09-01
Skip rendering Livewire component on certain test methods

gehrisandro:
## Description

This PR skips Livewire rendering on certain methods used while testing Filament forms. This improves the testing performance for Filament resources by around 15%.

The reason is that `TestForms::fillForm()` currently triggers three unnecessary renderings of the Livewire component.

## Visual changes

There are no visual changes.

## Functional changes

- [ ] ~~Documentation is up-to-date.~~ (no change required)

### DIFF
--- a/packages/forms/resources/views/components/rich-editor.blade.php
+++ b/packages/forms/resources/views/components/rich-editor.blade.php
@@ -66,7 +66,9 @@
                 </div>
             @endif
 
-            <div {{ $getExtraInputAttributeBag()->class(['fi-fo-rich-editor-main']) }}>
+            <div
+                {{ $getExtraInputAttributeBag()->class(['fi-fo-rich-editor-main']) }}
+            >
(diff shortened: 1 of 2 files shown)


danharrin:
Unfortunately I think this could break some tests. If elements of the HTML depend on the state of the form, it would result in `assertSee()` or `assertSeeHtml()` calls breaking. Anything backend rendered that relies on the form state could be affected.

gehrisandro:
@danharrin I don't think this is the case, as the component is rendered anyway at the end of the `fillForm()` method. See https://github.com/filamentphp/filament/blob/4.x/packages/forms/src/Testing/TestsForms.php#L69

My PR only prevents multiple renderings within a single `fillForm()` call.

danharrin:
Fair enough!

danharrin (approved):

PR #17589, opened 2025-09-01
Fix required validation on RichEditor

### LINKED ISSUE #17342: v4: Rich editor: required() validation not working
### Package

filament/filament

### Package Version

v4.0.1

### Laravel Version

v12.24.0

### Livewire Version

v3.6.4

### PHP Version

8.4.11

### Problem description

The `required()` validation rule on RichEditor is not working. The field label is properly marked as required (with the asterisk), but submitting the form with no value in the field will not trigger the validation error.

The error is reproducible even in the Filament demo: https://demo.filamentphp.com/blog/posts/create

### Expected behavior

It should not possible to submit the form and an error message should be visible.

### Steps to reproduce

1. Open the Livewire demo here: https://demo.filamentphp.com/blog/posts/create
2. Fill out all the required fields except the "Content".
3. Hit "Create"
4. It stores the new blog post with an empty content

### Relevant log output

```shell

```

gehrisandro:
## Description

This PR introduces an implementation of the `required()` rule for the `RichEditor` as the default implementation does not work, because an empty editor is technically not empty. It always contains at least an empty paragraph.

This fixes https://github.com/filamentphp/filament/issues/17342

As discussed in the issue, we need to check if the rendered HTML is an empty paragraph.

## Visual changes

With this change, the missing validation message appears and form submission is prevented.

## Functional changes

- [ ] ~~Documentation is up-to-date.~~ (No change required)

### DIFF
--- a/packages/forms/resources/views/components/rich-editor.blade.php
+++ b/packages/forms/resources/views/components/rich-editor.blade.php
@@ -66,7 +66,9 @@
                 </div>
             @endif
 
-            <div {{ $getExtraInputAttributeBag()->class(['fi-fo-rich-editor-main']) }}>
+            <div
+                {{ $getExtraInputAttributeBag()->class(['fi-fo-rich-editor-main']) }}
+            >
                 <div class="fi-fo-rich-editor-content fi-prose" x-ref="editor">
                     @foreach ($floatingToolbars as $nodeName => $buttons)
                         <div
--- a/packages/forms/src/Components/Concerns/CanBeValidated.php
+++ b/packages/forms/src/Components/Concerns/CanBeValidated.php
@@ -762,7 +762,7 @@ trait CanBeValidated
         return $this->evaluate($this->regexPattern);
     }
 
-    public function getRequiredValidationRule(): string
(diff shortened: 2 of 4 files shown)


danharrin (changes requested):
Maybe the empty paragraph check should be on the array content instead of `getHtml()` so its faster?

gehrisandro:
We use such an array check in our project, and it works fine, but this is maybe not always the case if extensions modify the array state but still render to an empty paragraph.

But if you prefer, I can easily update the PR to check for the array content.

danharrin:
I just worry that calling `getHtml()` without passing all the info the extensions need to properly render (custom block data, merge tag data, but it could be the same for other custom extensions) could cause the validation to fail. I would rather the validation passed when it shouldn't do, in this case.

So yes, please make it a simple array check.

gehrisandro:
The PR is updated.

The code to check the array looks a bit complicated, I know. But we can not check for a fixed array structure as it may contain other keys, for example information about font weight etc.

danharrin (approved):

danharrin:
Thanks!

PR #17590, opened 2025-09-01
fix: More predictable key inheritance

danharrin:
v4 made it more difficult to understand the key structure of components and actions inside schemas, by providing default keys for components like Section.

These keys are now not inherited by child components and actions.

This could break existing test suites that were changed during the v4 upgrade to work with the v4 key structure, but it will bring it more in line with v3 which should make future upgrades easier, and make the tests more readable.

I see this as more of a fix for an unintended v4 breaking change, rather than a breaking change itself. I did not intend for people to need to prefix their field/action assertions with the schema heading for example.


danharrin:
This could break existing test suites that were changed during the v4 upgrade to work with the v4 key structure, but it will bring it more in line with v3 which should make upgrades easier.

Mrkbingham:
Amazing, thanks @danharrin !

PR #17596, opened 2025-09-01
Rich editor content toText() renderer

tanthammar:
## Description

Sometimes you need the TipTap content as text without html formatting. Using the TipTap editors getText() function returns a better result than php `strip_tags()`.

This PR adds `toText()` to the `RichContentRenderer`

Example use case:

```php
TextColumn::make('description')
    ->formatStateUsing(fn ($record) => RichContentRenderer::make($record->description)->toText())
    ->limit(30),
```

## Functional changes


packages/forms/src/Components/RichEditor/RichContentRenderer.php
```diff
@@ -350,6 +350,17 @@ public function toHtml(): string
         return Str::sanitizeHtml($this->toUnsafeHtml());
     }
 
+    public function toText(): string
+    {
+        $editor = $this->getEditor();
+
+        $this->processCustomBlocks($editor);
+        $this->processFileAttachments($editor);
```
danharrin: I think these two lines probably aren't relevant for text output, right?

danharrin (changes requested):

packages/forms/src/Components/RichEditor/RichContentRenderer.php
tanthammar: done

danharrin (approved):

PR #17602, opened 2025-09-01
Fix placeholder alignment

CharlieEtienne:
## Description

This PR fixes an issue in TextColumn where placeholders weren't correctly aligned.
(Reported by luizcristino on discord)

To reproduce the bug just do the following:

```php
TextColumn::make('myFakeColumn')
    ->alignEnd()
    ->placeholder('Placeholder'),
```

The css classes like `fi-align-*` weren't in the output because the `blank($state)` condition is evaluated too early, before the alignment adds classes to attributes.

The fix is simply to move the alignment classes logic above this condition.

### BEFORE
before 2025-09-01 at 21 43 45

### AFTER
after 2025-09-01 at 21 43 30

## Functional changes

### DIFF
--- a/packages/tables/src/Columns/ColorColumn.php
+++ b/packages/tables/src/Columns/ColorColumn.php
@@ -23,10 +23,13 @@ class ColorColumn extends Column implements HasEmbeddedView
             $state = $state->all();
         }
 
+        $alignment = $this->getAlignment();
+
         $attributes = $this->getExtraAttributeBag()
             ->class([
                 'fi-ta-color',
                 'fi-inline' => $this->isInline(),
+                ($alignment instanceof Alignment) ? "fi-align-{$alignment->value}" : (is_string($alignment) ? $alignment : ''),
             ]);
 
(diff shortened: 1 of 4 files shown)


danharrin (changes requested):
I don't think `fi-ta-text-has-badges` should be applied to placeholders, so that needs to be merged in its original position

Also does `fi-wrapped` affect placeholders? If not, move that back too

danharrin (changes requested):
Please reflect the change to `ColorColumn`, `IconColumn` and `ImageColumn` in this PR

Please also do the same for `fi-in-placeholder` in infolist entries, as long as infolists also apply alignment correctly to the placeholder?

CharlieEtienne:
I pushed the changes for ColorColumn, IconColumn and Image Column.

Regarding infolists entries, the alignment classes are applied to the parent `fi-in-entry-content`, so it displays correctly.

danharrin (approved):

PR #17603, opened 2025-09-01
feat(translations): add support for heading and description translations

jeffersongoncalves:
Introduce `translateHeading` and `translateDescription` methods to handle conditional translation for headings and descriptions. Enhancements provide greater flexibility in localization by automatically translating string values when enabled.


danharrin:
Hi, sorry this is not a feature that I think provides a lot of enhancement over just `__()`

PR #17608, opened 2025-09-02
Introduced and tested stacked/pipeline dehydrateStateUsing

edstevo:
This pull request proposes an enhancement to how dehydrateStateUsing() functions can be applied in Filament form components.

Currently, calling dehydrateStateUsing() on a form field replaces any previously set callback. This makes it difficult to compose multiple transformations, such as removing whitespace and capitalising text for a single input.

For example, consider a postcode input where we want to:
	•	Remove extra spaces (" wo98  0AJ" → "WO98 0AJ")
	•	Capitalise the result

Using the current approach, only one transformation can be applied at a time. My proposed change allows stacking multiple dehydration callbacks, like a pipeline, which makes code more readable and reusable.

This aligns with the goal of improved code quality and reusability, as outlined in the Filament docs:
https://filamentphp.com/docs/4.x/resources/code-quality-tips

*Before*
```
TextInput::make('post_code')
    ->dehydrateStateUsing(fn ($state) => Str::squish($state));
```

*After (proposed)*
```
TextInput::make('post_code')
    ->removeWhitespace()
    ->capitalise();
```

This is enabled by extending the TextInput component and allowing multiple dehydration callbacks to be queued:

```
class TextInput extends BaseTextInput
{
    public function removeWhitespace(): static
    {
        return $this->appendDehydrateStateUsing(fn ($state) => Str::squish($state));
    }

    public function capitalise(): static
    {
        return $this->appendDehydrateStateUsing(fn ($state) => Str::upper($state));
    }
}
```

## Visual changes

None.

## Functional changes

- [ ] Introduced the ability to stack multiple dehydrateStateUsing() callbacks, enabling more composable field behaviour.
- [ ] This allows more expressive and reusable custom field components in extended schema classes.
- [ ] Documentation updated.

### DIFF
--- a/packages/forms/docs/01-overview.md
+++ b/packages/forms/docs/01-overview.md
@@ -1257,6 +1257,42 @@ TextInput::make('name')
     ->dehydrateStateUsing(fn (string $state): string => ucwords($state))
 ```
 
+#### Stacking dehydration functions
+
+You can also stack multiple dehydration functions using the `dehydrateStateUsing()` method.
+
+```php
+use Filament\Forms\Components\TextInput;
+
+TextInput::make('name')
+    ->required()
+    ->dehydrateStateUsing(fn (string $state): string => trim($state))
+    ->dehydrateStateUsing(fn (string $state): string => ucwords($state))
+```
+
+This is useful if you want to extract common logic into component classes, and then stack them on top of each other. For example:
+
+```php
+use Illuminate\Support\Str;
+
+class CustomTextInput extends TextInput
+{
+    public function removeWhitespace(): static
+    {
+        return $this->dehydrateStateUsing(fn ($state) => Str::squish($state));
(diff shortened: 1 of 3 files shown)


danharrin:
Hi, unfortunately we cannot introduce a breaking change like this. Currently users might be using `dehydrateStateUsing()` to overwrite the existing `dehydrateStateUsing()` defined on the base component class, for example.

edstevo:
Hey @danharrin,

That makes sense.

Would you accept another new method like `pipeDehydrateStateUsing()` so that we're not overwriting the existing method?

Thanks

danharrin:
I don't think so, sorry. At that point, you should probably just include all the pipes within one function.

edstevo:
Ok, thanks for taking the time to consider!

PR #17615, opened 2025-09-02
add toArray() to RichEditor

SOD96:
## Description

Brings the toJson functionality over (This existed in the awcodes tiptapeditor) and was pretty useful for converting things like Markdown HTML into what TipTap expected from a structure POV, to then iterate through. now as toArray as you can just json_encode the array if you still need that....

IE: 

Provides an array of the TipTap Structure of the content that you can go through
```
    $markdown = new Markdown('This is a **snippet** of _example_ Markdown.');

    $richContent = RichContentRenderer::make($markdown->toHtml())->toArray();
```

```
array:2 [▼
  "type" => "doc"
  "content" => array:1 [▶]
]
```

This is really only useful I think if you're doing edits before inserts and modifications outside of the typical filament panel. IE: Commands to bring in articles or format things differently. We use it all the time, so kinda need this.

## Functional changes

### DIFF
--- a/packages/forms/src/Components/RichEditor/RichContentRenderer.php
+++ b/packages/forms/src/Components/RichEditor/RichContentRenderer.php
@@ -360,14 +360,9 @@ class RichContentRenderer implements Htmlable
         }
 
         $editor = $this->getEditor();
-
-        if ($decoded) {
-            return json_decode($editor->getJSON(), true);
-        }
-
         $this->processMergeTags($editor);
 
-        return $editor->getJSON();
+        return $decoded ? json_decode($editor->getJSON(), true) : $editor->getJSON();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


packages/forms/src/Components/RichEditor/RichContentRenderer.php
```diff
@@ -350,6 +350,21 @@ public function toHtml(): string
         return Str::sanitizeHtml($this->toUnsafeHtml());
     }
 
+    /**
+     * @return string|array<string, mixed>
+     */
+    public function toJson(bool $decoded = false): string | array
```
danharrin: Please name it `toArray()` and remove the parameter & string return type. If you want to convert it to a string, you can do that yourself in the app.

danharrin (changes requested):

packages/forms/src/Components/RichEditor/RichContentRenderer.php
SOD96: @danharrin Appreciate the quick response, works for me! I've made the change :)

danharrin (approved):

PR #17622, opened 2025-09-02
Fix: drag prevention ineffective in `repeater` component during requests

People-Sea:
## Description

During the process of initiating requests for sorting, adding, and deleting, the disabled attribute has been added to the drag button, but it has not taken effect. If dragging and sorting are performed during this period, it may lead to incorrect data rendering.

As shown in the video, sorting during disabling caused the rendering of old items

# Before
https://github.com/user-attachments/assets/2646fd10-0893-4dd9-bd04-f3b45dbe8a22

# After

Code editor shows: Custom buttons have also been tested

https://github.com/user-attachments/assets/c72d2cc5-91f9-4254-abce-679dec137074

## Visual changes

Dragging is not allowed when the sort button is disabled

## Functional changes

### DIFF
--- a/packages/forms/resources/views/components/builder.blade.php
+++ b/packages/forms/resources/views/components/builder.blade.php
@@ -144,11 +144,8 @@
                                         class="fi-fo-builder-item-header-start-actions"
                                     >
                                         @if ($reorderActionIsVisible)
-                                            <li
-                                                x-sortable-handle
-                                                x-on:click.stop
-                                            >
-                                                {{ $reorderAction }}
+                                            <li x-on:click.stop>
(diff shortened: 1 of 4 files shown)


People-Sea:
This pr does not affect #17642

danharrin (changes requested):
Please use `extraAttributes(['x-sortable-handle' => true], merge: true)` so that existing extra attributes aren't overwritten

Does this need to be applied to the builder too?

People-Sea:
> Please use `extraAttributes(['x-sortable-handle' => true], merge: true)` so that existing extra attributes aren't overwritten
> 
> Does this need to be applied to the builder too?

Thank you for your reminder. I will use the merge parameter and synchronize the changes to the builder component

danharrin (approved):

PR #17623, opened 2025-09-02
fix: 03-tenancy.md getTenant example, v4 folder structures

geobehr:
Dan & Team, first PR here. I was working through multi-tenancy example getting error, needed getTenant to return Collection.

Also moved tenancy pages into v4-friendly 'app/Filament/Resources/Teams/Pages' folder structure

## Description
I was working through multi-tenancy example getting error, needed getTenant to return Collection.

Also moved tenancy pages into v4-friendly 'app/Filament/Resources/Teams/Pages' folder structure

## Functional changes

### DIFF
--- a/docs/07-users/03-tenancy.md
+++ b/docs/07-users/03-tenancy.md
@@ -71,6 +71,72 @@ public function panel(Panel $panel): Panel
 }
 ```
 
+### Setting up the 'Team' tenancy migrations and models
+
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks, however I feel like the new documentation you added here is quite opinionated and could cause confusion about what a user *needs* to do in order to use tenancy, compared to leaning on their existing models and migrations & app structure to do so. For example, they don't *need* to put the config in a resource.

PR #17626, opened 2025-09-02
Passes through missing arguments

scottzirkel:
## Description

The `mountAction` method accepts the method name, arguments, and context. The arguments are not passed into the method. This PR passes along the given arguments, should they exist.

## Visual changes

There are no visual changes.

## Functional changes

- [x] Documentation is up-to-date. (this bit is only documented from JavaScript)

### DIFF
--- a/packages/actions/src/Concerns/InteractsWithActions.php
+++ b/packages/actions/src/Concerns/InteractsWithActions.php
@@ -523,7 +523,9 @@ trait InteractsWithActions
                 return null;
             }
 
-            $resolvedAction = $this->{$methodName}();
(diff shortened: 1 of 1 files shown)


danharrin:
Hey, action arguments are not intended to be passed like this, and are instead injected as parameters into the different configuration functions available to the Action. For example

`Action::make('test')->action(fn ($arguments) => dd($arguments))`

Injecting them into the functions like this gives us a lot more control to change the arguments dynamically and keep a consistent experience when actions are used in different places, not just returned from Livewire methods.

scottzirkel:
Ah, I misunderstood that. Apologies, carry on.

PR #17642, opened 2025-09-03
Fix: update draggable index handling in sortable components

### LINKED ISSUE #17620: Repeater inside Builder Block can’t add or delete items
### Package

filament/filament

### Package Version

v4.0.6

### Laravel Version

v12.26.4

### Livewire Version

v3.6.4

### PHP Version

8.4.12

### Problem description

When using a **Builder** field and placing a **Repeater** inside a **Block**, adding or deleting repeater items does nothing. The UI briefly “flashes” as if the action is disallowed, but no items are added/removed. There are **no errors** in the browser console or Laravel logs.

This regression/bug prevents using repeaters within builder blocks.

**Things I Tried (Workarounds):**

- Moving the `Repeater` **outside** of the `Builder` → works normally.
- Using different field names / unique keys → no change.
- Toggling `->statePath()` custom paths on block/repeater → no change.
- Removing `->collapsed()` / `->defaultItems()` → no change.
- Disabling sortable features on Repeater → no change.

### Expected behavior

- Clicking **Add** should append a new repeater item within the block.
- Clicking **Delete** should remove the selected repeater item.
- State should update and persist normally.

### Steps to reproduce

1. Create a fresh Laravel app and install Filament v4 (Panel or just Forms).
2. Add a simple model (e.g., `Page`) with a JSON column `content`.
3. Create a Filament form where `content` is a Builder with a block that contains a Repeater.
4. In the form UI, try to **Add** an item to the Repeater → nothing happens (UI “flashes”).
5. Try to **Delete** an item (if you seeded one) → nothing happens (same UI flash).
6. Check browser console and Laravel logs → **no errors**.

https://github.com/deveti/filament-v4-builder-repeater-bug

### Relevant log output

```shell
No log or error output.
```

People-Sea:
Fix the error in the view that prevented sorting

Fixes #17620

## Description
The purpose of changing the dom structure of `builder.blade.php` is that if the add button in the middle of the item is to be kept, very complex js code is required to maintain its position .

Related issue: https://github.com/SortableJS/Sortable/issues/347

It's hard to maintain.  I don't think it should be adopted.  Besides, after my test, the solution in the following issue is not perfect 

If the dom structure is not updated, the problem shown in the video (dragging and not letting go, then putting it back in place) will occur in relatively marginal scenarios.

https://github.com/user-attachments/assets/3dc1472f-3b16-449d-999e-e32743aa14fd

Summary: If SortableJs is used, do not place any irrelevant elements in the sort box

After

https://github.com/user-attachments/assets/d33517c0-48fc-45e5-9aae-6c0c0f2f06d4

## Functional changes


danharrin:
> The purpose of changing the dom structure of builder.blade.php is that if the add button in the middle of the item is to be kept, very complex js code is required to maintain its position .

Can you explain how this is related to the issue please? We do need to preserve the button in the middle of blocks.

Is there just an easier way by reverting the broken PRs? Which PRs broke this?

People-Sea:
> > The purpose of changing the dom structure of builder.blade.php is that if the add button in the middle of the item is to be kept, very complex js code is required to maintain its position .
> 
> Can you explain how this is related to the issue please? We do need to preserve the button in the middle of blocks.
> 
> Is there just an easier way by reverting the broken PRs? Which PRs broke this?

There are two sub-issues in this s issue

1. The Builder \ Block nested within the Repeater actons could not use Repeater leading the pr: https://github.com/filamentphp/filament/pull/17590

2. When I was fixing the sorting, I didn't test the Builder\Block component. By default, Builder\Block enabled the buttons added between items, so it couldn't be used again after sorting once.

People-Sea:
Use what I call the use of complex javascript without changing the dom down complete ` between items add button ` can see here: https://jsbin.com/wagutowebi/edit?html,css,js,console,output 

Suppose the disabled item is the operation button between add

But there is a problem with it.  I tried to improve it but failed, which will lead to the program being not robust

People-Sea:
[PR](https://github.com/filamentphp/filament/pull/17604/files)  which improves smoothness but leads to the situation in the video where everything is re-rendered only when the position is truly changed

https://github.com/user-attachments/assets/5dce072a-f015-4b9d-a2cf-4d1654cc549f

packages/support/resources/js/sortable.js
people-sea, line 1: If there are no additional miscellaneous elements (add-betwein-Button) in the sort container, then newIndex == newDraggableIndex

People-Sea:
Hello @danharrin , I would like to disturb you again.  I just came up with a solution that can make the function work normally without changing any template file structure, and the script is relatively robust and easy to maintain , If possible, I will modify this pr in this way. 

The drawback is that the button in the middle seems misaligned during the drag-and-drop sorting process, but everything will be normal as soon as you release it

Simple demo： https://jsbin.com/jicopiluti/1/edit?html,js,console,output


---

This entire public relations process resolved all the `issue#17620` mentioned in the problem.  The main issue of being unable to operate nested components has been fixed.  For details, please refer to : 
https://github.com/filamentphp/filament/pull/17642/files#diff-9f5b69f902219ce73e2033d22fbc127969acf4aa608a11c527d4e24cfc4b11fb



The rest are all about fixing the front-end interaction not working

My words might be a bit jumbled.  Thank you for your patience in reading through ！

Odion-DS:
@danharrin. I would also like to briefly add my two cents. Since v4.0.5, actions that are in a path structure generally no longer work properly, meaning they are no longer executed at all. So it's not just the builder and repeater alone; his fix also fixes the other actions by adjusting the `packages/schemas/src/Concerns/HasComponents.php` file.

danharrin (approved):

danharrin:
The current solution looks good, thanks

PR #17662, opened 2025-09-04
[4.x] Feat: rich editor color extension

awcodes:
## Description

Adds support for applying color to selected text in the Rich Editor.

Defaults to the 500 shade of all available TW colors in the Color class, with support for overriding the select options on the RichEditor class with `->colors([])`.

## Visual changes

Screenshot 2025-09-04 at 1 49 09 PM
Screenshot 2025-09-04 at 1 49 02 PM

## Functional changes


CharlieEtienne:
Thanks for that @awcodes!

If I may add my two cents, what would be cool would be to have a circle with the current color instead of the swatch icon. 

For example:
image

Also, I would love something like your palette picker instead of the plain text select, but maybe that's more plugin territory...

Anyway, great addition (as always)

awcodes:
Yea, the idea was to keep it as simple as possible, ie, not having to maintain an actual color picker js implementation on top of it.

CharlieEtienne:
I see. Btw I wasn't suggesting a color picker, just seeing the colors themselves instead of text. Exactly like in your palette plugin, which should not involve any additional js, if I'm not wrong

awcodes:
Fair. Don't think Dan's willing to bring palette into core though. 😅

ahmed-fawzy99:
Amazing, waiting for this!

danharrin:
Replaced by #17888

PR #17672, opened 2025-09-05
fix: improve value formatting in `slider` component for better precision handling

### LINKED ISSUE #17559: Slider does not work with a Negative Minimum Value
### Package

filament/forms

### Package Version

v4

### Laravel Version

v12

### Livewire Version

_No response_

### PHP Version

PHP 8.4

### Problem description

When using the slider and you set the minimum range value to a negative value, the slider jumps around when the user releases the handle.

```php
Slider::make('slider')
  ->required()
  ->range(minValue: -25, maxValue: 200)
  ->default(50)
  ->step(1)
  ->tooltips(),
```

https://github.com/user-attachments/assets/d5e9e138-eeb7-4f63-a96a-2eb4776a4126

### Expected behavior

The slider to not jump around.

### Steps to reproduce

1. Create a filament project
2. Create a resource
3. Create a form with a slider
4. Set the slider range to have a minimum value of a negative number
5. Run Project
6. Drag Slider

https://github.com/mouse0270/filament-example-slider

### Relevant log output

```shell
Username: test@example.com
password: test@example.com
```

People-Sea:
## Description

When `format.from` is passed to `noUiSlider` for processing, it is guaranteed to be a Number.

Fixes #17559

## Visual changes

# Before

https://github.com/user-attachments/assets/d2f26c82-cd1d-4fd9-8315-20dcdce23a78

# After

https://github.com/user-attachments/assets/a30fe47f-84f8-48f5-892b-a73d5dc0d69d

## Functional changes


packages/forms/resources/js/components/slider.js
```diff
@@ -34,11 +34,8 @@ export default function sliderFormComponent({
                 direction: isRtl ? 'rtl' : 'ltr',
                 connect: fillTrack,
                 format: {
-                    from: (value) => value,
-                    to: (value) =>
-                        decimalPlaces !== null
-                            ? +value.toFixed(decimalPlaces)
-                            : value,
+                    from: (value) => +String(value),
```
danharrin: What is the benefit of `+String(value)` over `+value`?

packages/forms/resources/js/components/slider.js
```diff
@@ -34,11 +34,8 @@ export default function sliderFormComponent({
...
+                    to: (value) => +value.toFixed(+decimalPlaces),
```
danharrin: I'm not sure about stripping decimal places from the output like this if it is null, it feels like they should be preserved as they were before. If someone wants this behaviour they could do `decimalPlaces(0)` right?

danharrin (changes requested):

packages/forms/resources/js/components/slider.js
people-sea: That was my overthinking—+value is sufficient here.  Thank you for pointing it o

packages/forms/resources/js/components/slider.js
people-sea: I agree with your point—by default, we shouldn’t strip decimal places when decimalPlaces isn’t specified.  Thanks for bringing this up! 👍

danharrin (approved):

PR #17702, opened 2025-09-08
fix: prevent duplicate notifications

### LINKED ISSUE #17518: Duplicated notification in the Export Action
### Package

filament/filament

### Package Version

v4.0.3

### Laravel Version

v12.25.0

### Livewire Version

v3.6.4

### PHP Version

8.3.24

### Problem description

When using the Export Action provided by filament a notification is shown twice, most likely due to the `InteractsWithActions` and `CanExportRecords` traits both sending notifications (when not usinc a sync queue_driver).

CanExportRecords

```php
if (
    (filled($jobConnection) && ($jobConnection !== 'sync')) ||
    (blank($jobConnection) && (config('queue.default') !== 'sync'))
) {
    Notification::make()
        ->title($action->getSuccessNotificationTitle())
        ->body(trans_choice('filament-actions::export.notifications.started.body', $export->total_rows, [
            'count' => Number::format($export->total_rows),
        ]))
        ->success()
        ->send();
}
```

InteractsWithActions

```php
(match ($action->getStatus()) {
ActionStatus::Success => function () use ($action): void {
    $action->sendSuccessNotification();
    $action->dispatchSuccessRedirect();
},
ActionStatus::Failure => function () use ($action): void {
    $action->sendFailureNotification();
    $action->dispatchFailureRedirect();
},
})();
```

### Expected behavior

There should only be a single notifcation shown

### Steps to reproduce

- Clone repo
- Seed db with some users using the user factory
- Go to the customer resource
- Use the export action
- Notification is shown twice

After add the following to the action `->after(fn (Action $action) => $action->successNotificationTitle(null))` and run the export action again, now a single notification is shown because the action title is now null which prevents the `InteractsWithActions` check to pass

https://github.com/mikeslinkman/export-notifcation-reproduction

### Relevant log output

```shell

```

People-Sea:
## Description
Fixes #17518
Export notifications are uniformly notified using InteractsWithActions

## Visual changes

https://github.com/user-attachments/assets/f5d88cf2-8a9b-4a50-9cce-3aea3e78f435

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/CanExportRecords.php
+++ b/packages/actions/src/Concerns/CanExportRecords.php
@@ -292,8 +292,6 @@ trait CanExportRecords
             );
         });
 
-        $this->successNotificationTitle(__('filament-actions::export.notifications.started.title'));
-
         $this->defaultColor('gray');
 
         $this->modalWidth(static fn (ExportAction | ExportBulkAction $action): Width => match ($action->getColumnMappingColumns()) {
@@ -302,6 +300,8 @@ trait CanExportRecords
             3 => Width::FiveExtraLarge,
             default => Width::SevenExtraLarge,
         });
+
+        $this->successNotificationTitle(__('filament-actions::export.notifications.started.title'));
 
         if (! $this instanceof ExportBulkAction) {
             $this->model(fn (ExportAction $action): string => $action->getExporter()::getModel());
--- a/packages/actions/src/Concerns/CanExportRecords.php
+++ b/packages/actions/src/Concerns/CanExportRecords.php
@@ -156,12 +156,15 @@ trait CanExportRecords
             $maxRows = $action->getMaxRows() ?? $totalRows;
 
             if ($maxRows < $totalRows) {
-                $this->failureNotification(Notification::make()
-                    ->title(__('filament-actions::export.notifications.max_rows.title'))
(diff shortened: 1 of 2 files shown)


packages/actions/src/Concerns/CanExportRecords.php
```diff
@@ -269,17 +272,19 @@ protected function setUp(): void
                 )
                 ->dispatch();
 
+            $this->successNotificationTitle(__('filament-actions::export.notifications.started.title'));
```
people-sea: I think successNotification should also be used here.

Because in 
https://github.com/filamentphp/filament/blob/0e6383a0e170ad71e84220b05c78d68ff7241f5f/packages/actions/src/Concerns/InteractsWithActions.php#L269

https://github.com/filamentphp/filament/blob/0e6383a0e170ad71e84220b05c78d68ff7241f5f/packages/actions/src/Concerns/CanNotify.php#L119-L132

the `sendSuccessNotification()` method only uses the return value of `$this->getSuccessNotificationTitle()` if `$this->successNotification` has not been set


This will cause the `successNotification()` set in `->before` to "override" the default when the export mode is synchronous.

people-sea: This is on the assumption that the developer is not familiar with the source code

danharrin (approved):

PR #17709, opened 2025-09-09
fix interactive table column event bubbling and wire:navigate nested clicks

tanthammar:
## Description

Fixed click handling issues in interactive table columns and SPA mode navigation by using callbacks instead of direct navigation attributes.

### Interactive Table Columns

- **Fixed**: Clicks on action buttons and inputs in grid layout tables no longer bubble to parent `<a>` tags, preventing unintended navigation to edit/view pages
- **Affects**: Both SPA mode enabled and disabled

### SPA Mode Navigation

- **Fixed**: Nested element clicks with `stop` or `prevent` modifiers now work properly within `wire:navigate` links
- **Solution**: Using callbacks instead of `wire:navigate` or `x-navigate` attributes respects nested clicks with `wire:click` or `x-on:click` modifiers 
- **Not Fixed**: `wire:navigate.hover` issues remain due to Livewire/Alpine API limitations. (unable to do a .hover callback)

## Visual changes
none

## Functional changes

- [X] Non breaking change


packages/support/src/helpers.php
```diff
@@ -132,10 +132,11 @@ function generate_href_html(?string $url, bool $shouldOpenInNewTab = false, ?boo
         if ($shouldOpenInNewTab) {
             $html .= ' target="_blank"';
         } elseif ($shouldOpenInSpaMode ?? (FilamentView::hasSpaMode($url))) {
-            $html .= ' wire:navigate';
 
             if (FilamentView::hasSpaPrefetching()) {
-                $html .= '.hover';
+                $html .= ' wire:navigate.hover';
+            } else {
+                $html .= ' x-on:click.prevent="() => window.Alpine.navigate(' . "'{$url}'" . ')"';
```
danharrin: I don't think we should change this attribute for every single link, it seems too risky. Maybe just introduce a new `bool` param for `generate_href_html` for `$hasNestedClickEventHandler` which enables this?

danharrin (changes requested):

tanthammar:
Added tests for: 
  - Default behavior: When hasNestedClickEventHandler is false (default), SPA mode without prefetching now uses wire:navigate
  - When hasNestedClickEventHandler is true: SPA mode without prefetching uses Alpine navigation (x-on:click.prevent)
  - With prefetching enabled: Always uses wire:navigate.hover regardless of hasNestedClickEventHandler
  - With new tab or SPA disabled: The parameter has no effect

tanthammar:
Locations I updated to true:


`/tables/resources/views/index.blade.php`
`Line 1071`
`Line 1914:`

danharrin (approved):

danharrin (approved):

howdu:
@tanthammar just checking this PR doesn't break the event bubbling you fixed? 
https://github.com/filamentphp/filament/pull/18011

tanthammar:
@howdu Sorry for not replying to you. I collaborated w Dan via Discord on your PR. Then I forgot to reply here.

PR #17728, opened 2025-09-10
Fix:  Actions in builders no longer function since

### LINKED ISSUE #17725: Actions in builders no longer function since #17590
### Package

filament/filament

### Package Version

v4.0.9

### Laravel Version

v12

### Livewire Version

v3

### PHP Version

8.3

### Problem description

After the changes #17590 from actions placed within a builder block don't work.

The actions work in <=v4.0.4 but are broken in all versions above.

### Expected behavior

Clicking the action should trigger the php code to be run.

### Steps to reproduce

- Set up the example repo and go to create a new post record
    http://127.0.0.1:8000/admin/posts/create
- Add a heading block to the builder
- Click the action
- Nothing happens, the action's callback is not called

Image

https://github.com/oddvalue/filament-builder-actions

### Relevant log output

```shell

```

People-Sea:
## Description

fixes: #17725

added stricter checks when searching for actions within child schemas.   Now, the code only continues to search a child schema if the key matches exactly or starts with the child schema’s name, and skips the search if the key is blank or irrelevant.   This ensures more accurate and efficient action retrieval.

## Visual changes

https://github.com/user-attachments/assets/c90609b9-04fa-4161-87dc-16da8a646f1d

## Functional changes

### DIFF
--- a/packages/panels/src/Auth/MultiFactor/App/AppAuthentication.php
+++ b/packages/panels/src/Auth/MultiFactor/App/AppAuthentication.php
@@ -2,6 +2,8 @@
 
 namespace Filament\Auth\MultiFactor\App;
 
+use BaconQrCode\Renderer\ImageRenderer;
+use BaconQrCode\Writer;
 use Closure;
 use Filament\Actions\Action;
 use Filament\Actions\ActionGroup;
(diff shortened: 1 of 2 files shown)


ashleyshenton:
Just to add to this that we are having a similar issue where modals won't open when used inside repeaters. I tracked the cause of the issue down to the same PR referenced here. I've just pulled this branch down to test it locally, and this does fix our issue.

danharrin (approved):

PR #17729, opened 2025-09-10
[3.x] Add repeater item numbers

saade:
I was under the impression that repeaters had numbers just like Builders, but unfortunately that’s not the case.
Since I still have a long way to go before I can migrate my codebase to 4.x, I’m currently targeting 3.x.

Would you consider adding this feature even though 3.x is feature-complete? If needed, I’d be happy to submit a PR for 4.x as well.

Thanks!


danharrin (changes requested):
Once there is a v4 PR in place I will review and merge this

saade:
done https://github.com/filamentphp/filament/pull/17812

danharrin (approved):

PR #17731, opened 2025-09-10
feat(panels): add flag to toggle page head padding

danielhe4rt:
Recently I've decided to use more often the "Guest Panel" to build landing pages and every time that I start a new project, I need to copy the vendor `filament-panels::page.index` and add a new flag to remove the padding only.

See the white gap between the edges:

<details>

image

image

</details>

## Proposal

By adding a new component property `headerPadding`, we can easily solve it when creating custom pages that require the navbar.

```php
@props([
    'fullHeight' => false,
    'headerPadding' => true, // added item
])

@php
    use Filament\Pages\Enums\SubNavigationPosition;

    $subNavigation = $this->getCachedSubNavigation();
    $subNavigationPosition = $this->getSubNavigationPosition();
    $widgetData = $this->getWidgetData();
@endphp
// ... rest of the component

<div
        {{
            $attributes->class([
                'fi-page-header-main-ctn',
                'p-0' => ! $headerPadding,
            ])
        }}
    >
```

## Functional changes

Closes #17730

### DIFF
--- a/packages/panels/resources/views/components/page/index.blade.php
+++ b/packages/panels/resources/views/components/page/index.blade.php
@@ -1,5 +1,6 @@
 @props([
     'fullHeight' => false,
+    'headerPadding' => true,
 ])
 
 @php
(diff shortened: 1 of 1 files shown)


danharrin:
Sorry, there are a couple of reasons why I would not merge this. To start with, it doesn't work, since Tailwind classes are no longer present in the views in Filament v4, so the `p-0` is not compiled and does not apply unless you are explicitly using that class in a custom theme. And I would suggest that you instead just add this to your custom theme CSS file to do the same:

```css
.fi-page-header-main-ctn {
    @apply p-0;
}
```

Is the issue here that the `fi-page-header-main-ctn` element is output even when it has no content? In which case, I would accept a PR to remove it if it does not contain anything.

PR #17733, opened 2025-09-11
Allow relative links in RichEditor LinkAction

alwayshopeless:
Allow relative links in URL

## Visual changes

Reopen problem in Issue #12514 
Allow creating links with relative paths. For example, if the site is on several different domains, etc.

### DIFF
--- a/packages/forms/src/Components/RichEditor/Actions/LinkAction.php
+++ b/packages/forms/src/Components/RichEditor/Actions/LinkAction.php
@@ -23,8 +23,7 @@ class LinkAction
             ])
             ->schema([
                 TextInput::make('url')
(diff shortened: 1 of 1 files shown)


packages/forms/src/Components/RichEditor/Actions/LinkAction.php
danharrin, line 1: Please add `inputMode('url')` like the previous PR did

danharrin (changes requested):

packages/forms/src/Components/RichEditor/Actions/LinkAction.php
alwayshopeless: Oh, sorry. Fixed

danharrin (approved):

Issue #13335, opened 2024-06-20
Tooltip alignment


zepfietje:
Tooltip should align relative to actual table cell or text entry content.
This has been brought up before, so let's look into ways to fix this.

<img width="366" alt="Screenshot 2024-06-20 at 11 26 11" src="https://github.com/filamentphp/filament/assets/44533235/8aedcfd5-a0e8-4793-95bd-1df6dbe2618f">

zepfietje:
Do you think this should be v4 instead, @danharrin?

danharrin:
I don't think so?

zepfietje:
Cool, seems like a bug fix to me too.

colorants:
Hi, the problem seems to exist still. Are there any hacky ways to fix it or should I just ignore it for now.

ShaungBhone:
I've already fixed that, but we need to move the tooltip to `text-column.blade.php` instead of the entire `column.blade.php`. I'm not sure if you guys will accept this change or not.

![Image](https://github.com/user-attachments/assets/8f35b175-14af-45e1-8638-9cc3f408dd82)

jsheffers:
Would something like this work? Attaching the x-on:click to a child element?

Concept would be to split the attributes to container & inner attributes. 

<img width="2810" height="1350" alt="Image" src="https://github.com/user-attachments/assets/dc3e58fa-6d64-4460-aee3-d2323eeaab5d" />
<img width="922" height="86" alt="Image" src="https://github.com/user-attachments/assets/3f5853ed-1717-4144-9744-dbe3828f4bdc" />

<img width="1360" height="596" alt="Image" src="https://github.com/user-attachments/assets/026d11f6-12ca-4237-9773-621425b115a6" />


Styling updates
<img width="494" height="258" alt="Image" src="https://github.com/user-attachments/assets/05149560-ab1a-42a5-bb04-2dc479c797c2" />

jsheffers:
@danharrin any thoughts on this approach?

danharrin:
Does it work when there are multiple items (array state) in the column?

jsheffers:
Submitted a pull request 👆🏻

jsheffers:
> Does it work when there are multiple items (array state) in the column?

Yes, it does. I've attached a screen recording to the pull request.

PR #17747, opened 2025-09-11
Fix tooltip alignment by moving x-tooltip to a span.

jsheffers:
## Description

Tooltips for `TextEntry` and `TextColumn` don't align to the actual text content of the entry/column content. This new update adds the content as a span and attaches the x-tooltip to the span therefore allowing the tooltip to inherit it's position from the actual content and not the full width column.

Ref: https://github.com/filamentphp/filament/issues/13335

## Visual changes

**Old behavior**
https://github.com/user-attachments/assets/0b1e07e3-271e-4e95-bea4-80ec14b92ea5

**New behavior**
https://github.com/user-attachments/assets/e0015fce-a377-4914-b3c2-04b1bb91edb6

## Functional changes

### DIFF
--- a/packages/infolists/src/Components/TextEntry.php
+++ b/packages/infolists/src/Components/TextEntry.php
@@ -196,7 +196,7 @@ class TextEntry extends Entry implements HasAffixActions, HasEmbeddedView
                 <?php } ?>
             </div>
 
-        <?php return $this->wrapEmbeddedHtml(ob_get_clean());
+            <?php return $this->wrapEmbeddedHtml(ob_get_clean());
         }
 
         $shouldOpenUrlInNewTab = $this->shouldOpenUrlInNewTab();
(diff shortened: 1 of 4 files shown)


packages/infolists/src/Components/TextEntry.php
```diff
@@ -413,72 +439,70 @@ public function toEmbeddedHtml(): string
...
+                    </ul>
+
+                    <?php if ($stateOverListLimitCount) { ?>
+                        <div class="fi-in-text-list-limited-message">
+                            <?php if ($isLimitedListExpandable) { ?>
+                                <div
+                                    role="button"
+                                    x-on:click.prevent.stop="isLimited = false"
+                                    x-show="isLimited"
+                                    class="fi-link fi-size-xs">
```
danharrin: Please revert all formatting changes like this one, to make the Git diff as small as possible

danharrin (changes requested):

jsheffers:
I think I've addressed all formatting issues with my latest commits @danharrin

danharrin:
Thanks, but unfortunately I ran into quite a few bugs in this implementation, and felt like it could be fixed with fewer changes, so I decided to start over: #17916

jsheffers:
No worries! Thanks @danharrin

PR #17753, opened 2025-09-12
Render html in column manager

### LINKED ISSUE #17701: Columns with icon labels show the full HTML in ColumnManager
### Package

filament/filament

### Package Version

v4.0.8

### Laravel Version

v12.28.1

### Livewire Version

v3.6.4

### PHP Version

8.4.12

### Problem description

If you have a column with a label including HTML, like when adding a HeroIcon, if the ColumnManager is shown, the full HTML is displayed as the column label.

### Expected behavior

It should render the HTML the same way (which could be problematic in some cases, like in the example repository, that has a flex container for centering) or have a fallback value that displays instead.

### Steps to reproduce

Access the example repository with:

- admin@example.com
- admin

Go to the user resource (ignore the size of the icon, since the example doesn't have a custom theme with the necessary tailwind classes).

Expand the ColumnManager from the header.

https://github.com/RixzZ/filament-issue-01

### Relevant log output

```shell

```

archilex:
Fixes #17701 

When discussing PR #17443 we talked about adding a fallback `->columnManagerLabel()` but we decided against it so this PR just renders the html in the column manager label.

### DIFF
--- a/packages/tables/resources/views/components/column-manager.blade.php
+++ b/packages/tables/resources/views/components/column-manager.blade.php
@@ -81,7 +81,7 @@
                                         />
                                     @endif
 
-                                    <span x-text="column.label"></span>
+                                    <span x-html="column.label"></span>
                                 </label>
 
                                 @if ($hasReorderableColumns)
@@ -131,7 +131,7 @@
                                                 @endif
 
                                                 <span
-                                                    x-text="groupColumn.label"
+                                                    x-html="groupColumn.label"
                                                 ></span>
                                             </label>
 
@@ -165,7 +165,7 @@
                                     />
                                 @endif
 
-                                <span x-text="column.label"></span>
+                                <span x-html="column.label"></span>
                             </label>
 
                             @if ($hasReorderableColumns)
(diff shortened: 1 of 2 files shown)


danharrin (changes requested):
I think there might be some XSS risk there if a user uses HTML within a label. Currently, if they want to use HTML, they must use a `new HtmlString()` as the label I think. So maybe before putting the label in the `label` key of the column config array, wrap it in `e()` so that non-HtmlString values have escaped HTML entities?

archilex:
I had considered that but the column label isn't configurable/exposed to the user so didn't seem like an issue? And the one possible vector might be through tampering with the session since we store the column array there, including label, but even that's protected since on each request the $column->getLabel() is re-evaluated. 

But I'm not a security expert and in my testing using `e()` didn't hurt either so I went ahead and added it as another layer of defense.

danharrin:
If user input is being used to add a dynamic label to a column, like when there are custom fields configured by the user in the database, it could be a problem

danharrin (approved):

PR #17764, opened 2025-09-12
fix: Navigation group default alphabetical sorting

danharrin:
Fixes #17557

PR #17766, opened 2025-09-12
Revert "fix: Navigation group default alphabetical sorting"

danharrin:
Reverts filamentphp/filament#17764


stefket:
@danharrin What do you think about the idea of sorting navigation groups based on their enum values?

Since it’s already possible to assign an enum as a navigation group, we could extend this by introducing a getSort method in the enum. This method could then be used as the basis for determining the order of the groups.

danharrin:
They are already sorted by the enum cases in the order they are defined, I dont think we need an extra method.

stefket:
@danharrin Oh, I didn’t know that. That’s even better—thank you for your reply.

PR #17779, opened 2025-09-13
fix docs for the custom column table

CharrafiMed:
## Description

just converting `$record` to `$getRecord()`, since the `$record` is protected, and can be accessed using `$getRecord()`

## Visual changes
none

## Functional changes
fix a typo


danharrin:
I actually think we should fix the root of the problem and fix the code to match the docs here. Making it public will not help, we need to pass it into the view when the view is rendered. The same needs to happen for infolist entries.

danharrin:
Replaced by #17788

PR #17785, opened 2025-09-14
Use "state snapshot" for repeaters and builders

danharrin:
Currently, repeater and builder data is not cast before their state is used to determine item labels. This PR casts the state first to make it easier to use.

### DIFF
--- a/packages/forms/src/Components/Builder.php
+++ b/packages/forms/src/Components/Builder.php
@@ -1135,7 +1135,7 @@ class Builder extends Field implements CanConcealComponents, HasExtraItemActions
      */
     public function getRawItemState(string $key): array
     {
-        return $this->getChildSchema($key)->getRawState();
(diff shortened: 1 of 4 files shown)


tvercruysse:
@danharrin I just upgraded my app from v4.0.3 to v4.2.2 and noticed this MR broke the form i have.

Previously this worked:
`$this->setItemLabel(fn ($state) => new HtmlString('amount of items: ' . count($state['scheduled_items']));`

But since this MR `scheduled_items` first exists in the state, but then when dehydrating it calls `Arr::forget` on that key.

```
Schedule:
 - name
 - time
 - scheduled_items (repeater)

schedule_items
- item_name
- item_quantity
- item_alternatives (repeater)
```

danharrin:
Please open an issue with a simple reproduction repository

Issue #15347, opened 2025-01-15
Page sticky footer actions only sticky at certain window sizes/distance from action buttons


Mrkbingham:
### Package

filament/filament

### Package Version

v3.2.133

### Laravel Version

v10.48.25

### Livewire Version

v3.5.12

### PHP Version

PHP 8.3.15

### Problem description

When setting the `$formActionsAreSticky` property on pages, the footer is not positioned correctly due to the JS calculating whether or not the actions should be sticky/stuck.

I'm uncertain as to why this reversion was committed, however, using the old code here resolves my issue:
https://github.com/filamentphp/filament/commit/e99ed7188c836414ae95b251c4839b3212f9f43b

**I believe this is the current problem:**

The calculation of the action button's stickiness is based on the vertical scroll position `window.scrollY` and the height of the overall window.  However, the commit noted above takes into account the position of the element itself, determining stickiness based on the element's distance from it's original position.  

If there is a thread that I was unable to find related to the reversion of the commit linked above, I'd be happy to help open a new PR to address the issue.

Also, FWIW, the code is currently the same between v3 and v4 branches.

Mobile and desktop examples

https://github.com/user-attachments/assets/1f958afa-330d-49b6-a7d3-5cbab2c53b20

https://github.com/user-attachments/assets/339270d3-df37-4005-ab22-9b226e63dcc0

### Expected behavior

That the form action buttons would be sticky at the bottom of the screen until the page was scrolled to the position where the action buttons exist in a non-sticky position within the HTML of the page.

Mobile and desktop examples (Expected behavior):

Using this version of the `actions.blade.php` file: https://github.com/filamentphp/filament/blob/56450d423089609b3fb0125f45c504622451403c/packages/panels/resources/views/components/form/actions.blade.php

https://github.com/user-attachments/assets/78e0c4d0-8445-4e90-8b43-676c5d425f5c

https://github.com/user-attachments/assets/b5665989-5024-422e-b629-ea44760fd8f2

### Steps to reproduce

1. Add `\Filament\Pages\Page::stickyFormActions(true);` to your EventServiceProvider's boot method
2. Create a form that is tall enough to push the default save/cancel action buttons off the form
3. Load the page, and you will see the action buttons are sticky on a small screen, but as you scroll, once you get near to the action buttons, they are no longer sticky.

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/Mrkbingham/filament-sticky-footer-issue

tuimz:
[...]

PR #17794, opened 2025-09-15
bugfix(action-sticky): fix form action sticky issue

ki5an:
PR fix for: Page sticky footer actions only sticky at certain window sizes/distance from action buttons
https://github.com/filamentphp/filament/issues/15347


danharrin (changes requested):
Please prepare a fix inside `filamentActionsSchemaComponent`, instead of introducing new JS into the HTML.

ki5an (commented):
I moved the sticky logic into filamentActionsSchemaComponent. enableSticky and disableSticky are now triggered by the respective x-intersect events instead of using inline JS.

ki5an:
@danharrin, just checking if there’s anything else I should do. 
I prefer using an Alpine intercept so values stay consistent, triggering only when elements enter or leave the viewport. In one case, for example, if a table is below the form, the second condition ensures the sticky applies only below the form, not the table.

danharrin (approved):

PR #17825, opened 2025-09-16
fix: unset contain property for modals with file upload editor

### LINKED ISSUE #17354: [4.x] Image Editor not usable when opened from a modal
### Package

filament/filament

### Package Version

v4.0.1

### Laravel Version

v12.24.0

### Livewire Version

v3.6.4

### PHP Version

PHP 8.4

### Problem description

If you have an edit form that opens in a modal or slideOver and the form has a FileUpload element with the imageEditor. Opening the image editor either opens it in the slideOver or in the case of a modal opens it without it being centered on the screen.

### Expected behavior

This is making the image editor difficult to use because you have to mess with scroll bars to just be able to see anything useful and then you have to slide it back to see the other part. If you setup my reproduction and try to edit the image on a book, you'll see what I'm talking about. In v3 this worked fine and opened the image editor in a new modal that was centered on your screen.

### Steps to reproduce

1. Create a resource that doesn't have an edit page so it will use a modal for the edit
2. Add a FileUpload element to your form that includes the image() and imageEditor() methods
3. Setup your resource to use an infolist for the view and change the edit action to use a slide over
4. Make sure you have the edit action in your table rows so you can see the modal
5. Visit the table and click on the edit link for one of the rows. Then click on the edit pencil to open the image editor and notice that it opens up strange.
6. Visit the infolist view and click on the edit to open the slide over
7. Click on the edit pencil to edit your image and notice that it opens in the slide over.

https://github.com/dcaswel/filament-img-issue

### Relevant log output

```shell

```

People-Sea:
## Description
Do not use 'contain: unset;' when the image editor is included in the open modal. `

My suggestion is to simply remove 'contain: layout; `

Fix the issue where pressing Esc closes the modal window along with it and the error triggered by the blur event after the element disappears.

fixes: #17354

## Visual changes
image

## Functional changes


danharrin:
I think this breaks searchable Select field dropdowns inside modals, if the property is not there. Can you check please? We might need to change how we position the Select field dropdown if this property needs to be removed.

danharrin:
https://github.com/filamentphp/filament/issues/17024

People-Sea:
> 我认为，如果该属性不存在，模态框内可搜索的选择字段下拉菜单就会失效。您能帮我检查一下吗？如果需要移除此属性，我们可能需要更改选择字段下拉菜单的定位方式。

Gladly！

People-Sea:
I found the root cause: the offset returned by computePosition was incorrect because it did not account for the distance from the window to the modal.  (Currently, the modal is centered using layout instead of positioning.  After I switched to positioning for centering, the offset returned by computePosition became correct.  This should be the reason.)

People-Sea:
packages/support/resources/js/utilities/select.js

Just pseudo-code

image

People-Sea:
@danharrin plz review

People-Sea:
image
image
image
image

danharrin (approved):

PR #17829, opened 2025-09-16
Make TextInput type range, visible

tanthammar:
## Description

Fixes range input visibility

Example:

```php
TextInput::make('satisfaction')
                ->type('range')
                ->hint(fn ($state) => filled($state) ? __('form.range.selected', ['state' => $state]) : null)
                ->default(1)
                ->minValue(0)
                ->maxValue(10)
                ->prefix(1)
                ->suffix(10)
                ->step(2)
                ->datalist([0,2,4,6,8,10])
                ->live(),
```

## Visual changes

before
Skärmavbild 2025-09-16 kl  12 53 27

after
Skärmavbild 2025-09-16 kl  12 53 11

## Functional changes


packages/support/resources/css/components/input/index.css
```diff
@@ -36,3 +36,9 @@ input.fi-input {
         @apply text-justify;
     }
 }
+
+input[type='range'].fi-input {
+    margin-inline: auto;
+    width: 98%;
```
danharrin: Instead of a percentage width which will change with screen size, maybe add a fixed `margin-inline` of `1rem` instead?

Maybe `@apply mx-4 appearance-auto;`? How does that look?

danharrin (changes requested):

packages/support/resources/css/components/input/index.css
tanthammar: px or mx does not work. nor does margin-inline of rem or other values.

This is what it looks like on mobile screen size:
image

This is on 6xl screen size:
image


98% gives just a small space between the form border edges.

danharrin (approved):

PR #17830, opened 2025-09-16
fix: handle 'as' alias in orderBy correctly in Select component

f-liva:
## Description

Adjust the orderBy logic in the Select component's relationship method to correctly handle attributes that use aliases ('as' syntax). This ensures the query orders by the actual column name instead of the alias, preventing certain database errors. This change resolves issues when dealing with complex queries where attributes are aliased.

## Visual changes

No visual changes.

## Functional changes


danharrin (approved):

f-liva:
Thanks for making the change to the other components, they had slipped past me.

PR #17841, opened 2025-09-17
feat: adding extra content to `TextInputColumn` 

Baspa:
## Description

For a client of mine I needed to add extra content to the `TextInputColumn` similar to how it's done for the `TextInput`. I thought it might be nice to add this in Filament core as well. But it might need some changes. For example:

- I think the extra content is shown too close to the actual input. It also causes the table to "malformed" because the column with the extra content is higher than the other columns.
- I'm not sure if using the trait like this would be correct, it would allow us to implement similar functionality to the `SelectInput` however
- Docs might to be updated as well 

Would love some feedback on this. 

## Visual changes

Notice the mentioned spacing issues, this needs to be fixed first: 

Screenshot 2025-09-17 at 09 18 35

## Functional changes


danharrin:
I think I'm going to pass on this one for now, I think there aren't too many use cases for this and the alignment issues when one input has extra content and another does are quite problematic to be honest. If I get more feature requests I might reconsider in the future.

PR #17810, opened 2025-09-15
fix: State casting issues

danharrin:
Fixes #17809.

PR #17846, opened 2025-09-17
Adds IntegerStateCast class for state casting

Mrkbingham:
## Description

The state-casting change in https://github.com/filamentphp/filament/pull/17810 provides a more reliable way to ensure data is cast properly.  However, when dealing with relationship-data (e.g. storing IDs in the DB), I'd like to ensure as much type-consistency as possible.

With the change mentioned above, the relationship on a `Select` form field is being cast as a string, before then getting re-converted back to an integer.  My proposal here is to allow for an optional cast to an integer, which could be applicable to any field that uses cast, but in particular here if the relationship you are saving uses integer IDs in your table.  

## Visual changes

n/a

## Functional changes

### DIFF
--- a/packages/schemas/src/Components/StateCasts/IntegerStateCast
+++ b/packages/schemas/src/Components/StateCasts/IntegerStateCast
@@ -1,0 +1,30 @@
+<?php
+
+namespace Filament\Schemas\Components\StateCasts;
+
+use Filament\Schemas\Components\StateCasts\Contracts\StateCast;
+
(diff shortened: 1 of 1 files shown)


danharrin:
I think I would prefer a solution like #17849, what are your thoughts there?

Mrkbingham:
@danharrin looks good to me - [left one comment on that PR though](https://github.com/filamentphp/filament/pull/17849/files#r2360388096), unrelated to my original PR here, but thought it might be useful

PR #17849, opened 2025-09-17
fix: Cast select state to int

danharrin:


danharrin:
@gehrisandro what do you think of this change? If the state values of selects are integers, then cast them back to an integer. The internal state still needs to be a string, but people are probably expecting an integer back rather than an integer-string.

gehrisandro:
> @gehrisandro what do you think of this change? If the state values of selects are integers, then cast them back to an integer. The internal state still needs to be a string, but people are probably expecting an integer back rather than an integer-string.

I think this makes sense. Thank you!

tests/src/Schemas/Components/StateCasts/EnumArrayStateCastTest.php
```diff
@@ -61,7 +61,7 @@
     $cast = app(EnumArrayStateCast::class, ['enum' => IntegerBackedEnum::class]);
 
     expect($cast->set([IntegerBackedEnum::One, IntegerBackedEnum::Two, IntegerBackedEnum::Three]))
-        ->toBe([1, 2, 3]);
+        ->toBe(['1', '2', '3']);
```
Mrkbingham, line 64: Should this be casting an integer backed enum to strings?  For consistency I'd expect this to to have not changed in this PR since the base value is an integer - but is this due to ingesting the value as a string being a requirement for Enums in Filament somewhere else?

danharrin: Yes, this is just the internal state of the enum within Filament, the user never interacts with the string version. In the same way, integer options are also cast to strings automatically when enums are not used.

Mrkbingham: Right on!  Yeah looks great then

Miguel-Serejo:
This broke my use case where a select is used to choose from a list of client codes which have leading zeroes. 
Is there an easy way to revert the behavior back to using "string-integers"?

peterjbassi:
This change also broke form submissions for options where the value is a number greater than `PHP_INT_MAX`.  For example, If you try to use a value like `"922337203685477580800001"`, it will get converted to an integer and reduced to `PHP_INT_MAX`, which for 64-bit platforms is `9223372036854775807`, and fail validation.

This PR is problematic for other reasons though. My application can't be the only one built with the expectation of behavior that existed through `4.0.17` that integer option values would be returned as strings. While I agree with the spirit of this PR, I believe it represents a breaking change.

BenjaminRayner:
This also broke my use case where I have a select relationship from a list of alphanumeric ids. If an ID happens to have all number digits, it gets incorrectly casted to int and then the query fails when it tries to retrieve the relationship from the database.

Youssef-Amjad:
> @gehrisandro what do you think of this change? If the state values of selects are integers, then cast them back to an integer. The internal state still needs to be a string, but people are probably expecting an integer back rather than an integer-string.

Error fetching option label: Component not found: undefined
getLabelForSingleSelection @ select.js?v=5.0.0.0:1

PR #17862, opened 2025-09-19
Force render to bypass partial rendering

howdu:
This pull request introduces a mechanism to force a re-render of a  component and disable partial rendering. Useful in a RelationManager where a user clicks "create another" and expects to see the results behind the modal. 

## Usage
Add to all create actions used in a relation manager by adding this into a service provider which calls `forceRender` after the 'Create another' action has been called.
```php
CreateAction::configureUsing(function (CreateAction $action): void {
    $action->forceRenderAfterCreateAnother();
});
```

## Visual changes
https://github.com/filamentphp/filament/discussions/17545#discussioncomment-14244155

## Functional changes


howdu:
@danharrin if you approve? I could add `shouldForceRenderAfterCreateAnother()` into `CreateAction`,  so this can be added globally without losing the ability to use `after` elsewhere.

danharrin (changes requested):
I think you should add an early return to the `dehydrate()` method on the `PartialsComponentHook` if rendering is forced, to prevent the partials from rendering as well as the main HTML

danharrin:
> @danharrin if you approve? I could add `shouldForceRenderAfterCreateAnother()` into `CreateAction`, so this can be added globally without losing the ability to use `after` elsewhere.

Sure, that sounds fine to me

danharrin (approved):

PR #17871, opened 2025-09-19
Update 03-testing-tables.md - "Testing that a column can be sorted"

### LINKED ISSUE #17817: `sortBy()` does not work for MySQL database in sorting tests
### Package

Other

### Package Version

v4.0.15

### Laravel Version

v12.28.1

### Livewire Version

_No response_

### PHP Version

PHP 8.4.12

### Problem description

[I write tests to check if sorting works correctly](https://filamentphp.com/docs/4.x/testing/testing-tables#testing-that-a-column-can-be-sorted). After running my tests locally, with Postgresql set as my database, none of my tests failed. Then I pushed my changes to the repository and I found out that some tests are failing (which were the ones that tested the sorting functionality).

I checked which database the CI uses and it is MySQL, so I decided to change my `phpunit.xml` and set the database to MySQL. Yup, there it is, the same tests that failed in the pipeline also fail locally now.

The code for the test is:
```php
it('can sort users by `name` on index page', function () {
    // authenticate

    $users = User::factory()->count(3)->create();

    livewire(ListUsers::class)
        ->assertCanSeeTableRecords($users)
        ->sortTable('name')
        ->assertCanSeeTableRecords($users->sortBy('name'), inOrder: true) // it fails on this line
        ->sortTable('name', 'desc')
        ->assertCanSeeTableRecords($users->sortByDesc('name'), inOrder: true);
});
```

When I use the `orderBy` on the `QueryBuilder` instead the test never fails:
```php
it('can sort users by `name` on index page', function () {
    // authenticate

    User::factory()->count(3)->create();

    $orderedUsers = User::query()->orderBy('name')->get();
    $descOrderedUsers = User::query()->orderBy('name', 'desc')->get();

    livewire(ListUsers::class)
        ->assertCanSeeTableRecords($orderedUsers)
        ->sortTable('name')
        ->assertCanSeeTableRecords($orderedUsers, inOrder: true)
        ->sortTable('name', 'desc')
        ->assertCanSeeTableRecords($descOrderedUsers, inOrder: true);
});
```

Here is a test case with specific names set for the users. The test **always** fails with MySQL and **never** fails with Postgresql:

```php
it('can sort users by `name` on index page', function () {
    // authenticate

    $users = User::factory()->createMany([
        ['name' => 'kand. Emma Groen B'],
        ['name' => 'Lily Zum Vörde Sive Vörding'],
        ['name' => 'Lily Güler AD'],
    ]);

    livewire(ListUsers::class)
        ->assertCanSeeTableRecords($users)
        ->sortTable('name')
        ->assertCanSeeTableRecords($users->sortBy('name'), inOrder: true)
        ->sortTable('name', 'desc')
        ->assertCanSeeTableRecords($users->sortByDesc('name'), inOrder: true);
});
```

**When testing this in the browser we can see that the sorting strategy is different for both the databases.** 

### Expected behavior

So `->sortBy()` is a method for `Collections` the `->create()` method on the factory returns a `Collection` of `Models`. So this is not a bug on the `->sortBy()` method. 

This either should be documented differently (for now) or the sorting strategy in production should be changed and it should always use the `sortBy()` method of the `Collection` as the sorting strategy and not the database sorting strategy.

### Steps to reproduce

1. Create new laravel project
2. install filament
3. create `customer` resource (with model, factory, migration)
4. add test like above
5. run the test in mysql + pgsql

https://github.com/Merdin/filament-sorting-bug-in-test

### Relevant log output

```shell

```

Merdin:
## Description

See issue #17817

Currently, the documentation includes a flaky test as an example for sorting a column. This is because Filament sorts its table records according to the underlying database strategy.

Previously, the documentation suggested using the `sortBy()` method on the collection returned by the factory. However, this approach does not account for differences in how various databases handle sorting, which can lead to inconsistent test results.

With this commit, the documentation has been updated. The example test now uses a stable approach that explicitly orders records through the database query, ensuring reliable behavior across different database systems.

## Visual changes
N/A

## Functional changes

### DIFF
--- a/docs/10-testing/03-testing-tables.md
+++ b/docs/10-testing/03-testing-tables.md
@@ -118,7 +118,7 @@ it('can sort posts by title', function () {
     Post::factory()->count(10)->create();
 
     $sortedPostsAsc = Post::orderBy('title')->get();
-    $sortedPostsDesc = Post::orderBy('title', 'desc')->get();
+    $sortedPostsDesc = Post::orderByDesc('title')->get();
 
     livewire(PostResource\Pages\ListPosts::class)
         ->sortTable('title')
--- a/docs/10-testing/03-testing-tables.md
+++ b/docs/10-testing/03-testing-tables.md
@@ -115,15 +115,24 @@ Once the table is sorted, you can ensure that the table records are rendered in 
 use function Pest\Livewire\livewire;
 
 it('can sort posts by title', function () {
-    $posts = Post::factory()->count(10)->create();
+    Post::factory()->count(10)->create();
+
+    $sortedPostsAsc = Post::query()->orderBy('title')->get();
+    $sortedPostsDesc = Post::query()->orderBy('title', 'desc')->get();
 
     livewire(PostResource\Pages\ListPosts::class)
         ->sortTable('title')
-        ->assertCanSeeTableRecords($posts->sortBy('title'), inOrder: true)
+        ->assertCanSeeTableRecords($sortedPostsAsc, inOrder: true)
         ->sortTable('title', 'desc')
-        ->assertCanSeeTableRecords($posts->sortByDesc('title'), inOrder: true);
+        ->assertCanSeeTableRecords($sortedPostsDesc, inOrder: true);
 });
 ```
+
+<Aside variant="info">
+    Records are **sorted** according to the database's sorting strategy (MySQL and PostgreSQL use different sorting strategies).  
+    Therefore, you cannot use the `sortBy()` collection method on the factory results.  
+    Instead, make sure to order them explicitly in your query using `orderBy()`.
+</Aside>
 
 ### Testing the state of a column
 
--- a/docs/10-testing/03-testing-tables.md
+++ b/docs/10-testing/03-testing-tables.md
@@ -118,7 +118,7 @@ it('can sort posts by title', function () {
     Post::factory()->count(10)->create();
 
     $sortedPostsAsc = Post::query()->orderBy('title')->get();
-    $sortedPostsDesc = Post::query()->orderByDesc('title')->get();
+    $sortedPostsDesc = Post::query()->orderBy('title', 'desc')->get();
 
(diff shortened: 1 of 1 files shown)


docs/10-testing/03-testing-tables.md
```diff
@@ -115,16 +115,25 @@ Once the table is sorted, you can ensure that the table records are rendered in
...
-        ->assertCanSeeTableRecords($posts->sortByDesc('title'), inOrder: true);
+        ->assertCanSeeTableRecords($sortedPostsDesc, inOrder: true);
 });
 ```
 
+<Aside variant="info">
+    Records are **sorted** according to the database's sorting strategy (e.g. MySQL and PostgreSQL use different sorting strategies).  
+    Therefore, you cannot use the `sortBy()` collection method on the factory results.  
+    Instead, make sure to order them explicitly in your query using `orderBy()`.
+</Aside>
```
Merdin, line 135: Not sure if you want to mention something like this in the "main" documentation where sorting is explained.

docs/10-testing/03-testing-tables.md
```diff
@@ -115,16 +115,25 @@ Once the table is sorted, you can ensure that the table records are rendered in
...
-        ->assertCanSeeTableRecords($posts->sortBy('title'), inOrder: true)
+        ->assertCanSeeTableRecords($sortedPostsAsc, inOrder: true)
         ->sortTable('title', 'desc')
-        ->assertCanSeeTableRecords($posts->sortByDesc('title'), inOrder: true);
+        ->assertCanSeeTableRecords($sortedPostsDesc, inOrder: true);
 });
 ```
 
+<Aside variant="info">
+    Records are **sorted** according to the database's sorting strategy (e.g. MySQL and PostgreSQL use different sorting strategies).
```
Merdin: @danharrin  Not sure if you want to mention something like this in the "main" documentation where sorting is explained

danharrin (approved):

PR #17874, opened 2025-09-19
Fix keyboard navigation issues with markdown editors

### LINKED ISSUE #17854: Keyboard navigation issues with Markdown Editors
### Package

filament/filament

### Package Version

v4

### Laravel Version

v12

### Livewire Version

v3

### PHP Version

PHP 8.4

### Problem description

When there are adjacent Markdown Editors in a form, you are unable to use the `tab` key to navigate with the keyboard from one markdown editor to the next. This results in a keyboard trap.

Additionally using the `shift-tab` keyboard combination from within a markdown editor always moves focus back to the first focusable element in the form instead of to the previous one.

### Expected behavior

Using the `tab` should move to the next focusable element. Using the `shift-tab` keyboard combination should move to the previous focusable element.

### Steps to reproduce

1. Open a form that that has two or more adjacent Markdown Editors. (see reproduction repository for example)
2. Place focus into the first Markdown editor
3. Without selecting any text, press the `tab` key on the keyboard to attempt to move focus to the next Markdown Editor
4. Notice that focus remains in the current Markdown Editor
5. Using the keyboard press `shift-tab` keyboard combination
6. Notice that focus is moved to the first focusable element in the form instead of the previous one.

https://github.com/jobara/filament-markdown-editor-keyboard-trap

### Relevant log output

```shell

```

alex-elivate:
## Description
Fixes an issue where tab gets trapped if there are multiple markdown editors on the page.

## Visual changes
None

## Functional changes
Tab will not move to the next (or previous) item.  This issue was caused by trying to tab into hidden elements. Filtered out these elements when calculating where to land.
Fixes #17854


danharrin:
Thanks!

danharrin (approved):

PR #17882, opened 2025-09-21
Fix: ToggleButtons and Radio buttons lose keyboard focus

### LINKED ISSUE #16946: Accessibility Issue: ToggleButtons and Radio buttons lose keyboard focus after live updates
### Package

filament/filament

### Package Version

v3.3.31

### Laravel Version

v12

### Livewire Version

_No response_

### PHP Version

PHP 8.3

### Problem description

Hi,
I'm experiencing an accessibility issue with Filament form components that have live() enabled. When a ToggleButton or Radio button triggers dynamic updates (e.g., updating a Placeholder component), the form elements become unresponsive to keyboard navigation.

https://github.com/user-attachments/assets/1cb3e3ad-7288-4306-ba64-bca1b95dc0cb

### Expected behavior

Expected behavior:

- Should be able to change Radio button selections using arrow keys
- Should be able to toggle ToggleButtons using Space/Enter
- Focus should remain on the current element after live updates

Actual behavior:

- After the first selection, the Radio and Toggle button cannot be changed using keyboard
- The only workaround is to Tab to the next element and then Tab back

### Steps to reproduce

1. Create a form with ToggleButtons or Radio buttons that have live() enabled
2. Add dependent components (like Placeholder) that update based on the live values
3. Navigate using keyboard (Tab) and select an option
4. Try to change the selection using keyboard

https://github.com/alanost/filamentphp.accessibility-focus-with-live

### Relevant log output

```shell

```

alex-elivate:
## Description
Fixes #16946
I was unable to fix this with with Livewire. Focus state is supposed to be preserved, but I think the morphing process is causing the elements to be replaced in the browser with an evil twin.

## Visual changes
This fix causes live radio buttons to flash.  I understand if this doesn't make the cut.

https://github.com/user-attachments/assets/214ccaad-eb52-4b5e-83fd-71e2020668ff

## Functional changes
Lifecycle hook to refocus element after morph.

### DIFF
--- a/packages/forms/resources/views/components/radio.blade.php
+++ b/packages/forms/resources/views/components/radio.blade.php
@@ -36,9 +36,27 @@
                     ], escape: false);
             @endphp
 
-            <label class="fi-fo-radio-label">
+            <label class="fi-fo-radio-label" 
+                x-data="{
+                    init() {
+                        Livewire.hook('morph.updated', () => {
+                            if (window.focusAfterMorph) {
+                                const el = document.getElementById(window.focusAfterMorph)
+                                if (el) {
+                                    el.focus()
+                                }
+                                window.focusAfterMorph = null
+                            }
+                        })
+                    },
+                    handleChange(event) {
+                        window.focusAfterMorph = event.target.id
+                    }
+                }"
+                >
                 <input
                     type="radio"
+                    @change="handleChange"
                     {{
                         $inputAttributes->class([
                             'fi-radio-input',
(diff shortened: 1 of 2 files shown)


danharrin (changes requested):
If a button is focused, then another input is focused, I think this solution breaks on the next morph since it then refocuses the old input, right?

Is it worth closing this and maybe creating an issue discussion in the Livewire repo instead?

alex-elivate:
Agreed.  If a user selects another component before the hook is fired, the wrong component could be focused. I would love to learn more about how livewire tracks focus after the dom is updated. While this hack may be a usable workaround, it's not fixing the core issue.

PR #17895, opened 2025-09-22
Add table header cell render hook

archilex:
Similiar to the filter indicator renderhooks, this adds new renderhooks for the table header cell and the column cell. (the majority of the "additions" is just indenting).


danharrin (commented):
Can you benchmark the performance of this change please? Mainly to existing codebases that don't use this feature, but also if it would cause performance issues when it is used

archilex:
Here's my testing (averages):

1. When no view, conditional check for table header renderhook: 0.004 ms
2. When no view, conditional check for table cell renderhook: 0.004 ms
3. When view, checking and rendering the current table header html in a view using the renderhook: 0.147 ms
4. When view, checking and rendering the current table cell html in a view using the renderhook: 0.262 ms

The above was measured in the blade file using Benchmark::dd(). I didn't have a good way to benchmark the actual html portion of inside the index.blade.php since it's a mix of php and html and it wasn't playing nice with Benchmark so I don't have a comparisson point for those chunks of html compared to the view. (Spatie @measure was adding 25ms even without content so that didn't work either.)

I did however just refresh the page multiple times with debug bar, and the difference between current blade and using views was neglible. Both were averaging around 135ms/page load.

packages/tables/resources/views/index.blade.php
```diff
@@ -1865,68 +1872,81 @@ class="fi-ta-record-checkbox fi-checkbox-input"
...
-
-                                                        if ($columnWrapperTag === 'button') {
-                                                            if ($columnAction instanceof \Filament\Actions\Action) {
-                                                                $columnWireClickAction = "mountTableAction('{$columnAction->getName()}', '{$recordKey}')";
-                                                            } elseif ($columnAction) {
-                                                                $columnWireClickAction = "callTableColumnAction('{$column->getName()}', '{$recordKey}')";
-                                                            } else {
-                                                                if ($this->getTable()->getAction($recordAction)) {
-                                                                    $columnWireClickAction = "mountTableAction('{$recordAction}', '{$recordKey}')";
+                                                    @if (filled($tableCellView = FilamentView::renderHook(TablesRenderHook::TABLE_CELL, scopes: static::class, data: [
```
wychoong: will it be better if the checking is a hook exists for TABLE_CELL + scopes only run once per table/column instead of every single cell?

archilex:
@wychoong That's not possible with `FilamentView::renderHook()` as you have it as that actually renders the view which needs the $data. However, it's probably possible to introduce a method to get the table's renderhooks and see if it exists. That way it'd be possible to call it once per table.

That said, Dan, I'm reevaluating if I even need the cell renderhook. I definetly need the header renderhook, but I might have a way around needing the cell. 

Let me know if you want me to explore adding a new `hasRenderHook()` method.

wychoong:
> @wychoong That's not possible with `FilamentView::renderHook()` as you have it as that actually renders the view which needs the $data. However, it's probably possible to introduce a method to get the table's renderhooks and see if it exists. That way it'd be possible to call it once per table.
> 
> That said, Dan, I'm reevaluating if I even need the cell renderhook. I definetly need the header renderhook, but I might have a way around needing the cell.
> 
> Let me know if you want me to explore adding a new `hasRenderHook()` method.

Yup that’s what I meant

danharrin (changes requested):
Yeah a `hasRenderHook()` (cached into a variable in the table blade view) would be great to avoid checking for each cell

And of course if we can not introduce a cell hook at all that is even better

archilex:
Dan, I was able to work around needing the table cell view so I removed that. I went ahead and implemented the hasRenderHook method and then used that for the header cells. Not quite as necessary, but it'll still help.

danharrin (approved):

PR #17906, opened 2025-09-22
fix a11y radio and toggle button lose focus when live

### LINKED ISSUE #16946: Accessibility Issue: ToggleButtons and Radio buttons lose keyboard focus after live updates

alex-elivate:
## Description
Fixes #16946
While it's best practice to disable components on update, this removes focus as the component is disable during a livewire morph. The risk of a race condition for these components are much less due to the nature of the component.

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/forms/resources/views/components/radio.blade.php
+++ b/packages/forms/resources/views/components/radio.blade.php
@@ -31,7 +31,6 @@
                         'id' => $id . '-' . $value,
                         'name' => $id,
                         'value' => $value,
-                        'wire:loading.attr' => 'disabled',
                         $wireModelAttribute => $statePath,
                     ], escape: false);
             @endphp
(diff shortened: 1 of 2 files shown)


danharrin (approved):

danharrin:
Thanks for investigating :)

Issue #17382, opened 2025-08-18
SPA mode does not always update URL correctly


robinmiau:
### Package

filament/filament

### Package Version

v4.0.1

### Laravel Version

v12.3.0

### Livewire Version

v3.6.4

### PHP Version

PHP 8.3.24

### Problem description

As can be seen in the video, there is a point at which the URL no longer updates. This also happens even if you don't navigate so quickly.

https://github.com/user-attachments/assets/96db9595-1d49-4cb4-ab6d-b8299933ba5e

### Expected behavior

It should always display the correct URL.

### Steps to reproduce

1. Enable SPA mode (`->spa()` in AdminPanelProvider)
2. Navigate between pages; it does not necessarily have to be as fast as in the video.

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/filamentphp/demo

### Relevant log output

```shell

```

CorwinDev:
Related to https://github.com/livewire/livewire/discussions/9450, Livewire trys to call history.replaceState() to quickly resulting in the browser error `Throttling navigation to prevent the browser from hanging. See https://crbug.com/1038223. Command line switch --disable-ipc-flooding-protection can be used to bypass the protection`

saschaende:
> Related to [livewire/livewire#9450](https://github.com/livewire/livewire/discussions/9450), Livewire trys to call history.replaceState() to quickly resulting in the browser error `Throttling navigation to prevent the browser from hanging. See https://crbug.com/1038223. Command line switch --disable-ipc-flooding-protection can be used to bypass the protection`

Confirmed

But what I could also see:

- On a Filament panel as a GUEST I dont get this error, I can click around as fast as I want = no error
- After logging in with my account I get this error after ~4 fast clicks in the navigation, then the browser hangs with throtteling message in the console.

I have no explanation why this problem only appears, **when a user is logged in**. I hope this helps somebody to find the issue. Because it must be related to users/authentification.... idk.


### Solution (atually helped me)

```
        return $panel
            ->spa(hasPrefetching: true)
```

https://filamentphp.com/docs/4.x/panel-configuration#spa-mode

bzy107:
This issue may occur regardless of SPA.
At least in my local environment, there are cases where clicking the navigation does not respond.

ralphjsmit:
I can verify this issue (with SPA mode & prefetching _enabled_).

chrispage1:
[...]

PR #17911, opened 2025-09-23
Fix browser navigation throttling issues

### LINKED ISSUE #17880: [4.x] Browsers throttle navigation because of multiple `history.replaceState()` commands
### Package

filament/filament

### Package Version

v4.0.18

### Laravel Version

v12.30.1

### Livewire Version

v3.6.4

### PHP Version

PHP 8.4.12

### Problem description

Just switching pages can result in the browser throttling the navigation which means people are stuck on the page for some time and need to click again, which feels like the page is frozen.

Image

https://discord.com/channels/883083792112300104/1397514827278057613

### Expected behavior

Normal navigation. No freezes.

### Steps to reproduce

This can be used to add logging to `history.replaceState()`. The open any page and see that is runs `history.replaceState()` ~50 times. 

```php
    PanelsRenderHook::BODY_END,
    fn () => new HtmlString(<<<'HTML'
        <script>
            (() => {
                const original = window.history.replaceState;

                window.history.replaceState = function (...args) {
                    console.log('history.replaceState', args);

                    original.apply(this, args);
                }
            })();
        </script>
    HTML)
);
```

-

### Relevant log output

```shell

```

binaryfire:
Fixes https://github.com/filamentphp/filament/issues/17880 and https://github.com/filamentphp/filament/issues/17382

SPA mode navigation in Filament panels becomes throttled/frozen due to excessive `history.replaceState()` calls (often ~50 per navigation). This is caused by multiple Livewire components using the #[Url] attribute, each triggering their own history state update.

This PR adds client-side deduplication that intercepts `history.replaceState()` calls and skips duplicates by comparing the new state with the current state. This reduces those ~50 calls down to only the unique ones, which eliminates browser throttling issues. Fixing throttling also fixes problems with browser URLs not being updated correctly on page changes.

This can be removed once (if?) this is fixed in Livewire core. However there are multiple open issues for this going back a couple of years so I'm not sure if that will happen any time soon.


davidwhicker:
does this fix the issue for those not using SPA mode as well?

binaryfire:
> does this fix the issue for those not using SPA mode as well?

@davidwhicker This fixes all issues caused by hitting the history API too many times (in both SPA and non-SPA mode).

pxlrbt:
We should still push a fix in Livewire itself. I think it's a good workaround for now, but we should remove this as soon as possible again.

binaryfire:
> We should still push a fix in Livewire itself. I think it's a good workaround for now, but we should remove this as soon as possible again.

@pxlrbt Yeah, would definitely be better if this was addressed upstream. But there [was a PR](https://github.com/livewire/livewire/pull/9059/files) which took a similar deduplication approach and it was closed. Josh said he had other ideas on how to solve this, possibly as part of [this WIP PR](https://github.com/livewire/livewire/pull/9366), but that hasn't been updated since June so not sure what's happening there.

Livewire contributions from non-team members are rarely merged nowadays:
https://github.com/livewire/livewire/pulls?q=is%3Apr+is%3Aclosed
https://github.com/livewire/livewire/pulls
And the team almost never reply to discussions. 

Perhaps you or Dan will have more luck but the odds of any PR I'd make being merged are pretty low IMHO 🤷 Maybe it'll be addressed in v4.

pxlrbt:
Dan already contacted Josh. No idea how soon we can see a fix. But I guess we can merge this until then.

danharrin (approved):

binaryfire:
@danharrin I just realised this is breaking query params. Working on a fix now.

packages/panels/resources/js/history-state.js
```diff
@@ -0,0 +1,27 @@
...
+            return
+        }
+    } catch (error) {
+        // If comparison fails, proceed with the update
+    }
+
+    originalReplaceState.call(window.history, state, unused, url)
+}
+
+window.history.pushState = function (state, unused, url) {
```
danharrin, line 21: I think I understand the `replaceState()` override. However, I don't understand this `pushState()` one, which just appears to convert the URL to a string? @binaryfire

binaryfire: AFAIK, the `pushState` override is needed because Livewire is passing URL objects in the state. Without normalizing them to strings in both `pushState` and `replaceState`, the duplicate detection would fail. If you only normalize in `replaceState`, you'd end up with mixed formats in history:

```JS
// From pushState (without the override):
{ url: URL {href: "https://example.com", ...} }

// From replaceState (with the override):
{ url: "https://example.com" }
```

URL objects stringify to `{}` with `JSON.stringify()`, so when `replaceState` tries to compare states to detect duplicates, it would be comparing:

- Current state: `{"url":{}}` (the URL object from a previous pushState)
- Incoming state: `{"url":"https://example.com"}` (the normalized string)

Both methods need to normalize URL objects to strings so all history entries have consistent format and the comparison actually works.

That was my thinking anyway. It works with it and I didn't test it without it.

PR #17920, opened 2025-09-23
feat: add support for default collapsed grouping in table records.

GeminiDev1:
## Description

This PR implements the requested idea from:
https://github.com/filamentphp/filament/discussions/10717

Currently, groups using the `->collapsible()` method are expanded by default when the table loads.

This PR introduces a new `->groupingDefaultCollapsed()` method, which allows all groups to be collapsed by default.

## Visual changes

**Before**

image

**After**

Applied: `groupingDefaultCollapsed()` method.

image

## Functional changes


danharrin (changes requested):
Instead of getting all group titles and calling `toggleCollapseGroup` for each, which will then have to re-run for each pagination page when it changes, make it so that instead of `toggleCollapseGroup` adding the group to the array when it is collapsed, add the group to the array when it is expanded

So when the setting is disabled, it should act like it does now and track collapsed groups, and when the setting is enabled, it should instead track expanded groups

As such, all the conditions for checking if a group is collapsed or not should be flipped if the setting is enabled

Does that make sense?

GeminiDev1:
> Instead of getting all group titles and calling `toggleCollapseGroup` for each, which will then have to re-run for each pagination page when it changes, make it so that instead of `toggleCollapseGroup` adding the group to the array when it is collapsed, add the group to the array when it is expanded
> 
> So when the setting is disabled, it should act like it does now and track collapsed groups, and when the setting is enabled, it should instead track expanded groups
> 
> As such, all the conditions for checking if a group is collapsed or not should be flipped if the setting is enabled
> 
> Does that make sense?

Hi @danharrin ,

Thanks for the feedback, I have updated the implementation.

Let me know if you'd like me to tweak anything else.

danharrin (approved):

danharrin:
Thanks

PR #17933, opened 2025-09-24
Indicate that the $pollingInteral variable must be declared static

nickturrietta:
## Description

$pollingInterval must be declared static

## Visual changes

n/a


danharrin:
I'm not sure what issue you are hitting, but this property should not be static, I just checked the definition in the codebase.

nickturrietta:
Apologies. I was looking at v4 docs instead of v3.

PR #17955, opened 2025-09-26
fix: adjust select field wrapper for proper inline label alignment

### LINKED ISSUE #17951: Fieldset inlineLabel SelectInput not middle vertical aligned
### Package

filament/filament

### Package Version

v4.0.20

### Laravel Version

v12.31.1

### Livewire Version

v3.6.4

### PHP Version

PHP 8.4.11

### Problem description

```php
Select::make('currency_id')
    ->relationship('currency', 'name')
    ->selectablePlaceholder(false)
    ->required()
```

Image
* Notice the TextInput on left have the labels middle vertical aligned, while on the right the SelectInput does not

Image

Without required, same result

Image

### Expected behavior

All labels in the form should be aligned the same.

### Steps to reproduce

# Using attached repo
- Install as per normal
- Start php server (I use php artisan serve)
- Follow the README
- Visit URL: http://localhost:8000/admin/customers/10/edit
- should see the following...

Image

https://github.com/christoferd/filament-bug-demo-2025

### Relevant log output

```shell

```

People-Sea:
## Description

Fixes: #17951

The implementation method is consistent with the `text-input` component.

## Visual changes

# Before
image

# After

image

## Functional changes


People-Sea:
image

Additionally, many components are not vertically centered. I think when using inlineLabel, all labels should be aligned to the top instead of using padding-top to change their position.    

Vertical centering will make the form look very messy.

danharrin:
@People-Sea I have pushed changes to fix the consistency here

- Components with a fixed height (checkbox, toggle, text input) now have working vertical centering, so that the label is inline with the field
- Components without a fixed height, but with text that should be in line with the label (select placeholder/value should be inline with the label, textarea placeholder/value should be inline with the label) now have top padding on the label
- Components without a fixed height and without text that should be aligned (file upload, repeater, etc) have their label aligned to the top of the container

When I created the `inlineLabel()` feature, I didn't really intend for it to be used in a multi-column layout like you sent. This is why the label alignment is not the same across components, because it is optimised for a single column layout where vertical space is limited.

What do you think?

People-Sea:
@danharrin 
Thank you for the additional context on the PR — your explanation made it clear why inlineLabel looks inconsistent in multi-column layouts. Handling alignment based on whether a component has a fixed height makes sense, and the current changes look good. Users can still inject their own styles if they want a different alignment.

danharrin (approved):

PR #17964, opened 2025-09-26
Feature table hide select all checkbox

muesker:
## Description

Ive added the option to hide the table select all records checkbox.
The feature was also requested here: https://github.com/filamentphp/filament/discussions/5608

## Visual changes

image
image

## Functional changes

### DIFF
--- a/packages/tables/docs/04-actions.md
+++ b/packages/tables/docs/04-actions.md
@@ -257,15 +257,15 @@ BulkAction::make('delete')
                     if (($failureCount === 1) && ($totalCount === 1)) {
                         return 'One user failed to delete.';
                     }
-        
+
                     if ($failureCount === $totalCount) {
                         return 'All users failed to delete.';
                     }
-        
+
                     if ($failureCount === 1) {
                         return 'One of the selected users failed to delete.';
                     }
-        
+
                     return "{$failureCount} of the selected users failed to delete.";
                 },
             );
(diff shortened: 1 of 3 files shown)


muesker:
Maybe the function „selectAllAvailable“ is not that great because it could also mean that you select all available records in the list.
Ill think about another name if you guys think the same.

danharrin:
Hi, thanks for your contribution!

Since that discussion was opened, we implemented a feature called `selectCurrentPageOnly()`, which solves the same problem - instead of selecting all records in the table, it only selects the current page.

As such, I think we probably shouldn't go ahead with this PR since it solves the same problem.

Recently, we also implemented `maxSelectableRecords()` which could be useful to you.

If we get a lot of feedback indicating that this feature is required for many people, we will reconsider it again. But for now, it's important to keep the API consistent and simple, and not add too many methods which do very similar things.

Hope this helps!

muesker:
Hey, i get ur point so here is just my take why i need this feature:
I have a table with items from different customers but you can only select items from the same customer.
This is possible with my local patch where I live entangled the selection of records and after my first selection i only allow selection of the same customer but when you select all at once i cant handle it.
The live entanglement of selected records would be my next PR but if you think its not worth to have I will just build local patches.

danharrin:
You could use CSS to hide the elements you don't want, maybe?

PR #17966, opened 2025-09-26
fix(panels/auth): guard EmailVerificationPrompt for guests (redirect or 403)

### LINKED ISSUE #17885: [4.x] null returned when email verification is required on guest panel if user is not authunticated
### Package

filament/filament

### Package Version

v.4.0

### Laravel Version

12.30.1

### Livewire Version

_No response_

### PHP Version

PHP 8.4.0

### Problem description

For guest panel if `Authenticate::class` middleware is removed and added `emailVerification()` on the panel the below error is shown

```
Filament\Auth\Pages\EmailVerification\EmailVerificationPrompt::getVerifiable(): Return value must be of type Illuminate\Contracts\Auth\MustVerifyEmail, null returned
```

### Expected behavior

the EmailVerificationPrompt::class page should only be shown only if the user is logged in.

### Steps to reproduce

1. remove route `Route::get('/', ...)` from web.php
2. install filament "site" panel
3. update panel path to `/`
4. remove `AccountWidget::class`
5. remove `Authenticate::class` middleware
6. add `->emailVerification()` to panel

https://github.com/3adeling/filament-not-authenticated-email-verification-issue

### Relevant log output

```shell

```

kerimala:
## Description
When a guest visits a panel with `->emailVerification()` (and `Authenticate` middleware is not present), the EmailVerificationPrompt tried to resolve a verifiable user from `auth()` and crashed on `null`.

This PR adds an early guard:
- If no authenticated user exists, redirect to the **current panel’s** login route.
- If the panel has no login route, return **403**.
- Authenticated users that implement `MustVerifyEmail` still see the prompt as before.

**Scope:** minimal, page-level; no public API changes; multi-panel safe (uses the current panel’s login route).

## Visual changes
None.

## Functional changes
- Guest → **302** to panel login (or **403** if no login route).
- Authenticated (not `MustVerifyEmail`) → **403** (existing behavior).
- Authenticated (`MustVerifyEmail`) → **200**, prompt loads.

## Tests
Feature tests cover:
- Guest redirect (or 403 fallback).
- Non-verifiable authenticated user → 403.
- Verifiable authenticated user → 200.

## Notes
Fixes filamentphp/filament#17885

### DIFF
--- a/packages/panels/src/Auth/Pages/EmailVerification/EmailVerificationPrompt.php
+++ b/packages/panels/src/Auth/Pages/EmailVerification/EmailVerificationPrompt.php
@@ -27,26 +27,15 @@ class EmailVerificationPrompt extends SimplePage
 
     public function mount(): void
     {
-        if ($this->getVerifiable()->hasVerifiedEmail()) {
+        if ((! Filament::auth()->check()) || $this->getVerifiable()->hasVerifiedEmail()) {
             redirect()->intended(Filament::getUrl());
         }
     }
 
     protected function getVerifiable(): MustVerifyEmail
     {
-        $user = Filament::auth()->user();
-
-        if ($user === null) {
-            $panel = Filament::getCurrentOrDefaultPanel();
-
-            if ($panel?->hasLogin()) {
-                throw new HttpResponseException(new RedirectResponse($panel->route('auth.login')));
-            }
-
-            abort(403);
-        }
-
         /** @var MustVerifyEmail $user */
+        $user = Filament::auth()->user();
 
         return $user;
     }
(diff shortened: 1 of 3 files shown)


packages/panels/src/Auth/Pages/EmailVerification/EmailVerificationPrompt.php
danharrin, line 1: Any reason why this shouldn't be in `mount()` instead? I don't think `getVerifiable()` is the right place?

Allow the `getVerifiable()` to return `null` and then check what it returns?

danharrin (changes requested):

packages/panels/src/Auth/Pages/EmailVerification/EmailVerificationPrompt.php
danharrin: I'm also not sure that we should redirect them to login - the original issue specifically says that this is a guest panel so unauthenticated users can access. If there is no user, I guess we should redirect them to the panel?

danharrin:
@kerimala are you able to check my above comments? Thanks

danharrin (approved):

PR #17972, opened 2025-09-27
Update Field.php

tanthammar:
Not sure about the difference in 8.2+ php versions, but I get a warning from this...

image


danharrin:
Hmm I'm not seeing this on PHP 8.2 currently, and PHPStan is not flagging it either. If you get the error in the browser then I will add this, but I think it could be a PHPStorm error for you...

PR #18025, opened 2025-10-01
Update submitAction in wizards documentation

Michel-Verhoeven:
## Description

When using `<x-filament::button />` in the submitAction of the wizard you need to set the `wire:target` for the loading button to work. If you don't set the target the button in still disabled on loading but the default loading indicator isn't shown. This is only an enhancement of the documentation.


danharrin:
Thanks, but this is not necessarily the action name that is used every time a wizard is in a schema. In fact, it is only used when a wizard is inside a modal. Outside of modals, the submit target could be any Livewire action name.

PR #18028, opened 2025-10-01
Implement caching for component state path resolution

oddvalue:
## Description

This PR optimizes the performance of the `Get` helper and `getComponentByStatePath()` method, which were causing significant performance issues when used frequently in forms and schemas.

### Problem
The `getComponentByStatePath()` method performs a full recursive tree traversal on every call to locate components by their state path. When the `Get` helper is used extensively (e.g., in computed properties, validation, or reactive field updates), these repeated traversals create a substantial performance bottleneck.

### Solution
Implemented caching at two levels:

1. **`getComponentByStatePath()` caching** - Added a cache layer in `HasComponents` trait that stores component lookups by state path, hidden flag, and skip component parameters
2. **`Get` helper caching** - Added a static cache in the `Get` class to avoid repeated lookups for the same paths
3. **Early return optimization** - Added an early return in `Get` when the path refers to the component itself

### Safety
The caching is safe because:
- We cache **component object references**, not state values
- State is retrieved dynamically via `$component->getState()` on each `Get` call
- When `Set` updates a component's state, the cached reference still points to the same object, so changes are immediately visible
- Caches are properly cleared when components are modified via `components()` or `cloneComponents()` methods

## Visual changes

N/A - This is a performance optimization with no visual changes.

## Functional changes

### DIFF
--- a/packages/schemas/src/Components/Utilities/Get.php
+++ b/packages/schemas/src/Components/Utilities/Get.php
@@ -6,6 +6,11 @@ use Filament\Schemas\Components\Component;
 
 class Get
 {
+    /**
+     * @var array<string, Component | null>
+     */
+    protected static array $componentCache = [];
+
     public function __construct(
         protected Component $component,
     ) {}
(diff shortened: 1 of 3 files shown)


danharrin:
Hi, thanks for this. I have incorporated similar code into #18068.

However, we cannot introduce a static cache on the `Get` or `Set` helpers. Static array caches continually add array items throughout the request lifecycle, and in environments that use tools like Octane this introduces memory leaks, since the static array is preserved between multiple requests and only gets bigger over time. This is especially a problem when objects are stored in this array, and objects then get shared across multiple requests, which are stale references.

PR #18030, opened 2025-10-02
Fix/dropdown loading indicator

### LINKED ISSUE #13925: loading indicator not work on dropdown
### Package

filament/support

### Package Version

v3.2.98

### Laravel Version

v11.20.0

### Livewire Version

v3.5.4

### PHP Version

PHP 8.3.10

### Problem description

i have a Livewire component and it use filament dropdown, when i change the team, the loading indicator not showing, the chevron icon has disappear until the status updated and notification show

### Expected behavior

i want the loading indicator can be showing on dropdown list item

### Steps to reproduce

```blade 
<div>
    <x-filament::dropdown
            maxHeight="250px"
            placement="left-start"
            teleport="true"
    >
        <x-slot name="trigger">
            <x-filament::dropdown.header
                    class="font-semibold"
                    color="gray"
                    icon="heroicon-c-user-group"
            >
                {{ auth()->user()->currentTeam->name ?? 'Select Team' }}
            </x-filament::dropdown.header>
        </x-slot>

        <x-filament::dropdown.header
                class="font-medium"
                color="gray"
        >
            {{ __('Select Team') }}
        </x-filament::dropdown.header>


        <x-filament::dropdown.list>
            @foreach(auth()->user()->allTeams() as $team)
                        :color="auth()->user()->isCurrentTeam($team) ? 'primary' : null"
                        icon="heroicon-m-chevron-right"
                        tag="button"
                        :disabled="auth()->user()->isCurrentTeam($team) ?? false"
                        wire:click="changeTeam({{ $team }})"
                >
                    {{ $team->name }}
            @endforeach
        </x-filament::dropdown.list>
    </x-filament::dropdown>
</div>
```

```php
<?php

namespace App\Livewire;

use App\Models\Team;
use Filament\Notifications\Notification;
use Livewire\Component;

class SwitchTeam extends Component
{
    public function render(): View
    {
        return view('livewire.switch-team');
    }

    public function changeTeam(Team $team): void
    {
        auth()->user()->switchTeam($team);

        Notification::make()->title('Switched to '.$team->name)->success()->send();
    }
}

```

https://github.com/holiq/FilaTeam

### Relevant log output

_No response_

edmer664:
### Problem
Loading indicators in dropdown list items were not showing during Livewire operations. The chevron icon would disappear but no loading indicator would appear, making it unclear to users that an action was processing.

### Root Cause
The issue was caused by improper attribute escaping in the loading indicator component. The `wire:target` attribute was being escaped with backslashes (`\"`) instead of HTML entities (`&quot;`), which broke Livewire's ability to properly detect and bind loading states.

**Problematic output:**
```html
wire:target="changeTeam({\"id\":5,\"name\":\"DJ Banser\"})"
```

**Expected output:**
```html
wire:target="changeTeam({&quot;id&quot;:5,&quot;name&quot;:&quot;DJ Banser&quot;})"
```

### Solution
The fix aligns the loading indicator component with the icon component's attribute handling by removing an unnecessary `html_entity_decode()` call that was causing the escaping inconsistency.

**Changes:**
- Removed `html_entity_decode($wireTarget, ENT_QUOTES)` 
- Now passes `$wireTarget` directly through `ComponentAttributeBag`
- Both icon and loading indicator components now handle attributes consistently

### Testing
Tested with a Livewire component containing a 5-second processing delay:
- ✅ Loading indicator now shows correctly during wire operations
- ✅ Chevron icon properly disappears/reappears as expected
- ✅ No regression in existing dropdown functionality
- ✅ Proper HTML entity escaping maintained

### Files Changed
- `resources/views/components/dropdown/list/item.blade.php`

### Environment
- **Package**: filament/support v3.2.98
- **Laravel**: v11.20.0  
- **Livewire**: v3.5.4
- **PHP**: 8.3.10

This change ensures architectural consistency between components while maintaining Laravel's standard attribute escaping behavior.

## Visual changes

image

## Functional changes

### DIFF
--- a/packages/panels/resources/lang/lv/pages/auth/password-reset/request-password-reset.php
+++ b/packages/panels/resources/lang/lv/pages/auth/password-reset/request-password-reset.php
@@ -32,7 +32,7 @@ return [
 
     'notifications' => [
 
-         'sent' => [
+        'sent' => [
             'body' => 'Jūs saņemsiet e-pastu, ja Jūsu konts eksistē.',
         ],
 
--- a/packages/panels/src/GlobalSearch/GlobalSearchResult.php
+++ b/packages/panels/src/GlobalSearch/GlobalSearchResult.php
@@ -12,9 +12,9 @@ class GlobalSearchResult
      * @param  array<Action>  $actions
      */
     public function __construct(
-        readonly public string | Htmlable $title,
-        readonly public string $url,
-        readonly public array $details = [],
-        readonly public array $actions = [],
+        public readonly string | Htmlable $title,
+        public readonly string $url,
+        public readonly array $details = [],
+        public readonly array $actions = [],
     ) {}
 }
--- a/packages/panels/src/Resources/RelationManagers/RelationManagerConfiguration.php
+++ b/packages/panels/src/Resources/RelationManagers/RelationManagerConfiguration.php
@@ -9,7 +9,7 @@ class RelationManagerConfiguration
      * @param  array<string, mixed>  $properties
      */
     public function __construct(
-        readonly public string $relationManager,
+        public readonly string $relationManager,
         protected array $properties = [],
     ) {}
 
--- a/packages/support/resources/views/components/dropdown/list/item.blade.php
+++ b/packages/support/resources/views/components/dropdown/list/item.blade.php
@@ -90,7 +90,7 @@
     $hasLoadingIndicator = filled($wireTarget);
 
     if ($hasLoadingIndicator) {
-        $loadingIndicatorTarget = html_entity_decode($wireTarget, ENT_QUOTES);
+        $loadingIndicatorTarget = $wireTarget;
     }
 
     $hasTooltip = filled($tooltip);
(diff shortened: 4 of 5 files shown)


danharrin:
We use `html_entity_decode` in other places to encode targets, why is it only dropdowns that are affected by this problem?

edmer664:
@danharrin That's a good question.

I just tested it with the **button** component and it behaves the same as the dropdown. It doesn't render the `wire:target` correctly.
image

The reason why the icon component is being correctly rendered might be due to the use of the `@svg` directive to apply the attributes. In comparison to the loading indicator which does not use it that's why it is not rendering properly. 

_icon.blade.php_
```
    @svg(
        $icon,
        $class,
        array_filter($attributes->getAttributes()),
    )
 ```
 
image

How should we approach this? Should I just remove the `html_entity_decode` from the other components as well? Or do we wrap the `targets` with `e()` to make sure they are escaped when passed?

danharrin:
I don't think the HTML entities being output here are correct. In Filament core we don't experience the issue and we output arguments using `Js::from()`. I am going to comment this on the original issue.

PR #18031, opened 2025-10-02
Rich editor: content schemas and node placeholders

thijskuilman:
### Enforcing a content structure

It's now possible to pass a content schema to the editor, based on the example here: https://tiptap.dev/docs/examples/advanced/forced-content-structure.

I have updated the documentation to include instructions on how to use this.

### Node specific placeholders
You can now set placeholders for specific node types. The code implementation is based on the official example, [see the code example here](https://tiptap.dev/docs/editor/extensions/functionality/placeholder). I have updated the documentation accordingly.

### Show only current placeholder
I have added the [showOnlyCurrent property](https://tiptap.dev/docs/editor/extensions/functionality/placeholder#showonlycurrent) to RichEditor and updated the documentation. This allows you to show multiple placeholders at the same time.

## Visual changes

In the example below, I do the following:

- define a content schema `heading block+`
- applied a different node placeholder for `heading` and `paragraph`
- set `showOnlyCurrent` to `false`

afbeelding

## Functional changes

> [!NOTE]
> By default, Filament inserts an empty paragraph into RichEditor, which makes the validation and default state a bit cumbersome when working with content schemas. I have described workarounds for this in the README, but I'm curious what the reasons are for the default behaviour.


danharrin:
Hi, thanks for your PR.

For now, I am not going to introduce this new feature, as I have not had enough demand from the community to include it yet.

However, using #18049, you should be able to override the default Document and Placeholder extensions, to inject this behaviour. You should be able to accomplish this like that in your own app or plugin.

PR #18033, opened 2025-10-02
fix: Responsive chart height

### LINKED ISSUE #17302: Charts in Chart Widgets not resizing
### Package

filament/filament

### Package Version

v4.0.0

### Laravel Version

v12.0.0

### Livewire Version

_No response_

### PHP Version

8.3

### Problem description

When resizing the browser window and making it smaller, charts inside widgets shrink accordingly. However, when making the browser window larger again, the chart widgets grow again, but not the chart inside it.

This behaviour can be observed on the line chart widgets at https://demo.filamentphp.com/

### Expected behavior

Charts should grow in size when their widget grows.

### Steps to reproduce

Go to https://demo.filamentphp.com/, resize the window and make it smaller, then make it larger again.

### Relevant log output

```shell

```

danharrin:
Fixes #17302


somegooser:
Nice!! Thanks for this.

I think the StatsOverviewWidget is also missing this!

image

danharrin:
The stats charts already have their own height configuration, and I haven't experienced the issue you screenshotted. Can you please open an issue with a reproduction repository?

somegooser:
I wil try to do so!

This is a basic admin panel with Demo Data Seed from LunarPHP Core package!

danharrin:
If you look at the official Filament demo, it doesn't have the issue

somegooser:
I added additional text in .fi-wi-stats-overview-stat-description > span via element inspect tool on the official admin panel:

image

danharrin:
I don't see the problem here, the charts look like the correct height

somegooser:
I’m not sure if it’s the right term, but I meant the “StatsOverviewWidget”.

You can see three blocks: Revenue, New Customers, and New Orders.

The block with Revenue is not the same height as the other two blocks.

danharrin:
Sorry, that is not related to this PR

PR #18040, opened 2025-10-02
fix: [v4] allow custom relations with tenant.

mathieutu:
## Description

Hi, 
With the v4 migration, I now have an error when trying to create a new entry of a model linked to a tenant via a [BelongsToThrough](https://github.com/staudenmeir/belongs-to-through) relationship.

I've updated the hook to ensure that a non-existent `save` function is not called for any custom relationship class.

Thanks.

## Visual changes

N/A

## Functional changes

### DIFF
--- a/packages/panels/src/Resources/Resource/Concerns/BelongsToTenant.php
+++ b/packages/panels/src/Resources/Resource/Concerns/BelongsToTenant.php
@@ -10,6 +10,7 @@ use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Database\Eloquent\Relations\MorphTo;
 use Illuminate\Database\Eloquent\Relations\Relation;
 use LogicException;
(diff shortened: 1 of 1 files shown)


danharrin (changes requested):
Please explicitly check for the relation type rather than checking if the method exists

mathieutu:
Hey @danharrin.
The goal is precisely to handle any type of custom relation. We just ensure that the method exists before calling it. 

(We could add the specific type of relation I need right now, but the issue will be raised then with another custom type).

Why don't you like this way of doing? (On my side, I would even remove the type checking, and just keep the `save` test.)

danharrin:
I don't want other relation types without a `save()` method to silently fail, I want them to error so we can introduce other methods of saving for those if we need to.

mathieutu:
Understood, thanks. 

I've updated the PR.

danharrin (approved):

PR #18056, opened 2025-10-03
[Testing] Fix enum handling in table column state assertions

clarkewing:
## Description

Fixes a bug where `assertTableColumnStateSet()` and `assertTableColumnStateNotSet()` could throw an error when the expected or actual table column state was an enum.

Previously, the assertion message interpolated `$state` directly into a string, causing:

```log 
Error: Object of class App\Enums\SomeEnum could not be converted to string
```

This made testing the state of enum columns impossible as it systematically resulted in an error.

This PR updates the failure message to safely handle any type by using `get_debug_type()` for objects (including enums) and `var_export()` for scalars. Comparisons are unchanged; only the error message construction is affected.

This improves stability and makes failure output clearer.

## Visual changes

N/A — testing helpers only, no UI changes.

## Functional changes

### DIFF
--- a/packages/tables/src/Testing/TestsColumns.php
+++ b/packages/tables/src/Testing/TestsColumns.php
@@ -241,7 +241,7 @@ class TestsColumns
 
             $displayState = is_object($state) ? get_debug_type($state) : var_export($state, true);
 
-            Assert::assertEquals(
+            Assert::assertNotEquals(
                 $state,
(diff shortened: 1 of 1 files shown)


danharrin (changes requested):
Please revert the `sprintf` and introduce a new variable above for this

danharrin (approved):

PR #18058, opened 2025-10-03
fix: Nested relationship sorting

### LINKED ISSUE #17996: Table sorting tests with nested relationships
### Package

filament/filament

### Package Version

v4.1

### Laravel Version

v12.31.1

### Livewire Version

_No response_

### PHP Version

v8.4

### Problem description

Two-level (nested) relation fields (e.g. province.country.name) are not sortable in tests (works on table it self), while single-level relations (e.g. province.name) work correctly.

### Expected behavior

Testing nested relation fields (e.g. province.country.name) should be sortable in the same way single-level relations (e.g. province.name) are.

### Steps to reproduce

Set up a model hierarchy that goes two levels deep, e.g., Location → Province → Country.

https://github.com/CodeWithDennis/filament-nested-dot-bug

### Relevant log output

```shell

```

danharrin:
Fixes #17996.

### DIFF
--- a/packages/tables/src/Columns/Concerns/InteractsWithTableQuery.php
+++ b/packages/tables/src/Columns/Concerns/InteractsWithTableQuery.php
@@ -5,7 +5,6 @@ namespace Filament\Tables\Columns\Concerns;
 use Illuminate\Database\Connection;
 use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
 use Illuminate\Database\Eloquent\Relations\Relation;
-use Illuminate\Database\Query\Builder;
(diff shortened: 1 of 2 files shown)


alancolant:
This change introduces an issue where groups with a `null` key disappear.
It seems to be caused by the use of an `INNER JOIN` in `orderByPowerJoins`.
A `LEFT JOIN` would likely be more appropriate to preserve fallback items.

danharrin:
@alancolant can you submit a PR please?

alancolant:
> @alancolant can you submit a PR please?

I'll do

PR #18076, opened 2025-10-04
feat: add onlyAllowGroupSelection() method for group-only bulk selection

Baspa:
## Description

This PR introduces an option to restrict selection behavior in tables, allowing selection only within groups or for individual rows, rather than across the entire dataset.
This can be enabled by setting the onlyAllowGroupSelection option on the table.
 
The PR:
- Adds `onlyAllowGroupSelection()` method to HasBulkActions trait
- Add `selectsGroupOnly` property and getter method
- Update JavaScript to disable global select all when `selectsGroupOnly` is `true`
- Update Blade template to hide global select all button and page checkbox when `selectsGroupOnly` is enabled
- Maintains compatibility with existing `selectCurrentPageOnly` functionality

## Visual changes

This update modifies the selection UI so that users can only select records within a group or individually, rather than selecting all records at once.

## Functional changes


danharrin (approved):

packages/tables/resources/views/index.blade.php
```diff
@@ -606,16 +607,18 @@ class="fi-ta-selection-indicator"
...
-                                color="primary"
-                                tag="button"
-                                x-on:click="selectAllRecords"
-                                x-show="canSelectAllRecords()"
-                                {{-- Make sure the Alpine attributes get re-evaluated after a Livewire request: --}}
-                                :wire:key="$this->getId() . 'table.selection.indicator.actions.select-all.' . $allSelectableRecordsCount . '.' . $page"
-                            >
-                                {{ trans_choice('filament-tables::table.selection_indicator.actions.select_all.label', $allSelectableRecordsCount, ['count' => \Illuminate\Support\Number::format($allSelectableRecordsCount, locale: app()->getLocale())]) }}
-                            </x-filament::link>
+                            @if (! $selectsGroupsOnly)
```
nessimabadi, line 610: Hi @Baspa ,

I thinks there's an error here. When not using the selectGroupsOnly (just default select all), the value of $selectGroupOnly is `Filament\Tables\Table::selectsGroupsOnly(): bool {#4621 ▼ // vendor\filament\tables\resources\views/index.blade.php
  returnType: "bool"
  this: 
Filament\Tables
\
Table
 {#2575 …}
}` which doesn't apply the if clause and doesn't show the select all link

nessimabadi: $selectsCurrentPageOnly does returns null

danharrin: Resolved in ae6b15d46dec023c94b29077bcb22dca8d82f84e

nessimabadi: Thanks @danharrin, that was quick
Didn't let me finish debugging it hahahah

Thanks again for this amazing package

PR #18079, opened 2025-10-05
Adjust the input CSS for responsive text sizes on mobile devices

vinkla:
## Description

This PR will prevent the input fields from zooming in on iOS. [Text that is 16px or larger prevents form zoom on iOS.](https://css-tricks.com/16px-or-larger-text-prevents-ios-form-zoom/)

Please note that I was unable to build the CSS. If possible, could someone test this locally? Thanks!

## Visual changes

**Before:**

https://github.com/user-attachments/assets/62821ad7-432f-4150-9ed4-3089c06da6f6

**After:**

https://github.com/user-attachments/assets/d4dd32f1-05c7-4579-acd8-feff9add2819

## Functional changes


danharrin:
You can build the CSS to test it by running `npm i` and then `npm run build`

Can we apply this when `@supports (-webkit-touch-callout: none)` please? I really would rather avoid adding this to other devices, since it is so inconsistent with the other font sizes.

vinkla:
Updated and tested locally; it appears to work with `@supports (-webkit-touch-callout: none)`.

danharrin:
Is the same issue present in textareas? Or does Safari not zoom those? Or selects?

danharrin (approved):

PR #18084, opened 2025-10-05
Better query builder UX

danharrin:


robertvansteen:
@danharrin did this intentionally change that "Insert between blocks" is alway visible instead of being a more compact action that's only shown on hover? 
image

danharrin:
#18260

PR #18091, opened 2025-10-06
Dispatch events during record lifecycle (`Created`, `Creating`, `Saved`, `Saving`) and add related tests.

ropi-bc:
## Description

Adds events for creating, saving records in `EditRecord` and `CreateRecord` pages. Solves #18051

## Visual changes

No visual changes

## Functional changes


ropi-bc:
The unverified commit is from me; I just had a little blunder with my git config

Jubeki:
How about following the eloquent events pattern:

Creating, Created for new resources
Updating, Updated for existing resources
Saving, Saved for new and existing resources.

See https://laravel.com/docs/12.x/eloquent#events

ropi-bc:
@Jubeki so you're suggesting to remove the `Record` prefix? I don't mind it. The reason I named it `RecordCreating`, etc. is to follow the scheme of Action events (`ActionCalling`, `ActionCalled`). I find the intent is clearer this way.

Jubeki:
> @Jubeki so you're suggesting to remove the `Record` prefix? I don't mind it. The reason I named it `RecordCreating`, etc. is to follow the scheme of Action events (`ActionCalling`, `ActionCalled`). I find the intent is clearer this way.

No, what I meant is using the Suffixes according to Eloquent, so:
RecordCreating, RecordCreated for new records
RecordUpdating, RecordUpdated for existing records (currently RecordSaving and RecordSaved in your example)
RecordSaving, RecordSaved for new and existing records

ropi-bc:
How about something like this? @Jubeki ? I wanted to include `$data` too where it's appropriate. I find the result of `mutateFormDataBeforeCreate` to be valuable in a context like this.

Of course tests and docs aren't updated yet, I want to share my idea to you first

danharrin:
I think people are gonna be confused that actions on a simple resource (CreateAction / EditAction) do not call these events either.

However, for internal reasons I think it would be challenging to add support for simple resources to call "creatING and savING and updatING".

Do you have a use case for the ING events? Can I remove them please?

ropi-bc:
The idea was to follow the events an `Action` emits. I'm totally fine with removing ING events

danharrin:
I am also going to leave this undocumented, as I think in general it might confuse people to have 3 ways to hook into events (observers, page/action lifecycle hooks, and now also events). Laravel also doesn't really document internal events.

ropi-bc:
thank you, very insightful changes. what else do you need from me to move this forward?

danharrin:
@ropi-bc do my current changes work for you?

ropi-bc:
Absolutely, they do what I need them to do.

I suspect that when a new record is created in a modal, then both `ActionCalled` and `RecordCreated` are fired. Is this something we want? I don't mind if this is the case, but I want to talk about it, just in case.

danharrin:
Yeah, I think that is expected. Both things happened

ropi-bc:
Thank you both for being so proactive and helpful

PR #18099, opened 2025-10-06
Allow HTML in tooltip if user explicitly uses Htmlable interface

priithansen:
## Description

Allows HTML content in column tooltips if user explicitly uses Htmlable eg: 
`$column->tooltip(new HtmlString('<b>Example</b>'))`

## Visual changes

Usage of HTML in tooltips possible without hacking around it using extraAttributes() https://github.com/filamentphp/filament/discussions/14171

## Functional changes


packages/tables/src/Columns/CheckboxColumn.php
```diff
@@ -56,6 +57,7 @@ public function toEmbeddedHtml(): string
                     ? '{
                         content: ' . Js::from($tooltip) . ',
                         theme: $store.theme,
+                        allowHTML: ' . ($tooltip instanceof Htmlable ? 'true' : 'false') . ',
```
danharrin:
```suggestion
                        allowHTML: ' . Js::from($tooltip instanceof Htmlable) . ',
```

danharrin (changes requested):

packages/tables/src/Columns/CheckboxColumn.php
priithansen: Of course, updated all instances.

stefket:
This is a great improvement!
It would be great if this feature could also be made available in other places where tooltips are supported, such as Actions and Infolists, for consistency.

priithansen:
Makes sense yes! I will try to go over all the places today or tomorrow and update the PR.

priithansen:
All places where users can directly create tooltips should now be covered.

Although there is now a possiblity of breaking change with type mismatch. Not sure how this could be better handled.
```PHP
  Type of SomeCustomAction::$tooltip must be Illuminate\Contracts\Support\Htmlable|Closure|string|null (as in class Filament\Actions\Action)
```

Example from my own code where this would now need `| Htmlable |` added to work:
```PHP
class SomeCustomAction extends Action
{
    protected string | Closure | null $tooltip = 'Fancy typed tooltip';

    ...
```

danharrin (approved):

PR #18100, opened 2025-10-06
[4.x] Create RichContentFaker class

awcodes:
## Description

Adds a `RichContentFaker` class to allow for easier population of rich editor content in factories and tests.

## Functional changes

### DIFF
--- a/packages/forms/src/Components/RichEditor/RichContentFaker.php
+++ b/packages/forms/src/Components/RichEditor/RichContentFaker.php
@@ -1,0 +1,212 @@
+<?php
+
+namespace Filament\Forms\Components\RichEditor;
+
+use Faker\Factory;
+use Faker\Generator;
+use Illuminate\Support\Str;
+
+class RichContentFaker
+{
+    protected Generator $faker;
+
+    protected string $output = '';
+
+    protected ?RichContentRenderer $renderer = null;
+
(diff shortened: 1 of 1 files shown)


danharrin (changes requested):
Can we have some snapshot tests for the different options please?

awcodes:
Do you want it tested against the renderer output or the content in the editor in the browser?

danharrin:
Not sure what you mean? Tested against the output of the PHP function, no need for a browser test

awcodes:
Maybe I'm just not clear on what you mean by "snapshot" test.

I'm my mind snapshot implies browser. But as long as the RichContentRender outputs the correct html or json/array I think that satisfies the test. 

Am I wrong?

danharrin:
It's not a screenshot, it stores text: https://pestphp.com/docs/snapshot-testing

awcodes:
I'm not sure snapshot testing is going to work with this since each time the faker is called it would be different text, like in paragraphs which are using the $faker->paragraphs(). 

Wouldn't that cause the snapshot to fail every time?

danharrin:
Yeah thats true, I think we still need some sort of test

bambamboole:
You can seed faker so that it always creates the same fake data @awcodes

danharrin:

PR #18111, opened 2025-10-07
fix: restore nullable grouping behavior in table grouping

### LINKED ISSUE #18105: [4.x] Table defaultGroup fallback
### Package

filament/tables

### Package Version

v4.1.2

### Laravel Version

v12

### Livewire Version

_No response_

### PHP Version

PHP 8.4

### Problem description

I have a table for a resource defined as 
```
return $table
            ->columns([
              ......
            ])
            ->defaultGroup('group_hash')
```

However, not all my rows have a defined 'group_hash'. This leads to an error on the page where it is end of grouped products.

`Filament\Resources\Pages\ListRecords::getGroupedSelectableTableRecordKeys(): Argument #1 ($group) must be of type string, null given, called in /var/www/html/storage/framework/views/5505534bae45483ccd793c43b2e4g5c.php on line 2107`

### Expected behavior

Ideally in the method we can have a fallback in getGroupedSelectableTableRecordKeys to group rest of items and also
make the `public function getGroupedSelectableTableRecordKeys(string $group): array` `$group` argument optional.

### Steps to reproduce

Have a resource with the column set for defaultGroup be null for some fields and go through the pages. 

https://github.com/TonsiTT/r

### Relevant log output

```shell
Filament\Resources\Pages\ListRecords::getGroupedSelectableTableRecordKeys(): Argument #1 ($group) must be of type string, null given, called in /var/www/html/storage/framework/views/5505534bae45483ccd793c43b2e4g5c.php on line 2107
```

alancolant:
This fix restores the expected behavior for nullable  in table grouping.
Previously, the grouping relied on an `INNER JOIN`, which excluded rows with `NULL` values.
It now uses a `LEFT JOIN` to correctly include groups with nullable keys.

Previous: #18058
Also fix: #18105 allowing `getGroupedSelectableTableRecordKeys` and `scopeQueryByKey` to have `?string $key`

### DIFF
--- a/packages/tables/src/Concerns/HasBulkActions.php
+++ b/packages/tables/src/Concerns/HasBulkActions.php
@@ -133,7 +133,7 @@ trait HasBulkActions
     /**
      * @return array<string>
      */
-    public function getGroupedSelectableTableRecordKeys(string $group): array
+    public function getGroupedSelectableTableRecordKeys(?string $group): array
     {
         $query = $this->getFilteredTableQuery();
 
--- a/packages/tables/src/Grouping/Group.php
+++ b/packages/tables/src/Grouping/Group.php
@@ -350,7 +350,7 @@ class Group extends Component
         return $query;
     }
 
-    public function scopeQueryByKey(EloquentBuilder $query, string $key): EloquentBuilder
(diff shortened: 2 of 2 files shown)


danharrin:
Sorry, I also started working on this here #18113. My fix also sorts normal table column sorting & null relationships being selectable in groups.

PR #18123, opened 2025-10-07
fix: Charts with max height set should not have an aspect ratio

danharrin:


moskoweb:
The height limitation still doesn't work properly, it gets huge or gets small.

PR #18131, opened 2025-10-08
Feat/crud quietly

mikeslinkman:
## Description

Adds the option to quietly perform the following actions:

- Create
- Update
- Delete
- Replicate

Maybe we should consider adding it to the `bulkActions` as well, will do that if required

## Visual changes

N/A

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/CanQuietlyProcess.php
+++ b/packages/actions/src/Concerns/CanQuietlyProcess.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace Filament\Actions\Concerns;
+
+use Closure;
+
+trait CanQuietlyProcess
+{
+    protected ?Closure $quietly = null;
+
+    /**
+     * Set a condition to determine if a operation should be quiet
+     * @param Closure(static): bool $quietly
+     */
+    public function quietly(Closure $quietly): static
+    {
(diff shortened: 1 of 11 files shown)


packages/actions/src/Concerns/CanQuietlyProcess.php
```diff
@@ -0,0 +1,30 @@
...
+    {
+        if (! $this->quietly) {
+            return false;
+        }
+
+        if (is_bool($this->quietly)) {
+            return $this->quietly;
+        }
+
+        return $this->evaluate($this->quietly);
```
wychoong, line 24: Just this return would do

mikeslinkman: I had an error, but looking at code quality action I think I misread the error and incorrectly "fixed" it

danharrin:
Sorry, I know I asked for this but I think I'm going to decline for now. I don't think that I've had enough requests for this in the past for the feature to have so much footprint on the codebase, and I wouldn't want to add it for just one action. I also think you can accomplish this with just one line already:

```php
ReplicateAction::make()
    ->using(fn (Model $record) => $record->replicate()->saveQuietly())
```

PR #18139, opened 2025-10-09
Fix: URL-encode background parameter in UiAvatarsProvider

MACscr:
## Description

The `background` parameter in `UiAvatarsProvider` was not being URL-encoded, which causes issues when the background color contains spaces (e.g., OKLCH color values like `oklch(0.13 0.028 261.692)`).

When URLs with spaces are passed to `ImageColumn::getImageUrl()`, the `filter_var(FILTER_VALIDATE_URL)` validation fails, causing avatar URLs to be incorrectly treated as relative filesystem paths instead of external URLs. This results in avatars not displaying.

This PR adds `urlencode()` to the background parameter, matching the approach already used for the `name` parameter.

## Visual changes

N/A - This fixes existing functionality where avatars fail to display when using OKLCH colors.

## Functional changes


danharrin (approved):
I've never seen it fail so thats weird, not a problem though

PR #18141, opened 2025-10-09
[4.x.] fix: Notification background dark mode

pxlrbt:
## Description

When using `->color('primary')` on notifications, they show with a transparent background in dark mode.

## Visual changes

No `->color()`, solid background:
image

`->color('primary')`, transparent background:
image

fix:
image

## Functional changes


danharrin:
I think we might actually need a container around the element that applies a solid dark color, which we can then overlay the transparent color on. I think that is the bug that needs to be fixed, I might have removed that by accident when upgrading v3 to v4 styles.

pxlrbt:
You don't use a transparent color in light mode thought. So you want to change both? Does it really make a difference?

danharrin:
No, the container will be white in light mode. Yes, the color makes a difference, its more vibrant when the shade is lighter and the background is transparent.

NathanLochala:
@danharrin - Do you feel this is something you are ready to approve. I would really love for this bug to be fixed. Thank you!

danharrin:
No, it needs the adjustment that I mentioned

NathanLochala:
@pxlrbt - Is there any way we can get the PR adjusted so we can get this bug taken care of?

pxlrbt:
So in white they are pretty similar. It only makes a difference in black.

1. `white`
2. `white + `color-400/10`
3. `color-50`

image

1. `gray-950`
2. `gray-950 + `color-400/10`
3. `color-950`

image

This is what is looks with `black` as a base:

image

pxlrbt:
I adjusted the PR to add a container around the actual notification. I'd rather add something inside the notification but I don't want to change too much of the HTML structure

packages/notifications/resources/css/notification.css
```diff
@@ -46,7 +50,7 @@
     }
 
     &.fi-color {
-        @apply bg-color-50 dark:bg-color-400/10;
```
danharrin, line 49: I think this change should be reverted to return to v3 behaviour

packages/notifications/resources/css/notification.css
```diff
@@ -1,3 +1,7 @@
+.fi-no-notification-container:has(.fi-inline) {
+    @apply bg-white dark:black rounded-xl;
```
danharrin: Just checked in v3, it was `dark:bg-gray-900`

Please also run `cs` and rebuild the CSS

danharrin (changes requested):

danharrin:
I think we should actually return `fi-no-notification` as the outer element with an opaque white/gray BG, and then introduce `fi-no-notification-bg` as an inner element that provides the color.

Otherwise I think there is too much other CSS and JS that needs adjusting to account for `fi-no-notification` not being the outer element

pxlrbt:
I have another idea: We can just use the `color-mix()` function that's used for opacity anyway and combine it with the underlying color. No wrapper needed. I'll check this out this evening.

danharrin:
Good idea!

pxlrbt:
It's using `color-mix(in oklab)` now. The color is slightly different, but should be close enough. Color picker for dark mode says:

```
rgba(30.001, 38, 50)
rgba(30.001, 36, 46)
```


image

image

packages/notifications/resources/css/notification.css
```diff
@@ -46,7 +46,15 @@
     }
 
     &.fi-color {
-        @apply bg-color-50 dark:bg-color-400/10;
+        background-color: color-mix(in oklab, white 90%, var(--color-400));
+
+        &:where(.dark *) {
```
danharrin: Does this work?

```suggestion
        @variant dark {
```

danharrin (changes requested):

packages/notifications/resources/css/notification.css
pxlrbt: Looks good.

danharrin (approved):

PR #18148, opened 2025-10-09
propagate extraInputAttributes to searchable Select Butons

grafst:
## Description

When I want to omit a field from tabbing in a filament form i can add 
`->extraInputAttributes(['tabindex' => '-1'])`
so far so good.
However this doesn't work on a select with the `->searchable()` option, because that is not composed in the select.blade.php but rather in a javascript utility.

## Visual changes

tabindex

## Functional changes

We propagate the extra input Attributes via the selectFormComponent into the select utility and set them on the input and also on the clear button.

- [x] Documentation is up-to-date, since this would be expected behaviour.


danharrin (changes requested):
Some tests are failing on L11

Please ensure that classes are merged on the element, not replaced

grafst:
We merged the id attribute into the div. the other attributes (autofocus, disabled) can't be merged because they are not attributes on the searchable select but rather are handled by the select.js utility. 
Sorry if im missing or misunderstanding something.
please advise :)

danharrin:
That isn't what I meant, we don't need `->merge(['id'=>$id], escape: false))`

If a `class` attribute is passed, it should merge with the other classes on the element that are already there instead of replacing them. The merging needs to be done in JS

grafst:
Sorry, was on Hollyday
Yes, will look into merging it in the js file.

Wouldn't the merging of the id also be nice, to make the select behave similar to the other input fields and for example allow styling with css? Maybe in a different PR?

danharrin:
We shouldn't merge in the ID, since that is already present elsewhere in the component and IDs cannot be duplicate

danharrin:
Hi @grafst, are you able to make the required changes please?

PR #18049, opened 2025-10-03
Fix/rich editor plugins merge

bmspereira-07:
## Description

Fixes the rich editor plugin merging. If the user has custom plugins that will conflict with initial ones, it will replace them

## Functional changes

- [ ] Code style has been fixed by running the `composer cs` command.
- [ ] Changes have been tested to not break existing functionality.
- [ ] Documentation is up-to-date.

PR #18156, opened 2025-10-10
Extra customization for placeholder

thijskuilman:
## Description

> [!IMPORTANT]  
> **Why not use custom extension for this?**
> The [Fix/rich editor plugins merge](https://github.com/filamentphp/filament/pull/18049) allows you to overwrite default extensions. However, the Placeholder extension cannot be overwritten in this case because it’s already configured in `extensions.js`. Once an extension has been configured, it cannot be overwritten.
>
> For more details, see [ueberdosis/tiptap/issues/1044](https://github.com/ueberdosis/tiptap/issues/1044)

### Node specific placeholders
You can now set placeholders for specific node types. The code implementation is based on the official example, [see the code example here](https://tiptap.dev/docs/editor/extensions/functionality/placeholder). I have updated the documentation accordingly.

### Show only current placeholder
I have added the [showOnlyCurrent property](https://tiptap.dev/docs/editor/extensions/functionality/placeholder#showonlycurrent) to RichEditor and updated the documentation. This allows you to show multiple placeholders at the same time.

## Visual changes

In the example below, I do the following:

- applied a different placeholder for the `heading` node
- set `showOnlyCurrent` to `false`, so all placeholders are visible at once

```php
RichEditor::make('content')
  ->placeholder('Start typing...')
  ->nodePlaceholders([
      'heading' => 'Enter a title',
  ])
  ->showOnlyCurrentPlaceholder(false)
```

image

## Functional changes


thijskuilman:
Hi @danharrin ! I submitted a similar PR last week and [you pointed me towards the new extension merging mechanic](https://github.com/filamentphp/filament/pull/18031#issuecomment-3368187717). That approach worked great overall—thank you for the tip! However, I ran into an issue specifically when trying to extend the placeholder node (see the note in my PR description for details).

I’ve decided to resubmit just the placeholder customization portion from my previous PR. I’ve removed the content schema logic to keep the scope smaller and the impact lower. Hopefully this version is a bit easier to consider!

danharrin (changes requested):
Please simplify this to just `extraPlaceholderNodeOptions()`, which accepts an array or closure, doesnt transform in any way, and directly merges into the object. Then the user is free to implement whichever options they want, instead of being constrainted to the existing options, and also reduces the size of the PR.

danharrin:
Alternatively, is there a way to prevent the placeholder extension from being registered at all if an alternative placeholder extension is injected via a URL? I think this is a bug in Filament rather than TipTap, we can prevent the extension from being registered in the first place if it is overwritten.

If we can fix this, we don't need any new PHP methods

thijskuilman:
@danharrin 

> Alternatively, is there a way to prevent the placeholder extension from being registered at all if an alternative placeholder extension is injected via a URL? I think this is a bug in Filament rather than TipTap, we can prevent the extension from being registered in the first place if it is overwritten.

I agree, that would be an even better solution. Next week, I’ll investigate why this is happening and see if I can find a fix. If I’m unable to, I’ll add an `extraPlaceholderNodeOptions()` method.

I’ve set the PR to draft for now and will mark it ready once it’s good to go.

thijskuilman:
@danharrin After some hours of digging, I was finally able to pinpoint the problem with custom placeholders. I have created a [bug fix PR](https://github.com/filamentphp/filament/pull/18205). With this, we do not need any additional PHP methods anymore, so I am closing this PR.

PR #18166, opened 2025-10-11
fix: fix Italian translation in actions/delete.php

claudiocenghialta:
## Description

Fix Italian translations for actions/delete

## Visual changes

No visual changes

## Functional changes

No functional changes

- [x] Code style has been fixed by running the `composer cs` command.
- [x] Changes have been tested to not break existing functionality.
- [x] Documentation is up-to-date.

PR #18167, opened 2025-10-11
review missing translation for form package

danielebarbaro:
## Description

Review the missing Italian translation folder for the form package.

## Visual changes

No Visual changes :)

## Functional changes

### DIFF
--- a/packages/forms/resources/lang/it/components.php
+++ b/packages/forms/resources/lang/it/components.php
@@ -11,6 +11,7 @@ return [
             ],
 
             'add' => [
+
                 'label' => 'Aggiungi a :label',
 
                 'modal' => [
@@ -26,9 +27,11 @@ return [
                     ],
 
                 ],
+
             ],
 
             'add_between' => [
+
                 'label' => 'Inserisci tra i blocchi',
 
                 'modal' => [
@@ -44,6 +47,7 @@ return [
                     ],
 
                 ],
+
             ],
 
             'delete' => [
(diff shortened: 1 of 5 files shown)


danielebarbaro:
Can you kindly add the Hacktoberfest tag?

danharrin:
Can you please only make one PR for Italian translations instead of one for each package? Thanks

Let me know if this is the final one

danielebarbaro:
Yes, sure @danharrin. I’ll wait for this https://github.com/filamentphp/filament/pull/18166 to be merged, and then I’ll check if there are any missing keys.

danielebarbaro:
FYI: there are some issues here https://github.com/filamentphp/filament/blob/4.x/packages/panels/resources/lang/en/auth/multi-factor/recovery-codes-modal-content.php . The keys look wrong. 🙂

danharrin:
The keys are correct

danielebarbaro:
> The keys are correct

The keys are not in lowercase and snake case, so we have this:

Screenshot 2025-10-11 at 15 40 25

if we use `php artisan filament:check-translations **locale**`

Can we at least use something like :
```php

return [

    'actions' => [

        'Click to' => 'Click to',

        'copy' => [
            'label' => 'copy',
        ],

        'or' => 'or',

        'download' => [
            'label' => 'download',
        ],

        'all the codes at once.' => 'all the codes at once.',

    ],

    'messages' => [
        'copied' => 'Copied',
    ],

];

```

danharrin:
It is deliberate that they are numeric, it is a multi-part translation string and not all languages need all parts, so it doesn't make sense to give each a key since it doesn't translate directly.

danielebarbaro:
> It is deliberate that they are numeric, it is a multi-part translation string and not all languages need all parts, so it doesn't make sense to give each a key since it doesn't translate directly.

Wow, thanks for the kind reply! It’s clear to me now, and I’ve added a new commit.

danielebarbaro:
ok @danharrin the branch is now updated and ready for your review

danharrin (approved):

PR #18169, opened 2025-10-11
fix: Disable rich editor file attachments

### LINKED ISSUE #18157: RichEditor v4: files still upload via drag-and-drop / paste when attachFiles toolbar button is removed
### Package

filament/filament

### Package Version

v4.1.6

### Laravel Version

v12.33.0

### Livewire Version

v3.6.4

### PHP Version

PHP 8.4.13

### Problem description

When the RichEditor toolbar is configured without the attachFiles button, users can still upload files by dragging & dropping into the editor or pasting from the clipboard. This bypasses the toolbar configuration and results in unintended uploads/attachments being created.

### Expected behavior

If attachFiles is not present in toolbarButtons, then all file-attachment behaviors should be disabled (including drag-and-drop and paste). At minimum there should be an explicit API to disable the attachment extension entirely.

### Steps to reproduce

1. Create a form with a RichEditor that does not include attachFiles in toolbarButtons:

```
use Filament\Forms\Components\RichEditor;

RichEditor::make('content')
    ->toolbarButtons([
        'bold', 'italic', 'strike',
        'h2', 'h3',
        'bulletList', 'orderedList',
        'link', 'blockquote', 'codeBlock',
        'undo', 'redo',
    ]);
```

2. Open the form.

3. Drag & drop a PNG/JPG into the editor or paste an image from the clipboard (Ctrl/Cmd + V).

4. The file uploads and an attachment node is inserted, despite no attachFiles button.

**or from my repository:**

Open Post resource > Create - the form contains a RichEditor without attachFiles in toolbarButtons.

https://github.com/radimnemecek/filament-richeditor-attachment-bypass

### Relevant log output

```shell

```

danharrin:
Fixes #18157.


thijskuilman:
@danharrin In my project, I use a custom `ImagePlugin` that extends the editor’s image upload functionality. This means I use my own `attachImage` tool instead of the default `attachFiles` tool. However, after this PR change, uploads no longer work, since file uploads are now only enabled when `attachFiles` is explicitly present.

Would it be possible to make this check a bit more flexible to support custom file upload tools as well?

danharrin:
I'd accept a PR for a new method to `HasFileAttachments` like `fileAttachments()` which accepts a boolean OR NULL `$condition` and saves it to `$hasFileAttachments` (`null` default). Then the `hasFileAttachments()` method returns the boolean, or if it is null it returns `hasFileAttachmentsDefault()`. `hasFileAttachmentsDefault()` returns `true`. `hasFileAttachmentsDefault()` is overridden in `RichEditor` and `MarkdownEditor` to return `$this->hasToolbarButton('attachFiles')`. Then the Blade views for those components can be updated to use `$hasFileAttachments()` instead of `$hasToolbarButton('attachFiles')`

Issue #18172, opened 2025-10-11
[4.x] Sticky form actions exceed panel width


Saracaen:
### Package

filament/filament

### Package Version

v4.1.6

### Laravel Version

v12.33.0

### PHP Version

PHP 8.4.0

### Problem description

Sticky actions exceed the width of the panel content, and even the width of the panel if they become active on an EditRecord or CreateRecord page.

### Expected behavior

I expect the actions to become sticky, and to not exceed the width of the panel. Ideally, they should not exceed the width of the page content.

### Steps to reproduce

Run the reproduction repository and create or edit a customer. The '$formActionsAreSticky' is set to true for both the CreateCustomer and EditCustomer pages, and the Customer model has enough fields to overflow pretty much every screen.

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/Saracaen/filament-sticky-bug

### Relevant log output

```shell

```

danharrin:
I checked your repo and this is the intended design of the feature currently. I made sure that they do not cause the page to scroll horizontally, they just span the full width of the page.

I would accept a PR containing a new design suggestion if you would like to work on one.

Saracaen:
The main issue with this is that it breaks when you change the action alignment to end. The actions will then appear somewhere off the screen. I'm not sure of a better approach, as yet, since sticky does not work and the correct panel content width cannot yet reliably be established. I suspect some JavaScript could go a long way to resolving it, but I would need to look into that further.

PR #18176, opened 2025-10-12
Improves sticky schema actions combined with end aligned actions

Saracaen:
## Description

Improves behavior for #18172 

## Visual changes

**Before**
{7AF8AA60-A045-446B-B932-CBE5464B13FD}

**After**
{0BF05EBB-E30D-4FCD-B21C-3DCA982BC2F3}

## Functional changes

### DIFF
--- a/packages/schemas/resources/css/components/actions.css
+++ b/packages/schemas/resources/css/components/actions.css
@@ -11,12 +11,12 @@
 
     &.fi-sticky {
         & .fi-ac {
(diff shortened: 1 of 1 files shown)


danharrin:
Sorry, this is too breaking compared the original behaviour. If we change this, the bar needs to be full width.

danharrin:
If you want them to look like this I would suggest a custom theme

Saracaen:
I don't want them to look like this necessarily, but I'm afraid that the only reliable alternative to fix it on all layout sizes is to use JavaScript to fetch the parent container width and apply that to the sticky action container.

danharrin:
Let's do that then, as long as it also updates when you resize the window

PR #18184, opened 2025-10-12
Feature/navigation child items

webard:
## Description

This PR allows to define children navigation items related to resource. It allows to reach one more level of navigation.

Feature tests are not provided, because children navigation was available before in the Filament, but seems to be unintentionally omitted to configure.

Child navigation items was broken when displayed in sub navigation, what is fixed by this PR.

## Visual changes

Before:

https://github.com/user-attachments/assets/0601a468-3b34-43e4-9068-2f80c4ba0c69

After:

https://github.com/user-attachments/assets/40a1f0ab-bc38-4c36-8954-a0a1be3a463f

## Functional changes


danharrin:
The feature exists and is documented as "parent items" - https://filamentphp.com/docs/4.x/navigation/overview#grouping-navigation-items-under-other-items

I would accept a PR to fix the UI if it is broken as in your screenshots, but we don't need another way to configure it

webard:
@danharrin thank you for feedback. However, using "parent items" functionality, you can only add Resources as child navigation items, you cannot add custom links, pages, etc.

danharrin:
I think you can register custom navigation items in the panel provider, and set their parent item there?

webard:
Yes, of course, but in my opinion it's little more complicated and less intuitive than the method in Resource class, especially when you can set many other things regarding navigation in Resource class.

danharrin:
Agreed, but I think introducing another API for it just causes more confusion about what developers should use. So not going to proceed with it.

webard:
Understood, thank you.

PR #18193, opened 2025-10-13
fix undefined isSticky

howdu:
Fix js error when sticky is not enabled:
> Alpine Expression Error: isSticky is not defined
> Expression: "isSticky ? `width: ${width}px;` : ''"

## Functional changes


packages/schemas/resources/views/components/actions.blade.php
danharrin, line 1: Please save the result of `$isSticky()` at the top of the file so we dont run it twice

danharrin (changes requested):

danharrin (approved):

PR #18195, opened 2025-10-13
[4.x] Mounted action modal test helpers

### LINKED ISSUE #17857: [4.x] Action modal content not found in tests
### Package

filament/filament

### Package Version

v4.0.17

### Laravel Version

v12.30.0

### Livewire Version

v3.6.4

### PHP Version

PHP 8.4.12

### Problem description

When mounting an action on a form schema - either in the header, or as part of the form itself, the content of the modals cannot be seen.  The following examples show that the action exists, and can be successfully mounted, but the modal content itself cannot be confirmed.

After some investigation on my part it appears that the issue is likely related to rendering of the modal's contents as the tests provided ensure the modal content is accurate, and the action itself can be mounted, but the contents are never found in the HTML.

### Expected behavior

The expectation would be that the content of the modals would be rendered in the HTML, and the contents could be verified.

On the EditCustomer resource page:
```
{
    return [
        DeleteAction::make()->modalContent(new HtmlString('Header modal content')),
    ];
}
```

A test like this should pas:

```
it('cannot delete customer without confirmation', function () {
    // Create a customer
    $customer = Customer::factory()->create([
        'name' => 'Non-Deletable Customer',
    ]);

    // Try to delete the customer
    livewire(EditCustomer::class, [
        'record' => $customer->id,
    ])
        ->assertHasNoFormErrors()
        ->mountAction(DeleteAction::class)
        ->assertActionMounted(DeleteAction::class)
        ->assertSee('Header modal content');
});
```

But it fails to find the string `Header modal content` in the html.

### Steps to reproduce

Download the repo and `composer install`, then run: 
`./vendor/bin/pest --filter='cannot delete customer without confirmation'`

This will run through 4 example tests

https://github.com/Mrkbingham/filament-header-action-test-issue

### Relevant log output

```shell
FAILED  Tests\Feature\CustomerResourceTest > it cannot delete customer without confirmation - action class                                                                                                                 
  Expected: <div wire:snapshot="" wire:effects="" wire:id="Dusb7fcaQ80QoBaSdPmc"\n
      class="fi-page fi-resource-edit-record-page fi-resource-customers fi-resource-record-1"\n
  >\n
  ... (475 more lines)

  To contain: Header modal content

  at vendor/livewire/livewire/src/Features/SupportTesting/MakesAssertions.php:14
     10▕ {
     11▕     function assertSee($values, $escape = true, $stripInitialData = true)
     12▕     {
     13▕         foreach (Arr::wrap($values) as $value) {
  ➜  14▕             PHPUnit::assertStringContainsString(
     15▕                 $escape ? e($value): $value,
     16▕                 $this->html($stripInitialData)
     17▕             );
     18▕         }

      +1 vendor frames 
  2   tests/Feature/CustomerResourceTest.php:32
```

pxlrbt:
## Description

More assertion methods for testing modal content: #17857

## Functional changes

### DIFF
--- a/packages/actions/.stubs.php
+++ b/packages/actions/.stubs.php
@@ -86,13 +86,6 @@ namespace Livewire\Features\SupportTesting {
         public function parseNestedActionName(string | array $name): array {}
 
         public function parseNestedActions(string | TestAction | array $actions, array $arguments = [], bool $areRelativeToMountedActions = true): array {}
-
-        /**
-         * Helper method to get the modal content from partials.
-         *
-         * @internal
-         */
-        public function getMountedActionModalHtml(): string {}
     }
 
 }
--- a/packages/actions/.stubs.php
+++ b/packages/actions/.stubs.php
@@ -61,6 +61,14 @@ namespace Livewire\Features\SupportTesting {
 
         public function assertActionNotMounted(string | TestAction | array $actions = []): static {}
 
+        public function assertMountedActionModalSee(string | array $values, bool $escape = true): static {}
+
+        public function assertMountedActionModalDontSee(string | array $values, bool $escape = true): static {}
+
+        public function assertMountedActionModalSeeHtml(string | array $values): static {}
+
+        public function assertMountedActionModalDontSeeHtml(string | array $values): static {}
+
         public function assertActionHalted(string | TestAction | array $actions = []): static {}
 
         /**
@@ -78,6 +86,13 @@ namespace Livewire\Features\SupportTesting {
         public function parseNestedActionName(string | array $name): array {}
 
         public function parseNestedActions(string | TestAction | array $actions, array $arguments = [], bool $areRelativeToMountedActions = true): array {}
+
+        /**
+         * Helper method to get the modal content from partials.
+         *
+         * @internal
+         */
+        public function getMountedActionModalHtml(): string {}
     }
 
 }
--- a/docs/10-testing/05-testing-actions.md
+++ b/docs/10-testing/05-testing-actions.md
@@ -225,7 +225,7 @@ it('can send invoices to the primary contact by default', function () {
 
 ## Testing the content of an action modal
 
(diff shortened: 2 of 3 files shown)


danharrin (changes requested):
Currently, CI is failing

pxlrbt:
@danharrin PhpStan does not understand that those are macros and I used a helper method so I don't have to repeat getting the partial HTML. I can extract it into a helper method and place it over the class or inline and repeat the code. WDYT?

danharrin:
I think the methods need adding to the testing stubs file like the others. After that, we can add ignores where we are calling macroed functions.

packages/actions/.stubs.php
```diff
@@ -78,6 +86,13 @@ public function parseActionName(string | TestAction | array $actions): string {}
         public function parseNestedActionName(string | array $name): array {}
 
         public function parseNestedActions(string | TestAction | array $actions, array $arguments = [], bool $areRelativeToMountedActions = true): array {}
+
+        /**
+         * Helper method to get the modal content from partials.
+         *
+         * @internal
+         */
+        public function getMountedActionModalHtml(): string {}
```
danharrin: Do we need to include this in the stubs if we ignore all its usages anyway?

danharrin (changes requested):
If there is a request while the the modal is open, for example, if a live field is changed, I think the partial name becomes `action-modal.NESTING-INDEX` or something similar. Probably needs checking. `action-modals` is only used on a request where a modal is opened or closed

packages/actions/.stubs.php
pxlrbt: Honestly, I have no idea how this stub file is used. I thought it's a file for PhpStan (https://phpstan.org/user-guide/stub-files) but I don't find it in the PhpStan config.

pxlrbt:
You are right. I adjusted the code to use either the general key or one with a nesting index. Is there a scenario where multiple nested actions are rendered at once?

danharrin:
I don't think multiple actions are ever rendered at once, only the currently active one

danharrin (approved):

PR #18196, opened 2025-10-13
Fix default zoom on rich editor on iOS Safari

Saracaen:
## Description

The regular inputs are guarded against this by a font-size change via @supports (-webkit-touch-callout: none), but the rich editor was not. This brings the rich editor in line with the other inputs.

## Visual changes

**Before**

image

**After**

image

## Functional changes


danharrin (changes requested):
Please build the CSS and commit it

Just checking that you have tested this change locally too?

Saracaen:
Yes I tested this locally by using identical SASS selectors in my custom Filament theme. Pushed the compiled CSS.

Saracaen:
@danharrin I'm not 100% sure the compiled CSS is correct, or at least I would have also expected the dist index.css in the forms package to change as well. I had to do some tricks to get it to compile on Windows at all, because "build:forms-css" uses cp and rm, which are not supported in Windows. I fixed this locally by using shx, which should not change how the command behaves.

danharrin (approved):

PR #18202, opened 2025-10-14
Fix Action schema doesn't initialize when action is not sticky

aligulzar729:
## Description
The x-data is not initialized when action is not sticky and it throws error in console which makes smoke tests fail.
Good thing I had an PEST smoke test that detected it 

## Visual changes
Screenshot 2025-10-14 at 13 45 22

## Functional changes
moved the x-data initialization outside of $isSticky check

### DIFF
--- a/packages/schemas/resources/views/components/actions.blade.php
+++ b/packages/schemas/resources/views/components/actions.blade.php
@@ -13,8 +13,8 @@
 @endphp
 
 <div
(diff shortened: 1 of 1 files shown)


danharrin:
This was already fixed in another PR, and we dont need to init if its not sticky

aligulzar729:
but I see the error in my console

danharrin:
The fix isnt released yet

PR #18208, opened 2025-10-14
fix(mfa): fix notification on email MFA resend

kvnyhns:
## Description

Fixes a UX issue where the email authentication resend action always displayed a success notification, even when blocked by rate limiting.

### The Problem
When users attempted to resend their email authentication code more than twice within a minute, the rate limiter prevented the email from being sent, but the UI still showed "We've sent you a new code by email."

### The Solution
Modified `sendCode()` to return a boolean indicating success/failure. Resend actions now check this return value and display the appropriate notification:

- **Success** (not rate limited): "We've sent you a new code by email"
- **Rate limited**: "Too many resend attempts. Please wait before requesting another code."

Applied consistently across all three email MFA flows: login challenge, setup, and disable.

### Changes
- Modified `EmailAuthentication::sendCode()` to return `bool` instead of `void`
- Updated resend actions to check return value before showing notifications
- Added translation keys for rate-limited warnings
- Enhanced tests to verify UI notifications match actual behavior

## Functional changes

- [x] Code style has been fixed by running the `composer cs` command.
- [x] Changes have been tested to not break existing functionality.
- [x] Documentation is up-to-date.

PR #18211, opened 2025-10-14
refactor(mfa): DRY - Extract resend action logic into reusable method (builds on #18208)

imhayatunnabi:
Eliminates code duplication by extracting resend action logic into a reusable method.

Builds on #18208


danharrin:
I'm going to hold off doing this for now, I don't think there is enough duplication to warrant the refactor yet.

PR #18212, opened 2025-10-14
fix: make the code-editor component support live onBlur and debounce

### LINKED ISSUE #18207: CodeEditor Form Element does not support reactive on blur ->live(onBlur: true)
### Package

filament/filament

### Package Version

v4.0.6

### Laravel Version

v12.27.1

### Livewire Version

_No response_

### PHP Version

8.4

### Problem description

I have a `->live(onBlur: true)` on the CodeEditor element.
As soon as I start typing in the element, a request is sent.

I have the CodeEditor on the left and a kind of LivePreview on the right that renders Blade.
So when I start typing `{{fake()}}`, an exception is thrown immediately cause f is not a constant and fa and fak and so on.

### Expected behavior

When I set live onBlur, it should only update after I click outside the field.

This also prevents my error message saying that the constants do not exist because it only tries to render after I have finished typing.

### Steps to reproduce

Create a Form with:
```php
                            CodeEditor::make('html')
                                ->label(__('message.field.html'))
                                ->default(null)
//                                ->language(CodeEditor\Enums\Language::Php)
                                ->live(onBlur: true)
```

And than I have a custom element:
```php
                            MessagePreview::make('__preview')
                                ->label(__('message.field.preview')),
```php
<x-dynamic-component
    :component="$getFieldWrapperView()"
    :field="$field"
>
    <div
        x-data="{ state: $wire.$entangle(@js($getStatePath())).defer }"
        {{ $getExtraAttributeBag() }}>
        <input {{ $applyStateBindingModifiers('wire:model.blur') }}="{{ $getStatePath() }}" />
        <x-message-live-preview
            @viteReactRefresh
            :message="$record"
            :content="$get('html')"
            :subject="$get('subject')"
            :plainText="$get('plain_text')"
        />
    </div>
</x-dynamic-component>
```

Soo

Login, goto messages page. Create one and just type in the left field.
You will see it updates instantly.

Type something like `{{af}}` and it will throw imerdiently.

`->live(onBlur: true)` does not work
https://github.com/LvckyAPI/filament-debug/blob/6b285902906cc8c57efcd491451a7bf0caeeccce/laravel/app/Filament/Resources/Messages/Schemas/MessageForm.php#L16

https://github.com/LvckyAPI/filament-debug

### Relevant log output

```shell

```

People-Sea:
## Description
fixes:  #18207

## Visual changes

# Before

https://github.com/user-attachments/assets/776a19e6-18ea-4308-b609-cee5ef71e17b

# After

https://github.com/user-attachments/assets/d0e8098d-dcbc-4a39-ba47-4b62b38bba04

## Functional changes


danharrin (changes requested):
I would suggest looking at the rich editor or markdown editor for inspiration. We use the `isOptimisticallyLive: false` option for those, which means that debounce/blur settings do not add `.live` to the state `$entangle()`, so its much easier to handle and prevent the state updates.

I would suggest using that, since then you can immediately update the `state` property, and then manually call `this.$wire.commit()` when you want, which ends up being less code I think.

People-Sea:
I still need to keep the comparison of whether it's really updated when onBlur. This forces me to use one more attribute. Do you think it's necessary to keep it or just remove it?

danharrin:
The issue with preventing the state from being set immediately is that when the next Livewire request happens (e.g. to save), and the state is not set yet, the value is then not included in the request and gets lost.

In my suggestion, the only part that gets deferred is the `commit()` which happens on the next request automatically anyway, so you don't get situations where the state is not synced in time for the request.

Does that make a bit more sense?

We may still need the state comparison too, yes.

People-Sea:
Good idea. I'll start dealing with it as soon as I get home.

danharrin (approved):
Perfect, thanks

PR #18228, opened 2025-10-15
[4.x] Improved $get utility 

johanmolen:
Improve the `Get` utility.
This will introduce the following methods:

```php
function (Get $get): void {
    $get->string('name');
    $get->integer('count');
    $get->float('amount');
    $get->boolean('enabled');
    $get->array('setttings');
    $get->date('date_from');
    $get->enum('role', UserRoleEnum::class);
    $get->filled('name');
    $get->blank('name');
}
```


danharrin (changes requested):
Instead of introducing a new utility, how about just improve `$get` - `$get->string('foo')`.

We shouldn't even need to use Fluent internally, just introduce new methods directly on the `Get` class that have stronger types returned. `array`, `string`, `float`, `int` and `bool` should be enough

johanmolen:
@danharrin even better, I will PR it soon.

johanmolen:
@danharrin would you prefer using the `InteractsWithData` trait or standalone methods on the `Get` utility?

danharrin:
Let's just add a small subset of the useful methods in an optimised way

johanmolen:
@danharrin I made the changes but I feel like it can still be improved by a lot. Any suggestions?

```php
function (Get $get): void {
    $get->string('name');
    $get->integer('count');
    $get->float('amount');
    $get->boolean('enabled');
    $get->array('setttings');
    $get->date('date_from');
    $get->enum('role', UserRoleEnum::class);
    $get->filled('name');
    $get->blank('name');
}
```

danharrin (changes requested):
Please remove the `$default` parameters

Please return an actual string from `string()`, not `Stringable`

Instead of `filter_var` for `boolean()`, you should probably just `(bool) $value`

Please remove `collect()`

For `date()`, I don't think you need the `$format` and `$tz` params, if you `$get()` a `DateTimePicker` you already get the correct object. I don't think we need to do any casting in there.

johanmolen:
@danharrin I made the changes suggested, except the boolean change. Because there is a difference:

```php
filter_var('true', FILTER_VALIDATE_BOOLEAN); // true
filter_var(1, FILTER_VALIDATE_BOOLEAN); // true
filter_var('string', FILTER_VALIDATE_BOOLEAN); // false
(bool) 'true'; // true
(bool) 1; // true
(bool) 'string'; // true
```

If we want to keep it in line with Laravel we should use the `filter_var`.
I can still make the change if you prefer `(bool)`.

danharrin:
I would expect `'string'` to be true, not false, since it is truthy. Not sure why Laravel opted for that other behaviour tbh

johanmolen:
@danharrin I made the change 👍

packages/schemas/src/Components/Utilities/Get.php
```diff
@@ -31,4 +35,57 @@ public function __invoke(string | Component $path = '', bool $isAbsolute = false
...
+
+        return Date::parse($state);
+    }
+
+    /** @param class-string<BackedEnum> $enumClass */
+    public function enum(string $key, string $enumClass, bool $isAbsolute = false): ?BackedEnum
+    {
+        $state = $this($key, $isAbsolute);
+
+        if (! enum_exists($enumClass) || ! is_string($state)) {
```
danharrin: I think it should error if the enum doesn't exist, and also enums can have int backing values, not just strings

```suggestion
        if (blank($state)) {
```

packages/schemas/src/Components/Utilities/Get.php
```diff
@@ -31,4 +35,57 @@ public function __invoke(string | Component $path = '', bool $isAbsolute = false
...
+        $state = $this($key, $isAbsolute);
+
+        if (! is_string($state)) {
+            return null;
+        }
+
+        return Date::parse($state);
+    }
+
+    /** @param class-string<BackedEnum> $enumClass */
```
danharrin: Please put this docblock on multiple lines, like others in the codebase

packages/schemas/src/Components/Utilities/Get.php
```diff
@@ -31,4 +35,57 @@ public function __invoke(string | Component $path = '', bool $isAbsolute = false
...
+
+    public function date(string $key, bool $isAbsolute = false): ?CarbonInterface
+    {
+        $state = $this($key, $isAbsolute);
+
+        if (! is_string($state)) {
+            return null;
+        }
+
+        return Date::parse($state);
```
danharrin: I think use `Illuminate\Support\Carbon` instead of `Date` so its consistent with other areas

packages/schemas/src/Components/Utilities/Get.php
```diff
@@ -31,4 +35,57 @@ public function __invoke(string | Component $path = '', bool $isAbsolute = false
...
+    public function array(string $key, bool $isAbsolute = false): array
+    {
+        return (array) ($this($key, $isAbsolute) ?? []);
+    }
+
+    public function date(string $key, bool $isAbsolute = false): ?CarbonInterface
+    {
+        $state = $this($key, $isAbsolute);
+
+        if (! is_string($state)) {
```
danharrin:
```suggestion
        if (blank($state)) {
```

packages/schemas/src/Components/Utilities/Get.php
```diff
@@ -31,4 +35,57 @@ public function __invoke(string | Component $path = '', bool $isAbsolute = false
...
+    {
+        return (float) $this($key, $isAbsolute);
+    }
+
+    public function boolean(string $key, bool $isAbsolute = false): bool
+    {
+        return (bool) $this($key, $isAbsolute);
+    }
+
+    /** @return array<mixed, mixed> */
```
danharrin: Please format this on multiple lines

packages/schemas/src/Components/Utilities/Get.php
```diff
@@ -31,4 +35,57 @@ public function __invoke(string | Component $path = '', bool $isAbsolute = false
 
         return $component->getState();
     }
+
+    public function string(string $key, bool $isAbsolute = false): string
+    {
+        try {
+            return (string) ($this($key, $isAbsolute) ?? '');
```
danharrin: I think this should throw an error if its not a stringable value, so remove the `try`

danharrin (changes requested):

Estigy:
Please don't make `->boolean(...)` behave different from Laravel.
In Laravel it is implemented like this:

grafik

(See `InteractsWithInput` trait.)

I would strongly suggest to implement it the very same way in Filament. There is no good reason to handle it differently here when we are in a Laravel environment.

danharrin:
@Estigy it was my suggestion to implement it differently to Laravel.

I want this

```php
->visible(fn (Get $get) => $get('value'))
```

To behave the same as this

```php
->visible(fn (Get $get) => $get->bool('value'))
```

And if you have a text input with a truthy value in your version, the second would return false instead of true and behave different.

This is about consistency with Filament over consistency with Laravel.

@johanmolen I would like to see 2 new methods added though: `$get->filled()` and `$get->blank()`, which use the `filled()` and `blank()` helpers on the state to return a sane boolean based on the value.

Estigy:
@danharrin I see where you are coming from, but in my point of view those are two different things and it's ok (and even preferable) to have them yield different results:

In `->visible(fn (Get $get) => $get('value'))` the value is cast to boolean via PHP- it's not Filament that does the typecasting.
In `->visible(fn (Get $get) => $get->bool('value'))` on the other hand, we use a special method to get a boolean. In Laravel, this method on the Request class was *especially* introduced so *not* any string would yield `true` via PHP's typecasting, but only those that have some sort of special meaning. https://laravel.com/docs/12.x/requests#retrieving-boolean-input-values

When I see `$get->bool(...)`, I don't expect *any* string to yield bool because then the programmer could simply have omitted the method alltogether.

So when you see "I want this ... To behave the same as this ...", it's the complete opposite of what Laravel programmers would expect, as those two are not the same in Laravel.

danharrin:
I don't agree, sorry. I don't like the way Laravel handles this.

Estigy:
Ok :-)

johanmolen:
@danharrin Done!

danharrin (approved):

johanmolen:
Thanks a lot @danharrin

PR #18230, opened 2025-10-15
Toggle Buttons shouldn't overwrite colors

grafst:
## Description
When we set the ->boolean() after our custom colors it overwrites them.
toggle-color

## Functional changes
Only set the colors if they are still empty


danharrin:
I don't think I'm going to merge this at the moment, I don't think the problem is super serious, and the solution here is not perfect since if you pass a function to `colors()` that returns null or an empty array, the boolean colors are then broken.

grafst:
OK, for me it leads to unpredictable behaviour since i don't pay attention to the order of the methods and then some of the buttons are red and some are orange.
Would it be enough to check for null or an empty array?

PR #18234, opened 2025-10-16
feat(header): allow hiding title and show only breadcrumbs

pankajXdev:
## Description

This pull request addresses the inability to hide the page heading while keeping the breadcrumbs visible.

It introduces a new `isHeadingHidden` property on the `BasePage` class. When this property is set to `true` on a page, it will hide the heading and subheading, but the breadcrumbs will remain visible. This provides more flexibility for customizing the page header.

### Example usage:

 ```
class MyPage extends Page
  {
      protected bool $isHeadingHidden = true;
      // ...
  }
```

## Visual changes

**Before:**
before

**After:**
after

## Functional changes

### DIFF
--- a/packages/panels/resources/views/components/header/index.blade.php
+++ b/packages/panels/resources/views/components/header/index.blade.php
@@ -3,6 +3,7 @@
     'actionsAlignment' => null,
     'breadcrumbs' => [],
     'heading',
+    'isHeadingHidden' => false,
     'subheading' => null,
 ])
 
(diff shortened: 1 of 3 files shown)


danharrin:
I don't think we need a new method for this. You can make a new PR that changes the return type of `getHeading()` to allow `null`, and then check if it is blank before rendering the heading element.

pankajXdev:
Thank you, I thought of this but added a new property to keep it a bit cleaner to use.

PR #18236, opened 2025-10-16
Add extraInputAttributeBag to rich editor

goozkan:
## Description

Refactored the rich editor's content div to use $extraInputAttributeBag for applying classes. This allows for better extensibility.

Note: Tailwind classes applied via the attribute bag are not currently taking effect, though the editor's styling and functionality remain intact. Further investigation is needed to enable Tailwind utility class application. ex.('class'=>'h-20' is on the dom but not taking effect but 'style'=>'height:200px' is working)

## Functional changes


danharrin:
`extraInputAttributeBag` is already applied to the element above the one you applied it to, so this would result in duplicate attributes.

PR #18239, opened 2025-10-16
[4.x] Remove fieldset top-padding, when unlabeled and uncontained

caseydwyer:
## Description
The `fieldset` component is super-handy for _(a number of reasons, one of which is)_ controlling visibility for an entire set of fields. The only downside—if the `fieldset` is solely for that purpose and you have `hiddenLabel()` & `contained(false)`—is that you get some unnecessary top-padding applied (`pt-6`, from the `.fi-fieldset-not-contained`). 

This PR fixes that style issue and takes the `pt` one step further, by removing it entirely when both `.fi-fieldset-not-contained` and `.fi-fieldset-label-hidden` are present on the fieldset wrapper.

### Before:
image

### After:
image

## Functional changes

- [X] Changes have been tested to not break existing functionality. _(lmk if we need to add screenshot tests)_
- [X] Documentation is up-to-date. _(lmk if you'd like an example of this, in Schema docs or Forms reactive cookbook)_

### DIFF
--- a/packages/support/resources/css/components/fieldset.css
+++ b/packages/support/resources/css/components/fieldset.css
@@ -20,4 +20,8 @@
     &.fi-fieldset-not-contained {
         @apply pt-6;
     }
+
+    &.fi-fieldset-not-contained.fi-fieldset-label-hidden {
+        @apply pt-0;
+    }
 }


howdu:
Sorry if I'm missing the point - but can't you use a `Group` instead of `Fieldset`?

caseydwyer:
> Sorry if I'm missing the point - but can't you use a `Group` instead of `Fieldset`?

@howdu Ha, yeah you totally can—good call. 

I actually went looking for that exact thing (`Group`) in the docs and couldn't find it, hence this PR. Tested locally and confirmed the `Group` component does exactly what I was looking for...it just didn't show up in a search, so I didn't know about it.

Can you point me to any docs you're aware of that cover the `Group` component? Not trying to be difficult, I just genuinely cannot find them. If missing, I'm also happy to add a section on that...though I am (obviously) probably not the most qualified person for that job. 😅 

In any case, closing this PR. Thanks!

webard:
@caseydwyer https://filamentphp.com/docs/4.x/forms/overview#saving-data-to-relationships here is something about Group component.

caseydwyer:
@webard thanks, that's a great use-case that I didn't know about (the `relationship` config)—sounds like you can use that with other layout components, so would presumably work on a `Section` too? Regardless, appreciate you pointing that out. Would still be helpful/justified, IMHO, to give `Group` its own dedicated bit, probably somewhere in the Schema docs. I'll take a crack at that.

PR #18241, opened 2025-10-16
Docs: add Group overview to Schemas > Layouts

caseydwyer:
## Description

Adds some documentation for `Group` component, under **Schema > Layouts**. Loosely related: #18239 

## Visual changes
N/A

## Functional changes

### DIFF
--- a/packages/schemas/docs/02-layouts.md
+++ b/packages/schemas/docs/02-layouts.md
@@ -200,6 +200,21 @@ Grid::make([
     ])
 ```
 
+### Group component
+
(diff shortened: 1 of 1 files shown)


danharrin:
Any layout component, such as `Grid`, would allow you to control visibility all at once, without affecting the design.

I don't think we should document yet another option for this, it would probably result in more decision fatigue for the user to decide what component to use, when there aren't any real benefits except for a slightly shorter syntax.

caseydwyer:
@danharrin totally get that, especially re: decision fatigue. That makes sense. 👍 Should the `Group` component still be included, though, without the deeper example? Eg:

```php
Group::make([
    TextInput::make('title'),
    TextInput::make('subtitle'),
])
```
Just seemed like a component that should be included in the docs, and I couldn't find it referenced anywhere.

PR #18252, opened 2025-10-17
Support array route patterns for navigation items

wychoong:
## Description

Updated getNavigationItemActiveRoutePattern methods to allow returning string or array types, enabling more flexible route matching for navigation item activation. 

## Functional changes


danharrin (changes requested):
Please fix the v4 merge conflicts

Are you sure that returning an array works? `routeIs()` does not handle arrays, it handles multiple string arguments

wychoong:
> Please fix the v4 merge conflicts

not sure is this the right way to merge. how can I backport this to v3?

> Are you sure that returning an array works? `routeIs()` does not handle arrays, it handles multiple string arguments

`routeIs` pass to `Route->named()` which call to `Str::is($pattern)` that can handles array matching
https://github.com/laravel/framework/blob/500c8372dfd6eff4a3ef1399ed21575dd7d9c774/src/Illuminate/Routing/Route.php#L913-L926

https://github.com/laravel/framework/blob/500c8372dfd6eff4a3ef1399ed21575dd7d9c774/src/Illuminate/Support/Str.php#L519

I tested with my v3 app and it works. my use case is a custom page within a custom page. Probably the best way would be just rename the route name, but that would require to change all pages' route name with a common denominator and make the match pattern 'pages.settings.*'
but with this change its just a simple `return [static::getRouteName(), PageOne::getRouteName(), PageTwo::getRouteName()];`

danharrin (approved):

wychoong:
@danharrin should i create a PR for v3?

danharrin:
We don't usually introduce new features for v3, and technically you could still do this by overriding `getNavigationItems()` in v3

wychoong:
> We don't usually introduce new features for v3, and technically you could still do this by overriding `getNavigationItems()` in v3

that would require duplicate the whole `NavigationItem::make` or `getNavigationItems()[0]`, not too sure how reliable is the latter option. or is there a better alternative that can get the navigation item?

PR #18256, opened 2025-10-17
Feat/header hide title

pankajXdev:
## Description

This change introduces the ability to hide the page title without affecting the breadcrumbs or header actions.

 Previously, there was no way to hide just the title. This refactors the header component so that if `getHeading()` returns `null`, only the heading is hidden, while the rest of the header (breadcrumbs, actions) remains visible.

## Visual changes

**Before:** The title was always visible in the header.
before

 **After:** The title can be hidden by returning `null` from `getHeading()`, but the breadcrumbs and actions will still be displayed.
after

## Functional changes


danharrin (approved):

pankajXdev:
hi thank you very much for the clean up. I didn't actually think to look up the usages. I will keep that in mind in the future.

PR #18257, opened 2025-10-18
fix: prevent XSS vulnerabilities in markdown editor by sanitizing localStorage and preview content

imhayatunnabi:
## Description

- Issue 1: localStorage XSS
- Issue 2: Preview Content XSS
- Spacing on `expect  or` to `expect or` double spacing to single spacing

## Functional changes


danharrin:
Sorry, this PR does not make sense to me at all really.

To start with, the double spacing is deliberate, that is the spacing that the browser generates when they send the message.

Also, I don't think that the rendering functions that you modified to "prevent XSS" are even used by Filament. EasyMDE is a dependency that we use and not every feature is in use. It also seems that EasyMDE already has its own sanitisation built in, so I don't think this is even a problem.

I'm not sure what caused this PR, was it AI created with little direction? If so, please don't submit more of these in the future. If this was well intentioned, my apologies.

PR #18260, opened 2025-10-18
fix: Builder add between action spacing and visibility

danharrin:


theonaciri:
Hi Dan, thanks for your work.
I find this change not aesthetic, and prefered the button appearing on hover of last version.
The button is an optional shortcut but is now too intrusive I think.
If the change was made for mandatory accessibility and align mobile functionality with desktop, then rendering the buttons could be toggled after clicking on a button next to 'Expand all' button.
Example:
image


For now, I personally prefer to lose the functionnality. I will add to my theme.css or app.css:
```
.fi-fo-builder-add-between-items-ctn {
    display: none;
}
```

danharrin:
This PR adds the hover effect you mention so I am not sure I understand your reply

theonaciri:
Indeed this commit was actually made to restore the hover effect.
You work so fast I adressed the fix instead of the PR that made the change.

My sincere apologies, I had no intention to disturb. Keep it up !

PR #18274, opened 2025-10-20
Enable/disable file uploads in RichEditor and MarkdownEditor through ->fileAttachments(...)

thijskuilman:
## Description

This PR introduces a more flexible way to enable/disable file uploads for the `RichEditor` and `MarkdownEditor` fields. 

Previously, file uploads were only enabled when the `attachFiles` toolbar button was explicitly present. This caused issues for projects that extend the editor with custom image or file upload tools (e.g. a custom ImagePlugin using its own attachImage tool), see [here](https://github.com/filamentphp/filament/pull/18169#issuecomment-3409288975).

To address this, a new method fileAttachments() has been added, based on the feedback from Dan [here](https://github.com/filamentphp/filament/pull/18169#issuecomment-3414701667). 

```php
// Enable file uploads
RichEditor::make('content')
    ->fileAttachments(true)

// Disable file uploads
RichEditor::make('content')
    ->fileAttachments(false)

// File uploads are enabled if the 'attachFiles' tool is present
RichEditor::make('content')
    ->toolbarButtons(['attachFiles'])
```

## Functional changes


danharrin (approved):

Doriiaan:
I had a customAttachFiles tool, and it took me one hour to figure out what was causing my crash after composer update. I didn't have ->fileAttachments(true)...

Thanks for thinking to add this option, but it can be nice to think about backward compatibility.

PR #18279, opened 2025-10-20
Fix hint evaluation

martin-ro:
## Description

When evaluating a field hint with a closure it appears that it renders the container div even if the result of the closure is null or false. That creates an additional space above the field.

## Visual changes

Screenshot 2025-10-20 at 5 01 41 PM
Screenshot 2025-10-20 at 5 01 52 PM

## Functional changes

### DIFF
--- a/packages/forms/src/Components/Concerns/HasHint.php
+++ b/packages/forms/src/Components/Concerns/HasHint.php
@@ -51,7 +51,7 @@ trait HasHint
                         /** @var self $parentComponent */
                         $parentComponent = $component->getContainer()->getParentComponent();
 
-                        return filled($parentComponent->hasHint());
+                        return filled($parentComponent->getHint());
                     });
             }
 
--- a/packages/infolists/src/Components/Concerns/HasHint.php
+++ b/packages/infolists/src/Components/Concerns/HasHint.php
@@ -50,7 +50,7 @@ trait HasHint
                         /** @var self $parentComponent */
                         $parentComponent = $component->getContainer()->getParentComponent();
 
-                        return filled($parentComponent->hasHint());
+                        return filled($parentComponent->getHint());
                     });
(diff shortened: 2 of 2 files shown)


martin-ro:
Reproduction repo if needed: https://github.com/martin-ro/filament-v4-hint

packages/forms/src/Components/Concerns/HasHint.php
```diff
@@ -51,7 +51,7 @@ protected function setUpHint(): void
                         /** @var self $parentComponent */
                         $parentComponent = $component->getContainer()->getParentComponent();
 
-                        return filled($parentComponent->hasHint());
+                        return filled($parentComponent->getHint());
```
danharrin, line 54: Shouldn't it be `return $parentComponent->hasHint()`?

danharrin (changes requested):
You might also want to check the other HasHint trait, I think there are two

martin-ro:
> You might also want to check the other HasHint trait, I think there are two

Updated the 2nd trait.

packages/forms/src/Components/Concerns/HasHint.php
martin-ro: `hasHint` returns `bool`. So checking if it is filled wouldn't work.

The if-statement a  few lines above already checks `hasHint` and it would never get there.

danharrin: Yes, I'm just saying `return $parentComponent->hasHint()` not `return filled($parentComponent->hasHint())`

martin-ro: The closure to show a hint could return null so just returning it here wouldn't work I think

danharrin (approved):

PR #18282, opened 2025-10-20
[3.x]: add option to make notification non-stackable

reidsolon:
## Description

There's no option to make toast notifications to not stackable. It should still be stackable by default

## Visual changes

Screenshot 2025-10-20 at 8 47 09 PM

https://github.com/user-attachments/assets/08f2d7ad-72e8-45f8-a04c-17076c536441

https://github.com/user-attachments/assets/d42614fa-913b-444f-ba50-dc7a2f88687c

## Functional changes


danharrin:
Hi, thanks for the work you put in! However, I am not going to introduce this change for now. I don't think I have heard other requests for such a feature from other users, but if this changes in the future I may reconsider.

PR #18284, opened 2025-10-20
[4.x] Add string target support for when opening in new tabs

leek:
## Description

This PR extends Filament’s URL-handling logic to support custom window targets (e.g. named tabs) while maintaining full backward compatibility.

**No breaking changes** — all existing code using `true|false` continues to work as before.

## Example usage

```php
{
    return $table
        // Default behavior (unchanged)
        ->openRecordUrlInNewTab() 
        ->openRecordUrlInNewTab(true) 
        ->recordUrl(fn ($record) => Resource::getUrl('view', ['record' => $record]))
        ->recordUrl(fn ($record) => Resource::getUrl('view', ['record' => $record]), true)

        // New behavior: reuse named tab/window
        ->openRecordUrlInNewTab('record-123')
        ->recordUrl(fn ($record) => Resource::getUrl('view', ['record' => $record]), 'record-123')
    ;
}
```

## Functional changes

### DIFF
--- a/packages/actions/src/Action.php
+++ b/packages/actions/src/Action.php
@@ -706,7 +706,7 @@ class Action extends ViewComponent implements Arrayable
             label: $this->getLabel(),
             size: $this->getSize(),
             tag: $url ? $shouldPostToUrl ? 'form' : 'a' : 'button',
-            target: ($url && $this->shouldOpenUrlInNewTab()) ? '_blank' : null,
+            target: $url ? $this->getUrlTarget() : null,
             tooltip: $this->getTooltip(),
             type: $this->canSubmitForm() ? 'submit' : 'button',
         );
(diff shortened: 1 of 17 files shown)


danharrin:
Hi, thanks for your PR. I'm going to hold off for now, as I haven't had many requests for this in the past. If I get lots more requests I may reconsider in the future. For now, I think you can use `extraAttributes()` to add this custom attribute.

leek:
> Hi, thanks for your PR. I'm going to hold off for now, as I haven't had many requests for this in the past. If I get lots more requests I may reconsider in the future. For now, I think you can use `extraAttributes()` to add this custom attribute.

@danharrin Thanks for the review. Unfortunately, `extraAttributes()` doesn't actually work unless I am doing something wrong?

```php
{
    return $table
		// only accepts bool 😞
        // ->openRecordUrlInNewTab(true) 

		// adds `target="test"` to container and not the `<a>`
        ->extraAttributes(['target' => 'test']) 
    ;
}
```

I am surprised that everyone wants `target="_blank"` all of the time given the number of available acceptable options:

| Value | Description |
|--------|--------|
| _blank | Opens the linked document in a new window or tab |
| _self | Opens the linked document in the same frame as it was clicked (this is default) |
| _parent | Opens the linked document in the parent frame |
| _top | Opens the linked document in the full body of the window |
| _framename_ | Opens the linked document in the named frame | 

I would urge you to reconsider, especially since this PR should maintain 100% backwards compatability.
🙏🙏

danharrin:
I would accept a PR for more `extra*Attributes()` methods like `extraRecordLinkAttributes()` for example. But on an action for example, it should already work

PR #18298, opened 2025-10-21
Hide attach files toolbar button when file attachments are disabled

Saracaen:
## Description

Extends #18274 to hide the attachFiles toolbar button when file attachments are disabled.

## Visual changes

*Before*

{7A75080B-447B-4571-9182-E54D0B23F76C}

*After*

{2EA0CF7A-31F4-45C5-96AC-CC16CE11D27D}

## Functional changes

### DIFF
--- a/packages/forms/src/Components/Concerns/InteractsWithToolbarButtons.php
+++ b/packages/forms/src/Components/Concerns/InteractsWithToolbarButtons.php
@@ -168,4 +168,9 @@ trait InteractsWithToolbarButtons
 
         return false;
     }
+
+    public function hasCustomToolbarButtons(): bool
+    {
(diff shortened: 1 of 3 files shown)


danharrin (approved):

danharrin (changes requested):
I think this causes an infinite loop, since `hasFileAttachments` calls `hasFileAttachmentsByDefault` which calls `hasToolbarButton` which calls this method

Saracaen:
Should be fixed now, though not 100% sure on the fix. Basically I check if toolbarButtons is not set and return true in that case, or otherwise check if toolbarButtons contains 'attachFiles'. Which should no longer loop infinitely, because it should not call getDefaultToolbarButtons anymore.

danharrin (approved):

PR #18300, opened 2025-10-21
Feature: Add option to block password managers for text inputs

skaesdorf:
## Description

Some password managers may suggest filling stored credentials or addresses into a text field. This often is the case when a field is named something like `name`, `street`, `city` etc.

This behaviour can be blocked for some of the most common password managers like Lastpass of 1Password using HTML attributes or classes.

My current usecase is an application with an address database. Using this feature, I can disable the address suggestion when creating or editing addresses in my application.

## Functional changes

### DIFF
--- a/packages/forms/docs/02-text-input.md
+++ b/packages/forms/docs/02-text-input.md
@@ -248,6 +248,25 @@ TextInput::make('apiKey')
     This feature only works when SSL is enabled for the app.
 </Aside>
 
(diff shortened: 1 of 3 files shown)


danharrin:
Hi, thanks but I am going to suggest you use `extraInputAttributes()` instead for this, so we don't have to maintain the additional provider-specific options in core.

PR #18304, opened 2025-10-22
fix: when a FusedGroup is wrapped in a section, the input background color is inconsistent with a regular input

### LINKED ISSUE #18275: FusedGroup visual inconsistency
### Package

filament/filament

### Package Version

v4

### Laravel Version

v12

### Livewire Version

v3

### PHP Version

PHP 8.4

### Problem description

When using a `FusedGroup` inside a `Section`, the background color of the other inputs doesn’t match, causing a visual inconsistency.

This issue only appears in **dark mode** !.

Image

### Expected behavior

All the inputs should have exactly the same styling.

### Steps to reproduce

I reproduce it on Filament v4 DEMO repo at `app/Filament/Resources/Blog/Links/Schemas/LinkForm.php`

Code:

```php
 public static function configure(Schema $schema): Schema
    {
        return $schema
            ->components([
                Section::make()
                    ->schema([
                        FusedGroup::make()
                            ->schema([
                                TextInput::make('title')
                                    ->maxLength(255)
                                    ->required(),
                                ColorPicker::make('color')
                                    ->required()
                                    ->hex()
                                    ->hexColor(),
                            ])
                            ->label('Test')
                            ->columns(2),
                Textarea::make('description')
                    ->maxLength(1024)
                    ->required()
                    ->columnSpanFull(),
                TextInput::make('url')
                    ->label('URL')
                    ->required()
                    ->maxLength(255)
                    ->columnSpanFull(),
                SpatieMediaLibraryFileUpload::make('image')
                    ->collection('link-images')
                    ->acceptedFileTypes(['image/jpeg']),
                    ]),
            ]);
    }
```

https://github.com/MarkasCherry/demo/tree/bug/fused-group-styling-inconsistency

### Relevant log output

```shell

```

People-Sea:
## Description

fixes: #18275

Because the `fused-group` component uses background color to simulate the border between input boxes, adding another layer of `section` causes color inconsistency.  This PR has already used the preset colors of Tailwind CSS and adjusted the transparency to achieve the closest color.

## Visual changes

# Before 
image

# After 
image

## Functional changes


People-Sea:
I'll try again to simulate the middle line without using background

alexmanase:
Also, I think there is an inconsistence with the input height.

For example here:

image

The "Discount price" field is the fusion of 2 text inputs (which doesn't have the same height as a single input), and the "Hire date" is the fusion of 2 date inputs(which has the good height).

Also, you can see in your example that the height is not the same:

image

danharrin (changes requested):
Does this still work if a JS select dropdown is in a fused group, and that is inside a modal?

People-Sea:
I tried it out and found that this solution is indeed ineffective in the modal.  However, if overflow: hidden is not used, the class cannot be used to control the rounded corners (because this component supports custom column numbers).

People-Sea:
@danharrin How about this solution of using JavaScript for dynamic calculation and adding classes to achieve rounded corners and borders?

danharrin:
Replaced by #18372

PR #18348, opened 2025-10-29
Fix Repeater persistCollapsed

TheSkybreaker:
## Description

The `Repeater` has the `persistCollapsed` method in the `CanBeCollapsed` concern, but it does nothing at the moment. This add the check for it so it can stay collapsed (or not)

## Visual changes

//

## Functional changes

Now the Repeater item stays collapsed based on user choice if persistCollapsed is set to `true`

- [ ] Documentation is up-to-date. - I'm not sure the method is documented in the Repeater part of the documentation


danharrin (changes requested):
Please reflect the changes to Builder

danharrin (approved):

PR #18376, opened 2025-10-31
Improve visual separation of select optgroup headers

archilex:
## Description

When using optgroups with selects the headers look too much like clickable options causing confusion for users. This PR uses darker text to indicate they are not clickabled, but also semibold to indicate it's not just a disabled option.

`.fi-dropdown-header` was targeted from the select.css class instead of directly in`header.css` because `.fi-dropdown-header` is also used for the dropdown blade component and those styles shouldn't be affected.

## Visual changes

Screenshot 2025-10-31 at 12 27 19 PM

Screenshot 2025-10-31 at 12 26 51 PM

## Functional changes

### DIFF
--- a/packages/support/resources/css/components/input/select.css
+++ b/packages/support/resources/css/components/input/select.css
@@ -69,6 +69,10 @@ select.fi-select-input {
 
     & .fi-select-input-option-group {
         @apply divide-y divide-gray-100 dark:divide-white/5;
+
+        & .fi-dropdown-header {
+            @apply font-semibold text-gray-400 dark:text-gray-500;
+        }
     }
 
     & .fi-select-input-search-ctn {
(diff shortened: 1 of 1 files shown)


packages/support/resources/css/components/input/select.css
```diff
@@ -69,6 +69,10 @@ select.fi-select-input {
 
     & .fi-select-input-option-group {
         @apply divide-y divide-gray-100 dark:divide-white/5;
+
+        & .fi-dropdown-header {
+            @apply font-semibold text-gray-400 dark:text-gray-500;
```
danharrin:
```suggestion
            @apply font-medium text-gray-500 dark:text-gray-400;
```

danharrin (changes requested):

danharrin (approved):

PR #18394, opened 2025-11-03
Vasi rus/failsafe for getting notification on null

### LINKED ISSUE #18390: Call to a member function notifications() on null
### Package

filament/filament

### Package Version

v3.0, v4.0

### Laravel Version

v.12

### Livewire Version

v.3

### PHP Version

PHP 8.4

### Problem description

This typically occurs:
	•	On pages or components that attempt to fetch user notifications, such as the Filament notification dropdown or database notification query, before a user is authenticated.
	•	In background jobs or console commands that run outside a user context.
	•	When using a custom auth setup (e.g., multiple guards) and the expected user guard isn’t active.

💡 Root Cause

The code assumes that Filament::auth()->user() always returns a valid user model.
However, when null is returned, calling $user->notifications() immediately throws a fatal error.

Image

### Expected behavior

When the authenticated user model is not available (for example, no user is logged in), the system should gracefully handle notification queries without throwing an error.

Specifically:
	•	The code should check whether a user instance exists before attempting to access the notifications() relationship.
	•	If a user is authenticated, the system should return that user’s notifications as usual.
	•	If no user is authenticated, the system should instead return:
	•	an empty notification collection, or
	•	the notification model type (depending on implementation needs),
without causing a fatal exception.

### Steps to reproduce

	1.	Set up a Filament v4 panel that includes the notification component (for example, the user menu notification dropdown or a widget that calls auth()->user()->notifications()).
	2.	Log out of the application, or access the panel without an authenticated user (so that Filament::auth()->user() returns null).
	3.	Try to load a Filament page that references notifications — for example:
	•	The dashboard with the notification bell enabled.
	•	Any component, widget, or layout that calls auth()->user()->notifications().
	4.	Observe the error in your logs or browser: `Call to a member function notifications() on null`

https://github.com/vasi-rus/filament

### Relevant log output

```shell

```

vasi-rus:
## Description

This typically occurs:
• On pages or components that attempt to fetch user notifications, such as the Filament notification dropdown or database notification query, before a user is authenticated.
• In background jobs or console commands that run outside a user context.
• When using a custom auth setup (e.g., multiple guards) and the expected user guard isn’t active.

💡 Root Cause

The code assumes that Filament::auth()->user() always returns a valid user model.
However, when null is returned, calling $user->notifications() immediately throws a fatal error.

image

## Functional changes

- [ ] The issue was fixxed by addding a check for user instance.

### DIFF
--- a/packages/notifications/src/Livewire/DatabaseNotifications.php
+++ b/packages/notifications/src/Livewire/DatabaseNotifications.php
@@ -89,8 +89,14 @@ class DatabaseNotifications extends Component implements HasActions, HasSchemas
 
     public function getNotificationsQuery(): Builder | Relation
     {
-        /** @phpstan-ignore-next-line */
-        return $this->getUser()->notifications()->where('data->format', 'filament');
+        $user = $this->getUser();
+
+        if (!$user) {
+            // Return an empty builder to satisfy the return type
+            return DatabaseNotification::query()->limit(0);
+        }
+
+        return $user->notifications()->where('data->format', 'filament');
     }
 
     public function getUnreadNotificationsQuery(): Builder | Relation
--- a/packages/notifications/src/Livewire/DatabaseNotifications.php
+++ b/packages/notifications/src/Livewire/DatabaseNotifications.php
@@ -91,9 +91,8 @@ class DatabaseNotifications extends Component implements HasActions, HasSchemas
     {
         $user = $this->getUser();
 
-        if (!$user) {
-            /** @phpstan-ignore-next-line */
-            return DatabaseNotification::query()->limit(0);
+        if (! $user) {
+            abort(401);
         }
 
         /** @phpstan-ignore-next-line */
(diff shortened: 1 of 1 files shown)


packages/notifications/src/Livewire/DatabaseNotifications.php
```diff
@@ -89,8 +89,15 @@ public function isPaginated(): bool
 
     public function getNotificationsQuery(): Builder | Relation
     {
+        $user = $this->getUser();
+
+        if (!$user) {
+            /** @phpstan-ignore-next-line */
+            return DatabaseNotification::query()->limit(0);
```
danharrin: Instead, how about just using `abort(401)` or something?

danharrin (changes requested):

packages/notifications/src/Livewire/DatabaseNotifications.php
vasile-rus: I think is even better :)

danharrin:
```suggestion
        if (! $user) {
            abort(401);
```

danharrin (changes requested):

danharrin (approved):

PR #18396, opened 2025-11-03
feat(rich-editor): Implement hiddenLabel() modifier to enable tool labels

bbredewold:
## Description

This implements the hiddenLabel modifier to the RichEditorTool to allow the use of a text label next to the icon.

- When `->hiddenLabel(false)` is called on the `RichEditorTool`;
- the set label is now displayed right next to the icon, inside the `<button>`;
- and the tooltip is hidden.

## Visual changes
The text label appears next to the icon in the button.
BRB2025-11-03 at 10 35 50@2x

## Functional changes


saade:
Wouldn't it be better to use the native `->hiddenLabel(bool)`?

bbredewold:
> Wouldn't it be better to use the native `->hiddenLabel(bool)`?

Maybe, but the isLabelHidden property would be true by default. Not sure if that is pretty. What do you think?

saade:
In that case, I think it should be hidden by default. But let's hear what @danharrin thinks

danharrin:
Yes I think we should use that existing method, and just override the property to change the default.

bbredewold:
Updated to implement ->hiddenLabel()

PR #18404, opened 2025-11-03
feat(builder): Add option to prevent block add/edit modals from closing in builder form component

cobisimo:
## Description

In builder, block definition added option to prevent block add/edit modals from closing in builder form component when we are using preview mode.
By default, this modal was closable by default by clicking outside or pressing ESC button, which is not convenient in my use case, as the user is losing edit data and has to do that again if click outside by mistake.

## Visual changes

No visual changes

## Functional changes


packages/forms/src/Components/Builder.php
```diff
@@ -663,6 +669,8 @@ public function getEditAction(): Action
             ->label(__('filament-forms::components.builder.actions.edit.label'))
             ->modalHeading(__('filament-forms::components.builder.actions.edit.modal.heading'))
             ->modalSubmitActionLabel(__('filament-forms::components.builder.actions.edit.modal.actions.save.label'))
+            ->closeModalByClickingAway(fn (Builder $component): bool => ! $component->isModalPersistent())
+            ->closeModalByEscaping(fn (Builder $component): bool => ! $component->isModalPersistent())
```
cobisimo, line 673: I introduced single property for both click outside and ESC dialog close. I don't see use case when someone want one, but don't want another, and I wanted to keep it simple.

danharrin:
You should use `->editAction(fn (Action $action) => $action->closeModalByClickingAway()-> closeModalByEscaping())` etc. We should avoid introducing new proxy methods for all the different combinations of customisation methods, and use the existing APIs that give us full access to the underlying objects.

cobisimo:
Thanks, it makes perfect sense. I was not aware of that functionality.

PR #18418, opened 2025-11-05
Fix column manager polling and reordering

### LINKED ISSUE #18328: Column manager ignores “Apply columns” in polling tables
### Package

filament/filament

### Package Version

v4.1.10

### Laravel Version

v12.35.1

### Livewire Version

v3.6.4

### PHP Version

PHP 8.4.14

### Problem description

When you toggle a column in the column manager, the change is applied the next time the table is polled without having to press “Apply columns” first.

*In addition to the bug report, I would like to suggest making these apply buttons optional. The bug made me realize once again how much better it is when you don't have to confirm first.*

### Expected behavior

It should not be applied without first clicking on “Apply columns”.

### Steps to reproduce

1. Create a table with `->poll()`.
2. Open the column manager and hide/show any column.
3. The next time the table polls, the change will be applied without having to press “Apply columns” first.

### Relevant log output

```shell

```

archilex:
Fixes #18328 

I also noticed while fixing polling that reordering a column back to it's original location would not work as expected. This is due to the `hasReorderedTableColumns()` method returning false since the state is the same which is then persisted in the session to avoid calling the heavier `syncReorderableColumnsFromDefaultTableColumnState()`. 

However, when manually reordering columns `hasReorderedTableColumns` needs to be forced to return true so that when reordering back to the original order it is correctly applied.

### DIFF
--- a/packages/tables/resources/js/components/column-manager.js
+++ b/packages/tables/resources/js/components/column-manager.js
@@ -4,9 +4,13 @@ export default function filamentTableColumnManager({ columns, isLive }) {
 
         isLoading: false,
 
+        deferredColumns: [],
+
         columns,
 
         isLive,
+
+        hasReordered: false,
 
         init() {
             if (!this.columns || this.columns.length === 0) {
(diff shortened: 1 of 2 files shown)


danharrin (approved):

DaanTisg:
Just wanted to let you know this seems to break the Advanced Tables plugin:

Symfony\Component\ErrorHandler\Error\FatalError
vendor/archilex/filament-filter-sets/src/AdvancedTables.php:189

Declaration of Archilex\AdvancedTables\AdvancedTables::applyTableColumnManager(?array $state = null): void must be compatible with Filament\Resources\Pages\ListRecords::applyTableColumnManager(?array $state = null, bool $wasReordered = false): void

archilex:
@DaanTisg I've added compatibility in Advanced Tables v4.0.11. I was waiting for this to be released to be able to tag that release. Thanks.

PR #18419, opened 2025-11-05
Fix: Panel ID setter should tolerate repeats and surface tenant relationship options

imhayatunnabi:
## Summary
- prevent accidental duplicate panel ID assignments from exploding while blocking blank IDs
- make the missing panel ID exception actionable
- include relationship names when tenant ownership lookups fail for easier debugging
- commit the generated yarn.lock to lock dependency versions## Visual changes

## Functional changes


danharrin:
Thanks, but I don't think the extra exception information here is very helpful and could cause more confusion.

danharrin:
Please stop submitting AI PRs here @imhayatunnabi

Issue #13324, opened 2024-06-19
Tree reorderable on tables


danharrin:

johncarter-:
- Switch view toggle between standard table and tree editor.
 - Don't allow pagination.
 - If tree shows more than x rows show warning about performance?
 - Collapsible sub trees that can still be dragged with their children
 - Drag and drop trees between nested trees as well as being able to drag to top level.
 - Configurable attribute to act as title of draggable row/bar. E.g. title, slug, title and slug
 - Allow action / action group at end of draggable row/bar. To enable actions like delete, delete entire tree, create new child, etc.
 - Allow discard changes or save new order.

rvzug:
I've come across this issue/suggestion. Could this work with the [Nested Set package by Lazychaser](https://github.com/lazychaser/laravel-nestedset)? I want to dive into making Filament compatible with this package (including reorder/drag&drop into other records/drag&drop between other records/drag&drop to parent level etc.), but if you are already planning something. I may be waiting...

I'm searching for a way to make this type of table: 
![nested-set](https://cdn.productived.net/media/pages/reviews/mylifeorganized/6a68635eff-1632301567/tasks-1-1-0x300.webp)

GhostvOne:
- We should be able to access to the edit page of every items for relations manager

binaryfire:
@danharrin Do you think there's a way of getting [this](https://github.com/filamentphp/filament/pull/8916) working as part of this update? Reordering tables is already a bit hard to follow without the usual Sortable.js visual feedback. It'll be worse with trees.

Something like this would be a big UX improvement: https://sortablejs.github.io/Sortable/#nested

dmlogic:
Filament handling trees/nested sets would be amazing, and I'm happy to help implement any plan.

Our use case is Categories for which we use [laravel-nestedset](https://github.com/lazychaser/laravel-nestedset). We have two pain points:

Visually representing the tree when managing categories has been a challenge. With not a lot of available time we've ended up with:

* Viewing one level at a time with a breadcrumb generator in the infolist
* Relationship manager using a `children` `HasMany` relationship
* Custom actions to move up and down using the nested set methods

![Image](https://github.com/user-attachments/assets/c65595eb-de00-44b8-9b42-69f131ff68e0)

This works, but it's a poor representation of a tree.

[...]

PR #18423, opened 2025-11-05
feat: Add tree reordering functionality to tables

imhayatunnabi:
## Description https://github.com/filamentphp/filament/issues/13324

- add tree configuration API and disable pagination when tree mode is active
- wire nested drag and drop UI with Alpine and Sortable
- persist parent/position updates and add docs plus coverage for tree ordering
## Functional changes


danharrin:
Just taking an initial look here, I think this is quite far away from what we want. For example, none of the views are structured in a way that would work with the framework currently (fthey use Tailwind classes in the views instead of CSS files). As such, it seems like you may have just asked AI to create this feature, as if you would have tested locally then you would have found it to not work.

I may be wrong in my assumption here, but please can you avoid creating AI PRs like this, as they do not make the maintenance of the framework any easier for us. If this is hand-written code by you, I really appreciate the huge effort here, and I'm sorry.

I think I would like to tackle a large feature like this myself instead of accepting a community contribution, so that I fully understand how it works and I can maintain it going forward.

PR #18435, opened 2025-11-06
Update query column references for table context

iAmKevinMcKee:
## Description

We encountered a bug where if you search in a table, and then you try to sort a column that has another column with the same name (like 'name' and 'property.name'), you will get an exception. This ensures we are prepending the table name for the query.

## Functional changes


danharrin:
As I mentioned on Discord, we can't merge this one because of breaking changes with joined tables.

PR #18443, opened 2025-11-08
Add getOptionDescriptionFromRecordUsing() to CheckboxList

lucacastelnuovo:
## Description

Currently it is only possible to modify a records label using `getOptionLabelFromRecordUsing()`. I propose adding a `getOptionDescriptionFromRecordUsing()` to also be able to set a records description from an option’s Eloquent model.

## Functional changes

### DIFF
--- a/packages/forms/docs/06-checkbox-list.md
+++ b/packages/forms/docs/06-checkbox-list.md
@@ -270,10 +270,10 @@ CheckboxList::make('technologies')
 
 <Aside variant="warning">
     When using `disabled()` with `relationship()`, ensure that `disabled()` is called before `relationship()`. This ensures that the `dehydrated()` call from within `relationship()` is not overridden by the call from `disabled()`:
-    
+
     ```php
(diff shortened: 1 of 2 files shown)


lucacastelnuovo:
Should I add `getOptionDescriptionFromRecordUsing()` to `Select`, `ModalTableSelect`, `Type`, `SelectConstraint`, `IsRelatedToOperator`, `IsOperator`, `SelectColumn` and `SelectFilter` as well?

danharrin:
I don't think those other components have option descriptions in the first place, right? So no, we should not add this elsewhere

danharrin (changes requested):
Please fix the PHPStan issues, you probably need to save the results of the `has*` methods into vars when using them so the results stay the same

lucacastelnuovo:
> Please fix the PHPStan issues, you probably need to save the results of the `has*` methods into vars when using them so the results stay the same

Check.

danharrin (approved):

PR #18444, opened 2025-11-08
Ensure true mime type is passed as a header by default

joshembling:
## Description

I have recently discovered an issue where content that has first been downloaded as a stream, such as pdf's, png's, jpg's etc. do not keep their true mime type as the 'Content-Type' header is missing/incorrect. 

Example: 
- I have a pdf file with the mime type `application/pdf`
- I would expect upon upload that this mime type persists with this correct Content-Type header
- Instead the mime type is uploaded as `application/octet-stream` 

This means that after uploading such a file, the content is always streamed because of the way the file was retrieved the first time, which isn't expected behaviour. 

What I would expect: 
- The pdf file has the correct mime type `application/pdf`
- The upload is persisted with this same mime type by setting the correct header
- The file opens in the browser and doesn't download directly to one's device

This PR alleviates that issue by first passing in the correct mime type as a header, but can always be overwritten by the user e.g. 

```php
SpatieMediaLibraryFileUpload::make('attachments')
    ->multiple()
    ->customHeaders(['Content-Type' => 'application/octet-stream'])
```

Example file you can try this with is attached here:
[invoice_123_08_11_2025_test.pdf](https://github.com/user-attachments/files/23431904/invoice_123_08_11_2025_test.pdf)

Upload this file to the latest release of this package. Notice the mime_type in the database is `application/octet-stream`. 

Now install my changes. See the mime_type is correctly `application/pdf`.

## Functional changes

### DIFF
--- a/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
+++ b/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
@@ -140,7 +140,7 @@ class SpatieMediaLibraryFileUpload extends FileUpload
             $filename = $component->getUploadedFileNameForStorage($file);
 
             $media = $mediaAdder
-                ->addCustomHeaders($component->getCustomHeaders())
+                ->addCustomHeaders([...['ContentType' => $file->getMimeType()], ...$component->getCustomHeaders()])
                 ->usingFileName($filename)
                 ->usingName($component->getMediaName($file) ?? pathinfo($file->getClientOriginalName(), PATHINFO_FILENAME))
                 ->storingConversionsOnDisk($component->getConversionsDisk() ?? '')
--- a/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
+++ b/packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
@@ -140,7 +140,7 @@ class SpatieMediaLibraryFileUpload extends FileUpload
             $filename = $component->getUploadedFileNameForStorage($file);
 
             $media = $mediaAdder
-                ->addCustomHeaders([...['ContentType' => $file->getMimeType()], ...$component->getCustomHeaders()])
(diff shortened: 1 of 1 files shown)


danharrin:
I understand that this does fix the problem, but what is the root cause? It seems like it could be a bug in the Spatie Media Library package, but I might be wrong?

packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
```diff
@@ -140,7 +140,7 @@ protected function setUp(): void
             $filename = $component->getUploadedFileNameForStorage($file);
 
             $media = $mediaAdder
-                ->addCustomHeaders($component->getCustomHeaders())
+                ->addCustomHeaders([...['ContentType' => $file->getMimeType()], ...$component->getCustomHeaders()])
```
danharrin: Currently this is using `ContentType`, should it be this or should it be `Content-Type`? @joshembling

joshembling: Sorry for the delay @danharrin, been on holiday for the last week. 

Yes good catch, will amend that now!

joshembling:
> I understand that this does fix the problem, but what is the root cause? It seems like it could be a bug in the Spatie Media Library package, but I might be wrong?

@danharrin  Technically, I'm not sure this is a _certain_ bug, because it does persist a content type it was first associated with. But in every use case I can think of, the content type would never be expected to be anything else other than the file type, hence the unexpected behaviour.

I have dug through the spatie code, and I've yet to find the absolute root cause.

packages/spatie-laravel-media-library-plugin/src/Forms/Components/SpatieMediaLibraryFileUpload.php
danharrin: Can you test that the PR works locally please?

joshembling: Confirmed my original solution was the correct one. It should be `ContentType`. 

Amended in latest commit.

danharrin (approved):

PR #18446, opened 2025-11-08
[v4 Docs] Replace 'size' with 'iconSize' in icon-button examples

notAreYouScared:
## Description
Documentation changes only!

size is not used in icon buttons... so to change the actual size of the icon button you need to set iconSize

## Functional changes


danharrin:
Have a look at the Blade component source code and CSS, the `size` param does change the overall size of the button, but not the icon size which is controlled separately. This is consistent with the other button components, where the `size` param changes the size of the button but not the icon. I would accept a different PR to document `icon-size` separately, not replacing this.

PR #18472, opened 2025-11-11
[4.x] Fix table sorting

DirtyRacer1337:
## Description

Fixes duplicate alias error when relation column in same table

result sql goes like

from
```sql
select
  "sites".*
from
  "sites"
  left join "sites" on "sites"."parent_id" = "sites"."id"
order by
  "sites"."name" asc
```
to
```sql
select
  "sites".*
from
  "sites"
  left join "sites" as "sort" on "sites"."parent_id" = "sort"."id"
order by
  "sort"."name" asc
```


danharrin (changes requested):
Would you mind submitting a failing test PR for this please, so I can understand the use case better?

Also note the changes in #18499, we are moving away from joins when there is a simple non-nested relationship, so to reproduce when that is merged you might need a nested relationship.

DirtyRacer1337:
your pr works for me

PR #18476, opened 2025-11-11
Fix Textarea autosize causing scroll anchoring issues in Firefox

Saracaen:
## Description

When a Filament Textarea component with ->autosize() enabled is placed in a flex container (e.g., a chat input footer), it causes an adjacent, scrollable flex container (e.g., a message list) to lose its scroll position on every keystroke.

This "scroll jump" is specific to Firefox and makes for a very poor user experience in common layouts like chat interfaces.

## Functional changes


danharrin:
Please open an issue with a reproduction repository, just removing a line of code in a PR like this without further technical explanation is not safe for me to merge.

Saracaen:
@danharrin I can explain why this is happening. It has to do with the element being resized twice whenever a resize is happening. The setInitialHeight is already called in the init, which is expected behavior. But with it also being called for every resize what essentially happens is it grows/shrinks to the default size, before resizing to the intended size. This double resize causes a jump in Firefox that never recovers. Plus there is no reason or need to resize an element twice, just setting it to the desired size is more than sufficient.

I will see if I can create a reproduction repository.

danharrin:
I am quite sure that setting the element to the initial size is the fix for another issue.

Saracaen:
I have created a bug ticket with a reproduction repository, so I hope that helps! #18528

PR #18485, opened 2025-11-12
The navigation group won't show in RTL mode

hans-thomas:
## Description

After upgrading to v4, the navigation group no longer works in RTL mode.

## Visual changes

Before:

[Screencast From 2025-11-08 10-24-44.webm](https://github.com/user-attachments/assets/00275c19-9e2d-4d18-b239-0c4ba3c816d7)

After:

[Screencast From 2025-11-12 16-49-25.webm](https://github.com/user-attachments/assets/7b8cb4d5-1046-4f6c-a5cc-ea1ba538cd14)

## Functional changes

### DIFF
--- a/packages/panels/resources/views/components/sidebar/group.blade.php
+++ b/packages/panels/resources/views/components/sidebar/group.blade.php
@@ -64,7 +64,6 @@
     @if ($hasDropdown)
         <x-filament::dropdown
             :placement="(__('filament-panels::layout.direction') === 'rtl') ? 'left-start' : 'right-start'"
-            teleport
             x-show="! $store.sidebar.isOpen"
         >
             <x-slot name="trigger">
--- a/packages/panels/resources/views/components/sidebar/group.blade.php
+++ b/packages/panels/resources/views/components/sidebar/group.blade.php
@@ -29,13 +29,13 @@
         <div
             @if ($collapsible)
                 x-on:click="$store.sidebar.toggleCollapsedGroup(label)"
-            role="button"
+                role="button"
             @endif
             @if ($sidebarCollapsible)
                 x-show="$store.sidebar.isOpen"
-            x-transition:enter="fi-transition-enter"
-            x-transition:enter-start="fi-transition-enter-start"
-            x-transition:enter-end="fi-transition-enter-end"
+                x-transition:enter="fi-transition-enter"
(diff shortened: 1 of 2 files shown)


packages/panels/resources/views/components/sidebar/group.blade.php
```diff
@@ -64,7 +64,6 @@ class="fi-sidebar-group-collapse-btn"
     @if ($hasDropdown)
         <x-filament::dropdown
             :placement="(__('filament-panels::layout.direction') === 'rtl') ? 'left-start' : 'right-start'"
-            teleport
```
hans-thomas, line 67: The `teleport` parameter is `false` by default at

https://github.com/filamentphp/filament/blob/0926ea0677a70d1226f77a7fbf6bfcfe9fea48a6/packages/support/resources/views/components/dropdown/index.blade.php#L11

danharrin: This seems like a weird fix tbh, do you know what the original issue is?

I think at least we should only disable this conditionally:

Like `:teleport="__('filament-panels::layout.direction') !== 'rtl'"`

hans-thomas: I was thinking of disabling `teleport` conditionally, but when I commented this prop, I didn't observe any change. So, I just deleted this prop.

Okay then, I will disable it in RTL only.

hans-thomas:
Disabling the `teleport` prop will fix the issue in RTL mode, but in LTR mode, it seems unjustified.

With `teleport` prop:

image

without `teleport` prop:

image

danharrin (approved):

Youssef-Amjad:
Thank you for this great effort.

pedrofernando94:
This merged commit causes a problem where the topbar dropdown opens and closes immediately after clicking. Did the update on a project for the latest release and noticed it, after checking the releases changeling I noticed this Pull Request and that it was merged. I have now reverted the project back to previous version, however before I did it I added back the changes mades from this merged commit (removed teleport) directly in vendor and tested it, the problem does not appear without this changes.

danharrin:
@pedrofernando94 can you submit an issue with a reproduction repository please?

pedrofernando94:
@danharrin I created issue #18648 and provided reproduction repository.

PR #18487, opened 2025-11-12
fix: user menu drop down cannot be expanded with the keyboard

### LINKED ISSUE #18465: User menu drop down cannot be expanded with the keyboard
### Package

filament/filament

### Package Version

v4.2.0

### Laravel Version

v12.37

### Livewire Version

v3.6.4

### PHP Version

PHP 8.4.13

### Problem description

The user menu drop down cannot be expanded using only the keyboard. You are able to navigate to it with the keyboard, but the "enter" and "space bar" keys have no effect to expand it. You need to use the mouse to expand it. Once expanded, the contents are keyboard accessible.

### Expected behavior

The "enter" and "space bar" keys should be able to expand the user menu drop down so that a mouse isn't required.

### Steps to reproduce

1. Open the [Filament demo](https://demo.filamentphp.com/) (login if needed)
2. Using the keyboard press the "tab" key to move focus to the user menu drop down (DU in the top right hand corner)
3. Press the "enter" key
4. Press the "space bar" key
5. Notice that the drop down menu does not expand
6. Using the mouse click on the user menu (DU)
7. Notice the drop down menu expands

### Relevant log output

```shell

```

jobara:
## Description

resolves #18465 
- allows for the user menu dropdown to be expanded using either the mouse or the keyboard.

## Visual changes

No visual changes

## Functional changes


danharrin (changes requested):
I do still want to use `mousedown` as it is a faster interaction, can you please add a new event listener instead of modifying the existing one?

jobara:
> I do still want to use `mousedown` as it is a faster interaction, can you please add a new event listener instead of modifying the existing one?

@danharrin, I've put `mousedown` back and used `keyup.enter` and `keyup.space` for keyboard access.

danharrin (approved):

PR #18496, opened 2025-11-13
Extended Table empty state documentation with HtmlString example

monkeyphysics:
## Description

The documentation did not mention `Table::emptyState()` can take any Htmlable, which can be a simpler solution for users looking to customize the empty state than a custom view.

## Visual changes

None

## Functional changes


danharrin:
Thanks, but I would rather not suggest that users add raw HTML directly into their PHP classes if they can avoid it.

monkeyphysics:
@danharrin fair point

PR #18503, opened 2025-11-13
Revert "Hide attach files toolbar button when file attachments are disabled"

### LINKED ISSUE #18414: Adding disableToolbarButtons to RichEditor maxes out the CPU
### Package

filament/forms

### Package Version

v4.2.0

### Laravel Version

v12.36.1

### Livewire Version

v3.6.4

### PHP Version

PHP 8.4.12

### Problem description

Adding `disableToolbarButtons` to `RichEditor` appears to be broken, as it maxes out the CPU.

### Expected behavior

CPU not to be maxed out.

### Steps to reproduce

Add `disableToolbarButtons` to `RichEditor` component.

```
return $schema->schema([
    RichEditor::make('notes')
    ->disableToolbarButtons([
        'attachFiles',
        'codeBlock',
        'blockquote',
    ]),
]);
```

https://github.com/GeminiDev1/app-filament-issue-4.x

### Relevant log output

```shell

```

dododedodonl:
Reverts filamentphp/filament#18298

Resolves https://github.com/filamentphp/filament/issues/18414 

Verified the loop disappears when reverting line 35 of forms/src/Components/MarkdownEditor.php. Not sure what a proper fix would be to also hide the button when attachments are disabled.

### DIFF
--- a/packages/forms/src/Components/Concerns/InteractsWithToolbarButtons.php
+++ b/packages/forms/src/Components/Concerns/InteractsWithToolbarButtons.php
@@ -168,9 +168,4 @@ trait InteractsWithToolbarButtons
 
         return false;
     }
-
-    public function hasCustomToolbarButtons(): bool
-    {
-        return $this->evaluate($this->toolbarButtons) !== null;
-    }
 }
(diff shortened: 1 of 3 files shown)


danharrin:
Sorry, I'm not merging a PR that breaks existing behaviour like this, I would merge a fix to the underlying issue though.

dododedodonl:
> Sorry, I'm not merging a PR that breaks existing behaviour like this, I would merge a fix to the underlying issue though.

@danharrin I strongly disagree, the original PR breaks the previously existing simple behavior of for example `->disableToolbarButtons(['attachFiles'])`. You noticed this possiblility, asked the original author to fix this but they didn’t (correctly). Reverting the PR resolves a hard crash, while only showing an extra button one can hide manually with the exact statement that causes the crash.

Please reconsider closing this PR.

PR #18504, opened 2025-11-13
Fix for CPU usage when disabling toolbar buttons

### LINKED ISSUE #18414: Adding disableToolbarButtons to RichEditor maxes out the CPU

Saracaen:
## Description

This should prevent the infinite loop noted in #18414 when using disableToolbarButtons, by defaulting to true if no custom toolbar buttons have been configured yet. Fix is applied to both RichEditor and MarkdownEditor. No existing behavior is broken.

## Functional changes


danharrin:
I'm a little confused as to why this isn't caught by `hasCustomToolbarButtons`, what do you think?

Saracaen:
Yeah you are right, it should be caught by hasCustomToolbarButtons. I have a different idea for a potential fix, but I need to test it tonight to see what it does. My idea is to create a separate method to check if the toolbar buttons contains 'attachFiles', combined with a bool flag to detect/prevent recursion. In the case of recursion we should just return true.

Will update the PR and remove draft status once the changes are made + tested.

danharrin (approved):

dododedodonl:
Thank you for resolving this issue!

PR #18513, opened 2025-11-15
Optimize `IdentifyTenant` to reuse existing tenant if set

binaryfire:
Applications that resolve tenants via custom middleware currently can't avoid Filament re-querying the same tenant from the db.

This PR checks if a tenant has already been set and matches the expected tenant id. If yes, the existing tenant is used. This allows using `Filament::setTenant()` in custom upstream tenant resolution middleware to set the Filament tenant using an already resolved model.

Benefits:
- Allows custom tenant resolution middleware to be reused by Filament
- Saves 1 db query per request if tenant has already been resolved upstream
- Maintains all existing security checks (canAccessTenant, route validation)
- Backward compatible - no impact when tenant isn't pre-set

### DIFF
--- a/packages/panels/src/Http/Middleware/IdentifyTenant.php
+++ b/packages/panels/src/Http/Middleware/IdentifyTenant.php
@@ -29,7 +29,7 @@ class IdentifyTenant
             abort(404);
         }
 
-        $tenant = $panel->resolveTenantForRequest($request->route()->parameter('tenant'));
+        $tenant = $panel->getTenant($request->route()->parameter('tenant'));
 
         if (! $user->canAccessTenant($tenant)) {
             abort(404);
--- a/packages/panels/src/Http/Middleware/IdentifyTenant.php
+++ b/packages/panels/src/Http/Middleware/IdentifyTenant.php
@@ -29,18 +29,18 @@ class IdentifyTenant
             abort(404);
         }
 
-        $tenantRouteKey = $request->route()->parameter('tenant');
-        $slugAttribute = $panel->getTenantSlugAttribute();
+        $tenantRouteParameter = $request->route()->parameter('tenant');
+        $tenantSlugAttribute = $panel->getTenantSlugAttribute();
         $currentTenant = Filament::getTenant();
 
-        $currentTenantKey = filled($slugAttribute)
-            ? $currentTenant?->getAttribute($slugAttribute)
+        $currentTenantIdentifier = filled($tenantSlugAttribute)
+            ? $currentTenant?->getAttribute($tenantSlugAttribute)
             : $currentTenant?->getRouteKey();
 
-        if ($currentTenant && $currentTenantKey === $tenantRouteKey) {
+        if ($currentTenant && $currentTenantIdentifier === $tenantRouteParameter) {
             $tenant = $currentTenant;
         } else {
-            $tenant = $panel->getTenant($tenantRouteKey);
+            $tenant = $panel->getTenant($tenantRouteParameter);
             Filament::setTenant($tenant);
         }
 
--- a/packages/panels/src/Http/Middleware/IdentifyTenant.php
+++ b/packages/panels/src/Http/Middleware/IdentifyTenant.php
@@ -29,24 +29,13 @@ class IdentifyTenant
             abort(404);
         }
 
-        $tenantRouteParameter = $request->route()->parameter('tenant');
-        $tenantSlugAttribute = $panel->getTenantSlugAttribute();
-        $currentTenant = Filament::getTenant();
-
-        $currentTenantIdentifier = filled($tenantSlugAttribute)
-            ? $currentTenant?->getAttribute($tenantSlugAttribute)
-            : $currentTenant?->getRouteKey();
-
-        if ($currentTenant && $currentTenantIdentifier === $tenantRouteParameter) {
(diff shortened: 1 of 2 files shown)


packages/panels/src/Http/Middleware/IdentifyTenant.php
danharrin, line 1: Can you maybe just replace this middleware in the container? Idk, the route key check makes me feel weird. What if the tenant is identified based on domain, for example?

danharrin (changes requested):

packages/panels/src/Http/Middleware/IdentifyTenant.php
binaryfire: @danharrin This works for both path and domain-based tenancy because they both use route parameters. Domain patterns like `{tenant:slug}.example.com` create a route parameter called `tenant` that gets extracted via `$request->route()->parameter('tenant')`, just like path-based routing.

From `packages/panels/routes/web.php`:
```php
if (filled($tenantDomain)) {
    $routeGroup->domain($tenantDomain);  // e.g., {tenant:slug}.example.com or {tenant:domain}
} else {
    $routeGroup->prefix(/* ... */ . '{tenant' . /* ... */ . '}');  // e.g., /admin/{tenant}
}
```

Source: https://github.com/filamentphp/filament/blob/6fb113fb623d6e093a8f466a4eb7d9af323f837b/packages/panels/routes/web.php#L173

I've renamed the variable to `$tenantRouteParameter` to make it clearer.  I think this is a pretty straightforward change - the logic is exactly the same (the `tenant` route param is already what's used for both path and domain based resolution). This just checks if there's already a tenant set, and providing it matches the route param (a failsafe in case the wrong tenant was set upstream), uses it instead of hitting the DB again.

danharrin: I don't know, I'm still weird about it to be honest. Does replacing the middleware in the container work?

danharrin (changes requested):

packages/panels/src/Http/Middleware/IdentifyTenant.php
binaryfire: @danharrin No worries. Yeah I can swap it in the container. I'll have to keep an eye on it in case the core tenancy logic changes but that's fine. Just thought it'd be useful for anyone using a global tenancy package like `archtechx/tenancy` or `spatie/laravel-multitenancy`.

Take Spatie's package for example. Their middleware resolves the tenant once per request and memoizes it in the container:

```php
// Spatie's Multitenancy class
protected function determineCurrentTenant(): void
{
    $tenant = $tenantFinder->findForRequest($request); // DB query
    $tenant->makeCurrent(); // Stores in container
}

// Later retrieval is memoized
public static function current(): ?static
{
    return app($containerKey); // No DB query - returns cached instance
}
```

With this PR, you could add this to some custom middleware:
```php
public function handle(Request $request, Closure $next)
{
    if ($spatieCurrentTenant = \Spatie\Multitenancy\Contracts\IsTenant::current()) {
        \Filament\Facades\Filament::setTenant($spatieCurrentTenant);
    }
    return $next($request);
}
```

Then Filament's `IdentifyTenant` sees the tenant is already set, verifies it matches the route param, and reuses it. 1 DB query total.

Without this PR, there's no way to reuse the already-resolved tenant - Filament will always query it again. 2 DB queries. The only workaround is swapping `IdentifyTenant` in the container in every app that uses both packages. Or accepting the extra query on every request. 

But yeah, totally understand if you'd prefer people handle this themselves. Just seemed like a small change that'd be beneficial to others.

binaryfire: @danharrin Or what about an `->identifyTenant()` panel method?

binaryfire: @danharrin Just pushed a new version that uses a panel method

packages/panels/src/Panel/Concerns/HasTenancy.php
```diff
@@ -42,6 +42,8 @@ trait HasTenancy
 
     protected ?string $tenantOwnershipRelationshipName = null;
 
+    protected ?Closure $tenantIdentifier = null;
```
danharrin: Please call this `$resolveTenantUsing`

packages/panels/src/Panel/Concerns/HasTenancy.php
```diff
@@ -137,6 +139,13 @@ public function tenantRegistration(?string $page): static
         return $this;
     }
 
+    public function identifyTenant(?Closure $callback): static
```
danharrin: Please call this `resolveTenantUsing()`

packages/panels/src/Panel/Concerns/HasTenancy.php
```diff
@@ -202,6 +211,17 @@ public function getTenantRegistrationPage(): ?string
         return $this->tenantRegistrationPage;
     }
 
+    public function resolveTenantForRequest(string $key): Model
```
danharrin: Please merge this into the existing `getTenant()` method so we don't need another

danharrin (changes requested):

binaryfire:
Done

danharrin (approved):

PR #18515, opened 2025-11-15
Fix Htmlable icons support

zepfietje:
## Description

Changes that were missed in #18500.

## Functional changes


packages/support/resources/css/components/icon.css
danharrin, line 1: Please explain each of these changes and which situations they are needed in

danharrin (changes requested):

packages/support/resources/css/components/icon.css
zepfietje: The SVG element needs to inherit the parent span's size since otherwise it would be whatever size is defined by its viewBox attribute.

Even with just that change, when the SVG element is sized properly, the span element seems to grow larger than the defined width and height. Hence the max width and max height properties make sure the width and height are actually restricted.

danharrin: Can you send an example of the SVG where it doesnt inherit the size properly? `.fi-icon` should be applied to the SVG itself so I am confused.

zepfietje: Any SVG with a large viewBox should make the issue obvious, e.g.:
```svg
<svg fill="currentColor" viewBox="0 0 256 256"><path d="M230.92,212c-15.23-26.33-38.7-45.21-66.09-54.16a72,72,0,1,0-73.66,0C63.78,166.78,40.31,185.66,25.08,212a8,8,0,1,0,13.85,8c18.84-32.56,52.14-52,89.07-52s70.23,19.44,89.07,52a8,8,0,1,0,13.85-8ZM72,96a56,56,0,1,1,56,56A56.06,56.06,0,0,1,72,96Z"></path></svg>
```

The `fi-icon` class is applied to the wrapping `<span>` in fact, not the SVG element inside when using Htmlable icons, e.g.:
```php
new HtmlString('<svg fill="currentColor" viewBox="0 0 256 256"><path d="M230.92,212c-15.23-26.33-38.7-45.21-66.09-54.16a72,72,0,1,0-73.66,0C63.78,166.78,40.31,185.66,25.08,212a8,8,0,1,0,13.85,8c18.84-32.56,52.14-52,89.07-52s70.23,19.44,89.07,52a8,8,0,1,0,13.85-8ZM72,96a56,56,0,1,1,56,56A56.06,56.06,0,0,1,72,96Z"></path></svg>')
```

packages/support/resources/css/components/icon.css
```diff
@@ -1,27 +1,31 @@
...
-        @apply size-7;
+        @apply size-7 max-h-7 max-w-7;
     }
 
     &.fi-size-2xl {
-        @apply size-8;
+        @apply size-8 max-h-8 max-w-8;
+    }
+
+    svg {
```
danharrin:
```suggestion
    & > svg {
```

danharrin (changes requested):

danharrin (approved):

PR #18519, opened 2025-11-16
feat: Improve panel extensibility by supporting boot-level registration

alancolant:
## Description

I encountered a limitation when working on a scenario where Filament panels must be registered dynamically **during the application boot process**, based on the result of an Eloquent query.  
The goal was to allow external service providers to **extend or register panels conditionally**, only once database connectivity and runtime context were fully available.

However, Filament currently initializes and loads its panel routes **inside the `boot()` method** of `FilamentServiceProvider`.  
This design forces panel registration to happen **only in `register()`**, otherwise any panel registered later in the lifecycle — such as in `app()->booted()` — **will not receive routes**, since routing has already been processed at that stage.

This makes it impossible to build advanced or modular Filament extensions where panels:
- Depend on database content
- Need dynamic evaluation
- Are defined by external packages or plugins
- Must rely on runtime logic instead of static configuration

Since the database connection and models are not reliably available during service provider `register()` execution, I was unable to perform conditional logic and therefore **could not register panels in a valid state without losing their routing**.

## Resolution

The routing logic has been **moved out of the boot phase and is now executed at the moment a panel is registered**, meaning routes are generated directly when calling `app(PanelsRegistry::class)->register()` .  

This guarantees that route creation happens as soon as the panel is added — no matter if this takes place in `register()`, `boot()`, or even `app()->booted()`.

## Functional changes

- Updated `Panel.php` and `HasRoutes.php` 
- Removed the previous `routes/web.php` file and moved its logic into `HasRoutes` 
- Updated the `Filament` facade to automatically register a panel if the `PanelRegistry` has already been resolved:
  ```php
  if (static::getFacadeApplication()->resolved(PanelRegistry::class) === true) {
      app(PanelRegistry::class)->register(value($panel));
  }
  ```


danharrin:
Sorry, I don't think I can merge this. Before this service provider `boot()` runs, if someone attempts to generate a URL for a panel (using `route()`) and it is not registered, it will now throw an error, which is a breaking change.

What part of your panel config needs to be database-dependent? Let's work through that idea instead.

alancolant:
> Sorry, I don't think I can merge this. Before this service provider `boot()` runs, if someone attempts to generate a URL for a panel (using `route()`) and it is not registered, it will now throw an error, which is a breaking change.
> 
> What part of your panel config needs to be database-dependent? Let's work through that idea instead.


I have a modular system where each module is implemented as a provider and can be enabled or disabled individually, with its activation state stored in the database.

This setup is designed for a multi-tenant environment in which the tenant system acts as a dynamic provider bootstrapper to generate customized dynamic panels for each tenant, like injecting plugin)

danharrin:
You should probably use authorisation checks for this, in my opinion the set of routes registered should not be affected by the runtime of your app. It breaks route caching.

alancolant:
I had thought about that, but it means each plugin has to handle it, which isn’t the case for external plugins.

Another idea I had would be to completely re-bootstrap the application once the tenant is detected, so that the configuration can be injected very early in the cycle — but I was hoping to avoid needing that.

danharrin:
I think you'll just need to contribute authorisation to those external plugins. I really don't think changing the bootstrapping of the app is a good idea.

PR #18522, opened 2025-11-16
Method sorting rector

danharrin:


danharrin:
I don't think I am going to merge this into core for now as I don't think that there is a universally agreed upon order yet. If anyone in the community sees this and is interested, feel free to use this PR code as a starting point to create a community formatting package. I probably introduce an official implementation in the future, but I do not know if that will be within the next year or the next five years.

PR #18532, opened 2025-11-17
FIX: Remove toLowerCase on query in handleSearch (select.js)

### LINKED ISSUE #18306: Lowercase on Select search query
### Package

filament/filament

### Package Version

v4.1.10

### Laravel Version

v12.*

### Livewire Version

v3

### PHP Version

PHP 8.4.0

### Problem description

In the file `packages/support/resources/js/utilities/select.js`, we have this line:

```js
    handleSearch(event) {
        const query = event.target.value.trim().toLowerCase()
        this.searchQuery = query
```

This lowercasing of the query causes a couple of problems;

1. In a database with a case-sensitive character set, this won't work (not common, but possible)
2. It prevents the ability to do customizations on creating results with the search query.

Point two is the problem we're experiencing. We want to provide the custom option to "Create New Record" which will take the search query and create a record with that name. But the lowercasing of the search term is interfering with this.

If this isn't necessary, could we change this to not be lowercased?

### Expected behavior

The search comes into Laravel Backend the same case as typed

### Steps to reproduce

1. Create a new form with a select component
2. Make the Select component searchable
3. Add a custom getSearchResultsUsing
4. Log the output of `$search`
5. See that it is lowercased.

https://github.com/olliescase/filament-select-issue/blob/main/app/Filament/Pages/Auth/Login.php

### Relevant log output

```shell
[2025-10-22 13:24:50] local.DEBUG: Searching for: 'one'
```

pddevins:
## Description
Update handleSearch in select.js to not automatically lowercase the query.

Issue Reference : [Lowercase on Select search query](https://github.com/filamentphp/filament/issues/18306)

## Visual changes

Before: (search term is printed in dropdown using `getSearchResultsUsing`)
_String is transformed to lowercase_
Screenshot 2025-11-17 at 3 30 49 PM

After: (search term is printed in dropdown using `getSearchResultsUsing`)
_String maintains capitalization in original search query_
Screenshot 2025-11-17 at 3 47 46 PM

## Functional changes
The query term in a Select search is no longer converted to lowercase at the beginning of `handleSearch`. The query string is still converted to lowercase before comparisons with other lowercased options in `filterOptions`. Developers can use `strtolower` (or any other function/helper to modify capitalization) in `getSearchResultsUsing` to maintain the original functionality.


danharrin (approved):

danharrin:
Beautiful, thanks.

PR #18538, opened 2025-11-19
[4.x] Add `unsaved-changes-alert-setup` component for better reuse outside of panel

stefanzweifel:
## Description

This PR adds a new `<x-filament-panels::unsaved-changes-alert-setup />` Blade component, which contains the setup code for the [Unsaved Changes Alerts](https://filamentphp.com/docs/4.x/panel-configuration#unsaved-changes-alerts) feature.

We use the "Unsaved Changes Alerts" in our admin panel, but also use Filament Forms in our frontend and would like to bring the feature to our users as we think it's very convenient.

By extracting the logic into a component, we can simply add the new Blade component into our views and use the feature immediately.

## Functional changes

### DIFF
--- a/packages/panels/resources/views/components/page/index.blade.php
+++ b/packages/panels/resources/views/components/page/index.blade.php
@@ -147,25 +147,7 @@
 
     {{ \Filament\Support\Facades\FilamentView::renderHook(\Filament\View\PanelsRenderHook::PAGE_END, scopes: $this->getRenderHookScopes()) }}
 
-    @if (method_exists($this, 'hasUnsavedDataChangesAlert') && $this->hasUnsavedDataChangesAlert())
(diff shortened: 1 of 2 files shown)


danharrin:
Thanks, I don't think I want to make this a feature outside of panels though, then I have to deal with all the situations where it might break in an unpredictable setup. It also isn't really that convenient to use in its current form, it requires `hasUnsavedDataChangesAlert` on the Livewire component for example.

stefanzweifel:
Understandable. The need for `Filament\Pages\Concerns\HasUnsavedDataChangesAlert` on the component was already clear to me and is what we use aswell. (This also links the `unsavedChangesAlerts()`-setting on the panel with a Livewire component outside of the panel; not ideal.

Thanks for the quick feedback. ❤️

PR #18539, opened 2025-11-19
[4.x] Fix logic error in Select dynamic options check

gehrisandro:
## Description

This PR fixes an error in the logic to determine if a select has dynamic options.

Behavior before:
When a select has a relationship, is searchable and options are preloaded, the options are correctly preloaded into the DOM, but when opening the dropdown, the wrongly implemented check forces the options to reload.

New behavior:
Options are still correctly loaded into the DOM and used when the dropdown opens without reloading.

## Visual changes

No visual changes.

## Functional changes

- [ ] ~~Documentation is up-to-date.~~ No changes required

### DIFF
--- a/packages/forms/src/Components/Select.php
+++ b/packages/forms/src/Components/Select.php
@@ -1308,7 +1308,7 @@ class Select extends Field implements Contracts\CanDisableOptions, Contracts\Has
         }
 
         if ($this->hasRelationship()) {
-            return $this->isPreloaded();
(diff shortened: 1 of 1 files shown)


danharrin:
I don't think this is correct, the preloaded options could be scoped by a `$get` or similar, so they should reload. The condition is there so that they aren't fetched if they aren't preloaded, as the options array should be empty in that case.

PR #18550, opened 2025-11-20
Add translateLabel to ExportColumn

TheSkybreaker:
## Description

The ExportColumn does not have a `translateLabel` method. It simply adds that to translate column names

## Functional changes

- [ ] Documentation is up-to-date. - I can add to the label example if approved


danharrin:
Thanks for the PR, but I do not really want this feature in Filament. I wish I never allowed `translateLabel()` into Filament in the first place for forms and tables, since it encourages really bad translation practices. Structured translation keys are best to ensure the correct translation context is always used. Sorry.

PR #18551, opened 2025-11-20
Add shouldDispatchNotificationEvent method to exporter

TheSkybreaker:
## Description

When an export is completed it sends a notification to the user and writes a line on the log. While it can be useful, when using external channels for the logs it can be strange or bothersome to receive a message when an user completes an export (even more when there are a lot of them)
With this it can be controlled from the exporter if the notification should dispatch the DatabaseNotificationSent event.
Default is true to not change previous behaviour

## Visual changes

//

## Functional changes

- [ ] Documentation is up-to-date. - Can add to the documentation if and when approved

### DIFF
--- a/packages/actions/src/Exports/Exporter.php
+++ b/packages/actions/src/Exports/Exporter.php
@@ -242,4 +242,9 @@ abstract class Exporter
     {
         return $query;
     }
+
+    public static function shouldDispatchNotificationEvent(Export $export): bool
+    {
+        return true;
+    }
 }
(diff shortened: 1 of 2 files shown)


danharrin:
In this case, I think I would suggest that you just set `BROADCAST_CONNECTION=null` in your `.env`, so that broadcast events don't log. Let me know if this doesn't work for you and I can reconsider.

TheSkybreaker:
While for me personally COULD work, doing it with a method can be used in the case there's the need to broadcast messages, but not this specific one!

PR #18571, opened 2025-11-21
un-deprecate mountTableAction

robertvansteen:
## Description
This removes the `@deprecation` on the [testing stubs for `mountTableAction`](https://github.com/filamentphp/filament/blob/d7b51314c6c3000b7879ccc11a33185fca1edf15/packages/tables/.stubs.php#L83). It proposes to use the alternative `mountAction` however with that one it's not possible to pass the required context `['table' => true]` making it not a suitable alternative. I opted to remove the deprecation as I suspect this might be incorrect as the [underlying implementation](https://github.com/filamentphp/filament/blob/d7b51314c6c3000b7879ccc11a33185fca1edf15/packages/tables/src/Testing/TestsActions.php#L22) on `TestsActions` is not deprecated.

Appreciate this might not be the correct direction, happy to update `mountAction` to support passing context if the deprecation was intended.

## Visual changes

N/A

## Functional changes

### DIFF
--- a/packages/tables/.stubs.php
+++ b/packages/tables/.stubs.php
@@ -77,14 +77,8 @@ namespace Livewire\Features\SupportTesting {
 
         public function loadTable(): static {}
 
-        /**
-         * @deprecated Use `mountAction()` instead.
-         */
         public function mountTableAction(string | array $name, $record = null): static {}
 
-        /**
-         * @deprecated Use `unmountAction()` instead.
-         */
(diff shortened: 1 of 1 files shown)


danharrin:
Please see https://filamentphp.com/docs/4.x/testing/testing-actions#testing-table-actions, you should be able to use `mountAction(TestAction::make('create')->table())` etc. Please let me know if this doesn't work.

PR #18580, opened 2025-11-22
Adding details about notifications and export format

thomasLecler:
## Description

This PR updates the actions/export documentation page :

- Filament export requires notifications to be setup in a panel to be visible in common cases
- All formats are generated when none is provided, then it can be unexpected on big dataset

### DIFF
--- a/packages/actions/docs/12-export.md
+++ b/packages/actions/docs/12-export.md
@@ -6,9 +6,7 @@ import UtilityInjection from "@components/UtilityInjection.astro"
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


packages/actions/docs/12-export.md
```diff
@@ -337,7 +339,7 @@ ExportColumn::make('users_avg_age')
 
 ## Configuring the export formats
 
-By default, the export action will allow the user to choose between both CSV and XLSX formats. You can use the `ExportFormat` enum to customize this, by passing an array of formats to the `formats()` method on the action:
+By default, the export action will generate both CSV and XLSX formats and allow user to choose between it in the notification. You can use the `ExportFormat` enum to customize this, by passing an array of formats to the `formats()` method on the action:
```
danharrin:
```suggestion
By default, the export action will generate both CSV and XLSX formats and allow user to choose between them in the notification. You can use the `ExportFormat` enum to customize this, by passing an array of formats to the `formats()` method on the action:
```

danharrin (changes requested):

danharrin (approved):

PR #18591, opened 2025-11-24
Rich Editor: Floating toolbar JS extensions

thijskuilman:
## Description

This PR introduces a new method: `->floatingToolbarJsExtensions()`, which allows fully customizing the BubbleMenuPlugin per node type in the `RichEditor`.

**Why is this feature useful?**
I had a RichEditor like:

```php
RichEditor::make('content')
    ->floatingToolbars([
        'paragraph' => [
            'bold', 'italic', 'underline'
        ],
    )
```

With this, I noticed the `paragraph` toolbar is always appearing, even when no text is selected:

afbeelding

While this behavior works for most cases, I wanted the toolbar to appear only when text is selected. I realized there is no way to customize this behavior.

**Why did I pick this solution?**
I considered multiple approaches:
- **Introduce a boolean** to control whether certain floating toolbars only appear on text selection.
_Disadvantage_: This could become messy as more configuration options are added in the future.
- **Always require text selection for `paragraph` toolbars**.
_Disadvantage_: Not all users might want this behavior.

Ultimately, I figured the best solution is to allow custom JavaScript extensions. This gives full flexibility to define a custom `BubbleMenuPlugin` per node type. For example, with this feature I have customized the floating toolbar for a paragraph node like this:

```php
 RichEditor::make('content')
    ->floatingToolbarJsExtensions([
        FilamentAsset::getScriptSrc('extensions/custom-floating-paragraph-toolbar'),
    ])
    ->floatingToolbars([
        'paragraph' => [
            'bold', 'italic', 'underline'
        ],
    )
```

```javascript
import { BubbleMenuPlugin } from '@tiptap/extension-bubble-menu';

export const pluginKey = 'paragraph';

export default ({
    editor,
    element
}) => {
    return BubbleMenuPlugin({
        editor,
        element,
        pluginKey,
        shouldShow: ({ editor }) => {
            return editor.isFocused && !editor.state.selection.empty; // Only show on text selection
        },
        updateDelay: 50, // Decreased update delay, so the toolbar feels snappier
        options: {
            placement: 'top', // Changed the positioning
            offset: 15,
        },
    });
}
```

## Visual changes
None

## Functional changes


danharrin (changes requested):
> Always require text selection for paragraph toolbars.

I think this is the best path forward for now, can you please implement that instead?

thijskuilman:
> > Always require text selection for paragraph toolbars.
> 
> I think this is the best path forward for now, can you please implement that instead?

Sure thing! I created a new PR for that change, see [github.com/filamentphp/filament/pull/18675.](https://github.com/filamentphp/filament/pull/18675). I will close this one.

PR #18596, opened 2025-11-25
[4.x] Table of Contents for the Rich Editor

SOD96:
## Description

As this is replacing awcodes tiptapeditor, there are a few extra functions which it did. I believe these would be best placed being brought over. We require them in our workflow, so figured it's easier to add to this than try my own frankenstein dom searching maneuvers.

When you generate the HTML content you don't really have any way of chaining to get additional output, same as it worked on awcodes version, so largely this works the same with a couple of changes. You can call the Table of Contents method and you will get an ARRAY back only, it's up to you to style and implement. Previously he would provide the html, we're not doing that here. Similar to my previous work on the toArray.

In essence, all this does is add the Id Extension so we can automatically add id's to H2's, H3's etc. This allows them to then be anchored, put in a link and referenced on a page. I had a version working that added a # beside the Header so you could quick link to it - but I removed that as I figured that's heavily opinionated. 

This allows you to implement your own table of contents at the top of the page, that quick links to other sections. It's pretty neat! You can develop your own self referencing Table of Contents blocks separately to this functionality to actually insert it into the content

For content heavy pages, a table of contents is critical. I was adding a unique prefix to the headers, but I feel like that might cause issues maybe in regards to on-page SEO. So I've just left it as the header text. Happy to implement whatever works though.

I've also updated the docs to have this in it, and and the toArray functionality so people know it's here.

Any recommendations please let me know! 

## Visual changes

Example Implementation:
Screenshot 2025-11-25 at 12 38 30

## Functional changes

### DIFF
--- a/packages/forms/docs/10-rich-editor.md
+++ b/packages/forms/docs/10-rich-editor.md
@@ -51,6 +51,19 @@ class Post extends Model
 }
 ```
 
+### Output to Array
+Rather than working with the HTML string that we provide, you can also output to an array representation of the content by using the `toArray()` method:
+
+```php
+use Filament\Forms\Components\RichEditor\RichContentRenderer;
+
+$content = RichContentRenderer::make($record->content)->toArray()
+```
+
+You can then iterate through the `$content['content']` array to access the individual nodes.
+
+This can be useful if you want to manipulate the content in some way before rendering it, or if you want to output it in a different format (e.g., JSON for an API).
+
 ## Customizing the toolbar buttons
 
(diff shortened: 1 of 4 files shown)


danharrin:
Thanks for the suggestion, but I do not consider this to be a core feature of the rich content framework currently, and quite a large amount of code to maintain. You are able to inject your own extensions into the renderer though, and also use `processNodesUsing()` with your own custom function to transform the state though. You could try building it into a package using a macro, even.

Issue #7087, opened 2023-07-21
Native date time picker placeholder color is wrong in Safari


zepfietje:
https://github.com/tailwindlabs/tailwindcss-forms/issues/38

nathanheffley:
This issue seems to be recurring in v4, where the placeholder color is far too dark in light mode on Safari. These are two date inputs, one with a value selected and one with no value.

<img width="823" height="93" alt="Image" src="https://github.com/user-attachments/assets/e6ff39a1-8f6b-4f5b-ab38-8507d456bfd9" />

PR #18597, opened 2025-11-25
Fix Safari date picker placeholder color

nathanheffley:
## Description

The original fix in [0a2b41](https://github.com/filamentphp/filament/commit/0a2b41b73293d81d8031f0e25666a18ed9145778) is no longer working for Safari, which seems to have fixed the bug that fully transparent white is different from transparent, meaning that the issue described in #7087 is now recurring, at least in light mode.

This PR fixes that by making the fully transparent white ever so slightly less transparent (1%) so that there is no visual change but Safari renders the placeholder correctly.

## Visual changes

Before (light mode):

Screenshot 2025-11-25 at 12 24 03

After (light mode):

Screenshot 2025-11-25 at 12 22 03

Before (dark mode):

Screenshot 2025-11-25 at 12 23 56

After (dark mode):

Screenshot 2025-11-25 at 12 22 40

## Functional changes


danharrin (approved):

danharrin:
Thanks!

PR #18611, opened 2025-11-27
Update return type of StatsOverviewWidget

jsandfordhughescoop:
## Description

The `getStats` method on the `StatsOverviewWidget` component had a return type of `array<Stat>` however any component can be rendered within the `getStats` array.

## Visual changes

N/A

## Functional changes


danharrin:
While it *can* return any component, I can't guarantee that won't break in the future to be honest. I don't think its necessarily expected behaviour, so I don't want to suggest someone can do that. Also, having `Stat` in the docblock is a good way to show users how to use the method without them needing to check the docs.

PR #18615, opened 2025-11-28
Update 06-contributing.md

faisuc:
Updated documentation for running the translation tool.

### DIFF
--- a/docs/01-introduction/06-contributing.md
+++ b/docs/01-introduction/06-contributing.md
@@ -76,7 +76,7 @@ git clone git@github.com:filamentphp/filament.git
 composer install
 
 # Run the tool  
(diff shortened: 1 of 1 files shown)


danharrin (changes requested):
The file has `#! /usr/bin/env php` at the top so I don't think we need this, thoughts @pxlrbt?

pxlrbt:
@danharrin I just double checked with a fresh clone of the repo and that works, too. Windows doesn't support Shebangs, though. But the scripts uses Laravel Prompts which don't work on Windows anyway.

danharrin:
Thanks, seems like we shouldn't update the docs for Windows if the script doesnt work on Windows anyway.

PR #18621, opened 2025-11-28
Improve API of "counts" text columns

ropi-bc:
## Description

Allow the "relationshipToCount" to have boolean values. If this value is set to true, then infer the relationship based on column-name.

Introduce a new field called "attribute" which decouples the filament column's name, from the database column/model attribute. Default means original behavior.

Instead of:
```php
TextColumn('posts_count')->counts('posts');
```

You can now do:
```php
TextColumn('posts')->counts();
```

## Visual changes

No visual changes

## Functional changes

### DIFF
--- a/packages/support/src/Concerns/CanAggregateRelatedModels.php
+++ b/packages/support/src/Concerns/CanAggregateRelatedModels.php
@@ -14,9 +14,9 @@ trait CanAggregateRelatedModels
     protected string | array | Closure | null $relationshipToAvg = null;
 
     /**
-     * @var string | array<int | string, string | Closure> | Closure | null
+     * @var bool | string | array<int | string, string | Closure> | Closure | null
      */
-    protected string | array | Closure | null $relationshipsToCount = null;
+    protected bool | string | array | Closure | null $relationshipsToCount = null;
 
     /**
      * @var string | array<int | string, string | Closure> | Closure | null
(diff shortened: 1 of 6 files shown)


packages/support/src/Concerns/CanAggregateRelatedModels.php
```diff
@@ -56,12 +56,16 @@ public function avg(string | array | Closure | null $relationship, string | Clos
...
     /**
-     * @param  string | array<int | string, string | Closure> | Closure | null  $relationships
+     * @param  bool | string | array<int | string, string | Closure> | Closure | null  $relationships
      */
-    public function counts(string | array | Closure | null $relationships): static
+    public function counts(bool | string | array | Closure | null $relationships = true): static
     {
         $this->relationshipsToCount = $relationships;
 
+        if ($relationships === true) {
```
wychoong, line 65: doesnt work with `->counts(fn() => true)`

ropi-bc: is it supposed to? I don't think there's a usecase for that, because if it's possible that you don't want the count to be true, then you'll need to change the name to `..._count` anyway

danharrin:
Thanks, but I don't think I'm going to move forward with this. I don't think the existing API is *too* bad and introducing a new one could just cause confusion at this point. I might rethink this in the future.

PR #18626, opened 2025-11-30
fix: remove duplicate column in sql query on manual sorting

### LINKED ISSUE #18299: [v4.x] Duplicate column in SQL query on manual sorting
### Package

filament/filament

### Package Version

V4.1.9 

### Laravel Version

v12.34.0

### Livewire Version

v3.0

### PHP Version

PHP 8.3.0

### Problem description

using SQL Server!
Important. this isssue does NOT exist on sqlite as it accepts duplicate column names in the ORDER_BY sql query but SQL server does not and probably mysql or Mariadb will have this issue

When sorting manually on a column, Filament has a duplicate column name in the query

In the Resource file (OrdersResource for example) when you have set a TextColumn sortable() and also have defaultSort() for that SAME column, the first List load is fine but then sorting manually on that column will generate an SQL error 

SQLSTATE[42000]: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]A column has been specified more than once in the order by list. Columns in the order by list must be unique.
select top 10 * from [orders] order by [created_at] asc, [created_at] desc, [orders].[id] desc

If you remove defaultSort(), sorting manually works
The problem is the sorted column name is duplicate in the SQL query whicj is not allowed
in the example above you have "[created_at] asc, [created_at] desc" in the query

### Expected behavior

return the List ordered on the column I want to sort

### Steps to reproduce

Needs other database than sqlite!

In the resource file have defaultSort() with a column name that is sortable(). For example 

                    ->dateTime(config('filament.datetime_format'))
                    ->sortable(),

and have ->defaultSort('created_at)

Load the list end them manually sort on the column created_at
The same error is generated if the TEXTCOLUMN What is sortable() and defaultSort for that same column
TextColumn::make('name')->sortable() and defaultsort('name'), click the columnt to sort and it will generate the error

https://github.com/VincentDeVilder/app-filament-issue-4.x

### Relevant log output

```shell
[2025-10-21 13:52:14] local.ERROR: SQLSTATE[42000]: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]A column has been specified more than once in the order by list. Columns in the order by list must be unique. (Connection: sqlsrv, SQL: select top 10 * from [orders] order by [created_at] asc, [created_at] desc, [orders].[id] desc) {"view":{"view":"/var/www/html/vendor/filament/tables/resources/views/index.blade.php","data":[]},"userId":1,"exception":"[object] (Spatie\\LaravelIgnition\\Exceptions\\ViewException(code: 0): SQLSTATE[42000]: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]A column has been specified more than once in the order by list. Columns in the order by list must be unique. (Connection: sqlsrv, SQL: select top 10 * from [orders] order by [created_at] asc, [created_at] desc, [orders].[id] desc) at /var/www/html/vendor/laravel/framework/src/Illuminate/Database/Connection.php:824)
```

giacomomasseron:
## Description

When you have a defaultSort defined, and sorting manually on the same column, Filament has a duplicate column name in the query.

The code deleted is not needed because the default sort is applied anyway in the lines 88-95 of the same file.

## Functional changes


danharrin (commented):
Instead of deleting the entire block, check to see if the default sort string is the same as the normal sort string or not. We still need it if it is different.

giacomomasseron:
@danharrin I reintroduced the block and added the check

danharrin (approved):

PR #18631, opened 2025-12-01
feat: Add `persistColumnsInSession()` method.

Zamion101:
## Description

This adds a new method to `HasColumnManager` to conditionally configure the persistence of columns in user's session . 

Currently Filament stores the columns of a table in session regardless if the `SESSION_DRIVER` set to `file`, `cookie` etc. 
When the `SESSION_DRIVER` set to `cookie` this approach results to cookie size exceeding the allowed `4096` characters limit. 
With the addition of `persistColumnsInSession()` method this approach can be modified to not store the columns in session.

## Functional changes


danharrin (changes requested):
Please resolve the merge conflicts

Zamion101:
> Please resolve the merge conflicts

@danharrin should be resolved.

packages/tables/src/Concerns/HasColumnManager.php
```diff
@@ -152,10 +152,13 @@ protected function loadTableColumnsFromSession(): array
 
     protected function persistTableColumns(): void
     {
-        session()->put(
-            $this->getTableColumnsSessionKey(),
-            $this->tableColumns
-        );
+        $shouldPersistColumns = $this->getTable()->persistsColumnsInSession();
```
archilex: No need to set in a variable since it's not be reused and the method name is clear.

archilex (changes requested):

packages/tables/src/Concerns/HasColumnManager.php
Zamion101: Used it directly in the if statement and removed the `$shouldPersistColumns`.

archilex (approved):

danharrin (approved):

PR #18633, opened 2025-12-02
Update return types for `getTableFilterState` and `getTableFilterFormState`

alexmanase:
## Description

When using `getTableFilterState` and `getTableFilterFormState` methods, the return type can vary beyond just arrays or nulls, especially when custom filter schemas are used. Single-value filters (e.g., select fields) return strings, toggle filters return booleans, and multi-value filters return arrays. The current `?array` return type is too restrictive and causes type errors when filters return non-array values.

## Functional changes


danharrin:
Hey, this doesn't seem right to me. Even if a custom schema is used, it is still an array. There are many places within the codebase that depend on this being the case. The filters form data should be keyed by the names of the filters, and each value should be an array of the form fields just for that filter. This prevents conflicts between filter form fields by keeping them within their own "state path".

alexmanase:
You are right. My use case was that I needed to return a value of a fitler: `$livewire->getTableFilterState('filters.single_select')`. "filters" is a single filter with multiple components.

I will stick to `$livewire->getTableFilterState('filters')` and work with that returned array.

Thank you for the explanation!

Issue #18637, opened 2025-12-02
`CreateAction::make()->successNotification(null)` still shows the success notification on creation


danielh-official:
### Package

filament/filament

### Package Version

v4.2.4

### Laravel Version

v12.40.2

### PHP Version

PHP 8.4.15

### Problem description

This is a continuation of https://github.com/filamentphp/filament/issues/18594, which was cancelled due to not enough details being given.

The success notification was shown even though I disabled it.

### Expected behavior

When I set `CreateAction::make()->successNotification(null)`, it should disable the notification altogether, as stated under https://filamentphp.com/docs/4.x/actions/create#customizing-the-save-notification.

---

To disable the notification altogether, use the successNotification(null) method:

```
use Filament\Actions\CreateAction;

CreateAction::make()
    ->successNotification(null)
```

### Steps to reproduce

1. Clone the linked repository branch
2. Run `composer install`
3. Run `composer setup`
4. Create a user using `php artisan make:filament-user`
5. Run `composer dev`
6. Navigate to /admin/login and log in with the credentials from the user created in step 4
7. Navigate to /admin/customers
8. Click "New customer", enter a name, and click "Create"
9. You should see the "Created" notification at the top right

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/danielh-official/laravel/tree/filament-php-null-success-notification-does-not-disable-success-notification

### Relevant log output

```shell

```

danielh-official:
https://filamentphp.com/docs/4.x/resources/creating-records#customizing-the-save-notification

Found that notification disabling in panels is different from notification disabling in actions. Closing.

danielh-official:
I just switched to the modal version by commenting out the customer creation page in the resource.

See: https://github.com/danielh-official/laravel/commit/391e045234d31d79477a59b7dab777bc01e67115

What I found is that `->successNotification(null)` doesn't cancel the notification, but `->successNotificationTitle(null)` does.

Either I can fix whatever is causing `->successNotification(null)` not to work or we can update the related documentation at https://filamentphp.com/docs/4.x/actions/create#customizing-the-save-notification to specify using `->successNotificationTitle(null)` instead.

PR #18638, opened 2025-12-02
Fix method name for disabling notifications

danielh-official:
## Description

The proposed method for the current code is inaccurate. Unless fixed, the correct method to use is `successNotificationTitle(null)`.

For more context, see: https://github.com/filamentphp/filament/issues/18637

### DIFF
--- a/packages/actions/docs/04-create.md
+++ b/packages/actions/docs/04-create.md
@@ -112,13 +112,13 @@ CreateAction::make()
 
 <UtilityInjection set="actions" version="4.x" extras="Notification;;Filament\Notifications\Notification;;$notification;;The default notification object, which could be a useful starting point for customization.">As well as allowing a static value, the `successNotification()` method also accepts a function to dynamically calculate it. You can inject various utilities into the function as parameters.</UtilityInjection>
 
-To disable the notification altogether, use the `successNotification(null)` method:
+To disable the notification altogether, use the `successNotificationTitle(null)` method:
 
 ```php
 use Filament\Actions\CreateAction;
 
 CreateAction::make()
-    ->successNotification(null)
+    ->successNotificationTitle(null)
 ```
 
 ## Lifecycle hooks


Zamion101:
I don't think this is the correct thing to do @danharrin @danielh-official . 
`->successNotification(null)` should be the correct method to disable the success notification. I will be looking into why it does not work as intended. The problem should be in https://github.com/filamentphp/filament/blob/8f9380a85ef568fefefd8f3ec9134a18933aef50/packages/actions/src/Concerns/CanNotify.php#L119-L132

danielh-official:
> I don't think this is the correct thing to do @danharrin @danielh-official . `->successNotification(null)` should be the correct method to disable the success notification. I will be looking into why it does not work as intended. The problem should be in
> 
> https://github.com/filamentphp/filament/blob/8f9380a85ef568fefefd8f3ec9134a18933aef50/packages/actions/src/Concerns/CanNotify.php#L119-L132

Something like this?

```php
    public function sendSuccessNotification(): static
    {
        if (is_null($this->successNotification)) {
            return $this;
        }

        $notification = $this->evaluate($this->successNotification, [
            'notification' => $notification = Notification::make()
                ->success()
                ->title($this->getSuccessNotificationTitle()),
        ]) ?? $notification;

        if (filled($notification?->getTitle())) {
            $notification->send();
        }

        return $this;
    }
```

@Zamion101

Zamion101:
> Something like this?
> 
> ```
>     public function sendSuccessNotification(): static
>     {
>         if (is_null($this->successNotification)) {
>             return $this;
>         }
> 
>         $notification = $this->evaluate($this->successNotification, [
>             'notification' => $notification = Notification::make()
>                 ->success()
>                 ->title($this->getSuccessNotificationTitle()),
>         ]) ?? $notification;
> 
>         if (filled($notification?->getTitle())) {
>             $notification->send();
>         }
> 
>         return $this;
>     }
> ```

As far as I know `successNotification` is null initialized in the `CanNotify` 
https://github.com/filamentphp/filament/blob/8f9380a85ef568fefefd8f3ec9134a18933aef50/packages/actions/src/Concerns/CanNotify.php#L14

Won't this addition result of no notification being sent? 

I looked at built-in actions and most of the time when they do not want action to create a notification, both `successNotification(null)` and `successNotificationTitle(null)` used. I am suspecting that, this is done because `successNotification(null)` by itself does not worked at that time but nobody looked back to solve the problem. I think there needs to be a change in how the notifications are created (custom success and default notification that sent when there is no successNotification present).

Issue #18529, opened 2025-11-17
Select with searchable option empty state


valpuia:
### Package

filament/filament

### Package Version

4.2.2

### Laravel Version

12

### Livewire Version

3

### PHP Version

8.4

### Problem description

When using `searchable()` with `relationship()` and `preload()` options, it display data for easy access. but when we don't have any data, it display empty state like image. This empty state is same as using `->options()`

<img width="556" height="146" alt="Image" src="https://github.com/user-attachments/assets/481e098d-3fec-44a0-8110-6876ad99a039" />

```
Select::make('district_id')
    ->required()
    ->relationship(name: 'district', titleAttribute: 'name')
    // -> options(District::pluck('name', 'id')->toArray())
    ->searchable()
    ->preload()
```

### Expected behavior

Since we `preload()` the data, if it's empty it should display empty state like below, maybe different message

<img width="556" height="180" alt="Image" src="https://github.com/user-attachments/assets/7a68994a-f155-482f-b977-bab9761d7947" />

### Steps to reproduce

Use select on any resource with relationship and make the related empty

```
Select::make('district_id')
    ->required()
    ->relationship(name: 'district', titleAttribute: 'name') // should be null
    // -> options(District::pluck('name', 'id'))
    ->searchable()
    ->preload()
```

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/filamentphp/demo

### Relevant log output

```shell

```

chrizgx:
Hello! At this point in my implementation, it works exactly like the screenshot @valpuia shared in the _Expected Behavior_ section. I could swap the message for something more clear, but honestly I think the best approach would be to avoid rendering the search input at all when there aren’t any options available.

@danharrin @valpuia , please let me know if you’d prefer that approach (hiding the search input) or if I should instead just provide a clearer message for those cases.

Also — this is my first contribution to Filament and to open source in general, so I’m pretty excited about it 😄

danharrin:
Fixed by #18641.

PR #18641, opened 2025-12-03
Fix: Select with searchable and preloaded option empty state #18529

chrizgx:
## Description

The problem is well described here: https://github.com/filamentphp/filament/issues/18529

## Functional changes

Improves the UX of searchable selects by immediately displaying a “no results” message when the component has zero options and is preloaded.
Currently, the user must type something before any feedback appears. This update ensures the empty-state message is shown right away, providing clearer guidance.

If this approach is preferred, I can also implement a dedicated message specifically for this scenario.


danharrin (changes requested):
I think we need a dedicated method for this situation as the current message doesn't make sense when there is no active search.

danharrin (approved):

PR #18642, opened 2025-12-03
Update documentation on hiding shared table columns

shmurf:
Clarified usage of hiddenOn() method to hide columns for multiple relation managers.

## Description

Updates the docs to include mention of array support in the hiddenOn method of table columns

## Visual changes

None

## Functional changes

None


danharrin:
Thanks, but given this has been draft for a while, and I don't necessarily think we should add it to the docs, I am going to close this.

PR #18646, opened 2025-12-04
Fix topbar logo layout shift on page load with collapsible sidebar

Saracaen:
## Description

When allowing the sidebar to be collapsed on desktop using sidebarCollapsibleOnDesktop or sidebarFullyCollapsibleOnDesktop, the logo briefly shifts on page load and on every navigation when using the application in non-SPA mode. This causes a layout shift (CLS) that only appears for a split second, but is noticeable.

The issue occurs because the two toggle buttons use x-show to alternate visibility, but during Alpine initialization, neither button has been hidden yet, causing a brief moment where the layout is unstable. The x-cloak directive is necessary to prevent other visual glitches, but doesn't fully solve this timing issue.

My fix wraps the buttons in a container div with a fixed width (w-9 shrink-0) that reserves space for exactly one button, ensuring the logo position remains stable regardless of Alpine's state.

### Before

{7D66FDC5-C4BF-41AA-9821-6540B150B29E}

### After

{8E1A8D2F-6062-417D-B2D8-394902478E67}

## Functional changes


Saracaen:
Come to think of it, this extra container div should probably be styled with an actual CSS-class instead for consistency. Maybe something like .fi-topbar-start-btn or .fi-topbar-start-btn-ctn?

danharrin (changes requested):
It does need to be in its own CSS class, Tailwind classes in Blade files are not compiled in Filament. Maybe use `fi-topbar-collapse-sidebar-btn-ctn`

Saracaen:
I have pushed the changes and recompiled the theme.css!

danharrin (changes requested):
Panels with `sidebarFullyCollapsibleOnDesktop()` now have excess gap when the sidebar is collapsed

Image

danharrin (approved):

PR #18663, opened 2025-12-07
Improve `make:theme` command UX

danharrin:


packages/panels/src/Commands/MakeThemeCommand.php
```diff
@@ -76,12 +83,16 @@ protected function getOptions(): array
         ];
     }
 
-    public function handle(): int
+    public function handle(Filesystem $filesystem): int
```
hussamitani, line 86: This breaks other packages. Please make it nullable

PR #18665, opened 2025-12-08
Enhance table documentation with selection features

Umair-exe:
Added documentation for pre-selecting table rows and bulk actions in tables.

## Functional changes

### DIFF
--- a/packages/tables/docs/04-actions.md
+++ b/packages/tables/docs/04-actions.md
@@ -128,6 +128,45 @@ public function table(Table $table): Table
 }
 ```
 
+### Pre-selecting table rows
+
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks for the effort, but this is just an internal feature for now and may not continue to work in future releases.

Umair-exe:
> Thanks for the effort, but this is just an internal feature for now and may not continue to work in future releases.

Thanks for the clarification! I understand that this is an internal feature and may change or be removed in future releases. I just wanted to mention that I spent quite a bit of time researching and experimenting to get this working, so your confirmation really helps.

If this feature is likely to be removed later, could you please advise on the recommended or future-safe way to achieve pre-selection? I want to make sure I implement it in a way that aligns with Filament’s direction.

danharrin:
There isn't currently another way to handle pre-selection reliably. It probably won't be removed in a minor version, so just make sure it continues to work between major releases.

PR #18677, opened 2025-12-09
Fix: HTML in header tooltips

webard:
## Description

https://github.com/filamentphp/filament/pull/18562 introduced the `headerTooltip()` method to  table columns, which allows `Htmlable` as a argument, but HTML is not rendered correctly inside tooltip due to lack of `x-tooltip.html` attribute.

## Visual changes

Before:
image

After:
image

## Functional changes


packages/tables/resources/views/index.blade.php
```diff
@@ -1467,6 +1467,9 @@ class="fi-ta-header-cell-sort-btn"
                                                     >
                                                         @if (filled($columnHeaderTooltip))
                                                             <span
+                                                                @if ($columnHeaderTooltip instanceof \Illuminate\Contracts\Support\Htmlable)
+                                                                    x-tooltip.html
+                                                                @endif
                                                                 x-tooltip="{
```
danharrin: I might be wrong, but isn't it cleaner in the generated HTML to conditionally add the `.html` modifier to this line instead? Otherwise we have `x-tooltip.html x-tooltip`?

danharrin (changes requested):

webard:
Please look at the changed code, you mean something like this?

danharrin (approved):

PR #18684, opened 2025-12-09
getParentRepeaterItem could be null

recca0120:
getParentRepeaterItem could be null

## Description

getParentRepeaterItem() could be null

image
## Functional changes


danharrin (changes requested):
Please don't remove it, you will break PHPStan. Instead, change it to `assert(($repeater instanceof Repeater) || (! $repeater))`

danharrin (approved):

PR #18694, opened 2025-12-11
Support custom styles for toggle colors

Abdulmajeed-Jamaan:
## Description

Toggle components now accept color arrays for 'on' and 'off' states, enabling custom styles via get_component_custom_styles. Blade and helper updates ensure correct class and style application for both string and array color values.

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/forms/src/Components/Concerns/HasToggleColors.php
+++ b/packages/forms/src/Components/Concerns/HasToggleColors.php
@@ -6,10 +6,20 @@ use Closure;
 
 trait HasToggleColors
 {
+    /**
+     * @var string|array<string, string>|Closure|null
+     */
     protected string | array | Closure | null $offColor = null;
 
+    /**
+     * @var string|array<string, string>|Closure|null
+     */
     protected string | array | Closure | null $onColor = null;
 
+    /**
+     * @param  string|array<string, string>|Closure|null  $color
+     * @return $this
+     */
     public function offColor(string | array | Closure | null $color): static
     {
         $this->offColor = $color;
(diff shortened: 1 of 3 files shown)


danharrin:
Thanks, I don't think I really understand the use case though. Can this be done with CSS in a custom theme? If not, please @ me in this thread and let me know what this should be used for.

Abdulmajeed-Jamaan:
Hi @danharrin,

In our project we use it to allow the end user (using color picker) to specify the "on" and "off" colors.

The color picker generates hex value which then internally converted to array using `\Filament\Support\Colors\Color::generateV3Palette()`

danharrin:
Cool, you could change the color using CSS and set it using a CSS variable, right? And then output the CSS variable from the user setting

Abdulmajeed-Jamaan:
That what i tried at first, but found out the color toggling is [controlled by css classes](https://github.com/filamentphp/filament/blob/8a2395dc6013dcfa90749ba6aebee245df2ca4ff/packages/support/resources/views/components/toggle.blade.php#L18-L26) that overrides the `--color-*` variables, and they are toggled by alpine class binding.

Is there a way to pass on demand custom css variables for each state ?, i need a way to pass css variables for both on/off state, and have alpine js toggles between them same as how its happening with classes.

Or am i missing something ?

@danharrin

danharrin:
You could target `fi-toggle-on` or `fi-toggle-off` in your CSS and apply a different background color based on that

Abdulmajeed-Jamaan:
Thanks for the help @danharrin,

Figured out `extraAlpineAttributes` method, and [got it working like how we want it to](https://github.com/lara-zeus/bolt/pull/408/changes)  👍🏻.

PR #18704, opened 2025-12-15
Update docs: Add support for pagination links in custom widget and fix current-page-option-property initialization

khumam:
## Description

I worked around to figure out how to add pagination links to my custom widget without using a Filament table. Eventually, I found that pagination can be implemented in a custom widget by adding the WithPagination trait and moving the view variable into the render() method. So I think this new implementation should help other developers implement pagination in their custom widgets.

I also discovered that the initialization of the current-page-option-property may be incorrect. In the documentation, it uses :current-page-option-property but passes a string value, which causes an error because the perPage value is treated as a constant. Therefore, the colon should be removed.

## Visual changes

No visual changes.

## Functional changes

No functional changes.

### DIFF
--- a/.phpactor.json
+++ b/.phpactor.json
@@ -1,0 +1,4 @@
+{
+    "$schema": "/phpactor.schema.json",
+    "language_server_phpstan.enabled": true
+}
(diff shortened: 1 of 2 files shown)


danharrin:
Thanks, but I think this is quite a specific use case to describe in this detail in the docs, and could look a bit out of place to be honest.

Thanks for the attribute fix though, I have merged that myself into v4 👍

khumam:
Thanks Dan! 👍

PR #18705, opened 2025-12-15
fix: always apply filtering regardless of whether shouldDeselectAllRecordsWhenFiltered is enabled

### LINKED ISSUE #18693: Select all records is broken when using `deselectAllRecordsWhenFiltered`
### Package

filament/filament

### Package Version

v4.3.1

### Laravel Version

v12.42

### Livewire Version

v3.7.1

### PHP Version

PHP 8.4

### Problem description

The "Select all N records" ignores the table filter when using `deselectAllRecordsWhenFiltered` and selects entire records in the database.

The action for testing:

```php
->toolbarActions([
    BulkAction::make('count_selected')
        ->action(function (Collection $records) {
            Notification::make()
                ->title('Selected Users Count')
                ->body($records->count() . ' users selected.')
                ->send();
        })
]);
```

Image

Image

Image

Image

### Expected behavior

Select N records should only select N records based on table filters.

### Steps to reproduce

1. Create a table with some filters.
2. Add `->deselectAllRecordsWhenFiltered(false)` to the table.
3. Filter the table to see a limited number of records.
4. Select one record, then click "Select all N records".
5. Call a bulk action, the `$records` in the bulk action will contain all records, not selected records. 

https://github.com/amir9480/filament-table-select-all-bug-report

### Relevant class
https://github.com/amir9480/filament-table-select-all-bug-report/blob/main/app/Filament/Resources/Users/Tables/UsersTable.php#L44

People-Sea:
## Description

fixes: #18693 

I think `deselectAllRecordsWhenFiltered` should control the UI behavior only, should not affect the query result

related pr: #8591

## Functional changes

### DIFF
--- a/packages/tables/src/Concerns/HasBulkActions.php
+++ b/packages/tables/src/Concerns/HasBulkActions.php
@@ -280,9 +280,7 @@ trait HasBulkActions
                 }
             }
 
-            if ($table->shouldDeselectAllRecordsWhenFiltered()) {
-                $this->filterTableQuery($query);
-            }
+            $this->filterTableQuery($query);
 
             return $query;
         }


danharrin (changes requested):
Does this break when you filter, select a few records, apply a different filter, select other records that werent in the original filter, and then run a bulk action? Do the records from the first filter get included?

People-Sea:
@danharrin Thank you for your reminder. I tested that first apply the filter conditions, then select a piece of data, and then modify the filter conditions. Only when the data meets the current filter conditions will it be operated in batches


https://github.com/user-attachments/assets/fcbd2a31-b431-4fe2-949c-0b9da68d71ad


If this is in line with expectations, then the current error is only the selected data display error

image



If you think my analysis is correct, I will continue to fix the problems of the selected data.

People-Sea:
@danharrin   Thank you for your reminder. I have tested and reanalyzed the reason for the error. It was caused by checking "Select all"


This change will result in not allowing cross-condition batch operations

pr  test:

https://github.com/user-attachments/assets/d6542860-f26b-434e-a51d-786118d2e246



During batch operations, data that does not meet the current filter conditions is excluded

People-Sea:
This issue is a bit tricky. If "select all" is clicked, then "where in pk(id)" will not be used. I guess it's a performance problem.

What I'm not sure about now is the real expectation

danharrin:
Replaced by #18772.

PR #18716, opened 2025-12-16
Add onErrorUrl() method to Image component for fallback image support

mhortulanus:
## Description

This PR adds an `onErrorUrl()` method to the `Filament\Schemas\Components\Image` component, allowing developers to specify a fallback image URL that will be displayed when the primary image fails to load. This is useful for handling broken image links gracefully and providing a better user experience.

The implementation follows Filament's existing patterns:
- Supports `string | Closure | null` for dynamic URL generation
- Uses `$this->evaluate()` for closure evaluation
- Only adds error handling attributes when an error URL is provided
- Uses standard HTML `onerror` event handler for browser-native fallback functionality

**Usage example:**
Image::make($url, $alt)
    ->onErrorUrl($fallbackUrl)When the primary image fails to load, the browser will automatically replace it with the fallback URL specified via `onErrorUrl()`.

## Visual changes

No visual changes to the UI itself. The feature only affects behavior when an image fails to load - in that case, the fallback image will be displayed instead of a broken image icon.

## Functional changes

### DIFF
--- a/packages/schemas/docs/07-primes.md
+++ b/packages/schemas/docs/07-primes.md
@@ -414,6 +414,24 @@ Image::make(
 
 <AutoScreenshot name="primes/image/tooltip" alt="Image with a tooltip" version="4.x" />
 
+### Adding a fallback image URL
+
+You may specify a fallback image URL that will be displayed when the primary image fails to load using the `onErrorUrl()` method:
+
+```php
+use Filament\Schemas\Components\Image;
+
+Image::make(
+    url: asset('images/user-avatar.jpg'),
+    alt: 'User avatar',
+)
+    ->onErrorUrl(asset('images/default-avatar.jpg'))
+```
+
+<UtilityInjection set="schemaComponents" version="4.x">As well as allowing a static value, the `onErrorUrl()` method also accepts a function to dynamically calculate it. You can inject various utilities into the function as parameters.</UtilityInjection>
+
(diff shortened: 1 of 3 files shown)


danharrin:
Thanks, I suggest you just use `extraAttributes()` for this though I think. I haven't really heard many requests for this in the past and in the interest of not introducing too much bloat into the framework, I am going to hold off and maybe re-evaluate in the future if it gets mentioned more.

webard:
It can actually be very helpful. I had the same problem and the solution forced me to duplicate the `Image` component:

```php
ImageEntry::make('image')
    ->hiddenLabel()
    ->imageSize('100%')
    ->disk('public')
    ->visible(static fn (Product $record): bool => $record->image !== null && Storage::disk('public')->exists($record->image)),

ImageEntry::make('image')
    ->hiddenLabel()
    ->imageSize('100%')
    ->getStateUsing(static fn (): string => \asset('empty.png'))
    ->visible(static fn (Product $record): bool => $record->image === null || !Storage::disk('public')->exists($record->image)),
```

and this PR introduces the elegant way to do the same.

mhortulanus:
@danharrin the example of @webard illustrates exactly why I think this feature is necessary. I'm building a panel for the construction branch where we work extensively with profile pictures. Since not everyone has one yet, having a clean, structured fallback route feels much better than relying on extraAttributes().

danharrin:
That does seem like a different solution to the one proposed, and for a different component

PR #18717, opened 2025-12-16
Fix Postgres error when using searchable in tables

### LINKED ISSUE #18467: Postgres throws "undefined column" when using searchable in tables
### Package

filament/filament

### Package Version

v4.2.0

### Laravel Version

v12.37.0

### Livewire Version

v3.6.4

### PHP Version

8.4.1

### Problem description

An `Undefined column` error is thrown when searching a table which uses `->searchable([...])` on the `Table` object. It works fine when using `->searchable()` on `Column`s.

I found that adding a driver check  in the generate_search_column_expression helper method fixes it for me. It makes sure that for Postgres it always returns an expression, never a string, but I'm not sure if this breaks other things.
https://github.com/filamentphp/filament/blob/4.x/packages/support/src/helpers.php#L294

```php
if (
    $driverName === 'pgsql' ||
    str($column)->contains('(') || // This checks if the column name probably contains a raw expression like `lower()` or `json_extract()`.
    filled($collation)
) {
    return new Expression($column);
}
```

### Expected behavior

Searching a table should filter the table based on the input, instead of throwing an error.

### Steps to reproduce

- Use a Postgres database
- Create a resource like `CustomerResource`
- Add `->searchable(['any_column'])` to the table in the `CustomersTable` class
- Open the browser and try to search for something in the table
- The error is thrown

https://github.com/robinvda/filament-postgres-search-bug

### Relevant log output

```shell
Illuminate\Database\QueryException
vendor/laravel/framework/src/Illuminate/Database/Connection.php:824

SQLSTATE[42703]: Undefined column: 7 ERROR: column ""name"::text" does not exist LINE 1: ...ect count(*) as aggregate from "customers" where ("""name"":... ^ (Connection: pgsql, SQL: select count(*) as aggregate from "customers" where ("""name""::text"::text like %a%))
```

robinvda:
## Description

Fixes https://github.com/filamentphp/filament/issues/18467

Fixes a `undefined column` error from Postgres when using table search. Also fixes a bug where table search would always force the search term to be lowercase by adding `forceSearchCaseInsensitive()` to `Table` (same implementation as in `Column`).

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/support/src/helpers.php
+++ b/packages/support/src/helpers.php
@@ -272,11 +272,11 @@ if (! function_exists('Filament\Support\generate_search_column_expression')) {
 
         $isSearchForcedCaseInsensitive ??= match ($driverName) {
             'pgsql' => true,
-            default => str($column)->contains('json_extract('),
+            default => str($column)->contains('json_extract(') || str($column)->contains('->'),
         };
 
         if ($isSearchForcedCaseInsensitive) {
-            if (in_array($driverName, ['mysql', 'mariadb'], true) && str($column)->contains('->') && ! str($column)->startsWith('json_extract(')) {
+            if (in_array($driverName, ['mysql', 'mariadb', 'sqlite'], true) && str($column)->contains('->') && ! str($column)->startsWith('json_extract(')) {
                 [$field, $path] = invade($databaseConnection->getQueryGrammar())->wrapJsonFieldAndPath($column); /** @phpstan-ignore-line */
                 $column = "json_extract({$field}{$path})";
             }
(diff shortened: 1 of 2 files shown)


danharrin:
Can you explain more about the case insensitivity fix, I don't really understand it

robinvda:
Using Postgres, the search term was always forced to lowercase (using `Str::lower()`). So even if i search for `John`, it would only find results with `john`, not `John`. I'm on holiday for 2 weeks now so if necessary i can further check this when i get back.

danharrin:
In the query, is `lower(column)` or just `column` generated? It should be the first for Postgres

danharrin:
Replaced by #18791.

Issue #18017, opened 2025-10-01
RichEditor error in repeater


andyov17:
### Package

filament/filament

### Package Version

v4

### Laravel Version

v12

### Livewire Version

v3

### PHP Version

8.4

### Problem description

Similar to https://github.com/filamentphp/filament/issues/17472
if a RichEditor is in a repeater, subsequent saves fire the error:
> Filament\Forms\Components\RichEditor::{closure:Filament\Forms\Components\RichEditor::setUp():303}(): Argument #2 ($rawState) must be of type ?array, string given

### Expected behavior

no error

### Steps to reproduce

Have a repeater with a RichEditor in
initial saving of content works, but if you save without editing the RichEditor, the error shows

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/andyov17/filamentricheditor

### Relevant log output

```shell
Filament\Forms\Components\RichEditor::{closure:Filament\Forms\Components\RichEditor::setUp():303}(): Argument #2 ($rawState) must be of type ?array, string given
```

PR #17762, opened 2025-09-12
fix: Rich editor testing with HTML

danharrin:
Fixes #17472

Issue #17472, opened 2025-08-22
[v4.x] Error on testing a RichEditor component


alexmanase:
### Package

filament/forms

### Package Version

v4.0.3

### Laravel Version

v12.25.0

### Livewire Version

v3.6.4

### PHP Version

8.4.11

### Problem description

On testing a `RichEditor` component I get the following error:

```
Filament\Forms\Components\RichEditor::{closure:Filament\Forms\Components\RichEditor::setUp():303}(): Argument #2 ($rawState) must be of type ?array, string given
```

Even though I use an array instead of string, I am still unable to test this component.

### Expected behavior

The test should pass if I have something like this:

```php
// Form
->schema([
    RichEditor::make('note')
        ->required()
])

// Testing
->fillForm([
   'note' => 'foo'
])
->call('submit'); // I should get the state of the form
```

### Steps to reproduce

1. Clone `https://github.com/alexmanase/filament-testing-rich-editor-issue`
2. `composer install`
3. Run the test: `./vendor/bin/pest`
4. Observe that running `dd($this->form->getState())` isn't working.

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/alexmanase/filament-testing-rich-editor-issue

### Relevant log output

```shell
Filament\Forms\Components\RichEditor::{closure:Filament\Forms\Components\RichEditor::setUp():303}(): Argument #2 ($rawState) must be of type ?array, string given
```

Casmo:
I think I have a similar error:

<img width="1466" height="1078" alt="Image" src="https://github.com/user-attachments/assets/8bb7941e-e2c9-4539-b815-4d667f0b9436" />

Happening when working with dynamic inputs.

awcodes:
I'm not sure this is an actual bug. Even if the record data is stored as html it's converted to the Tiptap array during hydration and converted back during dehydration. So, in tests, I would expect the fillForm() to be the array representation and not the html string.

Does it still error if you use `<p>foo</p>`?

webplusmultimedia:
So m'i , F4.0.5 , php8.3 : 

<img width="1301" height="330" alt="Image" src="https://github.com/user-attachments/assets/da315320-2948-4c38-adaa-a60b6e5a587f" />

If use '', it's seem that's ok (hasErrors for me required)

__edit
sorry , when i use '' :

<img width="1301" height="330" alt="Image" src="https://github.com/user-attachments/assets/2b666bda-2db6-4c41-9665-85439cded2c8" />

[...]

PR #18718, opened 2025-12-16
Fix RichEditor type error when saving form without editing the field

iotron:
## Summary

This PR fixes the `TypeError: Argument #2 ($rawState) must be of type ?array, string given` error that occurs when saving a form that contains a RichEditor field in HTML mode (not using `->json()`) without editing that field.

**Related issues:**
- https://github.com/filamentphp/filament/issues/17472
- https://github.com/filamentphp/filament/issues/18017

## Problem

When a RichEditor stores content as HTML (the default mode), the `beforeStateDehydrated` callback registered in `setUp()` expects `?array $rawState`. However, when the form is saved without editing the RichEditor field:

1. `callAfterStateUpdated()` is NOT called (since the field wasn't updated)
2. `callBeforeStateDehydrated()` IS called (as part of the form save process)
3. `getRawState()` returns the HTML string from the database
4. The callback receives a string instead of an array, causing the type error

The existing fix in PR #17762 added state normalization to `callAfterStateUpdated()`, which only runs when the RichEditor field itself is edited. This doesn't cover the case where other fields are edited but the RichEditor is left untouched.

## Solution

Add the same state normalization logic to `callBeforeStateDehydrated()`, which runs during every form save regardless of which fields were edited. This ensures the raw state is converted to the expected array format before the `beforeStateDehydrated` callback is invoked.

## Test Case

1. Create a model with a RichEditor field storing HTML (not using `->json()` mode)
2. Save some HTML content to the field
3. Edit ANY other field in the form (not the RichEditor)
4. Save the form
5. Without this fix: `TypeError` is thrown
6. With this fix: Form saves successfully

### DIFF
--- a/packages/forms/src/Components/RichEditor.php
+++ b/packages/forms/src/Components/RichEditor.php
@@ -997,6 +997,24 @@ class RichEditor extends Field implements Contracts\CanBeLengthConstrained
         return parent::callAfterStateUpdated($shouldBubbleToParents);
     }
 
+    public function callBeforeStateDehydrated(array &$state = []): static
+    {
+        $rawState = $this->getRawState();
+
+        // https://github.com/filamentphp/filament/issues/17472
+        // The beforeStateDehydrated callback expects ?array but getRawState() returns
+        // string when RichEditor is in HTML mode and the field wasn't edited.
+        if (! is_array($rawState)) {
+            foreach ($this->getStateCasts() as $stateCast) {
+                $rawState = $stateCast->set($rawState);
+            }
+
+            $this->rawState($rawState);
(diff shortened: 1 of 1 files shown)


maarten-mald-digital:
Looking good! Waiting for this fix

danharrin:
Can you create a reproduction repository for this issue please? I don't understand why the state would be a string in the first place: the state cast should convert it to an array when the form is filled with the existing string, even if JSON mode is not being used.

danharrin:
Sorry, I need some more info here before I can merge this. Please open an issue with a reproduction repository.

PR #18721, opened 2025-12-16
Add "shouldTranslate" to notification title and body

Boy132:
## Description

Allows to use translation keys for notification title and body, similar to labels.

## Functional changes

### DIFF
--- a/packages/notifications/src/Concerns/HasBody.php
+++ b/packages/notifications/src/Concerns/HasBody.php
@@ -8,6 +8,8 @@ trait HasBody
 {
     protected string | Closure | null $body = null;
 
+    protected bool $shouldTranslateBody = false;
+
     public function body(string | Closure | null $body): static
     {
         $this->body = $body;
@@ -15,8 +17,19 @@ trait HasBody
         return $this;
     }
 
+    public function translateBody(bool $shouldTranslateBody = true): static
+    {
+        $this->shouldTranslateBody = $shouldTranslateBody;
+
+        return $this;
+    }
+
     public function getBody(): ?string
     {
-        return $this->evaluate($this->body);
+        $body = $this->evaluate($this->body);
+
+        return is_string($body) && $this->shouldTranslateBody
+            ? __($body)
+            : $body;
     }
 }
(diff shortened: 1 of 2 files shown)


danharrin:
Thanks for the PR, but I do not really want this feature in Filament. I wish I never allowed `translateLabel()` into Filament in the first place for forms and tables, since it encourages really bad translation practices. Structured translation keys are best to ensure the correct translation context is always used. Sorry.

Boy132:
> Thanks for the PR, but I do not really want this feature in Filament. I wish I never allowed `translateLabel()` into Filament in the first place for forms and tables, since it encourages really bad translation practices. Structured translation keys are best to ensure the correct translation context is always used. Sorry.

The problem in our case is that we send database notifications from user A to user B. And the language can be set per user. This creates a problem because now the notification for user B is displayed in the language of user A. We can work around that by using user B's locale in `trans` but that only works as long as we only have one recipient.

danharrin:
I'm not sure I understand the "only works as long as we only have one recipient", if that is the case how is this solution solving that part of the problem?

Boy132:
Assume you have three users: user A with english as their language, user B with german and user C with spanish.
Now, when user A triggers an action that sends a database notification to user B and user C it will use the language of user A:

```php
Notification::make()
    ->title(trans('notifications.test_title'))
    ->body(trans('notifications.test_body'))
    ->sendToDatabase([$userB, $userC]);
```

With this PR you could do this:

```php
Notification::make()
    ->title('notifications.test_title')
    ->body('notifications.test_body')
    ->translateTitle()
    ->translateBody()
    ->sendToDatabase([$userB, $userC]);
```

And the title and body would be correctly displayed for user B and user C. (because in our case we have a middleware that sets the app locale to the locale that the user selected)

Hope this makes it clearer!

danharrin:
I think this is the fix you are looking for

```php
Notification::make()
    ->title(fn (): string => trans('notifications.test_title'))
    ->body(fn (): string => trans('notifications.test_body'))
    ->sendToDatabase([$userB, $userC]);
```

Boy132:
Actually nevermind, I had a misconception. You were right to close this.
Sorry for the inconvenience! :D

PR #18730, opened 2025-12-18
[4.x - Feature] Adds a summaries() method to control the visibility of page and total summaries in tables

bezhanSalleh:
## Description

Adds a `summaries()` method to control the visibility of page and total summaries in tables.
```php
->summaries(pageSummary: bool|null, totalSummary: bool|null)
```

Both parameters default to `null`, preserving existing behavior.

> **Note:** This is a first attempt at implementing this feature. The API design and method names are open to feedback and can be adjusted if there's a better approach. If a more streamlined API would require breaking changes, that could be targeted for v5.x rather than v4.x(current).

## Motivation

When working with large datasets—particularly in inventory systems and transaction histories—having granular control over summary visibility becomes important. Different contexts call for different displays:

**Multi-page transaction history:** When viewing stock movements for a group of products, data often spans multiple pages. In this case, showing "This Page" summaries can be misleading since they only reflect a partial view of the data.

image

**Single product context:** When using the same table within a relation manager or for related records, neither summary may be needed—the product itself already has its own summarizer displaying the totals.

image

This change allows reusing the same table definition across different contexts while toggling summary visibility as needed.

## Functional changes


packages/tables/resources/views/components/summary/index.blade.php
```diff
@@ -24,10 +24,13 @@
             ->all();
     }
 
-    $hasPageSummary = (! $groupsOnly) && $records instanceof \Illuminate\Contracts\Pagination\Paginator && $records->hasPages();
+    $table = $this->getTable();
```
danharrin: Instead of calling methods on this object, please pass through the new parameters as attributes to the component like the rest

packages/tables/src/Table/Concerns/CanSummarizeRecords.php
```diff
@@ -7,8 +7,42 @@
 
 trait CanSummarizeRecords
 {
+    protected bool | Closure | null $hasPageSummary = null;
+
+    protected bool | Closure | null $hasTotalSummary = null;
+
+    public function summaries(bool | Closure | null $pageSummary = null, bool | Closure | null $totalSummary = null): static
```
danharrin: I think the args should be `$pageSummaryCondition` and `$totalSummaryCondition`

packages/tables/src/Table/Concerns/CanSummarizeRecords.php
```diff
@@ -7,8 +7,42 @@
...
+    {
+        if ($pageSummary !== null) {
```
danharrin: I don't think we should have these conditions

packages/tables/src/Table/Concerns/CanSummarizeRecords.php
danharrin: I think the defaults for the method args should be `true` instead of `null` too, but the property defaults can still be `null`

packages/tables/src/Table/Concerns/CanSummarizeRecords.php
```diff
@@ -7,8 +7,42 @@
...
+
+    public function hasTotalSummary(): bool
+    {
+        return (bool) $this->evaluate($this->hasTotalSummary ?? true);
+    }
+
     public function hasSummary(Builder | Closure | null $query): bool
     {
+        if (
+            $this->hasPageSummary === false &&
```
danharrin: What is this condition for? If we need to call `hasPageSummary` and `hasTotalSummary` in the view anyway, do we really need to check here too?

danharrin (changes requested):

packages/tables/resources/views/components/summary/index.blade.php
bezhanSalleh: Done.

packages/tables/src/Table/Concerns/CanSummarizeRecords.php
bezhanSalleh: if we pass the parameters through the attributes like you said, these won't be necessary anymore. which we have now.

danharrin (changes requested):
Please review the diff of this PR, there are many formatting changes

bezhanSalleh:
reverted to the actual fix, without pulling in the latest v4 changes.

danharrin (approved):

PR #18739, opened 2025-12-21
Fix: Scope persisted table filters session key by tenant

dieterwarson:
Scope persisted table filters session key by tenant

## Description

When persistFiltersInSession() is enabled, table filters are stored in the session using a key derived from the table/component class. In multi-tenant apps, switching tenants within the same session can unintentionally reuse the previous tenant’s persisted filters.

This PR scopes the table filters session key by the current tenant key (when a tenant is active), so persisted filters are isolated per tenant.

If tenant-scoped persistence is not the expected default behavior, this could be exposed as a configuration option on persistFiltersInSession() to allow opting into tenant-aware filter persistence.

### DIFF
--- a/packages/tables/src/Concerns/HasFilters.php
+++ b/packages/tables/src/Concerns/HasFilters.php
@@ -221,7 +221,15 @@ trait HasFilters
 
     public function getTableFiltersSessionKey(): string
     {
-        $table = md5($this::class);
+        $tenantKey = filament()->getTenant()?->getKey();
+
+        $namespace = $this::class;
+
+        if ($tenantKey) {
+            $namespace .= '|' . $tenantKey;
+        }
+
+        $table = md5($namespace);
 
         return "tables.{$table}_filters";
     }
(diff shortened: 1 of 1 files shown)


packages/tables/src/Concerns/HasFilters.php
```diff
@@ -221,7 +221,15 @@ public function parseTableFilterName(string $name): string
 
     public function getTableFiltersSessionKey(): string
     {
-        $table = md5($this::class);
+        $tenantKey = filament()->getTenant()?->getKey();
```
danharrin: `filament/filament` might not be installed when using `filament/tables`, so there needs to be some `class_exists()` check probably on the `Filament` facade. Then replace `filament()->` with `Filament::`

danharrin (changes requested):

packages/tables/src/Concerns/HasFilters.php
dieterwarson: Hi Dan, thanks for the feedback. I've pushed a new commit that takes this into account.

danharrin (approved):

PR #18741, opened 2025-12-22
Add current password field with validation message

0xxb:
## Description

Fix an issue where, in the admin panel’s personal profile page, the “Current Password” field was not correctly displayed in Chinese under the Chinese locale.

## Visual changes

before:
image

after:
image

## Functional changes

### DIFF
--- a/packages/panels/resources/lang/zh_CN/auth/pages/edit-profile.php
+++ b/packages/panels/resources/lang/zh_CN/auth/pages/edit-profile.php
@@ -22,6 +22,12 @@ return [
             'label' => '确认新密码',
         ],
 
+        'current_password' => [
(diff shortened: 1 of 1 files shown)


ziming:
does the hong kong locale have the same problem as well?

0xxb:
> does the hong kong locale have the same problem as well?

Yes, zh_HK has the same issue.
While checking, I also noticed that zh_CN, zh_HK, and zh_TW have a different directory / key structure compared to en, and multiple fields are missing.
This PR only addresses the specific missing field mentioned above. The broader inconsistencies across Chinese locales likely need to be handled in a separate PR.

danharrin (approved):

PR #18747, opened 2025-12-23
Feature: allow null brand name

webard:
## Description

This PR allows  `brandName()` to be nullable. When `brandName` is null, there is no logo and no suffix in page title.

Of course you can set empty string as brandName, but currently the ` - ` suffix is added to title, when brandName is empty, what looks bad.

This allows to use `PanelsRenderHook::TOPBAR_LOGO_BEFORE` to prepare own logo, in my case it is a dynamic logo that changes when you enter to other cluster of the panel.

This PR can be addressed to v5, because of breaking change that adds `->brandName(config('app.name'))` to generated service provider, to restore previous behavior.

Maybe good option will be separate method to set suffix in title, like `->titleSuffix()`? I can add this.

## Functional changes

### DIFF
--- a/packages/panels/resources/views/components/layout/base.blade.php
+++ b/packages/panels/resources/views/components/layout/base.blade.php
@@ -32,7 +32,7 @@
         @endphp
 
         <title>
-            {{ filled($title) ? "{$title} - " : null }} {{ $brandName }}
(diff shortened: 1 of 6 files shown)


danharrin:
To keep consistent with other parts of the codebase, please make `null` default to the app name like it did before, and allow `''` empty string to remove the brand name altogether.

PR #18750, opened 2025-12-24
[4.x] Fix to handle Block class in CanBeRepeated trait

cheesegrits:
## Description

The CanBeRepeated trait was not handling the Block class, so (for example) having a 'distinct' validation in a Block would blow up, because getParentRepeaterItem() couldn't find the parent repeating component.

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/schemas/src/Components/Concerns/CanBeRepeated.php
+++ b/packages/schemas/src/Components/Concerns/CanBeRepeated.php
@@ -2,19 +2,19 @@
 
 namespace Filament\Schemas\Components\Concerns;
 
-use Filament\Forms\Components\Builder\Block;
+use Filament\Forms\Components\Builder;
 use Filament\Forms\Components\Repeater;
 use Filament\Schemas\Schema;
 
 trait CanBeRepeated
 {
-    protected Schema | bool | null $cachedParentRepeaterItem = null;
+    protected Schema|bool|null $cachedParentRepeaterItem = null;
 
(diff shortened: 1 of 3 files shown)


danharrin (changes requested):
Would you mind adding a test for this please (and for the repeater version too if it doesn't exist), which will help me understand a bit more about the use and how it works with different (mixed) block types too? :)

cheesegrits:
Hard to know the best way of testing it.  Probably by implementing some tests for the 'distinct' validation on Repeater and Blocks?

And it occurred to me I probably need to test for the Builder class, not the Block class.

I'll @ ya when I have the testing done.

cheesegrits:
@danharrin OK, I added some tests in the Repeater and Builder tests.  They test the distinct() validation, which exercises the getParentRepeater() code.  I'm testing for distinct() in builder, repeater, builder inside a repeater, and repeater inside a builder.

BTW, the CanBeRepeated trait looked familiar, and I realized I'd written it for my distinct() validation PR on 3.x.  It doesn't appear that use of it has spread anywhere else, so I don't think changes to it will have any side effects.  The intended use is simply to find the parent repeating item, be it a Builder or a Repeater.   Code that uses it makes no other assumptions about the component type, it just cares about the state paths so it can check for uniqueness of values across multiple repeated instances.

danharrin (approved):
Great tests, good job

PR #18751, opened 2025-12-24
Database Notification Action Icon

e2d2-dev:
## Description

When a database notification with an action with an icon is sent, the enum will be serialized to just the value without "heroicon-" prefix to db.

This will add the prefix and icons are showing up.

## Visual changes

icon will show up

## Functional changes

none


danharrin (changes requested):
Instead of checking if the icon is a Heroicon, check if it has the `ScalableIcon` interface and call `getIconForSize()` to get the full icon string.

Also, do we need to do the same for the main notification icon in `toArray()`?

danharrin (approved):

PR #18778, opened 2025-12-29
fix: Dark mode input BG color

### LINKED ISSUE #18774: Search icon has a different background than search input
### Package

filament/filament

### Package Version

v4.3.1

### Laravel Version

v12.44.0

### Livewire Version

v3.7.3

### PHP Version

PHP 8.5.1

### Problem description

The search icon has a different background than the search input. *(I also suggest making the icon clickable)*

Image

Looking at the [Filament demo](https://demo.filamentphp.com/), which still uses v4.2.0, this seems to have appeared only recently.

### Expected behavior

All with the same background, and ideally make the entire field clickable.

### Steps to reproduce

Take a look at the search in the header, a table, etc.

### Relevant log output

```shell

```

danharrin:
Fixes #18774.


nessimabadi:
HI @danharrin , i think that by doing this the background is fixed for date, datetime, and time inputs but for the rest of the inputs it removes the background in dark mode.

Here's a screenshoot
image

I'm using the forms library not in a panel, but in a custom app

danharrin:
Hi, the class that I removed should just be a 1% opacity white, so I do not understand why it would cause the issue in your screenshot. Are you able to reproduce it in a fresh repository?

ziming:
i have the same issue

danharrin:
@ziming can I have an issue with a reproduction please?

ziming:
Not sure about the ingredients needed to reproduce but it is happening on about 3-4 of my production sites. (I have 8 sites in total).

danharrin:
I'm sorry, I cannot reproduce locally so you need to help me. This PR was created to fix a different bug.

ziming:
ok, if i figure out the right combination to reproduce it will let you know

nessimabadi:
I think i found the problem.
It occurs when using filament with the tailwind plugin "@tailwindcss/forms".

The plugin adds the following

image

Which sets the background to white on all inputs.

Before this pull request, all inputs had `bg-white/[1%]`, but the PR removed it for only the date related inputs, so now the base layer of the typography plugin applies.

I could create a test repository if needed. Just adding the typography plugin and setting it up will cause the bug.

The problem occurs because Filament 3 applications required the typography plugin, and Filament 4 applications don't. During the upgrade if the user did not remove the plugin then this will happen.

nessimabadi:
@danharrin I created an issue to move the conversation there with a reproduction repository

https://github.com/filamentphp/filament/issues/18874

Issue #18807, opened 2025-12-31
[v3] RepeatableEntry doesn't display values for array-based items


maytham553:
### Package

filament/filament

### Package Version

v3.3.45

### Laravel Version

v12.44.0

### Livewire Version

v3.7.3

### PHP Version

8.3.6

### Problem description

When using `RepeatableEntry` with array-based data (not Eloquent models) in infolists, child Entry components fail to display values. The repeater containers are created correctly, but all field values appear empty.

**Component:** `filament/infolists` - `RepeatableEntry`

### Expected behavior

When using `RepeatableEntry` with array-based data, each repeater item container should display its values correctly in child Entry components.

### Steps to reproduce

1. Create a Filament page with an infolist containing RepeatableEntry:

```
RepeatableEntry::make('family_members')
    ->label('Family Members')
    ->state([
        ['name' => 'John Doe', 'relation' => 'Father'],
        ['name' => 'Jane Doe', 'relation' => 'Mother'],
    ])
    ->schema([
        TextEntry::make('name')->label('Name'),
        TextEntry::make('relation')->label('Relation'),
    ])
```
2. View the page - the repeater displays containers but values are empty.
Child Entry components should display values from the array items:
- Container 1: name = "John Doe", relation = "Father"  
- Container 2: name = "Jane Doe", relation = "Mother"

But **Actual Behavior** 

Child Entry components display empty values:
- Container 1: name = (empty), relation = (empty)
- Container 2: name = (empty), relation = (empty)

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/maytham553/filament-repeater-entry-bug

### Relevant log output

```shell

```

danharrin:
Hi, can you please confirm that this is only an issue in v3 and is not an issue in v4?

maytham553:
Hi Dan,
I’ve tested this using a new [**v4** branch](https://github.com/maytham553/filament-repeater-entry-bug/tree/v4) in the reproduction repository, and the issue does not occur there — **everything works correctly**.
Based on this, it seems the issue is specific to Filament v3 only.

danharrin:
Perfect, thank you for confirming. I am not personally working on any further bug fixes to v3, especially when the issue has been fixed in v4 already. I am still working on security fixes to v3 if they are required. I will also merge any community contributes to v3 to fix bugs, so if you would like to submit a PR to fix this bug in v3 please do so :) Thanks!

maytham553:
@danharrin 
[...]

PR #18809, opened 2025-12-31
Add handling for array item data in RepeatableEntry and update state …

maytham553:
Fixes issue with `RepeatableEntry` not correctly handling array data in nested contexts, and improves state path resolution in `HasState` trait. more information here : https://github.com/filamentphp/filament/issues/18807

**Problem:**
1. When `RepeatableEntry` contains array data (not Model instances), child components couldn't access state because `getChildComponentContainers()` only handled Model instances via `record()` method.
2. `HasState::getState()` always used absolute paths, which broke nested structures where relative paths are needed.

## Functional changes

### DIFF
--- a/packages/infolists/src/Components/Concerns/HasState.php
+++ b/packages/infolists/src/Components/Concerns/HasState.php
@@ -69,7 +69,7 @@ trait HasState
 
             $state = $containerState instanceof Model ?
                 $this->getStateFromRecord($containerState) :
-                data_get($containerState, $this->getStatePath());
+                data_get($containerState, $this->getStatePath(false));
         }
 
         if (is_string($state) && ($separator = $this->getSeparator())) {
--- a/packages/infolists/src/Components/Concerns/HasState.php
+++ b/packages/infolists/src/Components/Concerns/HasState.php
@@ -69,7 +69,7 @@ trait HasState
 
             $state = $containerState instanceof Model ?
                 $this->getStateFromRecord($containerState) :
-                data_get($containerState, $this->getStatePath(false));
+                data_get($containerState, $this->getStatePath(isAbsolute: false));
         }
 
         if (is_string($state) && ($separator = $this->getSeparator())) {
--- a/packages/infolists/src/Components/RepeatableEntry.php
+++ b/packages/infolists/src/Components/RepeatableEntry.php
@@ -36,6 +36,8 @@ class RepeatableEntry extends Entry
 
             if ($itemData instanceof Model) {
                 $container->record($itemData);
+            } elseif (is_array($itemData)) {
+                $container->state($itemData);
             }
 
             $containers[$itemKey] = $container;


danharrin (approved):

dbpolito:
This made my infolist stop to work... maybe because i have nested repeatableentry, if i revert this line:

https://github.com/filamentphp/filament/pull/18809/changes#diff-3f63c2dc959e1eae881d60e4b4bd67f95f7df96cc8ed7652d8f0386ebde39930R72

it works again

dbpolito:
This doesn't work anymore, but worked before:

```php
               Infolists\Components\RepeatableEntry::make('reportMenuItems')
                    ->contained(false)
                    ->label('')
                    ->grid(2)
                    ->schema([
                        Infolists\Components\Section::make(fn ($record) => strtoupper($record->name))
                            ->collapsible()
                            ->hidden(fn ($record) => count($record->children) == 0)
                            ->schema([
                                Infolists\Components\RepeatableEntry::make('childrenTree')
                                    ->hiddenLabel()
                                    ->contained(false)
                                    ->schema([
                                        Infolists\Components\Grid::make()
                                            ->schema([
                                                Infolists\Components\TextEntry::make('name')
                                                    ->label('')
                                                    ->html()
                                                    ->url(static fn (MenuItem $model): string => $model->url, shouldOpenInNewTab: true),

                                                Infolists\Components\TextEntry::make('id')
                                                    ->label('')
                                                    ->formatStateUsing(fn () => '')
                                                    ->suffixAction(
                                                        Action::make('name')
                                                            ->icon('heroicon-m-star')
                                                            ->color(fn (MenuItem $record) => $reportMenuFavorites->find($record) ? 'primary' : 'gray')
                                                            ->action(fn (MenuItem $record, Component $livewire) => $this->star($record, $livewire, $reportMenuFavorites))
                                                    ),
                                            ]),
                                    ]),
                            ]),
                    ]),
```

and to add more context, `reportMenuItems` is a collection of models, and then inside of each model `childrenTree` is another collection of models

rasmuscnielsen:
Hey

Just chiming in here that this merge actually broke part of our dashboard too. 

**Resource:**

```php
RepeatableEntry::make('log')
    ->label('')
    ->contained(false)
    ->schema([
        ViewEntry::make('')
            ->view('filament.infolists.log-item-entry')
            ->label('')
    ]),
```

**log-item-entry.blade.php**

```php
$logItem = $getState();
```

After the update, `$getState()` suddenly was `null` instead of pointing to the log item. It's probably related to the fact that we were relying on providing an empty string to the data path, which would previously be pointing to the individual entry. That doesn't seem to be the case anymore.

A workaround was changing the view to this instead:

```php
$logItem = data_get($getRecord(), $getStatePath()); // statePath: log.0
```

I wonder what is the recommended way of changing the component/entry view  in a `RepeatableEntry` without using `schema`? I couldn't find it in the docs when I looked for it, so this was the easiest way I could find at the time.  

cc @danharrin 🙏

danharrin:
I think part of the problem there is that entries shouldn't have empty names like `''`. In your case, you can skip RepeatableEntry, and just have a ViewEntry for `log` which has a loop through `$getState()` and renders reach log item entry

dbpolito:
Following up on my earlier reproduction: this is unrelated to entries with empty names.

`RepeatableEntry::getChildComponentContainers()` explicitly handles model items by calling `$container->record($itemData)`. However, when the root Infolist uses `state()` and contains a collection of models, the updated `HasState::getState()` resolves `getStatePath(isAbsolute: false)` against the root array.

It therefore looks for paths such as `name` or `childrenTree` directly on the root state and ignores the current item record assigned by `RepeatableEntry`.

Minimal reproduction:

```php
$infolist->state([
    'items' => collect([$model]),
]);

RepeatableEntry::make('items')
    ->schema([
        TextEntry::make('name'),
    ]);
```

This rendered correctly before #18809 and now renders blank. Explicitly resolving state from the assigned record works around it:

```php
TextEntry::make('name')
    ->state(fn (Model $record) => $record->name);
```

Is a collection of models inside array-backed Infolist state intentionally unsupported? If so, the explicit model handling in `RepeatableEntry` and the expected migration path should be clarified. Otherwise, state resolution should use the assigned item record for model items.

danharrin:
Would using an array instead of a collection cause problems for you? Ideally state is normalised before it enters the schema

dbpolito:
Using a fully normalized array via `toArray()` fixes state resolution, but it does cause problems for this schema because it relies on the current `MenuItem $record` for URLs, nested relationships, full-path formatting, and favorite actions.

Using `$collection->all()` would only produce an `array<Model>` and appears to retain the same issue, since the child entry still resolves relative state against the root array.

Is the intended behavior that model items are only supported when the root Infolist uses `record()`, while `state()` must contain arrays recursively? If so, that answers the compatibility question. I was confused because `RepeatableEntry` explicitly handles model items with `$container->record($itemData)`.

dbpolito:
We’re using explicit `state()` callbacks to work around the issue, so we’re unblocked. I mainly wanted to point out the regression since it may affect others using model collections inside array-backed Infolist state, especially if the requirement to normalize state recursively is not clear in the documentation.

danharrin:
@dbpolito have you tried 4.x? I can't seem to be able to reproduce the same problem, the entire state internals for schemas were rewritten to handle more edge cases like that

dbpolito:
@danharrin not yet... i have quite a big app and that tailwind upgrade is being challenging to upgrade... hopefully i do it soon... 

it seems very edgy on v3, and if no one else is being bitten, all good i guess

PR #18844, opened 2026-01-03
[4.x] UI Change: Increase opacity for disabled buttons, add not-allowed cursor

Mrkbingham:
## Description

When viewing buttons either in a group, or co-located next to each other - in particular in the context of a form - it can be difficult to discern when they're disabled.  The default cursor acts as a small hint for users on desktop devices, however for mobile users this is particularly challenging to discern.

### Current Disabled UI

**Light mode:** 

Screenshot 2026-01-03 at 11 40 53 AM

**Dark Mode:**

Screenshot 2026-01-03 at 11 50 15 AM

**Un-grouped:**

Screenshot 2026-01-03 at 12 04 59 PM

### Proposed Disabled UI

**Light mode:** 

Screenshot 2026-01-03 at 11 40 45 AM

**Cursor** (only for buttons without tooltips):
Screenshot 2026-01-03 at 12 13 37 PM

**Dark Mode:**
Screenshot 2026-01-03 at 11 50 42 AM

**Un-grouped:**

Screenshot 2026-01-03 at 12 04 50 PM

As you can see, the contrast between active elements, labels and the disabled buttons is quite a bit more pronounced.  Of course, this would be a substantive change to the framework as it would impact every disabled button.  My belief is that this communicates a bit more effectively to users what can and cannot be modified in a form, preventing likely confusion regardless of device.

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/CanExportRecords.php
+++ b/packages/actions/src/Concerns/CanExportRecords.php
@@ -35,7 +35,6 @@ use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Bus;
 use Illuminate\Support\Number;
-use Illuminate\Support\Str;
 use Livewire\Component;
 use LogicException;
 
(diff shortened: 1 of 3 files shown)


danharrin:
Thanks, I think I am going to stick with how we have it currently though, there are many components that would need to be updated for this to be applied consistently and I think it would be too big of a change to do. Feel free to do it in your own custom theme CSS though.

Note: I would suggest that you open a discussion and tag me in these sorts of proposals first to avoid wasted time making the PR.

Mrkbingham:
@danharrin sounds good!  I'll keep the discussion vs PR in mind next time I come across a more wide-ranging change, thanks for the feedback.

PR #18847, opened 2026-01-03
Feature: Allow UnitEnum in authorization

webard:
## Description

Since [Laravel Gate contract](https://github.com/laravel/framework/blob/12.x/src/Illuminate/Contracts/Auth/Access/Gate.php) allows UnitEnum as a valid `$ability` argument, I think Filament should allow `UnitEnum` as well.

This can be helpful for developers who want to use Enums to control permissions instead of plain strings.

## Visual changes

None

## Functional changes


danharrin:
Sorry, I think this is too risky to merge, it could be a breaking change for those overriding these methods on their resource classes.

webard:
Thank you @danharrin - that makes sense to avoid breaking existing apps.

1. Would you consider accepting changes to HasAuthorization in Filament v5?
2. Would you consider accepting changes to the get_authorization_response() / authorize() helpers in v4? Since users can’t change the function signatures anyway, this shouldn’t be a breaking change, and it would let developers use Enums in their authorization flow.

Of course I prepare PRs after your response.

danharrin:
Yes to both of those 👍

PR #18848, opened 2026-01-03
Feature: check ability existence when Auth strict mode is true

webard:
## Description

When `$shouldCheckPolicyExistence` is set to false, the Gate takes full control over authorization without involving policies.

However, if no Gate definition exists and `strictAuthorization()` is set to true, the issue is silently ignored instead of throwing an Exception.

This PR fixes that behavior.

## Functional changes


danharrin:
Let me know if you think I'm wrong, but I believe this would break some cases. `$shouldCheckPolicyExistence` is not just used when a policy doesn't exist. A policy may exist, and in that case `inspect()` still works, but `has()` will return false. This means that the check will then throw an exception even if the policy exists.

webard:
@danharrin I dug deeper into Laravel code and you’re right - `has` method does not check for policy existence.

 I think I should check that `Gate::getPolicyFor($model)` is not null to cover described situation.

```php
if (
    Filament::isAuthorizationStrict()
    && ! Gate::forUser($user)->has($action)
    && ! Gate::getPolicyFor($model) === null
) {
    throw new LogicException("Strict authorization mode is enabled, but no ability [".enum_value($action)."] is defined.");
}
```

danharrin:
That is checking policy existence though, right? So surely if the user doesn't want that they would not have enabled that feature?

webard:
Correct, but when `isAuthorizationStrict` is set to true and `shouldCheckPolicyExistence` is set to false, one of the following must exist:

- ability
- policy

Or only an ability, if user has intentionally decided to disable policy existence checks, following your approach.

At the end I think that developer should be notified when some Gate definitions are missing, when `isAuthorizationStrict` is true.

danharrin:
Sure, I think I know where you're coming from now. Feel free to open a new PR with that refined implementation

PR #18850, opened 2026-01-04
feature: File upload aspect ratio forced cropping and validation

danharrin:


trovster:
This feature broke the aspect ratio values I was using for infolists, which were calculated as decimals.

```php
protected function aspectRatio(): string
{
    return strval($this->height() / $this->width());
}
```

danharrin:
@trovster can you expand please, what do you mean by infolist aspect ratios?

trovster:
Sorry, that was the wrong term. The display of the existing image when viewing the form. I have the following for my uploads. Since this merge, the *panel* display was not using the correct dimensions.

```php
FileUpload::make('still')
    ->panelAspectRatio($this->aspectRatio())
    ->imageCropAspectRatio($this->aspectRatio())
    ->imageEditorAspectRatios([
        $this->aspectRatio(),
    ]);

protected function aspectRatio(): string
{
    $height = 600;
    $width = 800;
    return strval($height / $width);
}
```

There are now `panelAspectRatio()` and `itemPanelAspectRatio()` options. It seems the cropping still works as I expect, but the panel isn't using the correct size. It seems it is `normalizeAspectRatio` which is causing the issue, as the aspect isn't in the **width:height** format.

I'm confused with all these image crop options now!

danharrin:
Hi @trovster, I think bfa8d28d68ebf7bf0776bca03fdb4c4ca92c583e should fix that

trovster:
Thanks @danharrin, I knew it'd be something simple to fix.

PR #18855, opened 2026-01-04
[4.x] Feature: Allow UnitEnum in auth helpers

webard:
## Description

According to #18847, this PR allows `UnitEnum` as an `$action` parameter of `authorize` and `get_authorization_response` helper functions. 

## Functional changes

### DIFF
--- a/packages/panels/src/helpers.php
+++ b/packages/panels/src/helpers.php
@@ -9,6 +9,7 @@ use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Gate;
 use LogicException;
+use UnitEnum;
 
 if (! function_exists('Filament\authorize')) {
     /**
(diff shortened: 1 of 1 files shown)


packages/panels/src/helpers.php
```diff
@@ -36,7 +39,7 @@ function get_authorization_response(string $action, Model | string $model, bool
 
         $policy = Gate::getPolicyFor($model);
 
-        if (filled($policy) && method_exists($policy, $action)) {
+        if (filled($policy) && method_exists($policy, enum_value($action))) {
```
danharrin: This `enum_value()` function is marked as internal in the Laravel codebase, we shouldn't use it. Introduce a match statement like what they have inside that function to handle fetching the correct value.

danharrin (changes requested):

packages/panels/src/helpers.php
```diff
@@ -49,7 +57,7 @@ function get_authorization_response(string $action, Model | string $model, bool
 
             throw new LogicException(blank($policyClass)
                 ? "Strict authorization mode is enabled, but no policy was found for [{$model}]."
-                : "Strict authorization mode is enabled, but no [{$action}()] method was found on [{$policyClass}].");
+                : 'Strict authorization mode is enabled, but no [' . $policyMethod . "()] method was found on [{$policyClass}].");
```
danharrin:
```suggestion
                : "Strict authorization mode is enabled, but no [{$policyMethod}()] method was found on [{$policyClass}].");
```

danharrin (changes requested):

danharrin (approved):

PR #18861, opened 2026-01-05
Revert "Fix Textarea autosize causing scroll anchoring issues in Firefox"

### LINKED ISSUE #18860: Textarea with autosize() ignores rows() on initial render (regression from #18595)
### Package

filament/filament

### Package Version

v4.4.0

### Laravel Version

v12.0.0

### Livewire Version

v3.6.3

### PHP Version

PHP 8.4.14

## Description
After PR #18595 ("Fix Textarea autosize causing scroll anchoring issues in Firefox"), textareas with `->autosize()` ignore the `->rows()` setting on initial render, displaying as a single line instead of the specified number of rows.

## Steps to Reproduce
1. Create a form with a Textarea field:
   ```php
   Textarea::make('script')
       ->rows(10)
       ->autosize()
   ```
2. Load the page
3. The textarea appears as a single line (~36px) instead of 10 rows (~252px)

## Expected Behavior
The textarea should initially display with the height specified by `->rows(10)`, then grow as content is added.

## Actual Behavior
The textarea displays as a single line. The `rows` attribute is present in HTML but the height is overridden to ~36px.

## Root Cause Analysis
In `textarea.js`, the `resize()` function is called via `x-intersect.once` which immediately overwrites the initial height set by `setInitialHeight()`:

1. `init()` → `setInitialHeight()` sets wrapper height to `15.75rem` (252px) ✓
2. `x-intersect.once="resize()"` triggers
3. `resize()` sets textarea height to `0px` temporarily
4. For empty textarea, `scrollHeight` = ~36px
5. Wrapper height is overwritten to `36px` ✗

Before #18595, `setInitialHeight()` was called at the start of `resize()`, which preserved the minimum height. After #18595, this call was removed.

## Suggested Fix
Either:
1. Restore `setInitialHeight()` call at the start of `resize()`, or
2. Add minimum height check in `resize()`:
   ```javascript
   const minHeightPx = parseFloat(initialHeight) * 16
   const newHeightPx = parseInt(newHeight)
   const finalHeight = Math.max(minHeightPx, newHeightPx) + 'px'
   ```

## Additional Context
- Browser: Chrome, Firefox, Safari (all affected)
- The bug does NOT occur when `->autosize()` is removed - `->rows(10)` works correctly without autosize

### Expected behavior

The textarea should initially display with the height specified by `->rows(10)` (approximately 252px / 15.75rem), then grow dynamically as content is added.

### Steps to reproduce

1. Create a form with a Textarea field using both `rows()` and `autosize()`:
```php
Textarea::make('script')
    ->rows(10)
    ->autosize()
```
2. Load the page containing this form
3. Observe the textarea height - it appears as a single line (~36px) instead of 10 rows (~252px)
4. Remove `->autosize()` and reload - the textarea now correctly shows 10 rows

https://github.com/eabay/filament-autosize-issue

### Relevant log output

```shell

```

danharrin:
Reverts filamentphp/filament#18595
Fixes #18860


Saracaen:
@danharrin Would it have been possible to solve this differently without breaking the scroll anchoring again, for example by adding a Math.max on the newHeight?

Something like:

```
const minHeightPx = parseFloat(initialHeight) * parseFloat(getComputedStyle(document.documentElement).fontSize)
const newHeight = Math.max(contentHeight, minHeightPx) + 'px'
```

I understand that the scroll anchoring issue is more of an edge case compared to the core rows functionality not working, so I understand the decision to revert the fix for that. But it would be nice to find a way to fix both issues, and I'm hoping that a fix like this could achieve that.

If you want I could test it and create a PR for it, provided that my theoretical fix works.

danharrin:
Yeah I don't really have time to test myself, you're welcome to submit a PR after you test it if you wish

Saracaen:
I have verified the fix locally, and both the scroll anchoring and the row height now work as expected, so the PR is ready.

PR #18866, opened 2026-01-05
Fix/table polling modals 

halaibrahim867:
## Description

Polling in Filament tables continues even while a table action modal is open. This can cause UI glitches, such as modals closing or action state being reset, because Livewire re-renders interrupt the action.  

This update pauses polling automatically whenever a table action is mounted (`mountedAction` or `mountedActions`) and resumes it once the action is closed, ensuring modals and actions work smoothly.


danharrin:
Can you please create a reproduction repository for the issue you are experiencing?

halaibrahim867:
Yes — a reproduction repository is already available.

Reproduction repository:
https://github.com/CharlieEtienne/poll-empty-search

danharrin:
I do not believe that this has been tested manually since `CanPollRecords` does not exist on the Livewire component so does not have access to `$this->mountedActions` ever. I think the component should still be able to make updates without causing the bug, so I think the solution lies elsewhere.

PR #18873, opened 2026-01-05
fix: ensure main container fills viewport with collapsible si…

Devonab:
This pull request aims to address the same issue reported in v4.x here: https://github.com/filamentphp/filament/pull/17646
for reference : https://github.com/Devonab/filament-easy-footer/issues/16#issuecomment-3694732307

## Description

When one sets up their panel using the one of following methods:

```
->sidebarCollapsibleOnDesktop()
->sidebarFullyCollapsibleOnDesktop()
```

The content is no longer at a minimum height of 100%, which can be a bit troublesome when one wishes to place an element in the renderHook PanelsRenderHook::FOOTER.

I replaced the `h-full` present in `packages/filament/packages/panels/resources/views/components/layout/index.blade.php`.

## Visual changes

Before :
image

After :
image

## Functional changes

### DIFF
--- a/packages/panels/resources/views/components/layout/index.blade.php
+++ b/packages/panels/resources/views/components/layout/index.blade.php
@@ -29,7 +29,7 @@
             @endif
             @class([
                 'fi-main-ctn w-screen flex-1 flex-col',
-                'h-full opacity-0 transition-all' => filament()->isSidebarCollapsibleOnDesktop() || filament()->isSidebarFullyCollapsibleOnDesktop(),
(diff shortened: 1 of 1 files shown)


danharrin:
I don't think we can merge this change unfortunately, it might break users' existing pages if the topbar height has been customised or something. This was less of a risk to merge early in a major version change like v4.

In your plugin, I would suggest injecting the CSS that does this to fix your user's problem.

Devonab:
Thank you for the follow-up, and all good, I understand. I will do what you suggest. it's a good idea as well :)

PR #18891, opened 2026-01-07
[4.x] Robots meta tag to prevent panel pages being indexed

chrispage1:
## Description

Firstly, Happy New Year! Hope you managed to get some time to relax @danharrin !

This is a simple addition to the Filament panels to prevent panel pages from being indexed by search engines. It adds a meta tag telling robots to not index the pages.

Currently, its possible for panel pages to be indexed within Google, exposing locations of Filament login URL's etc.

## Visual changes

NA

## Functional changes

NA

### DIFF
--- a/packages/forms/resources/js/components/textarea.js
+++ b/packages/forms/resources/js/components/textarea.js
@@ -41,7 +41,9 @@ export default function textareaFormComponent({
             const contentHeight = this.$el.scrollHeight
             this.$el.style.height = previousHeight
 
-            const minHeightPx = parseFloat(initialHeight) * parseFloat(getComputedStyle(document.documentElement).fontSize)
+            const minHeightPx =
+                parseFloat(initialHeight) *
+                parseFloat(getComputedStyle(document.documentElement).fontSize)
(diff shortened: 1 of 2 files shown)


ryangjchandler:
I wonder if this should be an opt-in feature or opt-out feature. For example, I'm building a platform inside of Filament so I don't care if the login page is indexed, if anything it could be useful for natural growth.

ziming:
Hope this is the default!

danharrin:
Yeah I definitely don't think this should be the default. I think the best approach I recommend is using a render hook for this, probably `HEAD_END`.

chrispage1:
@danharrin I've made some updates to allow it to be conditional - would you reconsider opening this or should I raise a new PR?

It allows the user to specify `->hideFromRobots()` against the panel and is backwards compatible by default being disabled.

danharrin:
Thanks, but I do not think I have had enough interest in this feature in the past to warrant introducing new configuration, especially when the existing render hook could be implemented in just one line. I might hear more interest in this in the future, and I may reconsider.

PR #18892, opened 2026-01-07
Feature: Sidebar always collapsed

webard:
## Description

This PR introduces a new Panel configuration method called sidebarAlwaysCollapsed().

When set to true, the sidebar navigation remains permanently collapsed to icons, with no option to expand it.

This can be particularly useful for systems that rely heavily on clusters, where the main navigation is used only to access high-level sections such as “PIM”, “CRM”, or “CMS”, and the actual resources are organized within clusters.

Additionally, this PR fixes an inconsistency in the mobile menu icon when `topbar()` is set to `false`. Previously, the icon was displayed as an arrow, while when `topbar()` was `true`, the icon was OutlinedBars3. This PR standardizes the icon behavior across both configurations.

Let me know if this change has a chance to be merged - if so, I’ll prepare the corresponding documentation updates.

## Visual changes

No arrow next to the logo when `sidebarAlwaysCollapsed()` is set to true

Fixed mobile menu icon when `topbar()` is `false`, before:
image

after:
image

## Functional changes


danharrin:
Thanks, but I am going to hold off on introducing new sidebar configurations for now, we already have quite a few and I haven't heard this one being requested very often. I may reassess in the future.

Are you sure that the bug icon fix you are suggesting doesn't affect any other situations? The arrow should still be used when there is a partially collapsible sidebar, but only on desktop

webard:
Sorry to hear that - unfortunately, there is no clean way to achieve similar behavior without “hacking” CSS.

@danharrin icons are fine, the issue is in the https://github.com/filafly/filament-phosphor-icons package.

PR #18899, opened 2026-01-07
[4.x] fix - handle Htmlable state in TextEntry/TextColumn toEmbeddedHtml blank check

SimonBroekaert:
## Description

When rendering the content of a `Filament\Forms\Components\RichEditor` field using a `Filament\Infolists\Components\TextEntry` when the model is set up with  the `Filament\Forms\Components\RichEditor\Models\Contracts\HasRichContent` contract and `Filament\Forms\Components\RichEditor\Models\Concerns\InteractsWithRichContent` trait, the **placeholder is never rendered**.

This is because the `blank($state)` check does return false because it is given a `Filament\Forms\Components\RichEditor\RichContentAttribute`.

This fix checks if the `$state` implements `Illuminate\Contracts\Support\Htmlable` and uses the rendered html in the blank check instead.

This fixes both `NULL` and `EMPTY` values.

## Functional changes

### DIFF
--- a/packages/infolists/src/Components/TextEntry.php
+++ b/packages/infolists/src/Components/TextEntry.php
@@ -185,7 +185,7 @@ class TextEntry extends Entry implements HasAffixActions, HasEmbeddedView
                 'fi-in-text',
             ]);
 
-        if (blank($state)) {
+        if (blank($state instanceof Htmlable ? $state->toHtml() : $state)) {
             $attributes = $attributes
                 ->merge([
                     'x-tooltip' => filled($tooltip = $this->getEmptyTooltip())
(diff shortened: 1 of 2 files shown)


danharrin (changes requested):
Does the same need to happen for `TextColumn`?

SimonBroekaert:
> Does the same need to happen for `TextColumn`?

Ah yes, the TextColumn can indeed also render the `RichContentAttribute`. I've tested it on my project and indeed, the placeholder is not rendering there either.

I've updated the PR to also add the change to that `blank` check.

Thanks! 🙏

danharrin (approved):

PR #18903, opened 2026-01-08
Fix achromatic colors producing incorrect hues in generatePalette

### LINKED ISSUE #18904: Achromatic colours produce incorrect hues in generatePalette
### Package

filament/filament

### Package Version

v4.x

### Laravel Version

v10.0.0

### Livewire Version

v3.0.0

### PHP Version

8.2.0

### Problem description

When converting achromatic colours like black (#000000), white, or greys to OKLCH, the chroma is zero and the resulting hue has no meaning. The current implementation of generatePalette extracts this meaningless hue and combines it with non‑zero chroma values from the palette constants, which causes the generated shades to take on red/pink tones instead of staying grey.

### Expected behavior

For achromatic colours such as black (#000000), white, and greys, converting to OKLCH yields a hue value of 0 because the chroma is 0. The current implementation of generatePalette extracts this hue and applies non‑zero chroma constants, resulting in palettes with unwanted red or pink tones instead of grayscale.
The function should detect achromatic input colours (chroma < 0.03) and keep the chroma at 0 in all generated shades. See PR #18903 for a proposed fix.

### Steps to reproduce

1. In a Filament project, call `Filament\Support\Colors\Color::generatePalette('#000000')`.
2. Observe that the returned palette contains shades with red or pink tones rather than neutral greys.
3. Repeat with other achromatic colours like `#ffffff` or `#888888` and note similar incorrect hues.

Expected: The palette should remain achromatic (greyscale) for these inputs.

https://github.com/filamentphp/filament

### Relevant log output

```shell

```

juliangums:
For achromatic colors like black (#000000), white, and grays, the OKLCH conversion produces a meaningless hue value (0) since chroma is also 0. The current code extracts this hue and applies non-zero chroma values in the generated palette, resulting in red/pink tones instead of grayscale. This fix detects achromatic input colors (chroma < 0.03) and keeps chroma at 0 for all generated shades.


danharrin:
Please update then snapshot tests

danharrin (approved):

PR #18910, opened 2026-01-08
[3.x] Allow HTML in radio and checkbox-list descriptions

cheesegrits:
## Description

Allow HTML in radio button and checkbox list descriptions.

## Functional changes

### DIFF
--- a/packages/forms/resources/views/components/checkbox-list.blade.php
+++ b/packages/forms/resources/views/components/checkbox-list.blade.php
@@ -215,7 +215,7 @@
                                 <p
                                     class="fi-fo-checkbox-list-option-description text-gray-500 dark:text-gray-400"
                                 >
-                                    {{ $getDescription($value) }}
+                                    {!! $getDescription($value) !!}
                                 </p>
                             @endif
                         </div>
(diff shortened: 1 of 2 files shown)


cheesegrits:
I would also like to get this into v4.  Not sure if changes on v3 are being merged forward, or I would need to submit a separate PR on v4.

danharrin:
Are you sure this isnt already possible by passing an HtmlString object as the description? We don't use `{!!` like this, HTML support is opt-in because it opens you up to XSS attacks, so you need to sanitise any user input in there.

danharrin:
I have just checked and I believe that the feature to use `HtmlString` / `Htmlable` does exist here.

cheesegrits:
@danharrin that first image is what happens when I provide an HtmlString().

There is a slight complication in that the descriptions are coming from an enum's getDescription() method, so maybe that's causing my issue.  I'll dig some more.

danharrin:
You should double check that the return type of `getDescription()` is correct so the `HtmlString` objects aren't cast to strings.

danharrin:
And if it doesn't work for you, I would merge a different PR that makes it work, but we can't merge a solution with `{!!`

PR #18918, opened 2026-01-09
Honor input color's lightness in generatePalette

juliangums:
The current generatePalette method extracts only the hue from the input color and applies fixed lightness values. This means:

- A bright color like #49D359 (lightness ~0.77) produces the same shade 500 lightness (0.68) as any other green
- A dark blue like #1a365d (lightness ~0.32) produces shade 500 at 0.68 lightness instead of respecting the dark tone the user chose

This fix:
- Extracts the input color's actual lightness
- Sets shade 500 to match the input's lightness
- Scales lighter shades (50-400) proportionally toward 0.98
- Scales darker shades (600-950) proportionally toward 0.15
- Scales chroma proportionally based on the input color's chroma

Result: When a user picks a dark blue, buttons appear dark blue. When they pick a bright green, buttons appear bright green. The palette is still perceptually uniform via OKLCH.

## Functional changes


danharrin:
Before:

Screenshot 2026-01-09 at 10 43 47

After:

Screenshot 2026-01-09 at 10 44 02

Aside from this being too big of a breaking change, I think there are many problematic palettes now generated where there is too little differentiation from 0-500 or 500-950.

I understand where you're coming from, and the current generator isn't perfect, but it has to balance many different base colors.

I would suggest just hard-coding the array of colors instead of using `generatePalette()` if you are unhappy.

juliangums:
@danharrin I get where you're coming from as well. It currently definitely generates palettes that are more pleasing to the eye. But for example in my app I have a setting so the user can set their brand colour and it is quite unexpected if the buttons are much lighter or darker than what they picked. I had users complain about it. I will create my own method to generate the palette for now as you suggested. Maybe this could be a setting to be set on the panel in future to generate colours either more harmonic (default, as it is now) or true to input? Happy to create a PR if wanted but I guess this would also need some more explanation in the docs etc

danharrin:
Thanks for the offer, I think since this is just a helper function that can be replaced though, its ok for a third party package to provide alternative generators, but not in core.

PR #18920, opened 2026-01-09
[4.x] Wizard programmatic step navigation without query string

andrefelipe18:
## Description

The Wizard component currently only supports programmatic forward navigation via the `nextStep()` method, which emits a browser event handled by Alpine. There is no native way to programmatically navigate to a specific step or move backwards without relying on query strings or directly mutating Alpine state.

This pull request introduces a new exposed Livewire method that allows navigating to a specific wizard step in a controlled and event-driven way. The implementation follows the existing Wizard architecture and event conventions, keeping the Wizard as the single source of truth for step state.

## Visual changes

No visual changes.

## Functional changes

### DIFF
--- a/packages/schemas/resources/js/components/wizard.js
+++ b/packages/schemas/resources/js/components/wizard.js
@@ -48,6 +48,23 @@ export default function wizardSchemaComponent({
             this.scroll()
         },
 
+        goToStep(stepKey) {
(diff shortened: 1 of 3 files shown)


danharrin (approved):

danharrin:
I have made it non-exposed because I don't see any reason for it to be exposed, especially when it also has the ability to skip validation.

PR #18934, opened 2026-01-10
Fix: Leak model data in Exception

webard:
## Description

Model data instead of model name can be visible in Exception message in some cases.

## Visual changes

None
## Functional changes


danharrin (changes requested):
How about line 72 too?

webard:
> How about line 72 too?

Fixed too

danharrin (approved):

PR #18944, opened 2026-01-11
Chore: PHPDoc for Actions component

webard:
## Description

`Actions` component can contain any `Component`, however, PHPStan currently reports an error when a developer uses anything other than `Action` or `ActionGroup`.

Example:
```php
Actions::make([
    Flex::make([
        Action::make('attach')
            ->iconButton()
            ->icon(Phosphor::Paperclip),

        Action::make('emoji')
            ->iconButton()
            ->icon(Phosphor::Smiley),
    ]),

    Flex::make([
        Action::make('record')
            ->iconButton()
            ->icon(Phosphor::MicrophoneFill)
            ->color(Color::Red),

        Action::make('send')
            ->iconButton()
            ->icon(Phosphor::PaperPlaneRightFill),
    ]),
])
    ->alignBetween()
```

New PHPDoc is taken from `Flex` component.

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/schemas/src/Components/Actions.php
+++ b/packages/schemas/src/Components/Actions.php
@@ -33,7 +33,7 @@ class Actions extends Component
     const BELOW_CONTENT_SCHEMA_KEY = 'below_content';
 
     /**
-     * @param  array<Action | ActionGroup> | Closure  $actions
+     * @param  array<Component | Action | ActionGroup | string | Htmlable> | Closure  $actions
      */
     final public function __construct(array | Closure $actions)
     {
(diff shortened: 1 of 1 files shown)


danharrin:
Honestly `Actions` doesn't really do anything if you are just embedding Flex, I think you might as well just replace it with a Flex. I think it makes sense for Actions to be strictly to directly embed action objects only.

webard:
@danharrin sounds reasonable. Is there any reason that `Flex` does not use `HasAlignment` trait? Because Actions and Flex are flexbox under the hood, but `HasAlignment` is available only in Actions.

danharrin:
Yeah it probably should, please PR

PR #18947, opened 2026-01-12
[4.x] Enhance text component to support icons

alanost:
## Description

Adds icon support for `Text` components in normal (non-badge) mode.

## Visual changes

Before: Icons only worked with `->badge()`  
After: Icons work in all modes

```php
Text::make('Label')
    ->icon('heroicon-m-star')
    ->size(TextSize::Large) // Icon scales automatically
```

image

## Functional changes

### DIFF
--- a/packages/schemas/resources/views/components/text.blade.php
+++ b/packages/schemas/resources/views/components/text.blade.php
@@ -17,8 +17,8 @@
     $weight = $getWeight();
     $fontFamily = $getFontFamily();
 
-    if (!$iconSize && $size instanceof TextSize) {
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks, I understand why you are suggesting this, but the idea behind prime components is that they are the simplest form of component, and you can combine them to make more complex things. For example, you can add a `Text` and and `Icon` component aside each other to get the same effect. The reason why we have the `icon()` method on the class already is just for badges, as you can't really make those any simpler since they have a background color and stuff to group them. Hopefully that makes sense.

PR #18959, opened 2026-01-14
feature/action-can-be-defined-as-closure

demobiel:
## Description

I have some specialised service classes, that I would like to be able to also return a Filament Action. I know this can be done by creating an Action class that calls the service classes, but for small projects this increases my number of classes :)
So if possible, I would like to add the feature that you can also specify an action by calling a closure that returns an action.

In that way I can do: 
```
{
    return $table
        ->recordActions([
            fn($record) => app(PdfRenderService::class,['document' => $record])->downloadAction()
        ]);
```
Where the dowloadAction method returns an Action.

## Visual changes

none

## Functional changes

### DIFF
--- a/packages/tables/docs/04-actions.md
+++ b/packages/tables/docs/04-actions.md
@@ -46,6 +46,22 @@ All methods on the action accept callback functions, where you can access the cu
 
 <AutoScreenshot name="tables/actions/simple" alt="Table with actions" version="4.x" />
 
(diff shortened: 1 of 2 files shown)


danharrin:
Thanks, but this wouldn't work since `pushRecordActions()` is called too early in the lifecycle to receive parameters like `$record`, evaluation needs to happen within getters instead of setters.

PR #18969, opened 2026-01-15
[4.x] fix: Disable wizard mode when steps resolve to null or empty array

### LINKED ISSUE #17: [Feature]: Ability to use a wizard in the modal form
### What happened?

The plugin is looking for a form method in the transition class. 
The form is inserted here : 
vendor/a909m/filament-statefusion/src/Concerns/ResolvesActionAttributes.php

```php 
protected function setActionAttributes(): void
    {

        // Model
        $this->requiresConfirmation();
        $this->modalDescription(fn () => $this->getTooltip());
        $this->modalIcon(fn () => $this->getIcon());
        $this->modalIconColor(fn () => $this->getColor());

        // Form
        $this->schema(function () {
            if ($this->hasTransitionClass() && method_exists($this->getTransitionClass(), 'form')) {
                return app($this->getTransitionClass())->form();
            }

            return null;
        });
}
```
This way we can't use the specific "->steps()" method to inject a wizard : 
https://filamentphp.com/docs/4.x/actions/modals#rendering-a-wizard-in-a-modal

This methods allows to hide submit modal action until the last step is reached.

### How to reproduce the bug

Insert a wizard into a modal

### Package Version

2.2.3

### PHP Version

8.4

### Laravel Version

12

### Which operating systems does with happen with?

_No response_

### Notes

_No response_

A909M:
## Summary

This PR fixes an issue where modal footer actions are hidden when `steps()` receives a Closure that returns `null` or an empty array or null.

## Problem

When using `steps()` with a Closure that conditionally returns `null`:

```php
->steps(function () {
    if ($someCondition) {
        return [Step::make('...'), ...];
    }
    
    return null; // This hides the modal footer actions!
})
```

The wizard mode is activated (`isWizard = true`) even when the Closure returns `null`, causing `getModalFooterActions()` to return an empty array and hide the submit/cancel buttons.

## Solution

- Evaluate the steps in `getSchema()` and disable wizard mode (`isWizard = false`) when steps resolve to `null` or an empty array
- Allow `steps()` to accept `null` and handle it gracefully

## Changes

- `HasSchema.php`: Added logic to disable wizard mode when schema is null or empty
- `HasWizard.php`: Added `$steps` property and updated `steps()` to accept nullable values

## Note

⚠️ This is a working proof-of-concept. The implementation may not follow Filament's coding standards perfectly. Feel free to refactor or improve the approach if there's a better way to handle this—I had limited time to polish the code.

## Related

- Fixes: https://github.com/A909M/Filament-StateFusion/issues/17

## Functional changes


danharrin:
Unfortunately we are not able to merge this as it is a breaking change, there are tests that used to work before and now fail in our test suite.

A909M:
> Unfortunately we are not able to merge this as it is a breaking change, there are tests that used to work before and now fail in our test suite.

@danharrin  Thanks for clarifying.

If I fix the failing tests and avoid the breaking change, would you be open to reviewing the PR again? I didn’t have time to look into the tests earlier, but I can work on that now.

danharrin:
If you can make it so that we aren't setting properties from getter methods which is a bit of an anti-pattern, and you aren't forcing the schema function to run twice, then sure.

PR #18977, opened 2026-01-15
Allow chart widgets to have a defined height

dayio:
## Description

Add height control for chart widgets

In V3, full-width widgets displayed correctly by filling the entire screen. In V4, they are too slim and barely readable.

Adding a `height` property would allow proper control over widget display and ensure readability.

## Visual changes

V3 

image

V4

image

V4 + Height

image

## Functional changes

### DIFF
--- a/packages/widgets/docs/03-charts.md
+++ b/packages/widgets/docs/03-charts.md
@@ -226,6 +226,14 @@ Alternatively, you may disable polling altogether:
 protected ?string $pollingInterval = null;
 ```
 
+## Setting a chart height
+
+To control the chart size, you can set a fixed height using the `$height`:
+
+```php
+protected ?string $height = '400px';
+```
+
 ## Setting a maximum chart height
 
 You may place a maximum height on the chart to ensure that it doesn't get too big, using the `$maxHeight` property:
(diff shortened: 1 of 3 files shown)


danharrin:
Please share a reproduction repository for this before I merge, as I cannot reproduce the problem myself.

dayio:
> Please share a reproduction repository for this before I merge, as I cannot reproduce the problem myself.

You can reproduce the behavior here : https://github.com/Bouhnosaure/filament-height-test

To quickly reproduce the PR, after installing the project and creating your user, go to the homepage there will be the widget.

Then here `vendor/filament/widgets/resources/views/chart-widget.blade.php` at the line `82` just add just above the `max-height` style property:

`'height: ' . $maxHeight => filled($maxHeight),` 

Reload, you'll see the difference in readability.

Thanks !

dayio:
Before 
Capture d’écran du 2026-01-15 17-28-54

After
Capture d’écran du 2026-01-15 17-29-01

marvinrabe:
@danharrin When setting the maxHeight we have the same problem with charts. What is specified in maxHeight is for some reason ignored and the chart gets squished into a very little stripe. Way way smaller than what would be allowed with the maxHeight setting. When leaving the maxHeight it will be bigger, but because it just fills space it might end up way to big.

danharrin:
Fixed the original bug in #19029 so we don't need to introduce any new features as workarounds

dayio:
Happy to help! Have a nice day.

PR #18994, opened 2026-01-17
Allow customizing the form component class in table filters

angus-mcritchie:
## Summary
- Adds `formComponent()` method to `SelectFilter` to allow using custom `Select` components
- Includes test and documentation

## Changes

This PR adds the ability to customize the form component class used by `SelectFilter` (and by extension `TernaryFilter` which inherits from it), following the same pattern already established in the base `Filter` class.

**Use case:** Users who have extended the base `Select` component with additional functionality (custom styling, behavior, or third-party integrations) can now use their custom component in select filters.

## Usage

```php
use App\Forms\Components\CustomSelect;
use Filament\Tables\Filters\SelectFilter;

SelectFilter::make('status')
    ->formComponent(CustomSelect::class)
    ->options([
        'draft' => 'Draft',
        'reviewing' => 'Reviewing',
        'published' => 'Published',
    ])
```

This also works with `TernaryFilter`.


danharrin:
Thanks, I think I'm going to decline for now though and suggest you use `->form()` instead as it provides more flexibility

PR #19000, opened 2026-01-18
Fallback to existing mention label if provider does not return a label

gehrisandro:
## Description

This PR changes the mention rendering to fallback to the stored label in the database according to the discussion here: https://github.com/filamentphp/filament/discussions/18984

In addition the label gets persisted to the database even if a `MentionProvider` is available (otherwise there is nothing to fall back to).

## Visual changes

Before (when user is not available anymore):
CleanShot 2026-01-18 at 12 08 01@2x

After:
CleanShot 2026-01-18 at 12 10 23@2x

_When the provider returns a label (in my case, the user still exists), there is no visual change._

## Functional changes

- [ ] ~~Documentation is up-to-date.~~(no change required)

### DIFF
--- a/packages/forms/src/Components/RichEditor/RichContentRenderer.php
+++ b/packages/forms/src/Components/RichEditor/RichContentRenderer.php
@@ -294,10 +294,6 @@ class RichContentRenderer implements Htmlable
                 return;
             }
 
-            if (filled($node->attrs->label ?? null)) {
-                return;
-            }
-
             $char = $node->attrs->char ?? '@';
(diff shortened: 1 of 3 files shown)


danharrin:
Can you please rebase this on 4.x?

gehrisandro:
> Can you please rebase this on 4.x?

Done

danharrin (approved):

danharrin:
@gehrisandro I removed a block that you modified because I don't think we need to remove the label from the mention in any case now

PR #19001, opened 2026-01-18
[5.x] docs(ai): update installation instructions for Blueprint v2.0

jeffersongoncalves:
Update composer commands in the documentation to specify the Blueprint version as "^2.0". Add a warning and alternate installation steps for Windows PowerShell users to address version constraint issues with `^`.


danharrin (commented):
I'm wondering, why do we need to specify the version as 2.0? In my opinion, we should leave out the version, so Blueprint is installed with whatever version is compatible with your current Filament installation. Let me know if I am missing something.

PR #19005, opened 2026-01-18
[4.x[ chore(git): add .gitignore and .gitattributes for subtree packages

jeffersongoncalves:
Introduce `.gitignore` and `.gitattributes` files for various subtree packages to manage ignored files and streamline exports during subtree splitting.


danharrin (changes requested):
Thanks, but I think most of these changes do not do anything in practice. Please remove all the new .gitignore files as none of the paths mentioned exist in each package, and are caught by the monorepo .gitignore. I think the only thing we should keep is the `/.github export-ignore` line in `.gitattributes` for each package, the rest can be removed.

danharrin:
Also its worth rebasing this on 4.x instead of 5.x

PR #19008, opened 2026-01-19
Update documentation for date formatting in TextColumn

mrezauli:
Clarified usage of date formatting methods in tooltips and emphasized the importance of chaining `dateTime()` before `since()` for proper output.

## Description

Here found the missing link of chaining methods of `dateTime()` before `since()`

## Visual changes

filamentAfter
filamentBefore

## Functional changes


danharrin:
Thanks, but I do not think this is correct. `dateTime()` is not required to use `since()`, you just need `since()`, and this is going to confuse people further I think. Let me know if `since()` doesn't work for you without `dateTime()` by opening a bug report.

PR #19019, opened 2026-01-20
Allow configuring default grid breakpoint

angus-mcritchie:
## Summary

This PR adds the ability to configure the default grid breakpoint used when passing integers to grid methods like `columns(2)`, `columnSpan(3)`, `columnStart(2)`, `columnOrder(1)`, and `grid(3)`.

Previously, the breakpoint was hardcoded to `'lg'`. Now users can:

1. **Configure via `config/filament.php`:**
   ```php
   'default_grid_breakpoint' => '2xl',
   ```

2. **Set programmatically (e.g., in a service provider):**
   ```php
   use Filament\Support\Facades\FilamentGrid;

   FilamentGrid::defaultBreakpoint('xl');
   ```

## Changes

- Added `GridManager` class to manage the default breakpoint
- Added `FilamentGrid` facade for easy access
- Updated all grid-related methods to use the configurable breakpoint
- Added config option with documentation
- Added tests for both config and programmatic approaches
- Updated layouts documentation

## Why?

Some applications need consistent grid behavior across different breakpoints. For example, an app optimized for large displays may want `'2xl'` as the default instead of `'lg'`.

This is fully backward compatible - the default remains `'lg'`.

### DIFF
--- a/packages/schemas/docs/02-layouts.md
+++ b/packages/schemas/docs/02-layouts.md
@@ -23,7 +23,7 @@ You may also [create your own custom layout components](custom-components#custom
 
 All layout components have a `columns()` method that you can use in a couple of different ways:
 
-- You can pass an integer like `columns(2)`. This integer is the number of columns used on the `lg` breakpoint and higher. All smaller devices will have just 1 column.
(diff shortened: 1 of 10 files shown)


danharrin:
Thanks, I am going to pass on this. To be honest, I don't really see the value of it, and I think it just makes the configuration feel more bloated than it does not. I may revisit in the future, but this is the first suggestion I have heard of something like this before.

PR #19025, opened 2026-01-20
Fix password reset link not displaying on login page

abournavaud:
## Description

The password reset link on the login page was no longer displaying when using the `hint()` method. This PR fixes the issue by using the new `afterLabel()` slot introduced in Filament v4, which properly positions the "Forgot password?" link next to the password field label.

## Visual changes

The password reset link now appears correctly after the password label instead of not showing at all.

## Functional changes


danharrin:
I cannot reproduce the problem, `hint()` should still work. Please open an issue with a reproduction repository.

Screenshot 2026-01-20 at 13 56 52

PR #19031, opened 2026-01-21
[4.x] Render hook classes

ryangjchandler:
## Description

Render hooks are commonly used to customize the Filament experience further, especially inside of panels.

One limitation that I personally find with the `renderHook()` method is that certain views require custom data and placing all of that logic inside of the panel provider or a service provider can become rather messy.

A workaround for this that already exists is creating a class to handle the rendering, but it still requires some manual wiring to work correctly, especially when dependency injection is involved.

This pull request introduces the concept of a `RenderHook` class with support for auto-discovery.

You create a class that extends `Filament\View\RenderHook` and implement the necessary methods. You specify the hook(s) you want to render for and the View or string that should be rendered.

Classes are resolved from the container so dependency injection works as expected and you can pull in service classes for data retrieval etc.

By adding `->discoverRenderHooks()` onto your `Panel` object, specifying the directory and namespace, you get the real benefit of auto-discovered render hooks, removing the need for that superfluous method call.

## Visual changes

N/A

## Functional changes

### DIFF
--- a/packages/panels/src/Panel/Concerns/HasRenderHooks.php
+++ b/packages/panels/src/Panel/Concerns/HasRenderHooks.php
@@ -4,6 +4,10 @@ namespace Filament\Panel\Concerns;
 
 use Closure;
 use Filament\Support\Facades\FilamentView;
+use Filament\View\RenderHook;
+use Illuminate\Filesystem\Filesystem;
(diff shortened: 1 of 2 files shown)


danharrin:

PR #19035, opened 2026-01-21
fix: One time code input attributes

### LINKED ISSUE #18990: 2FA always asking for code even we add the code
### Package

filament/filament

### Package Version

v5.0.0

### Laravel Version

v.12+

### Livewire Version

_No response_

### PHP Version

8.4

### Problem description

Even I am adding 2fa code from both email and authenticator app. Its always asking for 2FA code. 

http://shares.profitablemedia.com/20260117r954068b350.png

Here is my User Model:
`
namespace App\Models;

use Filament\Models\Contracts\FilamentUser;
use Filament\Panel;
use Illuminate\Foundation\Auth\User as Authenticatable;
use Illuminate\Notifications\Notifiable;
use Illuminate\Database\Eloquent\Factories\HasFactory;

// Filament v4 MFA Interfaces
use Filament\Auth\MultiFactor\App\Contracts\HasAppAuthentication;
use Filament\Auth\MultiFactor\App\Contracts\HasAppAuthenticationRecovery;

// Filament v4 MFA Traits
use Filament\Auth\MultiFactor\App\Concerns\InteractsWithAppAuthentication;
use Filament\Auth\MultiFactor\App\Concerns\InteractsWithAppAuthenticationRecovery;

use Filament\Auth\MultiFactor\Email\Contracts\HasEmailAuthentication;
use Filament\Auth\MultiFactor\Email\Concerns\InteractsWithEmailAuthentication;
use Illuminate\Contracts\Auth\MustVerifyEmail;

class User extends Authenticatable implements FilamentUser, HasEmailAuthentication, MustVerifyEmail, HasAppAuthentication, HasAppAuthenticationRecovery
{
    use HasFactory, Notifiable;
    use InteractsWithEmailAuthentication;
    use InteractsWithAppAuthentication;
    use InteractsWithAppAuthenticationRecovery;

    /**
     * Determine if the user can access the given Filament panel.
     */
    public function canAccessPanel(Panel $panel): bool
    {
        // Add your authorization logic here (e.g., email domain check)
        return true;
    }

    /**
     * Determine if the user has app-based MFA enabled.
     * This method is required by the HasAppAuthentication interface.
     */
    public function hasAppAuthentication(): bool
    {
        return (bool) $this->has_app_authentication;
    }

    protected $fillable = [
        'name',
        'email',
        'password',
        'app_authentication_secret',
        'app_authentication_recovery_codes',
        'has_app_authentication', 
        'has_email_authentication', // Ensure this is inside the array
    ]; // Remove the extra bracket from the middle

    protected $hidden = [
        'password',
        'remember_token',
        'app_authentication_secret',
        'app_authentication_recovery_codes',
    ];

    /**
     * Corrected casts for 2026 standards.
     */
    protected function casts(): array
    {
        return [
            'email_verified_at' => 'datetime',
            'password' => 'hashed',
            'has_app_authentication' => 'boolean',
            'has_email_authentication' => 'boolean', // Add this line
            'app_authentication_recovery_codes' => 'encrypted:array',
        ];
    }

    public function getAppAuthenticationSecret(): ?string
    {
        return $this->app_authentication_secret;
    }

    public function saveAppAuthenticationSecret(?string $secret): void
    {
        $this->app_authentication_secret = $secret;
        $this->save();
    }

    public function getAppAuthenticationHolderName(): string
    {
        return $this->email;
    }

    public function hasEmailAuthentication(): bool
    {
        return (bool) $this->has_email_authentication;
    }

    // Ensure this is also present to tell Filament where to send the code
    public function getEmailAuthenticationAddress(): string
    {
        return $this->email;
    }
}
`

What's wrong there?

### Expected behavior

Never works

### Steps to reproduce

After trying to setup 2FA code, Code accepting the code and not posting as well

https://github.com/pmediavictor/filament-2fa-code-issue

### Relevant log output

```shell

```

danharrin:


Baspa:
I still get this error on `5.x` is it possible that this never got merged into `5.x`? @danharrin

danharrin:
Its not possible, but some commit after might have reversed the change by accident

Mohammed-800:
I think there are two separate failure modes here:

1. The original bug from #18990 / #19035:
   The OneTimeCodeInput view was not passing `{{ $attributes }}` to the actual `<input>` element, so Livewire attributes such as `wire:model` were not applied correctly. In that case, the OTP field value is not submitted/dehydrated properly.

2. A separate setup-flow issue:
   In the authenticator app setup wizard, using replay-protected verification (`verifyKeyNewer()`) is too strict. The same valid TOTP can be validated more than once during the setup flow / wizard lifecycle, which causes a false "The code you entered is invalid" error even though the authenticator app code is correct.

Suggested fix:
- First, verify that the OneTimeCodeInput fix from PR #19035 is still present in the current 5.x branch and that the rendered input receives the Livewire attributes.
- Second, for authenticator setup only, use `verifyKey()` instead of `verifyKeyNewer()`.
- Keep `verifyKeyNewer()` for login challenge / replay-protected flows if needed.

So the issue may not be in Google Authenticator itself. It is either:
- the OTP input value not being bound/submitted correctly, or
- the setup flow using replay-safe verification in a context where the same code may be checked more than once.

danharrin:
I can't reproduce this problem on v4 or v5. https://github.com/filamentphp/filament/pull/19553 was introduced recently though which could be causing problems for you, but I can't imagine why it would cause problems to be honest... it is similar to what Fortify does.

danharrin:
Seems like it was only a problem when recovery codes were enabled, should be fixed by https://github.com/filamentphp/filament/pull/19644

PR #19040, opened 2026-01-22
fix: resolve table record without applying filter query conditions

### LINKED ISSUE #19026: Actions Fail in Modal When Table Filters Affect Action Visibility
### Package

filament/filament

### Package Version

v4.5.3

### Laravel Version

v12.47.0

### Livewire Version

v3.7.4

### PHP Version

PHP 8.2

### Problem description

When no filters are applied at the table level, opening a modal and executing any visible action works correctly.

Image

However, when a filter is applied to the table, the problem occurs only if the visibility of actions depends on whether the record is shown in the table. In this case, opening the modal and executing any visible action the first time works correctly, but once the modal re-renders, executing a newly visible action throws an error.

Image

### Expected behavior

Action execution should work correctly regardless of table filters or changes in action visibility.

### Steps to reproduce

1. Apply a filter to show only "Active" records.
2. Click "View Record" to open the modal.
3. Click the "Set Inactive" action _(this should update the record and hide the "Active" button)._
4. Click the "Set Active" action _(this should now throw an error)._

https://github.com/GeminiDev1/app-filament-issue-4.x/tree/modal-reload-issue

### Relevant log output

```shell

```

People-Sea:
### Problem: 

Actions fail when a record's state changes and no longer matches the active filter (e.g., restoring a trashed record while filtering "only trashed").

### Solution: 

resolveTableRecord() now only applies baseQuery() modifications (like removing SoftDeletingScope), not query() filter conditions.

## Functional changes

### DIFF
--- a/packages/tables/docs/03-filters/01-overview.md
+++ b/packages/tables/docs/03-filters/01-overview.md
@@ -200,6 +200,35 @@ TernaryFilter::make('trashed')
     ]))
 ```
 
+## Excluding filters when resolving records
+
+When a user interacts with a table record (e.g., clicking an action button), Filament resolves that record from the database. By default, all active filter conditions are applied, ensuring users cannot access records outside their filter scope.
+
+However, some filters like `TrashedFilter` modify global scopes rather than restricting access. When a record's state changes after the user saw it in the table, you may still want the user to interact with it.
+
+You may mark a filter to be excluded when resolving records using the `excludeWhenResolvingRecord()` method:
+
+```php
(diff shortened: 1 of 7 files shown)


danharrin:
I think excluding filters could be seen as a security issue, so I think it should be opt-in for individual filters. I am implementing that now.

danharrin (approved):

People-Sea:
@danharrin  Thanks for the improvement! The opt-in approach is much safer than my original implementation. Great solution! 👍

PR #19042, opened 2026-01-22
fix: hintActions field misalignment

### LINKED ISSUE #19038: [4.x] hintActions cause field misalignment due to unstyled wrapper div
### Package

filament/filament

### Package Version

4.5.3

### Laravel Version

12.48.1

### Livewire Version

_No response_

### PHP Version

8.4.17

### Problem description

- in Schema.php lines 198-203, Actions are wrapped in a `<div>` with no class
- this empty wrapper affects flex layout when used in `.fi-fo-field-label-ctn`
- fields with hintActions() are misaligned compared to adjacent fields

Image

### Expected behavior

Either don't render `<div>` wrapper for visible actions, or add display: contents style to `<div>` wrapper

Image

### Steps to reproduce

- add an hintAction to a field
- see misalignment of field

**Workaround**
```css
.fi-fo-field-label-ctn .fi-sc > div:not([class]) {
    display: contents;
}
```

https://github.com/zupolgec/filament-hint-action-misalignment

### Relevant log output

```shell

```

People-Sea:
## Description

Fixes #19038

Add `display: contents` to Action wrapper <div> in Schema, fixing field misalignment when using hintActions().

Thanks @zupolgec for reporting, the detailed reproduction, and suggesting the fix!

## Visual changes

# before

https://github.com/user-attachments/assets/4e83c72d-6c17-4ffd-b951-c4eb3f117b48

# after

https://github.com/user-attachments/assets/5475f624-fadf-4b99-be9f-e3112cbf262e

## Functional changes


danharrin (changes requested):
Unfortunately, we will need a solution that doesn't involve removing the `<div>`. This element is crucial for ensuring that DOM-diffing is reliable in Livewire when components are added and removed in updates. I think the solution is probably in CSS somewhere.

packages/schemas/src/Schema.php
```diff
@@ -196,11 +196,9 @@ function (Component | Action | ActionGroup $component) use (&$hasVisibleComponen
         <div <?= $attributes->toHtml() ?>>
             <?php foreach ($componentsWithVisibility as [$schemaComponent, $isSchemaComponentVisible]) { ?>
                 <?php if (($schemaComponent instanceof Action) || ($schemaComponent instanceof ActionGroup)) { ?>
-                    <div <?php if (! $isSchemaComponentVisible) { ?> class="fi-hidden"<?php } ?>>
```
zupolgec: @danharrin maybe just adding a contents class to the div? something like

```php
<div class="contents<?php if (! $isSchemaComponentVisible) { ?> fi-hidden<?php } ?>">
```

or 

```php
<?php
$wrapperAttributes = (new ComponentAttributeBag)->class([
    'contents',
    'fi-hidden' => ! $isSchemaComponentVisible,
]);
?>
<div <?= $wrapperAttributes->toHtml() ?>>
```

People-Sea:
> Unfortunately, we will need a solution that doesn't involve removing the `<div>`. This element is crucial for ensuring that DOM-diffing is reliable in Livewire when components are added and removed in updates. I think the solution is probably in CSS somewhere.

Thanks for the feedback, makes sense! I'll look for other solutions.

danharrin:
@People-Sea the Schema file controls all schema output, so now all schema actions have this new class around them I think

danharrin:
I have updated it so that even though the class applies to all schema actions, the new CSS only applies to inline schemas like the hint is. Otherwise it makes buttons etc grow to fill space in a grid

People-Sea:
Thanks for the refinement! That makes sense.

PR #19057, opened 2026-01-23
feature: Inter font preloading

### LINKED ISSUE #18936: FOUT (Flash of Unstyled Text) on page load. (And the default font is always loaded)
### Package

filament/filament

### Package Version

v4.5.2

### Laravel Version

v12.46.0

### Livewire Version

v3.7.3

### PHP Version

8.4.16

### Problem description

While the page loads, a slight change in the typography is noticeable as the browser finishes loading the font to be used.

### Expected behavior

The problem has a visual effect that makes navigating between pages less "clean". The same thing happens when using SPAs (only the first page loads).

https://github.com/user-attachments/assets/79bc151e-73d8-431f-982b-dcb9d02e5a04

**### Solution found (I don't know if it's the best)**
Preload the font, instructing the browser to prioritize loading. 
Using `<link rel="preload"`:

```<
head>
 <!-- ... -->
 <link rel="preload" href="/fonts/filament/filament/inter/inter-latin-wght-normal-NRMW37G5.woff2" as="font" type="font/woff2" crossorigin>
</head>
```

To implement this with Filament, add a hook to AppServiceProvider:

```
public function boot(): void
{
        PanelsRenderHook::STYLES_BEFORE,
        fn (): string =>
            '<link rel="preload" href="/fonts/filament/filament/inter/inter-latin-wght-normal-NRMW37G5.woff2" as="font" type="font/woff2" crossorigin />',
        );
}
```
**### Result:**

https://github.com/user-attachments/assets/8158bff1-f1e9-42f1-9f78-1fffcda7716b

### **Note:**
Another thing I noticed is that even if a different font is used and the local font provider is selected, Filament registers the "Inter" font in the FilamentAsset.
Image

_It does so in: vendor/filament/filament/src/FilamentServiceProvider.php packageBooted():

Image

### Steps to reproduce

With any new project. I'm using Windows 11 and the Chrome browser.

In AdminPanelProvider:

{
    {
        return $panel
            ->default()
            ->id('admin')
            ->path('admin')
            ->login()
            ->colors([
                'primary' => Color::Amber,
            ])
            **->topNavigation(true) 
            ->topbar(true)**
...

https://github.com/anvauy/demo-issue-fout-app.git

### Relevant log output

```shell

```

danharrin:
Closes #18936.


Youssef-Amjad:
```log
The resource http://127.0.0.1:8000/fonts/filament/filament/inter/inter-latin-wght-normal-NRMW37G5.woff2
was preloaded using link preload but not used within a few seconds from the window's load event.
Please make sure it has an appropriate `as` value and it is preloaded intentionally.
```

PR #19099, opened 2026-01-27
[4.x] Docs - update status enum

leandrocfe:
## Description

Docs - update status enum icons to use Heroicon constants

## Functional changes

### DIFF
--- a/docs/09-advanced/03-enums.md
+++ b/docs/09-advanced/03-enums.md
@@ -139,6 +139,7 @@ The `HasIcon` interface transforms an enum instance into an [icon](../styling/ic
 ```php
 use BackedEnum;
 use Filament\Support\Contracts\HasIcon;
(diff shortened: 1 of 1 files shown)


leandrocfe:
@danharrin, are you ok with replacing all examples that use strings with the Heroicon?

danharrin:
I think so, unless there is a specific reason we avoid the enum

leandrocfe:
> I think so, unless there is a specific reason we avoid the enum

Sounds good! I'll review all examples in the docs and open a new PR if it makes sense to replace the strings with Heroicons. Thanks!

danharrin:
@leandrocfe I would like to avoid using them in Blade files / for Blade component attributes, that is the only case I can think of

danharrin (approved):

PR #19114, opened 2026-01-28
Fix query builder constraints

### LINKED ISSUE #19112: v4 Filament Tables Query Builder Filters Not Applying Constraints Correctly in Standalone Livewire Components
### Package

filament/tables

### Package Version

v4.3.1

### Laravel Version

12.46.0

### Livewire Version

v3.7.6

### PHP Version

8.2

### Problem description

When I apply Query builder filter on any column

Image

Without entering anything in table column, click the edit action, the modal will appear. After closing the modal this error appears

Image

### Expected behavior

No error should appear.

### Steps to reproduce

Go to to /products route, the filament table will appear, apply query builder filter on any column without entering any data in the input.

Image

Now click on edit action action, the modal will appear, close the modal and this error will appear

Image

https://github.com/Umair-exe/Filament-tables-query-builder-bug

### Relevant log output

```shell

```

Umair-exe:
## Functional changes


People-Sea (approved):
LGTM. This defensive fix addresses the issue.

The root cause is a state/schema cache synchronization issue, but this null check is a reasonable and low-risk solution.

danharrin:
I would like to understand the cause of the error before choosing to silence it like this, I'm confused why the synchronisation issue is there based on the issue description

People-Sea:
> I would like to understand the cause of the error before choosing to silence it like this, I'm confused why the synchronisation issue is there based on the issue description

@danharrin  From my debugging, it seems related to statePath switching in applyFiltersToTableQuery() causing cachedDefaultChildSchemas to be built from the wrong statePath.

Umair-exe:
When getChildSchema($ruleIndex) is called with a non-default key (the rule index), and if there are no components registered for that key, the method returns null instead of returning an empty schema.

This happens in the Builder component when iterating through rules. The $ruleBuilder->getChildSchema($ruleIndex) call expects to get a child schema for each item/rule, but when the $ruleIndex doesn't have cached child schemas yet, it returns null.

danharrin:
I think #19128 addresses the root issue instead of silencing it

I would appreciate @Umair-exe to check if it still solves their problem and @People-Sea to review if the code is sane or not - thanks for your comment about the child schema caching, that made it easy to trace the problem

PR #19115, opened 2026-01-28
[4.x] Docs: use enum values instead of hardcoded icons

Rattone:
The resources generated by Filament use enums, so the documentation should do the same, guiding users toward best practices

### DIFF
--- a/docs/06-navigation/01-overview.md
+++ b/docs/06-navigation/01-overview.md
@@ -300,7 +300,7 @@ enum NavigationGroup implements HasIcon
     
     case Settings;
 
-    public function getIcon(): ?string
+    public function getIcon(): string | BackedEnum | Htmlable | null
     {
         return match ($this) {
             self::Shop => Heroicon::OutlinedShoppingCart,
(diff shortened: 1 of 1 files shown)


docs/06-navigation/01-overview.md
```diff
@@ -298,9 +303,9 @@ enum NavigationGroup implements HasIcon
     public function getIcon(): ?string
```
people-sea: Please update the return type to `Heroicon`.

People-Sea (changes requested):

docs/06-navigation/01-overview.md
webard: I think the 'BackedEnum' should be returned, it is forced by the HasIcon interface.

people-sea: Good point! To be consistent with the v5 docs, please use string | BackedEnum | Htmlable | null.

Rattone: I've aligned the return types with those in the HasIcon interface. Well done and thanks for the review.

danharrin (approved):

PR #19123, opened 2026-01-29
[4.x] Fix cross-domain URL detection in multi-domain apps

### LINKED ISSUE #19117: [Bug] v4.6.0 breaks cross-domain URL actions in multi-domain applications
## Bug Description

The `is_app_url()` helper function change in v4.6.0 (PR #19024) breaks cross-domain URL actions in multi-domain Laravel applications. External domain URLs are incorrectly classified as "internal", causing Filament to apply `wire:navigate` SPA navigation which fails for cross-domain requests.

## Steps to Reproduce

1. Configure a multi-domain Laravel application:
   - Filament admin panel runs on `portal.example.com`
   - Shop/frontend runs on `example.com`
   - `APP_URL` / `config('app.url')` is set to `https://example.com` (the main domain)

2. Create a table action with a URL pointing to the shop domain:
   ```php
   Action::make('view')
       ->url(fn (Product $record) => $record->shopUrl()) // Returns https://example.com/product/...
   ```

3. Click the action in Filament (running on `portal.example.com`)

**Expected:** Browser navigates to `https://example.com/product/...` (standard navigation)

**Actual:** Livewire attempts `wire:navigate` SPA navigation, which fails because it's a cross-domain request

## Root Cause

The change in `vendor/filament/support/src/helpers.php`:

```php
function is_app_url(string $url): bool
{
    if (str($url)->startsWith('/') && ! str($url)->startsWith('//')) {
        return true;
    }

    $appUrl = rtrim((string) config('app.url'), '/');

    return ($appUrl !== '' && str($url)->startsWith($appUrl))
        || rescue(fn () => str($url)->startsWith(request()->root()), false);
}
```

**Before v4.6.0:** Used `request()->root()` to determine if URL is internal. When running on `portal.example.com`, `request()->root()` returns `https://portal.example.com`, so shop URLs (`https://example.com/...`) were correctly treated as external.

**After v4.6.0:** Now **prefers `config('app.url')`** over `request()->root()`. If `APP_URL` is set to the main domain (`https://example.com`), shop URLs match and are incorrectly classified as "internal".

## Environment

- Filament Version: 4.6.0 (works correctly in 4.5.3)
- Laravel Version: 12.x
- Livewire Version: 3.x
- PHP Version: 8.4

## Suggested Fix

The `is_app_url()` function should prioritize `request()->root()` for the internal URL check, or provide a way to explicitly mark URLs as external. Multi-domain applications are a common pattern in Laravel (separate admin panel, API, shop domains).

Possible solutions:
1. Revert to using `request()->root()` as the primary check
2. Add an `->external()` method to actions to bypass `wire:navigate`
3. Check if the URL's host matches `request()->host()` rather than `config('app.url')`

People-Sea:
## Description

Fixes: #19117

Compare URL host with `request()->getHost()` instead of using `config('app.url')` prefix matching.

## Functional changes

### DIFF
--- a/packages/support/src/helpers.php
+++ b/packages/support/src/helpers.php
@@ -116,19 +116,13 @@ if (! function_exists('Filament\Support\is_slot_empty')) {
 if (! function_exists('Filament\Support\is_app_url')) {
     function is_app_url(string $url): bool
     {
-        $url = str($url);
-
-        if ($url->startsWith('/') && (! $url->startsWith('//'))) {
+        if (str($url)->startsWith('/') && ! str($url)->startsWith('//')) {
             return true;
         }
 
-        $appUrl = rtrim((string) config('app.url'), '/');
+        $urlHost = parse_url($url, PHP_URL_HOST);
 
-        if (($appUrl !== '') && $url->startsWith($appUrl)) {
-            return true;
-        }
-
-        return $url->startsWith(request()->root());
+        return (! $urlHost) || $urlHost === request()->getHost();
     }
 }
 


danharrin (changes requested):
Would you mind elaborating on the difference between `request()->root()` and `request()->getHost()`, and if the issue in #19024 will return?

Maybe @eslam-reda-div wants to have a look and see if this still fixes their problem

eslam-reda-div:
> Would you mind elaborating on the difference between `request()->root()` and `request()->getHost()`, and if the issue in #19024 will return?
> 
> Maybe @eslam-reda-div wants to have a look and see if this still fixes their problem


I plan to test this change across multiple runtime environments (Octane + FrankenPHP + Docker + Nginx) to make sure everything works properly.

From my initial review, this solution is better than the previous approach, as it handles multi-tenant setups, proxies, and subdomains more accurately and reliably.

I’ll share the test results once done. If any issues come up during testing, I’ll report them immediately or provide adjustments as needed.

The only important requirement is ensuring that trusted proxies are configured correctly so that request()->getHost() returns the correct host behind reverse proxies.

People-Sea:
@danharrin 
request()->root() returns full URL (https://example.com), request()->getHost() returns just the host (example.com). This fix compares hosts directly via parse_url() instead of prefix matching.
The scenario in #19024 is actually a configuration issue — reverse proxies need to forward the correct Host header, or TrustProxies needs to be configured. Most standard setups already handle this correctly.

danharrin (approved):

eslam-reda-div:
> @danharrin 
> request()->root() returns full URL (https://example.com), request()->getHost() returns just the host (example.com). This fix compares hosts directly via parse_url() instead of prefix matching.
> The scenario in #19024 is actually a configuration issue — reverse proxies need to forward the correct Host header, or TrustProxies needs to be configured. Most standard setups already handle this correctly.

I had already configured everything related to trusted proxies, but the issue would still fail in some runtime environments when multiple instances were behind a load balancer. The workaround I had implemented previously compared URLs to APP_URL, because in certain internal requests, Laravel Octane would send a different internal host than the main domain.

I’ll re-test this new fix to confirm that it correctly resolves the host internally within the runtime and that it won’t conflict with Octane. This should ensure that the host is consistently identified, even behind load balancers or in multi-instance setups.

PR #19126, opened 2026-01-29
update download version via composer

felipesantos2:
update the download version of spatie media in the doc.


danharrin (changes requested):
Please update the other Spatie plugin READMEs too

danharrin (approved):

PR #19129, opened 2026-01-29
Update 08-text-input.md

tomjamon:
## Description

The TextInputColumn documentation feels incomplete. For example, there is no mention that you can use `type('number')` for numeric fields (quantity, price, etc.), nor how to combine it with `inputMode('decimal')` and `step('any')` for decimal inputs. This PR adds a short section and example for `type('number')` so that users don’t have to dig into the vendor code to discover this.

## Visual changes

None (documentation only).

## Functional changes

### DIFF
--- a/packages/tables/docs/02-columns/08-text-input.md
+++ b/packages/tables/docs/02-columns/08-text-input.md
@@ -37,15 +37,29 @@ use Filament\Tables\Columns\TextInputColumn;
 TextInputColumn::make('background_color')->type('color')
 ```
 
-For numeric fields (e.g. quantity, price), use `type('number')`. You can combine it with `inputMode('decimal')` and `step('any')` for decimal values, or `step(1)` for integers:
+## Setting the HTML input mode
+
+You may use the `inputMode()` method to set the [HTML inputmode attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode), which hints at the type of keyboard to show on mobile. For numeric inputs, use `inputMode('decimal')` for decimal values or `inputMode('numeric')` for integers:
 
 ```php
+use Filament\Tables\Columns\TextInputColumn;
+
 TextInputColumn::make('quantity')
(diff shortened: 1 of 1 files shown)


danharrin (changes requested):
Please copy the docs code for these available features from the text input docs so its consistent. Update any references etc

tomjamon:
I can’t copy the TextInput docs code literally into the TextInputColumn docs because the behaviour isn’t the same.

In the Text Input docs, numeric behaviour is exposed via numeric() and integer(), which come from the CanHaveNumericState interface and the full TextInput implementation (e.g. getType() returning 'number' when isNumeric() is true, optional NumberStateCast, etc.). TextInputColumn does not implement that interface and has no numeric() or integer() methods; it only has type(), inputMode(), and step() (from the shared traits). So the APIs differ.

IMO that leaves two options:

**1. Keep the doc as in this PR**
Document the TextInputColumn way: type('number') plus inputMode('decimal') / inputMode('numeric'), step('any') / step(1), and rules(['numeric', …]) / rules(['integer', …]). Add a short note that the form TextInput provides the same behaviour via numeric() and integer(), while TextInputColumn requires these calls explicitly. No code is “copied” from the TextInput doc; the example stays specific to TextInputColumn.

**2. Align the APIs and then the docs**
Implement CanHaveNumericState (and numeric() / integer()) on TextInputColumn in the tables package so that it mirrors TextInput. Then the TextInputColumn doc can use the same patterns as the Text Input doc (e.g. ->numeric(), ->integer()), and the “copy” you asked for would be consistent.

danharrin:
I'm just asking for the docs in the same style, with headings etc. You can tweak them to match the current behaviour without changing the code.

tomjamon:
Thanks for the clarification. I’ve updated the docs to follow the same style as the TextInput form docs: added “Setting the HTML input mode” and “Setting the numeric step” sections with headings and kept the examples minimal.

danharrin (approved):

Issue #15394, opened 2025-01-21
SpatieMediaLibraryFileUpload panelLayout Grid


danielreales7:
### Package

filament/filament

### Package Version

v3.2.133

### Laravel Version

v11.38.2

### PHP Version

8.2

### Problem description

When I use ->panelLayout('grid') sometimes the styles load wrong

```php
SpatieMediaLibraryFileUpload::make('images_upload')
                                    ->collection('property_images')
                                    ->multiple()
                                    ->responsiveImages()
                                    ->image()
                                    ->imageEditor()
                                    ->label('Imágenes')
                                    ->imagePreviewHeight('250')
//                                    ->panelLayout('grid')
                                    ->columnSpanFull()
```

![Image](https://github.com/user-attachments/assets/b4175747-e925-4701-90bc-4f0c159079df)

![Image](https://github.com/user-attachments/assets/6fcfb733-96c1-448c-a2cf-0ecb4d0c102b)

![Image](https://github.com/user-attachments/assets/c5f18855-0945-45d9-833c-459c85a89d4e)

### Expected behavior

![Image](https://github.com/user-attachments/assets/06f68c58-fb79-4911-92fd-72eb4d4ec554)

### Steps to reproduce

```php
SpatieMediaLibraryFileUpload::make('images_upload')
                                    ->collection('property_images')
                                    ->multiple()
                                    ->responsiveImages()
                                    ->image()
                                    ->imageEditor()
                                    ->label('Imágenes')
                                    ->imagePreviewHeight('250')
                                    ->panelLayout('grid')
                                    ->columnSpanFull()
```

I have discovered that if the gallery appears at the moment it loads fine, but when using tabs as I show in the project is when it happens constantly

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/danielreales7/test-spatie-lmedia

### Relevant log output

```shell

```

danielreales7:
I have discovered that if I put activeTab (the gallery with the default images on the first load always appears fine)

https://github.com/user-attachments/assets/e8d9bc49-450c-41eb-92b7-0496619a5d07

However, if it is not active and I have to change between tabs, that is when the problem appears.

https://github.com/user-attachments/assets/d7d6e9af-65f7-42c4-8d02-961fd7128765

danielreales7:
I think this problem is related: [https://github.com/filamentphp/filament/pull/1997](url)

danielreales7:
Any update on this? @danharrin

[...]

PR #19139, opened 2026-01-30
[4.x] Fix `FileUpload` not working inside collapsed Repeater

### LINKED ISSUE #16253: FileUpload doesn't work as intended in Repeater of Relation Manager
### Package

filament/filament

### Package Version

v3.3.14

### Laravel Version

v12.13.0

### Livewire Version

_No response_

### PHP Version

8.2

### Problem description

I have Brand Resource with Categories Relation Manager.

In Categories Relation Manager, I have repeater for Menu Items that has File Upload for image.

After submitting the form:
It doesn't let me click on the 'x' button to remove the image. 
Also when changing the image, it doesn't work. Need to upload the image twice.

### Expected behavior

On clicking 'x' it should remove the image as it does normally.
Uploading image on edit should work as intended.

### Steps to reproduce

Use the repo I created.
In that first create a brand, then create category with menu items.
Now upload image for 1 menu item and save.
Now open the same category with menu item and try removing the image. Doesn't work.
Now try to change the image of that menu item, there will be an issue. Need to upload the same image twice.

https://github.com/faizanakbar10/filament-fileupload-repeater-issue

### Relevant log output

```shell

```

People-Sea:
### Fixes: #16253

FilePond calculates layout dimensions on init. When inside a collapsed Repeater (`display: none`), all dimensions are 0, causing mispositioned elements and broken interactions.

Fix: Defer initialization until the element becomes visible.

### Before

https://github.com/user-attachments/assets/93e17f0c-5649-4917-bf4d-b463871d216f

### After

https://github.com/user-attachments/assets/55b4c298-1267-4621-b677-98d927546ea9

## Functional changes


packages/forms/resources/js/components/file-upload.js
```diff
@@ -104,7 +104,25 @@ export default function fileUploadFormComponent({
...
         editor: {},
 
+        visibilityObserver: null,
+
         async init() {
+            if (this.pond) {
+                return
+            }
+
+            if (this.$el.offsetParent === null) {
```
danharrin, line 114: The only thing I can think of is maybe add `this.visibilityObserver?.disconnect()` again above here, so that if the component is initialised by Alpine at some point again when the observer is already registered, we don't have two instances active at once?

Apart from that, this looks like a great solution to the problem, thanks!

danharrin (changes requested):

danharrin:
Maybe it's worth checking if this fixes issue #15394 too, since its a bit of a similar problem?

People-Sea:
> Maybe it's worth checking if this fixes issue #15394 too, since its a bit of a similar problem?

@danharrin Thanks for the heads up! Will check if this resolves #15394 as well. ❤️

PR #19147, opened 2026-01-30
Top align section header icon

larsbo:
## Description

Top align section header icon.

## Visual changes

If a sections contains both a heading and a description the icon was vertically centered instead of staying at the top.

The example image in the documentation also represents the corrected behaviour: https://filamentphp.com/docs/4.x/schemas/sections#adding-an-icon-to-the-sections-header

icons

## Functional changes


danharrin (changes requested):
You'll need a solution that doesn't re-introduce #16935

danharrin:
@larsbo I did review it, please see my comment above

larsbo:
@danharrin In my opinion the center alignment introduced with #16935 is less helpful than the top alignment. The only case that benefits from it is the one from the screenshot: 2 text rows + action button.
All other possibilites like
- single line
- multiple lines
- single line + icon
- multiple lines + icon
- single line + action
- multiple lines + action
- ...
are working fine with a top alignment, while the centered one fails easily with longer content:
Bildschirmfoto 2026-01-30 um 21 19 35


And the top alignment variant seems to be more consistent compared with other components like modals and notifications with a top aligned close icon.

What do you think?

danharrin:
I agree that the icon should be top aligned, but I am not going to merge a PR that just introduces a separate issue. I think that the actions should be center aligned.

larsbo:
@danharrin I updated the PR. The actions part is now vertically centered again.

danharrin (approved):

PR #19150, opened 2026-01-30
Add assertions for extraImgAttributes to TestsColumn

Arne1303:
## Description

Right now it is possible to check for extraAttributes on columns using
- `assertTableColumnHasExtraAttributes()`
- `assertTableColumnDoesNotHaveExtraAttributes()`

However, the same functionality does not exist when using extraImgAttributes on the ImageColumn. I added
- `assertTableColumnHasExtraImgAttributes()`
- `assertTableColumnDoesNotHaveExtraImgAttributes()`

to add that functionality.

## Functional changes

### DIFF
--- a/docs/10-testing/03-testing-tables.md
+++ b/docs/10-testing/03-testing-tables.md
@@ -242,6 +242,20 @@ it('displays author in red', function () {
 });
 ```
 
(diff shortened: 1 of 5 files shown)


danharrin:
I'm not accepting new testing helpers like this, because then eventually we will end up with a testing helper for every single possible configuration method which is not practical.

Instead, I suggest using `assertTableColumnExists()` with a function to check if the column exists with the attributes that you specify

PR #19162, opened 2026-02-01
Feat: labels can grow when inline

CharlieEtienne:
## Description

This is my first attempt to address this problem:

Field wrapper, when having `inlineLabel()`, becomes a 3 columns grid, with input spanning on two columns.
Although it allows fields to be nicely aligned with each other, sometimes you can feel that a lot of space is lost, especially with checkbox/toggle and long descriptions:

Capture d’écran 2026-02-02 à 00 23 30

The goal is to have an option to have that instead:

Capture d’écran 2026-02-02 à 00 23 52

To avoid introducing a new trait and a new method, this proposal uses an optional `labelCanGrow` parameter in `inlineLabel` method:

```php
Toggle::make('Toggle')->inlineLabel(labelCanGrow: true)
```

## Visual changes

While it seems to work fine with most of the fields, obviously some fields like RichEditor would be unusable without an additional min-width (any help or advice would be appreciated!):

filament4playground test_admin_users_1_edit

## Functional changes


CharlieEtienne:
I set the PR back to draft because I feel it can be improved, but I'm not sure how. let me know what you think

danharrin:
In general, I'm skeptical. I see the appeal for some components like toggle. The fact that each component will need a min width to work like this feels dirty, and I question how it would work on mobile especially if the screen size doesn't allow the full min width to be used. And adding it just for toggle also feels a bit wrong as it's not consistent.

CharlieEtienne:
Just to clarify: here it won't affect style below sm breakpoint.

Ideas:

1. Add it for some fields only but we lose consistency
2. Add a very small min-width like 270px: RichEditor would not be usable below anyway
3. Leave it to user land and abandon the feature

danharrin:
I don't think this is suitable at the moment, sorry.

PR #19163, opened 2026-02-02
Fix: keep parent modal open when child action modal opens

leek:
## Summary

- When a nested action (e.g. repeater delete confirmation) opens inside a slide-over, the parent modal closes first — making it feel like the user lost their work
- This skips the parent close when action nesting depth is **increasing**, so the parent stays visible and the child renders on top
- CSS raises stacked child modals to `z-50` (above parent's `z-40`) with a lighter backdrop to avoid double-darkness

image

## Changes

**`packages/actions/resources/js/components/modals.js`** — 5 lines added

When `syncActionModals` detects nesting is increasing (parent → child), it no longer calls `closeModal()` on the parent. On nesting decrease (child dismissed), the parent is already open and the existing `(! isOpen)` guard in `open-modal` prevents a redundant reopen.

**`packages/support/resources/css/components/modal.css`** — 11 lines added

Sibling `.fi-modal-open` elements get `z-50` on their overlay and window container, plus a lighter backdrop (`bg-gray-950/25` / `dark:bg-gray-950/50`).

## Test plan

- [ ] Open any Create/Edit slide-over with a repeater
- [ ] Click delete on a repeater item → confirmation should stack ON TOP of the slide-over
- [ ] Cancel confirmation → slide-over stays with all form data intact
- [ ] Confirm delete → action executes, slide-over stays, item removed
- [ ] Close slide-over entirely → everything cleans up
- [ ] Test escape key and click-away on the confirmation modal
- [ ] Test with non-slide-over modals (standard centered modals with nested actions)
- [ ] Test deeper nesting (3+ levels) if applicable

### DIFF
--- a/packages/actions/resources/js/components/modals.js
+++ b/packages/actions/resources/js/components/modals.js
@@ -20,7 +20,12 @@ export default ({ livewireId }) => ({
             return
         }
 
-        if (this.actionNestingIndex !== null) {
+        const isNestingIncrease =
+            this.actionNestingIndex !== null &&
+            newActionNestingIndex !== null &&
+            newActionNestingIndex > this.actionNestingIndex
+
+        if (this.actionNestingIndex !== null && !isNestingIncrease) {
             this.closeModal()
         }
 
--- a/packages/support/resources/css/components/modal.css
+++ b/packages/support/resources/css/components/modal.css
@@ -100,6 +100,17 @@
         @apply fixed inset-0 z-40 bg-gray-950/50 dark:bg-gray-950/75;
     }
 
+    &.fi-modal-open ~ .fi-modal.fi-modal-open {
+        & > .fi-modal-close-overlay,
+        & > .fi-modal-window-ctn {
+            @apply z-50;
+        }
+
+        & > .fi-modal-close-overlay {
+            @apply bg-gray-950/25 dark:bg-gray-950/50;
+        }
+    }
+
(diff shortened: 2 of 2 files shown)


People-Sea:
I think this PR only fixes the surface problem and will introduce more breaking issues.
If you're using AI to help write code, you still need to actually test it before submitting.

danharrin:
Sorry, I don't think I can merge something like this generally, I think it is too much of a breaking change and a personal preference about how the default behaviour should work.

I would suggest proposing features in Discussions and @ ing me before making a PR to ensure you aren't spending time on something we wouldn't merge.

leek:
@People-Sea Rude.

@danharrin I'm not sure how much of a "personal preference" this is. Can you really tell me this is a good user experience?

**before:**
https://github.com/user-attachments/assets/05f5f262-a4e8-49c9-86b0-e9d3f2ca0d5a

**after:**
https://github.com/user-attachments/assets/5c0f50a2-9bdf-43a0-baa0-c532a5440121

danharrin:
For your specific case, it is definitely a better user experience, because the parent slideover is large and the child is a non-slideover confirmation modal. But this is not a representation of all other nested modal situations, and is a breaking change in behaviour which we cannot release.

I don't think there is a very straightforward way for this behaviour to be opt-in per nested modal, otherwise I would have suggested that.

PR #19183, opened 2026-02-03
fix: ensure filter state is dehydrated before persistence

### LINKED ISSUE #19170: dehydrateStateUsing is never called on form fields inside table filter schemas
### Package

filament/filament

### Package Version

v5.1.3

### Laravel Version

12.49.0

### Livewire Version

_No response_

### PHP Version

8.4.16

### Problem description

When using `formatStateUsing` and `dehydrateStateUsing` on a form field inside a table
  `Filter::make()->schema([...])`, the `dehydrateStateUsing` callback is never invoked. The raw user input is
  passed directly to the query() callback and persisted in the URL/session as-is.

  Meanwhile, `formatStateUsing` IS applied on every page hydration. Since the persisted filter state was
  never dehydrated, `formatStateUsing` keeps transforming an already-formatted value, causing it to compound
  on every page reload.

### Expected behavior

`dehydrateStateUsing` should be called before the value is passed to the query() callback and before the
  filter state is persisted in the URL/session

### Steps to reproduce

```php
Filter::make('foo')
      ->schema([
          TextInput::make('percentage')
              ->numeric()
              ->formatStateUsing(fn ($state) => $state * 100)
              ->dehydrateStateUsing(fn ($state) => $state / 100)
      ])
          if (filled($data['percentage'])) {
              $query->where('percentage', '>=', $data['percentage']);
          }
          return $query;
      })
```

  1. Enter 10 in the percentage filter field
  2. Observe the URL: ?filters[foo][percentage]=10 — expected 0.1
  3. The query() callback receives $data['percentage'] = 10 — expected 0.1
  4. Reload the page (Ctrl+R / F5)
  5. The input now shows 1000 (formatStateUsing applied: 10 * 100)
  6. Reload again → 100000, and so on — the value grows by 100x on every reload

https://github.com/marvinrabe/filament-demo/commit/25afcd459083211e805e31cc25c748bcbfdc6dd2

### Relevant log output

```shell

```

marvinrabe:
## Description

Fixes #19170.

`applyTableFilters()` now uses `getStateSnapshot()` to ensure dehydration runs before persisting state to `tableFilters`. The filter logic has been simplified to always use `tableDeferredFilters` as the form state path. This ensures `tableFilters` are always properly dehydrated while the raw (hydrated) form state remains in `tableDeferredFilters`. For non-deferred filter forms, live updates are handled via the `updatedTableDeferredFilters()` hook.

Added `dehydratedWhenHidden()` to the QueryBuilder constraint settings Group. Hidden components are excluded during dehydration by default, but operators without form fields (e.g., boolean `is true`) need their state preserved even when the settings Group is hidden.

Suggestion: The `tableDeferredFilters` state path name could be improved to `tableHydratedFilters`/`tableRawFilters`/... (or similar) to better reflect its purpose (holding raw / hydrated state before dehydration). However, this would be a breaking change for any code referencing this property directly.

## Visual changes

None.

## Functional changes


danharrin:
Sorry, upon reviewing this there seems to be multiple breaking changes that we cannot merge. I think it's probably better that I handle this PR myself instead.

PR #19187, opened 2026-02-03
feat: ability for `RichEditor` plugins to provide file attachment provider

ralphjsmit:
When working on a RichEditor plugin I noticed that it was not possible for a plugin to provide a file attachment provider. 

Right now, if a plugin needs a corresponding file attachment provider, the end user has to pass it manually:
```php
$this
    ->registerRichContent('content')
    ->plugins([$plugin])
    ->fileAttachmentProvider($plugin->getFileAttachmentProvider());
```

This PR adds a HasFileAttachmentProvider interface that plugins can implement alongside RichContentPlugin. When a plugin implements this interface, the RichContentAttribute and RichContentRenderer automatically resolve the file attachment provider from the plugin, so the end user only needs to register the plugin:

```php
$this
    ->registerRichContent('content')
    ->plugins([$plugin]);
```

The only "tricky" thing is that a rich editor can only have one file attachment provider instead of multiple. In this case it could make sense that the end user would need to manually select which file attachment provider to use. However, I think that if a plugin provides a file attachment provider for including media, then generally the user does not require any other file attachment providers.

Thanks!


danharrin (changes requested):
It doesn't seem like this has changed the rich editor to read from the plugin's file attachment provider, just changed the attribute and renderer. So the editor would only use the provider if a plugin is added to the attribute.

ralphjsmit:
> It doesn't seem like this has changed the rich editor to read from the plugin's file attachment provider, just changed the attribute and renderer. So the editor would only use the provider if a plugin is added to the attribute.

I see my example was out-of-date, but the reason for not doing this is (and doing it via the attribute/renderer instead) was that the file attachment provider requires an attribute to be set, like `attribute(...)`. So from that perspective it probably would not make sense to allow users to inject a custom fileAttachmentProvider() on the `RichEditor` itself if they are not using an attribute.

danharrin:
Please proceed with making it consistent, even if there currently aren't use cases. It will make work to expand the file attachment providers easier in the future

ralphjsmit:
@danharrin Updated the `RichEditor` class to also resolve the `fileAttachmentProvider` from the plugins in the same way so that it is consistent.

danharrin (changes requested):
Please fix the test failures

danharrin (approved):

PR #19230, opened 2026-02-08
WIP: CSP Support

4513:
## Description

As discussed in https://github.com/filamentphp/filament/discussions/7032, this PR adds a support for CSP.

The PR is in WIP state as more commits will be made.

## Visual changes

There are no visual changes expected to happen - an end user must not notice any change.

## Functional changes

### DIFF
--- a/packages/actions/resources/views/components/modals.blade.php
+++ b/packages/actions/resources/views/components/modals.blade.php
@@ -1,5 +1,5 @@
 @if ($this instanceof \Filament\Actions\Contracts\HasActions && (! $this->hasActionsModalRendered))
-    <style nonce="{{ \Filament\csp_nonce() }}">
+    <style nonce="{{ \Filament\csp_nonce() }}" scoped>
         .fi-internal-components-modals {height: 0;}
     </style>
     <div
--- a/packages/actions/resources/views/components/modals.blade.php
+++ b/packages/actions/resources/views/components/modals.blade.php
@@ -1,13 +1,10 @@
 @if ($this instanceof \Filament\Actions\Contracts\HasActions && (! $this->hasActionsModalRendered))
-    <style nonce="{{ \Filament\csp_nonce() }}" scoped>
-        .fi-internal-components-modals {height: 0;}
-    </style>
     <div
         wire:partial="action-modals"
         x-data="filamentActionModals({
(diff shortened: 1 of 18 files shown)


4513:
### Initialize commit

This commit creates a new API for developer to:
* configure CSP nonce for the application using Filament facade method 'useCspNonce'
  * it is expected to define the nonce using a closure within a ServiceProvider. An example using Spatie's library would be 'Filament::useCspNonce(static fn (): string => csp_nonce());`
  * the user is expected to configure Livewire too, following their documentation
* to access and print CSP nonce into Blade templates
  * example: `<script nonce="{{ \Filament\csp_nonce() }}"></script>`

### Nonces to current elements commit

To all style and script elements defined in Filament Blade templates a nonce has been added.

### Simple inlines to elements commit

Simple inline style attributes has been moved to their own style elements with nonce. Most classes are either internal (predefined) or unique (randomly generated, not expected to be used more than once). 

This commit does not edit `x-bind:style`

danharrin (changes requested):
As I said here https://github.com/filamentphp/filament/discussions/7032#discussioncomment-15693867, I don't think we should focus on CSP style support: it looks like most of the changes here are related to that. I do not think it's supported in Livewire / Alpine / TipTap. Should we not be focusing on inline scripts?

4513:
While making the library CSP compliant, I am trying to solve all issues that both are now relevant now and may be relevant in the future. I can undo the commit that focuses on inline styles tho.

The next steps will be focused on inline scripts, x-data attribute and pushing current CSP nonce via 'livewire/update' requests.

danharrin:
Please undo all style-related changes. While I appreciate the endeavour to solve everything, this will already be a large PR that is difficult to safely review, and including styles is going to increase the surface area when they aren't even able to be used.

danharrin:
Hi @4513, any update on this? Just want to make sure this doesn't go stale, or I would rather the PR was opened once it was ready for review. No expectations to finish if you don't want to, but if not I would rather close it. Thanks.

m-develops:
@4513 Need any help? Perhaps if we could divide the work we can make this work together?
Feel free to send me a PM.

danharrin:

PR #19237, opened 2026-02-09
feat: add modal and slideover support for column manager

marvinrabe:
## Description

This PR adds support for modal and slideover layouts for column managers, making it much easier to work with tables that have a shit ton of columns (see #18938).

The implementation follows the same pattern used for filter form layouts.

## Visual changes

Nothing noteworthy. Dropdown looks the same, modal visuals similar to filter modals.

## Functional changes

My Github actions failed mostly because of current Github outages ("This endpoint is temporarily being throttled. Please try again later."). Local runs were fine.


danharrin (changes requested):
This doesn't look like it was manually tested locally by a human, the heading and apply button for example is duplicated within the modal content

danharrin (approved):

packages/tables/resources/css/column-manager.css
```diff
@@ -1,9 +1,5 @@
 .fi-ta-col-manager {
-    @apply grid gap-y-4;
-
-    & .fi-ta-col-manager-ctn {
```
archilex, line 4: @marvinrabe Technically this change is correct and the container div isn't needed, but removing this class could be a breaking change for any dev that may be targeting it for custom themes/designs. @danharrin is that ok?

packages/tables/resources/views/components/column-manager/index.blade.php
```diff
@@ -0,0 +1,74 @@
...
+    'reorderAnimationDuration' => 300,
+    'resetActionPosition' => ColumnManagerResetActionPosition::Header,
+])
+
+<div
+    x-data="filamentTableColumnManager({
+                columns: $wire.entangle('tableColumns'),
+                isLive: {{ $applyAction->isVisible() ? 'false' : 'true' }},
+            })"
+    class="fi-ta-col-manager"
```
archilex, line 21: @marvinrabe Depending on what Dan says you may need to revert this to the previous class and re-add the container div with `fi-ta-col-manager`.

packages/tables/resources/views/index.blade.php
```diff
@@ -577,36 +579,100 @@ class="fi-ta-filters-dropdown"
...
+                                            {{ $columnManagerTriggerAction->getModalContent() }}
+
+                                            <div
+                                                x-data="filamentTableColumnManager({
+                                                            columns: $wire.entangle('tableColumns'),
+                                                            isLive: {{ $columnManagerApplyAction->isVisible() ? 'false' : 'true' }},
+                                                        })"
+                                                x-on:apply-table-column-manager.window="applyTableColumnManager()"
+                                                x-on:reset-table-column-manager.window="resetDeferredColumns()"
+                                                class="fi-ta-col-manager"
```
archilex, line 639: Depending on what Dan says you may need to change this to `fi-ta-col-manager-ctn` and then wrap it in the `fi-ta-col-manager` div

archilex (commented):
@marvinrabe Thanks for adding this feature. @danharrin Tested and reviewed and it looks good. Just need your feedback on the container div. If you're ok with its removal then this is ready to go.

danharrin:
Thanks @archilex, I did remove that element deliberately when I was fixing the feature as it was effectively duplicating the job of its parent and had the same gap styles.

archilex (approved):

marvinrabe:
@danharrin Thanks for the fix! I'm using a custom implementation in my project and forgot about testing this with the default `x-filament-tables::column-manager`.

PR #19238, opened 2026-02-10
fix: action modal partial rerender during poll destroying nested component state

### LINKED ISSUE #18681: Polling relationmanager table empty search input in tableSelect()
### Package

filament/filament

### Package Version

v4.3.1

### Laravel Version

v12

### Livewire Version

v3

### PHP Version

PHP v8.4

### Problem description

The search input in tableSelect is emptied when the table underneath has polling

### Expected behavior

polling a table should only affect this table

### Steps to reproduce

1. Create two resources with many to many relationships
2. Create a relationManager in one of those, with this table:
```php
return $table
            ->poll()
            ->headerActions([
                AttachAction::make()->tableSelect(AnyTable::class),
            ])
```
4. Click on "Attach" button in the relation manager
5. Try to search in table select

Detailed instructions in repo readme

https://github.com/CharlieEtienne/poll-empty-search

### Relevant log output

```shell

```

People-Sea:
## Description
Fixes #18681

When an action modal is open, pure poll requests still trigger shouldRenderMountedActionOnly() in dehydrate(), re-rendering the modal partial on every poll. This re-mounts nested Livewire components inside the modal, losing their state (e.g. search input gets cleared).

shouldSkipMountedActionRerender() skips partial rendering in dehydrate() when there are no calls or updates but an action is already mounted.

### Before

https://github.com/user-attachments/assets/59cf8577-7e84-421b-84c3-05591128b252

### After

https://github.com/user-attachments/assets/3c8d3591-97d9-48f0-94da-b69807f0e539

## Functional changes

### DIFF
--- a/packages/support/src/Livewire/Partials/PartialsComponentHook.php
+++ b/packages/support/src/Livewire/Partials/PartialsComponentHook.php
@@ -92,6 +92,24 @@ class PartialsComponentHook extends ComponentHook
         return $originallyMountedActionIndex === $mountedActionIndex;
     }
 
+    public function shouldSkipMountedActionRerender(): bool
+    {
+        if (! property_exists($this->component, 'mountedActions')) {
+            return false;
+        }
+
+        $mountedActionIndex = array_key_last($this->component->mountedActions);
+
+        if (blank($mountedActionIndex)) {
+            return false;
+        }
+
+        $updatesCount = $this->storeGet('updatesCount') ?? 0;
+        $callsCount = $this->storeGet('callsCount') ?? 0;
+
+        return ($updatesCount + $callsCount) === 0;
+    }
+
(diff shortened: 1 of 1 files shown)


danharrin (changes requested):
I think the actual bug here is that partial renders are not tracking nested components (within the rendered partial) correctly. While your fix does work as it prevents the partial render, it causes the relation manager table to also reload which it shouldn't need to.

I have tried looking into this problem before. I struggled to identify where the list of "nested" component IDs was being sent to Livewire, and how Livewire was determining whether or not to re-render those nested components on an update.

If we can work out how to preserve nested components within partials instead of replacing them with a new component every time, this should be fixed.

People-Sea:
@danharrin  Thanks Dan, you're right   skipping the render was too shallow. I'll dig into properly tracking children created during partial renders.

danharrin:
#19242

PR #19260, opened 2026-02-12
feat: ability for `RichEditor` plugins to provide enabled/disabled toolbar buttons

ralphjsmit:
This PR provides the ability for `RichEditor` plugins to provide enabled/disabled toolbar buttons. For example, for a media-related rich editor plugin, it would be convenient that when a user adds the plugin to a `RichEditor` instance, the plugin would automatically disable the default `attachFiles` toolbar button and enable the custom toolbar button of itself instead.

```php
use Filament\Forms\Components\RichEditor\Plugins\Contracts\HasToolbarButtons;
use Filament\Forms\Components\RichEditor\Plugins\Contracts\RichContentPlugin;

class HighlightRichContentPlugin implements RichContentPlugin, HasToolbarButtons
{
    // ... other methods ...

    /**
     * @return array<string | array<string | array<string>>>
     */
    public function getEnabledToolbarButtons(): array
    {
        return ['highlight', 'highlightWithCustomColor'];
    }

    /**
     * @return array<string>
     */
    public function getDisabledToolbarButtons(): array
    {
        return [];
    }
}
```

Thanks!


danharrin (changes requested):
Please rebase the commits on 4.x and retarget this PR

ralphjsmit:
@danharrin Re-targeted to 4.x.

danharrin (approved):

PR #19278, opened 2026-02-16
feat: add dropdown toolbar buttons to Rich Editor

leek:
image

image

## Summary

- Add dropdown toolbar buttons to group related tools (alignment, headings, etc.) into compact menus via associative array syntax in `toolbarButtons()` or by registering `RichEditorDropdownTool` instances in `tools()`
- `RichEditorDropdownTool` supports icon mode (inline row of icons) and select mode (vertical list with labels)
- Dropdown configs resolve from associative entries with proper cache invalidation and disable support — disabling options collapses a single-option dropdown back to a plain button; disabling the trigger removes the entire dropdown
- Add `paragraph` tool to core Rich Editor tools
- Add ARIA roles for accessibility (`menu`, `menuitem`, `aria-haspopup`)
- Add documentation for dropdown toolbar buttons and the `paragraph` tool

## Test plan

- [x] 12 new Pest tests covering dropdown creation, option disabling, collapse-to-plain-button, trigger removal, `hasToolbarButton()` for dropdown options, cache invalidation, and group interop
- [x] All 34 Rich Editor tests pass (`vendor/bin/pest tests/src/Forms/Components/RichEditorTest.php`)
- [x] PHPStan passes with no errors
- [x] Code style fixes applied (`composer cs`)


danharrin (changes requested):
Please refactor the implementation to use an object, such as `ToolbarButtonGroup`, to configure the groups instead of nested arrays. You could introduce a `textualButtons()` config method on that object instead of introducing `selectModeDropdowns()` for the entire editor.

danharrin (approved):

PR #19281, opened 2026-02-16
feat: show loading spinner on table column sort icon

leek:
image

## Summary

- Replaces the sort direction chevron icon with a spinning loading indicator while the `sortTable` Livewire request is in flight
- Scoped per-column via `wire:target="sortTable('{columnName}')"` so only the clicked column shows the spinner
- Follows the existing Filament loading indicator pattern used in buttons, badges, links, and icon buttons (`wire:loading` / `wire:loading.remove` with configurable delay)

## Test plan

- [x] Click a sortable column header — the chevron icon should be replaced by a spinner until the response completes
- [x] Verify only the clicked column shows the spinner, not all sortable columns
- [x] Verify the correct sort direction icon (up/down) returns after loading
- [x] Test in both light and dark modes
- [ ] Test with different column alignments (start, center, end)


Youssef-Amjad:
Nice ✨

danharrin (approved):

PR #19289, opened 2026-02-18
fix: prevent `array_key_exists()` null key deprecation on PHP 8.5

### LINKED ISSUE #19284: PHP 8.5: array_key_exists() with null key deprecation in HasTabs::modifyQueryWithActiveTab()
### Package

filament/filament

### Package Version

v5.2.1 

### Laravel Version

v12

### Livewire Version

v4

### PHP Version

PHP 8.5.2

### Problem description

In src/Resources/Concerns/HasTabs.php on line 75, array_key_exists($this->activeTab, $tabs) is called while $this->activeTab is null. PHP 8.5 deprecates passing null as the key parameter to array_key_exists():

`Warning: Using null as the key parameter for array_key_exists() is deprecated, use an empty string instead`

 This warning fires repeatedly on every table page load, causing noticeable performance degradation due to the volume of warnings generated.

### Expected behavior

No deprecation warning should be triggered. The null value of $this->activeTab should be caught before reaching array_key_exists().

### Steps to reproduce

 1. Use PHP 8.5
 2. Create any resource with a ListRecords page that has a table
 3. Do not define any tabs (use the default empty getTabs())
 4. Load the list page
 5. Observe repeated deprecation warnings in the log:
  
`Warning: Using null as the key parameter for array_key_exists() is deprecated in vendor/filament/filament/src/Resources/Concerns/HasTabs.php on line 75`

The issue is in modifyQueryWithActiveTab(). The guard on line 69 (blank(filled($this->activeTab))) should prevent null from reaching line 75, but it doesn't in all cases. A simple fix:

// Line 75, before:
`if (! array_key_exists($this->activeTab, $tabs)) {`

// After:
`if ($this->activeTab === null || ! array_key_exists($this->activeTab, $tabs)) {`

https://github.com/gldrenthe89/filamentv5-php-deprication-warning

### Relevant log output

```shell
warning – Using null as the key parameter for array_key_exists() is deprecated, use an empty string instead in /vendor/filament/filament/src/Resources/Concerns/HasTabs.php on line 75
```

People-Sea:
## Description

Fixes #19284 

guard array_key_exists() calls against null keys to avoid the PHP 8.5 deprecation warning, and fix a latent blank(filled()) typo in HasTabs.

## Functional changes

### DIFF
--- a/packages/panels/src/Auth/Pages/EditProfile.php
+++ b/packages/panels/src/Auth/Pages/EditProfile.php
@@ -263,8 +263,7 @@ class EditProfile extends Page
 
         if (
             (! is_array($recipient))
-            || ($currentEmail === null)
-            || (! array_key_exists($currentEmail, $recipient))
+            || (! array_key_exists($currentEmail ?? '', $recipient))
         ) {
             return $newEmail;
         }
--- a/packages/panels/src/Auth/Pages/EditProfile.php
+++ b/packages/panels/src/Auth/Pages/EditProfile.php
@@ -263,6 +263,7 @@ class EditProfile extends Page
 
         if (
             (! is_array($recipient))
+            || ($currentEmail === null)
             || (! array_key_exists($currentEmail, $recipient))
         ) {
             return $newEmail;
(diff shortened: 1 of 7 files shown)


danharrin (changes requested):
I'm wondering if the more consistent fix would just be to add `?? ''` after passing the key to `array_key_exists()`, so it always behaves like it used to do? Because an empty string can be an array key and we may have been relying on that before. What do you think?

danharrin (approved):

People-Sea:
@danharrin Good call, keeping the implicit coercion behavior is cleaner. Thanks for the fix!

PR #19292, opened 2026-02-18
fix: add null check

martin-ro:
## Description
User A deletes a record via relation manager. User B tries to delete the same record via relation manger:

```
TypeError
filament/packages/actions/src/DeleteAction.php:43
Filament\Actions\DeleteAction::{closure:Filament\Actions\DeleteAction::setUp():43}(): Argument #1 ($record) must be of type Illuminate\Database\Eloquent\Model, null given, called in /Users/martin/Projects/filament-v4/filament/packages/support/src/Concerns/EvaluatesClosures.php on line 36
```

[https://pastebin.com/jpz8S5RP](https://pastebin.com/jpz8S5RP)

## Visual changes

n/a

## Functional changes

## Reproduction repo
[https://github.com/martin-ro/filament-null-check](https://github.com/martin-ro/filament-null-check)

### DIFF
--- a/packages/actions/src/Concerns/InteractsWithActions.php
+++ b/packages/actions/src/Concerns/InteractsWithActions.php
@@ -598,6 +598,10 @@ trait InteractsWithActions
         if (filled($action['context']['recordKey'] ?? null)) {
             $record = $this->getTableRecord($action['context']['recordKey']);
 
+            if (! $record) {
+                throw new ActionNotResolvableException("Record [{$action['context']['recordKey']}] no longer exists.");
+            }
+
             $resolvedAction->getRootGroup()?->record($record) ?? $resolvedAction->record($record);
         }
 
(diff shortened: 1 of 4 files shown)


danharrin (approved):

fero8:
This introduces an issue in the case when a record is changed and is no longer returned by the table query, for example:
1. a modal window is opened from a record row action in a table
2. the modal changes some field of the record which causes it to no longer be returned by `getTableRecord()` which now returns null, then the exception is thrown and previously working flows are broken

Workaround is to override the `cacheMountedActions()` method and catch the exception in `ListRecords`

danharrin:
@fero8 please create a reproduction repository and post it here so I can test what is happening and what the behaviour was before, because from what I can see, the behaviour before would have also resulted in an exception, just a different one

andrewdwallo:
All this fix did for me was throw a different exception in the UI. For example, now it shows: 
```
Filament\Actions\Exceptions\ActionNotResolvableException
vendor/filament/actions/src/Concerns/InteractsWithActions.php:602

Record [696829] no longer exists. 
```

Update: to be clear the exception above is thrown when opening the delete modal in 2 browser tabs, deleting the record in one tab, switching back and clicking delete again in the other tab.

fero8:
As a matter of fact, it is not only about deleting. This exception "record no longer exist" will be thrown even if you return from any modal for a record that changed in the meantime OR changed as a result of an action done by the modal that results in the underlying record being removed from the current table filtered query.

I think this PR created a lot more issues than solving a somewhat edge case.

andrewdwallo:
@fero8 I created a PR #19401

PR #19296, opened 2026-02-18
feat: get item index in itemLabel on Repeater

pintend:
passes the index to `getItemLabel()` so we can do something like this

``` php
Repeater::make()
    ->itemLabel(function (int $index) {
        return match($index) {
            1 => 'Primary',
            default => 'Backup',
        };
    })
```

image

### DIFF
--- a/packages/forms/resources/views/components/repeater/index.blade.php
+++ b/packages/forms/resources/views/components/repeater/index.blade.php
@@ -91,7 +91,7 @@
             >
                 @foreach ($items as $itemKey => $item)
                     @php
-                        $itemLabel = $getItemLabel($itemKey);
+                        $itemLabel = $getItemLabel($itemKey, $loop->iteration);
                         $visibleExtraItemActions = array_filter(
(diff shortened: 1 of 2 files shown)


People-Sea:
Please rebase onto 4.x

packages/forms/resources/views/components/repeater/index.blade.php
```diff
@@ -91,7 +91,7 @@
             >
                 @foreach ($items as $itemKey => $item)
                     @php
-                        $itemLabel = $getItemLabel($itemKey);
+                        $itemLabel = $getItemLabel($itemKey, $loop->iteration);
```
people-sea, line 94: Should this be $loop->index instead? Since the injected parameter is named $index, it should start at 0 to match convention.

danharrin:
Please reopen on 4.x

pintend:
Replaced by https://github.com/filamentphp/filament/pull/19430

PR #19297, opened 2026-02-18
 fix: prevent contained styles from leaking into nested RepeatableEntry items

### LINKED ISSUE #19285: Contained function doesnt work in nested RepeatableEntry
### Package

filament/filament

### Package Version

v4.7.1

### Laravel Version

v12.51.0

### Livewire Version

v3.7.10

### PHP Version

PHP 8.4.18

### Problem description

When using a RepeatableEntry inside another RepeatableEntry, setting the nested RepeatableEntry to contained(false) does not remove its container styling as expected.

The nested component continues to render with the default container UI (card/box styling), even though contained(false) is explicitly set.

The container styling is only removed if the parent RepeatableEntry is also set to contained(false). This suggests that the nested component’s contained(false) configuration is being ignored unless the parent is also non-contained.   

However, the inverse configuration works correctly:
If the parent RepeatableEntry is set to contained(false) and the child RepeatableEntry is left as contained (default or contained(true)), the child correctly renders with container styling.

### Expected behavior

**Expected behavior**

When using a RepeatableEntry inside another RepeatableEntry, setting the nested RepeatableEntry to contained(false) should remove its container styling independently of the parent.

In other words, the nested RepeatableEntry should not retain the container UI styling when contained(false) is explicitly set, regardless of whether the parent RepeatableEntry is contained or not.   

### Steps to reproduce

Create a RepeatableEntry
Inside it, define another RepeatableEntry
Set the nested RepeatableEntry to contained(false)
Leave the parent RepeatableEntry with default behavior (contained = true)

```php
RepeatableEntry::make('parent')
    ->schema([
        RepeatableEntry::make('child')
            ->contained(false)
            ->schema([
                TextEntry::make('name'),
            ]),
    ]);
```

https://github.com/Odion-DS/filament-issue-contained-doesnt-work-in-nested-repeatable-entry-4x

### Relevant log output

```shell

```

People-Sea:
### Fixes: #19285 

Use direct child selector to prevent nested RepeatableEntry from inheriting parent's contained styling.

### before
image

### after
image

## Functional changes


webard:
I think that might have been the intended behavior. After this PR there is no way to distinguish these repeaters.

People-Sea:
> I think that might have been the intended behavior. After this PR there is no way to distinguish these repeaters.

@webard  Thanks for the feedback! However, the reverse case already works  --- child contained(true) inside parent contained(false) renders correctly. This asymmetry suggests it's a selector bug rather than intended behavior.

People-Sea:
That said, the lack of visual separation between items when contained(false) is used could be worth a separate UX  improvement.

danharrin (approved):

PR #19302, opened 2026-02-19
[3.x] fix: add missing autofocus attribute to Radio and ToggleButtons

alanost:
## Description

`->autofocus()` has no effect on Radio and ToggleButtons because their Blade templates don't render the `autofocus` HTML attribute. Other components like Checkbox and TextInput already do this correctly via `'autofocus' => $isAutofocused()` in their attribute merge.

This adds the missing `autofocus` attribute to the first option's `<input>` element for both components.

## Visual changes

None.

## Functional changes

### DIFF
--- a/packages/forms/resources/views/components/radio.blade.php
+++ b/packages/forms/resources/views/components/radio.blade.php
@@ -38,6 +38,7 @@
                         :attributes="
                             \Filament\Support\prepare_inherited_attributes($getExtraInputAttributeBag())
                                 ->merge([
+                                    'autofocus' => $isAutofocused() && $loop->first,
                                     'disabled' => $isDisabled || $isOptionDisabled($value, $label),
                                     'id' => $id . '-' . $value,
                                     'name' => $id,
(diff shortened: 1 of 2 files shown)


packages/forms/resources/views/components/toggle-buttons/index.blade.php
```diff
@@ -54,6 +54,7 @@
                 ])
             >
                 <input
+                    @if ($isAutofocused() && $loop->first) autofocus @endif
```
danharrin:
```suggestion
                    @if ($loop->first && $isAutofocused()) autofocus @endif
```

This reduces unnecessary calls to `$isAutofocused()` for non-first inputs

packages/forms/resources/views/components/radio.blade.php
```diff
@@ -38,6 +38,7 @@
                         :attributes="
                             \Filament\Support\prepare_inherited_attributes($getExtraInputAttributeBag())
                                 ->merge([
+                                    'autofocus' => $isAutofocused() && $loop->first,
```
danharrin:
```suggestion
                                    'autofocus' => $loop->first && $isAutofocused(),
```

danharrin (changes requested):

danharrin (approved):

PR #19327, opened 2026-02-23
feature: Additional rate limits by user ID / email

danharrin:


packages/panels/src/Auth/Pages/PasswordReset/ResetPassword.php
```diff
@@ -133,6 +138,30 @@ protected function getRateLimitedNotification(TooManyRequestsException $exceptio
             ->danger();
     }
 
+    protected function isResetPasswordRateLimited(?string $email): bool
+    {
+        if (blank($email)) {
+            return false;
+        }
+
+        $rateLimitingKey = "filament-reset-password:{$email}";
```
JaZo: I think it's good to hash the email so we don't use PII in rate limit keys.

packages/panels/src/Auth/Pages/Login.php
```diff
@@ -135,6 +144,46 @@ public function authenticate(): ?LoginResponse
...
+        }
+
+        RateLimiter::hit($rateLimitingKey);
+
+        return false;
+    }
+
+    protected function isLoginRateLimited(string $email): bool
+    {
+        $rateLimitingKey = "filament-login:{$email}";
```
JaZo: Ditto.

packages/panels/src/Auth/Pages/Register.php
```diff
@@ -118,6 +123,30 @@ protected function getRateLimitedNotification(TooManyRequestsException $exceptio
             ->danger();
     }
 
+    protected function isRegisterRateLimited(string $email): bool
+    {
+        if (blank($email)) {
+            return false;
+        }
+
+        $rateLimitingKey = "filament-register:{$email}";
```
JaZo: Ditto.

packages/panels/src/Auth/Pages/Login.php
```diff
@@ -135,6 +144,46 @@ public function authenticate(): ?LoginResponse
...
+        }
+
+        RateLimiter::hit($rateLimitingKey);
+
+        return false;
+    }
+
+    protected function isLoginRateLimited(string $email): bool
+    {
+        $rateLimitingKey = 'filament-login:' . sha1($email);
```
JaZo: My only concern is that this change could allow an attacker to lock out a user by repeatedly attempting to log in.

danharrin: Yeah, potentially. I think I've seen it elsewhere before though, like in the Laravel starter kits. Are there any alternatives?

JaZo: I'm affraid there's no alternative, except for removing this kind of rate limiting.

danharrin: I'm going to change the key to be a combination of IP and email instead

urufudev:
Hi 👋  I noticed an issue introduced after the merge feature: Additional rate limits by user ID / email.

When using a custom authentication setup that does not use email as the login identifier (for example, logging in with DNI), the login process throws:

ErrorException: Undefined array key "email"

Environment:

PHP 8.3.29

Laravel 12.53.0

danharrin:
Fixed by #19387

PR #19329, opened 2026-02-23
Add navigationGroup setter to Page and extend navigationGroup setter on Resource to also accept UnitEnum

jyrkidn:
## Description

The `$navigationGroup` property on a resource already accepts an UnitEnum, this has now been added to the setter.
For a page the `navigationGroup()` function has been added.

The setter on the page comes in handy to move the default Dashboard to another group without overriding the whole page + then it also aligns the Resource methods for `$navigationGroup`.

## Visual changes

/ 

## Functional changes

### DIFF
--- a/packages/panels/src/Pages/Page.php
+++ b/packages/panels/src/Pages/Page.php
@@ -182,14 +182,24 @@ abstract class Page extends BasePage
         return static::$navigationParentItem;
     }
 
-    public static function getActiveNavigationIcon(): string | BackedEnum | Htmlable | null
+    public static function navigationParentItem(?string $item): void
     {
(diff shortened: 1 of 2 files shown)


packages/panels/src/Pages/Page.php
danharrin, line 1: If you're going to introduce nav setters on the Page class, please do them all instead of just one

danharrin (changes requested):

jyrkidn:
The other navigation setters are taken over in the Page class. Also made sure that the order in both classes is the same of these functions.

danharrin (approved):

PR #19340, opened 2026-02-24
[3.x] fix: respect `deletable(false)` in repeater relationship save logic

alanost:
## Description

When a Repeater uses `->relationship()` and `->deletable(false)`, the `deletable(false)` only hides the delete button in the UI. However, the `saveRelationshipsUsing` callback still deletes related records that are missing from the repeater's current state.

This is unexpected — if a repeater is explicitly marked as non-deletable, it should not delete related records during save either.

### Steps to reproduce

```php
Repeater::make('members')
    ->relationship('members')
    ->deletable(false)  // UI: no delete button
    ->schema([...]);

// If a record exists in DB but is missing from the repeater state
// (e.g., added by another component), it gets deleted on save
// even though deletable(false) was set.
```

### Use case

This is common in Wizard forms where two repeaters reference the same relationship with different field schemas (e.g., Step A: personal data, Step B: income data). Step B uses `->deletable(false)` because only Step A should manage adding/removing records. Without this fix, Step B's save deletes records it doesn't have in its state.

## Visual changes

None.

## Functional changes

### DIFF
--- a/packages/forms/src/Components/Repeater.php
+++ b/packages/forms/src/Components/Repeater.php
@@ -909,21 +909,23 @@ class Repeater extends Field implements Contracts\CanConcealComponents, Contract
 
             $existingRecords = $component->getCachedExistingRecords();
 
-            $recordsToDelete = [];
+            if ($component->isDeletable()) {
+                $recordsToDelete = [];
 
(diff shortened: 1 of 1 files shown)


danharrin:
I don't think we can merge this as it's a breaking change to existing behaviour (users could be setting the repeater state even though its not deletable in the UI and expecting it to save), we'd need it to be an opt-in method, which you can PR to v4 if you wish.

PR #19360, opened 2026-02-26
feat: Add the ability to customize database notification's date.

ahmed-rashad-alnaggar:
Now represented database notification date can be customized:

```php
use Carbon\CarbonInterface;
use Filament\Enums\DatabaseNotificationsPosition;
use Filament\Panel;

{
    return $panel
        // ...
        ->databaseNotifications()
        ->databaseNotificationsFormatDateUsing(fn(CarbonInterface $date): string => $date->toDateTimeString());
}
```

I needed this feature to make Facebook like notifications:
same-day => `->diffForHumans()`
yesterday => `{yesterday} at h:i A`
older => `F j, h:i A`


danharrin:
Thanks for your contribution! However this time, I am going to decline it. While it is a relatively simple feature, I have not had many requests for it, and it introduces more global configuration methods onto the Panel object, which already has many (probably too many) configuration methods already. I think users will get overwhelmed if we introduce so many rarely-used configuration options to it.

To do this, you could instead set `databaseNotificationsLivewireComponent()` to your own LW component which extends `DatabaseNotifications`, and override the `formatNotificationDate()` method on there to change it. Then we don't need any changes to the core framework, and it is still simple to maintain for you.

Please see our docs about suggesting new features before implementing them, to save yourself time building something we won't merge. https://filamentphp.com/docs/5.x/introduction/contributing#development-of-new-features

ahmed-rashad-alnaggar:
Thank you for the detailed explanation — I really appreciate it.
And noted about the feature proposal process — I’ll make sure to follow the feature discussion guidelines before building similar changes in the future.

PR #19391, opened 2026-03-03
Add compact() method to Table for denser row layout

felipeArnold:
Description
Problem

Filament tables currently have a fixed row height based on default cell padding. There was no built-in way to reduce the visual density of a table,
making it difficult to display large amounts of data in a compact, scannable format without resorting to custom CSS overrides.

Solution

Introduces a compact() fluent method on the Table class that reduces the vertical padding of table rows, producing a denser layout suitable for
data-heavy interfaces.

What was added

CanBeCompact trait (packages/tables/src/Table/Concerns/CanBeCompact.php): follows the standard Filament boolean concern pattern — $isCompact property
defaults to false, a compact(bool | Closure $condition = true) setter, and an isCompact(): bool getter that resolves Closure values via evaluate().
Table.php: registers the new CanBeCompact trait alongside the existing concerns.
row.css: adds the .fi-compact hook class that applies py-2 padding to text cells, action cells, checkbox cells, and the selection cell — reducing row
height without affecting other layout properties.
index.blade.php: resolves $isCompact from the table and conditionally applies the fi-compact CSS class to each row.
CanBeCompactTest.php: unit tests covering the default state, enabling via compact(), disabling via compact(false), and both truthy and falsy Closure
variants.
Usage

Tables\Table::make()
->compact()
Or conditionally:

->compact(fn (): bool => $this->isCompactView)

image

### DIFF
--- a/packages/tables/resources/css/row.css
+++ b/packages/tables/resources/css/row.css
@@ -9,6 +9,26 @@
         @apply bg-gray-50 dark:bg-white/5;
     }
 
+    &.fi-compact {
+        & .fi-ta-text:not(.fi-inline) {
+            @apply py-2;
+        }
+
+        & .fi-ta-cell {
+            &:has(.fi-ta-actions) {
+                @apply py-2;
+            }
+
+            &:has(.fi-ta-record-checkbox) {
+                @apply py-2;
+            }
+
+            &.fi-ta-selection-cell {
+                @apply py-2;
+            }
+        }
+    }
+
     &.fi-collapsed {
         @apply hidden;
     }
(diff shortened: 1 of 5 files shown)


danharrin:
Thanks! At the moment, I'm going to keep this as a change that you can make in your theme rather than a setting on the table, but I may change my mind in the future. Please see our contributing guide: https://filamentphp.com/docs/5.x/introduction/contributing#development-of-new-features

DamienJolly:
I believe this would have clear practical value within Filament, particularly for table widgets on dashboards and analytics-focused pages. A compact table layout would reduce visual clutter and prevent tables from consuming excessive screen space, especially in views that contain multiple tables.

If you ultimately decide to keep this as a theme-level choice rather than a configurable setting, I completely understand. I simply wanted to share where I see tangible benefit and how I would personally make use of it.

PR #19396, opened 2026-03-03
feat: Add afterCreateUsing callback to Repeater for relationship items

dev:
## Description

 This PR adds an `afterCreateUsing()` callback to the Repeater component when used with `->relationship()`.

Currently, the Repeater provides callbacks for mutating data before operations: 
- `mutateRelationshipDataBeforeCreateUsing()` 
- `mutateRelationshipDataBeforeSaveUsing()` 
- `mutateRelationshipDataBeforeFillUsing()`

However, there is no way to execute code after a new relationship record is created and saved. This is needed when the record's ID is required to perform additional operations, such as attaching data via a pivot 
table.

The mutateRelationshipDataBeforeCreateUsing() callback cannot solve this because the record doesn't have an ID yet at that point. 

 **Note:** I also updated the documentation version tags from `4.x` to `5.x` as this appears to have been missed during the v5 release.
 
## Visual changes

N/A

## Functional changes

### DIFF
--- a/packages/forms/docs/12-repeater.md
+++ b/packages/forms/docs/12-repeater.md
@@ -28,7 +28,7 @@ Repeater::make('members')
     ->columns(2)
 ```
 
-<AutoScreenshot name="forms/fields/repeater/simple" alt="Repeater" version="4.x" />
+<AutoScreenshot name="forms/fields/repeater/simple" alt="Repeater" version="5.x" />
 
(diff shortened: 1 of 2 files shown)


ziming:
I thought it is the new filament 5 for a moment 😆

danharrin (changes requested):
Please reopen this to target 4.x, and introduce the same for updates and deletes

PR #19398, opened 2026-03-03
Security Issue: Rate Limiting Bypass and User Enumeration When Using Username-Based Authentication

jeffersongoncalves:
## Summary

PR #19387 ("fix: Login with username instead of email") addressed a bug related to PR #19327 ("Additional rate limits by user ID / email"), but the current implementation of `Login.php` on the `4.x` branch still contains **hardcoded references to `'email'`** in critical security-sensitive methods. This creates rate limiting bypass opportunities and broken error feedback when developers customize the login form to use `username` instead of `email` — a well-documented and commonly recommended customization pattern in the Filament ecosystem.

## Affected File

`packages/panels/src/Auth/Pages/Login.php` (branch `4.x`)

### 1. Hardcoded `'data.email'` in `throwFailureValidationException()` — Silent Authentication Failure

```php
protected function throwFailureValidationException(): never
{
    throw ValidationException::withMessages([
        'data.email' => __('filament-panels::auth/pages/login.messages.failed'),
    ]);
}
```

When a developer overrides the login form to use a `username` field (by overriding `getEmailFormComponent()` to return a `TextInput::make('username')`), this method throws the validation error against `data.email` — a field that **no longer exists** in the form.

**Security impact:**

- The "These credentials do not match our records" error message is **silently swallowed** — it never appears in the UI because there is no `email` field to attach it to.
- Attackers performing brute-force or credential-stuffing attacks receive **no visible feedback**, making it harder for legitimate users to notice suspicious activity on their accounts.
- This inconsistency in error behavior between email-based and username-based setups can be exploited as a **side-channel** to fingerprint the authentication backend configuration.

### 2. Hardcoded `$data['email']` in `getCredentialsFromFormData()` — Credential Mapping Failure

```php
protected function getCredentialsFromFormData(array $data): array
{
    return [
        'email' => $data['email'],
        'password' => $data['password'],
    ];
}
```

If a developer overrides `getEmailFormComponent()` to use `username` but forgets to also override `getCredentialsFromFormData()`, the method will attempt to access `$data['email']` which will be `null`. This means:

- The credential array sent to `retrieveByCredentials()` will contain `'email' => null`.
- Depending on the Eloquent/auth provider implementation, this could match unexpected records or throw unhandled exceptions, potentially leaking information about the database schema or user existence.

### 3. Per-User Rate Limiting Bypass (Related to PR #19327)

PR #19327 introduced additional rate limits keyed by user ID or email. If any of this per-user rate limiting logic relies on the form data key being `'email'`, then username-based login setups will **bypass the per-user rate limit entirely**, as the expected key is absent from the form data.

This means an attacker could:
- Circumvent per-user brute-force protection by targeting a panel configured with username-based login.
- Perform unlimited credential-stuffing attempts against a specific user, restricted only by the IP-based rate limit (5 attempts).

## Reproduction Steps

1. Create a custom Login page extending `Filament\Auth\Pages\Login`.
2. Override `getEmailFormComponent()` to return `TextInput::make('username')`.
3. Override `getCredentialsFromFormData()` to use `$data['username']`.
4. **Do NOT** override `throwFailureValidationException()` (most tutorials/guides don't mention this).
5. Attempt to log in with invalid credentials.
6. **Observe:** No error message appears in the UI — the validation error is attached to `data.email` which doesn't exist.
7. **Observe:** An attacker can repeatedly submit invalid credentials without the user ever seeing a warning.

## Suggested Fix

The `throwFailureValidationException()` method should dynamically resolve the field name used in the login form, rather than hardcoding `'data.email'`. For example:

```php
protected function throwFailureValidationException(): never
{
    // Dynamically resolve the login field name from the form component
    $loginField = $this->getEmailFormComponent()->getName();

    throw ValidationException::withMessages([
        "data.{$loginField}" => __('filament-panels::auth/pages/login.messages.failed'),
    ]);
}
```

Similarly, `getCredentialsFromFormData()` could be made more resilient by using the form component's name dynamically.

### DIFF
--- a/packages/panels/src/Auth/Pages/Login.php
+++ b/packages/panels/src/Auth/Pages/Login.php
@@ -78,6 +78,10 @@ class Login extends SimplePage
 
         $data = $this->form->getState();
 
+        if ($this->isLoginRateLimited($data[$this->getEmailFormComponent()->getName()])) {
+            return null;
+        }
+
         /** @var SessionGuard $authGuard */
         $authGuard = Filament::auth();
 
@@ -160,6 +164,26 @@ class Login extends SimplePage
         return false;
     }
 
+    protected function isLoginRateLimited(string $email): bool
+    {
+        $rateLimitingKey = 'filament-login:' . sha1(request()->ip() . '|' . $email);
+
+        if (RateLimiter::tooManyAttempts($rateLimitingKey, maxAttempts: 5)) {
+            $this->getRateLimitedNotification(new TooManyRequestsException(
+                static::class,
+                'authenticate',
+                request()->ip(),
+                RateLimiter::availableIn($rateLimitingKey),
+            ))?->send();
+
+            return true;
+        }
+
+        RateLimiter::hit($rateLimitingKey);
+
+        return false;
+    }
+
     protected function getRateLimitedNotification(TooManyRequestsException $exception): ?Notification
     {
         return Notification::make()
@@ -184,8 +208,10 @@ class Login extends SimplePage
 
     protected function throwFailureValidationException(): never
     {
+        $loginField = $this->getEmailFormComponent()->getName();
+
         throw ValidationException::withMessages([
-            'data.email' => __('filament-panels::auth/pages/login.messages.failed'),
+            "data.$loginField" => __('filament-panels::auth/pages/login.messages.failed'),
         ]);
     }
 


danharrin:
There is rate limiting at the start of this method, this is not a vuln.

PR #19400, opened 2026-03-03
[4.x] Allow hiding specific error notifications

coleshirley:
## Description

Allow error notifications with certain status codes to be hidden. This is useful for utilizing the built in filament error notifications for the majority of errors but implementing custom logic for specific http status codes

## Functional changes


danharrin (changes requested):
I'm not sure about the naming here. "Hidden" to me feels like the error is still caught (and doesn't show the error modal) but there isn't an error notification either. But in your implementation, the error modal isn't prevented. Maybe we need both options, with the current implementation being called something else. Thoughts?

coleshirley:
That makes sense. What do you think about `bypassErrorHandling` or `skipErrorHandling` for the current logic and then `hideErrorNotification` does what you suggest?

danharrin (approved):

PR #19401, opened 2026-03-03
fix: gracefully handle unresolvable actions when record state changes

andrewdwallo:
## Description

When a user clicks on an action whose associated record state has changed (e.g., the record was deleted or modified by another user/process), `cacheMountedActions()` and `getMountedAction()` throw an                      
  `ActionNotResolvableException`, causing an unhandled error. This change gracefully catches the exception in two places:

  - `InteractsWithTable::bootedInteractsWithTable()` — resets the mounted action state instead of crashing during the Livewire boot lifecycle.
  - `InteractsWithActions::callMountedAction()` — unmounts the action and returns `null` instead of throwing, consistent with how `mountAction()` and `unmountAction()` already handle this exception.

## Visual changes
N/A

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/InteractsWithActions.php
+++ b/packages/actions/src/Concerns/InteractsWithActions.php
@@ -194,9 +194,15 @@ trait InteractsWithActions
      */
     public function callMountedAction(array $arguments = []): mixed
     {
-        $action = $this->getMountedAction();
+        try {
+            $action = $this->getMountedAction();
+        } catch (ActionNotResolvableException $exception) {
+            $action = null;
+        }
 
         if (! $action) {
+            $this->unmountAction(canCancelParentActions: false);
+
             return null;
         }
 
--- a/packages/actions/src/Concerns/InteractsWithActions.php
+++ b/packages/actions/src/Concerns/InteractsWithActions.php
@@ -98,7 +98,9 @@ trait InteractsWithActions
 
         // Boot the InteractsWithTable trait first so the table object is available.
         if (! ($this instanceof HasTable)) {
-            $this->cacheMountedActions($this->mountedActions);
+            if (empty($this->cacheMountedActions($this->mountedActions))) {
+                $this->mountedActions = [];
+            }
         }
     }
 
@@ -483,8 +485,6 @@ trait InteractsWithActions
         try {
             return $this->cachedMountedActions = $this->resolveActions($mountedActions);
         } catch (ActionNotResolvableException) {
-            $this->mountedActions = [];
-
             return $this->cachedMountedActions = [];
         }
     }
--- a/packages/actions/src/Concerns/InteractsWithActions.php
+++ b/packages/actions/src/Concerns/InteractsWithActions.php
@@ -197,8 +197,6 @@ trait InteractsWithActions
         $action = $this->getMountedAction();
 
         if (! $action) {
-            $this->unmountAction(canCancelParentActions: false);
-
             return null;
         }
 
--- a/packages/actions/src/Concerns/InteractsWithActions.php
+++ b/packages/actions/src/Concerns/InteractsWithActions.php
@@ -194,11 +194,7 @@ trait InteractsWithActions
      */
     public function callMountedAction(array $arguments = []): mixed
     {
-        try {
-            $action = $this->getMountedAction();
-        } catch (ActionNotResolvableException $exception) {
-            $action = null;
-        }
+        $action = $this->getMountedAction();
 
         if (! $action) {
             $this->unmountAction(canCancelParentActions: false);
@@ -486,7 +482,13 @@ trait InteractsWithActions
      */
     protected function cacheMountedActions(array $mountedActions): array
     {
-        return $this->cachedMountedActions = $this->resolveActions($mountedActions);
+        try {
+            return $this->cachedMountedActions = $this->resolveActions($mountedActions);
+        } catch (ActionNotResolvableException) {
+            $this->mountedActions = [];
+
+            return $this->cachedMountedActions = [];
+        }
     }
 
     /**
@@ -597,6 +599,14 @@ trait InteractsWithActions
 
         if ($action['context']['bulk'] ?? false) {
             $resolvedAction = $this->getTable()->getBulkAction($action['name']);
+
+            if ($resolvedAction && ! empty($this->selectedTableRecords) && $this->getTable()->hasQuery()) {
+                $count = $this->getSelectedTableRecordsQuery(shouldFetchSelectedRecords: false)->count();
+
+                if ($count === 0) {
+                    throw new ActionNotResolvableException('The selected records for the bulk action no longer exist.');
+                }
+            }
         }
 
         $resolvedAction ??= $this->getTable()->getAction($action['name']) ?? throw new ActionNotResolvableException("Action [{$action['name']}] not found on table.");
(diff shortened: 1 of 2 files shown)


packages/actions/src/Concerns/InteractsWithActions.php
danharrin, line 1: Does it work if we add the try/catch in the boot hook of this trait around `cacheMountedActions()` instead, maybe?

We could even potentially move the try/catch into `cacheMountedActions()` itself?

danharrin (changes requested):

packages/actions/src/Concerns/InteractsWithActions.php
andrewdwallo: I moved it. Additionally, I added validation when resolving bulk actions for if the selected records no longer exist within the table's scoped query (e.g., they were deleted or fell out of a filtered view), the action is now gracefully unmounted instead of proceeding with an empty collection (which causes the same issues).

packages/actions/src/Concerns/InteractsWithActions.php
```diff
@@ -480,7 +482,13 @@ public function getMountedAction(?int $actionNestingIndex = null): ?Action
      */
     protected function cacheMountedActions(array $mountedActions): array
     {
-        return $this->cachedMountedActions = $this->resolveActions($mountedActions);
+        try {
+            return $this->cachedMountedActions = $this->resolveActions($mountedActions);
+        } catch (ActionNotResolvableException) {
+            $this->mountedActions = [];
```
danharrin: I don't like the fact that the cache method also overwrites the non-cached array, is this line definitely nessecary?

packages/actions/src/Concerns/InteractsWithActions.php
```diff
@@ -591,6 +599,14 @@ protected function resolveTableAction(array $action, array $parentActions): Acti
...
         if ($action['context']['bulk'] ?? false) {
             $resolvedAction = $this->getTable()->getBulkAction($action['name']);
+
+            if ($resolvedAction && ! empty($this->selectedTableRecords) && $this->getTable()->hasQuery()) {
+                $count = $this->getSelectedTableRecordsQuery(shouldFetchSelectedRecords: false)->count();
+
+                if ($count === 0) {
+                    throw new ActionNotResolvableException('The selected records for the bulk action no longer exist.');
+                }
+            }
```
danharrin: I am concerned that this is not strictly required and is causing an extra count query to be executed on each request. If this is not suppressing an exception, lets remove it.

packages/actions/src/Concerns/InteractsWithActions.php
```diff
@@ -197,6 +197,8 @@ public function callMountedAction(array $arguments = []): mixed
         $action = $this->getMountedAction();
 
         if (! $action) {
+            $this->unmountAction(canCancelParentActions: false);
```
danharrin: Can you explain if this is strictly required to supress an error?

From what I can tell, the only place we need to change in this PR is catching the exception in `cacheMountedActions()` to fix the original issue.

danharrin (changes requested):

packages/actions/src/Concerns/InteractsWithActions.php
andrewdwallo: The bulk action check is strictly required to suppress the same error. For single actions, `resolveTableAction()` already throws `ActionNotResolvableException` when the record no longer exists. But for bulk actions, there's no equivalent check. When the selected records fall out of the table's scoped query (e.g., a preset view changes while a bulk action modal is open), the action resolves successfully but downstream record access crashes. Without this fix users would need to specifically handle when a bulk action's collection of records is empty, which is typically not something they would need to handle considering a bulk action needs records just like the single actions need a non-null record. The `count()` query is the bulk action equivalent of the single action's `getTableRecord()` null check. Both hit the database to verify records still exist. If there's a less expensive way to verify the selected records still exist within the scoped query, I'm happy to use it, but as far as I can tell there's no cached version available at this point in the lifecycle.

packages/actions/src/Concerns/InteractsWithActions.php
andrewdwallo: Yes, this line is necessary. Without it, when the exception is caught and `cachedMountedActions` is set to `[]`, the `mountedActions` array still contains stale entries referencing actions that can no longer be resolved. On the next Livewire request cycle, code that iterates over `mountedActions` and tries to look up the corresponding cached action gets `null`, which causes a `Call to a member function getNestingIndex() on null` error in `PartialsComponentHook`.

To reproduce: open the same record's delete modal in two browser tabs, confirm the delete in the first tab, then confirm the delete in the second tab. Without clearing `mountedActions`, the second tab throws the `getNestingIndex() on null` exception instead of gracefully unmounting.

I understand the concern about a cache method modifying the source array, but the two arrays need to stay in sync. If `cachedMountedActions` is empty, `mountedActions` must also be empty, otherwise downstream code assumes there are mounted actions but finds nothing when it looks them up in the cache.

packages/actions/src/Concerns/InteractsWithActions.php
andrewdwallo: Yes you are right, that can be removed. I removed this.

packages/actions/src/Concerns/InteractsWithActions.php
danharrin: How about `cachedMountedActions` being set to null instead of an empty string? If that is the case, then the mounted actions will attempt caching again even if this method returns `[]`, and the downstream code doesn't assume anything?

packages/actions/src/Concerns/InteractsWithActions.php
danharrin: I think I disagree that an empty array of records is not being handled in many cases, at least currently any code that does not handle it already errors in this case, and in my experience, most bulk actions just loop through whatever records are passed and process them, or pass their keys to some query scope before updating. I think we can keep this how it was.

packages/actions/src/Concerns/InteractsWithActions.php
andrewdwallo: Setting `cachedMountedActions` to null causes a resolution loop (the exists() query runs 3 times as it keeps re-attempting), and eventually PartialsComponentHook calls `getNestingIndex()` on a null cached action. The stale mountedActions entries must be cleared to break the cycle.

danharrin: Okay, then can setting `$this->mountedActions` if the cached mounted actions are empty be done outside of this method? Again I just want to avoid side effects, there might be a better place for side effects, like the mount/call/boot action method or something?

andrewdwallo: That makes sense. There are 4 call sites for `cacheMountedActions()` — the two boot hooks (`bootedInteractsWithActions` and `bootedInteractsWithTable`), `getMountedActions()`, and `getMountedAction()`.   
 
Would you prefer:                                                                                                                                                                                            
                                                                                                                                                                                                               
1. Each caller checks if the result is empty and clears `mountedActions` itself
2. Only the two boot hooks handle the clearing (since that's where the issue manifests)

packages/actions/src/Concerns/InteractsWithActions.php
andrewdwallo: Okay I'll remove it

packages/actions/src/Concerns/InteractsWithActions.php
danharrin: In as few places as possible please (so probably boot hooks?), I just want to minimise impact of this PR since this is a delicate part of the system

andrewdwallo: I understand your concern. I updated it to clear mountedActions in the callers instead.

danharrin (approved):

Issue #13306, opened 2024-06-18
FileUpload: Some files are missing when uploading multiple files


uma-ac:
![FileUploadMissingFile](https://github.com/filamentphp/filament/assets/121853633/f25586a3-e569-46a0-92a6-3184d6b25db1)
<img width="744" alt="FileUploadMissingFile_temp" src="https://github.com/filamentphp/filament/assets/121853633/92d08d8d-9a91-4157-b063-33c1b310f0be">
<img width="991" alt="Screenshot 2024-06-18 at 10 33 41" src="https://github.com/filamentphp/filament/assets/121853633/50f20c4d-9b1c-4b7e-95b9-c66396523b83">
### Package

filament/filament

### Package Version

3

### Laravel Version

11

### Livewire Version

V3

### PHP Version

php 8

### Problem description

When I upload multiple files, some of the files are missing.

### Expected behavior

All the files should be present in the desired folder.

### Steps to reproduce

`Section::make('')
                    ->relationship('result_files')
                    ->schema([
                    FileUpload::make('attachments')
                        ->columnSpan(12)
                        ->multiple()
                        ->maxSize(100000)
                        ->disk('uploads')
                        ->directory('result_files')
                        ->acceptedFileTypes(["application/pdf"])
                        ->getUploadedFileNameForStorageUsing(function ($file): string {
                            $fileName = str($file->getClientOriginalName());
                            $fileName_str = preg_replace('/[^a-zA-Z0-9.]/','',iconv('UTF-8', 'ASCII//TRANSLIT',  $fileName));
                            return (string) str($fileName_str)->prepend(now() . '-');
                        })
                        ->downloadable()
                        ->rules([ new MalwareFile] )
                        ->label(__('application.resource.result_documents'))
                ]),`

`    FileUpload::make('attachments')
                        ->columnSpan(12)
                        ->multiple()
                        ->maxSize(100000)
                        ->disk('uploads')
                        ->visibility('private')
                        ->directory('application_files')
                        ->getUploadedFileNameForStorageUsing(function ($file): string {
                            $fileName = str($file->getClientOriginalName());
                            $fileName_str = preg_replace('/[^a-zA-Z0-9.]/','',iconv('UTF-8', 'ASCII//TRANSLIT',  $fileName));
                            return (string) str($fileName_str)->prepend(now() . '-');
                        })
                        ->downloadable()
                        ->rules([ new MalwareFile] )
                        ->label(__('application.resource.application_documents')),`

[...]

PR #19402, opened 2026-03-04
fix: Enable native picker for date/time TextInputColumn types

### LINKED ISSUE #15895: Datepicker not working with TextInputColumn
### Package

filament/filament

### Package Version

v3.3.4

### Laravel Version

v11.44.2

### Livewire Version

v3.6.2

### PHP Version

PHP 8.3.17

### Problem description

Since [release 3.2.134](https://github.com/filamentphp/filament/releases/tag/v3.2.134) the datepicker in TextInputColumn stopped working when used as `TextInputColumn::make('load_date')->type('date')`. The datepicker doesn't open.

It is caused by https://github.com/filamentphp/filament/pull/15340, specifically [this change](https://github.com/filamentphp/filament/pull/15340/files#diff-6b34c752663961dbdeeafe072594ffd77d5e6255eb2565838f84ecede3cfd910), when reverted, the column works fine.

### Expected behavior

When user clicks on the calendar icon, a datepicker should open.

### Steps to reproduce

1. add `TextInputColumn::make('date')->type('date')` to your Filament table
2. `composer require filament/filament:3.2.133 -W`
3. the datepicker opens and inline editing works
4. `composer require filament/filament:3.2.134 -W`
5. the datepicker doesn't open

https://github.com/jvitasek/app-filament-issue-3.x

### Relevant log output

```shell

```

isaackaara:
## Description

Fixes #15895

Since PR #15340 added `x-on:click.prevent.stop` to the `TextInputColumn` wrapper div, the native browser datepicker no longer opens for `TextInputColumn::make("field")->type("date")` (and other date/time types). The `.prevent` modifier calls `preventDefault()` which blocks the native picker popup.

### Root cause

PR #15340 changed `x-on:click.stop=""` to `x-on:click.prevent.stop` on the input wrapper to prevent stacked column clicks from triggering the row `<a>` link navigation. The `.stop` (stopPropagation) was already sufficient for preventing event bubbling, but `.prevent` (preventDefault) was added as a belt-and-suspenders measure. Unfortunately, `preventDefault()` also blocks native browser pickers (date, time, etc.) from opening.

### Fix

The codebase already handles `color` type inputs with an explicit `showPicker()` onclick handler. This PR extends that pattern to also cover `date`, `datetime-local`, `month`, `time`, and `week` input types, calling `showPicker()` to programmatically open the native picker despite `preventDefault()`.

### Steps to reproduce (before fix)

1. Add `TextInputColumn::make("some_date")->type("date")` to a Filament table
2. Click the calendar icon or input field
3. The native datepicker does not open

### After fix

The native datepicker opens correctly via `showPicker()`.

### Affected types

- `date`
- `datetime-local`
- `month`
- `time`
- `week`

## Visual changes

None (restores expected browser-native behavior).

## Functional changes

### DIFF
--- a/packages/tables/src/Columns/TextInputColumn.php
+++ b/packages/tables/src/Columns/TextInputColumn.php
@@ -303,7 +303,7 @@ class TextInputColumn extends Column implements Editable, HasEmbeddedView
 
                 <div class="fi-input-wrp-content-ctn">
                     <input
-                        <?php if (in_array($type, ['color'])) { ?>
+                        <?php if (in_array($type, ['color', 'date', 'datetime-local', 'month', 'time', 'week'])) { ?>
                             onclick="if (typeof this.showPicker === 'function') { this.showPicker() }"
                         <?php } ?>
                         x-model.lazy="state"
--- a/tests/src/Tables/Columns/TextInputColumnTest.php
+++ b/tests/src/Tables/Columns/TextInputColumnTest.php
@@ -25,6 +25,14 @@ it('can render', function (): void {
         ->assertCanRenderTableColumn('rating');
 });
 
+it('renders showPicker onclick for date type inputs', function (): void {
+    Post::factory()->count(1)->create();
+
+    livewire(TestTableWithDateTextInputColumn::class)
+        ->assertSuccessful()
+        ->assertSeeHtml('showPicker');
+});
+
 it('can display different values', function (): void {
     Post::factory()->create(['rating' => 1]);
     Post::factory()->create(['rating' => 5]);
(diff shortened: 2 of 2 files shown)


People-Sea:
@isaackaara 

Thanks for the PR.

I looked into this issue before and couldn’t reproduce file loss with unique filenames.

What I can reproduce is filename collision from custom getUploadedFileNameForStorageUsing:

unique names, with/without poll: all files stay
forced same name, with/without poll: paths collapse to one file
So I’m not convinced this PR is fixing the root issue in #13306.

Do you have a minimal repro (repo + exact steps) where files still go missing with unique filenames?
Also, this PR needs a regression test for that case.

isaackaara:
Hi Wen — just wanted to flag that this PR is specifically for the date/time picker issue in TextInputColumn (enabling the native browser picker which was blocked by the `.prevent` modifier). It's unrelated to file uploads. If you have any specific feedback on the picker implementation itself, happy to address it. Thanks for reviewing!

danharrin:
This breaks manual input in some browsers, see the previous PR.

Issue #14474, opened 2024-10-09
The FileUpload component when set as multiple() sometimes skips some files if the relation manager has ->poll().


danilogiacomi:
### Package

filament/forms

### Package Version

v3.2.116

### Laravel Version

v11.27.2

### Livewire Version

v3.5.4

### PHP Version

PHP 8.3.12

### Problem description

The FileUpload component when set as multiple() sometimes skips some files if the relation manager has ->poll().

In this scenario I have an action opening a modal with the FileUpload in its form and the creation of one entities for each uploaded file (within the action itself).

The Resource, then, has a relationManager with the entities created in the above-mentioned action.

If in the relationManager we add a ->poll('5s') then the issue appears, if we don't it works flawlessly.

I've created a repo with the two versions of the Resources so you can try it out both the working and the non-working version.

Please find the repository at the following URL, I've added a little instruction and additional information in the README and it also is shipped with some test files to be uploaded, for your convenience.

https://github.com/danilogiacomi/filamentFileUploadProblem

### Expected behavior

All the file uploaded to be present

### Steps to reproduce

Using the provided repository, got to /admin/polling-orders create a "New polling order", use the "Mass upload images" button to upload mulitple files (the repo also contains 32 sample files in the "exampleFiles" directory.

Note that the /admin/orders resource, which doesn't have the ->poll() in the resourceManager works flawlessly

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/danilogiacomi/filamentFileUploadProblem

### Relevant log output

```shell
N/A
```

NathanLochala:
I can confirm this issue. In fact, I can expand on it. When polling is set on a RelationManager table, any action/form associated with that table will "defocus" when the table polls.

danharrin:
I can't reproduce this anymore, so I think it was fixed by the modal key changes I made in #16488. Let me know if you still experience it though as I might have been doing something wrong

danilogiacomi:
Hi, 
I don't think it's fixed, if you clone the repository in the initial comment and try it out you'll find that the problem is still there.

danharrin:
I did clone the repository and test it out when I closed the issue, I can't reproduce it anymore. I will need more info

danilogiacomi:
> I did clone the repository and test it out when I closed the issue, I can't reproduce it anymore. I will need more info

The thing is that it's not sistematic, sometimes it loads all the 32 files, sometimes it only loads 31 of them, other times 28, etc...

[...]

PR #19403, opened 2026-03-04
fix: Prevent file loss during parallel multi-file uploads

### LINKED ISSUE #13306: FileUpload: Some files are missing when uploading multiple files

isaackaara:
## Description

Fixes #13306

When uploading multiple files via `FileUpload::make()->multiple()`, some files go missing. This is a race condition in the `shouldUpdateState` flag management during parallel uploads.

### Root cause

The `shouldUpdateState` flag is set to `false` when a file upload starts, and back to `true` when it completes. With `maxParallelUploads` defaulting to 2, this creates a race condition:

1. File A starts uploading -> `shouldUpdateState = false`
2. File B starts uploading -> `shouldUpdateState` already `false`
3. File A finishes -> `shouldUpdateState = true`
4. A Livewire poll or re-render triggers the `$watch("state")` handler
5. Since `shouldUpdateState` is now `true`, the handler overwrites the FilePond file list with server state
6. File B (still uploading) is lost

This explains why `->maxParallelUploads(1)` was reported as a workaround (serializes uploads, no race), and why `->poll()` on a relation manager exacerbates the issue (more frequent re-renders during the vulnerability window).

### Fix

Introduces an `activeUploads` counter that tracks the number of concurrent in-flight uploads. The `shouldUpdateState` flag is only re-enabled when the counter reaches zero (all uploads complete). This covers success, error, and abort paths.

### Before (race condition)
```
File A start -> shouldUpdateState = false
File B start -> shouldUpdateState = false (no-op)
File A done  -> shouldUpdateState = true  <-- VULNERABLE
[poll fires, state overwrites, File B lost]
File B done  -> shouldUpdateState = true
```

### After (fixed)
```
File A start -> activeUploads = 1, shouldUpdateState = false
File B start -> activeUploads = 2, shouldUpdateState = false
File A done  -> activeUploads = 1, shouldUpdateState stays false
[poll fires, state watcher skips update - safe]
File B done  -> activeUploads = 0, shouldUpdateState = true
```

### Regression tests

**PHP tests** (`tests/src/Forms/Components/FileUploadTest.php`):
- Validates that all files persist in state when uploading multiple files to a `->multiple()` FileUpload
- Tests sequential `fillForm` calls simulating incremental upload completions
- Verifies save works correctly with parallel multiple uploads

**JS tests** (`tests/js/file-upload-active-uploads.test.mjs`):
- Directly tests the `activeUploads` counter logic extracted from the Alpine component
- Validates that `shouldUpdateState` stays `false` while any upload is in flight
- Covers success, error, and abort code paths
- Simulates the exact #13306 scenario: poll fires after first file completes but before second finishes
- Zero dependencies, runs with `node --test tests/js/`

## Visual changes

None.

## Functional changes

- [x] Regression tests added for the race condition fix.

### DIFF
--- a/packages/forms/resources/js/components/file-upload.js
+++ b/packages/forms/resources/js/components/file-upload.js
@@ -80,6 +80,8 @@ export default function fileUploadFormComponent({
     uploadUsing,
 }) {
     return {
+        activeUploads: 0,
+
         fileKeyIndex: {},
 
         pond: null,
@@ -200,6 +202,7 @@ export default function fileUploadFormComponent({
                         progress,
                         abort,
                     ) => {
+                        this.activeUploads++
                         this.shouldUpdateState = false
 
                         let fileKey = (
@@ -220,16 +223,34 @@ export default function fileUploadFormComponent({
                             fileKey,
                             file,
                             (fileKey) => {
-                                this.shouldUpdateState = true
+                                this.activeUploads--
+
+                                if (this.activeUploads === 0) {
+                                    this.shouldUpdateState = true
+                                }
 
                                 load(fileKey)
                             },
-                            error,
+                            (errorMessage) => {
+                                this.activeUploads--
+
+                                if (this.activeUploads === 0) {
+                                    this.shouldUpdateState = true
+                                }
+
+                                error(errorMessage)
+                            },
                             progress,
                         )
 
                         return {
                             abort: () => {
+                                this.activeUploads--
+
+                                if (this.activeUploads === 0) {
+                                    this.shouldUpdateState = true
+                                }
+
                                 cancelUploadUsing(fileKey)
                                 abort()
                             },
--- a/tests/js/file-upload-active-uploads.test.mjs
+++ b/tests/js/file-upload-active-uploads.test.mjs
@@ -1,0 +1,180 @@
+/**
+ * Regression test for the activeUploads counter fix.
+ * https://github.com/filamentphp/filament/issues/13306
+ *
+ * Tests the counter logic that prevents shouldUpdateState from being set to
+ * true while parallel uploads are still in flight. Run with: node --test tests/js/
+ *
+ * BEFORE the fix, shouldUpdateState was a simple boolean toggle:
+ *   File A starts -> shouldUpdateState = false
+ *   File B starts -> shouldUpdateState = false (no-op)
+ *   File A done   -> shouldUpdateState = true  <-- VULNERABLE
+ *   [poll re-render fires here, state overwrites, File B lost]
+ *
+ * AFTER the fix, an activeUploads counter gates the flag:
+ *   File A starts -> activeUploads=1, shouldUpdateState = false
+ *   File B starts -> activeUploads=2, shouldUpdateState = false
+ *   File A done   -> activeUploads=1, shouldUpdateState stays false
+ *   [poll re-render fires here, state watcher skips -- safe]
+ *   File B done   -> activeUploads=0, shouldUpdateState = true
+ */
+
+import { describe, it } from 'node:test'
+import assert from 'node:assert/strict'
+
+/**
+ * Minimal simulation of the file-upload component's upload tracking logic.
+ * Extracted from packages/forms/resources/js/components/file-upload.js
+ */
+function createUploadTracker() {
+    return {
+        activeUploads: 0,
+        shouldUpdateState: true,
+
+        startUpload() {
+            this.activeUploads++
+            this.shouldUpdateState = false
+        },
+
+        finishUpload() {
+            this.activeUploads--
+            if (this.activeUploads === 0) {
+                this.shouldUpdateState = true
+            }
+        },
+
+        errorUpload() {
+            this.activeUploads--
+            if (this.activeUploads === 0) {
+                this.shouldUpdateState = true
+            }
+        },
+
+        abortUpload() {
+            this.activeUploads--
+            if (this.activeUploads === 0) {
+                this.shouldUpdateState = true
+            }
+        },
+    }
+}
+
+describe('activeUploads counter', () => {
+    it('should block state updates while any upload is in flight', () => {
+        const tracker = createUploadTracker()
+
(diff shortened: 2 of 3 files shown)


People-Sea:
@isaackaara 

I looked into this issue before and couldn’t reproduce file loss with unique filenames.

What I can reproduce is filename collision from custom getUploadedFileNameForStorageUsing:

unique names, with/without poll: all files stay
forced same name, with/without poll: paths collapse to one file
So I’m not convinced this PR is fixing the root issue in #13306.

Do you have a minimal repro (repo + exact steps) where files still go missing with unique filenames?
Also, this PR needs a regression test for that case.

isaackaara:
@People-Sea Thanks for testing this -- your findings are helpful and I think you have identified a second, separate issue.

**Filename collisions vs state race condition**

You are right that `getUploadedFileNameForStorageUsing` with `now()` can produce collisions when uploads run in parallel (same timestamp). That is a real bug, but it is distinct from what this PR addresses.

This PR targets the `shouldUpdateState` race condition that causes file loss specifically when `->poll()` is active on the page. The key reproduction path:

1. `FileUpload::make()->multiple()` (default `maxParallelUploads(2)`)
2. A relation manager or page with `->poll()` active
3. Upload several files at once

The poll triggers a Livewire re-render while uploads are still in flight. Because `shouldUpdateState` gets set back to `true` after the first file completes, the state watcher fires and overwrites the FilePond file list, dropping any uploads still in progress.

**Existing reproduction repo**

@danilogiacomi provided a reproduction repo in #14474 that demonstrates this exact scenario without any custom filename logic: https://github.com/danilogiacomi/filamentFileUploadProblem

The repo has two resources -- one with `->poll('5s')` and one without. File loss only occurs on the polling version. The README notes the issue is random (timing-dependent), which is consistent with a race condition.

**Community workarounds also support this**

Both workarounds reported in #13306 align with the race condition:
- `->maxParallelUploads(1)` works because it serializes uploads (no window where one file is done but another is still in flight)
- Removing `->poll()` works because it eliminates the re-render trigger during the vulnerability window

**Regression test**

I will add a test for this. The tricky part is that it is timing-dependent (needs a re-render to fire between individual file completions), so I will look at whether we can simulate that reliably in the test suite.

Let me know if you would like me to adjust anything in the approach.

isaackaara:
Thanks for the thorough review, Wen. You're right to distinguish the two issues — filename collisions from `getUploadedFileNameForStorageUsing` are a separate problem from the `shouldUpdateState` race condition this PR targets.

The race condition occurs when `->poll()` is active: a Livewire re-render can fire between individual file completions, resetting `shouldUpdateState` to `false` while uploads are still in flight. This causes FilePond state to be overwritten mid-upload, dropping whichever files hadn't finished yet. The fix uses an `activeUploads` counter to gate state updates until all concurrent uploads are complete.

I've now added regression tests to the PR:
- PHP tests covering multi-file state persistence across Livewire renders
- JS unit tests that directly simulate the race condition scenario (poll fires between file completions) and verify all 5 cases pass

Happy to provide additional context or tweak the approach if you have concerns about the implementation. Let me know!

danharrin:

PR #19405, opened 2026-03-04
fix: reinitialize charts when container becomes visible

### LINKED ISSUE #16599: v4 Charts not updating when in non-active tabs
### Package

filament/filament

### Package Version

4-beta6

### Laravel Version

12

### Livewire Version

3

### PHP Version

8.4

### Problem description

Not sure if I'm doing something wrong, but I can't get charts in nested non-active tabs to render properly on page load, or update properly. The video shows the problem using the reproduction repo. 

I'm using this to render the chart in 3 places:
```PHP
Livewire::make(StatChartWidget::class, fn ($record) => [
        'record' => $record,
    ])
    ->key(Str::random(10))
```

In the video, the following happens:

**1. Page load**
- Chart in the body loads ✅
- Chart in the visible tab loads ✅
- Chart in the non-visible nested "Data" tab is empty ❌

**2. "Refresh data" to update the model's `data` attribute _while the 'Data' tab is not active_**
- Chart in the body refreshes ✅
- Chart in the visible tab refreshes ✅
- Chart in the non-visible nested "Data" tab is not refreshed ❌

**2. "Refresh data" to update the model's `data` attribute _while the 'Data' tab is active_**
- Chart in the body refreshes ✅
- Chart in the visible tab refreshes ✅
- Chart in the non-visible nested "Data" tab refreshes ✅

I'm using `#[Reactive]` but I also tried refreshing via an event, and tried `->lazy()`. Neither helped.

https://github.com/user-attachments/assets/6cb1cb43-8c9b-4f3e-ab62-face3931473a

### Expected behavior

Charts / Livewire components in a page should render and update correctly regardless of what tab they're in.

### Steps to reproduce

1. Clone repo
2. Run `migrate:fresh --seed` to seed statistics table
3. Go to Statistics resource
4. See video and steps outlined in description to reproduce

https://github.com/binaryfire/filament-bug-reproduction

### Relevant log output

```shell

```

isaackaara:
## Problem

Charts placed in non-active tabs or hidden containers do not render correctly. When the tab containing a chart becomes active, the chart is either empty or has incorrect dimensions. Data updates sent while the chart is in a hidden tab are also lost.

## Root Cause

Chart.js relies on the canvas element's computed dimensions to render. When a chart is in a non-active tab (typically `display: none` or zero-height container), the canvas has zero dimensions. This causes:

1. Charts initialized while hidden render with zero width/height
2. `chart.update('resize')` called while hidden has no visible effect
3. When the tab becomes active, the chart is not re-rendered

## Fix

Added an `IntersectionObserver` to the chart Alpine component that:

1. Detects when the chart container enters the viewport (becomes visible)
2. If the chart instance exists but was updated while hidden, calls `chart.resize()` to recalculate dimensions
3. If the chart was never initialized (container was hidden at mount time), calls `initChart()` to create it with correct dimensions

Also updated the `updateChartData` handler to gracefully handle the case where the chart instance doesn't exist yet (deferred initialization), storing the data in `cachedData` so it's available when the chart is eventually initialized.

The observer is cleaned up in the `destroy()` method alongside the existing `ResizeObserver`.

## Steps to Reproduce

1. Place a chart widget inside a non-active tab
2. Load the page
3. Switch to the tab containing the chart
4. Chart is either empty or has wrong dimensions

## After Fix

1. Chart initializes correctly when its tab becomes active
2. Data updates received while the chart is hidden are applied when the tab becomes active

Fixes #16599

### DIFF
--- a/packages/widgets/resources/js/components/chart.js
+++ b/packages/widgets/resources/js/components/chart.js
@@ -21,16 +21,47 @@ export default function chart({ cachedData, maxHeight, options, type }) {
             this.initChart()
 
             this.$wire.$on('updateChartData', ({ data }) => {
+                cachedData = data
+
                 const chart = this.getChart()
 
                 if (!chart) {
+                    // Chart may not exist yet if the container was hidden
+                    // when updateChartData fired. It will be initialized
+                    // when the container becomes visible via the
+                    // IntersectionObserver below.
                     return
                 }
 
-                cachedData = data
                 chart.data = data
                 chart.update('resize')
             })
+
+            // Reinitialize the chart when its container becomes visible.
+            // Chart.js cannot render correctly in hidden containers (e.g.
+            // non-active tabs) because the canvas has zero dimensions.
+            this.intersectionObserver = new IntersectionObserver(
+                (entries) => {
+                    for (const entry of entries) {
+                        if (!entry.isIntersecting) {
+                            continue
+                        }
+
+                        const chart = this.getChart()
+
+                        if (chart) {
+                            // Chart exists but may have stale dimensions
+                            chart.resize()
+                        } else {
+                            // Chart was never initialized (container was
+                            // hidden on mount), initialize it now
+                            this.initChart()
+                        }
+                    }
+                },
+                { threshold: 0.1 },
+            )
+            this.intersectionObserver.observe(this.$el)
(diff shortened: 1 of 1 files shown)


danharrin:

PR #19417, opened 2026-03-04
feat: isolate dark mode preference per panel using prefixed localStorage key

jeffersongoncalves:
### Summary                                                   
																																																								 
- Prefix the localStorage theme key with the panel ID (theme-{panelId}) so each panel stores its dark mode preference independently
- When a project has multiple panels (e.g. admin and app) with different dark mode settings, changing the theme in one panel no longer overwrites the other
- Backwards compatible: reads from the legacy theme key as fallback, so existing users keep their preference after updating

### Problem

When a Laravel project has multiple Filament panels (e.g. admin with dark mode enabled and app with dark mode disabled), they all share a single localStorage key 'theme'. Navigating between panels causes one to overwrite the
other's preference, resulting in a broken dark mode experience.

### Solution

Each panel now uses theme-{panelId} as its localStorage key (e.g. theme-admin, theme-app). This is implemented across three files:

- base.blade.php: Exposes the panel ID via a --panel-id CSS custom property and prefixes all inline localStorage calls
- dark-mode.js: Reads --panel-id from CSS and uses it to build the storage key for all Alpine store operations
- theme-switcher/index.blade.php: Uses the prefixed key in x-data/x-init

All read operations include a fallback to the legacy 'theme' key, ensuring users who upgrade don't lose their existing preference.

### Test plan

- Create two panels (e.g. admin with ->darkMode() and app with ->darkMode(false))
- Verify switching theme in admin does not affect app
- Verify app always stays in light mode
- Verify a user with an existing theme key in localStorage (pre-update) has their preference migrated seamlessly
- Inspect localStorage and confirm keys are theme-admin / theme-app

### DIFF
--- a/packages/panels/resources/js/dark-mode.js
+++ b/packages/panels/resources/js/dark-mode.js
@@ -1,5 +1,13 @@
 document.addEventListener('alpine:init', () => {
+    const panelId = getComputedStyle(document.documentElement)
+        .getPropertyValue('--panel-id')
+        .trim()
+        .replace(/['"]/g, '')
+
+    const storageKey = panelId ? `theme-${panelId}` : 'theme'
+
     const theme =
+        localStorage.getItem(storageKey) ??
         localStorage.getItem('theme') ??
         getComputedStyle(document.documentElement).getPropertyValue(
             '--default-theme-mode',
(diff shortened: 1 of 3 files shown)


danharrin:
My opinion is that this is the desired experience and not broken. What do you think @People-Sea?

People-Sea:
@danharrin 
I agree with Dan. If I switch to dark mode late at night because the screen hurts my eyes, I don't wanna get blinded again when I open another panel lol

PR #19434, opened 2026-03-05
Add loading indicator SVG configuration

shaharyarhq:
Added configuration option for custom loading indicator SVG.

## Description
Previously, the loading indicator SVG was hardcoded and could not be 
customized without modifying the package source. This adds a 
`loading_indicator_svg` key to the config file, allowing users to 
define their own SVG icon for the loading indicator.

The `__attributes__` placeholder is injected into the SVG at runtime, 
so dynamic attributes (e.g. `wire:loading`, `class`) are still applied 
correctly regardless of which SVG is used.

## Visual changes
Before (Default Loader):

Screenshot 2026-03-04 180339

After (Lucide Loader):

Screenshot 2026-03-04 180421

## Functional changes

### DIFF
--- a/packages/support/config/filament.php
+++ b/packages/support/config/filament.php
@@ -117,4 +117,23 @@ return [
 
     'system_route_prefix' => 'filament',
 
+    /*
+    |--------------------------------------------------------------------------
+    | Custom Loading Indicator SVG
+    |--------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


danharrin:
Changes need to be in one PR #19435, as one should not be merged without the other

PR #19435, opened 2026-03-05
Refactor SVG handling for loading indicator

shaharyarhq:
Refactor loading indicator SVG handling to use a custom SVG if configured, otherwise use default SVG.

## Description
Refactored `generate_loading_indicator_html()` to support a custom SVG 
loading indicator via the `filament.loading_indicator_svg` config key.

Previously, the SVG was hardcoded inline. Now the function:
1. Builds the default SVG as before (no behaviour change if config is unset)
2. Checks `config('filament.loading_indicator_svg')` for a custom SVG string
3. If found, replaces the `__attributes__` placeholder with the resolved 
   `ComponentAttributeBag` attributes (classes, wire:loading, etc.) and 
   returns that instead

This keeps full backwards compatibility — existing apps with no config 
value set will see zero change in behaviour or appearance.

## Visual changes
If a user sets `loading_indicator_svg` in 
their `filament.php` config, their chosen SVG will render in place of 
the default spinner.
[See Visuals](https://github.com/filamentphp/filament/pull/19434)

## Functional changes

### DIFF
--- a/docs-assets/app/app/Filament/Clusters/Settings/Pages/ManageApi.php
+++ b/docs-assets/app/app/Filament/Clusters/Settings/Pages/ManageApi.php
@@ -3,6 +3,7 @@
 namespace App\Filament\Clusters\Settings\Pages;
 
 use App\Filament\Clusters\Settings\SettingsCluster;
+use BackedEnum;
 use Filament\Pages\Page;
 use Filament\Support\Icons\Heroicon;
 
@@ -14,7 +15,7 @@ class ManageApi extends Page
 
     protected static ?string $title = 'API';
 
-    protected static string | \BackedEnum | null $navigationIcon = Heroicon::OutlinedCodeBracket;
+    protected static string | BackedEnum | null $navigationIcon = Heroicon::OutlinedCodeBracket;
 
     protected static ?int $navigationSort = 8;
 }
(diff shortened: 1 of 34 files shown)


danharrin (changes requested):
I don't think that a config file is suitable for this sort of change to be honest. How about we ask the user to bind an Htmlable object to the container via a specific interface from Filament, that receives the attributes object in the constructor?

shaharyarhq:
> I don't think that a config file is suitable for this sort of change to be honest. How about we ask the user to bind an Htmlable object to the container via a specific interface from Filament, that receives the attributes object in the constructor?

Hi @danharrin, I've updated the implementation based on your feedback. Instead of a config string, the loading indicator is now resolved via Laravel's service container using a LoadingIndicator contract.

Users can override it in their AppServiceProvider:

```
<?php

use Filament\Support\Contracts\LoadingIndicator;
use App\Filament\Components\CustomLoadingIndicator;

public function register(): void
{
    $this->app->bind(LoadingIndicator::class, CustomLoadingIndicator::class);
}
```

Where CustomLoadingIndicator implements Filament\Support\Contracts\LoadingIndicator (which extends Htmlable) and receives the ComponentAttributeBag in its constructor.

Backwards compatible — default behaviour is unchanged. Let me know what you think!

PR #19439, opened 2026-03-06
Enhance documentation for action arguments in tables

RomainMazB:
Added information on executing table, header, and toolbar actions with `actionArguments` parameter.


danharrin:
Please reopen on 4.x

Please check your spelling

Please check that the example URL actually works for a table action, I think you need to pass context too

PR #19446, opened 2026-03-07
Fix crypto.randomUUID error in non-secure HTTP contexts

### LINKED ISSUE #19421: crypto.randomUUID() breaks error notifications in non-secure HTTP contexts
### Package

filament/notifications

### Package Version

v4.7.4

### Laravel Version

v12.53.0

### Livewire Version

v3.7.11

### PHP Version

PHP 8.4.1

### Problem description

PR #19332 replaced `uuid-browser` with the native `crypto.randomUUID()` API, which introduced a regression: 

`window.FilamentNotification` calls `crypto.randomUUID()` in its constructor, which is restricted to [secure contexts](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID)

When Filament is served in a non-secure context with `APP_DEBUG=false,` the built-in error notification silently crashes with:

> Uncaught (in promise) TypeError: crypto.randomUUID is not a function

Resulting in the user receiving no feedback when a Livewire component throws an error.

### Expected behavior

The error notification should appear regardless of whether the app is served over HTTPS. 

Two approaches I see working here:

1. Replace with [uuid ](https://www.npmjs.com/package/uuid) which uses crypto.getRandomValues() internally. This would replace the deprecated `uuid-browser` with a maintained alternative. 
2. When `crypto.randomUUID()` isn't available fall back to a manual implementation using crypto.getRandomValues():  

```js
const uuid = crypto.randomUUID?.() ?? ([1e7]+-1e3+-4e3+-8e3+-1e11).replace(/[018]/g, c =>
    (c ^ crypto.getRandomValues(new Uint8Array(1))[0] & 15 >> c / 4).toString(16)
);
```

If either of these are acceptable I'm happy to open a PR with a fix.

### Steps to reproduce

The attached reproduction repository has a README, but basically just serve a fresh v4.x Filament installation over HTTP on a local network IP, .test/.local hostname without certs or any non-HTTPS setup, make sure `APP_DEBUG=false` and get a Livewire component to throw an error. The error described above should appear in the console,

https://github.com/fvfvfvfv/filament-pr-19332-repro

### Relevant log output

```shell

```

04cb:
Fixes #19421. The native crypto.randomUUID() API is only available in secure contexts (HTTPS), causing error notifications to crash in non-secure HTTP contexts. This change adds a fallback using crypto.getRandomValues() which works in both secure and non-secure contexts.

### DIFF
--- a/packages/notifications/resources/js/Notification.js
+++ b/packages/notifications/resources/js/Notification.js
@@ -1,6 +1,17 @@
+const generateUuid = () => {
+    if (typeof crypto.randomUUID === 'function') {
+        return crypto.randomUUID()
+    }
+
+    // Fallback for non-secure contexts (HTTP) using crypto.getRandomValues()
+    return ([1e7] + -1e3 + -4e3 + -8e3 + -1e11).replace(/[018]/g, (c) =>
+        (c ^ crypto.getRandomValues(new Uint8Array(1))[0] & 15 >> c / 4).toString(16)
+    )
+}
+
 class Notification {
     constructor() {
-        this.id(crypto.randomUUID())
+        this.id(generateUuid())
 
         return this
     }


danharrin:
Clearly generated by some agent and not tested

PR #19467, opened 2026-03-09
fix: grouped buttons divider for RTL

thaqebon:
Fix divider shadow direction in grouped buttons for RTL layouts.
## Visual changes
Before
rtl-grouped-bottons1
After
rtl-grouped-bottons2

## Functional changes


danharrin (changes requested):
I would suggesting removing the `ltr:` prefixes and relying on those by default, so we don't change the specificity of these classes.

Then, please recompile the assets using `npm run build`.

danharrin (approved):

PR #19474, opened 2026-03-11
Add filament scripts attributes

helturkey:
## What does this PR do?

Adds an `attributes` parameter to the `@filamentScripts` Blade directive and `renderScripts()`, allowing arbitrary HTML attributes to be applied to every rendered `<script>` tag including the `window.filamentData` block.

## Usage

```blade
@filamentScripts(attributes: ['data-cfasync' => 'false'])
```

Or globally via `config/filament.php`:

```php
'scripts' => [
    'attributes' => [
        // 'data-cfasync' => 'false',
    ],
],
```

## Usage Case?

As an Example Cloudflare Rocket Loader defers JavaScript execution by default, which breaks Filament forms in blade, While admin panel could be filtered or bypassed using cloudflare rules, it may be difficult to bypass every single uri, you are using filament form in. Cloudflare requires `data-cfasync="false"` on **every** `<script>` tag to opt out. Without this, users had to override the entire `@filamentScripts` directive manually.

## Changes

I tried to make the minimal impacts or changes possible, so the changes are:
- `AssetManager::renderScripts()` now accepts `array $attributes = []`, merged with `config('filament.scripts.attributes')` (inline wins on conflict)
- `Js::getExtraAttributesHtml()` supports integer-keyed (valueless) attributes e.g. `['async']` renders as `async` not `0="async"`
- Attributes are applied to both `<script src=...>` tags and the `window.filamentData` inline block

## Tests

```bash
./vendor/bin/pest tests/src/Support/BladeDirectives/FilamentScriptsAttributesTest.php
```

### DIFF
--- a/packages/support/config/filament.php
+++ b/packages/support/config/filament.php
@@ -59,6 +59,22 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Script Attributes
+    |--------------------------------------------------------------------------
+    | This is an array of attributes that will be added to all script tags
+    | rendered by Filament. You can use this to add attributes like `data-cfasync` or `async` to
+    | all scripts, or to add custom data attributes that your application may require.
+    |
+    */
+
+    'scripts' => [
+        'attributes' => [
+            // 'data-cfasync' => 'false',
+        ],
+    ],
+
+    /*
+    |--------------------------------------------------------------------------
     | Cache Path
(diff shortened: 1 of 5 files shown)


danharrin (changes requested):
Please rebase this on 4.x and reopen. Please refactor to use a ComponentAttributeBag object to render and pass attributes between a class and view.

PR #19475, opened 2026-03-11
fix: Remove redundant window resize listener from chart widget

### LINKED ISSUE #19463: Widgets do not scale correctly when the container width changes
### Package

filament/widgets

### Package Version

v5.3.2

### Laravel Version

v12.53.0

### Livewire Version

v4.2.1

### PHP Version

8.5.3

### Problem description

Widgets do not scale correctly when the container width changes. I’m not referring to the window width - in that case everything works correctly. However, when only the container width changes, the widgets resize in a random way.

I know this is quite an unusual case, but I’m working on a slide-out chat plugin (similar to Shopify) that does not rely on a slideOver, but instead uses a permanently visible element next to the main content.

During testing, I noticed that widgets behave this way:

https://github.com/user-attachments/assets/1c659015-76cf-4184-a8a2-088203a4558a

The same issue occurs in Filament v4.

### Expected behavior

Widgets should scale as if the browser width had changed.

### Steps to reproduce

1. Clone repo
2. Click chat icon in header
3. Look at the widgets

https://github.com/webard-playground/filament-widget-resize-bug

### Relevant log output

```shell

```

people-sea:
### Fixes: #19463 

The chart component registered both a window resize listener and a ResizeObserver sharing the same debounced destroy/reinit handler. When the container width changes (e.g. the
  chat side panel in the reproduction repo), one chart's ResizeObserver callback gets randomly swallowed during the debounce, causing one chart to end up taller than the other.   
                                                                                                                                                                                   
  Removed the redundant window resize listener --- ResizeObserver already covers all cases including window resize.

### Before

https://github.com/user-attachments/assets/7e0d3af1-a6e7-4b09-9e0c-d0b5b090585d

### After

https://github.com/user-attachments/assets/066744b1-5769-40c1-a895-e692108d9d9d

## Functional changes


webard:
Thank you so much!

danharrin (approved):

PR #19511, opened 2026-03-16
add tabindex to login form

wietsewarendorff:
## Description

Add tabindex to login-form fields for easier form-usage with keyboard. this especially happens when "forget password" is enabled, then tab jumps from "e-mailaddress" to "forgot password". this fix  makes it jump to "password".

Also mentioned here: https://github.com/filamentphp/filament/discussions/7264

## Visual changes

none

## Functional changes

### DIFF
--- a/packages/panels/src/Auth/Pages/Login.php
+++ b/packages/panels/src/Auth/Pages/Login.php
@@ -248,7 +248,6 @@ class Login extends SimplePage
             ->email()
             ->required()
             ->autocomplete()
-            ->extraInputAttributes(['tabindex' => 1])
             ->autofocus();
     }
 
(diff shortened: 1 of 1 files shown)


wietsewarendorff:
looks as if using tabindex is not allowed at all... if so, let's forget this and close it

danharrin:
Since it fails the a11y testing, maybe do the opposite, and mark the elements that should be removed from the tab order with index -1?

wietsewarendorff:
that works, a negative index on the "Forgot password"-link makes the tab nicely jump from the e-mailaddress input-field to the password input-field

danharrin (approved):

PR #19517, opened 2026-03-16
Introduce ToolbarButton enum

ericmp33:
## Description

It adds the ability to use enum cases instead of hardcoded strings when working with toolbar buttons:
```php
RichEditor::make('txt')
    ->toolbarButtons([
        [ToolbarButton::Bold, ToolbarButton::Italic, ToolbarButton::Underline],
        [ToolbarButton::AttachFiles],
    ])
```

## Visual changes

\-

## Functional changes

### DIFF
--- a/packages/forms/docs/10-rich-editor.md
+++ b/packages/forms/docs/10-rich-editor.md
@@ -70,6 +70,19 @@ RichEditor::make('content')
 
 Each nested array in the main array represents a group of buttons in the toolbar.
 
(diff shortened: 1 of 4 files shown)


danharrin:
Since custom toolbar buttons are also supported, I don't really see the value in this to be honest aside from a little autocomplete for the default buttons. For custom buttons you would still need a string, so I don't think it's worthwhile to introduce.

PR #19531, opened 2026-03-18
Check for HasManyDeep in applySearchConstraint method

cheesegrits:
## Description

Added check for Staudenmeir's HasManyDeep in applySearchConstraints, as it needs to be treated the same way as BelongsToThrough, derefencing the actual target table name for passing to getJsonSafeColumnName.

I've tested this with a couple of uses of HasManyDeep, and I think it should "just work" regardless of depth.

Without this change, any use of a HasManyDeep column search just blows up, as it treats the dotted relation as a JSON path.

Not sure if I need to do the class_exists() check, and/or I should inline the namespace rather than a 'use'.

## Visual changes

None

## Functional changes


danharrin (changes requested):
I think the `class_exists()` check can be skipped, if an object is an instance of it then it definitely exists

danharrin (approved):

PR #19538, opened 2026-03-19
Chore: Openspout v5

webard:
This PR allows openspout v5.

One breaking change detected is that `Row::fromValues` has been replaced by `Row::fromValuesWithStyle`.

## Functional changes


danharrin:
Please reopen this once we have a Filament v6 branch in the works. We can't make this change because it would be breaking to anyone already using `Row::fromValues()` in Filament, which is in our docs, as they would get an automatic upgrade to the next OpenSpout version if they do not have it required explicitly in their `composer.json`.

PR #19539, opened 2026-03-19
Chore: Allow chillerlan/php-qrcode v6

webard:
Actually, I couldn’t find where this library is used, but this PR adds support for chillerlan/php-qrcode v6.

## Functional changes

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
         "ext-intl": "*"
     },
     "require-dev": {
-        "chillerlan/php-qrcode": "^5.0",
+        "chillerlan/php-qrcode": "^5.0|^6.0",
         "filament/support": "*",
         "larastan/larastan": "^3.0",
         "laravel/pint": "1.25.1",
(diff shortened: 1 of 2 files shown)


rickdubiel:
Why was this closed and not merged or have any action taken?

Luca-Lin:
Hi @webard —  I went digging and thought it might help to share what I found.

Filament doesn't reference any `chillerlan\…` class directly. The dependency is used transitively through the App-based MFA flow:

`Filament\Auth\MultiFactor\App\AppAuthentication::generateQrCodeDataUri()` calls `PragmaRX\Google2FAQRCode\Google2FA::getQRCodeInline()`, and `pragmarx/google2fa-qrcode` picks chillerlan as its rendering backend (its `QRCode/Chillerlan.php` adapter). So `chillerlan/php-qrcode` in Filament's `composer.json` is effectively just pinning that backend — there's no direct API usage to break.

Worth noting that the middle layer already allows both majors: `pragmarx/google2fa-qrcode` requires `chillerlan/php-qrcode: ^5.0|^6.0`, and its Chillerlan adapter is written against the version-agnostic `outputInterface => QRMarkupSVG::class` API (not the `outputType` constant that v6 removed). PHP floors line up too — v6 needs `^8.2`, same as Filament.

Given that google2fa-qrcode has already relaxed to `^5.0|^6.0`, it seems like Filament could safely do the same. Just a suggestion in case it helps move this along.

PR #19545, opened 2026-03-21
Add `$short` parameter to `since()` and `sinceTooltip()`

shaheenfawzy:
## Description

Adds a `$short` parameter to `since()` and `sinceTooltip()` on both `TextColumn` and `TextEntry`, enabling abbreviated relative timestamps via Carbon's `diffForHumans(short: true)` (e.g., "1h ago" instead of "1 hour ago").

```php
TextColumn::make('created_at')->since(short: true)
```

### Note on tests

There are no existing tests for `since()`, `sinceTooltip()`, or any of the date formatting methods in `CanFormatState`. Happy to add test coverage for the new `$short` parameter (and the existing methods) if desired.

## Visual changes

None

## Functional changes

### DIFF
--- a/packages/infolists/docs/02-text-entry.md
+++ b/packages/infolists/docs/02-text-entry.md
@@ -207,6 +207,15 @@ TextEntry::make('created_at')
     ->since()
 ```
 
+You may pass `short: true` to use abbreviated units (e.g., "1h ago" instead of "1 hour ago"):
+
+```php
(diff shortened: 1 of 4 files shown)


danharrin:
Thanks, I think I am going to pass on this for now though. I haven't really heard of this request before so I think it is pretty niche, and is quite easily achievable if you add `formatStateUsing()` to the column instead of `since()` with your own function.

PR #19547, opened 2026-03-22
fix: Only use x-show on logo when sidebar is collapsible

pxlrbt:
## Description

When switching to `->topbar(false)` the localStorage entry for `isOpen` might be false which leads to the logo being hidden in the sidebar. There is no way to change that without clearing users localStorage so we only should check for `isOpen` when the sidebar is actually collapsible.

## Visual changes

grafik

## Functional changes


danharrin (changes requested):
Sometimes conditional HTML tags can cause issues with prettier, please wrap the actual `x-show` only in the `@if` please

danharrin (approved):

pxlrbt:
Scusi, forgot about this one 🙈

PR #19557, opened 2026-03-23
fix: fall back to arguments() when invokedArguments is null in getJsClickHandler

ManukMinasyan:
## Description

When `getJsClickHandler()` was updated to use `getInvokedArguments()` (the property set by `__invoke()`), the fallback was changed from `$this->getArguments()` to `[]`. This means arguments set via the `arguments()` method are silently dropped from the generated `wire:click` handler — even though the action reports them correctly via `getArguments()`.

This breaks ActionGroup children used as repeater item actions. These actions cannot use `__invoke()` (which clones the action) because the cloned instances aren't connected to the ActionGroup's rendering pipeline. They must use `arguments()` directly, but the current code ignores that property when generating the JavaScript click handler.

**Regression from v3.x** where `getJsClickHandler()` used `$this->getArguments()` directly.

### Reproduction

1. Create a Repeater with `labeledItemActions` containing an `ActionGroup`
2. Inside the ActionGroup, add actions that need `$arguments['item']` (the repeater item key)
3. In the Blade view, set arguments on the flat actions via `$flatAction->arguments(['item' => $itemKey])`
4. Click an action inside the dropdown — `$arguments` is always `[]` in the callback

### Fix

Fall back to `$this->getArguments()` when `invokedArguments` is null, preserving the `invokedArguments` priority when present:

```diff
- if (count($arguments = $this->getInvokedArguments() ?? [])) {
+ if (count($arguments = $this->getInvokedArguments() ?? $this->getArguments())) {
```

## Visual changes

No visual changes.

## Functional changes

### DIFF
--- a/docs-assets/app/app/Filament/Clusters/Settings/Pages/ManageApi.php
+++ b/docs-assets/app/app/Filament/Clusters/Settings/Pages/ManageApi.php
@@ -3,6 +3,7 @@
 namespace App\Filament\Clusters\Settings\Pages;
 
 use App\Filament\Clusters\Settings\SettingsCluster;
+use BackedEnum;
 use Filament\Pages\Page;
 use Filament\Support\Icons\Heroicon;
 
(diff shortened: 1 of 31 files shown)


ManukMinasyan:
Closing - found a way to properly use invokedArguments without requiring a core change. The fix clones the ActionGroup per repeater item and invokes child actions via __invoke() to set invokedArguments correctly.

PR #19559, opened 2026-03-24
feat: expose TipTap/ProseMirror globals for custom rich editor extensions

leek:
## Problem

Custom extensions loaded via `RichContentPlugin::getTipTapJsExtensions()` are fetched at runtime via dynamic `import()`. Each extension bundle must currently include its own copy of TipTap/ProseMirror because there's no way to reference the copy already bundled in `rich-editor.js`. This leads to:

- **Multiple ProseMirror instances** on the same page
- **`instanceof` failures** across bundles (e.g. `DecorationSet`, `Node`, `Mark`) — requiring monkey-patches to work around
- **~150-200 KB of redundant JS per custom extension** (each re-bundles `@tiptap/core`, `prosemirror-state`, etc.)

This is particularly painful for Tiptap Pro extensions (Pages, TableKit) which import heavily from `@tiptap/core` and `@tiptap/pm/*`.

## Solution

Expose `window.Filament.TipTap` with the core ProseMirror modules that are already bundled in `rich-editor.js`:

```js
window.Filament.TipTap = {
    'core': /* @tiptap/core */,
    'pm/state': /* @tiptap/pm/state */,
    'pm/view': /* @tiptap/pm/view */,
    'pm/model': /* @tiptap/pm/model */,
}
```

Custom extension builds can then mark `@tiptap/*` as external and resolve from this global at runtime, sharing the exact same ProseMirror instance as the host editor.

### Example: esbuild plugin for custom extensions

```js
// In your project's build script
const tiptapSharedPlugin = {
    name: 'tiptap-shared',
    setup(build) {
        build.onResolve({ filter: /^@tiptap\// }, args => ({
            path: args.path,
            namespace: 'tiptap-shared',
        }))

        build.onLoad({ filter: /.*/, namespace: 'tiptap-shared' }, async (args) => {
            const key = args.path.replace('@tiptap/', '')
            const realModule = await import(args.path)
            const namedExports = Object.keys(realModule)
                .filter(k => k !== '__esModule' && k !== 'default')

            let code = `const __m = window.Filament.TipTap[${JSON.stringify(key)}];\n`
            if (namedExports.length) {
                code += `export const { ${namedExports.join(', ')} } = __m;\n`
            }
            code += `export default __m?.default ?? __m;\n`
            return { contents: code, loader: 'js' }
        })
    },
}
```

## Changes

- **`packages/forms/resources/js/components/rich-editor-entry.js`** — new thin wrapper that exposes `window.Filament.TipTap` and re-exports the existing Alpine component
- **`bin/build.js`** — points the rich-editor build at the wrapper entry point
- **`packages/forms/dist/components/rich-editor.js`** — rebuilt (bundle size unchanged — the modules were already included, we're just assigning them to `window`)

## Non-breaking

Existing custom extensions that bundle their own TipTap continue to work unchanged. This simply gives extension authors the *option* to externalize `@tiptap/*` and eliminate duplicate ProseMirror instances.


awcodes:
I think this is a solid addition to core and solves the issues I have run into too trying to offer syntax highlighting in code blocks where externally it creates pm version conflicts when the editor plugin is executed at run time.

I do think it needs some documentation though.

danharrin (commented):
@leek please clarify why you chose the extra wrapper instead of modifying the main rich editor JS file. I am refactoring that away, so please let me know if there was a reason

danharrin (approved):

leek:
> @leek please clarify why you chose the extra wrapper instead of modifying the main rich editor JS file. I am refactoring that away, so please let me know if there was a reason

@danharrin No hard reason - purely a choice. Wanted to isolate the `window` assignment as a bundle-entry concern so `rich-editor.js` stayed a pure component module, no global side effects on import. Your refactor is safe 👍

PR #19565, opened 2026-03-25
fix: parent modal re opens for brief moment

### LINKED ISSUE #19561: overlayParentActions feature re-opens parent modal after close
### Package

filament/filament

### Package Version

v4.7

### Laravel Version

12

### Livewire Version

v3

### PHP Version

8.4

### Problem description

When using `overlayParentActions` for actions, closing a child modal and parent modal at once by pressing Escape twice closes both the modals which is fine. However, there is a flicker, it re-opens the parent modal for a fraction of second and then closes it. This happens quite often in large or complex data driven modals, but it can be seen in simple modals by throttling the network to slow 4g in browser console networks tab.

Here is a video showing the problem:

https://github.com/user-attachments/assets/11c3ab7a-0efd-4783-a510-1281448dffd3

### Expected behavior

It should close both the modals without re-opening them again.

### Steps to reproduce

1. Install the repo by following regular Laravel installation guide
2. Open `/` route and login using email: test@example.com and password: password
3. Open browser console, and set network throttling to 4g Slow inside networks tab.
4. Click On Modal Test page link in the sidebar navigation
5. Click on Test action button and then click on Open Child Modal action button.
6. Hit Escape twice quickly and it would close both the Modals but for a fraction of a second the parent modal would re-open and then close again.

https://github.com/mansoorkhan96/filament-modal-bug

### Relevant log output

```shell

```

mansoorkhan96:
Fixes the issue: https://github.com/filamentphp/filament/issues/19561

## What the original problem is caused by

When `overlayParentActions()` is used, both parent and child modals are open at the same time. Pressing Escape twice quickly creates a race condition between the local Alpine.js state and the server side Livewire state.

Here is what happens:

- The first Escape closes the child modal instantly on the client side and sends a livewire request to unmount it on server.
- The second Escape fires before the server responds. It closes the parent modal on the client side and queues another Livewire unmount request.
- At this point, both modals appear closed to the user. However, the server has not processed either request yet.
- When the server responds to the first request (child unmount), the parent action is still in the mounted actions stack. So `syncActionModals()` tells the client to open the parent modal. Since the parent's `isOpen` is already `false` (the user closed it), the open guard passes and the parent modal briefly flickers back open.
- When the server responds to the second request , the parent modal closes again.

The root cause is that `syncActionModals()` in `modals.js` unconditionally calls `openModal()` when reverting to the parent's nesting index, with no awareness that the user already closed that modal locally while the Livewire request was in-flight.

## How the solution works

A `dismissedModals` Set is added to the `filamentActionModals` Alpine component. It listens for `modal-closed` events comig up from child modal elements. The `modal-closed` event only fires from the `close()` method (user initiated via Escape, click-away, or close button) and never from `closeQuietly()` (programmatic closed or synced). This distinction is key.

When `syncActionModals()` is about to open a modal, it checks if that modal's ID is in the `dismissedModals`. If that's true, it skips the `openModal()` call and removes the entry. The pending Livewire request will handle the server side unmount on its own. The Set is cleared when nesting reaches `null` (all actions unmounted).

## Why this solution over alternatives

- **Tried Modal level flag (`isUserClosed` in `modal.js`)**: Added a flag directly on each modal Alpine component, but it breaks non-action modals. If a user closes a standalone trigger based modal with Escape and then clicks the trigger button again, the flag would block reopening since the `open-modal` handler can't distinguish "open from trigger" vs "open from sync".

- **Tried Server side fix**: Not viable. The PHP `syncActionModals()` runs inside `unmountAction()` and has no knowledge of local Alpine state. Livewire also processes the two `$wire.unmountAction(false)` calls sequentially (not batched), so the intermediate state that causes the flicker is unavoidable on server side.

So, the `dismissedModals` Set approach is minimal, handles arbitrary nesting depths,and relies on an existing event distinction (`close()` vs `closeQuietly()`) that naturally separates user-initiated dismissals vs programmatic closes.

https://github.com/user-attachments/assets/832fd17d-9726-41f8-8d34-1cccf5439524


danharrin (changes requested):
@mansoorkhan96 in the PR description, please provide a technical explanation of what the original problem is caused by, how the solution works and why you chose this over others

mansoorkhan96:
> @mansoorkhan96 in the PR description, please provide a technical explanation of what the original problem is caused by, how the solution works and why you chose this over others

Updated the PR description.

danharrin:
Reopening child modals now takes two attempts after they are closed, I don't think this is the right solution, sorry. It is too heavy handed.

https://github.com/user-attachments/assets/a668af6d-52cb-405d-9378-f0d632a38794

People-Sea:
https://github.com/user-attachments/assets/0c99fa9c-c297-448d-9ef7-5de3154ad40a

I think we can consider a behavioral change in the next major release: specifically, having the frontend close the modal only after a successful uninstallation. This would naturally solve the flickering. Currently, it's just a UX issue—the final outcome is correct, but the process feels unpolished. What do you think? @danharrin

danharrin:
@People-Sea I think I would like to avoid that as we know that the user's intention is 100% to close the modal, so the "optimistic UI" approach would be to immediately close it rather than wait for a server response. The server response will never indicate that the modal should remain open, you know?

People-Sea:
> @People-Sea I think I would like to avoid that as we know that the user's intention is 100% to close the modal, so the "optimistic UI" approach would be to immediately close it rather than wait for a server response. The server response will never indicate that the modal should remain open, you know?

Thanks, I see your point. I agree, and I'll try a different approach.

PR #19571, opened 2026-03-26
i18n: make RichEditor text color labels translatable

abournavaud:
## Description

Currently, `TextColor::getDefaults()` uses `Str::ucwords($name)` to generate color labels (e.g. "Red", "Slate"), which are hardcoded in English. This makes the color dropdown in the RichEditor unusable for non-English locales.

This PR replaces the hardcoded labels with Laravel's `__()` translation helper, allowing color names to be translated via language files. English and French translations are included. Other locales will fall back to the English translation by default.

- `TextColor::getDefaults()` now uses `__('filament-forms::components.rich_editor.actions.text_color.modal.form.color.options.{name}')` instead of `Str::ucwords($name)`
- Added color name translations in `lang/en/components.php` (26 colors)
- Added color name translations in `lang/fr/components.php` (26 colors + full `text_color` action block)

## Visual changes

No visual changes for English users. French users will now see translated color names in the text color dropdown.

## Functional changes


danharrin (changes requested):
Please fix the merge conflicts

Please restore the previous behaviour of auto-generating an option label from the color name if the translation is missing from this list. You could fetch the entire list of options first in an array, and then check if the key exists before using the translation.

danharrin (approved):

PR #19573, opened 2026-03-27
fix: hide table pagination for empty records

ahmed-rashad-alnaggar:
Hi,
Pagination is shown even when there are no records if the table is not using the default pagination mode or when the paginator is not a `LengthAwarePaginator`.
This PR uses `\Illuminate\Contracts\Pagination\Paginator` or `\Illuminate\Contracts\Pagination\CursorPaginator` `isNotEmpty()` function instead of the old behavior.

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -125,7 +125,7 @@
     $secondLevelHeadingTag = $heading ? $getHeadingTag(1) : $headingTag;
     $pluralModelLabel = $getPluralModelLabel();
     $records = $isLoaded ? $getRecords() : null;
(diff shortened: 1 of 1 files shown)


danharrin:
@ahmed-rashad-alnaggar I have rewritten the line to be more specific and ensure that for `LengthAwarePaginator`, the paginator UI is shown if there are records on other pages, in case the user visits a page without records. Can you check that the new code still fixes your original problem please?

danharrin (approved):

ahmed-rashad-alnaggar:
@danharrin It works, but I do not understand this part:

> ensure that for `LengthAwarePaginator`, the paginator UI is shown if there are records on other pages, in case the user visits a page without records.

What is the page without records? Isn't it (the pagination UI) should be hidden if there is no items in the records collection?
I have searched through the internal `Paginator` and `LengthAwarePaginator` to see if I miss something but I couldn't figure it out.

Thank you for your concern!

danharrin:
There may not be records on the current page, but there are records on a previous page, so you need UI to switch back to another page. Thats the difference between the total check and empty check, the total check checks all pages, the empty check just checks the current page.

PR #19579, opened 2026-03-29
feat: add chart widget empty state

ahmed-rashad-alnaggar:
@danharrin as discussed in #13311 I have added a feature to display an empty state on chart widgets.

This is a heavy re-work of #13408, solves its issues and compatible with new Filament v4 and v5 code style.

Please consider merging this PR as it make `TableWidget` and `ChartWidget` looks the same when there is no data to display, which feels consistent and professional.


ahmed-rashad-alnaggar:
Hi @danharrin,
I noticed a latency of displaying the chart on a slow network devices, this is due to the async loading of the chart.js AFTER going from empty to non-empty data at first page load, like if the default filters have no data then adjusting them to a setting that has data.
I can't think of many suitable solutions but the one in my mind right now is:
instead of conditionally rendering the chart using `@if` directive is to just render it with a conditional style directive on a wrapper `div` to show and hide it, so do you have a better one or should I just implement it?

ahmed-rashad-alnaggar:
This PR is opened as a draft for now, as it depends on the outcome of another related PR. I’ll keep this in draft state until that one is reviewed and its direction (merge or not) is decided.

Once that is clarified, I’ll update or finalize this PR accordingly.

ahmed-rashad-alnaggar:
@danharrin ready to review!

ahmed-rashad-alnaggar:
@danharrin Is there any changes required?

danharrin:
I will review this when I next prepare a minor release as it is a new feature, currently I am busy with patch releases.

danharrin (approved):

PR #19585, opened 2026-03-29
feat: `ActionSeparator` for visual dividers between actions

leek:
image

## Summary

> _@danharrin I realize this may not be something you want to include in Filament, but I wanted to submit it as a PR before trying to release it bundled in a plugin or something._

Adds `ActionSeparator`, a lightweight component that renders a visual divider between actions. Works in page header actions, table header actions, and within action groups — adapting its appearance to the context:

- **Inline (header actions, button groups):** Renders a vertical line
- **Dropdown menus:** Splits items into separate `fi-dropdown-list` sections, using the native Filament divider pattern

### Usage

```php
use Filament\Actions\ActionSeparator;

// Page header actions
{
    return [
        ActionGroup::make([...])
            ->icon('heroicon-o-cog-6-tooth')
            ->color('gray'),
        ActionSeparator::make(),
        Action::make('create')
            ->label('New Record'),
    ];
}

// Inside a dropdown — creates a visual section break
ActionGroup::make([
    Action::make('edit')->label('Edit'),
    Action::make('duplicate')->label('Duplicate'),
    ActionSeparator::make(),
    Action::make('delete')->label('Delete')->color('danger'),
])
```

### Implementation

- `ActionSeparator` extends `ViewComponent` (same base as `Action` and `ActionGroup`)
- Uses `BelongsToGroup` and `CanBeHidden` concerns for visibility control and group nesting
- Overrides `isHiddenInGroup()` to skip authorization checks (not applicable to separators)
- Renders as `<div class="fi-ac-separator">` with a CSS hook
- In dropdown `ActionGroup`, splits the action list into separate `fi-dropdown-list` sections (same mechanism that nested `ActionGroup` uses)
- `ActionGroup` type unions updated to accept `ActionSeparator` in `actions()`, `getActions()`, `toEmbeddedHtml()`, and `cloneActions()`
- Separators are excluded from `flatActions` since they are visual-only

### DIFF
--- a/packages/actions/docs/03-grouping-actions.md
+++ b/packages/actions/docs/03-grouping-actions.md
@@ -99,7 +99,32 @@ ActionGroup::make([
 
 ## Adding dividers between actions
 
-You may add dividers between groups of actions by using nested `ActionGroup` objects:
+You may add visual dividers between actions using `ActionSeparator`. It adapts to its context automatically — rendering as a vertical line between inline actions, or splitting a dropdown into separate sections:
+
+```php
+use Filament\Actions\Action;
+use Filament\Actions\ActionGroup;
+use Filament\Actions\ActionSeparator;
+
+// Between inline actions (e.g., page header actions)
+[
+    ActionGroup::make([
+        // Array of actions
+    ]),
+    ActionSeparator::make(),
+    Action::make('create'),
+]
+
+// Inside a dropdown — creates a visual section break
+ActionGroup::make([
+    Action::make('view'),
+    Action::make('edit'),
+    ActionSeparator::make(),
+    Action::make('delete'),
+])
+```
+
+Alternatively, you may add dividers inside a dropdown by using nested `ActionGroup` objects:
 
 ```php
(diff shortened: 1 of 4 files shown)


archilex:
@leek This is already supported in Filament: https://filamentphp.com/docs/5.x/actions/grouping-actions#adding-dividers-between-actions

danharrin:
Yeah, given the existing syntax in place, I would accept this feature for separating actions without a dropdown, but the syntax would be very different:

```php
use Filament\Actions\ActionGroup;

ActionGroup::make([
    ActionGroup::make([
        // Array of actions
    ])->dropdown(false),
    ActionGroup::make([
        // Array of actions
    ])->dropdown(false),
])->dropdown(false), // Pass this to the outer group too to add separators, maybe?
```

leek:
> @leek This is already supported in Filament: https://filamentphp.com/docs/5.x/actions/grouping-actions#adding-dividers-between-actions

This PR is more about the vertical one but always something new to learn!

PR #19592, opened 2026-03-30
[4.x] Fix all-table summaries

ajnsn:
## Description

When using ->summaries(pageCondition: false, allTableCondition: false) to show only group summaries ([Link](https://filamentphp.com/docs/4.x/tables/summaries#hiding-summary-rows)) group summary rows were incorrectly hidden because $hasSummary short-circuited to false.

Removed the redundant page/all-table condition check from $hasSummary in index.blade.php, since components/summary/index.blade.php already gates those independently.

## Visual changes

n/a

## Functional changes


danharrin (changes requested):
I think the summary around line 1134 of index.blade.php needs adjusting, as surely now the `<table>` wrapper is going to render even if there is no page or all table summary?

ajnsn:
Good catch, should be fixed now.

PR #19593, opened 2026-03-30
[4.x] Fix last group summary

ajnsn:
## Description

Fix grouped table summaries at pagination boundaries by rendering the trailing summary when the next page starts a different group.

The previous hasMorePages() check was too broad and could drop the final summary for the last group on a page.

## Visual changes

n/a

## Functional changes


packages/tables/src/Concerns/CanSummarizeRecords.php
```diff
@@ -4,13 +4,53 @@
...
+        }
+
+        $query = $this->getFilteredSortedTableQuery();
+
+        if ($query === null) {
+            return true;
+        }
+
+        $nextPageFirstRecord = (clone $query)
+            ->skip($records->currentPage() * $records->perPage())
```
danharrin: Just checking, has this been verified to work with cursor pagination? I haven't tested it myself, but I assume that cursor pagination does not track a numeric page number like this.

danharrin (changes requested):

danharrin (approved):

PR #19598, opened 2026-03-31
Fix dashboard route overwritten by home redirect when using tenant domain binding field

People-Sea:
## Description
   When `tenantDomain()` uses a binding field like `{tenant:domain}`, the root route collision check did not normalize the domain portion of the route key.

  Since Laravel stores routes by normalized domain and URI, this caused the dashboard route to be missed and the `home` redirect to be registered as well.

  This updates the check to use `RouteUri::parse()` for both the domain and URI so it matches Laravel's route normalization.

## Functional changes


People-Sea:
Follow-up to #19584

danharrin (approved):

PR #19604, opened 2026-03-31
Fix: set hasModal in slideOver

webard:
## Description

Currently, using `slideOver()` without `modal()` has no effect. If this is intentional behavior, I understand, but it doesn’t seem to make much sense and can be confusing.

This PR changes this behavior by enabling the modal automatically when `slideOver()` is called.

## Visual changes

None

## Functional changes


danharrin:
You don't need to manually set `modal()`, by default the modal mode will be based on if there is any modal content like a modal header/description/form. Merging this change would be breaking for some niche use cases where `slideOver()` is part of the global configuration for actions, and it would cause actions that previously didn't open modals to start doing it.

PR #19652, opened 2026-04-07
refactor: chart widget chart.js implementation

ahmed-rashad-alnaggar:
Hi, this PR is focused on improving performance of the chart.js implementation by not destroying and re-initializing the chart on every event, and also improve animating chart by removing the default `animation.duration = 0` as this should not be the default case as the charts should animate to introduce a beautiful data visualization, handled missing or miss configured parts while initializing the chart, for example in the discussion #19609 with @danharrin I understood that the desired behavior is to rely on `ResizeObserver` not chart responsiveness options, but actually the old implementation was still relies on it by not setting `options.responsive` to `false` so the chart was resized and calculations are done by many "providers".
English is not my native language so i used AI to generate a well detailed description of the changes, rather than that I spent hours developing (only the `updateChartData` function is AI driven) and testing the code with all the possibilities to guarantee it is working (going from no data to data and vice versa with applying different cases like resizing). now the chart is animated when initialized for the first time or by updating data (not by a reset ugly animation but by a beautiful update and data changed animation). also the performance has significantly improved.

START OF AI DESCRIPTION

### Overview

This PR refactors the Filament Chart Widget's Alpine.js component and its companion Blade template. The primary goals are:

- Eliminate unnecessary `Chart` destroy/recreate cycles on resize and theme changes, replacing them with targeted, in-place updates.
- Extract cohesive methods (`updateChartData`, `updateChartTheme`, `getChartFallbackColors`, `whenChart`) to reduce duplication and improve readability.
- Stop mutating global `Chart.defaults` for per-instance concerns (colors, font), scoping those values to individual chart options instead.
- Remove the `maxHeight` prop from the JS component API, detecting it from the canvas element's own style at runtime.
- Use `@style` directive to apply `width`, `height` and `max-height` css properties so the chart can animate properly.

END OF AI DESCRIPTION

Feel free to ask any question!


People-Sea:
https://github.com/user-attachments/assets/7ff9f8c7-0fe9-49c0-8453-36a5051cf117



@ahmed-rashad-alnaggar 
Hi, could you test this scenario too? Opening a right-side slide-over sometimes distorts the Doughnut Chart, and the other charts resize a bit late. I attached a screen recording.

ahmed-rashad-alnaggar:
@People-Sea
Of course! Can you please provide a reproduction repo or just each chart settings?

People-Sea:
@ahmed-rashad-alnaggar https://github.com/People-Sea/test

ahmed-rashad-alnaggar:
Fixed!
The cause of only one to resize is splitting observing the element from creating the observer, something with alpine I guess.
And for the delay in resizing, it was due to the debounce, actually I left it as it is as the old implementation does, as I thought it could be to mitigate a bug or something, but after removing it and testing, everything is actually works very well!
@People-Sea can you test it now?

ahmed-rashad-alnaggar:
@People-Sea Any updates?

danharrin:
Please have patience, the core team is very busy.

ahmed-rashad-alnaggar:
Sorry, take your time!

People-Sea (commented):
I took a closer look at this PR after my initial pass, and I don’t think
  it’s safe to merge as-is.

  I understand the goal here is to improve the resize path, but the current
  implementation goes beyond that. It replaces the previous destroy/recreate
  model with a partial manual state sync layer, and that seems to introduce
  multiple regressions.

  I was able to reproduce at least these issues in the repro app:
  - object-shaped dataset updates break
  - theme grid colors get stuck
  - the decimation/plugin update path becomes stale



https://github.com/user-attachments/assets/c94505d5-8243-49e8-a511-2738a3e70a38





https://github.com/user-attachments/assets/91bbc46b-a8b8-4967-8856-36db2f7da4b8

ahmed-rashad-alnaggar:
Working on it!

ahmed-rashad-alnaggar:
Can you update the reproduction repo?

People-Sea:
@ahmed-rashad-alnaggar 
Sorry if my review comments were not clear enough.

What I mean is that this PR has become too broad and is no longer focused on the original problem. It includes too many additional changes, including state-sync and lifecycle rewrites around chart behavior.

Because of that, I don’t think this PR should continue moving forward in its current form.

ahmed-rashad-alnaggar:
@People-Sea
This PR focuses on refactoring the entire Chart.js implementation in Filament, as the title indicates—it’s not limited to “resizing.” You may have gotten that impression because I didn’t explain it clearly earlier (I’m a non-native speaker—sorry about that 😅).

I did consider splitting this into multiple PRs with step-by-step changes, but I felt that might be overengineering, since the main goal is to refactor how Filament implements Chart.js.

Let me know what you think. If needed, I can split this into multiple smaller refactor PRs if that approach is preferred.

danharrin:
If there is a simple way we can resolve the original concerns that @ahmed-rashad-alnaggar and I previously talked about in #19609, then I would accept a new PR with just what we discussed (as long as it doesn't cause problems). But I agree that this PR now includes way too much extra logic to fix problems that we did not previously discuss.

PR #19681, opened 2026-04-11
refactor: Chart.js implementation in chart widgets

ahmed-rashad-alnaggar:
This is **Phase 1 of 4**.
 
Adhering to the feedback from @People-Sea [comment](https://github.com/filamentphp/filament/pull/19652#issuecomment-4200856921), this PR refactors the Chart widget's Chart.js implementation. **This is a pure refactor; no functional changes or new features are introduced** [(except animation)](#animation-default-change).
 
---

## What’s Changed

- **Removed Global Defaults Mutation:** Stopped the practice of overriding global `Chart.defaults` (e.g. `backgroundColor`, `borderColor`, `color`, `font.family`) inside `initChart()`. Visual fallbacks (colors, fonts, etc.) are now applied locally to the per-chart `options` object instead. The two genuinely global defaults (`legend.labels.boxWidth` and `legend.position`) are hoisted to module level, where they belong.

- <a id="animation-default-change"></a>**Removed Animation Duration Default:** The line `Chart.defaults.animation.duration = 0` was removed.

- **Cleaner Event Handling:** Moved logic for theme changes, data updates, and resizing into dedicated methods (`updateChartTheme`, `updateChartData`, `_resizeChart`) instead of bulky inline anonymous functions.

- **`initChart` renamed to `_initChart()` and its `data` parameter removed:** The leading
  underscore signals that this is an internal method not intended to be called directly.
  The `data = null` parameter was also dropped — `_initChart` now always reads from
  `cachedData` directly, since `updateChartData` handles live data updates independently.

- **Helper-Driven Architecture:**
  - Introduced `whenChart()` to centralize null-safety checks before interacting with the Chart.js instance, eliminating three near-identical guard blocks that were duplicated across event handlers.
  - Extracted all `getComputedStyle` calls into `getChartFallbackColors()` for better readability and easier testing.
 
- **Decoupled Aspect Ratio Logic:** `options.maintainAspectRatio` is now derived by directly checking `this.$refs.canvas.style.maxHeight !== ''` instead of relying on the `maxHeight` prop passed from outside. This decouples the logic from external property drilling and removes an unnecessary function parameter.

- **Minor — `destroy()` uses optional chaining:** Replaced the manual `if (this.resizeObserver)` guard with `this.resizeObserver?.disconnect()`.


ahmed-rashad-alnaggar:
## Upcoming Phases

To keep pull requests concise and maintainable, the following updates will be handled in separate stages:

- **Phase 2:** Implement dynamic **chart resizing** (eliminating the need to re-initialize the instance).

- **Phase 3:** Enable seamless **theme updates** without full re-initialization.

- **Phase 4:** Add logic for reactive **chart data updates**.

**Note:** This staged approach ensures each PR remains focused and specific, as requested.

People-Sea (commented):
Thanks. I think I need to see where the rest of this is going before I can judge this phase confidently on its own.

The part I’m most concerned about is the manual dataset/state diff/state-sync path. I understand it is being introduced to support in-place updates, but if this approach is going to work, that part needs to be extremely robust. It is also the same area that caused issues in the previous PR, and a big reason that PR felt too broad to me.

Since this is already being presented in phases, this intermediate step is hard to evaluate in isolation without seeing the intended end state more clearly.

If you continue with the rest of the plan, it would help a lot if you could include the test scenarios you’re validating, along with a few short videos. That would make the review much easier.

ahmed-rashad-alnaggar:
Just to confirm I understood correctly — we’re good to proceed with this phase, and for the upcoming update/state-sync part you’d like me to include test scenarios and short videos demonstrating the behavior, right?

People-Sea:
Yes, you can continue and submit the full implementation.

But it’s worth noting that, if the full approach still relies on the same manual dataset/state diff/state-sync shadow layer as before, then I would still consider it a broad and high-risk change. If that layer is not clearly proven stable first, I don’t think we’re even at the point of meaningfully validating whether moving to update()/resize() avoids regressions.

Put simply, I think this approach can only move forward if you can show that the frontend state is being maintained correctly and does not introduce new issues. We should not introduce instability just to improve performance or visual behavior.

ahmed-rashad-alnaggar:
This is **Phase 2 of 4**.

Building on the refactor in Phase 1, this PR implements dynamic chart resizing — eliminating the destroy-and-reinitialize cycle that previously occurred on every resize event.

---

## What's Changed

- **`_resizeChart()` no longer destroys the instance:** Previously, `_resizeChart()` called
  `chart.destroy()` followed by `_initChart()`, which created a completely new Chart.js instance
  on every resize. It now calls `chart.resize()` directly, letting Chart.js handle dimension
  updates in-place.

- **`options.responsive` set to `false`:** With `responsive: true` (Chart.js's default),
  Chart.js monitors and controls the canvas size itself — which conflicts with our
  `ResizeObserver`-driven approach. Setting `responsive: false` hands full sizing control to the
  CSS layer, making `chart.resize()` the single explicit trigger for dimension updates.

- **Manual DPR (device pixel ratio) handling introduced:** With `options.responsive` set to `false`, Chart.js no longer listens for device pixel ratio changes (e.g. zooming). To preserve visual sharpness, a debounced `resize` listener is added that checks `window.devicePixelRatio` against `chart.currentDevicePixelRatio` and resizes the chart when a mismatch is detected.

- **Canvas receives explicit CSS dimensions via `@style`:** The canvas `style` attribute is
  replaced with a `@style([...])` directive that always applies `width: 100%`, and conditionally
  applies either `height: 100%; max-height: 100%` (when no `$maxHeight` is set) or
  `max-height: {$maxHeight}` (when one is). This gives the canvas a reliable, CSS-defined size
  at all times — a prerequisite for `chart.resize()` to read the correct dimensions.

- **`hasMaxHeight` condition tightened:** The check in `_initChart()` was updated from `!== ''`
  to `!== '100%'` to align with the new canvas style output, where a canvas without a
  user-defined max-height now explicitly receives `max-height: 100%` rather than an empty style.

- **`_initChart()` and `ResizeObserver` setup deferred to `$nextTick`:** Both calls are now
  wrapped in `$nextTick()` inside `init()`, ensuring the DOM is fully rendered before the chart
  is initialized and the observer is attached. This also resolves an unwanted behavior caused by
  `Alpine.effect`'s immediate execution: in the old code, `_initChart()` ran first, then
  `Alpine.effect` was defined — and because `Alpine.effect` invokes its callback once on
  definition, it immediately triggered `updateChartTheme()`, which destroyed and re-initialized
  the freshly created chart instance on every page load for no reason. By deferring
  `_initChart()` to `$nextTick`, the effect's initial run finds no instance yet and exits early
  via `whenChart()`, leaving the chart to be initialized exactly once.

- The `Alpine.debounce()` wrapper on the `ResizeObserver` callback was removed — aside from being
  unnecessary now that resizing is a lightweight `chart.resize()` call rather than a full
  destroy-and-reinitialize, the debounce delay was itself the source of a visible flicker during
  resize.

 - ~~**Minor — `filled($maxHeight)` replaced with `$maxHeight`:** The Blade class condition for
  `fi-wi-chart-canvas-ctn-no-aspect-ratio` was simplified from `filled($maxHeight)` to
  `$maxHeight`, since `$maxHeight` is already a falsy-safe value in this context.~~

- **Introduced `$hasMaxHeight` for clearer sizing semantics:** Instead of relying directly on `filled($maxHeight)`, a dedicated `$hasMaxHeight` flag is now computed as `filled($maxHeight) && $maxHeight !== '100%'`. This makes the intent explicit and ensures that `100%` is treated as the default/unconstrained case rather than a user-defined max height.

## Screen Recording

> **Note:** Any perceived slowness in the recordings is due to hardware limitations of the recording device, not the implementation itself.

### Before

[filament-chart-resize-old.webm](https://github.com/user-attachments/assets/cdc8fbf1-eb7e-4b71-b35a-0c5dca270dd0)

### After

[filament-chart-resize-new.webm](https://github.com/user-attachments/assets/4a1a2423-e6a6-4219-9a28-3065a7791529)

ahmed-rashad-alnaggar:
This is **Phase 3 of 4.**

Building on the in-place resize introduced in Phase 2, this PR eliminates the remaining destroy-and-reinitialize cycle — the one that fired on every theme change.

---

## What's Changed

- **`updateChartTheme()` no longer destroys the instance:** Previously, `updateChartTheme()` called `chart.destroy()` followed by `this._initChart()`, discarding the entire Chart.js instance on every light/dark switch or system theme change. It now directly mutates `chart.options` with the updated color values and calls `chart.update('none')` — applying the new theme in-place with no instance recreation.

- **Three new properties:** `userBackgroundColor`, `userBorderColor`, and `userTextColor` — are now captured from the incoming `options` object at component initialization, mirroring the existing pattern for `userPointBackgroundColor` and the grid color properties, for `updateChartTheme()` to correctly re-apply user-supplied colors after a theme change

## Screen Recording

[filament-chart-theme-change.webm](https://github.com/user-attachments/assets/8c61a0cb-da64-4e03-8c57-e5e42826bc7f)

ahmed-rashad-alnaggar:
@People-Sea

I think it would make sense for this to be the final phase in this PR.

The remaining work in Phase 4 (updating/syncing chart data with reactive and animation logic) feels like a separate concern from the refactor and lifecycle improvements covered in Phases 1–3. It might be cleaner to handle that in a dedicated follow-up PR, so this one stays focused and easier to review.

packages/widgets/resources/views/chart-widget.blade.php
```diff
@@ -87,15 +87,17 @@ class="fi-wi-chart-filter-content-actions-ctn"
...
                 }}
             >
                 <canvas
                     x-ref="canvas"
-                    @if ($maxHeight)
-                        style="max-height: {{ $maxHeight }}"
-                    @endif
+                    @style([
+                        'width: 100%',
+                        'height: 100%; max-height: 100%' => !$maxHeight,
```
people-sea: I think `height: 100%` changes the sizing semantics here. It moves the chart away from the old ratio-driven sizing path and toward a CSS/container-driven height path, so `aspectRatio` from `getOptions()` no longer has the same control over the rendered height.




https://github.com/user-attachments/assets/2d011900-6e54-4e61-846c-2e2eb48133cc

packages/widgets/resources/js/components/chart.js
```diff
@@ -92,34 +60,19 @@ export default function chart({ cachedData, maxHeight, options, type }) {
...
 
             options ??= {}
+            options.backgroundColor = this.userBackgroundColor ?? backgroundColor
+            options.borderColor = this.userBorderColor ?? borderColor
+            options.color = this.userTextColor ?? textColor
+            options.font ??= {}
+            options.font.family ??= fontFamily
             options.borderWidth ??= 2
-            options.maintainAspectRatio ??= !!maxHeight
+            options.responsive ??= false
```
people-sea, line 91: I can reproduce visible blur by moving the window between displays with different DPRs, which suggests responsive: false is dropping Chart.js’s DPR resync path.

image

People-Sea:
Additionally, I found a new behavior change: pie/doughnut charts no longer preserve the visible separator between slices in this PR.

image

image

ahmed-rashad-alnaggar:
> Additionally, I found a new behavior change: pie/doughnut charts no longer preserve the visible separator between slices in this PR.

This is actually an intentional change rather than a regression.

Previously, Filament was setting:

```js
Chart.defaults.borderColor = borderColor
```

However, this never affected pie/doughnut charts in practice, because those chart types already define their own default `borderColor` (`#fff`) at the [dataset level](https://www.chartjs.org/docs/latest/charts/doughnut.html#dataset-properties). Since a dataset-level value takes precedence, the global default wasn’t actually used.

In this PR, the behavior was updated to:

```js
options.borderColor = this.userBorderColor ?? borderColor
```

So instead of relying on a global default, the border color is now applied explicitly via the chart options, which aligns more closely with what Filament seemed to intend.

The separator between slices still exists, but it’s now less noticeable because both the slice background and border default colors are derived from the same Filament primary/chart-color palette (e.g. `primary-400`, etc.), making them visually very close.

```css
& .fi-color {
        & .fi-wi-chart-bg-color {
            @apply text-color-50 dark:text-color-400/10;
        }

        & .fi-wi-chart-border-color {
            @apply text-color-500 dark:text-color-400;
        }
    }
```

If preserving a stronger visual separator as default is important, we could explicitly set a contrasting `borderColor`, but the previous implementation wasn’t actually responsible for that effect.

Capture

packages/widgets/resources/views/chart-widget.blade.php
ahmed-rashad-alnaggar: I wasn’t able to reproduce this on my end.

There might be a specific layout or configuration where this changes. Could you share the setup or steps you used so I can try to reproduce it?

People-Sea:
This does fix the `'100%'` case, but it doesn't really address the reason I raised the comment.

Before this PR, when no `maxHeight` was set, we left the canvas height alone and Chart.js stayed on its normal aspect-ratio sizing path. In this PR, the default case still gives the canvas `height: 100%; max-height: 100%`, and then `maintainAspectRatio` is inferred from whether `style.maxHeight !== '100%'`. So this is still relying on a different sizing model, not restoring the previous behavior.

`responsive: false` is also still opting out of Chart.js' own responsive / DPR handling, which was the other part of the concern here.

So while this makes the `'100%'` branch consistent with the current implementation, it still doesn't convince me that the resize path is correct or equivalent to what we had before.

packages/widgets/resources/views/chart-widget.blade.php
people-sea: I think this is more of an approach problem than a reproduction problem.


The concern isn't that I found one app where this looks wrong. It's that this PR is taking over sizing/responsive behavior at a different layer: giving the canvas an explicit CSS height by default, inferring chart behavior back from that style, and opting out of Chart.js' own responsive handling with `responsive: false`.

ahmed-rashad-alnaggar:
Thanks for the clarification.

After pushing the fix commit, I wanted to validate the behavior thoroughly before responding, so I held off commenting until I had tested it properly.

Regarding `responsive: false`, from looking into Chart.js internals, enabling `responsive` wires up multiple observers (resize observer, mutation observer, and a window resize listener for DPR changes). In the previous Filament implementation, those observers would all trigger, and since the chart was being destroyed/reinitialized on resize, that resulted in repeated teardown/recreate cycles.

In this PR, `responsive: false` is intentional to avoid that path and keep resizing under manual control. However, I’m aware this also opts out of Chart.js’ built-in DPR syncing, so I’m working on finding the best way to handle this without introducing regressions.

Because of that, the new resize model explicitly sets the canvas `width` (and `height` when no `maxHeight` is provided). Without these, Chart.js doesn’t have stable dimensions at initial render when `responsive` is disabled, which can break layout and animations (actually the animations is the big part, which didn't work at all even after setting `Chart.defaults.animation.duration = 0`). Setting them ensures the chart fits its container at render time and behaves correctly from the first draw. It also results in the chart preserving its initial aspect ratio, so subsequent manual resizes remain consistent.

I explained the `responsive: false` part first since it directly informs why the resize model is implemented this way. While this model moves away from the destroy/reinitialize cycle to a manual resize approach, I haven’t observed any behavioral differences in practice compared to the previous implementation — the rendered result and resizing behavior remain consistent from a user perspective.

ahmed-rashad-alnaggar:
If the new model doesn’t introduce any regressions, side effects, or observable behavioral differences — and it maintains (or even improves) animation smoothness, performance, and the overall user experience — I’m trying to better understand what concern remains with adopting it.

From my perspective, the goal is to avoid the previous destroy/reinitialize cycle while preserving the same rendering behavior as much as possible. If that goal is met without introducing drawbacks, I’d like to understand if there are still specific cases or risks I might be missing.

ahmed-rashad-alnaggar:
Working on DPR....

ahmed-rashad-alnaggar:
With the latest changes, the DPR-related issue is now resolved.

The current approach effectively mimics `responsive: true`, but with animations working correctly and in a more performant way.

ahmed-rashad-alnaggar:
@People-Sea There’s one detail I’m still unsure about.

When collapsing and expanding the chart, there’s a scaling animation. I’m not fully convinced whether this is desirable behavior or not.

One option would be to prevent that entirely (e.g. via an `IntersectionObserver` or similar), so the chart only renders once it’s visible and avoids the scaling effect. The other option is to leave it as-is.

I’m leaning either way here, so wanted to get your opinion before deciding.

danharrin (changes requested):
The entry animation is super laggy. Does the entire PR depend on the entry animation being enabled?

This is what it was like to load a chart before

https://github.com/user-attachments/assets/16748e4d-2da3-4380-ab8a-dd00f5782d3e

And this is what it is like with this PR, notice the stutters?

https://github.com/user-attachments/assets/086a1707-d7d8-43fb-8800-212f52a9c107

If entry animation can be disabled and this PR is still useful, lets do that. If not, we won't be able to consider these changes.

ahmed-rashad-alnaggar:
One thing that might be contributing here is the default [easing](https://www.chartjs.org/docs/latest/configuration/animations.html#easing) (`easeOutQuart`). It can sometimes feel a bit heavy, especially on larger datasets or during initial render.

Would you mind trying to override the easing (e.g. setting it to `linear`) by passing `animation.easing` in the `getOptions()` method, and seeing if the issue still reproduces on your end? That would help isolate whether this is primarily an easing-related performance issue.

- If switching easing resolves the stutter, we could consider changing the default easing to something lighter.
- If the issue persists, we can fall back to disabling the entry animation by default (similar to the previous behavior).

Just to clarify, this PR doesn’t depend on the entry animation — it’s more of an added enhancement rather than a requirement, so disabling it wouldn’t affect the core value of the changes.

Let me know what you find 👍

ahmed-rashad-alnaggar:
@danharrin Did that works?

danharrin:
Linear animation does not look natural and I am honestly ok without any animation, so I have disabled animation out of the box again. You can enable it using the chart options if you wish.

danharrin (approved):

PR #19683, opened 2026-04-12
docs: add local development quickstart for contributors

dhrupo:
## Summary

This PR adds a short local development quickstart to the root README for contributors.

## Why

The current contribution guidance links to the docs, but new contributors still need to figure out the basic monorepo setup, how to run the demo app, and which test command to use. This adds a small copy-paste-friendly section to reduce that friction.

## Changes

- add `composer install` and `npm install` setup steps
- document `composer test:sqlite`
- document `npm run build` for frontend-related changes
- document how to run the demo app from `docs-assets/app`

## Testing

- no runtime code changes
- docs-only change

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,8 +77,20 @@ composer install
 npm install
 cp .env.example .env
 php artisan key:generate
+touch database/database.sqlite
+# Update `.env` to use SQLite before migrating.
 php artisan migrate
+```
+
+Run the dev servers in separate terminals:
+
+```bash
+cd docs-assets/app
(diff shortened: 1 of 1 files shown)


danharrin:
Thanks for the effort, but the `docs-assets` app is not meant to aid development like this, it is meant for an internal script that takes screenshots, hence why it is not documented like that. The instructions linked on https://filamentphp.com/docs/5.x/introduction/contributing are correct, you probably should set up a separate test app.

Issue #19707, opened 2026-04-15
Double notification when using a searchable Select in Action schema (first is correct, second showing after first is incorrect and empty)


mrk-j:
### Package

filament/filament

### Package Version

v5.5.1

### Laravel Version

v13.5.0

### Livewire Version

v4.2.4

### PHP Version

PHP 8.5.4

### Problem description

I noticed that empty notifications in my Filament app, when investigating I found out it they appeared when using `->searchable()` on a `Select` in an `Action` like this:

```php
Action::make('updateName')
    ->successNotificationTitle('Name updated!')
    ->successRedirectUrl(function (Model $record) {
        return static::getUrl(['record' => $record]);
    })
    ->mountUsing(function (Schema $form, Model $record) {
        $form->fill([
            'name' => $record->name,
        ]);
    })
    ->schema([
        Select::make('name')
            ->searchable()
            ->options([
                'Name A' => 'Name A',
                'Name B' => 'Name B',
                'Name C' => 'Name C',
            ]),
    ])
    ->action(function (array $data, Model $record) {
        $record->name = $data['name'];
        $record->save();
    }),
```

It seems to be the specific combination of `Action::successRedirectUrl()` and `Select::searchable()`.

I did some investigating and my guess it is related to this (`NotificationsServiceProvider`):

```
        on('dehydrate', function (Component $component): void {
            if (! Livewire::isLivewireRequest()) {
                return;
            }

            if (store($component)->has('redirect')) {
                return;
            }

            $notifications = session()->pull('filament.notifications') ?? [];

            if (count($notifications) <= 0) {
                return;
            }

            session()->put('filament.notifications.claimed', $notifications);

            $component->dispatch('notificationsSent');
        });
```

The empty notification is due to the `claimed` session becoming a nested array: `['claimed' => ['claimed' => []]`. I can fix the issue by renaming the `'filament.notifications'` session to another not related string like: `'filament.notifications-test'` and leaving all `claimed` variants in place.

### Expected behavior

There should not be a second empty notification.

### Steps to reproduce

In the sample app you go to the `EditUser` page:
- Then click the `Update name` action and just hit `Save`. 
- You will then see a notification showing `Name updated!`.
- When this fades out an empty notification opens.

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/mrk-j/filament-double-notification-issue

### Relevant log output

```shell

```

PR #19715, opened 2026-04-16
fix: database notifications reacting to flash notification events

### LINKED ISSUE #19705: DatabaseNotifications::removeNotification() fails on PostgreSQL when flash notification ID is not a valid UUID
### Package

filament/notifications

### Package Version

v5.5.1

### Laravel Version

v13.5.0

### Livewire Version

v4.2.4

### PHP Version

PHP 8.5.3

### Problem description

We're seeing `SQLSTATE[22P02]: Invalid text representation` errors in our production logs on a PostgreSQL application that uses `->databaseNotifications()`. The error originates from `DatabaseNotifications::removeNotification()` passing an unvalidated `id` string to a query against the `uuid`-typed `notifications.id` column.

**Our production setup:** We use Pusher for broadcasting and have `->databaseNotifications()` enabled on our panel. The reason we have both broadcast and database notifications active is that we use broadcast notifications for real-time user-facing flash/toast alerts, while database notifications are used by Filament exports. This means both the `Notifications` and `DatabaseNotifications` Livewire components are present on every page.

The error occurs because:

1. Both `Filament\Livewire\Notifications` and `Filament\Livewire\DatabaseNotifications` listen for the `notificationClosed` Livewire event.
2. `Notification::fromArray()` falls back to `Str::random()` (a 16-char alphanumeric string, not a UUID) when `$data['id']` is missing — this can occur when broadcast notification data is reconstructed without the original ID.
3. When the user closes a flash notification that has a non-UUID ID, `DatabaseNotifications::removeNotification()` receives that ID and runs a `DELETE` query. PostgreSQL rejects the non-UUID string.

While we haven't been able to pinpoint the exact sequence that causes broadcast notification data to arrive at `fromArray()` without an `id` key, we can confirm from our production Sentry logs that it does happen — the Livewire POST bodies show flash notifications in the `Notifications` component state with 16-character `Str::random()` IDs (e.g. `SXQ0lDh1M4WiicaT`) instead of UUIDs. It may be related to how Pusher delivers the data, or an edge case in `BroadcastNotificationCreated::broadcastWith()` overwriting the Filament notification ID.

Regardless of the upstream cause, the two code paths below make the bug applicable:

**`Notification::fromArray()`** (`vendor/filament/notifications/src/Notification.php:99`):
```php
```

**`DatabaseNotifications::removeNotification()`** (`vendor/filament/notifications/src/Livewire/DatabaseNotifications.php:42-46`):
```php
#[On('notificationClosed')]
public function removeNotification(string $id): void
{
    $this->getNotificationsQuery()
        ->where('id', $id)
        ->delete();
}
```

MySQL does not surface this bug because it stores notification IDs as `char(36)` and silently matches zero rows on format mismatch.

### Expected behavior

Closing a flash notification should not cause a database error, even when `DatabaseNotifications` is active. Either `fromArray()` should generate a valid UUID as its fallback, or `removeNotification()` should validate the ID format before querying.

### Steps to reproduce

1. Clone the reproduction repository and start it with Sail:
   ```bash
   git clone https://github.com/V13Axel/filament-notification-uuid-bug.git
   cd filament-notification-uuid-bug
   cp .env.example .env
   composer install
   ./vendor/bin/sail up -d
   ./vendor/bin/sail artisan key:generate
   ./vendor/bin/sail artisan migrate
   ./vendor/bin/sail artisan db:seed
   ```
2. Navigate to http://localhost:9991/admin/login
3. Log in with `admin@example.com` / `password`
4. Click **"Reproduce Bug"** in the sidebar
5. Click the red **"1. Send notification with Str::random() ID"** button
6. Close the warning notification that appears
7. Observe the `SQLSTATE[22P02]` error

The reproduction calls `Notification::make(Str::random())` directly to simulate what `fromArray()` does when data arrives without an `id` key — this is the same ID format we see in our production Sentry logs.

https://github.com/V13Axel/filament-notification-uuid-bug

### Relevant log output

```
SQLSTATE[22P02]: Invalid text representation: 7 ERROR: invalid input syntax for type uuid: "SXQ0lDh1M4WiicaT"
(Connection: pgsql, SQL: delete from "notifications" where "notifications"."notifiable_type" = user
and "notifications"."notifiable_id" = 1 and "notifications"."notifiable_id" is not null
and "id" = SXQ0lDh1M4WiicaT)
```

People-Sea:
## Functional changes


danharrin:
I think this is a side effect of #19707 rather than its own problem. Technically, flash notifications with the same ID as database notifications should close both, this is deliberate so that users don't see duplicate notifications.

People-Sea:
I think I may have misunderstood this initially, so I want to confirm one thing: should `DatabaseNotifications` ignore shared event IDs that aren't database notification IDs?

danharrin:
@People-Sea I did reopen the original issue as I misunderstood what was being reported. There shouldn't be an error if a notification is dismissed and there is no matching DB notification, you are correct.

The use case is: sending a broadcast notification and a database notification to a user at the same time (with the same notification ID), so if the user is currently using the app, they receive a broadcast flash notification immediately (through websockets), but if they are not connected then it is in their database notifications list. The idea is that if the user is using the app and dismisses the broadcast flash notification, that should also delete the database notification, otherwise it would be duplicative.

PR #19726, opened 2026-04-17
feat: add ->size() support to Filament\Schemas\Components\Icon

Waguilar33:
## Description

`Filament\Schemas\Components\Icon` has no way to control icon size. `IconEntry` (infolists) and `IconColumn` (tables) both support `->size(IconSize)`, but the schema `Icon` component does not, making it impossible to adjust size when using `Icon` in schema contexts.

A common use case is decorating form labels with an info-hint icon:

```php
Field::macro('infoHint', fn (string $hint): Field => $this->afterLabel(
    Schema::start([
        Icon::make('heroicon-o-information-circle')
            ->tooltip($hint)
            ->size(IconSize::Small), // previously not possible
    ])
));
```

This PR adds `->size()` and `->getSize()` to `Icon`, matching the `IconEntry`/`IconColumn` API. `->iconSize()` was not used — that convention is for components that *contain* an icon alongside other content (e.g. `Callout`, `Section`). `Icon` *is* the icon, so `->size()` is the correct name.

String values are coerced via `IconSize::tryFrom()` in `toEmbeddedHtml()` before being passed to `generate_icon_html()`, which requires `?IconSize`. A `null` size falls through to `generate_icon_html()`'s own default (`IconSize::Medium`), so existing rendered output is unchanged.

## Visual changes

No visual change when `->size()` is not called (defaults to `IconSize::Medium` as before). When called, the `fi-size-{value}` class is applied, matching the output of `IconEntry` and `IconColumn`.

## Functional changes


Waguilar33:
Not sure why tests are failing, seem unrelated to PR

danharrin (dismissed):

PR #19731, opened 2026-04-19
feature: `FileUpload`/`RichEditor` file path protection

danharrin:


samuelwei:
@danharrin Tampered file path are dropped not rejected. I would expect a validation error. If we set the file upload as required and manipulate the file path, the required validation is bypassed.

danharrin:
@samuelwei I have submitted PR #19759 for this

PR #19749, opened 2026-04-21
feat: allow passing a confirmation message directly into copyable()    

EG-Mohamed:
## Description

Previously, making a column or entry copyable with a custom confirmation message required two separate calls:

```php
TextColumn::make('email')
    ->copyable()
    ->copyMessage('Email address copied');
```

This PR introduces a string overload for `copyable()`, allowing the message to be passed inline:

```php
TextColumn::make('email')
    ->copyable('Email address copied');
```

The change is implemented in `Filament\Support\Concerns\CanBeCopied::copyable()` and automatically applies to all consumers of the trait, including:

* `TextColumn`
* `ColorColumn`
* `TextEntry`
* `ColorEntry`
* `CodeEntry`
* Text schema component

All existing call signatures continue to behave exactly as before:

* `copyable()`
* `copyable(true)`
* `copyable(false)`
* `copyable(fn () => ...)`

The standalone `->copyMessage()` method remains unchanged and takes precedence when both are used.

This patch also applies cleanly to the `5.x` branch. A follow-up PR can be opened once this is merged.

## Visual Changes

No visual changes. The rendered output is identical to using:

```php
->copyable()->copyMessage('...')
```

## Functional Changes

* [x] Code style has been fixed by running `composer cs`.
* [x] Changes have been tested and do not break existing functionality.
* [x] Documentation is up to date.

### DIFF
--- a/packages/infolists/docs/02-text-entry.md
+++ b/packages/infolists/docs/02-text-entry.md
@@ -756,6 +756,16 @@ TextEntry::make('apiKey')
     ->copyMessageDuration(1500)
 ```
 
+As a shorthand, you may pass the confirmation message directly into `copyable()`:
+
+```php
+use Filament\Infolists\Components\TextEntry;
+
+TextEntry::make('apiKey')
+    ->label('API key')
+    ->copyable('Copied!')
+```
+
 <AutoScreenshot name="infolists/entries/text/copyable" alt="Text entry with a button to copy it" version="4.x" />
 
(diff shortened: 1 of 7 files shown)


danharrin:
Thanks, but I think I will pass on this for now. In general, I am not a fan of parameters that do more than one thing based on the type passed, and try to avoid them where possible in the codebase currently.

PR #19750, opened 2026-04-21
feat: add DateRangeFilter with inline layout, native toggle, and per-picker customization

EG-Mohamed:
## Description

Filament currently lacks a built-in date range filter. Users must either implement a custom filter from scratch or rely on the `QueryBuilder`, both of which introduce unnecessary complexity for a very common use case.

This PR introduces a `DateRangeFilter` class to the tables package:

```php
DateRangeFilter::make('created_at');
```

This renders **"From"** and **"Until"** date pickers in the filter form and automatically scopes the query using `whereDate`. No custom `query()` callback is required for standard use cases.

---

## API

```php
DateRangeFilter::make('created_at')
    ->attribute('published_at') // Override the queried column (defaults to the filter name)
    ->fromLabel('Start date') // Custom label for the "From" picker
    ->untilLabel('End date') // Custom label for the "Until" picker
    ->inline() // Render both pickers side by side in a 2-column grid
    ->native(false) // Use the custom JS date picker instead of the browser native input
    ->withoutUntil() // From-only mode — hides the "Until" picker
    ->modifyFromDatePickerUsing(
        fn (DatePicker $datePicker) => $datePicker->minDate('2020-01-01'),
    )
    ->modifyUntilDatePickerUsing(
        fn (DatePicker $datePicker) => $datePicker->maxDate(now()),
    );
```

A custom `query()` callback is still supported for non-standard filtering logic.

## What’s Included

* `DateRangeFilter` class with `whereDate` scoping for both bounds
* Active filter indicators for each picker independently, with per-field removal support
* Full test coverage in `tests/src/Tables/Filters/DateRangeFilterTest.php`
* Documentation at `packages/tables/docs/03-filters/06-date-range.md`
* Translation keys under `filament-tables::table.filters.date_range`

## Visual Changes

No visual changes. The filter uses existing `DatePicker` and `Grid` components already present in Filament. 

## Functional Changes

* [x] Code style has been fixed by running `composer cs`.
* [x] Changes have been tested and do not break existing functionality.
* [x] Documentation is up to date.

### DIFF
--- a/packages/tables/docs/03-filters/06-date-range.md
+++ b/packages/tables/docs/03-filters/06-date-range.md
@@ -1,0 +1,122 @@
+---
+title: Date range filters
+---
+import AutoScreenshot from "@components/AutoScreenshot.astro"
+
+## Introduction
+
+Date range filters allow users to scope records by a start date, an end date, or both. To use one, create a filter using the `DateRangeFilter` class:
+
+```php
+use Filament\Tables\Filters\DateRangeFilter;
+
+DateRangeFilter::make('created_at')
+```
+
+This renders two date pickers — "From" and "Until" — in the filter form. When either is filled, the query is scoped using `whereDate` on the filter's attribute. You do not need to provide a custom `query()` callback for the common case.
+
+## Customizing the column used by a date range filter
+
+The column name used to scope the query defaults to the name of the filter. To customize this, you may use the `attribute()` method:
+
+```php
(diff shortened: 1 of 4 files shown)


danharrin:
Thanks for your contribution, but I am going to pass at the moment. We don't have a dedicated date range input currently, so the UX isn't great using two date pickers. When I build one of these in the future, I may also introduce a new filter type too. I am tracking that in #19802.

PR #19754, opened 2026-04-22
[4.x] fix(actions): batch cancel, failed() handler, explicit retry defaults

JoshSalway:
Two-commit PR on the five queueable jobs in `packages/actions`.

## Commit 1: `ExportCsv` and `ImportCsv`

Pure additions. No existing property touched; `$maxExceptions = 5`, `retryUntil()`, `backoff()`, `middleware()`, `tags()` delegation all preserved.

- **Batch-cancel early-return in `handle()`.** `Bus::findBatch($id)->cancel()` now stops in-flight chunks on the next worker cycle instead of running them all to completion.
- **`failed()` handler.** Scalar-only log context: `export_id` / `import_id`, `page`, `batch_id`, exception class, message.

Deliberately not touching `$tries`, `$timeout`, or `$maxExceptions` on these two jobs. They already delegate `retryUntil()` / `backoff()` / `middleware()` to the user's `Exporter` / `Importer`, and the existing `$maxExceptions = 5` was a deliberate choice in [`1c1fe80`](https://github.com/filamentphp/filament/commit/1c1fe80). Happy to revisit retry policy separately if you want; keeping this PR to additive changes.

## Commit 2: `PrepareCsvExport`, `ExportCompletion`, `CreateXlsxFile`

None of these three have ever had retry properties set. None delegate `retryUntil()`, so their attempt bound today is whatever the worker inherits (varies: `queue:work` 1, `vapor:work` runtime `SQS_TRIES ?? 3`, Horizon supervisor configs). Makes intent explicit per each job's idempotency and runtime profile.

| Job | `$tries` | `$backoff` | `$timeout` | Why |
|---|---|---|---|---|
| `PrepareCsvExport` | 1 | N/A | inherit | Retry would re-dispatch children via `batch()->add()`. Runtime scales with source-query size, keep worker-inherited timeout rather than 60s cap. |
| `ExportCompletion` | 1 | N/A | inherit | Retry would double-notify. Single `Notification::send()` / `sendToDatabase()` call, sub-second runtime. |
| `CreateXlsxFile` | 3 | `[30, 60, 300]` | inherit | Idempotent (overwrites same file); transient S3/disk blips recover. XLSX runtime is O(row count); large exports need worker-inherited timeout (Vapor Lambda ceiling, Horizon supervisor) rather than 60s cap. |

All three also get `$maxExceptions = 0` explicitly (the framework default is `null`, which disables the exception-counter check entirely and allows unbounded catchable-failure retries when paired with `retryUntil`) and a `failed()` handler. `PrepareCsvExport` gets the batch-cancel early-return. `queue:work` users see no change on the first two.

Can be dropped without affecting commit 1.

## Tests

Four narrow unit tests in `tests/src/Actions/`:

- `{Export,Import}CsvBatchCancellationTest`: uses `withFakeBatch(cancelledAt: ...)` + Mockery `shouldNotReceive('getAttribute')->with('user')` to prove the early-return fires.
- `{Export,Import}CsvFailedHandlerTest`: calls `failed(new RuntimeException(...))` against `Log::shouldReceive('error')` to pin the structured log context shape.

Full test plan: `composer test:phpstan` clean, `vendor/bin/pint --test --dirty` clean, `vendor/bin/pest tests/src/Actions` 470 tests / 1358 assertions green on PHP 8.4. Branch merges cleanly with `6bc4609`.

## Compatibility

Laravel 12.x / 13.x, PHP 8.2+. Public API unchanged on every job. No behaviour change on `ExportCsv` / `ImportCsv`: all additions, no existing property modified. On the three standalone jobs, `queue:work` users see no `$tries` change on `PrepareCsvExport` / `ExportCompletion`; `CreateXlsxFile` moves from worker-inherited default to explicit `3`.

Happy to tune at review time.

---

Context: <https://joshsalway.com/articles/improving-queue-safety-in-laravel>


danharrin (approved):

JoshSalway:
@danharrin thanks 👍

PR #19786, opened 2026-04-29
docs(forms): document visibility handling in relationship condition with saveRelationshipsWhenHidden

Sngt25:
Added section on using visibility in condition functions for relationship data.

## Description

This PR adds documentation for using component visibility checks within relationship condition functions. It includes an example showing how to use `$component->isVisible()` in the condition, and highlights the critical requirement of calling `saveRelationshipsWhenHidden()` to preserve data when the component is hidden. This method is currently not indexed in the official docs, so the documentation provides important clarification.

## Visual changes

N/A - Documentation only update

## Functional changes


danharrin:
I made some tweaks in 059cd013ccb8dfccf57301d13e89c08096aa4ff6

PR #19792, opened 2026-04-29
fix: hide actions with authorization feedback when denial has no message

intrepidws:
## Description
Currently `authorizationTooltip()` and `authorizationNotification()` always reveal a denied action, even when the policy returned a bare false. This can be lead to undesirable UX. For instance, simply showing a user a delete action (even in a disabled state) can lead to them questioning why they don't have that ability. While responses are helpful, it's sometimes preferrable to entirely hide the action.

This PR adds two new "or hidden" variants that give the message-bearing path the same UX as before, but fall back to hiding the action otherwise. 
- `authorizationTooltipOrHidden()` - disables the action with a tooltip when the deny response has a message; hides the action otherwise.
- `authorizationNotificationOrHidden()` - keeps the action clickable and shows a notification when the deny response has a message; hides the action otherwise.

These variants are meant for scenarios where there is nothing useful to tell the user or it is desirable to entirely hide the existence of the action from the user:

## Visual changes
When using either of these two new methods, the user will no longer see the action in scenarios where a message is not returned from the policy.

There are no changes (visual or otherwise) for current methods `authorizationTooltip()` and `authorizationNotification`.

## Functional changes


danharrin (changes requested):
In your opinion, should this just be default behaviour rather than opt-in?

intrepidws:
@danharrin Yes, absolutely. But I figured that could be construed as a breaking change of sorts?

danharrin:
Potentially, but I don't think its necessarily expected behaviour that the button should be visible and disabled either in this case. The method is an opt-in feature that is meant to be used with response messages. When response messages aren't being used, I think we might have some flexibility to fix the behaviour.

intrepidws:
That's a fair point. Whatever you think makes sense here but I would agree that this seems like a smart default. Let me know if you want me to do anything.

danharrin:
Yeah, lets make it the default behaviour instead of introducing a config option please

intrepidws:
@danharrin The changes have been made as requested.

danharrin (approved):

PR #19796, opened 2026-04-30
Fix table push columns

leandrocfe:
## Description

- Fix [`Table::pushColumns()`](https://filamentphp.com/docs/4.x/tables/overview#adding-new-columns-alongside-existing-columns) losing columns when `columns()` is called afterwards.
- Tests: adds `pushColumns` describe block covering the `Table::configureUsing()` global-push scenario, asserting      
  `created_at` and `updated_at` remain after a component calls `columns(['title'])`

## Functional changes

### DIFF
--- a/packages/tables/src/Table/Concerns/HasColumns.php
+++ b/packages/tables/src/Table/Concerns/HasColumns.php
@@ -24,6 +24,11 @@ trait HasColumns
      */
     protected array $columnsLayout = [];
 
+    /**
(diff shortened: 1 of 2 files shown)


danharrin:
Hmm sorry, I don't think this is how it should work. Push should push only, not push and affect how other methods operate in the future. You could instead call `->pushColumns()` later as well instead of `->columns()` again.

PR #19799, opened 2026-05-01
feat: add translate*() methods and translatable() for heading, description, tooltip, and placeholder

Waguilar33:
## Description

Extends Filament's existing `translateLabel()` pattern to four additional string-bearing traits that currently have no translation support. Each trait gains a `translate*()` boolean-flag method following the same pattern already established in `HasLabel`.

**Traits updated:**
- `HasHeading` (schemas) → `translateHeading()`
- `HasDescription` (schemas) → `translateDescription()`
- `HasTooltip` (support, actions, tables/columns, infolists) → `translateTooltip()`, `translateHeaderTooltip()`, `translateEmptyTooltip()`
- `HasPlaceholder` (support, forms, tables/filters) → `translatePlaceholder()`

A new `translatable()` convenience method on the base schema `Component` calls all available `translate*()` methods via `method_exists`, so applications can enable full translation for a component type in a single `configureUsing` call:

```php
// Enable globally for all sections
Section::configureUsing(function (Section $section): void {
    $section->translatable();
});

// Raw translation keys now work without __()
Section::make('app/labels.details')
    ->description('app/labels.details_description')
    ->schema([...])
```

```php
// Enable globally for all columns
Column::configureUsing(function (Column $column): void {
    $column->translatePlaceholder()->translateTooltip();
});

// Raw keys work directly
TextColumn::make('title')
    ->placeholder('app/labels.not_applicable')
    ->tooltip('app/labels.title_tooltip')
```

## Visual changes

N/A — behaviour-only change. No UI is altered; strings passed as raw translation keys are now passed through `__()` when the relevant flag is enabled. Without the flag, behaviour is identical to before.

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/HasTooltip.php
+++ b/packages/actions/src/Concerns/HasTooltip.php
@@ -9,9 +9,18 @@ trait HasTooltip
 {
     protected string | Htmlable | Closure | null $tooltip = null;
 
+    protected bool $shouldTranslateTooltip = false;
+
     public function tooltip(string | Htmlable | Closure | null $tooltip): static
     {
         $this->tooltip = $tooltip;
+
+        return $this;
+    }
+
+    public function translateTooltip(bool $shouldTranslateTooltip = true): static
+    {
+        $this->shouldTranslateTooltip = $shouldTranslateTooltip;
 
         return $this;
     }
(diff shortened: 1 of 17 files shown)


Waguilar33:
@danharrin This error in `InteractsWithRecord.php` was pre-existing in 4.x and unrelated to this PR, but it was blocking CI. Fixed the `nullsafe ?->` call on a type PHPStan can prove is non-nullable at that point.

danharrin:
Thanks, but this isn't something I am interested in introducing. Unlike labels, which are auto-generated by the framework, heading/description/tooltip etc are all passed by the user anyway, so they can use `__()` in the place they pass the string. I don't think the extra methods add value.

PR #19809, opened 2026-05-03
feat: enhance discover_app_classes to include models from app/Models …

3bd-ulrahman:


danharrin:
The class map will already contain models from that namespace, once the class map is built. I don't think we need to complicate it by scanning that directory again.

3bd-ulrahman:
it's annoying to make every time composer auto-load file to get the models during development

danharrin:
If you want to submit a separate PR to trigger the autoload automatically before this code runs then I would accept that

3bd-ulrahman:
i understand that you want composer to be single source of truth but trigger it will make composer scan all directories, that will be slow instead of just scanning app/Models directory.
also may add method to AdminPanelProvider to specify directories that need to be scanned for models

danharrin:
This is not a method that is called from requests or jobs AFAIK, so it doesn't matter if it is slow to load really.

danharrin:
Models can live anywhere in the directory tree, module/domain namespaces, inside vendor package code, etc, so we shouldn't just scan App\Models

3bd-ulrahman:
Yeah, I agree.
But I still think triggering Composer autoload generation could slow things down.

What about adding a method to PanelProvider that allows developers to specify which folder paths should be scanned? I think this would be much faster.

danharrin:
How much slower do you think it will make things? I think we could easily cope with an extra 2 or 3 seconds considering this is a CLI command.

3bd-ulrahman:
as you see, but at least for me i don't like 2 or 3 seconds need it in ms 
also composer may be in docker not in local machine or composer may throw error for any reason
so still see scan directories is better

PR #19812, opened 2026-05-03
fix(forms): customBlock TipTap node loses data-config + data-type on getHTML()

### LINKED ISSUE #19811: RichEditor customBlock loses data-config + data-type on editor.getHTML()
### Package

`filament/filament` (forms — RichEditor)

### Package version

4.x (current)

### Laravel version

n/a (bug is in JS extension)

### Livewire version

n/a

### PHP version

n/a

### Problem description

The `customBlock` TipTap node extension at `packages/forms/resources/js/components/rich-editor/extension-custom-block.js` loses two attributes when the editor serializes its content via `editor.getHTML()`:

1. **`data-config`** is rendered as `config="[object Object]"` — both the `data-` prefix is dropped and the object is stringified via `String()` instead of `JSON.stringify()`.
2. **`data-type="customBlock"`** is dropped entirely from the rendered output.

The visible symptom appears in any plugin that reads `editor.getHTML()` — most commonly the `awcodes/richer-editor` SourceCode plugin, which displays editor content in a CodeMirror modal. Users see:

```html
<div config="[object Object]" data-id="callout"/>
```

instead of the expected:

```html
<div data-type="customBlock" data-id="callout" data-config='{"text":"..."}'></div>
```

If the user then clicks Submit on the SourceCode modal with the mangled output displayed, the customBlock's `data-config` is permanently overwritten with `"[object Object]"` and the block's data is destroyed on save.

The same bug also breaks `parseHTML` round-trips: if any code path re-parses the renderHTML output, `JSON.parse('[object Object]')` throws.

### Expected behavior

`editor.getHTML()` should produce the same DOM structure that `parseHTML()` expects to consume, so customBlock nodes survive round-trip serialization.

### Steps to reproduce

1. Register a custom block via `RichEditor::make('content')->customBlocks([SomeBlock::class])`
2. Insert the block via the Blocks panel and configure it (e.g. add some text/values)
3. Open the SourceCode plugin from `awcodes/richer-editor` (or any tool that calls `editor.getHTML()`)
4. Observe the rendered HTML showing `config="[object Object]"` with no `data-type="customBlock"` attribute

### Root cause

In `packages/forms/resources/js/components/rich-editor/extension-custom-block.js`:

**Bug 1** — the `config` attribute (line ~130) has no `renderHTML` callback, so TipTap falls back to `String(value)`:

```js
config: {
    default: null,
    parseHTML: (element) =>
        JSON.parse(element.getAttribute('data-config')),
    // missing renderHTML
},
```

For comparison, the `id` attribute right below it has the correct pattern with a `renderHTML` callback that emits `data-id`.

**Bug 2** — the node's `renderHTML()` (line ~177) doesn't emit `data-type="customBlock"`, even though `parseHTML()` matches on it:

```js
renderHTML({ HTMLAttributes }) {
    return ['div', mergeAttributes(HTMLAttributes)];
},
```

**Bonus bug** — `addNodeView` (line 34) sets the DOM attribute via `setAttribute('data-config', node.attrs.config)`, which also coerces objects to `[object Object]`. The visual editor doesn't suffer from this because the node view reads `node.attrs.config` directly, not the DOM attribute. But this would bite copy/paste round-trips and any future code that reads the DOM.

Both attribute bugs are already correctly handled in the **PHP-side TipTap extension** for the same node, in the same package — `packages/forms/src/Components/RichEditor/TipTapExtensions/CustomBlockExtension.php` (lines 41-48 and 65-74). The fix mirrors the PHP behavior in JS.

### Why this is safe to fix

- **Public page rendering** (`RichContentRenderer::toHtml`) is unaffected — it transforms `customBlock` → `renderedCustomBlock` before serialization (see `RichContentRenderer.php::processCustomBlocks()`).
- **Visual editor display** is unaffected — `addNodeView` reads `node.attrs.config` directly, not the DOM attribute.
- **Editor save** is unaffected — saves go through `editor.getJSON()`, not `editor.getHTML()`.
- The fix only changes paths that are currently broken (Source Code mode display, copy/paste round-trips, any other consumer of `editor.getHTML()`).

### PR

I have a fix ready in iotron/filament@fix/custom-block-render-html that mirrors the PHP extension. Will open a PR linked to this issue.

iotron:
Closes #19811.

## What

Mirror the PHP-side `CustomBlockExtension` behavior in the JS-side `extension-custom-block.js` so the customBlock node round-trips correctly through `editor.getHTML()`.

## Why

The JS extension was missing:

1. A `renderHTML` callback on the `config` attribute → TipTap falls back to `String(obj)` → `[object Object]`
2. A `data-type="customBlock"` emission in the node's own `renderHTML()` → the rendered HTML can't be re-parsed by `parseHTML` (which matches on `div[data-type="customBlock"]`)

Both pieces are correctly implemented in the PHP twin: `packages/forms/src/Components/RichEditor/TipTapExtensions/CustomBlockExtension.php` lines 41-48 and 65-74. This PR ports the same logic to JS.

Bonus fix: `addNodeView` was setting `dom.setAttribute('data-config', node.attrs.config)`, also coercing to `[object Object]` — fixed to `JSON.stringify(node.attrs.config)`.

## Impact

| Path | Before | After |
|---|---|---|
| Source Code mode display (e.g. `awcodes/richer-editor`) | `config="[object Object]"`, missing `data-type` | Correct JSON-encoded `data-config` + `data-type` |
| `editor.getHTML()` round-trip via parseHTML | Throws (`JSON.parse('[object Object]')`) | Works |
| Public page render (`RichContentRenderer::toHtml`) | unaffected (transforms via `processCustomBlocks` before serialization) | unchanged |
| Visual editor display (`addNodeView`) | unaffected (uses internal `node.attrs`) | unchanged |
| Editor save (`editor.getJSON()`) | unaffected | unchanged |

No regression surface — the fix only changes paths that are currently broken. The 3 conceptual changes:

1. `addNodeView`: `dom.setAttribute('data-config', JSON.stringify(node.attrs.config))` — was passing the raw object
2. `config` attribute: added `renderHTML` callback that emits `data-config` with `JSON.stringify`
3. Node's `renderHTML()`: explicitly merge in `data-type="customBlock"`

## Reproduction

1. RichEditor with a custom block registered (e.g., `RichEditor::make('content')->customBlocks([SomeBlock::class])`)
2. Insert via the Blocks panel and configure
3. Open the SourceCode plugin from `awcodes/richer-editor`
4. Before: `<div config="[object Object]" data-id="..."/>`
5. After: `<div data-type="customBlock" data-id="..." data-config='{"...":"..."}'></div>`
6. Submit → reload → custom block survives intact

## Tests

Manual repro confirmed end-to-end against the actual issue. Happy to add Cypress coverage if maintainers point me at the right test file for the extension.


People-Sea (commented):
Please rebuild the assets and commit the compiled changes.

danharrin (approved):

PR #19820, opened 2026-05-06
Filament issue reproduction project

ijpatricio:
@danharrin  hey!

Since the `filament-issue` binary is going to be un when a Filament release is done, from my POV, we have 2 ways to approach it:
-  the release happens as usual, and then we dispatch another workflow, which polls packagist for that tag, creating the zip with a regular `composer require`
- we do it all in sync, no Packagist, adding the `packages` directory to the composer.json's repositories

This PR adheres to do it in sync.

Considerations:
- The workflow is running manually only.
- The last step should in theory, attach the zip to the release with same tag.
  - As I've seen immutable setting in Filament's release, this needs testing.

I have 5.x ready to PR as well, but maybe we finish this one first.

Can we process with some manual testing first in 4.x?


ijpatricio:
@danharrin also, check the releases on the fork, should you want to download/run 

https://github.com/ijpatricio/filament/releases

image

danharrin (approved):

PR #19823, opened 2026-05-06
fix: add wire keys to grouped action items

### LINKED ISSUE #19798: Bug: Incorrect loading indicators on table actions after changing default record
### Package

filament/filament

### Package Version

v5.6.1

### Laravel Version

v13.6.0

### Livewire Version

v4.2.4

### PHP Version

PHP 8.3

### Problem description

I have a Filament table listing languages. Each record has two actions:
- **Delete**
- **Make default language**

#### Behavior
- For the default language:
  - **Delete** is hidden
  - **Make default** is disabled

---

### Issue

After changing the default language, interacting with the *previous* default record causes inconsistent loading indicators:

1. Clicking **Make default language** → **no loading indicator appears**
2. Clicking **Delete** → **both actions show loading indicators at the same time**

---

### Important detail

- The issue always affects **the record that was default on initial page load**
- Even after switching the default multiple times, the problem stays tied to that original record
- If the page is refreshed, the issue resets and affects the new default at load

---

### Screen Recording

https://github.com/user-attachments/assets/ac14d083-d86d-4682-a921-c2b4ed0f81cf

### Expected behavior

- Each action should show its own correct loading state
- Loading indicators should not leak between actions or records

### Steps to reproduce

**Table Setup**

```php
return $table
            ->records( static fn (): array => /* languages api */ )
            ->columns([
                TextColumn::make('language')
                    ->label(__('columns.language.label')),

                TextColumn::make('is_default')
                    ->label(__('columns.is_default.label'))
                    ->formatStateUsing(static fn (bool $state): ?string => $state ? __('columns.is_default.badge') : null)
                    ->badge(static fn (bool $state): bool => $state)
                    ->color('success')
                    ->icon(static fn (bool $state): ?\BackedEnum => $state ? Heroicon::Star : null)
                    ->size(TextSize::Medium)
            ])
            ->recordActions(
                ActionGroup::make([
                    MakeLanguageDefaultAction::make(),
                    DeleteLanguageAction::make()
                ])->color('gray')
            )
            ->reorderable(
                column: 'order',
                condition: static fn () => count($websiteSettingsService->getLanguages()) > 1
            )
            ->paginated(false);
```

**Make Default Language Action Setup**

```php
return Action::make('make_language_default')
            ->label(__('label'))
            ->color('success')
            ->icon(Heroicon::Star)
            ->hidden(static fn (array $record): bool => $record['is_default'])
            ->action(static function (Component $livewire, array $record): void {
               // Set default language

                $livewire->resetTable();
            });
```

**Delete Language Action Setup**

```php
return DeleteAction::make('delete_language')
            ->modal()
            ->modalHeading(static fn (array $record) =>
                __('modal.heading', ['language' => $record['language']->getName()]))
            ->hidden(false)
            ->disabled(static fn (array $record): bool => $record['is_default'])
            ->successNotification(null)
            ->action(static function (Component $livewire, array $record): void {
               // delete language

                $livewire->resetTable();
            });
```

https://github.com/ahmed-rashad-alnaggar/filament-tables-actions-issue

### Relevant log output

```shell

```

people-sea:
## Description

Fixes #19798

Adds `wire:key` to grouped action items.

Without a key, Livewire can reuse a previous dropdown item when the visible actions in a group change, which can leave the loading indicator pointing at the wrong action.

I noticed the same kind of loading state issue may happen with a few other action layouts too.

Keeping this one scoped to grouped actions. If this approach makes sense, I can open a separate PR for the other layouts.

### Before

https://github.com/user-attachments/assets/66bbca46-f942-4b40-b7bf-1536a558a4e4

### After

https://github.com/user-attachments/assets/06643f6f-6fbe-4b33-9630-4190c64c7b7e

## Functional changes


danharrin (changes requested):
Thanks for taking this on. Two thoughts before merging:

**1. Scope to table-row actions.** For header and bulk actions, the visible set is stable per render (same auth checks, same buttons in the same position) — the morph problem you're solving is specifically about row-level dropdowns where the visible action set varies per record (e.g., delete hidden on protected rows) and rows can reorder/filter. Page-level header actions on `EditRecord`/`ViewRecord` also have a record but don't need the key. Easiest gate: only emit when `$this->getRecord()` and `$this->getTable()` are both set. Keeps the cost on the path that actually needs it.

**2. Other action layouts** — you mentioned the same issue exists elsewhere. Happy to fold those in here once #1 settles, or as a follow-up — your call.

people-sea:
Thanks Dan, I’ve scoped this to table row actions and added a few recordings showing the fix across the other action views too.



https://github.com/user-attachments/assets/0f6b0a18-202c-4ed7-b6f6-84f23101fdf9

danharrin (approved):

PR #19825, opened 2026-05-06
Refocus autofocus field after Create & create another

intrepidws:
## Description
After clicking **Create & create another**, the form is reset to its default state but focus stays on the submit button, so users have to click back into the first field before typing the next record. This change refocuses the field marked with `autofocus()` automatically.

A few implementation notes:
- Both `CreateAction` (modal/relation manager) and `CreateRecord` (resource page) now dispatch an `autofocus-form-fields` Livewire event once the form has been rehydrated with defaults.
- A new `autofocusFormFields` handler on the root `Schema`'s Alpine component listens for the event and focuses the first reachable `[autofocus]` element. The event payload includes `livewireId` so nested/sibling Livewire components ignore events that aren't theirs.

## Visual changes
The only visual change is the focus styles on the `autofocus()` field.

## Functional changes


packages/actions/src/CreateAction.php
```diff
@@ -136,6 +136,9 @@ protected function setUp(): void
                 $hydratedDefaultState = null;
                 $schema->hydrateState($hydratedDefaultState, shouldCallHydrationHooks: false);
 
+                $livewire = $this->getLivewire();
+                $livewire->dispatch('autofocus-form-fields', livewireId: $livewire->getId());
```
people-sea: Since #19817 added a generic schema client-side reset event, I think we should reuse that here instead of dispatching a separate autofocus-specific event.

intrepidws: Oh very nice, I wasn't aware of that PR. Will refactor to use that.

intrepidws: OK this is in place now, thanks again for the suggestion

danharrin (approved):

PR #19860, opened 2026-05-14
fix(tables): preserve addSelect() subqueries when resolving BelongsToMany records

Orrison:
## Description

When resolving a single record for a `BelongsToMany` relationship table — e.g. `assertTableColumnStateSet()`, edit/view action records, or anything routing through `resolveTableRecord()` — virtual columns added by user scopes via `modifyQueryUsing(fn ($query) => $query->addSelect([...]))` are silently dropped. `getStateUsing()` callbacks reading those virtual attributes return `null`.

The table rendering path (`getTableRecords()`) is unaffected, which makes this easy to miss: columns display correctly when listing, but the same column is `null` when reading state for a single record.

This mostly caused issues with `assertTableColumnStateSet` during tests.

## Root cause

In `resolveTableRecord()` (`packages/tables/src/Concerns/HasRecords.php:215-227`) the order of operations is:

1. `$relationship->getQuery()` — raw relationship query
2. `$table->applyQueryScopes(...)` — runs user `modifyQueryUsing()` callbacks, including any `addSelect()` subqueries
3. `$table->selectPivotDataInQuery($query)` — called `$query->select([...])`, which **replaces** the column list and wipes step 2's `addSelect`

The list path goes through `getQuery() → getRelationshipQuery()`, where `selectPivotDataInQuery()` runs *first* (HasQuery.php:101) and `applyQueryScopes()` runs *after* — so the bug never surfaces there.

## Fix

Change `$query->select($columns)` → `$query->addSelect($columns)` in `selectPivotDataInQuery()` at `packages/tables/src/Table/Concerns/HasQuery.php:130`.

Why this is safe:
- Eloquent's `Builder::addSelect()` initialises `$query->columns` from `null` when no prior `select()` was made, so callers that previously had no columns set end up with the exact same column list as before — `select()` and `addSelect()` produce identical SQL on a fresh query.
- Callers that *did* add columns earlier (the bug case) now have those columns preserved alongside the pivot/base selections.
- The method already explicitly includes `$query->getModel()->getTable() . '.*'` in `$columns`, so base-table columns are always selected.

The alternative — reordering operations in `resolveTableRecord()` — only patches one call site and leaves the same hazard for future callers (e.g. `HasBulkActions::getSelectedTableRecordsQuery()` at line 341). The one-line `addSelect` change addresses the root cause.

## Test

Added a regression test in `tests/src/Panels/Resources/RelationManagerTest.php` backed by a new fixture `DepartmentsWithModifiedQueryRelationManager` that:

- Calls `->modifyQueryUsing(fn ($query) => $query->addSelect([DB::raw("'preserved' as virtual_label")]))`
- Exposes the virtual column via a `TextColumn`
- Asserts `assertTableColumnStateSet('virtual_label', 'preserved', $department->getKey())`

Verified the test **fails** on the pre-fix code (reads `null`) and **passes** with the fix.

## Visual changes

N/A

## Functional changes

### DIFF
--- a/packages/tables/src/Table/Concerns/HasQuery.php
+++ b/packages/tables/src/Table/Concerns/HasQuery.php
@@ -127,7 +127,7 @@ trait HasQuery
             ];
         }
 
-        $query->select($columns);
+        $query->addSelect($columns);
 
         return $query;
     }
--- a/tests/src/Fixtures/Resources/Tickets/RelationManagers/DepartmentsWithModifiedQueryRelationManager.php
+++ b/tests/src/Fixtures/Resources/Tickets/RelationManagers/DepartmentsWithModifiedQueryRelationManager.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Filament\Tests\Fixtures\Resources\Tickets\RelationManagers;
+
+use Filament\Resources\RelationManagers\RelationManager;
+use Filament\Tables\Columns\TextColumn;
+use Filament\Tables\Table;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Support\Facades\DB;
+
+class DepartmentsWithModifiedQueryRelationManager extends RelationManager
+{
+    protected static string $relationship = 'departments';
+
+    public function table(Table $table): Table
+    {
+        return $table
+            ->modifyQueryUsing(fn (Builder $query): Builder => $query->addSelect([
+                DB::raw("'preserved' as virtual_label"),
+            ]))
+            ->columns([
+                TextColumn::make('virtual_label'),
+            ]);
+    }
+}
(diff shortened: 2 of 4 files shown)


danharrin (approved):

melvinmihaylovnulabg:
Hey @Orrison, I found a regression in Filament v5.6.4 from PR #19860 / commit 476b838eb344f9f461981b5fa5f85e427dfdd0dc.

The change in `packages/tables/src/Table/Concerns/HasQuery.php` from:

`$query->select($columns);`

 to:

`$query->addSelect($columns);`

breaks BelongsToMany relation manager row-action record resolution when the pivot table has its own `id`.

In our case:
- relation manager uses `belongsToMany(User::class, 'team_user')`

Orrison:
> Hey @Orrison, I found a regression in Filament v5.6.4 from PR #19860 / commit [476b838](https://github.com/filamentphp/filament/commit/476b838eb344f9f461981b5fa5f85e427dfdd0dc).
> 
> The change in `packages/tables/src/Table/Concerns/HasQuery.php` from:
> 
> `$query->select($columns);`
> 
> to:
> 
> `$query->addSelect($columns);`
> 
> breaks BelongsToMany relation manager row-action record resolution when the pivot table has its own `id`.
> 
> In our case:
> 
> * relation manager uses `belongsToMany(User::class, 'team_user')`

Ah, whoops! Looks like a fix for this was already raised and merged. And I believe @danharrin will be pushing out the release soon.

PR #19861, opened 2026-05-14
Fix schema component load listener for deferred dispatches

masskrdjn:
## Description

Fixes a JavaScript error triggered when a deferred schema component dispatch waits for a schema component to load.

In some cases, such as using a `RichEditor` inside a `Flex` component, the Livewire commit hook payload does not expose the component through the `component` variable used by the listener. This caused:

```text
ReferenceError: component is not defined
```

The schema dispatch listener now resolves the Livewire component from either the hook payload or the message payload before querying schema partials and registering the `schema-component-*-loaded` event listener.

## Visual changes

No visual changes.

## Functional changes

`composer cs` could not be run locally because the available CLI PHP version is 8.2.30, while the locked dependencies require PHP 8.3+ / 8.4 and the `zip` and `sockets` extensions.


danharrin:
Replaced by #19865

Issue #19857, opened 2026-05-12
richeditor link error while in flex component


EpicKau:
### Package

filament/filament

### Package Version

5.6.3

### Laravel Version

13.8.0

### Livewire Version

4.3.0

### PHP Version

PHP 8.4

### Problem description

When I place a richeditor field inside a Flex component the link function dont work.

I get the following error when I try to set a link in a richeditor field on a custom tenant profile page. 
`filament/schemas: ReferenceError: component is not defined` (schemas.js:133)

The code: 
`schema-component-${component.id}-${dispatch.params.awaitSchemaComponent}-loaded` 

This should be
`schema-component-${message.component.id}-${dispatch.params.awaitSchemaComponent}-loaded` 

### Expected behavior

No js error, standard link behavior

### Steps to reproduce

create a richeditor inside a flex component and try to set a link.
`Flex::make([
RichEditor::make('text') ])`

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/EpicKau/filamentphp-bug

### Relevant log output

```shell

```

danharrin:
Should be fixed by #19865.

PR #19865, opened 2026-05-15
fix: use correct Livewire id for awaited schema component events

people-sea:
### Fixes: #19857

  This fixes a v5-only rich editor link action error where awaited schema component events referenced an undefined `component` variable instead of the Livewire message component.

## Functional changes


danharrin:
Is this a replacement for #19861?

people-sea:
@danharrin 
Yes, this should replace #19861. #19857 is v5 / Livewire 4, and this was
  missed when the listener was updated to use `message.component`.

danharrin (approved):

PR #19867, opened 2026-05-15
feat: defer schema loading

KianAcquoy:
Solves: https://github.com/filamentphp/filament/discussions/5401

## Description

Builder components with many items are slow to render because every block's schema is evaluated and sent to the browser on initial page load, regardless of whether the user ever opens those items.

This PR adds opt-in lazy loading to the `Builder` component. Calling `->lazy()` makes the builder collapsible and starts every item collapsed. Each item renders only its header and a loading placeholder on first load. When the user expands an item, a Livewire call triggers `loadItem`, which renders the block's schema and injects it into the page.

```php
Builder::make('content')
    ->blocks([/* ... */])
    ->lazy()
```

State is always kept in the parent Livewire data array, values survive collapse/expand cycles and page reloads. Validation still runs against unloaded items: if a required field is missing in a collapsed block, the item is automatically expanded and loaded on submit.

**Per-block opt-out:** `Block::make(...)->lazy(false)` on an individual block forces it to render eagerly even inside a lazy builder.

**Unload on collapse:** `->lazy(unloadOnCollapse: true)` removes a block's schema from the DOM when the user collapses it again.

## Visual changes

_Instead of instantly opening the collapsed block, it will open with a loading state._
image

## Functional changes


KianAcquoy:
Hey @danharrin ! We think this addition would make a really good change to the current block builder, since we noticed that it can be quiet slow when using a lot of blocks.
Is there something like a roadmap, or do you have an estimate of when you have time to review this?

danharrin:
This is a new feature so needs to be released in a new minor version. I have only released patch versions since the PR was submitted. It should be reviewed at least once before the next minor release, but I am incredibly busy and new features are not a priority currently over bug and security fixes.

danharrin (changes requested):
Please merge the latest 4.x in, it refactors field views into HTML rendering functions on the component classes, so quite a lot of work is required to get this PR up to date.

Just to let you know, the next minor release will be made within the next few days, so there is a bit of urgency here if you want to see the change in Filament soon. Otherwise, the PR will remain open and will be addressed in the following minor release. It is up to you.

KianAcquoy:
> Please merge the latest 4.x in, it refactors field views into HTML rendering functions on the component classes, so quite a lot of work is required to get this PR up to date.
> 
> Just to let you know, the next minor release will be made within the next few days, so there is a bit of urgency here if you want to see the change in Filament soon. Otherwise, the PR will remain open and will be addressed in the following minor release. It is up to you.

I've merged 4.x into this branch & made changes to the code. Its fully up to date now

KianAcquoy:
@danharrin Hey! Do you have any indication on when you may have time to check this out?

danharrin:
Hi, it is the oldest PR so it will be first on my list to review for the next minor release. I dont know if you'll get a review next week or in 3 weeks, but it will be before any of the other new features are reviewed.

danharrin:
I think my plan for this is to refactor it to a feature where any part of a schema can be marked as hidden until it is visible in the viewport, when it gets lazy loaded via a partial render. Should be quite simple, I will push my changes to this branch I think.

I pushed #20369 initially to make the way that we toggle "concealed" components like tabs/wizard/collapsible section/collapsible repeater/collapsible builder is consistent, and then we can use an intersection observer consistently for the feature.

The builder being able to collapse items is just one example of a schema being able to be lazily loaded.

danharrin:
I will ship a patch release probably this week or next week containing #20369 and a few other bug fixes, hope that the community doesnt report any issues with components in concealed schemas, before shipping a minor release a couple of weeks later containing this new feature I think. #20369 reverts historical fixes for certain components that didnt work when concealed in another, but since we use the `display: none` technique elsewhere I think we might have fixed those issues in other ways since. And being able to use `display: none` means that the intersection observer will work everywhere and we can have a generic implementation.

KianAcquoy:
Alright, sounds good. Let me know if you need any further assistance!

danharrin:
I pushed the new generic implementation to this branch

KianAcquoy:
Looking solid! Tested it out, and works good.

danharrin (approved):

ahmed-rashad-alnaggar:
@danharrin Heads up: this update introduces a breaking change — the `schema` method's signature in `Filament\Schemas\Components\Concerns\HasChildComponents` (used by `Filament\Schemas\Components\Component`) now also accepts `Filament\Schemas\Schema`. Not sure if there are other cases like this, but I ran into it via the [Filament Flex Fields](https://github.com/janczakb/filament-flex-fields) package. Might be worth adding a notice about it somewhere.

PR #19897, opened 2026-05-22
fix(select): correct dropdown position in nested modals

IKetutWidiyane:
Fix incorrect dropdown positioning for Select component when used inside nested Filament modals.

1. Problem
Dropdown position breaks in nested modal scenarios due to absolute positioning context inherited from parent modal.

2. Solution
- Introduce usesFixedDropdownPositioning() helper
- Detect nested modal structure
- Force fixed positioning when Select is inside nested modals
- Ensure Floating UI strategy and DOM positioning are consistent

3. Impact
- No changes to normal Select usage
- No changes to modal behavior in single-level modals
- Only affects Select inside nested modals

### DIFF
--- a/packages/support/resources/js/utilities/select.js
+++ b/packages/support/resources/js/utilities/select.js
@@ -1362,17 +1362,35 @@ export class Select {
         return false
     }
 
+    usesFixedDropdownPositioning() {
+        const fixedPositioningContext = this.selectButton.closest(
+            '.fi-fixed-positioning-context',
+        )
+        const absolutePositioningContext = this.selectButton.closest(
+            '.fi-absolute-positioning-context',
+        )
+
+        if (
+            fixedPositioningContext !== null &&
+            absolutePositioningContext === null
+        ) {
+            return true
+        }
+
+        const modal = this.selectButton.closest('.fi-modal')
+
+        return (
+            modal !== null &&
+            (modal.parentElement?.closest('.fi-modal') ?? null) !== null
+        )
+    }
+
     async openDropdown() {
         // Make dropdown visible but with position absolute by default, or fixed in containers with .fi-fixed-positioning-context class, and opacity 0 for measurement
         this.dropdown.style.display = 'block'
         this.dropdown.style.opacity = '0'
 
-        // Check if the select is inside a container that opts in to fixed positioning
-        const useFixedPositioning =
-            this.selectButton.closest('.fi-fixed-positioning-context') !==
-                null &&
-            this.selectButton.closest('.fi-absolute-positioning-context') ===
-                null
+        const useFixedPositioning = this.usesFixedDropdownPositioning()
         this.dropdown.style.position = useFixedPositioning
             ? 'fixed'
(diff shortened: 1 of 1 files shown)


danharrin:
Do you have an issue or reproduction repository for this problem?

IKetutWidiyane:
> Do you have an issue or reproduction repository for this problem?

Yes, I can provide a reproduction.

Steps:
1. Create a nested modal scenario in Filament
2. Place a Select component inside the inner modal
3. Open dropdown

Issue:
Dropdown position is misaligned due to absolute positioning context from parent modal interfering with Floating UI calculations.

I will prepare a minimal reproduction repository if needed.

danharrin:
Yes, please do provide a minimal reproduction repo

IKetutWidiyane:
> Yes, please do provide a minimal reproduction repo

Here is a minimal reproduction repository:

https://github.com/IKetutWidiyane/filament-nested-modal-select-bug

The issue can be reproduced by following the steps in the README.

people-sea (commented):
https://github.com/user-attachments/assets/4008f0a1-347b-4876-8861-24515169d009

The repro does not match the PR description or the original issue. These action modals are rendered as sibling `.fi-modal` elements, not nested modals.

I don’t think we should merge this. Forcing `fixed` positioning here bypasses the absolute positioning context that modals already set up, and could regress dropdown positioning in scrollable modals.

danharrin:
Yeah this is all a bit strange, the fix doesn’t really match the problem either.

IKetutWidiyane:
> rp.mov 
> The repro does not match the PR description or the original issue. These action modals are rendered as sibling `.fi-modal` elements, not nested modals.
> 
> I don’t think we should merge this. Forcing `fixed` positioning here bypasses the absolute positioning context that modals already set up, and could regress dropdown positioning in scrollable modals.

Thanks for the clarification. I see now that the issue isn't actually related to nested modals. The current approach with fixed positioning is probably not the right solution here.

I'll take another look at the root cause and try a different approach.

danharrin:
@IKetutWidiyane can you attach a screen recording of your bug as I don't think we understand how to reproduce it

IKetutWidiyane:
> @IKetutWidiyane can you attach a screen recording of your bug as I don't think we understand how to reproduce it
I’ve attached a screen recording here showing the issue more clearly:

https://streamable.com/237uk4

After reviewing it more carefully, I think the issue is not actually related to nested modals themselves. The real problem seems to be the Select dropdown positioning inside the second modal.

The dropdown overflows outside the modal and appears visually misaligned instead of staying properly aligned with the Select field inside the active modal.

I also realize now that my previous `position: fixed` approach was probably targeting the wrong root cause.

danharrin:
Hmm, I don't see any misalignment on my end in that screen recording.

IKetutWidiyane:
Thanks for checking. Looking at it again, this may be more related to my environment or viewport than an actual Filament issue. I’ll investigate further before attempting another fix.

PR #19899, opened 2026-05-23
include `RestrictsFileUploadsToSchemaComponents` to components example

valpuia:
## Description

Include recent changes for [security about file uploads](https://github.com/filamentphp/filament/pull/19886) on Components example.

Why: when I create [custom pages](https://filamentphp.com/docs/5.x/navigation/custom-pages#creating-a-page), after it generated I go and check components docs and copy paste what I want to display on that page, so copying `RestrictsFileUploadsToSchemaComponents` traits will make it more secure

## Functional changes

### DIFF
--- a/docs/12-components/02-action.md
+++ b/docs/12-components/02-action.md
@@ -41,6 +41,7 @@ You must use the `InteractsWithActions` and `InteractsWithSchemas` traits, and i
 use Filament\Actions\Concerns\InteractsWithActions;
 use Filament\Actions\Contracts\HasActions;
 use Filament\Schemas\Concerns\InteractsWithSchemas;
+use Filament\Schemas\Concerns\RestrictsFileUploadsToSchemaComponents;
 use Filament\Schemas\Contracts\HasSchemas;
 use Livewire\Component;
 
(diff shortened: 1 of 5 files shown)


docs/12-components/02-action.md
valpuia, line 38: I am not sure whether we need to add `RestrictsFileUploadsToSchemaComponents` traits here

ziming:
will the code be different if it is spatie media upload field to s3?

valpuia:
> will the code be different if it is spatie media upload field to s3?

Not sure, but the docs say this:

> This means an attacker who can reach the page can tamper with a Livewire request to upload files to arbitrary property paths on any page that uses `InteractsWithSchemas`, even pages that do not display an upload field at all

So it doesn't matter about spatie media or not, it can happens when using `InteractsWithSchemas` traits, this is how I understand

danharrin (approved):

samuelwei:
image

Hi @valpuia, was `RestrictsFileUploadsToSchemaComponents` in combination with `InteractsWithSchemas` working for you?

image

valpuia:
@samuelwei After reading about security, I figure out that I don't need `InteractsWithSchemas` at all on my use-case, so I remove it from my project (after I create this PR).

And after checking your message, I try to use it and it doesn't work as well.. maybe opening an issue might be better

danharrin:
I have fixed the trait clash in 3d0903a3223fbd3c76367cbfc371dab792da7808

PR #19900, opened 2026-05-23
Add Bash, Kotlin, Ruby, and Rust to Language enum

nb3n:
## Description

This PR extends the existing Language enum (defined as string) by adding support for additional programming languages: Bash, Kotlin, Ruby, and Rust.

These languages were missing from the enum, which caused limitations when tagging or categorizing content such as blog articles and code examples. For example, Bash-related articles could not be properly labeled due to the absence of a corresponding enum value.

This change ensures more accurate classification and better coverage of commonly used programming languages.
## Visual changes

No UI changes are introduced by this update.

## Functional changes

- [x] New languages added to the Language enum: Bash, Kotlin, Ruby, Rust
- [ ] Code style has been fixed by running the composer cs command

### DIFF
--- a/packages/forms/src/Components/CodeEditor/Enums/Language.php
+++ b/packages/forms/src/Components/CodeEditor/Enums/Language.php
@@ -4,6 +4,8 @@ namespace Filament\Forms\Components\CodeEditor\Enums;
 
 enum Language: string
 {
+    case Bash = 'bash';
+
     case Cpp = 'cpp';
 
     case Css = 'css';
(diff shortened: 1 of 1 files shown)


danharrin:
This looks like an untested LLM-generated contribution. The `Language` enum has a single runtime consumer (`CodeEditor`), and the JavaScript side hard-codes which languages it can highlight — adding enum cases without the corresponding `@codemirror/lang-*` imports just produces values that fall back to plain text. The PR description references a use case (tagging blog articles / code examples) that this enum doesn't exist to support.

PR #19984, opened 2026-05-28
fix: stale dependent fields after item actions in `live()` Repeater & Builder

### LINKED ISSUE #19914: State is not updated when Repeater item si deleted
### Package

filament/filament

### Package Version

v5.6.5 / v4.11.5

### Laravel Version

v13.12.0

### Livewire Version

v4.3.0

### PHP Version

PHP 8.5

### Problem description

When an item is removed from a ->live() Repeater, the status for other fields is not updated

### Expected behavior

Other fields dependent statuses (computed through $get) should update accordingly

### Steps to reproduce

1. Start the reproduction project:

```bash
composer setup
php artisan serve
```

2. Log in at http://127.0.0.1:8000/admin with test@example.com / password.
3. Go in Users→ New User
4. Fill the first element label → its label is added in Main element's options:

Image

5. Add a new Element and fill its label → the label is added in Main element's options:

Image

6. Delete the first element → it's label still present in Main element's options:

Image

**bonus** 

if you add another element → select options are not updated (Element 1 still present):

Image

if you edit an element label → select options are updated (Element 1 is removed)

Image

**Another interesting thing:**

when an item is deleted, the "afterStateUpdated" of the repeater is called, but other fields computed values closures are not

https://github.com/fabio-ivona/filament-repeater-delete-trigger-issue

### Relevant log output

```shell

```

people-sea:
### Fixes: #19914 
Item actions on a `->live()` Repeater only call `partiallyRender()` on the Repeater itself, so sibling fields computed via `$get` are not re-evaluated.

Default `shouldPartiallyRenderAfterActionsCalled()` to `false` on `->live()` Repeaters so the parent Livewire component re-renders fully and dependent fields are recomputed.

### Before

https://github.com/user-attachments/assets/a5425c4b-6397-4f13-839d-204d6c9c067a

### After

https://github.com/user-attachments/assets/e10b8064-3acf-4aec-acb5-af7f273ae3f6

## Functional changes

### DIFF
--- a/packages/forms/src/Components/Builder.php
+++ b/packages/forms/src/Components/Builder.php
@@ -1213,16 +1213,10 @@ class Builder extends Field implements CanConcealComponents, HasExtraItemActions
 
     public function shouldPartiallyRenderAfterActionsCalled(): bool
     {
-        if (($condition = $this->evaluate($this->shouldPartiallyRenderAfterActionsCalled)) !== null) {
+        $condition = $this->evaluate($this->shouldPartiallyRenderAfterActionsCalled);
+
+        if ($condition !== null) {
             return (bool) $condition;
-        }
-
-        if (is_bool($this->isLive)) {
-            return ! $this->isLive;
-        }
-
-        if (! isset($this->container)) {
-            return true;
         }
 
         return ! $this->isLive();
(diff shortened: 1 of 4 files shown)


fabio-ivona:
thanks @People-Sea :heart:

danharrin (approved):

PR #19990, opened 2026-05-29
fix: render sidebar navigation item children when parent has no URL

### LINKED ISSUE #19989: [BUG] NavigationItem not showing children
### Package

filament/filament

### Package Version

v5.6.6

### Laravel Version

v13.12.0

### Livewire Version

v4.3.0

### PHP Version

PHP 8.5

### Problem description

When registering a custom Navigation item that has children, none of the children item are shown. 

The problem is the same when using sidebar or topbar, the only difference is the for topbar it will show children if you add `->group()`

Image

### Expected behavior

Expected output is to see an item in sidebar or topbar that has a dropdown menu showing it's children.

Bonus expected would be to be able to still use stuff like `->link` on the parent.  

### Steps to reproduce

```php
->navigationItems([
				NavigationItem::make('tools')
					->label(__('landing.nav.tools'))
					->icon(Heroicon::WrenchScrewdriver)
					->childItems([
						NavigationItem::make('all')
						->label('test'),
					]),
			])
```

https://github.com/gigili/filament-navitem-bug

### Relevant log output

```shell
No console or laravel logs created
```

truffle-dev:
## Description

Closes #19989.

When a top-level \`NavigationItem\` is registered via \`->navigationItems([...])\` with \`->childItems([...])\` and no \`->url(...)\` on the parent, the children never render in the sidebar. The reporter's repro (also in https://github.com/gigili/filament-navitem-bug):

\`\`\`php
->navigationItems([
    NavigationItem::make('tools')
        ->label(__('landing.nav.tools'))
        ->icon(Heroicon::WrenchScrewdriver)
        ->childItems([
            NavigationItem::make('all')->label('test'),
        ]),
])
\`\`\`

The gate at \`packages/panels/resources/views/components/sidebar/item.blade.php:115\` requires the parent or one of its children to be \`active\` before the children-\`<ul>\` is emitted:

\`\`\`blade
@if (($active || $activeChildItems) && $childItems)
\`\`\`

For a parent with no URL, \`$active\` is always false (no URL to match against the current route) and \`$activeChildItems\` is false (the children have no URLs either). So the gate fails and the children-\`<ul>\` is never written to the DOM.

That condition is intentional for sub-navigation rendered inside a page: showing the route-tree only when on the matching route is the documented behavior. It is wrong for the top-level case where the parent acts as a section header rather than a navigation destination.

This change adds \`blank(\$url)\` to the OR chain so the children render whenever the parent has no URL:

\`\`\`blade
@if (\$childItems && (blank(\$url) || \$active || \$activeChildItems))
\`\`\`

When the parent has a URL, the existing gate fires unchanged; sub-navigation behavior is preserved. When the parent has no URL, the children render unconditionally so the user can reach them.

## Visual changes

Before: a top-level \`NavigationItem\` with no URL and \`childItems\` renders only the parent button. Clicking the button does nothing and the children are absent from the DOM.

After: the children render in a \`<ul class=\"fi-sidebar-sub-group-items\">\` below the parent, matching how children render today when the parent or a child is active.

## Functional changes

- [x] Code style has been fixed by running the \`composer cs\` command.

### DIFF
--- a/packages/panels/resources/views/components/sidebar/item.blade.php
+++ b/packages/panels/resources/views/components/sidebar/item.blade.php
@@ -112,7 +112,7 @@
         @endif
     </a>
 
-    @if (($active || $activeChildItems) && $childItems)
+    @if ($childItems && (blank($url) || $active || $activeChildItems))
         <ul class="fi-sidebar-sub-group-items">
             @foreach ($childItems as $childItem)
                 @php


GeminiDev1:
Should a parent item with a `->url(...)` still render its `childItems`?

Or is that intentional to hide `childItems` when parent URL is set?

truffle-dev:
URL'd parents are intentionally unchanged. The new clause is `$childItems && (blank($url) || $active || $activeChildItems)`: `blank($url)` only opens the gate when no URL is set; with a URL, the original `$active || $activeChildItems` still gates the children-`<ul>`, so URL'd parents continue to show their children only on the matching route (the documented sub-navigation pattern). The added path is purely the top-level case where a parent acts as a section header rather than a destination.

danharrin (approved):

PR #20011, opened 2026-06-05
Fix Select multiple state synchronization before form validation

### LINKED ISSUE #20005: Bug Report: Select::make()->multiple() state not syncing before form validation
### Package

filament/filament

### Package Version

5.6.6

### Laravel Version

12.61.1

### Livewire Version

4.3.1

### PHP Version

8.3.30

### Problem description

When using Select::make('field')->multiple()->searchable(), the selected values are not reliably flushed from Tom Select's internal JS state to the Livewire component property before form validation runs on wire:submit. This results in a validation error (e.g., The field is required) even though items are selected in the UI.

Root Cause Investigation:
In vendor/filament/forms/resources/views/components/select.blade.php:

PHP
$isNative = (! ($isSearchable  $isMultiple 
 $isHtmlAllowed) && $isNative());
Because $isMultiple forces $isNative to false, it always renders the Tom Select Alpine component. The deferred $wire.$entangle() binding isn't flushing before the form submission action triggers validation.

Attempted Fixes (All Failed):

Plain ->multiple()->required()

Adding ->live() or ->live(onBlur: true)

Explicit statePath() targeting an isolated array property

What Works:

Bypassing Filament with a raw Blade <select wire:model.live="..." multiple> works perfectly, confirming it's an entangle/lifecycle sync issue with the custom component wrapper.

Using CheckboxList::make() works as a temporary UI workaround.

Has anyone else run into this synchronization race condition with multi-selects on v5, or found a clean way to force the state flush on submit?

### Expected behavior

When the user selects options in the UI and submits the form, those selected values should reliably sync to the component's state (e.g., $this->data['section_ids']) so that validation passes and the data can be processed.

### Steps to reproduce

1. Create a Filament form with a multiple, searchable, and required select component:

   ```php
   Select::make('section_ids')
       ->multiple()
       ->searchable()
       ->required()
       ->options([...])

2. Open the form in the browser, click the dropdown, and select one or more options.
3. Immediately click the form's submit button.
4. Observe the validation error indicating the field is required, and note that the backend state remains empty.

https://github.com/Elliot-fsc2/drdi/blob/main/resources/views/rdo/announcement/%E2%9A%A1section-create.blade.php

### Relevant log output

```shell

```

hamdyelbatal122:
When using `Select::make()-\>multiple()-\>searchable()`, the component renders a custom Alpine.js component instead of the native HTML select. Under the hood, this component uses `$wire.$entangle()` to sync the selection state.

### Cause
By default, `@entangle`/`$wire.entangle` uses Alpine watchers to synchronize state between Alpine and Livewire. However, because Alpine watchers run asynchronously in the microtask queue, a race condition occurs when the form is submitted immediately after selecting an option (or when submitting via pressing Enter on selection). The form submit event triggers backend validation on the server before the asynchronous entangle watcher has had a chance to flush the updated state to Livewire's JS client. Consequently, backend validation fails (e.g., throwing a false "required" validation error) even though the value is selected in the UI.

### Solution
Synchronously update Livewire's client-side state inside the `onStateChange` callback in the select Alpine component using `this.$wire.set(statePath, newState, false)`. Passing `false` as the third parameter ensures that Livewire's internal client-side state is updated immediately during the event execution tick (preventing timing issues) while maintaining the deferred network behavior (the network request is deferred until the form submits, unless the field is set to `live()`).

Fixes #20005


danharrin:
Sorry, I think this fix is too hacky for us to merge, we don't do similar things in other places IIRC. There must be another way to fix this.

hamdyelbatal122:
Hello @danharrin ,

I have restored the branch for this pull request on my fork:

https://github.com/hamdyelbatal122/filament/tree/fix/select-multiple-state-sync-validation

Latest commit SHA: 584c58ff12952892eca0ac966d924039ffc4da1d

Because the original fork was deleted, GitHub did not automatically re-link the head repository and the PR remains closed. If you would like to review and merge these changes without creating a new PR, you can import the branch locally using the commands below:

```
  git fetch https://github.com/hamdyelbatal122/filament.git fix/select-multiple-state-sync-validation
  git checkout -b fix/select-multiple-state-sync-validation FETCH_HEAD

```
(If you want to create a branch on the upstream repository first, run:)
```
  git push origin fix/select-multiple-state-sync-validation
```


If you prefer, I can open a new pull request from my fork with the same branch — please let me know.

Thank you for your time!

danharrin:
Hi, I am not concerned about where the `set()` is being called from. The problem wasn't that it was in the JS file. I am concerned about the `set()` itself, it bypasses Livewire's native state binding which we already have set up for this component. It doesn't feel like the right fix. I don't know how Livewire will react now or in future releases when we manually set the state like that.

PR #20012, opened 2026-06-07
fix: `WithFileUploads` trait clash in test fixtures

people-sea:
## Description
Removed the redundant `use WithFileUploads` causing a trait clash in these two fixtures.
## Functional changes


danharrin:
Thanks, this test failure actually surfaced a much bigger problem I didn't realist though (my original changes were a BC because of how PHP handles trait conflicts), so now I switched implementation again: 78a21a8f62ea6bfa5a25b7eb37d21a92d326d40c

PR #20018, opened 2026-06-08
fix: guard null child schema in Repeater::getItemLabel()

### LINKED ISSUE #19997: Repeater with `itemLabel()` crashes with "Call to a member function getStateSnapshot() on null"
### Package

filament/forms

### Package version

v5.6.6

### Laravel version

v13.12.0

### Livewire version

v4.3.0

### PHP version

8.4.21

### Problem description

```
```

```php
{

        // ...
        // ...
    ]);
}
```

### Expected behavior

```php
{

        'container' => $container,
        'item' => $container,
        'key' => $key,
        'schema' => $container,
        'uuid' => $key,
        'index' => $index,
    ]);
}
```

### Steps to reproduce

```php

```

### Reproduction repository

truffle-dev:
## Description

Closes #19997.

A \`Repeater\` configured with \`->itemLabel(...)\` can fatally crash a Livewire render with:

\`\`\`
\`\`\`

\`Repeater::getItemLabel(\$key)\` resolves the child container via \`\$this->getChildSchema(\$key)\` and then dereferences \`\$container->getStateSnapshot()\` unconditionally. \`getChildSchema()\` (in \`HasChildComponents\`) returns the schema only when \`\$key\` is present in the lazily-built \`cachedDefaultChildSchemas\`; for any other key it falls through and returns \`null\`.

The blade view (\`repeater/index.blade.php\`) iterates \`\$getItems()\` (recomputed fresh per render) and calls \`\$getItemLabel(\$itemKey)\` per item. When the cached child schemas were warmed earlier in the same request with a different set of keys than the current raw state (e.g. a sibling component called \`\$set()\` into the repeater path, a modal action wrote the parent array via \`fillForm()\`, or two overlapping Livewire requests left a stale cache), \`getChildSchema(\$itemKey)\` returns \`null\` and the next line fatally crashes the whole render. Only repeaters using \`->itemLabel()\` are affected, since \`getItemLabel()\` is the only caller that dereferences the child schema without a null check.

The reporter provided a public reproduction at https://github.com/MadBox-99/filament-repeater-itemlabel-crash with a Pest test that triggers the crash deterministically.

The method's return signature is already \`string | Htmlable | null\`, and the only consumer (the blade view at line 108) tests visibility with \`filled(\$itemLabel)\`, so \`null\` is a legitimate "no label for this item" return value. This change adds an explicit early return when the child schema cannot be resolved, instead of letting the closure run against a null container:

\`\`\`php
\$container = \$this->getChildSchema(\$key);

if (\$container === null) {
    return null;
}
\`\`\`

The new regression test exercises the failing path directly: a repeater with a single default item is asked for the label of a key that was never cached, and is expected to return \`null\` rather than crash.

## Visual changes

None. \`null\` was already a permitted return value, and the blade view treats a null item label as "no label", which is what the crash-free path now yields when the cache disagrees with the current raw state. Repeaters that hit this path no longer take the entire render down.

## Functional changes

- [x] Code style has been fixed by running the \`composer cs\` command.

### DIFF
--- a/packages/forms/src/Components/Repeater.php
+++ b/packages/forms/src/Components/Repeater.php
@@ -1231,6 +1231,10 @@ class Repeater extends Field implements CanConcealComponents, HasExtraItemAction
     {
         $container = $this->getChildSchema($key);
 
+        if ($container === null) {
+            return null;
+        }
+
         return $this->evaluate($this->itemLabel, [
             'container' => $container,
             'item' => $container,
--- a/tests/src/Forms/Components/RepeaterTest.php
+++ b/tests/src/Forms/Components/RepeaterTest.php
@@ -2341,6 +2341,24 @@ describe('item labels', function (): void {
         expect($repeater->getItemLabel($itemKeys[2], 2))->toBe('Item 2: third');
     });
 
+    it('returns `null` for `getItemLabel()` when the child schema is missing for the given key', function (): void {
+        Schema::make(Livewire::make())
+            ->statePath('data')
+            ->components([
+                $repeater = Repeater::make('items')
+                    ->schema([
+                        TextInput::make('name'),
+                    ])
+                    ->itemLabel(static fn (array $state): string => "Item: {$state['name']}")
+                    ->default([
+                        ['name' => 'first'],
+                    ]),
+            ])
+            ->fill();
+
+        expect($repeater->getItemLabel('this-key-has-no-cached-child-schema', 0))->toBeNull();
+    });
+
     it('can set `itemNumbers()` with a `Closure`', function (): void {
         $repeater = Repeater::make('items')
             ->itemNumbers(static fn (): bool => true);


danharrin:
Sorry, I don't think this fix is correct. The schema should always exist if the original method is invoked. I think this is patching over a different problem. Please let me know how the original problem is not fixable if you would like me to reopen this.

truffle-dev:
Thanks for the look. I walked the cache lifecycle to understand which path you have in mind.

`HasChildComponents::cachedDefaultChildSchemas` is cleared in two places: `HasState::rawState()` when state is loaded into the schema (line 668), and `HasState::callAfterStateUpdatedHooks()` after `afterStateUpdated` fires (line 217). Both cover the Filament-driven mutation surface. The reproducer in #19997 reaches the null branch by mutating the repeater's raw state through `data_set($livewire, $repeater->getStatePath(), $raw)`, a direct Livewire-property write that bypasses both, so the cache holds the prior keys.

Reading your "patching over a different problem" two possible ways:

1. The consumer-side path is the bug. Mutating Repeater state via `data_set()` on the Livewire instance is the wrong API; consumers should reach for a path that runs through `rawState()` (which clears the cache). A leaf null-guard papers over the wrong API choice and the right shape is a docblock or `Repeater::itemLabel()` note pointing at the correct mutation surface. In that case the report is closeable as user-error.

2. `Repeater::getChildSchema()` should be coherent with `getItems()` on every call. The blade renders fresh items from current raw state, but the label lookup goes through the lazy cache and can miss. A fix at this layer would rebuild from raw state on cache miss inside Repeater (override `getChildSchema()`, or detect divergence and clear), so the schema is always coherent with current raw state regardless of how it got there.

If you can point at which one you have in mind I will take it. If it is neither and the report itself is invalid, happy to leave the close as-is.

PR #20020, opened 2026-06-08
fix: restore page scroll position on browser back in SPA mode

### LINKED ISSUE #20017: Filament v5 does not restore the previous scroll position when navigating back in SPA mode
### Package

filament/filament

### Package Version

 v5.6.7

### Laravel Version

v13.14.0

### Livewire Version

v4.3.1

### PHP Version

8.5.7

### Problem description

In Filament v5, when I click the browser’s Back button, the page scrolls to the top instead of returning to the position from which I clicked the link. This issue only occurs in SPA mode.

Importantly, this issue does not occur in Filament v4. In fact, Filament v4 seems to be more responsive in this regard overall.

Filament v5 (bad):

https://github.com/user-attachments/assets/b6550234-c156-4f45-b7ae-1cd765603722

Filament v4 (good):

https://github.com/user-attachments/assets/8e8083a8-3b2b-4c1f-8da0-d50977eb4d98

### Expected behavior

The scroll position should return to the point from which I clicked the link. This is the natural behavior of web browsers, so it should be replicated in SPA mode.

### Steps to reproduce

1. Clone repo
2. Run `artisan db:seed`
3. Display more rows (e.g. 50)
4. Go to "view" on some bottom rows
5. Press "back" button in your browser

https://github.com/webard/filament-scroll-problem

### Relevant log output

```shell

```

hamdyelbatal122:
## Summary

- Fixes scroll position not being restored when pressing the browser back button in SPA mode ([#20017](https://github.com/filamentphp/filament/issues/20017))
- Livewire's navigate plugin stores scroll on `document.body.scrollTop`, but Filament pages scroll via the viewport (`window.scrollY` / `document.documentElement.scrollTop`)
- Adds `scroll-restoration.js` that:
  - Saves the viewport scroll position on forward `livewire:navigate` (when leaving a page)
  - Skips saving on history navigation, since the viewport is already at the top and would overwrite the stored value
  - Restores the saved position via `livewire:navigating` `onSwap()` so it runs after Livewire resets scroll during the page swap
  - Falls back to restoring again on `livewire:navigated` with multiple delayed attempts to win any remaining races

This follows the same event-driven pattern already used for sidebar scroll preservation in `stores/sidebar.js`.

Fixes #20017

## Test plan

- [x] Enable SPA mode on a panel (`->spa()`)
- [x] Open a resource list with enough records to scroll (e.g. 50+ rows)
- [x] Scroll down and click "View" on a record near the bottom
- [x] Press the browser back button
- [x] Confirm the list page restores the previous scroll position instead of jumping to the top
- [x] Confirm forward navigation via links still scrolls to the top as expected


people-sea (commented):
I think this should be fixed upstream in Livewire rather than patched in Filament.

danharrin:
Agree with @people-sea, this is too much new code to maintain when the fix should be in Livewire.

PR #20028, opened 2026-06-12
fix: Support --panel option in make:filament-user command

### LINKED ISSUE #20025: make:filament-user --panel=id crashes when no default panel is set
### Package

filament/filament

### Package Version

v4.x and v5.x

### Laravel Version

v11.0.0 / v12.0.0

### Livewire Version

v3.0.0

### PHP Version

PHP 8.3

### Problem description

Running `php artisan make:filament-user --panel=seller` crashes with `NoDefaultPanelSetException` when no panel provider has `->default()` set. 
`configurePanel()` via `HasPanel` correctly resolves the `--panel` option and stores the panel in `$this->panel`. But `MakeUserCommand:84` ignores it and calls `Filament::getCurrentOrDefaultPanel()` which only works during HTTP requests. In CLI, `getCurrentPanel()` returns null, so it falls to `getDefaultPanel()` → throws because no panel is marked default.
The `$this->panel` was already correctly resolved. Line 84's check is a stale leftover from before `HasPanel` and `--panel` existed (PR #

Image

18661 added `HasPanel` but never touched line 84).

### Expected behavior

`php artisan make:filament-user --panel=seller` should work without requiring any panel to have `->default()` set.

### Steps to reproduce

1. Register two panel providers, neither with `->default()`.
2. Run `php artisan make:filament-user --panel=<panel id>`.

https://github.com/ashraful19/filament-demo/tree/5x-bug-makefilament-user-default-panel

### Relevant log output

```shell
NoDefaultPanelSetException: No default Filament panel is set.

at Filament\PanelRegistry::getDefault()
at Filament\Facades\Filament::getCurrentOrDefaultPanel()
at Filament\Commands\MakeUserCommand::handle()  ← line 84
Root cause: packages/panels/src/Commands/MakeUserCommand.php:84 should check $this->panel instead of Filament::getCurrentOrDefaultPanel().
```

sawirricardo:
## Problem

Running `php artisan make:filament-user --panel=seller` crashes with `NoDefaultPanelSetException` when no panel provider has `->default()` set. `Filament::getCurrentOrDefaultPanel()` only works during HTTP requests — in CLI, `getCurrentPanel()` returns null and `getDefaultPanel()` throws.

Fixes #20025

## Solution

- Added `HasPanel` trait to `MakeUserCommand`, consistent with other panel commands (`MakeResourceCommand`, `MakePageCommand`, etc.)
- Added `--panel` option so users can specify the target panel
- `handle()` now calls `configurePanel()` to resolve the panel from the `--panel` option or prompt
- Auth guard and login URL resolve from the selected panel when available, with fallback to `Filament::getCurrentOrDefaultPanel()`

## Changes

1. `use HasPanel` trait added to `MakeUserCommand`
2. `--panel` InputOption added to `getOptions()`
3. `configurePanel()` called in `handle()` before checking panel availability
4. `getAuthGuard()` and `sendSuccessMessage()` use `$this->panel` when set

## Testing

1. Register two panel providers, neither with `->default()`
2. Run `php artisan make:filament-user --panel=seller`
3. Should complete without `NoDefaultPanelSetException`
4. Without `--panel`, should prompt for panel selection (consistent with other commands)


danharrin:
#20026

PR #20035, opened 2026-06-13
Fix `getItemLabel` null dereference and distinct validation for booleans in repeaters

### LINKED ISSUE #19995: Repeater distinct() forces selection for Checkbox/Toggle when multiple items exist despite nullable/optional state
### Package

filament/forms

### Package Version

v5.6.6

### Laravel Version

v12.61.0

### Livewire Version

v4.3.0

### PHP Version

PHP 8.5.6

### Problem description

When using a Repeater with a Checkbox input (or also with a Toggle input) with `->distinct()` and two or more entries in the repeater, then there has to be one Checkbox (or Toggle) selected, even though `->required(false)` or `->nullable(true)` is set.

Image

Image

Additionally, if the repeater only has one entry, there doesn't have to be one selected and it doesn't give a validation message and can be saved.

Image

### Expected behavior

The behaviour should be consistent no matter how many entries the repeater has.
It should respect `->nullable(true)` or `->required(false)` or any other suitable method for not forcing the user to select one Checkbox/Toggle.

### Steps to reproduce

Setup a repeater like this:
```
Repeater::make('roles')
    ->schema([
        Checkbox::make('primary')
            ->distinct(),
    ])
```

In the reproduction repository it is reproducible in the Create-Action of the User Resource. 

Image

https://github.com/Timeo04/issue-repeater-distinct-not-required

### Relevant log output

```shell

```

### LINKED ISSUE #19997: Repeater with `itemLabel()` crashes with "Call to a member function getStateSnapshot() on null"

pranab-acharya:
This PR fixes two bugs with the `Repeater` component:

1. **Fixes #19997 (Null dereference when evaluating repeater item label without warmed child schema cache):**
When the repeater's cache is missing the item's key, `getChildSchema` evaluates to a fallback `ComponentContainer`. If a cache miss occurs, calling `getStateSnapshot()` on it caused a fatal error. The fix introduces the null-safe operator `$container?->getStateSnapshot()` in `Repeater::getItemLabel()`, gracefully preventing fatal errors. A test for this is added to `RepeaterTest.php`.

2. **Fixes #19995 (Repeater distinct() forces selection for Checkbox/Toggle when multiple items exist despite nullable/optional state):**
When using `->distinct()` on a boolean field (like Checkbox) inside a repeater, Filament was erroneously forcing at least one item to be selected, even if the field wasn't explicitly `required`. This fix adds a check `if ($component->isRequired())` around the `distinct.must_be_selected` failure, meaning that `distinct()` on an optional boolean will now appropriately behave as "at most one can be selected". A test case for this logic is added to `CheckboxTest.php`.

### DIFF
--- a/packages/forms/src/Components/Concerns/CanBeValidated.php
+++ b/packages/forms/src/Components/Concerns/CanBeValidated.php
@@ -685,7 +685,9 @@ trait CanBeValidated
                         return;
                     }
 
-                    $fail(__($validationMessages['distinct.must_be_selected'] ?? 'filament-forms::validation.distinct.must_be_selected', ['attribute' => $component->getValidationAttribute()]));
+                    if ($component->isRequired()) {
+                        $fail(__($validationMessages['distinct.must_be_selected'] ?? 'filament-forms::validation.distinct.must_be_selected', ['attribute' => $component->getValidationAttribute()]));
+                    }
 
                     return;
                 }
--- a/packages/forms/src/Components/Repeater.php
+++ b/packages/forms/src/Components/Repeater.php
@@ -1236,7 +1236,7 @@ class Repeater extends Field implements CanConcealComponents, HasExtraItemAction
             'item' => $container,
             'key' => $key,
             'schema' => $container,
-            'state' => $container->getStateSnapshot(),
+            'state' => $container?->getStateSnapshot(),
             'uuid' => $key,
             'index' => $index,
         ]);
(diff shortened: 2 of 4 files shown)


pranab-acharya:
Closing this PR because one of the issues it addressed (#19997) already has an active PR (#20018).

To keep things clean and follow best practices, I've isolated the fix for the remaining issue (#19995) into a new, dedicated Pull Request: #20036.

PR #20037, opened 2026-06-13
Fix #19997: Ensure Repeater cache coherence when state diverges

### LINKED ISSUE #19997: Repeater with `itemLabel()` crashes with "Call to a member function getStateSnapshot() on null"

pranab-acharya:
Fixes #19997.

## Problem

A `Repeater` configured with `->itemLabel(...)` can fatally crash a Livewire render when its raw state is updated directly via Livewire properties (e.g. `data_set()`) instead of Filament's internal state mutation APIs.

This happens because the Repeater caches the structural schemas of its child items during a request. When state is updated directly through Livewire, the internal child schema cache (`$cachedDefaultChildSchemas`) is not cleared. During the render cycle, the Blade view iterates over the newly added raw items and asks the Repeater for the label of an un-cached key. The cache lookup misses, returns `null`, and `getItemLabel()` blindly calls `getStateSnapshot()` on that null value, crashing the app.

A previous PR (#20018) attempted to fix this by adding a simple `if ($container === null)` check inside `getItemLabel()`. As @danharrin pointed out, this was just patching over a symptom: "The schema should always exist if the original method is invoked." 

## Solution

This PR fixes the root architectural issue by ensuring **Cache Coherence**.

Instead of papering over the crash, we now override `getChildSchema()` in the `Repeater` component to make it intelligent:
If a schema is requested for a key that exists in the true raw state but *is missing from the cache*, it indicates that the state has diverged due to a direct state mutation. When this divergence is detected, the Repeater flushes its outdated cache and forces a fresh rebuild from the true raw state. 

This ensures `getItemLabel` always gets a valid schema container and effectively fixes the state mismatch regardless of how the user mutated the Livewire data array.

I have also added a comprehensive Pest test to explicitly simulate this scenario (`data_set()`) to guarantee it no longer crashes.

### DIFF
--- a/packages/forms/src/Components/Repeater.php
+++ b/packages/forms/src/Components/Repeater.php
@@ -891,6 +891,20 @@ class Repeater extends Field implements CanConcealComponents, HasExtraItemAction
         return $this->evaluate($this->isReorderableWithButtons) && $this->isReorderable();
     }
 
+    public function getChildSchema($key = null): ?Schema
+    {
+        if (
+            filled($key) &&
+            (! array_key_exists($key, $this->cachedDefaultChildSchemas ?? [])) &&
+            is_array($state = $this->getState()) &&
+            array_key_exists($key, $state)
+        ) {
+            $this->clearCachedDefaultChildSchemas();
+        }
+
+        return parent::getChildSchema($key);
+    }
+
     public function isAddable(): bool
     {
         if ($this->isDisabled()) {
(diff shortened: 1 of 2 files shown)


danharrin:
#20039

PR #20038, opened 2026-06-13
Fix: keep the submit button disabled while any field is still processing

SkHCrusher:
## Description

The shared `formButton` Alpine component tracks field processing (e.g. file uploads) with a single boolean `isProcessing`. When more than one field reports processing at the same time — for example several `FileUpload` fields uploading concurrently — the first `form-processing-finished` event sets `isProcessing` to `false`, re-enabling the submit button and hiding the processing message even though the other fields are still uploading. The form can then be submitted before all uploads have finished.

Each `FileUpload` only inspects its own FilePond instance before dispatching `form-processing-finished`, so it has no knowledge of other fields that are still processing. The shared `formButton` therefore has to count concurrent processing events instead of storing a single boolean.

This replaces the boolean with a counter: `form-processing-started` increments it, `form-processing-finished` decrements it, and `isProcessing` stays `true` until the count returns to zero.

### Steps to reproduce

1. A form (or action modal) with two separate `FileUpload` fields.
2. Drop a file into each at roughly the same time → the submit button shows the "Uploading file..." message and is disabled.
3. When the first upload finishes, the button is re-enabled and the message disappears, although the second upload is still running — so the form can be submitted before every file has uploaded.

## Visual changes
The form has two `FileUpload` fields (front + back image), both uploaded at once.
**Both files still uploading — button correctly disabled** (unchanged by this PR)
iScreen Shoter - Google Chrome - 260613143505

**Before — first upload finished, second still running: button wrongly re-enabled**
As soon as the first file finishes, `isProcessing` is reset to `false`, so the submit button becomes clickable again and the *Uploading file…* message disappears although the second file is still uploading.
iScreen Shoter - Google Chrome - 260613143307

**After — first upload finished, second still running: button stays disabled**
With the processing counter the button stays disabled until the last upload has finished.
iScreen Shoter - 20260613143812363

## Functional changes


danharrin (changes requested):
Please rebase this commit on 4.x and force push

SkHCrusher:
Hey @danharrin, thanks for the review!

I've rebased onto the latest `4.x` and force-pushed, so it's now a single clean commit with no merge commit, exactly as you requested.

Whenever you get a chance to take another look, that'd be much appreciated!

danharrin (approved):

PR #20051, opened 2026-06-17
fix: respect custom selectability in bulk actions when selecting all

### LINKED ISSUE #20050: Bulk actions execute on non-selectable records when using 'Select all matching'
### Description

When configuring custom selectability on a table using `checkIfRecordIsSelectableUsing()`, the checkboxes for non-selectable records are correctly disabled in the UI. 

However, if a user selects all records on the page and then clicks the banner link to **"Select all matching records"** (which switches Livewire to track deselected records), the bulk action ends up executing on the non-selectable records as well. 

This happens because the backend database query has no awareness of the PHP closure defined in `checkIfRecordIsSelectableUsing()`. 

To prevent this, Filament currently disables deselection tracking entirely when custom selectability is active (`canTrackDeselectedRecords()` returns `false`), forcing the client to download all individual IDs. But this workaround causes severe performance issues/memory limits on large datasets, and doesn't actually prevent the backend from resolving non-selectable records if deselection tracking is somehow active or simulated.

---

### Reproduction Repository

https://github.com/hamdyelbatal122/filament-reproduction-20050

### Steps to Reproduce

1. Clone the reproduction repository and install dependencies:
   ```bash
   composer install
   ```
2. Setup database, migrations, and seed the demo data:
   ```bash
   composer setup
   ```
3. Start the server:
   ```bash
   php artisan serve
   ```
4. Log in at `http://127.0.0.1:8000/login` with:
   * **Email:** `admin@filamentphp.com`
   * **Password:** `password`
5. Navigate to the posts page: `http://127.0.0.1:8000/blog/posts`
6. Notice that the checkbox for "Draft" posts is disabled (configured using `checkIfRecordIsSelectableUsing` in `PostsTable.php`).
7. Check the master checkbox in the header to select all posts on the page, then click the banner link **"Select all 83 posts matching..."**.
8. Click the bulk actions dropdown and trigger the **"Test Select All"** action.
9. You will see an notification saying:
   > **Executed on 101 records. Drafts included: 18**
   
   *(This shows that the bulk action executed on all records, including the 18 draft posts that are non-selectable).*

hamdyelbatal122:
Fixes #20050

This PR addresses the issue where bulk actions could run on non-selectable records when using the "Select all matching records" feature.

Instead of disabling deselection tracking across pages (which was the previous workaround and caused heavy performance/memory overhead on large tables), this PR:

1. Restores deselection tracking when custom record selectability is defined.
2. Filters out non-selectable records on the backend using the table's `isRecordSelectable()` check right after the selected records are resolved (for both collections and lazy collections).
3. Safely computes `totalSelectedRecordsCount` using the resolved collection's count instead of a direct database query count, since the database count doesn't account for the PHP selectability closure.

Added a unit test in `BulkActionTest` to cover this scenario and verify that non-selectable records are ignored when executing bulk actions with `isTrackingDeselectedTableRecords = true`.


people-sea (commented):
Can you explain how this solves the query-level problem mentioned in your issue? SQL still cannot know about the PHP selectability closure, and this PR appears to filter only after records are resolved.

hamdyelbatal122:
@people-sea 
You're right, SQL doesn't know about the PHP selectability closure, so the query itself still retrieves all records. Since the selectability check is a PHP closure, there's no way to filter at the database/query level.

The main issue this PR fixes is the performance/memory overhead. Previously, if custom selectability was defined, Filament disabled deselection tracking entirely (forcing Livewire to track and pass thousands of individual IDs over the wire, which hits memory limits on large tables). 

By filtering the records in PHP after they are resolved (in memory or chunked via lazy collections), we can safely restore deselection tracking. This prevents Livewire from bloating with IDs, while ensuring the bulk action is never executed on non-selectable records.

people-sea:
I think this direction is unsafe: it re-enables deselection tracking while only filtering the resolved `$records` path, so query-based bulk actions  can still include non-selectable records.

hamdyelbatal122:
You're completely right @people-sea. I overlooked query-based bulk actions (like `DeleteBulkAction` with `fetchSelectedRecords(false)`), which would run on the unfiltered query.

To fix this, I updated `getSelectedTableRecordsQuery()` to evaluate selectability via `lazy()` chunking and restrict the query builder to the selectable record IDs. This ensures both model-based and query-based bulk actions are completely safe while keeping deselection tracking enabled. 

I also added a unit test to verify this.

danharrin (approved):

PR #20052, opened 2026-06-17
fix(translation): improve Arabic translations, pluralization, and spelling

hamdyelbatal122:
This PR improves Arabic translation files in several packages (`actions`, `support`, and `tables`) by correcting spelling errors and aligning pluralization rules with proper Arabic grammar.

### What's Changed:
- **Spelling Corrections**:
  - Replaced "جاري" with the grammatically correct "جارٍ" (e.g., "جارٍ التحميل...", "جارٍ البحث...").
  - Fixed incorrect hamzas (e.g., changed "إتجاه" to the correct "اتجاه").
  - Corrected "لم يتم حذفه" to the more natural "تعذر حذف".

- **Pluralization Fixes (Dual/Plural Alignment)**:
  - Corrected translation keys (like pagination overview, table selection indicator, and action notifications) to follow Arabic dual rules (`{2} ... نتيجتين` and `{2} ... سجلين`) instead of treating them as generic plurals.
  - Aligned plural ranges (like `[3,10] ... صفوف` vs `[11,*] ... صفاً`) so that numbering ranges are grammatically accurate in Arabic.


packages/tables/resources/lang/ar/table.php
```diff
@@ -28,15 +28,15 @@
 
         'select' => [
 
-            'loading_message' => 'جاري التحميل...',
```
EG-Mohamed, line 31: I think **"جارٍ التحميل..."** is 100% correct in Modern Standard Arabic, but it's less familiar and less immediately readable for most users.

**"جاري التحميل..."** is more commonly used in software interfaces and is easier to recognize at a glance, which makes it a better choice from a UX perspective even if it's less grammatically precise.

I'd prefer to keep **"جاري التحميل..."** for consistency and readability.

packages/tables/resources/lang/ar/table.php
```diff
@@ -28,15 +28,15 @@
...
+            'loading_message' => 'جارٍ التحميل...',
 
             'no_options_message' => 'لا توجد خيارات متاحة.',
 
             'no_search_results_message' => 'لا توجد خيارات مطابقة لبحثك.',
 
             'placeholder' => 'اختر',
 
-            'searching_message' => 'جاري البحث...',
```
EG-Mohamed, line 39: Same here

EG-Mohamed (commented):
I’d recommend keeping **"جاري التحميل..."** instead of **"جارٍ التحميل..."**.

While **"جارٍ التحميل..."** is more accurate in Modern Standard Arabic, **"جاري التحميل..."** is more common in UI/UX contexts, easier to read quickly, and clearer for most users.

hamdyelbatal122:
Thanks for the feedback, @EG-Mohamed  

I get that "جاري" is visually common because a lot of informal sites/apps use it, but it's better to stick to standard Arabic grammar. 

Linguistically, the word is an "اسم منقوص نكرة مرفوع", so the grammatically correct spelling is strictly "جارٍ". 

Most big tech companies actually follow this standard in their official localizations. For example:
- **Google** uses "جارٍ التحميل..." on Android/Chrome.
- **Microsoft** uses "جارٍ المعالجة..." in Windows/Office.
- **Apple** uses "جارٍ الاتصال..." in iOS.

So keeping it as "جارٍ" is the standard for professional software interfaces. Let me know what you think!

packages/tables/resources/lang/ar/table.php
hamdyelbatal122: While "جاري" might be common in informal or amateur contexts, it is unequivocally a spelling and grammatical error (خطأ إملائي) in this specific syntax. In professional software localization, we do not intentionally introduce typos into the codebase just because they are "common" among some users.

EG-Mohamed:
@hamdyelbatal122  Thanks for the detailed clarification, I really appreciate it.

Just to be clear, I’m not against the change at all. I understand that **"جارٍ التحميل..."** is the grammatically correct form, and your explanation makes sense from a professional localization perspective.

My original point was only from the perspective of an end user. Since **"جاري التحميل..."** is visually more common in many Arabic interfaces, I felt it may be easier for some users to recognize quickly.

That said, I agree that **"جارٍ التحميل..."** is the correct standard Arabic form, and I’m totally fine with following the project’s localization style.

I’ll leave the final decision to the maintainers @danharrin based on what fits the project best: strict Arabic correctness or common UI familiarity.

Thanks again for taking the time to explain it.

hamdyelbatal122:
Thank you for your understanding and for the constructive discussion, @EG-Mohamed 

I completely understand your point regarding UI familiarity with commonly used forms.

That said, I'd like to add a small note from a UI/UX perspective: major platforms actually enforce the grammatically correct Arabic form in their everyday interfaces, which gradually helps users become familiar with the proper spelling over time.
As a quick example, here's Google Translate natively using "جارٍ الترجمة..." out of the box:

Screenshot_18-6-2026_18441_
Screenshot_18-6-2026_184441_

At the end of the day, we both want what's best for Filament. I'm more than happy to defer the final decision to @danharrin and the translation managers, trusting them to choose the direction that best aligns with the framework's quality standards.
Thanks again for your time and for the friendly, thoughtful review!

danharrin (approved):

PR #20055, opened 2026-06-18
fix: Navigation group enum respects `collapsible()` when `collapsed`

### LINKED ISSUE #20054: NavigationGroup::fromEnum doesn't use Collapsible interface correctly
### Package

filament/filament

### Package Version

v5.6.7

### Laravel Version

v13.15.0

### Livewire Version

v4.3.1

### PHP Version

v8.4.22

### Problem description

When `\Filament\Navigation\NavigationGroup::fromEnum` is checking if the Enum implements `Collapsible`, the order of operations appears wrong within the `if` statement.

https://github.com/filamentphp/filament/blob/5.x/packages/panels/src/Navigation/NavigationGroup.php#L143-L146

The NavigationGroup::collapsed() method also sets collapsible to whatever value collapsed is.  But because the `fromEnum` method calls `collapsible` first, then `collapsed`, you can never set a navigation group to be collapsible, but NOT collapsed initially.

I believe the order of the calls in `fromEnum` should be flipped, so that `collapsed` is called first, and then `collapsible`.

### Expected behavior

Here's an enum example that I would expect to result in a Navigation group that is collapsible, but not initially collapsed.  But what actually happens it that the group is not collapsible at all.

### Steps to reproduce

Example Enum:

```php
<?php

namespace App\Enums;

use Filament\Support\Contracts\Collapsible;
use Filament\Support\Contracts\HasLabel;

enum AppNavigationGroups implements HasLabel, Collapsible
{
    case GROUP;

    public function getLabel(): Htmlable|string|null
    {
        return match ($this) {
            self::GROUP => 'Group',
        };
    }

    public function isCollapsible(): bool
    {
        return match ($this) {
            default => true,
        };
    }

    public function isCollapsed(): bool
    {
        return match ($this) {
            default => false,
        };
    }
}
```

Within a Panel, use it:

```php
$panel->navigationGroups(\App\Enums\AppNavigationGroups::class)
```

https://github.com/mattzuba/filament-navigation-group-issue

### Relevant log output

```shell

```

people-sea:
### Fixes: #20054 

`collapsed()` also sets the collapsible state, so calling it after `collapsible()` in `fromEnum()` overwrote the enum's `isCollapsible()`. Swap the order.

### Before

image

### After
image

## Functional changes

### DIFF
--- a/packages/panels/src/Navigation/NavigationGroup.php
+++ b/packages/panels/src/Navigation/NavigationGroup.php
@@ -141,8 +141,8 @@ class NavigationGroup extends Component
         }
 
         if ($case instanceof Collapsible) {
-            $group->collapsible($case->isCollapsible());
             $group->collapsed($case->isCollapsed());
+            $group->collapsible($case->isCollapsible());
         }
 
         return $group;
--- a/tests/src/Panels/Navigation/NavigationTest.php
+++ b/tests/src/Panels/Navigation/NavigationTest.php
@@ -5,6 +5,7 @@ use Filament\Navigation\NavigationGroup;
 use Filament\Navigation\NavigationItem;
 use Filament\Pages\Page;
 use Filament\Resources\Pages\ListRecords;
+use Filament\Support\Contracts\Collapsible;
 use Filament\Tests\Fixtures\Clusters\UserManagement;
 use Filament\Tests\Fixtures\Clusters\UserManagement\Pages\ManageAdmins;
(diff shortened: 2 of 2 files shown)


danharrin (changes requested):
Would you mind adding a regression test for this please?

people-sea:
Done, thanks!

danharrin (approved):

PR #20057, opened 2026-06-18
fix(notifications): guard against NaN/empty transitionDuration and transitionEasing in animate()

wadimsewo:
## What is the problem?

When rapidly triggering actions that send notifications (4–5 times in quick succession), Livewire commits pile up. During these rapid commits, `window.getComputedStyle(el).transitionDuration` and `transitionTimingFunction` can return empty strings — causing `parseFloat('')` to yield `NaN` and leaving the easing as `""`.

Passing these invalid values to the Web Animations API throws:

```
TypeError: Failed to execute 'animate' on 'Element': duration must be non-negative or auto
TypeError: Failed to execute 'animate' on 'Element': Easing may not be the empty string
```

This crashes the Alpine.js component and blocks all subsequent Livewire requests, causing the UI to fully hang until a page reload.

## What is the solution?

Apply fallback guards at the point of assignment in `init()`:

- `parseFloat(...) * 1000 || 0` — falls back to `0` when the computed value is `NaN`
- `transitionTimingFunction || 'ease'` — falls back to `'ease'` when the computed value is an empty string

## Steps to reproduce

1. Open any Filament page with a table action that sends a notification on completion
2. Submit the action 4–5 times rapidly without waiting for notifications to dismiss
3. Observe the `TypeError` in the browser console and the UI hanging


people-sea (commented):
This looks more like a fallback than a root-cause fix. I tried several ways to reproduce this locally, but couldn’t trigger the error naturally.

If you’re seeing this in a real app, could you share the exact reproduction steps, along with a screenshot or video of the error? That would help us understand what’s causing the invalid transition values and fix the underlying issue, instead of adding a guard that may hide it.
Thanks!

wadimsewo:
I tried to reproduce this with a fresh installation but wasn't able to. I then noticed that the Livewire JS files were outdated — after publishing them with `php artisan livewire:publish --assets`, the error no longer occurred. It did seem to be connected to DB notifications as well. Since I wasn't able to reliably reproduce the issue, I'll close this PR for now. Thanks!

Issue #20063, opened 2026-06-20
createOptionForm() resets parent modal scroll position to top


AchmadFathoni:
### Package

filament/filament

### Package Version

v5.6.7

### Laravel Version

v13.16.1

### Livewire Version

v4.3.1

### PHP Version

PHP 8.4.22

### Problem description

When using `createOptionForm()` on a `Select` inside a form rendered in a modal, creating **or canceling** the nested action modal resets the parent modal's scroll position to the top.

Even clicking **Cancel** triggers the reset — despite Cancel generating zero Livewire requests and being intended as a no-op. After the nested modal closes, the parent modal DOM is visibly torn down and re-built (causing a brief flick), and the scroll container is a brand-new DOM node with `scrollTop = 0`.

Chrome also logs a scroll anchoring failure during the morph:

```
Scroll anchoring was disabled in a scroll container because of too many
consecutive adjustments (10) with too little total distance
(0.928 px average, 9.283 px total).
```

### Expected behavior

The parent modal's scroll position should be preserved after the nested `createOptionForm` modal closes — both on Create and on Cancel.

### Steps to reproduce

1. Clone the reproduction repository and set it up:

```bash
git clone https://github.com/AchmadFathoni/issue-reproduction-template
cd issue-reproduction-template
composer setup
php artisan serve
```

2. Log in at `http://127.0.0.1:8000/admin` with `test@example.com` / `password`.
3. Navigate to **Items** in the sidebar.
4. Click the edit button on "Test Item" — opens a `4xl` modal.
5. **Scroll to the bottom** — the `Category` Select with the `+` button is the last field, beneath 16 other fields (text inputs and textareas that force overflow).
6. Click the `+` button on the Category Select to open the create-option modal.
7. Click **Cancel**.
8. **Observe:** the parent modal scrolls back to the top.

Same result if you fill in the form and click **Create**.

The reproduction form (`ItemForm::configure()`) looks like:

```php
return $schema
    ->columns(1)
    ->components([
        TextInput::make('name')->required(),
        Textarea::make('description')->nullable()->rows(4),
        TextInput::make('price')->numeric()->nullable()->prefix('$'),
        TextInput::make('quantity')->numeric()->nullable(),
        TextInput::make('sku')->nullable(),
        TextInput::make('status')->nullable(),
        Textarea::make('notes')->nullable()->rows(4),
        // 8 padding textareas (3 rows each) to force modal scroll overflow
[...]

PR #20070, opened 2026-06-22
fix: modal scroll position after closing nested modals

VernSG:
## Description

Fix nested modal scroll behavior when closing a child modal inside another modal.

Previously, when a child modal was opened from within a parent modal and then closed, the parent modal could
jump back to the top because the scroll position was lost during Livewire morph/remount. This change
preserves modal scroll position across remounts and prevents browser scroll anchoring from forcing the
container back to the top.

This was verified in the issue reproduction template and matches the UX expected in #20063 

## Visual changes

https://github.com/user-attachments/assets/7ddd295d-2e33-4b12-80d6-b5b072b2ce9b

## Functional changes


people-sea (commented):
Thanks
This is actually about focusing on the restoration issue, and we shouldn't keep adding scroll bar state machines.

VernSG:
> Thanks This is actually about focusing on the restoration issue, and we shouldn't keep adding scroll bar state machines.

Ah, that makes sense. The browser is scrolling because it's trying to restore focus to the previous element when the nested modal closes. I agree that adding a scroll state machine is not the best approach here.

I'll drop the scroll state tracking and look into fixing the focus restoration behavior itself so it doesn't cause a scroll jump. I'll update the PR once I find a cleaner fix

VernSG:
I've tested PR #20077 locally and it perfectly solves the scroll jump issue by properly handling the manual focus restoration for nested traps. Since #20077 addresses the root cause more elegantly, I'll close this PR in favor of that one. Thanks @people-sea

PR #20071, opened 2026-06-22
fix: pre-populate `tableColumnSearches` to prevent JS `.length` collision on individual column search

### LINKED ISSUE #15073: Attribute with name length is not giving the desired output when using ->searchable(isIndividual: true)
### Package

filament/filament

### Package Version

v3.2.130

### Laravel Version

v.11.35.0

### Livewire Version

_No response_

### PHP Version

PHP 8.3

### Problem description

When you have an attribute named length in the database and you use the ->searchable(isIndividual: true) in the tables of filament you get a zero in the search field and the search field is not working. When deleting the zero to filter, the zero keeps coming back.

image

### Expected behavior

That the zero is not there and that I am able to filter individual on an attribute named length.

### Steps to reproduce

1. Create a table with a column that is named length with type integer or float
2. Create a standard resource of the model
3. Make the column length individual searchable

```php
    {
        return $table
            ->columns([
                Tables\Columns\TextColumn::make('id')
                    ->label('ID'),
                Tables\Columns\TextColumn::make('length')
                    ->suffix(' mm')
                    ->searchable(isIndividual: true)
                    ->sortable(),
                Tables\Columns\TextColumn::make('total_meters')
                    ->suffix(' m')
                    ->sortable(),
                Tables\Columns\TextColumn::make('supplier_reference')
                    ->label('Reference')
                    ->sortable(),
            ])
            ->filters([
                //
            ])
            ->actions([
            ])
            ->bulkActions([
                Tables\Actions\BulkActionGroup::make([
                    Tables\Actions\DeleteBulkAction::make(),
                ]),
            ]);
    }
```

https://github.com/Geoffry304/filament-length-attribute-issue

### Relevant log output

_No response_

lazerg:
### Fixes #15073

When a table column is named `length` (or any JS reserved property) and has individual search enabled, the input shows `0` on load and resets to `0` after clearing.

The problem is that `$tableColumnSearches` serializes as a JSON array `[]` when empty, so Alpine reads `[].length === 0` instead of the search value. Fixed by pre-populating all individually searchable column keys with `''` on mount so it always serializes as a JSON object `{}`, and replacing `Arr::forget` with `Arr::set($key, '')` on clear to keep the key present.


lazerg:
@danharrin mind taking a look when you get a chance? Small fix that pre-populates `tableColumnSearches` so a `length` column individual search stops colliding and showing `0`.

danharrin (approved):

PR #20075, opened 2026-06-23
feat: allow navigation parent items to be referenced by class name

ahmed-rashad-alnaggar:
This PR adds support for referencing parent navigation items by their class name, while preserving the existing behavior of referencing them by their label.

Using class names provides a more hierarchical and refactor-friendly approach.

## Example

Previously, child items had to reference the parent label:

```php
protected static ?string $navigationParentItem = 'Notifications';

protected static string | UnitEnum | null $navigationGroup = 'Settings';
```

With this PR, they can also reference the parent page or resource directly:

```php
protected static ?string $navigationParentItem = NotificationsResource::class;

protected static string | UnitEnum | null $navigationGroup = 'Settings';
```

Label-based references continue to work unchanged.

## Implementation

- `Page::getNavigationItems()`
- `Resources\Pages\Page::getNavigationItems()`
- `Resources\Resource\Concerns\HasNavigation::getNavigationItems()`

now return navigation items keyed by `static::class`:

```php
return [
    static::class => NavigationItem::make(...),
];
```

`NavigationManager` was updated to preserve collection keys during grouping and resolve parent items using either:

1. the item class name (collection key), or
2. the item label (existing behavior).

This preserves backward compatibility while allowing parent-child relationships to be defined independently from labels.

### DIFF
--- a/docs/06-navigation/01-overview.md
+++ b/docs/06-navigation/01-overview.md
@@ -124,7 +124,21 @@ All items in the same navigation group will be displayed together under the same
 
 ### Grouping navigation items under other items
 
-You may group navigation items as children of other items, by passing the label of the parent item as the `$navigationParentItem`:
+You may group navigation items as children of other items by setting the `$navigationParentItem` property.
+
+You may reference the parent item either by its label or by its page/resource class:
+
+```php
+use App\Filament\Resources\NotificationsResource;
+use UnitEnum;
+
+// By class name
+protected static ?string $navigationParentItem = NotificationsResource::class;
+
+protected static string | UnitEnum | null $navigationGroup = 'Settings';
+```
+
+You may still reference the parent by its label:
 
 ```php
 use UnitEnum;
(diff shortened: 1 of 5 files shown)


danharrin:
Unfortunately I don't think I can merge this, I know projects where users override `getNavigationItems()` and call `parent::getNavigationItems()[0]` to customize an item, and this would break cases like that.

ahmed-rashad-alnaggar:
@danharrin how about having a new method just for the navigation manager, so `getNavigationItems()` still unchanged?

danharrin:
What would that look like?

ahmed-rashad-alnaggar:
The same as `getNavigationItems` just for the navigation manager something like `getKeyedNavigationItems`.

Also I have a question, why it is `getNavigationItems(): array` not `getNavigationItem(): ?NavigationItem`, as I couldn't find a real case for a page or a resource to register more than one navigation item, was this a design mistake or I am missing something.

ahmed-rashad-alnaggar:
@danharrin Actually I figured a better approach, should I update this PR or create a new one?

And I still have the question I asked.

danharrin:
Yes it is deliberate that a resource or page can register multiple nav items. What is your proposal for fixing the problem?

ahmed-rashad-alnaggar:
@danharrin One approach would be to keep a keyed collection of navigation items in `NavigationManager` and use it while building the navigation in `NavigationManager::get()`. The collection could be populated either in `NavigationManager::mountNavigation()` or during `Page`/`Resource::registerNavigationItems()`.

danharrin:
How about adding `->key(static::class)` to `NavigationItem` when it is created and then the parent nav item matches by key or label

ahmed-rashad-alnaggar:
🏃 On it

ahmed-rashad-alnaggar:
@danharrin done! => #20107

PR #20090, opened 2026-06-29
DateTimePicker: add missing Korean (ko) dayjs locale

corean:
## Description

The JavaScript (non-native) date/time picker localizes its calendar via dayjs
locales registered in `packages/forms/resources/js/components/date-time-picker.js`.
The `locales` map already includes many languages (e.g. `ja`, `zh-cn`, `zh-hk`,
`zh-tw`), but **Korean (`ko`) is missing** — even though dayjs ships an official
`dayjs/locale/ko`.

As a result, when the picker is used with `->native(false)` and a Korean locale
(`->locale('ko')` or an app locale of `ko`), the calendar's month and weekday
labels fall back to English instead of rendering in Korean.

This PR registers the Korean locale so the JavaScript picker renders month/weekday
names in Korean:

```js
// packages/forms/resources/js/components/date-time-picker.js
const locales = {
    // ...
    km: require('dayjs/locale/km'),
    ko: require('dayjs/locale/ko'), // added (alphabetically between `km` and `ku`)
    ku: require('dayjs/locale/ku'),
    // ...
};


danharrin (changes requested):
Please rebase this on 4.x, recompile the assets, and force push

danharrin (approved):

PR #20107, opened 2026-07-03
feat: allow navigation parent items to be referenced by navigation item key

ahmed-rashad-alnaggar:
Rework on #20075 as @danharrin suggested!


danharrin (changes requested):
You are now keying by class name only, this breaks compatibility with the existing feature when users reference parents by label

Also please add two tests for label-based and class-based parent references to the existing navigation test file

ahmed-rashad-alnaggar:
Thanks for the feedback — fixed. Parent items can now be matched by either their key (class name) or their label, so this no longer breaks the existing label-based reference feature.

Also renamed the PR title from "feat: allow navigation parent items to be referenced by class name" to "feat: allow navigation parent items to be referenced by navigation item key" — while fixing the compatibility issue, I extended the same key-or-label matching to `HasSubNavigation::processParentNavigationItems()` as well, so sub-navigation parent items can be referenced by key too, not just label. Wanted the title to reflect that this applies consistently across both, rather than sounding like it's only about class names in the main navigation.

I've added the two tests you asked for (label-based and class-based parent references), plus a few more covering the merge behavior (children referencing the same parent via both key and label now get merged instead of the second set overwriting the first) and sort ordering of the merged children, for both the main navigation and sub-navigation code paths.

ahmed-rashad-alnaggar:
One implementation detail worth calling out from the initial commit: `getKey()` never returns `null` — when no explicit key is set, it falls back to the label, so it always returns a value.

danharrin (dismissed):

danharrin (dismissed):

packages/panels/src/Pages/Concerns/HasSubNavigation.php
ahmed-rashad-alnaggar, line 1: I don't see a necessary for this, as the items not keyed anymore.

I mean `return $items->values();` should be `return $items;`

danharrin (approved):

PR #20121, opened 2026-07-07
fix: Prevent in-flight uploads from being wiped when parallel uploads complete

### LINKED ISSUE #20122: FileUpload multiple(): removing a completed file while another upload is in flight wipes the in-flight upload
### Package

filament/filament

### Package Version

v4.11.8

### Laravel Version

v13.18.1

### Livewire Version

v3.8.2

### PHP Version

8.4.19

### Problem description

With `FileUpload::make()->multiple()` and the default `maxParallelUploads: 2`, removing an already-completed file with its ✕ ("tap to undo") button while another file is still uploading also discards the file that is still in flight: it disappears from the list mid-upload and is never stored.

Root cause: `shouldUpdateState` flips back to `true` as soon as the *first* upload finishes, while other uploads are still in flight. Removing the completed file then deletes its temporary `livewire-file:` value from the state, so the resulting state update passes the guard in the `$watch('state')` handler, which rebuilds the FilePond list from server-confirmed uploads only — wiping the in-flight file. On any real-world connection where uploads take more than an instant, this is trivially hit.

The race is also verified present on v3.3.54. Related to #13306: same underlying race and the same symptom ("files missing when uploading multiple files"). The poll-based path reported there did not reproduce for me on current 4.x, current 3.3, or the originally reported Filament 3.2.93 + Livewire 3.5.2. A fix is proposed in #20121; happy to backport to 3.x once the approach is approved.

### Expected behavior

Only the file whose ✕ button was clicked is removed; the in-flight upload continues and completes.

### Steps to reproduce

1. Throttle your upload speed (Chrome DevTools → Network → custom throttling profile, ~100 kbit/s up) — or use any real slow connection. On an unthrottled localhost the race window is only milliseconds wide, which is why this class of issue is hard to reproduce there.
2. Open a form with `FileUpload::make()->multiple()` and select a 1.5 MB and a 300 KB file together.
3. When the small file shows "Upload complete" while the big one is still uploading, click the small file's ✕ ("tap to undo") button.
4. The big file vanishes from the list too and is never stored.

https://github.com/Bramvzw/filament-13306-repro

### Relevant log output

_No response_

Bramvzw:
## Description

Fixes #20122. Related to #13306.

With `FileUpload::make()->multiple()` and the default `maxParallelUploads: 2`, `shouldUpdateState` flips back to `true` as soon as the *first* file finishes uploading, while other uploads are still in flight. Any state update without `livewire-file:` values that reaches the `$watch('state')` handler during that window rebuilds the FilePond list from server-confirmed uploads only — silently wiping the in-flight files.

**Reproduction (fully organic, see #20122):** upload two files on a connection where uploads take more than an instant; when the first finishes, remove it with its ✕ ("tap to undo") button while the second is still uploading. The removal deletes the completed file's temporary `livewire-file:` value from the state, the state update passes the guard, and the in-flight file is wiped and never stored. Verified on v4.11.8 and v3.3.54 (the race exists on both; happy to backport this fix to 3.x).

**Why this class of issue has been hard to reproduce:** on an unthrottled localhost the race window is milliseconds wide. Under a throttled or real-world connection it is as wide as the remaining upload time of the slower file. A single-threaded dev server (`php artisan serve`) also serializes the uploads, hiding the race entirely.

**The fix:** an `activeUploads` counter; `shouldUpdateState` only returns to `true` once *all* in-flight uploads have finished. The counter decrements on the success, error, and abort paths, so a failed or cancelled upload cannot leave the flag stuck. (Note: programmatic `pond.removeFile()` never aborts an in-flight Livewire upload — pre-existing FilePond behaviour, unrelated to this change.)

To be upfront about the history of this PR: my initial recordings simulated the state update with a scripted call, because I had assumed rather than verified the poll-based path reported in #13306. Testing showed that path does not reproduce on current 4.x, current 3.3, or the originally reported 3.2.93 + Livewire 3.5.2 — but it surfaced the trigger above, which does reproduce on both supported majors. The recordings below use only real user actions.

This takes the same counter approach as #19403 (credit to @isaackaara for the original diagnosis); this PR adds the reproduction, recordings, and edge-case verification that were missing there.

## Visual changes

Before/after recordings in the comment below: identical scenario and clicks, the only variable changed between the two recordings is this PR's built `file-upload.js`. Reproduction repository: https://github.com/Bramvzw/filament-13306-repro

## Functional changes

- [x] Code style has been fixed by running the `composer cs` command. (JS-only change; `npx prettier --check` passes)
- [x] Changes have been tested to not break existing functionality. (single upload, error path, abort path, remove-after-error verified in-browser under php-fpm)
- [ ] Documentation is up-to-date. (no documentation changes needed — behavioural bug fix)


Bramvzw:
Before:
https://github.com/user-attachments/assets/0c2df307-b991-40b4-85a8-81b50968f995

After:
https://github.com/user-attachments/assets/87ff317c-8cb2-4802-8e1a-2121ae20cd21

people-sea:
@Bramvzw 
Thanks for working on this.

Could you share the repo/branch, or even just the patch, that you used for the videos? I’m not asking for another recording ---- I’d like to run the same setup locally.

I think plain parallel uploads should not cause this by themselves. In FilePond, `load()` in `server.process` only gives FilePond the server id for that item; FilePond still owns the queued/processing items after that. On the Livewire side, finished temporary uploads come back as `livewire-file:...`, and Filament skips rebuilding the FilePond list while those values are present.

So I think this would require a very specific state update during the upload, not just one parallel upload finishing before another. I’d like to verify the exact code used for the video before changing the FileUpload state handling.

Bramvzw:
@People-Sea You were right to push on this. Plain parallel uploads don't cause loss by themselves, and the poll-based path from #13306 did not reproduce for me either — not on current 4.x, current 3.3, or the originally reported Filament 3.2.93 + Livewire 3.5.2. The `livewire-file:` guard happens to cover the window in every configuration I could build.

To be upfront: my initial recordings simulated the state update with a scripted call, because I had assumed rather than verified the reported path. Digging into your question surfaced a trigger that *is* fully organic and reproduces on both supported majors: removing the completed file (its ✕ / "tap to undo" button) while the second file is still uploading. The removal strips the guarding `livewire-file:` value from the state, and the resulting state update wipes the in-flight upload. I've opened #20122 for it and replaced the recordings with that scenario — real clicks only, and the only variable changed between the before/after recordings is this PR's built `file-upload.js`.

Repro setup: https://github.com/Bramvzw/filament-13306-repro — deliberately bare (no custom middleware, no polling, no custom JS). Practical note if you run it: throttle the upload in DevTools (~100 kbit/s) or use a multi-worker server; on an unthrottled localhost the race window is only milliseconds wide, and `php artisan serve` serializes the uploads entirely — which I suspect is why this class of issue has been so hard to reproduce.

danharrin (approved):

Bramvzw:
The bug also reproduces on 3.x, and this patch applies cleanly there — the only difference is that the 3.x `process` callback has no abort hook, so the counter is decremented on the success and error paths. Verified with the same before/after reproduction on a bare Filament 3 app (unpatched: the in-flight upload is wiped when the completed file is removed; patched: it completes).

Happy to open a backport PR to 3.x if you'd accept one — otherwise feel free to ignore this.

PR #20124, opened 2026-07-08
Add bulk actions (attach, detach, manage) to Spatie Laravel Tags Plugin

intrepidws:
## Description

This PR adds table bulk actions to the `filament/spatie-laravel-tags-plugin` package. Until now the plugin shipped a form field (`SpatieTagsInput`), a table column (`SpatieTagsColumn`), and an infolist entry (`SpatieTagsEntry`), but there was no way to tag many records at once. These actions fill that gap, mirroring the behavior and options of the existing `SpatieTagsInput`.

Three new bulk actions are added:
- `AttachSpatieTagsBulkAction` — opens a modal with a tags input and attaches the chosen tags to every selected record. Existing tags on a record are preserved (via `syncWithoutDetaching`), and tags that don't exist yet are created.
- `DetachSpatieTagsBulkAction` — removes the chosen tags from the selected records without deleting the tags themselves from the database.
- `ManageSpatieTagsBulkAction` — a single modal with two tags inputs, so users can attach some tags and detach others in one pass. Each record's attach + detach is wrapped in a `DB::transaction` so a mid-record failure can't leave it half-updated.

```php
use Filament\Actions\AttachSpatieTagsBulkAction;
use Filament\Actions\DetachSpatieTagsBulkAction;
use Filament\Actions\ManageSpatieTagsBulkAction;
use Filament\Tables\Table;

{
    return $table
        ->toolbarActions([
            AttachSpatieTagsBulkAction::make(),
            DetachSpatieTagsBulkAction::make(),
            // ...or, instead of the two above:
            ManageSpatieTagsBulkAction::make(),
        ]);
}
```

## Visual changes
Screenshot 2026-07-08 at 1 45 45 PM
Screenshot 2026-07-08 at 1 44 00 PM
Screenshot 2026-07-08 at 1 44 11 PM
Screenshot 2026-07-08 at 1 44 22 PM

## Functional changes

### DIFF
--- a/packages/spatie-laravel-tags-plugin/README.md
+++ b/packages/spatie-laravel-tags-plugin/README.md
@@ -187,7 +187,7 @@ AttachSpatieTagsBulkAction::make()
 
 When no `type()` is passed, the actions match tags across every type, in the same way as the `SpatieTagsInput` field. The [security caveats around tag types and privilege namespaces](#security-tag-types-and-privilege-namespaces) apply equally to these bulk actions, so pass `->type(...)` (or `->type(null)` for untyped tags only) when tag types are used as a privilege namespace in your application.
 
-You may authorize each selected record individually using [`authorizeIndividualRecords()`](https://filamentphp.com/docs/actions/create#authorization), passing the name of a policy method. Records that fail the authorization check are skipped, and the user is notified:
+You may authorize each selected record individually using [`authorizeIndividualRecords()`](https://filamentphp.com/docs/actions/overview#authorization), passing the name of a policy method. Records that fail the authorization check are skipped, and the user is notified:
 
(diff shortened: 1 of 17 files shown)


danharrin:
Thanks for you work here, I think I am going to pass though. In my opinion customisability is quite important for tagging UIs as tags can be used in a variety of ways, and I think having users define these in their apps is probably not too bad. I would suggest opening discussion threads in the future and tagging me to suggest big features like this as it helps us validate it first before you spend time on the implementation.

intrepidws:
No worries @danharrin I understand. I had to implement this for a project of mine, so extracting it into this PR wasn't a ton of work. Thanks for your work as always.

PR #20125, opened 2026-07-09
fix: restore default action badge colors

people-sea:
## Description

Restored the default `primary` badge color in optimized action rendering paths when an action has a badge but no explicit badge color.

This matches the existing Blade component defaults and ensures default action badges receive the expected `fi-color-primary` classes.

### Before
image

### After

image

## Functional changes

### DIFF
--- a/packages/support/src/View/Concerns/CanGenerateButtonHtml.php
+++ b/packages/support/src/View/Concerns/CanGenerateButtonHtml.php
@@ -53,6 +53,7 @@ trait CanGenerateButtonHtml
         string | Htmlable | null $tooltip = null,
         ?string $type = 'button',
     ): string {
+        $badgeColor ??= 'primary';
         $color ??= 'primary';
 
         if (! $iconPosition instanceof IconPosition) {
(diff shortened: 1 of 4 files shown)


people-sea:
Sorry for the confusion. I overlooked that action badges are intended to be gray by default in 4.x.

danharrin:
@people-sea I'm not sure about that tbh - the Blade components seem to say the opposite, right? Or am I misunderstanding?

Could just be a regression introduced by 4.x, we basically need to make these HTML generation functions act exactly the same as the Blade component versions

people-sea:
@danharrin 
Yeah, you're right. I got this wrong while going through the optimization commits. I later checked further back and confirmed it was a regression in 4.x, but then I saw the gray badge in the docs screenshot and assumed that was intentional, so I closed the PR.

I can reopen it if you'd like.

danharrin:
I will make the commit, no worries!

danharrin:
Done in 732795a5ec775001b42bafccf1bb2d56e1f8cde8

PR #20127, opened 2026-07-10
fix: prevent embedded field label slots from being escaped

people-sea:
## Functional changes


danharrin:
Sorry, didn't see this and just committed c79263016deb6c04bf9fba8a3d80e9b6eaf268b6! What do you think? I can definitely see the merits of your approach too, it might be more consistent.

danharrin:
I think I am going to keep my approach, I think there is slightly less chance that a user could be bitten by unescaped HTML in these slots by needing to wrap the content in `HtmlString`, you know?

people-sea:
> I think I am going to keep my approach, I think there is slightly less chance that a user could be bitten by unescaped HTML in these slots by needing to wrap the content in `HtmlString`, you know?

I agree too!

PR #20128, opened 2026-07-10
feat: add global default alignment for header actions

hammadzafar05:
Header actions only had a per-page instance property for controlling their mobile alignment, with a getter but no setter. Form actions already have a global default alignment API on BasePage (alignFormActionsStart/Center/End(), formActionsAlignment()). This adds the equivalent for header actions on InteractsWithHeaderActions, plus a fluent instance setter that was also missing.

## Visual changes

Before:
image

After:
image

## Functional changes

### DIFF
--- a/docs/06-navigation/02-custom-pages.md
+++ b/docs/06-navigation/02-custom-pages.md
@@ -59,12 +59,33 @@ protected function getHeaderActions(): array
 
 #### Aligning header actions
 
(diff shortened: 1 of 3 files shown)


danharrin:
Thanks, I think I am going to pass on this though, I am just conscious of introducing yet more config properties and methods to these huge classes.

PR #20130, opened 2026-07-10
Feature: additional classes when row has no actions

webard:
## Description

This pull request adds additional CSS classes to the table header and table rows when a row has no available actions.

For example, the Leek UI Plus plugin provides sticky actions:
https://filamentphp.com/plugins/leek-ui-plus

However, when a row has no available actions, an empty actions column is still rendered. These new CSS classes make it possible to hide the actions column when actions are not available for all rows.

## Visual changes

None

## Functional changes


danharrin:
Hmm, please let me know if I am missing something. But this seems like a feature introduced to fix a singular specific problem, rather than just fixing the problem in the first place? Can we not fix the problem instead?

webard:
@danharrin This problem can’t really be solved without marking the relevant <td> elements with appropriate CSS classes.

This PR doesn’t introduce any behavioral changes. It only adds a few CSS classes that allow external packages to apply `display: none` to the table header and the corresponding table cells when the actions column isn’t needed.

danharrin:
I mean, the CSS classes are being added without any behaviour in this PR though, that is my problem. If you had submitted a PR with new CSS classes that had behaviour attached to them, that would be a bug fix in this case, rather than a new feature to add classes for one task?

webard:
I didn’t want to change the existing behavior.

However, do you think it would be acceptable to hide both the actions header cell and the corresponding actions cells in rows when no actions are available, directly in Filament? That would solve the problem completely, regardless of whether an external package is being used.

danharrin:
Yes, I even believe that the entire actions column is already being hidden when there are no row actions, right? This is just an issue with conditional visibility?

danharrin:
Ideally, we should solve this in PHP by not rendering the col at all if all rows dont have actions in that cell, rather than CSS

webard:
That makes sense. I’ll open a new PR with that approach instead. Thanks!

PR #20133, opened 2026-07-10
Feature: hide actions column when no actions are available for any row

webard:
## Description

This PR prevents the actions column from being rendered when no actions are available for any row.

This is an improved version of #20130, implementing the solution by preventing the column from being rendered in PHP rather than relying on additional CSS classes.

## Visual changes

The empty actions column is no longer rendered when no actions are available for any row.

## Functional changes


danharrin (approved):

danharrin:
@webard I pushed a commit to reduce memory usage by avoiding computing the visible actions once we know that at least one row has visible actions, let me know what you think and if you have any problems while testing it

webard:
I’ll test it, thanks!

PR #20139, opened 2026-07-11
Feature: place filters in multiple locations

### LINKED ISSUE #14956: Investigate adding filters in multiple places on a table at once

webard:
## Description

This Pull Request introduces the ability to place filters in multiple locations and closes #14956.

Filter placement is configured using the `placement()` method available on the filter object, which accepts the existing `FiltersLayout` enum as its argument.

```php
TernaryFilter::make('email_verified_at')
    ->label('Email verified')
    ->nullable()
    ->placement(FiltersLayout::AboveContent)
```

The implementation has been designed to maintain full backward compatibility.

Intentional design decisions:
- `Apply filters` button applies filters from all placements, regardless of which placement’s button is clicked.
- Reset filters inside a specific placement resets only the filters belonging to that placement.
- The Reset filters button ("X" button above the table) resets all filters across all placements.
- `Modal` and `Dropdown` placements display two filter icons. I believe this is acceptable, as developers can either assign different icons to each placement or simply use only one of the placement.
- The badge displayed on the filter icons for `Modal` and `Dropdown` placements shows only the number of active filters within that specific placement.

This PR is still in draft because one important question remains: how should each placement be configured individually? I currently see two possible approaches.

**1. Extend the existing configuration methods**
Methods such as `filtersFormColumns`, `filtersApplyAction`, etc. could accept an additional `placement` argument. This would allow the same method to be called multiple times, configuring a different placement each time:
```php
$table
    ->filtersFormColumns(3)
    ->filtersFormColumns(3, placement: FiltersLayout::AboveContent);
```

**2. Introduce a dedicated configuration method**
```php
$table->filtersFormPlacementConfiguration(
    FiltersLayout::AboveContent,
    fn (FiltersFormPlacementConfiguration $config) => $config->columns(4),
);
```

I think the second approach is cleaner. It would also make it possible, in a future major release, to deprecate the table-level configuration methods and configure placements exclusively through `filtersFormPlacementConfiguration()`.

Placement-specific configuration would take precedence over the global configuration. For example:
```php
$table->filtersLayout(FiltersLayout::Dropdown)
    ->filtersFormColumns(3)
    ->filtersFormPlacementConfiguration(
        FiltersLayout::Dropdown,
        fn (FiltersFormPlacementConfiguration $config) => $config->columns(4),
    );
```

In this case, the filter form displayed in the dropdown would use 4 columns, not 3.

Before I proceed with the implementation, I’d like to know whether this Pull Request has a realistic chance of being merged into the main branch.

Personally, I believe this is a very useful feature. It makes it much easier to organize tables with a large number of filters by keeping the most important filters visible while moving less frequently used ones into a dropdown or modal.

The implementation was developed with the assistance of AI and has been carefully reviewed and tested.

## Functional changes


danharrin:
Thanks for taking a proper run at this, and for stopping to ask about API direction instead of ploughing on. The unknowns were always tricky, which is why I closed #14956 in the first place.

The good bits: leaving the state model alone (one filters form, one `tableFilters` state path, placement as pure presentation) is the right foundation, and the backward compatibility story holds up.

## What I'd merge

A `FilterPanel` container object, passed straight into `filters()`:

```php
use Filament\Tables\Enums\FiltersLayout;
use Filament\Tables\Filters\FilterPanel;

$table
    ->filters([
        FilterPanel::make(FiltersLayout::AboveContent, [
            SelectFilter::make('status'),
            Filter::make('is_published'),
        ])
            ->columns(4),

        FilterPanel::make(FiltersLayout::Dropdown, [
            SelectFilter::make('category'),
            SelectFilter::make('author'),
        ]),
    ]);
```

Rules:

1. `make()` takes the location and the filters. Both required. A panel is meaningless without either, so neither is optional or deferred to a fluent call (`->filters()` exists, `make()` just passes through to it).
2. All or nothing. If one item in the `filters()` array is a `FilterPanel`, every item must be. Mixing panels and loose filters throws, including via `pushFilters()`, so a plugin appending filters to a panel-mode table registers its own panel. The panel is a logical partition of all filters, not a modifier for some of them, and this kills every "implicit default panel" question (where it renders, trigger order, what it inherits) before it exists. Not a breaking change, since mixed arrays were never valid anyway.
3. No per-filter `placement()` method. The array structure is the single source of truth for where a filter lives. Keeping both invites a real ambiguity: if a filter points at a location that also has a configured panel, which one does it join? Sugar can be added later if people want it. It can't be removed once shipped.
4. Your defaults idea stays, that part of your proposal is good. Configuration resolves panel method → table-level `filtersForm*` method → per-layout default. The table-level methods are never deprecated; they permanently mean "defaults for every panel". So `FilterPanel` gets nullable counterparts for the relevant options (`columns()`, `width()`, `maxHeight()`, `resetActionPosition()`, `triggerAction()`), each falling through when unset.
5. Normalise internally so there's only one rendering model. A flat `filters()` array becomes a single implicit `FilterPanel` at the table's `filtersLayout()` before rendering, so the view layer only ever sees a list of panels. One code path, no flat-versus-grouped branching. `filtersLayout()` keeps its meaning purely through that normalisation.

Your instincts here were right and I'd keep them: filter indicators, the "remove all" button, and the deferred-filters apply button all stay global. Filter state is a single Livewire property, so a per-panel apply button would lie about what actually gets submitted.

I know this is a bigger lift than either of your proposals. The container class needs flattening so everything keyed by filter name keeps working, and the table view restructures around a panel loop. But every cheaper option locks in the enum-bucket model for good, and if the feature is worth doing, this is the version I'm willing to maintain. If you're up for reworking the PR in this direction I'll gladly review it. If not, fair enough, this comment can stand as the design record for the future.

Issue #20140, opened 2026-07-11
Table toolbar configuration API


danharrin:
Note: I am not looking for a community member to build this feature, it will be handled by myself or another member of the Filament core team when we have availability.

PR #20150, opened 2026-07-11
Feature: FilterPanel component to place filters in multiple locations

### LINKED ISSUE #14956: Investigate adding filters in multiple places on a table at once

webard:
## Description

This Pull Request introduces the ability to place filters in multiple locations and closes #14956.

It is a rework of the original Pull Request #20139. Thanks @danharrin for taking the time to think through the API properly.

The PR introduces a new `FilterPanel` component. Its first argument accepts a `FiltersLayout`, and the second argument is an array of filters that should be displayed in that location:

```php
$table
    ->filters([
        FilterPanel::make(FiltersLayout::AboveContent, [
            SelectFilter::make('status'),
            SelectFilter::make('category'),
        ])->columns(2),

        FilterPanel::make(FiltersLayout::Dropdown, [
            SelectFilter::make('author'),
        ]),
    ]);
```

FilterPanel exposes the same configuration methods that are currently available on the table via the `filtersForm*` / `filters*` methods:

- `columns()`
- `width()`
- `maxHeight()`
- `resetActionPosition()`
- `triggerAction()`

`filtersApplyAction()` was intentionally not included, as the Apply action behaves globally regardless of which filter panel it is triggered from.

If a setting is not explicitly overridden, it falls back to the global table configuration.

**Intentional design decisions**
- `Apply filters` button applies filters from all panels, regardless of which panel’s button is clicked.
- Reset filters within a panel resets only the filters belonging to that panel (this behavior has been preserved from the first version).
- The Reset filters button (the “X” button above the table) resets all filters across all panels.
- `Modal` and `Dropdown` panels display two filter icons. I believe this is acceptable, as developers can assign different icons to each panel or simply use only one of them.
- The badge displayed on the filter icons for `Modal` and `Dropdown` panels shows only the number of active filters within that panel.
- Creating multiple `FilterPanel` instances for the same `FiltersLayout` automatically merges their filters.
- `pushFilters()` also merges filters into the specified panel.
- Mixing `FilterPanel` instances with standard filters is intentionally not supported. The following code will throw `LogicException`:
```php
$table
    ->filters([
        FilterPanel::make(FiltersLayout::AboveContent, [
            SelectFilter::make('status'),
            SelectFilter::make('category'),
        ])->columns(2),

        SelectFilter::make('author'),
    ]);
```

Tested and verified - everything works as expected. Backward compatibility is preserved.

## Functional changes

### DIFF
--- a/packages/tables/docs/03-filters/06-layout.md
+++ b/packages/tables/docs/03-filters/06-layout.md
@@ -245,9 +245,30 @@ The `Dropdown` and `Modal` locations each render their own trigger, so a table w
 
 Each panel has its own reset action that clears only the filters shown in that panel. To clear every filter at once, use the "remove all" button in the active filter indicators above the table. The apply button (when deferring filters) and the indicators are always global - filter state is a single value, so applying always submits every panel's filters.
 
-### Adding filters to a location that already has a panel
+### Merging panels that share a location
 
-If a panel targets a location that another panel already occupies - for example a plugin appending `FilterPanel::make(FiltersLayout::Dropdown, [...])` to a table that already has a dropdown panel — its filters are **merged into the existing panel**, keeping the existing panel's configuration. This lets extensions add filters to a shared location without conflicting.
+Declaring more than one panel for the same location - whether twice in the same `filters()` array, or by appending one later with `pushFilters()` (for example from a plugin) — **merges their filters into a single panel** for that location. The configuration of the **first** panel at that location wins; any configuration set on the later panels is ignored:
+
+```php
+use Filament\Tables\Enums\FiltersLayout;
+use Filament\Tables\Filters\FilterPanel;
+use Filament\Tables\Filters\SelectFilter;
+
+$table
+    ->filters([
+        FilterPanel::make(FiltersLayout::AboveContent, [
+            SelectFilter::make('status'),
+        ])->columns(2),
+
+        // The same location again: its filters are merged into the panel above,
+        // and its own configuration (such as `columns()`) is ignored.
+        FilterPanel::make(FiltersLayout::AboveContent, [
+            SelectFilter::make('category'),
+        ]),
+    ]);
+```
+
+This lets a plugin add filters to a shared location (such as the dropdown) without conflicting with the table's own panels.
 
 <Aside variant="danger">
     A [custom filter form schema](#customizing-the-filter-form-schema) (`filtersFormSchema()`) cannot be combined with `FilterPanel` objects, as they are competing layout mechanisms. Use `filtersFormSchema()` with a flat `filters()` array only.
--- a/packages/tables/docs/03-filters/06-layout.md
+++ b/packages/tables/docs/03-filters/06-layout.md
@@ -226,24 +226,67 @@ FilterPanel::make(FiltersLayout::Dropdown, [
     ->triggerAction(fn (Action $action): Action => $action->label('Filter'));
 ```
 
-### Using the dropdown and a modal together
-
-The `Dropdown` and `Modal` locations each render their own trigger, so a table with a panel in both shows two filter icons — one opening a dropdown, the other opening a modal:
+Each option also accepts a closure, which can inject `$table` and `$livewire` like anywhere else in Filament:
 
 ```php
+use Filament\Tables\Contracts\HasTable;
+use Filament\Tables\Enums\FiltersLayout;
+use Filament\Tables\Filters\FilterPanel;
+
+FilterPanel::make(FiltersLayout::AboveContent, [
+    // ...
+])->columns(fn (HasTable $livewire): int => $livewire->isTableReordering() ? 1 : 3);
+```
+
+<Aside variant="info">
+    `triggerAction()`, `width()` and `maxHeight()` only apply to a panel that opens - see [supported location combinations](#supported-location-combinations). A panel that is always visible, such as `AboveContent`, ignores them.
+</Aside>
+
+### Supported location combinations
+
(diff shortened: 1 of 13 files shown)


webard:
I’m not entirely sure about the behavior of the Reset action. Wouldn’t it be better for it to reset all filters, regardless of which location it’s triggered from? Let me know if you think Reset should reset filters globally, and I’ll change it.

I’m also not entirely convinced about the number of additional variables in `index.blade.php`, but it seems like the clearest approach without hiding the logic.

danharrin:

danharrin (changes requested):
Thanks for your work here.

I do not think we need to support arbitrary combinations of `FiltersLayout`. Please enforce and document this rule:

- A table may have at most one interactive filter location:
  - `Dropdown`
  - `Modal`
  - `AboveContentCollapsible`
  - `BeforeContent`
  - `BeforeContentCollapsible`
  - `AfterContent`
  - `AfterContentCollapsible`
- That location may be combined with `AboveContent`, `BelowContent`, and `Hidden`.
- `AboveContent` cannot be combined with `AboveContentCollapsible`, since they occupy the same physical location.
- Multiple panels using the same exact layout may continue to merge.

Treat before/after layouts as interactive even when they are not explicitly collapsible, since they use the filter trigger to open as floating panels on mobile. This intentionally means that combinations such as dropdown + modal, before + after, and dialog + side filters are unsupported. Please throw a clear `LogicException` for unsupported combinations rather than silently ignoring a panel.

There are a few implementation issues to address within that supported subset:

- Resolve collapsibility from the relevant panel rather than using one global `$hasCollapsibleFilters` value. Currently, for example, `BeforeContentCollapsible` also makes an ordinary `AboveContent` panel collapsible.
- Wire the panel configuration through to the rendered UI. `triggerAction()` is currently never applied, `resetActionPosition()` resolves correctly but the view still uses the table-level value, and before/after panels can use the width of the first registered panel instead of their own.
- Handle conditionally hidden filters. If every filter in the panels is hidden, `getFilterPanels()` returns an empty array and the form configuration accesses index `0`, causing an exception. Panel badges also count hidden filters even though the table does not apply them.
- Keep the filters argument to `FilterPanel::make()` required. An empty panel is not meaningful.
- Bind each panel to its table so closure-based panel configuration supports the normal `$table` and `$livewire` injections.

With deferred filters, resetting one panel currently calls `applyTableFilters()`, which applies pending changes from every other panel. After resetting the panel’s child schema, copy only that panel’s filter keys from `tableDeferredFilters` into `tableFilters`, leave the other deferred values untouched, and run the normal filter update handling. The Apply action should remain global.

Please add behavioral tests for:

- accepted and rejected layout combinations, including mobile side-panel behavior;
- panel configuration affecting rendered output rather than only its getters;
- all-hidden panels and hidden-filter badge counts;
- resetting one deferred panel without applying another panel’s pending changes.

Once those are addressed, please rebase onto the current `4.x` branch. The overall direction is good; I just want the supported layout contract and these public behaviors settled before we ship it.

And if you think that the direction I stated about unsupported panel combinations means that this PR is no longer useful to most apps, please let me know.

webard:
Thanks for the feedback - everything has been implemented.

I’d suggest allowing `Dropdown` + `Modal` as a valid combination, since they don’t conflict technically and can now have distinct triggers. I know it’s a somewhat unusual combination, but I believe developers may find a use case for it.
 
I’d also suggest merging `AboveContent`and `AboveContentCollapsible` into a single section, with the collapsible filters revealed inline next to the always-visible ones.
 
Next to the “Apply filters” button, we could add a “Show more filters” button-link that expands the list of filters. If any filter from the expanded list is applied, the list would be expanded by default.

Please let me know what you think.

danharrin:
What do you think about the fact that clicking "apply" on one panel applies all panels? It seems a bit problematic for me from a behavioural standpoint, but I don't see a straightforward fix either that won't have edge cases. It might be a showstopper for me, to be honest.

webard:
Actually, I thought it felt more natural. When deferred filters are enabled, I can set filters in multiple places and then click "Apply" wherever I happen to finish setting them. To be honest, that felt more natural to me during testing.

danharrin:
I am very sorry as I know you worked hard on this, but I think that is a deal breaker for me. If you make changes to filters in one location they should not carry unsaved changes from other locations. I am certain that this would be reported as a bug if we released the feature. I also don't think there is a way around that limitation unfortunately.

webard:
I’m a little disappointed because:

1. There are several people who are waiting for this.
2. The feature is completely opt-in and doesn’t change the default behavior in any way.
3. The behavior of the “Apply” button is a design decision that can be documented.
4. There is no way to achieve similar behavior through a plugin.

But I accept your decision. Thank you for all the reviews you’ve done.

danharrin:
I understand that the feature is opt-in and doesn't change default behaviour. I think the only way we will be able to support this is if we separate the state of each filter panel so that each has its own deferred and applied state path separate and schema. Also makes resetting easier. If we introduce this feature in the current shape, changing it in the future to fix the broken behaviour (by changing the state paths of separate filter panels) could force us into introducing breaking changes. This isn't an easy change to make though, and I need some other things to ship first before I can be confident enough to do that. I might even save it for 6.x.

I am also going to personally work on #20140 which will allow us to correctly ship multiple filter buttons in the toolbar, unlocking the ability to have multiple filter panels using the same layout mode (like 2 dropdown filter panels) in the future, and removing the constraints I asked for around certain layouts not being compatible with others.

I'm still keeping the issue open as this feature is something I eventually want to incorporate, but I don't think it's the right time.

I did feel bad when I closed this PR actually, and wrote down my thoughts earlier today https://x.com/danjharrin/status/2098357785637290170

Thanks again

PR #20152, opened 2026-07-12
feat:Add HasRenderHooks concern, enabling Pages, Widgets, RelationManagers, and Resources to declare their own render hooks

ahmed-rashad-alnaggar:
## Summary

Adds a `Filament\Livewire\Concerns\HasRenderHooks` trait that lets a Filament
Livewire component declare its own render hooks directly on the class,
automatically scoped to itself, instead of registering them externally from
a service provider or middleware.

- `Filament\Pages\BasePage` uses it directly.
- `Filament\Widgets\TableWidget` uses it directly.
- `Filament\Resources\RelationManagers\RelationManager` uses it directly.
- `Filament\Resources\Resource` is not a Livewire component, so it gets a
  separate, static-only counterpart:
  `Filament\Resources\Resource\Concerns\HasRenderHooks`, exposing a static
  `getRenderHooks()` with no booting logic.
- `Filament\Resources\Pages\Page` bridges the two by overriding
  `bootHasRenderHooks()` to merge its resource's static hooks with its own
  instance hooks before registering them, all scoped to the page class:

  ```php
  public function bootHasRenderHooks(): void
  {
      $renderHooks = array_merge(static::getResource()::getRenderHooks(), $this->getRenderHooks());

      foreach ($renderHooks as $name => $hook) {
          FilamentView::registerRenderHook($name, $hook, static::class);
      }
  }
  ```

  Because the page's own hooks are merged in second, a hook name declared on
  both the resource and the page resolves in favor of the page — the more
  specific declaration wins.

## Motivation

Registering a render hook scoped to a specific page, widget, relation
manager, or resource currently requires a service provider call:

```php
    PanelsRenderHook::PAGE_START,
    fn (): View => view('warning-banner'),
    scopes: EditUser::class,
);
```

This works, but separates the hook from the class it belongs to. Plugin
authors and anyone shipping reusable pages, widgets, relation managers, or
resources benefit from declaring hooks alongside the class itself, the same
way `getHeaderActions()` or `getHeaderWidgets()` already work.

## Usage

On a page, table widget, or relation manager:

```php
use Filament\View\PanelsRenderHook;

class EditUser extends EditRecord
{
    protected function getRenderHooks(): array
    {
        return [
            PanelsRenderHook::PAGE_START => fn (): View => view('warning-banner'),
        ];
    }
}
```

On a resource:

```php
use Filament\Resources\Resource;
use Filament\View\PanelsRenderHook;

class UserResource extends Resource
{
    public static function getRenderHooks(): array
    {
        return [
            PanelsRenderHook::PAGE_START => fn (): View => view('resource-banner'),
        ];
    }
}
```

Hooks declared on `UserResource` are automatically picked up by every page
belonging to that resource, and registered scoped to each individual page
class. If a page declares a hook under the same name, the page's hook takes
precedence over the resource's for that page.

### DIFF
--- a/docs/09-advanced/01-render-hooks.md
+++ b/docs/09-advanced/01-render-hooks.md
@@ -257,6 +257,76 @@ FilamentView::registerRenderHook(
 );
 ```
 
+## Registering render hooks from within a class
+
+Instead of registering a scoped render hook from a service provider, you can
+declare it directly on the class it belongs to, using the
+`Filament\Livewire\Concerns\HasRenderHooks` trait. This trait is already
+included by default on:
+
+- Pages
+- Table widgets
+- Relation managers
+
+To use it, override `getRenderHooks()`, returning an array that maps render
+hook names to their callbacks:
+
+```php
+use Filament\View\PanelsRenderHook;
+use Illuminate\Contracts\View\View;
+
+class EditUser extends EditRecord
+{
+    protected function getRenderHooks(): array
+    {
+        return [
+            PanelsRenderHook::PAGE_START => fn (): View => view('warning-banner'),
+        ];
+    }
+}
+```
+
+Each hook registered this way is automatically [scoped](#scoping-render-hooks)
+to the declaring class, so you don't need to pass a `scopes` argument
+yourself. This is functionally equivalent to calling:
+
+```php
+FilamentView::registerRenderHook(
+    PanelsRenderHook::PAGE_START,
+    fn (): View => view('warning-banner'),
+    scopes: EditUser::class,
+);
+```
+
+### Registering render hooks on a resource
+
+`Filament\Resources\Resource` is not a Livewire component, so instead it uses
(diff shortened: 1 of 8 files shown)


danharrin:

ahmed-rashad-alnaggar:
@danharrin Just wanted to add a quick note (not trying to nag or pressure you at all!) — I think this one should be pretty low-risk to slot in, since it's in a different area than what you're currently focused on stabilizing for the release, so it shouldn't interact with or introduce any issues in that scope.

No worries either way if it doesn't make it in time — happy for it to wait for review whenever you have bandwidth. Just thought it might be worth a look if there's a spare moment before the cut-off, since it seems independent of the current beta work.

danharrin:
Thanks for the contribution. Unfortunately, render hooks are stored in a request-wide registry by class scope, while this registers them per Livewire component instance. Multiple instances of the same widget class would therefore share and duplicate each other’s hooks. Fixing that would require a different, instance-local design, so I’m going to close this PR. Thanks again for exploring the idea.

PR #20153, opened 2026-07-12
feat: Add strict hierarchical breadcrumbs option

ahmed-rashad-alnaggar:
## Summary

Closes discussion #19515.

Adds an opt-in **strict hierarchical breadcrumbs** mode, configurable per-panel via `Panel::breadcrumbs()`:

```php
$panel->breadcrumbs(strictHierarchical: true);
```

By default, breadcrumbs for a page are built from its resource and parent record relationships (e.g. an `EditRecord` page's breadcrumbs come from its resource's list/view pages, plus any parent resources for nested resources). When strict hierarchical mode is enabled, breadcrumbs are instead (additionally) built by walking the page's full navigational hierarchy — cluster ancestry, navigation group, and navigation parent item — and combining that chain with the existing resource/parent-record breadcrumbs.

## Changes

- **`Filament\Panel\Concerns\HasBreadcrumbs`**: extended `breadcrumbs()` with a second `bool|Closure $strictHierarchical = false` parameter, and added `hasStrictHierarchicalBreadcrumbs()` alongside the existing `hasBreadcrumbs()`.
- **`Filament\Pages\Page`**: added `$breadcrumb` property, `getBreadcrumb()`, and `getHierarchicalBreadcrumbs()`, which recursively resolves cluster, navigation group, and navigation parent item breadcrumbs for a page. `getBreadcrumbs()` now checks `Filament::getCurrentOrDefaultPanel()->hasStrictHierarchicalBreadcrumbs()` and prepends the hierarchical chain when enabled.
- **`Filament\Resources\Pages\Page`**: `getResourceBreadcrumbs()` now checks strict mode first and, when enabled, prepends `static::getHierarchicalBreadcrumbs()` ahead of the existing resource/parent-record breadcrumbs; falls back to the previous cluster-only behavior otherwise.

## Behavior

- **Strict mode disabled (default):** unchanged — breadcrumbs resolve exactly as before, via cluster + resource/parent-record relationships.
- **Strict mode enabled:** breadcrumbs are `[...hierarchical chain, ...resource/parent-record breadcrumbs, page's own breadcrumb]`. The hierarchical chain and resource/record breadcrumbs are combined, not one replacing the other.

### DIFF
--- a/docs/06-navigation/01-overview.md
+++ b/docs/06-navigation/01-overview.md
@@ -639,6 +639,25 @@ public function panel(Panel $panel): Panel
 }
 ```
 
+## Strict hierarchical breadcrumbs
+
+By default, breadcrumbs for a page are built from its resource and parent record relationships - for example, an `EditRecord` page's breadcrumbs are derived from its resource's list and view pages, and any parent resources in the case of nested resources.
+
+If you'd prefer breadcrumbs to reflect the full navigational hierarchy of a page instead - including its [cluster](clusters), [navigation group](#grouping-navigation-items), and [navigation parent item](#grouping-navigation-items-under-other-items) - you can enable strict hierarchical breadcrumbs in your [configuration](../05-panel-configuration):
+
+```php
+use Filament\Panel;
+
+public function panel(Panel $panel): Panel
+{
+    return $panel
+        // ...
+        ->breadcrumbs(strictHierarchical: true);
+}
+```
+
+When enabled, a page's breadcrumbs will be built by walking up through its cluster, its navigation group, and its navigation parent item, before falling back to the page's own breadcrumb. This is particularly useful if you rely heavily on [navigation parent items](#grouping-navigation-items-under-other-items) to create deep, third-level navigation, and want the breadcrumb trail to mirror that structure exactly.
+
 ## Reloading the sidebar and topbar
 
 Once a page in the panel is loaded, the sidebar and topbar are not reloaded until you navigate away from the page, or until a menu item is clicked to trigger an action. You can manually reload these components to update them by dispatching a `refresh-sidebar` or `refresh-topbar` browser event.
--- a/packages/panels/src/Clusters/Cluster.php
+++ b/packages/panels/src/Clusters/Cluster.php
@@ -3,6 +3,7 @@
 namespace Filament\Clusters;
 
 use Filament\Facades\Filament;
+use Filament\Navigation\NavigationItem;
 use Filament\Pages\Page;
 use Filament\Pages\PageConfiguration;
 use Filament\Panel;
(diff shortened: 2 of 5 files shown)


ahmed-rashad-alnaggar:
Before this potentially gets merged, I wanted to float an idea for feedback.

Right now `strictHierarchical` only affects breadcrumbs. I was thinking it might make sense to generalize this into a panel-wide "strict hierarchical navigation" concept, rather than a breadcrumbs-only flag — so the same setting also governs URL/slug generation (e.g. nesting a resource's slug under its navigation parent item / group / cluster), keeping breadcrumbs and slugs consistent with each other instead of having them independently configurable and potentially out of sync.

Something like:

```php
$panel->strictHierarchicalNavigation(true);
```

which `hasStrictHierarchicalBreadcrumbs()` (and a future slug-generation check) would both read from, instead of breadcrumbs having its own separate toggle.

Wanted to raise this in case it changes the shape of the API here — happy to rework this PR to fit that direction if it seems worth pursuing, or keep it scoped to breadcrumbs only if slugs should stay a separate concern. Let me know what you think.

danharrin:

ahmed-rashad-alnaggar:
Converting to draft to fix bugs regarding clusters and sub navigation.

ahmed-rashad-alnaggar:
Fixed!

danharrin (changes requested):
Thanks! Please add tests for the breadcrumb behavior introduced here before we review the implementation further. At minimum, cover:

- Existing breadcrumb behavior when strict mode is disabled.
- Custom pages and resource pages with navigation groups and parent items.
- Pages inside clusters, including clusters with sub-navigation disabled.
- Panels using a custom `NavigationBuilder`.
- Grouped pages that do not register a navigation item.

The implementation should gracefully fall back to the page’s existing breadcrumbs when the corresponding navigation group or item cannot be found.

I tested the latter cases against the current implementation. A custom `NavigationBuilder` currently throws `Undefined array key "N;"`, disabled cluster sub-navigation throws `Undefined array key 0`, and an unregistered grouped page throws for its serialized group key. Adding these tests should reproduce those failures.

ahmed-rashad-alnaggar:
@danharrin Sorry I was too busy recently that I didn't manage to look into this.

I am still busy right now, but I think I can have time to also support strict slug/URL as it is really a more professional looking feature, as I said here:

> Before this potentially gets merged, I wanted to float an idea for feedback.
> 
> Right now `strictHierarchical` only affects breadcrumbs. I was thinking it might make sense to generalize this into a panel-wide "strict hierarchical navigation" concept, rather than a breadcrumbs-only flag — so the same setting also governs URL/slug generation (e.g. nesting a resource's slug under its navigation parent item / group / cluster), keeping breadcrumbs and slugs consistent with each other instead of having them independently configurable and potentially out of sync.
> 
> Something like:
> 
> ```
> $panel->strictHierarchicalNavigation(true);
> ```
> 
> which `hasStrictHierarchicalBreadcrumbs()` (and a future slug-generation check) would both read from, instead of breadcrumbs having its own separate toggle.
> 
> Wanted to raise this in case it changes the shape of the API here — happy to rework this PR to fit that direction if it seems worth pursuing, or keep it scoped to breadcrumbs only if slugs should stay a separate concern. Let me know what you think.

Should I?

danharrin:
@ahmed-rashad-alnaggar lets stick to this breadcrumbs feature for now, I consider this part of Filaments "navigation" system, but allowing a navigation group to affect the URL is kinda out of the realm of navigation IMO. I might change my mind in the future but I don't think it fits.

danharrin (approved):

PR #20165, opened 2026-07-12
feat(tables): persist grouping in the user's session

IGedeon:
## Description

Tables can already persist their sort, search, filters and columns in the user's session, but not their grouping. This adds a `persistGroupInSession()` method to the table configuration so a user's selected grouping is restored on their next visit, mirroring the existing `persistSortInSession()`:

```php
{
    return $table
        ->groups([
            'status',
            'category',
        ])
        ->persistGroupInSession();
}
```

While implementing this I found that the `updatedTableGroupColumn()` Livewire hook was dead code: the underlying property is `$tableGrouping`, so Livewire resolves the hook as `updatedTableGrouping()` and the old name never fired. Renaming it fixes that dormant hook (grouping changes now also reset the pagination page correctly) and is what enables writing the grouping to the session.

## Visual changes

None. The behavior is opt-in and there is no UI change.

## Functional changes

### DIFF
--- a/packages/tables/docs/07-grouping.md
+++ b/packages/tables/docs/07-grouping.md
@@ -60,6 +60,24 @@ public function table(Table $table): Table
 }
 ```
 
+### Persisting the grouping in the user's session
+
+To persist the grouping in the user's session, use the `persistGroupInSession()` method:
+
+```php
+use Filament\Tables\Table;
+
+public function table(Table $table): Table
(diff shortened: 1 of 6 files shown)


danharrin (changes requested):
Please rebase this on 4.x and force push

danharrin (approved):

PR #20176, opened 2026-07-13
feat: Implement empty state behavior into stats

ahmed-rashad-alnaggar:
## Summary

Adds an `emptyState()` method to `Stat`, letting you define a fallback value/content to render when a stat's `value` resolves to `null`.

```php
Stat::make('Unique views', $uniqueViews)
    ->emptyState('-')
```

The blade view now does:

```blade
{{ $getValue() ?? $getEmptyState() }}
```

## Why not just handle this with `value()`?

You *can* already get the same visual result today without this PR, just by doing the null-coalescing yourself:

```php
Stat::make('Unique views', null)
    ->value($uniqueViews ?? '-')
```

That works, but it means every stat needs this handled ad-hoc at the call site. Adding a dedicated `emptyState()` method:

- separates "what the value is" from "what to show when there's no value," which reads more clearly
- can be configured **globally** via `Stat::configureUsing()`, so an app can set one default empty-state placeholder (e.g. `'-'` or `'N/A'`) for every stat without repeating `?? '-'` everywhere

```php
use Filament\Widgets\StatsOverviewWidget\Stat;

Stat::configureUsing(function (Stat $stat): void {
    $stat->emptyState('-');
});
```

That global-default use case isn't really reachable with the manual `value($x ?? '-')` approach, since it has to be repeated per-stat rather than configured once.


danharrin (changes requested):
I don't think I would accept this as a useful new method considering it is already possible (albeit without the global config capabilities).

However, you could reframe it as a new `placeholder()` method which does the same but formats it with a lighter gray text color?

ahmed-rashad-alnaggar:
Done!

ahmed-rashad-alnaggar:
Just wanted to add a quick note (not trying to nag or pressure you at all!) — I think this one should be pretty low-risk to slot in, since it's in a different area than what you're currently focused on stabilizing for the release, so it shouldn't interact with or introduce any issues in that scope.

No worries either way if it doesn't make it in time — happy for it to wait for review whenever you have bandwidth. Just thought it might be worth a look if there's a spare moment before the cut-off, since it seems independent of the current beta work.

danharrin (approved):

PR #20180, opened 2026-07-14
fix: use dehydrated settings for query builder constraints

people-sea:
## Description
Query builder constraints bypass schema state casts when applying raw settings, causing incorrect datetime filters when the Filament and app timezones differ.

Use dehydrated settings for queries while retaining raw settings for summaries.

### Before

image

### After

image

## Functional changes

### DIFF
--- a/packages/tables/src/Filters/QueryBuilder.php
+++ b/packages/tables/src/Filters/QueryBuilder.php
@@ -138,6 +138,7 @@ class QueryBuilder extends BaseFilter
                 function (Operator $operator) use ($ruleIndex, &$summaries): void {
                     $summaries[$ruleIndex] = $operator->getSummary();
                 },
+                shouldUseRawSettings: true,
             );
         }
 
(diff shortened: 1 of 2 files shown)


danharrin (changes requested):
I don't think state snapshot existed when I first built the query builder, but now that it does I probably would have used it from the start, should we consider making this the default behaviour without a param? Or is there a case when we cant use it?

people-sea:
> I don't think state snapshot existed when I first built the query builder, but now that it does I probably would have used it from the start, should we consider making this the default behaviour without a param? Or is there a case when we cant use it?

Yeah, making it the default makes sense. I think summaries may be one exception: with deferred filters, the schema can contain changes that haven't been applied yet, while the summary should reflect the applied state.

Would it make sense to use snapshots by default and keep raw settings for summaries?

danharrin (approved):

PR #20191, opened 2026-07-16
Fix action inside livewireProperty tab never running

### LINKED ISSUE #20190: Actions inside Tab with `livewireProperty` are broken
### Package

filament/filament

### Package Version

v4.11.8

### Laravel Version

v13.20.0

### Livewire Version

_No response_

### PHP Version

v8.4.20

### Problem description

A Filament `Action` placed **inside a `Tabs::make()->livewireProperty(...)` tab** never executes its `->action()` closure. The action is silently skipped: no download is returned, no modal opens, nothing happens. The **same action placed at the top level** (outside the tabs) works correctly.

### Expected behavior

Action working fine.

### Steps to reproduce

Place any action inside Tabs with `livewireProperty()`

https://github.com/pxlrbt/filament-action-issue

### Relevant log output

https://github.com/user-attachments/assets/693fd18a-4ac8-4f73-92f1-cfb387ebec72

pxlrbt:
**DISCLAIMER: As you can see Claude did find this fix. I am not too sure about the internals here, so please check whether it's the best approach.**

## Description

An action placed inside a `Tabs::make()->livewireProperty(...)` tab never executed its `->action()` callback: no modal opened, no download was returned, nothing happened. The same action at the top level worked.

## Changes

Two things combined to cause it:

1. A livewireProperty tab is keyed by its array key only at render time (`toEmbeddedHtmlForLivewireProperty()`). During `mountAction()` — which resolves the action before the schema renders — the tab still carried its auto-generated key, so `Schema::getAction()` could not resolve the nested schemaComponent key (e.g. `content.first`) and returned null.

2. `HasKey::key()` did not flush the cached absolute key, so once a tab's absolute key had been cached, re-keying it had no effect on `getKey(isAbsolute: true)` — the value the action button's schemaComponent is built from. The button then emitted a stale, unresolvable key.

Fix both: key each tab with its array key when the child schema is built (`configureChildSchema()`), so resolution works at mount time as well as at render time; and flush the cached absolute keys in `key()` so a re-key always takes effect.

Fixes https://github.com/filamentphp/filament/issues/20190

### DIFF
--- a/packages/schemas/src/Components/Concerns/HasKey.php
+++ b/packages/schemas/src/Components/Concerns/HasKey.php
@@ -23,6 +23,9 @@ trait HasKey
         $this->key = $key;
         $this->isKeyInheritable = $isInheritable;
 
+        $this->flushCachedAbsoluteKey();
+        $this->flushCachedAbsoluteInheritanceKey();
+
         return $this;
     }
 
--- a/packages/schemas/src/Components/Tabs.php
+++ b/packages/schemas/src/Components/Tabs.php
@@ -7,6 +7,7 @@ use Filament\Schemas\Components\Concerns\CanPersistTab;
 use Filament\Schemas\Components\Concerns\HasLabel;
 use Filament\Schemas\Components\Tabs\Tab;
 use Filament\Schemas\Contracts\HasRenderHookScopes;
+use Filament\Schemas\Schema;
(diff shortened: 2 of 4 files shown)


danharrin:
Nearly! #20192

pxlrbt:
Nice, thanks!

PR #20226, opened 2026-07-20
Fix stale repeatable entry items after relationship refresh

### LINKED ISSUE #20222: RepeatableEntry renders a deleted related row until a page refresh (regression in v5.7.0)
### Package

filament/infolists

### Package Version

v5.7.1

### Laravel Version

v13.20.0

### Livewire Version

v4.3.3

### PHP Version

8.5.8

### Problem description

A `RepeatableEntry` on a `ViewRecord` infolist keeps rendering a related row that an action deleted within the same request. The record is removed from the database, but its row stays in the rendered HTML until a full page refresh.

This regressed between `filament/*` v5.6.8 (correct) and v5.7.0 (stale), and is still present in v5.7.1. It bisects to #19550 ("Optimize large form rendering"), which added a `$cachedItems` memoization to `RepeatableEntry::getItems()`:

```php
// filament/infolists/src/Components/RepeatableEntry.php
protected ?array $cachedItems = null;

public function getItems(): array
{
    if ($this->cachedItems !== null) {
        return $this->cachedItems; // short-circuits before re-reading getState()
    }
    // ...builds the row containers from getState()...
    return $this->cachedItems = $containers;
}
```

After the action mutates the relation, `getItems()` returns the pre-action cache because `clearCachedChildSchemas()` (which resets `$cachedItems`) is not called on the entry during the post-action re-render within the same request. On v5.6.8 there was no cache, so `getState()` was re-resolved on every render. The stale row is in the visible DOM (a rendered `fi-in-text-item` cell), not only in a serialized `wire:snapshot`.

### Expected behavior

After the delete action, the deleted row is removed from the `RepeatableEntry` without a manual page refresh, as it was on v5.6.8.

### Steps to reproduce

1. `Post` hasMany `Comment`; a `ViewPost` infolist with `RepeatableEntry::make('comments')` and a per-row delete `Action` that runs `$record->delete(); $post->setRelation('comments', $post->comments()->get());`.
2. Open the View page for a post that has two comments.
3. Click the delete action on one comment and confirm.
4. The deleted comment's row is still rendered until the page is reloaded.

The reproduction repository includes a feature test proving it: running `php artisan test` passes on `filament/filament:5.6.8` and fails on `filament/filament:^5.7`.

https://github.com/byjesper/filament-repeatableentry-stale-repro

### Relevant log output

```shell
The final assertion in tests/Feature/RepeatableEntryStaleAfterDeleteTest.php fails on v5.7.0+
(the two expectations above it pass, confirming the row is deleted from the database and only
the rendered RepeatableEntry is stale):

Failed asserting that '<rendered ViewPost html>' does not contain 'Delete me without a refresh.'.
```

olipayne:
## Description

Fixes #20222.

`RepeatableEntry::getItems()` now compares the current state with the state used to build its memoized item schemas. This keeps the 5.7 memoization optimization for unchanged state while rebuilding the schemas when a loaded Eloquent relationship is refreshed after an action.

A regression test deletes a related record, refreshes the parent model, and verifies that the repeatable entry immediately reflects the remaining relationship items.

## Visual changes

None.

## Functional changes

- [x] Code style has been fixed with Pint and verified with Rector.
- [x] Documentation is up-to-date (no documentation changes are required).

## Tests

- `vendor/bin/pest --configuration=phpunit.sqlite.xml tests/src/Infolists/Components/RepeatableEntryTest.php --compact`
- `php -d memory_limit=1G vendor/bin/phpstan analyse packages/infolists/src/Components/RepeatableEntry.php --no-progress`

### DIFF
--- a/packages/infolists/src/Components/RepeatableEntry.php
+++ b/packages/infolists/src/Components/RepeatableEntry.php
@@ -31,6 +31,8 @@ class RepeatableEntry extends Entry implements HasEmbeddedView
      */
     protected ?array $cachedItems = null;
 
+    protected mixed $cachedItemsState = null;
+
     /**
      * Configure table columns for display
      *
@@ -66,13 +68,15 @@ class RepeatableEntry extends Entry implements HasEmbeddedView
      */
     public function getItems(): array
     {
-        if ($this->cachedItems !== null) {
+        $state = $this->getState() ?? [];
+
+        if (($this->cachedItems !== null) && ($this->cachedItemsState === $state)) {
             return $this->cachedItems;
         }
 
         $containers = [];
 
-        foreach ($this->getState() ?? [] as $itemKey => $itemData) {
+        foreach ($state as $itemKey => $itemData) {
             $container = $this
                 ->getChildSchema()
                 ->getClone()
@@ -87,6 +91,8 @@ class RepeatableEntry extends Entry implements HasEmbeddedView
 
             $containers[$itemKey] = $container;
         }
+
+        $this->cachedItemsState = $state;
 
         return $this->cachedItems = $containers;
     }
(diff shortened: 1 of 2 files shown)


danharrin:
More complete fix for 4.x too: #20230

PR #20227, opened 2026-07-21
feat(tables): allow summaries to be positioned at the top of the table

quint-pruijssers:
## Description

Table summaries are always rendered below the records. On long tables that means scrolling past every row before seeing the totals, while for many reporting tables the totals are exactly what users open the page for.

This adds a `position()` method to summarizers, together with a new `Filament\Tables\Enums\SummarizerPosition` enum (`Top` / `Bottom`), so a summary can be rendered directly below the column headers instead:

```php
use Filament\Tables\Enums\SummarizerPosition;

{
    return $table
        ->columns([
            TextColumn::make('views_count')
                ->summarize(
                    Sum::make()
                        ->position(SummarizerPosition::Top),
                ),
        ]);
}
```

Each summarizer decides individually where it is rendered, so some summaries can be shown above the records and others below on the same table. The summary section is now rendered in up to two places — directly below the column header row and at the bottom as today — and each section only appears when at least one visible summarizer targets that position. Both the page summary and the all-table summary rows respect the position; group summaries are unaffected.

The default is `Bottom`, so existing tables render identical output when the API is unused. The position filter is threaded through `getSummarizers()` / `hasSummary()` as an optional parameter, keeping the existing signatures backwards compatible.

New tests cover the API (default, enum value, `Closure`, resetting with `null`), a render test asserts the top/bottom ordering around the records, and a browser test asserts `assertNoAccessibilityIssues()` in light and dark mode.

## Visual changes

A summary row can now appear directly below the column header row.

## Functional changes

### DIFF
--- a/packages/tables/docs/06-summaries.md
+++ b/packages/tables/docs/06-summaries.md
@@ -424,6 +424,24 @@ public function table(Table $table): Table
 }
 ```
 
+## Positioning summaries
+
+By default, summaries are displayed at the bottom of the table. You may render a summarizer above the records instead, using the `position()` method with the `SummarizerPosition` enum:
+
+```php
+use Filament\Tables\Columns\Summarizers\Sum;
+use Filament\Tables\Columns\TextColumn;
+use Filament\Tables\Enums\SummarizerPosition;
+
+TextColumn::make('views_count')
+    ->summarize(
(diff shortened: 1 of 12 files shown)


danharrin:

PR #20228, opened 2026-07-21
feat(tables): add color support to table groups

quint-pruijssers:
## Description

Group headers currently always use the neutral gray styling. When a table contains many groups, or when groups have a semantic meaning (statuses, categories, warning levels), a color helps users scan the table. Badges, buttons and icons already support the color system; this brings the same to table groups.

This adds a `color()` method to the `Group` object, which colors the group header using any color registered in the app:

```php
{
    return $table
        ->groups([
            Group::make('status')
                ->color('info'),
        ]);
}
```

The method also accepts a function, which receives the first `$record` of each group, so every group can resolve its own color from its records:

```php
Group::make('status')
    ->color(fn (Post $record): ?string => match ($record->status) {
        'published' => 'success',
        'draft' => 'gray',
        default => null,
    }),
```

The color is applied to the group header row background in the table layout and to the group header block in the stacked/mobile layout, with a matching heading text color in light and dark mode. The color map is registered through a new `GroupHeaderComponent`, following the same pattern as `BadgeComponent`, so custom registered colors work as well. Returning `null` (the default) keeps the current neutral styling, so existing tables are unchanged.

New tests cover the API (static value, `Closure` per record, default `null`), a render test asserts the color class on the header, and a browser test asserts `assertNoAccessibilityIssues()` in light and dark mode.

## Visual changes

Group header rows get a colored background and heading color when `color()` is used.

## Functional changes


danharrin:

PR #20229, opened 2026-07-21
feat(tables): allow summarizers to render in the group header row

quint-pruijssers:
## Description

Grouped tables can already display summaries, but only as a separate row at the end of each group. In reporting-style tables it is often more useful to show the totals in the group header itself: the totals are the first thing you read, and they stay visible when a collapsible group is collapsed.

This adds an `inGroupHeader()` method to summarizers, so each summarizer can opt in to being rendered inside the group header row instead of the trailing group summary row:

```php
{
    return $table
        ->columns([
            TextColumn::make('views_count')
                ->summarize(Sum::make()->inGroupHeader()),
        ])
        ->defaultGroup('category');
}
```

Each summarizer decides individually where it is rendered, so both placements can be mixed on the same table: summarizers using `inGroupHeader()` are rendered in per-column cells inside the group header row (aligned with their columns, using the same heading-colspan logic as the existing summary rows), while the remaining summarizers stay in the trailing group summary row. The trailing row is suppressed entirely when no summarizers remain in it.

In the table layout the header row switches from a single `colspan` cell to a `<th scope="row">` heading cell plus per-column `<td>` summary cells; in the stacked/mobile layout the summaries render below the group heading. Selection checkboxes and action column placeholders are preserved. Without `inGroupHeader()`, the output is unchanged.

New tests cover the API, the header/trailing placement (including mixed usage), and a browser test asserts `assertNoAccessibilityIssues()` in light and dark mode.

## Visual changes

The group header row gains per-column summary cells when at least one summarizer uses `inGroupHeader()`.

## Functional changes


danharrin:

PR #20245, opened 2026-07-22
Add stickyScrollSidebar() to make the sidebar co-scroll with the page

the14thsky:
## Description

By default the desktop sidebar has a fixed height and its navigation scrolls **independently inside it** (a nested scrollbar). For panels with long navigation this results in two separate scroll contexts on one screen, and lower nav groups sit behind a subtle inner scrollbar.

This PR adds an opt-in panel option, `stickyScrollSidebar()`, that makes the desktop sidebar **scroll together with the page** instead. When the menu is taller than the viewport, the sidebar's bottom pins to the bottom of the viewport, so scrolling the page reveals the rest of the menu — no nested scrollbar.

```php
use Filament\Panel;

{
    return $panel
        // ...
        ->stickyScrollSidebar();
}
```

It is off by default, so existing behavior is unchanged.

Behavior when enabled:

- **Desktop (`lg`+):** the sidebar grows to its content height (no inner scroll) and is `position: sticky`.
  - Fits within the viewport → pins just below the topbar.
  - Taller than the viewport → its bottom pins to the viewport bottom, so every nav item stays reachable by scrolling the page.
- **Mobile / collapsed:** unchanged.

The topbar offset is **measured at runtime** (from `.fi-topbar-ctn`) and published as the `--fi-sidebar-sticky-scroll-offset` custom property, so the sidebar adapts to any header height — a taller brand logo, a notification bar, a wrapped header — rather than assuming a fixed height.

Implementation: a `stickyScrollSidebar()` setter + `hasStickyScrollSidebar()` getter on the `HasSidebar` concern (delegated through `FilamentManager` and the `Filament` facade); the sidebar Blade adds a `fi-sidebar-sticky-scroll` class; CSS makes the desktop sidebar tall/sticky with a page-scrolling nav; and `sticky-scroll-sidebar.js` measures the topbar and sets the sticky `top` (pinning the bottom when the sidebar is taller than the viewport). There is no per-scroll JS — CSS `position: sticky` handles the scrolling; the script only runs on `resize`, a `ResizeObserver` (sidebar + topbar), and `livewire:navigated`, and is inert for panels that don't enable the option.

## Visual changes

https://github.com/user-attachments/assets/59a16a20-6dc7-4fe8-aa62-fec6eb4a672a

**Before** (default — nav scrolls inside a fixed-height sidebar):

**After** (`stickyScrollSidebar()` — sidebar co-scrolls with the page; bottom groups revealed on scroll):

## Functional changes


danharrin:
Hi, thanks for your contribution. I think that this probably works better as a plugin rather than a core feature. There are definitely some UX tradeoffs to using this, like forcing the user to scroll through a long navigation list, which make me think it will not be opted-in to very often.

In the future, please @ me in a discussion before creating a PR for a feature so we can chat about it first.

PR #20247, opened 2026-07-22
Add `$activeWizardStepIndex` utility injection to schema components

wit3:
## Description

Schema components rendered inside a `Wizard` currently have no first-class way to know which step is active from within their configuration functions. Reaching the wizard means manually walking the component tree (`getContainer()->getParentComponent()` repeatedly), which is brittle and undocumented.

This PR adds a small utility, mirroring the existing `$parentRepeaterItemIndex` injection:

- A new `CanBeContainedByWizard` concern on the base schema `Component` (sibling of `CanBeRepeated`) that resolves the closest ancestor `Wizard`.
- `getParentWizard(): ?Wizard` — the closest ancestor wizard, or `null`.
- A new injectable **`$activeWizardStepIndex`** utility that resolves to the wizard's current zero-based step index (`Wizard::getCurrentStepIndex()`), or `null` when the component is not inside a wizard.

### Why

A common need is to defer expensive work until the step that needs it is actually active — e.g. only building a large option set once the user reaches that step, instead of on every render of the (mounted-but-hidden) later steps:

```php

Select::make('authorId')
    ->options(fn (?int $activeWizardStepIndex): array => ($activeWizardStepIndex === 1)
        ? Author::query()->pluck('name', 'id')->all()
        : [])
```

The value tracks the current step as known on the server (it updates as the user navigates, and combines with `persistStepInQueryString()` when it needs to survive a full page load).

## Changes

- `packages/schemas/src/Components/Concerns/CanBeContainedByWizard.php` — new concern (`getParentWizard()`, `getActiveWizardStepIndex()`).
- `packages/schemas/src/Components/Component.php` — use the concern; inject `activeWizardStepIndex` in `resolveDefaultClosureDependencyForEvaluationByName()`.
- `tests/src/Schemas/Components/WizardTest.php` — tests for injection (inside a step, `null` outside a wizard) and reactivity across navigation.
- `packages/schemas/docs/05-wizards.md` — documentation.

## Tests

New tests pass (`pest --configuration=phpunit.sqlite.xml tests/src/Schemas/Components/WizardTest.php`). Code passes `pint --config pint-strict-imports.json` and `phpstan` (level 6).

## Notes

Targets `5.x` to match a current release line; happy to retarget or forward-port if maintainers prefer.

### DIFF
--- a/packages/schemas/docs/05-wizards.md
+++ b/packages/schemas/docs/05-wizards.md
@@ -193,6 +193,23 @@ Wizard::make([
 
 <UtilityInjection set="schemaComponents" version="5.x">As well as allowing a static value, the `persistStepInQueryString()` method also accepts a function to dynamically calculate it. You can inject various utilities into the function as parameters.</UtilityInjection>
 
+## Accessing the current step index inside a step's schema
+
+Any component inside a wizard can inject the `$activeWizardStepIndex` utility into its configuration functions. It resolves to the zero-based index of the wizard's currently active step, or `null` when the component is not contained by a wizard. This is useful to defer expensive work until the step that needs it is actually active:
+
+```php
+use Filament\Forms\Components\Select;
+
+Select::make('authorId')
+    ->options(fn (?int $activeWizardStepIndex): array => ($activeWizardStepIndex === 1)
+        ? Author::query()->pluck('name', 'id')->all()
+        : [])
+```
+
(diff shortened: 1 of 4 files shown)


danharrin:
Thanks for the PR! I don’t think we should add this, since the wizard’s active step is managed client-side and a server-side index cannot remain synchronized across Livewire requests or all navigation paths. The performance use case is now handled more directly by #19867, which adds `deferLoading()` support for wizard step schemas so expensive content is only loaded when the step becomes active.

PR #20253, opened 2026-07-22
refactor(support): resolve vendor path via Composer Factory and NullIO

emyassine:
Dynamically resolve the project vendor directory using `Composer\InstalledVersions` and enforce strict path prefix matching with a trailing directory separator.

## Description

This PR refactors `discover_app_classes()` to dynamically determine the location of the Composer `vendor` directory at runtime using reflection on `Composer\InstalledVersions`, with a fallback to `base_path('vendor')`. It also fixes potential false positives in vendor path comparisons.

### Changes
* **Runtime Vendor Resolution:** Replaced hardcoded relative vendor path lookups with reflection on `Composer\InstalledVersions::class`. This avoids introducing dev dependencies like `Composer\Factory` while guaranteeing full production compatibility.
* **Custom Vendor Support:** Seamlessly supports non-standard Composer configurations where `vendor-dir` has been relocated (e.g., `third_party/`).
* **Strict Path Exclusion:** Appends a trailing `DIRECTORY_SEPARATOR` to `$vendorPathPrefix` and performs checks against both raw and `realpath()` resolved paths to prevent unintended matching of sibling directories (e.g., `/var/www/vendor2`).

> **Note:** This change specifically fixes class discovery in `discover_app_classes()`. However, other areas in the codebase may still contain legacy hardcoded `vendor/` references or assumptions about directory structures that could benefit from similar refactoring for complete custom `vendor-dir` support.

## Visual changes

None. Purely backend/architectural refactor.

## Functional changes

- [x] Documentation is up-to-date (checked, but as far as I know it has never been mentioned in the docs).

### DIFF
--- a/packages/support/src/helpers.php
+++ b/packages/support/src/helpers.php
@@ -351,7 +351,7 @@ if (! function_exists('Filament\Support\discover_app_classes')) {
     {
         $vendorDir = class_exists(\Composer\InstalledVersions::class)
             ? dirname((new \ReflectionClass(\Composer\InstalledVersions::class))->getFileName(), 2)
-            : base_path('vendor');
+            : 'vendor';
 
         $autoloadPath = $vendorDir . '/autoload.php';
 
--- a/packages/support/src/helpers.php
+++ b/packages/support/src/helpers.php
@@ -3,8 +3,8 @@
 namespace Filament\Support;
 
 use BackedEnum;
-use Composer\Factory;
-use Composer\IO\NullIO;
+use Composer\InstalledVersions;
+use ReflectionClass;
 use Filament\Support\Contracts\LoadingIndicator;
 use Filament\Support\Contracts\ScalableIcon;
 use Filament\Support\Enums\IconSize;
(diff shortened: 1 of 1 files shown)


emyassine (commented):
Refactor `discover_app_classes()` to safely resolve the Composer vendor directory at runtime using `Composer\InstalledVersions` instead of introducing a dependency on `Composer\Factory` or relying on hardcoded `vendor/` relative paths.

### Changes
* **Runtime Safety:** Replaced `Composer\Factory` with reflection on `Composer\InstalledVersions` (provided by `composer-runtime-api`). This prevents fatal `Class "Composer\Factory" not found` errors in production environments where `composer/composer` is not installed.
* **Custom `vendor-dir` Support:** Dynamically locates the active vendor directory regardless of custom Composer configurations (e.g., `third_party/`).
* **Path Prefixing Fix:** Appended a trailing `DIRECTORY_SEPARATOR` to the resolved vendor path prefix (`$vendorPathPrefix`) to prevent unintended exclusion of directories with overlapping prefixes (e.g., `/path/vendor2/`).

### Note
This change specifically resolves vendor paths for `discover_app_classes()`. Other areas across the codebase may still contain legacy hardcoded `vendor/` references that could benefit from similar refactoring for complete `vendor-dir` customization.

PR #20258, opened 2026-07-22
`RichEditor` `minHeight()`/`maxHeight()` support

pxlrbt:
## Description

`Textarea` natively supports `->rows()` but there was no option for `RichEditor` so far. Since this is not a native feature, I thought it was useful to add both min and max limits. It uses `lh` unit which is supported in baseline: https://caniuse.com/wf-lh

## Visual changes

https://github.com/user-attachments/assets/e79cba76-5044-4963-bef4-58d361fb148a

## Functional changes


danharrin (changes requested):
Unlike a textarea where line height is fixed, the rich editor has a variety of line heights based on the element used (heading, paragraph) like the markdown editor.

The markdown editor has min/max height, right? So should we not mirror that pattern for the rich editor instead of min/max rows?

pxlrbt:
You're right, it's probably the better variant. I only compared it to the Textarea and forgot about the MarkdownEditor. I still like the idea of "rows" for a simple RichEditor but I can just pass the `3lh` then

pxlrbt:
The approach I took is a bit different from the markdown editor. While that one works via JS I decided to use plain CSS. Because we need to apply the height restrictions to the content and not the outer field container, I pass them down via CSS vars. Might look a bit weird but works.

danharrin (dismissed):

danharrin (approved):

Issue #20262, opened 2026-07-23
Create action does nothing after using browser Back from a resource create page


CodeWithDennis:
### Package

filament/filament

### Package Version

v5.7.2

### Laravel Version

v13.21.1

### Livewire Version

v4.3.3

### PHP Version

v8.5.8

### Problem description

On a resource list page, opening Create, then pressing the browser Back button, and clicking Create again does nothing. The URL stays on the list page and the create page never opens.

This happens consistently after Back → Create. A hard refresh of the list page makes Create work again until the sequence is repeated. 

**Tested in Helium and Brave on macOS.**

### Expected behavior

Clicking Create after returning via the browser Back button should navigate to the create page again, just like the first time.

### Steps to reproduce

 1. Log in
 2. Open Posts
 3. Click New post
 4. Press the browser Back button
 5. Click New post again
 
Bug: nothing happens — the create page does not open.

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/CodeWithDennis/filament-create-after-browser-back

### Relevant log output

```shell

```

CodeWithDennis:
https://github.com/user-attachments/assets/310a278c-1df5-45b2-83d8-bfc9a31d3eb1

hamdyelbatal122:
PR created to resolve this issue: https://github.com/filamentphp/filament/pull/20272

danharrin:
I was a bit confused by this issue because the text mentions the list page a lot but it doesn't make sense in that context, and the screen recording only shows the create page. So I fixed it on the create page.

PR #16910, opened 2025-07-10
Updated the Create Action to no longer enable double clicking during creation and therefore creating duplicate records

PixelBeardCo:
## Description

When using the default CreateAction, slower network conditions can cause the submit button to be re-enabled before the redirect completes. This brief window allows users to click the button again, potentially resulting in duplicate record creation.

This fix introduces a new isCreating boolean flag to prevent multiple submissions. The flag is set when creation begins and immediately halts any subsequent attempts while the process is in progress. It is reset if an exception occurs to ensure normal functionality resumes.

## Visual changes

https://github.com/user-attachments/assets/ed4005c0-7e86-48ed-9ced-bf9a470d321e

## Functional changes

- [✔️ ] Code style has been fixed by running the `composer cs` command.
- [✔️] Changes have been tested to not break existing functionality.
- [✔️] Documentation is up-to-date.

PR #20272, opened 2026-07-26
fix: prevent Create action unresponsiveness after browser back navigation

hamdyelbatal122:
Fixes filamentphp/filament#20262

### Description
This PR fixes an issue where clicking the Create action on a resource list page becomes unresponsive after navigating to the Create page and returning via the browser Back button.

### Root Cause
1. **Redundant Livewire Click Handler**: Non-optimized action renderers were attaching `wire:click` handlers to actions even when a navigation `url` was configured. When navigating via `wire:navigate`, the combined click event and SPA navigation caused Livewire event listener state to conflict upon browser back navigation.
2. **Locked `$isCreating` State**: On `CreateRecord.php`, `$isCreating` was set to `true` upon form submission and not reset prior to calling `$this->redirect(...)`. As a result, Livewire saved a component snapshot with `$isCreating = true` in the browser history cache, causing subsequent form submit attempts after browser back navigation to return early.

### Changes Made
- Updated `Action::getLivewireClickHandler()` to return `null` if the action has a URL set and does not post to URL.
- Reset `$this->isCreating = false` right before calling `$this->redirect(...)` in `CreateRecord.php`.
- Added unit test in `ActionTest.php` to verify `getLivewireClickHandler()` behavior when an action has a URL.

### Testing
- Added unit test passes cleanly (`pest tests/src/Actions/ActionTest.php`).
- Verified code formatting with Pint (`vendor/bin/pint`).

### DIFF
--- a/packages/actions/src/Action.php
+++ b/packages/actions/src/Action.php
@@ -370,7 +370,7 @@ class Action extends ViewComponent implements Arrayable
             return null;
         }
 
-        if (filled($this->getUrl()) && (! $this->shouldPostToUrl()) && (! $this->hasAction())) {
+        if (filled($this->getUrl()) && (! $this->shouldPostToUrl())) {
             return null;
         }
 
--- a/packages/actions/src/Action.php
+++ b/packages/actions/src/Action.php
@@ -370,7 +370,7 @@ class Action extends ViewComponent implements Arrayable
             return null;
         }
 
-        if (filled($this->getUrl()) && (! $this->shouldPostToUrl())) {
+        if (filled($this->getUrl()) && (! $this->shouldPostToUrl()) && (! $this->hasAction())) {
             return null;
         }
 
--- a/packages/actions/src/Action.php
+++ b/packages/actions/src/Action.php
@@ -370,6 +370,10 @@ class Action extends ViewComponent implements Arrayable
             return null;
         }
 
+        if (filled($this->getUrl()) && (! $this->shouldPostToUrl())) {
+            return null;
+        }
+
         return $this->getJsClickHandler();
     }
 
--- a/packages/panels/src/Resources/Pages/CreateRecord.php
+++ b/packages/panels/src/Resources/Pages/CreateRecord.php
@@ -72,6 +72,8 @@ class CreateRecord extends Page
     public function hydrate(): void
     {
         $this->authorizeAccess();
+
+        $this->isCreating = false;
     }
 
     protected function fillForm(): void
@@ -162,8 +164,6 @@ class CreateRecord extends Page
         }
 
         $redirectUrl = $this->getRedirectUrl();
-
-        $this->isCreating = false;
 
         $this->redirect($redirectUrl, navigate: FilamentView::hasSpaMode($redirectUrl));
     }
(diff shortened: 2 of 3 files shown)


danharrin:
Your change reverts #16910 which is introducing another issue here.

hamdyelbatal122:
Thanks for catching that, @danharrin 
 You're completely right.

I've updated the PR to resolve this:

Kept $isCreating = true during the redirect phase to preserve the double-click protection from #16910.
Stored a timestamp in session on successful creation, and check it inside create(). If subsequent attempts are made after 15 seconds (such as returning via browser back), it resets $isCreating = false to allow submitting again.
Refined Action::getLivewireClickHandler() to only omit the handler when it's a pure navigation link without a server action.

danharrin:
Reopened to review again, I don't think this is the correct fix either, there shouldn't be an arbitrary number of seconds before the create button works again. We must work out how to ensure that some temporary state in props like this is not restored when going back in history.

Also, I think your fix in `Action.php` doesn't have anything to do with the issue so we should remove that, the create button doesn't have both sets of attributes as far as I can see.

hamdyelbatal122:
@danharrin 
I've updated the PR based on your review:

1. **Reverted `Action.php` changes**: Removed the unrelated URL check in `getLivewireClickHandler()` and its corresponding test.
2. **Removed Session Check & Reset `$isCreating` before Redirect**: Reverted the session timestamp workaround in `CreateRecord.php`. Instead, `$this->isCreating = false` is now set right before calling `$this->redirect(...)`.

**Why this fixes the issue:**
When `create()` completes, setting `$this->isCreating = false` before `$this->redirect(...)` ensures that Livewire serializes `isCreating: false` into the component snapshot. Since Alpine/Livewire uses this snapshot for history restoration, navigating back in browser history now restores the page with `isCreating: false`, making the Create button immediately functional again.

danharrin:
That redirect change is what you did before and it breaks the previous bug fix I mentioned

hamdyelbatal122:
@danharrin, thank you for clarifying.
Ah, that makes complete sense

I have updated the fix to reset `$this->isCreating = false` inside `hydrate()` in `CreateRecord.php` instead of before redirecting:

1. **Preserves Double-Submit Protection during Redirect**: `$this->isCreating` remains `true` throughout the `create()` execution and `$this->redirect(...)` call, so double-clicks during the redirect response stay protected.
2. **Resets Flag on New Requests / Back Navigation**: When navigating back in browser history and performing an action, component hydration calls `hydrate()`, which resets `$this->isCreating = false` at the start of the request, allowing the Create button to function again immediately.

danharrin:
I don't think you're testing these theories

hamdyelbatal122:
I've tested the flow thoroughly to ensure both requirements are satisfied without conflict.
### Root Cause
When `create()` completes, `$isCreating = true` was getting serialized into the component snapshot sent back with the redirect response. Alpine caches this snapshot when navigating away. When clicking browser Back, the page restores from that history snapshot with `isCreating: true`, disabling the Create button.
### Fix
Added a `dehydrate()` hook to `CreateRecord`:
```php
public function dehydrate(): void
{
    $this->isCreating = false;
}

danharrin:
That regresses the duplicate click issue again

PR #20285, opened 2026-07-27
Custom Downloader to allow for people running Laravel Vapor to suppor…

bretto36:
…t downloading of files

## Description

Allow the ability to customise the Downloader for an export action. I use Laravel Vapor, and large files get blocked by the Load Balancer/Lambda/API Gateway - https://docs.vapor.build/projects/development#binary-responses - Streamed responses don't seem to work for larger files. 

This allows someone to specify a new downloader on their Export - I am planning on making a Filament Plugin for this so that people can simply use a Trait from the package. 

I'd like your thoughts on the viability of this. Best case scenario is changing the existing system to automatically recognise Vapor and return a redirect to an S3 url instead of trying to stream it. But the last PR for that seemed to go no where. I figured this extends the system so that people can do whatever they want. Perhaps people could stream it if it's below a threshold or upload and return the s3 url if not. 

## Functional changes


packages/actions/src/Exports/Downloaders/CsvDownloader.php
```diff
@@ -17,21 +17,9 @@ public function __invoke(Export $export): StreamedResponse
...
-                    continue;
-                }
-
-                if (! str($file)->endsWith('.csv')) {
-                    continue;
-                }
-
-                echo $disk->get($file);
+        return response()->streamDownload(function () use ($export): void {
+            foreach (app(CsvExportContent::class)($export) as $chunk) {
```
bretto36, line 21: I moved the creation of the content to a separate class so that someone can extend the CSVDownloader/XsxDownloader and only override the invoke method, without having to touch the actual creation of the rows

packages/actions/src/Exports/Exporter.php
```diff
@@ -222,6 +223,11 @@ public function getFormats(): array
         return [ExportFormat::Csv, ExportFormat::Xlsx];
     }
 
+    public static function getDownloader(ExportFormatInterface $format): Downloader
```
bretto36, line 226: Allows someone to specify their own downloader depending on the Formatter used

bretto36:
Thanks @danharrin  love the extensions you've added

danharrin (approved):

PR #20287, opened 2026-07-27
[feat]: `->topbarOnMobile()` to display global search, noti & user menu on topbar on mobile view when use with topbar(false)

valpuia:
Create new `->topbarOnMobile()` to display global search, notification and user menu on topbar on mobile view instead of sidebar, while the desktop view still render on sidebar when `->topbar(false)` is used. 

## Visual changes

Before:

https://github.com/user-attachments/assets/6bc5714f-da69-4015-80bd-aa8b3211c302

After:

https://github.com/user-attachments/assets/c67c6426-dba5-4e47-928a-8a4ff0ae7b52

## Functional changes


danharrin:
Hi, thanks, but I think that I would like to wait longer for more community demand before introducing the feature. The PR is quite large for such a small change, and adding a new method to the navigation customisation API does make it even more cluttered than it was before, which could cause confusion.

Please open discussions in the future about feature proposals and @ me so we can chat about if it would be accepted

PR #20297, opened 2026-07-28
fix: SPA mode resource create page back button cache

### LINKED ISSUE #20262: Create action does nothing after using browser Back from a resource create page

danharrin:
Fixes #20262.


hamdyelbatal122:
Checked out the changes and the Dusk E2E browser test in `CreateRecordTest.php`—the pairing of `component.init` to detect history-restored snapshots with the `#[Renderless]` `resetCreation()` method handles the SPA history restoration cleanly while keeping double-submit protection active during redirect.

Looks solid!

PR #20306, opened 2026-07-30
Fix badge color condition in getNavigationBadgeColor

unctom:
## Description

The getNavigationBadgeColor() example in the documentation returns 'warning' when the record count is greater than 10, and 'primary' otherwise. This seems backwards for a typical use case a "warning" badge color is usually meant to draw attention when a value drops below an expected threshold (e.g. low stock, too few records), not when it exceeds one. The current example effectively flags "warning" for having more records, which is counter to how most implementations of this pattern are used in practice.

### Before
```php
public static function getNavigationBadgeColor(): ?string
{
    return static::getModel()::count() > 10 ? 'warning' : 'primary';
}
```
image

### After
```php
public static function getNavigationBadgeColor(): ?string
{
    return static::getModel()::count() < 10 ? 'warning' : 'primary';
}
```
image

## Functional changes

### DIFF
--- a/docs/06-navigation/01-overview.md
+++ b/docs/06-navigation/01-overview.md
@@ -85,7 +85,7 @@ If a badge value is returned by `getNavigationBadge()`, it will display using th
 ```php
 public static function getNavigationBadgeColor(): ?string
 {
-    return static::getModel()::count() > 10 ? 'warning' : 'primary';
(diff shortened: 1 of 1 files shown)


danharrin:
The example in the screenshot is `Orders`, I don't think this is confusing enough to change.

PR #20314, opened 2026-08-01
Fix(Forms): Defer MarkdownEditor initialization in collapsed sections and hidden containers

hamdyelbatal122:
Fixes #20313

### Cause
When a `MarkdownEditor` component is placed inside a collapsed `Section` or an inactive tab/container, `EasyMDE` (CodeMirror 5) eagerly attempts to measure character coordinates while the element has no rendered DOM dimensions (`visibility: hidden` or `height: 0`). This triggers `TypeError: Cannot read properties of null (reading 'top')` or `TypeError: Cannot read properties of undefined (reading 'map')`.

### Solution
1. **Deferred Initialization**: Refactored `markdownEditorFormComponent` to inspect container visibility on `init()`. If the editor starts inside a hidden/collapsed container, `EasyMDE` instantiation is deferred until the section is expanded and revealed via `ResizeObserver`.
2. **Init Hook Protection**: Added a safe `CodeMirror.defineInitHook` in `markdown-editor.js` to prevent unhandled selection measurement errors on zero-rect elements before `refresh()` runs.

PR #20320, opened 2026-08-02
fix: use raw Markdown editor instance when refreshing

### LINKED ISSUE #20313: Uncaught TypeError: Cannot read properties of undefined (reading 'map')   on markdown-editor.js
### Package

filament/filament

### Package Version

v5.7.5

### Laravel Version

13.23.0 

### Livewire Version

4.3.4

### PHP Version

PHP 8.4

### Problem description

When i'm using the MarkdownEditor field in a Filament collapsed section. I cant use click the field and edit it. It throws console errors and make the Markdown Editor usable. 

### Expected behavior

Just like a non collapsed section or outside a section that it works like expected.

### Steps to reproduce

1) Open up the given repository. 
2) Navigate to the create form form a user. 
3) open up chrome devtools and make sure u are on the console tab. 
4) Open up the collapsed section and try to use the Markdown editor fields

https://github.com/vl-wbk/markdown-editor-issue/tree/5.x

### Relevant log output

```shell
Uncaught TypeError: Cannot read properties of undefined (reading 'map')
    at Qr (markdown-editor.js?v=5.7.5.0:5:10087)
    at Xt (markdown-editor.js?v=5.7.5.0:5:14244)
    at _o (markdown-editor.js?v=5.7.5.0:5:21727)
    at Ya (markdown-editor.js?v=5.7.5.0:5:21664)
    at st.prepareSelection (markdown-editor.js?v=5.7.5.0:20:1059)
    at xf (markdown-editor.js?v=5.7.5.0:10:8858)
    at vf (markdown-editor.js?v=5.7.5.0:10:7877)
    at markdown-editor.js?v=5.7.5.0:10:7747
    at Zc (markdown-editor.js?v=5.7.5.0:5:3526)
    at Ir (markdown-editor.js?v=5.7.5.0:10:7676)
Qr @ markdown-editor.js?v=5.7.5.0:5
Xt @ markdown-editor.js?v=5.7.5.0:5
_o @ markdown-editor.js?v=5.7.5.0:5
Ya @ markdown-editor.js?v=5.7.5.0:5
st.prepareSelection @ markdown-editor.js?v=5.7.5.0:20
xf @ markdown-editor.js?v=5.7.5.0:10
vf @ markdown-editor.js?v=5.7.5.0:10
(anonymous) @ markdown-editor.js?v=5.7.5.0:10
Zc @ markdown-editor.js?v=5.7.5.0:5
Ir @ markdown-editor.js?v=5.7.5.0:10
(anonymous) @ markdown-editor.js?v=5.7.5.0:10
```

people-sea:
### Supersedes: #20314

The CodeMirror instance was refreshed through Alpine's reactive proxy.

Use `Alpine.raw()` to refresh the underlying editor instance.

### Before

https://github.com/user-attachments/assets/96fc23e4-060a-4775-a89e-c25a2132055e

### After

https://github.com/user-attachments/assets/03372f3c-6ec7-4a06-ba7f-4901f3273d2e

## Functional changes


hamdyelbatal122:
@people-sea, thanks for taking a look at this issue!

I noticed a couple of things that might be worth discussing:

1. Source file not updated The changes here only touch the compiled dist/ file. The source at packages/forms/resources/js/components/markdown-editor.js is unchanged, which means the fix would be overwritten on the next npm run build.

2. The root cause is still there The core issue is that CodeMirror tries to measure character coordinates while the editor is inside a hidden container (collapsed section, inactive tab, etc.). Adding Alpine.raw() to the refresh() call helps avoid the proxy issue, but the editor still initializes eagerly inside the hidden container and CodeMirror will still throw when it can't measure anything.

For example, if you put a MarkdownEditor inside a collapsed Section and load the page — CodeMirror will still fire up immediately and try to call getBoundingClientRect() on elements that have zero dimensions. That's where the TypeError comes from, and Alpine.raw() alone doesn't prevent that from happening.

In #20314, I took a different approach: the editor simply doesn't initialize until its container is actually visible. This way CodeMirror never runs into the measurement problem in the first place. 

Happy to discuss if you see it differently!

danharrin (approved):

PR #20334, opened 2026-08-05
Fix: don't reserve three columns for a stats widget that has fewer stats

webard:
## Description

A stats widget holding a single stat renders that stat in the first of three columns as soon as the widget itself is wide enough, leaving two thirds of it blank and squeezing the value until it overflows the card.

## Visual changes

Before (width > 1878px):
CleanShot 2026-08-05 at 08 48 15

After (width > 1878px):
CleanShot 2026-08-05 at 08 49 21

## Functional changes

### DIFF
--- a/packages/widgets/src/StatsOverviewWidget.php
+++ b/packages/widgets/src/StatsOverviewWidget.php
@@ -68,7 +68,9 @@ class StatsOverviewWidget extends Widget implements HasSchemas
         $count = count($this->getCachedStats());
 
         if ($count < 3) {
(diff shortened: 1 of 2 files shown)


danharrin:
Hmm, not sure I follow. The screenshot at the top shows 3 stats widgets aside each other which isn't really a standard config. A standard config would be one stats widget at full width. If we make the change you propose, I think that would make it so there is one card that spans the full width of the dashboard if there is just one widget which would look weird I think and would be breaking change.

Please help me understand if you disagree that this is a breaking change

Issue #20339, opened 2026-08-06
InvalidArgumentException: View [components.rich-editor] not found.


sagautam5:
### Package

filament/filament

### Package Version

v5.7.3

### Laravel Version

v13

### Livewire Version

v4

### PHP Version

v8.4

### Problem description

Previously, we were using filament `v5.6.7` version and did some view customization for rich text editor.

`app/Filament/Components/CustomRichEditor.php`

 ```php
    protected string $view = 'components.custom-rich-editor';
```

`resources/views/components/custom-rich-editor.blade.php`

```blade
<div class="my-class">
    @include ('filament-forms::components.rich-editor')
</div>
```

It was working fine on this version. Recently, we just upgraded the filament version to `v5.7.3`, a minor upgrade and we are getting this error.
```
InvalidArgumentException   PHP 8.4.6         
View [components.rich-editor] not found.
 -> @include ('filament-forms::components.rich-editor')
```

I went through  `Filament\Forms\Components\RichEditor` and noticed that previous $view property was missing there. May be core logic behind component views are changed but I could not find any documentation for the changes. May be I am misunderstanding how filament resolves views for components in recent versions.

### Expected behavior

Customized RichEditor should load correctly and its general functionalities in filament form should work correctly.

### Steps to reproduce

1. Use filament version `v5.6.7` or lower
2. Extend filament rich editor and make custom rich editor
3. Load custom blade view for the rich editor
4. Add following code on the blade view
```blade
<div>
    @include ('filament-forms::components.rich-editor')
</div>
```
5. Use this editor in some filament resource form
6. Upgrade filament version to v5.7.3 
7. Visit form where this custom component is used, you will get above error

Screenshots:

v5.6.7

<img width="1920" height="902" alt="Image" src="https://github.com/user-attachments/assets/a5f126c1-57dc-4c4c-a43a-fdd6cf1e9fee" />

v5.7.3:

<img width="1729" height="904" alt="Image" src="https://github.com/user-attachments/assets/27b0ec73-46ed-46fc-a140-10a55716ac18" />

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/sagautam5/editorbug

### Relevant log output

```shell
[...]

PR #20340, opened 2026-08-06
docs: explain RichEditor custom view wrappers

drewmt:
Related to #20339.

## Summary

- Document the supported custom wrapper pattern for RichEditor.
- Show how a custom field view delegates to its embedded markup instead of including Filament's internal Blade view.
- Add regression coverage for the documented pattern.

This gives applications upgrading from the previous internal view a stable wrapper path without restoring a duplicate Blade alias.

## Testing

- ./vendor/bin/pest tests/src/Forms/Components/RichEditorTest.php
- ./vendor/bin/pint --test --dirty

### DIFF
--- a/packages/forms/docs/10-rich-editor.md
+++ b/packages/forms/docs/10-rich-editor.md
@@ -17,6 +17,31 @@ RichEditor::make('content')
 
 <AutoScreenshot name="forms/fields/rich-editor/simple" alt="Rich editor" version="5.x" />
 
+## Wrapping the editor in a custom Blade view
+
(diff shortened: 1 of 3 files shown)


danharrin:
Thanks, but I would not recommend this approach. We actually have a "field wrapper" feature where you can do this across any type of field: call the `fieldWrapper()` method on the field and pass your own Blade component view with a slot for the content.

Alternatively if you don't need to completely wrap the field HTML you can inject any extra content including `view()` objects like this: https://filamentphp.com/docs/5.x/forms/overview#adding-extra-content-to-a-field

And for clarity: none of Filament's Blade views are protected from breaking changes really unless they are explicitly called out in the docs like our set of Blade components. If you want to be safe, you should fully define the Blade view in your own app rather than relying on the internals of ours which could change.

PR #20343, opened 2026-08-07
fix: expose table filter indicators as an accessible list

cannycookie:
## Description

The active filter indicator badges are rendered in a plain container with no list semantics, so assistive technologies announce them as loose, unrelated text — a screen-reader user can't tell how many filters are active or that the badges form a single group.

This adds `role="list"` and an accessible name to the badges container, and `role="listitem"` to each badge, so screen readers convey the count and grouping of active filters.

- **WCAG 2.1 — 1.3.1 Info and Relationships (Level A):** the visual grouping of active filters is now programmatically determinable.
- ARIA roles are used rather than `<ul>`/`<li>` because the container is a flex row, and browsers drop native list semantics from `display: flex` lists — `role="list"`/`role="listitem"` reliably preserve them.
- The `aria-label` reuses the existing `filament-tables::table.filters.indicator` translation (already shown as the visible group label), so no new translation strings are introduced.

The per-indicator remove button already has an accessible label, and the "reset all" control is an Action with its own name — so the container/badge roles were the only remaining gap in this block.

## Visual changes

None — the change is semantic only and produces no visual difference. The effect is in the accessibility tree:

**Before**
```
text "Active filters"
text "Status: Active"
text "Created: last 7 days"
```

**After**
```
text "Active filters"
list "Active filters"
  listitem  "Status: Active"
  listitem  "Created: last 7 days"
```

## Functional changes

### DIFF
--- a/packages/tables/resources/views/index.blade.php
+++ b/packages/tables/resources/views/index.blade.php
@@ -839,7 +839,10 @@
                                 {{ __('filament-tables::table.filters.indicator') }}
                             </span>
 
-                            <div class="fi-ta-filter-indicators-badges-ctn">
+                            <div
+                                class="fi-ta-filter-indicators-badges-ctn"
+                                role="list"
+                            >
                                 @foreach ($filterIndicators as $indicator)
                                     @php
                                         $indicatorColor = $indicator->getColor();
@@ -847,6 +850,7 @@
 
                                     <x-filament::badge
                                         :color="$indicatorColor"
+                                        role="listitem"
                                     >
                                         {{ $indicator->getLabel() }}
 


danharrin (changes requested):
Do we really want the `text "Active filters"` before the list of the same name? Or should we hide that for screenreaders?

Also please rebase your commit on 4.x, force push the branch, and set the target branch of the PR to 4.x

danharrin:
Hi @cannycookie, any update on this please?

cannycookie:
hey @danharrin  Sorry for the delay, have been maxed out.

Good point — I've dropped the aria-label instead. Screen readers mostly ignore names on lists anyway, so the visible text was already doing that job. No duplication now, and role="list"/role="listitem" still fix the grouping for WCAG2.2. 

Rebased onto 4.x and retargeted

danharrin (approved):

PR #20345, opened 2026-08-07
chore(boost): scoped directive

miteyema:
## Description

From Laravel Boost 2.5 onwards, path-scoped guidelines are possible.

## Visual changes

Leaner CLAUDE.md file when running `boost:install` with `BOOST_RULES_SCOPED_GUIDELINES=true`

## Functional changes


danharrin:
Thanks, but not all Filament code is in this directory. There is an open PR that moves it to a skill, I think we will go down that route instead soon.

PR #20346, opened 2026-08-07
Fix: Crash when showing all validation messages for per-element rules

webard:
## Description

A field with a per-element validation rule crashes the entire page instead of rendering a validation message, whenever `showAllValidationMessages()` is enabled on it:

```php
use Filament\Forms\Components\CheckboxList;
use Filament\Schemas\Contracts\HasSchemas;
use Filament\Schemas\Schema;
use Livewire\Component;

class SomeForm extends Component implements HasForms, HasSchemas
{
    use InteractsWithForms;

    public array $data = [];

    public function form(Schema $schema): Schema
    {
        return $schema
            ->statePath('data')
            ->components([
                CheckboxList::make('choices')
                    ->showAllValidationMessages()
                    ->options(['alpha' => 'Alpha', 'beta' => 'Beta']),
            ]);
    }

    public function save(): void
    {
        $this->form->getState();
    }
}

// call
livewire(SomeForm::class)
    ->fillForm(['choices' => ['gamma']])
    ->call('save');
```

Expected behavior: 
Validation message: "The selected choices.0 is invalid."

Actual behavior:
```
TypeError: htmlspecialchars(): Argument #1 ($string) must be of type string, array given
```

## Functional changes

### DIFF
--- a/packages/forms/resources/views/components/field-wrapper.blade.php
+++ b/packages/forms/resources/views/components/field-wrapper.blade.php
@@ -54,7 +54,7 @@
 
     if ($hasError && filled($statePath) && blank($errorMessage) && blank($errorMessages)) {
         if ($shouldShowAllValidationMessages) {
-            $errorMessages = $errors->has($statePath) ? $errors->get($statePath) : ($hasNestedRecursiveValidationRules ? $errors->get("{$statePath}.*") : []);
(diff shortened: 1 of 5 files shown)


danharrin (approved):
I reproduced the crash through the Livewire validation lifecycle and confirmed that wildcard validation message buckets are flattened in all core rendering paths. The targeted tests, browser coverage, Pint, and PHPStan pass.

PR #20347, opened 2026-08-07
fix: avoid dispatching duplicate failed authentication events

shaheenfawzy:
## Description

`Login::authenticate()` verifies the submitted password twice on every successful sign-in.

The first `Timebox` retrieves the user and calls `validateCredentials()`, so by the time it returns the credentials are known to be good and the panel access check has already run:

```php
$user = app(Timebox::class)->call(function (Timebox $timebox) use (...): ?Authenticatable {
    // retrieveByCredentials() + validateCredentials() + isUserAllowedToAccessPanel()
    return $user;
}, $timeboxDuration);
```

Those same credentials are then handed to the guard:

```php
if (! $authGuard->attemptWhen($credentials, fn (Authenticatable $user): bool => $this->isUserAllowedToAccessPanel($user), $remember)) {
```

`attemptWhen()` runs `retrieveByCredentials()` and `validateCredentials()` from scratch, so every sign-in pays two bcrypt comparisons and two `SELECT`s, fires `Attempting` twice, and runs the panel access check twice. At a production bcrypt cost of 12 the redundant hash alone is around 200ms.

Since nothing is left for the guard to establish, the user is logged in directly with `$authGuard->login()`. The behaviour the guard would have contributed is kept:

- `SessionGuard::hasValidCredentials()` dispatches `Validated`, and is protected, so `Validated` is now dispatched from `authenticate()` at the point the guard would have dispatched it.
- `SessionGuard::$rehashOnLogin` is protected, so the rehash reads `config('hashing.rehash_on_login', true)` directly, which is the same value `AuthManager` passes to the guard when it constructs it.

One deliberate difference: `SessionGuard::$lastAttempted` is protected with no setter, so it is no longer populated by a panel login. Nothing in Filament reads it.

Reproduction: https://github.com/shaheenfawzy/filament-login-double-hash (`composer setup && php artisan test --filter=LoginPerformanceTest`)

The same method also pads every sign-in with the multi-factor timebox; that is a separate concern and is fixed in #20348.

## Visual changes

None.

## Functional changes

`tests/src/Panels/Auth/LoginTest.php` gains three tests: `Attempting` and `Validated` each fire exactly once on a successful sign-in, a password stored at a lower bcrypt cost is rehashed, and it is left alone when `hashing.rehash_on_login` is disabled. The first fails on `5.x` without this change, seeing `Attempting` twice. The two rehash tests pass either way and are regression guards, since `attemptWhen()` rehashed too.

### DIFF
--- a/packages/panels/src/Auth/Pages/Login.php
+++ b/packages/panels/src/Auth/Pages/Login.php
@@ -30,7 +30,6 @@ use Filament\Support\Enums\Alignment;
 use Filament\View\PanelsRenderHook;
 use Illuminate\Auth\Events\Attempting;
 use Illuminate\Auth\Events\Failed;
-use Illuminate\Auth\Events\Validated;
 use Illuminate\Auth\SessionGuard;
 use Illuminate\Contracts\Auth\Authenticatable;
 use Illuminate\Contracts\Auth\Guard;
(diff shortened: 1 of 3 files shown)


danharrin (approved):
Thanks for prompting a closer review of the authentication events here. During the security review, we confirmed that the post-MFA `attemptWhen()` call must remain: it reloads and revalidates the user so password and panel-access changes made during the challenge are observed before a session is created. Removing it allowed an in-flight password reset to be bypassed and could overwrite the new password during rehashing.

I revised the branch to retain that security boundary and its intentional second `Attempting` event, while removing Filament’s redundant `Failed` dispatch after `attemptWhen()` has already dispatched it. The added regression coverage verifies early failures, unknown users, denied panel access, MFA and non-MFA final revalidation, password-reset preservation, and exactly one `Failed` event per failure.

The targeted authentication tests, PHPStan, and formatting checks pass.

PR #20354, opened 2026-08-09
Feature: Pausing and resuming actions

webard:
## Description

This pattern already exists: people build it today on top of `halt()`. An action checks in `before()` whether something is satisfied; if not, it mounts an additional action and interrupts itself with `halt()`. The child modal collects whatever was missing, then removes itself from the stack and calls `callMountedAction()` back on the parent.

This works, but every retry is a new attempt, not a continuation, from Filament’s perspective, so the whole process runs again. There are three effects that cannot be fixed from the outside:

- Limiter: `->rateLimit(3)` effectively gives you one successful flow, because interrupting and resuming each consume one hit. The limiter runs before any hook the author controls.
- `beforeFormValidated()` and `afterFormValidated()` run twice. There is no way to prevent them from being called; at best, you can make your own code idempotent, which the documentation does not mention.
- `ActionCalling` event fires twice, which breaks audit flows based on that event.

There is also an extensibility problem: `before()` is a single closure, so a plugin cannot hook into it without overwriting the application’s hook. The only non-overridable extension point today is the undocumented `ActionCalling` event.

### What changes
An action can now be paused. A paused action remains mounted together with the data the user has already entered, and the next call is a resume of that action: stages that have already run are not repeated.
Conditions are registered through `pauseWhen()`. They are evaluated after schema validation and before the `before()` hook:

```php
Action::make('transferFunds')
    ->schema([
        // ...
    ])
    ->pauseWhen(fn (): bool => ! session()->has('identity_confirmed_at'))
    ->action(function (array $data): void {
        // ...
    })
```

Conditions accumulate instead of overwriting each other, so a plugin can pause an action without stepping on a condition registered by the application, and vice versa. Alternatively, `$action->pause()` can be called imperatively from `before()` or from the action body.

### Use cases

- Data freshness. Someone opened an edit form, and the record changed underneath them in the meantime. You pause, show the diff, and let the user decide. Today this cannot be done cleanly: `requiresConfirmation()` runs before the form, while this is only known after validation.

- Conflict detection: another user has already managed to perform the operation.

- Impact preview: "This operation will affect 1,240 records and delete 3 related invoices. Continue?", calculated on the server after validation.

- Re-authentication before a sensitive operation: password, OTP code.

- Missing data: the action needs a delivery address that the customer does not have. A modal adds it and fills the parent form, and the user does not lose anything they have already entered.

- Limits or billing: "You’re out of credits", top-up modal, resume.

### Security

The stage at which the action stopped is stored in `#[Locked] public array $pausedMountedActions`, so the browser cannot forge it to bypass the limiter or validation hooks.

Pause conditions are recalculated on every call, including on resume. This is intentional: the browser can call `callMountedAction()` on a paused action at any moment, even after cancelling the child modal, so execution is decided by the conditions, not by saved state.

### Backward compatibility

Nothing changes for actions that do not use pausing. One public Livewire property has been added: an empty array until something is paused.

### Demo

Working demo: [https://github.com/webard/filament-pauseable-otp-demo](https://github.com/webard/filament-pauseable-otp-demo)

```bash
git clone -b filament-pauseable-extracted-otp https://github.com/webard/filament.git filament
git clone https://github.com/webard/filament-pauseable-otp-demo.git demo
cd demo && composer setup && php artisan serve
```

Both clones need to sit next to each other, because the demo points to Filament through relative paths: `../filament/packages/*`.

> Note: the demo branch merges this PR with [#20353](https://github.com/filamentphp/filament/pull/20353), which extracts the multi-factor challenge outside the login page.

### Proof of concept

https://github.com/webard/filament-otp-actions

The package adds a `requiresOtpConfirmation()` macro to actions, which confirms an action using a one-time code from the multi-factor method configured by the user. It is built entirely on the public API introduced by this PR: it registers a condition through `pauseWhen()`, mounts its own challenge modal, and resumes the action after a valid code.

> Note: this package uses functionality from [#20353](https://github.com/filamentphp/filament/pull/20353), which extracts the multi-factor challenge outside the login page.

## Functional changes

### DIFF
--- a/packages/actions/docs/01-overview.md
+++ b/packages/actions/docs/01-overview.md
@@ -508,6 +508,140 @@ Action::make('delete')
     })
 ```
 
+## Pausing an action
+
+An action can be paused, which stops it before it runs, but keeps it mounted along with any data that the user has already entered. The next time it is called, it resumes instead of starting again: the stages that already completed, such as rate limiting and the hooks around form validation, are not repeated.
+
+This is useful when an action needs something more from the user before it may run - confirming their identity, for example - and you want to ask for it once their form has been validated, without the action running in the meantime.
+
+Register a condition with `pauseWhen()`. It is evaluated after the action's schema is validated, and before the `before()` hook:
+
+```php
+use Filament\Actions\Action;
+
+Action::make('transferFunds')
+    ->schema([
+        // ...
+    ])
+    ->pauseWhen(fn (): bool => ! session()->has('identity_confirmed_at'))
+    ->action(function (array $data): void {
+        // ...
+    })
+```
+
+Conditions accumulate rather than replace each other, so a plugin may pause an action without overwriting a condition that you registered, and the other way around.
+
+Alternatively, `$action->pause()` may be called from inside the `before()` hook or the action itself:
+
+```php
+use Filament\Actions\Action;
+
+Action::make('transferFunds')
+    ->before(function (Action $action): void {
+        if (session()->has('identity_confirmed_at')) {
+            return;
+        }
+
+        $action->pause();
+    })
+    ->action(function (): void {
+        // ...
+    })
+```
+
+<Aside variant="info">
+    A hook that pauses is not called again when the action resumes, since it already ran up to the point that it paused. Pause conditions are the exception: they are evaluated every time, including on resumption. They are what keeps a paused action paused, so an action whose conditions are still met simply pauses again.
+</Aside>
+
+### Asking the user for something before an action resumes
+
+Since a paused action stays mounted, another action can be mounted on top of it, and its modal is stacked above the paused action's modal. Once the user has done what the pause was waiting for, unmount the child action and call the parent action again, which resumes it:
+
+```php
+use Filament\Actions\Action;
+
+Action::make('transferFunds')
+    ->schema([
+        // ...
+    ])
+    ->pauseWhen(function (Action $action): bool {
+        if (session()->has('identity_confirmed_at')) {
+            return false;
+        }
+
+        $action->getLivewire()->mountAction('confirmIdentity');
+
+        return true;
+    })
+    ->registerModalActions([
+        Action::make('confirmIdentity')
+            ->schema([
+                // ...
+            ])
+            ->action(function (Action $action): void {
+                session()->put('identity_confirmed_at', now());
+
+                $livewire = $action->getLivewire();
+
+                array_pop($livewire->mountedActions);
+
+                $livewire->callMountedAction();
+            }),
+    ])
+    ->action(function (array $data): void {
+        // ...
+    })
+```
+
+The parent action's data survives the round trip, so it receives everything that the user entered before it paused.
+
+The nested action can also write to it. The paused action's schema state lives on the Livewire component at `mountedActions.{nestingIndex}.data`, so a nested action can fill in a field that the user left empty, and the paused action receives it once it resumes:
+
+```php
+use Filament\Actions\Action;
+
+Action::make('createInvoice')
+    ->schema([
+        TextInput::make('title')
+            ->required(),
+        TextInput::make('reference'),
+    ])
+    ->pauseWhen(function (Action $action): bool {
+        if (filled($action->getData()['reference'] ?? null)) {
+            return false;
+        }
+
+        $action->getLivewire()->mountAction('generateReference');
+
+        return true;
+    })
+    ->registerModalActions([
+        Action::make('generateReference')
+            ->schema([
+                TextInput::make('prefix')
+                    ->required(),
+            ])
+            ->action(function (Action $action, array $data): void {
+                $livewire = $action->getLivewire();
+
+                data_set($livewire->mountedActions, '0.data.reference', "{$data['prefix']}-123");
+
+                array_pop($livewire->mountedActions);
+
+                $livewire->callMountedAction();
+            }),
+    ])
+    ->action(function (array $data): void {
+        // `$data['reference']` is now filled in.
+    })
+```
+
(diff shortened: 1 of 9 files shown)


danharrin:
Thanks for putting so much work into this, including the demo and proof of concept. I have spent some time tracing the lifecycle and trying the use cases with nested actions. Unfortunately, I do not think we should add general pausing and resuming to the action pipeline.

For application-owned flows, I think the nested action should be responsible for the final operation rather than resuming the parent action. For example, the parent can collect the input while a nested action validates the confirmation, revalidates and reauthorizes the parent input, performs the operation, and calls `cancelParentActions()`:

```php
Action::make('transferFunds')
    ->schema([
        // ...
    ])
    ->modalSubmitAction(false)
    ->extraModalFooterActions([
        Action::make('confirmTransfer')
            ->schema([
                // Password, OTP, etc.
            ])
            ->action(function (array $data, array $mountedActions): void {
                $input = $mountedActions[0]->getRawData();

                // Validate and authorize `$input` again, verify any
                // server-side preconditions, then perform the operation.
            })
            ->cancelParentActions(),
    ])
```

The parent data exposed through `$mountedActions` is raw, so the nested action must not trust an earlier request’s validation. It should validate the input it consumes and recompute any security-sensitive conditions immediately before performing the operation.

When a confirmation depends on validated parent data, such as an impact preview, the parent can run once as a preflight action, compute the preview, mount the nested confirmation action, and `halt()`. The nested action then recomputes the impact and performs the operation itself. The parent is not resumed, so its lifecycle does not run twice. Put the operation’s limiter and auditing on the nested action that actually performs it.

The other examples appear to fit the same model:

- For stale data or conflicts, let the nested action show the conflict and perform a compare-and-swap after checking the record again.
- For missing data or topping up credits, let the child finish normally and return to the parent for an explicit resubmission, or let the child perform the final operation if automatic continuation is important.
- For an unavailable external service, `halt()` already leaves the form mounted so the user can retry. That is a new attempt, so rerunning its limiter and hooks seems correct.

More importantly, I found several problems with the proposed checkpoint state:

- `pausedMountedActions` binds the completed stage only to a nesting index, not to the identity of the action that paused. `#[Locked]` prevents the browser from inventing a checkpoint, but it does not prevent a legitimate checkpoint from being applied to another action. I reproduced pausing action A, changing `mountedActions.0.name` to action B, and calling it. Action B had already exhausted its `rateLimit(1)`, but it still ran and its `before()` hook was skipped. The normal disabled and authorization checks still run, but the checkpoint can bypass lifecycle safeguards belonging to a different action.
- The checkpoint is pulled and removed before resumed validation. If validation fails, it is not restored. The following submission is treated as a fresh call, so the limiter and validation hooks become eligible to run again. I reproduced the validation hooks running twice this way.
- This is not continuation when `pause()` is called imperatively. If `before()` pauses, everything after `pause()` in that hook is skipped permanently because the whole hook is skipped on the next call. If the action body pauses, the body starts again from the beginning when resumed. I reproduced both behaviors. That makes side effects before or after `pause()` particularly difficult to reason about.
- A pause also ends the current database transaction by committing or rolling it back. Resumption starts another transaction in another request, so the operation is not a continuation of the original transactional unit. If the pause rolls back, earlier stages may nevertheless be skipped when it resumes; if it commits, a later failure cannot roll those effects back.

The state that satisfies a pause condition is also left to each consumer. The documentation recommends `session()->has('identity_confirmed_at')` and then `session()->put(...)`. That confirmation is shared by every tab and action flow using the same Laravel session. This may be intentional for a session-wide recent-authentication grant, but it does not confirm one particular action invocation. The demo’s fixed acceptance keys have the same cross-tab and concurrent-flow problem. The locked Livewire property only remembers where execution stopped; it does not provide invocation-scoped confirmation state.

Finally, the documented happy path still requires consumers to mutate `mountedActions` directly with `array_pop()`, write data using a hard-coded nesting index, and call `callMountedAction()` manually. That suggests the abstraction is not removing the stack-management complexity so much as adding lifecycle checkpointing around it.

The one requirement nested actions do not satisfy transparently is a plugin adding something like `requiresOtpConfirmation()` to an arbitrary existing action while preserving that action’s complete hooks, transaction, notifications, redirects, and closure. I agree that this is a real extensibility gap, but it is effectively asking for middleware or continuations around arbitrary action code. I do not think this stage-based state machine is safe enough to provide that.

If core support is needed, I would prefer a narrower API that allows a nested action to submit or revalidate its parent schema and receive the validated data. That would improve the responsible-child flow without persisting and selectively replaying the parent lifecycle. The reusable MFA extraction in [#20353](https://github.com/filamentphp/filament/pull/20353) can still be considered separately.

webard:
Understood, and agreed. Thanks for digging into it @danharrin. Would this shape be acceptable instead? Conditions accumulate, so an application and a plugin do not overwrite each other:

```php
->requiresConfirmationFromAction(
    Action::make('topUpCredits')
        ->modalHeading('Not enough credits')
        ->schema([/* ... */])
        ->action(function (Action $action): void {
            // Top up, then hand control back only if it actually succeeded.
            $action->submitParentAction();
        }),
    fn (array $data): bool => auth()->user()->credits < $data['budget'],
)
->requiresConfirmationFromAction(
    OtpChallengeAction::make(),
)
```

Nothing is persisted and no stage is skipped - the action validates its schema and runs every hook on every submission. The condition is evaluated after validation and before `before()`, and the nested action hands control back with `submitParentAction()` once its own transaction has committed. It also gets `getValidatedData()` and `fillParentActionData()`, so nothing touches `mountedActions` by hand.

If this looks reasonable, I would open it as a separate PR.

danharrin:
I think I would prefer not to add this abstraction at this stage. If nested actions are missing a smaller, independently useful capability, such as safely validating or filling parent action data, we could consider that?

PR #20361, opened 2026-08-12
Fix: forget a closed action's cached schema when it is unmounted

webard:
## Description

`unmountAction()` pops the action off `mountedActions`, but leaves its schema in `cachedSchemas`
under `mountedActionSchema{nestingIndex}`. `getMountedActionSchema()` reads that cache before it
builds anything, so an action mounted at the same nesting index later in the same request is handed the schema of
the action that has just closed.

This causes actions to have their own header and footer, but someone else’s form. The bug appeared while testing the `modalClickThrough()` method. It occurred when I opened one action and then, without closing the current one, opened another.

## Functional changes

### DIFF
--- a/packages/actions/src/Concerns/InteractsWithActions.php
+++ b/packages/actions/src/Concerns/InteractsWithActions.php
@@ -421,12 +421,12 @@ trait InteractsWithActions /** @phpstan-ignore trait.unused */
         $this->mountedActions = [];
         $this->cachedMountedActions = null;
 
-        $this->unCacheMountedActionSchemas();
+        $this->forgetCachedMountedActionSchemas();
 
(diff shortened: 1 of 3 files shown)


webard:
I realize that the same fix can be achieved in userland with the `replaceMountedAction()` method, but that is not entirely intuitive, and not handling this automatically introduces a silent bug.

However, if `replaceMountedAction()` is intended to be the only way to fix this problem, I fully agree with that and will close the PR.

danharrin:
`replaceMountedAction()` is still the correct way to intentionally replace an action. This fix also keeps the schema cache consistent whenever an action is unmounted, preventing stale schemas in other valid flows. Thank you!

danharrin (approved):

PR #20378, opened 2026-08-20
fix: \`TextInput::copyable()\` copies stale state instead of current …

### LINKED ISSUE #20374: TextInput::copyable() copies stale initial state
### Package

filament/forms

### Package Version

v5.7.6

### Laravel Version

v11.55.0

### Livewire Version

v4.4.0

### PHP Version

PHP 8.4.8

### Problem description

TextInput::copyable() copies the field state from the last Livewire render instead of the current value in the input.

Example:

```
TextInput::make('name')
    ->copyable();
```

Steps to reproduce:

1. Open a Create or Edit form with a copyable TextInput.
2. Change the field value.
3. Click the copy button.

Expected: the currently entered value is copied.

Actual: the initial/last-rendered state is copied. For create action, it returns null. For edit, it always returns the initial state.

The issue appears to come from CopyAction generating the clipboard handler from the server-side $state:

```
$this->alpineClickHandler(function (mixed $state): string {
    $copyableState = Js::from($state);

    return <<<JS
        window.navigator.clipboard.writeText({$copyableState})
        JS;
});
```

For non-live fields, this state becomes stale. The copy action should use the current client-side field value at click time.

### Expected behavior

The currently entered value is copied.

### Steps to reproduce

1. Open a Create or Edit form with a copyable TextInput.
2. Change the field value.
3. Click the copy button.

### Relevant log output

```shell

```

harshp-fabninjas:
## Description

`TextInput::copyable()` copied the field's value from the last Livewire render instead of whatever was currently typed in the input. On a Create form it copied `null`; on Edit it always copied the original value, ignoring any edits made since the last server round trip.

This happened because `CopyAction` built its clipboard click handler from the server-side `$state` via `Js::from($state)`, baking that value into the generated JS at render time. For non-`live()` fields, that value goes stale as soon as the user types.

The fix reads the input's current value straight from the DOM at click time instead of a value frozen in at render time: `$el.closest('.fi-input-wrp')?.querySelector('input')?.value`. The button and its `<input>` share a common wrapper element (`.fi-input-wrp`), so this reliably resolves to the field's own input — the same pattern already used elsewhere in the codebase (`x-on:focus-input.stop="$el.querySelector('input')?.focus()"`).

Fixes #20374

## Visual changes

N/A — no visual/markup changes, only the generated JavaScript behavior of the copy button.

## Functional changes

### DIFF
--- a/packages/forms/src/Components/TextInput/Actions/CopyAction.php
+++ b/packages/forms/src/Components/TextInput/Actions/CopyAction.php
@@ -26,12 +26,11 @@ class CopyAction extends Action
         $this->defaultColor('gray');
 
         $this->alpineClickHandler(function (mixed $state): string {
-            $copyableState = Js::from($state);
             $copyMessageJs = Js::from($this->getCopyMessage($state));
             $copyMessageDurationJs = Js::from($this->getCopyMessageDuration($state));
 
             return <<<JS
-                window.navigator.clipboard.writeText({$copyableState})
+                window.navigator.clipboard.writeText(\$el.closest('.fi-input-wrp')?.querySelector('input')?.value ?? '')
                 \$tooltip({$copyMessageJs}, {
                     theme: \$store.theme,
                     timeout: {$copyMessageDurationJs},
--- a/packages/forms/src/Components/TextInput/Actions/CopyAction.php
+++ b/packages/forms/src/Components/TextInput/Actions/CopyAction.php
@@ -30,7 +30,7 @@ class CopyAction extends Action
             $copyMessageDurationJs = Js::from($this->getCopyMessageDuration($state));
 
             return <<<JS
-                window.navigator.clipboard.writeText(\$el.closest('.fi-input-wrp')?.querySelector('input')?.value ?? '')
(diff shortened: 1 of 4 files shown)


packages/forms/src/Components/TextInput/Actions/CopyAction.php
```diff
@@ -26,12 +26,11 @@ protected function setUp(): void
         $this->defaultColor('gray');
 
         $this->alpineClickHandler(function (mixed $state): string {
-            $copyableState = Js::from($state);
             $copyMessageJs = Js::from($this->getCopyMessage($state));
             $copyMessageDurationJs = Js::from($this->getCopyMessageDuration($state));
 
             return <<<JS
-                window.navigator.clipboard.writeText({$copyableState})
+                window.navigator.clipboard.writeText(\$el.closest('.fi-input-wrp')?.querySelector('input')?.value ?? '')
```
people-sea: I don’t think we should fall back to '' here. If a DOM change or another issue prevents the input from being resolved, it should fail instead of clearing the user’s clipboard.

danharrin (approved):

PR #20387, opened 2026-08-22
fix: apply the connection table prefix to number constraint aggregate subqueries

### LINKED ISSUE #20382: Query builder: number constraint aggregate ignores the connection table prefix
### Package

filament/filament

### Package Version

v5.7.6

### Laravel Version

v13.26.1

### Livewire Version

v4.4.1

### PHP Version

8.4.23

### Problem description

When a `NumberConstraint` targets a relationship (`children.length`) and an aggregate is picked in the "Aggregate" select, the generated sub-query references the related table without the connection prefix, so the query fails:

```
SQLSTATE[HY000]: General error: 1 no such column: items.length

select * from "fo_items"
where (select cast(min(items.length) as real)
       from "fo_items"
       where "fo_items"."parent_id" = "fo_items"."id") >= 700
```

The table is `fo_items`, but `min()` gets `items.length`. In the same statement
`whereColumn()` does get the prefix.

Without an aggregate the same constraint works.

### Expected behavior

The aggregate sub-query runs and filters the records, like the same constraint does without an aggregate.

### Steps to reproduce

1. Set a table prefix on the connection: `config/database.php` → `'prefix' => 'fo_'`.
2. One table `items` with a self relation: `id`, `parent_id`, `name`, `length`;
   model `Item` with `children(): HasMany`.
3. Table filter:

```php
QueryBuilder::make()->constraints([
    NumberConstraint::make('children.length'),
])
```

4. Open `/admin/items`, add the "Length" rule, operator "Minimum", value `700`.
   The filter works.
5. Pick any option in the "Aggregate" select — the query fails with
   `no such column: items.length`.

https://github.com/albyzip/filament

### Relevant log output

```shell

```

lazerg:
## Description

When a `NumberConstraint` on a relationship uses an aggregate, the column inside the aggregate is interpolated straight into `selectRaw()`, so the connection's table prefix is never applied and the query fails with `no such column: items.length` on a prefixed connection. Passing the qualified column through the grammar's `wrap()` gives `min("fo_items"."length")`, matching what `whereColumn()` in the same subquery already produces.

Fixes #20382

## Visual changes

None.

## Functional changes

- Aggregate subqueries built by `NumberConstraint` now resolve against the prefixed table name, so the filter works on connections that set a prefix.

### DIFF
--- a/packages/query-builder/src/Constraints/NumberConstraint/Operators/Concerns/CanAggregateRelationships.php
+++ b/packages/query-builder/src/Constraints/NumberConstraint/Operators/Concerns/CanAggregateRelationships.php
@@ -68,7 +68,7 @@ trait CanAggregateRelationships
         $relationship = $query->getModel()->{$relationshipName}();
 
         $relatedModel = $relationship->getModel();
-        $attributeForQuery = $relatedModel->qualifyColumn($attributeForQuery);
+        $attributeForQuery = $relatedModel->getConnection()->getQueryGrammar()->wrap($relatedModel->qualifyColumn($attributeForQuery));
         $castType = $this->getNumericCastType($query);
 
         if ($relationship instanceof BelongsToMany) {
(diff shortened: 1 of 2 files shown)


danharrin:
Duplicate of #20386

PR #20402, opened 2026-08-26
fix: handle closure conditions in searchable table columns

ropi-bc:
## Description

Fixes closures passed to a table column `searchable()` method being treated as search column names instead of conditions.

This caused global searches to pass a `Closure` to `getJsonSafeColumnName()`, resulting in a `TypeError`. It also meant that a closure returning `false` still marked the column as searchable.

Closures now follow the same condition path as boolean values. Tests cover both a searchable column returning `true` during a global search and a condition returning `false`.

## Visual changes

None.

## Functional changes


ropi-bc:
sorry about the confusion with the users again, hopefully I'll be able to merge these 2 accounts soon

danharrin (approved):

PR #20414, opened 2026-08-29
chore: run `composer run pint`

jackbayliss:
## Description

IT IS I, I AM FINALLY HERE

This runs `composer run pint` 

There are two separate pint scripts
1) The cs script 
2) The pint script  

One has strict imports, the other doesnt.

Perhaps there's a reason for the two being diff?  Anyway, I ran the strict one seeing as everything else adheres to it

Could we just copy the strict json into the pint.json? That way, free strictness? (unless theres a reason theres 2 idk)

## Functional changes

### DIFF
--- a/packages/tables/src/Columns/Column.php
+++ b/packages/tables/src/Columns/Column.php
@@ -15,6 +15,7 @@ use Filament\Support\Concerns\HasPlaceholder;
 use Filament\Support\Concerns\HasVerticalAlignment;
 use Filament\Support\Concerns\HasWidth;
 use Filament\Support\Enums\Alignment;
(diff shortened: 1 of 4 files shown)


danharrin:
Heeeeey!

The strict Pint config breaks the Blade Prettier plugin. And I do want strict imports by default, but you cant specify a custom Pint config path for the Blade Prettier plugin AFAIK. So we are stuck with running Pint with the custom strict config :/

danharrin (approved):

danharrin:
I do have a GH action that should format the code, did we misconfigure it to not use the strict config?

jackbayliss:
@danharrin Possibly! I'll give it a poke in a sep PR

PR #20419, opened 2026-08-29
chore: use parallel and bump pint

jackbayliss:
## Description

This bumps pint to the latest sem ver version and uses the --parallel flag meaning CI will be much faster  which means happy life.

laravel/pint v1.27.1 had the fix to support the parallel flag when using the config flag, but thought I'd just update it the latest sem ver.

Can't update to 1.3x cause they bumped the php version

Tested locally and all gucci. (probably worth running CI workflow too maybe)

## Visual changes

None.

## Functional changes

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         "danharrin/date-format-converter": "^0.3.1",
         "filament/support": "*",
         "larastan/larastan": "^3.0",
-        "laravel/pint": "1.25.1",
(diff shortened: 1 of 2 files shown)


danharrin (approved):

danharrin:
All green, thanks! https://github.com/filamentphp/filament/actions/runs/33298370792/job/99221834626

PR #20435, opened 2026-09-02
fix: scope the wizard next action loading indicator to its own request

people-sea:
## Description

The Wizard Next action currently targets `callSchemaComponentMethod`, so unrelated schema component requests can also trigger its loading indicator.

This scopes the action target to the full `nextStep` calls, including the Wizard key and each possible `currentStepIndex`. Parameterized targets are only applied when the action still uses the Wizard's default target, so custom targets configured through `nextAction()` are preserved.

The wrapper still needs a broad loading state because other Livewire requests should prevent moving to the next step. It now uses `inert`, which blocks both pointer and keyboard interaction without changing the button's appearance. Only the Next action's own request disables the button visually and shows its loading indicator.

Support for multiple parameters in `wire:target` requires Livewire 3.7, so the minimum Livewire version has been raised to `^3.7`.

A browser test covers both cases: loading dynamic options does not show the Next action's loading indicator, while clicking the Next action does.

While investigating this, I found that the permanently stuck loading state was caused by loading state leaking between concurrent requests. That issue has already been fixed upstream.

### Before

https://github.com/user-attachments/assets/5b0ff4e2-5c19-46e6-b53c-27509e962baa

### After

https://github.com/user-attachments/assets/1e738f18-0bb0-47cf-8881-4bd5d1cb9a86

## Functional changes


packages/schemas/resources/css/components/wizard.css
```diff
@@ -147,7 +147,7 @@
         }
 
         & > .fi-disabled {
-            @apply pointer-events-none opacity-70;
+            @apply pointer-events-none;
```
people-sea: This ensures that the Next button cannot be clicked while another request is running, and the cursor remains the default arrow to indicate that it is not clickable.

danharrin (approved):

PR #20446, opened 2026-09-05
feat: Add relationship support to `ColumnGroup`

DamienJolly:
## Description

This allows a relationship to be defined on a column group so that child columns are resolved relative to that relationship.

For example:

```php
ColumnGroup::make('Author', [
    TextColumn::make('name'),
    TextColumn::make('email'),
    TextColumn::make('company.name'),
])
->relationship('author')
```

The columns will resolve as:

```php
author.name
author.email
author.company.name
```

Columns that are already prefixed with the configured relationship are left unchanged.

This makes `ColumnGroup` more useful for defining reusable groups of columns that may be used directly on a model or through a relationship.

Related discussion:
#20445 

## Visual changes

No visual changes.

This is an API/functional change to `ColumnGroup`.

## Functional changes

### DIFF
--- a/packages/tables/docs/02-columns/01-overview.md
+++ b/packages/tables/docs/02-columns/01-overview.md
@@ -827,6 +827,30 @@ ColumnGroup::make('Website visibility')
     ->wrapHeader()
 ```
 
+If all the columns in a group belong to the same relationship, you can use the `relationship()` method to prefix each child column's name with the relationship name, instead of having to do it manually for each column:
+
+```php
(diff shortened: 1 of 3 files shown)


danharrin:
Thanks for the contribution! I can see how this would reduce repetition when reusing column definitions, but I'm going to pass on adding this to core. Explicit names like `customer.name` keep the relationship clear, and I don't think the convenience here warrants another API on `ColumnGroup`.

PR #20469, opened 2026-09-08
Add search to Builder block picker

KianAcquoy:
## Description

Adds an opt-in search field to the `Builder` block picker, to quickly filter large block lists. It reuses the `CanBeSearchable` concern from `CheckboxList`, so `searchPrompt()`, `noSearchResultsMessage()` and `searchDebounce()` are also supported. The field is auto-focused when the picker opens, using a new reactive `isOpen` property on the shared dropdown Alpine component.

```php
Builder::make('content')
    ->searchable()
    ->blocks([
        // ...
    ])
```

## Visual changes

_Without `searchable`_
image

_With `searchable`_
image
image

## Functional changes


danharrin (changes requested):
Thanks for this! A few things need addressing before we merge:

- Please keep the search labels in sync when the available blocks change, for example when a block reaches `maxItems()`. Alpine currently retains the original labels array after a Livewire morph, so searches can hide the wrong blocks.
- Escape currently closes the picker as well as clearing the search. The dropdown handles Escape before the input’s `stopPropagation()`. Please make the first Escape clear a nonempty search, then close on the next Escape and return focus to the trigger.
- Please search HTML labels as they appear to the user. Stripping tags leaves entities encoded, so a label displaying `R&D` doesn’t match a search for `R&D`.
- Please add tests for the new configuration and browser behavior, including changing available blocks, keyboard interaction, and accessibility checks in light and dark modes.
- Please define the supported search configuration directly on `Builder` rather than using `CanBeSearchable`. That concern also exposes methods such as `searchValues()` and `searchLabels()` which have no effect here.

KianAcquoy:
Thank for the feedback, Dan. I've addressed all points in my latest commit.

While moving off CanBeSearchable I noticed CheckboxList has the same issue: it uses the concern but never reads searchingMessage(), noOptionsMessage(), searchLabels() or searchValues(). Maybe we should resolve that in a follow-up PR?

danharrin:
Yeah a follow up PR would be good thanks

danharrin (changes requested):
Thanks for the updates! A few keyboard interactions still need addressing:

- Please prevent Enter in the search input from submitting the surrounding form.
- With a nonzero `searchDebounce()`, pressing Escape immediately after typing can close the picker instead of clearing the input, because the handler checks the debounced state.
- When the search is empty, Escape should close only the picker and return focus to its trigger. It currently also reaches an enclosing action modal’s Escape handler.

Please also make clearing/focusing on reopen wait until the panel is actually visible, rather than relying on a single timeout. The Escape handler similarly assumes it registers before the dropdown’s listener, which isn’t guaranteed while the Builder asset is loading.

Could you add focused browser tests for these cases and document `searchDebounce()`?

danharrin (approved):

PR #20472, opened 2026-09-08
fix: scope table session state to the table configuration of a `TableSelect`

### LINKED ISSUE #20468: Two TableSelect pickers share one session state key, breaking each other's filters
### Package

filament/tables

### Package Version

v5.8.1

### Laravel Version

13.30.1

### Livewire Version

v4.4.4

### PHP Version

8.4.23

### Problem description

Table session state (filters, columns, sort, search) is keyed by the Livewire component class:

// Filament\Tables\Concerns\HasFilters::getTableFiltersSessionKey()
$namespace = $this::class;
$table = md5($namespace);

return "tables.{$table}_filters";

TableSelect mounts every picker through the same component, TableSelectLivewireComponent, so every TableSelect in the app — regardless of its tableConfiguration() — reads and writes the same session keys. State from one picker is applied to another that has a completely different table.

For columns this degrades silently: unknown columns are dropped. For the query builder it throws, because a stored rule has no matching block in the other table:

LogicException: No query builder block found for [email]
vendor/filament/tables/src/Filters/QueryBuilder.php:176

The page then returns HTTP 500 on every visit until the session is cleared, since the state is restored from the session on each request.

### Expected behavior

Each TableSelect should keep its own table state. Applying a filter in one picker should not affect a picker that uses a different tableConfiguration(), and a stored filter state should never take the page down with a 500.

### Steps to reproduce

1. Create two models with different columns: Author (name) and Publisher (email).
2. Create a resource for each.
3. In each resource's table, add a query builder filter with a single constraint on its own column, and call ->persistFiltersInSession().
4. In AuthorForm, add TableSelect::make('picked')->tableConfiguration(PublishersTable::class).
5. In PublisherForm, add TableSelect::make('picked')->tableConfiguration(AuthorsTable::class).
6. Open Authors → New author, open the picker, apply the Email constraint, close the modal.
7. Open Publishers → New publisher and open the picker.

The second picker throws LogicException: No query builder block found for [email], HTTP 500.

https://github.com/albyzip/filament-table-select

### Relevant log output

```shell

```

lazerg:
## Description

Every `TableSelect` picker mounts through the same `TableSelectLivewireComponent`, and the table session keys are built from the Livewire component class. Two pickers with different `tableConfiguration()` classes therefore share one set of persisted filters, columns, sort and search. Applying a filter in one picker changes the other, and a stored query builder rule that has no matching block in the other table throws `LogicException: No query builder block found for [...]`, which returns a 500 on every visit until the session is cleared.

The session keys now come from a `getTableSessionKeyNamespace()` method, which `TableSelectLivewireComponent` overrides to include its table configuration class.

Only `TableSelect` keys change. For every other component the namespace is still the Livewire component class, so existing persisted state keeps working.

Making `QueryBuilder` tolerant of stale persisted rules in general is left out of scope here, since that is a separate behavioural decision rather than part of this bug.

Fixes #20468

## Functional changes

### DIFF
--- a/packages/forms/src/Components/TableSelect/Livewire/TableSelectLivewireComponent.php
+++ b/packages/forms/src/Components/TableSelect/Livewire/TableSelectLivewireComponent.php
@@ -139,6 +139,11 @@ class TableSelectLivewireComponent extends Component implements HasActions, HasF
         return $table;
     }
 
+    public function getTableSessionKeyNamespace(): string
+    {
+        return $this::class . '|' . base64_decode($this->tableConfiguration);
+    }
+
     /**
      * @return array<mixed>
      */
--- a/packages/tables/src/Concerns/CanGroupRecords.php
+++ b/packages/tables/src/Concerns/CanGroupRecords.php
@@ -83,7 +83,7 @@ trait CanGroupRecords
 
     public function getTableGroupingSessionKey(): string
     {
-        $table = md5($this::class);
+        $table = md5($this->getTableSessionKeyNamespace());
 
         return "tables.{$table}_grouping";
     }
--- a/packages/tables/src/Concerns/CanPaginateRecords.php
+++ b/packages/tables/src/Concerns/CanPaginateRecords.php
@@ -93,7 +93,7 @@ trait CanPaginateRecords
 
     public function getTablePerPageSessionKey(): string
     {
-        $table = md5($this::class);
+        $table = md5($this->getTableSessionKeyNamespace());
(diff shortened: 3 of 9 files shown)


danharrin:
I don't think this is the right solution, I think we should just completely disable all session features on tables that are used with the table select.

lazerg:
> I don't think this is the right solution, I think we should just completely disable all session features on tables that are used with the table select.

yes, agreed...

PR #20473, opened 2026-09-09
fix: validate Livewire temp uploads from sidecar metadata

uvera:
## Description

`FileUpload` validated new uploads with Laravel's `mimetypes`, `max`, and `min` rules. Those call `TemporaryUploadedFile::getMimeType()` and `getSize()`, which in production read Flysystem metadata instead of the JSON sidecar Livewire already writes next to the temp file.

That causes two failures:

- Real images are rejected against `image()` / `image/*` when the disk reports a generic Content-Type such as `application/octet-stream`.
- A missing `file_size` throws `UnableToRetrieveMetadata` (HTTP 500) instead of a validation error.

This change reads `name` / `type` / `size` from the sidecar first, then the original filename and `mimeTypeMap()`, and only then Flysystem inside a try/catch. Missing temp files fail validation. PHP-family extensions (`.php`, `.phtml`, `.phar`, and similar) are still blocked, matching Laravel's `mimetypes` rule.

The same metadata helper is used for rich editor file attachments. There is no API change.

## Visual changes

None. Validation behavior only; FilePond UI is unchanged.

## Functional changes

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/forms/resources/lang/en/validation.php
+++ b/packages/forms/resources/lang/en/validation.php
@@ -9,4 +9,6 @@ return [
 
     'tampered_file_path' => 'The :attribute field contains a file path that is not permitted.',
 
+    'temporary_upload_missing' => 'The :attribute is no longer available. Please upload the file again.',
+
 ];
--- a/packages/forms/src/Components/BaseFileUpload.php
+++ b/packages/forms/src/Components/BaseFileUpload.php
@@ -3,6 +3,7 @@
 namespace Filament\Forms\Components;
 
 use Closure;
+use Filament\Forms\Support\TemporaryUploadedFileMetadata;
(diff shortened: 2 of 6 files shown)


danharrin:
Thanks for the PR, but I’m going to close this. Falling back to the filename means a file named `avatar.jpeg` can pass image validation even when its contents aren’t an image—the new test actually checks that plain text is accepted this way. Current Livewire 4.x already detects MIME types from the file contents, so it’s not clear that the reported issue needs a fix here. The missing-size exception may still be a real issue, but please open an issue with a minimal reproduction with the Livewire version and storage configuration so we can investigate that separately without replacing Laravel’s file validation.

PR #20479, opened 2026-09-10
fix: Resolve nested resource index URLs to parent relation pages

OccTherapist:
## Description

When editing a nested resource record, the breadcrumb for the nested resource (e.g. “Lessons”) uses `Resource::getIndexUrl()`. Today that only looks for a **parent page keyed exactly like the relationship name**. If the relation page is registered as `manageLessons` (or another custom key), lookup fails and Filament falls back to the parent **edit/view** page.

This change resolves the parent relation page in this order:

1. Explicit `ParentResourceRegistration::page('…')`
2. Parent page keyed like the relationship name (existing behavior)
3. First parent `ManageRelatedRecords` page whose relationship matches
4. Existing view → edit → index fallback

`getRouteName()` is unchanged (still used for nested child route prefixes/names).

Docs updated for nesting + relation pages. Same code path exists on `5.x` / `6.x` and should be forward-ported.

## Visual changes

None.

## Related issues

N/A — discovered while using nested resources with `ManageRelatedRecords` pages registered under custom `getPages()` keys.

## Checklist

- [x] Code style uses Laravel Pint
- [x] Tests added for relationship key, discovery, explicit `page()`, and edit/view fallback
- [x] Documentation updated

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/03-resources/07-managing-relationships.md
+++ b/docs/03-resources/07-managing-relationships.md
@@ -1143,6 +1143,10 @@ public static function getPages(): array
 }
 ```
 
+<Aside variant="tip">
+    Use the relationship name as the page key (for example `addresses`). Nested resources use that key when building breadcrumbs and other “index” links back to the relation page. See [nested resource breadcrumbs and index URLs](nesting#nested-resource-breadcrumbs-and-index-urls).
+</Aside>
+
 <Aside variant="warning">
     When using a relation page, you do not need to generate a relation manager with `make:filament-relation-manager`, and you do not need to register it in the `getRelations()` method of the resource.
 </Aside>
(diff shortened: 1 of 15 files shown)


danharrin (changes requested):
Please remove the automatic page scanning and only support an explicit `page()` override, preserving the existing relationship-key fallback.

OccTherapist:
Thanks — updated. Automatic \`ManageRelatedRecords\` page scanning is removed. Resolution is now only:

1. Explicit \`ParentResourceRegistration::page()\`
2. Relationship-key fallback (unchanged)

Docs and tests adjusted accordingly.

danharrin (approved):

PR #20482, opened 2026-09-10
feat(actions): add `enum()` to `ImportColumn`

MadBox-99:
## Description

Mapping an import column onto an enum currently takes two manual steps, and getting either one wrong fails at runtime.

**1. The example CSV.** To show the user which values are accepted, the cases have to be listed by hand. Passing the cases themselves reads naturally:

```php
ImportColumn::make('status')
    ->examples(Status::cases())
```

…but it breaks the `downloadExample` modal action. `getExamples()` is handed straight to the `league/csv` writer, which casts every value to a string, so an enum instance throws `Object of class App\Enums\Status could not be converted to string` and the download responds with a 500. Nothing in the signature of `examples(mixed $examples)` hints at that.

**2. Validation.** Without an explicit `Rule::enum()`, an unknown value in the CSV reaches the model's cast and throws a `ValueError: "..." is not a valid backing value for enum`, rather than being reported as a failed row the user can read and correct.

This PR adds `enum()`, which covers both:

```php
use App\Enums\Status;
use Filament\Actions\Imports\ImportColumn;

ImportColumn::make('status')
    ->enum(Status::class)
```

- The cases become the example data for the column, unless `examples()` is set, in which case that still wins.
- The state is validated against the enum.
- Pure enums are supported, using their case names.
- A `multiple()` column validates each item, through the nested recursive rules the importer already applies to array columns.

It reuses the existing `HasEnum` concern, so the method has the same name and signature as the `enum()` that `SelectColumn` already exposes.

Nothing changes for a column that does not call `enum()`.

## Visual changes

None — this affects the contents of the example CSV and the validation rules of an import column, not the UI.

## Functional changes

### DIFF
--- a/packages/actions/docs/11-import.md
+++ b/packages/actions/docs/11-import.md
@@ -165,6 +165,28 @@ Any rows that do not pass validation will not be imported. Instead, they will be
 
 <UtilityInjection set="importColumns" version="4.x">As well as allowing a static value, the `rules()` method also accepts a function to dynamically calculate it. You can inject various utilities into the function as parameters.</UtilityInjection>
 
+### Validating a column against an enum
+
+If a column maps onto a [backed enum](https://www.php.net/manual/en/language.enumerations.backed.php), you can pass the enum class to the `enum()` method. Any row whose value is not a case of the enum will fail validation, instead of reaching the model's cast and throwing a `ValueError`:
+
+```php
+use App\Enums\Status;
+use Filament\Actions\Imports\ImportColumn;
+
+ImportColumn::make('status')
+    ->enum(Status::class)
+```
+
+The cases of the enum are also used as the [example CSV data](#providing-example-csv-data) for the column, so the user can see which values are accepted without you listing them by hand. Passing `examples()` yourself overrides this.
+
+Pure enums are supported too, in which case the case names are validated and used as the example data.
+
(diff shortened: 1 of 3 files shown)


danharrin (dismissed):

MadBox-99:
Sorry for the stale review — I pushed one more fix after you approved.

**The bug:** an optional `enum()` column failed every row whose cell for it was empty. `castStateItem()` casts a blank cell to `null`, and `Importer::__invoke()` casts the data before it validates, so `Rule::enum()` received `null` and rejected it. A two-row import with one blank cell imported 1 row and wrote the other to the failed-rows CSV with "The selected status is invalid."; removing `->enum()` from the same column imported both.

**The fix:** mark the column `nullable` alongside the enum rule — the same pairing `CanBeValidated::getValidationRules()` already uses in the forms package, where `getRequiredValidationRule()` returns `nullable` for a field that is not required. A `required` rule still fails a blank cell, because `required` is implicit and `nullable` does not suppress it.

**Tests:** `it('does not validate a blank state against the enum')` covers the single and `multiple()` paths against the real `castState('')` output, and `it('imports a row whose optional enum column is a blank CSV cell')` runs the `ImportCsv` job end to end. Both fail without the fix.

PR #20483, opened 2026-09-10
fix: make rich editor content area fully clickable

KianAcquoy:
## Description

Since #20258 the rich editor's editable area does not stretch to fill the configured min height, so an empty or short editor only focuses when you click the first line. This makes the contenteditable element fill the full content area so clicking anywhere in the field focuses the editor. The `10rem` default min height from #20258 is retained as requested.

## Visual changes

No visual changes

## Functional changes


danharrin (changes requested):
Please revert and retain the new default for the min height of editors, this was a deliberate change for consistency with the markdown editor.

This PR can still fix the bug with not being able to click past the first line.

KianAcquoy:
Wasn't sure if it was a deliberate change, I should ask about it in the future haha. Reverted the change, now only the fix is present.

danharrin:
Yeah usually I wouldn't make "breaking changes" but this was more of a consistency bug with the editor compared to the expectations of the framework.

danharrin (approved):

PR #20485, opened 2026-09-10
fix: stop the collapsed sidebar from shrinking below its icons

### LINKED ISSUE #20480: Collapsed sidebar icons become clipped after configured width starts being respected
### Package

filament/filament

### Package Version

5.8.1

### Laravel Version

13.0.0

### Livewire Version

4.0.0

### PHP Version

8.4

### Problem description

After updating from Filament 5.7.8 to 5.8.1, our collapsed desktop sidebar started clipping navigation icons.

Our panel has this configuration:

return $panel
    ->sidebarWidth('15rem')
    ->collapsedSidebarWidth('2rem')
    ->sidebarCollapsibleOnDesktop();

This configuration predates the update. Previously, the collapsed sidebar displayed the full icons.

The collapsed sidebar becomes too narrow for its navigation icons and padding. The right side of each icon is clipped.

Image

Image

Commit - https://github.com/filamentphp/filament/commit/87f375722299a20530eead64dda73e8121dac412
 began applying --collapsed-sidebar-width to the collapsed desktop sidebar which in my opinion causes the issue. 

### Expected behavior

Navigation icons should remain fully visible when the sidebar is collapsed. If the configured width cannot accommodate the icons and padding, a minimum supported width or documentation explaining that constraint would help.

### Steps to reproduce

Steps to reproduce
1. Configure a panel with the settings above.
2. Add navigation items with icons.
3. Open the panel at a desktop viewport width.
4. Collapse the sidebar.

### Relevant log output

```shell

```

lazerg:
## Description

Since #20436 made the collapsed desktop sidebar honour `--collapsed-sidebar-width`, a configured width that is narrower than a navigation icon clips the icons. The reporter's panel uses `->collapsedSidebarWidth('2rem')`, which worked before because the width was driven by the content; now the sidebar is 32px wide while the nav's leading padding already pushes the first icon to 24px, so the right half of every icon is cut off.

This adds `min-w-max` alongside the configured width, so the collapsed sidebar can never be narrower than the icons and padding it has to display. Wide enough values are unaffected: with the default `4.5rem` the sidebar stays 72px and nothing moves, and `2rem` now falls back to the content width of 63px instead of clipping.

Fixes #20480.

## Functional changes


danharrin:
https://github.com/filamentphp/filament/issues/20480#issuecomment-5645144113

PR #20486, opened 2026-09-11
Disable session persistence for tables inside a TableSelect

### LINKED ISSUE #20468: Two TableSelect pickers share one session state key, breaking each other's filters

pxperr:
## Description

Every `TableSelect` (and `ModalTableSelect`, which wraps one) renders through
the same `TableSelectLivewireComponent`, and the table session keys are built
from the Livewire component class. So two pickers with different table
configurations end up sharing the same persisted filters, search, sort,
grouping and column state. Apply a filter in one and it shows up in the other,
and if a query builder rule gets persisted from one picker the other one throws
`LogicException: No query builder block found for [...]` and 500s until the
session is cleared.

Following the direction in #20472, this turns off session persistence on the
table used by a `TableSelect` rather than trying to scope the keys. There's a
new `persistInSession()` method on `Table` that toggles every persistence
option in one go (filters, search, column searches, sort, grouping and
columns), and the component calls `persistInSession(false)` once the
configuration class has run. Having it on the table means there's one place
to update when a new session feature comes along, rather than another line
to remember in the component.

I've also made `loadTableColumnsFromSession()` return the default column state
when `persistsColumnsInSession()` is false. Column persistence is on by
default, so every picker has already been writing to the shared columns key.
The write was guarded by the flag but the read wasn't, so without this pickers
would keep picking up that stale state after upgrading until the session
expired.

Records per page is still stored in the session as there's no table-level
switch for it. It's validated against the pagination options when read, so at
worst it falls back to the default.

Tests cover `persistInSession()` turning everything on and off, and the
`TableSelect` component having all the options off, writing nothing to the
session, not inheriting another picker's state, and ignoring stale column
state already in the session. The tables suite and the TableSelect,
ModalTableSelect, AttachAction and RelationManager tests all pass (the browser
tests in `ColumnTest` and `QueryBuilderTest` didn't run locally as I don't have
Playwright installed). Pint, Rector and PHPStan are clean.

`persistInSession()` is documented in the tables overview, and the modal table
select docs now mention that persistence is always off inside a picker.

Fixes #20468

## Visual changes

None.

## Functional changes


packages/forms/src/Components/TableSelect/Livewire/TableSelectLivewireComponent.php
```diff
@@ -87,6 +87,12 @@ public function table(Table $table): Table
             ->maxSelectableRecords(is_array($this->state) ? $this->maxSelectableRecords : 1)
             ->deselectAllRecordsWhenFiltered(false)
             ->disabledSelection($this->isDisabled)
+            ->persistFiltersInSession(false)
```
danharrin: Please create some method that toggles all of these at once (`persistInSession(false)`) and is owned by the Table object so that it's less likely we will forget to turn off new session features here in the future.

danharrin (changes requested):

packages/forms/src/Components/TableSelect/Livewire/TableSelectLivewireComponent.php
pxperr: Done - added `persistInSession()` on `Table` that toggles all of them, and the component now calls that instead.

danharrin (approved):

PR #20495, opened 2026-09-11
Support Widget::configureUsing() for global polling config

jeffersongoncalves:
## Summary
- Add `Configurable` (and `EvaluatesClosures`) to `Widget`, so `Widget::configureUsing()` — and subclass-scoped variants like `ChartWidget::configureUsing()` / `StatsOverviewWidget::configureUsing()` — can globally disable or change polling across every widget.
- `CanPoll` now exposes a fluent `poll()` setter with `Closure` support, matching the existing `CanPoll` trait in the `schemas` and `tables` packages.
- `ComponentManager::configure()` now accepts `object` instead of the concrete `Filament\Support\Components\Component` class, since `Widget` extends `Livewire\Component` rather than that base class.
- Documented the new global config option in `packages/widgets/docs/02-stats-overview.md` and `packages/widgets/docs/03-charts.md`.

## Test plan
- [x] `vendor/bin/pest tests/src/Widgets` (94 passed, excluding pre-existing browser accessibility test infra failures)
- [x] `vendor/bin/pest tests/src/Support` (727 passed)
- [x] `vendor/bin/phpstan analyse` (no errors)
- [x] `vendor/bin/pint --dirty` (clean)

### DIFF
--- a/packages/support/src/Components/ComponentManager.php
+++ b/packages/support/src/Components/ComponentManager.php
@@ -98,7 +98,7 @@ class ComponentManager implements ScopedComponentManager
         }
     }
 
-    public function configure(Component $component, Closure $setUp): void
+    public function configure(object $component, Closure $setUp): void
     {
         $componentClass = $component::class;
 
--- a/packages/support/src/Components/Contracts/ScopedComponentManager.php
+++ b/packages/support/src/Components/Contracts/ScopedComponentManager.php
@@ -9,7 +9,7 @@ interface ScopedComponentManager
 {
     public function configureUsing(string $component, Closure $modifyUsing, ?Closure $during = null, bool $isImportant = false): mixed;
 
-    public function configure(Component $component, Closure $setUp): void;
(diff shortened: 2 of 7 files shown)


danharrin:
Thanks for working on this. While the `configureUsing()` API looks consistent with our other components, after testing the behavior more thoroughly, I don’t think global widget polling configuration fits naturally with the existing API.

Widgets already support overriding the protected `$pollingInterval` property and `getPollingInterval()` method. A global configuration callback can unexpectedly override those existing customizations. Preserving their precedence across parent and child widget classes would require a fairly complicated set of inheritance and configuration-ordering rules, which would be difficult to explain and maintain relative to the benefit of the feature.

Supporting `Closure` values on the property would also be incompatible with existing widgets that redeclare the documented `?string` property.

I’d rather keep polling configuration explicit on each widget than introduce these hidden precedence rules, so I’m going to close this.

PR #20502, opened 2026-09-14
Feature: nested action parent data

webard:
## Description

This adds the smaller, independently useful capability suggested in #20354: letting a nested action safely validate and fill the data of the action it was mounted from.

Today a nested action that consumes its parent's data only has `getRawData()`, which -  as the documentation states -  returns whatever the browser last sent and has been validated by nothing. There is no supported way to validate it: `getMountedActionSchema()` is `protected`, and calling `getState()` with its default arguments runs the hooks that persist file uploads, so a read would have the side effects of a submission.

### `getValidatedData()`

Validates an action's schema and returns the validated data, without running the action:

```php
Action::make('first')
    ->schema([
        TextInput::make('foo')->required(),
    ])
    ->action(function () {
        // ...
    })
    ->extraModalFooterActions([
        Action::make('second')
            ->action(function (array $mountedActions) {
                $data = $mountedActions[0]->getValidatedData();

                // ...
            }),
    ])
```

A `ValidationException` is thrown when the schema is invalid. The hooks that run before dehydration are skipped, the same way `Repeater::getItemState()` skips them, so reading an action's data never has the side effects of submitting it - a file upload comes back as it was sent rather than as the path it is stored at. The documentation states this, along with `mutateDataUsing()` and the form-validation hooks not running.

### `fillParentActionData()`

Writes into the schema data of the action a nested action was mounted from. It delegates to `Schema::fillPartially()`, so the data is hydrated as it is for any other partial fill, and the parent validates it with its own rules when submitted:

```php
Action::make('createInvoice')
    ->schema([
        TextInput::make('title')->required(),
        TextInput::make('reference')->required(),
    ])
    ->action(function (array $data) {
        // `$data['reference']` is filled in.
    })
    ->extraModalFooterActions([
        Action::make('generateReference')
            ->schema([
                TextInput::make('prefix')->required(),
            ])
            ->action(function (Action $action, array $data) {
                $action->fillParentActionData([
                    'reference' => "{$data['prefix']}-123",
                ]);
            }),
    ])
```

The parent is taken from the mounted stack rather than `getParentAction()`, which is only set for actions registered on a modal and not for actions registered on a component inside one. A `LogicException` is thrown when the action was not mounted from a parent, rather than failing silently.

Both delegate to `getValidatedMountedActionData()` and `fillMountedActionData()` on the Livewire component, which are public so they can be used directly.

### `getParentActionValidatedData()`

A nested action that only needs the action it was mounted from can read it directly, instead of reaching into `$mountedActions`:

```php
Action::make('second')
    ->action(function (Action $action) {
        $data = $action->getParentActionValidatedData();

        // ...
    })
```

Like `fillParentActionData()`, the parent is taken from the mounted stack, so this works for an action registered on a component inside the modal as well as one registered on the modal itself. A `LogicException` is thrown when the action was not mounted from another action.

## Functional changes


danharrin:
@webard thoughts on the changes now? Does everything still work for your use cases?

danharrin (approved):

webard:
Yes, this still covers my needs. Thank you for taking the time to refine it - this will be really useful.

PR #20503, opened 2026-09-14
fix: Dropdown `aria-controls` loop when the component is initialised twice on the same element

kman86:
## Description

If `filamentDropdown` is initialised a second time on an element that still has a live instance, the tab freezes. The component has no `destroy()`, so the first instance's `MutationObserver` on the trigger survives, and the panel `id` is generated per instance, so the second instance picks a different one. Each observer then sees the other's `aria-controls` write as a change and puts its own value back. `setAttributeIfChanged()` cannot stop this, because from each instance's point of view the value did change. We hit it after a Livewire morph of a list with dropdowns, where Alpine occasionally runs `destroyTree` + `initTree` on an item that is still there: 399,602 synchronous `setAttribute` calls in 27 seconds on a page with 25 dropdowns.

The fix keeps the panel `id` and the observer on the element instead of on the instance, disconnects a previous observer before creating a new one, and adds a `destroy()` that releases the observer and the `livewire:navigate` listener. Alpine runs `destroy()` synchronously inside `destroyTree`, before `initTree` creates the new instance, so a teardown never touches its replacement.

Reproduce on any page with a Filament dropdown, in the browser console:

```js
const root = document.querySelector('[x-data="filamentDropdown"]')
const trigger = root.querySelector('.fi-dropdown-trigger button')
const panel = Alpine.$data(root).$refs.panel

Alpine.destroyTree(root)
panel.removeAttribute('id')
trigger.removeAttribute('aria-controls')
Alpine.initTree(root)
trigger.setAttribute('aria-controls', 'tampered')
```

Before: the tab freezes. After: `aria-controls` is restored to the original panel `id`. `dropdown.js` is identical on `4.x` and `5.x` at the time of writing.

## Visual changes

None.

## Functional changes

- [x] Code style has been fixed by running the `composer cs` command. (No PHP touched.)
- [x] Changes have been tested to not break existing functionality. (Browser test with the patched component: after a forced second `initTree` the trigger and panel settle in under 20 attribute writes, against 5,000+ before. Opening and closing checked manually.)
- [x] Documentation is up-to-date. (No documented behaviour changes.)


danharrin (changes requested):
Thanks for the fix. Could you add a browser test that reproduces this through a real user interaction, without calling JS like `Alpine.destroyTree()` or `Alpine.initTree()` directly? Without a real-world reproduction, we can’t be sure this can occur in a Filament panel or verify that the underlying behavior won’t regress.

kman86:
Done in 4d43f91: a browser test that reproduces the loop through user interaction only. `tests/src/Support/DropdownTest.php` visits a fixture page (`DropdownBrowserTest`) with a keyed list of 25 rows, each holding an `<x-filament::dropdown>` and an `x-text` binding, and clicks a "prepend row" button twice. No `Alpine.destroyTree()` or `initTree()` anywhere; the morph does that itself.

Result on the previous `dropdown.js`: `{"reinitialisedRows":23,"writes":5003,"limitHit":true}`, red in 3 of 3 runs (all retry attempts). With the fix: `{"reinitialisedRows":23,"writes":100,"limitHit":false}`, green in 3 of 3. The test asserts that existing rows really were initialised a second time, so it cannot pass vacuously, and it waits for each morph's DOM result instead of sleeping. The counter swallows `setAttribute` above 5000 writes so a looping page can still report rather than freeze the renderer.

Why a plain morph initialises the component twice, since Alpine's observer normally dedupes a node that is removed and re-inserted within one batch:

1. Prepending to a keyed list makes the keyed diff move every retained row (`remove()` plus `replaceWith()` down the list): removed and re-inserted, not recreated.
2. `Alpine.morph`'s `patch()` calls `Alpine.cloneNode(from, to)` for every row, which initialises the incoming node's directives in clone mode. An `x-text` (or `x-show`) effect runs synchronously there and calls `mutateDom`, which does `observer.takeRecords()` mid-morph.
3. The removal and re-insertion of a row therefore land in different mutation batches. Alpine processes "removed" first (`destroyTree`, marker gone) and then "added" (no marker, so `initTree`) on the same element. With one `x-text` per row, 23 of the 24 retained rows are re-initialised per click.
4. The morph keys elements by `el.id` when there is no `wire:key`, so the panel (which carries the client-generated `id`) is replaced while the trigger `<button>` survives. The old instance never disconnected its observer, so it keeps watching the live trigger. Depending on batch order the new instance either adopts the old `id` or generates its own; in the latter case the next morph strips `aria-controls` and the two observers ping-pong.

That explains why it was intermittent in the production app where we hit it (5 of 8 saves froze the tab, 400k `setAttribute` calls in 27 seconds) and why it only surfaced on a timeline whose rows have a collapsible "show more" text. The fixture keeps one `x-text` per row for the same reason.

The rebuilt `packages/support/dist/index.js` is included so the test runs against the fixed component. Happy to trim the explanation in the test comments or move the fixture if you prefer it leaner.

people-sea:
This seems related to [alpinejs/alpine#4868](https://github.com/alpinejs/alpine/pull/4868). Since the reproduced morph already calls destroy() before reinitializing the dropdown, I think the Filament-side fix only needs to clean up the observer and navigation listener there, without storing state on the DOM node.


https://github.com/user-attachments/assets/85d11c66-93ca-48e8-93ac-e898e000096f

danharrin (approved):

PR #20510, opened 2026-09-15
Fix duplicate banner landmark in panel sidebar

Orrison:
## Description

Panel pages render two native `<header>` elements, which produces two implicit ARIA `banner` landmarks on the same page:

1. `packages/panels/resources/views/livewire/sidebar.blade.php` renders the sidebar's header as `<header class="fi-sidebar-header">`.
2. `packages/panels/resources/views/components/header/index.blade.php` (the `<x-filament-panels::header>` component used for the page title, breadcrumbs, and actions) renders `<header {{ $attributes->class(['fi-header', ...]) }}>`.

Neither is nested inside `<article>`, `<aside>`, `<main>`, `<nav>`, or `<section>`, so both get the implicit `banner` role. Assistive technology expects exactly one `banner` landmark per page so users can reliably skip to or from the site header. Having two competing `banner` landmarks fails WCAG 2.4.1 Bypass Blocks (flagged by axe-core as `landmark-no-duplicate-banner` / `landmark-unique`).

I checked the topbar (`packages/panels/resources/views/livewire/topbar.blade.php`) as well, it already uses `<nav aria-label="...">`, not `<header>`, so there is no third landmark to fix there.

### Fix

Changed the sidebar header's tag from `<header>` to `<div>`. It is not the page's primary banner landmark, it is a container for the logo and sidebar collapse controls. The `fi-sidebar-header` class and all other attributes are unchanged, and nothing in the CSS or JS keys off the tag name (only the class), so this does not affect styling or behavior.

The page header (`fi-header`) is untouched and remains the single `banner` landmark.

### Before / after landmark structure

Before: `<header class="fi-sidebar-header">` (sidebar) and `<header class="fi-header">` (page) both expose a `banner` landmark, two banner landmarks per page.

After: `<div class="fi-sidebar-header">` (sidebar, no landmark role) and `<header class="fi-header">` (page), one banner landmark per page.

### Alternative approach

If keeping the sidebar element as a `<header>` tag is preferred (for example to minimize the diff or avoid any risk to third-party code relying on the tag), the same fix can be achieved by keeping `<header>` and adding `role="presentation"` (or `role="none"`) to it instead. An explicit role always overrides the implicit landmark role, so this would also leave only the page header as the `banner` landmark. Happy to switch to that approach if maintainers prefer it.

## Visual changes

No visual changes. This only changes an element's tag name and does not affect layout, styling, or behavior.

## Functional changes

Verified with the existing browser test in `tests/src/Panels/Auth/LoginTest.php` (`can fill the login form, authenticate, and redirect to the dashboard in the browser`), which visits the authenticated dashboard in both light and dark mode and calls `assertNoAccessibilityIssues()`. This test passes with the fix applied and exercises the full sidebar and page header together.


danharrin (approved):

danharrin (changes requested):
The change does not cause any problems necessarily. However I think the premise may be incorrect: a `<header>` only exposes the `banner` landmark when scoped to the `<body>`. When it is inside `<main>` or sectioning content, it does not expose another banner ([HTML-AAM](https://www.w3.org/TR/html-aam-1.0/#el-header-ancestorbody), [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/header#accessibility)).

Filament renders the page header inside [`<main id="fi-main-content">`](https://github.com/filamentphp/filament/blob/4.x/packages/panels/resources/views/components/layout/index.blade.php#L109-L122). Therefore, `fi-header` should not be a `banner`. The sidebar header is outside `<main>`, making it the existing page-level banner. Changing that element to `<div>` appears to leave the page with no banner rather than resolving a duplicate.

The other native headers appear correctly scoped too: `fi-simple-header` is inside the simple layout’s `<main>`, while both section-header implementations are inside `<section>`.

Is there a chance that the a11y tool is wrong here?

Orrison:
Looking into whether or not this is a false positive on the tools end. Going to close this for now till I find out more details as your findings seem to be right to me.

PR #20516, opened 2026-09-16
fix: query builder filter indicators when a rule is deleted from the deferred filters form

### LINKED ISSUE #20513: QueryBuilder indicators throw "No query builder block found" when an applied rule is removed in the deferred filters form
### Package

filament/tables

### Package Version

v5.8.2 (also v5.8.1)

### Laravel Version

v13.32.0

### Livewire Version

v4.4.5

### PHP Version

PHP 8.4.23

### Problem description

With deferred filters (the default), a `QueryBuilder` filter throws a `LogicException` during a full render when a rule is applied but has since been deleted from the filters form without applying again:

```
No query builder block found for [price]. (View: vendor/filament/tables/resources/views/index.blade.php)
at vendor/filament/tables/src/Filters/QueryBuilder.php:100
```

`QueryBuilder::getRuleSummaries()` (called from `indicateUsing()`) receives the rules from the *applied* state (`tableFilters`, via `getTableFilterState()`), but looks each rule up with `$ruleBuilder->getChildSchema($ruleIndex)`. The `RuleBuilder` lives in the filters form, whose state path is `tableDeferredFilters`, so its child schemas are built from the *deferred* state. When the two differ structurally (rule deleted in the form, not applied yet), the lookup returns `null` and the exception is thrown.

`applyFiltersToTableQuery()` already handles this mismatch by temporarily switching the form state path to `tableFilters`; building the indicators does not.

It only surfaces on a full render. The builder's delete action renders partially, so clicking delete on its own is fine. But when the same request also carries a pending deferred `wire:model` update (e.g. the user changed another field in the filters form first), `PartialsComponentHook` falls back to a full render, the table view calls `getFilterIndicators()`, and the request crashes. In production the logged Livewire method was `callMountedAction`.

### Expected behavior

The table keeps rendering. The indicators should be built from the applied state with a rule builder that matches that state (or rules without a matching block should be skipped) instead of throwing.

### Steps to reproduce

1. A resource table with deferred filters, a `SelectFilter` and a `QueryBuilder`:

```php
->filters([
    SelectFilter::make('status')
        ->options([
            'active' => 'Active',
            'archived' => 'Archived',
        ]),
    QueryBuilder::make('extra')
        ->constraints([
            NumberConstraint::make('price'),
        ]),
])
```

2. Open the filters, add a rule "Price is minimum 10" and apply.
3. Open the filters again, change the Status select (don't apply), then click delete on the price rule.
4. The request fails with `No query builder block found for [price]`.

The reproduction repository contains a failing test for this (`tests/Feature/QueryBuilderIndicatorsTest.php`):

```php
$rule = ['type' => 'price', 'data' => ['operator' => 'isMin', 'settings' => ['number' => 10]]];

$component = Livewire::test(ListProducts::class)
    ->set('tableDeferredFilters.extra.rules', ['rule-uuid' => $rule])
    ->call('applyTableFilters')
    ->mountAction(TestAction::make('delete')
        ->schemaComponent('extra.rules', schema: 'tableFiltersForm')
        ->arguments(['item' => 'rule-uuid']));

// A pending deferred update in the same request forces a full render, as in the browser
$component->update(
    calls: [['method' => 'callMountedAction', 'params' => [], 'path' => '']],
    updates: ['tableDeferredFilters.status.value' => 'archived'],
);
```

Workaround we use for now, a subclass that skips rules the builder doesn't know:

```php
public function getRuleSummaries(array $rules, RuleBuilder $ruleBuilder, int $iteration = 1): array
{
    $knownRules = array_filter(
        $rules,
        fn (mixed $rule, int|string $ruleIndex): bool => $ruleBuilder->getChildSchema($ruleIndex) !== null,
        ARRAY_FILTER_USE_BOTH,
    );

    return parent::getRuleSummaries($knownRules, $ruleBuilder, $iteration);
}
```

https://github.com/MaartenJC/filament-querybuilder-repro

Run `php artisan test --filter=QueryBuilderIndicatorsTest`.

### Relevant log output

```
Illuminate\View\ViewException: No query builder block found for [price]. (View: vendor/filament/tables/resources/views/index.blade.php) (View: vendor/filament/tables/resources/views/index.blade.php)
at vendor/filament/tables/src/Filters/QueryBuilder.php:100
```

lazerg:
## Description

With deferred filters, a `QueryBuilder` filter throws `No query builder block found for [x]` during a full render once a rule has been deleted in the filters form but not applied yet.

The indicator summaries are built from the applied state (`tableFilters`), but the rule builder they look each rule up in belongs to the filters form, whose state path is `tableDeferredFilters`. When the user deletes a rule without applying, the two no longer match and the lookup fails. `applyFiltersToTableQuery()` already avoids this by temporarily switching the form to the applied state path; building the indicators did not.

The summaries now resolve the rule builder from the applied state the same way, so they keep describing the filters the table is actually using.

Fixes #20513

## Functional changes


tests/src/Tables/Filters/QueryBuilderTest.php
```diff
@@ -4081,6 +4082,53 @@ function applyQueryBuilderFilter(array $rules)
...
+            ->assertCanSeeTableRecords([$latePost])
+            ->assertCanNotSeeTableRecords([$earlyPost])
+            ->assertSee('Published at is after Mon, Jul 13, 2026 15:00:00')
+            // Delete the rule from the deferred form, *without* applying it.
+            ->mountAction(TestAction::make('delete')
+                ->schemaComponent('query_builder.rules', schema: 'tableFiltersForm')
+                ->arguments(['item' => 'rule-uuid']));
+
+        // A pending deferred update in the same request forces a full render, as it does in the browser.
+        $component->update(
```
people-sea: I don't think we should construct this request manually. The issue steps are slightly off—the failure occurs after deleting the rule and then sorting Price.

lazerg: Fixed.

danharrin (approved):

PR #20520, opened 2026-09-17
Feature: full-width toggle buttons

caseydwyer:
## Description

This PR adds a new `->fullWidth()` method to ToggleButton inputs, and does what it says on the tin—allowing the contained buttons to expand, so they take up the full available width of the container. This works across all variants:

- **stacked** (default) - each button expands to 100% of the available width
- **grouped** - buttons expand/grow to collectively fill available space, _with known bug (see below)_
- **inline** - buttons expand/grow to collectively fill available space

Closes #20509.

## Visual changes

full-width-assortment

## Functional changes

- [ ] Generate screenshot for docs - deferred as separate commit (_unsure if that's needed/wanted_)

## Bugs

While working on the sample screenshot above, I got curious about how it handled lengthy labels and text-wrapping. It works fine/as expected on `inline` and `stacked` variants, but the standard, non-full-width `grouped` variant appears to break out of its parent container—unless, coincidentally, this new `->fullWidth()` method is applied, which seems to ***fix*** the issue:

broken-container

When prompted about that issue, here's what Claude said, in case it's helpful (_I have not confirmed this bit, but decided it was beyond the scope of this PR, due to the impact beyond ToggleButton component. Let me know if you'd like an attempted fix on this issue in a separate pull_):

> Confirmed the root cause. .fi-btn-group is a shared class in filament/packages/support/resources/css/components/button.css — used not just by ToggleButtons::grouped(), but also by ActionGroup, FileUpload, and the generic <x-filament::button.group> component. So this bug (and any fix) has a larger blast radius than your edits.
\
The bug, in button.css:

```css
.fi-btn-group {
    @apply grid grid-flow-col rounded-lg shadow-sm ring-1 ring-gray-950/10 dark:ring-white/20;

    & > .fi-btn {
        @apply flex-1 rounded-none;
        ...
    }
}
```

> - `grid grid-flow-col` puts every button into its own auto-generated grid column, but with no `grid-auto-columns` override, the default track sizing is `auto`—which behaves like `max-content` and never lets the track shrink below the text's unwrapped width. That's why a long label just keeps growing the column (and the whole grid) instead of wrapping.
> - The `flex-1` on `.fi-btn inside .fi-btn-group` is dead code—`flex-1` only means anything on a flex item, but the parent here is `display: grid`, not flex, so it's silently doing nothing. It looks like it was intended to make the buttons equal-width, which might never have actually worked.
> The fix: give the grid explicit flexible tracks—`grid-auto-columns: minmax(0, 1fr)` (Tailwind arbitrary value `auto-cols-[minmax(0,1fr)]`) instead of the implicit `auto`. The minmax(0, ...) lets a track shrink below its content's natural width (enabling wrap), and 1fr makes all columns equal width—which is what the (currently broken) `flex-1` was trying to achieve. Once that's in place, the `flex-1` on `.fi-btn` becomes true dead code and should be removed.


getupkid:
Nice, I only just implemented my own local solution to this problem the other day. Good to see a solution being committed to the core.

danharrin:
Also pushed a fix for the overflow issue mentioned

danharrin (dismissed):

Issue #17557, opened 2025-08-28
Navigation groups not sorted alphabetically


robinmiau:
### Package

filament/filament

### Package Version

v4.0.4

### Laravel Version

v12.3.1

### Livewire Version

v3.6.4

### PHP Version

PHP 8.3.24

### Problem description

Since Filament v4, navigation groups are no longer sorted alphabetically. Instead, it appears that they are sorted based on the first navigation item in the group.

<img width="306" height="711" alt="Image" src="https://github.com/user-attachments/assets/60706d28-fd59-4521-985d-a37a2e299db0" />

### Expected behavior

They should remain sorted alphabetically unless you explicitly rearrange them with `->navigationGroups([])` in the panel provider.

### Steps to reproduce

1. Add resources with different navigation groups.
2. The navigation groups are not sorted alphabetically by their label.

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/filamentphp/demo

### Relevant log output

```shell

```

harman-codes:
Hi, you can try this. As $navigationGroup accepts UnitEnum, create an enum. For example: NavGroupEnum.php 

<php
use Filament\Support\Contracts\HasLabel;
use Illuminate\Contracts\Support\Htmlable;

enum NavGroupEnum implements HasLabel
{
    case Attendance;
    case Leave;
    case Recruitment;

    public function getLabel(): string|Htmlable|null
    {
        return match ($this) {
            self::Attendance => 'Attendance',
            self::Leave => 'Leave Management',
            self::Recruitment => 'Recruitment',
        };
    }
}

and on the resource or page , you can use it as : 

On AnnouncementResource.php , add it as : 
protected static string | UnitEnum | null $navigationGroup = NavGroupEnum::Announcement;

Now, you can arrange cases as per your requirement and groups will be arranged automatically. 

For example, if you will arrange cases like : 
case Recruitment;    
case Leave;
case Attendance;

Groups will be arranged in the same order. 

As in your case, if you want to arrange them in Alphabetical order, arranges the cases in enum alphabetically.

robinmiau:
@harman-codes I know about enums, but it should still sort alphabetically by default with no extra configuration, as was the case in v3. Thanks anyway!

danharrin:
[...]

Issue #15498, opened 2025-02-01
[Feature Request] Add ability to control navigation groups order


oldinosor:
In Filament 3.2, there doesn't seem to be a way to control the order of navigation groups. They are currently ordered alphabetically by default.

**Current behavior:**
- Navigation groups are ordered alphabetically
- Methods like `sort()`, `sortOrder()`, and `navigationGroups()` don't affect the order
- Using invisible characters or prefixes shows in the UI

**Expected behavior:**
- Ability to explicitly set the order of navigation groups
- Similar to how we can order individual navigation items within a group using `navigationSort`

**Example of desired functionality:**
```php
Panel::configureUsing(function (Panel $panel): void {
    $panel->navigationGroups([
        NavigationGroup::make()
            ->label('Deal Pipeline')
            ->order(1),
        NavigationGroup::make()
            ->label('Settings')
            ->order(2),
        NavigationGroup::make()
            ->label('Administration')
            ->order(3),
    ]);
});

oldinosor:
This would greatly help in organizing the navigation structure in a more logical way rather than being constrained by alphabetical ordering.

Additional context: We've tried several approaches without success:

Using sort() and sortOrder()
Using invisible Unicode characters
Using the NavigationBuilder
Using prefixes (but they show in the UI)
Having a proper way to control the order would be very helpful for creating intuitive admin interfaces.

abrardev99:
Would love to send PR for this onc if no one already working on it.

saade:
It does respect the order you define in `->navigationGroups()` ? It should.

oldinosor:
@abrardev99 Thanks for your help.
@saade Yes, you are right. It works. Thanks.

The navigationGroups() functionality is documented in the Panels section, but it's not prominently featured in the documentation.
In Filament v3.2 documentation, there are several ways to configure navigation:
At the Resource level using navigationGroup()
At the Panel level using navigationGroups()
At the individual item level using navigationSort()

Having encountered this documentation gap myself, I'd like to contribute by submitting a small documentation improvement. The goal is to make the navigation group ordering feature more discoverable and to provide clear examples of how to use it effectively. This could help other developers avoid the same confusion I experienced.

danharrin:
Hi, when registering groups in a panel you can order them by placing them in a different order in `$panel->navigationGroups()`. We currently have no plans to support otherwise.

oldinosor:
@danharrin Thanks for your rewiew.

PR #20525, opened 2026-09-18
Add sort weight for unregistered navigation groups

jeffersongoncalves:
## Summary

- A navigation group created ad hoc (an item's `->group('Label')` that's never listed in `Panel::navigationGroups()`) currently has no way to control where it ends up: all unregistered groups tie on the same weight and fall back to the order their first item happened to be discovered in. This makes it impossible for a package to say "my group goes near the end" without the consuming app explicitly registering it.
- Adds `NavigationGroup::sort()` / `getSort()`, mirroring the existing `NavigationItem::sort()` / `getSort()` pattern.
- Adds `NavigationItem::groupSort()` / `getGroupSort()` and `$navigationGroupSort` / `getNavigationGroupSort()` on resources and pages, so a resource/page can declare its group's weight without the item's own `group` value needing to become an object (avoids threading a `NavigationGroup` instance through `NavigationManager`'s `serialize()`-based grouping key).
- In `NavigationManager::get()`, an unregistered group now sorts by `count($registeredGroups) + $group->getSort()` instead of always `count($registeredGroups)`. Registered groups (via `navigationGroups()`) are unaffected — their order still comes entirely from array position.
- Additive and backwards compatible: `getSort()` defaults to `0`, so groups that don't opt in keep today's exact tie value and relative order.

Related: #10130, #15498, #17557 (previous requests for group-level ordering control).

## Test plan

- [x] `vendor/bin/pest tests/src/Panels/Navigation/NavigationTest.php` — added a new test asserting an unregistered group with a high `groupSort()` sorts after one without it, even when its item's own `sort()` is lower.
- [x] `vendor/bin/pest tests/src/Panels/Navigation/NavigationBuilderTest.php tests/src/Panels/Configuration/NavigationTest.php tests/src/Panels/Configuration/ResourceConfigurationTest.php tests/src/Panels/Configuration/PageConfigurationTest.php tests/src/Panels/ComponentCachingTest.php` — no regressions (83 passed).
- [x] `vendor/bin/phpstan analyse` on the touched files — no errors.
- [x] `vendor/bin/pint --test` on the touched files — no new violations (pre-existing CRLF/style flags on these files are unrelated to this change).


danharrin:
Sorry, no plans to make this part of the system even more complex than it already is. We already have multiple ways to sort groups which interact and cause problems. It's best to just make plugin users register groups explicitly so they have total control, and that registration API is not difficult or complex enough to warrant this. Please propose features in discussions and @ me instead of creating PRs to start with.

